PR #161, opened 2024-01-18
store: added cli query commands

hackaugusto:
why: I'm adding the control plane and want an easy way to inspect the store.

### DIFF
--- a/store/src/cli/mod.rs
+++ b/store/src/cli/mod.rs
@@ -1,6 +1,10 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, str::FromStr};
 
-use clap::{Parser, Subcommand};
+use anyhow::anyhow;
+use clap::{Args, Parser, Subcommand};
+use hex::FromHex;
+use miden_crypto::{Felt, FieldElement, StarkField};
+use miden_node_proto::digest::Digest;
 use miden_node_store::config;
 
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Parser)]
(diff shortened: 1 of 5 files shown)


store/src/cli/mod.rs
```diff
@@ -16,4 +20,74 @@ pub struct Cli {
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Subcommand)]
 pub enum Command {
     Serve,
+
+    #[command(subcommand)]
+    Query(Query),
+}
+
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Subcommand)]
+pub enum Query {
```
bobbinth, line 29: Should we add some simple comments here so that we get basic hints in the CLI about what the commands do? Probably fine to limit this to top-level commands (i.e., `serve` and `query`).

store/src/main.rs
```diff
@@ -17,7 +30,114 @@ async fn main() -> Result<()> {
...
+        Command::Query(command) => query(config, command).await?,
     }
 
     Ok(())
 }
+
+async fn query(
+    config: StoreTopLevelConfig,
+    command: Query,
+) -> Result<()> {
```
bobbinth, line 42: Let's add some basic comments to this function to explain what it does at the high level (e.g., that it is not going directly to the database but rather is making a request to the stope's gRPC API).

bobbinth (approved):
Looks good! Thank you! I left a couple of comment-related comments inline - but other than that, this is good to go.

Issue #106, opened 2023-12-13
Verify transaction proofs


hackaugusto:
ref: https://discord.com/channels/893151281848406016/1184427810736058420/1184532717564608602

bobbinth:
There are two potential places to verify transaction proofs:

1. In the block producer as a part of [verify_tx()](https://github.com/0xPolygonMiden/miden-node/blob/main/block-producer/src/state_view/mod.rs#L46).
2. In the RPC as a part of [submit_prove_transaction()](https://github.com/0xPolygonMiden/miden-node/blob/main/rpc/src/server/api.rs#L82) handler. 

I think we definitely should do this in the RPC. This will require deserializing transaction and checking the proof, but it will also prevent transactions with invalid proofs getting to the block producer. This is desirable because RPC component is much easier to scale (i.e., by running on multiple instances with a load balancer).

It may also be good to re-verify transactions in the block producer - just in case (the proofs will be verified as a part of the batching process, but it may be good to do it via `verify_tx()` as well). If we do this, I think we should have a configuration option for the block producer - something like `check_tx_proofs`.

bobbinth:
Closed by #220.

PR #171, opened 2024-01-23
Add proof verification on the `RPC` and `Block Producer`

phklive:
In this PR I propose the addition of the verification of tx prooofs on the `RPC` and `Block Producer`

This will make sure that the locally proven tx's have a valid proof of execution.

closing: #106


phklive:
Looking into it, tests fail giving this error when proof is verified: https://github.com/0xPolygonMiden/miden-node/pull/171/files#diff-a6ed0c785d4590bc95c216bcf514384eee6765b1c2b732d0b0a1ad7e14d3204aR50

This error comes from the winter-verifier crate:
https://github.com/facebook/winterfell/blob/a450b818f7ec70e7d40628c789845a93d6e0c030/verifier/src/errors.rs#L54

bobbinth:
> Should we use the same name on both declarations?

We should try to have different names - but don't have great ideas at the moment. Maybe the one in the block producer should be called `TransactionValidator`?

> Is there a better way to verify a transaction in Miden? For now I am using the `verify()` method of the `TransactionVerifier` of the miden-tx crate. It seems suited for this situation.

Yes, that's the right struct/method.

> When using the `TransactionVerifier` from the miden-tx crate, you need to instantiate it with a certain "security_level", what is the security level that should be used for proof verification in the miden-node?

For now, let's use 96 bits. We will move to 120+ bits in the future, but probably not in the next couple of years.

> Which could potentially be normal considering that it's `dummy`.

Yes, dummy proofs won't pass verification. We have two options, I think:

1. Disable proof verification while testing using feature flags.
2. Generate valid transaction proofs for tests.

The second option would be ideal, but I also don't know how much more effort it is. If it is not too difficult, we should go with this option. But if it will take quite a bit of work, I'd turn off proof verification for tests and create an issue to fix this later.

phklive:
> We should try to have different names - but don't have great ideas at the moment. Maybe the one in the block producer should be called TransactionValidator?

Changed the naming to `TransactionValidator` in the block producer.

> For now, let's use 96 bits. We will move to 120+ bits in the future, but probably not in the next couple of years.

Understood, set it to 96.

> Yes, dummy proofs won't pass verification. We have two options, I think: 

I think that we could merge this PR with a `#[cfg(not(test))]` flag. And open an issue and make another PR that would add a functional prover to the node tests. What do you think?

phklive:
@bobbinth, what do you think about this error: https://github.com/0xPolygonMiden/miden-node/pull/171/files#diff-e5a2367dc1bde32c05ec20d7a483b14afdba6c744bd048ad9db9391169e3c4eeR65

There was no error that truly suited / represented the error situation of the proof being invalid. 

Added a new error type in the `VerifyTxError` enum called `InvalidTransactionProof`. Initially I wanted it to take as input the `ExecutionProof` but clippy says it's too large to be used or should be boxed (because of the size of the proof). 

Do you think that this error message is okay or should we add information?

bobbinth:
> Do you think that this error message is okay or should we add information?

Maybe add `TransactionId` to the enum?

phklive:
> > Do you think that this error message is okay or should we add information?
> 
> Maybe add `TransactionId` to the enum?

Added.

phklive:
Testing the node with the client with the new transaction proof validation. For now it's erroring out. Looking into it: 

SCR-20240123-llsf

phklive:
For now it fails: 

My suspiscion is that the --feature testing is creating invalid proofs, hence why failing the proof verification step. 

SCR-20240123-lrye

I am working on running the node and client without the testing flag to see if this would solve the issue. 

`My computer is burning atm... 🔥`

bobbinth:
> I am working on running the node and client without the testing flag to see if this would solve the issue.

This may take some time. I do think it should work fine with the `testing` flag as well, or if not, we should fix it. First step here is to make sure that proof verification works locally (i.e., we generate and verify the proof immediately in the same test). I noticed that in our "end-to-end" tests (e.g., [here](https://github.com/0xPolygonMiden/miden-base/blob/main/miden-tx/tests/p2id_script_test.rs) we only execute transactions, but don't try to prove and verify them.

Could you make a PR in miden base to make sure we can prove and verify transactions locally?

bobbinth:
btw, the error you listed usually (but not always) happens when the proof was generated with one set of inputs but is being verified against another set of inputs.

phklive:
> > I am working on running the node and client without the testing flag to see if this would solve the issue.
> 
> This may take some time. I do think it should work fine with the `testing` flag as well, or if not, we should fix it. First step here is to make sure that proof verification works locally (i.e., we generate and verify the proof immediately in the same test). I noticed that in our "end-to-end" tests (e.g., [here](https://github.com/0xPolygonMiden/miden-base/blob/main/miden-tx/tests/p2id_script_test.rs) we only execute transactions, but don't try to prove and verify them.
> 
> Could you make a PR in miden base to make sure we can prove and verify transactions locally?

Sure will do so. Then your advice is: 
- Make the PR on miden base
- Make sure that the Proof verification works as intended
- Merge this PR. 

right?

bobbinth:
Yes, I think we should merge this PR only after we get the end-to-end flow working, and for this we need to make sure things in miden base work correctly first.

PR #173, opened 2024-01-23
sync_state: remove block_path, fixes #164

### LINKED ISSUE #164: Remove `block_path` from `SyncStateResponse`
This would make `sync-state` endpoint more efficient and will improve handling of responses on the client. See https://github.com/0xPolygonMiden/miden-node/issues/152#issuecomment-1898116360 for more info.

hackaugusto:
as per the title, this removes `block_path` and fixes #164

### DIFF
--- a/proto/proto/requests.proto
+++ b/proto/proto/requests.proto
@@ -31,14 +31,34 @@ message GetBlockHeaderByNumberRequest {
 }
 
 // State synchronization request.
+//
+// Specifies state updates the client is intersted in. The server will return the first block which
+// contains a note matching `note_tags` or the chain tip. And the corresponding updates to
+// `nullifiers` and `account_ids` for that block range.
 message SyncStateRequest {
-    // Send updates to the client starting at this block.
+    // Last block known by the client. The response will contain data starting from the next block,
+    // until the first block which contains a note of matching the requested tag, or the chain tip
+    // if there are no notes.
     uint32 block_num = 1;
 
+    // Accounts' hash to include in the response.
+    //
+    // An account hash will be included if-and-only-if it is the latest update. Meaning it is
+    // possible there was an update to the account for the given range, but if it is not the latest,
+    // it won't be included in the response.
     repeated account.AccountId account_ids = 2;
 
-    // Tags and nullifiers are filters, both filters correspond to the high 16 bits of the real values.
+    // Determines the tags which the client is interested in. These are only the 16high bits of the
+    // note's complete tag.
+    //
+    // The above means it is not possible to request an specific note, but only a "note family",
+    // this is done to increase the privacy of the client, by hiding the note's the client is
+    // intereted on.
     repeated uint32 note_tags = 3;
+
+    // Determines the nullifiers the client is interested in.
+    //
+    // Similarly to the note_tags, this determins only the 16high bits of the target nullifier.
     repeated uint32 nullifiers = 4;
 }
 
--- a/proto/proto/responses.proto
+++ b/proto/proto/responses.proto
@@ -12,8 +12,7 @@ import "tsmt.proto";
 message ApplyBlockResponse {}
 
 message CheckNullifiersResponse {
-    // Each requested nullifier has its corresponding nullifier proof at the
-    // same position.
+    // Each requested nullifier has its corresponding nullifier proof at the same position.
     repeated tsmt.NullifierProof proofs = 1;
 }
 
@@ -39,19 +38,16 @@ message SyncStateResponse {
     // block header of the block with the first note matching the specified criteria
     block_header.BlockHeader block_header = 2;
 
-    // data needed to update the partial MMR from `block_num` to `block_header.block_num`
+    // data needed to update the partial MMR from `block_num + 1` to `block_header.block_num`
     mmr.MmrDelta mmr_delta = 3;
 
-    // Merkle path in the updated chain MMR to the block at `block_header.block_num`
-    merkle.MerklePath block_path = 4;
-
-    // a list of account hashes updated after `block_num` but not after `block_header.block_num`
+    // a list of account hashes updated after `block_num + 1` but not after `block_header.block_num`
     repeated AccountHashUpdate accounts = 5;
 
     // a list of all notes together with the Merkle paths from `block_header.note_root`
     repeated note.NoteSyncRecord notes = 6;
 
-    // a list of nullifiers created between `block_num` and `block_header.block_num`
+    // a list of nullifiers created between `block_num + 1` and `block_header.block_num`
     repeated NullifierUpdate nullifiers = 7;
 }
 
--- a/store/README.md
+++ b/store/README.md
@@ -131,11 +131,10 @@ contains excessive notes and nullifiers, client can make additional filtering of
 
 * `chain_tip`: `uint32` – number of the latest block in the chain.
 * `block_header`: `BlockHeader` – block header of the block with the first note matching the specified criteria.
-* `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `block_num` to `block_header.block_num`.
-* `block_path`: `MerklePath` – Merkle path in the updated chain MMR to the block at `block_header.block_num`.
-* `accounts`: `[AccountHashUpdate]` – a list of account hashes updated after `block_num` but not after `block_header.block_num`.
+* `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `block_num + 1` to `block_header.block_num`.
+* `accounts`: `[AccountHashUpdate]` – a list of account hashes updated after `block_num + 1` but not after `block_header.block_num`.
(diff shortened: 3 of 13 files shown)


igamigo:
I think, in order to keep the sync_state protocol consistent, that we would need to construct the MmrDelta slightly different, as explained in this comment: https://github.com/0xPolygonMiden/miden-node/issues/152#issuecomment-1898116360 (unless this was planned for a future change or there si anything I'm missing)

store/src/state.rs
```diff
@@ -342,27 +342,19 @@ impl State {
...
             .get_state_sync(block_num, account_ids, note_tag_prefixes, nullifier_prefixes)
             .await?;
 
-        let (delta, path) = {
-            let delta = inner
-                .chain_mmr
-                .get_delta(block_num as usize, state_sync.block_header.block_num as usize)?;
+        let delta = inner
+            .chain_mmr
+            .get_delta(block_num as usize, state_sync.block_header.block_num as usize)?;
```
bobbinth, line 355: I think we are missing a `+1` here. This should be something like:
```Rust
let start_forest = block_num as usize + 1;
let end_forest =state_sync.block_header.block_num as usize; 
let delta = inner.chain_mmr.get_delta(start_forest, end_forest)?;
```

For example, let's say the MMR has 7 leaves. And the latest block known to the client is block 4 (i.e., `forest=5`). In the request, the client would use `block_num=4` (the number of the last block known to the client). But the delta they need is to go from 5-leaf MMR to 6-leaf MMR - that is `get_delta(5, 6)` (we don't go to forest 7 because we also send block header for block 6).

This will also require special handling for cases when `start_forest >= end_forest` (the case when `start_forest > end_forest` won't be organically handled by the `Mmr`, I believe). If this happens, we just send an empty delta.

We should also document this very well, both here and in the readme docs.

hackaugusto: I'm missing something here.

- `block_num` is 0-indexed
- `forest` is 1-indexed
- to convert from one to the other, we need `+1`

The above means we have to change to:

```rust
get_delta(block_num + 1, state_sync.block_header.block_num + 1)?;
```

- The Mmr `get_delta` is inclusive on both values.
- The request is non-inclusive, on the users's `block_num`, but inclusive on the `block_header.block_num`.

Meaning we have to increase the start by another `+1`:

```rust
get_delta(block_num + 2, state_sync.block_header.block_num + 1)?;
```

But, I guess @igamigo has tested this already, and it works? So maybe I missing something here?

hackaugusto: Humm, I guess it is because we are returning the `Mmr` delta to the previous block, because of the self-referential issue, so we also have a `-1` on both because of that.

bobbinth: Yes, it is related to the self-referential issue but even without it, it could still make sense to have it work like this. Specifically:

1. For the `to_forest`, we probably still don't want to use `state_sync.block_header.block_num + 1` because we are sending the block header for `state_sync.block_header.block_num` to the client anyway and thus sending the delta for `forest = state_sync.block_header.block_num` makes sense.
2. For the `from_forest` we know that on the previous request the client had synced up to `forest = block_num + 1` (because we sent the delta up to `forest = block_num` + the block header for block at `block_num`), and so this is the value we should use for the starting forest on the new request.

hackaugusto: Added the `+1` and some docs.

store/src/state.rs
```diff
@@ -342,27 +342,27 @@ impl State {
...
+        //
+        // - The Mmr forest is 1-indexed whereas the block number is 0-indexed. The Mmr root
+        // contained in the block header always lag behind by one block, this is because the Mmr
+        // leaves are hashes of block headers, and we can't have self-referential hashes. These two
+        // points cancel out and don't require adjusting.
+        // - Mmr::get_delta is inclusive, whereas the sync_state request block_num is defined to be
+        // exclusive, so the from_forest has to be adjusted with a +1
+        let delta = inner
+            .chain_mmr
+            .get_delta((block_num + 1) as usize, state_sync.block_header.block_num as usize)?;
```
bobbinth, line 363: Wouldn't this fail if `block_num + 1 > state_sync.block_header.block_num` (because of [this](https://github.com/0xPolygonMiden/crypto/blob/next/src/merkle/mmr/full.rs#L174) condition)?

This could happen if both client and node are at block $n$, and the client makes a request using $n$ as `block_num`.

hackaugusto: ah, good point, added code to handle the corner case.

proto/proto/requests.proto
```diff
@@ -31,14 +31,37 @@ message GetBlockHeaderByNumberRequest {
 }
 
 // State synchronization request.
+//
+// Specifies state updates the client is intersted in. The server will return
+// the first block which contains a note matching `note_tags`. And the
+// corresponding updates to `nullifiers` and `account_ids` for that block
```
igamigo, line 37: Very minor detail, but it might be worth mentioning here (on the request instead of `block_num`) that if there is no match, the chain tip is returned?

hackaugusto: updated

store/src/state.rs
```diff
@@ -342,27 +342,35 @@ impl State {
...
+            // - The Mmr forest is 1-indexed whereas the block number is 0-indexed. The Mmr root
+            // contained in the block header always lag behind by one block, this is because the Mmr
+            // leaves are hashes of block headers, and we can't have self-referential hashes. These two
+            // points cancel out and don't require adjusting.
+            // - Mmr::get_delta is inclusive, whereas the sync_state request block_num is defined to be
+            // exclusive, so the from_forest has to be adjusted with a +1
+            let from_forest = (block_num + 1) as usize;
+            let to_forest = state_sync.block_header.block_num as usize;
+            inner.chain_mmr.get_delta(from_forest, to_forest)?
         };
```
bobbinth, line 371: Should we also add a check here handling potential malicious input? Specifically, what will happen here if `block_num` is much bigger than `state_sync.block_header.block_num`? We should probably catch this and return an error.

bobbinth: Or I guess this will happen naturally inside `ChainMmr::get_delta()`?

hackaugusto: > Or I guess this will happen naturally inside ChainMmr::get_delta()?

Yep

proto/proto/requests.proto
```diff
@@ -31,14 +31,37 @@ message GetBlockHeaderByNumberRequest {
 }
 
 // State synchronization request.
+//
+// Specifies state updates the client is intersted in. The server will return
+// the first block which contains a note matching `note_tags` or the chain tip.
+// And the corresponding updates to `nullifiers` and `account_ids` for that
+// block range.
```
bobbinth, line 38: Awesome comments!

minor nit: we usually wrap lines after 100 chars. Is there a reason to wrap them after 80 chars here?

store/README.md
```diff
@@ -132,7 +132,6 @@ contains excessive notes and nullifiers, client can make additional filtering of
 * `chain_tip`: `uint32` – number of the latest block in the chain.
 * `block_header`: `BlockHeader` – block header of the block with the first note matching the specified criteria.
 * `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `block_num` to `block_header.block_num`.
```
bobbinth, line 134: Should we change to:

`mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `block_num + 1` to `block_header.block_num`.

And the propagate this change to other files?

proto/proto/responses.proto
```diff
@@ -42,9 +42,6 @@ message SyncStateResponse {
     // data needed to update the partial MMR from `block_num` to `block_header.block_num`
     mmr.MmrDelta mmr_delta = 3;
 
-    // Merkle path in the updated chain MMR to the block at `block_header.block_num`
-    merkle.MerklePath block_path = 4;
-
     // a list of account hashes updated after `block_num` but not after `block_header.block_num`
     repeated AccountHashUpdate accounts = 5;
```
bobbinth, line 46: Similar comment here: should we use `block_num + 1` here? Also, should we say `request.block_num` (here and in other filed comments) - otherwise, I'm not sure it is clear that `block_num` is related to the request.

bobbinth (approved):
Looks great! Thank you! I left a couple of small comments inline - otherwise it is good to merge.

Unless, you want to document `SyncStateResponse` just as thoroughly :)

proto/proto/requests.proto
hackaugusto: update `*.proto` to 100 chars

store/README.md
hackaugusto: updated this file, which other files do you have in mind?

proto/proto/responses.proto
hackaugusto: updated

store/README.md
bobbinth: `rpc/README.md`

hackaugusto: updated

PR #178, opened 2024-01-24
db: add order by clauses

### LINKED ISSUE #175: select_accounts_by_block_range should order results by block number
The query below is not ordering the results:

https://github.com/0xPolygonMiden/miden-node/blob/58be2733cae265c8937391acc72cdc52d98af117/store/src/db/sql.rs#L429-L440

~If one account has more than one account hash change for the same region. The order of the results is undetermined.~ This is not true, I forgot the `account_id` is the primary key, so we only have the latest state.

hackaugusto:
As per the title. This adds order by to the select statements.

fixes #175

### DIFF
--- a/store/src/db/sql.rs
+++ b/store/src/db/sql.rs
@@ -63,7 +63,8 @@ pub fn insert_nullifiers_for_block(
 pub fn select_nullifiers(
     conn: &mut Connection
 ) -> Result<Vec<(RpoDigest, BlockNumber)>, anyhow::Error> {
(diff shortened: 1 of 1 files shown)


bobbinth (approved):
Looks good! Thank you!

The main motivation here is that this makes the output more predictable at the expense of slightly reduced efficiency?

hackaugusto:
> The main motivation here is that this makes the output more predictable at the expense of slightly reduced efficiency?

pretty much. In some cases it can prevent bugs from happening, but I haven't found any here.

PR #180, opened 2024-01-25
docs: add documentation to the store's State

hackaugusto:
as per the title

### DIFF
--- a/store/src/state.rs
+++ b/store/src/state.rs
@@ -134,6 +134,7 @@ impl From<NullifierStateForTransactionInput> for NullifierTransactionInputRecord
 }
 
 impl State {
+    /// Loads the state from the `db`.
     #[instrument(skip(db))]
     pub async fn load(mut db: Db) -> Result<Self, anyhow::Error> {
         let nullifier_tree = load_nullifier_tree(&mut db).await?;
@@ -320,6 +321,9 @@ impl State {
         Ok(())
     }
 
+    /// Queries a [BlockHeader] from the database.
+    ///
+    /// If [None] is given as the value of `block_num`, the latest [BlockHeader] is returned.
     pub async fn get_block_header(
         &self,
         block_num: Option<BlockNumber>,
@@ -327,6 +331,10 @@ impl State {
         self.db.select_block_header_by_block_num(block_num).await
     }
 
+    /// Generates membership proofs for each one of the `nullifiers` against the latest nullifier
+    /// tree.
+    ///
+    /// Note: these proofs are invalidated once the nullifier tree is modified, i.e. on a new block.
     pub async fn check_nullifiers(
         &self,
         nullifiers: &[RpoDigest],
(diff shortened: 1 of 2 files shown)


store/src/state.rs
```diff
@@ -444,16 +466,20 @@ impl State {
         Ok((account, nullifier_blocks))
     }
 
+    /// Lists known nullifiers with their inclusion blocks.
```
bobbinth, line 469: For this, and the 2 method below, I wonder if we should say "Lists all known..." to emphasize that all elements are returned.

Also, should we add a note that these methods are intended primarily for testing?

store/src/state.rs
```diff
@@ -411,6 +432,7 @@ impl State {
         Ok((latest, peaks, account_states))
     }
 
+    /// Returns data needed by the block producer to verify transactions proofs.
```
bobbinth, line 435: I would say: "Returns the data needed by the block producer to verify transaction validity." Because this data is not really needed to verify the proofs but rather that the proofs apply to valid account/nullifier states.

store/src/state.rs
```diff
@@ -320,13 +321,17 @@ impl State {
...
+    ///
+    /// If [None] is given as the value of `block_num`, the latest [BlockHeader] is returned.
     pub async fn get_block_header(
         &self,
         block_num: Option<BlockNumber>,
     ) -> Result<Option<block_header::BlockHeader>, anyhow::Error> {
         self.db.select_block_header_by_block_num(block_num).await
     }
 
+    /// Generates membership proofs for each one of the `nullifiers`.
```
bobbinth, line 334: Should we add a note to the effect of: "The membership proofs are generated against the most recent state of the nullifier tree and will become invalid as soon as the tree is updated".

Also note for the future: I wonder if as a part of this response we should also return the block number for the block against which the proofs were generated. Otherwise, the client may not be able to tell for which root the proofs are valid.

store/src/state.rs
```diff
@@ -335,6 +340,22 @@ impl State {
...
+    /// # Arguments
+    ///
+    /// - `block_num`: The last block *know* by the client, updates start from the next block.
+    /// - `account_ids`: Include the account's hash if their _last change_ was in the result's block
+    /// range.
+    /// - `note_tag_prefixes`: Only the 16 high bits of the tags the client is interested in, result
+    /// will include notes with matching prefixes, the first block with a matching note determines
+    /// the block range.
+    /// - `nullifier_prefixes`: Only the 16 high bits of the nullifiers the client is intersted in,
+    /// results will cinlude nullifiers matching prefixes produced in the given block range.
```
bobbinth, line 358: minor nit: we usually indent bullet-point lists a bit differently:
```Rust
/// - `block_num`: The last block *know* by the client, updates start from the next block.
/// - `account_ids`: Include the account's hash if their _last change_ was in the result's block
///   range.
/// - `note_tag_prefixes`: Only the 16 high bits of the tags the client is interested in, result
///   will include notes with matching prefixes, the first block with a matching note determines
///   the block range.
/// - `nullifier_prefixes`: Only the 16 high bits of the nullifiers the client is intersted in,
///   results will cinlude nullifiers matching prefixes produced in the given block range.
```

bobbinth (approved):
Looks great! Thank you for doing this! I left a few small nits inline. After these, we are good to merge.

store/src/state.rs
hackaugusto: changed

store/src/state.rs
hackaugusto: changed

store/src/state.rs
hackaugusto: changed

store/src/state.rs
hackaugusto: changed

PR #181, opened 2024-01-25
sql: add unit test for queries

hackaugusto:
As per the PR title. I wanted to add a test for the boundary case of a sync request where the client's block_num is the same as the chain tip, I also added a few tests to some of the existing functions that lacked tests, and tried to organized some of the code per function.


bobbinth (approved):
Looks good! Thank you!

PR #182, opened 2024-01-25
node: add error message on endpoint resolution failure

hackaugusto:
otherwise the code would just panic on the `addrs[0]`


bobbinth (approved):
Looks good! Thank you!

PR #188, opened 2024-01-25
block_builder: using borrows instead of refcount

hackaugusto:
The main code base didn't need the reference counting, and could be implemented with slices only. So this PR removes the `Arc<T>`s in favor of borrowing. The idea is that the types are a bit more straight forward (less wrappers), and we save a bit in atomic counting (not that important, most likely negligible).

### DIFF
--- a/block-producer/src/batch_builder/errors.rs
+++ b/block-producer/src/batch_builder/errors.rs
@@ -1,15 +1,31 @@
+use miden_objects::transaction::ProvenTransaction;
 use miden_vm::crypto::MerkleError;
 use thiserror::Error;
 
 use crate::MAX_NUM_CREATED_NOTES_PER_BATCH;
 
-#[derive(Error, Debug, PartialEq)]
+/// Error that may happen while building a transaction batch.
+///
+/// These errors are returned from the batch builder to the transaction queue, instead of
+/// dropping the transactions, they are included into the error values, so that the transaction
+/// queue can re-queue them.
+#[derive(Error, Debug)]
 pub enum BuildBatchError {
     #[error(
         "Too many notes in the batch. Got: {0}, max: {}",
         MAX_NUM_CREATED_NOTES_PER_BATCH
     )]
-    TooManyNotesCreated(usize),
+    TooManyNotesCreated(usize, Vec<ProvenTransaction>),
+
     #[error("failed to create notes SMT: {0}")]
-    NotesSmtError(#[from] MerkleError),
+    NotesSmtError(MerkleError, Vec<ProvenTransaction>),
+}
+
+impl BuildBatchError {
+    pub fn into_transactions(self) -> Vec<ProvenTransaction> {
+        match self {
+            BuildBatchError::TooManyNotesCreated(_, txs) => txs,
+            BuildBatchError::NotesSmtError(_, txs) => txs,
+        }
+    }
 }
--- a/block-producer/src/batch_builder/mod.rs
+++ b/block-producer/src/batch_builder/mod.rs
@@ -5,13 +5,13 @@ use tokio::{sync::RwLock, time};
 use tracing::{debug, info, instrument, Span};
 
 use self::errors::BuildBatchError;
-use crate::{block_builder::BlockBuilder, SharedProvenTx, SharedRwVec, SharedTxBatch, COMPONENT};
+use crate::{block_builder::BlockBuilder, ProvenTransaction, SharedRwVec, COMPONENT};
 
 pub mod errors;
 #[cfg(test)]
 mod tests;
 
-mod batch;
+pub mod batch;
 pub use batch::TransactionBatch;
 use miden_node_utils::logging::{format_array, format_blake3_digest};
 
@@ -32,7 +32,7 @@ pub trait BatchBuilder: Send + Sync + 'static {
     /// Start proving of a new batch.
     async fn build_batch(
         &self,
-        txs: Vec<SharedProvenTx>,
+        txs: Vec<ProvenTransaction>,
     ) -> Result<(), BuildBatchError>;
 }
 
@@ -50,7 +50,7 @@ pub struct DefaultBatchBuilderOptions {
 
 pub struct DefaultBatchBuilder<BB> {
     /// Batches ready to be included in a block
-    ready_batches: SharedRwVec<SharedTxBatch>,
+    ready_batches: SharedRwVec<TransactionBatch>,
 
     block_builder: Arc<BB>,
 
@@ -97,7 +97,7 @@ where
     /// Note that we call `build_block()` regardless of whether the `ready_batches` queue is empty.
     /// A call to an empty `build_block()` indicates that an empty block should be created.
     async fn try_build_block(&self) {
-        let mut batches_in_block: Vec<SharedTxBatch> = {
+        let mut batches_in_block: Vec<TransactionBatch> = {
             let mut locked_ready_batches = self.ready_batches.write().await;
 
             let num_batches_in_block =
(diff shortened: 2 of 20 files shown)


block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -155,15 +154,17 @@ impl BlockWitness {
...
                     |(
                         account_id,
                         AccountUpdate {
                             initial_state_hash,
                             final_state_hash: _,
                             proof,
                         },
-                    )| { (u64::from(account_id), initial_state_hash, proof) },
+                    )| {
+                        (u64::from(*account_id), *initial_state_hash, proof.clone())
```
plafer, line 166: This `proof.clone()` could be expensive - any reason why we change `self` -> `&self`?

plafer (approved):
LGTM (modulo few suggestions)!

The initial reason for using `Arc` was that, in the case where all components run in the same process, I thought it would give the implementor the opportunity to own the data if needed at no cost (i.e. no `clone()` needed).

However this PR made me realize that we can always tell when the implementor would need to own the data (e.g. `TransactionVerifier::verify_tx` should never need to own the data).

Left suggestions to add a reference where I thought it was appropriate.

block-producer/src/txqueue/mod.rs
```diff
@@ -159,7 +159,7 @@ where
 
     /// Divides the queue in groups to be batched; those that failed are appended back on the queue
     async fn try_build_batches(&self) {
-        let txs: Vec<SharedProvenTx> = {
+        let txs: Vec<ProvenTransaction> = {
             let mut locked_ready_queue = self.ready_queue.write().await;
 
             if locked_ready_queue.is_empty() {
```
bobbinth, line 165: If I remember correctly, the main reason why we used `SharedProvenTx` rather than just `ProvenTransaction` was to avoid cloning of `ProvenTransaction` objects. These objects could be pretty larger - e.g., 50KB - 100KB in size.

If I'm understanding correctly, with the current approach line 179 below would result in cloning of all transactions currently in the queue. Is the thinking that performance impact of this cloning is negligible?

block-producer/src/batch_builder/mod.rs
```diff
@@ -94,7 +94,7 @@ where
             locked_ready_batches.drain(..num_batches_in_block).collect()
         };
 
-        match self.block_builder.build_block(batches_in_block.clone()).await {
+        match self.block_builder.build_block(&batches_in_block).await {
             Ok(_) => {
                 // block successfully built, do nothing
             },
```
bobbinth, line 100: Here, I think it is fine as we don't introduce any extra cloning (block builder does not take ownership of the batches, and just clones relevant pieces from them as needed).

block-producer/src/state_view/mod.rs
```diff
@@ -138,7 +138,7 @@ where
 ///    `already_consumed_nullifiers`
 #[instrument(target = "miden-block-producer", skip_all, err)]
 fn ensure_in_flight_constraints(
-    candidate_tx: SharedProvenTx,
+    candidate_tx: ProvenTransaction,
```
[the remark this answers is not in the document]
hackaugusto: thanks! changed.

block-producer/src/state_view/mod.rs
```diff
@@ -172,7 +172,7 @@ fn ensure_in_flight_constraints(
 
 #[instrument(target = "miden-block-producer", skip_all, err)]
 fn ensure_tx_inputs_constraints(
-    candidate_tx: SharedProvenTx,
+    candidate_tx: ProvenTransaction,
```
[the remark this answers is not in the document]
hackaugusto: changed

block-producer/src/store/mod.rs
```diff
@@ -30,7 +30,7 @@ pub trait Store: ApplyBlock {
     /// TODO: add comments
     async fn get_tx_inputs(
         &self,
-        proven_tx: SharedProvenTx,
+        proven_tx: ProvenTransaction,
```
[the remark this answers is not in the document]
hackaugusto: changed

block-producer/src/txqueue/mod.rs
```diff
@@ -83,7 +83,7 @@ pub trait TransactionVerifier: Send + Sync + 'static {
     ///    perform the check above.
     async fn verify_tx(
         &self,
-        tx: SharedProvenTx,
+        tx: ProvenTransaction,
```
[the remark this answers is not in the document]
hackaugusto: changed

block-producer/src/txqueue/mod.rs
hackaugusto: removed the copy, the `txs` are now returned with the `Error`.

block-producer/src/block_builder/prover/block_witness.rs
hackaugusto: humm, I don't recall why I changed into a non-consuming version. Let me fix that

hackaugusto:
@bobbinth the comments have been addressed, @plafer has approved it already, but I'm not sure if you're against merging it. Let me know if it is okay to merge this PR.

block-producer/src/txqueue/mod.rs
```diff
@@ -160,16 +162,20 @@ where
...
                     },
-                    Err(_) => {
+                    Err(BuildBatchError::TooManyNotesCreated(_, mut txs)) => {
+                        // batch building failed, add txs back at the end of the queue
+                        ready_queue.write().await.append(&mut txs);
+                    },
+                    Err(BuildBatchError::NotesSmtError(_, mut txs)) => {
                         // batch building failed, add txs back at the end of the queue
                         ready_queue.write().await.append(&mut txs);
                     },
```
bobbinth, line 181: Would it make sense to implement something like `BuildBatchError::into_transactions()` to get the list of transactions regardless of the variant? This way, we won't have to modify this code as more variants are added `BuildBatchError` in the future.

block-producer/src/batch_builder/errors.rs
```diff
@@ -1,15 +1,17 @@
+use miden_objects::transaction::ProvenTransaction;
 use miden_vm::crypto::MerkleError;
 use thiserror::Error;
 
 use crate::MAX_NUM_CREATED_NOTES_PER_BATCH;
 
-#[derive(Error, Debug, PartialEq)]
+#[derive(Error, Debug)]
 pub enum BuildBatchError {
```
bobbinth, line 8: Could we add a comment here explaining what the internal `Vec<ProvenTransaction>` represents?

Also, in the future, if the number of variants gets big, we may want to refactor this to look something like this:
```Rust
pub struct BuildBatchError {
    err_type: BuildBatchErrorType,
    proposed_txs: Vec<ProvenTransaction>,
}
```

bobbinth (approved):
Looks good! Thank you! I left a couple of small comments inline. After these, we are good to merge.

hackaugusto:
One important note: I made some big assumptions with this PR, that I think it is worth writing down:

- I assume we don't want to include the same `tx` into multiple batches, as that would mean one of the batches is invalid (because they conflict with one another, and will have conflicting account hashes).
- I also assume we don't want to use the same `tx` for different purposes on two different subsystems, meaning that a single copy is sufficient, as the `tx` flows from one subsystem to another.

With the above in mind, it makes sense to use ownership when necessary, and exploiting Rust's affine type system to prevent the above.

bobbinth:
> I assume we don't want to include the same `tx` into multiple batches, as that would mean one of the batches is invalid (because they conflict with one another, and will have conflicting account hashes).

Correct - the same transaction should not be in two different batches.

> I also assume we don't want to use the same `tx` for different purposes on two different subsystems, meaning that a single copy is sufficient, as the `tx` flows from one subsystem to another.

I think this is correct as well, but I'm less certain here than with the previous assumption (mostly because I haven't thought about it too much).

block-producer/src/txqueue/mod.rs
hackaugusto: done

block-producer/src/batch_builder/errors.rs
hackaugusto: done

PR #189, opened 2024-01-26
(Logs) Build transactions batch

polydez:
Added logs for transaction batch building flow.

Sample output:

```
2024-01-29T09:01:42.062953Z  INFO block_producer:try_build_batches: miden-block-producer: block-producer/src/txqueue/mod.rs:162: new
2024-01-29T09:01:42.063159Z  INFO block_producer:try_build_batches:batch_builder: miden-block-producer: block-producer/src/txqueue/mod.rs:192: new
2024-01-29T09:01:42.063304Z  INFO block_producer:try_build_batches:batch_builder:build_batch: miden-block-producer: block-producer/src/batch_builder/mod.rs:116: new
2024-01-29T09:01:42.063354Z  INFO block_producer:try_build_batches:batch_builder:build_batch: miden-block-producer: block-producer/src/batch_builder/mod.rs:123: Building a transaction batch, num_txs: 1
2024-01-29T09:01:42.063412Z  INFO block_producer:try_build_batches:batch_builder:build_batch:new_batch: miden-block-producer: block-producer/src/batch_builder/batch.rs:41: new
2024-01-29T09:01:42.072082Z  INFO block_producer:try_build_batches:batch_builder:build_batch:new_batch: miden-block-producer: block-producer/src/batch_builder/batch.rs:41: close, time.busy: 8.14ms, time.idle: 526µs
2024-01-29T09:01:42.072189Z  INFO block_producer:try_build_batches:batch_builder:build_batch: miden-block-producer: block-producer/src/batch_builder/mod.rs:128: Transaction batch built
2024-01-29T09:01:42.072302Z  INFO block_producer:try_build_batches:batch_builder:build_batch: miden-block-producer: block-producer/src/batch_builder/mod.rs:116: close, time.busy: 8.95ms, time.idle: 50.4µs batch_id: "0x8a73a469f0af54c4c32588af08f7fd840f80a25e690a3247f252500fddc6f534"
2024-01-29T09:01:42.092294Z  INFO block_producer:try_build_batches:batch_builder: miden-block-producer: block-producer/src/txqueue/mod.rs:192: close, time.busy: 29.0ms, time.idle: 141µs
2024-01-29T09:01:42.092359Z  INFO block_producer:try_build_batches: miden-block-producer: block-producer/src/txqueue/mod.rs:162: close, time.busy: 174µs, time.idle: 29.3ms
```

### DIFF
--- a/block-producer/src/batch_builder/batch.rs
+++ b/block-producer/src/batch_builder/batch.rs
@@ -132,14 +132,12 @@ impl TransactionBatch {
     // --------------------------------------------------------------------------------------------
 
     fn compute_id(txs: &[SharedProvenTx]) -> BatchId {
-        let mut result = BatchId::default();
-
+        let mut buf = Vec::with_capacity(32 * txs.len());
         for tx in txs {
-            let tx_hash = Blake3_256::hash(&tx.id().as_bytes());
-            result = Blake3_256::merge(&[result, tx_hash]);
+            let mut tx_id = tx.id().as_bytes().to_vec();
+            buf.append(&mut tx_id);
         }
-
-        result
+        Blake3_256::hash(&buf)
     }
 }
 
--- a/block-producer/src/batch_builder/batch.rs
+++ b/block-producer/src/batch_builder/batch.rs
@@ -78,7 +78,7 @@ impl TransactionBatch {
             )
         };
 
-        let id = Self::calc_id(&txs);
+        let id = Self::compute_id(&txs);
 
         Ok(Self {
             id,
@@ -128,11 +128,14 @@ impl TransactionBatch {
         self.created_notes_smt.root()
     }
 
-    fn calc_id(txs: &[SharedProvenTx]) -> BatchId {
+    // HELPER FUNCTIONS
+    // --------------------------------------------------------------------------------------------
+
+    fn compute_id(txs: &[SharedProvenTx]) -> BatchId {
         let mut result = BatchId::default();
 
         for tx in txs {
-            let tx_hash = Blake3_256::hash_elements(tx.id().inner().as_elements());
+            let tx_hash = Blake3_256::hash(&tx.id().as_bytes());
             result = Blake3_256::merge(&[result, tx_hash]);
         }
 
--- a/block-producer/src/batch_builder/batch.rs
+++ b/block-producer/src/batch_builder/batch.rs
@@ -1,11 +1,14 @@
 use std::collections::BTreeMap;
 
+use miden_crypto::hash::blake::{Blake3Digest, Blake3_256};
 use miden_objects::{accounts::AccountId, notes::NoteEnvelope, Digest};
 use miden_vm::crypto::SimpleSmt;
 use tracing::instrument;
 
 use super::errors::BuildBatchError;
 use crate::{SharedProvenTx, CREATED_NOTES_SMT_DEPTH, MAX_NUM_CREATED_NOTES_PER_BATCH};
+
+pub type BatchId = Blake3Digest<32>;
 
 // TRANSACTION BATCH
 // ================================================================================================
@@ -16,6 +19,7 @@ use crate::{SharedProvenTx, CREATED_NOTES_SMT_DEPTH, MAX_NUM_CREATED_NOTES_PER_B
 /// Note: Until recursive proofs are available in the Miden VM, we don't include the common proof.
 #[derive(Debug)]
 pub struct TransactionBatch {
+    id: BatchId,
     updated_accounts: BTreeMap<AccountId, AccountStates>,
     produced_nullifiers: Vec<Digest>,
     created_notes_smt: SimpleSmt<CREATED_NOTES_SMT_DEPTH>,
@@ -34,7 +38,7 @@ impl TransactionBatch {
     /// - The number of created notes across all transactions exceeds 4096.
     ///
     /// TODO: enforce limit on the number of created nullifiers.
-    #[instrument(target = "miden-block-producer", skip_all, err)]
+    #[instrument(target = "miden-block-producer", name = "new_batch", skip_all, err)]
     pub fn new(txs: Vec<SharedProvenTx>) -> Result<Self, BuildBatchError> {
         let updated_accounts = txs
             .iter()
@@ -74,7 +78,10 @@ impl TransactionBatch {
             )
         };
 
+        let id = Self::calc_id(&txs);
+
         Ok(Self {
+            id,
             updated_accounts,
             produced_nullifiers,
             created_notes_smt,
@@ -84,6 +91,11 @@ impl TransactionBatch {
 
     // PUBLIC ACCESSORS
     // --------------------------------------------------------------------------------------------
+
+    /// Returns the batch ID.
+    pub fn id(&self) -> BatchId {
+        self.id
+    }
 
     /// Returns an iterator over (account_id, init_state_hash) tuples for accounts that were
     /// modified in this transaction batch.
@@ -114,6 +126,17 @@ impl TransactionBatch {
     /// Returns the root of the created notes SMT.
     pub fn created_notes_root(&self) -> Digest {
         self.created_notes_smt.root()
+    }
+
+    fn calc_id(txs: &[SharedProvenTx]) -> BatchId {
+        let mut result = BatchId::default();
+
+        for tx in txs {
+            let tx_hash = Blake3_256::hash_elements(tx.id().inner().as_elements());
+            result = Blake3_256::merge(&[result, tx_hash]);
+        }
+
+        result
     }
 }
 
--- a/block-producer/src/batch_builder/batch.rs
+++ b/block-producer/src/batch_builder/batch.rs
@@ -2,6 +2,7 @@ use std::collections::BTreeMap;
 
 use miden_objects::{accounts::AccountId, notes::NoteEnvelope, Digest};
 use miden_vm::crypto::SimpleSmt;
+use tracing::instrument;
 
 use super::errors::BuildBatchError;
 use crate::{SharedProvenTx, CREATED_NOTES_SMT_DEPTH, MAX_NUM_CREATED_NOTES_PER_BATCH};
@@ -33,6 +34,7 @@ impl TransactionBatch {
     /// - The number of created notes across all transactions exceeds 4096.
     ///
     /// TODO: enforce limit on the number of created nullifiers.
+    #[instrument(target = "miden-block-producer", skip_all, err)]
     pub fn new(txs: Vec<SharedProvenTx>) -> Result<Self, BuildBatchError> {
         let updated_accounts = txs
             .iter()
--- a/block-producer/src/batch_builder/batch.rs
+++ b/block-producer/src/batch_builder/batch.rs
@@ -134,8 +134,7 @@ impl TransactionBatch {
     fn compute_id(txs: &[SharedProvenTx]) -> BatchId {
         let mut buf = Vec::with_capacity(32 * txs.len());
         for tx in txs {
-            let mut tx_id = tx.id().as_bytes().to_vec();
-            buf.append(&mut tx_id);
+            buf.extend_from_slice(&tx.id().as_bytes());
         }
         Blake3_256::hash(&buf)
     }
--- a/block-producer/src/batch_builder/mod.rs
+++ b/block-producer/src/batch_builder/mod.rs
@@ -2,7 +2,7 @@ use std::{cmp::min, sync::Arc, time::Duration};
 
 use async_trait::async_trait;
 use tokio::{sync::RwLock, time};
-use tracing::{debug, info, instrument};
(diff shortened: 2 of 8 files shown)


store/src/server/mod.rs
```diff
@@ -12,7 +12,7 @@ mod api;
 // STORE INITIALIZER
 // ================================================================================================
 
-#[instrument(target = "miden-store", skip_all)]
+#[instrument(target = "miden-store", name = "store::serve", skip_all)]
```
bobbinth, line 15: Should we change the name just to `store` here? If I'm understanding how this works correctly, this will mean that we get `store` prefix on all requests to the store and then maybe we don't need to prefix them with `store` in other places?

rpc/src/server/mod.rs
```diff
@@ -11,7 +11,7 @@ mod api;
 
 // RPC INITIALIZER
 // ================================================================================================
-#[instrument(target = "miden-rpc", skip_all)]
+#[instrument(target = "miden-rpc", name = "rpc::serve", skip_all)]
```
bobbinth, line 14: Similar comment to the one above: maybe this should be just `rpc`?

block-producer/src/server/mod.rs
```diff
@@ -23,7 +23,7 @@ pub mod api;
 // ================================================================================================
 
 /// TODO: add comments
-#[instrument(target = "miden-block-producer", skip_all)]
+#[instrument(target = "miden-block-producer", name = "block_producer::serve", skip_all)]
```
bobbinth, line 26: Similar comment to the one above: maybe this can be just `block_producer` and then we can remove this prefix from other places?

block-producer/src/server/mod.rs
```diff
@@ -52,19 +52,15 @@ pub async fn serve(config: BlockProducerConfig) -> Result<()> {
...
-    tokio::spawn(async move {
-        queue
-            .run()
-            .instrument(info_span!(target: COMPONENT, "transaction_queue_start"))
-            .await
-    });
+    tokio::spawn(
+        async move { queue.run().await }
+            .instrument(info_span!(target: COMPONENT, "transaction_queue")),
+    );
```
bobbinth, line 58: I'm not sure we need to instrument this. I think batch building methods are named explicitly enough so that we won't confuse them with something else.

block-producer/src/server/mod.rs
```diff
@@ -52,19 +52,15 @@ pub async fn serve(config: BlockProducerConfig) -> Result<()> {
...
-    tokio::spawn(async move {
-        batch_builder
-            .run()
-            .instrument(info_span!(target: COMPONENT, "batch_builder_start"))
-            .await
-    });
+    tokio::spawn(
+        async move { batch_builder.run().await }
+            .instrument(info_span!(target: COMPONENT, "batch_builder")),
+    );
```
bobbinth, line 63: Similar comment to the one above: I'm not sure we need to instrument this.

block-producer/src/txqueue/mod.rs
```diff
@@ -148,6 +148,7 @@ where
         }
     }
 
+    #[instrument(target = "miden-block-producer", skip_all)]
```
bobbinth, line 151: I'm not sure instrumenting this gives us much additional value as compared to instrumenting `try_build_batches` below. So, I'd probably not do it.

block-producer/src/txqueue/mod.rs
```diff
@@ -175,17 +177,22 @@ where
...
+                        Err(_) => {
+                            // batch building failed, add txs back at the end of the queue
+                            ready_queue.write().await.append(&mut txs);
+                        },
+                    }
                 }
-            });
+                .instrument(
+                    info_span!(target: COMPONENT, "block_producer::batch_builder::build_batch"),
+                ),
```
bobbinth, line 194: I would probably remove this as well.

block-producer/src/batch_builder/mod.rs
```diff
@@ -111,17 +112,19 @@ impl<BB> BatchBuilder for DefaultBatchBuilder<BB>
 where
     BB: BlockBuilder,
 {
+    #[instrument(target = "miden-block-producer", skip_all)]
```
bobbinth, line 115: Should we not log an error here as well?

block-producer/src/batch_builder/mod.rs
```diff
@@ -111,17 +112,19 @@ impl<BB> BatchBuilder for DefaultBatchBuilder<BB>
...
     async fn build_batch(
         &self,
         txs: Vec<SharedProvenTx>,
     ) -> Result<(), BuildBatchError> {
         let num_txs = txs.len();
 
+        info!(target: COMPONENT, num_txs, "Batch building started");
```
bobbinth, line 122: nit: I'd change the message to "Building a transaction batch".

block-producer/src/batch_builder/mod.rs
```diff
@@ -111,17 +112,19 @@ impl<BB> BatchBuilder for DefaultBatchBuilder<BB>
...
+        debug!(target: COMPONENT, txs = %format_array(txs.iter().map(|tx| tx.id().to_hex())));
+
         let batch = Arc::new(TransactionBatch::new(txs)?);
         self.ready_batches.write().await.push(batch);
 
-        info!(COMPONENT, "batch built with {num_txs} txs");
-
```
bobbinth, line 124: I would not remove this but modify to log "Transaction batch built". We should log the batch ID (see next comment) and the total number of batches now in the queue (similar to what we did in the submit proven transaction flow).

For login batch ID, we could also do it as a part of `instrument` on line 115. For this, we could declare a field there and provide the value of this field later (see [here](https://docs.rs/tracing/latest/tracing/attr.instrument.html) under "Adding an empty field to be recorded later").

block-producer/src/batch_builder/batch.rs
```diff
@@ -33,6 +34,7 @@ impl TransactionBatch {
     /// - The number of created notes across all transactions exceeds 4096.
     ///
     /// TODO: enforce limit on the number of created nullifiers.
+    #[instrument(target = "miden-block-producer", skip_all, err)]
     pub fn new(txs: Vec<SharedProvenTx>) -> Result<Self, BuildBatchError> {
         let updated_accounts = txs
             .iter()
```
bobbinth, line 40: As a part of this constructor, I think we should also compute "batch ID". This would be a new field that we could add to the `TransactionBatch` struct.

A simple way to define a unique ID per batch is to compute a hash of all transaction IDs. For this, we could use some fast hash function. Probably the best option is [Blake3](https://github.com/0xPolygonMiden/crypto/blob/main/src/hash/blake/mod.rs#L107) which we already have in `miden-crypto`.

Lastly, I would probably change the name of the span to `new_batch`.

bobbinth (changes requested):
Thank you! Looks good! I added some comments inline - most of them are related to removing some intermediary spans which I don't think add much value. Basically, in my mind the deepest span path could look like so:
```
block_producer:try_build_batches:build_batch:new_batch
```
As compared to the current:
```
block_producer::serve:transaction_queue:run:try_build_batches:block_producer::batch_builder::build_batch:build_batch:new
```

block-producer/src/txqueue/mod.rs
polydez: @bobbinth I tried to remove this and then it starts to log children as root spans. I'll keep it for now, just reduce to `block_producer`, but need to investigate its behavior more.

bobbinth: In this case, I'd probably use `batch_builder` rather than `block_producer`.

Does this span get closed properly? (or is it just a few lines missing in the logs you uploaded?)

polydez: @bobbinth thank you! I've fixed it. Yes, I missed 2 lines of logs, now it should be good.

block-producer/src/server/mod.rs
```diff
@@ -52,19 +52,9 @@ pub async fn serve(config: BlockProducerConfig) -> Result<()> {
...
-    });
+    tokio::spawn(async move { queue.run().await });
 
-    tokio::spawn(async move {
-        batch_builder
-            .run()
-            .instrument(info_span!(target: COMPONENT, "batch_builder_start"))
-            .await
-    });
+    tokio::spawn(async move { batch_builder.run().await });
```
bobbinth, line 57: Should we put just `info` events here - e.g., `transaction queue started` and `batch builder started`?

block-producer/src/batch_builder/batch.rs
```diff
@@ -113,6 +127,17 @@ impl TransactionBatch {
     pub fn created_notes_root(&self) -> Digest {
         self.created_notes_smt.root()
     }
+
+    fn calc_id(txs: &[SharedProvenTx]) -> BatchId {
```
bobbinth, line 131: A couple nits:

1. Let's add a section separator above called "HELPER FUNCTIONS" as this is not a public accessor.
2. Let's call this function `compute_id()`.

block-producer/src/batch_builder/batch.rs
```diff
@@ -113,6 +127,17 @@ impl TransactionBatch {
...
+        let mut result = BatchId::default();
+
+        for tx in txs {
+            let tx_hash = Blake3_256::hash_elements(tx.id().inner().as_elements());
+            result = Blake3_256::merge(&[result, tx_hash]);
+        }
```
bobbinth, line 137: The way this look currently works will result in quite a bit of unnecessary hashing. What I would probably is:

1. Build a byte vector (using `tx.id().as_bytes`).
2. Hash this vector using `Blake3_256::hash()`.

My guess is that this will reduce the amount of hashing by about 3x.

bobbinth (approved):
Thank you! Looks good! I left a couple of small comments inline. After these, we are good to merge.

block-producer/src/batch_builder/mod.rs
```diff
@@ -111,16 +112,23 @@ impl<BB> BatchBuilder for DefaultBatchBuilder<BB>
...
+        debug!(target: COMPONENT, txs = %format_array(txs.iter().map(|tx| tx.id().to_hex())));
+
         let batch = Arc::new(TransactionBatch::new(txs)?);
-        self.ready_batches.write().await.push(batch);
 
-        info!(COMPONENT, "batch built with {num_txs} txs");
+        info!(target: COMPONENT, "Transaction batch built");
+        Span::current().record("batch_id", format_blake3_digest(batch.id()));
+
+        self.ready_batches.write().await.push(batch);
```
bobbinth, line 132: Let's add another info event here: "Transaction batch added to the batch queue" and add a filed for this event with the total number of batches in the queue.

block-producer/src/batch_builder/batch.rs
```diff
@@ -128,11 +128,14 @@ impl TransactionBatch {
...
+    // --------------------------------------------------------------------------------------------
+
+    fn compute_id(txs: &[SharedProvenTx]) -> BatchId {
         let mut result = BatchId::default();
 
         for tx in txs {
-            let tx_hash = Blake3_256::hash_elements(tx.id().inner().as_elements());
+            let tx_hash = Blake3_256::hash(&tx.id().as_bytes());
             result = Blake3_256::merge(&[result, tx_hash]);
         }
```
bobbinth, line 140: This would still result in too much hashing. Please see my suggestion from the previous comment.

block-producer/src/server/mod.rs
```diff
@@ -53,17 +53,13 @@ pub async fn serve(config: BlockProducerConfig) -> Result<()> {
     let block_producer = api_server::ApiServer::new(api::BlockProducerApi::new(queue.clone()));
 
     tokio::spawn(async move {
-        queue
-            .run()
-            .instrument(info_span!(target: COMPONENT, "transaction_queue_start"))
-            .await
+        info!(target: COMPONENT, "Transaction queue started");
+        queue.run().await
     });
```
bobbinth, line 58: Should these be in reverse order? For exmaple:
```Rust
tokio::spawn(async move {
    queue.run().await;
    info!(target: COMPONENT, "Transaction queue started");
});
```

polydez: No, since `queue.run().await` starts infinite loop. But we could move this info into the `run` method.

block-producer/src/batch_builder/batch.rs
```diff
@@ -113,6 +127,18 @@ impl TransactionBatch {
...
+
+    // HELPER FUNCTIONS
+    // --------------------------------------------------------------------------------------------
+
+    fn compute_id(txs: &[SharedProvenTx]) -> BatchId {
+        let mut buf = Vec::with_capacity(32 * txs.len());
+        for tx in txs {
+            let mut tx_id = tx.id().as_bytes().to_vec();
+            buf.append(&mut tx_id);
+        }
```
bobbinth, line 139: Here, I would do something like:
```Rust
but.extend_from_slice(&tx.id().as_bytes());
```

PR #192, opened 2024-01-26
block_producer: add docs for BatchBuilder trait

hackaugusto:
As per the PR title.

I made an assumption of the design goal while writing the docs, namely that the trait is abstracting over a batch proving strategy, and it can represent a pool of distributed machines instead of an unique machine.

### DIFF
--- a/block-producer/src/batch_builder/mod.rs
+++ b/block-producer/src/batch_builder/mod.rs
@@ -17,9 +17,18 @@ pub use batch::TransactionBatch;
 // BATCH BUILDER
 // ================================================================================================
 
+/// Abstraction over batch proving of transactions.
+///
+/// Transactions are aggregated into batches prior to being added to blocks. This trait abstracts
+/// over this responsibility. The trait's goal is to be implementation agnostic, allowing for
+/// multiple implementations, e.g.:
+///
+/// - in-process cpu based prover
+/// - out-of-process gpu based prover
+/// - distributed prover on another machine
 #[async_trait]
 pub trait BatchBuilder: Send + Sync + 'static {
-    /// TODO: add doc comments
+    /// Start proving of a new batch.
     async fn build_batch(
         &self,
         txs: Vec<SharedProvenTx>,


block-producer/src/batch_builder/mod.rs
```diff
@@ -17,9 +17,18 @@ pub use batch::TransactionBatch;
 // BATCH BUILDER
 // ================================================================================================
 
+/// Abstraction over batch proving of transactions.
+///
+/// Transactions are aggregated into batches prior to being added to blocks. This trait abstracts
+/// over this responsability. The trait's goal is to be implementation agnostic, allowing for
```
plafer, line 23:
```suggestion
/// over this responsibility. The trait's goal is to be implementation agnostic, allowing for
```

plafer (approved):
LGTM! Thank you for doing this 🙂

block-producer/src/batch_builder/mod.rs
hackaugusto: oops, fixed!

> LGTM! Thank you for doing this 🙂

ofcourse :)

block-producer/src/batch_builder/mod.rs
```diff
@@ -17,9 +17,18 @@ pub use batch::TransactionBatch;
...
+///
+/// Transactions are aggregated into batches prior to being added to blocks. This trait abstracts
+/// over this responsibility. The trait's goal is to be implementation agnostic, allowing for
+/// multiple implementations, e.g.:
+///
+/// - in-process cpu based prover
+/// - out-of-process gpu based prover
+/// - distributed prover on another machine
 #[async_trait]
 pub trait BatchBuilder: Send + Sync + 'static {
```
bobbinth, line 30: Looking at this now, I wonder if we should structure things a bit differently. Specifically, this is currently responsible for both building batches and adding them to the "batch queue" (e.g., `build_batch()` doesn't return anything). We probably should separate these into:

1. Batch builder - which takes a set of transactions and returns a transaction batch.
2. Batch queue/pool - which maintains a set of built batches.

PR #197, opened 2024-01-30
txqueue: document where empty blocks come from

hackaugusto:
this behavior is a bit surprising, i think it deserves a small comment.

### DIFF
--- a/block-producer/src/txqueue/mod.rs
+++ b/block-producer/src/txqueue/mod.rs
@@ -155,6 +155,8 @@ where
         let txs: Vec<ProvenTransaction> = {
             let mut locked_ready_queue = self.ready_queue.write().await;
 
+            // If there are no transactions in the queue, this call is a no-op. The [BatchBuilder]
+            // will produce empty blocks if necessary.
             if locked_ready_queue.is_empty() {
                 return;
             }


bobbinth (approved):
Looks good! Thank you!

Curious what was surprising about this behavior? (i.e., what would you have expected to happen here?)

hackaugusto:
> Curious what was surprising about this behavior? (i.e., what would you have expected to happen here?)

Ah, I didn't mean the code has surprising behavior. It is very much straightforward, just an `if` and a `return`.

What was surprising is the difference in behavior among the components, they all behave/are implemented differently:

- txqueue is a no-op on empty queues
- batch builder is building things on an empty queue
- block builder doesn't have a queue

When I was reading the code I knew that we produce empty blocks, and I kinda knew the implementation because I had read the batch builder multiple times. So when I noticed a `return` on an empty queue, my first reaction was that it had to be a bug, instead it is just a difference among the components.

bobbinth:
> What was surprising is the difference in behavior among the components, they all behave/are implemented differently:
> 
> * txqueue is a no-op on empty queues
> * batch builder is building things on an empty queue
> * block builder doesn't have a queue
> 
> When I was reading the code I knew that we produce empty blocks, and I kinda knew the implementation because I had read the batch builder multiple times. So when I noticed a `return` on an empty queue, my first reaction was that it had to be a bug, instead it is just a difference among the components.

Makes sense. Maybe we can make the methodologies more regular, though some if it is inherent in the requirements. For example:

1. We need to produce blocks regularly, and so, if there is nothing to put into a block, we produce an empty block.
2. Unlike with blocks, we don't need to produce batches regularly, and so, producing empty batches would be a waste of resources.

Issue #176, opened 2024-01-24
Verification fails for transactions from new accounts


igamigo:
While testing the client, I'm trying to send transactions to the node with new accounts (in this case, creating a note with assets from a fungible faucet). Local execution works fine, but when submitting the proof to the node I'm getting this:

```
rpc api error: rpc request failed for submit_proven_transaction: status: InvalidArgument, message: "VerificationFailed(IncorrectAccountInitialHash { tx_initial_account_hash: RpoDigest([7804508306418852326, 6348915401079023739, 4425441608650618110, 7471845003988793641]), store_account_hash: Some(RpoDigest([0, 0, 0, 0])) })", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "content-length": "0", "date": "Wed, 24 Jan 2024 16:05:19 GMT"} }
```

Looks like the node checks the account's hash and because it has a blank hash for new accounts, it never validates.

phklive:
@igamigo will look into it, interestingly the verification of proofs also fail when sending a `ProvenTransaction` from the client to the node. 

I am working on adding the proof verification on the node `RPC` and `Block Producer` in this PR: https://github.com/0xPolygonMiden/miden-node/pull/171

phklive:
For now this is the error that we are getting when trying to verify a proof from the client on the node.

<img width="629" alt="SCR-20240126-muup" src="https://github.com/0xPolygonMiden/miden-node/assets/42912740/902a89c6-76bb-4efd-a639-a553da130ebf">

igamigo:
Is that by using accounts that are not in the genesis state, or by importing the accounts beforehand (or maybe it happens in both scenarios)?

phklive:
> Is that by using accounts that are not in the genesis state, or by importing the accounts beforehand (or maybe it happens in both scenarios)?

By importing accounts beforehand and making a tx, sending it to the running node and trying to verify the proof. 

In this PR: #171

phklive:
Added a verifier with 96 min_security in the miden_client for testing just after the proof is executed. It errors out. That rules out the possibility of error in the message transmission, serialisation / deserialisation.

```rust
Verif error: TransactionVerificationFailed(
    VerifierError(
        InconsistentOodConstraintEvaluations,
    ),
)
```

bobbinth:
[...]

PR #200, opened 2024-01-30
Fix failed verification for new accounts

phklive:
The checks that were performed on new txs received by the node were preventing new accounts to submit txs. 

Because the `account_hash` of the tx account did not match the `account_hash` contained in the DB for this specific tx account considering that it had never be added. 

By default it was `Digest::default()`

Closes: #176

### DIFF
--- a/block-producer/src/state_view/mod.rs
+++ b/block-producer/src/state_view/mod.rs
@@ -179,7 +179,9 @@ fn ensure_tx_inputs_constraints(
 
     match tx_inputs.account_hash {
         Some(store_account_hash) => {
-            if candidate_tx.initial_account_hash() != store_account_hash {
+            if candidate_tx.initial_account_hash() != store_account_hash
+                && store_account_hash != Digest::default()
+            {
                 return Err(VerifyTxError::IncorrectAccountInitialHash {
                     tx_initial_account_hash: candidate_tx.initial_account_hash(),
                     store_account_hash: Some(store_account_hash),
--- a/block-producer/src/state_view/mod.rs
+++ b/block-producer/src/state_view/mod.rs
@@ -179,13 +179,17 @@ fn ensure_tx_inputs_constraints(
 
     match tx_inputs.account_hash {
         Some(store_account_hash) => {
-            if candidate_tx.initial_account_hash() != store_account_hash
-                && store_account_hash != Digest::default()
-            {
-                return Err(VerifyTxError::IncorrectAccountInitialHash {
-                    tx_initial_account_hash: candidate_tx.initial_account_hash(),
-                    store_account_hash: Some(store_account_hash),
-                });
+            if candidate_tx.initial_account_hash() != store_account_hash {
+                // Prevents this check from returning an Err() in case a new account
+                // that has not yet been stored in the DB submits a transaction. By default
+                // the `store_account_hash` of a new account would be `Digest::default()`
+                if store_account_hash == Digest::default() {
+                } else {
+                    return Err(VerifyTxError::IncorrectAccountInitialHash {
+                        tx_initial_account_hash: candidate_tx.initial_account_hash(),
+                        store_account_hash: Some(store_account_hash),
+                    });
+                }
             }
         },
         None => {
--- a/block-producer/src/state_view/mod.rs
+++ b/block-producer/src/state_view/mod.rs
@@ -178,8 +178,8 @@ fn ensure_tx_inputs_constraints(
     debug!(target: COMPONENT, %tx_inputs);
 
     match tx_inputs.account_hash {
-        // Having `Some(store_account_hash)` here signifies that we are operating
-        // on the state of an existing account.
+        // if the account is present in the Store, make sure that the account state hash
+        // from the received transaction is the same as the one from the Store
         Some(store_account_hash) => {
             if candidate_tx.initial_account_hash() != store_account_hash {
                 return Err(VerifyTxError::IncorrectAccountInitialHash {
@@ -188,8 +188,7 @@ fn ensure_tx_inputs_constraints(
                 });
             }
         },
-        // Having `None` here signifies that we are operating
-        // on the state of a new account
+        // if the account is not present in the Store, it must be a new account
         None => {
             // TODO: check if the transaction was executed against new account
         },
--- a/block-producer/src/state_view/mod.rs
+++ b/block-producer/src/state_view/mod.rs
@@ -178,25 +178,20 @@ fn ensure_tx_inputs_constraints(
     debug!(target: COMPONENT, %tx_inputs);
 
     match tx_inputs.account_hash {
+        // Having `Some(store_account_hash)` here signifies that we are operating
+        // on the state of an existing account.
         Some(store_account_hash) => {
             if candidate_tx.initial_account_hash() != store_account_hash {
-                // Prevents this check from returning an Err() in case a new account
-                // that has not yet been stored in the DB submits a transaction. By default
-                // the `store_account_hash` of a new account would be `Digest::default()`
-                if store_account_hash == Digest::default() {
-                } else {
-                    return Err(VerifyTxError::IncorrectAccountInitialHash {
-                        tx_initial_account_hash: candidate_tx.initial_account_hash(),
-                        store_account_hash: Some(store_account_hash),
-                    });
-                }
+                return Err(VerifyTxError::IncorrectAccountInitialHash {
+                    tx_initial_account_hash: candidate_tx.initial_account_hash(),
+                    store_account_hash: Some(store_account_hash),
+                });
             }
         },
+        // Having `None` here signifies that we are operating
+        // on the state of a new account
         None => {
-            return Err(VerifyTxError::IncorrectAccountInitialHash {
-                tx_initial_account_hash: candidate_tx.initial_account_hash(),
-                store_account_hash: None,
-            })
+            // TODO: check if the transaction was executed against new account
         },
     }
 
--- a/block-producer/src/store/mod.rs
+++ b/block-producer/src/store/mod.rs
@@ -181,9 +181,15 @@ impl Store for DefaultStore {
             nullifiers.into_iter().collect()
         };
 
-        let tx_inputs = TxInputs {
-            account_hash,
-            nullifiers,
+        let tx_inputs = match account_hash {
+            Some(hash) if hash == Digest::default() => TxInputs {
+                account_hash: None,
+                nullifiers,
+            },
+            _ => TxInputs {
+                account_hash,
+                nullifiers,
+            },
         };
 
         debug!(target: COMPONENT, %tx_inputs);
--- a/block-producer/src/store/mod.rs
+++ b/block-producer/src/store/mod.rs
@@ -183,10 +183,14 @@ impl Store for DefaultStore {
             nullifiers.into_iter().collect()
         };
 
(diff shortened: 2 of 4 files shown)


block-producer/src/state_view/mod.rs
```diff
@@ -179,7 +179,9 @@ fn ensure_tx_inputs_constraints(
 
     match tx_inputs.account_hash {
         Some(store_account_hash) => {
-            if candidate_tx.initial_account_hash() != store_account_hash {
+            if candidate_tx.initial_account_hash() != store_account_hash
+                && store_account_hash != Digest::default()
+            {
```
bobbinth, line 184: I would document this much better and would probably split into two branches for clarity (even if ` store_account_hash != Digest::default()` branch is a no-op)

phklive: This PR should have been in draft. 

Absolutely, added your demands.

block-producer/src/state_view/mod.rs
```diff
@@ -180,10 +180,16 @@ fn ensure_tx_inputs_constraints(
...
+                if store_account_hash == Digest::default() {
+                } else {
+                    return Err(VerifyTxError::IncorrectAccountInitialHash {
+                        tx_initial_account_hash: candidate_tx.initial_account_hash(),
+                        store_account_hash: Some(store_account_hash),
+                    });
+                }
             }
         },
         None => {
```
bobbinth, line 195: After looking at this a bit more, a few thoughts:

- It seems like there is some inconsistency here between how store and block producer work. Specifically, if the account is not found in the store, `account_hash` should be `None` rather than `Digest::default()`. So, `Digest::default()` case should never happen and we should not handle it in a special way.
- If the account is not present in the store, this means that the transaction was executed against a new account. Ideally, we should handle this case separately. Specifically, `ProvenTransaction` should have a flag indicating whether it was executed against a new account or not, and this flag should be enforced in the transaction kernel. I'll create an issue for this in `miden-base`.

Overall, this section should look something like this:
```Rust
match tx_inputs.account_hash {
    Some(store_account_hash) => {
        if candidate_tx.initial_account_hash() != store_account_hash {
            return Err(VerifyTxError::IncorrectAccountInitialHash {
                tx_initial_account_hash: candidate_tx.initial_account_hash(),
                store_account_hash: Some(store_account_hash),
            });
        }
    },
    None => {
        // check if the transaction was executed against new account
    }
}
```

bobbinth (changes requested):
Thank you! Looks good! I left a comment inline. Basically, I think the issue actually needs to be fixed in the `store` and we also need to make a small modification to the transaction kernel to make this work properly.

block-producer/src/state_view/mod.rs
phklive: My understanding of the situation is the following: 

1. In the miden-node we are calling the `get_tx_inputs` function on the `Store` trait that has been redefined in the Block Producer crate. 

https://github.com/0xPolygonMiden/miden-node/blob/3f657a0f45ab22b36382c16446bdd66b4d74715f/block-producer/src/store/mod.rs#L117-L192

2. This function sends a request to the API of the Miden Store. Which makes a call to the `get_tx_inputs` function of the `State` of the Miden Store.

https://github.com/0xPolygonMiden/miden-node/blob/3f657a0f45ab22b36382c16446bdd66b4d74715f/store/src/state.rs#L488-L541

3. This is the function responsible for building the response which is:

```rust
Result<(AccountState, Vec<NullifierStateForTransactionInput>), anyhow::Error>
```

4. This response is then used in the `Block Producer` to build the `TxInputs` struct. The part that we care about in the construction of this response is the `account_hash` of the `AccountState`: 

https://github.com/0xPolygonMiden/miden-node/blob/3f657a0f45ab22b36382c16446bdd66b4d74715f/store/src/state.rs#L499-L503

5. This `account_hash` is built from the `account_tree` using the `account_id`. the `open` method of the `account_tree` returns a `ValuePath` where we isolate the `value`

https://github.com/0xPolygonMiden/crypto/blob/c86bdc6d515046513a86ec331da5a0f77fd29cf2/src/merkle/path.rs#L164-L171

6. We then use into to turn this `value` which is an `RpoDigest` into a `Word`.

We understand here that the information that we receive in the Miden Node by calling `get_tx_inputs` is a `TxInputs` built through this process. 

Which also clarifies the fact that there is no possibility here for a Digest to be `None` considering that the `account_hash` that is built is an `RpoDigest` turned into a `Word`.

phklive: I have the following questions: 

What is the behaviour that we should have in the Miden Store and in the Miden Node?

Should the `account_tree` return `None` if the account does not exist in the tree?

bobbinth: > Should the `account_tree` return `None` if the account does not exist in the tree?

I don't think we need to modify the `account_tree` (I believe we use `Smt` for this).

I guess my point was more about `TxInputs` struct and how it is built from the store response. Basically, it look like this right now:
```Rust
pub struct TxInputs {
    pub account_hash: Option<Digest>,
    pub nullifiers: BTreeMap<Digest, bool>,
}
```
If we use `Digest::default()` to indicate that the account is not present in the store, then `account_hash` shouldn't be an option. But I think it would actually better if it is `Option` and we initialize it appropriately based on the store's response (when we construct `TxInputs`).

phklive: An `Option<Digest>` is already what we have:

https://github.com/0xPolygonMiden/miden-node/blob/3f657a0f45ab22b36382c16446bdd66b4d74715f/block-producer/src/store/mod.rs#L52-L60

What I understand then is that we do not need to change the Store but we should change the way that we build our `TxInputs` in the Node:

https://github.com/0xPolygonMiden/miden-node/blob/3f657a0f45ab22b36382c16446bdd66b4d74715f/block-producer/src/store/mod.rs#L144-L164

Hence adding the fact that if `account_hash` == `Digest::default()` then we return `None`, does that seem correct to you?

bobbinth: > Hence adding the fact that if `account_hash` == `Digest::default()` then we return `None`, does that seem correct to you?

Yes - i think that would be the correct way to do it.

phklive: > After looking at this a bit more, a few thoughts:
> 
> * It seems like there is some inconsistency here between how store and block producer work. Specifically, if the account is not found in the store, `account_hash` should be `None` rather than `Digest::default()`. So, `Digest::default()` case should never happen and we should not handle it in a special way.
> * If the account is not present in the store, this means that the transaction was executed against a new account. Ideally, we should handle this case separately. Specifically, `ProvenTransaction` should have a flag indicating whether it was executed against a new account or not, and this flag should be enforced in the transaction kernel. I'll create an issue for this in `miden-base`.
> 
> Overall, this section should look something like this:
> 
> ```rust
> match tx_inputs.account_hash {
>     Some(store_account_hash) => {
>         if candidate_tx.initial_account_hash() != store_account_hash {
>             return Err(VerifyTxError::IncorrectAccountInitialHash {
>                 tx_initial_account_hash: candidate_tx.initial_account_hash(),
>                 store_account_hash: store_account_hash,
>             });
>         }
>     },
>     None => {
>         // check if the transaction was executed against new account
>     }
> }
> ```

Could you clarify here what you meant by: `// check if the transaction was executed against new account`?

bobbinth: For now, this would be just a `TODO`. Once we implement the transaction kernel change I mentioned above (I still need to create an issue for this), we'll come back and handle it properly.

phklive:
SCR-20240201-khhm

Fixed.

phklive:
This test seems to have been built upon the wrong assumption that we should block accounts that are not existent in the Store. Even though accounts that are not in the Store if I am correct should be new accounts.

https://github.com/0xPolygonMiden/miden-node/blob/c4344cb837199f468bbf442faac35c80cb5d70d7/block-producer/src/state_view/tests/verify_tx.rs#L116-L145

Will remove / modify this test to handle correct behaviour.

block-producer/src/store/mod.rs
```diff
@@ -181,9 +181,15 @@ impl Store for DefaultStore {
...
-            nullifiers,
+        let tx_inputs = match account_hash {
+            Some(hash) if hash == Digest::default() => TxInputs {
+                account_hash: None,
+                nullifiers,
+            },
+            _ => TxInputs {
+                account_hash,
+                nullifiers,
+            },
```
bobbinth, line 192: Under what conditions would it be valid for account hash to come back as `None` from the store? Basically, if have cases here:

1. `Some(account_hash)` and `account_hash != Digest::default()` - this means the account is in the store.
2. `Some(account_hash)` and `account_hash == Digest::default()` - this means the account is new.
3. `None` - shouldn't this be an error?

phklive: Indeed it should be an error, the store should not return a None value.

phklive: This new change should fix it.

block-producer/src/store/mod.rs
```diff
@@ -181,9 +181,20 @@ impl Store for DefaultStore {
...
+            Some(hash) => TxInputs {
+                account_hash: Some(hash),
+                nullifiers,
+            },
+            None => {
+                return Err(TxInputsError::MalformedResponse(
+                    "incorrect account hash returned from the store. Got None.".to_string(),
+                ))
+            },
         };
```
bobbinth, line 198: small nit: I would probably write this as:
```Rust
let account_hash = match account_hash {
    Some(hash) if hash == Digest::default() => None,
    Some(hash) => Some(hash),
    None => {
        return Err(TxInputsError::MalformedResponse(
            "incorrect account hash returned from the store. Got None.".to_string(),
        ))
    },
};

let tx_inputs = TxInputs { account_hash, nullifiers };
```
This logically separates processing of account hash and construction of tx inputs, and makes the code a bit easier to follow.

I would also add a short comment about the assumptions we are making when parsing account hash (i.e., that we assume that default digest means the store does not have the account).

block-producer/src/state_view/mod.rs
```diff
@@ -178,6 +178,8 @@ fn ensure_tx_inputs_constraints(
     debug!(target: COMPONENT, %tx_inputs);
 
     match tx_inputs.account_hash {
+        // Having `Some(store_account_hash)` here signifies that we are operating
+        // on the state of an existing account.
         Some(store_account_hash) => {
```
bobbinth, line 183: I would probably make the comments more about "why" we are doing what we are doing. For example: "if the account is present in the store, make sure that the account state which went into the transaction is the same as the one in the store."

block-producer/src/state_view/mod.rs
```diff
@@ -186,11 +188,10 @@ fn ensure_tx_inputs_constraints(
                 });
             }
         },
+        // Having `None` here signifies that we are operating
+        // on the state of a new account
```
bobbinth, line 192: Similarly, here I would change the comment slightly to maybe something like this: "if the account is not in the store, it must be a new account".

bobbinth (approved):
Looks good! Thank you! I left a couple of small comments inline. Once these are addressed, we are good to merge.

phklive:
Added requested changes + documented more thoroughly.

bobbinth (approved):
Looks good! Thank you! I pushed a small commit which updates comment formatting (to make sure comments don't go beyond 100 chars per line).

Issue #125, opened 2023-12-24
TxQueue tests are flaky


bobbinth:
`test_build_batch_success` seems to fail sometimes in CI. The assert that triggers the failure is [here](https://github.com/0xPolygonMiden/miden-node/blob/59984a5d4969053c55edf148c978a30b0ada23e9/block-producer/src/txqueue/tests/mod.rs#L109) (i.e., a batch with no transactions is sent to the batch builder).

plafer:
This test is really dependent on proper timing of things (e.g. I expect that after X ms I will only have sent 1 batch, etc etc). There's probably a better way to write it. This *might* involve rewriting the [interval tick logic](https://github.com/0xPolygonMiden/miden-node/blob/59984a5d4969053c55edf148c978a30b0ada23e9/block-producer/src/txqueue/mod.rs#L107) so that tests can control when ticks happen, independent of timing.

hackaugusto:
IMO we should split the driver and the code. This is not always possible, but here it seems simple enough to have the `run` loops moved all the way to the top server. The unit tests should never have to spawn threads or have synchronization to work correctly (as in latches or sleeps). We should also have some way of moving time forward backwards in tests, meaning the code should not inspect the system's time directly.

PR #201, opened 2024-01-30
block_builder: fix flaky tests (fixes: #125)

hackaugusto:
The test `test_build_batch_success` was flaky on the CI because of lack of coordination among the test and the code. This commit fixes the issue by using tokio's paused time loop, with explicity time advances, and a channel to synchronize the batch builder task with the test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
         with:
           command: doc
           args: ${{matrix.args}}
-  
+
   test:
     name: test ${{matrix.toolchain}} on ${{matrix.os}} with ${{matrix.args}}
     runs-on: ${{matrix.os}}-latest
@@ -102,3 +102,12 @@ jobs:
         with:
           command: test
           args: ${{matrix.args}}
+      - name: Test repeat
+        shell: bash
+        run: |
+          for i in $(seq 100); do
+            cargo test --all-features || {
+              echo "FAILURE";
+              break
+            };
+          done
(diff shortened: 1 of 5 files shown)


hackaugusto:
I'm not 100% sure if this fixes the issue, since I can't reproduce the error locally, even after 100 runs with:

```shell
for i in $(seq 100); do cargo test --all-features -- test_build_batch_success test_tx_verify_failure || { echo "FAILURE"; break}; done
```

So opening this as draft to do a few CI runs.

hackaugusto:
from [GH's docs](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits):

> Job execution time - Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete.

The base run takes about 2m15s to run, so 100 runs should take about 3h45min. I think the job wont be killed.

hackaugusto:
The current main has about a 30% failure rate: https://github.com/0xPolygonMiden/miden-node/pull/202

hackaugusto:
huh, seems like GH's docs is not right, the job was killed:

> Error: The operation was canceled.

At about the 27m54s mark.

There was no failure. The initial test took 2m15s, so it should have been about 12 runs, with zero failures. If there is flakiness it is below 9%, which is better than the previous 30%. So I'm calling this a win. Opening for review.

hackaugusto:
huh, I have never seen this before, the last push is stuck:

bobbinth:
> At about the 27m54s mark.

I think the test job is configured to time out after 30 mins. Maybe that's the reason?

> huh, I have never seen this before, the last push is stuck:

Happened to me once last week. I did the same thing (closed the PR and opened a new one).

PR #202, opened 2024-01-30
Test ci

hackaugusto:


hackaugusto:
opened this as a test to run the flaky test a 100 times, to compare against https://github.com/0xPolygonMiden/miden-node/pull/201 and see if that makes the ci less flaky

hackaugusto:
I forgot to add the number of iterations, but this failed after 5min (the CI mark is green because I forgot to forward the error code in the shell script, the script failed successfully).

Given it took 5m19s to fail, and the previous run took 2m14s, we can assume it was on the 3rd run. so that is about a 30% failure rate.

Issue #74, opened 2023-11-27
Decide where to define the "domain types" corresponding to the proto types


plafer:
See [discussion](https://github.com/0xPolygonMiden/miden-node/pull/55#discussion_r1385684223) and [comment](https://github.com/0xPolygonMiden/miden-node/pull/55#discussion_r1405368470)

plafer:
## Structure
I suggest the following structure, based on how the ibc-rs repository handles it:

- `proto` crate: this crate contains the `.proto` files, as well as the generated versions. It does *not* contain conversions.
  - this should probably live in `miden-base`, or in a repository of its own
  - this is analogous to [ibc-proto-rs](https://github.com/cosmos/ibc-proto-rs/tree/main)
- Any type that has an analogous protobuf type defines the conversion to-from the protobuf type next to its definition

## ibc-rs example
In the ibc-rs ecosystem, we'll take the `ChannelEnd` type as an example. 
- The protobuf definition lives [here](https://github.com/cosmos/ibc-proto-rs/blob/9c8be3df748e9b7f9509e561d30d4ed5a3fc47e3/src/prost/ibc.core.channel.v1.rs#L11-L28) (this is the generated version, the protobuf definition actually lives in another repository, but we don't need to do that). We call this the "raw" type
- The "domain" type lives [here](https://github.com/cosmos/ibc-rs/blob/de2c530f941eaf993badce2fccc4f3cb352f95b8/ibc-core/ics04-channel/types/src/channel.rs#L105-L111), in its separate crate, which has the `proto` crate as a dependency. The conversion code lives with the domain type [here](https://github.com/cosmos/ibc-rs/blob/de2c530f941eaf993badce2fccc4f3cb352f95b8/ibc-core/ics04-channel/types/src/channel.rs#L125-L166)

## Benefits of proposed structure
I prefer this system over ours because
- The conversion code lives next to the domain type; everything is in one place. Contrast with our [conversion.rs](https://github.com/0xPolygonMiden/miden-node/blob/f4eefddb4d8e4c12ce8659f0a6ddc084482a269d/proto/src/conversion.rs) module which contains every and all conversions of all types. This makes it much harder to understand how each protobuf type relates to other "domain types" in our system.
- You can implement `TryFrom<Raw> for Domain` where the `Error` type lives in the domain crate
[...]

PR #207, opened 2024-02-02
Improved errors in the node

polydez:
These changes are intended to improve the way we work with errors in the node. We should use error enums in node internals with help of `thiserror`. For RPC API and  CLIs we continue using of `anyhow` for convenience.


proto/src/error.rs
```diff
@@ -1,51 +1,24 @@
 use miden_crypto::merkle::MmrError;
+use thiserror::Error;
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Error, Debug, PartialEq)]
 pub enum ParseError {
```
bobbinth, line 5: Question: do we not need to implement `Clone` if we implement `Error`?

store/src/db/errors.rs
bobbinth, line 1: Let's move all error types from here and also from other submodules into a single module `store/src/errors.rs`. This is how we usually do it in other crates - but for some reason didn't follow this convention here.

store/src/db/errors.rs
```diff
@@ -3,28 +3,89 @@ use std::io;
...
+    GetAccountsTaskFailed(String),
+    #[error("Get block header task failed: {0}")]
+    GetBlockHeaderTaskFailed(String),
+    #[error("Get block headers task failed: {0}")]
+    GetBlockHeadersTaskFailed(String),
+    #[error("Get account hashes task failed: {0}")]
+    GetAccountHashesTaskFailed(String),
+    #[error("Apply block task failed: {0}")]
+    ApplyBlockTaskFailed(String),
+}
```
bobbinth, line 69: I'm not sure I get the naming for this error and its variants. If I'm understanding correctly, this error happens when we are trying to query the database, right? If so, should we just call it `QueryError`?

As for variants, should we make them consistent with function names? For example: `SelectAccountsError` instead of `GetAccountsTaskFailed`.

store/src/errors.rs
```diff
@@ -1,64 +1,63 @@
...
+};
 use miden_node_proto::error::ParseError;
+use thiserror::Error;
+use tokio::sync::oneshot::error::RecvError;
 
-#[derive(Clone, Debug)]
+use crate::db::errors::DbError;
+
+#[derive(Error, Debug)]
 pub enum StateError {
```
bobbinth, line 12: There are quite a few error variants here. I would probably try to break this up into some logical error groups. For example, we could have `ApplyBlockError` and then State::apply_block()` would return this error.

Also, a lot of variants seem to be related to the parsing of protobuf messages (e.g., `MissingAccountId`). In the future, hopefully most of them will be encapsulated in the `proto` crate (see #74), but maybe for now we just create one or two variant for errors like this (e.g., `MessageMissingField(String)`).

block-producer/src/txqueue/mod.rs
```diff
@@ -21,16 +19,19 @@ mod tests;
 // TRANSACTION VERIFIER
 // ================================================================================================
 
-#[derive(Debug, PartialEq)]
+#[derive(Error, Debug, PartialEq)]
 pub enum VerifyTxError {
```
bobbinth, line 23: Let's move this, and other error types, into `src/errors.rs` module.

bobbinth (commented):
Thank you! Looks good! This is not a full review - but I wanted to leave a few preliminary comments. The main thing I think is to come up with a better way of grouping errors and splitting them up into smaller, more targeted types.

proto/src/error.rs
polydez: We need. I missed it, will revert back.

store/src/db/errors.rs
polydez: @bobbinth 
> I'm not sure I get the naming for this error and its variants. If I'm understanding correctly, this error happens when we are trying to query the database, right? If so, should we just call it QueryError?

These errors are used for wrapping errors raised by deadpool's `interact` method. The only two cases of such errors are panic and abortion of user's callback closure.

> As for variants, should we make them consistent with function names? For example: SelectAccountsError instead of GetAccountsTaskFailed.

Done.

block-producer/src/errors.rs
bobbinth, line 1: Let's add section separators between different error types.

block-producer/src/errors.rs
```diff
@@ -0,0 +1,133 @@
...
+    transaction::{InputNotes, ProvenTransaction},
+    TransactionInputError,
+};
+use miden_vm::ExecutionError;
+use thiserror::Error;
+
+use crate::{
+    block_builder::prover::block_witness::CREATED_NOTES_TREE_INSERTION_DEPTH,
+    MAX_NUM_CREATED_NOTES_PER_BATCH,
+};
```
bobbinth, line 17: I would move `CREATED_NOTES_TREE_INSERTION_DEPTH` to `/src/lib.rs`.

block-producer/src/lib.rs
```diff
@@ -8,7 +8,8 @@ use tokio::sync::RwLock;
 pub mod test_utils;
 
 mod batch_builder;
-mod block_builder;
+pub(crate) mod block_builder;
```
bobbinth, line 11: After the previous comment, this change should not be necessary.

store/src/errors.rs
bobbinth, line 1: Let's add section separators between different error types.

store/src/errors.rs
```diff
@@ -1,64 +1,154 @@
...
+    #[error("Select account hashes task failed: {0}")]
+    SelectAccountHashesTaskFailed(String),
+    #[error("Get state sync task failed: {0}")]
+    GetStateSyncTaskFailed(String),
+    #[error("Apply block task failed: {0}")]
+    ApplyBlockTaskFailed(String),
+}
+
+#[derive(Debug, Error)]
+pub enum GenesisBlockError {
```
bobbinth, line 74: nit: I would rename this into just `GenesisError`.

Also, let's implement `Clone` an `PartialEq` here and for all other error types.

store/src/state.rs
```diff
@@ -176,7 +189,7 @@ impl From<NullifierStateForTransactionInput> for NullifierTransactionInputRecord
 impl State {
     /// Loads the state from the `db`.
     #[instrument(target = "miden-store", skip_all)]
-    pub async fn load(mut db: Db) -> Result<Self, anyhow::Error> {
+    pub async fn load(mut db: Db) -> Result<Self> {
```
bobbinth, line 192: I would probably create a special error type for this function. Something like: `StateInitializationError`.

store/src/state.rs
```diff
@@ -347,7 +363,9 @@ impl State {
                 .await
         });
 
-        acquired_allowed.await?;
+        acquired_allowed
+            .await
+            .map_err(ApplyBlockError::BlockApplyingBrokenBecauseOfClosedChannel)?;
```
bobbinth, line 368: nit: I'd rename this variant to `ApplyBlockFailedClosedChannel`.

store/src/state.rs
```diff
@@ -411,7 +429,7 @@ impl State {
         account_ids: &[AccountId],
         note_tag_prefixes: &[u32],
         nullifier_prefixes: &[u32],
-    ) -> Result<(StateSyncUpdate, MmrDelta), anyhow::Error> {
+    ) -> Result<(StateSyncUpdate, MmrDelta)> {
```
bobbinth, line 432: I would probably create a separate error type for this function. Something like `StateSyncError`.

store/src/db/errors.rs
bobbinth: > These errors are used for wrapping errors raised by deadpool's `interact` method. The only two cases of such errors are panic and abortion of user's callback closure.

What happens if there is error with the query itself (e.g., malformed SQL statement)? For example:
```Rust
pub async fn select_accounts(&self) -> Result<Vec<AccountInfo>> {
    self.pool
        .get()
        .await?
        .interact(sql::select_accounts)
        .await
        .map_err(|err| InteractionTaskError::SelectAccountsTaskFailed(err.to_string()))?
}
```
What would happen if `select_accounts` returns an error? (does it get swallowed somewhere?)

bobbinth (changes requested):
Thank you! Looks good! Not a full review yet - but I did leave some questions/comments inline.

store/src/db/errors.rs
polydez: No-no, it doesn't get swallowed, it will be returned in `Result` of `select_accounts`. In `sql::select_accounts` this `rusqlite::Error` error will be wrapped with `DbError::SqliteError(err)` and passed through the parent `select_accounts`. In a case of `InteractionTaskError`, it will be also converted to `DbError::InteractionTaskError(err)` (look at question mark in the line with `map_err`) and passed through.

store/src/errors.rs
polydez: > Also, let's implement `Clone` an `PartialEq` here and for all other error types.

I think we can implement `PartialEq`, but will have difficulties with implementing `Clone`, because of underlying error types, which don't implement it. Do we really need `Clone` for errors?

polydez: I've checked, `PartialEq` also wasn't implemented in the most of underlying error types.

store/src/db/errors.rs
bobbinth: Ah - I see. So the value returned from `.interact().await` is `Result<Result<_, DbError>, InteractError>` - right?

polydez: @bobbinth right!

store/src/errors.rs
```diff
@@ -1,64 +1,198 @@
...
+    GenesisBlockError(#[from] GenesisError),
+    #[error("State error: {0}")]
+    StateError(Box<StateError>),
+    #[error("Conversion error: {0}")]
+    ConversionError(#[from] ConversionError),
+    #[error("Decoding nullifier from database failed: {0}")]
+    NullifierDecodingError(DeserializationError),
+    #[error("Block applying was broken because of closed channel on state side: {0}")]
+    ApplyBlockFailedClosedChannel(RecvError),
+}
```
bobbinth, line 85: Let's rename this error type to `DatabaseError`. Also, lets remove unused variants (e.g., I think `StateError` variant is not used).

In generally, would be great to make this error specifically about interacting with the database and reduce the number of variants (e.g., I don't think `GenesisBlockError` should be here, but that'll require a bigger refactoring). Let's create an issue for this.

store/src/errors.rs
```diff
@@ -1,64 +1,198 @@
...
+#[derive(Error, Debug)]
+pub enum ConversionError {
+    #[error("Parse error: {0}")]
+    ParseError(#[from] ParseError),
+    #[error("Field `{field_name}` required to be filled in protobuf representation of {entity}")]
+    MissingFieldInProtobufRepresentation {
+        entity: &'static str,
+        field_name: &'static str,
+    },
+}
```
bobbinth, line 27: This variant will probably go away after #74.

store/src/state.rs
```diff
@@ -596,28 +625,29 @@ async fn load_nullifier_tree(db: &mut Db) -> Result<TieredSmt> {
...
+async fn load_mmr(db: &mut Db) -> Result<Mmr, StateInitializationError> {
     let block_hashes: Result<Vec<RpoDigest>, ParseError> = db
         .select_block_headers()
         .await?
         .into_iter()
         .map(|b| b.try_into().map(|b: BlockHeader| b.hash()))
         .collect();
 
-    let mmr: Mmr = block_hashes?.into();
+    let mmr: Mmr = block_hashes.map_err(ConversionError::ParseError)?.into();
```
bobbinth, line 636: I think we should do something similar to what we did for `load_accounts` and `load_nullifier_tree` here. Specifically, use `StateInitializationError::FailedToCreateChainMmr` variant.

store/src/errors.rs
```diff
@@ -1,64 +1,198 @@
...
+    BlockDbIsEmpty,
+    #[error("Failed to get MMR delta: {0}")]
+    FailedToGetMmrDelta(MmrError),
+}
+
+// Block applying errors
+// =================================================================================================
+
+#[derive(Error, Debug)]
+pub enum ApplyBlockError {
```
bobbinth, line 161: Let's create an issue to revisit variants of this error. I think we can consolidate and normalize them a bit better than how they are now.

bobbinth (changes requested):
Thank you! Looks good! I left a few more comments inline - but overall, we can merge this pretty soon and create a few follow-up issues to address specific points in more targeted PRs.

store/src/errors.rs
polydez: @bobbinth I've extracted several variants into a new `DatabaseSetupError` class and slightly refactored `ensure_genesis_block` function.

store/src/errors.rs
polydez: @bobbinth thank you. I've created an issue: https://github.com/0xPolygonMiden/miden-node/issues/214
Please, help me to fill it by providing more context.

PR #208, opened 2024-02-02
domain: moved conversion code to domain crate.

hackaugusto:
Issue: #74

As requested in the issue above, quote:

> Move conversions between protobuf messages and native types into proto/src/domain module. Here, instead of having a single large module, we should split the code up into sub-modules - one per domain type. For example: proto/src/domain/account_id.rs, proto/src/domain/digest.rs, proto/src/domain/block_header.rs etc.


bobbinth (commented):
Looks good! Thank you! Since @hackaugusto is on vacation this week, @polydez will take over this and related PRs.

A few comments:

1. Instead of encapsulating each struct in a sub-module under the `domain` module, I think we should group them logically. For example, we could have `accounts`, `notes`, `block` etc. modules each containing relevant structs and conversions.
2. The modules should contain other things about domain objects as well - for example, implementations of `Display` traits would go here too.
3. We should also add domain types for responses (and maybe requests). For example, in the block producer we have [TxInputs](https://github.com/0xPolygonMiden/miden-node/blob/main/block-producer/src/store/mod.rs#L54) struct which is basically the same as `GetTransactionInputsResponse` protbuf message. Then, in the store we have [get_transaction_inputs()](https://github.com/0xPolygonMiden/miden-node/blob/main/store/src/state.rs#L494) which basically returns the same data but as a tuple. What we should do instead is to have a single `TxInputs` (or `TransactionInputs`) domain object and implement to/from conversions between it and `GetTransactionInputsResponse` struct. Then `get_transaction_inputs()` method in the store would return this domain object and then in the block producer, we would convert the response into `TxInputs` in a single line [here](https://github.com/0xPolygonMiden/miden-node/blob/main/block-producer/src/store/mod.rs#L136).
4. I wonder if we still need `proto/src/hex.rs` module.
5. Maybe for another PR, but from `proto/src/lib.rs` the main exports should be the domain object rather than generated objects (generated objects can still be exported, but probably under submodules).

proto/src/domain/nullifiers.rs
```diff
@@ -0,0 +1,41 @@
...
+};
+
+// NULLIFIER INPUT RECORD
+// ================================================================================================
+
+#[derive(Clone, Debug)]
+pub struct NullifierWitness {
+    pub nullifier: Digest,
+    pub proof: SmtProof,
+}
```
bobbinth, line 15: The type of `nullifier` should probably be `Nullifier` rather than `Digest` (and this is relevant to all other similar places).

proto/src/domain/blocks.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    /// MMR peaks for the current chain state
+    pub chain_peaks: MmrPeaks,
+
+    /// The hashes of the requested accounts and their authentication paths
+    pub account_states: Vec<AccountInputRecord>,
+
+    /// The requested nullifiers and their authentication paths
+    pub nullifiers: Vec<NullifierWitness>,
+}
```
bobbinth, line 97: Since this struct is used only in the block producer, I would move it there (probably into `src/block.rs` file).

Also, I would probably change the `account_states` and `nullifiers` into maps because this is what is happening in the block witness constructor anyway. So, the struct would look something like this:
```Rust
pub struct BlockInputs {
    pub block_header: BlockHeader,
    pub chain_peaks: MmrPeaks,
    pub accounts: BTreeMap<AccountId, AccountWitness>,
    pub nullifiers: BTreeMap<Digest, SmtProof>,
}

pub struct AccountWitness {
    pub hash: Digest,
    pub proof: MerklePath,
}
```

proto/src/domain/transactions.rs
```diff
@@ -0,0 +1,83 @@
...
+/// Information needed from the store to verify a transaction.
+#[derive(Debug)]
+pub struct TransactionInputs {
+    /// The account state in the store corresponding to tx's account ID
+    pub account_state: AccountState,
+
+    /// Maps each consumed notes' nullifier to block number, where the note is consumed
+    /// (`zero` means, that note isn't consumed yet)
+    pub nullifiers: BTreeMap<Digest, u32>,
+}
```
bobbinth, line 27: Looking at this now, I no longer think it make sense to have the same object shared between the store and the block producer (primarily because it forces us to build a `BTreeMap` in the store, when we really don't need to).

So, I'm thinking we should:
- Move this struct to the block producer crate.
- On the store side, either create a separate struct or build ` GetTransactionInputsResponse` directly.

I would also simplify `TransactionInputs` (and maybe the whole message) to look as follows:
```Rust
pub struct TransactionInputs {
    pub account_id: AccountId,
    pub account_hash: Option<Digest>,
    pub nullifiers: BTreeMap<Digest, u32>,
}
```

bobbinth (approved):
Looks good! Thank you! I am going to merge this PR as it is already quite big, but I left a few comment inline. We can address these comments in a separate PR.

Also, one big piece that is missing is still missing is storing domain objects (rather than protobuf messages) in the store. This will probably lead us to creating a few more domain objects on the store side.

PR #209, opened 2024-02-04
Block producer: compute new nullifiers root

### LINKED ISSUE #195: nullifier DB: Use new `Smt` (which replaced `TieredSmt`)
Precursor to #140. The node should compile and run using [the new `Smt`](https://github.com/0xPolygonMiden/crypto/pull/254)

plafer:
Closes #195
Closes: #210


plafer:
@bobbinth I still need to test sending transactions with `miden-client`, but that won't change the overall structure, so this is ready for review!

proto/proto/responses.proto
```diff
@@ -61,7 +62,7 @@ message AccountBlockInputRecord {
 // A nullifier returned as a response to the GetBlockInputs
 message NullifierBlockInputRecord {
     digest.Digest nullifier = 1;
-    merkle.MerklePath proof = 2;
+    smt.SmtOpening opening = 2;
```
bobbinth, line 65: Instead of creating a separate message for `SmtOpening` would it be simpler to just serialize it as bytes (using our `Serializabe`/`Deserializable` interfaces)?

plafer: I personally prefer using protobuf types instead of `bytes`; for readability, and extra type safety (attempting to deserialize into anything other than an `SmtProof` will cause a compile-time error).

What do you see as advantages of using `bytes` instead?

bobbinth: Mostly just relative simplicity (given that we already have tested serialization code), and potentially easier maintainability - e.g., if we switch from Merkle paths to compact Merkle path, the messages still remain the same. Though, this could be a double-edged sword.

Not a strong preference from my side though.

plafer: I will leave as is (for above-mentioned reasons), and if it becomes a pain, we can switch to `bytes`. But my fear is that we end up doing that in multiple `proto` files, up to a point where they barely convey any useful information (due to the excessive use of `bytes`).

block-producer/src/block_builder/prover/mod.rs
```diff
@@ -106,6 +109,38 @@ proc.compute_note_root
...
+        # Prepare stack for `smt::set`
+        movdn.12 movupw.2 dupw.2
+        #=> [NULLIFIER_VALUE, NULLIFIER_i, ROOT_i, NULLIFIER_VALUE, num_nullifiers_left_to_update, ... ]
+
+        exec.smt::set
+        #=> [OLD_VALUE, ROOT_{i+1}, NULLIFIER_VALUE, num_nullifiers_left_to_update, ... ]
+
+        # Check that OLD_VALUE == 0 (i.e. that nullifier was indeed not previously produced)
+        padw assert_eqw
+        #=> [ROOT_{i+1}, NULLIFIER_VALUE, num_nullifiers_left_to_update, ... ]
```
bobbinth, line 132: small optimization observation: here we just want to assert that top 4 elements of the stack are zeros, right? If so, doing `assertz assertz assertz assertz` would be a bit faster (8 cycles vs. 19).

block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -33,6 +33,8 @@ pub struct BlockWitness {
     pub(super) updated_accounts: BTreeMap<AccountId, AccountUpdate>,
     /// (batch_index, created_notes_root) for batches that contain notes
     pub(super) batch_created_notes_roots: BTreeMap<usize, Digest>,
+    pub(super) produced_nullifiers: BTreeMap<Digest, MerklePath>,
+    pub(super) nullifier_smt_leaves: Vec<SmtLeaf>,
```
bobbinth, line 37: Would it make sense to consolidate these into a single field? Since `StmProof` is basically leaf + path, maybe something like this would work:
```Rust
pub(super) produced_nullifiers: BTreeMap<Digest, SmtProof>,
```

bobbinth (commented):
Looks good! Thank you! I left a couple of small comments/questions inline.

Another comment: I would have probably had fewer protobuf messages and then there would be less code in the `conversions.rs` file - but as we agreed in the previous comment, I think the current approach is fine too.

block-producer/src/block_builder/prover/block_witness.rs
plafer: I had this initially, but then we would need to add an accessor `SmtProof::get_leaf() -> &SmtLeaf` (i.e. splitting in 2 variables is a result of only having `into_parts(self)`).

Should I open a PR on `miden-crypto` to address that?

block-producer/src/block_builder/prover/mod.rs
plafer: yup! Fixed

block-producer/src/block_builder/prover/block_witness.rs
plafer: Refactored!

block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -189,11 +130,12 @@ impl BlockWitness {
         }
 
         Self::validate_account_states(block_inputs, batches)?;
+        Self::validate_nullifiers(block_inputs, batches)?;
```
bobbinth, line 133: We should probably remove the `TODO` on line 125 above.

block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -246,6 +188,147 @@ impl BlockWitness {
...
+        StackInputs::new(stack_inputs)
+    }
+
+    /// Builds the advice inputs to the block kernel
+    fn build_advice_inputs(self) -> Result<AdviceInputs, BlockProverError> {
+        let advice_map: Vec<_> = self
+            .produced_nullifiers
+            .values()
+            .map(|proof| (proof.leaf().hash().as_bytes(), proof.leaf().to_elements()))
+            .collect();
```
bobbinth, line 298: Adding nullifier-related info the the advice inputs happens here and also down starting at line 317. I wonder if we should encapsulate it better. Maybe do something similar to what we do with the chain MMR? Specifically, we could have a separate small helper function - something like `add_nullifiers_to_advice_inputs()` and put all relevant logic there.

bobbinth (approved):
Looks good! Thank you! I left a couple more small comments inline. After these (and after we confirm that the end-to-end process is working), we can merge.

block-producer/src/block_builder/prover/block_witness.rs
plafer: done

block-producer/src/block_builder/prover/block_witness.rs
plafer: went with a slightly different approach. `add_nullifiers_to_advice_inputs()` would not have worked as well because it needs to modify the `MerkleStore`, and `AdviceInputs::extend_merkle_store()` takes an iterator of `InnerNodeInfo` (which would be a pain to generate). I also didn't like that we would build the `AdviceInputs`, and then mutate them once again (I would even remove that functionality from `AdviceInputs`'s API - but that's another discussion).

This new approach instead combines the building of the merkle store, and the building of the advice map; as opposed to grouping nullifiers and accounts together, respectively. I find that even better actually - when investigating problems with the store (or advice map), you only have to look in one place.

plafer:
@bobbinth switched a nullifier's value to be `[block_num, 0, 0, 0]` (which is the right way to write it for `block_num` to be the least significant element of the word).

I also confirmed that everything works as expected with the `miden-client`. Specifically, running `miden-client tx new consume-notes ...` updates the `nullifiers` table in the db.

I think this is ready to merge! 🎉

store/src/state.rs
```diff
@@ -583,9 +583,9 @@ impl State {
 // UTILITIES
 // ================================================================================================
 
-/// Returns the nullifier's block number given its leaf value in the TSMT.
+/// Returns the nullifier's block number given its leaf value in the SMT.
 fn block_to_nullifier_data(block: BlockNumber) -> Word {
-    [Felt::new(block as u64), Felt::ZERO, Felt::ZERO, Felt::ZERO]
+    [Felt::ZERO, Felt::ZERO, Felt::ZERO, Felt::new(block as u64)]
```
bobbinth: nit: I would use `Felt::from(block)` to avoid the cast.

store/src/state.rs
```diff
@@ -585,7 +585,7 @@ impl State {
 
 /// Returns the nullifier's block number given its leaf value in the SMT.
 fn block_to_nullifier_data(block: BlockNumber) -> Word {
-    [Felt::ZERO, Felt::ZERO, Felt::ZERO, Felt::new(block as u64)]
+    [Felt::new(block as u64), Felt::ZERO, Felt::ZERO, Felt::ZERO]
```
bobbinth, line 588: Same comment as the one above: conversion from `u32` would probably be a bit cleaner.

bobbinth (approved):
Looks good! Thank you! I left a couple of very small nits inline. After these, let's merge.

bobbinth:
@igamigo - now that this PR has been merged, the nullifier database should be updated when notes are consumed. Could you verify that things are working on the client side as expected?

Issue #205, opened 2024-02-01
Change default config file path to `miden-node.toml`


bobbinth:
Currently, the default config file is `miden.toml`. This conflicts with `miden.toml` that miden client also uses as a default config file (see https://github.com/0xPolygonMiden/miden-client/issues/133). I think we should change the default to `miden-node.toml`.

phklive:
What do you think of: 
- Client => `client.toml`
- Node => `node.toml`

It's the main configuration file hence making sense to have the same name as the crate. The redundant use of `miden`  doesn't really add value here.

PR #212, opened 2024-02-06
Changed config filename from `miden.toml` to `node.toml`

phklive:
I had previously mistakenly pushed this PR to main, I have reverted it promptly and opened this PR in a new branch here.

@bobbinth should we include branch protection rules on our repos to prevent push to main or other errors related to branching from happening in the future?

This PR closes: #205


phklive:
In this PR I have change the naming from `miden.toml` to `node.toml` if requested I can change it to `miden-node.toml`, I have added my rational here: https://github.com/0xPolygonMiden/miden-node/issues/205#issuecomment-1929333596

I believe that we should have a naming convention accross repos for config files.

polydez (approved):
Looks good to me! :)

bobbinth:
Not a strong preference, but I do like a slightly more explicit name better (i.e., `miden-node.toml`). Curious what @igamigo and @plafer think as well.

igamigo:
I can see some benefit in removing it, but I too tend to prefer keeping the `miden-` prefix for less ambiguity and more consistency (even if marginal), since binary names are currently `miden-node` and `miden-client`. It can also help in searching and identifying files more easily from the outside.

PR #217, opened 2024-02-08
Fully remove `TieredSmt`

### LINKED ISSUE #216: Remove `tsmt.proto`
#209 deprecated the use of TSMT for nullifiers (in favor of `Smt`).

We should remove `proto/proto/tsmt.proto` (and fix the corresponding endpoints that use it, which I believe are meant for testing). I think this requires a similar change in the client, but I'm not sure.

plafer:
Closes #216


bobbinth (approved):
Looks good! Thank you!

PR #219, opened 2024-02-09
Get nullifier prefix from the most significant element

### LINKED ISSUE #218: Fix nullifer prefixes
In order for clients to receive information about spent nullifiers, they request nullifier data by their prefix.
This prefix comes from reducing the nullifiers of notes to the higher 16 bits, but currently this is being done incorrectly on the node: https://github.com/0xPolygonMiden/miden-node/blob/main/store/src/db/sql.rs#L50. We would want to index the word by their last element and not the first since it's little-endian.

bobbinth:
This PR fixes #218.

### DIFF
--- a/store/src/db/sql.rs
+++ b/store/src/db/sql.rs
@@ -3,7 +3,8 @@ use std::rc::Rc;
 
 use miden_crypto::{
     hash::rpo::RpoDigest,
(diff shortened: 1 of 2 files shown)


igamigo (approved):
LGTM! EDIT: Although it seems one of the modified tests now fails because it matches multiple nullifiers.

PR #220, opened 2024-02-12
Add tx proof verification in `RPC` and `Block Producer`

phklive:
In this PR I propose the addition of the verification of tx proofs on the `RPC` and `Block Producer` of the Miden node.

This will make sure that the locally proven tx's have a valid proof of execution.

Closing: #106 

Would be better to merge after: https://github.com/0xPolygonMiden/miden-base/pull/460

### DIFF
--- a/block-producer/src/config.rs
+++ b/block-producer/src/config.rs
@@ -16,11 +16,12 @@ pub struct BlockProducerConfig {
     /// Store gRPC endpoint in the format `http://<host>[:<port>]`.
     pub store_url: String,
 
-    /// Enables or Disables the verification of proofs in the Block Producer
-    /// We could want to disable proof verification for the following reasons:
-    /// - Proofs take non-negligible time to verify
-    /// - Proof verification is already made at the RPC level
-    /// - Batch Builder will verify tx proofs before batch proof generation
+    /// Enable or disable the verification of transaction proofs before they are accepted into the
+    /// transaction queue.
+    ///
+    /// Disabling transaction proof verification will speed up transaction processing as proof
+    /// verification may take ~15ms/proof. This is OK when all transactions are forwarded to the
+    /// block producer from the RPC component as transaction proofs are also verified there.
     pub verify_tx_proofs: bool,
 }
 
--- a/block-producer/src/config.rs
+++ b/block-producer/src/config.rs
@@ -15,6 +15,8 @@ pub struct BlockProducerConfig {
 
     /// Store gRPC endpoint in the format `http://<host>[:<port>]`.
     pub store_url: String,
+
+    pub check_tx_proofs: bool,
 }
 
 impl BlockProducerConfig {
@@ -62,6 +64,7 @@ mod tests {
                 r#"
                     [block_producer]
                     store_url = "http://store:8000"
+                    check_tx_proofs = true
 
                     [block_producer.endpoint]
                     host = "127.0.0.1"
@@ -81,6 +84,7 @@ mod tests {
                             port: 8080,
                         },
                         store_url: "http://store:8000".to_string(),
+                        check_tx_proofs: true
                     }
                 }
             );
--- a/block-producer/src/config.rs
+++ b/block-producer/src/config.rs
@@ -16,6 +16,11 @@ pub struct BlockProducerConfig {
     /// Store gRPC endpoint in the format `http://<host>[:<port>]`.
     pub store_url: String,
 
+    /// Enables or Disables the verification of proofs in the Block Producer
+    /// We could want to disable proof verification for the following reasons:
+    /// - Proofs take non-negligible time to verify
+    /// - Proof verification is already made at the RPC level
+    /// - Batch Builder will verify tx proofs before batch proof generation
     pub verify_tx_proofs: bool,
 }
 
--- a/block-producer/src/config.rs
+++ b/block-producer/src/config.rs
@@ -16,7 +16,7 @@ pub struct BlockProducerConfig {
     /// Store gRPC endpoint in the format `http://<host>[:<port>]`.
     pub store_url: String,
 
-    pub check_tx_proofs: bool,
+    pub verify_tx_proofs: bool,
 }
 
 impl BlockProducerConfig {
@@ -64,7 +64,7 @@ mod tests {
                 r#"
                     [block_producer]
                     store_url = "http://store:8000"
-                    check_tx_proofs = true
+                    verify_tx_proofs = true
 
                     [block_producer.endpoint]
                     host = "127.0.0.1"
@@ -84,7 +84,7 @@ mod tests {
                             port: 8080,
                         },
                         store_url: "http://store:8000".to_string(),
-                        check_tx_proofs: true
+                        verify_tx_proofs: true
                     }
                 }
             );
--- a/block-producer/src/errors.rs
+++ b/block-producer/src/errors.rs
@@ -44,8 +44,7 @@ pub enum VerifyTxError {
     TransactionInputError(#[from] TransactionInputError),
 
     /// Failed to verify the transaction execution proof
-    /// TODO: Add transaction id in error message
-    #[error("Invalid transaction proof error")]
+    #[error("Invalid transaction proof error for transaction: {0}")]
     InvalidTransactionProof(TransactionId),
 }
 
--- a/block-producer/src/errors.rs
+++ b/block-producer/src/errors.rs
@@ -4,7 +4,7 @@ use miden_objects::{
     accounts::AccountId,
     crypto::merkle::MerkleError,
     notes::Nullifier,
-    transaction::{InputNotes, ProvenTransaction},
+    transaction::{InputNotes, ProvenTransaction, TransactionId},
     Digest, TransactionInputError,
 };
 use miden_vm::ExecutionError;
@@ -42,6 +42,11 @@ pub enum VerifyTxError {
 
     #[error("Transaction input error: {0}")]
     TransactionInputError(#[from] TransactionInputError),
+
+    /// Failed to verify the transaction execution proof
+    /// TODO: Add transaction id in error message
+    #[error("Invalid transaction proof error")]
+    InvalidTransactionProof(TransactionId),
 }
 
 // Transaction adding errors
--- a/block-producer/src/state_view/mod.rs
+++ b/block-producer/src/state_view/mod.rs
@@ -3,10 +3,13 @@ use std::{collections::BTreeSet, sync::Arc};
 use async_trait::async_trait;
 use miden_node_proto::TransactionInputs;
 use miden_node_utils::formatting::format_array;
-use miden_objects::{accounts::AccountId, notes::Nullifier, transaction::InputNotes, Digest};
+use miden_objects::{
+    accounts::AccountId, notes::Nullifier, transaction::InputNotes, Digest,
+    MIN_PROOF_SECURITY_LEVEL,
+};
+use miden_tx::TransactionVerifier;
 use tokio::sync::RwLock;
 use tracing::{debug, instrument};
-use {miden_objects::MIN_PROOF_SECURITY_LEVEL, miden_tx::TransactionVerifier};
 
 use crate::{
     block::Block,
--- a/block-producer/src/state_view/mod.rs
+++ b/block-producer/src/state_view/mod.rs
@@ -4,12 +4,10 @@ use async_trait::async_trait;
 use miden_node_proto::TransactionInputs;
 use miden_node_utils::formatting::format_array;
 use miden_objects::{accounts::AccountId, notes::Nullifier, transaction::InputNotes, Digest};
-
-#[cfg(not(test))]
-use {miden_objects::MIN_PROOF_SECURITY_LEVEL, miden_tx::TransactionVerifier};
-
 use tokio::sync::RwLock;
 use tracing::{debug, instrument};
+#[cfg(not(test))]
+use {miden_objects::MIN_PROOF_SECURITY_LEVEL, miden_tx::TransactionVerifier};
 
 use crate::{
     block::Block,
--- a/block-producer/src/state_view/mod.rs
+++ b/block-producer/src/state_view/mod.rs
@@ -6,7 +6,6 @@ use miden_node_utils::formatting::format_array;
 use miden_objects::{accounts::AccountId, notes::Nullifier, transaction::InputNotes, Digest};
 use tokio::sync::RwLock;
 use tracing::{debug, instrument};
-#[cfg(not(test))]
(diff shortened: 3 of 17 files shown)


phklive:
@bobbinth do we have constants in other places than mock?

It would be useful to be able to import constant values from miden-base to use in other repos like this one: https://github.com/0xPolygonMiden/miden-base/blob/bbb11913810e7506ede3cdfe2f9f8155705877bc/mock/src/constants.rs#L32

I would use it here: 
https://github.com/0xPolygonMiden/miden-node/pull/220/files#diff-e5a2367dc1bde32c05ec20d7a483b14afdba6c744bd048ad9db9391169e3c4eeR68

hackaugusto:
> It would be useful to be able to import constant values from miden-base to use in other repos like this one: https://github.com/0xPolygonMiden/miden-base/blob/bbb11913810e7506ede3cdfe2f9f8155705877bc/mock/src/constants.rs#L32

+1 to reusing the constant. (I grepped the codebase, we only have that value defined in the mock, perhaps add it along side the `TransactionVerifier`?)

block-producer/src/state_view/mod.rs
```diff
@@ -54,7 +56,22 @@ where
...
+            //
+            // This check makes sure that the transaction proof that has been attached to the transaction
+            // is valid, errors out if the proof is invalid.
+            let tx_verifier = TransactionVerifier::new(96);
+            tx_verifier
+                .verify(candidate_tx.clone())
+                .map_err(|_| VerifyTxError::InvalidTransactionProof(candidate_tx.id()))?;
+        }
+
+        // 2. soft-check if `tx` violates in-flight requirements.
```
hackaugusto, line 74: Suggestion: Remove the counter. We don't need to know the step count (doesn't help with understanding things). And having them is just extra work (this clutters the git history, and can easily get out of sync if someone doesn't notice they should be changed).

bobbinth:
> > It would be useful to be able to import constant values from miden-base to use in other repos like this one: https://github.com/0xPolygonMiden/miden-base/blob/bbb11913810e7506ede3cdfe2f9f8155705877bc/mock/src/constants.rs#L32
> 
> +1 to reusing the constant. (I grepped the codebase, we only have that value defined in the mock, perhaps add it along side the `TransactionVerifier`?)

Yes - let's define it in `miden-bsae`. I would probably put it in [this file](https://github.com/0xPolygonMiden/miden-base/blob/main/objects/src/lib.rs#L56).

block-producer/src/state_view/mod.rs
```diff
@@ -54,7 +56,22 @@ where
         &self,
         candidate_tx: &ProvenTransaction,
     ) -> Result<(), VerifyTxError> {
-        // 1. soft-check if `tx` violates in-flight requirements.
+        #[cfg(not(test))]
```
hackaugusto, line 59: I'm not sure which test is failing, but this probably means this class should be generic over the `TransactionVerifier` implementation. Then:

- The code bellow would be a `DefaultVerifier` of sorts, that runs the actual verify code.
- The `DummyProver` would have two verifier counter parts. Possible a `SuccessfulDummyVerifier` and a `FailureDummyVerifier`. Then the tests would use this implementation.

The changes above would allow one to test the code path were a transaction verification fails. And would save us the time to generate an actual proof in the tests.

rpc/src/server/api.rs
```diff
@@ -117,6 +118,20 @@ impl api_server::Api for RpcApi {
...
+        let tx = ProvenTransaction::read_from_bytes(&request.transaction)
+            .map_err(|_| Status::invalid_argument("Invalid transaction"))?;
+
+        let tx_verifier = TransactionVerifier::new(96);
+
+        tx_verifier.verify(tx.clone()).map_err(|_| {
+            Status::invalid_argument(format!(
+                "Invalid transaction proof for transaction: {}",
+                tx.id().to_hex()
+            ))
```
bobbinth, line 132: with the latest changes in `miden-base`, we can probably get rid of `to_hex()` part.

block-producer/src/errors.rs
```diff
@@ -42,6 +42,11 @@ pub enum VerifyTxError {
 
     #[error("Transaction input error: {0}")]
     TransactionInputError(#[from] TransactionInputError),
+
+    /// Failed to verify the transaction execution proof
+    /// TODO: Add transaction id in error message
+    #[error("Invalid transaction proof error")]
+    InvalidTransactionProof(TransactionId),
```
bobbinth, line 49: With the latest changes in `miden-base`, we can probably fix this TODO.

block-producer/src/state_view/mod.rs
```diff
@@ -54,7 +56,22 @@ where
...
+        // TODO: Add valid proof generation in tests
+        {
+            // 1. Verify transaction proof
+            //
+            // This check makes sure that the transaction proof that has been attached to the transaction
+            // is valid, errors out if the proof is invalid.
+            let tx_verifier = TransactionVerifier::new(96);
+            tx_verifier
+                .verify(candidate_tx.clone())
+                .map_err(|_| VerifyTxError::InvalidTransactionProof(candidate_tx.id()))?;
```
bobbinth, line 71: As mentioned in the issue, let's add a config setting - e.g., something like `check_tx_proofs`. For now, it could be hard-coded to `true` and we can expose it via `miden-node.toml` as a separate issue.

Also, let's define a constant for `96` in `miden-base` as discussed in the other comment.

block-producer/src/state_view/mod.rs
bobbinth: Yes, agree that we should handle it via traits. Let's do this in a different PR though.

bobbinth (changes requested):
Thank you! Looks good. I left a few comments inline.

block-producer/src/state_view/mod.rs
phklive: I have not clearly understood what you were referring to when saying `a config setting`.

Something like so: 
```rust
/// Block producer specific configuration
#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
pub struct BlockProducerConfig {
    pub endpoint: Endpoint,

    /// Store gRPC endpoint in the format `http://<host>[:<port>]`.
    pub store_url: String,

    pub check_tx_proofs: bool
}
```

Also I don't fully understand when we wouldn't want to not verify proofs ( except in tests ).

phklive:
> > > It would be useful to be able to import constant values from miden-base to use in other repos like this one: https://github.com/0xPolygonMiden/miden-base/blob/bbb11913810e7506ede3cdfe2f9f8155705877bc/mock/src/constants.rs#L32
> > 
> > 
> > +1 to reusing the constant. (I grepped the codebase, we only have that value defined in the mock, perhaps add it along side the `TransactionVerifier`?)
> 
> Yes - let's define it in `miden-bsae`. I would probably put it in [this file](https://github.com/0xPolygonMiden/miden-base/blob/main/objects/src/lib.rs#L56).

Done in this PR: https://github.com/0xPolygonMiden/miden-base/pull/465

block-producer/src/state_view/mod.rs
hackaugusto: Here is the issue: https://github.com/0xPolygonMiden/miden-node/issues/229

block-producer/src/state_view/mod.rs
bobbinth: > Something like so:
> 
> ```rust
> /// Block producer specific configuration
> #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
> pub struct BlockProducerConfig {
>     pub endpoint: Endpoint,
> 
>     /// Store gRPC endpoint in the format `http://<host>[:<port>]`.
>     pub store_url: String,
> 
>     pub check_tx_proofs: bool
> }
> ```

Yes, this is correct.

> Also I don't fully understand when we wouldn't want to not verify proofs ( except in tests ).

I briefly mentioned this in https://github.com/0xPolygonMiden/miden-node/issues/106#issuecomment-1902598703, but to elaborate more:
- Proofs take non-negligible time to verify (10 - 15ms).
- We already verify proofs in the PRC - so, by the time they get to the block producer they should be already verified.
- Block producer will also verify them later when building a batch (as a part of batch proof generation).

phklive: Understood, hence the `check_tx_proof` configuration would enable/disable the `verify_tx` proof verification. 

But the batch proof verification would always be done.

Right?

phklive:
In the latest two commits: 
- Added the `verify_tx_proofs` configuration argument in the `toml` and `rust` files, adding a new config enabling us to enable or disable tx proof verification in `verify_tx`.
- This also solves the situation of tests failing because of generating `DummyProvenTx` with fake tx proofs, we can pass `false` as a config to the `DefaultStateView` hence disabling the tx proof verification.
- Changed from `check_tx_proofs` to `verify_tx_proofs`, I feel that it makes it clearer.

block-producer/src/config.rs
```diff
@@ -15,6 +15,8 @@ pub struct BlockProducerConfig {
 
     /// Store gRPC endpoint in the format `http://<host>[:<port>]`.
     pub store_url: String,
+
+    pub verify_tx_proofs: bool,
```
bobbinth, line 19: Let's add a doc comment here. I would also add a brief description of the motivation for this setting.

node/miden-node.toml
```diff
@@ -4,6 +4,7 @@
 # port defined as: sum(ord(c)**p for (p, c) in enumerate('miden-block-producer', 1)) % 2**16
 endpoint = { host = "localhost", port = 48046 }
 store_url = "http://localhost:28943"
+verify_tx_proofs = true
```
bobbinth, line 7: Let's add a comment here to explain what `verify_tx_proofs` does. This should be a user-facing comment.

block-producer/src/txqueue/mod.rs
```diff
@@ -22,14 +22,14 @@ mod tests;
...
     /// Method to receive a `tx` for processing.
     ///
     /// This method should:
     ///
-    /// 1. Verify the transaction is valid, against the current's rollup state, and also against
+    /// Verify the transaction is valid, against the current's rollup state, and also against
     ///    in-flight transactions.
-    /// 2. Track the necessary state of the transaction until it is commited to the `store`, to
+    /// Track the necessary state of the transaction until it is commited to the `store`, to
     ///    perform the check above.
```
bobbinth, line 33: We should still keep this as bullet points (even if as un-ordered list). Also, let's make sure to update the formatting.

block-producer/src/state_view/mod.rs
```diff
@@ -136,9 +157,9 @@ where
 // -------------------------------------------------------------------------------------------------
 
 /// Ensures the constraints related to in-flight transactions:
-/// 1. the candidate transaction doesn't modify the same account as an existing in-flight
+/// - the candidate transaction doesn't modify the same account as an existing in-flight
 ///    transaction (issue: #186)
-/// 2. no consumed note's nullifier in candidate tx's consumed notes is already contained in
+/// - no consumed note's nullifier in candidate tx's consumed notes is already contained in
 ///    `already_consumed_nullifiers`
```
bobbinth, line 163: minor nit: the formatting should be updated a bit, to look like so:
```
/// Ensures the constraints related to in-flight transactions:
/// - the candidate transaction doesn't modify the same account as an existing in-flight
///   transaction (issue: #186)
/// - no consumed note's nullifier in candidate tx's consumed notes is already contained in
///   `already_consumed_nullifiers`
```

block-producer/src/state_view/mod.rs
```diff
@@ -33,28 +40,42 @@ impl<S> DefaultStateView<S>
 where
     S: Store,
 {
-    pub fn new(store: Arc<S>) -> Self {
+    pub fn new(
+        store: Arc<S>,
+        verify_tx_proofs: bool,
+    ) -> Self {
```
bobbinth, line 46: This is probably fine for now - but I wonder if we should pass a config struct into this (either new or existing).

bobbinth (changes requested):
Thank you! Looks good! I left a few more comments inline.

block-producer/src/txqueue/mod.rs
phklive: What do you imply by "update the formatting"?

block-producer/src/state_view/mod.rs
phklive: I could add a struct as follows: 

```rust
pub struct DefaultStateViewConfig<S> {
    store: Arc<S>,
    verify_tx_proofs: bool,
}

pub struct DefaultStateView<S> {
    store: Arc<S>,

    /// Enables or disables the verification of transaction proofs in `verify_tx`
    verify_tx_proofs: bool,

    /// The account ID of accounts being modified by transactions currently in the block production
    /// pipeline. We currently ensure that only 1 tx/block modifies any given account (issue: #186).
    accounts_in_flight: Arc<RwLock<BTreeSet<AccountId>>>,

    /// The nullifiers of notes consumed by transactions currently in the block production pipeline.
    nullifiers_in_flight: Arc<RwLock<BTreeSet<Digest>>>,
}

impl<S> DefaultStateView<S>
where
    S: Store,
{
    pub fn new(config: DefaultStateViewConfig<S>) -> Self {
        Self {
            store: config.store,
            verify_tx_proofs: config.verify_tx_proofs,
            accounts_in_flight: Arc::new(RwLock::new(BTreeSet::new())),
            nullifiers_in_flight: Arc::new(RwLock::new(BTreeSet::new())),
        }
    }
}
```

But I don't fully understand why we would want to do so, I might not have understood your request.

hackaugusto: > This is probably fine for now - but I wonder if we should pass a config struct into this (either new or existing).

To drop my 2 cents. IMO we should give all the settings as configuration variables, wrapped in a [`watch`](https://docs.rs/tokio/latest/tokio/sync/watch/index.html). This would make it possible to have config hot reload.

bobbinth: > I could add a struct as follows

I don't think we need to do this yet - but if we need to pass in more config settings into the state view, a dedicated struct may be needed.

> To drop my 2 cents. IMO we should give all the settings as configuration variables, wrapped in a [`watch`](https://docs.rs/tokio/latest/tokio/sync/watch/index.html). This would make it possible to have config hot reload.

Let's create an issue to enable hot config reloads. I don't think this is a priority right now - but would be a cool feature to support down the line.

bobbinth (approved):
Looks good! Thank you! I made a small commit with minor wording changes to some of the comments, but other than that, it is good to merge.

Issue #215, opened 2024-02-06
Create a faucet website for node testing


phklive:
For testing purposes it would be useful to be able to initialise that node and make a simple frontend available where any user could input their account ID and get a note generated with some amount of tokens. 

When initialising the node for testing the faucet could be automatically created and the faucet website started. We could have both the node and website living on the same machine. 

**Goal**: Create a simple website frontend faucet enabling users to get assets.

phklive:
Depends on: #223

Dominik1999:
Will this happen this week?

phklive:
> Will this happen this week?

That's the goal, working on it actively.

Dominik1999:
This is done, right?

https://ethdenver.polygonmiden.io/

phklive:
It works indeed but the code has not yet been merged with the miden-node repo. Will close once it's merged.

https://github.com/0xPolygonMiden/miden-node/pull/247

gubloon:
> This is done, right?
> 
> https://ethdenver.polygonmiden.io/

@Dominik1999 
Can we repurpose this domain(polygonmiden.io) for playground or if we are not going to use http://miden.io/ anymore?

Dominik1999:
we can use miden.io indeed.

Dominik1999:
> It works indeed but the code has not yet been merged with the miden-node repo. Will close once it's merged.
> 
> [#247](https://github.com/0xPolygonMiden/miden-node/pull/247)

It looks like the PR was closed by not merged. Can you update this issue here accordingly?

phklive:
Yes this PR has not yet been merged it needs some more cleanup, will merge by this week.

This is the new PR: #270

Dominik1999:
What is missing here? Can we merge that today?

phklive:
> What is missing here? Can we merge that today?

This PR is ready to be reviewed, waiting for an approval to merge.

Dominik1999:
Closed by #270

PR #224, opened 2024-02-12
Add faucet for miden-node

phklive:
- a Faucet account
- a regular account

closes: #215


bobbinth:
Awesome! One question: would it make sense to add this feature into the `miden-node` crate so that the faucet could be started together with the node (maybe using a command line arg).

One benefit of this is fewer components for users to manage. The downside is potential extra complexity in the node. I'm not sure how these balance each other.

PR #226, opened 2024-02-13
Changed types of protobuf fields

### LINKED ISSUE #112: List of type changes of protobuf structs' fields
Agreed that there are no benefits in terms of encoding efficiency - but I'd say that specifying an upper bound is still useful for a variety of reasons.

Also, in this case specifically, we should probably be using `fixed32` data type as we'll get to needing 4 bytes of variable encoding within about a month (assuming 3-second block times).

I went through the other protobuf definitions, and I think we should make the following changes:
- In the [BlockHeader](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/block_header.proto#L6) message
  - Change `block_num` type to `fixed32`.
  - I'm also wondering if we should change `timestamp` to `fixed32` as well because in our case variable-length encoding will always take 5 bytes. But I maybe saving one byte is not worth it.
- In the [Note](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/note.proto), [NoteCreated](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/requests.proto#L13), and [NoteSyncRecord](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/responses.proto#L34) messages:
  - Change `block_num` type to `fixed32`.
  - Change `sender` type to `AccountId`.
  - Change `tag` type to `fixed64` as in most cases this will be set to account ID.
  - Ideally, `num_assets` would be a `fixed8` (because it would always be a single byte), but there is no such type in protobuf.
- In the [GetBlockHeaderByNumberRequest](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/requests.proto#L32), [AccountHashUpdate](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/responses.proto#L23), [NullifierUpdate](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/responses.proto#L29), [NullifierTransactionInputRecord](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/responses.proto#L29), [SyncStateRequest](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/requests.proto#L40), and [NullifierLeaf](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/tsmt.proto#L23) messages:
  -  Change `block_num` type to `fixed32`.
- In the [SyncStateResponse](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/proto/responses.proto#L43) message:
  - Change `chain_tip` to `fixed32`.

> anything that uses the block_num with a u64 / Felt needs conversion. example [1](https://github.com/0xPolygonMiden/miden-base/blob/main/objects/src/block/header.rs#L24) [2](https://github.com/0xPolygonMiden/miden-node/blob/main/store/src/state.rs#L413-L416).

I would say that in the first example, we should probably change `block_num` (as well as `timestamp` and `version`) to non-`Felt` data types.

In the second example, I think the conversion cannot be avoided as we are converting from `Word` to either `u64` or `u32`.

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-node/issues/84#issuecomment-1848867480_

polydez:
Changed types of protobuf fields. This resolves #112.

### DIFF
--- a/proto/proto/block_header.proto
+++ b/proto/proto/block_header.proto
@@ -23,5 +23,5 @@ message BlockHeader {
     // specifies the version of the protocol.
     uint32 version = 9;
     // the time when the block was created.
-    fixed32 timestamp = 10;
+    uint64 timestamp = 10;
 }
--- a/proto/proto/block_header.proto
+++ b/proto/proto/block_header.proto
@@ -7,7 +7,7 @@ message BlockHeader {
     // the hash of the previous blocks header.
     digest.Digest prev_hash = 1;
     // a unique sequential number of the current block.
-    uint32 block_num = 2;
+    fixed32 block_num = 2;
     // a commitment to an MMR of the entire chain where each block is a leaf.
     digest.Digest chain_root = 3;
     // a commitment to account database.
@@ -23,5 +23,5 @@ message BlockHeader {
     // specifies the version of the protocol.
     uint32 version = 9;
     // the time when the block was created.
-    uint64 timestamp = 10;
+    fixed32 timestamp = 10;
 }
--- a/proto/proto/note.proto
+++ b/proto/proto/note.proto
@@ -14,7 +14,6 @@ message Note {
     merkle.MerklePath merkle_path = 7;
 }
 
-// TODO: change `sender` to AccountId
 message NoteSyncRecord {
     uint32 note_index = 1;
     digest.Digest note_id = 2;
@@ -23,7 +22,6 @@ message NoteSyncRecord {
     merkle.MerklePath merkle_path = 6;
 }
 
-// TODO: change `sender` to AccountId
 message NoteCreated {
     uint32 note_index = 1;
     digest.Digest note_id = 2;
--- a/proto/proto/note.proto
+++ b/proto/proto/note.proto
@@ -3,13 +3,14 @@ package note;
 
 import "digest.proto";
 import "merkle.proto";
+import "account.proto";
 
 message Note {
-    uint32 block_num = 1;
+    fixed32 block_num = 1;
     uint32 note_index = 2;
     digest.Digest note_id = 3;
-    fixed64 sender  = 4;
-    uint64 tag = 5;
+    account.AccountId sender  = 4;
+    fixed64 tag = 5;
     merkle.MerklePath merkle_path = 7;
 }
 
@@ -17,8 +18,8 @@ message Note {
 message NoteSyncRecord {
     uint32 note_index = 1;
     digest.Digest note_id = 2;
-    fixed64 sender  = 3;
-    uint64 tag = 4;
+    account.AccountId sender  = 3;
+    fixed64 tag = 4;
     merkle.MerklePath merkle_path = 6;
 }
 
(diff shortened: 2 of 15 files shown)


proto/proto/note.proto
```diff
@@ -3,29 +3,30 @@ package note;
...
     digest.Digest note_id = 3;
-    fixed64 sender  = 4;
-    uint64 tag = 5;
+    account.AccountId sender  = 4;
+    fixed64 tag = 5;
     merkle.MerklePath merkle_path = 7;
 }
 
 // TODO: change `sender` to AccountId
 message NoteSyncRecord {
```
bobbinth, line 18: I think the `TODO` can be removed here.

proto/proto/note.proto
```diff
@@ -3,29 +3,30 @@ package note;
...
     digest.Digest note_id = 2;
-    fixed64 sender  = 3;
-    uint64 tag = 4;
+    account.AccountId sender  = 3;
+    fixed64 tag = 4;
     merkle.MerklePath merkle_path = 6;
 }
 
 // TODO: change `sender` to AccountId
 message NoteCreated {
```
bobbinth, line 27: I think the `TODO` can be removed here.

proto/proto/block_header.proto
```diff
@@ -23,5 +23,5 @@ message BlockHeader {
     // specifies the version of the protocol.
     uint32 version = 9;
     // the time when the block was created.
-    uint64 timestamp = 10;
+    fixed32 timestamp = 10;
```
bobbinth, line 26: Let's keep `timestamp` as `uint64` for now.

proto/src/domain/blocks.rs
```diff
@@ -23,8 +23,10 @@ impl From<BlockHeader> for block_header::BlockHeader {
             proof_hash: Some(header.proof_hash().into()),
             version: u64::from(header.version())
                 .try_into()
-                .expect("TODO: BlockHeader.version should be u64"),
-            timestamp: header.timestamp().into(),
+                .expect("Failed to convert BlockHeader.version into u32"),
+            timestamp: u64::from(header.timestamp())
+                .try_into()
+                .expect("Failed to convert BlockHeader.timestamp into u32"),
```
bobbinth, line 29: If we roll back `timestamp` to 64-bit values, this will revert too.

store/src/state.rs
```diff
@@ -480,7 +480,12 @@ pub fn build_notes_tree(
     for note in notes.iter() {
         let note_id =
             note.note_id.clone().ok_or(NoteCreated::missing_field(stringify!(note_id)))?;
-        let account_id = note.sender.try_into().or(Err(ApplyBlockError::InvalidAccountId))?;
+        let account_id = note
+            .sender
+            .clone()
+            .ok_or(NoteCreated::missing_field(stringify!(sender)))?
+            .try_into()
+            .or(Err(ApplyBlockError::InvalidAccountId))?;
```
bobbinth, line 488: In the future, this conversion will need to be done in the `domain` module.

bobbinth (approved):
Looks good! Thank you! I left some comments inline. After these are addressed, we can merge.

proto/proto/block_header.proto
polydez: Are we going to store milliseconds in timestamp? Currently it seems to store seconds from Linux epoche, `u32` should be enough.

bobbinth (approved):
Looks good! Thank you!

Let's create an issue to consider converting `timestamp` to `u32` in the future.

PR #230, opened 2024-02-13
Fix `TODO` for new accounts

phklive:
In this PR I propose to remove the TODO that checks if the transaction was executed against a new account.

I made sure that the transactions initial account hash is indeed equal to `Digest::default()` considering that we received `None` back from the store, meaning that this is a new account.

### DIFF
--- a/block-producer/src/state_view/mod.rs
+++ b/block-producer/src/state_view/mod.rs
@@ -194,7 +194,14 @@ fn ensure_tx_inputs_constraints(
         },
         // if the account is not present in the Store, it must be a new account
         None => {
-            // TODO: check if the transaction was executed against new account
+            // if the initial account hash is not equal to `Digest::default()` it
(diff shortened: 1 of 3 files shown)


phklive:
Question: 

How should we handle old accounts that interact with the node and that are also not recorded in the store?
- Should we simple error out and prevent them from interacting? 
- Is there another thing to do?

bobbinth:
I believe @polydez is working on something that touches the same files. Maybe best to wait until his changes are merged before tackling this one.

> How should we handle old accounts that interact with the node and that are also not recorded in the store?

This should not be possible. If a transaction is against an existing account (i.e., account with `nonce != 0`) but this account is not present in the store, the transaction should be considered invalid.

phklive:
@polydez what's your status on that issue / file?

bobbinth (approved):
Looks good! Thank you!

Issue #223, opened 2024-02-12
Accounts created with `make_genesis` command have nonce 0


phklive:
We define as `new accounts` accounts with a nonce of 0. 

An account that has a nonce of 0 should not be stored in the DB considering that it has never created a tx and hence has never been able to interact with the rollup to end up in the DB. 

We need to fix the way we create accounts through the `make_genesis` function to make sure they have nonce = 1.

https://github.com/0xPolygonMiden/miden-node/blob/db51f392257abcf2fd5a703a15b830e4d8a5f7ab/node/src/commands/genesis/mod.rs#L44

phklive:
This issue has been discussed here: https://github.com/0xPolygonMiden/miden-client/issues/139

The following situation arises:

- We have updated the way we create `ProvenTransaction` in miden-base hence passing `Digest::default()` as `init_account_hash` when the account is new ( nonce 0 )

- Accounts created via `make_genesis` function are added automatically to the DB even though they have never interacted with the DB. Hence they have an `init_account_hash` that is not `Digest::default()` stored in the DB at genesis

- When wanting to make a tx with an account that was created at genesis by importing them into a client it errors out with the following error: 
```
rpc api error: rpc request failed for submit_proven_transaction: status: InvalidArgument, message: "VerificationFailed(IncorrectAccountInitialHash { tx_initial_account_hash: RpoDigest([0, 0, 0, 0]), store_account_hash: Some(RpoDigest([2194417484513961168, 8441993850887968501, 2612373643818268576, 12735460985696479504])) })", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "content-length": "0", "date": "Tue, 13 Feb 2024 10:00:16 GMT"} }
```

- The constraint check fails because the `init_account_hash` in the `ProvenTransaction` is `Digest::default()` ( considering the account is new ) but it is not `Digest::default()` in the Nodes store, hence erroring out.

Some ways to solve this  could be:
- initialise the accounts with nonce of 1 ( Not sure if that is possible without executing a tx )
- Not import the accounts in the nodes DB but wait for a tx
- Not import the accounts in the nodes DB but automatically make 2 tx's adding the accounts into the db

bobbinth:
> initialise the accounts with nonce of 1 ( Not sure if that is possible without executing a tx )

I think this is the way to go. The current `create_basic_fungible_faucet()` etc. functions create accounts with the nonce set to $0$ - so, I think we have 2 options here:
[...]

Issue #222, opened 2024-02-12
Add End-to-End tests for the Miden Node


Dominik1999:
We need to define and run integration tests for the Miden Node.

phklive:
Linked to: 
- https://github.com/0xPolygonMiden/miden-node/issues/93
- https://github.com/0xPolygonMiden/miden-node/issues/144
- https://github.com/0xPolygonMiden/miden-node/issues/145

hackaugusto:
I think this issue is more about an end-to-end test (E2E) instead of the Rust's [integration tests](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests). Here are a few notes that I believe are important for these types of tests:

1. CI:
    1. The E2E only run after successful unit tests
    2. Logs should be aggressively collected and persisted by the CI
2. Tests:
    1. Have overridable timeouts for environment setup/teardown and test execution. default timeout for tests of around 30secs
    2. Can skip the teardown on test failure
    3. Must have automatic retries with a ressetable test timeout configurable by test and disabled by default
3. Environment:
    1. Must be containerized

The rationale of the above is:

1.
     i. E2E tests are super slow, if the unit tests fail it doesn't make sense to waste CI time on the E2e tests
     ii. Debugging E2E tests from CI can be challenging, and the logs are the only tool we have in some cases (specially when tests fail because the CI machines are not that powerful).
2.
     i. When debugging/running a single test, the setup/teardown time will dominate the test runtime, this means these settings should be different. The test timeout is important to detect issues that are outside of the test runner, and to incentivize good test writting.
     ii. When debugging a test, it is useful to keep the environment around to inspect further.
     iii. Flakiness is a given, retries on CI will be needed. Retries should be disabled by default because otherwise super badly written tests can get in, this setting is just a last resort kinda of tool.
3.
     i. it is impossible to do environment cleanup properly without it (e.g. double fork creates a daemon and there is no reliable way of caughting it without a cgroup). Using containers also allows for parallel test execution (useful when working on different branches).

Ideally we would also have:

- The ability to run a new E2E test multiple times when it is introduced, say 5 times. To detect flakiness before merging a PR.
- Collect some metrics about the tests, to know what is slow/flaky and needs an eye to improve
- Test should be written with no assumption they have a clean slate. This would require a setup/teardown per test, which is super slow.

With the above said. I think the best approach here is to have a hand written test runner, that will setup the container environment first, and then call the tests (which can be a rust binary).

[...]

PR #232, opened 2024-02-14
Set nonce of accounts to 1 at genesis

phklive:
In this PR I propose to set the nonce of Miden-Accounts imported at genesis to 1. 

This will prevent errors that we had on checks because of accounts imported at genesis breaking the node constraints.

Closes: #223

### DIFF
--- a/node/src/commands/genesis/mod.rs
+++ b/node/src/commands/genesis/mod.rs
@@ -18,7 +18,7 @@ use miden_node_utils::config::load_config;
 use miden_objects::{
     accounts::{Account, AccountData, AccountType, AuthData},
     assets::TokenSymbol,
-    Felt,
+    Felt, ONE,
 };
 
 mod inputs;
@@ -166,7 +166,7 @@ fn create_accounts(
             }
         }
 
-        account_data.account.set_nonce(Felt::new(1))?;
+        account_data.account.set_nonce(ONE)?;
 
         account_data.write(path)?;
 
--- a/node/src/commands/genesis/mod.rs
+++ b/node/src/commands/genesis/mod.rs
@@ -121,7 +121,7 @@ fn create_accounts(
 
     for account in accounts {
         // build account data from account inputs
-        let account_data = match account {
+        let mut account_data = match account {
             AccountInput::BasicWallet(inputs) => {
                 print!("Creating basic wallet account...");
                 let init_seed = hex_to_bytes(&inputs.init_seed)?;
@@ -165,6 +165,8 @@ fn create_accounts(
                 return Err(anyhow!("Failed to generate account file {} because it already exists. Use the --force flag to overwrite.", path.display()));
             }
         }
+
+        account_data.account.set_nonce(Felt::new(1))?;
 
         account_data.write(path)?;
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,12 +1,20 @@
 [workspace]
-members = ["block-producer", "node", "proto", "rpc", "store", "utils", "test-macro"]
+members = [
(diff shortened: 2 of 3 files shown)


phklive:
After this is merged: https://github.com/0xPolygonMiden/miden-base/pull/466

Need to change back miden-base to main

phklive:
SCR-20240214-nbha

Error from genesis accounts imported into client fixed.

node/src/commands/genesis/mod.rs
```diff
@@ -166,6 +166,8 @@ fn create_accounts(
             }
         }
 
+        account_data.account.set_nonce(Felt::new(1))?;
```
bobbinth, line 169: minor nit: we can import `ONE` constant from `miden-base` and use it here instead of `Felt::new(1)` (that's how we usually do it).

phklive: Questions:
- I saw that we had `ONE` and `ZERO`, curious why we have them. Are they simply shorthands to `Felt::new(0) and 1` or are they useful in another manner? Why only `ONE` and `ZERO`? 
- We can import as an example `ONE` but also many other types from `miden-objects` but also `miden-crypto` why is that? Should we find a way to make sure that all crates depend only on `miden-objects` for global types?

bobbinth: > I saw that we had `ONE` and `ZERO`, curious why we have them. Are they simply shorthands to `Felt::new(0) and 1` or are they useful in another manner? Why only `ONE` and `ZERO`?

Yes - these are just shorthands. We defined only these ones as they are used quite frequently in various places.

> We can import as an example `ONE` but also many other types from `miden-objects` but also `miden-crypto` why is that? Should we find a way to make sure that all crates depend only on `miden-objects` for global types?

Yes - the intent is so that most crates only need to depend on `miden-base` and are not even aware of `miden-crypto`. Doesn't always work out this way, but I try to make it work whenever possible.

bobbinth (approved):
Looks good! Thank you! Could you test this with `miden-client` to make sure that executing transactions still works as expected? If all is good, we can merge.

phklive:
> Looks good! Thank you! Could you test this with `miden-client` to make sure that executing transactions still works as expected? If all is good, we can merge.

Yes it's here, the image I had added. Previously we had an error that is documented here: https://github.com/0xPolygonMiden/miden-node/issues/223#issuecomment-1941119962

phklive:
> > Yes it's here: [#232 (comment)](https://github.com/0xPolygonMiden/miden-node/pull/232#issuecomment-1943801220) the image I had added. Previously we had an error that is documented here: [#223 (comment)](https://github.com/0xPolygonMiden/miden-node/issues/223#issuecomment-1941119962)
> 
> Thank you! And if you sync after minting it also look all good?

Process: 
- Import 2 account, 1 fungible faucet, 1 basic
- sync
- mint transaction from faucet
- sync
- consume tokens in basic acc
- create new basic account
- mint transaction from faucet
- sync
- consume tokens in basic acc
- p2id from basic acc 1 to basic acc 2
- sync

Everything works correctly, no error on node, no error on client.

bobbinth:
> * Import 2 account, 1 fungible faucet, 1 basic
> * sync
> * mint transaction from faucet
> * sync
> * consume tokens in basic acc
> * create new basic account
> * mint transaction from faucet
> * sync
> * consume tokens in basic acc
> * p2id from basic acc 1 to basic acc 2
> * sync
> * consume tokens in basic acc 1
> * sync

Thank you! btw, this could be one of E2E tests we implement as a part of #222.

PR #233, opened 2024-02-14
Using `Nullifier` instead of `Digest` for nullifiers

polydez:
This addresses the comment from [#208](https://github.com/0xPolygonMiden/miden-node/pull/208):

https://github.com/0xPolygonMiden/miden-node/pull/208#discussion_r1485563702

### DIFF
--- a/block-producer/src/block_builder/prover/block_witness.rs
+++ b/block-producer/src/block_builder/prover/block_witness.rs
@@ -4,6 +4,7 @@ use miden_node_proto::domain::blocks::BlockInputs;
 use miden_objects::{
     accounts::AccountId,
     crypto::merkle::{EmptySubtreeRoots, MerklePath, MerkleStore, MmrPeaks, SmtProof},
+    notes::Nullifier,
     vm::{AdviceInputs, StackInputs},
     BlockHeader, Digest, Felt, ZERO,
 };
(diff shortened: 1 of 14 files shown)


block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -308,7 +309,7 @@ impl BlockWitness {
             merkle_store
                 .add_merkle_paths(self.produced_nullifiers.iter().map(|(nullifier, proof)| {
                     // Note: the initial value for all nullifiers in the tree is `[0, 0, 0, 0]`
-                    (u64::from(nullifier[3]), Digest::default(), proof.path().clone())
+                    (u64::from(nullifier.inner()[3]), Digest::default(), proof.path().clone())
```
bobbinth, line 312: Ideally, instead of doing `nullifier.inner()[3]` we should add a method on the `Nullifier` to get the most significant element. Maybe something like `Nullifier::mso()` or `Nullifier::most_significant_felt()`?

bobbinth (commented):
Looks good! Thank you! I left one comment above - not related to this PR per se, but I think might be good to fix as a part of this (we'll need to make another small PR in `miden-base`).

Also, some tests seem to be failing.

bobbinth (approved):
Looks good! Thank you!

Issue #65, opened 2023-11-16
`DefaultTransactionQueue`: Ensure at most 2^12 created notes per batch


plafer:
Related: #64

bobbinth:
I think this may be as simple as making sure that if batch builder gets a list of more than $n$ transactions (i.e. [here](https://github.com/0xPolygonMiden/miden-node/blob/main/block-producer/src/batch_builder/mod.rs#L130), it would return an error. $n$ should probably be defined in `miden-base` (i.e., [here](https://github.com/0xPolygonMiden/miden-base/blob/main/objects/src/lib.rs#L57)). We could call it `MAX_TRANSACTIONS_PER_BATCH` and it would be computed based on `NOTE_TREE_DEPTH`.

bobbinth:
> We could call it `MAX_TRANSACTIONS_PER_BATCH` and it would be computed based on `NOTE_TREE_DEPTH`.

Actually, we need one more constant to support this: `MAX_BATCHES_PER_BLOCK`. Then, `MAX_TRANSACTIONS_PER_BATCH` is defined as `2.pow(NOTE_TREE_DEPTH - MAX_BATCHES_PER_BLOCK.ilog2())`.

PR #239, opened 2024-02-15
Ensure the number of created notes in a batch don't exceed the maximum

hackaugusto:
fixes: #65 

The batch kernel appends notes in order to a merkle tree, the tree has a fixed depth of 13 levels, which limits the maximum number of created notes in a batch.

This PR ensures the proposed batches don't exceed the above limit.

### DIFF
--- a/block-producer/src/txqueue/mod.rs
+++ b/block-producer/src/txqueue/mod.rs
@@ -1,6 +1,7 @@
 use std::{sync::Arc, time::Duration};
 
 use async_trait::async_trait;
+use miden_objects::MAX_NOTES_PER_BATCH;
 use tokio::{sync::RwLock, time};
 use tracing::{debug, info, info_span, instrument, Instrument};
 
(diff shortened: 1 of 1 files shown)


block-producer/src/txqueue/mod.rs
```diff
@@ -100,15 +101,35 @@ where
...
+            let mut batch = Vec::with_capacity(self.options.batch_size);
+            let mut notes_in_batch = 0;
+
+            while let Some(tx) = txs.pop() {
+                notes_in_batch += tx.output_notes().num_notes();
+
+                debug_assert!(
+                    tx.output_notes().num_notes() <= MAX_NOTES_PER_BATCH,
+                    "Sanity check, the number of output notes of a single transaction must never be larger than the batch maximum",
+                );
```
hackaugusto, line 114: The real limit is much lower, defined by:

https://github.com/0xPolygonMiden/miden-base/blob/353589bc20c41db8138e11d6b5eecb419e693043/objects/src/constants.rs#L19-L20

This is just a sanity check. That's why it is only a `debug_assert`

polydez (approved):
Looks good to me. The only doubt I have, what if there are too many notes in one single transaction? If this is possible, batches building will return empty batches every time. But I hope, that such situation is impossible.

hackaugusto:
> Looks good to me. The only doubt I have, what if there are too many notes in one single transaction? If this is possible, batches building will return empty batches every time. But I hope, that such situation is impossible.

If everything is working as expected, that [is not possible](https://github.com/0xPolygonMiden/miden-base/blob/353589bc20c41db8138e11d6b5eecb419e693043/objects/src/transaction/outputs.rs#L92-L97).

PR #241, opened 2024-02-15
tracing_forest: fixed issues to use logging with dev envs

### LINKED ISSUE #231: Make sure tracing-forest can be used with top-level binaries
The binaries have a `forest` feature flag that enables [`tracing-forest`](https://lib.rs/crates/tracing-forest) log formatting. This crate holds the log emission when inside a `span` and pretty-prints everything in a tree format. This means that logs are held until the `span` quits, which is an issue for `instrumentation` over a top level function, like [`serve`](https://github.com/0xPolygonMiden/miden-node/blob/3ccd036fe384687bafa10a71ab6d5ac9cb6ea6a1/store/src/server/mod.rs#L15).

This issue is about ensuring logs are emitted while the server is running, instead of only when the program exits.

Context: https://github.com/0xPolygonMiden/miden-node/pull/204

hackaugusto:
Fixed a few issues I found with the logging and also fixes #231 while I was at it.

- Remove the spans that would swallow all the logs when using tracing forest. These were the spans around main and tasks' entry points.
- Added feature flags to enable tracing_forest. The previous approach did not work, since the compilation profile doesn't change the dependencies.
- Added support for `RUST_LOG` and changed the default from `TRACE` to `INFO`.

### DIFF
--- a/utils/src/logging.rs
+++ b/utils/src/logging.rs
@@ -1,5 +1,9 @@
 use anyhow::Result;
-use tracing::subscriber::{self, Subscriber};
+use tracing::{
+    level_filters::LevelFilter,
+    subscriber::{self, Subscriber},
+};
+use tracing_subscriber::EnvFilter;
 
 pub fn setup_logging() -> Result<()> {
     subscriber::set_global_default(subscriber())?;
@@ -9,8 +13,7 @@ pub fn setup_logging() -> Result<()> {
 
 #[cfg(not(feature = "tracing-forest"))]
 pub fn subscriber() -> impl Subscriber + core::fmt::Debug {
-    use tracing::level_filters::LevelFilter;
-    use tracing_subscriber::{fmt::format::FmtSpan, EnvFilter};
+    use tracing_subscriber::fmt::format::FmtSpan;
 
     tracing_subscriber::fmt()
         .pretty()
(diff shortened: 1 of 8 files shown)


utils/src/logging.rs
```diff
@@ -9,8 +13,8 @@ pub fn setup_logging() -> Result<()> {
 
 #[cfg(not(feature = "tracing-forest"))]
 pub fn subscriber() -> impl Subscriber + core::fmt::Debug {
-    use tracing::level_filters::LevelFilter;
-    use tracing_subscriber::{fmt::format::FmtSpan, EnvFilter};
+    println!("using default");
```
polydez, line 16: If this line added not only for testing, we need to have more informative message here.

utils/src/logging.rs
```diff
@@ -30,8 +34,13 @@ pub fn subscriber() -> impl Subscriber + core::fmt::Debug {
 
 #[cfg(feature = "tracing-forest")]
 pub fn subscriber() -> impl Subscriber + core::fmt::Debug {
+    println!("using tracing-forest");
```
polydez, line 37: Same as above.

polydez (approved):
As I understood, top-level instrumenting affects badly to all logging. If so, I agree with changes. I've left nit comments here. All remaining is fine.

utils/src/logging.rs
hackaugusto: oops, it was for testing only, removed it :)

utils/src/logging.rs
hackaugusto: removed :)

PR #247, opened 2024-02-23
Adding a faucet to the miden-node

phklive:
- a Faucet account
- a regular account

closes: #215


phklive:
https://github.com/0xPolygonMiden/miden-node/pull/224#issuecomment-1942932254

To answer your question @bobbinth I believe that it would better to have the faucet as a separate crate to not deeply couple it to the node. Considering that in the future we could have other implementations / backends that could need a faucet. 

Hence leaving it as a standalone component could enable people with configurations to spin up a faucet rapidly and ergonomically.

phklive:
Things that could be done / improved now: 
- Improve error handling in the faucet server code
- Improve error handling / reporting of server errors on the frontend
- Fix different import of `rusqlite` features in the node <> client
- Enable user configuration of the endpoints through config file

PR #248, opened 2024-02-26
Mock proven TX improvements

polydez:
In this PR we got rid of `DummyProvenTxGenerator` in favour of `MockProvenTxBuilder`. The latter was also improved in order to provide all needed features.

Look into https://github.com/0xPolygonMiden/miden-node/issues/79 for context.

### DIFF
--- a/block-producer/src/test_utils/proven_tx.rs
+++ b/block-producer/src/test_utils/proven_tx.rs
@@ -3,7 +3,6 @@
 use std::sync::{Arc, Mutex};
 
 use miden_air::{ExecutionProof, HashFunction};
-use miden_mock::constants::ACCOUNT_ID_REGULAR_ACCOUNT_UPDATABLE_CODE_ON_CHAIN;
 use miden_objects::{
     accounts::AccountId,
     notes::{NoteEnvelope, NoteMetadata, Nullifier},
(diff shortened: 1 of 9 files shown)


block-producer/src/test_utils/proven_tx.rs
```diff
@@ -33,29 +24,58 @@ static NUM_NOTES_CREATED: Lazy<Arc<Mutex<u64>>> = Lazy::new(|| Arc::new(Mutex::n
...
+    final_account_hash: Digest,
     notes_created: Option<Vec<NoteEnvelope>>,
     nullifiers: Option<Vec<Nullifier>>,
 }
 
 impl MockProvenTxBuilder {
     pub fn new() -> Self {
-        let account_index: u32 = {
+        let mock_account: MockPrivateAccount = {
             let mut locked_num_accounts_created = NUM_ACCOUNTS_CREATED.lock().unwrap();
```
hackaugusto, line 37: I'm against this approach. This is adding synchronization among tests. The results of these values will depend on the number and order the tests are executed. Which can easily lead to brittle tests and flakiness.

IMO these values should be inject.

hackaugusto (approved):
Approving, as this PR does what the linked issue is asking for, and the tests are green.

With that said, I really think we shouldn't have global state for tests if we can avoid it. I also think we should reuse code, and for the moment this re-implements a chunk of the mock available in the miden-base.

PR #253, opened 2024-02-29
Improvements in `MockStoreSuccessBuilder`

### LINKED ISSUE #79: `block_builder` tests: fix and use new infrastructure
#77 introduced a new infrastructure useful especially for the `block_builder` tests. It is currently incomplete, and needs to be fixed before every `block_builder` test can be updated to use it. Specifically

- `MockStoreSuccessBuilder` and `build_expected_block()` need to compute the correct notes root
- `MockStoreSuccessBuilder`: `new()` needs to be replaced with a few different constructors (including `from_batches()`) which allow it to know the initial account IDs and state.

Once this is done, use this new test code everywhere, and remove the obsolete `DummyProvenTxGenerator`.

polydez:
- `MockStoreSuccessBuilder` and `build_expected_block_header()` now compute produced notes root
- `MockStoreSuccessBuilder::new()` was replaced with two constructors: `from_batches` and `from_accounts` which allow it to fill initial account IDs and state and also nullifiers and produced notes.

This will resolve #79

### DIFF
--- a/block-producer/src/block_builder/prover/mod.rs
+++ b/block-producer/src/block_builder/prover/mod.rs
@@ -224,7 +224,7 @@ impl BlockProver {
         let proof_hash = Digest::default();
         let timestamp: Felt = SystemTime::now()
             .duration_since(UNIX_EPOCH)
-            .expect("today is expected to be before 1970")
+            .expect("today is expected to be after 1970")
             .as_millis()
             .try_into()
             .expect("timestamp is greater than or equal to the field modulus");
--- a/block-producer/src/test_utils/store.rs
+++ b/block-producer/src/test_utils/store.rs
@@ -42,17 +42,11 @@ impl MockStoreSuccessBuilder {
         };
 
         let created_notes = note_created_smt_from_batches(batches.iter().cloned());
-        let produced_nullifiers = batches
-            .iter()
-            .cloned()
-            .flat_map(TransactionBatch::produced_nullifiers)
-            .map(|nullifier| nullifier.inner())
-            .collect();
 
         Self {
             accounts: Some(accounts_smt),
             notes: Some(created_notes),
-            produced_nullifiers: Some(produced_nullifiers),
+            produced_nullifiers: None,
             chain_mmr: None,
             block_num: None,
         }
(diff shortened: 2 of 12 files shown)


block-producer/src/block_builder/prover/mod.rs
```diff
@@ -224,7 +224,7 @@ impl BlockProver {
         let proof_hash = Digest::default();
         let timestamp: Felt = SystemTime::now()
             .duration_since(UNIX_EPOCH)
-            .expect("today is expected to be before 1970")
+            .expect("today is expected to be after 1970")
```
plafer, line 227: Good point

block-producer/src/test_utils/store.rs
```diff
@@ -3,63 +3,91 @@ use std::collections::BTreeSet;
...
+            SimpleSmt::<ACCOUNT_TREE_DEPTH>::with_leaves(accounts).unwrap()
+        };
+
+        let created_notes = note_created_smt_from_batches(batches.iter().cloned());
+        let produced_nullifiers = batches
+            .iter()
+            .cloned()
+            .flat_map(TransactionBatch::produced_nullifiers)
+            .map(|nullifier| nullifier.inner())
+            .collect();
```
plafer, line 50: This should be removed, and fixes the test (as you mentioned offline).

The problem is that you're initializing the store with the "new" nullifiers that the transaction wants to update. That is, when a batch produces nullifiers (i.e. sets their value to `[block_num, 0, 0, 0]` in the store), they should have a value of `[0, 0, 0, 0]` in the store. Thus, in this function, we should *not* take the produced nullifiers and add them to the store - the block kernel will be doing that.

The error `BlockProverFailed(ProgramExecutionFailed(MerkleStoreLookupFailed(RootNotInStore(...))))` is doesn't give much insight into why this fails (maybe we can fix that). When we initialize the Merkle store from the block witness, we basically assume that the value is `0` in the store ([here](https://github.com/0xPolygonMiden/miden-node/blob/f323ffa15d240388a41bc68953b5995ba47ac68f/block-producer/src/block_builder/prover/block_witness.rs#L292)). But since you set the value to something else here, the Merkle store root becomes inconsistent with the actual store's root (due to the fact that 2 nullifiers are set in the Merkle store, and not set in the actual store). So when we try to `smt::set` the new nullifier [here](https://github.com/0xPolygonMiden/miden-node/blob/f323ffa15d240388a41bc68953b5995ba47ac68f/block-producer/src/block_builder/prover/mod.rs#L127), we provide the incorrect Merkle store root, which the VM complains about.

polydez: That was my main concern! Thank you! I've fixed this and CI is expected to pass now. Please, make a review if you have a chance!

plafer (approved):
LGTM!

Issue #93, opened 2023-12-09
Add full docker-compose


hackaugusto:
The full production system will need some auxiliary services. Namely metrics, tracing, log collection, and in the future maybe a DB. These systems are big and requires a bit of configuration effort, and we should have them available in a testing environment for a few reasons 1. ensure it works in production 2. having these capabilities can be super useful to debug, optimize, and have general insight of the system.

These services are inherently stateful, and once in a while we need to test multiple versions (say, for an upgrade, or change in schema in a dev version that should run on a new instance). The best way to isolate and manage these details in via container technology.

Docker compose is one possibility. But it may be useful to do some researching, MacOS seems to finally got support to container technologies and there may be some better alternatives out-there. For linux I would recommend going with podman, which is compatible with docker/docker-compose.

hackaugusto:
it seems podman actually supports macos nowadays: https://www.redhat.com/sysadmin/run-containers-mac-podman , there are also lima, minikube, rancher, and possibly other options that I missed.

okcan:
Hi agusto, I have a question, why is there a search for macos?

hackaugusto:
@okcan do you mean search of a container stack for macos? for development, most of the team is using macos. running the system inside container for E2E testing makes it easier to test against new environments (i.e. new OS images), to do process cleanup (i.e. just shutdown the cgroup), to test in parallel (by isolating the network/ports in cgroups), and so forth.

okcan:
Sorry, I misunderstood for a moment haha :D

phklive:
Considering the potential refactor of the node from 4 binaries to 1 binary we will close this issue and look into adding a Docker compose in the future if needed. #92

PR #257, opened 2024-03-01
Dockerize node

phklive:
Docker will help us end-to-end test our code & enable easier sharing and installation of our code by our users.

Closes: #93


Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Miden-node Dockerfile
+
+#### Setup image builder
+FROM rust:1.75.0-bullseye AS builder
```
hackaugusto, line 4: questions:

- why bullseye and not bookworm? (bookworm is the latest release)
- have you tried the slim version instead? (for smaller image sizes, even some other base image, like [alpine](https://hub.docker.com/_/alpine))

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN apt-get update && apt-get -y upgrade
+RUN apt-get install -y gcc llvm clang bindgen pkg-config
+
+# Setup workdir
+WORKDIR /app
+COPY . miden-node
+RUN cd miden-node && make
+
+### Run Miden-Node
+FROM ubuntu:22.04
```
hackaugusto, line 16: questions:

- why ubuntu here instead of debian like the builder image? if we can, debian slim would be better
- why 22.04 instead of 24.04?

Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Miden-node Dockerfile
+
+#### Setup image builder
+FROM rust:1.75.0-bullseye AS builder
+
+# Install dependencies
+RUN apt-get update && apt-get -y upgrade
+RUN apt-get install -y gcc llvm clang bindgen pkg-config
```
hackaugusto, line 8: please document why these dependencies are necessary. Something along the lines of:

```
# Require to build sqlite: gcc llvm clang bindgen pkg-config
```

Also, why do we need both gcc and clang?

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Setup workdir
+WORKDIR /app
+COPY . miden-node
+RUN cd miden-node && make
+
+### Run Miden-Node
+FROM ubuntu:22.04
+RUN apt-get update && apt-get -y upgrade && apt-get install -y make libssl-dev libsqlite3-dev curl
```
hackaugusto, line 17: Here too, please document why the dependencies are needed.

Why do we need `-dev` here? The builder should be compiling everything, no?

scripts/start-miden-node.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+miden-node make-genesis --inputs-path genesis.toml
```
hackaugusto, line 2: I think we can generate the genesis in the image generation and cache it.

IOW, this should be a `RUN` directive in the `Dockerfile` instead of being here

Makefile
```diff
@@ -0,0 +1,8 @@
+FEATURES=testing
+
+install:
+	cargo install --features=${FEATURES} --path node
+
+run:
+	miden-node make-genesis --inputs-path node/genesis.toml
+	miden-node start --config node/miden-node.toml
```
hackaugusto, line 8: The goal is to move from Makefile to `cargo make`, could you try using that instead?

reference: https://github.com/0xPolygonMiden/miden-node/issues/251

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY --from=builder /app/miden-node/scripts/start-miden-node.sh start-miden-node.sh
+COPY --from=builder /app/miden-node/node/miden-node.toml miden-node.toml
+COPY --from=builder /app/miden-node/node/genesis.toml genesis.toml
+COPY --from=builder /usr/local/cargo/bin/miden-node /usr/local/bin/miden-node
+RUN chmod +x start-miden-node.sh
+EXPOSE 57291
+CMD [ "./start-miden-node.sh" ]
```
hackaugusto, line 24: This deserves some documentation. Specifically why this is sufficient as a PID 1

Dockerfile
hackaugusto, line 8: Let's merge these two layers so that we use less disk:

```suggestion
RUN apt-get update && apt-get -y upgrade && apt-get install -y gcc llvm clang bindgen pkg-config
```

Dockerfile
hackaugusto: This image should have [`LABEL`s](https://docs.docker.com/reference/dockerfile/#label). I suggest using [open containers labels](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys):

```
LABEL org.opencontainers.image.authors=miden@polygon.io \
      org.opencontainers.image.url=https://0xpolygonmiden.github.io/ \
      org.opencontainers.image.documentation=https://github.com/0xPolygonMiden/miden-node \
      org.opencontainers.image.source=https://github.com/0xPolygonMiden/miden-node \
      org.opencontainers.image.vendor=Polygon \
      org.opencontainers.image.licenses=MIT

ARG CREATED
ARG VERSION
ARG COMMIT
LABEL org.opencontainers.image.created=$CREATED \
      org.opencontainers.image.version=$VERSION \
      org.opencontainers.image.revision=$COMMIT
```

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+FROM rust:1.75.0-bullseye AS builder
+
+# Install dependencies
+RUN apt-get update && apt-get -y upgrade
+RUN apt-get install -y gcc llvm clang bindgen pkg-config
+
+# Setup workdir
+WORKDIR /app
+COPY . miden-node
+RUN cd miden-node && make
```
hackaugusto, line 13: I think it makes sense to use a [`RUN --mount=type=cache`](https://docs.docker.com/reference/dockerfile/#run---mounttypecache) so that we can reuse the host's cargo cache. And just copy the final binary to the image.

Dockerfile
hackaugusto: We need to add a [`VOLUME`](https://docs.docker.com/reference/dockerfile/#volume) to have the store's db.

hackaugusto (commented):
Thanks! I left a few questions and suggestions.

Dockerfile
phklive: Bullseye seemed the most used, how can I know which one is the latest one?

hackaugusto: The images are [listed here](https://hub.docker.com/_/debian),

Dockerfile
phklive: I have followed what was written in the README:

https://github.com/0xPolygonMiden/miden-node/blob/3c3ebbecbece73b330fe01cfa2b6a41c825c8db1/README.md?plain=1#L35

Indeed I don't think that we need both, should we make a modification?

Dockerfile
phklive: Also taken from the README: 

https://github.com/0xPolygonMiden/miden-node/blob/3c3ebbecbece73b330fe01cfa2b6a41c825c8db1/README.md?plain=1#L35

scripts/start-miden-node.sh
phklive: Indeed, changing it.

Makefile
phklive: Changing to that

phklive: Will do so in another PR that we will merge before this one. 

Closing this issue: #251

Dockerfile
phklive: Could you please clarify what should be documented, I believe that the node should be the running process.

Dockerfile
hackaugusto: @polydez do we need both compilers? If not, I would simplify this.

Dockerfile
hackaugusto: Okay, so the idea of the builder final image split is to have smaller image sizes, and a smaller surface area for issues. The goal is to have the build dependencies only in the builder, and then copy only the final artifacts to the last image.

So I think we should remove this.

Dockerfile
hackaugusto: The PID 1 is special, it has to handle zombie processes and signals properly, otherwise the container misbehaves. This is okay as long as we don't spawn new processes on our architecture, and this should be documented. Some like:

```
# Miden node does not spawn sub-processes, so it can be used as the PID1
```

Another approach is to just fix the node to actually do the zombie cleanup

phklive: What do you mean by "fixing the node to do the zombie cleanup" ?

hackaugusto: Here are some docs:

https://github.com/docker-library/official-images?tab=readme-ov-file#init
https://github.com/krallin/tini/issues/8
https://www.padok.fr/en/blog/docker-processes-container

Dockerfile
phklive: Removed gcc and kept clang

Dockerfile
bobbinth, line 1: Does this and `.dockerignore` files need to be in the root directory? Given that they are specifically for the `node` crate, would it make sense to put them into the `node` directory?

Dockerfile
polydez: @hackaugusto @phklive can't answer without testing, but might work.

Dockerfile
phklive: Yes I will move the Dockerfile to the node crate. This is the first step, second step is the docker composition, each crate will have their own `Dockerfile` enabling individual managment and spin-up.

Dockerfile
phklive: To add a volume we would need to modify where the node reads and writes it's database files in order to persist them to the volume.

polydez (approved):
I've left one nit comment inline, but overall it looks good!

Makefile.toml
```diff
@@ -3,6 +3,7 @@
 [env]
 CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = true
 
+# linting
```
hackaugusto, line 6: no need to change on this PR, but for the future, these comments that only tag code are not necessary. This is repeating the context we have in the `.toml` itself.

Dockerfile
hackaugusto: > To add a volume we would need to modify where the node reads and writes it's database files in order to persist them to the volume.


The [`WORKDIR`](https://docs.docker.com/reference/dockerfile/#workdir) setting should handle that.

node/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+RUN miden-node make-genesis --inputs-path node/genesis.toml
+
+# Run Miden node
+FROM debian:bookworm-slim
+
+# Install required packages
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    libssl-dev \
+    libsqlite3-dev \
```
hackaugusto, line 31: Do we need these dependencies? If so, can you please add some docs about it?

hackaugusto (approved):
Thanks! LGTM

node/Dockerfile
hackaugusto: This is the runner image, no?

Makefile.toml
phklive: This is needed by cargo-make to enable the "workspace" mode, if this flag is not set cargo-make does not work properly.

node/Dockerfile
phklive: Yes it is indeed. Running the program failed when these were not installed on the machine.

Makefile.toml
hackaugusto: ah, thanks for the explanation, i didn't know that. Can you please give links to the docs about this?

node/Dockerfile
hackaugusto: That is interesting. Can you please add a comment above the deps. It seems a bit off that the headers are needed to run the compiled binary. Also, what was the error?

phklive: 2024-03-12 16:55:51 miden-node: error while loading shared libraries: libsqlite3.so.0: cannot open shared object file: No such file or directory

Makefile.toml
phklive: Sure, I will add them to the code.

node/Dockerfile
hackaugusto: Ah, for that you shouldn't need the headers, you only need the library. `libsqlite3` should be enough, it is only marginally better though (to avoid confusion like I had here)

hackaugusto: Here is what I have been proposing:

- keep dependencies to a minimum
- document why they are needed

why am I proposing the above?

- it is hard to maintain script files without this information, for example, the discussion of `gcc` and `clang` being needed is something that happened because of lack of documentation. I'm trying to prevent that
- the above is specially a problem when upgrade the file down the line, and usually causes issues were one doesn't know why something is there, so we default to "better not touch it since it is working", which can compound over time

so here, since `libsqlite3-dev` is not needed, these are the headers, we would remove that dependency and have `libsqlite3` instead. since that is the `.so` file that is needed to run the software, and add a comment `# sqlite3: used by the store`.

That way we have a minimum number of dependencies and some traceability on them, making it easier to maintain it.

Dockerfile
phklive: This is the last thing pending, how not yet been able to make it work correctly. Looking into it now.

phklive: Managed to fix this, now the node creates a db folder to put it's sqlite files, this volume is mounted from the host as a volume.

.dockerignore
```diff
@@ -0,0 +1,36 @@
...
+**/charts
+**/docker-compose*
+**/compose*
+**/Dockerfile*
+**/node_modules
+**/npm-debug.log
+**/secrets.dev.yaml
+**/values.dev.yaml
+/bin
+/target
```
hackaugusto, line 30: This is not working for me. Maybe it is a different among podman and docker. But this is causing me to upload 11G of data to the docker VM, and at first sign it looked like podman wasn't working.

Can you change the paths from `/bin` and `/target` to `./bin` and `./target`? And also test on your machine?

hackaugusto: I now wonder how the `--mount` is implemented on `MacOs`, because I may have given bad advice with the cargo cache below, if the data has to be copied to the VM.

node/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+    apt-get install -y llvm clang bindgen pkg-config libssl-dev libsqlite3-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+# Copy source code
+WORKDIR /app
+COPY . .
+
+# Cache Cargo dependencies
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    cargo fetch --manifest-path node/Cargo.toml
```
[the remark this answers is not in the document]
hackaugusto: Yeah, I just tested, the cache is not working, the VM is downloading everything fresh:

```
[1/2] STEP 10/12: RUN --mount=type=cache,target=/usr/local/cargo/registry     cargo fetch --manifest-path node/Cargo.toml
info: syncing channel updates for '1.75-aarch64-unknown-linux-gnu'
info: latest update on 2023-12-28, rust version 1.75.0 (82e1608df 2023-12-21)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
```

(also compiling)

node/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+# Build the node crate
+RUN cargo install --features testing --path node
+RUN miden-node make-genesis --inputs-path node/genesis.toml
+
+# Run Miden node
+FROM debian:bookworm-slim
```
hackaugusto, line 40: Note: The final image has an okay size:

```
REPOSITORY                  TAG                 IMAGE ID      CREATED             SIZE
localhost/miden-node-image  latest              63b9223ac3cb  About a minute ago  124 MB
```

hackaugusto: Note: `start`/`stop` is working, the state is not lost

node/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+LABEL org.opencontainers.image.authors=miden@polygon.io \
+      org.opencontainers.image.url=https://0xpolygonmiden.github.io/ \
+      org.opencontainers.image.documentation=https://github.com/0xPolygonMiden/miden-node \
+      org.opencontainers.image.source=https://github.com/0xPolygonMiden/miden-node \
+      org.opencontainers.image.vendor=Polygon \
+      org.opencontainers.image.licenses=MIT
+
+ARG CREATED
+ARG VERSION
+ARG COMMIT
```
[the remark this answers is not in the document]
hackaugusto: I think this is not a good idea. When I'm debugging the DockerFile, I will rebuild the same image multiple times using the same `COMMIT`, the same `VERSION`, and different `CREATED` timestamps.

The `CREATED` timestamps change will invalid the cache, and force a rebuild of everything that follows it. Moving this to the top of the file means we never reuse the caching system when rebuilding images, and we spend tons of time with rebuilds.

Please move it down, as far as you can. Here is the [docs](https://docs.docker.com/reference/dockerfile/#impact-on-build-caching)

Makefile.toml
```diff
@@ -43,3 +42,14 @@ dependencies = [
...
+
+# testing
+[tasks.test]
+command = "cargo"
+args = ["test", "--all-features", "--workspace", "--", "--nocapture"]
+
+# docker
+[tasks.docker-build-node]
+command = "docker"
+args = ["build", "-f", "node/Dockerfile", "-t", "miden-node-image", "."]
```
hackaugusto, line 54: This is not setting the `ENV` vars:

```
time="2024-03-14T13:29:04+01:00" level=warning msg="missing \"CREATED\" build argument. Try adding \"--build-arg CREATED=<VALUE>\" to the command line"
time="2024-03-14T13:29:04+01:00" level=warning msg="missing \"VERSION\" build argument. Try adding \"--build-arg VERSION=<VALUE>\" to the command line"
time="2024-03-14T13:29:04+01:00" level=warning msg="missing \"COMMIT\" build argument. Try adding \"--build-arg COMMIT=<VALUE>\" to the command line"
```

node/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+    apt-get -y upgrade && \
+    apt-get install -y --no-install-recommends \
+    sqlite3 \
+    && rm -rf /var/lib/apt/lists/*
+
+# Define volume for SQLite database files
+VOLUME db
+
+# Copy artifacts from the builder stage
+COPY --from=builder /app/node/miden-node.toml miden-node.toml
```
hackaugusto, line 54: This means we can't change the configuration without having to rebuild all images. This is not good. This should be mounted from the host.

hackaugusto (changes requested):
Thank you! I got the node started and played around with it a bit :) Super neat. We have some few things to fix though.

node/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+      org.opencontainers.image.source=https://github.com/0xPolygonMiden/miden-node \
+      org.opencontainers.image.vendor=Polygon \
+      org.opencontainers.image.licenses=MIT
+
+ARG CREATED
+ARG VERSION
+ARG COMMIT
+LABEL org.opencontainers.image.created=$CREATED \
+      org.opencontainers.image.version=$VERSION \
+      org.opencontainers.image.revision=$COMMIT
```
hackaugusto, line 19: I don't think we need the labels for the builder image. The image that will be published is the image with the node. Please move it down to the other one, and move the `ARG` down as further down as possible.

Makefile.toml
phklive: Indeed, I am curious from what data you would want to populate these fields ?

node/Dockerfile
phklive: > Note: The final image has an okay size:
> 
> ```
> REPOSITORY                  TAG                 IMAGE ID      CREATED             SIZE
> localhost/miden-node-image  latest              63b9223ac3cb  About a minute ago  124 MB
> ```

Do you have ideas in mind to improve on it?

phklive: > Note: `start`/`stop` is working, the state is not lost

This is positive right?

node/Dockerfile
phklive: I enabled the manual mounting of the file using the `-v` flag.

Makefile.toml
phklive: I added the version and created from the crate version and `date` shell command. 

I am not sure how you would like to get the commit info.

node/Dockerfile
hackaugusto: > Do you have ideas in mind to improve on it?

I think this size is good enough. I was just adding context to the PR.

> Note: start/stop is working, the state is not lost

Yes. This was not a request. I was just adding context to the PR.

Makefile.toml
hackaugusto: The `COMMIT` should have the git commit.
The `VERSION` should have the node's software version.
The `CREATED` should have the created timestamp.

node/Dockerfile
hackaugusto: ah, makes sense, so this file is the default, and the user can overwrite it with the flag 👍

Makefile.toml
```diff
@@ -43,3 +42,36 @@ dependencies = [
...
+command = "cargo"
+args = ["test", "--all-features", "--workspace", "--", "--nocapture"]
+
+# docker
+[tasks.docker-build-node]
+workspace = false
+script = '''
+CREATED=$(date)
+VERSION=$(cat node/Cargo.toml | grep -m 1 '^version' | cut -d '"' -f 2)
+COMMIT=1
```
hackaugusto, line 57:
```suggestion
COMMIT=$(git rev-parse HEAD)
```

should do the trick for this one

hackaugusto (approved):
LGTM, thanks!

node/Dockerfile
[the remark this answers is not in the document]
phklive: Indeed for now the caching does not work. I am looking into doing it manually as we did or using `cargo-chef`: https://github.com/LukeMathWalker/cargo-chef

What do you think?

phklive:
Tested. Last thing we could do would be to improve caching to reduce build time of the image, created a issue so we can merge this PR and go forward: https://github.com/0xPolygonMiden/miden-node/issues/277

miden-node.toml
bobbinth, line 1: Quick question: why do we need to duplicate this file here?

phklive: We do not need to, fixed it in the next PR.

PR #261, opened 2024-03-05
gh: add issue templates

hackaugusto:
This adds issue templates. The goals here are:

- Make sure that we communicate what we want in the issue. The goal is for the issue to have enough context so that another dev can pick the issue up and start working on it.
- Add structure to issues when opened by people outside of the team. The goal is to make the issue template structure in such a way it is helpful for us to understand the external request, and to help any external contributor to give us the context we need to understand the issue/feature.

---

GH has moved away from the markdown templates to yaml based templates. So this PR is using the later. The syntax is [documented here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema).

I have kept the option of writing a blank issue, just in case the existing templates don't cover all possibilities, but I added an old style markdown template to give it at least some structure.

### DIFF
--- a/.github/ISSUE_TEMPLATE/3-task.yml
+++ b/.github/ISSUE_TEMPLATE/3-task.yml
@@ -1,0 +1,48 @@
+name: "Task"
+description: "Work item"
+title: "[Task]: "
+body:
+  - type: markdown
+    attributes:
+      value: |
+        A task should be less than a week worth of work!
+  - type: checkboxes
+    id: component
+    attributes:
+      label: "Node's components affected by this task"
+      options:
+        - RPC
+        - Block producer
+        - Store
+        - Protobuf messages
+        - Testing
+    validations:
+      required: true
+  - type: textarea
+    id: task-what
+    attributes:
+      label: "What should be done?"
+      placeholder: "Add healthchecking to the servers"
+    validations:
+      required: true
+  - type: textarea
+    id: task-how
+    attributes:
+      label: "How should it be done?"
+      placeholder: "Add a /health endpoint to each of the servers, that returns a small JSON reponse and `200` while the service is healthy"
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/3-task.yml
```diff
@@ -0,0 +1,48 @@
+name: "Task"
+description: "Work item"
+title: "[Task]: "
+body:
+  - type: markdown
+    attributes:
+      value: |
+        A task should be less than a week worth of work!
```
Dominik1999, line 8: Love this!

.github/ISSUE_TEMPLATE/3-task.yml
```diff
@@ -0,0 +1,48 @@
...
+    attributes:
+      label: "What should be done?"
+      placeholder: "Add healthchecking to the servers"
+    validations:
+      required: true
+  - type: textarea
+    id: task-how
+    attributes:
+      label: "How should it be done?"
+      placeholder: "Add a /health endpoint to each of the servers, that returns a small JSON reponse and 200 while the service is healthy"
```
Dominik1999, line 32:
```suggestion
      placeholder: "Add a /health endpoint to each of the servers, that returns a small JSON reponse and `200` while the service is healthy"
```

Dominik1999 (approved):
Love it, looks good to me.

Let's do that for `miden-base` and `miden-client`

hackaugusto:
Half seems to be working:

bug:

image

feature:

image

The task form and the issue template are not working :(

bobbinth:
This looks great! One suggestion: maybe we should also have something like "Other" template in case the issues don't fit into one of the 3 categories?

PR #262, opened 2024-03-05
gh: fix task template

hackaugusto:
I found a issue with the task template. Maybe this is the reason why it is not rendering.

There should also be a issue template validation somewhere, at least the errors are [documented here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms). But I couldn't find the tool, I guess it is only available for the repo's admins.

Edit: Ahhhh, that was simple. I just had to navigate to the yaml file:

image

And yeah, the issue was the lack of a `label` in there.


Dominik1999 (approved):
lgtm thanks

PR #264, opened 2024-03-06
Fixed `nullifier_value_to_block_num` function

### LINKED ISSUE #249: `SubmitProvenTransaction` does not error on already-spent notes
Currently, a client can double-spend a note by executing 2 consecutive similar transactions and not syncing in between (if there was a sync and the client knows the note was consumed, we can simply error before attempting execution). What happens on the node is the following: `SubmitProvenTransaction` does not error out (ie, it does not check double-spending at the RPC call), but the block builder will then fail iteratively, bringing the node into a failure state since it won't be able to construct blocks anymore. 

The error is the following:

````
2024-02-26T14:33:58.016077Z  INFO try_build_block:build_block:execute_program: miden_processor: .cargo/git/checkouts/miden-vm-3db6be6805d1979c/c718a0a/processor/src/lib.rs:119: close, time.busy: 30.3ms, time.idle: 54.1µs
2024-02-26T14:28:48.008614Z ERROR try_build_block:build_block: miden-block-producer: block-producer/src/block_builder/mod.rs:69: error: failed to compute new block: program execution failed
````

### PR DESCRIPTION
**Pull request by polydez**
This bug leaded to incorrect checking of already consumed notes.

This will resolve #249

### DIFF
--- a/proto/src/domain/mod.rs
+++ b/proto/src/domain/mod.rs
@@ -1,5 +1,3 @@
-use miden_objects::Word;
-
 pub mod accounts;
 pub mod blocks;
 pub mod digest;
@@ -24,12 +22,4 @@ where
     From: TryInto<To, Error = E>,
 {
     from.into_iter().map(|e| e.try_into()).collect()
-}
-
-/// Given the leaf value of the nullifier SMT, returns the nullifier's block number.
-///
-/// There are no nullifiers in the genesis block. The value zero is instead used to signal absence
-/// of a value.
-pub fn nullifier_value_to_block_num(value: Word) -> u32 {
-    value[0].as_int().try_into().expect("invalid block number found in store")
 }
--- a/proto/src/domain/mod.rs
+++ b/proto/src/domain/mod.rs
@@ -31,5 +31,5 @@ where
 /// There are no nullifiers in the genesis block. The value zero is instead used to signal absence
 /// of a value.
 pub fn nullifier_value_to_block_num(value: Word) -> u32 {
-    value[3].as_int().try_into().expect("invalid block number found in store")
+    value[0].as_int().try_into().expect("invalid block number found in store")
 }
--- a/store/src/state.rs
+++ b/store/src/state.rs
@@ -487,7 +487,7 @@ fn block_num_to_nullifier_value(block: BlockNumber) -> Word {
 ///
 /// There are no nullifiers in the genesis block. The value zero is instead used to signal absence
 /// of a value.
-pub fn nullifier_value_to_block_num(value: Word) -> BlockNumber {
+fn nullifier_value_to_block_num(value: Word) -> BlockNumber {
     value[0].as_int().try_into().expect("invalid block number found in store")
 }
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: proto/src/domain/mod.rs [hunk 1]
```diff
@@ -31,5 +31,5 @@ where
 /// There are no nullifiers in the genesis block. The value zero is instead used to signal absence
 /// of a value.
 pub fn nullifier_value_to_block_num(value: Word) -> u32 {
```
plafer, line 33: This function is duplicated in a few places in the codebase, such as [here](https://github.com/0xPolygonMiden/miden-node/blob/3c3ebbecbece73b330fe01cfa2b6a41c825c8db1/store/src/state.rs#L482-L484). Ideally we'd only have it in one place - and probably on the `Nullifier` type somehow

plafer (approved):
LGTM!

proto/src/domain/mod.rs
polydez: Thanks, I was thinking about it too and gone to the same conclusion. :) Let me implement it. :)

polydez: @plafer I've moved it into the `store`.

store/src/state.rs
```diff
@@ -479,10 +479,18 @@ impl State {
...
+fn block_num_to_nullifier_value(block: BlockNumber) -> Word {
     [Felt::from(block), Felt::ZERO, Felt::ZERO, Felt::ZERO]
 }
 
+/// Given the leaf value of the nullifier SMT, returns the nullifier's block number.
+///
+/// There are no nullifiers in the genesis block. The value zero is instead used to signal absence
+/// of a value.
+pub fn nullifier_value_to_block_num(value: Word) -> BlockNumber {
+    value[0].as_int().try_into().expect("invalid block number found in store")
```
hackaugusto, line 491: Oh, I see, the bug was using `3` instead of `0` here?

Lets please add a test for this.

proto/src/domain/mod.rs
plafer: That's fine for now, but I think the ideal solution would be to
1. Use miden-base's [`Nullifier`](https://github.com/0xPolygonMiden/miden-base/blob/9a1b1263b1faf661623d74e00878cf51371b1088/objects/src/notes/nullifier.rs#L23) type in the node
2. Add a `Nullifier::from_block_num(u32) -> Self` method to `Nullifier`, or something like that.

If you agree, can you create an issue for that? Thanks!

polydez: @plafer no, we use `Nullifier` now for storing nullifier's hash, but these functions for encoding/decoding nullifier's value. Nullifier's value stores just block number, where nullifier was consumed.

polydez: We can add different type for that, something like `NullifierData`.

store/src/state.rs
polydez: @hackaugusto I've added tests.

proto/src/domain/mod.rs
plafer: Ah yes you're right, my bad! Agreed with something like `NullifierData`

PR #265, opened 2024-03-06
config: add systemd service files

hackaugusto:
This adds a two minimal systemd files to run a `miden-node` and `miden-faucet`.

These two service files are far from perfect, but it is nice to have them around if we need them once more.

For the moment the `faucet` is pigbacking in the node for the account details (from the genesis), and it removes the client's db on every restart. Both are not great, but to improve these the faucet needs to be improved. Which is at the moment is under development.

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -1,0 +1,13 @@
+## Service configuration files
+
+- `miden-faucet.service`: simple configuration to setup a faucet service monitored via systemd.
+- `miden-node.service`: simple configuration to setup full node service monitored via systemd.
+
+Install the above files to your system's unit directory (e.g.
+`/etc/systemd/system/`), and run the following commands to start the service:
+
+```sh
+systemctl daemon-reload                 # ask the systemd to load the new configuration files
+systemctl enable --now miden-node       # enable and start the node
+systemctl enable --now miden-faucet     # enable and start the faucet
+```
--- a/config/miden-faucet.service
+++ b/config/miden-faucet.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Miden faucet
+Wants=network-online.target
+Wants=miden-node.service
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Type=exec
+Environment="RUST_LOG=info"
+ExecStart=miden-faucet import --asset-amount 100 --faucet-path /srv/miden-node/accounts/account1.mac
+ExecStartPre=-rm store.sqlite3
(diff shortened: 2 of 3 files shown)


bobbinth:
I am not very familiar with how systemd works - these files are so that we are able to run node/faucet as services, right? If so, maybe worth adding a bit more info about how they are to be used (in the relevant documentation files)?

Also, is `config` directory in the root the right place for them? For example, would it make more sense to put the relevant files into `node` and `faucet` directories?

phklive:
I think that we should not merge this PR for now, I agree that we should have managment for the faucet and node as services but we should wait for this issue to be closed first to make sure we have all the right paths and structure, etc...: #215

hackaugusto:
> I am not very familiar with how systemd works - these files are so that we are able to run node/faucet as services, right? If so, maybe worth adding a bit more info about how they are to be used (in the relevant documentation files)?

Added a bit of docs.

> Also, is config directory in the root the right place for them? For example, would it make more sense to put the relevant files into node and faucet directories?

I have no preference, where do you want them?

> I think that we should not merge this PR for now, I agree that we should have managment for the faucet and node as services but we should wait for this issue to be closed first to make sure we have all the right paths and structure, etc...: https://github.com/0xPolygonMiden/miden-node/issues/215

These files can be updated if changes are necessary.

bobbinth:
> > Also, is config directory in the root the right place for them? For example, would it make more sense to put the relevant files into node and faucet directories?
> 
> I have no preference, where do you want them?

I actually don't know what the best practices on this are. I'm assuming these are needed for node deployment but are separate from the docker build? Basically, how would these be used during deployment? For example, we have something like this:

1. Download/install Rust.
2. Download/install required dependencies.
3. Run `cargo install miden-node`.
4. Generate the genesis file.
5. Run `miden-node start`.

Would these basically replace step 5 in the above with something like?:

1. Download `miden-node.service` from Github.
2. Install it into `/etc/systemd/system`. (is this done by just copying the file?)
3. Run `systemctl daemon-reload` and `systemctl enable --now miden-node`.

bobbinth:
> Yeah, that would work too, but we would need to distribute the `miden-node` binary too.

So - eventually, they would go into building of a `.deb` package, right? In this case, I think putting them into their respective directories makes sense - e.g., `miden-node.service` would go into `node` directory.

Also, should we create an issue for building `.deb` packages?

hackaugusto:
super seeded by https://github.com/0xPolygonMiden/miden-node/pull/312

PR #266, opened 2024-03-07
Grouped nullifier encode/decode methods into `NullifierValueEncoder`

polydez:
Resolved discussion started here: https://github.com/0xPolygonMiden/miden-node/pull/264#discussion_r1514358952

### DIFF
--- a/store/src/state.rs
+++ b/store/src/state.rs
@@ -164,7 +164,8 @@ impl State {
             // update nullifier tree
             let nullifier_tree = {
                 let mut nullifier_tree = inner.nullifier_tree.clone();
-                let nullifier_data = block_num_to_nullifier_value(block_header.block_num());
+                let nullifier_data =
+                    NullifierValueEncoder::encode_block_num(block_header.block_num());
                 for nullifier in nullifiers.iter() {
                     nullifier_tree.insert(*nullifier, nullifier_data);
                 }
@@ -443,7 +444,7 @@ impl State {
             .cloned()
             .map(|nullifier| {
                 let value = inner.nullifier_tree.get_value(&nullifier);
-                let block_num = nullifier_value_to_block_num(value);
+                let block_num = NullifierValueEncoder::decode_block_num(&value);
 
                 NullifierInfo {
                     nullifier: nullifier.into(),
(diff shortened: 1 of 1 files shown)


plafer:
Now that the repo has a `next` branch, shouldn't this PR target `next`?

store/src/state.rs
```diff
@@ -478,17 +479,21 @@ impl State {
 // UTILITIES
 // ================================================================================================
 
-/// Returns the nullifier's leaf value in the SMT by its block number.
-fn block_num_to_nullifier_value(block: BlockNumber) -> Word {
-    [Felt::from(block), Felt::ZERO, Felt::ZERO, Felt::ZERO]
-}
+struct NullifierValueEncoder;
```
plafer, line 482: I was thinking more along the lines of having a `NullifierValue` struct

```rust
struct NullifierValue(Word);
```

With conversions to/from `BlockNum`, etc. Basically the store and block producer would use this type instead of `Word` for nullifier values.

polydez: @plafer it is needed to store this value in `Smt` and it doesn't support anything but `Word`.

polydez:
> Now that the repo has a `next` branch, shouldn't this PR target `next`?

@plafer since it just a refactoring, I don't think it matters, but let ask @bobbinth for his opinion.

store/src/state.rs
plafer: Ah, good point. Created this issue to [resolve this](https://github.com/0xPolygonMiden/crypto/issues/288). Until that is implemented, I personally prefer just a function over this `NullifierValueEncoder` struct - so IMO we should just close this PR.

What do you think?

polydez: @plafer yes, I agree with you, it will be good solution. I'll close this PR, thank you!

polydez:
Closed since we decided to make `Smt` generic over `AsRef<Word>`: https://github.com/0xPolygonMiden/crypto/issues/288

bobbinth:
While I think https://github.com/0xPolygonMiden/crypto/issues/288 could be independently interesting, I wonder if we should use a different approach for this specific issue. Specifically, we could create a newtype `NullifierDb`or `NullifierTree` which would look something like this:
```Rust
pub struct NullifierTree(Smt);

impl NullifierTree {

    pub add(&mut self, nullifier: Nullifier, block_num: u32)-> Result<(), NullifierTreeError> {
        // returns an error if the nullifier is already in the tree
    }

    pub fn get_block_num(&self, nullifier: &Nullifier) -> Option<u32> {
        // returns None if the nullifier is not in the tree
    }
}
```

This should help abstract away the logic of how block numbers are associated with nullifiers and would leave more flexibility for the future (e.g., we could build a specialized implementation of the `NullifierTree` which assumes that leaves are just 32-bit values).

polydez:
@bobbinth great idea, I like it!

PR #267, opened 2024-03-11
Add reproducible builds

phklive:
This improves the reliability of the builds for the miden-node enabling more consistency and reproducibility.

closes: #251


phklive:
@hackaugusto for now we are not using the matrix capabilities in the CI. Should we still keep the matrix structure for future proofing or should we switch to a simpler CI structure?

hackaugusto:
> @hackaugusto for now we are not using the matrix capabilities in the CI. Should we still keep the matrix structure for future proofing or should we switch to a simpler CI structure?

I don't mind either way.

phklive:
@hackaugusto should we also add doc and test in cargo make file? 

Considering that these can be ran without flags for now and easier: `cargo test` against `cargo make test`

What do you think?

phklive:
> > @hackaugusto for now we are not using the matrix capabilities in the CI. Should we still keep the matrix structure for future proofing or should we switch to a simpler CI structure?
> 
> I don't mind either way.

Removed the matrix to simplify CI for now

hackaugusto (approved):
LGTM, thanks!

.gitignore
```diff
@@ -8,7 +8,7 @@ target/
 
 # Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
 # More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
-Cargo.lock
+# Cargo.lock
```
bobbinth, line 11: Is this comment still relevant?

rust-toolchain
```diff
@@ -0,0 +1 @@
+1.75
```
bobbinth, line 1: We should probably set MSRV in the workspace `Cargo.toml` file and I'm curious if there is a way to ensure that one and this one are in sync.

Makefile.toml
bobbinth, line 1: Should we add tasks for testing to this file as well?

phklive: We do have one here: https://github.com/0xPolygonMiden/miden-node/pull/267/files/43b7147397c498bfd44030bc75c819fa7bf80b47#diff-9375fd04332c86472d7be397ef09428cb86babd8826880a5835bd1d1c1bdbc08R30

Are you thinking of something else?

rust-toolchain
phklive: Will add it to the cargo.toml, i can check but  I do not think that there is a standard way to check that both versions match.

.gitignore
phklive: Will remove it.

Makefile.toml
bobbinth: Ah - I missed it! n/m

rust-toolchain
phklive: After doing a quick research i have not found a standard way of checking that all crates match the version of the rust-toolchain file. Hence decided to create a PR to address that and fix your comments:
https://github.com/0xPolygonMiden/miden-node/pull/272

PR #270, opened 2024-03-11
 Add faucet enabling testing

phklive:
- a Faucet account
- a regular account

closes: #215


igamigo:
One thing to keep in mind is that [this PR](https://github.com/0xPolygonMiden/miden-client/pull/189) will be merged soon (and the Client will be released to crates.io either today or tomorrow) so that might imply some minor changes here.

Dominik1999:
Let merge :) 

@igamigo can you review this today or tomorrow?

igamigo:
> Let merge :)
> 
> @igamigo can you review this today or tomorrow?

Will be checking it out in a few hours!

faucet/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+    // let config: FaucetTopLevelConfig =
+    //     load_config(cli.config.as_path()).extract().map_err(|e| {
+    //         io::Error::new(
+    //             io::ErrorKind::InvalidData,
+    //             format!("Failed to load configuration file: {}", e),
+    //         )
+    //     })?;
+
+    // // Instantiate Miden client
+    // let mut client = utils::build_client(config.faucet.database_filepath.clone());
```
igamigo, line 55: Should this be removed?

faucet/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+                    io::ErrorKind::InvalidData,
+                    format!("Failed to import faucet account: {}", e),
+                )
+            })
+        },
+    }?;
+
+    // Sync client
+    client.sync_state().await.map_err(|e| {
+        io::Error::new(io::ErrorKind::ConnectionRefused, format!("Failed to sync state: {e:?}"))
```
igamigo, line 115: nit: I don't think the client failing to sync implies a `ConnectionRefused` necessarily, but I'm not familiar with the enum so not sure if there is a better match

faucet/src/utils.rs
```diff
@@ -0,0 +1,100 @@
...
+use miden_objects::{
+    accounts::{Account, AccountData},
+    assets::TokenSymbol,
+    crypto::dsa::rpo_falcon512::KeyPair,
+    utils::serde::Deserializable,
+    Felt,
+};
+
+/// Instantiates the Miden client
+pub fn build_client(database_filepath: String) -> Client<TonicRpcClient, SqliteStore> {
```
igamigo, line 22: Seeing as `Client<TonicRpcClient, SqliteStore>` is used in various places, it could be benefitial to define a newtype that makes it more succinct.

faucet/src/handlers.rs
```diff
@@ -0,0 +1,100 @@
...
+#[get("/faucet_id")]
+pub async fn faucet_id(state: web::Data<FaucetState>) -> HttpResponse {
+    let response = FaucetIdResponse {
+        faucet_id: state.id.to_string(),
+    };
+
+    HttpResponse::Ok().json(response)
+}
+
+#[get("/get_tokens")]
```
igamigo, line 30: Feels like maybe this should be a `POST` action

faucet/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+actix-web = "4"
+actix-files = "0.6.5"
+actix-cors = "0.7.0"
+derive_more = "0.99.17"
+figment = { version = "0.10", features = ["toml", "env"] }
+miden-lib = { workspace = true }
+miden-client = { version = "0.1.0", features = ["testing","concurrent"] }
```
igamigo, line 15: Should the `testing` be optional for the client? If the node is deployed without it, the client would not work correctly (although this probably won't be needed anytime soon?)

igamigo (approved):
LGTM! Left some non-blocking considerations.
A couple more general points:

- For the website, it seems that the button should post to the server and so you would get some feedback from the browser in terms of the request. The deployed website did not show any feedback and this would be confusing in terms of waiting for the response
- The deployed website is still up, should it be taken offline for now?

faucet/miden-faucet.toml
```diff
@@ -0,0 +1,5 @@
+[faucet]
+# port defined as: sum(ord(c)**p for (p, c) in enumerate('miden-store', 1)) % 2**16
```
mFragaBA, line 2: is this updated? just ran it and got a different result:

```python
>>> sum(ord(c)**p for (p, c) in enumerate('miden-store', 1)) % 2**16
28943
```

Also, does it refer to the `endpoint`'s port or the `rpc_url`'s port?

faucet/src/config.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// Top-level config
+// ================================================================================================
+
+/// Faucet top-level configuration.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+pub struct FaucetTopLevelConfig {
+    pub faucet: FaucetConfig,
+}
```
mFragaBA, line 46: why do we need a top level config?

faucet/src/main.rs
mFragaBA, line 1: I don't think it's strictly needed, but shouldn't the server do a sync_state every once in a while? (I'm not sure whether it's better to do it periodically or on every `get_tokens` request though)

.gitignore
```diff
@@ -24,3 +24,4 @@ genesis.dat
 *.sqlite3
 *.sqlite3-shm
 *.sqlite3-wal
+db/
```
mFragaBA, line 27: we should either track the directory, make the CLI create it or mention somewhere in the documentation that the user needs to create a `db` directory. Otherwise you get an error because the path doesn't exist

faucet/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+            decimals,
+            max_supply,
+            config: faucet_config,
+        }) => {
+            config = load_config(faucet_config.as_path()).extract().map_err(|e| {
+                io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!("Failed to load configuration file: {}", e),
+                )
+            })?;
```
mFragaBA, line 75: It's possible to load the config before the match, isn't it?

faucet/src/static/index.html
```diff
@@ -0,0 +1,25 @@
...
+<body>
+    <div id="navbar">
+        <h1 id="title">Miden Faucet</h1>
+      <div id="faucetId"></div>
+    </div>
+    <div id="center-container">
+        <h2 id="subtitle">Request test POL tokens</h2>
+        <span id="error-message"></span>
+        <input type="text" id="account-id" placeholder="Hex encoded Account id" required>
+        <button id="button">Send me 100 tokens!</button>
```
mFragaBA, line 20: This should change depending on the faucet configuration I think

faucet/src/static/index.html
```diff
@@ -0,0 +1,25 @@
...
+    <link rel="stylesheet" href="./index.css">
+</head>
+
+<body>
+    <div id="navbar">
+        <h1 id="title">Miden Faucet</h1>
+      <div id="faucetId"></div>
+    </div>
+    <div id="center-container">
+        <h2 id="subtitle">Request test POL tokens</h2>
```
mFragaBA, line 17: This too

faucet/src/main.rs
phklive: Indeed.

faucet/src/config.rs
phklive: Changed the file to: 

```toml
endpoint = { protocol = "http", host = "localhost", port = 8080 }
rpc_url = "http://localhost:57291"
database_filepath = "db/miden-faucet.sqlite3"
```

faucet/miden-faucet.toml
phklive: Removed the comment.

faucet/src/main.rs
phklive: I think it should be okay on each request for now.

phklive: I don't see much value at the moment to sync it periodically.

.gitignore
phklive: Indeed. Will revert back the config file so it automatically creates it at the root.

bobbinth:
@phklive - could you add a short README to the `faucet` crate? Just something very simple which explains its purpose and how to use it.

faucet/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "miden-node-faucet"
+version = "0.1.0"
+edition = "2021"
```
bobbinth, line 4: We should do this the same way as we do it in all other crates (e.g., see [here](https://github.com/0xPolygonMiden/miden-node/blob/next/rpc/Cargo.toml)).

faucet/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
bobbinth, line 6: nit: this can probably be deleted.

faucet/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+miden-lib = { workspace = true }
+miden-client = { version = "0.1.0", features = ["concurrent"] }
+miden-node-proto = { path = "../proto", version = "0.2" }
+miden-node-utils = { path = "../utils", version = "0.2" }
+miden-objects = { workspace = true }
+serde = { version = "1.0", features = ["derive"] }
+clap = { version = "4.5.1", features = ["derive"] }
+async-mutex = "1.4.0"
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
```
bobbinth, line 27: nit: let's alphabetize the dependencies.

PR #272, opened 2024-03-11
 Add script to check rust versions + cleanup

phklive:
In this pr after @bobbinth 's review on #251 added a few improvements:
- bash script enabling the check of correct rust versions in the whole repo ( rust-toolchain and Cargo.tomls ) we will have other bash scripts in the repo in the future for end-to-end testing for example, hence I believe that it does not hurt to add it to the repo now.
- added bash script version check to ci ( will be able to add it to pre-commit in the future )
- added msrv in all cargo.toml files
- removed unnecessary comments from .gitignore file and improved formatting


Cargo.toml
```diff
@@ -9,6 +9,7 @@ members = [
   "test-macro",
 ]
 resolver = "2"
+rust-version = "1.75"
```
bobbinth, line 12: Question: since we add MSRV here, do we need to also have it in all individual `Cargo.toml` files? Or would this get inherited automatically?

scripts/check-rust-version.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Define the expected Rust version
+EXPECTED_VERSION="1.75"
```
bobbinth, line 4: Question: instead of defining the version here, is it possible to read it from the `rust-toolchain` file?

bobbinth (commented):
Looks good! Thank you! I left a couple of questions inline.

Also, should we add "Testing" section to the main readme to explain how to run tests?

Cargo.toml
phklive: It get's inherited by the crates of the workspace. I will then remove them from the individual crates. The reason to have it in individual crates would be in case you would want to override this version.

scripts/check-rust-version.sh
phklive: Yes it is possible, updated the script to do so.

phklive:
> Looks good! Thank you! I left a couple of questions inline.
> 
> Also, should we add "Testing" section to the main readme to explain how to run tests?

Added a section for testing in the Contributing.md

block-producer/Cargo.toml
```diff
@@ -1,12 +1,15 @@
 [package]
 name = "miden-node-block-producer"
-version = "0.2.0"
+version.workspace = true
```
bobbinth, line 3: I would not pull the `version` field from the workspace as it may be different for different crates. This is also relevant to other crates.

bobbinth: I would also move `description` field to be right after `version`. field.

block-producer/Cargo.toml
```diff
@@ -23,9 +26,9 @@ async-trait = { version = "0.1" }
 clap = { version = "4.3", features = ["derive"] }
 figment = { version = "0.10", features = ["toml", "env"] }
 itertools = { version = "0.12" }
-miden-node-proto = { path = "../proto", version = "0.2" }
-miden-node-store = { path = "../store", version = "0.2" }
-miden-node-utils = { path = "../utils", version = "0.2" }
+miden-node-proto = { path = "../proto" }
+miden-node-store = { path = "../store" }
+miden-node-utils = { path = "../utils" }
```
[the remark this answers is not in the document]
phklive: Just made a new PR, I made these changes before you explained to me the versioning conventions that we were adopting. it's fixed if you refresh.

block-producer/Cargo.toml
phklive: I took example on how reth does it: https://github.com/paradigmxyz/reth

they pin everything to the release of the main workspace. We can definitely do differently just tell me what you think is best.

bobbinth: > Here you mean that we could release a new version of the block-producer before releasing a new version of the store for example ?

Yes - we sometimes may do patch releases - e.g., `v0.1.1` while other crates are on `v0.1.0`.

phklive: Added the individual versioning back.

scripts/check-rust-version.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Check rust-toolchain file
+TOOLCHAIN_VERSION=$(cat rust-toolchain)
+
+# Check each Cargo.toml file
+CARGO_VERSION=$(cat Cargo.toml | grep "rust-version" | cut -d '"' -f 2)
+if [ "$CARGO_VERSION" != "$TOOLCHAIN_VERSION" ]; then
+    echo "Mismatch in $file. Expected $TOOLCHAIN_VERSION, found $CARGO_VERSION"
+    exit 1
+fi
```
bobbinth, line 11: The comment is obsolete here, right? (i.e., we don't check each `Cargo.toml.` file)

utils/Cargo.toml
```diff
@@ -1,14 +1,15 @@
 [package]
 name = "miden-node-utils"
-version = "0.2.0"
-description = "Miden node's shared utilities"
-authors = ["miden contributors"]
-readme = "README.md"
-license = "MIT"
-repository = "https://github.com/0xPolygonMiden/miden-node"
+version.workspace = true
```
bobbinth, line 3: `version` not be pulled from workspace.

bobbinth (commented):
Thank you! I left a few comments inline. These comments are applicable to most `Cargo.toml` files in this repo.

CONTRIBUTING.md
```diff
@@ -80,6 +80,13 @@ For example, a new change to the `miden-node-store` crate might have the followi
 
 You can find more information about the `cargo make` commands in the [Makefile](Makefile.toml)
 
+### Testing
+- After writing code different types of tests (unit, integration, end-to-end) are required to make sure that the correct behavior has been achieved and that no bugs have been introduced. You can run tests using the following command:
```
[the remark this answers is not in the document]
phklive: We don't specifically need to but it's the format that has been using throughout the file, hence I copied it.

We could change it but then it would meed reformatting the whole file to make styling consistent and make sure that code blocks are shown correctly. ( I can't just remove the bullet points, it would break the code blocks ).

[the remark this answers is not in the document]
bobbinth: I think in other sections the bullet points make sense as there are multiple bullet points. In this section, this is not really a bullet point but a single paragraph - so, I think removing bullet point from here is fine (i.e., we don't need to reformat anything else).

CONTRIBUTING.md
```diff
@@ -80,6 +80,13 @@ For example, a new change to the `miden-node-store` crate might have the followi
 
 You can find more information about the `cargo make` commands in the [Makefile](Makefile.toml)
 
+### Testing
+After writing code different types of tests (unit, integration, end-to-end) are required to make sure that the correct behavior has been achieved and that no bugs have been introduced. You can run tests using the following command:
+
+    ```
+    cargo make test
+    ```
```
bobbinth, line 88: nit: we probably should decrease the indent here.

bobbinth (approved):
Looks good! Thank you! I left one minor nit inline. Once this is addressed, we can merge.

PR #275, opened 2024-03-12
Implemented nullifier tree wrapper over `Smt`

polydez:
Resolves: https://github.com/0xPolygonMiden/miden-node/issues/269

### DIFF
--- a/store/src/errors.rs
+++ b/store/src/errors.rs
@@ -6,11 +6,28 @@ use miden_objects::{
         merkle::{MerkleError, MmrError},
         utils::DeserializationError,
     },
-    AccountError, BlockHeader, Digest,
+    notes::Nullifier,
+    AccountError, BlockHeader,
 };
 use rusqlite::types::FromSqlError;
 use thiserror::Error;
 use tokio::sync::oneshot::error::RecvError;
+
+use crate::types::BlockNumber;
+
+// INTERNAL ERRORS
+// =================================================================================================
+
+#[derive(Debug, Error)]
+pub enum NullifierTreeError {
+    #[error("Merkle error: {0}")]
+    MerkleError(#[from] MerkleError),
+    #[error("Nullifier {nullifier} for block #{block_num} already exists in the nullifier tree")]
+    NullifierAlreadyExists {
+        nullifier: Nullifier,
+        block_num: BlockNumber,
+    },
+}
 
 // DATABASE ERRORS
 // =================================================================================================
@@ -41,7 +58,7 @@ pub enum StateInitializationError {
     #[error("Database error: {0}")]
     DatabaseError(#[from] DatabaseError),
     #[error("Failed to create nullifiers tree: {0}")]
-    FailedToCreateNullifiersTree(MerkleError),
+    FailedToCreateNullifiersTree(NullifierTreeError),
     #[error("Failed to create accounts tree: {0}")]
     FailedToCreateAccountsTree(MerkleError),
 }
@@ -106,7 +123,7 @@ pub enum ApplyBlockError {
     #[error("Received invalid nullifier root")]
     NewBlockInvalidNullifierRoot,
     #[error("Duplicated nullifiers {0:?}")]
-    DuplicatedNullifiers(Vec<Digest>),
+    DuplicatedNullifiers(Vec<Nullifier>),
     #[error("Unable to create proof for note: {0}")]
     UnableToCreateProofForNote(MerkleError),
     #[error("Block applying was broken because of closed channel on database side: {0}")]
@@ -117,6 +134,8 @@ pub enum ApplyBlockError {
     DbBlockHeaderEmpty,
     #[error("Failed to get MMR peaks for forest ({forest}): {error}")]
     FailedToGetMmrPeaksForForest { forest: usize, error: MmrError },
+    #[error("Failed to update nullifier tree: {0}")]
+    FailedToUpdateNullifierTree(NullifierTreeError),
 }
 
 #[derive(Error, Debug)]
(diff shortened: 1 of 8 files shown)


store/src/errors.rs
```diff
@@ -41,7 +58,7 @@ pub enum StateInitializationError {
     #[error("Database error: {0}")]
     DatabaseError(#[from] DatabaseError),
     #[error("Failed to create nullifiers tree: {0}")]
-    FailedToCreateNullifiersTree(MerkleError),
+    FailedToCreateNullifiersTree(NullifierTreeError),
```
bobbinth, line 61: minor nit: I'd use `FailedToCreateNullifierTree` (i.e., `Nullifier` without the `s`).

store/src/server/api.rs
```diff
@@ -388,10 +388,10 @@ fn invalid_argument<E: core::fmt::Debug>(err: E) -> Status {
 }
 
 #[instrument(target = "miden-store", skip_all, err)]
-fn validate_nullifiers(nullifiers: &[generated::digest::Digest]) -> Result<Vec<RpoDigest>, Status> {
+fn validate_nullifiers(nullifiers: &[generated::digest::Digest]) -> Result<Vec<Nullifier>, Status> {
     nullifiers
         .iter()
-        .map(TryInto::try_into)
+        .map(|digest_pb| Ok(Nullifier::from(RpoDigest::try_from(digest_pb)?)))
```
bobbinth, line 394: Instead of going through `RpoDigest` - should we just implement `From<digest::Digest> for Nullifier` in the `miden-node-proto` crate?

store/src/state.rs
```diff
@@ -39,9 +39,85 @@ pub struct TransactionInputs {
     pub nullifiers: Vec<NullifierInfo>,
 }
 
+/// Nullifier SMT.
+#[derive(Debug, Clone)]
+struct NullifierTree(Smt);
+
+impl NullifierTree {
```
bobbinth, line 46: I would move `NullifierTree` into a separate file.

store/src/state.rs
```diff
@@ -39,9 +39,85 @@ pub struct TransactionInputs {
...
+        [Felt::from(block), Felt::ZERO, Felt::ZERO, Felt::ZERO]
+    }
+
+    /// Given the leaf value of the nullifier SMT, returns the nullifier's block number.
+    ///
+    /// There are no nullifiers in the genesis block. The value zero is instead used to signal
+    /// absence of a value.
+    fn nullifier_value_to_block_num(value: Word) -> BlockNumber {
+        value[0].as_int().try_into().expect("invalid block number found in store")
+    }
```
bobbinth, line 115: nit: I don't think "nullifier_value" is the right word here (block number is not the value of a nullifier). I'm thinking better names for these functions would be `block_num_to_leaf_value` and `leaf_value_to_block_num`.

bobbinth (changes requested):

store/src/server/api.rs
polydez: We actually have this implemented, my bad. Thanks for noticing!

store/src/nullifier_tree.rs
```diff
@@ -0,0 +1,110 @@
...
+    ) -> Result<(), NullifierTreeError> {
+        let value = self.0.insert(nullifier.inner(), Self::block_num_to_leaf_value(block_num));
+        if value == EMPTY_WORD {
+            return Ok(());
+        }
+
+        Err(NullifierTreeError::NullifierAlreadyExists {
+            nullifier: *nullifier,
+            block_num: Self::leaf_value_to_block_num(value),
+        })
```
hackaugusto, line 57: This error is misleading. It seems to imply the `insert` failed, because the nullifier already exists. But the existing nullifier is being replaced, since `insert` behaves like an `insert_or_replace`.

What was the intention here? To me, it seems that if we want the same behaviour, we should not have this error. OR, if we want to `Err`, then we have to do a check before the insert.

store/src/nullifier_tree.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// Returns block number stored for the given nullifier or `None` if the nullifier wasn't
+    /// consumed.
+    pub fn get_block_num(
+        &self,
+        nullifier: &Nullifier,
+    ) -> Option<BlockNumber> {
+        let value = self.0.get_value(&nullifier.inner());
+        if value == EMPTY_WORD {
+            return None;
+        }
```
hackaugusto, line 69: This feels like a design flaw in `SparseMerkleTree`. We are required to encode the lack of a value as an `EMPTY_WORD`, but that implementation detail is leaking, and now on higher level abstractions we are having to special case it.

https://github.com/0xPolygonMiden/crypto/issues/289

hackaugusto (approved):
LGTM, I left one nit about the insert API that would be nice to iron out though.

store/src/nullifier_tree.rs
plafer: Building on https://github.com/0xPolygonMiden/crypto/issues/289#issuecomment-1997266377, the correct check here would be `if value == Smt::EMPTY_VALUE`, since `Smt::EMPTY_VALUE` is part of the public API, but the fact that `Smt::EMPTY_VALUE == EMPTY_WORD` is an implementation detail.

store/src/nullifier_tree.rs
bobbinth: Oh! Great catch! Yes, definitely we should do a check and only then do the insert. Basically, if the error is returned, the state of the nullifier tree should not change.

store/src/nullifier_tree.rs
bobbinth: Yes, let's change this to `if value == Smt::EMPTY_VALUE`.

bobbinth (approved):
Looks good! Thank you!

Issue #268, opened 2024-03-11
[Feature]: Separate CI into multiple files 


phklive:
### To whom is this feature for?

App developer

### Why is is this feature needed?

With the growth of the tasks that we need to perform in the CI it would be good to have multiple files that manage these tasks

### How is this feature used?

Instead of having 1 CI file we will have multiple files like lint, integration, end-to-end, docker, etc...

PR #278, opened 2024-03-14
Separate CI into multiple files

phklive:
In this PR I propose the separation of the CI jobs into multiple files enabling more scaling and better separation of concerns.

closes: #268


polydez (approved):
Looks good, thanks! How about running these tasks not only for `main`, but also for `next` branch?

bobbinth:
Question: how would this affect the CI badge that we have in the main readme file (see [here](https://github.com/0xPolygonMiden/miden-node/blob/main/README.md?plain=1#L4)).

phklive:
> Question: how would this affect the CI badge that we have in the main readme file (see [here](https://github.com/0xPolygonMiden/miden-node/blob/main/README.md?plain=1#L4)).

Changed the badge from CI to UNIT test to make sure it is still relevant. 

In the future we can or: 
- Remove the badge
- Add more badges ( lint, doc, docker, ... )

Fun fact, hovered around the reth codebase and saw that their badge was also broken haha: https://github.com/paradigmxyz/reth/issues/7158

bobbinth:
> Changed the badge from CI to UNIT test to make sure it is still relevant.

Is there a way to change the label on the badge to `tests`?

Not sure if we need to add lint and doc badges - but maybe some others would be interesting in the future (e.g., `docker` could be interesting). But I do think `tests` we should keep for sure.

phklive:
> > Changed the badge from CI to UNIT test to make sure it is still relevant.
> 
> Is there a way to change the label on the badge to `tests`?
> 
> Not sure if we need to add lint and doc badges - but maybe some others would be interesting in the future (e.g., `docker` could be interesting). But I do think `tests` we should keep for sure.

Changed it to `test` to keep naming convention with the other files being singular.

Issue #92, opened 2023-12-09
feature flags to disable/enable services


hackaugusto:
we have implemented a node binary, that runs all 3 services (`rpc`, `block_producer`, and `store`) in the same process. this simplifies a lot of things including stdin/stdout/stderr, process management, shutdown, testing.

now that we have a single binary, we can add flags or subcommands to control which services are executed, e.g. `miden-node services run`, `miden-node store run`, `miden-store rpc run`. 

This approach simplifies a few things:

- deployment, specially with static linking since a single binary can be copied to multiple machines, and only the process management config need to change to choose which service is executed.
- compilation, since we don't have to compile 3 binaries, but a single one, we should get a small speed up
- testing, since we no longer have to add infrastructure to build and test multiple binaries
- user onboarding, we can do releases of a single binary per platform, and an operator can just download that to do testing

plafer:
Relating to #100, this means that we only ever need just one config file. If someone wants to run e.g. only the `store`, then they only need to have the `[store]` header in their config.

bobbinth:
Closed by #323.

Kosi23:
Hi @Mirko-von-Leipzig — I’d like to take a crack at this one if it’s still open.
My read of the issue plus the discussion thread on #1846: the asymmetry between SubmitProvenBatch (which validates the reference block via a store lookup) and SubmitProvenTransaction (which delegates to the block-producer) means clients get different error semantics for the same bad input, and every batch submission round-trips to the store even though reference blocks cluster heavily around the tip.
Plan I’d propose:
	1.	Add a small LRU header cache in the RPC component (sized to recent blocks — capacity configurable, sensible default), wrapping the existing SubmitProvenBatch reference-block lookup so misses fall through to the store and populate the cache.
	2.	Move the same reference-block check onto SubmitProvenTransaction’s RPC path, returning the same error variant SubmitProvenBatch already returns so clients see consistent semantics.
	3.	Tests covering hit/miss/eviction on the cache and an integration test asserting both endpoints reject a non-existent reference block uniformly.
Two questions before I start:
	•	Any preference between lru and a hand-rolled bounded cache here? I noticed the codebase already pulls in parking_lot so I’d default to a parking_lot::Mutex<LruCache<_, _>> unless the lookup path is fully async, in which case tokio::sync::Mutex.
[...]

PR #279, opened 2024-03-14
Add docker-compose.yml

phklive:
In this PR I propose to add a full Docker compose to handle the spin up of 3 Docker containers managing respectively:
- block-producer
- rpc
- store

This will enable an easy spin-up and management of the different docker containers.

Closes: #93


bobbinth:
Should we put this on-hold for now? My reasoning is as follows:

- I don't think this is a high priority issue (i.e., we are not yet running the components independently).
- How we run them may change depending on how we resolve of #92.

phklive:
> Should we put this on-hold for now? My reasoning is as follows:
> 
> * I don't think this is a high priority issue (i.e., we are not yet running the components independently).
> * How we run them may change depending on how we resolve of [feature flags to disable/enable services #92](https://github.com/0xPolygonMiden/miden-node/issues/92).

Impressive. This is exactly the discussion we were having with @hackaugusto. 

This PR is now on hold. Will focus further down the line on the refactor of the node to go from 4 to 1 binary.

Anyway have been able to test the node as separate components each in their own Docker container and it worked pretty well.

PR #280, opened 2024-03-19
chore: update archived action on ci

mFragaBA:
[As it was done for the miden-client workflows](https://github.com/0xPolygonMiden/miden-client/pull/222/files), this PR replaces the usage of `actions-rs/toolchain` which is archived as of now with the updated `dtolnay/rust-toolchain` action.


phklive (approved):
Thanks, LGTM!

PR #282, opened 2024-03-20
Add end-to-end testing workflow

phklive:
In this PR I add end-to-end tests for the Miden node.

closes: #222


Dominik1999:
Should we merge this?

phklive:
> Should we merge this?

I just want to go over, cleanup and should be ready for review & merge.

.github/workflows/test.yml
```diff
@@ -24,3 +24,25 @@ jobs:
...
+      - name: Install cargo make
+        run: cargo install cargo-make
+      - name: cargo make - docker-build-node
+        run: cargo make docker-build-node
+      - name: cargo make - docker-run-node
+        run: cargo make docker-run-node
+      - name: cargo install - miden-client
+        run: cargo install miden-client --features concurrent,testing
+      - name: cargo make - test-end-to-end
+        run: cargo make test-end-to-end
```
bobbinth, line 48: Question: is the docker image built with `concurrent,testing` by default? Should we specify these features explicilty?

.github/workflows/test.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   unit-and-integration:
-    name: test stable on ubuntu-latest
+    name: unit and integration tests stable on ubuntu-latest
```
bobbinth, line 13: minor nit: I'd probably still keep it as "test stable on ubuntu-latest"

bobbinth (commented):
Thank you! Looks great! I did a very quick review and left a couple of nits/questions inline. I defer a more detailed review to @Dominik1999 or @hackaugusto.

One note for the future: these end-to-end tests will work only when node and client versions are in sync. But we won't be able to use them when the node has some breaking changes which haven't been propagated to the client yet. To address this, we'll need to create a separate "test runner" which will not rely on `miden-client`. Let's create an issue for this.

scripts/test-end-to-end.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
hackaugusto, line 1: Run the binary from the env:
```suggestion
#!/usr/bin/env bash
```
For bash this is not that big of a deal I guess, but it is better to use the system's env path instead of assuming hardcoding it.

scripts/test-end-to-end.sh
```diff
@@ -0,0 +1,43 @@
...
+
```
hackaugusto, line 2: I find shell scripts super annoying to debug. So I think we should always have the following flags enabled:

```
set -o errexit
set -o pipefail
```

docs are here: https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin

scripts/test-end-to-end.sh
```diff
@@ -0,0 +1,43 @@
...
+miden-client account new fungible-faucet --token-symbol JACK --decimals 3 --max-supply 1000
+
+# Capture IDs
+basic_account=$(miden-client account -l | grep -oE ' 0x[a-fA-F0-9]{16} ' | head -n 1)
+faucet_account=$(miden-client account -l | grep -oE ' 0x[a-fA-F0-9]{16} ' | head -n 2 | tail -n 1)
+
+# Sync client and create tx between faucet and account
+miden-client sync
+miden-client tx new mint $basic_account $faucet_account 777
+sleep 15
```
hackaugusto, line 26: why is this needed, can we get rid of this? `sleep` in tests is to me a big no-no.

Makefile.toml
```diff
@@ -48,6 +48,12 @@ dependencies = [
 command = "cargo"
 args = ["test", "--all-features", "--workspace", "--", "--nocapture"]
 
+[tasks.test-end-to-end]
+workspace = false
+script = '''
+./scripts/test-end-to-end.sh
```
hackaugusto, line 54: I usually think of E2E as a category of tests, and not a single test. I would call this more of a smoketest. But eventually we will need something more robust to run additional tests.

hackaugusto (approved):
LGTM

.github/workflows/test.yml
phklive: Indeed the docker image is built with the `concurrent,testing` features enabled by default.

Could you clarify by specify these features explicitly?

phklive:
> Thank you! Looks great! I did a very quick review and left a couple of nits/questions inline. I defer a more detailed review to @Dominik1999 or @hackaugusto.
> 
> One note for the future: these end-to-end tests will work only when node and client versions are in sync. But we won't be able to use them when the node has some breaking changes which haven't been propagated to the client yet. To address this, we'll need to create a separate "test runner" which will not rely on `miden-client`. Let's create an issue for this.

Created here: https://github.com/0xPolygonMiden/miden-node/issues/291

Makefile.toml
phklive: The idea is that that the bash scripts can be extended with other flows enabling more testing opportunities. For now there is only 1 indeed.

scripts/test-end-to-end.sh
phklive: The sleep is needed to leave some time for the rollup to include the transaction in a block and the block to be applied to the db. If not then the note does not have a commit height and it cannot be consumed by the Miden client. 

Do you see a better solution than sleeping?

Makefile.toml
hackaugusto: What I was trying to say with "something more robust" is that shell script is not a good language to write E2E tests. It works for a single test like this, which does a good job of covering the happy path and being a smoke test. But once we have a handful of tests we should be able to do other things, for example:

- have test selection, similar to how we do `cargo test -- <test_name>`
- have tests run in parallel, similar to how `cargo test` uses runners
- have the ability to debug the test scripts, and here shell scripts really lose the upper hand

To get the above, we will eventually have to translate the `.sh` into a `.rs` or `.py`. But I don't think we have to do this now.

scripts/test-end-to-end.sh
hackaugusto: Polling would be the best option here. Probably the miden-client should have a command similar to `miden-client wait-for-note <note_id>` where the client would poll until the note_id is available.

igamigo: A couple of other ways to implement this now:

- Wait for two blocks to be created before moving on (`miden-client info` returns the current block number; should be fairly easy to parse the response and sync until you are one or two blocks above the original count)
- Wait until the transaction has been committed (although there is not any super simple way to programmatically get this from the client, but you can use `tx list` and check that the transaction ID is now committed)

.github/workflows/test.yml
bobbinth: On line 46 we explicitly build `miden-client` with `concurrent,testing` features. Could we do something similar on line 42?

The main motivation is to make sure that that changing the defaults for docker build does not break this task.

phklive:
This PR does not encompass what we need for end-to-end testing on the Miden node. Some of these tests are already covered by the Miden client. We will implement more robust end-to-end testing when the Miden node and general Miden crates are ready.

PR #283, opened 2024-03-21
bugfix: table name is block_headers instead of block_header

hackaugusto:
The table in the references clause had a typo.


phklive (approved):
Indeed should be `block_headers` and not `block_header`.

Great job and great eyes as always.

PR #284, opened 2024-03-21
sqlite: use bundled version to enable FK checks

hackaugusto:
I'm not sure why the PRAGMA wasn't sufficient. But this was caught by @phklive when integrating the faucet. The feature flag seems to be required

### DIFF
--- a/store/Cargo.toml
+++ b/store/Cargo.toml
@@ -30,7 +30,7 @@ miden-node-utils = { path = "../utils", version = "0.2" }
 miden-objects = { workspace = true }
 once_cell = { version = "1.18.0" }
 prost = { version = "0.12" }
-rusqlite = { version = "0.30", features = ["array", "buildtime_bindgen"] }
+rusqlite = { version = "0.30", features = ["array", "buildtime_bindgen", "bundled"] }
 rusqlite_migration = { version = "1.0" }
 serde = { version = "1.0", features = ["derive"] }
 thiserror = { workspace = true }
--- a/store/src/db/migrations.rs
+++ b/store/src/db/migrations.rs
@@ -25,9 +25,9 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
             merkle_path BLOB NOT NULL,
 
             PRIMARY KEY (block_num, note_index),
+            CONSTRAINT fk_block_num FOREIGN KEY (block_num) REFERENCES block_headers (block_num),
             CONSTRAINT notes_block_number_is_u32 CHECK (block_num >= 0 AND block_num < 4294967296),
-            CONSTRAINT notes_note_index_is_u32 CHECK (note_index >= 0 AND note_index < 4294967296),
-            FOREIGN KEY (block_num) REFERENCES block_headers (block_num)
+            CONSTRAINT notes_note_index_is_u32 CHECK (note_index >= 0 AND note_index < 4294967296)
         ) STRICT, WITHOUT ROWID;
 
         CREATE TABLE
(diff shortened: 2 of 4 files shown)


store/Cargo.toml
```diff
@@ -30,7 +30,7 @@ miden-node-utils = { path = "../utils", version = "0.2" }
 miden-objects = { workspace = true }
 once_cell = { version = "1.18.0" }
 prost = { version = "0.12" }
-rusqlite = { version = "0.30", features = ["array", "buildtime_bindgen"] }
+rusqlite = { version = "0.30", features = ["array", "buildtime_bindgen", "bundled"] }
```
bobbinth, line 33: Should we bump-up the version to v0.31 (I think that's the latest one now).

hackaugusto: Not for the moment, Deadpool has the SQLite version pinned

phklive, line 33: This is the feature used by the Miden client. I am curious if the bug could be related, needing to have the same features enabled on all crates.

store/src/db/migrations.rs
```diff
@@ -50,16 +50,16 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
...
+            CONSTRAINT nullifiers_block_number_is_u32 CHECK (block_number >= 0 AND block_number < 4294967296)
         ) STRICT, WITHOUT ROWID;
         ",
     )])
 });
 
 #[test]
 fn migrations_test() {
-    assert!(MIGRATIONS.validate().is_ok());
+    assert_eq!(MIGRATIONS.validate(), Ok(()));
```
phklive, line 64: Why is this change needed? Isn't it equivalent?

store/src/db/tests.rs
```diff
@@ -18,12 +18,36 @@ fn create_db() -> Connection {
...
+    let block_header = BlockHeader::new(
+        num_to_rpo_digest(1),
+        block_num,
+        num_to_rpo_digest(3),
+        num_to_rpo_digest(4),
+        num_to_rpo_digest(5),
+        num_to_rpo_digest(6),
+        num_to_rpo_digest(7),
+        num_to_rpo_digest(8),
+        9_u8.into(),
```
phklive, line 34: Why are we using 9_u8.into() here instead of num_to_rpo_digest ?

store/src/db/tests.rs
```diff
@@ -382,6 +416,9 @@ fn test_db_account() {
 fn test_notes() {
     let mut conn = create_db();
 
+    let block_num1 = 1;
```
phklive, line 419: Would the idiomatic way to write this var name be `block_num_1` ?

store/src/db/tests.rs
```diff
@@ -434,14 +470,17 @@ fn test_notes() {
         &mut conn,
         &[(tag >> 48) as u32],
         &[],
-        block_num - 1,
+        block_num1 - 1,
     )
     .unwrap();
     assert_eq!(res, vec![note.clone()]);
 
+    let block_num2 = note.block_num + 1;
```
phklive, line 478: Same here regarding idiomatic naming.

phklive (approved):
Added a few questions and nits, overall LGTM.

store/src/db/migrations.rs
hackaugusto: the result of `validate()` is a bool, that erases the error message. Comparing against `Ok(())` preserver the sqlite3 error if the test fails

store/Cargo.toml
hackaugusto: which bug do you mean? This by itself wasn't a bug in itself, but it allowed one to slip through the cracks (the name mismatch on the other PR).

store/src/db/tests.rs
hackaugusto: I _think_ we don't have a `Into` conversion from u8 to RpoDigest. I think this is just wrapping the type to create the block header

phklive: What do you call `wrapping` ?

store/src/db/tests.rs
phklive: I prefer it the idiomatic way personally.

store/src/db/tests.rs
hackaugusto: when you have a unit type struct, wrapping an inner type.

store/src/db/tests.rs
hackaugusto: changed

PR #287, opened 2024-03-26
On-chain accounts protobuf & domain objects

polydez:
This PR is intended to be 1 of 3 PRs regarding support of on-chain (public) accounts in Miden node. This PR includes only implementation of protobuf and domain objects for on-chain accounts.


proto/proto/account.proto
```diff
@@ -15,3 +15,86 @@ message AccountInfo {
...
+    optional uint64 nonce = 3;
+}
+
+message AccountDetails {
+    // Details enumeration for public accounts.
+    oneof details {
+        account.AccountFullDetails full = 1;
+        account.AccountDelta delta = 2;
+    }
+}
```
bobbinth, line 100: I'm wondering: do we actually need to provide such granular messages for all underlying sub-components of `AccountDetails`? Maybe we should just use `bytes` for it? Or do we lose some functionality this way?

polydez: @bobbinth as for me, the only reason we need it is using protobuf as communication protocol. Otherwise, all our communication would devolute to just `data: bytes`. At the moment, we deserialize the whole protobuf message at once and don't need to read just separate fields.

proto/src/domain/notes.rs
```diff
@@ -25,7 +25,7 @@ impl From<(u64, NoteEnvelope)> for note::NoteCreated {
         Self {
             note_id: Some(note.note_id().into()),
             sender: Some(note.metadata().sender().into()),
-            tag: note.metadata().tag().into(),
+            tag: u32::from(note.metadata().tag()).into(),
```
hackaugusto, line 28: I think we should add the missing conversions to `miden-base`.

polydez: @hackaugusto nice idea, done: https://github.com/0xPolygonMiden/miden-base/pull/559

bobbinth: Eventually (in a different PR), tag here would be `u32` as well, right

proto/proto/account.proto
bobbinth: I don't think we need to move everything to bytes, but I do wonder how useful it is to add all this code - especially if we don't really use it. Maybe for now we just use bytes here and change this to more detailed messages if the need arises.

polydez: @bobbinth okay, I've got your point, I will remove all unnecessary protobuf messages relative to account details/deltas and use just bytes here.

PR #292, opened 2024-03-29
Refactor node from 4 to 1 binary and add feature flags

phklive:
In this PR I propose the refactor of the node internal structure to enable/disable services using flags and reduce from 4 to 1 binary.

closes: #92


Cargo.toml
```diff
@@ -1,14 +1,15 @@
...
-  "faucet"
+  "bin/node",
+  "crates/block-producer",
+  "crates/proto",
+  "crates/rpc",
+  "crates/faucet",
+  "crates/store",
+  "crates/utils",
+  "crates/test-macro",
 ]
```
bobbinth, line 11: What is the motivation for creating this structure?

Also, if we go with it, shouldn't `faucet` go under `bin` too?

phklive: It clearly separates the binary and it's components ( crates ) making the navigation of the repo clearer. This structured is inspired by RETH: https://github.com/paradigmxyz/reth

imo what I will do in this PR is to refactor the faucet to become a lib of the Node as with the other components ( block-producer, rpc, store ) and enable/disable the faucet using commang line arguments, EX: 

`cargo run start node --faucet`

PR #293, opened 2024-03-30
Added support for on-chain (public) accounts in store

polydez:
In this PR database structure updated and only read-only access is implemented. Inserting of account details into the database is implemented in the next PR with block applying.


rpc/src/server/api.rs
```diff
@@ -132,4 +132,21 @@ impl api_server::Api for RpcApi {
 
         self.block_producer.clone().submit_proven_transaction(request).await
     }
+
+    /// Returns details for public (on-chain) account by id.
+    #[instrument(
+        target = "miden-rpc",
```
hackaugusto, line 138: can't we use the `COMPONENT` constant here?

store/src/db/migrations.rs
```diff
@@ -39,21 +39,35 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
...
+        (
+            account_id INTEGER NOT NULL,
+            nonce INTEGER NOT NULL,
+            vault BLOB NOT NULL,
+            storage BLOB NOT NULL,
+            code BLOB NOT NULL,
+
+            PRIMARY KEY (account_id),
+            CONSTRAINT account_details_nonce_non_negative CHECK (nonce >= 0),
+            FOREIGN KEY (account_id) REFERENCES accounts (account_id)
```
hackaugusto, line 56: Please give it a contraint name, this makes the error messages more readable.

store/src/db/sql.rs
```diff
@@ -113,6 +115,35 @@ pub fn select_accounts_by_block_range(
...
+    account_id: AccountId,
+) -> Result<Account> {
+    let mut stmt = conn.prepare(
+        "SELECT nonce, vault, storage, code FROM account_details WHERE account_id = ?1;",
+    )?;
+
+    let mut rows = stmt.query(params![u64_to_value(account_id)])?;
+    let Some(row) = rows.next()? else {
+        return Err(DatabaseError::AccountNotOnChain(account_id));
+    };
```
hackaugusto, line 134: Nit: I think this is a little more readable:

```suggestion
    let row = rows.next()?.ok_or(Err(DatabaseError::AccountNotOnChain(account_id)))?;
```

rpc/src/server/api.rs
```diff
@@ -132,4 +132,21 @@ impl api_server::Api for RpcApi {
...
+        ret(level = "debug"),
+        err
+    )]
+    async fn get_public_account_details(
+        &self,
+        request: tonic::Request<GetPublicAccountDetailsRequest>,
+    ) -> std::result::Result<Response<GetPublicAccountDetailsResponse>, Status> {
+        debug!(target: COMPONENT, request = ?request.get_ref());
+
+        self.store.clone().get_public_account_details(request).await
```
hackaugusto, line 150: I think we should validate the account_id here

hackaugusto (approved):
LGTM, left only a few nits.

rpc/src/server/api.rs
polydez: Done

store/src/db/sql.rs
```diff
@@ -113,6 +115,33 @@ pub fn select_accounts_by_block_range(
     Ok(result)
 }
 
+/// Select the latest account details by account id from the DB using the given [Connection].
+///
+/// # Returns
+///
+/// The latest account details, or an error.
+pub fn get_account_details(
```
hackaugusto, line 123: Nit: The functions that do a single `SELECT` are named with `select_`. Maybe lets rename it so that it is consistent?

(the `get_state_sync` is named so because it does much more work that a single `SELECT`)

hackaugusto (approved):
LGTM

rpc/src/server/api.rs
```diff
@@ -132,4 +133,30 @@ impl api_server::Api for RpcApi {
...
+        err
+    )]
+    async fn get_public_account_details(
+        &self,
+        request: tonic::Request<GetPublicAccountDetailsRequest>,
+    ) -> std::result::Result<Response<GetPublicAccountDetailsResponse>, Status> {
+        debug!(target: COMPONENT, request = ?request.get_ref());
+
+        // Validating account using conversion:
+        let _account_id: AccountId = request
```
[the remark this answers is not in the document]
polydez: @igamigo nice catch, thanks!

store/src/errors.rs
```diff
@@ -46,8 +47,32 @@ pub enum DatabaseError {
...
     #[error("Block applying was broken because of closed channel on state side: {0}")]
     ApplyBlockFailedClosedChannel(RecvError),
+    #[error("Public account ({0}) details not found on-chain")]
+    AccountNotOnChain(AccountId),
+    #[error("Failed to apply block because of on-chain account final hashes mismatch (expected {expected}, \
+        but calculated is {calculated}")]
+    ApplyBlockFailedAccountHashesMismatch {
+        expected: RpoDigest,
+        calculated: RpoDigest,
+    },
```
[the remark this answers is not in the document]
polydez: @hackaugusto moved it to the next PR where it should be introduced, thank you!

store/src/db/migrations.rs
```diff
@@ -39,21 +39,35 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
...
+            account_id INTEGER NOT NULL,
+            nonce INTEGER NOT NULL,
+            vault BLOB NOT NULL,
+            storage BLOB NOT NULL,
+            code BLOB NOT NULL,
+
+            PRIMARY KEY (account_id),
+            CONSTRAINT account_details_nonce_non_negative CHECK (nonce >= 0),
+            CONSTRAINT fk_account_id FOREIGN KEY (account_id) REFERENCES accounts (account_id)
         ) STRICT, WITHOUT ROWID;
```
bobbinth, line 57: I thought we were going to replace this with `details` field in the `accounts` table. Or did you decide to keep it as a separate table?

store/src/server/api.rs
```diff
@@ -159,6 +159,32 @@ impl api_server::Api for StoreApi {
...
+            .get_account_details(
+                request.account_id.ok_or(invalid_argument("Account missing id"))?.into(),
+            )
+            .await
+            .map_err(internal_error)?;
+
+        Ok(Response::new(GetPublicAccountDetailsResponse {
+            details: account.to_bytes(),
+        }))
+    }
```
bobbinth, line 186: Here, we should probably check if the the request is made for a public account (this can be determined by account ID) and return an error if not.

But also, see overall review comment - maybe we should make this endpoint a bit more general.

bobbinth (commented):
Thank you! Looks good! I left a couple of comments inline.

Also, I'm wondering if we should make this endpoint a bit more general. Specifically, instead `GetPublicAccountDetails` it could be just `GetAccountDetails` and it would work as follows:

- For off-chain accounts, it would return just the hash.
- For public accounts, it would return the hash and the details.

So, the `GetAccountDetailsResponse` message would look like so:
```proto
message GetAccountDetailsResponse {
    digest.Digest account_hash = 1;
    bytes details = 2;
}
```
Handling of this in the store should be pretty simple: based on the incoming account ID we can figure out whether the request is being made for an off-chain or public account and then build the response accordingly.

store/README.md
```diff
@@ -136,6 +136,18 @@ contains excessive notes and nullifiers, client can make additional filtering of
 * `notes`: `[NoteSyncRecord]` – a list of all notes together with the Merkle paths from `block_header.note_root`.
 * `nullifiers`: `[NullifierUpdate]` – a list of nullifiers created between `block_num + 1` and `block_header.block_num`.
 
+### GetPublicAccountDetails
```
phklive, line 139: You added the documentation to the `store` but not to the `rpc` is that voluntary?

store/src/db/migrations.rs
polydez: @bobbinth yes, as we decided earlier, I will change DB structure and remove this table.

store/README.md
polydez: This endpoint was removed later.

store/src/server/api.rs
polydez: Finally I renamed it to `GetAccountDetails` and it supports all account types, so it doesn't validate account type.

store/src/db/sql.rs
```diff
@@ -82,7 +87,10 @@ pub fn select_accounts_by_block_range(
     let mut stmt = conn.prepare(
         "
         SELECT
-            account_id, account_hash, block_num
+            account_id,
+            account_hash,
+            block_num,
+            details
```
bobbinth, line 93: This is probably fine for now, but this function is used to build a `sync_state` response where account details are not needed. So, we are doing potentially quite a bit of extra work here by fetching details from the database and parsing them.

Let's create an issue to address this in the future.

bobbinth (approved):
Looks good! Thank you! I left a few comments inline. Once these are addressed, we can merge.

store/src/db/sql.rs
polydez: I refactored this, now there are two different structures (`AccountHashUpdate` is reused), so details aren't provided here anymore.

proto/proto/account.proto
```diff
@@ -10,8 +10,13 @@ message AccountId {
...
+    account.AccountId account_id = 1;
     digest.Digest account_hash = 2;
-    fixed32 block_num = 3;
+    uint32 block_num = 3;
+}
+
+message AccountInfo {
+    AccountHashUpdate update = 1;
+    optional bytes details = 2;
 }
```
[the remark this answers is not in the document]
polydez: We already have `AccountDetails` structure in miden-objects, maybe need to end up with different name.

PR #294, opened 2024-03-30
Implemented writing on-chain accounts details on block applying

polydez:
This is the last PR of this series. Here is details of on-chain accounts is written to database.


block-producer/src/batch_builder/batch.rs
```diff
@@ -108,12 +110,14 @@ impl TransactionBatch {
...
-    /// Returns an iterator over (account_id, new_state_hash) tuples for accounts that were
+    /// Returns an iterator over (account_id, details, new_state_hash) tuples for accounts that were
     /// modified in this transaction batch.
-    pub fn updated_accounts(&self) -> impl Iterator<Item = (AccountId, Digest)> + '_ {
-        self.updated_accounts
-            .iter()
-            .map(|(account_id, account_states)| (*account_id, account_states.final_state))
+    pub fn updated_accounts(
+        &self
+    ) -> impl Iterator<Item = (AccountId, Option<AccountDetails>, Digest)> + '_ {
```
hackaugusto, line 117: Can you add a struct type for the return tuple? It should make the code a bit easier to read.

block-producer/src/state_view/mod.rs
```diff
@@ -121,24 +121,23 @@ where
         &self,
         block: Block,
     ) -> Result<(), ApplyBlockError> {
-        self.store.apply_block(block.clone()).await?;
+        let account_ids_in_block: Vec<AccountId> =
+            block.updated_accounts.iter().map(|(account_id, _, _)| *account_id).collect();
+        let produced_nullifiers = block.produced_nullifiers.clone();
+
+        self.store.apply_block(block).await?;
```
hackaugusto, line 128: I assume that `store.apply_block` encodes the `Block` into a protobuf message. Meaning it doesn't need ownership of `Block`. If that is the case, I think we should change it to receive a reference, and then here we don't need the copies.

proto/proto/responses.proto
```diff
@@ -113,3 +113,8 @@ message ListNotesResponse {
     // Lists all notes of the current chain
     repeated note.Note notes = 1;
 }
+
+message GetPublicAccountDetailsResponse {
+    // Public account full details.
```
hackaugusto, line 118: Please add a bit of documentation explaining what is the type of the encoded data. Just enough context for someone that will use this type to know how to decode it.

store/src/db/sql.rs
```diff
@@ -125,16 +155,83 @@ pub fn select_accounts_by_block_range(
...
+    for (account_id, details, account_hash) in accounts.iter() {
+        let account_id = *account_id;
+        count += acc_stmt.execute(params![
+            u64_to_value(account_id),
+            account_hash.to_bytes(),
+            block_num
+        ])?;
+        if let Some(details) = details {
+            match details {
+                AccountDetails::Full(account) => {
```
hackaugusto, line 184: nit: this could be just:

```rust
match details {
  Some(AccountDetails::Full(account)) => { ... },
  Some(AccountDetails::Delta(delta)) => { ... },
  None => (),
}
```

Not a huge improvement, but this makes the indentation shallower which IMO is a bit easier to read.

store/src/db/sql.rs
```diff
@@ -125,16 +155,83 @@ pub fn select_accounts_by_block_range(
...
+
+                    let mut account = Account::new(
+                        account_id.try_into()?,
+                        AssetVault::read_from_bytes(row.get_ref(1)?.as_blob()?)?,
+                        AccountStorage::read_from_bytes(row.get_ref(2)?.as_blob()?)?,
+                        AccountCode::read_from_bytes(row.get_ref(3)?.as_blob()?)?,
+                        column_value_as_u64(row, 0)?
+                            .try_into()
+                            .map_err(DatabaseError::CorruptedData)?,
+                    );
```
hackaugusto, line 211: Request: use `select_account_details` instead

store/src/server/api.rs
```diff
@@ -190,11 +221,15 @@ impl api_server::Api for StoreApi {
             .accounts
             .into_iter()
             .map(|account_update| {
+                let account_details: Option<AccountDetails> =
+                    <Option<AccountDetails>>::read_from_bytes(&account_update.details)
+                        .map_err(|err| Status::invalid_argument(err.to_string()))?;
```
hackaugusto, line 226: This should have validation:

- if the account is on-chain, there must be details
- if the account is off-chain, there must NOT be any details

hackaugusto (commented):
partial review, since this is built in another pr.

proto/proto/responses.proto
polydez: This code was removed later.

store/src/db/sql.rs
polydez: This code was removed later.

block-producer/src/state_view/mod.rs
hackaugusto: The linked code is creating new objects. At first sight it seems none of them require ownership.

store/src/db/sql.rs
```diff
@@ -156,25 +156,60 @@ pub fn upsert_accounts(
...
-        "
-        INSERT OR REPLACE INTO
-            accounts (account_id, account_hash, block_num, details)
-        VALUES (?1, ?2, ?3, ?4);
-    ",
+    let mut upsert_stmt = transaction.prepare(
+        "INSERT OR REPLACE INTO accounts (account_id, account_hash, block_num, details) VALUES (?1, ?2, ?3, ?4);",
     )?;
+    let mut select_details_stmt =
+        transaction.prepare("SELECT details FROM accounts WHERE account_id = ?1;")?;
```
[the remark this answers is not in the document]
polydez: This preparation statement intentionally moved out of the loop for performance reasons.

store/src/server/api.rs
```diff
@@ -215,18 +218,38 @@ impl api_server::Api for StoreApi {
...
+                    (true, false) => {
+                        return Err(Status::invalid_argument("On-chain account must have details"));
+                    },
+                    (false, true) => {
+                        return Err(Status::invalid_argument(
+                            "Off-chain account must not have details",
+                        ));
+                    },
+                    _ => (),
+                }
```
bobbinth, line 237: I wonder if we should do a more robust check here to endure that for on-chain accounts, not only are the details provided, but also that the details are valid (i.e., applying the details to the previous state of the account gives us the current state). There are a few open questions here which I don't have the answers to - so, let's create an issue to discuss.

store/src/db/mod.rs
```diff
@@ -347,8 +346,14 @@ impl Db {
...
                         let accounts: Vec<_> = account_smt
                             .leaves()
-                            .map(|(account_id, state_hash)| (account_id, None, state_hash.into()))
-                            .collect();
+                            .map(|(account_id, state_hash)| {
+                                Ok(AccountDetailsUpdate {
+                                    account_id: account_id.try_into()?,
+                                    final_state_hash: state_hash.into(),
+                                    details: None,
+                                })
```
bobbinth, line 354: Does this assume that there are no public accounts in the genesis file? If so, we should probably create an issue about handling public accounts in the genesis state.

store/src/db/sql.rs
```diff
@@ -667,8 +694,35 @@ fn account_hash_update_from_row(row: &rusqlite::Row<'_>) -> Result<AccountHashUp
...
+
+    account.apply_delta(delta)?;
+
+    let actual_hash = account.hash();
+    if &actual_hash != final_state_hash {
+        return Err(DatabaseError::ApplyBlockFailedAccountHashesMismatch {
+            calculated: actual_hash,
+            expected: *final_state_hash,
+        });
+    }
```
bobbinth, line 725: Ah - I see. We are performing the check here.

store/src/db/sql.rs
```diff
@@ -153,28 +157,51 @@ pub fn select_account(
...
+    for update in accounts.iter() {
+        let account_id = update.account_id.into();
+        let full_account = match &update.details {
+            None => None,
+            Some(AccountDetails::Full(account)) => {
+                debug_assert!(account.is_new());
+                debug_assert_eq!(account_id, u64::from(account.id()));
+
+                Some(Cow::Borrowed(account))
+            },
```
bobbinth, line 179: Should we not check here if final account hash matches the account hash from details and return an error if not?

bobbinth (changes requested):
Thank you! Looks good! I left some comments inline.

The main open question is how to verify that the account details included with a public account are valid. The most problematic part of this is to verify account delta for existing accounts. Ideally, this should be happening in the RPC so that a proven transaction with malformed account delta would not make it into the block producer. But I'm not seeing an efficient way of doing this. So, the current approach is fine, but let's create an issue to figure out how to validate account delta more holistically.

store/src/db/mod.rs
polydez: Yes, we don't support public accounts in the genesis fill, will create an issue about it, thank you!

polydez: Created an issue: https://github.com/0xPolygonMiden/miden-node/issues/305

rpc/README.md
```diff
@@ -111,7 +110,7 @@ contains excessive notes and nullifiers, client can make additional filtering of
 - `chain_tip`: `uint32` – number of the latest block in the chain.
 - `block_header`: `BlockHeader` – block header of the block with the first note matching the specified criteria.
 - `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `block_num + 1` to `block_header.block_num`.
-- `accounts`: `[AccountHashUpdate]` – a list of account hashes updated after `block_num + 1` but not after `block_header.block_num`.
+- `accounts`: `[AccountSummary]` – a list of account hashes updated after `block_num + 1` but not after `block_header.block_num`.
```
bobbinth, line 113: nit: I would probably change the description to something like: "account summaries for accounts updated after ...".

store/README.md
```diff
@@ -157,7 +156,7 @@ contains excessive notes and nullifiers, client can make additional filtering of
 - `chain_tip`: `uint32` – number of the latest block in the chain.
 - `block_header`: `BlockHeader` – block header of the block with the first note matching the specified criteria.
 - `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `block_num + 1` to `block_header.block_num`.
-- `accounts`: `[AccountHashUpdate]` – a list of account hashes updated after `block_num + 1` but not after `block_header.block_num`.
+- `accounts`: `[AccountSummary]` – a list of account hashes updated after `block_num + 1` but not after `block_header.block_num`.
```
bobbinth, line 159: Same nit as above.

bobbinth (approved):
Looks good! Thank you! I left a couple of very small nits inline. After these, we can merge.

PR #295, opened 2024-04-02
Implemented support for note trees in Miden node

polydez:
Updated code to use `BlockTreeNotes` and `BatchTreeNotes`, introduced `batch_index` in order to reduce bugs with calculations.

### DIFF
--- a/store/src/db/migrations.rs
+++ b/store/src/db/migrations.rs
@@ -18,6 +18,7 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
             notes
         (
             block_num INTEGER NOT NULL,
+            batch_index INTEGER NOT NULL,
             note_index INTEGER NOT NULL,
             note_hash BLOB NOT NULL,
             sender INTEGER NOT NULL,
@@ -27,6 +28,7 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
             PRIMARY KEY (block_num, note_index),
             CONSTRAINT fk_block_num FOREIGN KEY (block_num) REFERENCES block_headers (block_num),
             CONSTRAINT notes_block_number_is_u32 CHECK (block_num >= 0 AND block_num < 4294967296),
+            CONSTRAINT notes_batch_index_is_u32 CHECK (batch_index BETWEEN 0 AND 0xFFFFFFFF)
             CONSTRAINT notes_note_index_is_u32 CHECK (note_index >= 0 AND note_index < 4294967296)
         ) STRICT, WITHOUT ROWID;
 
--- a/store/src/db/migrations.rs
+++ b/store/src/db/migrations.rs
@@ -18,8 +18,8 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
             notes
         (
             block_num INTEGER NOT NULL,
-            batch_index INTEGER NOT NULL,
-            note_index INTEGER NOT NULL,
+            batch_index INTEGER NOT NULL,   -- Index of batch in block, starting from 0
+            note_index INTEGER NOT NULL,    -- Index of note in batch, starting from 0
             note_hash BLOB NOT NULL,
             sender INTEGER NOT NULL,
             tag INTEGER NOT NULL,
--- a/block-producer/src/batch_builder/batch.rs
+++ b/block-producer/src/batch_builder/batch.rs
@@ -2,12 +2,10 @@ use std::collections::BTreeMap;
 
 use miden_objects::{
     accounts::AccountId,
-    crypto::{
-        hash::blake::{Blake3Digest, Blake3_256},
-        merkle::SimpleSmt,
-    },
+    batches::BatchNoteTree,
+    crypto::hash::blake::{Blake3Digest, Blake3_256},
     notes::{NoteEnvelope, Nullifier},
-    Digest, BATCH_OUTPUT_NOTES_TREE_DEPTH, MAX_NOTES_PER_BATCH,
+    Digest, MAX_NOTES_PER_BATCH,
 };
 use tracing::instrument;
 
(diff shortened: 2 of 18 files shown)


store/src/db/migrations.rs
```diff
@@ -18,6 +18,7 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
             notes
         (
             block_num INTEGER NOT NULL,
+            batch_index INTEGER NOT NULL,
             note_index INTEGER NOT NULL,
```
bobbinth, line 22: We should probably add comments here to clarify what `batch_index` and `note_index` represent.

store/src/db/migrations.rs
```diff
@@ -27,6 +28,7 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
             PRIMARY KEY (block_num, note_index),
             CONSTRAINT fk_block_num FOREIGN KEY (block_num) REFERENCES block_headers (block_num),
             CONSTRAINT notes_block_number_is_u32 CHECK (block_num >= 0 AND block_num < 4294967296),
+            CONSTRAINT notes_batch_index_is_u32 CHECK (batch_index BETWEEN 0 AND 0xFFFFFFFF)
             CONSTRAINT notes_note_index_is_u32 CHECK (note_index >= 0 AND note_index < 4294967296)
```
bobbinth, line 32: If we are using `BETWEEN` for `batch_index`, probably makes sense to do the same for `block_num` and `note_index`.

faucet/Cargo.toml
```diff
@@ -15,11 +15,11 @@ actix-files = "0.6.5"
...
 derive_more = "0.99.17"
 figment = { version = "0.10", features = ["toml", "env"] }
-miden-lib = { workspace = true }
+miden-lib = { version = "0.1.0" }
 miden-client = { version = "0.1.0", features = ["concurrent"] }
 miden-node-proto = { path = "../proto", version = "0.2" }
 miden-node-utils = { path = "../utils", version = "0.2" }
-miden-objects = { workspace = true }
+miden-objects = { version = "0.1.0" }
 serde = { version = "1.0", features = ["derive"] }
```
bobbinth, line 23: Why are these changes needed?

Cargo.toml
```diff
@@ -24,15 +24,15 @@ exclude = [".github/"]
...
 [workspace.dependencies]
 miden-air = { version = "0.8", default-features = false }
-miden-lib = { version = "0.1" }
-miden-objects = { version = "0.1" }
+miden-lib = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "polydez-notes-tree" }
+miden-objects = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "polydez-notes-tree" }
 miden-processor = { version = "0.8" }
 miden-stdlib = { version = "0.8", default-features = false }
-miden-tx = { version = "0.1" }
+miden-tx = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "polydez-notes-tree" }
```
bobbinth, line 31: These can probably be changed to the `next` branch now.

bobbinth (approved):
Looks good! Thank you! I left a few small comments inline. After these are addressed, we can merge.

store/src/db/migrations.rs
polydez: @bobbinth yes, I did it, but in different PR: https://github.com/0xPolygonMiden/miden-node/pull/293/files#diff-aaf55c2ae683cbe885dbb5cebb9f8526e16dba91d863f854f4ecca5c366f900aR14-R17

faucet/Cargo.toml
polydez: @bobbinth I faced incompatibility of miden-client with the `next` branch of miden-base and decided not to migrate it yet, just use previous version in the faucet. Today @igamigo contacted me and provided updated code, so I will check and probably update the deps.

polydez: @bobbinth update: migration of the fauced seems not a straightforward, needs to rewrite part of code regarding using of new miden-client. For this PR I leave old version of the client.
@igamigo @phklive FYI

polydez:
@igamigo 
> I did not check out the changes in `miden-base` in depth. I'm a bit confused by the introduction of the batch index and what advantages it brings but the overall code LGTM.

We need batch index for block note tree in order to mitigate bugs with note index calculations we faced. Our previous approach was to write note id and metadata hash into two adjacent leaves of `SimpleSmt`, which indexes were calculated as 
```
MAX_NUMBER_OF_BATCHES * batch_index + note_index_in_batch * 2
```
and +1 for metadata hash. This code was duplicated several times and each implementation was different, some of them were buggy. This PR should force to provide batch index and node index in batch as separate arguments and tree implementation will calculate all the rest by itself in one single place.

faucet/Cargo.toml
polydez: @phklive some of structures/arguments were changed in miden-client, thus, miden-faucet code needs to be also updated.

phklive: Understood. We will soon rework the faucet + end-to-end testing code to remove the dependency on the client. We will implement transaction creation, execution and submission directly using the base, making sure that we do not have these types of problems in the future.

igamigo: Didn't the faucet use the released client version? There shouldn't have been any changes in that regard so I wonder  what the issue was. One thing that will break on the node when bringing everything to `next` is the account generation code which still uses `KeyPair`. I will take a look.

igamigo:
```rust
miden-client = { version = "0.1.0", features = ["concurrent"] }
miden-node-proto = { path = "../proto", version = "0.2" }
miden-node-utils = { path = "../utils", version = "0.2" }
```
Seems like the faucet should probably use the release versions of `miden-node-proto` in order to be fully compatible with the client. https://github.com/0xPolygonMiden/miden-node/blob/next/faucet/Cargo.toml

PR #297, opened 2024-04-04
Migration to the `next` version of Miden dependencies

polydez:


block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -261,7 +261,7 @@ impl BlockWitness {
                 .expect("updated accounts number is greater than or equal to the field modulus"),
         );
 
-        StackInputs::new(stack_inputs)
+        StackInputs::new(stack_inputs).map_err(BlockProverError::FailedToConstructStackInputs)
```
bobbinth, line 264: Under what circumstances could this error happen? Shouldn't this be just `.expect()` since we know for sure we won't construct invalid stack inputs above?

block-producer/src/errors.rs
```diff
@@ -84,21 +84,23 @@ impl BuildBatchError {
...
     #[error("Received invalid merkle path")]
     InvalidMerklePaths(MerkleError),
-    #[error("program execution failed")]
+    #[error("Program execution failed")]
     ProgramExecutionFailed(ExecutionError),
-    #[error("failed to retrieve {0} root from stack outputs")]
+    #[error("Failed to retrieve {0} root from stack outputs")]
     InvalidRootOutput(&'static str),
+    #[error("Failed to construct stack inputs: {0}")]
+    FailedToConstructStackInputs(InputError),
```
bobbinth, line 96: Related to the previous comment: I think this error type can go away.

faucet/Cargo.toml
```diff
@@ -15,11 +15,11 @@ actix-files = "0.6.5"
 actix-cors = "0.7.0"
 derive_more = "0.99.17"
 figment = { version = "0.10", features = ["toml", "env"] }
-miden-lib = { workspace = true }
+miden-lib = { version = "0.1.0" }
 miden-client = { version = "0.1.0", features = ["concurrent"] }
 miden-node-proto = { path = "../proto", version = "0.2" }
 miden-node-utils = { path = "../utils", version = "0.2" }
-miden-objects = { workspace = true }
+miden-objects = { version = "0.1.0" }
```
bobbinth, line 22: lLet's add comments here of why we are keeping old versions.

node/src/commands/genesis/mod.rs
```diff
@@ -183,11 +184,14 @@ fn parse_auth_inputs(
 ) -> Result<(AuthScheme, AuthData)> {
     match auth_scheme_input {
         AuthSchemeInput::RpoFalcon512 => {
-            let auth_seed = hex_to_bytes(auth_seed)?;
-            let keypair = KeyPair::from_seed(&auth_seed)?;
+            let auth_seed: [u8; 40] = hex_to_bytes(auth_seed)?;
+            // Get the first 32 bytes, ignore remaining:
+            let auth_seed: [u8; 32] = auth_seed[..32].try_into().expect("Unreachable");
```
bobbinth, line 189: Instead of doing this, we should probably just shorten the `auth_seed` string `genesis.toml` file and then we can just do `let auth_seed: [u8; 32] = hex_to_bytes(auth_seed)?;`

bobbinth (changes requested):
Thank you! Looks good! I left a few small comments inline.

bobbinth (approved):
All looks good! Thank you!

Issue #288, opened 2024-03-26
Update storage to support on-chain notes


bobbinth:
This is needed to support on-chain notes, which require storing full `Note` objects in the database. There are several ways to do:

#### Approach 1
Add `details` field to the `notes` table. This field will contain serialized of the `Note` object. For now, this will include the metadata - but once we remove it from the `Note`, it will be just `script`, `inputs`, `assets`, and `serial_num`. The `details` field would be nullable - so, for off-chain notes it will be set to NULL.

A potential downside of this approach is that for standardized scripts, we'll be duplicating bunch of data.

#### Approach 2
Similar to approach 1 (i.e., we add `details` field to the `notes` table), but now we serialize the `Note` in such a way that we serialize only `inputs`, `assets`, `serial_num`, and `script_root`. The actual script would be stored in a new table - `note_scripts`, which could look as follows:

```
CREATE TABLE
    note_scripts
(
    script_root BLOB NOT NULL,
    script_code BLOB NOT NULL,

    PRIMARY KEY (script_root),
)  STRICT, WITHOUT ROWID;
```
This would avoid script duplication for common scripts, but may make lookups a bit more involved.

#### Approach 3
Add a separate table for `note_details` which would have `note_id` as the primary key and would have all fields (`script`, `inputs`, `assets`, and `serial_num`) materialized. Similar to approach 2, we could also have a separate table for `note_scripts` to reduce storage footprint.

The advantage of this approach is that it completely separates note details from the original note table - so, we could imagine `note_details` living on a separate disk in the future.

Overall, my thinking is that approach 1 is probably the simplest to go with for now and then we can get more sophisticated in the future with either approach 2 or 3.

phklive:
@bobbinth, regarding  Approach 1: we add a `details` field to the `notes` table in the database. This field contains the whole `Note` serialised. Would be filled for on-chain notes and would be null for off-chain notes. 

If we take the example of this PR it adds a new table for on-chain accounts called `account_details`. Should I follow your approach or follow the same pattern as in this PR for notes?

https://github.com/0xPolygonMiden/miden-node/pull/293

bobbinth:
Ah - good point! I think we should be consistent and either put everything into a single field in the `accounts` table or have separate `note_details` table. I am fine with either - but I do think not adding extra tables would probably be simpler for now.

@polydez - what do you think?

polydez:
[...]

PR #300, opened 2024-04-05
Add support for on-chain `notes` to the database

phklive:
In this PR I propose an update to the database to provide support for on-chain notes.

Closes: #288


igamigo:
Do we want to base this PR off of `phklive-get-notes-by-id`?

phklive:
> Do we want to base this PR off of `phklive-get-notes-by-id`?

Does that bring any issues? 

These PR's are very linked. And the first PR is needed to correctly test this one.

proto/proto/note.proto
```diff
@@ -12,6 +12,7 @@ message Note {
     account.AccountId sender  = 4;
     fixed64 tag = 5;
     merkle.MerklePath merkle_path = 7;
+    optional bytes details = 8;
```
igamigo, line 15: Can we document somewhere what `details` is meant to contain and when the endpoint will return it?

rpc/README.md
```diff
@@ -60,6 +60,18 @@ Retrieves block header by given block number.
...
+Returns a list of Note's matching a certain list of NoteId's.
+
+**Parameters**
+
+* `note_ids`: `NoteId` - ID of the Note that we want to query.
+
+**Returns**
+
+* `notes`: `Note` - List of notes matching the list of NoteId's.
+
```
igamigo, line 74: nit: I think this is missing the changes done related to inputs and outputs being lists.

igamigo:
> > Do we want to base this PR off of `phklive-get-notes-by-id`?
> 
> Does that bring any issues?
> 
> These PR's are very linked. And the first PR is needed to correctly test this one.

I might be missing something but it seems this PR includes some changes that are also on the PR for `phklive-get-notes-by-id`, which makes it a a bit harder to tell which changes belong to this PR as opposed to the other.

phklive:
> > > Do we want to base this PR off of `phklive-get-notes-by-id`?
> > 
> > 
> > Does that bring any issues?
> > These PR's are very linked. And the first PR is needed to correctly test this one.
> 
> I might be missing something but it seems this PR includes some changes that are also on the PR for `phklive-get-notes-by-id`, which makes it a a bit harder to tell which changes belong to this PR as opposed to the other.

We had a misunderstanding, you were right indeed. 

I re-based the PR on `phklive-get-notes-by-id` and did the same for this PR: #303

store/src/db/sql.rs
```diff
@@ -369,7 +376,8 @@ pub fn select_notes_since_block_by_tag_and_sender(
             note_hash,
             sender,
             tag,
-            merkle_path
+            merkle_path,
+            details
```
bobbinth, line 380: This is probably OK for now, but in the future we should refactor this. The main reason is that `select_notes_since_block_by_tag_and_sender()` is used to build a `sync_state` response and for this response we don't need not details. With this approach, we'd be doing a lot of unnecessary work (e.g., reading the details form the database, parsing them etc.).

Let's create an issue to address this in the future.

proto/src/domain/notes.rs
```diff
@@ -28,6 +28,7 @@ impl From<(usize, usize, NoteEnvelope)> for note::NoteCreated {
             note_id: Some(note.note_id().into()),
             sender: Some(note.metadata().sender().into()),
             tag: note.metadata().tag().into(),
+            details: None,
```
bobbinth, line 31: When do we use this conversion? It may be good to add a comment on why it is ok to always set `details` to `None`.

store/src/db/sql.rs
```diff
@@ -333,6 +339,7 @@ pub fn insert_notes(
             u64_to_value(note.note_created.sender),
             u64_to_value(note.note_created.tag),
             note.merkle_path.to_bytes(),
+            note.note_created.details.to_bytes()
```
bobbinth, line 342: It seems like we are writing details as an `Option` to the database. If so, I think a better approach would be to make the column nullable (as I mentioned in one of the previous comments) and then write null on `None` and the actual value of details on `Some`.

store/src/db/sql.rs
```diff
@@ -275,6 +276,9 @@ pub fn select_notes(conn: &mut Connection) -> Result<Vec<Note>> {
         let merkle_path_data = row.get_ref(6)?.as_blob()?;
         let merkle_path = deserialize(merkle_path_data)?;
 
+        let details_data = row.get_ref(7)?.as_blob()?;
+        let details = deserialize(details_data)?;
```
bobbinth, line 280: Related to the above comment: if we make column nullable and store `NULL` on `None`, we might need to handle this slightly differently.

bobbinth (changes requested):
Thank you! Looks good! I left a few comments inline. The man one is about potentially making the `details` column nullable.

node/miden-node.toml
```diff
@@ -17,5 +17,5 @@ store_url = "http://localhost:28943"
 [store]
 # port defined as: sum(ord(c)**p for (p, c) in enumerate('miden-store', 1)) % 2**16
 endpoint = { host = "localhost", port = 28943 }
-database_filepath = "db/miden-store.sqlite3"
+database_filepath = "database/miden-store.sqlite3"
```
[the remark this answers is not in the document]
phklive: Because we added `db/` as a prefix to the file path of the db files we needed to add `db/` to the `.gitignore` this caused problems with editors and git in general with the other `db` folder that we have in the `store` crate.

store/src/db/sql.rs
phklive: Should I just remove the `details` field from this query ?

rpc/README.md
phklive: fixed by merging next.

proto/src/domain/notes.rs
polydez: Also consider adding `Option<Vec<u8>>` as `details` argument for conversion:
```rust
impl From<(usize, usize, NoteEnvelope, Option<Vec<u8>>)> for note::NoteCreated {
    fn from((batch_idx, note_idx, note, details): (usize, usize, NoteEnvelope, Option<Vec<u8>>)) -> Self {
```
It would be more flexible and clearer to understand.

store/src/db/sql.rs
polydez: @phklive in public accounts we had the same issue and I removed this field from query and structure. I also added structure for specific endpoint for public accounts where this field is useful.

store/src/db/sql.rs
phklive: Does that only apply to notes or does it also apply to accounts? I have the understanding that we are not following this pattern when inserting accounts: https://github.com/0xPolygonMiden/miden-node/blob/f5a7bd9611ce56a4511030171389916a2ca1311b/store/src/db/sql.rs#L159-L214

polydez: @phklive if I understood you correctly, in line #205 I process `None` details as `NULL`.

phklive: > @phklive if I understood you correctly, in line 205 I process `None` details as `NULL`.

using `as_ref` ?

rpc/README.md
bobbinth: Seems like something is still off here. This content is different from what's in `next` (see [here](https://github.com/0xPolygonMiden/miden-base/blob/next/miden-tx/src/host/account_delta_tracker.rs)). Also, it shouldn't appear as new changes.

bobbinth:
@phklive - how close are we to getting this reviewed/merged?

phklive:
> @phklive - how close are we to getting this reviewed/merged?

Should be ready for review, finalized in the plane. Just arrived safely in Athens 🇬🇷

proto/src/domain/notes.rs
phklive: // This is `None` for now as this conversion is used by the block-producer
            // when using `apply_block`. The block-producer has not yet been updated to support
            // on-chain notes. Will be set to the correct value when the block-producer is
            // up-to-date.

store/src/db/sql.rs
phklive: To the extent of my understanding as of now I do not think that it would be wise to simply remove the details field and set all note details to `None` this would be misleading. Hence I prefer to open a separate issue to refactor this flow. 

Before opening the issue @bobbinth could you please clarify the reason why we would not need note details when syncing the state? I thought that we would also want to collect some new notes from the network using sync_state and that some of these notes could happen to be on-chain.

bobbinth: > To the extent of my understanding as of now I do not think that it would be wise to simply remove the details field and set all note details to `None` this would be misleading. Hence I prefer to open a separate issue to refactor this flow.

Agreed. We should have a separate struct which is like note but without details (that's how it works for accounts, I believe). And yes, we can do it in a separate PR.

> Before opening the issue @bobbinth could you please clarify the reason why we would not need note details when syncing the state? I thought that we would also want to collect some new notes from the network using sync_state and that some of these notes could happen to be on-chain.

State sync could return many irrelevant notes (because we do a kind of "fuzzy" matching based on tags) and sending details for such notes could be expensive. So, the process is that with state sync we get minimal note info (Id + metadata + inclusion proofs) and if some of the notes happen to be on-chain, we can request details for them separately via the `GetNotesById` endpoint (this is actually one of the main reasons why we add this endpoint).

bobbinth (approved):
Looks good! Thank you!

Let's make sure to create an issue to remove note details from state sync responses.

Dominik1999:
Can we merge?

phklive:
> Can we merge?

Yes we can. Opening the issue now.

store/src/db/sql.rs
phklive: Opened issue here: https://github.com/0xPolygonMiden/miden-node/issues/309

igamigo: Regarding this, with the current information we get from the sync, there is no way to tell whether the note is on chain or not, so we should add that information to the response.

bobbinth: Does it not come with note metadata? If it is not there yet, we should make sure the protobuf messages include `note_type`.

igamigo: Correct, the sync endpoint specifically does not include the note type in any form.

Issue #141, opened 2024-01-11
Implement support for on-chain notes


bobbinth:
Currently, the node can handle only off-chain notes. To support on-chain notes we'd need to do the following:

1. Add necessary tables to the store database to store on-chain note info.
2. Update block producer to make sure it extracts on-chain note data from proven transactions and send this data to the store via `apply_block`.
3. Update the store to record the on-chain note data in the DB.
4. Update the store and the RPC to provide an endpoint for users to retrieve on-chain note details
  a. Maybe also update `sync_state` endpoint to return on-chain note details for such notes (though, this requires a bit more thinking).

This work requires https://github.com/0xPolygonMiden/miden-base/issues/403 to be completed.

hackaugusto:
For some context:

related issues in miden-base: https://github.com/0xPolygonMiden/miden-base/issues/403 https://github.com/0xPolygonMiden/miden-base/issues/402
related PRs so far: https://github.com/0xPolygonMiden/miden-base/pull/481 https://github.com/0xPolygonMiden/miden-base/pull/485

Dominik1999:
We break it down into smaller issues @hackaugusto @Dominik1999 

- issue to create an endpoint
- issue for DB
- ...

Dominik1999:
@bobbinth see comment above

bobbinth:
I broke out points 1 and 4 into separate issues: #286 and #288. So, this issue is now only for points 2 and 3 of the original description. I think the order of implementation could be:

1. #288
2. #286 
3. And then this issue.

Issue #291, opened 2024-03-29
[Feature]: Create `Test Runner` that does not rely on Miden client


phklive:
### To whom is this feature for?

Rollup operator

### Why is is this feature needed?

For now we are relying on the Miden client for the end-to-end tests of the Miden node. This could cause problems when breaking changes arise in those crates. To address this we want to remove this dependency and create a `Test runner` that would have the same functionality and enable the end-to-end tests to be performed.

### How is this feature used?

It will be used in the testing suite of the Miden node

### Anything else?

You can find the original PR and comment here: https://github.com/0xPolygonMiden/miden-node/pull/282#pullrequestreview-1961665075

PR #303, opened 2024-04-05
Add support for on-chain notes to `block-producer`

phklive:
In this PR I propose to update the block-proposer to enable addition of on-chain notes to the `store` through `apply_block` api

Closes: #141


phklive:
@hackaugusto as we discussed note and tx validation together a few days ago I have the feeling that it is a full other PR in of itself. Should we open another issue to address that thoroughly?

phklive:
@bobbinth Indeed I wasn't fully getting the extent of the trouble we could get into when using the Miden client as a dep for faucet or end-to-end testing. Here is an example where not having a dep on the client to send ProvenTransactions to the Block-Producer for testing could be very useful. 

That should be something to address: #291

block-producer/src/batch_builder/batch.rs
```diff
@@ -62,23 +64,52 @@ impl TransactionBatch {
...
+                                note_id: Some(note.note_id().into()),
+                                sender: Some(note.metadata().sender().into()),
+                                tag: note.metadata().tag().into(),
+                                details: None,
+                            }
+                        }
+                    })
+                })
+                .collect();
+
```
igamigo, line 98: This will likely have to change after https://github.com/0xPolygonMiden/miden-base/pull/572. Maybe it's best to point at that branch for now to not have to redo this soonish.

Dominik1999:
Nice, only this is remaining, I suppose?

phklive:
> Nice, only this is remaining, I suppose?

Yes exactly, rebasing atm.

block-producer/src/batch_builder/batch.rs
phklive: Thanks for the heads up, fixed the implementation.

phklive:
This PR has been rebased on continued here: https://github.com/0xPolygonMiden/miden-node/pull/310

block-producer/src/block.rs
```diff
@@ -19,7 +19,7 @@ use crate::store::BlockInputsError;
 pub struct Block {
     pub header: BlockHeader,
     pub updated_accounts: Vec<UpdatedAccount>,
-    pub created_notes: Vec<(usize, usize, NoteEnvelope)>,
+    pub created_notes: Vec<(usize, usize, NoteCreated)>,
```
polydez, line 22: If we put `NoteCreated` here, we don't need to provide additionally `batch_index` and `note_index` here.

proto/src/domain/notes.rs
```diff
@@ -20,14 +18,15 @@ impl From<note::Note> for note::NoteSyncRecord {
 // NoteCreated
 // ================================================================================================
 
-impl From<(usize, usize, NoteEnvelope)> for note::NoteCreated {
-    fn from((batch_idx, note_idx, note): (usize, usize, NoteEnvelope)) -> Self {
+impl From<(usize, usize, NoteCreated)> for note::NoteCreated {
```
polydez, line 21: I'm not sure we need such conversion.

store/src/db/migrations.rs
```diff
@@ -24,6 +24,7 @@ pub static MIGRATIONS: Lazy<Migrations> = Lazy::new(|| {
             sender INTEGER NOT NULL,
             tag INTEGER NOT NULL,
             merkle_path BLOB NOT NULL,
+            details BLOB NOT NULL,
```
polydez, line 27: Please, consider making `details` nullable.

polydez (changes requested):

block-producer/src/batch_builder/batch.rs
```diff
@@ -26,9 +27,10 @@ pub struct TransactionBatch {
     id: BatchId,
     updated_accounts: BTreeMap<AccountId, AccountStates>,
     produced_nullifiers: Vec<Nullifier>,
-    created_notes_smt: BatchNoteTree,
     /// The notes stored `created_notes_smt`
```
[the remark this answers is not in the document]
phklive: That is also what I thought, I did not know if I had to remove it. 

Removed.

PR #304, opened 2024-04-06
fix: Send correct note index on sync response

igamigo:
(related to https://github.com/0xPolygonMiden/miden-base/pull/574)
As part of trying to solve the problem that we are seeing on the client for concurrent transactions on integration tests, we noticed that the note indices might not be being sent correctly from the node.

After #295, the meaning of `note_index` that is passed around between processes changed a little bit. It now describes the position of the note local to the correct subtree (based on batch index), instead of the absolute position in the note tree:
https://github.com/0xPolygonMiden/miden-node/pull/295/files#diff-dfb6dcf6fa51cd46bba3f909bc3700a7dc800fa7b92d6a8e90724cd86a5012e7R86-R89.

This causes the sync response to return incorrectly as it does _not_ include the batch index, so there is no way to calculate the correct position. This PR attempts a fix where, for `NoteSyncRecord` (which is sent as part of the sync response), the `note_index` is set to the absolute position of the note. This, however, does not fix the problem so there might be something else that we are missing (or maybe this PR is wrong altogether).


store/src/db/mod.rs
```diff
@@ -52,6 +53,14 @@ pub struct NoteCreated {
     pub tag: u64,
 }
 
+impl NoteCreated {
+    /// Returns the absolute position on the note tree based on the batch index
+    /// and local-to-the-subtree index
+    pub fn absolute_note_index(&self) -> u32 {
+        BlockNoteTree::leaf_index(self.batch_index as usize, self.note_index as usize) as u32
+    }
```
bobbinth, line 61: I wonder if this should be:
```Rust
BlockNoteTree::leaf_index(self.batch_index as usize, self.note_index as usize) as u32 / 2
```

This would give us the index of the node which is the parent of both note leaf and metadata leaf, and I think that maybe this is what the client is expecting?

igamigo: That very much sounds like it. I tested it and now I don't get [the transaction executor's `is_in_block()` error](https://github.com/0xPolygonMiden/miden-base/blob/69ec4dc4e1a6f78d3d370c6b3b2bd25b5887ebc8/objects/src/transaction/inputs.rs#L84-L88). I get a VM assertion error

```rust
called `Result::unwrap()` on an `Err` value: TransactionExecutionError(ExecuteTransactionProgramFailed(FailedAssertion { clk: 5456, err_code: 131075, err_msg: None }))
```

bobbinth (approved):
Looks good! Thank you! Let's update the `Cargo.toml` after https://github.com/0xPolygonMiden/miden-base/pull/575 is merge, and then merge this one as well.

proto/src/domain/notes.rs
```diff
@@ -2,21 +2,6 @@ use miden_objects::notes::NoteEnvelope;
...
-    fn from(value: note::Note) -> Self {
-        Self {
-            note_index: value.note_index,
-            note_id: value.note_id,
-            sender: value.sender,
-            tag: value.tag,
-            merkle_path: value.merkle_path,
-        }
-    }
-}
```
bobbinth, line 18: Question: why do we no longer need this?

bobbinth (approved):
Looks good! Thank you for fixing these again! I left just one question inline.

proto/src/domain/notes.rs
igamigo: I was just trying to make sure that we were not setting the incorrect `note_index` and checked all references which brought me to this. I deleted it to see if it would bring up any place where we were using this and saw that it was not being used at all. I can bring it back.

bobbinth: If it is not used anywhere - no need to bring it back.

igamigo: Agreed, that's what I thought

PR #310, opened 2024-04-10
Add support for on-chain notes to `block-producer`

phklive:
In this PR I propose to update the block-proposer to enable addition of on-chain notes to the `store` through `apply_block` api

Closes: #141


block-producer/src/batch_builder/batch.rs
```diff
@@ -27,9 +28,10 @@ pub struct TransactionBatch {
     id: BatchId,
     updated_accounts: BTreeMap<AccountId, AccountStates>,
     produced_nullifiers: Vec<Nullifier>,
-    created_notes_smt: BatchNoteTree,
     /// The notes stored `created_notes_smt`
```
polydez, line 31: It seems this comment is not in the right place now. And I think, it is confusing a bit, maybe it would be better to get rid of this comment at all?

block-producer/src/batch_builder/batch.rs
```diff
@@ -27,9 +28,10 @@ pub struct TransactionBatch {
...
-    created_notes: Vec<NoteEnvelope>,
+    created_notes_smt: BatchNoteTree,
+    created_notes: Vec<NoteCreated>,
+    created_note_envelopes: Vec<NoteEnvelope>,
```
polydez, line 34: Do we use these envelopes anywhere else?

block-producer/src/batch_builder/batch.rs
```diff
@@ -64,14 +66,38 @@ impl TransactionBatch {
         let produced_nullifiers =
             txs.iter().flat_map(|tx| tx.input_notes().iter()).cloned().collect();
 
-        let (created_notes, created_notes_smt) = {
-            let created_notes: Vec<NoteEnvelope> = txs
+        let (created_notes, created_notes_smt, created_note_envelopes) = {
```
polydez, line 69: I think we shouldn't produce `NoteCreated` here, since we don't know all needed fields at the moment (we don't know `batch_index`). We can provide note envelopes here and serialized details and construct `NoteCreated` in the block builder.

So I would rewrite this, something like:
```rust
        let (created_notes_smt, created_note_envelopes) = {
            let created_note_envelopes: Vec<(NoteEnvelope, Option<Vec<u8>>)> = txs
                .iter()
                .flat_map(|tx| tx.output_notes().iter())
                .map(|note| {
                    match note {
                        OutputNote::Public(note) => (note.into(), Some(note.to_bytes())),
                        OutputNote::Private(envelope) => (*envelope, None),
                    }
                })
                .collect();

            if created_note_envelopes.len() > MAX_NOTES_PER_BATCH {
                return Err(BuildBatchError::TooManyNotesCreated(created_note_envelopes.len(), txs));
            }

            // TODO: document under what circumstances SMT creating can fail
            (
                BatchNoteTree::with_contiguous_leaves(
                    created_note_envelopes
                        .iter()
                        .map(|(note_envelope, _)| (note_envelope.id(), note_envelope.metadata())),
                )
                .map_err(|e| BuildBatchError::NotesSmtError(e, txs))?,
                created_note_envelopes,
            )
        };

        Ok(Self {
            id,
            updated_accounts,
            produced_nullifiers,
            created_notes_smt,
            created_notes,
        })
```

block-producer/src/batch_builder/batch.rs
```diff
@@ -27,9 +28,10 @@ pub struct TransactionBatch {
     id: BatchId,
     updated_accounts: BTreeMap<AccountId, AccountStates>,
     produced_nullifiers: Vec<Nullifier>,
-    created_notes_smt: BatchNoteTree,
     /// The notes stored `created_notes_smt`
-    created_notes: Vec<NoteEnvelope>,
+    created_notes_smt: BatchNoteTree,
+    created_notes: Vec<NoteCreated>,
```
polydez, line 33: We possibly don't need this field, see comments below.

block-producer/src/block.rs
```diff
@@ -19,7 +19,7 @@ use crate::store::BlockInputsError;
 pub struct Block {
     pub header: BlockHeader,
     pub updated_accounts: Vec<AccountUpdateDetails>,
-    pub created_notes: Vec<(usize, usize, NoteEnvelope)>,
+    pub created_notes: Vec<(usize, usize, NoteCreated)>,
```
polydez, line 22: If we put `NoteCreated` here, we don't need to provide additionally `batch_index` and `note_index` here.

block-producer/src/block_builder/mod.rs
```diff
@@ -82,16 +83,17 @@ where
 
         let updated_accounts: Vec<_> =
             batches.iter().flat_map(TransactionBatch::updated_accounts).collect();
-        let created_notes = batches
+
+        let created_notes: Vec<(usize, usize, NoteCreated)> = batches
```
polydez, line 87: We don't need these two additional `usize`s, because we already have `batch_index` and `note_index` fields in `NoteCreated` structure. The task is to fill them here. 

I think, the best solution would be to not to create `NoteCreated` in the batch builder (see comments above), but just to add serialized details field in the `TransactionBatch::created_note_envelopes` vector:
```rust
created_note_envelopes_with_details: Vec<(NoteEnvelope, Option<Vec<u8>>)>,
```
And after that we would be able to produce `NoteCreated` here and init its `batch_index` and `note_index` fields during creation.

polydez (changes requested):
Thanks you! I left several inline comments. The main idea here that we don't need to produce `NoteCreated` in the batch builder when we don't ready to fill its fields, but later in the block producer. We also can get rid of several fields.

block-producer/src/batch_builder/batch.rs
phklive: Removed.

block-producer/src/batch_builder/batch.rs
phklive: Indeed.

block-producer/src/block.rs
phklive: Indeed.

block-producer/src/batch_builder/batch.rs
phklive: With the new refactor we can directly use the `created_note_envelopes_with_details` everywhere.

block-producer/src/block_builder/mod.rs
```diff
@@ -84,13 +84,20 @@ where
...
+        let created_notes: Vec<NoteCreated> = batches
             .iter()
             .enumerate()
-            .flat_map(|(batch_idx, batch)| {
-                batch.created_notes().enumerate().map(move |(note_idx_in_batch, note)| {
-                    (batch_idx, note_idx_in_batch, note.clone())
-                })
+            .flat_map(|(batch_index, batch)| {
+                batch.created_note_envelopes_with_details().enumerate().map(
+                    move |(note_index, (note_envelope, details))| NoteCreated {
```
polydez, line 92: Similar conversion is used 2 times in code. Consider writing `From` implementation in proto/domain.

polydez (changes requested):
Looks good to me, thank you! I've left a couple comments here. Please also remove unused conversion from module proto/src/domain/notes.rs.

block-producer/src/block_builder/mod.rs
phklive: Could you point me to the seconde instance?

block-producer/src/batch_builder/batch.rs
```diff
@@ -66,52 +64,32 @@ impl TransactionBatch {
         let produced_nullifiers =
             txs.iter().flat_map(|tx| tx.input_notes().iter()).cloned().collect();
 
-        let (created_notes, created_notes_smt, created_note_envelopes) = {
-            let created_note_envelopes: Vec<NoteEnvelope> = txs
+        let (created_note_envelopes_with_details, created_notes_smt) = {
+            let created_note_envelopes_with_details: Vec<(NoteEnvelope, Option<Vec<u8>>)> = txs
```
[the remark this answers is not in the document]
phklive: You would leave the outer var and struct name the same but only change the inner var to something shorter right?

proto/src/domain/notes.rs
```diff
@@ -1,23 +1,17 @@
...
-
-use crate::generated::note;
+use crate::generated::note::{self, NoteCreated};
 
 // NoteCreated
 // ================================================================================================
 
-impl From<&(usize, usize, NoteEnvelope)> for note::NoteCreated {
-    fn from((batch_idx, note_idx, note): &(usize, usize, NoteEnvelope)) -> Self {
+impl From<&(usize, usize, NoteCreated)> for note::NoteCreated {
```
polydez, line 6: This conversion is not used in code. It should be removed.

block-producer/src/block_builder/mod.rs
polydez: > Could you point me to the seconde instance?

Please, see comments below.

block-producer/src/test_utils/block.rs
```diff
@@ -140,15 +140,27 @@ impl MockBlockBuilder {
     }
 
     pub fn build(self) -> Block {
-        let created_notes = self.created_notes.unwrap_or_default();
+        let created_note_envelopes = self.created_note_envelopes.unwrap_or_default();
+
+        let created_notes = created_note_envelopes
+            .iter()
+            .map(|(batch_index, note_index, note_envelope)| NoteCreated {
```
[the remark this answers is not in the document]
phklive: Yes I saw that one, considering it was in test_utils I did not find it crucial to DRY an implement something that would make the code slightly less readable.

What do you think?

block-producer/src/batch_builder/batch.rs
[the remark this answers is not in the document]
phklive: In the future I would go through the files to refactor, I think we should simplify this concept of what is a Note:
- Note
- NoteEnvelope
- CreatedNote
- ect...

For now this would require a too large refactor to do in this PR. 

I have changed the name simply to `created_note` as an internal var it stays readable.

block-producer/src/test_utils/block.rs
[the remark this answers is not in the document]
phklive: These 2 situations are not similar if you look into the code. 

We will need to refactor the code in the future to be able to get the same architecture everywhere.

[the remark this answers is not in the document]
polydez: My idea was to implement such conversion:
```rust
impl From<(usize, usize, &NoteEnvelope, Option<Vec<u8>>)> for NoteCreated {
...
```
This conversion fill fit well for both cases. The only concern I have, that tuple with 4 fields might be too long.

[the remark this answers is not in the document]
phklive: I don't think that this conversion is truly needed in this case. And I do believe that the tuple is too long to be used as is if we decide to add a `From` then I would recommend using a dedicated type to represent this tuple, but I believe that it should not be needed because the code is clear enough for now and the code that we are optimizing for right now will disapear in a near refactor.

igamigo (approved):
LGTM

block-producer/src/block.rs
```diff
@@ -19,7 +19,7 @@ use crate::store::BlockInputsError;
 pub struct Block {
     pub header: BlockHeader,
     pub updated_accounts: Vec<AccountUpdateDetails>,
-    pub created_notes: Vec<(usize, usize, NoteEnvelope)>,
+    pub created_notes: Vec<NoteCreated>,
```
bobbinth, line 22: Related to the previous comment: this could have probably have been something like `Vec<(usize, usize, NoteEnvelope)>`. In general, I think we should try to avoid using protobuf types inside domain types.

A thought for the future: we should probably move the `Block` struct into `miden-base` at some point.

bobbinth (changes requested):
Looks good! Thank you! I left a couple of comments inline which I think should simplify things a little.

Also, as a part of this PR, let's also address [this](https://github.com/0xPolygonMiden/miden-node/blob/phklive-block-producer-onchain-notes/store/src/state.rs#L504) TODO.

bobbinth:
I've merged @igamigo PR into this. There we introduced [this](https://github.com/0xPolygonMiden/miden-node/blob/14a6b3e573e572ce1a156f37660f49ac4a4d26dc/store/src/db/mod.rs#L59) temporary change. I think we should address it properly as a part of this PR.

bobbinth:
> Also, as a part of this PR, let's also address [this](https://github.com/0xPolygonMiden/miden-node/blob/phklive-block-producer-onchain-notes/store/src/state.rs#L504) TODO.

Seems like @igamigo has already addressed this TODO in on of the commits in this PR.

mFragaBA:
@polydez  can we pull `next` into this? There's a fix we need to run the client's integration tests (we're currently pointing to this branch to develop and without the fix the integration tests run indefinitely on the CI)

polydez:
> @polydez can we pull `next` into this? There's a fix we need to run the client's integration tests (we're currently pointing to this branch to develop and without the fix the integration tests run indefinitely on the CI)

Sure!

bobbinth (commented):
Looks good! Thank you! I left some more comments inline. Some of these will require some changes in `miden-base`. I'll work on these and then we can finish this PR.

store/src/db/sql.rs
```diff
@@ -358,8 +359,9 @@ pub fn select_notes(conn: &mut Connection) -> Result<Vec<Note>> {
                 batch_index: row.get(1)?,
                 note_index: row.get(2)?,
                 note_id,
-                sender: column_value_as_u64(row, 4)?,
-                tag: column_value_as_u64(row, 5)?,
+                note_type: row.get::<_, u8>(4)?.try_into()?,
```
igamigo, line 362: These don't look too good versus a specific function, but I think it's good enough.

block-producer/src/test_utils/block.rs
```diff
@@ -90,7 +91,7 @@ pub struct MockBlockBuilder {
     last_block_header: BlockHeader,
 
     updated_accounts: Option<Vec<AccountUpdateDetails>>,
-    created_notes: Option<Vec<(usize, usize, NoteEnvelope)>>,
+    created_note_envelopes: Option<Vec<NoteBatch>>,
```
bobbinth, line 94: we should probably rename this back into `created_notes`.

block-producer/src/test_utils/block.rs
```diff
@@ -121,11 +122,11 @@ impl MockBlockBuilder {
...
     }
 
-    pub fn created_notes(
+    pub fn created_note_envelopes(
         mut self,
-        created_notes: Vec<(usize, usize, NoteEnvelope)>,
+        created_note_envelopes: Vec<NoteBatch>,
     ) -> Self {
-        self.created_notes = Some(created_notes);
+        self.created_note_envelopes = Some(created_note_envelopes);
```
bobbinth, line 129: Not sure if this is needed any more. Or at least the name should be different.

bobbinth (approved):
Looks good! Thank you! I left a couple of test-related comments inline - but they are non-blocking.

PR #311, opened 2024-04-10
Add note type to `NoteSyncRecord`

igamigo:
Basing it off of `phklive-block-producer-onchain` to make it simpler to test from the client.

### DIFF
--- a/proto/proto/note.proto
+++ b/proto/proto/note.proto
@@ -10,7 +10,7 @@ message Note {
     uint32 note_index = 2;
     digest.Digest note_id = 3;
     account.AccountId sender  = 4;
-    fixed64 tag = 5;
+    fixed32 tag = 5;
     merkle.MerklePath merkle_path = 7;
     // This field will be present when the note is on-chain.
     // details contain the `Note` in a serialized format.
@@ -21,7 +21,8 @@ message NoteSyncRecord {
     uint32 note_index = 1;
     digest.Digest note_id = 2;
     account.AccountId sender  = 3;
-    fixed64 tag = 4;
+    fixed32 tag = 4;
+    fixed32 note_type = 5;
     merkle.MerklePath merkle_path = 6;
 }
 
@@ -30,7 +31,7 @@ message NoteCreated {
     uint32 note_index = 2;
     digest.Digest note_id = 3;
     account.AccountId sender  = 4;
-    fixed64 tag = 5;
+    fixed32 tag = 5;
     // This field will be present when the note is on-chain.
     // details contain the `Note` in a serialized format.
     optional bytes details = 6;
--- a/store/src/db/mod.rs
+++ b/store/src/db/mod.rs
@@ -5,7 +5,7 @@ use miden_node_proto::domain::accounts::{AccountInfo, AccountSummary, AccountUpd
 use miden_objects::{
     block::BlockNoteTree,
     crypto::{hash::rpo::RpoDigest, merkle::MerklePath, utils::Deserializable},
-    notes::{NoteId, Nullifier},
+    notes::{NoteId, NoteType, Nullifier},
     BlockHeader, GENESIS_BLOCK,
 };
 use rusqlite::vtab::array;
@@ -44,15 +44,24 @@ pub struct NoteCreated {
     pub note_index: u32,
     pub note_id: RpoDigest,
     pub sender: AccountId,
-    pub tag: u64,
+    pub tag: u32,
     pub details: Option<Vec<u8>>,
 }
 
 impl NoteCreated {
     /// Returns the absolute position on the note tree based on the batch index
-    /// and local-to-the-subtree index
+    /// and local-to-the-subtree index.
     pub fn absolute_note_index(&self) -> u32 {
         BlockNoteTree::note_index(self.batch_index as usize, self.note_index as usize) as u32
+    }
+
+    // Returns the note type based on available details.
+    pub fn note_type(&self) -> NoteType {
+        if self.details.is_some() {
+            NoteType::Public
+        } else {
+            NoteType::OffChain
+        }
     }
 }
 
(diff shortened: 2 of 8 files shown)


proto/proto/note.proto
```diff
@@ -22,6 +22,7 @@ message NoteSyncRecord {
     digest.Digest note_id = 2;
     account.AccountId sender  = 3;
     fixed64 tag = 4;
+    uint32 note_type = 5;
```
bobbinth, line 25: Should we use an [Enum](https://protobuf.dev/programming-guides/proto3/#enum) here?

igamigo: Wasn't familiar with it, took a look at the other types and did not see anything better so I used u32. When I added the enum originally I think I was hitting this: https://github.com/tokio-rs/prost/issues/927. But I can delete some of the codegen and it works fine.

````rust
  --> proto/src/generated/note.rs:60:41
   |
59 | #[derive(Eq, PartialOrd, Ord, Hash)]
   |          -- first implementation here
60 | #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord,...
   |                                         ^^ conflicting implementation for `note::NoteType`
   |
   = note: this error originates in the derive macro `Eq` (in Nightly builds, run with -Z macro-backtrace for more info)
````

File: proto/proto/note.proto [hunk 2]
```diff
@@ -22,6 +22,7 @@ message NoteSyncRecord {
     digest.Digest note_id = 2;
     account.AccountId sender  = 3;
     fixed64 tag = 4;
+    note.NoteType note_type = 5;
     merkle.MerklePath merkle_path = 6;
 }
```
bobbinth, line 27: btw, while we re at it: the `tag` is now a `u32` value - so, we should probably use `fixed32` for it (here and in other messages). We could change it in a different PR though (and we'll need to handle 32-bit tags on the client as well).

igamigo: Done. We have to deal with the other changes from the client anyway.

store/src/db/mod.rs
```diff
@@ -44,16 +47,25 @@ pub struct NoteCreated {
...
     }
+
+    // Returns the note type based on available details.
+    pub fn note_type(&self) -> NoteType {
+        if self.details.is_some() {
+            NoteType::Public
+        } else {
+            NoteType::OffChain
+        }
+    }
```
bobbinth, line 68: This is probably OK for now, but note type should be a part of `NoteCreated` message (we'll do this in a different PR).

store/src/state.rs
```diff
@@ -507,7 +507,7 @@ pub fn build_note_tree(notes: &[NoteCreated]) -> Result<BlockNoteTree, ApplyBloc
             note_type,
             note.tag
                 .try_into()
-                .map_err(|_| NoteError::InconsistentNoteTag(note_type, note.tag))?,
+                .map_err(|_| NoteError::InconsistentNoteTag(note_type, note.tag.into()))?,
```
bobbinth, line 510: Would probably be better to change the error variant to take `u32` as the second value in a tuple.

bobbinth (approved):
Looks good! Thank you! I left a couple of comments inline (only one of the is actionable for now).

proto/proto/note.proto
igamigo: Never mind, it breaks the build obviously. Not sure how to fix this. Seems we generate the attributes on `build.rs` and this conflicts with some that prost is automatically adding?

```rust
 // Generate the stub of the user facing server from its proto file
    tonic_build::configure()
        .file_descriptor_set_path(&file_descriptor_path)
        .type_attribute(".", "#[derive(Eq, PartialOrd, Ord, Hash)]")
        .skip_protoc_run()
        .out_dir("src/generated")
        .compile_with_config(prost_config, protos, includes)
        .into_diagnostic()?;
```

store/src/state.rs
igamigo: This is on `miden-base`. I can make the change though.

store/src/db/mod.rs
igamigo: Yeah, was counting on that as it seemed a bit more involved and depending on the leftover refactor for removing duplicate data.

store/src/state.rs
bobbinth: Ah - in that case, it's fine to keep as is.

proto/proto/note.proto
bobbinth: OK - I'm thinking of doing the following:

- Use `fixed32` in the message.
- Use the enum in the domain object.
- Create an issue to address this in the future.

store/src/db/tests.rs
```diff
@@ -624,7 +623,7 @@ fn test_notes() {
     // test no updates
     let res = sql::select_notes_since_block_by_tag_and_sender(
         &mut conn,
-        &[(tag >> 48) as u32],
+        &[(tag >> 16)],
```
bobbinth, line 626: I don't think we need these shifts any more.

bobbinth (commented):
Looks good! Thank you!

I did just realize that we need to make one more change: in `select_notes_since_block_by_tag_and_sender`, we should remove the shift operations in SQL statement. We are now basically filtering by the entire tag. So, the SQL should look something like this:
```Rust
SELECT
    block_num,
    batch_index,
    note_index,
    note_hash,
    sender,
    tag,
    merkle_path,
    details
FROM
    notes
WHERE
    -- find the next block which contains at least one note with a matching tag
    block_num = (
        SELECT
            block_num
        FROM
            notes
        WHERE
            tag IN rarray(?1) OR sender IN rarray(?2)) AND
            block_num > ?3
        ORDER BY
            block_num ASC
        LIMIT
            1
    ) AND
    -- filter the block's notes and return only the ones matching the requested tags
    tag IN rarray(?1) OR sender IN rarray(?2));
```

bobbinth (approved):
All looks good! Thank you!

PR #317, opened 2024-04-12
fix: Return block tree absolute note index on `GetNotesById`

igamigo:


bobbinth (approved):
Looks good! Thank you!

PR #319, opened 2024-04-12
docs: cleaning up artefacts

Dominik1999:
Cleaning the docs artefacts the same way as in miden-base. The script still works and it builds the docs.


bobbinth (approved):
Looks good! Thank you!

Issue #90, opened 2023-12-08
sqlite, ensure all migrations are applied on development mode


hackaugusto:
it seems that migrations are tracked with a single integer, if the migration file is modified while in development the counter may not change, and the changes may not be applied to the local db file.

possible fixes:

- don't rely on the counter
- use a fresh db file on every dev run
    - using a in-memory db
    - using a temporary file directory

PR #321, opened 2024-04-12
Improved work with database migrations

polydez:
This solution calculates migration chain hash for current database version and compares it with expected chain hash (i.e. for SQL scripts in store's binary). If hash differs from expected, node exits with error message.

This will help us to avoid mistakes with running node with obsolete database (incompatible with migration script(s)). If one or more migration scripts were updated, this will give different migration chain hash and program fail to start. Formatting is ignored during hash calculation. Also, in the future will probably ignore comments in SQL code.

What also useful in this PR:

1. Moved migrations from string literals to separate SQL files. This will allow syntax hilighting of SQL scripts, formatting and improves readability.
2. Introduced database settings storage. This was needed for storing migration chain hash, but might be useful for other stuff.

Resolves: #90


store/src/db/migrations.rs
```diff
@@ -1,71 +1,90 @@
...
+            accumulator = Blake3_160::merge(&[accumulator, script_hash]);
+            accumulator
+        })
+        .collect()
+}
+
+fn preprocess_sql(sql: &str) -> String {
+    // TODO: We can also remove all comments here (need to analyze the SQL script in order to remain comments
+    //       in string literals).
+    remove_spaces(sql)
```
hackaugusto, line 80: This is a tricky one, one possibility is to use `PRAGMA table_info(?table)` to get the parsed table, but that would required creating a new DB, applying all the transactions, then getting the results of `table_info` from the fresh DB and comparing against the "online" DB.

I'm not sure how much work it is, so I suppose this is the best approach for now (in terms of effort to get it working).

store/src/db/migrations.rs
```diff
@@ -1,71 +1,90 @@
...
+        );
+
+        if actual_hash != Some(expected_hash) {
+            return Err(DatabaseError::UnsupportedDatabaseVersion);
+        }
+    }
+
+    MIGRATIONS.to_latest(conn).map_err(DatabaseError::MigrationError)?;
+
+    if version_before != MIGRATIONS.current_version(conn)? {
```
hackaugusto, line 52: This doesn't seem to be catching errors when the number of migrations is the same. I did the following:

- run the node to create the DB
- opened the DB via the CLI, and ran `DROP TABLE accounts; CREATE TABLE accounts (a int);` to simulate conflicting migrations
- ran the node, and it didn't detect the issue, so the code ran, and I got an error inside the code `SQLite error: no such column: account_id in SELECT account_id, account_hash FROM accounts ORDER BY block_num ASC; at offset 7`

I think we should always check the hash.

polydez: This solution is not intended to check changing of the database structure by the users outside of our migration mechanism. If you think, this worth it, I can redesign my solution, but it would require to apply all migrations to in-memory database and compare hashes with hash calculated for current database structure. For hash calculation we would need to use pragmas: `table_list`, `table_info` (or even `table_xinfo`), `index_list`, `index_xinfo`, `foreign_key_list`, and calculate hashes of the returned results. And I don't know, if it's possible to extract constraint checks. I think, this would be too complex for the problem we try to solve.

A much simpler solution might be just storing schema version after migration in `settings` table and comparing it on the next start. If database schema was changed outside of our migrations, we can raise an error and fail to start.

hackaugusto: > This solution is not intended to check changing of the database structure by the users outside of our migration mechanism

Agreed, and that is not what I was trying to propose. I did the manually edit as a test, to show that changing the migration files is not caught by this change, unless I'm missing something.

> A much simpler solution might be just storing schema version after migration in settings table and comparing it on the next start. If database structure was changed outside of our migrations, the schema version will be changed, so we can raise an error and fail to start.

Ye, that is what I was trying to get at, this `if` statement is preventing the comparison against the settings table.

polydez: > Yes, that is what I was trying to get at, this `if` statement is preventing the comparison against the settings table.

I probably misunderstand you, but this `if` is after migration is happened to the latest supported version and all checks were successful.

hackaugusto: > I probably misunderstand you, but if statement you hilighted is positioned after migration is happened to the latest supported version and all checks succeeded.

Ah, my bad. This is indeed the wrong `if`. I haven't looked deeper why the hash check didn't catch the issue.

bobbinth (approved):
Looks good! Thank you! Not a deep review from my end - but I left one minor nit inline.

crates/store/src/db/settings.rs
```diff
@@ -0,0 +1,29 @@
...
+pub struct Settings;
+
+impl Settings {
+    pub fn exists(conn: &Connection) -> Result<bool> {
+        table_exists(conn, "settings")
+    }
+
+    pub fn get_value<T: FromSql>(conn: &Connection, name: &str) -> Result<Option<T>> {
+        conn.query_row("SELECT value FROM settings WHERE name = $1", params![name], |row| {
+            row.get/*::<_, T>*/(0)
```
hackaugusto, line 14: left over from testing?

hackaugusto (approved):
LGTM

crates/store/src/db/migrations/001-init.sql
```diff
@@ -0,0 +1,67 @@
+CREATE TABLE
+    settings
+(
+    name  TEXT NOT NULL,
+    value BLOB,
+
+    PRIMARY KEY (name),
+    CONSTRAINT settings_name_is_not_empty CHECK (length(name) > 0)
+) WITHOUT ROWID;
```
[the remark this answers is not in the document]
polydez: @hackaugusto with `STRICT` flag it won't work as I expected. In order to make data human-readable and simplify the code, I use native SQLite feature for writing different types of values to the same field. I prefer strong typing, but for this case it's very convenient.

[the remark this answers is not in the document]
hackaugusto: I don't follow the rationale, can you give an example? (also, not that big of a deal, if you say it is better like this, fine with me)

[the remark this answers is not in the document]
polydez: I use settings for storing schema version (integer) and migration hash (string). I had several options for writing this into a table:
1. Write values as is (don't use `STRICT` flag). This is current approach.
2. Store values as binary encoded BLOB (in this case we would need to deserialize values from binary in order to examine them).
3. Store values as STRING (this is also human-readable, but we would need to require implementation of `FromStr`/`ToString` traits from data, serialize/parse and map errors).

According to the current usage, I think, the 3rd solution is also good, so I can rewrite `Settings` to 3rd approach.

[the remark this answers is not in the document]
polydez: I tried 3rd approach and it's not so good, as I expected due to `FromStr` error type, which is associated type (but there is also error from rusqlite, which is also should be returned). Thus, it would be easier to parse string outside of settings. Now I think, the 3rd approach is worse, than the 1st.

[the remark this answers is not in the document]
hackaugusto: Ah, I see, I was under the impression that converting to from/bytes would be easy to implement.

If that is the case, then maybe lets follow the approach [documented here](https://www.sqlite.org/flextypegood.html#attribute_tables), and instead of calling the type `BLOB` lets call `ANY` and add a comment explaining it?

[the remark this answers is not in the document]
polydez: @hackaugusto I didn't know about `ANY` datatype, thank you! Fixed it!

Issue #322, opened 2024-04-15
[Bug]: Faucet has been broken by miden-base and miden-client v0.2.0


phklive:
### Version

v0.2.0

### Other packages versions

NA

### What happened?

The Faucet has been broken by the release of miden-base and miden-client v0.2.0

### What should have happened?

It should not be broken

### How can this be reproduced?

Try to run the faucet with the miden-client at v0.2.0 it will break.

PR #323, opened 2024-04-15
Refactor node from 4 to 1 binary

phklive:
In this PR I propose the refactor of the node internal structure to enable/disable services using flags and reduce from 4 to 1 binary.

closes: #92 #322


polydez (changes requested):
Looks good to me, thank you! I've left a few inline comments. The main idea here is that we should move config parsing and service starting functionality to the node binary (leave only `serve` in the libraries). We also should union top-level configs in one single structure in the node binary with optional field for each corresponding service:

```rust
struct Config {
    rpc: Option<RpcConfig>,
    faucet: Option<FaucetConfig>,
   ....
}```

File: crates/faucet/src/handlers.rs [hunk 1]
```diff
@@ -46,45 +49,53 @@ pub async fn get_tokens(
...
-    // Serialize note for transport
-    let bytes = input_note.to_bytes();
+    let note_id: NoteId;
+
+    // Serialize note into bytes
+    let bytes = match created_notes.first() {
+        Some(note) => {
+            note_id = note.id();
+            note.to_bytes()
+        },
```
igamigo, line 90: Have you tested this on the client? I think in order to import this you might need an `InputNoteRecord`.

crates/faucet/src/config.rs
```diff
@@ -0,0 +1,36 @@
...
+        self.endpoint.to_string()
+    }
+}
+
+// Top-level config
+// ================================================================================================
+
+/// Faucet top-level configuration.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Deserialize)]
+pub struct FaucetTopLevelConfig {
```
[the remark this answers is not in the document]
phklive: Agreed

crates/block-producer/src/lib.rs
```diff
@@ -42,3 +44,14 @@ const SERVER_BUILD_BATCH_FREQUENCY: Duration = Duration::from_secs(2);
 
 /// Maximum number of batches per block
 const SERVER_MAX_BATCHES_PER_BLOCK: usize = 4;
+
+// MAIN FUNCTION
+// =================================================================================================
+
+pub async fn start_block_producer(config_filepath: &Path) -> Result<()> {
```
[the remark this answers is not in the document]
polydez: > Could you explain where in the binary you would add such functions?

I think, `commands/start` would be a good place for them.

> Also to note that the faucet is not a Tonic server but an Actix web server.

Not a problem, it can return Actix's status messages. The idea is to keep flexible errors in the binary, but return error codes from the library.

crates/block-producer/Cargo.toml
```diff
@@ -11,19 +11,12 @@ readme.workspace = true
...
-name = "miden-node-block-producer"
-path = "src/main.rs"
-bench = false
-doctest = false
-
 [features]
 tracing-forest = ["miden-node-utils/tracing-forest"]
 
 [dependencies]
 anyhow = { version = "1.0" }
```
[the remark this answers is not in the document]
polydez: Using `anyhow` or `eyre` in libraries is considered as a bad practice, because it's hard to handle different errors by library users. But binary is a different case: we usually just print error message from the binary to the user and don't need to handle it, so `anyhow` is suitable there.

crates/block-producer/src/lib.rs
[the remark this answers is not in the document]
phklive: Understood. Changed.

crates/faucet/src/config.rs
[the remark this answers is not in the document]
phklive: Done.

crates/rpc/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+pub mod server;
+
+// CONSTANTS
+// =================================================================================================
+pub const COMPONENT: &str = "miden-rpc";
+
+// MAIN FUNCTION
+// =================================================================================================
+
+pub async fn start_rpc(config_filepath: &Path) -> Result<()> {
```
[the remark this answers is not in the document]
phklive: Yes.

crates/store/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+pub mod types;
+
+// CONSTANTS
+// =================================================================================================
+pub const COMPONENT: &str = "miden-store";
+
+// MAIN FUNCTION
+// =================================================================================================
+
+pub async fn start_store(config_filepath: &Path) -> Result<()> {
```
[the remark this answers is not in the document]
phklive: Yes.

crates/faucet/src/handlers.rs
phklive: Indeed, fixed.

phklive:
Fixed all requested changes and comments. 

I have used `.expect` quite extensively considering that multiple states are irrecoverable ( example if we load the configuration file incorrectly or if the server fails to run ) let me know what you think about it or if you have a better solution in mind.

Now the commands used to run the node are the following: 

```sh
cargo run --release --features testing start node (or faucet, store, block-producer, rpc. You can also pass the --config flag to start to specify a specific config file usually located in bin/node/miden-node.toml)
```

Cargo.toml
```diff
@@ -25,6 +26,13 @@ exclude = [".github/"]
 [workspace.dependencies]
 miden-air = { version = "0.9", default-features = false }
 miden-lib = { version = "0.2"}
+miden-node-block-producer = { path = "crates/block-producer", version = "0.2" }
```
polydez, line 29: I think, it's not needed to specify version here.

crates/block-producer/src/server/mod.rs
```diff
@@ -56,9 +55,11 @@ pub async fn serve(config: BlockProducerConfig) -> Result<()> {
 
     let addr = config
         .endpoint
-        .to_socket_addrs()?
+        .to_socket_addrs()
+        .expect("Failed to turn address into socket address.")
```
polydez, line 59: Please, avoid panics in the library code. Consider adding new error enumeration, if this function can return different error types.

polydez (changes requested):
Looks good, thanks! I've added several inline comments here.

Cargo.toml
phklive: removed.

igamigo (approved):
This LGTM in general. I'm a bit confused with some of the error handling changes on component code but once that's handled it should be good to merge. Another thing that we might want to look at in the future is to have top-level commands for each component instead of them being a subcommand of the `start` (ie, `faucet start` instead of `start faucet`; could allow for more component-specific commands in the future), but probably not many good reasons right now to do so.

crates/block-producer/src/server/mod.rs
```diff
@@ -56,9 +55,11 @@ pub async fn serve(config: BlockProducerConfig) -> Result<()> {
...
         .next()
-        .ok_or(anyhow!("Couldn't resolve server address"))?;
+        .expect("Failed to resolve address.");
```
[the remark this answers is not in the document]
phklive: I agree, @polydez requested to remove the anyhow inside the libs. Looking into a better way to handle errors atm.

crates/faucet/src/handlers.rs
```diff
@@ -46,45 +51,53 @@ pub async fn get_tokens(
...
     // Instantiate transaction template
-    let tx_template = TransactionTemplate::MintFungibleAsset { asset, target_account_id };
+    let tx_template = TransactionTemplate::MintFungibleAsset(asset, target_account_id, note_type);
+
+    // Instantiate transaction request
+    let tx_request = client
+        .lock()
+        .await
+        .build_transaction_request(tx_template)
+        .expect("Failed to build transaction request.");
```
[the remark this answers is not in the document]
phklive: How do you handle situations like missing config file or others? It is going to panic at one point right?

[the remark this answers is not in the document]
igamigo: In general I think handling error is better overall, although I agree that missing config files would be an anomaly. However what I meant is that in this scenario specifically where we are handling an HTTP request, there should ideally be no Rust panics but rather the server should return the HTTP error code that best describes the situation. In this case the transaction template to transaction request conversion should work fine but it does rely on various validations, etc. that could fail and so the code should ideally not panic.

crates/utils/src/errors.rs
```diff
@@ -0,0 +1,14 @@
+#[derive(Debug)]
+pub enum ApiError {
+    ApiInitialisationFailed(String),
+    ApiServeFailed(String),
+    AddressResolutionFailed(String),
+    EndpointToSocketFailed(String),
```
polydez, line 6: We can use `io::Error` here instead of `String`.

crates/utils/src/errors.rs
```diff
@@ -0,0 +1,14 @@
...
+    DatabaseConnectionFailed(String),
+}
+
+impl std::fmt::Display for ApiError {
```
polydez, line 10: Please use `thiserror` for `ApiError` and write display implementation using its macro. We use intensively `thiserror`, for example: https://github.com/0xPolygonMiden/miden-node/blob/next/block-producer/src/errors.rs#L16-L47

crates/utils/src/errors.rs
```diff
@@ -0,0 +1,14 @@
+#[derive(Debug)]
+pub enum ApiError {
+    ApiInitialisationFailed(String),
+    ApiServeFailed(String),
+    AddressResolutionFailed(String),
```
polydez, line 5: Currently this `String` is used only for error message about address resolution fail. It would be better to not to provide `String` here, but write error message for this variant in `Display` implementation.

crates/utils/src/errors.rs
```diff
@@ -0,0 +1,14 @@
+#[derive(Debug)]
+pub enum ApiError {
+    ApiInitialisationFailed(String),
```
polydez, line 3: Here would be better to keep `tonic::Error`, don't need to convert it to string.

crates/utils/src/errors.rs
```diff
@@ -0,0 +1,14 @@
...
+    ApiServeFailed(String),
```
polydez, line 4: The same as above.

polydez (changes requested):
I'd like to improve error representation here:
1. Use `thiserror` for all error enumerations and write `Display` representation for each error variant.
2. Try to keep correct error chain without converting concrete types to `String`. Sometimes, `String` is also okay, but not for cases when we have a single error structure/enumeration which caused the error.

I've hilighted some of examples inline.

bin/node/src/commands/start.rs
```diff
@@ -0,0 +1,75 @@
...
+use tokio::task::JoinSet;
+
+use crate::config::NodeConfig;
+
+// START
+// ===================================================================================================
+
+pub async fn start_node(config: NodeConfig) -> Result<()> {
+    let mut join_set = JoinSet::new();
+    let db = Db::setup(config.store.clone().context("Missing store configuration.")?).await?;
```
polydez, line 17: I would rather spawn `start_*` functions here. This would help us to reduce/simplify code.

crates/faucet/src/errors.rs
```diff
@@ -0,0 +1,58 @@
...
+    error,
+    http::{header::ContentType, StatusCode},
+    HttpResponse,
+};
+use miden_client::errors::ClientError;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum FaucetError {
+    /// Client has submitted a bad request
```
polydez, line 11: I think, these docs are excessive here.

crates/utils/src/errors.rs
polydez: The issue is still here. It would be better to just remove this `String` argument.

phklive:
https://github.com/0xPolygonMiden/miden-node/pull/323#discussion_r1570533631 

Which errors are you talking about? All of them or specifically this one?

polydez:
> [#323 (comment)](https://github.com/0xPolygonMiden/miden-node/pull/323#discussion_r1570533631)
> 
> Which errors are you talking about? All of them or specifically this one?

Just the pointed error variant. If you look at the code which raises this error, you will see that this string initialized with error message similar to what is implemented in `Display` for this enum variant: https://github.com/0xPolygonMiden/miden-node/blob/ebc4c0d8932cf5393f2e5a6ecaafc3c0150991c2/crates/store/src/server/mod.rs#L32

I'd probably set this string with `endpoint.to_string()` or `Debug` representation of endpoint in order to give information to the user about which endpoint was failed to convert.

bin/node/src/commands/start.rs
```diff
@@ -0,0 +1,76 @@
...
+    ));
+
+    // Wait for block-producer to start & start rpc
+    tokio::time::sleep(Duration::from_secs(1)).await;
+    join_set.spawn(start_rpc(config.rpc.context("Missing rpc configuration.")?));
+
+    // block on all tasks
+    while let Some(res) = join_set.join_next().await {
+        // For now, if one of the components fails, crash the node
+        res.unwrap().unwrap();
```
polydez, line 34: I think, it would be better to raise error here:
```rust
res??;
```

polydez (changes requested):
Looks good, thanks! I've left a couple of nit comments.

crates/rpc/src/server/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    info!(target: COMPONENT, "Server initialized");
+
+    let addr = config
+        .endpoint
+        .to_socket_addrs()
+        .map_err(ApiError::EndpointToSocketFailed)?
+        .next()
+        .ok_or(config.endpoint.to_string())
+        .map_err(|err| ApiError::AddressResolutionFailed(err))?;
```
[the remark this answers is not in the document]
polydez: The second disadvantage of current solution is that we format `endpoint` as string even if no errors occurred.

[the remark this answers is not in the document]
phklive: Fixed

polydez (approved):
Looks good to me, thank you!

crates/rpc/src/config.rs
```diff
@@ -3,32 +3,29 @@ use std::fmt::{Display, Formatter};
...
-    pub database_filepath: String,
+    /// Store gRPC endpoint in the format `http://<host>[:<port>]`.
+    pub store_url: String,
+    /// Block producer gRPC endpoint in the format `http://<host>[:<port>]`.
+    pub block_producer_url: String,
 }
 
-impl FaucetConfig {
+impl RpcConfig {
     pub fn as_url(&self) -> String {
```
bobbinth, line 19: nit: maybe a better name would be `endpoint_url` (rather than `as_url`) since we are not converting the entire struct to a single URL string? This would be applicable to other config structs as well.

bin/node/miden-node.toml
bobbinth, line 1: I think we need to think through the structure of this file a bit more to prevent some possible "easy-to-make" mistakes. Let's create an issue for this.

Specifically, there are some redundant properties which are assumed to be synchronized. For example, in the file:
```toml
[block_producer]
verify_tx_proofs = true

[rpc]
endpoint = { host = "0.0.0.0", port = 57291 }
block_producer_url = "http://localhost:48046"

[store]
```
We assume that `block_producer.store_url` is the same as `store.endpoint` - but if these diverge, I don't think we report an error. A potentially better approach would be something like:
```toml
[block_producer]
verify_tx_proofs = true

[rpc]
endpoint = { host = "0.0.0.0", port = 57291 }

[store]
```
Here, relevant URLs are inferred from other present components.

bobbinth (commented):
Looks great! Thank you! I know this has already been merged, but I left a few small comments. Most of these can be separate issues which we can address in future PRs.

bin/node/miden-node.toml
```diff
@@ -19,3 +19,12 @@ store_url = "http://localhost:28943"
...
 genesis_filepath = "genesis.dat"
+
+[faucet]
+endpoint = { host = "localhost",  port = 8080 }
+database_filepath = "store.sqlite3"
+# Data used to construct the faucet account of the faucet
+asset_amount = 333
+token_symbol = "POL"
+decimals = 8
+max_supply = 1000000
```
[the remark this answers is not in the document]
phklive: Opened a pull request here: https://github.com/0xPolygonMiden/miden-node/pull/332

bin/node/miden-node.toml
phklive: Opened an issue here: https://github.com/0xPolygonMiden/miden-node/issues/334

crates/rpc/src/config.rs
phklive: Fixed in this pull request: #332

bin/node/miden-node.toml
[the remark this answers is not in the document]
phklive: Opened an issue here: #333

README.md
```diff
@@ -42,15 +42,15 @@ sudo apt install llvm clang bindgen pkg-config libssl-dev libsqlite3-dev
...
 This will install the executable `miden-node` in your PATH, at `~/.cargo/bin/miden-node`.
 
 Otherwise, if only to try the node out for testing, run:
 
 ```sh
-cargo install --features testing --path node
+cargo install --features testing --path bin/node
 ```
 
 Currently, the only difference between the two is how long the `make-genesis` command will take to run (see next subsection).
```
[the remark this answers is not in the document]
phklive: This would be the right command to use in case you would want to run the software using the cloned repo. 

If you would like to use the binary directly installed from crates.io then yes indeed `cargo install miden-node` would be the right command. 

I believe that both command could / should be documented for completeness ( having multiple ways to download / install the software ) 

Let me know what is your pov.

Cargo.toml
```diff
@@ -25,6 +26,13 @@ exclude = [".github/"]
 [workspace.dependencies]
 miden-air = { version = "0.9", default-features = false }
 miden-lib = { version = "0.2"}
+miden-node-block-producer = { path = "crates/block-producer" }
+miden-node-faucet = { path = "crates/faucet" }
+miden-node-proto = { path = "crates/proto" }
+miden-node-rpc = { path = "crates/rpc" }
+miden-node-store = { path = "crates/store" }
+miden-node-test-macro = { path = "crates/test-macro" }
+miden-node-utils = { path = "crates/utils" }
```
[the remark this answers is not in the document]
phklive: I believe that the version is inferred from the workspace cargo.toml that has the version number added.

README.md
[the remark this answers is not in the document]
bobbinth: I think for `cargo install` downloading from crates.io is probably the main use case - but I'm not against also having a section which describes how to deploy a node by cloning the repo. In general, we should probably update the section about deploying the node because we now have multiple ways (installing from crates.io, via docker, via `.deb` package, by cloning the repo) and we should describe in a more comprehensive way than we do now.

PR #325, opened 2024-04-18
Improve CI & builds

phklive:
In this PR I propose improvements on the builds and CI. 

I have found new ways of improving ci and builds, hence adding them to the node.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -16,7 +16,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
-      - name: Install cargo make
-        run: cargo install cargo-make
+      - uses: davidB/rust-cargo-make@v1
       - name: cargo make - doc
         run: cargo make doc
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,9 +14,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-        with:
-          profile: minimal
-          override: true
       - name: check rust versions
         run: ./scripts/check-rust-version.sh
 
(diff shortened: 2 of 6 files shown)


polydez (approved):
Looks good, thank you!

.github/workflows/doc.yml
```diff
@@ -16,7 +16,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
-      - name: Install cargo make
-        run: cargo install cargo-make
+      - uses: davidB/rust-cargo-make@v1
```
bobbinth, line 19: I'm still not sure we should use `davidB/rust-cargo-make`. It doesn't seem to be a very popular package (e.g., just 31 stars on Github) and I'm always weary of using such packages, mostly due to the risk of supply chain attacks.

How much faster does the CI run because of this?

.github/workflows/lint.yml
bobbinth, line 1: Same comment in this file as the one above.

.github/workflows/test.yml
```diff
@@ -17,7 +17,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
-      - name: Install cargo make
-        run: cargo install cargo-make
+      - uses: davidB/rust-cargo-make@v1
```
bobbinth, line 20: Same comment as the one above.

.github/workflows/doc.yml
phklive: It reduced the CI running time by 1 min 39 sec approximately: 
SCR-20240419-odbd

From 1 min40 sec to 1 sec thanks to a simple use of gh actions. 

But I believe that 1 min 40 sec gain in the CI is a marginal earning compared to the risk of a supply chain attack compromising the security of our chain. 

What is your point of view? Should we revert?

bobbinth: 1 min 40 sec is not negligible - but also, not something that I think will impact our processes much.

On the other hand, the potential security risk is rather small (it only affects the CI) - so, maybe not worth worrying about it too much.

Overall, I'm on the fence - but leaning toward keeping this as they are now (i.e., keep using `davidB/rust-cargo-make`).

PR #328, opened 2024-04-23
Updated the code for moved `Block` and `AccountUpdateData` to `miden-base`

polydez:
This resolves: https://github.com/0xPolygonMiden/miden-base/issues/592


crates/store/src/server/api.rs
```diff
@@ -275,32 +274,28 @@ impl api_server::Api for StoreApi {
...
-                    (true, false) => {
+                let details = AccountUpdateDetails::read_from_bytes(&account_update.details)
+                    .map_err(|err| Status::invalid_argument(err.to_string()))?;
+
+                match (account_state.account_id.is_on_chain(), details.is_private()) {
+                    (true, true) => {
                         return Err(Status::invalid_argument("On-chain account must have details"));
                     },
-                    (false, true) => {
+                    (false, false) => {
```
hackaugusto, line 284: I think in a follow up we should introduce enums to make this easier to read, something like:

```suggestion
                    (State::OnChain, Account::Private) => {
```

(I haven't put much thought on the naming, just sharing the desire to make the code better)

hackaugusto (approved):
lgtm

crates/proto/proto/requests.proto
```diff
@@ -9,9 +9,8 @@ import "note.proto";
...
 message AccountUpdate {
     account.AccountId account_id = 1;
-    digest.Digest account_hash = 2;
-    // Details for public (on-chain) account.
-    optional bytes details = 3;
+    digest.Digest final_state_hash = 2;
+    bytes details = 3;
 }
 
 message ApplyBlockRequest {
```
bobbinth, line 16: I'm wondering if we should change `ApplyBlockRequest` message to be just:
```proto
message ApplyBlockRequest {
    bytes block = 1
}
```
And the payload would be just the serialized version of the `Block` struct.

The main motivation is that it should reduce the complexity of handling different domain objects (e.g., we just reuse domain objects from `miden-base` instead of having protobuf messages and local domain objects for them).

crates/store/src/server/api.rs
```diff
@@ -258,76 +257,22 @@ impl api_server::Api for StoreApi {
...
-                    tag: note.tag,
-                    details: note.details,
-                })
-            })
-            .collect::<Result<Vec<_>, Status>>()?;
-
-        let _ = self.state.apply_block(block_header, nullifiers, accounts, notes).await;
+        let block = Block::read_from_bytes(&request.block).map_err(|err| {
+            Status::invalid_argument(format!("Block deserialization error: {err}"))
+        })?;
```
bobbinth, line 263: Question: do we still get all the same validations as we had here before? (I think we do, but wanted to double-check).

crates/store/src/db/sql.rs
```diff
@@ -490,7 +488,6 @@ pub fn select_notes_since_block_by_tag_and_sender(
         let note = Note {
             block_num,
             note_created: NoteCreated {
```
bobbinth, line 490: It is a bit confusing to me why we have these `Note` and `NoteCreated` structs. We'll probably need to revisit this in the future.

crates/store/src/state.rs
```diff
@@ -471,20 +478,10 @@ impl State {
...
-            (note.note_id, note_metadata),
-        ));
-    }
+pub fn build_note_tree<'a>(
+    notes: impl Iterator<Item = (BlockNoteIndex, &'a OutputNote)>,
+) -> Result<BlockNoteTree, ApplyBlockError> {
+    let entries = notes.map(|(note_index, note)| (note_index, note.id().into(), *note.metadata()));
 
     BlockNoteTree::with_entries(entries).map_err(ApplyBlockError::FailedToCreateNoteTree)
 }
```
bobbinth, line 487: I wonder if this should be a method on the `Block` struct - e.g., something like:
```Rust
pub fn build_note_tree(&self) -> BlockNoteTree
```

crates/store/src/state.rs
```diff
@@ -154,62 +148,70 @@ impl State {
...
-            // build notes tree
-            let note_tree = build_note_tree(&notes)?;
-            if note_tree.root() != block_header.note_root() {
+            // build note tree
+            let note_tree = build_note_tree(block.notes())?;
+            if note_tree.root() != block.header().note_root() {
                 return Err(ApplyBlockError::NewBlockInvalidNoteRoot);
             }
 
-            drop(span);
```
bobbinth, line 199: Do we no longer need this `drop(span)`?

bobbinth (commented):
Looks good! Thank you! I left some small comments inline (and also some comments for the future).

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -93,15 +92,11 @@ where
...
-            updated_accounts,
-            created_notes,
-            produced_nullifiers,
-        };
+        let block =
+            Block::new(new_block_header, updated_accounts, created_notes, produced_nullifiers);
 
         // TODO: Change to block.hash(), once it implemented
-        let block_hash = block.header.hash();
+        let block_hash = block.header().hash();
```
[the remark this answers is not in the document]
polydez: No, we had one before, but it was unimplemented, so I removed it in order to not confuse developers.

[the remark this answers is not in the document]
bobbinth: I added it recently. If you do `cargo update` it should appear :)

crates/store/src/server/api.rs
polydez: Yes, we check it during validation of proven transaction: https://github.com/0xPolygonMiden/miden-base/blob/a01390c8fc7dd5cb4937d58e17fea665b5b4d9e8/objects/src/transaction/proven_tx.rs#L84-L99

crates/store/src/db/sql.rs
polydez: Ah, you're right, it's useless now, so I merged them into one struct `Note`.

crates/store/src/state.rs
polydez: You're right, I removed it during refactoring, but reverted the part after drop, so I should restore this `drop`, thank you!

crates/store/src/server/api.rs
bobbinth: Do we not also do these checks during block deserialization?

crates/store/src/state.rs
polydez: Got it, here is PR: https://github.com/0xPolygonMiden/miden-base/pull/630

crates/store/src/server/api.rs
polydez: We don't need it, since this endpoint is not accessible outside and block-producer does this validation before block building.

crates/store/src/db/mod.rs
```diff
@@ -39,29 +43,38 @@ pub struct NullifierInfo {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-pub struct NoteCreated {
-    pub batch_index: u32,
-    pub note_index: u32,
+pub struct Note {
```
bobbinth, line 46: nit: should we rename `Note` into `NoteRecord`?

crates/store/src/db/mod.rs
```diff
@@ -39,29 +43,38 @@ pub struct NullifierInfo {
...
 
-impl NoteCreated {
+impl Note {
     /// Returns the absolute position on the note tree based on the batch index
     /// and local-to-the-subtree index.
     pub fn absolute_note_index(&self) -> u32 {
-        BlockNoteTree::note_index(self.batch_index as usize, self.note_index as usize) as u32
+        self.note_index.note_index() as u32
     }
 }
```
bobbinth, line 63: nit: I'm wondering if we actually need this `absolute_note_index()`. Seems like we can just call `note.note_index.note_index()` (i think it would have been better if it was named not `note_index()` but something like `to_absolute_index()`.

bobbinth (commented):
Looks good! Thank you! I left a couple more nits inline. Overall, once https://github.com/0xPolygonMiden/miden-base/pull/630 is merged, I think we can update and merge this one as well.

crates/store/src/server/api.rs
bobbinth: I think it would still be good to make sure the block is well-formed here - but I also think this is already happening. Are there any checks that we did in the removed code which we are not doing any more?

crates/store/src/db/mod.rs
polydez: Okay, I renamed function in that PR: https://github.com/0xPolygonMiden/miden-base/pull/630

PR #332, opened 2024-04-25
Transform `Faucet` from lib to binary

phklive:
In this PR I transform the miden-faucet from a library to a binary. 

The behaviour of the faucet was resembling more a binary than a lib that should be attached to the miden-node hence it's de-coupling.

Closing: #333


bin/node/miden-node.toml
bobbinth, line 1: Should we remove the `faucet` section from this file?

bobbinth (approved):

bin/node/miden-node.toml
```diff
@@ -10,7 +10,7 @@ verify_tx_proofs = true
 
 [rpc]
 # port defined as: sum(ord(c)**p for (p, c) in enumerate('miden-rpc', 1)) % 2**16
-endpoint = { host = "0.0.0.0", port = 57291 }
```
polydez, line 13: This suppress exposing port to the network in favour of local host connections. Is it was intended for RPC?

bin/node/miden-node.toml
phklive: Indeed, removed.

bin/node/miden-node.toml
phklive: These are some default configs for the node. A user can decide to modify them to their will afterwards. Do you believe we should leave the as is?

polydez: IMO we should provide defaults which are correct for the most frequent use cases. It's not a big mistake, but I'd prefer to keep it as is.

PR #338, opened 2024-04-29
feat: Implement `NoteMetadata` protobuf message and `NoteType` enum

### LINKED ISSUE #314: [Task]: Address `note_type` enum codegen issues

- [X] RPC
- [ ] Block producer
- [ ] Store
- [X] Protobuf messages
- [ ] Testing

### What should be done?

On https://github.com/0xPolygonMiden/miden-node/pull/311 we added `note_type` to `NoteSyncRecord`. Ideally this would be an enum in the protobuf messages, but when using something like this:

```
enum NoteType {
  PUBLIC = 0;
  OFF_CHAIN = 1;
  ENCRYPTED = 2;
}
```

The codegen creates conflicting definitions of some types due to deriving some attributes twice. See [this](https://github.com/0xPolygonMiden/miden-node/pull/311#discussion_r1559919146) for more information.

### How should it be done?

Fix the codegen when using enums

### When is this task done?

When on a clean build the types are generated correctly and the protobuf definition files contain the enums.

### Additional context

The linked PR contains mor econtext.

### LINKED ISSUE #315: [Task]: Refactor note-related messages to include a `NoteMetadata`

- [X] RPC
- [ ] Block producer
- [X] Store
- [X] Protobuf messages
- [ ] Testing

### What should be done?

Currently note-related messages in protobuf contain the same fields. These should be refactored out into its own `NoteMetadata` object

### How should it be done?

https://github.com/0xPolygonMiden/miden-node/blob/25529ac9bc629e4ac1f0d2c8bc9739b0d1bc2866/proto/proto/note.proto

Change messages to include a new `NoteMetadata` field that looks something like this:

````protobuf
message NoteMetadata {
    account.AccountId sender  = 1;
    uint32 note_type = 2;
    fixed32 tag = 3;
    fixed64 aux = 4; // unsure if we want/need this for a first iteration
}
````

and replaces the pertinent fields.

### When is this task done?

When note messages contain NoteMetadata as a specific field.

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by igamigo**
Closes #315 
Closes #314

### CODE REVIEW

File: crates/proto/src/generated/note.rs [hunk 1]
```diff
@@ -31,11 +40,37 @@ pub struct NoteSyncRecord {
...
-    pub tag: u32,
-    #[prost(uint32, tag = "5")]
-    pub note_type: u32,
-    #[prost(message, optional, tag = "6")]
+    pub metadata: ::core::option::Option<NoteMetadata>,
+    #[prost(message, optional, tag = "4")]
     pub merkle_path: ::core::option::Option<super::merkle::MerklePath>,
 }
+#[derive(Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
```
igamigo, line 48: I have yet to avoid these conflicting derives from happening without making something too custom on `builld.rs`.

igamigo: Turns out we did not need those custom attributes, I thought I had tested it but apparently not (probably was testing it along other modifications I had made).

crates/proto/src/domain/notes.rs
```diff
@@ -0,0 +1,33 @@
...
+    type Error = ConversionError;
+
+    fn try_from(value: crate::generated::note::NoteMetadata) -> Result<Self, Self::Error> {
+        let sender = value
+            .sender
+            .ok_or_else(|| crate::generated::note::NoteMetadata::missing_field("Sender"))?
+            .try_into()?;
+        let note_type = NoteType::try_from(value.note_type as u64)?;
+        let tag = NoteTag::from(value.tag);
+        let aux = Felt::new(value.aux);
```
bobbinth, line 18: We should probably use `try_from()` here as well as this will silently truncate `u64` range to field element range.

crates/store/src/db/sql.rs
```diff
@@ -396,15 +401,14 @@ pub fn insert_notes(transaction: &Transaction, notes: &[NoteRecord]) -> Result<u
...
             note.block_num,
             note.note_index.batch_idx(),
             note.note_index.note_idx_in_batch(),
             note.note_id.to_bytes(),
-            note.note_type as u8,
-            u64_to_value(note.sender),
-            note.tag,
+            note.metadata.note_type() as u8,
+            u64_to_value(note.metadata.sender().into()),
+            note.metadata.tag().inner(),
```
bobbinth, line 411: Seems like we are not storing `metadata.aux` field in the database yet. We should probably make an issue to do that (unless one already exists?).

crates/store/src/db/sql.rs
```diff
@@ -475,22 +479,27 @@ pub fn select_notes_since_block_by_tag_and_sender(
...
         let merkle_path = MerklePath::read_from_bytes(merkle_path_data)?;
         let details_data = row.get_ref(8)?.as_blob_or_null()?;
         let details = details_data.map(<Vec<u8>>::read_from_bytes).transpose()?;
 
+        let metadata = NoteMetadata::new(
+            sender.try_into()?,
+            NoteType::try_from(note_type)?,
+            tag.into(),
+            Default::default(),
+        )?;
```
bobbinth, line 495: I'd leave a `TODO` here to properly handle `metadata.aux` field.

crates/store/src/db/sql.rs
```diff
@@ -538,14 +547,19 @@ pub fn select_notes_by_id(conn: &mut Connection, note_ids: &[NoteId]) -> Result<
         let details_data = row.get_ref(8)?.as_blob_or_null()?;
         let details = details_data.map(<Vec<u8>>::read_from_bytes).transpose()?;
 
+        let note_type = row.get::<_, u8>(4)?.try_into()?;
+        let sender = column_value_as_u64(row, 5)?;
+        let tag: u32 = row.get(6)?;
+
+        let metadata =
+            NoteMetadata::new(sender.try_into()?, note_type, tag.into(), Default::default())?;
```
bobbinth, line 555: I'd leave a `TODO` here to properly handle `metadata.aux` field.

bobbinth (commented):
Thank you! Looks good! I left a few comments inline. Most are minor - but three is one in which I wonder if using enum for note type was actually a good idea.

crates/store/src/db/sql.rs
igamigo: Created https://github.com/0xPolygonMiden/miden-node/issues/342

crates/proto/build.rs
```diff
@@ -7,11 +7,9 @@ fn main() -> miette::Result<()> {
     // Compute the directory of the `proto` definitions
     let cwd: PathBuf = env::current_dir().into_diagnostic()?;
     let proto_dir: PathBuf = cwd.join("proto");
-
```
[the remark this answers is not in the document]
igamigo: Reverted!

crates/proto/proto/note.proto
```diff
@@ -5,24 +5,33 @@ import "digest.proto";
 import "merkle.proto";
 import "account.proto";
 
+enum NoteType {
+  PUBLIC = 0;
+  OFF_CHAIN = 1;
+  ENCRYPTED = 2;
+}
```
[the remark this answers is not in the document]
igamigo: Yeah, not sure about this. There cannot be protobuf enums that don't start with 0 so it's either creating a phantom variant, creating a custom mapping function (from protobuf's note type to base note type), or just removing it altogether and using a `fixed32`. For making sure they stay synced I can add a test that asserts this.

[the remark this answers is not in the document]
igamigo: On second thought it's not worth the added complexity if we can just convert the note type number. Removed in favor of a `fixed32`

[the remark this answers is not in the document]
bobbinth: Agreed! I came to the same conclusion. The only question - should we use `int32` instead? (this way, encoding would take 1 byte instead of 4, but otherwise should be the same).

crates/store/src/db/sql.rs
```diff
@@ -347,13 +347,19 @@ pub fn select_notes(conn: &mut Connection) -> Result<Vec<NoteRecord>> {
         let details_data = row.get_ref(8)?.as_blob_or_null()?;
         let details = details_data.map(<Vec<u8>>::read_from_bytes).transpose()?;
 
+        let note_type = row.get::<_, u8>(4)?.try_into()?;
+        let sender = column_value_as_u64(row, 5)?;
+        let tag: u32 = row.get(6)?;
+
+        // TODO: Properly handle note metadata
```
bobbinth, line 354: nit: I would change the comment to "TODO: properly handle metadata aux".

bobbinth (approved):
Looks good! Thank you! In addition to the small comment I left inline, I think we should change the type of `note_type` in protobuf message from `fixed32` to `int32`.

crates/proto/proto/note.proto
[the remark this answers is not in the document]
igamigo: Changed to `uint32`. Seemed similar in terms of encoding low values but marginally more correct for the use case.

PR #339, opened 2024-05-02
feat: add the option to mint public notes with the faucet

### LINKED ISSUE #329: [Feature]: Add public notes to the faucet

End user

Alice loves the new faucet and that she can get a private note with some tokens. Now, she wants also to check out the second flow and receive a public note via the blockchain. 

### How is this feature used?

Alice wants to be able to click either a button "Public notes" or a button "Private notes". 

If she clicks "Public notes" the faucet should execute a transaction using `--note-type public`. If she clicks "Private notes" the type should be `--note-type private`.

### Anything else?

We could add a short explanation how to consume the note then. But we can also add this later

tomyrd:
closes #329 

The previous faucet only minted private notes which had to be downloaded and imported into the client. Now with the implementation of public notes the flow can be much simpler. Just mint the note, sync and consume. This PR adds the option to mint public notes with the faucet and adds some useful information about how to consume it.

The initial screen looks like this:
Screenshot 2024-05-02 at 14-48-05 Miden Faucet

After minting a private note this information box is showed:
Screenshot 2024-05-02 at 14-47-27 Miden Faucet

If the note is public the commands are slightly different:
Screenshot 2024-05-02 at 14-47-50 Miden Faucet


tomyrd:
Which type of note (public or private) do you think should be selected by default?

bin/faucet/src/static/index.js
```diff
@@ -56,6 +78,21 @@ document.addEventListener('DOMContentLoaded', function () {
         }
     }
 
+    function toggleVisibility(clickedPrivate) {
```
igamigo, line 81: If this function changes the state of the global variable used in the request, I'd maybe name it something like "onNoteTypeButtonClick".

bin/faucet/src/static/index.js
```diff
@@ -56,6 +78,21 @@ document.addEventListener('DOMContentLoaded', function () {
...
+        if (clickedPrivate) {
+            privateButton.classList.add('active');
+            privateButton.classList.remove('inactive');
```
igamigo, line 84: nit: Maybe this could be factored out to a function since it occurs 4 times

bin/faucet/src/static/index.js
```diff
@@ -34,19 +45,30 @@ document.addEventListener('DOMContentLoaded', function () {
...
             }
 
             const blob = await response.blob();
-            downloadBlob(blob, 'note.mno');
+            if(isPrivateNote) {
+                importCommand.style.display = 'block';
+                downloadBlob(blob, 'note.mno');
+            }
+
+            const noteId = response.headers.get('Content-Disposition').split('filename=')[1].replace(/"/g, '');
```
igamigo, line 68: This looks a bit complicated but also it might be fine for now. I guess an alternative would be for the name could be sent directly through a custom header to make it a bit simpler.
Alternatively the server could return a JSON response that contains the type of response, and (depending on the user selection) the encoded note details.

bobbinth:
One suggestions:

I would make the amount selectable (and maybe give a few options there), but would make public note and private notes into  separate button.

So, basically:
- We'd have 1 drop-down with amount selections - something like: 100, 500, 1000 (default 100).
- We'd have 2 buttons - "Send public note" and "Send private note".

mFragaBA:
>     * We'd have 1 drop-down with amount selections - something like: 100, 500, 1000 (default 100).

If this is done, we should validate these amounts on the server side so no one consumes an arbitrary amount of tokens, shouldn't we?

bobbinth:
> If this is done, we should validate these amounts on the server side so no one consumes an arbitrary amount of tokens, shouldn't we?

I would just hard-code a few amounts and not allow free entry (mostly for simplicity). But I guess we will need to have a check on the server side if we send amounts with the request.

bin/faucet/src/handlers.rs
```diff
@@ -105,7 +113,7 @@ pub async fn get_tokens(
         .append_header(header::ContentDisposition {
             disposition: actix_web::http::header::DispositionType::Attachment,
             parameters: vec![actix_web::http::header::DispositionParam::Filename(
-                "note.mno".to_string(),
+                note_id.to_string(),
```
mFragaBA, line 122: can we keep the `.mno` extension?

tomyrd:
> I would make the amount selectable (and maybe give a few options there), but would make public note and private notes into separate button.

Done! The new screenshots are pasted in the PR description.

> If this is done, we should validate these amounts on the server side so no one consumes an arbitrary amount of tokens, shouldn't we?

I added the possible options to the `miden-faucet.toml` file and the amount requested is always checked we should be safe from arbitrary requests.

bin/faucet/src/static/index.js
mFragaBA: how about something like `toggleNoteTypeButton` and rename the parameter to `createPrivate` or similar? I think it describes better wheat the function does

tomyrd: Yes, that name wasn't very descriptive but in the new version we don't need that function anymore.

bobbinth:
> > I would make the amount selectable (and maybe give a few options there), but would make public note and private notes into separate button.
> 
> Done! The new screenshots are pasted in the PR description.

Looks great! Thank you!

Probably not for this PR: we should add some basic explanations about what public vs. private notes are (maybe on hover of the respective buttons, but maybe in some other way).

bin/faucet/src/static/index.js
tomyrd: Done! Didn't know about custom headers.

bin/faucet/src/static/index.html
```diff
@@ -16,8 +16,23 @@ <h1 id="title">Miden Faucet</h1>
...
+            <br>
+            <button id="button-public">Send Public Note</button>
+        </div>
+    </div>
+    <div id="info-container">
+        <p id="info-message">You can consume the note by running the following commands in your shell</p>
+        <p id="client-commands">
+            <span id="import-command" style="display: none;">miden-client input-notes import ./path/to/downloaded/note<br></span>
+            miden-client sync <br>
+            miden-client tx new consume-notes <span id="command-account-id"></span> <span id="note-id"></span>
```
mFragaBA, line 34: this is because now a `--account` must precede the account id

```suggestion
            miden-client tx new consume-notes --account <span id="command-account-id"></span> <span id="note-id"></span>
```

mFragaBA (commented):
LGTM! Tested locally and had no issues. I left a small comment regarding the commands used to consume the note.
Also on a side note, now we'll need to keep in mind that changes to the client might impact here as well

bobbinth:
@tomyrd - oh, and let's update changelog.

igamigo (approved):
Agreed, let's merge after the changelog is updated.

mFragaBA (approved):
agreed on the changelog

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.2.1 (2024-04-27)
 
+* Added option to mint pulic notes in the faucet (#339).
```
bobbinth, line 5: This should go into `v0.3.0` section (you may need to rebase from the latest `next`).

bobbinth:
The tests are failing because of the recent changes in `miden-base`. I'll merge this PR as is and we'll fix tests in a separate issue.

phklive:
@tomyrd I believe that this PR has removed the loading spinner when the user clicks on one of the buttons. 

This creates a UX where the user does not know if the app is loading waiting for a response from the rollup or if the website is broken.

I believe that we should add it back in another PR.

phklive:
Great job btw.

tomyrd:
> I believe that we should add it back in another PR.

Done. The PR is #362 .

PR #340, opened 2024-05-02
Implemented raw block storage and RPC endpoint for retrieving raw block by number

polydez:
Resolves https://github.com/0xPolygonMiden/miden-node/issues/316


crates/store/src/config.rs
```diff
@@ -17,6 +17,8 @@ pub struct StoreConfig {
     pub database_filepath: PathBuf,
     /// Genesis file
     pub genesis_filepath: PathBuf,
+    /// Blocks directory
+    pub blocks_dir: PathBuf,
```
bobbinth, line 21: I'm wondering if `blockstore_dir` is a better name.

crates/store/src/server/api.rs
```diff
@@ -265,6 +268,8 @@ impl api_server::Api for StoreApi {
         // TODO: Why the error is swallowed here? Fix or add a comment with explanation.
         let _ = self.state.apply_block(block).await;
 
+        self.block_storage.save_block(block_num, &request.block).await?;
```
bobbinth, line 271: Should this not be a part of `state.apply_block()`? The main reason is that if `state.apply_block()` succeeds, but `block_storage.save_block()` fails, we'll end up in an inconsistent state.

The downside is that we'll need to serialize the block (instead of just getting `request.block`) but maybe that's OK?

crates/store/src/blocks.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    fn block_path(&self, block_num: u32) -> PathBuf {
+        self.blocks_dir.join(format!("{block_num:08x}.bin"))
+    }
+}
+
+impl BlockStorage for BlockStorageDefault {
+    async fn save_block(&self, block_num: u32, data: &[u8]) -> Result<(), std::io::Error> {
+        tokio::fs::write(self.block_path(block_num), data).await
```
bobbinth, line 35: If the file exists, this would overwrite it, right?

bin/node/miden-node.toml
```diff
@@ -19,3 +19,4 @@ store_url = "http://localhost:28943"
 endpoint = { host = "localhost", port = 28943 }
 database_filepath = "miden-store.sqlite3"
 genesis_filepath = "genesis.dat"
+blocks_dir = "blocks"
```
bobbinth, line 22: Similar to one of the previous comments: I would change "blocks" to "blockstore".

crates/store/src/blocks.rs
```diff
@@ -0,0 +1,45 @@
...
+
+impl BlockStorageDefault {
+    pub async fn new(blocks_dir: PathBuf) -> Result<Self, std::io::Error> {
+        tokio::fs::create_dir_all(&blocks_dir).await?;
+
+        Ok(Self { blocks_dir })
+    }
+
+    fn block_path(&self, block_num: u32) -> PathBuf {
+        self.blocks_dir.join(format!("{block_num:08x}.bin"))
```
bobbinth, line 29: Couple nits:

- I would change the extension from `.bin` to `.dat` because `.bin` could be a bit ambiguous.
- I would just use `block_` prefix (instead of `block_num:`). So, `block_12345.dat`.

crates/proto/proto/rpc.proto
```diff
@@ -12,4 +12,5 @@ service Api {
     rpc SyncState(requests.SyncStateRequest) returns (responses.SyncStateResponse) {}
     rpc SubmitProvenTransaction(requests.SubmitProvenTransactionRequest) returns (responses.SubmitProvenTransactionResponse) {}
     rpc GetAccountDetails(requests.GetAccountDetailsRequest) returns (responses.GetAccountDetailsResponse) {}
+    rpc GetBlockByNumber(requests.GetBlockByNumberRequest) returns (responses.GetBlockByNumberResponse) {}
```
bobbinth, line 15: nit: let's alphabetize this list.

bobbinth (commented):
Looks good! Thank you! I left a few comments inline, the main one is about making sure the store always stays in a consistent state (even if block writing fails).

bobbinth (approved):
Looks good! Thank you! I left a couple of non-blocking comments inline.

Also:
- Let's create an issue to figure out how to do streaming.
- The number of block we'll create per year is probably going to be over 10 million and I'm not sure much of a problem it would be. Regardless of this, we may want to figure out some strategy for storing blocks in sub-directories (e.g., define an epoch and store blocks for each epoch in a separate sub-directory). Let's create an issue for this.

crates/store/src/state.rs
```diff
@@ -79,6 +79,11 @@ impl State {
         Ok(Self { db, inner, writer })
     }
 
+    /// Returns a reference to the database.
+    pub fn db(&self) -> Arc<Db> {
+        self.db.clone()
+    }
```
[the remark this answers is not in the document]
polydez: I was used outside of the state, but now removed.

polydez:
@bobbinth:
> * The number of block we'll create per year is probably going to be over 10 million and I'm not sure much of a problem it would be. Regardless of this, we may want to figure out some strategy for storing blocks in sub-directories (e.g., define an epoch and store blocks for each epoch in a separate sub-directory). Let's create an issue for this.

Yes, I agree with you, such a big number of files in a single directory might be inconvenient to manage. I introduced epochs in this PR, please take a look! The main idea here is that we get first 16 bits of block number as an epoch, it means, each epoch will store 65536 blocks, it's about 7.5 days per epoch with current block issue frequency.

crates/store/src/state.rs
```diff
@@ -236,17 +241,10 @@ impl State {
...
-                notes,
-                block.created_nullifiers().to_vec(),
-                block.updated_accounts().to_vec(),
-            )
-            .await
-        });
+        let handle =
+            tokio::spawn(
+                async move { db.apply_block(allow_acquire, acquire_done, block, notes).await },
+            );
```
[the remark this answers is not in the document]
polydez: I moved `blockstore` into `State`. Regarding simplification, if you mean removing `tokio::spawn` and communication using channels here, I encourage that. I also proposed to remove this construction long time ago. :) I don't know what benefits we get from this now, since once transaction commit fails, we still put in-memory storage into inconsistent state with the database.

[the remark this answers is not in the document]
bobbinth: > I don't know what benefits we get from this now, since once transaction commit fails, we still put in-memory storage into inconsistent state with the database.

I don't think that's the case: if transaction commit fails, the in-memory data structures are not updated - i.e., the `mem::replace` statements a few lines down are not executed (or at least, that's the intent).

In general, this mechanism is used to support the following process:

1. We start the database transaction and update the database. While this is going on, reads against the database are allowed as they read the previous state. So, the store can still serve all requests.
2. Once the data has been written to the database, we lock the in-memory structures for update. At this point, the store blocks on all requests.
3. We update in-memory structures. While this is going on, the store cannot serve any requests - so, it is critical to make this section as fast as possible.
4. We commit the database transaction and release the lock on in-memory structures.

Because database transaction and in-memory structures live in different places, we need this channel - but I am wondering if there is some way we can refactor the structure to make the code simpler.

crates/store/src/blocks.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    // HELPER FUNCTIONS
+    // --------------------------------------------------------------------------------------------
+
+    fn block_path(&self, block_num: u32) -> PathBuf {
+        let epoch = block_num >> 16;
+        let epoch_dir = self.store_dir.join(format!("{epoch:04x}"));
+        epoch_dir.join(format!("block_{block_num:08x}.dat"))
+    }
```
bobbinth, line 41: I think this is fine for now, but I would still create an issue to discuss what should be the length of an epoch. For example, our plan is to reduce block times to 3 - 4 seconds, in which case, the current approach would mean that an epoch is just 2 - 3 days. This may be OK but also may be too short.

bobbinth (commented):
Thank you! Looks good! I left a few more comments inline.

crates/store/src/state.rs
```diff
@@ -236,16 +243,9 @@ impl State {
...
-            db.apply_block(
-                allow_acquire,
-                acquire_done,
-                block.header(),
-                notes,
-                block.created_nullifiers().to_vec(),
-                block.updated_accounts().to_vec(),
-            )
-            .await
+            db.apply_block(allow_acquire, acquire_done, block_store, block, notes).await
```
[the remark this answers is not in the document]
bobbinth: Actually, this may need to happen before line 258 as I think that's when the database transaction gets committed (might be worth adding a comment there).

crates/store/src/state.rs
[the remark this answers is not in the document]
polydez: Thank you for explanation, now it's much clearer for me. According to the explanation, current implementation still has bugs (3 actually):
1. When we lock in-memory data, we don't wait until commit is succeed and start to update in-memory. In a case of failed commit, in-memory data will be updated.
2. Once we finish update in-memory data, we unlock it instantly without waiting for finishing of database update.
3. Error result of `db.apply_block` in `tokio::spawn` is swallowed.

crates/store/src/blocks.rs
polydez: If we use epoch only for limiting file count in the directory, I think, current epoch size is good. Even if we want to use it for something else, 2-3 days is still good (for example, in Solana each epoch lasts for about 2 days).

crates/store/src/state.rs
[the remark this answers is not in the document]
bobbinth: > 1. When we lock in-memory data, we don't wait until commit is succeed and start to update in-memory. Even in a case of failed commit, in-memory data will be updated. By the way, we have a comment about this below.

I'm not sure that's correct. Currently, we make copies of the data and start updating them (this is very inefficient, and we'll need to improve it in the future, but that's how it works now). Then, once the transaction is committed, we do `mem::replace` to swap out the updated copies.

> 2. Once we finish update in-memory data, we unlock it instantly without waiting for finishing of database update.

I think this is similar to the point above: in-memory updates don't actually happen until `mem::replace` calls, and these happen after the database transaction is committed.

> 3. Error result of `db.apply_block` in `tokio::spawn` only logged, but not re-raised or handled.
> 
> The simplest solution would be just moving all these `mem::replace`s into `Ok(Ok(()))` branch of result handling below (line 283).

This part may be correct - but it is difficult for me to tell how error propagation works here. So, probably another good reason to try to refactor the code (and document it more) to make it super clear what's happening.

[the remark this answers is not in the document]
polydez: @bobbinth, it actually works differently. 

Please look at these lines: https://github.com/0xPolygonMiden/miden-node/blob/816d668f03f3652183a1f16a65c1d6c59faf2cf9/crates/store/src/db/mod.rs#L306-L309

Here our spawned task sends notification that it's a time for acquiring the lock and then waits for the response. Response is sent by the caller task just after locking (the caller awaits for the notification first, then acquires the lock and sends another notification, after that it immediately starts to update in-memory data - without waiting for commit): https://github.com/0xPolygonMiden/miden-node/blob/816d668f03f3652183a1f16a65c1d6c59faf2cf9/crates/store/src/state.rs#L253-L269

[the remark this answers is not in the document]
bobbinth: Hmmm - I'll need to look into this more. Let's create an issue (and link this discussion there). We can address it as a part of the next milestone.

crates/store/src/state.rs
```diff
@@ -52,8 +53,12 @@ struct InnerState {
 
 /// The rollup state
 pub struct State {
+    /// The underlying database.
     db: Arc<Db>,
```
bobbinth, line 57: I would update the comment to something like: "The database which stores block headers, nullifiers, notes, and the latests states of accounts."

crates/store/src/state.rs
```diff
@@ -52,8 +53,12 @@ struct InnerState {
...
 
+    /// The underlying block store.
+    block_store: Arc<BlockStore>,
```
bobbinth, line 60: I would change the comment to something like: "The block store which stores full block contents for all blocks."

bobbinth (approved):
Looks good! Thank you! I left a couple of minor comments inline. Once these are addressed, let's fix the merge conflicts, and then merge.

Also, let's create an issue for refactoring of the `apply_block()` process as we discussed in the comments above.

crates/store/src/blocks.rs
bobbinth: Sounds good! I guess we can always introduce a higher-level periods that would be something like top 14 bits.

One potential change here: should epoch be formatted as: `{epoch:08x}` rather than `{epoch:04x}`?

PR #345, opened 2024-05-06
feat: Return block authentication data with `GetBlockHeaderByNum`

### LINKED ISSUE #256: Usability: Add an easy way for a user to retrieve an inclusion proof after having synced over its block
Given the following scenario:

- Note `N` is created and included in block `X`
- Current block height is `X+N`
- User syncs with a block higher than `X`
- User learns about note `N` and wants to consume it

The `sync` endpoint, as it is, doesn't cover the situation above. So a user may never see the inclusion proof for the block `N`, and therefor never be able to consume it.

igamigo:
Closes #256

### DIFF
--- a/crates/proto/proto/requests.proto
+++ b/crates/proto/proto/requests.proto
@@ -14,6 +14,8 @@ message CheckNullifiersRequest {
     repeated digest.Digest nullifiers = 1;
 }
 
+// Returns the block header corresponding to the requested block number, as well as the merkle
+// path that validates the block's inclusion in the chain.
 message GetBlockHeaderByNumberRequest {
     // The block number of the target block.
     //
--- a/crates/proto/proto/requests.proto
+++ b/crates/proto/proto/requests.proto
@@ -24,7 +24,7 @@ message GetBlockHeaderByNumberRequest {
     // If not provided, means latest know block.
     optional uint32 block_num = 1;
     // Whether or not to return authentication data for the block header.
-    bool include_authentication = 2;
+    optional bool include_mmr_proof = 2;
 }
 
 // State synchronization request.
--- a/crates/proto/proto/requests.proto
+++ b/crates/proto/proto/requests.proto
@@ -17,7 +17,7 @@ message CheckNullifiersRequest {
 // Returns the block header corresponding to the requested block number, as well as the merkle
 // path and current forest which validate the block's inclusion in the chain.
 //
-// The merkle path is an MMR proof for the block's leaf, based on the current forest.
+// The Merkle path is an MMR proof for the block's leaf, based on the current chain length.
 message GetBlockHeaderByNumberRequest {
     // The block number of the target block.
     //
--- a/crates/proto/proto/requests.proto
+++ b/crates/proto/proto/requests.proto
@@ -15,7 +15,7 @@ message CheckNullifiersRequest {
 }
 
 // Returns the block header corresponding to the requested block number, as well as the merkle
-// path that validates the block's inclusion in the chain.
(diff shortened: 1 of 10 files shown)


crates/proto/proto/responses.proto
```diff
@@ -17,7 +17,14 @@ message CheckNullifiersResponse {
...
 
 message GetBlockHeaderByNumberResponse {
+    // The requested block header
     block_header.BlockHeader block_header = 1;
+
+    // Merkle path to verify the block's inclusion in the MMR at the returned `forest`
+    optional merkle.MerklePath proof = 2;
+
+    // Current value of the MMR forest
+    optional fixed32 forest = 3;
```
bobbinth, line 27: Some nits about naming:

- I would probably rename `proof` into `mmr_path`.
- I would rename `forest` into something like `chain_length`.

Alternatively, we could introduce `MmrProof` message which would be similar to the `MmrProof` struct from `miden-crypto` - but `position` there would be somewhat redundant with the data in the block header itself. So, not sure if that's a good idea.

bobbinth (commented):
Looks good! Thank you! I left a couple more comments inline.

crates/proto/proto/responses.proto
igamigo: I thought about reducing it to `MmrProof` but didn't because of what you mentioned. It does not seem necessary to me right now but I don't mind adding it if anyone has a different opinion.

crates/proto/proto/responses.proto
```diff
@@ -17,7 +17,14 @@ message CheckNullifiersResponse {
...
 
 message GetBlockHeaderByNumberResponse {
+    // The requested block header
     block_header.BlockHeader block_header = 1;
+
+    // Merkle path to verify the block's inclusion in the MMR at the returned `forest`
+    optional merkle.MerklePath mmr_path = 2;
+
+    // Current value of the MMR forest
+    optional fixed32 chain_length = 3;
```
bobbinth, line 27: nit: update mentions of the `forest` in the comments.

crates/proto/proto/requests.proto
```diff
@@ -14,11 +14,17 @@ message CheckNullifiersRequest {
     repeated digest.Digest nullifiers = 1;
 }
 
+// Returns the block header corresponding to the requested block number, as well as the merkle
+// path and current forest which validate the block's inclusion in the chain.
+//
+// The merkle path is an MMR proof for the block's leaf, based on the current forest.
```
bobbinth, line 20: nit: update mention of `forest` in the comments.

crates/rpc/README.md
```diff
@@ -34,7 +34,7 @@ Gets a list of proofs for given nullifier hashes, each proof as a sparse Merkle
 
 ### GetBlockHeaderByNumber
 
-Retrieves block header by given block number.
+Retrieves block header by given block number, optionally alongside a Merkle path and the current MMR forest to validate its inclusion.
```
bobbinth, line 37: nit: mention of `forest`.

crates/store/README.md
```diff
@@ -48,7 +48,7 @@ Get a list of proofs for given nullifier hashes, each proof as a sparse Merkle T
 
 ### GetBlockHeaderByNumber
 
-Retrieves block header by given block number.
+Retrieves block header by given block number. Optionally, it also returns the MMR path and current forest to authenticate the block's inclusion.
```
bobbinth, line 51: nit: mention of `forest`.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - all related to updating the use of `forest` in docs/comments.

PR #347, opened 2024-05-08
Save blocks asynchronously

polydez:
I noticed that after refactoring, block saving, which is blocking, was moved from background thread to the async runtime. This can affect other async tasks, so I implemented async method as well. Unfortunately we can't simply get rid of blocking method now, since it is used in synchronous runtime (in genesis block storing).

### DIFF
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -257,7 +257,7 @@ impl State {
             let mut inner = self.inner.write().await;
 
             // Save the block before transaction is committed:
-            self.block_store.save_block(block_num, &block_data)?;
+            self.block_store.save_block(block_num, &block_data).await?;
 
             let _ = inform_acquire_done.send(());
 
--- a/crates/store/src/blocks.rs
+++ b/crates/store/src/blocks.rs
@@ -12,23 +12,30 @@ impl BlockStore {
         Ok(Self { store_dir })
     }
 
-    pub fn save_block(&self, block_num: u32, data: &[u8]) -> Result<(), std::io::Error> {
-        let block_path = self.block_path(block_num);
-        let epoch_path = block_path.parent().ok_or(std::io::Error::from(ErrorKind::NotFound))?;
-
-        if !epoch_path.exists() {
-            std::fs::create_dir_all(epoch_path)?;
-        }
-
-        std::fs::write(block_path, data)
-    }
-
     pub async fn load_block(&self, block_num: u32) -> Result<Option<Vec<u8>>, std::io::Error> {
         match tokio::fs::read(self.block_path(block_num)).await {
             Ok(data) => Ok(Some(data)),
             Err(err) if err.kind() == std::io::ErrorKind::NotFound => Ok(None),
             Err(err) => Err(err),
         }
+    }
+
+    pub async fn save_block(&self, block_num: u32, data: &[u8]) -> Result<(), std::io::Error> {
+        let (epoch_path, block_path) = self.epoch_block_path(block_num)?;
+        if !epoch_path.exists() {
+            tokio::fs::create_dir_all(epoch_path).await?;
+        }
+
+        tokio::fs::write(block_path, data).await
+    }
+
+    pub fn save_block_blocking(&self, block_num: u32, data: &[u8]) -> Result<(), std::io::Error> {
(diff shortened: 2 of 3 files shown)


polydez:
@hackaugusto 
> Just for my understanding, what is the motivation around saving the blocks as files instead of in the DB? And why multiple files instead of a single append-only one? (Lots of file systems don't do well with too many files in a single folder)

We didn't want our database to grow such fast: database gives no benefits to us in such case, since we store only blobs at the moment. It would also be more expensive to shrink the DB when garbage collecting of obsolete blocks. Storing in a single append-only file has also its drawbacks: removing obsolete blocks becomes much more expensive and we also would need to index block offsets within it. As a compromise, it would be grouping blocks by chunks (for example, 1k blocks per file), but we decided for implement easiest and flexible solution for now. By the way, the number of blocks in each folder is limited now to 65536 and we will lower this number, if face any difficulties with this matter.

hackaugusto:
> We didn't want our database to grow such fast

I don't understand the concern here. The one file per block uses more disk space since disk writes are page aligned. This doesn't introduce compression and removes the packing done by the DB resulting in a worse occupancy rate.

> It would also be more expensive to shrink the DB when garbage collecting of obsolete blocks

I don't follow. I would estimate the cost here to be 1. Number of pages touched by the changes 2. Number of syscalls needed to do the update.

1. I believe the number of pages with a DB would be smaller, since the data is packed in a b-tree and saved one page at the time, whereas one file per header has no packing and would have one page per file
3. This should also be better for the DB, it should be something like 6 syscalls for the DB (write/sync/flush for the wal + db file) vs. `n+2` for the individual files (open + read to list files + one remove per file)

> Storing in a single append-only file has also its drawbacks: removing obsolete blocks becomes much more expensive and we also would need to index block offsets within it in the DB. As a compromise solution, it would be grouping blocks by chunks (for example, 1k blocks per file), but we decided to implement easiest but still flexible solution for now

Yeah, a chunk file would make sense if removal is to be expected.

hackaugusto:
As a follow up to the above comment. IMO it makes sense to:

- Have a single file with the headers of a single epoch
- The above file would be named with the epoch, so to get a block it as simple as `open(block_num / blocks_per_epoch)` followed by `seek(block_num % blocks_per_epcoh * block_in_bytes)`
- Having a single file per epoch also makes it easier to send data across the network with `send_file`, to delete data in batches (by deleting a single file), increases occupancy and so on

I personally don't have strong feeling about the consistency with the DB.

polydez:
@hackaugusto we store full blocks, not only block headers. But grouping block sets in files is a good solution, we discussed it with @bobbinth while were choosing right architecture. For storing blocks with arbitrary size it will require indexing I mentioned before, so we chose not ideal, but the simplest solution for now. Let's create an issue for this improvement in the future.

By the way, now I realized that we duplicate data by storing block headers into the DB and full blocks (including block headers) in files. I think, we should store only full blocks and read block headers from that files (we can read only header's part from the file).

hackaugusto:
> @hackaugusto we store full blocks, not only block headers

I see. So the issues with packing is not as bad. Thanks for the explanation

bobbinth:
> By the way, now I realized that we duplicate data by storing block headers into the DB and full blocks (including block headers) in files. I think, we should store only full blocks and read block headers from that files (we can read only header's part from the file).

I'm not sure that's a big issue. Block headers are fairly compact (about 250 bytes) - so, even if we store 10M headers per year, that's just slightly more than 2.5GB of extra data. For comparison, at 100 TPS, note commitments (`node_id + metadata`) would probably require 250GB per year.

crates/store/src/state.rs
```diff
@@ -257,7 +257,7 @@ impl State {
             let mut inner = self.inner.write().await;
 
             // Save the block before transaction is committed:
-            self.block_store.save_block(block_num, &block_data)?;
+            self.block_store.save_block(block_num, &block_data).await?;
```
bobbinth, line 260: Not for this PR, but in the future we could move this call up and to right before we start updating in-memory structure, but await it still here. This way, writing to file could happen at the same time as memory structures are updated (and it is OK if we encounter an error during update, the incorrect file would be overwritten on the next valid update).

bobbinth (approved):
Looks good! Thank you! I left one comment inline for the future.

PR #348, opened 2024-05-08
Make proof verification error messages more detailed in the RPC

bobbinth:
This small PR makes is so that we return more detailed error messages from the PRC when proof verification fails.


igamigo (approved):
LGTM

Issue #346, opened 2024-05-06
[Task]: Modify gRPC Tonic Server to Allow HTTP/1.1 Requests


dagarcia7:
### Node's components affected by this task

- [X] RPC
- [ ] Block producer
- [ ] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

Implement functionality within the gRPC Tonic server to accept and process HTTP/1.1 requests from gRPC-Web clients. This will enable direct handling of requests without the need for the client to use an external proxy like Envoy.

* Integrate [tonic-web](https://crates.io/crates/tonic-web) with the server setup to enable this functionality. This integration should allow the Tonic server to translate HTTP/1.1 requests into the HTTP/2 format expected by gRPC without needing an external proxy. Here is an example of using `tonic-web` with the current RPC server configuration:

* Ensure that the server can handle both traditional gRPC (HTTP/2) and gRPC-Web (HTTP/1.1) requests.

### How should it be done?

* Integrate the [tonic-web](https://crates.io/crates/tonic-web) crate into the existing Tonic server configuration. This crate facilitates the server's ability to understand and respond to HTTP/1.1 requests as per the gRPC-Web specification. Here is an example of how to modify the existing gRPC server configuration per [tonic-web's documentation](https://docs.rs/tonic-web/0.11.0/tonic_web/).

```Rust
// RPC INITIALIZER
// ================================================================================================

pub async fn serve(config: RpcConfig) -> Result<(), ApiError> {
    info!(target: COMPONENT, %config, "Initializing server");

    let api = api::RpcApi::from_config(&config)
        .await
        .map_err(ApiError::ApiInitialisationFailed)?;
    let rpc = api_server::ApiServer::new(api);

    info!(target: COMPONENT, "Server initialized");

    let addr = config
        .endpoint
        .to_socket_addrs()
        .map_err(ApiError::EndpointToSocketFailed)?
        .next()
        .ok_or_else(|| ApiError::AddressResolutionFailed(config.endpoint.to_string()))?;

    Server::builder()
        .accept_http1(true)
        .add_service(tonic_web::enable(rpc))
        .serve(addr)
        .await
        .map_err(ApiError::ApiServeFailed)?;

    Ok(())
}
```

### When is this task done?

* The task is complete when the server can natively handle gRPC-Web requests sent over HTTP/1.1 alongside traditional gRPC requests over HTTP/2.

[...]

PR #352, opened 2024-05-10
rpc: enable tonic-web

hackaugusto:
Enables grpc-web #346

### DIFF
--- a/crates/rpc/src/server/mod.rs
+++ b/crates/rpc/src/server/mod.rs
@@ -30,7 +30,9 @@ pub async fn serve(config: RpcConfig) -> Result<(), ApiError> {
         .ok_or_else(|| ApiError::AddressResolutionFailed(config.endpoint.to_string()))?;
 
     Server::builder()
-        .add_service(rpc)
+        .accept_http1(true)
+        .add_service(rpc.clone())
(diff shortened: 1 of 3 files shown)


crates/rpc/src/server/mod.rs
```diff
@@ -30,7 +30,9 @@ pub async fn serve(config: RpcConfig) -> Result<(), ApiError> {
         .ok_or_else(|| ApiError::AddressResolutionFailed(config.endpoint.to_string()))?;
 
     Server::builder()
-        .add_service(rpc)
+        .accept_http1(true)
+        .add_service(rpc.clone())
```
hackaugusto, line 34: This should work because `rpc` has basically no state.

bobbinth: Does this allow us to serve both HTTP/1.1 and HTTP/2.0 requests? That is, what would happen if we did just:
```Rust
Server::builder()
    .accept_http1(true)
    .add_service(tonic_web::enable(rpc))
    .serve(addr)
    .await
    .map_err(ApiError::ApiServeFailed)?;
```

bobbinth (approved):

PR #353, opened 2024-05-10
config: unified config files

### LINKED ISSUE #318: Remove duplicates of sample config files for the node
After merging #312, we will have sample config files (`miden-node.toml` and `genesis.toml`) stored in 2 different places (i.e., in `node` and in `packaging` directories). This is not a big deal, but I think ideally we should have a "single source of truth" for these.  Assuming we want to do this, we have the following options:

1. Keep the only copies in the `node` directory. I think we'll need to update the packaging task to work with this.
2. Keep the only copies in the `packaging` directory. We may need to update our docker build scripts for this (and maybe something else?).
3. Put these files into some dedicated directory (e.g., `config`). We'll need to update both the Debian package task and docker build script.

I'm not sure what's the best approach here is - maybe option 3?

hackaugusto:
fixes #318


bin/node/Dockerfile
bobbinth, line 1: Question: why remove the comments form this file?

config/miden-node.toml
```diff
@@ -17,6 +17,6 @@ store_url = "http://localhost:28943"
 [store]
 # port defined as: sum(ord(c)**p for (p, c) in enumerate('miden-store', 1)) % 2**16
 endpoint = { host = "localhost", port = 28943 }
-database_filepath = "miden-store.sqlite3"
-genesis_filepath = "genesis.dat"
-blockstore_dir = "blocks"
+database_filepath = "/opt/miden/miden-store.sqlite3"
+genesis_filepath = "/opt/miden/genesis.dat"
+blockstore_dir = "/opt/miden/blocks"
```
bobbinth, line 22: This is now assuming that we are on Linux-based systems, right? (i.e., this won't work on Windows). This is fine as we are currently not targeting Windows anyways, but we should probably mention this more explicitly in the docs.

README.md
```diff
@@ -72,18 +72,13 @@ By default this will generate 1 file and 1 folder in the current directory:
 
 ### Running the node
 
-To run the node you will need to provide a configuration file. We have an example config file in [node/miden-node.toml](bin/node/miden-node.toml). Then, to run the node, run:
+Create a configuration file based on [node/miden-node.toml](bin/node/miden-node.toml), then create the necessary directories and start the node:
 
 ```sh
+mkdir -p /opt/miden
 miden-node start --config <path-to-config-file> <component-to-be-started>
 ```
```
bobbinth, line 80: I think we may need to update the docs more comprehensively here as the previous steps assume that everything is in the current directory (rather than in `/opt/miden`). In general, I think we have the following ways of installing the node:

1. By cloning the repo and compiling the source.
2. By running `cargo install`.
3. By building and running docker container (I wonder if these should be two separate steps and we'd provide docker image for download).
4. By downloading and running `.deb` package.

I think only the first 2 (or 3 if we include building of Docker image) require Rust to be installed. And for `cargo install`, I think we should mention explicitly which files need to be copied where.

bobbinth (changes requested):
Thank you! Looks good! I left a couple of comments inline - the main comment is about updating "Usage" section more comprehensively so that steps are still trivially easy to follow.

bin/node/Dockerfile
hackaugusto: They seemed unnecessary. For example:

```
# Copy source code
WORKDIR /app
COPY . .
```

The comment doesn't add much to the existing `COPY` command.

config/miden-node.toml
hackaugusto: What do you have in mind? The docs mention this is an _example_:

```
To run the node you will need to provide a configuration file. We have an example config file in [node/miden-node.toml](https://github.com/0xPolygonMiden/miden-node/blob/main/node/miden-node.toml). Then, to run the node, run:
```

Not sure what else is necessary there.

README.md
hackaugusto: created an issue: https://github.com/0xPolygonMiden/miden-node/issues/355

bobbinth (approved):

PR #356, opened 2024-05-13
fix: `AccountData` seed fix

### LINKED ISSUE #349: [Bug]: make-genesis Command Creates Account Keys with Different RNG than When Importing
### Version

v0.0

### Other packages versions

miden-client: v0.2.1

### What happened?

When attempting to run this new transaction command via the miden-client

```
miden-client tx new mint <regular-account-ID-A> <faucet-account-id> 1000 --note-type private
```

 with the regular account and faucet account generated via `make-genesis` command in this repo (with the default `genesis.toml` as input), I ran into this error

```
transaction executor error: ExecuteTransactionProgramFailed(AdviceMapKeyNotFound([2688260003581148592, 5885449917191890350, 9370402774682726881, 17460876200495716001]))
```

**Potential Solution:**
I believe the problem stems from the fact that the `make-genesis` command generates a `SecretKey` via the rng from the [ChaCha20Rng](https://github.com/0xPolygonMiden/miden-node/blob/c4747bb81a8f6037dd634c165958b67ecab9564e/bin/node/src/commands/genesis/mod.rs#L186) library.

After this, [create_basic_wallet](https://github.com/0xPolygonMiden/miden-node/blob/c4747bb81a8f6037dd634c165958b67ecab9564e/bin/node/src/commands/genesis/mod.rs#L129C47-L129C66) is called which [sets the generated public key as part of the account code](https://github.com/0xPolygonMiden/miden-base/blob/38340550a077a488d0466118c6868fe1144a38fc/miden-lib/src/accounts/wallets/mod.rs#L52).

Finally, when `import_account` is called in the miden-client, [the account object (containing the key generated previously) is directly passed through to store](https://github.com/0xPolygonMiden/miden-client/blob/4bf6ad3943d60a4c7735cb4caa35a702ea81f8a7/src/client/accounts.rs#L107C21-L107C42), but [a new `SecretKey` is generated](https://github.com/0xPolygonMiden/miden-client/blob/4bf6ad3943d60a4c7735cb4caa35a702ea81f8a7/src/client/accounts.rs#L88) this time using `RpoRandomCoin` as the rng which is stored separately. 

I believe the error downstream when creating a new transactions happens because of the discrepancy in the keys, which is a result of the different RNGs used on the provided seed. 

The easiest fix would be to generate the `SecretKey` in the same fashion in both places (`make-genesis` and `import_account`). For example, in `make-genesis`, you could do this

```
fn parse_auth_inputs(
    auth_scheme_input: AuthSchemeInput,
    auth_seed: &str,
) -> Result<(AuthScheme, AuthData)> {
    match auth_scheme_input {
        AuthSchemeInput::RpoFalcon512 => {
            let auth_seed: [u8; 32] = hex_to_bytes(auth_seed)?;
            let auth_seed_digest = Digest::try_from(&auth_seed)?.into();
            let mut rng = RpoRandomCoin::new(auth_seed_digest);
            let secret = SecretKey::with_rng(&mut rng);

            let auth_scheme = AuthScheme::RpoFalcon512 { pub_key: secret.public_key() };
            let auth_info = AuthData::RpoFalcon512Seed(auth_seed);

            Ok((auth_scheme, auth_info))
        },
    }
}
```

and ensure that `RpoRandomCoin` and `Digest` are correctly imported. However, I wanted to file this as a bug instead of suggest a PR as I'm sure there are different refactors possible to allow the same RNG to be used in both places instead of only using `RpoRandomCoin`. 

After generating both the regular and faucet account through the make-genesis command, I should have been able to import them via the miden-client and followed rest of the instructions here https://crates.io/crates/miden-client to create a new transaction to mint a fungible asset for the regular account.

* Clone the miden-repo and ensure you're on the main branch
* `cd` into the repository and run `cargo install --features testing --path bin/node`
* `cd` into `bin/node` and run `miden-node make-genesis` to generate the two accounts specified in the default `genesis.toml` file 
* run `miden-node start node` to start the node
* Install the miden-client CLI via `cargo install miden-client --features testing,concurrent`
* run `miden-client init` and select all the default options
* Import the generated accounts via `miden-client account import <path-to-accounts-directory>/*.mac`
* run `miden-client sync`
* run `miden-client tx new mint <regular-account-ID-A> <faucet-account-id> 1000 --note-type private` and this should cause the error described. 

### Relevant log output

```shell
transaction executor error: ExecuteTransactionProgramFailed(AdviceMapKeyNotFound([2688260003581148592, 5885449917191890350, 9370402774682726881, 17460876200495716001]))
```

igamigo:
Closes #349.
This fix defines `RpoRandomCoin` as the canonical RNG for Falcon keys for `AccountData` descriptions. A small doc update was issued for the enum here: https://github.com/0xPolygonMiden/miden-base/pull/689.

This was tested with the client and node on `next` and issuing transactions worked fine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,7 +103,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.61",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -216,7 +216,7 @@ dependencies = [
  "actix-router",
  "proc-macro2",
  "quote",
- "syn 2.0.61",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -387,7 +387,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.61",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -398,7 +398,7 @@ checksum = "c6fa2087f2753a7da8cc1c0dbfcf89579dd57458e36769de5ac750b4671737ca"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.61",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -514,7 +514,7 @@ dependencies = [
  "regex",
  "rustc-hash",
  "shlex",
- "syn 2.0.61",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -696,7 +696,7 @@ dependencies = [
  "heck 0.5.0",
  "proc-macro2",
  "quote",
- "syn 2.0.61",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -1483,7 +1483,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "regex-syntax 0.6.29",
- "syn 2.0.61",
+ "syn 2.0.63",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


igamigo:
Superseded by #357

PR #357, opened 2024-05-14
chore: Use `miden-base` 0.3

igamigo:
Also updates Rust version based on the new `miden-base` requirements, and updates the account generation on genesis.
This does not update the node's crates versions but I can add that change if needed.


bobbinth (approved):
Looks good! Thank you!

PR #359, opened 2024-05-14
fix: remove duplicate rpc server

mFragaBA:
on #352 we added support for webrpc, but it causes the node to not start as it has the route definitions duplicated. I tested removing the line and running the miden client's integration test to make sure we're still supporting http2 requests.

### DIFF
--- a/crates/rpc/src/server/mod.rs
+++ b/crates/rpc/src/server/mod.rs
@@ -31,7 +31,6 @@ pub async fn serve(config: RpcConfig) -> Result<(), ApiError> {
 
     Server::builder()
         .accept_http1(true)
-        .add_service(rpc.clone())
         .add_service(tonic_web::enable(rpc))
         .serve(addr)
         .await


igamigo:
Here's the error that would occur upon running the node:

```shell
2024-05-14T17:49:38.026870Z  INFO miden-rpc: crates/rpc/src/server/mod.rs:23: Server initialized
thread 'tokio-runtime-worker' panicked at /Users/ignacioamigo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/axum-0.6.20/src/routing/mod.rs:145:9:
Invalid route "/rpc.Api/*rest": insertion failed due to conflict with previously registered route: /rpc.Api/*rest
```

As Martin mentioned, the client's integration tests work well without the removed line, so it should be safe to remove.

igamigo (approved):

bobbinth (approved):
Looks good! Thank you for fixing this!

PR #360, opened 2024-05-16
feat: Use newer revisions for faucet

igamigo:
Migrates faucet's revisions to the newer ones.
As a consequence of changes in how the store is handled between the client and `miden-base`'s `TransactionExecutor`, it is not `Send` and so a `Client` is instantiated on each request for now (which might be a net positive for how the faucet is supposed to work at least?).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1570,19 +1570,19 @@ dependencies = [
 
 [[package]]
 name = "miden-client"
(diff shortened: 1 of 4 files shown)


bobbinth (approved):
Looks good! Thank you!

Let's create an issue in `miden-client` to change `Rc` to `Arc` there.

PR #361, opened 2024-05-16
Update faucet instructions with newer CLI commands

igamigo:


bobbinth (approved):
Looks good! Thank you!

PR #362, opened 2024-05-17
fix: various faucet fixes for next

tomyrd:
This PR fixes a few things related to the faucet:
- Brings some faucet fixes from `main` to `next`.
- Fixes the docker local node.
- Fixes a problem with the config file path that made running the node somewhat awkward.
- Adds a loading spinner to the UI (change requested in [this comment](https://github.com/0xPolygonMiden/miden-node/pull/339#issuecomment-2116843172)).
- It adds some instructions on how to run the faucet.

Here is a video showing the new spinner:

https://github.com/0xPolygonMiden/miden-node/assets/43424983/405b1d45-3a91-4cc7-b4ec-f08b29ed9134


bin/faucet/README.md
```diff
@@ -4,3 +4,17 @@ This crate contains a binary for running a Miden rollup faucet.
...
+cargo make docker-build-node
+cargo make docker-run-node
+```
+
+2. From the "miden-node" repo root run the faucet:
+```bash
+cargo run --bin miden-faucet  --features testing --release
+```
+
+After a few secons you may go to `http://localhost:8080` and see the faucet UI.
```
igamigo, line 20: nit:
```suggestion
After a few seconds you may go to `http://localhost:8080` and see the faucet UI.
```

igamigo (approved):
LGTM! I think it would be good if we could add two changes:

- If the spinner is active, make sure the buttons are deactivated/don't do anything when clicked in order to avoid people spamming requests.
- Judging by the video it looks like the spinner moves the page layout a bit which is probably not ideal.

bin/faucet/README.md
```diff
@@ -4,3 +4,17 @@ This crate contains a binary for running a Miden rollup faucet.
 
 ## License
 This project is [MIT licensed](../../LICENSE).
+
+## Running the faucet
```
mFragaBA, line 8: I'd put this above the License section

mFragaBA (commented):
I noticed the docker is using rust 1.76, could that be?

image

Should we bump it up to 1.78?

bin/faucet/src/handlers.rs
```diff
@@ -6,12 +6,18 @@ use miden_objects::{
...
 use serde::{Deserialize, Serialize};
 use tracing::info;
 
-use crate::{errors::FaucetError, utils::FaucetState};
+use crate::{
+    errors::FaucetError,
+    utils::{build_client, FaucetState},
+};
+
+const TOKEN_AMOUNT_OPTIONS: [u64; 3] = [100, 500, 1000];
```
mFragaBA, line 20: wasn't this already part of the config? I can still see it in `miden-faucet.toml`

bin/faucet/src/handlers.rs
```diff
@@ -23,14 +29,14 @@ struct FaucetRequest {
 #[derive(Serialize)]
 struct FaucetMetadataReponse {
     id: String,
-    asset_amount_options: Vec<u64>,
+    asset_amount_options: [u64; 3],
```
mFragaBA, line 32: why do we need the change to a fixed size array?

mFragaBA (commented):
I left some small comments. I'm waiting for the docker image to build so I can test it 😅

bin/faucet/src/handlers.rs
igamigo: This was one of the changes I left on my PR to `main`, but I think @tomyrd is going to roll it back into using the configuration field.

bin/faucet/README.md
tomyrd: Done.

bin/faucet/src/handlers.rs
tomyrd: Done.

phklive (approved):
Tested and LGTM.

Only thing I could say is that I believe that you should run `cargo update` and commit before merging because we are now also tracking the cargo.lock file.

bobbinth (approved):
Looks good! Thank you!

PR #365, opened 2024-05-21
feat: add config to make onchain genesis accounts

### LINKED ISSUE #305: [Feature]: Support for public accounts in the genesis file

End user

It might be useful to create public (on-chain) accounts in genesis. At the moment we don't support public accounts in genesis file. We should discuss, if we need it.

### How is this feature used?

TBD

### Anything else?

_No response_

mFragaBA:
closes #305.

changes:

- added the possibility to specify whether the genesis accounts are on_chain or not. If not specified defaults to false ([11994b6](https://github.com/0xPolygonMiden/miden-node/pull/365/commits/11994b60dffd25118db344b2e00be0afe702514e))
- changed the `into_block` function of `GenesisState` so it choses the appropiate `AccountUpdateDetails` ([7bcfe25](https://github.com/0xPolygonMiden/miden-node/pull/365/commits/7bcfe25806dce9c621b67b35e612cb97422fd652))

## Steps To Test

1. initialize two miden clients
2. create a new genesis uncommenting the `on_chain = true` lines in the config for both accounts
3. start the node
4. import the accounts on each client
5. run a mint transaction on one of the clients, wait for a bit so the tx gets included in a block and sync on the other client. You should receive a nonce update on the faucet account

### DIFF
--- a/bin/node/src/commands/genesis/inputs.rs
+++ b/bin/node/src/commands/genesis/inputs.rs
@@ -25,6 +25,8 @@ pub struct BasicWalletInputs {
     pub init_seed: String,
     pub auth_scheme: AuthSchemeInput,
     pub auth_seed: String,
+    #[serde(default = "_default_true")]
+    pub on_chain: bool,
 }
 
 #[derive(Debug, Clone, Deserialize)]
@@ -35,9 +37,15 @@ pub struct BasicFungibleFaucetInputs {
     pub token_symbol: String,
     pub decimals: u8,
     pub max_supply: u64,
+    #[serde(default = "_default_true")]
+    pub on_chain: bool,
 }
 
 #[derive(Debug, Clone, Copy, Deserialize)]
 pub enum AuthSchemeInput {
     RpoFalcon512,
+}
+
+const fn _default_true() -> bool {
+    true
 }
--- a/bin/node/src/commands/genesis/inputs.rs
+++ b/bin/node/src/commands/genesis/inputs.rs
@@ -25,8 +25,7 @@ pub struct BasicWalletInputs {
     pub init_seed: String,
     pub auth_scheme: AuthSchemeInput,
     pub auth_seed: String,
-    #[serde(default = "_default_true")]
-    pub on_chain: bool,
+    pub storage_mode: String,
 }
 
 #[derive(Debug, Clone, Deserialize)]
@@ -37,15 +36,10 @@ pub struct BasicFungibleFaucetInputs {
     pub token_symbol: String,
     pub decimals: u8,
     pub max_supply: u64,
-    #[serde(default = "_default_true")]
-    pub on_chain: bool,
+    pub storage_mode: String,
 }
 
 #[derive(Debug, Clone, Copy, Deserialize)]
 pub enum AuthSchemeInput {
     RpoFalcon512,
-}
-
-const fn _default_true() -> bool {
-    true
 }
--- a/bin/node/src/commands/genesis/mod.rs
+++ b/bin/node/src/commands/genesis/mod.rs
@@ -126,11 +126,17 @@ fn create_accounts(
                 let (auth_scheme, auth_secret_key) =
                     parse_auth_inputs(inputs.auth_scheme, &inputs.auth_seed)?;
 
+                let storage_type = if inputs.on_chain {
+                    AccountStorageType::OnChain
+                } else {
+                    AccountStorageType::OffChain
+                };
+
                 let (account, account_seed) = create_basic_wallet(
                     init_seed,
                     auth_scheme,
                     AccountType::RegularAccountImmutableCode,
-                    AccountStorageType::OffChain,
+                    storage_type,
                 )?;
 
                 AccountData::new(account, Some(account_seed), auth_secret_key)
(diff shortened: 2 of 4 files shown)


mFragaBA:
> @mFragaBA - this is ready for review, right?

yes

bin/node/src/commands/genesis/inputs.rs
```diff
@@ -25,6 +25,8 @@ pub struct BasicWalletInputs {
     pub init_seed: String,
     pub auth_scheme: AuthSchemeInput,
     pub auth_seed: String,
+    #[serde(default = "_default_true")]
+    pub on_chain: bool,
```
bobbinth, line 29: Instead of a boolean flag, should we use `AccountStorageType` here?

Also, what is the motivation setting the default value to true?

bin/node/src/commands/genesis/inputs.rs
```diff
@@ -35,9 +37,15 @@ pub struct BasicFungibleFaucetInputs {
     pub token_symbol: String,
     pub decimals: u8,
     pub max_supply: u64,
+    #[serde(default = "_default_true")]
+    pub on_chain: bool,
```
bobbinth, line 41: Similar comment to the one above.

config/genesis.toml
```diff
@@ -16,3 +18,5 @@ auth_seed = "0xd123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
 token_symbol = "POL"
 decimals = 12
 max_supply = 1000000
+# To make this account on_chain uncomment the following
+# on_chain = true
```
bobbinth, line 22: Similar to the previous comments: I would probably call this `storage_mode` (we should probably rename `AccountStorageType` into `AccountStorageMode` - but that's for another PR) and not comment it out. Also, I'd move it right after `type` on line 14.

bobbinth (commented):

bin/node/src/commands/genesis/inputs.rs
mFragaBA: > Instead of a boolean flag, should we use AccountStorageType here?

That can work as well, just seemed simpler to use a bool

> Also, what is the motivation setting the default value to true?

Oh, now that you mention it I wanted to set it as false by default (that is so current configs wouldn't break and would behave as always)

bobbinth: I'd prefer to set all the required properties explicitly (and I think it's OK to break things for now).

mFragaBA: should be changed now

bin/node/src/commands/genesis/inputs.rs
mFragaBA: changed

config/genesis.toml
mFragaBA: moved this up, uncommented and did the change suggested. Now it's `storage_mode` and it can have the values `on-chain` and `off-chain`. I wasn't able to use `AccountStorageType` straight up because it said it didn't implement Deserialize (though it might be fixable and I didn't find a way), so I ended up writing a `parse_storage_mode`

mFragaBA: I also updated the test because it was failing and added a check on it asserting that the accounts have the proper storage mode

bin/node/src/commands/genesis/mod.rs
```diff
@@ -195,6 +199,14 @@ fn parse_auth_inputs(
     }
 }
 
+fn parse_storage_mode(storage_type: &str) -> Result<AccountStorageType> {
+    match storage_type.to_lowercase().as_str() {
+        "on-chain" => Ok(AccountStorageType::OnChain),
+        "off-chain" => Ok(AccountStorageType::OffChain),
+        lowercase_type => Err(anyhow!("The provided value for storage_type ({}) does not match the expected values (on-chain, off-chain)", lowercase_type))
```
bobbinth, line 206: minor nit: could we inline the variable into the string? (e.g., "The provided value for storage_type ({lowercase_type}) does not match the expected values (on-chain, off-chain)".

Also, should we rename this to just `mode`?

config/genesis.toml
```diff
@@ -4,12 +4,14 @@ timestamp = 1672531200
...
 [[accounts]]
 type = "BasicWallet"
+storage_mode = "off-chain"
 init_seed = "0xa123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
 auth_scheme = "RpoFalcon512"
 auth_seed = "0xb123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
 
 [[accounts]]
 type = "BasicFungibleFaucet"
+storage_mode = "off-chain"
```
bobbinth, line 14: Should we default the faucet to `on-chain`?

bobbinth (approved):
Looks good! Thank you! I left a couple of small comments inline. After these are addressed, we can merge.

bin/node/src/commands/genesis/mod.rs
mFragaBA: done!

config/genesis.toml
mFragaBA: I don't see why not. Done!

Issue #363, opened 2024-05-21
[Feature]: Standardise Makefile and CI across Miden repositories


phklive:
### To whom is this feature for?

Rollup operator

### Why is is this feature needed?

For now we are using different configurations for the Makefile and CI across the Miden repositories.

We want to standardise this:
- use Make instead of Cargo make
- use Make commands in the CI
- use the same configuration for commands in all repositories

### How is this feature used?

This will improve management of repositories

Dominik1999:
Closed by https://github.com/0xPolygonMiden/miden-node/pull/367

PR #367, opened 2024-05-27
Standardise CI and Makefile across repositories

phklive:
In this PR I propose a standardisation of the CI and Makefile across Miden repositories. 

With the following changes: 
- Cargo make to Make
- CI using Cargo make to Make

Once this first one is approved I will port it to the following repos:
- VM
- Crypto
- Client
- Compiler

Closes: #363


hackaugusto (approved):
LGTM, thanks :)

Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+# --- installing --------------------------------------------------------------------------------
+
+.PHONY: install-node
+install-node: ## Installs node with testing feature enabled
+	cargo install --features testing --path bin/node
+
+.PHONY: install-faucet
+install-faucet: ## Installs faucet with testing feature enabled
+	cargo install --features testing --path bin/faucet
```
bobbinth, line 61: Should we have 2 versions of these commands? One for installing with `testing` feature and one without?

bobbinth (approved):
Looks good! Thank you! I left one comment inline. Once this is addressed, we can merge.

Makefile
phklive: Sure.

PR #368, opened 2024-05-27
Remove client dependency from faucet

### LINKED ISSUE #343: Remove client dependency from the faucet
Currently, the faucet uses `miden-client` to create and submit transactions to the node. When the node and client versions diverge, this becomes problematic. So, instead, we should refactor the faucet to use structures from `miden-base` and directly submit transactions to the node (via RPC).

tomyrd:
closes #343 

This PR removes almost all of the functionality implemented by the `miden-client` and replaces it with it's own implementation that doesn't depend on the crate. The new `FaucetClient` is stored in the state and wrapped with a mutex to allow concurrency (a feature that was removed in the previous version).

Some caveats:
- We still have a small dependency. The `InputNoteRecord` type is needed in order to be serialized and imported by the user. We could change the `miden-client` to accept the `Note` type as it's import or move the `InputNoteRecord` to `miden-base`.
- There's a small issue where if multiple transactions are sent at the same time, the node accepts the first one and rejects all others. This happens because of [this verification](https://github.com/0xPolygonMiden/miden-node/blob/4e0b36753892e9c554e0643951a589e60a22d012/crates/block-producer/src/state_view/mod.rs#L157-L160).


bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+        let script = self.executor.compile_tx_script(script, vec![], vec![]).map_err(|err| {
+            FaucetError::InternalServerError(format!("Failed to compile script: {}", err))
+        })?;
+
+        let mut transaction_args = TransactionArgs::new(Some(script), None, AdviceMap::new());
+        transaction_args.extend_expected_output_notes(vec![output_note.clone()]);
+
+        self.executor
+            .load_account(self.id)
+            .map_err(|err| FaucetError::InternalServerError(err.to_string()))?;
```
igamigo, line 119: The account can probably be loaded once on initialization instead of on every request

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+    rng: RpoRandomCoin,
+}
+
+unsafe impl Send for FaucetClient {}
+unsafe impl Sync for FaucetClient {}
+
+impl FaucetClient {
+    pub async fn new(config: FaucetConfig) -> Result<Self, FaucetError> {
+        let (rpc_api, root_block_header, root_chain_mmr) =
+            initial_connection(config.clone()).await?;
```
igamigo, line 51: nit: The name for this function is a bit misleading, maybe something like `initialize_faucet_client`? Since it requests the genesis block, initliazes the tonic client, etc.

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    let request = GetBlockHeaderByNumberRequest {
+        block_num: Some(0),
+        include_mmr_proof: Some(true),
+    };
+    let response = rpc_api.get_block_header_by_number(request).await.unwrap();
+    let root_block_header: BlockHeader =
+        response.into_inner().block_header.unwrap().try_into().unwrap();
+    let root_chain_mmr =
+        ChainMmr::new(PartialMmr::from_peaks(MmrPeaks::new(0, Vec::new()).unwrap()), Vec::new())
```
igamigo, line 300: Can we turn these `unwrap()`s into `expect()`s, stating why they are safe to perform?

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+        .await
+        .map_err(|err| FaucetError::InternalServerError(err.to_string()))?;
+
+    let request = GetBlockHeaderByNumberRequest {
+        block_num: Some(0),
+        include_mmr_proof: Some(true),
+    };
+    let response = rpc_api.get_block_header_by_number(request).await.unwrap();
+    let root_block_header: BlockHeader =
+        response.into_inner().block_header.unwrap().try_into().unwrap();
```
igamigo, line 298: Perhaps instead of panicking here (and on line 296), the function should return a `FaucetError::InitializationError` or something similar, indicative of what the problem might be

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+    let root_chain_mmr =
+        ChainMmr::new(PartialMmr::from_peaks(MmrPeaks::new(0, Vec::new()).unwrap()), Vec::new())
+            .unwrap();
+
+    Ok((rpc_api, root_block_header, root_chain_mmr))
+}
+// TODO: Remove the falcon signature function once it's available on base and made public
```
igamigo, line 305: This `TODO` seems duplicate. Also, if it's not complicated we should publish this function since this has various occurrences already (I think I made the function `pub` but forgot to re-export it)

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+        let request = SubmitProvenTransactionRequest {
+            transaction: proven_transaction.to_bytes(),
+        };
+
+        self.rpc_api
+            .submit_proven_transaction(request)
+            .await
+            .map_err(|err| FaucetError::InternalServerError(err.to_string()))?;
+
+        self.data_store.update_faucet_account(&delta).unwrap();
```
igamigo, line 149: This should probably return an error instead of panicking

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+    let (faucet_account, account_seed) = create_basic_fungible_faucet(
+        seed,
+        token_symbol,
+        config.decimals,
+        Felt::try_from(config.max_supply)
+            .map_err(|err| FaucetError::InternalServerError(err.to_string()))?,
+        AccountStorageType::OffChain,
+        auth_scheme,
+    )
+    .unwrap();
```
igamigo, line 276: This is probably safe to do but let's change this to an `expect()` explaining why it is so.

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// HELPER FUNCTIONS
+// ================================================================================================
+
+fn build_account(config: FaucetConfig) -> Result<(Account, Word, SecretKey), FaucetError> {
+    let token_symbol = TokenSymbol::new(config.token_symbol.as_str())
+        .map_err(|err| FaucetError::AccountCreationError(err.to_string()))?;
+
+    let seed: [u8; 32] = [0; 32];
```
igamigo, line 260: Should this be configurable somehow? Maybe not for this PR since it was like this before anyway

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+
+        let mut transaction_args = TransactionArgs::new(Some(script), None, AdviceMap::new());
+        transaction_args.extend_expected_output_notes(vec![output_note.clone()]);
+
+        self.executor
+            .load_account(self.id)
+            .map_err(|err| FaucetError::InternalServerError(err.to_string()))?;
+
+        let executed_tx =
+            self.executor.execute_transaction(self.id, 0, &[], transaction_args).unwrap();
```
igamigo, line 122: This should likely also return a `FaucetError`

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,353 @@
...
+                .replace("{amount}", &Felt::new(asset.amount()).to_string()),
+        )
+        .expect("shipped MASM is well-formed");
+
+        let script = self.executor.compile_tx_script(script, vec![], vec![]).map_err(|err| {
+            FaucetError::InternalServerError(format!("Failed to compile script: {}", err))
+        })?;
+
+        let mut transaction_args = TransactionArgs::new(Some(script), None, AdviceMap::new());
+        transaction_args.extend_expected_output_notes(vec![output_note.clone()]);
```
igamigo, line 115: Perhaps this can be factored out to its own function to make the mint function more readable

igamigo (commented):
Looking good! Left some comments

bin/faucet/src/client.rs
tomyrd: Oops, I forgot some unwraps from the testing phase

tomyrd: Done.

bin/faucet/src/client.rs
tomyrd: Done.

bin/faucet/src/client.rs
tomyrd: Done.

bin/faucet/src/client.rs
tomyrd: Done.

bin/faucet/Cargo.toml
```diff
@@ -23,12 +22,15 @@ actix-web = "4"
 async-mutex = "1.4.0"
 derive_more = "0.99.17"
 figment = { version = "0.10", features = ["toml", "env"] }
-miden-client = { version = "0.3", features = ["concurrent"] }
+miden-client = { version = "0.3" }
 miden-lib = { version = "0.3" }
 miden-node-proto = { workspace = true }
 miden-node-utils = { workspace = true }
 miden-objects = { version = "0.3" }
+miden-tx = { version = "0.3" }
```
bobbinth, line 30: Should we not import this with `concurrent` feature enabled? (to make transaction proving faster)

bin/faucet/Cargo.toml
```diff
@@ -14,7 +14,6 @@ repository.workspace = true
 [features]
 # Makes `make-genesis` subcommand run faster. Is only suitable for testing.
 # INFO: Make sure that all your components have matching features for them to function.
-testing = ["miden-client/testing"]
```
bobbinth, line 17: I think we might still need `testing` feature to make sure faucet uses the same PoW settings as node/client. This feature would probably need to be applied to `miden-objects` and `miden-tx`.

bobbinth:
> * We still have a small dependency. The `InputNoteRecord` type is needed in order to be serialized and imported by the user. We could change the `miden-client` to accept the `Note` type as it's import or move the `InputNoteRecord` to `miden-base`.

I didn't see where in the code we are currently using `InputNoteRecord`, but if we need it, my preference would be to create a struct in `miden-objects` (e.g., maybe something like `NoteFile` which is used specifically for serializing different note variants.

bin/faucet/src/client.rs
tomyrd: Done.

bin/faucet/src/client.rs
tomyrd: [PR](https://github.com/0xPolygonMiden/miden-base/pull/714) ready. I can remove the TODO+function once it's merged.

bin/faucet/src/client.rs
tomyrd: Done. Changed ir to an `AccountCreationError`.

bin/faucet/src/client.rs
tomyrd: Done.

bin/faucet/src/client.rs
tomyrd: Like storing the account seed in the config file?

bin/faucet/Cargo.toml
tomyrd: Done. Omitted `miden-tx` as it doesn't have a `testing` feature.

bin/faucet/Cargo.toml
tomyrd: Done.

tomyrd:
> I didn't see where in the code we are currently using `InputNoteRecord`

[Here](https://github.com/0xPolygonMiden/miden-node/blob/bace872defb0a325c92600f5feaf220b9a6fe65a/bin/faucet/src/handlers.rs#L66). It's needed because the client can only import serialized `InputNoteRecord` files.

> maybe something like `NoteFile` which is used specifically for serializing different note variants.

Created an [issue](https://github.com/0xPolygonMiden/miden-base/issues/715) for this in `miden-base`.

bin/faucet/src/client.rs
igamigo: Yeah, although it's likely unnecessary for now. Let's defer until it's needed.

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    pub fn get_faucet_id(&self) -> AccountId {
+        self.id
+    }
+}
+
+#[derive(Clone)]
+pub struct FaucetDataStore {
+    faucet_account: Rc<RefCell<Account>>,
```
mFragaBA, line 145: If we changed `Rc<RefCell<Account>>` for `Arc<<Mutex<Account>>` (maybe even `Arc<RwLock<Account>>`) would we still need the `unsafe impl Send` and `unsafe impl Sync` from above?

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,382 @@
...
+        }
+
+        let module_ast = account.code().module().clone();
+        Ok(module_ast)
+    }
+}
+
+struct FaucetAuthenticator {
+    secret_key: AuthSecretKey,
+    rng: RefCell<ChaCha20Rng>,
```
mFragaBA, line 212: as for what was said above, what happens if we wrap this with a `Mutex`/ `RwLock` instead of a `RefCell`?

bin/faucet/src/handlers.rs
```diff
@@ -51,64 +40,30 @@ pub async fn get_tokens(
...
-        .map_err(|err| FaucetError::InternalServerError(err.to_string()))?;
-
-    // Run transaction executor & execute transaction
-    let tx_result = client
-        .new_transaction(tx_request)
-        .map_err(|err| FaucetError::InternalServerError(err.to_string()))?;
-
-    // Get created notes from transaction result
-    let created_notes = tx_result.created_notes().clone();
+    // Execute transaction
```
mFragaBA, line 53: can we add some `info!` calls in between these steps (executing mint and proving + submitting)?

mFragaBA (commented):
Tested locally with the dockerized node, worked like a charm. In fact, it feels faster than before (it's still restricted by block time since we can only do one tx per block.

Again unrelated to the PR, but could it be possible to do a mint for multiple accounts at a time by "aggregating" many mint requests into a single Tx that generates all mint notes? It would probably change the flow a bit since we'd need to switch to a polling scheme or similar, but it would allow the faucet to mint for several accounts per block.

bin/faucet/src/client.rs
tomyrd: Yes because the `TransactionExecutor` still wouldn't be `Send+Sync`

bin/faucet/src/client.rs
tomyrd: [It was suggested](https://github.com/0xPolygonMiden/miden-base/pull/714#pullrequestreview-2087440701) to remove this struct in favor of the `BasicAuthenticator` which currently also uses a `RefCell`. [In the future](https://github.com/0xPolygonMiden/miden-client/issues/358#issuecomment-2135305518) this may change and we could remove the `unsafe impl Send+Sync`.

bin/faucet/src/handlers.rs
tomyrd: Done.

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,382 @@
...
+/// - The secret key is malformed due to either incorrect length or failed decoding.
+/// - The signature generation failed.
+///
+/// TODO: once this gets made public in miden base, remve this implementation and use the one from
+/// base
+fn get_falcon_signature(
+    key: &rpo_falcon512::SecretKey,
+    message: Word,
+    rng: &mut ChaCha20Rng,
+) -> Result<Vec<Felt>, AuthenticationError> {
```
bobbinth, line 358: I believe it was mentioned in one of the comments in `minde-base` that we can get rid of this function and use `BasicTransactionAuthenticator` instead.

bin/faucet/src/client.rs
bobbinth, line 1: minor nit: could we add our usual section separators to this file?

bobbinth (commented):

Also, I think as a part of this PR we should get rid of the `miden-client` dependency entirely. So, before merging this, we should probably address https://github.com/0xPolygonMiden/miden-base/issues/715 and propagate the changes here and to `miden-client`.

bin/faucet/src/client.rs
tomyrd: Done.

bin/faucet/src/client.rs
tomyrd: Done. Also added some missing doc comments.

tomyrd:
> So, before merging this, we should probably address https://github.com/0xPolygonMiden/miden-base/issues/715 and propagate the changes here and to miden-client.

Creted a [PR](https://github.com/0xPolygonMiden/miden-base/pull/721) adding the new `NoteFile` object. Drafting this PR until that one is merged. We will also need to update the client to use the new serialized file but it's not a blocking change for the faucet.

bin/faucet/Cargo.toml
```diff
@@ -23,12 +23,14 @@ actix-web = "4"
 async-mutex = "1.4.0"
 derive_more = "0.99.17"
 figment = { version = "0.10", features = ["toml", "env"] }
-miden-client = { version = "0.3", features = ["concurrent"] }
-miden-lib = { version = "0.3" }
+miden-lib = { package = "miden-lib", git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "next" }
```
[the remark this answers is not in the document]
tomyrd: Done.

bin/faucet/Cargo.toml
```diff
@@ -23,12 +23,14 @@ actix-web = "4"
...
 miden-node-proto = { workspace = true }
 miden-node-utils = { workspace = true }
-miden-objects = { version = "0.3" }
+miden-objects = { package = "miden-objects", git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "next" }
+miden-tx = { package = "miden-tx", git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "next",  features = ["concurrent"] }
```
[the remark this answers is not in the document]
tomyrd: Done.

bin/faucet/Cargo.toml
```diff
@@ -23,12 +23,14 @@ actix-web = "4"
...
+tonic = { version = "0.11" }
```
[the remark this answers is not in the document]
tomyrd: Done.

bobbinth:
@tomyrd - thank you!

@igamigo, @mFragaBA - could one of you take another look and then we can merge?

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,331 @@
...
+            TransactionExecutor::new(data_store.clone(), Some(Rc::new(authenticator)));
+
+        executor
+            .load_account(id)
+            .map_err(|err| FaucetError::InternalServerError(err.to_string()))?;
+
+        Ok(Self { data_store, rpc_api, executor, id })
+    }
+
+    /// Gets [RpoRandomCoin] from the client
```
igamigo, line 81: nit: This comment is a bit misleading

bin/faucet/src/client.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn get_random_coin(&self) -> RpoRandomCoin {
+        let mut rng = thread_rng();
+        let coin_seed: [u64; 4] = rng.gen();
+
+        RpoRandomCoin::new(coin_seed.map(Felt::new))
+    }
```
igamigo, line 87: Correct me if I'm wrong but I think you added this function because when passing a long-lived RNG object, the object would not be modified and the serial number would be the same for all created notes, right?
@bobbinth is there any reason why [`create_p2id_note()`](https://github.com/0xPolygonMiden/miden-base/blob/727916915ddb9a237ffb975a1f7cc2f9f69139dc/miden-lib/src/notes/mod.rs#L31) does not take `&mut R` instead of owning the object? In that case the faucet could have one `RpoRandomCoin` and pass that every time, no?

igamigo (approved):
LGTM! Left a couple of comments but nothing blocking.

bin/faucet/src/client.rs
tomyrd: Yes, I had to change the rng to replicate what the client does to generate random numbers. Befor this fix the serial number was always the same so the note ID didn't change for the same account+asset amout.

bobbinth: > @bobbinth is there any reason why [`create_p2id_note()`](https://github.com/0xPolygonMiden/miden-base/blob/727916915ddb9a237ffb975a1f7cc2f9f69139dc/miden-lib/src/notes/mod.rs#L31) does not take `&mut R` instead of owning the object? In that case the faucet could have one `RpoRandomCoin` and pass that every time, no?

hmmm - good catch! There is no good reason why it is `mut R` rather than `&mut R`. @tomyrd - could you make a PR in `miden-base` to fix this and then update this PR accordingly?

tomyrd: On it

bobbinth: @tomyrd - we should be able to update this now - right?

bobbinth: oh - n/m - I see that it is already done. thank you!

bobbinth:
I see that this was merged into `mFragaBA-next-0.4`. Should we merge it into `next`?

mFragaBA:
> I see that this was merged into `mFragaBA-next-0.4`. Should we merge it into `next`?

it will change the miden-base dependencies to point to the next branch since it was merged into `mFragaBA-next-0.4`, if we're ok with that then sure (we'll need to change it before release)

bin/faucet/Cargo.toml
```diff
@@ -23,12 +23,14 @@ actix-web = "4"
...
 miden-node-utils = { workspace = true }
-miden-objects = { version = "0.3" }
+miden-objects = { workspace = true , features = ["concurrent"] }
+miden-tx = { workspace = true,  features = ["concurrent"] }
+rand = { version = "0.8.5" }
 rand_chacha = "0.3"
 serde = { version = "1.0", features = ["derive"] }
-tracing = { workspace = true }
 thiserror = { workspace = true }
+tonic = { version = "0.11" }
```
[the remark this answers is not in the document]
bobbinth: Yes - let's do it.

bobbinth:
> it will change the miden-base dependencies to point to the next branch since it was merged into `mFragaBA-next-0.4`, if we're ok with that then sure (we'll need to change it before release)

Yes - I think it is fine to have `next` branch here point to the `next` branch in `miden-base`.

crates/store/src/state.rs
```diff
@@ -226,7 +226,7 @@ impl State {
                 .map(|(note_index, note)| {
                     let details = match note {
                         OutputNote::Full(note) => Some(note.to_bytes()),
-                        OutputNote::Header(_) => None,
+                        OutputNote::Header(_) | OutputNote::Partial(_) => None,
```
bobbinth, line 229: Getting a partial note here should be an error. But this is probably not for this PR. Let's create an issue to come back to this in the future.

bobbinth (approved):

mFragaBA (approved):
LGTM! (tested locally with a dockerized node + faucet and an account from a local client. Created multiple public notes and was able to consume them). We'd still need https://github.com/0xPolygonMiden/miden-client/pull/375 for private ones to import private notes into the client

crates/store/src/state.rs
tomyrd: Created a PR but then closed it as it is being fixed in #377

PR #370, opened 2024-05-30
Synchronize block applying

polydez:
This resolves https://github.com/0xPolygonMiden/miden-node/issues/344

Now the in-memory data is being locked for the whole database commitment operation. Block saving started earlier, and it's possible to save unapplied block, so possible inconsistency is resolved by introducing "latest block number" variable which is updated in the same critical section, as commitment. Blocks later than latest block number are considered as candidates (unapplied blocks).

### DIFF
--- a/crates/store/src/errors.rs
+++ b/crates/store/src/errors.rs
@@ -142,6 +142,8 @@ pub enum ApplyBlockError {
     DatabaseError(#[from] DatabaseError),
     #[error("I/O error: {0}")]
     IoError(#[from] io::Error),
+    #[error("Task join error: {0}")]
+    TokioJoinError(#[from] tokio::task::JoinError),
     #[error("Concurrent write detected")]
     ConcurrentWrite,
     #[error("New block number must be 1 greater than the current block number")]
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -2,7 +2,7 @@
 //!
 //! The [State] provides data access and modifications methods, its main purpose is to ensure that
 //! data is atomically written, and that reads are consistent.
-use std::{mem, sync::Arc};
+use std::sync::Arc;
 
 use miden_node_proto::{domain::accounts::AccountInfo, AccountInputRecord, NullifierWitness};
 use miden_node_utils::formatting::{format_account_id, format_array};
@@ -21,7 +21,7 @@ use tokio::{
     sync::{oneshot, Mutex, RwLock},
     time::Instant,
 };
-use tracing::{error, info, info_span, instrument};
+use tracing::{info, info_span, instrument};
 
 use crate::{
     blocks::BlockStore,
@@ -49,6 +49,7 @@ struct InnerState {
     nullifier_tree: NullifierTree,
     chain_mmr: Mmr,
     account_tree: SimpleSmt<ACCOUNT_TREE_DEPTH>,
+    latest_block_num: BlockNumber,
 }
 
 /// The rollup state
@@ -79,8 +80,14 @@ impl State {
         let nullifier_tree = load_nullifier_tree(&mut db).await?;
         let chain_mmr = load_mmr(&mut db).await?;
         let account_tree = load_accounts(&mut db).await?;
+        let latest_block_num = load_latest_block_num(&mut db).await?;
 
-        let inner = RwLock::new(InnerState { nullifier_tree, chain_mmr, account_tree });
+        let inner = RwLock::new(InnerState {
+            nullifier_tree,
+            chain_mmr,
+            account_tree,
+            latest_block_num,
+        });
 
         let writer = Mutex::new(());
         let db = Arc::new(db);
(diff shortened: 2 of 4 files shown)


crates/store/src/state.rs
```diff
@@ -49,6 +49,7 @@ struct InnerState {
     nullifier_tree: NullifierTree,
     chain_mmr: Mmr,
     account_tree: SimpleSmt<ACCOUNT_TREE_DEPTH>,
+    latest_block_num: BlockNumber,
```
bobbinth, line 52: I think we can get the same info from `chain_mmr` - so, `latest_block_num` may not be necessary. That is, we can add a method `InnerState::latest_block_num()` but this method would get this info from `chain_mmr`.

crates/store/src/state.rs
```diff
@@ -252,40 +268,25 @@ impl State {
...
+        block_save_task.await??;
+
         // scope to update the in-memory data
         {
             let mut inner = self.inner.write().await;
 
-            // Save the block before transaction is committed:
-            self.block_store.save_block(block_num, &block_data).await?;
-
             let _ = inform_acquire_done.send(());
```
bobbinth, line 278: Could we add a bit more docs to this and the previous lines? If I understand correctly:

- Line 276: we lock the in-memory state; at this point all read requests to the store will start blocking.
- Line 278: we send a signal to commit the transaction to the database. We do this after locking the in-memory state so that there are no in-consistent reads where the state of the database has been updated but the in-memory state hasn't.

bobbinth (commented):
Looks good! Thank you! I left a couple of comments inline: the main comment is that I'm not sure we need `latest_block_num` as a separate filed as this data should be already available in `chain_mmr`.

crates/store/src/state.rs
polydez: Good point, thank you!

crates/store/src/state.rs
polydez: Yes, you're absolutely correct. I've updated the comments, please take a look!

crates/store/src/errors.rs
```diff
@@ -142,6 +142,8 @@ pub enum ApplyBlockError {
     DatabaseError(#[from] DatabaseError),
     #[error("I/O error: {0}")]
     IoError(#[from] io::Error),
+    #[error("Task join error: {0}")]
+    TokioJoinError(#[from] tokio::task::JoinError),
```
bobbinth, line 146: Do we still need to add this error variant?

crates/store/src/state.rs
```diff
@@ -229,63 +242,48 @@ impl State {
...
             .map_err(ApplyBlockError::BlockApplyingBrokenBecauseOfClosedChannel)?;
 
-        // scope to update the in-memory data
+        // Awaiting the block saving task to complete without errors
+        block_save_task.await??;
+
+        // Scope to update the in-memory data
         {
+            // We need to hold the write lock here, so that the in-memory state is not changed
             let mut inner = self.inner.write().await;
```
bobbinth, line 271: I'm not sure the comment here is entirely correct: we hold the lock not so much so that in-memory state is not changed (this would be guaranteed by locking the `writer` field on line 112) but because we want to prevent inconsistent reads. That is, if we commit the DB transaction before acquiring a lock, a read request to the store could read the old in-memory state but new database state.

bobbinth (approved):
Thank you, looks good! I left a couple of doc-related comments inline.

Also, we should probably update the comments on lines 99 - 108 to make sure they are up to date.

crates/store/src/errors.rs
polydez: Yes, this error might be thrown by `tokio::spawn` for `block_save_task` and `db_update_task`.

crates/store/src/state.rs
polydez: Yes, you're right, thank you for noticing!

mFragaBA (approved):
LGTM! I'm guessing there is no problem with this but wanted to check, now it's possible to write the tentative block to a file but the commit transaction might fail. Is there an issue with having that block around there?

I left a small commend for a log we might want to keep. Also ran the integration tests from the miden client against this node and they were all passing so that's also a good sign.

It was also a nice PR to review to get familiar with how the node applies blocks!

crates/store/src/state.rs
```diff
@@ -229,63 +246,50 @@ impl State {
...
-                    is_panic = err.is_panic(),
-                    COMPONENT,
-                    "apply_block task joined with an error"
-                );
-            },
-            Ok(Err(err)) => {
-                error!(err = err.to_string(), COMPONENT, "apply_block failed with a DB error");
-            },
-            Ok(Ok(())) => {
-                info!(%block_hash, block_num, COMPONENT, "apply_block successful");
```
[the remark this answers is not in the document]
polydez: Good catch, thank you!

polydez:
@mFragaBA,
> now it's possible to write the tentative block to a file but the commit transaction might fail. Is there an issue with having that block around there?

Yes, now we should check if requested block is less or equal to the latest applied block for such case. This guaranties to not get block candidate.

bobbinth (approved):
Looks good! Thank you!

PR #377, opened 2024-06-06
Return the list of committed transactions in state sync endpoint

polydez:
Resolves: https://github.com/0xPolygonMiden/miden-node/issues/371
Related issue: https://github.com/0xPolygonMiden/miden-base/issues/722


crates/block-producer/src/block_builder/prover/mod.rs
```diff
@@ -279,4 +279,20 @@ impl BlockProver {
...
         ))
     }
+
+    fn compute_tx_hash(&self, witness: &BlockWitness) -> Digest {
+        let elements: Vec<Felt> = witness
+            .updated_accounts
+            .iter()
+            .flat_map(|(&account_id, update)| {
+                update.transactions.iter().flat_map(move |&tx| {
+                    let mut result = vec![Felt::ZERO, Felt::ZERO, Felt::ZERO, account_id.into()];
```
mFragaBA, line 289: why do we need to add the account id to the elements to hash?

crates/store/src/db/sql.rs
```diff
@@ -640,6 +640,83 @@ pub fn select_block_headers(conn: &mut Connection) -> Result<Vec<BlockHeader>> {
...
+        SELECT
+            transaction_id
+        FROM
+            transactions
+        WHERE
+            block_num > ?1 AND
+            block_num <= ?2 AND
+            account_id IN rarray(?3)
+        ORDER BY
+            transaction_id ASC
```
mFragaBA, line 703: does the order by transaction id add anything? was it supposed to be ordered by block number?

crates/store/src/db/tests.rs
```diff
@@ -95,6 +95,32 @@ fn test_sql_insert_nullifiers_for_block() {
...
+    let mut conn = create_db();
+
+    let transactions = query_transactions(&mut conn);
+
+    assert!(transactions.is_empty(), "No elements must be initially in the DB");
+
+    let count = insert_transactions(&mut conn);
+
+    assert_eq!(count, 2, "Two elements must have been inserted");
+}
```
mFragaBA, line 122: can we also query after inserting the transactions and make sure it's also 2?

mFragaBA (commented):
code looks good! I left some minor questions. I'll test this against the client to have another layer of validation

crates/block-producer/src/block_builder/prover/mod.rs
polydez: Please look into this issue for clarifications: https://github.com/0xPolygonMiden/miden-base/issues/722

crates/store/src/db/sql.rs
polydez: I just wanted to have reproducible results. If for client it's matter to have ordering by block number, I can change it.

crates/store/src/db/tests.rs
polydez: Good catch, thank you!

Cargo.toml
```diff
@@ -25,18 +25,18 @@ readme = "README.md"
 
 [workspace.dependencies]
 miden-air = { version = "0.9", default-features = false }
-miden-lib = { version = "0.3" }
+miden-lib = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "polydez-block-tx-ids" }
```
bobbinth, line 28: These can probably just point to `next` now.

crates/store/src/state.rs
```diff
@@ -226,7 +223,7 @@ impl State {
                 .map(|(note_index, note)| {
                     let details = match note {
                         OutputNote::Full(note) => Some(note.to_bytes()),
-                        OutputNote::Header(_) => None,
+                        OutputNote::Partial(_) | OutputNote::Header(_) => None,
```
bobbinth, line 226: Similar to the comment I left in https://github.com/0xPolygonMiden/miden-node/pull/368#discussion_r1630144939 - getting a `Partial` note here should be an error. Let's create an issue for this.

polydez: I've updated the code in order to return error in a case of invalid output note type. If I don't catch your point, please let me know, I will create an issue for this.

crates/proto/proto/responses.proto
```diff
@@ -45,11 +45,15 @@ message SyncStateResponse {
     // List of account hashes updated after `block_num + 1` but not after `block_header.block_num`
     repeated account.AccountSummary accounts = 5;
 
+    // List of transaction IDs which changed the state of requested accounts
+    // between `block_num + 1` and `block_header.block_num`
+    repeated digest.Digest transactions = 6;
```
bobbinth, line 50: I would update the comment to be: "List of IDs for transactions executed against requested accounts between `block_num + 1` and `block_header.block_num`."

crates/block-producer/src/block_builder/prover/mod.rs
bobbinth, line 289: Let's position `account_id` as the first element. Specifically:
```Rust
let mut result = vec![account_id.into(), ZERO, ZERO, ZERO];
```
(we can import `ZERO` from `miden-base`).

Also, we could probably rewrite this as:
```Rust
let mut elements = Vec::new();
for (&account_id, update) in witness.updated_accounts.iter() {
    for tx in update.transactions.iter() {
        elements.extend_from_slice(&[account_id.into(), ZERO, ZERO, ZERO]);
        elements.extend_from_slice(tx.as_elements());
    }
}
```
This is a bit more compact and may avoid a bunch of vector allocations.

bobbinth (changes requested):

Also, I think one thing that is missing is verifying that `tx_hash` is consistent with transaction info in the block during [apply_block](https://github.com/0xPolygonMiden/miden-node/blob/next/crates/store/src/state.rs#L115) (we check that all other data in the block is consistent with the header, would be good to check that transaction data is consistent as well).

Lastly, what do you think of moving transaction IDs into `AccountSummary` (something along the lines what is suggested in https://github.com/0xPolygonMiden/miden-node/issues/371#issuecomment-2143112925). I don't think we should do it in this PR, but curious how much work it would be.

bobbinth:
Oh - also, let's resolve merge conflicts and update the changelog.

polydez:
@bobbinth 
> Lastly, what do you think of moving transaction IDs into `AccountSummary` (something along the lines what is suggested in https://github.com/0xPolygonMiden/miden-node/issues/371#issuecomment-2143112925). I don't think we should do it in this PR, but curious how much work it would be.

I looked into the code and it seems to be pretty easy to implement this.

crates/block-producer/src/block_builder/prover/mod.rs
```diff
@@ -279,4 +279,16 @@ impl BlockProver {
...
+        let mut elements = Vec::with_capacity(witness.updated_accounts.len() * 8);
+        for (&account_id, update) in witness.updated_accounts.iter() {
+            for tx in update.transactions.iter() {
+                elements.extend_from_slice(&[account_id.into(), ZERO, ZERO, ZERO]);
+                elements.extend_from_slice(tx.as_elements());
+            }
+        }
+
+        Hasher::hash_elements(&elements)
+    }
```
igamigo, line 293: Should this be a method of `BlockWitness`?

crates/store/src/state.rs
```diff
@@ -574,3 +571,15 @@ async fn load_accounts(
     SimpleSmt::with_leaves(account_data)
         .map_err(StateInitializationError::FailedToCreateAccountsTree)
 }
+
+fn compute_tx_hash(updated_accounts: &[BlockAccountUpdate]) -> Digest {
```
igamigo, line 575: This logic for this function is very similar to the one for `BlockWitness`, should it be factored out to avoid problems if it changes? (although likely tests would catch it so if it's more convenient this way it's fine)

igamigo (approved):
LGTM, left a couple of comments.

crates/store/src/state.rs
polydez: The logic is similar, but it operates with different structures with different layouts. Thus, it's not easy to write abstractions to deal with both of them. Luckily, one of the methods is being used to check another one, so if one of them is changed, this check will fail.

mFragaBA (commented):
One thing I noticed we might still have problems with in the client is that we're still not getting the info of _when_ the transaction was included in a block. This is due to the fact that we return a sync state response with all nullifiers and transaction ids in the range (request.block_header, next_block_header) where block header is the block of the first note matching any of the requests' tags or account ids. So, if many TXs that only consume notes are sent and then we sync after a while, we'll still receive the TXs ids but won't know which got included in which block. A similar thing happens with nullifiers as well.

crates/store/src/state.rs
bobbinth: I do agree with @igamigo that having this logic in two different places is not ideal. And now that I think about it, we should probably move it to `miden-base`. Specifically, we could do the following:

1. Implement a generic function `compute_tx_hash()` which takes an iterator over `(TransactionId, AccountId)` tuples and returns the hash.
2. Implement `Block::transactions()` method and `BlockWitness::transactions()` methods which would return iterators over these tuples.
3. I would probably also implement `Block::compute_tx_hash()` method which returns the tx hash.

bobbinth:
> One thing I noticed we might still have problems with in the client is that we're still not getting the info of _when_ the transaction was included in a block. This is due to the fact that we return a sync state response with all nullifiers and transaction ids in the range (request.block_header, next_block_header) where block header is the block of the first note matching any of the requests' tags or account ids. So, if many TXs that only consume notes are sent and then we sync after a while, we'll still receive the TXs ids but won't know which got included in which block. A similar thing happens with nullifiers as well.

Ah - good point! We could return `(TransactionId, block_num)` and `(Nullifier, block_num)` tuples. But that's separate from this PR. Could you create a new issue for this?

bobbinth (changes requested):
Looks good! Thank you!

Before merging this, let's do the changes I mentioned [here](https://github.com/0xPolygonMiden/miden-node/pull/377#discussion_r1631682221) (will require a PR in `miden-base`).

In a follow-up PR we can update the structure of returned data (i.e., move transaction IDs into account summaries, return block numbers for nullifiers and transaction IDs).

mFragaBA:
also, I think after https://github.com/0xPolygonMiden/miden-base/pull/732 got merged in base it's breaking compilation.

bobbinth:
> also, I think after [0xPolygonMiden/miden-base#732](https://github.com/0xPolygonMiden/miden-base/pull/732) got merged in base it's breaking compilation, idk if we should fix it here or should I open a new PR (I still need these changes to compile the node)

Ah, yes! @polydez could you update this PR to make sure it works with latest `next` from `miden-base`? For example, the `format_input_notes()` function could look something like this:
```Rust
pub fn format_input_notes(notes: &InputNotes<InputNoteCommitment>) -> String {
    format_array(notes.iter().map(|c| match c.note_id() {
        Some(note_id) => format!("({}, {})", c.nullifier().to_hex(), note_id.to_hex()),
        None => format!("({})", c.nullifier().to_hex()),
    }))
}
```

After this, we'll merge this PR and can address my comments from https://github.com/0xPolygonMiden/miden-node/pull/377#discussion_r1631682221 in a separate PR.

mFragaBA:
> > One thing I noticed we might still have problems with in the client is that we're still not getting the info of _when_ the transaction was included in a block. This is due to the fact that we return a sync state response with all nullifiers and transaction ids in the range (request.block_header, next_block_header) where block header is the block of the first note matching any of the requests' tags or account ids. So, if many TXs that only consume notes are sent and then we sync after a while, we'll still receive the TXs ids but won't know which got included in which block. A similar thing happens with nullifiers as well.
> 
> Ah - good point! We could return `(TransactionId, block_num)` and `(Nullifier, block_num)` tuples. But that's separate from this PR. Could you create a new issue for this?

created #380

bobbinth (approved):
Looks good! Thank you! I've pushed a couple of small commits to point this back at `next` in `miden-base` and also did some minor refactoring.

bobbinth:
> created #380. I'm open to taking it

Thank you! I'll assign it to you.

PR #379, opened 2024-06-07
fix: return error if partial note is found

tomyrd:


tomyrd:
Closing as this is fixed in #377

PR #382, opened 2024-06-11
fix: Preserve transaction ordering in `BlockWitness`

igamigo:
We found that the integration tests on the client were sometimes failing after the recent changes related to adding the transactions to the block structures. @mFragaBA investigated a bit and saw that the node was logging an `InvalidTxHash` error which implied problems with checking the consistency between the header and the block's transactions, concluding that the tests were flaky because this would only occur when a couple of transactions for different accounts are batched into the same block. This is because the `BlockWitness` keeps the transaction data in a `BTreeMap`, which breaks ordering with the iterator. There are structs that would preserve it while providing dictionary functionality, but in this case it seems we can just replace it with a `Vec`.
Ran the tests with these changes locally and [in the CI](https://github.com/0xPolygonMiden/miden-client/actions/runs/9469968083/job/26090499353) and it seems tests consistently pass again.

### DIFF
--- a/crates/block-producer/src/block_builder/prover/block_witness.rs
+++ b/crates/block-producer/src/block_builder/prover/block_witness.rs
@@ -21,7 +21,7 @@ use crate::{
 /// Provides inputs to the `BlockKernel` so that it can generate the new header.
 #[derive(Debug, PartialEq)]
 pub struct BlockWitness {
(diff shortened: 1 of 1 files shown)


bobbinth (approved):
Nice catch! Thank you!

Issue #380, opened 2024-06-10
Return block height info along nullifiers and transaction ids


mFragaBA:
>> One thing I noticed we might still have problems with in the client is that we're still not getting the info of _when_ the transaction was included in a block. This is due to the fact that we return a sync state response with all nullifiers and transaction ids in the range (request.block_header, next_block_header) where block header is the block of the first note matching any of the requests' tags or account ids. So, if many TXs that only consume notes are sent and then we sync after a while, we'll still receive the TXs ids but won't know which got included in which block. A similar thing happens with nullifiers as well.
>
>Ah - good point! We could return `(TransactionId, block_num)` and `(Nullifier, block_num)` tuples. But that's separate from this PR. Could you create a new issue for this?

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-node/issues/377#issuecomment-2155540500_

bobbinth:
As a part of this, let's do the following:

1. Investigate moving transaction data into `AccountSummary` (rather than having it as a separate filed in the response).
2. Update endpoint docs in readme files (i.e., [here](https://github.com/0xPolygonMiden/miden-node/tree/next/crates/rpc#syncstate) and [here](https://github.com/0xPolygonMiden/miden-node/blob/next/crates/store/README.md#syncstate)).

mFragaBA:
small update:

- I checked and the node already sends the block height for each `Nullifier`, we just were not using the info on the client side
- as for the transactions, I made a test branch named `﻿mFragaBA-return-tx-and-nullifier-block-height-info` with the "naive" implementation (just adding the block height along with the ID)
- I'm gonna follow up with another branch where I change the AccountSummary message as said [here](https://github.com/0xPolygonMiden/miden-node/issues/371#issuecomment-2143112925). I think it's doable, though I still don't quite understand the use case (my thought process is that a client will only care about the txs generated by them, so they should have all the info needed).

- I still need to update the docs, but I'll do that once we decide for either option
- I'll also update the client's code once we decide on either option

igamigo:
When you mention "either option", do you mean coupling the block height to the transaction ID versus updating `AccountSummary`?
I disagree that the client would only care about transactions created locally. I think I mentioned it in the issue or we discussed offline, but for public accounts it makes sense that you can understand if and how many transactions were executed. In any case it should be marginal impact for a good amount of usable information: once you are tracking more than one account you wouldn't be able to tell how many transactions were executed on each account otherwise.

[...]

PR #383, opened 2024-06-12
feat: return tx block height info

mFragaBA:
addresses #380. 

Includes the block where a transaction was included and the account it was executed against in the state sync response. In this PR, we include that information within the existing `AccountSummary`. A few issues arise from this though:

- first, I couldn't just modify the `AccountSummary` protobuf message as it's reused in other places. Hence I made a wrapper named `AccountUpdate` (similar to the `NullifierUpdate` name).
  - `AccountUpdate` contains an `AccountSummary` and a list of `TransactionInfo` (which in turn contains the transaction ID and block number)
- I had to change a bit `select_transactions_by_accounts_and_block_range`. Now it returns a mapping `account_id => TransactionInfo`
- Then, on the sync state handler we need to "merge" the `TransactionInfo`s with the fetched account summaries. This poses a new issue as some transactions might not have a matching account summary if a client with old state does a sync request (@bobbinth mentioned situation where this happens in the issue). 
  - By just wrapping the `AccountSummary` struct, one problem is that the account_id for `AccountUpdate` is within the optional `AccountSummary`. So that will require to either modify `AccountSummary`, to define a new message that stores almost the same info as `AccountSummary` or to embed the fields within `AccountUpdate`

**Note**: This is a POC implementation, made so we can compare this vs. [a decoupled approach](https://github.com/0xPolygonMiden/miden-node/compare/mFragaBA-return-tx-and-nullifier-block-height-info?expand=1). 

TODO:

- [ ] code cleanup
- [ ] update documentation


bobbinth:
> **Note**: This is a POC implementation, made so we can compare this vs. [a decoupled approach](https://github.com/0xPolygonMiden/miden-node/compare/mFragaBA-return-tx-and-nullifier-block-height-info?expand=1).

Could you create a PR for the "decoupled approach" as well? Would make it a bit easier to compare the two.

mFragaBA:
> > **Note**: This is a POC implementation, made so we can compare this vs. [a decoupled approach](https://github.com/0xPolygonMiden/miden-node/compare/mFragaBA-return-tx-and-nullifier-block-height-info?expand=1).
> 
> Could you create a PR for the "decoupled approach" as well? Would make it a bit easier to compare the two.

https://github.com/0xPolygonMiden/miden-node/pull/386 done!

mFragaBA:
closing this PR since we ended up going with the decoupled approach on #386

Issue #342, opened 2024-05-03
[Task]: Store Note's.`aux` field in the database


igamigo:
### Node's components affected by this task

- [ ] RPC
- [X] Block producer
- [X] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

We are currently hardcoding `aux` fields of the note metadata internally. We should change the code to pass the value accordingly and storing it in the database.

### How should it be done?

Add a column in the notes SQL table schema for the `aux` field, and store/retrieve it accordingly.

### When is this task done?

When the SQL table includes the `aux` field and it is stored and retrieved accordingly.

bobbinth:
Closed by #384.

PR #384, opened 2024-06-12
feat: store and retrieve aux field

mFragaBA:
addresses #342.

I tried testing this using the integration tests on miden-client but they fail because I still can't use any other value than 0 for the aux field until https://github.com/0xPolygonMiden/miden-base/issues/731 is done. 

That being said, I somewhat tested this change by changing the hardcoded 0 value in the transaction kernel's create_note and using the same value for notes on the integration tests. They ran successfully with some fixes present in https://github.com/0xPolygonMiden/miden-client/pull/381 (mainly removing hardcoded values from the client side)

- the client was able to consume the notes
- I checked the node's db using `sqlite3` and the aux column of `notes` had the proper values

### DIFF
--- a/crates/store/src/db/sql.rs
+++ b/crates/store/src/db/sql.rs
@@ -8,8 +8,8 @@ use miden_objects::{
     block::{BlockAccountUpdate, BlockNoteIndex},
     crypto::{hash::rpo::RpoDigest, merkle::MerklePath},
     notes::{NoteId, NoteMetadata, NoteType, Nullifier},
-    utils::{serde::{Deserializable, Serializable}, DeserializationError},
-    BlockHeader, Word,
+    utils::serde::{Deserializable, Serializable},
+    BlockHeader,
 };
 use rusqlite::{
     params,
@@ -352,10 +352,9 @@ pub fn select_notes(conn: &mut Connection) -> Result<Vec<NoteRecord>> {
         let sender = column_value_as_u64(row, 5)?;
         let tag: u32 = row.get(6)?;
         let aux: u64 = row.get(7)?;
-        let aux = aux.try_into().map_err(|err| DatabaseError::DeserializationError(DeserializationError::InvalidValue(err)))?;
+        let aux = aux.try_into().map_err(DatabaseError::InvalidFelt)?;
 
-        let metadata =
-            NoteMetadata::new(sender.try_into()?, note_type, tag.into(), aux)?;
+        let metadata = NoteMetadata::new(sender.try_into()?, note_type, tag.into(), aux)?;
 
         notes.push(NoteRecord {
             block_num: row.get(0)?,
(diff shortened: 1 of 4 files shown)


crates/store/src/db/sql.rs
```diff
@@ -483,17 +486,19 @@ pub fn select_notes_since_block_by_tag_and_sender(
...
-        let merkle_path_data = row.get_ref(7)?.as_blob()?;
+        let aux: u64 = row.get(7)?;
+        let aux = aux.try_into().map_err(|err| DatabaseError::DeserializationError(DeserializationError::InvalidValue(err)))?;
+        let merkle_path_data = row.get_ref(8)?.as_blob()?;
         let merkle_path = MerklePath::read_from_bytes(merkle_path_data)?;
-        let details_data = row.get_ref(8)?.as_blob_or_null()?;
+        let details_data = row.get_ref(9)?.as_blob_or_null()?;
         let details = details_data.map(<Vec<u8>>::read_from_bytes).transpose()?;
 
         // TODO: Properly retrieve aux
```
igamigo, line 496: We should be good to remove this `TODO`!

crates/store/src/db/sql.rs
```diff
@@ -483,17 +486,19 @@ pub fn select_notes_since_block_by_tag_and_sender(
         let note_type = row.get::<_, u8>(4)?;
         let sender = column_value_as_u64(row, 5)?;
         let tag: u32 = row.get(6)?;
-        let merkle_path_data = row.get_ref(7)?.as_blob()?;
+        let aux: u64 = row.get(7)?;
+        let aux = aux.try_into().map_err(|err| DatabaseError::DeserializationError(DeserializationError::InvalidValue(err)))?;
```
igamigo, line 490: I'm not sure this error represents the problem correctly here. Looking at the definition it seems that it's related to problems deserializing blobs where here, the `u64` was already deserialized from the database. But also, not sure if there is a better representation there. Maybe introducing a new `InvalidFelt` variant?

crates/store/src/db/sql.rs
igamigo: Also, these same changes need to be applied to the other note queries

crates/store/src/db/sql.rs
mFragaBA: done!

crates/store/src/db/sql.rs
mFragaBA: done! I removed the TODO and applied the same change to `select_note` and `select_note_by_id`

bobbinth (approved):
All looks good! Thank you!

@Dominik1999 - could you tackle https://github.com/0xPolygonMiden/miden-client/pull/381 next so that we can test/close this as well?

Dominik1999:
There you go https://github.com/0xPolygonMiden/miden-base/pull/752

bobbinth:
@mFragaBA - now that https://github.com/0xPolygonMiden/miden-base/pull/752 is merged, could you confirm that all is working here as well? After this we can merge this PR as well.

igamigo:
Was able to test this works correctly by doing the whole circuit from the client: minting an asset, consuming and sending it with different `aux` values, and the node data validates correctly:

image

PR #386, opened 2024-06-13
feat: return tx block height info (decoupled approach)

mFragaBA:
addresses #380 

This is an alternative approach to #383


crates/proto/proto/responses.proto
```diff
@@ -32,6 +32,12 @@ message NullifierUpdate {
     fixed32 block_num = 2;
 }
 
+message TransactionUpdate {
+    digest.Digest transaction_id = 1;
+    fixed32 block_num = 2;
+    fixed64 account_id = 3;
+}
```
bobbinth, line 39: I wonder if `TransactionSummary` would be a better name here (would be parallel with `AccountSummary`).

Also, as discussed in the issue, let's create a separate message for `TransactionId`.

crates/proto/proto/responses.proto
```diff
@@ -47,7 +53,7 @@ message SyncStateResponse {
 
     // List of IDs for transactions executed against requested accounts between
     // `block_num + 1` and `block_header.block_num`
-    repeated digest.Digest transactions = 6;
+    repeated TransactionUpdate transactions = 6;
```
bobbinth, line 56: To make things more clear - should we say `request.block_num + 1` and `self.block_header.block_num`? Otherwise, I'm not sure it is clear what `block_num` refers to.

Also, similar comment for other doc comments for this message.

bobbinth (approved):

As mentioned in the issue, I think this is a better approach as additional code complexity is pretty low and the downsides of doing this vs. putting transactions into the `AccountSummary` are pretty small.

bobbinth:
Oh - and let's update the readme docs and the changelog.

crates/proto/proto/responses.proto
```diff
@@ -32,6 +32,12 @@ message NullifierUpdate {
     fixed32 block_num = 2;
 }
 
+message TransactionUpdate {
+    digest.Digest transaction_id = 1;
+    fixed32 block_num = 2;
+    fixed64 account_id = 3;
```
igamigo, line 38: Is there any reason why this is `fixed64` instead of `account.AccountId`?

igamigo (approved):
LGTM! Left a comment/question.

crates/proto/proto/responses.proto
mFragaBA: done! One kind of annoying thing I learned is that proto3 doesn't have a required field, which makes it kind of annoying to have all the `Option`s laying around 😅

crates/proto/proto/responses.proto
mFragaBA: nope, changed it for account.AccountId

crates/proto/proto/responses.proto
bobbinth: yep - `request` / `response` works just fine.

crates/store/src/server/api.rs
```diff
@@ -141,7 +142,15 @@ impl api_server::Api for StoreApi {
             })
             .collect();
 
-        let transactions = state.transactions.into_iter().map(Into::into).collect();
+        let transactions = state
+            .transactions
+            .into_iter()
+            .map(|transaction_info| TransactionSummary {
```
bobbinth, line 148: minor nit: variable name could probably be `transaction_summary` or just `transaction`.

crates/store/src/db/mod.rs
```diff
@@ -48,6 +49,13 @@ pub struct NullifierInfo {
     pub block_num: BlockNumber,
 }
 
+#[derive(Debug, PartialEq)]
+pub struct TransactionInfo {
```
bobbinth, line 53: Should we rename this into `TransactionSummary` as well?

crates/store/README.md
```diff
@@ -149,10 +149,11 @@ contains excessive notes and nullifiers, client can make additional filtering of
 
 - `chain_tip`: `uint32` – number of the latest block in the chain.
 - `block_header`: `BlockHeader` – block header of the block with the first note matching the specified criteria.
-- `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `block_num + 1` to `block_header.block_num`.
-- `accounts`: `[AccountSummary]` – account summaries for accounts updated after `block_num + 1` but not after `block_header.block_num`.
-- `notes`: `[NoteSyncRecord]` – a list of all notes together with the Merkle paths from `block_header.note_root`.
-- `nullifiers`: `[NullifierUpdate]` – a list of nullifiers created between `block_num + 1` and `block_header.block_num`.
+- `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `request.block_num + 1` to `self.block_header.block_num`.
+- `accounts`: `[AccountSummary]` – account summaries for accounts updated after `request.block_num + 1` but not after `self.block_header.block_num`.
+- `transactions`: `[TransactionSummary]` – transaction summaries for transactions included after `request.block_num + 1` but not after `self.block_header.block_num`.
```
bobbinth, line 154: I would also say that each transaction summary is a tuple `(transaction_id, account_id, block_num)`.

crates/store/README.md
```diff
@@ -149,10 +149,11 @@ contains excessive notes and nullifiers, client can make additional filtering of
...
+- `notes`: `[NoteSyncRecord]` – a list of all notes together with the Merkle paths from `self.block_header.note_root`.
+- `nullifiers`: `[NullifierUpdate]` – a list of nullifiers created between `request.block_num + 1` and `self.block_header.block_num`.
```
bobbinth, line 156: I would make it clear that each entry here is a tuple `(nullifier, block_num)`.

crates/rpc/README.md
```diff
@@ -105,10 +105,11 @@ contains excessive notes and nullifiers, client can make additional filtering of
...
-- `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `block_num + 1` to `block_header.block_num`.
-- `accounts`: `[AccountSummary]` – account summaries for accounts updated after `block_num + 1` but not after `block_header.block_num`.
-- `notes`: `[NoteSyncRecord]` – a list of all notes together with the Merkle paths from `block_header.note_root`.
-- `nullifiers`: `[NullifierUpdate]` – a list of nullifiers created between `block_num + 1` and `block_header.block_num`.
+- `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `request.block_num + 1` to `self.block_header.block_num`.
+- `accounts`: `[AccountSummary]` – account summaries for accounts updated after `request.block_num + 1` but not after `self.block_header.block_num`.
+- `transactions`: `[TransactionSummary]` – transaction summaries for transactions included after `request.block_num + 1` but not after `self.block_header.block_num`.
+- `notes`: `[NoteSyncRecord]` – a list of all notes together with the Merkle paths from `self.block_header.note_root`.
+- `nullifiers`: `[NullifierUpdate]` – a list of nullifiers created between `request.block_num + 1` and `self.block_header.block_num`.
```
bobbinth, line 112: Similar comments as above.

bobbinth (approved):
Looks good! Thank you! I left a few comments inline - mostly related to naming/docs.

Should we take this PR out of "draft" mode?

crates/proto/proto/responses.proto
mFragaBA: changed to request/response

crates/rpc/README.md
mFragaBA: done!

crates/store/README.md
mFragaBA: done!

crates/store/README.md
mFragaBA: done!

crates/store/src/db/mod.rs
mFragaBA: sure, done!

crates/store/src/server/api.rs
mFragaBA: renamed to `transaction_summary`

mFragaBA:
I'll close the other PR (#383)

bobbinth (approved):
All looks good! Thank you!

Do we have an issue in `miden-client` to make use of these changes?

PR #387, opened 2024-06-14
fix: Add `aux` input to faucet

tomyrd:
With the [recent change](https://github.com/0xPolygonMiden/miden-base/pull/752) in `miden-base` the faucet stopped working. This PR fixes the note script the faucet uses to send assets.

### DIFF
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -105,6 +105,7 @@ impl FaucetClient {
             target_account_id,
             vec![asset.into()],
             note_type,
+            Default::default(),
             &mut self.rng,
         )
         .map_err(|err| FaucetError::InternalServerError(err.to_string()))?;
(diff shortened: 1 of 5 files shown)


bin/faucet/src/client.rs
```diff
@@ -316,6 +317,7 @@ fn build_transaction_arguments(
         &DISTRIBUTE_FUNGIBLE_ASSET_SCRIPT
             .replace("{recipient}", &recipient)
             .replace("{note_type}", &Felt::new(note_type as u64).to_string())
+            .replace("{aux}", &Felt::default().to_string())
```
igamigo, line 320: Might be best to do `note.metadata().aux().to_string()` here so that it comes from the created note, even if it's hardcoded there as well.

igamigo (approved):
LGTM! Left a small comment

bin/faucet/src/client.rs
tomyrd: Done!

bobbinth (approved):
Looks good! Thank you! But the tests seem to be failing?

PR #390, opened 2024-06-24
Implemented future notes support in block-producer

polydez:
Resolves https://github.com/0xPolygonMiden/miden-node/issues/381


bobbinth (commented):
Thank you! Looks good. This is not a full review, but I left some comments inline.

crates/block-producer/src/state_view/mod.rs
```diff
@@ -163,23 +195,39 @@ fn ensure_in_flight_constraints(
...
+
     Ok(())
 }
 
+/// Ensures the constraints related to transaction inputs:
+/// - the candidate transaction's initial account state hash must be the same as the one
+///   in the Store or empty for new accounts
+/// - input notes must not be already consumed
+///
+/// Returns a list of input notes that were not found in the Store
```
bobbinth, line 225: I would write the last sentence as: "Returns a list of unauthenticated input notes that were not found in the store."

crates/block-producer/src/txqueue/mod.rs
```diff
@@ -94,7 +94,7 @@ where
                 return;
             }
 
-            locked_ready_queue.drain(..).collect()
+            locked_ready_queue.drain(..).rev().collect()
```
bobbinth, line 97: Why do we need this `.rev()` here?

crates/block-producer/src/store/mod.rs
```diff
@@ -61,6 +76,8 @@ pub struct TransactionInputs {
     /// Maps each consumed notes' nullifier to block number, where the note is consumed
     /// (`zero` means, that note isn't consumed yet)
     pub nullifiers: BTreeMap<Nullifier, u32>,
+    /// List of notes that were not found in the store
+    pub missing_notes: Vec<NoteId>,
```
bobbinth, line 80: Let's rename this to `missing_unauthenticated_notes` and update the comment accordingly.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -129,6 +158,18 @@ where
...
+        let dangling_notes = self.find_dangling_notes(&txs).await;
+        if !dangling_notes.is_empty() {
+            let missing_notes = match self.store.get_missing_notes(&dangling_notes).await {
+                Ok(notes) => notes,
+                Err(err) => return Err(BuildBatchError::GetMissingNotesRequestError(err, txs)),
+            };
+
+            if !missing_notes.is_empty() {
+                return Err(BuildBatchError::UnauthenticatedNotesNotFound(missing_notes, txs));
+            }
```
bobbinth, line 170: As discussed offline, let's invert this and use `GetNotesById` to retrieve notes which are already in the store.

Let's also create an issue to make a specialized endpoint later which would return both note IDs and note inclusion paths.

bobbinth (commented):

Also, we'll need to:
- Rebase from the latest `next` as protobuf files are now in a different place.
- Make update needed to account for changes introduced in [0xPolygonMiden/miden-base#772](https://github.com/0xPolygonMiden/miden-base/pull/772).
- Update the changelog.

crates/block-producer/src/batch_builder/mod.rs
polydez: I've created an issue for this: https://github.com/0xPolygonMiden/miden-node/issues/395

crates/block-producer/src/txqueue/mod.rs
polydez: @bobbinth our previous solution processed transactions in reverse order (LIFO) and shuffled them in a case of error during batch building. I changed this ordering to FIFO with keeping the same order in a case of error. Processing transaction in FIFO order is important in order to reduce batch building errors with unathorized notes. 

For example:
1. Let transaction `B` consumes unauthorized note from transaction `A`. 
2. If transaction `B` is processed first and the distance between these transaction in the queue is high enough for them to be put into separate batches, transaction `A` won't be in the ready batches list at the time of processing transaction `B`.
3. This will lead to batch building error and unprocessed transaction will be inserted to the end of the queue.
4. Then transaction `A` will be processed and added to the ready batches list.
5. Eventually transaction `B` will be also processed.

My current solution relies on transaction ordering in order to eliminate such errors. My only concern, is it possible  for transaction `B` to be put into the queue before transaction `A`? In this case the whole queue processing will stop because of constant "unauthorized notes not found" errors. If such is possible, I'd propose to do shuffling of the queue as we did before my changes.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -44,38 +47,50 @@ impl TransactionBatch {
...
+        // Populate batch created notes, filtering out notes consumed in the same batch.
+        // Consumed notes are also removed from the unauthenticated input note map.
+        let output_notes: Vec<_> = txs
             .iter()
-            .flat_map(|tx| tx.input_notes().iter())
-            .map(|note| note.nullifier())
+            .flat_map(|tx| tx.output_notes().iter())
+            .filter(|&note| !unauthenticated_input_notes.remove(&note.id()))
+            .cloned()
             .collect();
```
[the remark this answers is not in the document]
polydez: @bobbinth:
1. This situation is impossible because such transaction will not be included in the transaction queue at the very beginning of the pipeline, because we require input unauthorized note to be present in in-flight notes or in the store.
2. Good catch, it might be possible.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -44,38 +47,50 @@ impl TransactionBatch {
...
+            // TODO: we need to handle a possibility that a batch contains multiple transactions against
+            //       the same account (e.g., transaction `x` takes account from state `A` to `B` and
+            //       transaction `y` takes account from state `B` to `C`). These will need to be merged
+            //       into a single "update" `A` to `C`.
+            updated_accounts.push((tx.id(), tx.account_update().clone()));
+
+            for note in tx.input_notes() {
+                produced_nullifiers.push(note.nullifier());
+                if let Some(note_id) = note.note_id() {
+                    unauthenticated_input_notes.insert(note_id);
```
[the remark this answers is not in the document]
polydez: @bobbinth, according to my current solution, unauthenticated note `x` would be removed from the set on the first consuming and the second attempt to consume it would raise an error. But let's also forbid producing note with the same ID twice, it might help with debugging of such cases.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -44,38 +47,61 @@ impl TransactionBatch {
...
+                produced_nullifiers.push(note.nullifier());
+                if let Some(header) = note.header() {
+                    if !unauthenticated_input_notes.insert(header.id()) {
+                        return Err(BuildBatchError::DuplicateUnauthenticatedNote(
+                            header.id(),
+                            txs,
+                        ));
+                    }
+                }
+            }
```
bobbinth, line 70: With this logic, even if a note was produced and consumed in the same batch, we would still generate a nullifier for it. This means for all "ephemeral" notes we'll need to update the nullifier tree. It does simplify some things, but also may create quite a bit of an overhead. I'm still thinking through whether we should do something else here.

bobbinth (changes requested):
Looks good! Thank you. I left some comments inline. The main ones are about not loading all the notes into memory in the store.

Also, I'm still thinking about how the logic for batching notes will work. Will write more thoughts on this later.

bobbinth (approved):
I'm going to merge this PR to "unbreak" the `next` branch - but let's create a new PR to address the outstanding things. These are:

1. Address some minor comments from the previous review (e.g., [this one](https://github.com/0xPolygonMiden/miden-node/pull/390#discussion_r1659364585) and [this one](https://github.com/0xPolygonMiden/miden-node/pull/390#discussion_r1659408181)).
2. Remove loading note into memory from the store.
3. Fix transaction batching logic.

The last point refers to [this comment](https://github.com/0xPolygonMiden/miden-node/pull/390#discussion_r1659429275), but to describe things more concretely (and add some new info):

When we combine transactions into a batch, we want to preserve the following properties:

1. If a note was produced by one transaction and consumed by another transaction in the same batch, the note should not be added to the list of output notes for the batch (this already happens) and the nullifier of this note should not be added to the list of nullifiers for the batch.
  a. The matching should be done based on note ID (i.e., we compare note IDs to figure out if two notes are the same), but if input/output notes with the same IDs have different hashes (i.e., the metadata is different) we should return an error.
2. If an unauthenticated note is present in the store at the time when we create the batch, this note should no longer be included in the list of unauthenticated notes.
  a. The assumption here is that we'll verify authentication info for this note in the batch kernel.
4. We should make sure that the list of unauthenticated input notes and output notes across all transactions in a batch is unique. This should be done before we match an eliminate notes which are produced and consumed in the same batch.
  a. We may relax this restriction later, but I think for now it is the easiest to enforce.

When we aggregate batches into a block, the logic should be slightly different. Specifically, notes which are produced and consumed in different batches of the same block do note get "erased". That is, they are still added to the block's note tree and their nullifiers are added to the list of nullifiers produced by the block.

One thing I was thinking is that maybe instead of tracking unauthenticated notes in a separate field of the `TransactionBatch` struct, we could do it more like we do in the `ProvenTransaction` struct. Specifically, we `TransactionBatch` could look something like this:
```rust
pub struct TransactionBatch {
    id: BatchId,
    updated_accounts: Vec<(TransactionId, TxAccountUpdate)>,
    input_notes: Vec<InputNoteCommitment>,
    output_notes: Vec<OutputNote>,
    output_notes_smt: BatchNoteTree,
}
```
Where `InputNoteCommitment` is the same as the one we use in the `ProvenTransaction`. The list of input notes will include the following:

1. All authenticated notes (i.e., the ones missing the `header` field) would be in the list in its original form.
2. All unauthenticated notes which could be matched against output notes in the same batch would be removed from the list.
3. All unauthenticated notes which were in the store at the time when the batch was created, would be transformed into authenticated notes (i.e., their `header` field would be cleared).

Issue #388, opened 2024-06-20
[Task]: Add WASM Support for miden-node-proto Crate


dagarcia7:
### Node's components affected by this task

- [X] RPC
- [ ] Block producer
- [ ] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

Currently, adding the `miden-node-proto` crate as a dependency for any crate that attempts to support WASM builds fails due to having `tonic` as a dependency. This has had consequences in both the `miden-client` repo and the `miden-wasm` crate that is currently under development and not yet in a miden repository. 

**How this affects the `miden-client` crate:**
* Imports like these https://github.com/0xPolygonMiden/miden-client/blob/68aed5e91705ab7dc980c3b05ba7921e5bc37efc/src/errors.rs#L6, don't work when building the `miden-client` for WASM targets so we had to add a feature to conditionally compile this import for non-wasm builds.
* We also had to copy and paste code https://github.com/0xPolygonMiden/miden-client/blob/68aed5e91705ab7dc980c3b05ba7921e5bc37efc/src/errors.rs#L22 directly from the `miden-node-proto` repo to get it to compile for wasm builds. 

**How this affects the `miden-wasm` crate:**
* Because the `ApiClient` and `ApiServer` are generated via `tonic-build` and thus create auto-generated code that relies on `tonic`, I am also unable to import this code or the `miden-node-proto` crate in general from the `miden-wasm` crate. 
* This code is important to be accessible for this crate because instead of duplicating this code in the `miden-wasm` crate and creating my own wasm-compatible rpc client, there should be a way to support both the regular tonic rpc client that the `miden-client` CLI uses, and a wasm compatible rpc client that can be consumed by the `miden-wasm` crate. 

### How should it be done?

These are just preliminary thoughts, but a potential approach so solve this problem could be:
* Add a feature to the `miden-node-proto` crate that will allow us to conditionally add (or exclude) the `tonic` crate and any other WASM-incompatible crates. 
* Use this feature in combination with https://crates.io/crates/tonic-web-wasm-client to create an auto-generated `ApiClient` (called `WebApiClient`, `WasmApiClient`, or something similar) so that this client can be imported into the `miden-wasm` crate without any issues. 

### When is this task done?

This task is done when the following two are possible: 
* `miden-node-proto` can be listed as a dependency in the `miden-client` for both WASM and non-WASM targets
[...]

PR #391, opened 2024-06-25
feat: Distribute protobuf files through a new crate

igamigo:
Addresses #388


Cargo.toml
```diff
@@ -3,6 +3,7 @@ members = [
   "bin/node",
   "bin/faucet",
   "crates/block-producer",
+  "crates/miden-rpc-proto",
```
bobbinth, line 6: nit: since we don't prefix other crates with `miden-`, I'm not sure we need to do it here either. We should probably change this to `rpc-proto` (the name of the crate would still be `miden-rpc-proto`).

crates/miden-rpc-proto/README.md
```diff
@@ -0,0 +1,9 @@
+# Miden node utils
```
bobbinth, line 1: nit: the name here should be updated.

crates/miden-rpc-proto/src/lib.rs
```diff
@@ -0,0 +1,22 @@
+use std::{fs::{self, File}, io::Write, path::Path};
+
+mod proto_files;
+pub use proto_files::PROTO_FILES;
+
+/// Writes the RPC protobuf file into `target_dir`.
+pub fn write_proto(target_dir: &Path) -> Result<(), String> {
+    if !target_dir.exists() {
+        fs::create_dir_all(target_dir).unwrap();
```
bobbinth, line 9: Should we return an error here explaning that we failed to create the directory? Same goes for some `unwraps()` later in this file.

crates/miden-rpc-proto/README.md
igamigo: Not a nit at all haha , fixed.

crates/miden-rpc-proto/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+This crate contains the RPC protobuf message definitions.
```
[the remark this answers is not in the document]
igamigo: Changed

crates/miden-rpc-proto/src/lib.rs
igamigo: Yeah, was planning on removing these. Done!

igamigo:
PR implementing this on the client: https://github.com/0xPolygonMiden/miden-client/pull/395

crates/miden-rpc-proto/README.md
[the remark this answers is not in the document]
igamigo: Although in its current state, this PR adds all message definitions, not just the RPC-related ones.

crates/rpc-proto/src/lib.rs
bobbinth, line 1: Should we make the crate `no_std` compatible and have a default feature which enabled `std` (like we do in other crates)? `write_proto()` function than can be gated by the `std` feature.

igamigo: Done

mFragaBA: If the crate is only used on build scripts and not on runtime you don't need to make it no_std-compatible. You can just add the dependency as a [build-dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html?highlight=build-dependencies#build-dependencies)

crates/rpc-proto/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+default = ["std"]
+std = []
+
+[dependencies]
+hex = { version = "0.4" }
+miden-node-utils = { workspace = true }
+miden-objects = { workspace = true }
+prost = { version = "0.12" }
+thiserror = { workspace = true }
+tonic = { workspace = true }
```
bobbinth, line 24: Do we need all these dependencies here?

crates/rpc-proto/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "miden-rpc-proto"
+version = "0.3.0"
+description = "Miden RPC message definitions"
```
bobbinth, line 4: nit: I would change the description to "Miden node RPC message definitions"

mFragaBA (approved):
LGTM! I was able to build this without default features. I wasn't able to do so for `wasm32_unknown_unknown` target though, but it shouldn't be relevant since we're only using the crate in build scripts.
 
 I like the idea of exposing the plain protobuf files, but I also think clients will end up copy-pasting the code to generate the files from the `.proto` files. Couldn't we expose both the proto files as well as the generated files?
 
 Another thing that came up to mind is whether we'd be able to, from miden-node-proto, generate the files without the `ApiClient` stuff (I think just commenting:
 
 ```rust
     // Compile the proto file for all servers APIs
    let protos = &[
        proto_dir.join("block_producer.proto"),
        proto_dir.join("store.proto"),
        proto_dir.join("rpc.proto"),
    ];
 ```
 
 from `proto/build.rs` based on a flag might work), and if those files end up being no_std compatible.
 
 We can iterate on the approach though and I think exposing the proto files is something we just want anyways.

crates/rpc-proto/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+readme = "README.md"
+keywords = ["miden", "node", "protobuf", "rpc"]
```
bobbinth, line 6: In addition to `keywords`, let's add `categories` field - it could look something like this:
```toml
categories = ["no-std"]
```
(not sure if any other categories would be applicable)

crates/rpc-proto/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+prost = { version = "0.12" }
+thiserror = { workspace = true }
+tonic = { workspace = true }
+
+[build-dependencies]
+miette = { version = "7.0", features = ["fancy"] }
+prost = { version = "0.12" }
+prost-build = { version = "0.12" }
+protox = { version = "0.6" }
+tonic-build = { version = "0.11" }
```
bobbinth, line 31: Do we also need all of these build dependencies?

crates/rpc-proto/src/lib.rs
bobbinth: Yeah - in most cases, I think that's how it'll work. I wasn't sure if that would be relevant in all cases though (e.g., maybe there is kind of just-in-time protobuf compilation that may use it as a runtime dependency).

bobbinth:
> I like the idea of exposing the plain protobuf files, but I also think clients will end up copy-pasting the code to generate the files from the `.proto` files. Couldn't we expose both the proto files as well as the generated files?

Wouldn't generated files need to assume the framework used for generation (e.g., `tonic` or something else)?

mFragaBA:
> > I like the idea of exposing the plain protobuf files, but I also think clients will end up copy-pasting the code to generate the files from the `.proto` files. Couldn't we expose both the proto files as well as the generated files?
> 
> Wouldn't generated files need to assume the framework used for generation (e.g., `tonic` or something else)?

Hmmm, I'm not sure. Looking at the generated files, the only ones that depend on having tonic are `rpc`, `store` and `block_producer`. The other ones only require having `prost`, which I think is wasm compatible, isn't it?

crates/rpc-proto/src/lib.rs
igamigo: That's how I was using it on the client as well. In any case it's very simple to keep it `no_std` compatible.

crates/rpc-proto/Cargo.toml
igamigo: Removed these

crates/rpc-proto/Cargo.toml
igamigo: Done

crates/rpc-proto/Cargo.toml
igamigo: Done

crates/rpc-proto/Cargo.toml
igamigo: Removed

igamigo:
> I like the idea of exposing the plain protobuf files, but I also think clients will end up copy-pasting the code to generate the files from the `.proto` files. Couldn't we expose both the proto files as well as the generated files?
> 
> Another thing that came up to mind is whether we'd be able to, from miden-node-proto, generate the files without the `ApiClient` stuff (I think just commenting:
> 
> ```rust
>     // Compile the proto file for all servers APIs
>    let protos = &[
>        proto_dir.join("block_producer.proto"),
>        proto_dir.join("store.proto"),
>        proto_dir.join("rpc.proto"),
>    ];
> ```
> 
> from `proto/build.rs` based on a flag might work), and if those files end up being no_std compatible.

I'm not sure that works, I did a quick test and didn't seem to. Additionally I'm not sure how valuable those created types would be without being wired to some sort of an API client (or similar) and as Bobbin mentioned that probably makes or breaks the possibility of consuming the API properly.

I think eventually we might want to make the API client and independent component (both from the node and `miden-client`) but as it stands, keeping it in the client can be a good first step.

crates/rpc-proto/src/lib.rs
bobbinth, line 1: We should add `#![no_std]` at the top of this file.

bobbinth (approved):
Looks good! Thank you! I left one last comment inline. After this, we can merge.

crates/rpc-proto/src/lib.rs
```diff
@@ -0,0 +1,31 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
bobbinth, line 1: Do we need to gate it by `std` feature? In other crates we just do `#![no_std]` and it seems to work.

igamigo: If you link `alloc` and use that, then it'd work. Here we can replace `String` and `format!` with their `alloc` counterparts, but not things like `fs`/`Path` so it would fail to compile correctly by default (since it would include the `std` feature).

bobbinth: Ah - I see. Another way we should be able to do it is:
```rust
#![no_std]

#[cfg(feature = "std")]
extern crate std;
```
This is similar to how it is done in `miden-vm` [assembler](https://github.com/0xPolygonMiden/miden-vm/blob/main/assembly/src/lib.rs).

igamigo: This makes the function a bit more verbose with the imports since it seems the prelude is not loaded automatically, but it works fine.

bobbinth: I think that's OK.

bobbinth (approved):
All looks good! Thank you!

bobbinth:
oh - i forgot: let's also update the changelog in a small follow-up PR.

PR #394, opened 2024-06-26
chore: replace println with tracing logs

### LINKED ISSUE #165: Add logging to `make_genesis` command
The `make_genesis` command uses `println!()` for now to display information to the user. 

It would be better if we would change that by logging now that it has been setup.

h3lio5:
fixes #165


mFragaBA (approved):
LGTM! @bobbinth do we need a second approve to merge?

bobbinth (approved):
Looks good! Thank you for tackling these!

PR #396, opened 2024-07-01
Addressed review comments for #390

polydez:
This PR is created in order to address remaining review comments in https://github.com/0xPolygonMiden/miden-node/pull/390. We had to merge that PR in order to make miden-note compilable against the latest miden-base.

Here we remove in-memory notes map in favor of direct queries to database. Output notes consumed in the same batch now removed from output notes list and SMT, corresponding input notes also removed from a batch and don't affect produced nullifiers set. Input unauthenticated notes are checked for duplicates as well, as output notes. Hashes of all corresponding output/input notes are required to be equal.

Resolves: https://github.com/0xPolygonMiden/miden-node/issues/381

### DIFF
--- a/crates/block-producer/src/batch_builder/batch.rs
+++ b/crates/block-producer/src/batch_builder/batch.rs
@@ -6,7 +6,7 @@ use miden_objects::{
     block::BlockAccountUpdate,
     crypto::hash::blake::{Blake3Digest, Blake3_256},
     notes::{NoteId, Nullifier},
-    transaction::{OutputNote, TransactionId, TxAccountUpdate},
+    transaction::{InputNoteCommitment, OutputNote, TransactionId, TxAccountUpdate},
     Digest, MAX_NOTES_PER_BATCH,
 };
 use tracing::instrument;
@@ -26,8 +26,7 @@ pub type BatchId = Blake3Digest<32>;
 pub struct TransactionBatch {
     id: BatchId,
     updated_accounts: Vec<(TransactionId, TxAccountUpdate)>,
-    unauthenticated_input_notes: BTreeSet<NoteId>,
-    produced_nullifiers: Vec<Nullifier>,
+    input_notes: Vec<InputNoteCommitment>,
     output_notes_smt: BatchNoteTree,
     output_notes: BTreeMap<NoteId, OutputNote>,
 }
@@ -80,7 +79,7 @@ impl TransactionBatch {
         // This still allows transaction `A` to consume an unauthenticated note `x` and output note `y`
         // and for transaction `B` to consume an unauthenticated note `y` and output note `x`
         // (i.e., have a circular dependency between transactions), but this is not a problem.
-        let mut produced_nullifiers = vec![];
+        let mut input_notes = vec![];
         for input_note in txs.iter().flat_map(|tx| tx.input_notes().iter()) {
             // Header is presented only for unauthenticated notes.
             if let Some(input_note_header) = input_note.header() {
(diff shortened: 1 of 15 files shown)


crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -74,14 +78,24 @@ impl TransactionBatch {
...
+                let consumed_in_place = unauthenticated_input_notes.remove(&note.id());
+                if !consumed_in_place {
+                    if let Some(nullifier) = candidate_nullifiers.remove(&note.id()) {
+                        produced_nullifiers.push(nullifier);
+                    }
+                }
+                !consumed_in_place
+            })
             .cloned()
             .collect();
```
bobbinth, line 100: I think this logic should works, but I wonder if a simpler way to do this would be to reverse the order. Specifically:

1. First, instead of iterating over input notes (as is done on lines 61 - 74 above), we iterate over output notes and build a map `BTreeMap<NoteId, OutputNote>`.
2. Then, we iterate over input notes to build a list of input notes for the batch. At the same time, we try to remove any unauthenticated note from the `BTreeMap`.
  a. Here, we can also perform the check that hashes of input/output notes match.
3. Lastly, the remaining values of the `BTeeMap` become the output notes of the batch.

bobbinth (commented):
Looks good! Thank you. I left a few comments inline.

crates/store/src/state.rs
```diff
@@ -527,13 +529,14 @@ impl State {
             })
             .collect();
 
-        let missing_unauthenticated_notes = filter_found_notes(unauthenticated_notes, &inner.notes);
+        let missing_unauthenticated_notes =
+            select_note_ids(&self.db, unauthenticated_notes).await?;
```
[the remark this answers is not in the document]
polydez: Good catch, thank you!

bobbinth (changes requested):
Thank you! Looks good. I left some more comments inline.

crates/block-producer/src/test_utils/block.rs
```diff
@@ -169,6 +169,7 @@ pub(crate) fn note_created_smt_from_note_batches<'a>(
     BlockNoteTree::with_entries(note_leaf_iterator).unwrap()
 }
 
+#[cfg(test)]
```
bobbinth, line 172: Why is this needed here? I thought this entire module is enabled only for tests.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -134,14 +182,14 @@ impl TransactionBatch {
         })
     }
 
-    /// Returns unauthenticated input notes set consumed by the transactions in this batch.
-    pub fn unauthenticated_input_notes(&self) -> &BTreeSet<NoteId> {
-        &self.unauthenticated_input_notes
+    /// Returns input notes list consumed by the transactions in this batch.
+    pub fn input_notes(&self) -> &[InputNoteCommitment] {
```
bobbinth, line 186: I would also mention in the comment that any unauthenticated input notes which have matching output notes within this batch are not included in this list.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -26,80 +32,122 @@ pub type BatchId = Blake3Digest<32>;
...
+        txs: Vec<ProvenTransaction>,
+        found_unauthenticated_notes: Option<BTreeMap<NoteId, MerklePath>>,
+    ) -> Result<Self, BuildBatchError> {
         let id = Self::compute_id(&txs);
 
+        // Populate batch output notes and updated accounts.
         let mut updated_accounts = vec![];
-        let mut produced_nullifiers = vec![];
+        let mut output_notes = vec![];
+        let mut output_note_index = BTreeMap::new();
```
bobbinth, line 66: I think it could improve the readability of the code quite a bit if we combined these two variables into a dedicated struct - e.g., something like `OutputNoteTracker`. We could then implement various convenience methods on this struct. For example:
```rust
impl OutputNoteTracker {
    pub fn add_notes_from_tx(&mut self, tx: &ProvenTransaction) -> Result<(), BuildBatchError> {
         todo!()
    }

    pub fn remove_note(&mut self, note_header: &NoteHeader) -> Result<bool, BuildBatchError> {
        todo!()
    }

    pub fn into_notes(&self) -> Vec<OutputNote> {
        todo!()
    }
}
```

bobbinth (approved):
Looks good! Thank you! I left a couple of comments inline - the main one is for a refactoring which could improve readability of the code. Once these are addressed, we can merge.

Also, we need to add tests to make sure the unauthenticated note handling works as expected. Let's create an issue for this and add tests in a subsequent PR.

polydez:
> Also, we need to add tests to make sure the unauthenticated note handling works as expected. Let's create an issue for this and add tests in a subsequent PR.

Done: https://github.com/0xPolygonMiden/miden-node/issues/398

PR #400, opened 2024-07-08
chore: use workspace inheritance for crate versions

Mirko-von-Leipzig:
This removes the need to edit each crate's toml file when bumping the version.

This is under the assumption that these are always in lockstep.


bobbinth (approved):
Looks good! Thank you!

We do sometimes make patch releases of individual components (e.g., if we identify and fix a bug in a given component) - but so far this has been pretty rare and we can handle these on one-off basis.

PR #401, opened 2024-07-08
feat(node): normalize configuration

### LINKED ISSUE #334: [Feature]: Refactor configuration file for improved inference

End user

As of now in the configuration file of the miden-node file we have some dependencies between components:

```toml
[block_producer]
verify_tx_proofs = true

[rpc]
endpoint = { host = "0.0.0.0", port = 57291 }
block_producer_url = "http://localhost:48046"

[store]
```

`block_producer` and `rpc` have the url of the `store` for example.

This could lead to easy mistakes when typing out the configuration or by having wrong url's. 

A way to solve this would be to let the components infer the url of the other components from the configuration file as follows: 

```toml
[block_producer]
verify_tx_proofs = true

[rpc]
endpoint = { host = "0.0.0.0", port = 57291 }

[store]
```

### How is this feature used?

This will enable to have slimmer and more readable configuration files, it will also reduce the amount of possible errors.

### Anything else?

Initial comment here: https://github.com/0xPolygonMiden/miden-node/pull/323#discussion_r1574248236

Mirko-von-Leipzig:
Simplifies node configuration by removing all redundant options.

The following options are removed 

- block_producer.store_url
- rpc.store_url
- rpc.block_producer_url

as they are already covered by `store.endpoint` and `block_producer.endpoint`.

Note that this only impacts the configuration options as seen by the caller of the node binary. Individual node components (rpc, store, block_producer) still have the original configuration objects. This PR only strips out the duplicate options for the user.

In addition, this PR removes the optional indirection in `NodeConfig`. ~Maybe this is incorrect, and the `figment` crate somehow requires this to handle missing options properly? I don't think so; but maybe there was a reason for this beyond old changes :)~

Closes #334.

------------------------------------

Update after this PR was reworked (the above all still holds true).

Configuration of the various `miden-node start xxx` options are now handled separately. This gives us better type safety without any options.

In addition, we now disallow unknown configuration options. I think this is safer as otherwise users are left wondering why their inputs are doing nothing.

`load_config` is now also generic which makes it more ergonomic imo.

### DIFF
--- a/bin/node/src/config.rs
+++ b/bin/node/src/config.rs
@@ -23,8 +23,6 @@ impl Default for NodeConfig {
 
 #[cfg(test)]
 mod tests {
-    use std::path::PathBuf;
-
     use figment::Jail;
     use miden_node_block_producer::config::BlockProducerConfig;
     use miden_node_rpc::config::RpcConfig;
@@ -58,8 +56,7 @@ mod tests {
                 "#,
             )?;
 
-            let config: NodeConfig =
-                load_config(PathBuf::from(NODE_CONFIG_FILE_PATH).as_path()).extract()?;
+            let config: NodeConfig = load_config(NODE_CONFIG_FILE_PATH)?;
 
             assert_eq!(
                 config,
--- a/bin/node/src/config.rs
+++ b/bin/node/src/config.rs
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 
 /// Node top-level configuration.
 #[derive(Clone, Default, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
 pub struct NodeConfig {
     block_producer: NormalizedBlockProducerConfig,
     rpc: NormalizedRpcConfig,
@@ -14,12 +15,14 @@ pub struct NodeConfig {
 
 /// A specialized variant of [RpcConfig] with redundant fields within [NodeConfig] removed.
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct NormalizedRpcConfig {
     endpoint: Endpoint,
 }
 
 /// A specialized variant of [BlockProducerConfig] with redundant fields within [NodeConfig] removed.
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct NormalizedBlockProducerConfig {
     endpoint: Endpoint,
     verify_tx_proofs: bool,
--- a/bin/node/src/config.rs
+++ b/bin/node/src/config.rs
@@ -1,52 +1,95 @@
 use miden_node_block_producer::config::BlockProducerConfig;
 use miden_node_rpc::config::RpcConfig;
 use miden_node_store::config::StoreConfig;
+use miden_node_utils::config::Endpoint;
 use serde::{Deserialize, Serialize};
 
 /// Node top-level configuration.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+#[derive(Clone, Default, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
 pub struct NodeConfig {
-    pub block_producer: Option<BlockProducerConfig>,
-    pub rpc: Option<RpcConfig>,
-    pub store: Option<StoreConfig>,
+    block_producer: NormalizedBlockProducerConfig,
+    rpc: NormalizedRpcConfig,
+    store: StoreConfig,
 }
 
-impl Default for NodeConfig {
+/// A specialized variant of [RpcConfig] with redundant fields within [NodeConfig] removed.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+struct NormalizedRpcConfig {
+    endpoint: Endpoint,
+}
+
+/// A specialized variant of [BlockProducerConfig] with redundant fields within [NodeConfig] removed.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+struct NormalizedBlockProducerConfig {
+    endpoint: Endpoint,
+    verify_tx_proofs: bool,
+}
+
+impl Default for NormalizedRpcConfig {
     fn default() -> Self {
-        Self {
-            block_producer: Some(Default::default()),
-            rpc: Some(Default::default()),
-            store: Some(Default::default()),
-        }
+        // Ensure we stay in sync with the original defaults.
+        let RpcConfig {
+            endpoint,
+            store_url: _,
+            block_producer_url: _,
+        } = RpcConfig::default();
+        Self { endpoint }
+    }
+}
+
+impl Default for NormalizedBlockProducerConfig {
+    fn default() -> Self {
+        // Ensure we stay in sync with the original defaults.
+        let BlockProducerConfig { endpoint, store_url: _, verify_tx_proofs } =
+            BlockProducerConfig::default();
+        Self { endpoint, verify_tx_proofs }
+    }
+}
+
+impl NodeConfig {
+    pub fn into_parts(self) -> (BlockProducerConfig, RpcConfig, StoreConfig) {
+        let Self { block_producer, rpc, store } = self;
+
+        let block_producer = BlockProducerConfig {
+            endpoint: block_producer.endpoint,
+            store_url: store.endpoint_url(),
+            verify_tx_proofs: block_producer.verify_tx_proofs,
+        };
+
+        let rpc = RpcConfig {
+            endpoint: rpc.endpoint,
+            store_url: store.endpoint_url(),
+            block_producer_url: block_producer.endpoint_url(),
+        };
+
+        (block_producer, rpc, store)
     }
 }
 
 #[cfg(test)]
 mod tests {
     use figment::Jail;
-    use miden_node_block_producer::config::BlockProducerConfig;
-    use miden_node_rpc::config::RpcConfig;
     use miden_node_store::config::StoreConfig;
     use miden_node_utils::config::{load_config, Endpoint};
 
     use super::NodeConfig;
-    use crate::NODE_CONFIG_FILE_PATH;
+    use crate::{
+        config::{NormalizedBlockProducerConfig, NormalizedRpcConfig},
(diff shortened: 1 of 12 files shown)


bin/node/src/config.rs
```diff
@@ -41,54 +77,47 @@ mod tests {
...
-                        },
-                        store_url: "http://store:8000".to_string(),
-                        block_producer_url: "http://block_producer:8001".to_string(),
-                    }),
-                    store: Some(StoreConfig {
+                    },
+                    rpc: RpcConfig {
+                        endpoint: Endpoint { host: "rpc_host".to_string(), port: 4321 },
+                        store_url: "http://store_host:1234".to_string(),
+                        block_producer_url: "http://block_host:1133".to_string(),
```
Mirko-von-Leipzig, line 110: I'm also curious - is our assumption that these are always `http`?

bin/node/src/config.rs
```diff
@@ -1,35 +1,71 @@
...
 }
 
-impl Default for NodeConfig {
-    fn default() -> Self {
-        Self {
-            block_producer: Some(Default::default()),
-            rpc: Some(Default::default()),
-            store: Some(Default::default()),
+impl NodeConfig {
+    pub fn load(config_file: impl AsRef<Path>) -> figment::Result<Self> {
```
Mirko-von-Leipzig, line 18: The `figment::Result` is required to properly test using `figment::jail` which is a pity as this adds `figment` to our dependencies.

This is probably fine though - I imagination a binary should contain configuration code :). Alternatively we create some better abstractions or hide `figment` somehow, but I don't really think that's worth the effort. lmk.

bin/node/src/config.rs
```diff
@@ -1,35 +1,71 @@
...
+        }
+
+        /// A normalized version of [NodeConfig] i.e. with all redundant options
+        /// removed. This helps prevent user configuration issues.
+        #[derive(Deserialize)]
+        struct UserConfig {
+            block_producer: UserBlockProducerConfig,
+            rpc: UserRpcConfig,
+            store: StoreConfig,
         }
```
bobbinth, line 39: nit: I would probably move these structs out of the function definition.

Also, question: how would this work when parts of the config file are missing? For example, say we want to run just the store component of the node (i.e., via `miden-node start store` command). Our config file may look like so:
```toml
[store]
database_filepath = "/opt/miden/miden-store.sqlite3"
genesis_filepath = "/opt/miden/genesis.dat"
blockstore_dir = "/opt/miden/blocks"
```
Similarly, if we wanted to run just the RPC component, the config file may look like so:
```toml
[rpc]
endpoint = { host = "0.0.0.0", port = 57291 }
block_producer_url = "http://block_producer_machine:48046"
store_url = "http://store_machine:28943"
```
I think the above would currently fail to parse, right?

I wonder if we should make the logic a bit more sophisticated. Specifically: if a dependency for a component is present, we try to infer the relevant fields from it or verify that all fields are consistent. To give a few examples (using only block producer and store components):

If the config file has just the `block_producer` component, it could look like so:
```toml
[block_producer]
verify_tx_proofs = true
```

If it has both `block_producer` and `store` components, it could look like so:
```toml
[block_producer]
verify_tx_proofs = true

[store]
database_filepath = "/opt/miden/miden-store.sqlite3"
genesis_filepath = "/opt/miden/genesis.dat"
blockstore_dir = "/opt/miden/blocks"
```
In the above, we infer `block_producer.store_url` from the `store.endpoint`. 

But the following would still be valid:
```toml
[block_producer]
verify_tx_proofs = true

[store]
database_filepath = "/opt/miden/miden-store.sqlite3"
genesis_filepath = "/opt/miden/genesis.dat"
blockstore_dir = "/opt/miden/blocks"
```
This is because `block_producer.store_url` is consistent with the `store.endpoint`.

However, if they were inconsistent (like in the below) - we should get an error:
```toml
[block_producer]
store_url = "http://localhost:12345"
verify_tx_proofs = true

[store]
database_filepath = "/opt/miden/miden-store.sqlite3"
genesis_filepath = "/opt/miden/genesis.dat"
blockstore_dir = "/opt/miden/blocks"
```

The above are just examples because having just 2 sections in a file would not be legal. In my mind, the legal combinations are:
- Either way have all 3 sections present (this is when the entire node runs in a single process), or
- Only one of the sections is present (this is when individual components run in separate processes).

bin/node/src/config.rs
```diff
@@ -1,35 +1,71 @@
...
+
+        let config =
+            miden_node_utils::config::load_config(config_file.as_ref()).extract::<UserConfig>()?;
```
bobbinth, line 42: Since figment is now a dependency - should we just call it directly (i.e., `Figment::from(Toml::file(config_file))`). In fact, i'm not sure our `load_config()` function is all that useful any more - so, maybe it can be removed altogether and we just use `NodeConfig::load()` everywhere?

bobbinth (commented):
Looks good! Thank you. I added a couple of comments inline. The main one is about potentially making the logic more robust to handle partial configuration files.

A couple of other comments:
- Let's update the changelog.
- We should also probably update the sample config files [here](https://github.com/0xPolygonMiden/miden-node/tree/main/config). For example, `miden-node.toml` would probably look something like this:

```toml
# This is an example configuration file for the Miden node.

[rpc]
endpoint = { host = "0.0.0.0", port = 57291 }

[block_producer]
# enables or disables the verification of transaction proofs before they are accepted into the
# transaction queue.
verify_tx_proofs = true

[store]
database_filepath = "/opt/miden/miden-store.sqlite3"
genesis_filepath = "/opt/miden/genesis.dat"
blockstore_dir = "/opt/miden/blocks"
```

bin/node/src/config.rs
bobbinth: hmmm - not sure. Probably, we should not assume this.

bin/node/src/config.rs
bobbinth: Yep - I think that's fine. I left a comment inline about potentially using `Figment` directly from the node rather than having a utility function for it in `miden-node-utils`.

bin/node/src/config.rs
Mirko-von-Leipzig: Ah I had assumed that running individual components used the individual configurations structs only i.e. that `miden-node start rpc` used `RpcConfig` and not `NodeConfig`.

I'll dig a little deeper into the code and come up with something more elegant :). Moving PR to draft for now.

bin/node/src/config.rs
Mirko-von-Leipzig: I thought maybe the `load_config` was isolated so that there would be a single place where we manipulate the configuration provider e.g. if we wanted to add env flag support then this would be the single place to add it.

I think probably its a good idea to encapsulate this at least somewhat if we have many config structs. I'll see how things look after fixing the component configurations.

bin/node/src/config.rs
bobbinth: One option is to change the config into an enum to look something like:
```rust
pub enum NodeConfig {
    Full(FullNodeConfig),
    Rpc(RpcConfig),
    BlockProducer(BlockProducerConfig),
    Store(StoreConfig)
}

pub struct FullNodeConfig {
    pub block_producer: BlockProducerConfig,
    pub rpc: RpcConfig,
    pub store: StoreConfig,
}
```
But there are probably other ways too.

bin/node/src/config.rs
Mirko-von-Leipzig: I've reworked this so that this is no longer necessary.

bin/node/src/config.rs
Mirko-von-Leipzig: I went a different route. I handle the different start "modes" configs as separate type so we have full type safety.

Mirko-von-Leipzig:
I think I've fixed and addressed the issues. I did do a rebase so the original commits were not really related to the current state.

bin/node/src/config.rs
```diff
@@ -1,54 +1,92 @@
...
-            rpc: Some(Default::default()),
-            store: Some(Default::default()),
+            endpoint: original.endpoint,
+            verify_tx_proofs: original.verify_tx_proofs,
         }
     }
 }
 
+impl NodeConfig {
+    pub fn into_components(self) -> (BlockProducerConfig, RpcConfig, StoreConfig) {
```
bobbinth, line 47: minor nit: we usually name such functions `into_parts()`.

bin/node/src/config.rs
```diff
@@ -1,54 +1,92 @@
...
+        // Ensure we stay in sync with the original defaults.
+        Self { endpoint: RpcConfig::default().endpoint }
+    }
 }
 
-impl Default for NodeConfig {
+impl Default for NormalizedBlockProducerConfig {
     fn default() -> Self {
+        // Ensure we stay in sync with the original defaults.
+        let original = BlockProducerConfig::default();
```
bobbinth, line 38: I wonder if we should destructure the result here - e.g., something like this:
```rust
let BlockProducerConfig { endpoint, verify_tx_proof, ... } = BlockProducerConfig::default();
```
This way, if `BlockProducerConfig` changes, we'd get a compile-time error here.

bin/node/src/config.rs
```diff
@@ -1,54 +1,92 @@
...
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+struct NormalizedBlockProducerConfig {
+    endpoint: Endpoint,
+    verify_tx_proofs: bool,
+}
+
+impl Default for NormalizedRpcConfig {
+    fn default() -> Self {
+        // Ensure we stay in sync with the original defaults.
+        Self { endpoint: RpcConfig::default().endpoint }
```
bobbinth, line 31: Similar comment as above re destructuring.

bin/node/src/config.rs
```diff
@@ -1,54 +1,92 @@
...
-    pub rpc: Option<RpcConfig>,
-    pub store: Option<StoreConfig>,
+    block_producer: NormalizedBlockProducerConfig,
+    rpc: NormalizedRpcConfig,
+    store: StoreConfig,
+}
+
+/// A specialized variant of [RpcConfig] with redundant fields within [NodeConfig] removed.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+struct NormalizedRpcConfig {
```
bobbinth, line 17: Should we also add `#[serde(deny_unknown_fields)]` here and to other config structs in this module?

bobbinth (approved):
Looks good! Thank you! I left a few small comments inline - after these are addressed, we can merge.

bin/node/src/config.rs
Mirko-von-Leipzig: Yes I somehow missed this, thanks.

Mirko-von-Leipzig:
I marked the changes as `[BREAKING]` in the changelog since configuration is public facing. Phrasing could maybe be better as well.

I made the other requested changes and rebased.

bobbinth (approved):
All looks good! Thank you!

PR #406, opened 2024-07-12
feat(block-producer): nullifier map type safety

Mirko-von-Leipzig:
Improve the type safety of `TransactionInputs::nullifiers` mapping by using `Option<NonZeroU32>` instead of relying on zero to indicate none.

Shower thought: is there utility in allowing a nullifier set to be included in the genesis block? Maybe for bootstrapping test networks..

(I'm happy to also just drop this PR, its not a major thing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [BREAKING] Configuration files with unknown properties are now rejected (#401).
 * [BREAKING] Removed redundant node configuration properties (#401).
+* Improve type safety of the transaction inputs nullifier mapping (#406).
 
 ## 0.4.0 (2024-07-04)
 
--- a/crates/block-producer/src/state_view/mod.rs
+++ b/crates/block-producer/src/state_view/mod.rs
@@ -252,7 +252,7 @@ fn ensure_tx_inputs_constraints(
     let infracting_nullifiers: Vec<Nullifier> = tx_inputs
         .nullifiers
         .into_iter()
-        .filter_map(|(nullifier_in_tx, block_num)| (block_num != 0).then_some(nullifier_in_tx))
+        .filter_map(|(nullifier_in_tx, block_num)| block_num.is_some().then_some(nullifier_in_tx))
         .collect();
 
     if !infracting_nullifiers.is_empty() {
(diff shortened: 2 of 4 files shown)


bobbinth (approved):
Looks good! Thank you! Let's update the changelog and merge.

> Shower thought: is there utility in allowing a nullifier set to be included in the genesis block? Maybe for bootstrapping test networks..

Hmmm - I haven't thought about it much but doing something like that might be tricky. The main reason is that when we build the nullifier SMT the default value for missing nullifiers is 0. This works because we assume that no nullifier can exist in the genesis block (block 0). But if in fact some nullifiers can exist in the genesis block, we'll either need to use a different default value for missing nullifiers, or do something else to distinguish missing nullifiers from the ones that were created in the genesis block.

Mirko-von-Leipzig:
> Looks good! Thank you! Let's update the changelog and merge.

Done. I don't think I'm seeing the correct merge option - I assume we're doing fast-foward merges? However, I'm only seeing merge with merge commit or squash + merge.

bobbinth:
> I don't think I'm seeing the correct merge option - I assume we're doing fast-foward merges? However, I'm only seeing merge with merge commit or squash + merge.

In most cases, we use the "squash and merge" option. But this is also something to be discussed in the styles and processes discussion I mentioned earlier today.

bobbinth:
Oh - another thing, we usually create branches directly in the repos (rather than in forked repos).

Mirko-von-Leipzig:
> Oh - another thing, we usually create branches directly in the repos (rather than in forked repos).

The [`CONTRIBUTING.md`](https://github.com/0xPolygonMiden/miden-node/blob/main/CONTRIBUTING.md#flow) said to use forked repos; but I definitely prefer not dealing with origin/upstream remotes so that's good news

bobbinth:
> > Oh - another thing, we usually create branches directly in the repos (rather than in forked repos).
> 
> The [`CONTRIBUTING.md`](https://github.com/0xPolygonMiden/miden-node/blob/main/CONTRIBUTING.md#flow) said to use forked repos; but I definitely prefer not dealing with origin/upstream remotes so that's good news

Yep - that's the flow for external contributors. But for the core team, we can create branches in the repos directly.

PR #407, opened 2024-07-12
feat(block-producer): multiple inflight account txs

### LINKED ISSUE #186: Allow multiple in-flight transactions for an account.
Currently for simplicity only one transaction can be in-flight per account.

Mirko-von-Leipzig:
Support multiple inflight transactions affecting the same account by tracking the latest inflight state of modified accounts. This is then checked against new txns.

This is likely more a progress check than a final PR - I suspect there are more indirect changes that must be made?

Probably missing: tests ensuring that secondary inflight transactions are rejected if they have the wrong initial state.

Closes #186

### DIFF
--- a/crates/block-producer/src/batch_builder/batch.rs
+++ b/crates/block-producer/src/batch_builder/batch.rs
@@ -57,6 +57,11 @@ impl AccountUpdate {
 
     /// Merges the transaction's update into this account update.
     fn merge_tx(&mut self, tx: &ProvenTransaction) {
+        assert!(
+            self.final_state == tx.account_update().init_state_hash(),
+            "Transacion's initial state does not match current account state"
+        );
+
         self.final_state = tx.account_update().final_state_hash();
         self.transactions.push(tx.id());
 
@@ -170,15 +175,15 @@ impl TransactionBatch {
     pub fn account_initial_states(&self) -> impl Iterator<Item = (AccountId, Digest)> + '_ {
         self.updated_accounts
             .iter()
-            .map(|(&account, update)| (account, update.init_state))
+            .map(|(&account_id, update)| (account_id, update.init_state))
     }
 
     /// Returns an iterator over (account_id, details, new_state_hash) tuples for accounts that were
     /// modified in this transaction batch.
     pub fn updated_accounts(&self) -> impl Iterator<Item = BlockAccountUpdate> + '_ {
-        self.updated_accounts.iter().map(|(&account, update)| {
+        self.updated_accounts.iter().map(|(&account_id, update)| {
             BlockAccountUpdate::new(
-                account,
+                account_id,
                 update.final_state,
                 update.details.clone(),
                 update.transactions.clone(),
--- a/crates/block-producer/src/batch_builder/batch.rs
+++ b/crates/block-producer/src/batch_builder/batch.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeMap, BTreeSet},
+    collections::{btree_map::Entry, BTreeMap, BTreeSet},
     mem,
 };
 
@@ -13,7 +13,7 @@ use miden_objects::{
     },
     notes::{NoteHeader, NoteId, Nullifier},
     transaction::{InputNoteCommitment, OutputNote, TransactionId},
-    Digest, MAX_NOTES_PER_BATCH,
+    AccountDeltaError, Digest, MAX_NOTES_PER_BATCH,
 };
 use tracing::instrument;
 
@@ -56,7 +56,7 @@ impl AccountUpdate {
     }
 
     /// Merges the transaction's update into this account update.
-    fn merge_tx(&mut self, tx: &ProvenTransaction) {
+    fn merge_tx(&mut self, tx: &ProvenTransaction) -> Result<(), AccountDeltaError> {
         assert!(
             self.final_state == tx.account_update().init_state_hash(),
             "Transacion's initial state does not match current account state"
@@ -64,8 +64,9 @@ impl AccountUpdate {
 
         self.final_state = tx.account_update().final_state_hash();
         self.transactions.push(tx.id());
+        self.details = self.details.clone().merge(tx.account_update().details().clone())?;
 
-        // TODO: merge account update details. Waiting on implementation in.. miden-objects.
+        Ok(())
     }
 }
 
@@ -98,10 +99,18 @@ impl TransactionBatch {
         let mut unauthenticated_input_notes = BTreeSet::new();
         for tx in &txs {
             // Merge account updates so that state transitions A->B->C become A->C.
-            updated_accounts
-                .entry(tx.account_id())
-                .and_modify(|x| x.merge_tx(tx))
-                .or_insert(AccountUpdate::new(tx));
+            match updated_accounts.entry(tx.account_id()) {
+                Entry::Vacant(vacant) => {
+                    vacant.insert(AccountUpdate::new(tx));
+                },
+                Entry::Occupied(occupied) => occupied.into_mut().merge_tx(tx).map_err(|error| {
+                    BuildBatchError::AccountUpdateError {
+                        account_id: tx.account_id(),
+                        error,
+                        txs: txs.clone(),
+                    }
+                })?,
+            };
 
             // Check unauthenticated input notes for duplicates:
             for note in tx.get_unauthenticated_notes() {
--- a/crates/block-producer/src/batch_builder/batch.rs
+++ b/crates/block-producer/src/batch_builder/batch.rs
@@ -6,7 +6,6 @@ use std::{
 use miden_objects::{
     accounts::{delta::AccountUpdateDetails, AccountId},
     batches::BatchNoteTree,
-    block::BlockAccountUpdate,
     crypto::{
         hash::blake::{Blake3Digest, Blake3_256},
         merkle::MerklePath,
@@ -38,11 +37,11 @@ pub struct TransactionBatch {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-struct AccountUpdate {
-    init_state: Digest,
-    final_state: Digest,
-    transactions: Vec<TransactionId>,
-    details: AccountUpdateDetails,
+pub struct AccountUpdate {
+    pub init_state: Digest,
+    pub final_state: Digest,
+    pub transactions: Vec<TransactionId>,
+    pub details: AccountUpdateDetails,
 }
 
 impl AccountUpdate {
@@ -189,15 +188,8 @@ impl TransactionBatch {
 
     /// Returns an iterator over (account_id, details, new_state_hash) tuples for accounts that were
     /// modified in this transaction batch.
-    pub fn updated_accounts(&self) -> impl Iterator<Item = BlockAccountUpdate> + '_ {
-        self.updated_accounts.iter().map(|(&account_id, update)| {
-            BlockAccountUpdate::new(
-                account_id,
-                update.final_state,
-                update.details.clone(),
-                update.transactions.clone(),
-            )
-        })
+    pub fn updated_accounts(&self) -> impl Iterator<Item = (&AccountId, &AccountUpdate)> + '_ {
+        self.updated_accounts.iter()
     }
 
     /// Returns input notes list consumed by the transactions in this batch. Any unauthenticated
--- a/crates/block-producer/src/batch_builder/batch.rs
+++ b/crates/block-producer/src/batch_builder/batch.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use miden_objects::{
-    accounts::AccountId,
+    accounts::{delta::AccountUpdateDetails, AccountId},
     batches::BatchNoteTree,
     block::BlockAccountUpdate,
     crypto::{
@@ -12,7 +12,7 @@ use miden_objects::{
         merkle::MerklePath,
     },
     notes::{NoteHeader, NoteId, Nullifier},
-    transaction::{InputNoteCommitment, OutputNote, TransactionId, TxAccountUpdate},
+    transaction::{InputNoteCommitment, OutputNote, TransactionId},
     Digest, MAX_NOTES_PER_BATCH,
 };
 use tracing::instrument;
@@ -31,10 +31,37 @@ pub type BatchId = Blake3Digest<32>;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct TransactionBatch {
     id: BatchId,
-    updated_accounts: Vec<(TransactionId, TxAccountUpdate)>,
+    updated_accounts: BTreeMap<AccountId, AccountUpdate>,
     input_notes: Vec<InputNoteCommitment>,
     output_notes_smt: BatchNoteTree,
     output_notes: Vec<OutputNote>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct AccountUpdate {
+    init_state: Digest,
+    final_state: Digest,
+    transactions: Vec<TransactionId>,
+    details: AccountUpdateDetails,
+}
+
+impl AccountUpdate {
+    fn new(tx: &ProvenTransaction) -> Self {
+        Self {
+            init_state: tx.account_update().init_state_hash(),
+            final_state: tx.account_update().final_state_hash(),
+            transactions: vec![tx.id()],
+            details: tx.account_update().details().clone(),
+        }
+    }
+
+    /// Merges the transaction's update into this account update.
+    fn merge_tx(&mut self, tx: &ProvenTransaction) {
+        self.final_state = tx.account_update().final_state_hash();
+        self.transactions.push(tx.id());
+
+        // TODO: merge account update details. Waiting on implementation in.. miden-objects.
+    }
 }
 
 impl TransactionBatch {
@@ -62,14 +89,15 @@ impl TransactionBatch {
 
         // Populate batch output notes and updated accounts.
         let mut output_notes = OutputNoteTracker::new(&txs)?;
-        let mut updated_accounts = vec![];
+        let mut updated_accounts = BTreeMap::<AccountId, AccountUpdate>::new();
         let mut unauthenticated_input_notes = BTreeSet::new();
         for tx in &txs {
-            // TODO: we need to handle a possibility that a batch contains multiple transactions against
-            //       the same account (e.g., transaction `x` takes account from state `A` to `B` and
-            //       transaction `y` takes account from state `B` to `C`). These will need to be merged
-            //       into a single "update" `A` to `C`.
-            updated_accounts.push((tx.id(), tx.account_update().clone()));
+            // Merge account updates so that state transitions A->B->C become A->C.
+            updated_accounts
+                .entry(tx.account_id())
+                .and_modify(|x| x.merge_tx(tx))
+                .or_insert(AccountUpdate::new(tx));
+
             // Check unauthenticated input notes for duplicates:
             for note in tx.get_unauthenticated_notes() {
                 let id = note.id();
@@ -142,18 +170,18 @@ impl TransactionBatch {
     pub fn account_initial_states(&self) -> impl Iterator<Item = (AccountId, Digest)> + '_ {
         self.updated_accounts
             .iter()
-            .map(|(_, update)| (update.account_id(), update.init_state_hash()))
+            .map(|(&account, update)| (account, update.init_state))
     }
 
     /// Returns an iterator over (account_id, details, new_state_hash) tuples for accounts that were
     /// modified in this transaction batch.
     pub fn updated_accounts(&self) -> impl Iterator<Item = BlockAccountUpdate> + '_ {
-        self.updated_accounts.iter().map(|(transaction_id, update)| {
+        self.updated_accounts.iter().map(|(&account, update)| {
             BlockAccountUpdate::new(
-                update.account_id(),
-                update.final_state_hash(),
-                update.details().clone(),
-                vec![*transaction_id],
+                account,
+                update.final_state,
+                update.details.clone(),
+                update.transactions.clone(),
             )
         })
     }
--- a/crates/block-producer/src/block_builder/prover/block_witness.rs
+++ b/crates/block-producer/src/block_builder/prover/block_witness.rs
@@ -1,7 +1,8 @@
 use std::collections::{BTreeMap, BTreeSet};
 
 use miden_objects::{
-    accounts::AccountId,
+    accounts::{delta::AccountUpdateDetails, AccountId},
+    block::BlockAccountUpdate,
     crypto::merkle::{EmptySubtreeRoots, MerklePath, MerkleStore, MmrPeaks, SmtProof},
     notes::Nullifier,
     transaction::TransactionId,
@@ -34,7 +35,7 @@ impl BlockWitness {
     pub fn new(
         mut block_inputs: BlockInputs,
         batches: &[TransactionBatch],
-    ) -> Result<Self, BuildBlockError> {
+    ) -> Result<(Self, Vec<BlockAccountUpdate>), BuildBlockError> {
         if batches.len() > MAX_BATCHES_PER_BLOCK {
             return Err(BuildBlockError::TooManyBatchesInBlock(batches.len()));
         }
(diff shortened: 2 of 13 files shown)


crates/block-producer/src/state_view/mod.rs
```diff
@@ -92,7 +95,12 @@ where
         // identify a set of unauthenticated input notes which are not in the store yet; we'll use
         // this set to verify that these notes are currently in flight (i.e., they are output notes
         // of one of the inflight transactions)
-        let tx_inputs = self.store.get_tx_inputs(candidate_tx).await?;
+        let mut tx_inputs = self.store.get_tx_inputs(candidate_tx).await?;
+        if let Some(inflight) = self.accounts_in_flight.read().await.get(&candidate_tx.account_id())
+        {
+            // Merge the inflight account state into the store state.
+            tx_inputs.account_hash = Some(*inflight);
+        }
```
Mirko-von-Leipzig, line 103: This feels a little hacky, but maybe just a better comment explaining why we are aggregating these is good enough?

crates/block-producer/src/state_view/tests/verify_tx.rs
Mirko-von-Leipzig, line 1: Where can I find descriptions of what these test cases are referring to? i.e. where are the requirements defined?

crates/block-producer/src/errors.rs
```diff
@@ -31,10 +31,10 @@ pub enum VerifyTxError {
     UnauthenticatedNotesNotFound(Vec<NoteId>),
 
     /// The account's initial hash did not match the current account's hash
-    #[error("Incorrect account's initial hash ({tx_initial_account_hash}, stored: {})", format_opt(.store_account_hash.as_ref()))]
+    #[error("Incorrect account's initial hash ({tx_initial_account_hash}, actual: {})", format_opt(.actual_account_hash.as_ref()))]
     IncorrectAccountInitialHash {
         tx_initial_account_hash: Digest,
-        store_account_hash: Option<Digest>,
+        actual_account_hash: Option<Digest>,
```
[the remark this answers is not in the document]
bobbinth: I probably wouldn't call it `actual_account_hash`, but also `onchain` does not feel like the right term is the state technically may not be on chain yet. Maybe something like `latest_known_account_hash` but that's a bit long.

crates/block-producer/src/state_view/mod.rs
```diff
@@ -141,7 +152,7 @@ where
 
         // Remove account ids of transactions in block
         for update in block.updated_accounts() {
-            let was_in_flight = locked_accounts_in_flight.remove(&update.account_id());
+            let was_in_flight = locked_accounts_in_flight.remove(&update.account_id()).is_some();
```
bobbinth, line 155: Is this modification sufficient? Let's say we have two in-flight transactions agains the same account. Let's say the account ID is `a`, the first transaction changes account state from `X` to `Y`, and the second transaction changes account state from `Y` to `Z`. So, the entry in the in-flight accounts map would be `(a, Z)`.

Now, let's say we produce a block which includes only the first transaction. If I'm understanding the code correctly, the entry `(a, Z)` would be removed from the map. So, we'll end up in a situation when the second transaction is still in flight, but there is no entry for the corresponding account in the in-flight accounts map.

crates/block-producer/src/state_view/mod.rs
bobbinth: I think this should be OK as long as we document it here and also in `ensure_tx_inputs_constraints()` as there we no longer check just against the store but more generally against the latest known account state.

As a side note (and probably not for this PR): I find the name `TransactionInputs` really confusing as what we are getting are not really inputs into the transaction but various elements needed to verify transaction validity. I don't have a better suggestion though.

crates/block-producer/src/state_view/mod.rs
```diff
@@ -31,7 +34,7 @@ pub struct DefaultStateView<S> {
 
     /// The account ID of accounts being modified by transactions currently in the block production
     /// pipeline. We currently ensure that only 1 tx/block modifies any given account (issue: #186).
-    accounts_in_flight: Arc<RwLock<BTreeSet<AccountId>>>,
+    accounts_in_flight: Arc<RwLock<BTreeMap<AccountId, Digest>>>,
```
bobbinth, line 37: nit: the doc comment here needs to be updated.

But also, I'm wondering if we should introduce a wrapper struct around the `BTreeMap`. Maybe something like `AccountStateTracker` (though, probably a better name is needed). For now, it could look something like this:

```rust
pub struct AccountStateTracker(BTreeMap<AccountId, Digest>);

impl AccountStateTracker {
    pub fn add(id: AccountId, init_state: Digest, final_state: Digest) -> Result<(), SomeError> {
        // if init_state is inconsistent with what is already being tracked, return an error
        todo!("implement")
    }

    pub fn remove(id: AccountId, final_state: Digest) -> Result<(), SomeError> {
        // removes the entry for id from the internal map, only if final_state matches the
        // corresponding value
        todo!("implement")
    }

    // could also have other convenience methods here - e.g., for checking if an account update
    // is in flight
}
```
We could also use a slightly more sophisticated approach internally and define the tracker as:
```rust
pub struct AccountStateTracker(BTreeMap<AccountId, Vec<Digest>>);
```
Here, the vector of digests would keep a chain of updates currently in-flight. Not strictly necessary, but I think it will make checking for inconsistencies easier.

bobbinth (commented):
Thank you! Looks good! I left a few comments inline.

We also need to make sure that the processes of combining transactions into a batch and combining batches into a block work correctly (e.g., see [here](https://github.com/0xPolygonMiden/miden-node/blob/next/crates/block-producer/src/batch_builder/batch.rs#L68) and [here](https://github.com/0xPolygonMiden/miden-node/blob/next/crates/block-producer/src/block_builder/prover/block_witness.rs#L24)).

crates/block-producer/src/state_view/mod.rs
Mirko-von-Leipzig: Maybe something something `TransactionState`-like? Its sort of the initial (subset of all) state required by the transaction.

crates/block-producer/src/state_view/mod.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: This will be handled internally by the account state tracker.

Mirko-von-Leipzig: You're right; I didn't consider partial application of inflight transactions.

I'll handle this using the account state tracker abstraction you recommended.

crates/block-producer/src/state_view/mod.rs
Mirko-von-Leipzig: I'll implement something like this using a `VecDeque`. New transaction are appended and states are removed from the front once they complete.

crates/block-producer/src/state_view/account_state.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// The latest state value of the given account.
+    pub fn get(&self, id: &AccountId) -> Option<&Digest> {
+        self.0.get(id).and_then(|states| states.back())
+    }
+
+    #[cfg(test)]
+    /// Number of accounts with inflight transactions.
+    pub fn len(&self) -> usize {
```
bobbinth, line 70: nit: we usually put attributes after comments. For example, here it would be:
```rust
/// Number of accounts with inflight transactions.
#[cfg(test)]
pub fn len(&self) -> usize {
```

crates/block-producer/src/state_view/account_state.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        // New accounts have a default state.
+        let current_state =
+            self.0.get(&id).and_then(|states| states.back()).copied().unwrap_or_default();
+
+        if current_state != init_state {
+            return Err(current_state);
+        }
+
+        self.0.entry(id).or_default().push_back(final_state);
```
bobbinth, line 35: Is this logic correct? What if we try to add an update for an existing account? That is, the `init_state` is not empty but the internal map has no entry for this account.

I think the high-level logic should be:
- If the entry for the account is already in the map, make sure the last state is the same as `init_state` and then append the `final_state` to the queue.
- If the entry is not in the map yet, insert a new entry for the account id with `final_state` in the single element in the queue.

Mirko-von-Leipzig: Ah. Right there is no way to test for an empty initial state because this is local only and excludes state that could be in the store.

Thanks!

crates/block-producer/src/state_view/account_state.rs
Mirko-von-Leipzig: I was hoping to find a lint to enforce this. Looks like its stuck in limbo despite being agreed upon :/ Bummer..

https://github.com/rust-lang/rustfmt/issues/3744
https://github.com/rust-lang/style-team/issues/72

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -31,12 +31,39 @@ pub type BatchId = Blake3Digest<32>;
...
+            details: tx.account_update().details().clone(),
+        }
+    }
+
+    /// Merges the transaction's update into this account update.
+    fn merge_tx(&mut self, tx: &ProvenTransaction) {
+        self.final_state = tx.account_update().final_state_hash();
+        self.transactions.push(tx.id());
+
+        // TODO: merge account update details. Waiting on implementation in.. miden-objects.
```
Mirko-von-Leipzig, line 63: I need to merge `self.details` here but its not possible to do with the public API provided by `miden-objects` for `AccountUpdateDetails`.

Presumably such merge code would best live in that crate anyway since its fallible and involved.

Mirko-von-Leipzig:
I've made the suggested changes, and also added support for multiple transactions in the transaction batcher (missing the detail merging as mentioned [here](https://github.com/0xPolygonMiden/miden-node/pull/407#discussion_r1679166952)).

I looked into any required `BlockWitness` changes, but I think those might already be covered by my batcher changes which didn't change the API? Might be wrong - still wrapping my head around things.

crates/block-producer/src/state_view/mod.rs
```diff
@@ -82,17 +82,25 @@ where
...
             candidate_tx,
-            &*self.accounts_in_flight.read().await,
             &*self.nullifiers_in_flight.read().await,
             &*self.notes_in_flight.read().await,
             &[],
         )?;
+        self.accounts_in_flight.read().await.verify_update(
+            candidate_tx.account_id(),
+            candidate_tx.account_update().init_state_hash(),
+        )?;
```
Mirko-von-Leipzig, line 92: This check is now external to `ensure_in_flight_constraints` which I'm not entirely happy with.

Its easy to move it back inside but then the second check that performs the update would do extra work (verifying twice). 

Options:
1. Leave as is
2. Move check back inside, checking twice is not the end of the world
3. Remove the check from `verify_and_add` (not a fan of this)
4. Some more type safe API which only lets you update after verification is confirmed. I did toy around with this, but its substantially more code.

Wasn't quite happy with any of these. Open to suggestions.

crates/block-producer/src/batch_builder/batch.rs
bobbinth: Could you make a PR with required changes in `miden-base`? I think the valid merge scenarios are:

1. `Private` + `Private` => `Private` (this should be trivial).
2. `Delta` + `Delta` => `Delta` (this is probably the most involved one since we'd need to merge 2 deltas).
3. `New` + `Delta` => `New` (this should be pretty easy because we can do `Account::apply_delta()`).

All other scenarios, as far as I can tell, should result in an error (i.e., `Delta` + `New` should not be possible as an account must first be created and then updated).

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -31,12 +31,39 @@ pub type BatchId = Blake3Digest<32>;
...
+            final_state: tx.account_update().final_state_hash(),
+            transactions: vec![tx.id()],
+            details: tx.account_update().details().clone(),
+        }
+    }
+
+    /// Merges the transaction's update into this account update.
+    fn merge_tx(&mut self, tx: &ProvenTransaction) {
+        self.final_state = tx.account_update().final_state_hash();
+        self.transactions.push(tx.id());
```
bobbinth, line 61: I would probably add an `assert!` somewhere here to make sure that `self.final_state == tx.account_update().init_state_hash()`.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -144,18 +172,18 @@ impl TransactionBatch {
     pub fn account_initial_states(&self) -> impl Iterator<Item = (AccountId, Digest)> + '_ {
         self.updated_accounts
             .iter()
-            .map(|(_, update)| (update.account_id(), update.init_state_hash()))
+            .map(|(&account, update)| (account, update.init_state))
```
bobbinth, line 175: nit: I would probably name the variable `account_id` rather than just `account`.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -144,18 +172,18 @@ impl TransactionBatch {
...
     }
 
     /// Returns an iterator over (account_id, details, new_state_hash) tuples for accounts that were
     /// modified in this transaction batch.
     pub fn updated_accounts(&self) -> impl Iterator<Item = BlockAccountUpdate> + '_ {
-        self.updated_accounts.iter().map(|(transaction_id, update)| {
+        self.updated_accounts.iter().map(|(&account, update)| {
```
bobbinth, line 181: Same nit as above.

crates/block-producer/src/state_view/mod.rs
bobbinth: I think I'd prefer the second option (do the check twice).

We could also change `InflightAccountStates::verify_and_add()` into just `InflightAccountStates::add()` - but I think the check is light enough where doing it twice shouldn't have a meaningful impact on things.

bobbinth (commented):

> I looked into any required `BlockWitness` changes, but I think those might already be covered by my batcher changes which didn't change the API?

I think the biggest thing we need to make sure is happening is when we have 2 batches with updates to the same account, we'd need to merge their `BlockAccountUpdate`'s. If this isn't happening already, this could be a fairly involved change, and I'm thinking we can create a separate issue for this and tackle it in a future PR.

crates/block-producer/src/state_view/account_state.rs
```diff
@@ -0,0 +1,117 @@
+use std::collections::{BTreeMap, VecDeque};
+
+use miden_objects::{accounts::AccountId, Digest};
+
+use crate::errors::VerifyTxError;
+
+/// Tracks the list of inflight account updates.
+///
+/// New transactions can be registered with [Self::verify_and_add]. States that are no longer considered inflight
+/// (e.g. due to being applied) may be removed using [Self::remove].
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I had assumed this would be handled by the lint/fmt so I never bother checking comments. I'll try figure out why this was ignored.

crates/block-producer/src/batch_builder/batch.rs
Mirko-von-Leipzig: Which immediately broken some tests 😬 (`test_build_batch_success`) as it built a batch out of identical transactions.

I "fixed" by making the account IDs unique per tx in a batch, however probably would be a good idea to support building consecutive transactions on the same account. This would make test setup easier.

crates/block-producer/src/batch_builder/batch.rs
Mirko-von-Leipzig: Opened [#797](https://github.com/0xPolygonMiden/miden-base/pull/797)

crates/block-producer/src/state_view/account_state.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: We don't actually specify this in our fmt configuration. We're missing (in all repos afaict):

```
wrap_comments = true
comment_width = 100
```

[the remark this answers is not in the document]
Mirko-von-Leipzig: Re-running fmt with these enabled has quite a large diff so I'll avoid pushing it into this PR.

Mirko-von-Leipzig:
I think this PR could do with another review :) I think its complete.

I also took a stab at merge support in the block witness in the last commit. I think this is done correctly, though of course there are no tests. I'll add those tomorrow -- just wanted input on my implementation first? No point in testing the wrong assumption :D

crates/block-producer/src/state_view/account_state.rs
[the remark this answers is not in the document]
bobbinth: Yeah - let's change this in a follow-up PR.

crates/block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -37,37 +37,62 @@ impl BlockWitness {
...
-            let mut account_initial_states: BTreeMap<AccountId, Digest> =
-                batches.iter().flat_map(TransactionBatch::account_initial_states).collect();
+            // Each accounts initial state in the context of the block i.e. only the first
+            // transaction affecting an account matters.
+            let mut account_initial_states = BTreeMap::<AccountId, Digest>::new();
+            for (account_id, initial_state) in
+                batches.iter().flat_map(TransactionBatch::account_initial_states)
+            {
+                account_initial_states.entry(account_id).or_insert(initial_state);
+            }
```
bobbinth, line 47: I think the assumption we make here is that account updates across batches are ordered chronologically. That is, if the same account is updated from `a -> b` in one batch and from `b -> c` in another. The batch with `a -> b` update comes first - right?

If so, I think we should state this assumption explicitly, and maybe in the future we can relax this condition somehow because batches are not necessarily meant to be sequential.

crates/block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -37,37 +37,62 @@ impl BlockWitness {
...
+            // Conversely, take the final state from the last transaction per account.
+            let mut account_final_states = BTreeMap::<AccountId, Digest>::new();
+            let mut account_transactions = BTreeMap::<AccountId, Vec<TransactionId>>::new();
+            for update in batches.iter().flat_map(TransactionBatch::updated_accounts) {
+                account_final_states.insert(update.account_id(), update.new_state_hash());
+                account_transactions
+                    .entry(update.account_id())
+                    .or_default()
+                    .extend_from_slice(update.transactions());
+            }
```
bobbinth, line 58: Similar comment as above.

bobbinth (approved):

Also agreed - would be good to add some tests.

crates/block-producer/src/block_builder/prover/block_witness.rs
Mirko-von-Leipzig: Ah. I'll try fix that and otherwise just add a comment.

bobbinth: I think it is important to note down somewhere that:

- We currently expect transactions in a batch to be ordered chronologically.
- We do not expect the batches in the block to be ordered chronologically.

Though I think a good question would be why the difference? I'm not sure I have a good answer to this yet :)

bobbinth: I guess maybe one justification:

It should be pretty easy to order transactions chronologically for a batch. It may not be so easy (or may be even impossible) to order batches in the same way in the block.

For example, let's say we have accounts `a` and `b`. Batch 1 updates account `a` from `a0 -> a1` and account `b` from `b1 -> b2`. Batch 2 updates account `a` from `a1 -> a2` and account `b` from `b0 -> b1`. Not sure if such situation can arise in practice, but theoretically, it should be possible. And in this case, two batches cannot be ordered chronologically with respect to each other.

Mirko-von-Leipzig: My (attempt at a) justification:

A transaction is rejected if it is not chronologically ordered i.e. it is rejected if the nonce and state transitions are not correct with the latest state on the node. Transactions are therefore placed into batches in chronological order.

There are many possible strategies for placing transactions into batches e.g.
- merge as many transactions together (compression)
- delay some transactions for MEV purposes
- minimize coupling between batches

A very simple batch strategy might just be round-robin style. This can result in a cyclic coupling between batches which means they cannot be ordered. Example: two batches `b`, two accounts `x`,`y`, with two transactions each `x0`, `x1`, `y0`, `y1`.

If the transactions arrive as `x0`,`y0`,`y1`,`x1` then we get batches:
```
b0 = [x0,y1]
b1 = [y0,x1]
```
which means there is no way to chronologically order the batches. This could be disallowed by tracking a batch dependency graph -- and there might be an argument that we need this anyway so that if a batch fails, we know which other batches must wait for it to be rebuilt.

Mirko-von-Leipzig: Put differently: it is already illegal for a batch to contain non-sequential transactions for the same account. It might be simplest to also disallow inter-account cycles.

bobbinth: Yes - agreed. In terms of how to implement this, one option is to do the following:

1. For each account, get the initial state from `BlockInputs`. Set the latest known state of this account to this value.
2. Iterate over all batches and find a batch where the account's initial state and the latest known sate match. Then set the latest known state to the final account state from this batch.
3. Repeat the above process until no batches have the initial state of the account match the current latest state.

This is probably not super efficient (and there could be ways to improve on this) - but I think it should work.

Mirko-von-Leipzig: I was thinking of that; but it would involve some emptiness checks and care to avoid leaving something behind.

What I'm going to try is to first collect all batches into a nested map: `Map<AccountId, Map<InitState, Update>>`. I can then start at the accounts input state and iterate through the inner update map until done.

I had started doing this; but then realised I should be doing this merging in the block updater instead - since it has similar issues of assuming batches are independent (and also invokes the witness builder).

Mirko-von-Leipzig:
Block witness and producer have been updated to support multiple txs as well. I implemented this in the least intrusive way I could find; however the result is a bit ugly imo.

The witness now outputs the block deltas since it is responsible for merging batches. My initial attempts at moving this out met with resistance from the "code shape" being wrong in several locations. A better flow can probably be found by taking a higher level view but that's a large diff left best left for another PR.

crates/block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -77,13 +48,83 @@ impl BlockWitness {
...
+        // Build account witnesses.
+        let mut account_witnesses = Vec::with_capacity(updated_accounts.len());
+        let mut block_updates = Vec::with_capacity(updated_accounts.len());
+
+        for (account_id, mut updates) in updated_accounts {
+            let (initial_state_hash, proof) = block_inputs
+                .accounts
+                .remove(&account_id)
+                .map(|witness| (witness.hash, witness.proof))
+                .ok_or(BuildBlockError::InconsistentAccountIds(vec![account_id]))?;
```
Mirko-von-Leipzig, line 71: We lost some error information because this performs the check for one account instead of all of them. If its important that we catch the entire set of IDs then that is possible, just tedious.

crates/block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -77,13 +48,83 @@ impl BlockWitness {
...
+                    .remove(&current_hash)
+                    .ok_or(BuildBlockError::InconsistentAccountStates(vec![account_id]))?;
+
+                transactions.extend(update.transactions);
+                current_hash = update.final_state;
+
+                details = Some(match details {
+                    None => update.details,
+                    Some(details) => {
+                        details.merge(update.details).expect("Should this be an error?")
```
Mirko-von-Leipzig, line 89: I wasn't sure if detail merging could fail at this stage? If yes, what error does this become? Till now only store and batch updates could mismatch, now batches can mismatch as well..

crates/block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -31,44 +33,13 @@ pub struct BlockWitness {
...
-                        },
-                    )
-                })
-                .collect()
-        };
+    ) -> Result<(Self, Vec<BlockAccountUpdate>), BuildBlockError> {
+        if batches.len() > MAX_BATCHES_PER_BLOCK {
+            return Err(BuildBlockError::TooManyBatchesInBlock(batches.len()));
+        }
+        Self::validate_nullifiers(&block_inputs, batches)?;
```
Mirko-von-Leipzig, line 42: The validate function became a bit of a liability - it can no longer be performed at the start but instead needs to be done in parts since account state can only be checked after we've merged the batch updates.

crates/block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -31,44 +33,13 @@ pub struct BlockWitness {
...
-                            initial_state_hash,
-                            final_state_hash: update.new_state_hash(),
-                            proof,
-                            transactions: update.transactions().to_vec(),
-                        },
-                    )
-                })
-                .collect()
-        };
+    ) -> Result<(Self, Vec<BlockAccountUpdate>), BuildBlockError> {
```
Mirko-von-Leipzig, line 38: I dislike the return type, but this was the smallest delta I could find.

bobbinth (approved):

Overall though, I wonder if we should simplify the whole approach somehow. I think things are relatively clean at the batch level, but when we get to the block level, some of the inter-dependencies we need to check worry me a bit.

A good time to think through and potentially refactor this logic would be when we migrate batch/block construction logic to `miden-base`. Basically, I think `miden-base` should contain the following logic:

1. Given a list of proven transactions and some additional data (e.g., inclusion paths for unauthenticated note), output a new `TransactionBatch`.
2. Given a list of transaction batches and some additional data (i.e., what is now called `BlockInputs`), output a new `Block`.

On my TODO list is to write up issues about these.

---
Separately, cc @igamigo - we should probably add integration tests in the client to try to test multiple in-flight transactions for the same account.

crates/block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -77,13 +48,83 @@ impl BlockWitness {
...
+                details.expect("Must be some by now"),
+                transactions,
+            ));
+        }
+
+        if !block_inputs.accounts.is_empty() {
+            return Err(BuildBlockError::InconsistentAccountIds(
+                block_inputs.accounts.keys().copied().collect(),
+            ));
+        }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Done. Follow-up question: should this actually be a fatal error?

Technically we can continue onward as there is probably just a minor bug in the store.. Counter argument is that this is weird behaviour and we should abort to be safe?

PR #408, opened 2024-07-12
Implemented test for unauthenticated input notes

polydez:
We implemented unauthenticated input notes support in these PRs: https://github.com/0xPolygonMiden/miden-node/pull/390 and https://github.com/0xPolygonMiden/miden-node/pull/396.

We need to add tests to make sure the unauthenticated note handling in the whole pipeline (from posting transaction to block building) works as expected.

I've also refactored some methods and improved some mocks needed for tests.

One of tests fails at the moment, will try to find a bug.

Resolves https://github.com/0xPolygonMiden/miden-node/issues/398


crates/block-producer/src/state_view/tests/verify_tx.rs
```diff
@@ -228,3 +229,80 @@ async fn test_verify_tx_vt5() {
...
+        Ok(()),
+        "Dangling unauthenticated notes must be found in the in-flight notes after previous tx verification"
+    );
+}
+
+/// Tests that `verify_tx()` fails when the unauthenticated input note not found not in the in-flight
+/// notes nor in the store
+#[tokio::test]
+#[miden_node_test_macro::enable_logging]
+async fn test_verify_tx_stored_unauthenticated_notes() {
```
Mirko-von-Leipzig, line 273: Nit: test name seems unrelated, maybe something like
```suggestion
async fn test_verify_tx_rejects_unknown_unauthenticated_note() {
```

crates/block-producer/src/state_view/tests/verify_tx.rs
```diff
@@ -228,3 +229,80 @@ async fn test_verify_tx_vt5() {
         Err(VerifyTxError::InputNotesAlreadyConsumed(vec![nullifier_in_both_txs]))
     );
 }
+
+/// Tests that `verify_tx()` succeeds when the unauthenticated input note found in the in-flight notes
+#[tokio::test]
+#[miden_node_test_macro::enable_logging]
+async fn test_verify_tx_in_flight_notes() {
```
Mirko-von-Leipzig, line 236: Nit, though these are getting fairly long..
```suggestion
async fn test_verify_tx_unauthenticated_inflight_notes_are_accepted() {
```

crates/block-producer/src/batch_builder/tests/mod.rs
```diff
@@ -146,6 +148,159 @@ async fn test_batches_added_back_to_queue_on_block_build_failure() {
...
+
+    let build_block_result = batch_builder
+        .block_builder
+        .build_block(&batch_builder.ready_batches.read().await)
+        .await;
+    assert_eq!(build_block_result, Ok(()));
+}
+
+#[tokio::test]
+async fn test_block_builder_missing_notes() {
```
Mirko-von-Leipzig, line 257:
```suggestion
async fn test_block_builder_fails_if_notes_are_missing() {
```

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -97,13 +96,13 @@ impl TransactionBatch {
...
-                    match found_unauthenticated_notes {
-                        Some(ref found_notes) => match found_notes.get(&input_note_header.id()) {
-                            Some(_path) => input_note.nullifier().into(),
-                            None => input_note.clone(),
-                        },
-                        None => input_note.clone(),
-                    }
+                    // If an unauthenticated note was found in the store, transform it to an authenticated one
+                    // (i.e. erase additional note details except the nullifier)
+                    found_unauthenticated_notes
```
Mirko-von-Leipzig, line 101: Nit: it feels weird having an optional container as input.

We can remove a layer of indirection by dropping the `Option<_container_>`. Callers can transform their optional container with `.unwrap_or_default` if necessary.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -244,3 +244,145 @@ impl OutputNoteTracker {
...
+    use crate::test_utils::{
+        mock_proven_tx,
+        note::{mock_note, mock_output_note, mock_unauthenticated_note_commitment},
+    };
+
+    #[test]
+    fn test_output_note_tracker_duplicate_output_notes() {
+        let mut txs = mock_proven_txs();
+
+        OutputNoteTracker::new(&txs).expect("Something went wrong, no error expected");
```
Mirko-von-Leipzig, line 260: Nit: may as well use `unwrap` if there is no meaningful expectation message

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -244,3 +244,145 @@ impl OutputNoteTracker {
...
+            vec![],
+            vec![duplicate_output_note.clone(), mock_output_note(8), mock_output_note(4)],
+        ));
+
+        match OutputNoteTracker::new(&txs) {
+            Err(BuildBatchError::DuplicateOutputNote(note_id, _)) => {
+                assert_eq!(note_id, duplicate_output_note.id())
+            },
+            res => panic!("Unexpected result: {res:?}"),
+        }
```
Mirko-von-Leipzig, line 275: Pity that `assert_matches!` is not yet stable; but maybe we should use the [assert_matches crate](https://docs.rs/assert_matches/latest/assert_matches/) until it is. 

Its a common occurrence I think? If not then these once off matches are fine.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -244,3 +244,145 @@ impl OutputNoteTracker {
...
+    fn test_output_note_tracker_remove_in_place_consumed_note() {
+        let txs = mock_proven_txs();
+        let mut tracker = OutputNoteTracker::new(&txs).unwrap();
+
+        let note_to_remove = mock_note(4);
+
+        assert!(tracker.remove_note(note_to_remove.header(), &txs).unwrap());
+        assert!(!tracker.remove_note(note_to_remove.header(), &txs).unwrap());
+
+        // Check that output notes are in the expected order and consumed note was removed
```
Mirko-von-Leipzig, line 288: Is note order important - sincere noob question

Mirko-von-Leipzig (approved):
I think this looks good though I'm not 100% familiar with all the corner cases yet.

I left a few test naming nits/suggestions - feel free to ignore since its a style thing, but will elaborate below.

In general the tests name the condition being tested, but not the expected result. So as someone unfamiliar with the code I need to read the test carefully to know what's happening.

As an example:
```rust
fn missing_notes()
```
versus
```rust
fn block_with_missing_notes_is_rejected()
```

crates/block-producer/src/batch_builder/batch.rs
polydez: Yes, we need to keep notes order.

crates/block-producer/src/batch_builder/batch.rs
polydez: Nice catch, but I don't think this worth of using one more crate. Let's rewrite it once `assert_matches!` is stabilized.

crates/block-producer/src/state_view/tests/verify_tx.rs
polydez: I don't think suggested name is more correct.

crates/block-producer/src/batch_builder/batch.rs
polydez: I don't have strong opinion on this matter, but your suggestion can help to simplify code a bit, so I decided to get rid of `Option`, thank you!

crates/block-producer/src/state_view/tests/verify_tx.rs
Mirko-von-Leipzig: Naming is hard 😬

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -244,3 +243,145 @@ impl OutputNoteTracker {
...
+    fn test_convert_unauthenticated_note_to_authenticated() {
+        let txs = mock_proven_txs();
+        let found_unauthenticated_notes =
+            BTreeMap::from_iter([(mock_note(5).id(), Default::default())]);
+        let batch = TransactionBatch::new(txs, found_unauthenticated_notes).unwrap();
+
+        let expected_input_notes =
+            vec![mock_unauthenticated_note_commitment(1), mock_note(5).nullifier().into()];
+        assert_eq!(batch.input_notes.len(), expected_input_notes.len());
+        assert_eq!(batch.input_notes, expected_input_notes);
```
bobbinth, line 367: The check for the lengths to match seems a bit redundant as if the actual vectors don't match, the lengths won't match either.

bobbinth (approved):
Looks good! Thank you! I left a coupe of comments inline - but they are all pretty minor.

One general thing (not specific to this PR): I find tests in this repo pretty hard to read and follow. We should probably start thinking about refactoring of the testing framework soon. Let's create an issue for this.

PR #411, opened 2024-07-16
feat(faucet): embed static website

Mirko-von-Leipzig:
This PR simplifies faucet deployment by embedding the website within the binary. This removes the need to separately deploy the website resources.

To facilitate this, this PR also fixes a paper-cut with the javascript callbacks. Previously these were hardcoded to use `localhost:8080` as the server address. These would have to be manually edited to the correct URL (which might be a proxy) during deployment. Instead we now use `window.location.href` which gives the correct url of the current page.

Note: I am not a web expert so if anyone has any input on the above that would be welcome.

------------------------

The static embedding is handled using a build script + additional dependencies, which I dislike but it seems this is the "blessed" way to achieve this using `actix-web`.

I tested this locally but without a proxy. Though I don't see why this would matter :)

### DIFF
--- a/bin/faucet/Cargo.toml
+++ b/bin/faucet/Cargo.toml
@@ -32,11 +32,11 @@ miden-tx = { workspace = true,  features = ["concurrent"] }
 rand = { version = "0.8" }
 rand_chacha = "0.3"
 serde = { version = "1.0", features = ["derive"] }
-static-files = "0.2.1"
+static-files = "0.2"
 thiserror = { workspace = true }
 toml = { version = "0.8" }
 tonic = { workspace = true }
 tracing = { workspace = true }
 
 [build-dependencies]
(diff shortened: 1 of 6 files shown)


bin/faucet/Cargo.toml
```diff
@@ -32,7 +32,11 @@ miden-tx = { workspace = true,  features = ["concurrent"] }
 rand = { version = "0.8" }
 rand_chacha = "0.3"
 serde = { version = "1.0", features = ["derive"] }
+static-files = "0.2.1"
```
bobbinth, line 35: nit: we usually don't specify patch versions for dependencies (i.e., it would be `0.2` instead of `0.2.1`).

bobbinth (approved):
Looks good! Thank you! I left one tiny nit inline.

PR #412, opened 2024-07-17
fix: only rebuild proto on changes

Mirko-von-Leipzig:
The proto files are being "rebuilt" on every compile at the moment. This is due to the 
```
    println!("cargo:rerun-if-changed=generated");
```
which is a directory that does not exist, causing it to trigger on every build.

----------------------------------------

There were several attempts at this; none were really good enough. My last attempt is changing very little but adds a few consistency improvements.

Notably, it adds a simple CI check to ensure changes the main `proto` files is propagated to all repo's which use them.

To aid this, I've made the build script outputs more repeatable by

- deleting target files and folders
- sorting outputs


Mirko-von-Leipzig:
As a side note - it feels weird putting "fixes" under `Enhancements` in the changelog. Should there also be a `Fixes`?

crates/proto/build.rs
```diff
@@ -8,24 +8,24 @@ use prost::Message;
 /// This is done only if BUILD_PROTO environment variable is set to `1` to avoid running the script
 /// on crates.io where repo-level .proto files are not available.
 fn main() -> miette::Result<()> {
-    println!("cargo:rerun-if-changed=generated");
-    println!("cargo:rerun-if-changed=../../proto");
+    let crate_root = PathBuf::from_str(std::env!("CARGO_MANIFEST_DIR")).into_diagnostic()?;
+    let workspace_root =
+        crate_root.parent().and_then(|crates| crates.parent()).ok_or_else(|| {
+            miette::miette!("Failed to navigate two directories up into workspace root")
+        })?;
```
bobbinth, line 15: Would this still work when we publish to crates.io? (there may not be a workspace root there).

One potential option is to move this code after we check for `BUILD_PROTO` environment variable. But I'm not sure how it will affect things.

bobbinth (approved):

> I am however unsure why the generated directory is included at all -- maybe to rebuild it if it accidentally got deleted? Though that seems excessive as the dev can see this happened using git.

I believe the motivation for this is described in https://github.com/0xPolygonMiden/miden-node/pull/9#discussion_r1267109540 - but I'm not sure if it is still valid.


> As a side note - it feels weird putting "fixes" under `Enhancements` in the changelog. Should there also be a `Fixes`?

Yes, there should be `Fixes` section as well. Overall, I think we should have 3 categories (or maybe more) in the changelog: `Features`, `Enhancements`, and `Fixes` (we had 2 of these in the [last](https://github.com/0xPolygonMiden/miden-node/releases/tag/v0.4.0) release).

crates/proto/build.rs
Mirko-von-Leipzig: Ah I see and it worked before because the publish abort check was earlier. I'll see if I can't just move the check to the beginning.

Alternative thought: lets just remove the build script entirely. We already have the generated files checked into git. So really its only buying us the convenience of not having to think about things when we fiddle with the proto files (which should be fairly rare). And if you're changing the proto files you can just manually run protoc (which we can simply add as a script file for ease of use, or to the readme as an instruction). For safety we can add a CI step that checks that protoc output still matches the generated files.

bobbinth: If not too difficult, I think the alternative approach would be preferred. Specifically:

- Add something like `proto` command (and an associated script) to the Makefile so that we could run `make proto` to rebuild the generated file.
- Add a CI check to make sure that if any of the `.proto` files changed, the generated files are up to date.

One comment though: if we go this route, we should also make sure we can do the same for the [build script](https://github.com/0xPolygonMiden/miden-node/blob/main/crates/rpc-proto/build.rs) in the `rpc-proto` crate.

Mirko-von-Leipzig: Who/what uses the `write_proto` function?

bobbinth: https://github.com/0xPolygonMiden/miden-client/blob/main/crates/rust-client/build.rs#L21

Mirko-von-Leipzig: I've updated the description for my new attempt. Might be a bit too extreme of a change though it does seem to work.

bin/compile-proto/src/main.rs
```diff
@@ -0,0 +1,72 @@
+use std::{fs, io::Write, path::PathBuf};
+
+use miette::IntoDiagnostic;
+use prost::Message;
+
+/// Compiles the `miden-node-proto` crate's proto definitions.
+///
+/// Check the readme for more information.
```
bobbinth, line 8: Doesn't seem like this crate has a readme file.

Cargo.toml
```diff
@@ -2,15 +2,15 @@
...
   "crates/block-producer",
   "crates/proto",
-  "crates/rpc-proto",
   "crates/rpc",
   "crates/store",
   "crates/utils",
   "crates/test-macro",
 ]
-default-members = ["bin/faucet", "bin/node", "crates/proto", "crates/rpc-proto"]
+default-members = ["bin/faucet", "bin/node", "crates/proto"]
```
bobbinth, line 13: I'm wondering: do we need to have `default-members` at all? If we don't have it, `cargo build` etc. commands will be applied to all crates, right? Not sure that's a bad thing.

bobbinth (changes requested):
Thank you! This is not a super-deep review, but I did leave a couple of small comments inline. Regarding the general approach, I have the following comments:

1. I would prefer to keep `proto` and `rpc-proto` crates separate. The main reason is that `rpc-proto` is meant to be the interface definition only for the RPC component of the node (it doesn't quite work like that right now, but that's a separate issue) and it is meant to be consumed by external users. While `proto` is mean to contain interfaces of all node components and to be used primarily internally.
2. I do like adding the `make proto` command to the Makefile and having this be the only way to re-build generated files and copy protobuf files into `rpc-proto`. Whether the implementation of this command relies on a separate binary or build scripts gated by an environment variable is not super important - but I do think that using just the build scripts may be cleaner here.
3. I think it would be great if we added a CI check to make sure that updates to `.proto` files are properly propagated (i.e., someone didn't forget to run the `make proto` command).

So, my recommendations for this PR would be:

1. Revert the merging of `proto` and `rpc-proto` crates.
2. Add `proto` command to the Makefile.
  a. If possible, rely on the build scripts to recompile/copy the protobuf files. For example, maybe this command would just do `BUILD_PROTO=1 cargo build` and the build scripts would be fully conditional on `BUILD_PROTO` variable.
  b. If the above doesn't work for some reason, create a separate binary to perform this tasks (similar to what is already in this PR).
3. Add a CI check to make sure `make proto` command does not cause a diff.

Cargo.toml
Mirko-von-Leipzig: Correct; its actually somewhat annoying to have as you always have to include the package details for non-default-members.

Mirko-von-Leipzig:
I really wish github allowed replying to top level comments..

@bobbinth what about a more explicit approach of moving ownership of the proto definitions to the `proto` crate.  We can then

1. Remove the env flag trigger from the `proto/build.rs` since the definitions are now part of the crate filesystem (I think that was the issue?).
2. Add the functionality required by `rpc-proto` to `proto` under some feature flag (sort of how my current impl merged the two). Really this would only require the actual file contents be made available under a `nostd` implementation.
3. Have `rpc-proto` use `proto` as a dependency.

Benefits:

- A single copy of the proto definitions.
- A single "pure" `build.rs` with no bool flags, no copying around.
- The current (temporary) dependency of `rpc-proto` is made explicit.
- Non-breaking change for both crates.

I'll attempt this so long, but today is a slow day for me it seems so maybe you'll be able to prevent me from wasting my time if you spot any cracks.

Mirko-von-Leipzig:
I've implemented the latest idea I proposed [here](https://github.com/0xPolygonMiden/miden-node/pull/412#issuecomment-2238882055).

I'm waiting on CI to double-check that I don't need some `nostd` related feature-gating.

crates/rpc-proto/src/lib.rs
```diff
@@ -3,8 +3,7 @@
 #[cfg(feature = "std")]
 extern crate std;
 
-mod proto_files;
-pub use proto_files::PROTO_FILES;
+pub use miden_node_proto::PROTO_FILES;
```
Mirko-von-Leipzig, line 6: Is exporting a `nostd` compliant `const` from an otherwise non-compliant dependency fine? I think it is -- haven't experimented much with `nostd`.

crates/proto/build.rs
```diff
@@ -3,53 +3,116 @@ use std::{env, fs, path::PathBuf};
...
+
+fn generate_proto_file_getter() -> std::io::Result<()> {
+    // Unfortunately doing relative paths is convoluted, so this function harcodes
+    // the paths instead.
+
+    use std::io::Write;
+
+    let mut target = fs::File::create("src/proto_files.rs")?;
+
+    writeln!(target, "// File generated by buid.rs")?;
```
igamigo, line 98:
```suggestion
    writeln!(target, "// File generated by build.rs")?;
```

crates/proto/build.rs
```diff
@@ -3,53 +3,116 @@ use std::{env, fs, path::PathBuf};
...
+
+    for entry in std::fs::read_dir("proto")? {
+        let entry = entry?;
+        let path = entry.path();
+        if path.is_file() {
+            let filename = path
+                .file_name()
+                .and_then(|f| f.to_str())
+                .expect("Filename should exist and  be utf8");
+            writeln!(target, r#"    ("{filename}.rs", include_str!("../proto/{filename}")),"#)?;
```
igamigo, line 111: Is the `.rs` suffix in `{filename}.rs` correct? The map would then have keys such as `requests.proto.rs`

Mirko-von-Leipzig: Thanks; I think I've successfully confused myself with strings and extensions.

Mirko-von-Leipzig:
We can still add a CI check if we want; but should be difficult to get wrong with the build script

bobbinth:
I think this approach works, but I'm not sure I like it better than what we had before (assuming we get rid of the un-needed rebuilds).

One reason is that I think there is benefit to keeping `.proto` files separate from the Rust code. These files describe the interfaces of the node components regardless of the language these components are implemented in. Granted, we don't have any plans right now for any non-Rust implementations, but if someone wanted to grab these files to build a Go implementation of the node, it would be a bit more natural to get them from the repo root rather than from one of the internal crates.

Another reason is that as I mentioned, ideally `rpc-proto` shouldn't pull in any of the files related to storage/block producer. We don't have this working yet because of the way we wrote the `.proto` files and refactoring them is a pretty low priority, but it feels a bit hacky to couple `proto` and `rpc-proto` crates.

So, unless there are some complications (i.e., conditional build scripts don't work for some reason), I'd still prefer to keep the `.proto` files in the repo root.

Mirko-von-Leipzig:
I've reverted to the original format and placing. I've just added a CI check which rebuilds and ensures there is no diff.

crates/proto/build.rs
```diff
@@ -47,9 +57,41 @@ fn main() -> miette::Result<()> {
...
         .compile_with_config(prost_config, protos, includes)
         .into_diagnostic()?;
 
+    generate_mod_rs(&dst_dir).into_diagnostic()?;
+
     Ok(())
 }
+
+/// Generate `mod.rs` which includes all files in the folder as submodules.
+fn generate_mod_rs(directory: impl AsRef<Path>) -> std::io::Result<()> {
```
bobbinth, line 70: Nice catch!

PR #413, opened 2024-07-17
Added warning on `CHANGELOG.md` modification

phklive:
In this PR I propose to add a warning in the CI if the `CHANGELOG.md` file has not been modified.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,17 +9,6 @@ on:
     types: [opened, reopened, synchronize]
 
 jobs:
-  version:
-    name: check rust version consistency
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@main
-        with:
-          profile: minimal
-          override: true
-      - name: check rust versions
-        run: ./scripts/check-rust-version.sh
-
   rustfmt:
     name: rustfmt check nightly on ubuntu-latest
     runs-on: ubuntu-latest
@@ -41,3 +30,43 @@ jobs:
           rustup update --no-self-update nightly
           rustup +nightly component add clippy
           make clippy
+
+  doc:
+    name: doc stable on ubuntu-latest
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+      - name: Build docs
+        run: |
+          rustup update --no-self-update
+          make doc
+
+  version:
+    name: check rust version consistency
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+        with:
+          profile: minimal
+          override: true
+      - name: check rust versions
+        run: ./scripts/check-rust-version.sh
+
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@main
+        with:
+          fetch-depth: 0
+      - name: Check if CHANGELOG.md is modified
+        run: |
+          # Get the list of changed files in the PR
+          changed_files=$(git diff --name-only origin/${{ github.event.pull_request.base.ref }}...${{ github.sha }})
+          # Check if CHANGELOG.md is in the list of changed files
+          if echo "$changed_files" | grep -q '^CHANGELOG.md$'; then
+            echo "CHANGELOG.md has been modified."
+          else
+            echo $'::warning file=CHANGELOG.md::CHANGELOG.md has not been modified.\n This warning can be ignored if is has been explicitely decided not to log changes.\n Except in this situation, make sure to add log changes.'
+            exit 1
+          fi


.github/workflows/lint.yml
```diff
@@ -40,4 +29,44 @@ jobs:
...
+        run: |
+          # Get the list of changed files in the PR
+          changed_files=$(git diff --name-only origin/${{ github.event.pull_request.base.ref }}...${{ github.sha }})
+          # Check if CHANGELOG.md is in the list of changed files
+          if echo "$changed_files" | grep -q '^CHANGELOG.md$'; then
+            echo "CHANGELOG.md has been modified."
+          else
+            echo $'::warning file=CHANGELOG.md::CHANGELOG.md has not been modified.\n This warning can be ignored if is has been explicitely decided not to log changes.\n Except in this situation, make sure to add log changes.'
+            exit 1
+          fi
```
bobbinth, line 72: Before merging this, let's figure out if anything needs to be adjusted based on https://github.com/0xPolygonMiden/miden-vm/pull/1340#issuecomment-2233824292.

.github/workflows/lint.yml
```diff
@@ -40,4 +29,44 @@ jobs:
...
+  version:
+    name: check rust version consistency
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+        with:
+          profile: minimal
+          override: true
+      - name: check rust versions
+        run: ./scripts/check-rust-version.sh
```
bobbinth, line 53: While we are at it, let's also change `rust-toolchain` to `rust-toolchain.toml` and update the `check-rust-version.sh` script accordingly.

bobbinth (commented):

.github/workflows/lint.yml
Mirko-von-Leipzig: I don't quite understand why we're restricting the current toolchain to the MSRV of the workspace. I've read through #267, #272 and the related miden-vm PRs and issues, and discord.

From what I can tell its issues due to using the nightly compiler -- which afaik we're not using?

.github/workflows/lint.yml
```diff
@@ -40,4 +29,44 @@ jobs:
...
+      - name: Check if CHANGELOG.md is modified
+        run: |
+          # Get the list of changed files in the PR
+          changed_files=$(git diff --name-only origin/${{ github.event.pull_request.base.ref }}...${{ github.sha }})
+          # Check if CHANGELOG.md is in the list of changed files
+          if echo "$changed_files" | grep -q '^CHANGELOG.md$'; then
+            echo "CHANGELOG.md has been modified."
+          else
+            echo $'::warning file=CHANGELOG.md::CHANGELOG.md has not been modified.\n This warning can be ignored if is has been explicitely decided not to log changes.\n Except in this situation, make sure to add log changes.'
+            exit 1
```
Mirko-von-Leipzig, line 71: A downside to this is that this marks the overall CI run as a failure; somewhat destroying the glance value. If we think it will be rare to have a no-op changelog PR then this is probably fine.

Did you look into existing github actions we could use? e.g. this [one](https://github.com/tarides/changelog-check-action) allows you to skip the check by adding a `no changelog` label to the PR. I haven't actually audited the action itself though.

Mirko-von-Leipzig (approved):
Some questions, but the addition itself seems fine (modulo @bobbinth's comments).

I would also consider using dtolnay's [toolchain action](https://github.com/dtolnay/rust-toolchain) but that can be a separate issue.

.github/workflows/lint.yml
bobbinth: I'm a bit hesitant to use actions that are not very popular - but skipping changelog check by adding `no changelog` label seems cool. Is this something that can be fairly easily emulated locally?

.github/workflows/lint.yml
bobbinth: I think the primary motivation is that everyone is using the same toolchain as we've had issues when things suddenly break if a different version is used. As an example, in https://github.com/facebook/winterfell/issues/277 code that worked fine prior to 1.78 started failing with 1.78 onwards (this was due to the bug in the code, but still).

Another example is something like https://github.com/0xPolygonMiden/miden-vm/issues/917, where nothing broke but there was a significant performance degradation when moving from 1.69 to 1.70 (which fixed itself after 1.74).

bobbinth:
> I would also consider using dtolnay's [toolchain action](https://github.com/dtolnay/rust-toolchain) but that can be a separate issue.

We've had this before, I believe, but decided to switch to them methodology used by [rust-analyzer](https://github.com/rust-lang/rust-analyzer/blob/99718d0c8bc5aadd993acdcabc1778fc7b5cc572/.github/workflows/ci.yaml#L63-L67).

.github/workflows/lint.yml
Mirko-von-Leipzig: Fair enough; I believe we should be able to simply copy the actions code. Its very short, label is added as an env variable and then simply used in an `if then` within the script.

```
env:
    NO_CHANGELOG_LABEL: ${{ contains(github.event.pull_request.labels.*.name, 'no changelog') }}
```

.github/workflows/lint.yml
phklive: Opened this PR on the VM with the rationale behind the changes: https://github.com/0xPolygonMiden/miden-vm/pull/1406

.github/workflows/lint.yml
phklive: In this PR I inspired myself with the action code and merged it with our existing one, enabling us to get the `no changelog` label feature: https://github.com/0xPolygonMiden/miden-vm/pull/1406

And  not having to depend on not widely used actions as @bobbinth mentioned

bobbinth (approved):
Looks good! Thank you! Let's do the following and then merge:

- Confirm that https://github.com/0xPolygonMiden/miden-vm/pull/1406 worked and apply the same changes here.
- Update `rust-toolchain` to be a TOML file.

phklive:
@bobbinth what is the toolchain that we want to use for our repos? 

https://github.com/0xPolygonMiden/compiler/blob/44e2d383f4871c8b2ed5c86e7a4b3393cac34a5a/rust-toolchain.toml#L3

bobbinth (approved):
Looks good! Thank you!

> @bobbinth what is the toolchain that we want to use for our repos?
> 
> https://github.com/0xPolygonMiden/compiler/blob/44e2d383f4871c8b2ed5c86e7a4b3393cac34a5a/rust-toolchain.toml#L3

Let's use the one from the `miden-vm` repo.

PR #417, opened 2024-07-24
Rerun build scripts for protobuf on env changed

polydez:
Build scripts for code generation from *.proto files depend on `BUILD_PROTO` environment variable. But when user sets value of this variable to "1" after previous run, build scripts runs are skipped. This PR fixes that.


Mirko-von-Leipzig (approved):
Oh interesting - nice catch!

PR #419, opened 2024-07-25
feat: implement `CheckNullifiersByPrefix` endpoint

### LINKED ISSUE #404: [Feature]: Add an endpoint to check for nullifiers given block bounds and prefixes
> > If we were to have an endpoint to validate whether a list of notes have been consumed / nullifiers are present in the nullifier SMT, we could add a single step after both the normal sync and the ignored notes sync where we see if any of those got consumed.
> 
> We do have a [CheckNullifiers](https://github.com/0xPolygonMiden/miden-node/blob/main/proto/rpc.proto#L9) endpoint in the node - thought, it requires sending full nullifiers and returns back authentication paths for the nullifiers found in the node. So, this is a bit more heaving then needed and also reduced privacy. Maybe it is worth adding an endpoint to the node to return a list of nullifiers based on some prefix and block bounds.

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-client/issues/405#issuecomment-2212513725_

tomyrd:
closes #404

This is my first time implementing something in the node, I originally made this to better understand the node internals. Be sure to correct me if I missed something important in terms of validation or security.

This PR looks to add a new endpoint to the node that lets users check if a note is consumed without directly exposing its nullifier. In the next hours I'll also be adding a PR in the client that uses this endpoint to deal with imported notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Enhancements
 
+- Added `CheckNullifiersByPrefix` endpoint (#419).
 - [BREAKING] Configuration files with unknown properties are now rejected (#401).
 - [BREAKING] Removed redundant node configuration properties (#401).
 - Improve type safety of the transaction inputs nullifier mapping (#406).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 ### Enhancements
 
-- Added `CheckNullifiersByPrefix` endpoint (#419).
 - [BREAKING] Configuration files with unknown properties are now rejected (#401).
 - [BREAKING] Removed redundant node configuration properties (#401).
 - Improve type safety of the transaction inputs nullifier mapping (#406).
@@ -13,6 +12,7 @@
 - Added warning on CI for `CHANGELOG.md` (#413).
 - Now accounts for genesis are optional. Accounts directory will be overwritten, if `--force` flag is set (#420).
 - Added `GetAccountStateDelta` endpoint (#418).
+- Added `CheckNullifiersByPrefix` endpoint (#419).
 
 ### Fixes
 
--- a/crates/store/src/db/sql.rs
+++ b/crates/store/src/db/sql.rs
@@ -353,6 +353,17 @@ pub fn select_nullifiers_by_block_range(
     Ok(result)
 }
 
+/// Select nullifiers created that match the `nullifier_prefixes` filter using the given
+/// [Connection].
+///
+/// Each value of the `nullifier_prefixes` is only the `prefix_len` most significant bits
+/// of the nullifier of interest to the client. This hides the details of the specific
+/// nullifier being requested.
+///
+/// # Returns
+///
+/// A vector of [NullifierInfo] with the nullifiers and the block height at which they were
+/// created, or an error.
 pub fn select_nullifiers_by_prefix(
     conn: &mut Connection,
     _prefix_len: u32,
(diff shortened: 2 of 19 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Enhancements
 
+- Added `CheckNullifiersByPrefix` endpoint (#419).
```
bobbinth, line 7: nit: we usually sort these in descending order (so new PRs get added at the end of the list).

proto/responses.proto
```diff
@@ -17,6 +17,11 @@ message CheckNullifiersResponse {
     repeated smt.SmtOpening proofs = 1;
 }
 
+message CheckNullifiersByPrefixResponse {
+    // List of nullifiers matching the 16-bit prefixes specified in the request.
+    repeated NullifierUpdate nullifiers = 1;
+}
```
bobbinth, line 23: nit: I'd probably omit mentioning of the 16 bits here.

crates/store/src/db/sql.rs
```diff
@@ -307,6 +307,51 @@ pub fn select_nullifiers_by_block_range(
...
+///
+/// # Returns
+///
+/// A vector of [NullifierInfo] with the nullifiers and the block height at which they were
+/// created, or an error.
+pub fn select_nullifiers_by_prefix(
+    conn: &mut Connection,
+    _prefix_len: u32,
+    nullifier_prefixes: &[u32],
+) -> Result<Vec<NullifierInfo>> {
```
bobbinth, line 325: I would mention that currently only 16-bit prefixes are supported and also add an `assert_eq!` in the function to enforce this.

bobbinth (commented):
Looks good! Thank you! Not a super in-depth review - but I did leave a few small comments inline.

crates/store/src/db/sql.rs
```diff
@@ -307,6 +307,51 @@ pub fn select_nullifiers_by_block_range(
...
+    let nullifier_prefixes: Vec<Value> =
+        nullifier_prefixes.iter().copied().map(u32_to_value).collect();
+
+    let mut stmt = conn.prepare(
```
Mirko-von-Leipzig, line 329: @bobbinth separate note: we should consider using [prepare_cached](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.prepare_cached). Parsing a sql statement is actually a non-trivial cost.

I'm not suggesting this PR changes this since `prepare` is used in all the other queries as well.

proto/rpc.proto
```diff
@@ -6,6 +6,7 @@ import "requests.proto";
 import "responses.proto";
 
 service Api {
+    rpc CheckNullifiersByPrefix(requests.CheckNullifiersByPrefixRequest) returns (responses.CheckNullifiersByPrefixResponse) {}
```
Mirko-von-Leipzig, line 9: Nit: this should move one down to be properly sorted alphabetically (shorter string comes 1st).

proto/store.proto
```diff
@@ -9,6 +9,7 @@ import "responses.proto";
 
 service Api {
     rpc ApplyBlock(requests.ApplyBlockRequest) returns (responses.ApplyBlockResponse) {}
+    rpc CheckNullifiersByPrefix(requests.CheckNullifiersByPrefixRequest) returns (responses.CheckNullifiersByPrefixResponse) {}
```
Mirko-von-Leipzig, line 12: Nit: same alphabetical sorting (one down)

bobbinth (approved):
All looks good! Thank you!

crates/store/src/db/sql.rs
bobbinth: Agreed! Let's create an issue for this.

Mirko-von-Leipzig: #423

PR #420, opened 2024-07-26
Make accounts optional for genesis generation

polydez:
Resolves: https://github.com/0xPolygonMiden/miden-node/issues/198

With this PR user won't be forced to create at least one account in genesis. Also I slightly changed behavior working with `accounts` directory during genesis creation: if this directory already exists, program will fail. If `--force` flag is provided, directory will be overwritten.


bobbinth (approved):
Looks good! Thank you!

PR #421, opened 2024-07-26
feat(store): `GetNoteAuthenticationInfo` endpoint

polydez:
Resolves: https://github.com/0xPolygonMiden/miden-node/issues/395

This PR adds `GetNoteAuthenticationInfo` endpoint. Prior this we had to use `GetNotesById` endpoint for getting inclusion proofs, but it returns a lot of unnecessary information for such task.

Also `GetBlockInputs` endpoint now returns not only found requested note IDs, but also their inclusion proofs.


proto/note.proto
```diff
@@ -23,6 +23,11 @@ message Note {
     optional bytes details = 6;
 }
 
+message NoteInclusionProof {
+    digest.Digest note_id = 1;
+    merkle.MerklePath merkle_path = 2;
+}
```
bobbinth, line 29: This needs to include a bit more data. Specifically, note inclusion proof should be sufficient to prove that a note is present in the chan. This requires:

1. A Merkle path from the leaf to the root of note Merkle tree of a block.
2. Index of the note in this block.
3. Block's note root + sub hash (or alternatively, the entire block header).
4. MMR path to the block in the current chain MMR.

Since several notes could be from the same block, it may make sense to de-duplicate the data. Though, maybe it doesn't matter too much and at least for now we can send all required data on per note basis.

proto/rpc.proto
```diff
@@ -10,6 +10,7 @@ service Api {
     rpc GetAccountDetails(requests.GetAccountDetailsRequest) returns (responses.GetAccountDetailsResponse) {}
     rpc GetBlockByNumber(requests.GetBlockByNumberRequest) returns (responses.GetBlockByNumberResponse) {}
     rpc GetBlockHeaderByNumber(requests.GetBlockHeaderByNumberRequest) returns (responses.GetBlockHeaderByNumberResponse) {}
+    rpc GetNoteInclusionProofs(requests.GetNotesByIdRequest) returns (responses.GetNoteInclusionProofsResponse) {}
```
bobbinth, line 13: Even if the requests are the same - we should probably create a different request message (i.e., we probably shouldn't reuse `GetNotesByIdRequest` request here).

proto/store.proto
```diff
@@ -14,6 +14,7 @@ service Api {
     rpc GetBlockByNumber(requests.GetBlockByNumberRequest) returns (responses.GetBlockByNumberResponse) {}
     rpc GetBlockHeaderByNumber(requests.GetBlockHeaderByNumberRequest) returns (responses.GetBlockHeaderByNumberResponse) {}
     rpc GetBlockInputs(requests.GetBlockInputsRequest) returns (responses.GetBlockInputsResponse) {}
+    rpc GetNoteInclusionProofs(requests.GetNotesByIdRequest) returns (responses.GetNoteInclusionProofsResponse) {}
```
bobbinth, line 17: Same comment as above.

bobbinth (commented):
Thank you! Looks good. This is not a full review yet - but I left a couple of comments inline. The main one is that note inclusion proof is actually a bit more complex than just a single Merkle path.

proto/note.proto
polydez: @bobbinth thank you for clarification!
> 4. MMR path to the block in the current chain MMR.
Do we need this field?
I found implementation of `NoteInclusionProof` in `miden-base` and there is no such field:
https://github.com/0xPolygonMiden/miden-base/blob/f80276e93521f896b2ff5544f42ddaba6e50038e/objects/src/notes/origin.rs#L23-L35

crates/store/src/db/sql.rs
```diff
@@ -665,6 +665,58 @@ pub fn select_notes_by_id(conn: &mut Connection, note_ids: &[NoteId]) -> Result<
...
+
+        let note_id_data = row.get_ref(1)?.as_blob()?;
+        let note_id = NoteId::read_from_bytes(note_id_data)?;
+
+        let batch_index = row.get(2)?;
+        let note_index = row.get(3)?;
+        let node_index_in_block = BlockNoteIndex::new(batch_index, note_index).to_absolute_index();
+
+        let merkle_path_data = row.get_ref(4)?.as_blob()?;
+        let merkle_path = MerklePath::read_from_bytes(merkle_path_data)?;
```
Mirko-von-Leipzig, line 710: Meta: something to consider is adding typed getters using an extension trait for the sqlite row. We did something similar in pathfinder [here](https://github.com/eqlabs/pathfinder/blob/7ec71dcbc35372a4424f4772b3a02ecfe7e19c5c/crates/storage/src/params.rs#L151) which I think added quite a lot in terms of consistency and readability.

Mirko-von-Leipzig (approved):
LGTM

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -88,7 +85,7 @@ impl TransactionBatch {
     #[instrument(target = "miden-block-producer", name = "new_batch", skip_all, err)]
     pub fn new(
         txs: Vec<ProvenTransaction>,
-        found_unauthenticated_notes: BTreeMap<NoteId, MerklePath>,
+        found_unauthenticated_notes: BTreeMap<NoteId, NoteInclusionProof>,
```
bobbinth, line 88: Related to the previous comments: I don't think `BTreeMap<NoteId, NoteInclusionProof>` would be sufficient to capture all required information. We'll probably need to create a separate struct which contains all additional info (i.e., block headers and MMR proofs).

The questions is what to call this struct. One option is `UnauthenticatedNoteWitness` - but I don't love this name.

bobbinth (changes requested):
Thank you! Looks good. Not a full review but I did leave some comments inline. The main one is I think just `NoteInclusionProof` objects are not sufficient we are missing the MMR part of the proofs.

proto/rpc.proto
```diff
@@ -12,6 +12,7 @@ service Api {
     rpc GetAccountStateDelta(requests.GetAccountStateDeltaRequest) returns (responses.GetAccountStateDeltaResponse) {}
     rpc GetBlockByNumber(requests.GetBlockByNumberRequest) returns (responses.GetBlockByNumberResponse) {}
     rpc GetBlockHeaderByNumber(requests.GetBlockHeaderByNumberRequest) returns (responses.GetBlockHeaderByNumberResponse) {}
+    rpc GetNoteInclusionProofs(requests.GetNoteInclusionProofsRequest) returns (responses.GetNoteInclusionProofsResponse) {}
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Removed.

proto/responses.proto
```diff
@@ -119,6 +119,11 @@ message GetNotesByIdResponse {
     repeated note.Note notes = 1;
 }
 
+message GetNoteInclusionProofsResponse {
+    // Lists note inclusion proofs returned by the database
+    repeated note.NoteInclusionProof proofs = 1;
+}
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I've updated this to include block MMR proofs.

proto/responses.proto
```diff
@@ -88,8 +88,8 @@ message GetBlockInputsResponse {
     // The requested nullifiers and their authentication paths
     repeated NullifierBlockInputRecord nullifiers = 4;
 
-    // The list of requested notes which were found in the database
-    repeated digest.Digest found_unauthenticated_notes = 5;
+    // The list of requested unauthenticated notes with their inclusion proofs which were found in the database
+    repeated note.NoteInclusionProof found_unauthenticated_notes = 5;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I've ignored this for now, I'll update this once the other endpoint's api is set.

crates/store/src/db/sql.rs
Mirko-von-Leipzig: Created #442

crates/block-producer/src/batch_builder/batch.rs
Mirko-von-Leipzig: Ignored for now

crates/store/src/db/mod.rs
```diff
@@ -224,6 +224,24 @@ impl Db {
             })?
     }
 
+    /// Loads multiple block headers from the DB.
+    #[instrument(target = "miden-store", skip_all, ret(level = "debug"), err)]
+    pub async fn select_many_block_headers(
```
Mirko-von-Leipzig, line 229: Or maybe `multiple_` instead? I also considered overloading the `select_block_headers` to take an optional vec instead.

crates/store/src/state.rs
```diff
@@ -397,6 +403,51 @@ impl State {
...
+        // We also avoid accessing the db in the block as this would delay
+        // dropping the guard.
+        let merkle_paths = {
+            let state = self.inner.read().await;
+            let forest = state.chain_mmr.forest();
+
+            blocks
+                .iter()
+                .map(|block_num| {
+                    let proof = state.chain_mmr.open(todo!(), todo!())?.merkle_path;
```
Mirko-von-Leipzig, line 431: How do I convert an absolute index into a tree + offset index as required by `open`?

My brain tells me this is simple; but for some reason I'm glitching here. Probably just some modulo variation? Or does one have to go bit searching?

crates/store/src/types.rs
```diff
@@ -1,2 +1,12 @@
...
 pub type BlockNumber = u32;
 pub type AccountId = u64;
+
+/// Data required to verify a block's MMR proof.
+pub struct BlockMmrProof {
+    pub block_number: BlockNumber,
+    pub mmr_path: MerklePath,
+    pub sub_hash: Digest,
+    pub note_root: Digest,
+}
```
Mirko-von-Leipzig, line 12: Is this living here okay?

crates/store/src/state.rs
```diff
@@ -397,6 +403,54 @@ impl State {
...
 
+    /// Queries all the note inclusion proofs matching a certain Note IDs from the database.
+    pub async fn get_note_inclusion_proofs(
+        &self,
+        note_ids: Vec<NoteId>,
+    ) -> Result<BTreeMap<NoteId, NoteInclusionProof>, DatabaseError> {
+        self.db.select_block_note_inclusion_proofs(note_ids).await
+    }
+
+    pub async fn get_block_mmr_proofs(
```
bobbinth, line 414: I would rename this into `get_block_inclusion_proofs()`.

crates/proto/src/domain/mod.rs
```diff
@@ -9,18 +9,20 @@ pub mod transactions;
...
 
-pub fn convert<T, From, To>(from: T) -> Vec<To>
+pub fn convert<T, From, To, R>(from: T) -> R
 where
     T: IntoIterator<Item = From>,
     From: Into<To>,
+    R: FromIterator<To>,
 {
     from.into_iter().map(|e| e.into()).collect()
 }
```
bobbinth, line 19: Similar to one of the previous comments, maybe this could be something like:
```rust
pub fn convert<From, To>(from: impl IntoIterator<Item = From>) -> impl Iterator<Item = To>
where
    From: Into<To>
```
This way, the caller would be responsible for calling `.collect()` or could use the returned iterator in some other ways.

Though, I'm wondering if we are getting much convenience from this function at all (and maybe it make sense to remove it). Though, this would probably be for another PR.

crates/proto/src/domain/notes.rs
```diff
@@ -1,17 +1,21 @@
...
+    crypto::hash::rpo::RpoDigest,
+    notes::{NoteId, NoteInclusionProof, NoteMetadata, NoteTag, NoteType},
     Felt,
 };
 
-use crate::errors::{ConversionError, MissingFieldHelper};
+use crate::{
+    errors::{ConversionError, MissingFieldHelper},
+    generated::note::{NoteInclusionProof as NoteInclusionProofPb, NoteMetadata as NoteMetadataPb},
+};
```
bobbinth, line 10: nit (and not for this PR): I think it would be cleaner to add `Proto` suffix instead of `Pb` to the aliases. For example, `NoteMetadata as NoteMetadataProto`.

crates/block-producer/src/store/mod.rs
```diff
@@ -56,15 +55,15 @@ pub trait Store: ApplyBlock {
...
     ///
-    /// TODO: right now this return only Merkle paths per note, but this will need to be updated to
-    /// return full authentication info.
+    /// TODO: right now this returns only Merkle paths per note, but this will need to be updated to
+    ///       return full authentication info.
     async fn get_note_authentication_info(
         &self,
         notes: impl Iterator<Item = &NoteId> + Send,
-    ) -> Result<BTreeMap<NoteId, MerklePath>, NotePathsError>;
+    ) -> Result<BTreeMap<NoteId, NoteInclusionProof>, NotePathsError>;
```
bobbinth, line 66: Maybe we can use a new struct for the return type here - something like:
```rust
pub struct NoteAuthenticationInfo {
    block_proofs: Vec<BlockInclusionProof>,
    note_proofs: BTreeMap<NoteId, NoteInclusionProof>,
}
```
This can the also be used in the `BlockInputs` struct.

bobbinth (changes requested):
Looks good! Thank you! I left some comments inline (and I think answered all outstanding questions - but let me know if I missed something).

crates/store/src/state.rs
bobbinth: Opened https://github.com/0xPolygonMiden/crypto/pull/324 to improve this in the future.

crates/proto/src/domain/mod.rs
Mirko-von-Leipzig: I suspect removal and handing control back to the caller ito allocations and `iter().map(..)` chaining makes the most sense. Though I don't have the full sense of the ergonomics here yet.

crates/proto/src/domain/notes.rs
Mirko-von-Leipzig: I'll see if it doesn't make more sense to alias the module instead

```rust
use crate::generated::note as proto;

impl From<T> for proto::T {...}
```

crates/store/src/server/api.rs
```diff
@@ -255,6 +256,54 @@ impl api_server::Api for StoreApi {
...
+        request: tonic::Request<GetNoteInclusionProofsRequest>,
+    ) -> Result<Response<GetNoteInclusionProofsResponse>, Status> {
+        info!(target: COMPONENT, ?request);
+
+        let note_ids = request.into_inner().note_ids;
+
+        let note_ids: Vec<RpoDigest> = try_convert(note_ids)
+            .map_err(|err| Status::invalid_argument(format!("Invalid NoteId: {}", err)))?;
+
+        let note_ids: Vec<NoteId> = note_ids.into_iter().map(From::from).collect();
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: #449

crates/store/src/db/sql.rs
```diff
@@ -718,6 +770,31 @@ pub fn select_block_header_by_block_num(
...
+    let mut result = Vec::with_capacity(blocks.len());
+
+    for block in blocks {
+        let header: Vec<u8> = stmt
+            .query_row([block], |row| row.get(0))
+            .optional()?
+            .ok_or(DatabaseError::BlockNotFoundInDb(block))?;
+        let header = BlockHeader::read_from_bytes(&header)?;
+        result.push(header);
+    }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: You're right, but I had concerns with the output not aligning with the input since the input is a slice and not a set.

And selecting using `IN` would return a set.

I'll see if I can't improve this and the function parameters.

crates/store/src/server/api.rs
```diff
@@ -255,6 +256,54 @@ impl api_server::Api for StoreApi {
...
+        // Massage into shape required by protobuf
+        let note_proofs = note_proofs.iter().map(Into::into).collect();
+        let block_proofs = block_proofs
+            .into_iter()
+            .map(|proof| BlockMmrProof {
+                block_num: proof.block_number,
+                sub_hash: Some(proof.sub_hash.into()),
+                note_root: Some(proof.note_root.into()),
+                mmr_path: Some((&proof.mmr_path).into()),
+            })
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Do you mean the `proto/domain` module, or the store's `state` module? If the former then we also need to move this type into `proto` or `miden-objects`.f

I'm going to assume the latter for now.

crates/proto/src/domain/notes.rs
Mirko-von-Leipzig: Ended up using `generated` for the most part.

crates/store/src/db/sql.rs
```diff
@@ -665,6 +665,58 @@ pub fn select_notes_by_id(conn: &mut Connection, note_ids: &[NoteId]) -> Result<
...
+/// Select note inclusion proofs matching the NoteId, using the given [Connection].
+///
+/// # Returns
+///
+/// - Empty map if no matching `note`.
+/// - Otherwise, note inclusion proofs, which `note_id` matches the `NoteId` as bytes.
+pub fn select_block_note_inclusion_proofs(
+    conn: &mut Connection,
+    note_ids: &[NoteId],
+) -> Result<BTreeMap<NoteId, NoteInclusionProof>> {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I've changed the input to `BTreeSet` instead, due to concerns given [here](https://github.com/0xPolygonMiden/miden-node/pull/421#discussion_r1723223858).

crates/store/src/db/mod.rs
```diff
@@ -307,6 +325,25 @@ impl Db {
             })?
     }
 
+    /// Loads all the note inclusion proofs matching a certain Note IDs from the database,
+    /// grouped by block.
+    #[instrument(target = "miden-store", skip_all, ret(level = "debug"), err)]
+    pub async fn select_block_note_inclusion_proofs(
+        &self,
+        note_ids: Vec<NoteId>,
+    ) -> Result<BTreeMap<NoteId, NoteInclusionProof>> {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: https://github.com/0xPolygonMiden/miden-node/pull/421#discussion_r1723223858

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: I'm busy implementing this; but I'm starting to feel like we should have split this PR up much more.

e.g. add this functionality to the store api, and only handle consuming it in the block producer in a separate PR.

crates/proto/src/domain/notes.rs
bobbinth: I actually prefer adding type aliases as I think it makes the code a bit easier to read - e.g., instead of `generated::merkle::MerklePath` we get `MerklePathProto`. But let's keep things as they currently are in this PR.

crates/store/src/db/sql.rs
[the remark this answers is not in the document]
bobbinth: You mean that `blocks` parameter could contain duplicate block numbers? I think we should think of this as an error - though not sure we should check for it (at least at this level).

I'm also not sure we should enforce that for every passed-in block number there must be a record in the DB. So, I think this is OK that this function returns a vector of different size as compared to the passed-in vector?

Basically, the way I think of this function is that given a list of block numbers, it returns a set of matching block headers from the database. But it can also return no results if none block headers match.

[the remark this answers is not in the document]
bobbinth: And to add: it would then be a responsibility of the caller to make sense of these results.

crates/store/src/db/sql.rs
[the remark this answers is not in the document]
bobbinth: One potential concern I have with using `BTreeMap` and `BTreeSet` where we can avoid them is performance (inserting into these should be a bit more expensive than inserting into a vector). But maybe that's a case of premature optimization.

But in general, I'm not sure we need to align the input/output types here - there is no guarantee here that for every passed-in note ID we'll get an inclusion proof back. So, I think it is fine to pass in a slice of notes and get back a vector of results (possibility of different length and in different order). But maybe I'm missing something?

crates/store/src/server/api.rs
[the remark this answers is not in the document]
bobbinth: I meant moving `BlockMmrProof` (or what is now `BlockInclusionProof`) into `proto/domain` as I assumed we'll need to use it in the `block-producer` crate as well. But if it is only used here, we can keep it in this module.

crates/store/src/state.rs
```diff
@@ -87,6 +94,21 @@ struct InnerState {
     account_tree: SimpleSmt<ACCOUNT_TREE_DEPTH>,
 }
 
+/// Data required to verify a block's inclusion proof.
+pub struct BlockInclusionProof {
+    pub block_header: BlockHeader,
+    pub mmr_path: MerklePath,
+}
+
+impl From<BlockInclusionProof> for miden_node_proto::generated::block::BlockInclusionProof {
```
bobbinth, line 103: Following one of the above comments: I think it would be a bit easier to read this line if it was:
```rust
impl From<BlockInclusionProof> for BlockInclusionProofProto
```
or
```rust
impl From<BlockInclusionProof> for proto::BlockInclusionProof
```

crates/block-producer/src/store/mod.rs
bobbinth: I'm fine with splitting this into multiple PRs (I too find that it has become a bit difficult to follow) and doing `store` part in one PR and `block-producer` part in another makes sense to me.

crates/store/src/state.rs
Mirko-von-Leipzig: Ah yeah my `grep` missed this one. Thanks.

crates/store/src/db/sql.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: I think I just got lost along the way and forgot that the input and output set are already misaligned ie not all inputs have an output. I'll change it to vectors when I get a chance again.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -141,10 +139,11 @@ impl TransactionBatch {
...
                     // except the nullifier)
-                    found_unauthenticated_notes
-                        .get(&input_note_header.id())
-                        .map(|_path| InputNoteCommitment::from(input_note.nullifier()))
-                        .unwrap_or_else(|| input_note.clone())
+                    if found_unauthenticated_notes.contains_note(&input_note_header.id()) {
+                        InputNoteCommitment::from(input_note.nullifier())
+                    } else {
+                        input_note.clone()
+                    }
```
Mirko-von-Leipzig, line 146: What isn't clear to me is where/what the note inclusion proof should be used for?

crates/block-producer/src/test_utils/block.rs
```diff
@@ -22,7 +22,14 @@ pub async fn build_expected_block_header(
...
 ) -> BlockHeader {
-    let last_block_header = *store.last_block_header.read().await;
+    let last_block_header = *store
+        .block_headers
+        .read()
+        .await
+        .iter()
+        .max_by_key(|(block_num, _)| *block_num)
+        .unwrap()
+        .1;
```
Mirko-von-Leipzig, line 32: This is really awkward to do everywhere, but I didn't think its safe to extract this as a method since it holds a lock. So calling it elsewhere could easily result in an accidental dead-lock.

Mirko-von-Leipzig (commented):
I've rebased and I *think* did the block producer portion as well.

proto/store.proto
```diff
@@ -16,6 +16,7 @@ service Api {
     rpc GetBlockByNumber(requests.GetBlockByNumberRequest) returns (responses.GetBlockByNumberResponse) {}
     rpc GetBlockHeaderByNumber(requests.GetBlockHeaderByNumberRequest) returns (responses.GetBlockHeaderByNumberResponse) {}
     rpc GetBlockInputs(requests.GetBlockInputsRequest) returns (responses.GetBlockInputsResponse) {}
+    rpc GetNoteInclusionProofs(requests.GetNoteInclusionProofsRequest) returns (responses.GetNoteInclusionProofsResponse) {}
```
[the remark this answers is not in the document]
bobbinth: Yes, correct - this endpoint is only for internal usage.

proto/block.proto
```diff
@@ -25,3 +26,8 @@ message BlockHeader {
     // the time when the block was created.
     fixed32 timestamp = 10;
 }
+
+message BlockInclusionProof {
+    BlockHeader block_header = 1;
+    merkle.MerklePath mmr_path = 2;
+}
```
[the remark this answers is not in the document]
bobbinth: Not a strong preference, but it does feel odd having this message not be "self-contained". If we do want to keep it like that, we should probably add some comments explaining why. But I also don't think it is a big deal if we add a `chain_length` field here - yes, it'll be duplicated, but I think the implications of this are negligible.

proto/responses.proto
```diff
@@ -136,6 +136,10 @@ message GetNotesByIdResponse {
     repeated note.Note notes = 1;
 }
 
+message GetNoteInclusionProofsResponse {
+    note.NoteInclusionProofs proofs = 1;
+}
```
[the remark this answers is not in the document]
bobbinth: Yeah - I think I suggested it somewhere. I think `NoteAuthenticationInfo` may be a bit better (one reason is that we use it in the context of authenticating "unauthenticated notes") - but also not a strong preference by any means.

crates/proto/src/domain/blocks.rs
```diff
@@ -1,14 +1,14 @@
-use miden_objects::BlockHeader;
+use miden_objects::{crypto::merkle::MerklePath, BlockHeader};
 
 use crate::{
     errors::{ConversionError, MissingFieldHelper},
-    generated::block_header,
+    generated::block,
```
bobbinth, line 5: It is a bit difficult to tell what's what by looking at `block::BlockHeader` types below. I would probably alias this as:
```rust
generated::block as proto
```
And then everywhere below we could say just `proto::BlockHeader`.

crates/store/src/state.rs
```diff
@@ -87,6 +96,43 @@ struct InnerState {
     account_tree: SimpleSmt<ACCOUNT_TREE_DEPTH>,
 }
 
+#[derive(Clone, Default, Debug)]
+pub struct NoteAuthenticationInfo {
+    pub block_proofs: Vec<BlockInclusionProof>,
+    pub note_proofs: BTreeMap<NoteId, NoteInclusionProof>,
+    pub chain_length: BlockNumber,
+}
```
bobbinth, line 104: Not for this PR, but this should actually be in the domain objects of `miden-proto` crate (this is because this struct is used by both store and block producer). If we do that, we won't need to add `miden-store` as a dependency to `miden-block-producer` (I think).

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -114,7 +114,9 @@ where
             .await?;
 
         let missing_notes: Vec<_> = dangling_notes
-            .difference(&block_inputs.found_unauthenticated_notes)
+            .difference(
+                &block_inputs.found_unauthenticated_notes.note_proofs.keys().copied().collect(),
```
bobbinth, line 118: I would probably add a method on `NoteAuthenticationInfo` to make this a single method invocation (e.g., maybe something like `NoteAuthenticationInfo::note_ids()`).

crates/block-producer/src/batch_builder/tests/mod.rs
```diff
@@ -260,12 +261,19 @@ async fn test_block_builder_no_missing_notes() {
 async fn test_block_builder_fails_if_notes_are_missing() {
     let accounts: Vec<_> = (1..=4).map(MockPrivateAccount::<3>::from).collect();
     let notes: Vec<_> = (1..=6).map(mock_note).collect();
+    // We require mmr for the note authentication to succeed.
+    //
+    // We also need two blocks worth of mmr because the mock store skips genesis.
+    let mut mmr = Mmr::new();
+    mmr.add(Digest::new([1u32.into(), 2u32.into(), 3u32.into(), 4u32.into()]));
+    mmr.add(Digest::new([1u32.into(), 2u32.into(), 3u32.into(), 4u32.into()]));
```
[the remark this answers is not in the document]
bobbinth: We should probably start thinking about how to refactor our testing framework.

crates/block-producer/src/batch_builder/batch.rs
bobbinth: They are not currently used, but once we have batch/block kernels, they will be used in them for proving that these notes are actually authenticated.

bobbinth (approved):
Looks good! Thank you! I left a few small comments inline. Once these are addressed, we can merge.

I also left one comment for a subsequent PR. I think we should do it right after this PR (it shouldn't be too complex).

crates/store/src/state.rs
Mirko-von-Leipzig: #455

proto/responses.proto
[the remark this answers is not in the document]
Mirko-von-Leipzig: I've made this change -- should the proto method name change as well to `GetNoteAuthenticationInfo`?

I'm still a bit murky on the nomenclature

proto/block.proto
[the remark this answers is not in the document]
Mirko-von-Leipzig: I've made this change; I like local reasoning.

Mirko-von-Leipzig:
Changes made - one final naming question [here](https://github.com/0xPolygonMiden/miden-node/pull/421#discussion_r1728703554) and then we're good to go I think.

proto/responses.proto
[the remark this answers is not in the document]
bobbinth: I could go either way - so, basically, up to you :)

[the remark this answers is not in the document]
Mirko-von-Leipzig: changed it so that it isn't the only outlier :)

PR #424, opened 2024-07-31
feat(rpc): `SyncNotes` endpoint

### LINKED ISSUE #409: Endpoint for note syncing
As described in https://github.com/0xPolygonMiden/miden-client/issues/405#issuecomment-2229651865, it may be beneficial to have an endpoint which works similar to the `SyncState` but is limited only to retrieving note data. 

The endpoint could be called `NoteSync` and could look as follows:

#### Parameters
- `block_num`: `uint32` – send notes created starting at this block.
- `note_tags`: `[uint32]` – send notes matching these tags.

#### Returns

- `mmr_delta`: `MmrDelta` – data needed to update the partial MMR from `request.block_num + 1` to `response.block_header.block_num`.
- `notes`: `[NoteSyncRecord]` – a list of all notes together with the Merkle paths from `response.block_header.note_root`.

One open question is how exactly we should handle the `mmr_delta` here (or if there is a better way to send this info to the client).

Mirko-von-Leipzig:
Closes #409.

The implementation is essentially a clone of SyncState with some parts left off. Which makes me believe I've misunderstood the assignment :)

### DIFF
--- a/crates/rpc-proto/proto/responses.proto
+++ b/crates/rpc-proto/proto/responses.proto
@@ -69,8 +69,8 @@ message SyncNoteResponse {
     // Block header of the block with the first note matching the specified criteria
     block_header.BlockHeader block_header = 2;
 
-    // Data needed to update the partial MMR from `request.block_num + 1` to `response.block_header.block_num`
-    mmr.MmrDelta mmr_delta = 3;
+    // Proof for block header's MMR with respect to the chain tip.
+    merkle.MerklePath mmr_path = 3;
 
     // List of all notes together with the Merkle paths from `response.block_header.note_root`
     repeated note.NoteSyncRecord notes = 4;
--- a/crates/rpc-proto/proto/responses.proto
+++ b/crates/rpc-proto/proto/responses.proto
@@ -62,6 +62,20 @@ message SyncStateResponse {
     repeated NullifierUpdate nullifiers = 8;
 }
 
+message SyncNoteResponse {
+    // Number of the latest block in the chain
+    fixed32 chain_tip = 1;
+
+    // Block header of the block with the first note matching the specified criteria
+    block_header.BlockHeader block_header = 2;
+
+    // Data needed to update the partial MMR from `request.block_num + 1` to `response.block_header.block_num`
+    mmr.MmrDelta mmr_delta = 3;
+
+    // List of all notes together with the Merkle paths from `response.block_header.note_root`
+    repeated note.NoteSyncRecord notes = 4;
+}
+
 // An account returned as a response to the GetBlockInputs
 message AccountBlockInputRecord {
     account.AccountId account_id = 1;
--- a/crates/rpc-proto/proto/responses.proto
+++ b/crates/rpc-proto/proto/responses.proto
@@ -70,6 +70,11 @@ message SyncNoteResponse {
     block_header.BlockHeader block_header = 2;
 
     // Proof for block header's MMR with respect to the chain tip.
+    //
+    // More specifically, the full proof consists of `forest`, `position` and `path` components. This
+    // value constitutes the `path`. The other two components can be obtained as follows: 
+    //   - `position` is simply `resopnse.block_header.block_num`
+    //   - `forest` is the same as `response.chain_tip + 1`
     merkle.MerklePath mmr_path = 3;
 
     // List of all notes together with the Merkle paths from `response.block_header.note_root`
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -91,6 +91,13 @@ pub struct StateSyncUpdate {
     pub nullifiers: Vec<NullifierInfo>,
 }
 
+#[derive(Debug, PartialEq)]
+pub struct NoteSyncUpdate {
+    pub notes: Vec<NoteRecord>,
+    pub block_header: BlockHeader,
+    pub chain_tip: BlockNumber,
+}
+
 impl Db {
     /// Open a connection to the DB, apply any pending migrations, and ensure that the genesis block
     /// is as expected and present in the database.
@@ -286,6 +293,25 @@ impl Db {
             .await
             .map_err(|err| {
                 DatabaseError::InteractError(format!("Get state sync task failed: {err}"))
+            })?
+    }
+
+    #[instrument(target = "miden-store", skip_all, ret(level = "debug"), err)]
+    pub async fn get_note_sync(
+        &self,
+        block_num: BlockNumber,
+        note_tag_prefixes: &[u32],
+    ) -> Result<NoteSyncUpdate, StateSyncError> {
+        let note_tag_prefixes = note_tag_prefixes.to_vec();
+
+        self.pool
+            .get()
+            .await
+            .map_err(DatabaseError::MissingDbConnection)?
+            .interact(move |conn| sql::get_note_sync(conn, block_num, &note_tag_prefixes))
+            .await
+            .map_err(|err| {
+                DatabaseError::InteractError(format!("Get notes sync task failed: {err}"))
             })?
     }
 
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -26,7 +26,7 @@ use crate::{
     blocks::BlockStore,
     config::StoreConfig,
     db::migrations::apply_migrations,
-    errors::{DatabaseError, DatabaseSetupError, GenesisError, StateSyncError},
+    errors::{DatabaseError, DatabaseSetupError, GenesisError, NoteSyncError, StateSyncError},
     genesis::GenesisState,
     types::{AccountId, BlockNumber},
     COMPONENT,
@@ -291,7 +291,7 @@ impl Db {
         &self,
         block_num: BlockNumber,
         note_tags: Vec<u32>,
-    ) -> Result<NoteSyncUpdate, StateSyncError> {
+    ) -> Result<NoteSyncUpdate, NoteSyncError> {
         self.pool
             .get()
             .await
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -269,26 +269,16 @@ impl Db {
     pub async fn get_state_sync(
         &self,
         block_num: BlockNumber,
-        account_ids: &[AccountId],
-        note_tag_prefixes: &[u32],
-        nullifier_prefixes: &[u32],
+        account_ids: Vec<AccountId>,
+        note_tags: Vec<u32>,
+        nullifier_prefixes: Vec<u32>,
     ) -> Result<StateSyncUpdate, StateSyncError> {
-        let account_ids = account_ids.to_vec();
-        let note_tag_prefixes = note_tag_prefixes.to_vec();
-        let nullifier_prefixes = nullifier_prefixes.to_vec();
-
         self.pool
             .get()
             .await
             .map_err(DatabaseError::MissingDbConnection)?
             .interact(move |conn| {
-                sql::get_state_sync(
-                    conn,
-                    block_num,
-                    &account_ids,
-                    &note_tag_prefixes,
-                    &nullifier_prefixes,
-                )
+                sql::get_state_sync(conn, block_num, &account_ids, &note_tags, &nullifier_prefixes)
             })
             .await
             .map_err(|err| {
@@ -300,15 +290,13 @@ impl Db {
     pub async fn get_note_sync(
         &self,
         block_num: BlockNumber,
-        note_tag_prefixes: &[u32],
+        note_tags: Vec<u32>,
     ) -> Result<NoteSyncUpdate, StateSyncError> {
-        let note_tag_prefixes = note_tag_prefixes.to_vec();
-
         self.pool
             .get()
             .await
             .map_err(DatabaseError::MissingDbConnection)?
-            .interact(move |conn| sql::get_note_sync(conn, block_num, &note_tag_prefixes))
+            .interact(move |conn| sql::get_note_sync(conn, block_num, &note_tags))
             .await
             .map_err(|err| {
                 DatabaseError::InteractError(format!("Get notes sync task failed: {err}"))
--- a/crates/store/src/db/sql.rs
+++ b/crates/store/src/db/sql.rs
@@ -964,9 +964,9 @@ pub fn get_state_sync(
 pub fn get_note_sync(
     conn: &mut Connection,
     block_num: BlockNumber,
-    note_tag_prefixes: &[u32],
+    note_tags: &[u32],
 ) -> Result<NoteSyncUpdate, StateSyncError> {
-    let notes = select_notes_since_block_by_tag(conn, note_tag_prefixes, block_num)?;
+    let notes = select_notes_since_block_by_tag(conn, note_tags, block_num)?;
 
     let (block_header, chain_tip) = if !notes.is_empty() {
         let block_header = select_block_header_by_block_num(conn, Some(notes[0].block_num))?
--- a/crates/store/src/db/sql.rs
+++ b/crates/store/src/db/sql.rs
@@ -521,9 +521,9 @@ pub fn insert_notes(transaction: &Transaction, notes: &[NoteRecord]) -> Result<u
 ///
 /// # Returns
 ///
-/// - Empty vector if no tag created after `block_num` match `tags` or `account_ids`.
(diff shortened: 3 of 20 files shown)


proto/requests.proto
```diff
@@ -68,6 +68,25 @@ message SyncStateRequest {
     repeated uint32 nullifiers = 4;
 }
 
+// State synchronization request.
+//
+// Specifies state updates the client is intersted in. The server will return the first block which
+// contains a note matching `note_tags` or the chain tip. And the corresponding updates to
+// `nullifiers` and `account_ids` for that block range.
+message SyncNoteRequest {
```
bobbinth, line 76: The comments here needs to be updated.

proto/requests.proto
```diff
@@ -68,6 +68,25 @@ message SyncStateRequest {
...
+    // Last block known by the client. The response will contain the data starting from the next block,
+    // until the 
+    fixed32 block_num = 1;
```
bobbinth, line 79: The comment here seems incomplete.

proto/requests.proto
```diff
@@ -68,6 +68,25 @@ message SyncStateRequest {
...
+
+    // Determines the tags which the client is interested in. These are only the 16high bits of the
+    // note's complete tag.
+    //
+    // The above means it is not possible to request an specific note, but only a "note family",
+    // this is done to increase the privacy of the client, by hiding the note's the client is
+    // intereted on.
+    repeated uint32 note_tags = 3;
```
bobbinth, line 87: I don't think this comment is accurate (this applies to the state sync endpoint as well). Specifically, note tags are now always 32 bits.

This also makes me thing that we should changed `uint32` to `fixed32`.

crates/store/src/server/api.rs
```diff
@@ -220,6 +220,45 @@ impl api_server::Api for StoreApi {
...
+        &self,
+        request: tonic::Request<SyncNoteRequest>,
+    ) -> Result<Response<SyncNoteResponse>, Status> {
+        let request = request.into_inner();
+
+        let (state, delta) = self
+            .state
+            .sync_state(request.block_num, &[], &request.note_tags, &[])
+            .await
+            .map_err(internal_error)?;
```
bobbinth, line 241: While this works, I think we should have a dedicated method which does not make extra requests to the database or maybe modify the existing method to avoid extra requests if `account_ids` and `nullifier_prefixes` are not provided. Could be done in a follow-up PR though.

bobbinth (commented):

Overall, I think this looks right - but there are two questions I have and probably @igamigo and @tomyrd are the ones best positioned to answer them:

- Do we need to include the `transactions` field in the response (similar to how we have it for state sync). It seems to me that we don't - but I also haven't thought it through completely.
- Would the client be able to use the MMR info returned from this endpoint to update its MMR correctly? On the client side, this will probably work somewhat differently from how state sync works because this endpoint would be invoked at the time when the client's MMR is "in the future" compared to the requested data.

proto/requests.proto
Mirko-von-Leipzig: I can make that change but it will be a breaking on iiuc. 

Just to make sure I understand - note tags used to be the 16 high bits, but are now the 32 high bits?

bobbinth: Yeah, a breaking change is fine as I think we already assume on the client that note tags are 32 bits.

At some point in the past we did take only the high 16 bits of the tags in the request (at that point, tags were 64 bits, I believe). But now we've changed this so that tags are always 32 bits and we always request things by the exact tag (so, no need to deal with prefixes any more).

A bit more info about note tags is [here](https://github.com/0xPolygonMiden/miden-base/blob/ab7e030c3df76c8e1fe06ad4e747c28dd13f8a13/objects/src/notes/note_tag.rs#L67).

crates/store/src/server/api.rs
Mirko-von-Leipzig: I've implemented this with all of the lower down pieces - I was surprised at how many parts there were to add and invoke a single db query.

crates/store/src/db/sql.rs
```diff
@@ -604,6 +606,84 @@ pub fn select_notes_since_block_by_tag_and_sender(
...
+        FROM
+            notes
+        -- find the next block which contains at least one note with a matching tag
+        WHERE
+            tag IN rarray(?1) AND
+            block_num > ?2
+        ORDER BY
+            block_num ASC
+        LIMIT
+            1;
```
Mirko-von-Leipzig, line 650: The similar query in `select_notes_since_block_by_tag_and_sender` above contains a level of nesting -- I'm not sure why the nested select is there?

crates/store/src/db/mod.rs
```diff
@@ -289,6 +296,25 @@ impl Db {
             })?
     }
 
+    #[instrument(target = "miden-store", skip_all, ret(level = "debug"), err)]
+    pub async fn get_note_sync(
+        &self,
+        block_num: BlockNumber,
+        note_tag_prefixes: &[u32],
+    ) -> Result<NoteSyncUpdate, StateSyncError> {
+        let note_tag_prefixes = note_tag_prefixes.to_vec();
```
Mirko-von-Leipzig, line 305: I've noticed this pattern of taking by reference/slice and then immediately cloning. Bit of an anti-pattern imo

proto/requests.proto
Mirko-von-Leipzig: Does that not negate the privacy aspect? Or is the hope now that there are many notes with a matching tag?

bobbinth: Yes, the idea is that there will be many notes matching a specific tag. And the way it currently works, when a note is directed to a specific account, only the first 16 bits of the account ID are included in the tag (the rest are set to 0). So, not much privacy if we have only 10K accounts in the system - but as soon as there are a few million, things should become quite a bit more private.

crates/store/src/server/api.rs
bobbinth: Let's create an issue to refactor this.

igamigo:
> Do we need to include the transactions field in the response (similar to how we have it for state sync). It seems to me that we don't - but I also haven't thought it through completely.

I don't think this is necessary.

> Would the client be able to use the MMR info returned from this endpoint to update its MMR correctly? On the client side, this will probably work somewhat differently from how state sync works because this endpoint would be invoked at the time when the client's MMR is "in the future" compared to the requested data.

After reviewing it a little bit, I think returning a Merkle path here is better and makes for a cleaner handling of the response on the client side. More so if we assume the client is going to call the endpoint after having synced past the block of the note's creation.

bobbinth:
> After reviewing it a little bit, I think returning a Merkle path here is better and makes for a cleaner handling of the response on the client side. More so if we assume the client is going to call the endpoint after having synced past the block of the note's creation.

Yes - I came to a similar conclusion. Basically, we just need the `MmrProof` for the returned block header. And we can get this directly from the in-memory MMR.

proto/requests.proto
```diff
@@ -54,20 +54,28 @@ message SyncStateRequest {
...
 
-    // Determines the tags which the client is interested in. These are only the 16high bits of the
-    // note's complete tag.
-    //
-    // The above means it is not possible to request an specific note, but only a "note family",
-    // this is done to increase the privacy of the client, by hiding the note's the client is
-    // intereted on.
-    repeated uint32 note_tags = 3;
+    // Determines the tags which the client is interested in.
+    repeated fixed32 note_tags = 3;
```
bobbinth, line 58: nit: should we say "Specifies" rather than "Determines"?

proto/requests.proto
```diff
@@ -54,20 +54,28 @@ message SyncStateRequest {
...
+//
+// Specifies note tags that client is intersted in. The server will return the first block which
+// contains a note matching `note_tags` or the chain tip.
+message SyncNoteRequest {
+    // Last block known by the client. The response will contain data starting from the next block,
+    // until the first block which contains a note of matching the requested tag.
+    fixed32 block_num = 1;
+
+    // Determines the tags which the client is interested in.
+    repeated fixed32 note_tags = 2;
```
bobbinth, line 76: Same nit as above.

proto/responses.proto
```diff
@@ -62,6 +62,20 @@ message SyncStateResponse {
...
 
+message SyncNoteResponse {
+    // Number of the latest block in the chain
+    fixed32 chain_tip = 1;
+
+    // Block header of the block with the first note matching the specified criteria
+    block_header.BlockHeader block_header = 2;
+
+    // Data needed to update the partial MMR from `request.block_num + 1` to `response.block_header.block_num`
+    mmr.MmrDelta mmr_delta = 3;
```
bobbinth, line 73: Based on recent discussions, I think this should be just the `MmrProof` for the included block header against the latest chain MMR on the node.

crates/store/src/state.rs
```diff
@@ -451,6 +451,51 @@ impl State {
         Ok((state_sync, delta))
     }
 
+    /// Loads data to synchronize a client's notes.
+    ///
+    /// The client's request contains a list of tag prefixes, this method will return the first
+    /// block with a matching tag, or the chain tip. All the other values are filter based on this
+    /// block range.
```
bobbinth, line 458: nit: I would change "tag prefixes" to just "tags".

crates/store/src/state.rs
```diff
@@ -451,6 +451,51 @@ impl State {
...
+    ///
+    /// # Arguments
+    ///
+    /// - `block_num`: The last block *know* by the client, updates start from the next block.
+    /// - `note_tag_prefixes`: Only the 16 high bits of the tags the client is interested in, result
+    ///   will include notes with matching prefixes, the first block with a matching note determines
+    ///   the block range.
```
bobbinth, line 465: typo: `know` -> `known`.

Also the description needs to be updated to mention full tags rather than 16-bit prefixes.

crates/store/src/state.rs
```diff
@@ -451,6 +451,51 @@ impl State {
...
+    #[instrument(target = "miden-store", skip_all, ret(level = "debug"), err)]
+    pub async fn sync_notes(
+        &self,
+        block_num: BlockNumber,
+        note_tag_prefixes: &[u32],
```
bobbinth, line 470: Similar to the above comment: this should be just `note_tags`.

crates/store/src/db/sql.rs
bobbinth: Without nested select, wouldn't we just get a single row returned from here (the `LIMIT 1` would apply to the whole query, right?)?

What we want to do is get all relevant notes from the first block which hash at least one relevant note. So, we first need to find this block (this is done via the inner select) and then get all matching notes in this blocks (that's the outer select).

crates/store/src/db/sql.rs
```diff
@@ -877,6 +957,35 @@ pub fn get_state_sync(
...
 }
 
+// NOTE SYNC
+// ================================================================================================
+
+/// Loads the data necessary for a note sync.
+pub fn get_note_sync(
+    conn: &mut Connection,
+    block_num: BlockNumber,
+    note_tag_prefixes: &[u32],
```
bobbinth, line 967: nit: this should be `note_tags`.

crates/store/src/db/mod.rs
bobbinth: Agreed - if a function needs to clone something immediately into a vector, we should just pass it as a vector and let the caller do the cloning, if needed.

Also, we should rename `note_tag_prefixes` into just `note_tags`.

bobbinth (commented):

crates/store/src/db/sql.rs
Mirko-von-Leipzig: Ah. Yes, I had (incorrectly) assumed that the array contained all the notes. But of course the array is just all of the tags for a single note. Thanks!

crates/rpc-proto/proto/responses.proto
```diff
@@ -62,6 +62,20 @@ message SyncStateResponse {
...
+    // Block header of the block with the first note matching the specified criteria
+    block_header.BlockHeader block_header = 2;
+
+    // Proof for block header's MMR with respect to the chain tip.
+    merkle.MerklePath mmr_path = 3;
+
+    // List of all notes together with the Merkle paths from `response.block_header.note_root`
+    repeated note.NoteSyncRecord notes = 4;
+}
+
```
Mirko-von-Leipzig, line 78: Is there a rationale for specifying `optional` in some other responses? e.g. in `GetBlockHeaderByNumberResponse` some of these are optional.

And these would be none if there are no matching notes iiuc.

crates/store/src/state.rs
```diff
@@ -451,6 +450,34 @@ impl State {
...
+        block_num: BlockNumber,
+        note_tag_prefixes: Vec<u32>,
+    ) -> Result<(NoteSyncUpdate, MmrProof), NoteSyncError> {
+        let inner = self.inner.read().await;
+
+        let note_sync = self.db.get_note_sync(block_num, note_tag_prefixes).await?;
+
+        let mmr_proof = inner
+            .chain_mmr
+            .open(note_sync.block_header.block_num() as usize, inner.chain_mmr.forest())?;
```
Mirko-von-Leipzig, line 476: I just stole this from another method - is this correct?

crates/store/src/state.rs
```diff
@@ -451,6 +450,34 @@ impl State {
         Ok((state_sync, delta))
     }
 
+    /// Loads data to synchronize a client's notes.
+    ///
+    /// The client's request contains a list of tag prefixes, this method will return the first
```
bobbinth, line 455: "tag prefixes" should be changed to just "tags"

crates/store/src/state.rs
```diff
@@ -451,6 +450,34 @@ impl State {
...
+    /// # Arguments
+    ///
+    /// - `block_num`: The last block *known* by the client, updates start from the next block.
+    /// - `note_tags`: The tags the client is interested in, resulting notes are restricted to
+    ///    the first block containing a matching note.
+    #[instrument(target = "miden-store", skip_all, ret(level = "debug"), err)]
+    pub async fn sync_notes(
+        &self,
+        block_num: BlockNumber,
+        note_tag_prefixes: Vec<u32>,
```
bobbinth, line 468: `note_tag_prefixes` should be just `note_tags`.

crates/store/src/state.rs
bobbinth: Yes, this is correct.

proto/responses.proto
```diff
@@ -62,6 +62,20 @@ message SyncStateResponse {
...
 
+message SyncNoteResponse {
+    // Number of the latest block in the chain
+    fixed32 chain_tip = 1;
+
+    // Block header of the block with the first note matching the specified criteria
+    block_header.BlockHeader block_header = 2;
+
+    // Proof for block header's MMR with respect to the chain tip.
+    merkle.MerklePath mmr_path = 3;
```
bobbinth, line 73: I would add some more comments here to explain the context of the path since MMR proof consists of 3 elements (`forest`, `position`, `path`). In our case:

- `forest` is the same as `response.chain_tip + 1` (though would be good to double-check this).
- `position` is the same as `response.block_header.block_num`.
- `path` is the `response.mmr_path`.

Most of this should be almost self-explanatory, but doesn't hurt to spell this out explicitly.

bobbinth (approved):
Looks good! Thank you! I left a few more minor comments inline. After these are addressed we can merge.

crates/rpc-proto/proto/responses.proto
bobbinth: I think we rarely have response fields as optional. In the specific case of `GetBlockHeaderByNumberResponse`, these are optional because we have an optional `include_mmr_proof` filed in the request. If this is set to `false` then, MMR proof-related fields are omitted from the response.

crates/store/src/db/sql.rs
```diff
@@ -604,6 +606,93 @@ pub fn select_notes_since_block_by_tag_and_sender(
     Ok(res)
 }
 
+/// Select notes matching the tag search criteria using the given [Connection].
+///
+/// # Returns
+///
+/// - Empty vector if no tag created after `block_num` match `tags`.
+/// - Otherwise, notes which the 16 high bits match `tags`, or the `sender` is one of the
+///   `account_ids`.
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I tried writing it out - my brain is a bit foggy today so let me know if you'd prefer wording instead.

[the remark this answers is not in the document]
bobbinth: I think it looks good.

bobbinth (approved):
All looks good! Thank you!

PR #426, opened 2024-08-01
ci: workflow improvements

Mirko-von-Leipzig:
This PR adds a few CI niceties and general improvements as listed below. PR can be viewed by commit to see each change in isolation. Some of these were briefly discussed in #425.

1. Remove default workspace members. Having default members causes `cargo build` et al without any target specified to only compile these default members. This can lead to confusion and brings little benefit.
2. Actually wrap comments at 100 characters. This includes actually applying the formatting and is the majority of the diff.
3. Limit the concurrency of each CI workflow to 1 per PR.
4. Cache compilation artifacts. I'll expand on my approach below.

### Caching

I've elected to cache only the `next` branch itself. Since _"all"_ branches begin and end with `next` this gives a solid cache hit ratio for each PR and greatly minimizes cache churn and space requirements.

Alternate strategies include caching each branch/PR which will improve CI times even more. But at a much greater storage cost. github limits the cache to 10GB after which cache churn will increase. I imagine our builds are still small enough that we could get away with it, but I think only caching `next` will give more than enough bang for buck unless you're dramatically changing dependencies.

This can only really be evaluated after using it for a while, and we can tweak it to a different strategy if required. Hard to judge how large the cache of a single workflow will be. The current strategy also has a cache per job, so it gets multiplied.


Mirko-von-Leipzig:
I'm unsure if these type of PRs should be added to the change log.. i.e. is the change log for us or for users who won't care about such internals.

crates/store/src/db/migrations.rs
```diff
@@ -100,8 +100,8 @@ fn compute_migration_hashes() -> Vec<Hash> {
 }
 
 fn preprocess_sql(sql: &str) -> String {
-    // TODO: We can also remove all comments here (need to analyze the SQL script in order to remain comments
-    //       in string literals).
+    // TODO: We can also remove all comments here (need to analyze the SQL script in order to remain
+    // comments       in string literals).
```
bobbinth, line 104: nit: some extra spaces on the second line.

bobbinth (approved):
Looks good! Thank you! I left just one nit inline - but also I'm far from the expert on this so my review is more like a rubber stamp here.

crates/store/src/db/migrations.rs
Mirko-von-Leipzig: Pity one can't customise rustfmt with additional rules like aligning indents for todo's etc. Thanks!

PR #427, opened 2024-08-01
feat(store): cache sql statements

### LINKED ISSUE #423: [Task]: Use cached sqlite statements

- [ ] RPC
- [ ] Block producer
- [X] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

We currently use `prepare` to create sqlite statements, however we should use [`prepare_cached` ](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.prepare_cached)as parsing a statement is a non-trivial cost.

### How should it be done?

Analyze the sql statements in `crates/store/src/db/sql.rs` and change every common one to use `prepare_cached`.

Exceptions should be made for statements which should be rare. Alternatively (and maybe additionally), we should [change the cache size](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.set_prepared_statement_cache_capacity) which [currently defaults to 16](https://docs.rs/rusqlite/latest/src/rusqlite/lib.rs.html#152) to something large enough to handle all statements..

### When is this task done?

All relevant `prepare` are changed to `prepare_cached` and the default cache size has been changed (or at least address why it should not be changed).

### Additional context

https://github.com/0xPolygonMiden/miden-node/pull/419#discussion_r1692742549

Mirko-von-Leipzig:
Replace all `conn.prepare` with `conn.prepare_cached` and double the cache size limit from the [default of 16](https://docs.rs/rusqlite/latest/src/rusqlite/lib.rs.html#152) to 32. We currently have 22 different statements so this should be adequate for a while.

Closes #423.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Added `GetAccountStateDelta` endpoint (#418).
 - Added `CheckNullifiersByPrefix` endpoint (#419).
 - Support multiple inflight transactions on the same account (#407).
+- Cache sql statements (#427).
 
 ### Fixes
 
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -29,7 +29,7 @@ use crate::{
     errors::{DatabaseError, DatabaseSetupError, GenesisError, StateSyncError},
     genesis::GenesisState,
     types::{AccountId, BlockNumber},
-    COMPONENT,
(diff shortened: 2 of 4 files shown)


polydez (approved):
I left only one nit comment, but overall looks good to me, thanks!

bobbinth (approved):
Looks good! Thank you!

PR #430, opened 2024-08-06
ci: auto-close linked issues

### LINKED ISSUE #431: Close linked issues when PR is merged
Github's builtin system for auto-closing linked issues on PR merge only works when merging into the default branch. Our working branch is `next` so unfortunately this auto-closing feature does not work.

It should be possible to add a workflow which does this automatically instead.

Mirko-von-Leipzig:
Playing with a workflow to close issues on merge into `next`.

Example action [here](https://medium.com/@martatatiana/github-projects-change-issue-status-based-on-pull-request-change-45dcacab9fb7).

Closes #431

### DIFF
--- a/.github/workflows/issue.yml
+++ b/.github/workflows/issue.yml
@@ -2,17 +2,19 @@
 # Runs changelog related jobs.
 # CI job heavily inspired by: https://github.com/tarides/changelog-check-action
 
-name: changelog
+name: auto-close
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig:
Interestingly this does work for the manually linked issue. But not for the text based one unfortunately, though this would be possible to parse manually so overall this could work. Not really super important though, so I'll park this for now :)

PR #434, opened 2024-08-06
fix: faucet build after changes in base

igamigo:


bobbinth (approved):
Looks good! Thank you!

PR #436, opened 2024-08-08
chore: update installation instructions

### LINKED ISSUE #355: [Task]: Update README to have a more comprehensive installation guide

- [ ] RPC
- [ ] Block producer
- [ ] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

Update the installation steps in the README.md.

### How should it be done?

Review and clarify the README.
Remove assumptions on file paths and OSs.

### When is this task done?

When the text is concise, correct, and easy to follow by a user

### Additional context

https://github.com/0xPolygonMiden/miden-node/pull/353#discussion_r1597512754

Mirko-von-Leipzig:
Closes #355

The various 
```
> #[!TIP]
> ...
```
annotations can be viewed by inspecting the whole file and not the diff which imo is a better way of reading this PR :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ miden-node make-genesis \
 which will create `genesis.dat` and an `accounts` directory containing account data based on the `genesis.toml` file.
 
 > [!NOTE]
-> `make-genesis` will take a long time if you're running the production version of `miden-node`, see the tip in (node installation section)[#installing-the-node].
+> `make-genesis` will take a long time if you're running the production version of `miden-node`, see the tip in the [installation](#installing-the-node) section.
 
 Modify the `miden-node.toml` configuration file such that the `[store]` paths point to our `<..>` folder:
 
--- a/README.md
+++ b/README.md
@@ -27,9 +27,15 @@ The diagram below illustrates high-level design of each component as well as bas
 
 ![Architecture diagram](./assets/architecture.png)
 
-## Usage
+## Installation
 
-Before you can build and run the Miden node or any of its components, you'll need to make sure you have Rust [installed](https://www.rust-lang.org/tools/install). Miden node requires Rust version **1.78** or later.
+Official releases are available as debian packages which can be found under our [releases](https://github.com/0xPolygonMiden/miden-node/releases) page.
+
+Alternatively, the Rust package manager `cargo` can be used to install on non-debian distributions or to compile from source.
+
+### Intall using `cargo`
+
+Install Rust version **1.78** or greater using the official Rust installation [instructions](https://www.rust-lang.org/tools/install).
 
 Depending on the platform, you may need to install additional libraries. For example, on Ubuntu 22.04 the following command ensures that all required libraries are installed.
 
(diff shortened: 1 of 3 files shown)


bobbinth (commented):
Looks good! Thank you! I left one comment for the future inline. Also, one general comment:

We currently have several ways (3?) of installing the node: (1) from crates.io, (2) from Docker image, and (3) from Debian package (a 4th potential way is by cloning and compiling the repo). I am thinking it would be good to say this upfront in the "Usage" section and then have dedicated sub-sections for each approach (and maybe common sub-sub sections for configuring the node - though, not sure to what extent they are common).

Basically, right now we dive right into installing from crates.io and only towards the end we mention other approaches. And it also is not super clear whether "Configuration" section, for example, is applicable only to the crates.io approach, or if, for example, it should also be used in the Debian package approach.

README.md
```diff
@@ -39,61 +39,83 @@ sudo apt install llvm clang bindgen pkg-config libssl-dev libsqlite3-dev
...
 
 ```sh
-miden-node init
+miden-node init \
+  --config-path <..>/miden-node.toml \
+  --genesis-path <..>/genesis.toml  
 ```
-To generate the genesis block, run:
+
+which will generate `miden-node.toml` and `genesis.toml` files. The latter controls the accounts that the genesis block will be spawned with and by default contains a basic wallet account and a basic fungible faucet account. You can modify this file to add/remove accounts as desired.
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: #446

Mirko-von-Leipzig:
> Looks good! Thank you! I left one comment for the future inline. Also, one general comment:
> 
> We currently have several ways (3?) of installing the node: (1) from crates.io, (2) from Docker image, and (3) from Debian package (a 4th potential way is by cloning and compiling the repo). I am thinking it would be good to say this upfront in the "Usage" section and then have dedicated sub-sections for each approach (and maybe common sub-sub sections for configuring the node - though, not sure to what extent they are common).
> 
> Basically, right now we dive right into installing from crates.io and only towards the end we mention other approaches. And it also is not super clear whether "Configuration" section, for example, is applicable only to the crates.io approach, or if, for example, it should also be used in the Debian package approach.

What do you think about restricting our officially supported options here? Restrict the guide to formal packages for Debian, `cargo` and an official docker image uploaded to `dockerhub`?

Advanced users are welcome to compile from source, or to create their own docker image but they shouldn't really need a guide for that beyond the `makefile` and `dockerfile`.

Mirko-von-Leipzig:
Hmm. I think our docker file is missing the `init` configuration file bit. It currently only copies the genesis data during the build process.

In addition, its missing the `--locked` so it can break when `miden-base` has breaking upstream changes (happened to me during testing on an old commit).

bobbinth:
> How do you feel about restricting our officially supported options here? Restrict the guide to formal packages for Debian, `cargo` and an official docker image uploaded to `dockerhub`?
> 
> Advanced users are welcome to compile from source, or to create their own docker image but they shouldn't really need a guide for that beyond the `makefile` and `dockerfile`.

Yep - that's fine. We could even the number of options to 2 for now if bring up the docker option to the latest will take non-trivial amount of work.

Mirko-von-Leipzig:
I've rewritten the instructions to only focus on the debian packages and `cargo install` - though the latter does include from source instructions as well.

I've removed docker (pending re-adding it as a separate section if we fix the build).

I've also removed mention of the separate process vs single process. What I'm thinking is (and please correct me if this is a bad idea):

Make the single process version very simple. Anything complicated ito configuration can be handled using the multi-process setup instead.

I'm thinking the simple version could just take in the `rpc` endpoint and a store directory. At this point it also doesn't really make sense to take in config file, so maybe we just have that subcommand take in `--rpc-endpoint` and `--data-directory` cli args. The store and block-producer endpoints don't have to public for a simple setup, so we can just `localhost:0` them in the code instead.

The biggest change here is really just that store and block-producer endpoints can't be configured.

README.md
```diff
@@ -27,143 +27,124 @@ The diagram below illustrates high-level design of each component as well as bas
...
 
-## Usage
+## Installation
+
+Official releases are available as debian packages which can be found under our [releases](https://github.com/0xPolygonMiden/miden-node/releases) page.
+
+Alternatively, the Rust package manager `cargo` can be used to install on non-debian distributions or to compile from source.
 
-Before you can build and run the Miden node or any of its components, you'll need to make sure you have Rust [installed](https://www.rust-lang.org/tools/install). Miden node v0.2 requires Rust version **1.78** or later.
+### Intall using `cargo`
```
bobbinth, line 36: Before this section, should we add a short "Installing debian package"? It could a big part of the old "Debian Packages" section.

README.md
```diff
@@ -27,143 +27,124 @@ The diagram below illustrates high-level design of each component as well as bas
...
 
-Please make sure you have the sha256sum program installed, for most linux operating systems this is already installed. If you wish to install it on your macOS, you can use brew:
+## Updating
 
-```sh
-brew install coreutils
-```
+We currently make no guarantees about backwards compatibility. Updating the node software therefore consists of wiping all existing data and re-installing the node's software again. This includes regenerating the configuration files and genesis block as these formats may have changed. This effectively means every update is a complete reset of the blockchain.
 
-## Testing
```
bobbinth, line 159: I think it doesn't hurt to keep the "Testing" section.

README.md
```diff
@@ -27,143 +27,124 @@ The diagram below illustrates high-level design of each component as well as bas
...
-
-   ```sh
-   make docker-build-node
-   ```
-
-   This command will build the docker image for the Miden node and save it locally.
+## Usage
 
-2. Run the Docker container
+### Setup
```
bobbinth, line 87: I don't remember for sure, but I think when installing a debian package some default location is used for storage and config files. Maybe it is worth mentioning it here somewhere?

bobbinth (commented):
Looks good! Thank you! I left a couple of comments inline - mostly about adding a bit more info about debian packages.

> Make the single process version very simple. Anything complicated ito configuration can be handled using the multi-process setup instead.
> 
> I'm thinking the simple version could just take in the `rpc` endpoint and a store directory. At this point it also doesn't really make sense to take in config file, so maybe we just have that subcommand take in `--rpc-endpoint` and `--data-directory` cli args. The store and block-producer endpoints don't have to public for a simple setup, so we can just `localhost:0` them in the code instead.

I think that's fine - but I would create a separate issue for this that we'd resolve in a subsequent PR.

README.md
Mirko-von-Leipzig: I was worried users would think it applies to testing their deployed node i.e. that its part of the user guide. I'll see if I can make it clear instead.

README.md
Mirko-von-Leipzig: Ah. Yeah it automagically sets up a systemd service which has expected locations.

The package also copies over the `toml` configuration files which I don't quite like since one still has to run `make-genesis` anyway, making the initialization only half complete.

I think I'll remove this so that normal installation and configuration are the same for all ways?

Mirko-von-Leipzig: Hmm. Actually I'm unsure now. I see that the post-install script adds a `miden` user but I have no idea why -- I'm not that clued up on this sort of devops.

Could be that its required for file permissions in which case creating the files afterwards wouldn't be correct? i.e. `make-genesis` won't really work.

😕

Mirko-von-Leipzig: Ah. The systemd file expects the `user` to be `miden` which is probably the only reason it was added. I hope.

README.md
```diff
@@ -27,142 +27,166 @@ The diagram below illustrates high-level design of each component as well as bas
...
-```sh
-sudo apt install llvm clang bindgen pkg-config libssl-dev libsqlite3-dev
-```
+Alternatively, the Rust package manager `cargo` can be used to install on non-debian distributions or to compile from source.
 
-### Installing the node
+### Debian package
 
-To install for production use cases, run:
+Debian packages are available and are the fastest way to install the node on Debian based system. Currently only `amd64` architectures are supported.
```
bobbinth, line 40: spelling: should it be "... on a Debian-based system."?

Also, probably "... only `amd64` architecture is supported."

README.md
```diff
@@ -27,142 +27,166 @@ The diagram below illustrates high-level design of each component as well as bas
...
-```sh
-cargo install --path bin/node
-```
+These packages can be found under our [releases](https://github.com/0xPolygonMiden/miden-node/releases) page along with a checksum.
 
-This will install the executable `miden-node` in your PATH, at `~/.cargo/bin/miden-node`.
+Note that this includes a `systemd` service called `miden-node` (disabled by default).
 
-Otherwise, if only to try the node out for testing, run:
+To install, download the desired releases `.deb` package and checksum files. Install using
```
bobbinth, line 46: Should we mention here (or somewhere else) that the user needs to be `miden`? Or is this not a requirement?

bobbinth (approved):
Looks good! Thank you! I left a couple of comments/questions inline.

README.md
Mirko-von-Leipzig: afaict its not a requirement - the `miden` user is created as part of the post-install instructions. Which I initially found alarming but is apparently industry practice.. go figure.

README.md
Mirko-von-Leipzig: Thanks. I should figure out how to integrate a spell-checker into my editor..

PR #438, opened 2024-08-12
fix: Fix build by adding note's execution hints

igamigo:
This fixes the faucet's and node build by doing the following:

- Adding the new `NoteExecutionHint` for all note's metadata
- Adding the new `after_block_num` field to note exports on the faucet

Note that this does not store/retrieve the execution hints correctly, it hardcodes `NoteExecutionHint::None` for all notes right now in order to have a working build and unblock ongoing work. Correctly storing the hint is to be done separately (#439).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,9 +36,9 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.8.0"
+version = "3.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


Mirko-von-Leipzig (commented):
iiuc this is bumping the various `miden-base` dependencies? 

What's the driving rationale? Is this required for ongoing client work?

SantiagoPittella (approved):

igamigo:
Superseded by #439

PR #441, opened 2024-08-12
feat: Add note `execution_hint` to store

### LINKED ISSUE #439: Add `NoteExecutionHint` to the store

- [ ] RPC
- [ ] Block producer
- [X] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

Recently, `NoteExecutionHint` was added to the `NoteMetadata` struct. We need to store and retrieve this accordingly.

### How should it be done?

Add the field into the SQL databases and queries, and store/retrieve it accordingly for transaction outputs.

### When is this task done?

When we can return the correct note metadata on note queries.

### Additional context

_No response_

SantiagoPittella:
This fixes the faucet's and node build by doing the following:

- Adding the new NoteExecutionHint for all note's metadata
- Adding the new after_block_num field to note exports on the faucet
- Adding the `execution_hint` field to Notes table.

Closes #439 .

This PR is built on top of https://github.com/0xPolygonMiden/miden-base/pull/827 .

It is meant to be merged AFTER that PR.

### DIFF
--- a/crates/proto/src/domain/notes.rs
+++ b/crates/proto/src/domain/notes.rs
@@ -26,8 +26,15 @@ impl From<NoteMetadata> for crate::generated::note::NoteMetadata {
         let sender = Some(val.sender().into());
         let note_type = val.note_type() as u32;
         let tag = val.tag().into();
+        let execution_hint: u64 = Felt::from(val.execution_hint()).into();
         let aux = val.aux().into();
 
-        crate::generated::note::NoteMetadata { sender, note_type, tag, aux }
+        crate::generated::note::NoteMetadata {
+            sender,
+            note_type,
+            tag,
+            execution_hint,
+            aux,
+        }
     }
 }
--- a/crates/proto/src/domain/notes.rs
+++ b/crates/proto/src/domain/notes.rs
@@ -16,11 +16,7 @@ impl TryFrom<crate::generated::note::NoteMetadata> for NoteMetadata {
         let note_type = NoteType::try_from(value.note_type as u64)?;
         let tag = NoteTag::from(value.tag);
 
-        // TODO: Conversion/helper functions should be provided for these conversions
-        let execution_hint_tag = (value.execution_hint & 0xFF) as u8;
-        let execution_hint_payload = ((value.execution_hint >> 8) & 0xFFFFFF) as u32;
-        let execution_hint =
-            NoteExecutionHint::from_parts(execution_hint_tag, execution_hint_payload)?;
+        let execution_hint = NoteExecutionHint::try_from(value.execution_hint)?;
 
         let aux = Felt::try_from(value.aux).map_err(|_| ConversionError::NotAValidFelt)?;
 
--- a/crates/proto/src/domain/notes.rs
+++ b/crates/proto/src/domain/notes.rs
@@ -15,9 +15,16 @@ impl TryFrom<crate::generated::note::NoteMetadata> for NoteMetadata {
             .try_into()?;
         let note_type = NoteType::try_from(value.note_type as u64)?;
         let tag = NoteTag::from(value.tag);
+
+        // TODO: Conversion/helper functions should be provided for these conversions
+        let execution_hint_tag = (value.execution_hint & 0xFF) as u8;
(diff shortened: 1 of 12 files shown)


crates/store/src/db/sql.rs
```diff
@@ -589,9 +600,15 @@ pub fn select_notes_since_block_by_tag_and_sender(
         let merkle_path = MerklePath::read_from_bytes(merkle_path_data)?;
         let details_data = row.get_ref(9)?.as_blob_or_null()?;
         let details = details_data.map(<Vec<u8>>::read_from_bytes).transpose()?;
+        let execution_hint = row.get_ref(10)?.as_i64()? as u64;
```
igamigo, line 603: I think we can use `column_value_as_u64()` here

crates/proto/src/domain/notes.rs
```diff
@@ -26,8 +33,15 @@ impl From<NoteMetadata> for crate::generated::note::NoteMetadata {
         let sender = Some(val.sender().into());
         let note_type = val.note_type() as u32;
         let tag = val.tag().into();
+        let execution_hint: u64 = Felt::from(val.execution_hint()).into();
```
igamigo, line 36: I think we can now replace this with 
```rust
        let execution_hint: u64 = val.execution_hint().into();
```

igamigo (approved):
LGTM! Left a couple minor comments. I also don't know if it's possible to easily test this (checking that storing/retrieval is correct) but it would be nice

crates/proto/src/domain/notes.rs
```diff
@@ -15,9 +15,16 @@ impl TryFrom<crate::generated::note::NoteMetadata> for NoteMetadata {
             .try_into()?;
         let note_type = NoteType::try_from(value.note_type as u64)?;
         let tag = NoteTag::from(value.tag);
+
+        // TODO: Conversion/helper functions should be provided for these conversions
+        let execution_hint_tag = (value.execution_hint & 0xFF) as u8;
+        let execution_hint_payload = ((value.execution_hint >> 8) & 0xFFFFFF) as u32;
+        let execution_hint =
+            NoteExecutionHint::from_parts(execution_hint_tag, execution_hint_payload)?;
```
Mirko-von-Leipzig, line 23: Something to consider as part of the discussion in #435

Mirko-von-Leipzig (approved):
LGTM

crates/store/src/db/migrations/001-init.sql
```diff
@@ -23,16 +23,17 @@ CREATE TABLE
...
+    batch_index    INTEGER NOT NULL, -- Index of batch in block, starting from 0
+    note_index     INTEGER NOT NULL, -- Index of note in batch, starting from 0
+    note_id        BLOB    NOT NULL,
+    note_type      INTEGER NOT NULL, -- 1-Public (0b01), 2-Private (0b10), 3-Encrypted (0b11)
+    sender         INTEGER NOT NULL,
+    tag            INTEGER NOT NULL,
+    aux            INTEGER NOT NULL,
+    merkle_path    BLOB    NOT NULL,
+    details        BLOB,
+    execution_hint INTEGER,
```
bobbinth, line 36: A couple of small comments:

- We should probably move `execution_hint` to be right after `aux` field since logically it is part of the note's metadata.
- I think `execution_hint` should be `NOT NULL` field as we always have an integer value for it.

crates/store/src/db/sql.rs
```diff
@@ -424,7 +424,8 @@ pub fn select_notes(conn: &mut Connection) -> Result<Vec<NoteRecord>> {
             tag,
             aux,
             merkle_path,
-            details
+            details,
+            execution_hint
```
bobbinth, line 428: Similar to the other comment, here, and in other places in this file, I'd probably move `execution_hint` to be right after the `aux` field.

crates/proto/src/domain/notes.rs
bobbinth: I think instead of doing this, we should actually use `NoteExecutionHint::try_from()` here (I believe this is now available in `miden-base`).

igamigo (approved):
LGTM!

PR #443, opened 2024-08-15
ci: refactor deploy workflow

Mirko-von-Leipzig:
This PR removes testnet deployment until we are confident that devnet works.

To facilitate this testing, both are now triggered manually but with the testnet option removed.

I've also changed the secrets naming to follow the github conventions, and also hidden the IPs.

And renamed workflow file to `deploy.yml`


Mirko-von-Leipzig:
Superseded by #445

PR #444, opened 2024-08-15
docs: Update `responses.proto`

igamigo:
Seems like the doc comments for the returned MMR path in the `SyncNoteResponse` was referring to an `MmrProof`, but a `MerklePath` is returned instead.

### DIFF
--- a/proto/responses.proto
+++ b/proto/responses.proto
@@ -69,12 +69,7 @@ message SyncNoteResponse {
     // Block header of the block with the first note matching the specified criteria
     block_header.BlockHeader block_header = 2;
 
-    // Proof for block header's MMR with respect to the chain tip.
-    //
-    // More specifically, the full proof consists of `forest`, `position` and `path` components. This
-    // value constitutes the `path`. The other two components can be obtained as follows: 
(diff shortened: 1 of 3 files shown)


SantiagoPittella (approved):

Mirko-von-Leipzig (approved):

proto/responses.proto
```diff
@@ -69,12 +69,7 @@ message SyncNoteResponse {
...
     block_header.BlockHeader block_header = 2;
 
-    // Proof for block header's MMR with respect to the chain tip.
-    //
-    // More specifically, the full proof consists of `forest`, `position` and `path` components. This
-    // value constitutes the `path`. The other two components can be obtained as follows: 
-    //   - `position` is simply `resopnse.block_header.block_num`
-    //   - `forest` is the same as `response.chain_tip + 1`
+    // Merkle path to verify the block's inclusion in the MMR at the returned `chain_tip`
     merkle.MerklePath mmr_path = 3;
```
bobbinth, line 73: This is a non-blocking comment: would it still make sense to mention how MMR proof can be constructed from the response?

igamigo: Added some context related to this

PR #445, opened 2024-08-16
Devops 2881 test

BrucePolygon:
Updated the workflow with changing ssh to sim


Mirko-von-Leipzig (approved):
I made two small changes

1. Enabled testnet option
2. Instance ID must get echo'd into github env for it to be available for subsequent steps

PR #448, opened 2024-08-20
ci: add locking and remove extra nightly

### LINKED ISSUE #447: Infrastructure files are missing `--locked`
The `makefile`, `dockerfile` and possibly some CI workflows are missing the `--locked` specification which often results in broken builds. In particular, this can happen often because we are relying on `miden-base/next` which of course moves pretty fast :)

We should add `--locked` to all of these locations.

Mirko-von-Leipzig:
This PR adds `--locking` to all infrastructure files where applicable. It also removes all `nightly` toolchains from non-fmt workflows.

Closes #447.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,48 +15,50 @@ concurrency:
 
 jobs:
   rustfmt:
-    name: rustfmt check nightly on ubuntu-latest
+    name: rustfmt
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@main
-      - uses: Swatinem/rust-cache@v2
-        with:
-          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
-      - name: Rustfmt
+      - name: Rustup
         run: |
           rustup update --no-self-update nightly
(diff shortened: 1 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -15,48 +15,50 @@ concurrency:
...
-          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
-      - name: Rustfmt
+      - name: Rustup
         run: |
           rustup update --no-self-update nightly
           rustup +nightly component add rustfmt
-          make format-check
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
```
bobbinth, line 28: Is there any significance to ordering? Previously we had `uses: Swatinem/rust-cache@v2` before `Rustup` - but now it is after - does that make a difference?

bobbinth (approved):
Looks good! Thank you! I left just one question inline.

Could you also propagate these changes to https://github.com/0xPolygonMiden/miden-base/pull/818?

cc @igamigo - we should probably do something similar on the client.

PR #450, opened 2024-08-20
ci: fix debian package workflow

Mirko-von-Leipzig:
I broke this in #448.

This workflow also has some strange triggers. Should we remove some of them?

I somehow thought it should only run for formal releases.

-------------------

edit: workflow now only triggers on releases

### DIFF
--- a/.github/workflows/deb_packager.yml
+++ b/.github/workflows/deb_packager.yml
@@ -32,7 +32,7 @@ jobs:
         run: cargo clean
 
       - name: Building for amd64
-        run: cargo build --release --path bin/node --locked
+        run: cargo build --release --locked --bin miden-node
 
       - name: create package directories
         run: |
--- a/.github/workflows/deb_packager.yml
+++ b/.github/workflows/deb_packager.yml
@@ -1,13 +1,8 @@
 name: deb_packager
 
 on:
-  push:
-    branches: [main, next]
-    paths:
-      - '**'
-    tags:
-      - 'v*.*.*'
-      - 'v*.*.*-*'
+  release:
+    types: [released, prereleased]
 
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig:
Actually I'm fairly sure we are overwriting the release package for the latest release on every push onto next?

yep - here is the workflow from #444 overwriting release 0.4.0 assets: https://github.com/0xPolygonMiden/miden-node/actions/runs/10422889196/job/28868391306.

Mirko-von-Leipzig:
A question then - when do we actually want to trigger this workflow?

Two options I think

1. On a "release" tag ie `v.x.y.z` etc
2. On a github release

Probably (2) makes the most sense as we want the package to get appended to it?

bobbinth (approved):
Looks good! Thank you!

> A question then - when do we actually want to trigger this workflow?
> 
> Two options I think
> 
> 1. On a "release" tag ie `v.x.y.z` etc
> 2. On a github release
> 
> Probably (2) makes the most sense as we want the package to get appended to it? Bleh although then it won't update the we change the tag to point to somewhere else..

I think (2) also makes the most sense to me (and I don't think we change tags to point to something else - under what circumstances may we want to do this?).

Mirko-von-Leipzig:
> I think (2) also makes the most sense to me (and I don't think we change tags to point to something else - under what circumstances may we want to do this?).

I've had this happen once or twice when I realized I had tagged the wrong commit immediately post release, and another time where I clicked release on a release draft and it created the tag on the wrong branch.

So essentially only when there was a manual "bug" involved.

But one can also "fix" by deleting the release and starting over so maybe not so much of an issue.

PR #451, opened 2024-08-20
ci: devnet deployment workflow

sshrihar:
Fix for miden devnet environment deployment.


.github/workflows/deploy_devnet.yml
```diff
@@ -0,0 +1,230 @@
+# Deploy the miden node for devnet.
+name: Node deployment fix for devnet
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
```
Mirko-von-Leipzig, line 8:
```suggestion
    branches:
      - next
  workflow_dispatch:
```

.github/workflows/deploy_devnet.yml
```diff
@@ -0,0 +1,230 @@
...
+      
+      - name: Install prerequisites
+        id: install_prereq
+        run: |
+          COMMAND_ID=$(aws ssm send-command \
+            --instance-ids ${{ secrets.DEVNET_INSTANCE_ID }} \
+            --document-name "AWS-RunShellScript" \
+            --parameters '{"commands":[
+                  "sudo apt-get update -y",
+                  "sudo apt-get install -y git gcc openssl libbz2-dev libffi7 make cargo",
```
Mirko-von-Leipzig, line 68:
```suggestion
                  "sudo apt-get install -y git gcc openssl libbz2-dev libffi7 make",
```
I think cargo probably shouldn't be here since we are using `rustup` to manage cargo

.github/workflows/deploy_devnet.yml
```diff
@@ -0,0 +1,230 @@
...
+                  "sudo apt install -y build-essential",
+                  "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y",
+                  "source $HOME/.cargo/env",
+                  "rustup update",
+                  "sudo apt-get install -y clang cmake",
+                  "sudo rm -rf /opt/miden/miden-node/genesis.dat",
+                  "sudo rm -rf /opt/miden/miden-node/miden-store.sqlite3",
+                  "sudo rm -rf /opt/miden/miden-node/miden-store.sqlite3-*",
+                  "sudo rm -rf /opt/miden/miden-node/accounts",
+                  "sudo rm -rf /opt/miden/miden-node/blocks"
```
Mirko-von-Leipzig, line 78:
```suggestion
                  "sudo rm -rf /opt/miden/miden-node/",
                  "sudo mkdir  /opt/miden/miden-node/",
```
Probably best to nuke the whole directory in case we change the data required.

.github/workflows/deploy_devnet.yml
```diff
@@ -0,0 +1,230 @@
...
+                --query "CommandInvocations[0].Status" \
+                --output text)
+            echo "Command Status: $STATUS"
+            
+            OUTPUT=$(aws ssm list-command-invocations \
+                --command-id ${{ steps.install_prereq.outputs.command_id }} \
+                --details \
+                --query "CommandInvocations[0].CommandPlugins[0].Output" \
+                --output text)
+            echo "Command Output: $OUTPUT"
```
Mirko-von-Leipzig, line 99: When called in a loop, does this only contain the new output, or is it the full output each time?

As in, if I call it twice it will print the same thing twice or only the new output is appended? I assume the former.

.github/workflows/deploy_devnet.yml
```diff
@@ -0,0 +1,230 @@
...
+                "sudo systemctl enable miden-faucet",
+                "sudo systemctl start miden-node",
+                "sleep 5",
+                "sudo systemctl start miden-faucet"
+              ]}' \
+            --output text \
+            --query "Command.CommandId")
+          echo "command_id=$COMMAND_ID" >> $GITHUB_OUTPUT
+
+      - name: Check Start services command status and retrieve output
```
Mirko-von-Leipzig, line 211: Is it possible to somehow extract these checks steps? Probably not from what I remember without creating a custom github action?

.github/workflows/deploy_devnet.yml
Mirko-von-Leipzig, line 1: Is the idea to have a completely separate file for testnet? Could we instead combine them so we don't have so much duplication?

Mirko-von-Leipzig (commented):
Some minor suggestions, I'd love to find a way to deduplicate/simplify this workflow but I suspect there just isn't a way

.github/workflows/deploy_devnet.yml
Mirko-von-Leipzig: Looks like it should be possible to create a nice wrapper using a composite action. Though I can attempt that at some other time. Right now its more important to just get this up and running.

.github/workflows/deploy_workflow.yml
```diff
@@ -0,0 +1,252 @@
...
+            echo "Command Output: $OUTPUT"
+            
+            if [ "$STATUS" == "Success" ]; then
+              echo "Command completed successfully."
+              break
+            elif [ "$STATUS" == "Failed" ] || [ "$STATUS" == "Cancelled" ]; then
+              echo "Command failed with status: $STATUS"
+              exit 1
+            else
+              sleep 10  # Wait for 30 seconds before checking again
```
Mirko-von-Leipzig, line 65:
```suggestion
              sleep 10  # Wait for 10 seconds before checking again
```

Mirko-von-Leipzig: And elsewhere as well. Probably best to just remove the comment :D

.github/workflows/deploy_workflow.yml
```diff
@@ -0,0 +1,252 @@
...
+        run: |
+          COMMAND_ID=$(aws ssm send-command \
+            --instance-ids ${{ secrets.instance_id }} \
+            --document-name "AWS-RunShellScript" \
+            --parameters '{"commands":[
+                  "sudo apt-get update -y",
+                  "sudo apt-get install -y git gcc openssl libbz2-dev libffi7 make",
+                  "sudo apt install -y build-essential",
+                  "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y",
+                  "source $HOME/.cargo/env",
```
Mirko-von-Leipzig, line 80: Is this required because the rust install script doesn't update the current sessions paths?

.github/workflows/deploy_workflow.yml
```diff
@@ -0,0 +1,252 @@
...
+          done
+
+      - name: Cargo Install
+        id: cargo_install
+        run: |
+          COMMAND_ID=$(aws ssm send-command \
+            --instance-ids ${{ secrets.instance_id }} \
+            --document-name "AWS-RunShellScript" \
+            --parameters '{"commands":[
+                "/home/ssm-user/.cargo/bin/cargo install --git https://github.com/0xPolygonMiden/miden-node miden-node ${{ inputs.build_prefix }} --bin miden-node --locked --features ${{ inputs.features }}",
```
Mirko-von-Leipzig, line 125: Do we not get `cargo` added to the path? Not an issue; just curious.

.github/workflows/deploy_testnet.yml
```diff
@@ -0,0 +1,20 @@
+name: testnet deploy
+
+on:
+  release:
+    types: [created]
+  workflow_dispatch:
```
Mirko-von-Leipzig, line 6: Manually triggering this probably won't work as `github.ref_name` won't be a useable git tag.

Mirko-von-Leipzig (approved):
Some minor nits and a question about the testnet triggers.

Overall this looks really good though!

PR #452, opened 2024-08-20
feat(faucet): add `after_block_num` to the faucet

### LINKED ISSUE #440: Add `after_block_num` to the faucet

- [ ] RPC
- [ ] Block producer
- [ ] Store
- [ ] Protobuf messages
- [ ] Testing
- [X] Faucet

### What should be done?

The faucet does not currently sync with the node. This means it has no idea of where the blockchain is, so it cannot correctly fill in the `after_block_num` field for the note export file (which is meant to be a sensible starting point for clients to sync the note correctly when importing).

### How should it be done?

Add a way to periodically sync the faucet (perhaps on every transaction) and send the latest known block number on the `after_block_num` field when exporting notes.

### When is this task done?

When exported notes contain a valid `after_block_num` field.

### Additional context

_No response_

SantiagoPittella:
This PR introduces a `current_block_number` to the `FaucetClient` and spawns a thread while running the faucet that is in charge of updating the `current_block_number` after `CHAIN_TIP_UPDATER_INTERVAL` time.

This PR also addresses an issue with the `.masm` file adding the execution hint field.

Closes #440

### DIFF
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -78,14 +78,18 @@ impl FaucetClient {
 
         let current_block_number = 0;
 
-        Ok(Self {
+        let mut faucet_client = Self {
             data_store,
             rpc_api,
             executor,
             id,
             rng,
             current_block_number,
-        })
+        };
+
+        faucet_client.update_current_block_number().await?;
+
+        Ok(faucet_client)
     }
 
     /// Executes a mint transaction for the target account.
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -45,6 +45,7 @@ pub struct FaucetClient {
     data_store: FaucetDataStore,
     id: AccountId,
     rng: RpoRandomCoin,
+    current_block_number: u32,
 }
 
 unsafe impl Send for FaucetClient {}
@@ -74,7 +75,17 @@ impl FaucetClient {
         let mut rng = thread_rng();
         let coin_seed: [u64; 4] = rng.gen();
         let rng = RpoRandomCoin::new(coin_seed.map(Felt::new));
-        Ok(Self { data_store, rpc_api, executor, id, rng })
+
+        let current_block_number = 0;
+
+        Ok(Self {
+            data_store,
+            rpc_api,
+            executor,
+            id,
+            rng,
+            current_block_number,
+        })
     }
 
     /// Executes a mint transaction for the target account.
@@ -149,6 +160,22 @@ impl FaucetClient {
 
     pub fn get_faucet_id(&self) -> AccountId {
         self.id
+    }
+
+    pub fn get_chain_tip(&self) -> u32 {
+        self.current_block_number
+    }
+
+    pub async fn update_current_block_number(&mut self) -> Result<(), FaucetError> {
+        let request = GetBlockHeaderByNumberRequest { block_num: None, include_mmr_proof: None };
+
+        let response = self.rpc_api.get_block_header_by_number(request).await.map_err(|err| {
+            FaucetError::InternalServerError(format!("Failed to get block header: {}", err))
+        })?;
+
+        self.current_block_number = response.into_inner().block_header.unwrap().block_num;
+
+        Ok(())
     }
 }
 
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -323,13 +323,15 @@ fn build_transaction_arguments(
 
     let tag = output_note.metadata().tag().inner();
     let aux = output_note.metadata().aux().inner();
+    let execution_hint = output_note.metadata().execution_hint().into();
 
     let script = &DISTRIBUTE_FUNGIBLE_ASSET_SCRIPT
         .replace("{recipient}", &recipient)
         .replace("{note_type}", &Felt::new(note_type as u64).to_string())
         .replace("{aux}", &Felt::new(aux).to_string())
         .replace("{tag}", &Felt::new(tag.into()).to_string())
-        .replace("{amount}", &Felt::new(asset.amount()).to_string());
+        .replace("{amount}", &Felt::new(asset.amount()).to_string())
+        .replace("{execution_hint}", &Felt::new(execution_hint).to_string());
 
     let script = TransactionScript::compile(script, vec![], TransactionKernel::assembler())
         .map_err(|err| {
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -45,7 +45,6 @@ pub struct FaucetClient {
     data_store: FaucetDataStore,
     id: AccountId,
     rng: RpoRandomCoin,
-    current_block_number: u32,
 }
 
 unsafe impl Send for FaucetClient {}
@@ -76,16 +75,7 @@ impl FaucetClient {
         let coin_seed: [u64; 4] = rng.gen();
         let rng = RpoRandomCoin::new(coin_seed.map(Felt::new));
 
-        let current_block_number = 0;
-
-        Ok(Self {
-            data_store,
-            rpc_api,
-            executor,
-            id,
-            rng,
-            current_block_number,
-        })
+        Ok(Self { data_store, rpc_api, executor, id, rng })
     }
 
     /// Executes a mint transaction for the target account.
@@ -161,12 +151,6 @@ impl FaucetClient {
 
     pub fn get_faucet_id(&self) -> AccountId {
         self.id
-    }
-
-    pub fn set_current_block_number(&mut self, new_block_height: u32) -> Result<(), FaucetError> {
-        self.current_block_number = new_block_height;
-
-        Ok(())
     }
 }
 
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -78,18 +78,14 @@ impl FaucetClient {
 
         let current_block_number = 0;
 
-        let mut faucet_client = Self {
+        Ok(Self {
             data_store,
             rpc_api,
             executor,
             id,
             rng,
             current_block_number,
-        };
-
-        faucet_client.update_current_block_number().await?;
-
-        Ok(faucet_client)
+        })
     }
 
     /// Executes a mint transaction for the target account.
@@ -136,7 +132,7 @@ impl FaucetClient {
     pub async fn prove_and_submit_transaction(
         &mut self,
         executed_tx: ExecutedTransaction,
-    ) -> Result<(), FaucetError> {
+    ) -> Result<u32, FaucetError> {
         let transaction_prover = TransactionProver::new(ProvingOptions::default());
 
         let delta = executed_tx.account_delta().clone();
@@ -150,7 +146,8 @@ impl FaucetClient {
             transaction: proven_transaction.to_bytes(),
         };
 
-        self.rpc_api
+        let response = self
+            .rpc_api
             .submit_proven_transaction(request)
             .await
             .map_err(|err| FaucetError::InternalServerError(err.to_string()))?;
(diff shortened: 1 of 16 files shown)


bin/faucet/src/client.rs
```diff
@@ -79,7 +80,17 @@ impl FaucetClient {
         let mut rng = thread_rng();
         let coin_seed: [u64; 4] = rng.gen();
         let rng = RpoRandomCoin::new(coin_seed.map(Felt::new));
-        Ok(Self { data_store, rpc_api, executor, id, rng })
+
+        let current_block_number = 0;
```
Mirko-von-Leipzig, line 84: You should be able to initialize this to the on-chain value using the `initialize_faucet_client` function

bin/faucet/src/main.rs
```diff
@@ -72,6 +73,26 @@ async fn main() -> Result<(), FaucetError> {
 
             info!(target: COMPONENT, %config, "Initializing server");
 
+            let client_state_clone = Arc::clone(&faucet_state.client);
+
+            info!("Initializing chain tip updater");
+            thread::spawn(move || {
+                let sys = actix_web::rt::System::new();
```
Mirko-von-Leipzig, line 80: Nit: consider moving the clone inside the spawn

```suggestion
            let client_state_clone = Arc::clone(&faucet_state.client);

            info!("Initializing chain tip updater");
            thread::spawn({
                let client_state = Arc::clone(&faucet_state.client);
                move || {
                    let sys = actix_web::rt::System::new();
```

bin/faucet/src/main.rs
```diff
@@ -72,6 +73,26 @@ async fn main() -> Result<(), FaucetError> {
...
+                sys.block_on(async move {
+                    loop {
+                        let state_clone_inner = Arc::clone(&client_state_clone);
+                        actix_web::rt::spawn(async move {
```
Mirko-von-Leipzig, line 84: I'm not very familiar with actix, but do you need to spawn here? Isn't it already in a future so you could just call it without the additional spawn?

bin/faucet/src/main.rs
```diff
@@ -72,6 +73,26 @@ async fn main() -> Result<(), FaucetError> {
...
+                            let mut state = state_clone_inner.lock().await;
+                            state.update_current_block_number().await.unwrap();
+                        })
+                        .await
+                        .unwrap();
+
+                        thread::sleep(Duration::from_secs(CHAIN_TIP_UPDATER_INTERVAL));
```
Mirko-von-Leipzig, line 91: Feels weird using a normal sleep in async, shouldn't this be a tokio sleep? Or does it not matter because actix is single threaded?

Mirko-von-Leipzig (commented):
Thanks - some comments and questions for now

bin/faucet/src/main.rs
SantiagoPittella: Oh, no. You are right. This spawn is not needed. I'm removing it. Thanks!

bin/faucet/src/main.rs
SantiagoPittella: Yes, it should use a Tokio interval and a actix_web task should be enough too. Im changing it.

bin/faucet/src/main.rs
SantiagoPittella: I can't move the close inside the spawn since it will take ownership of the `faucet_state`

Mirko-von-Leipzig: Does it? Hmm. I thought maybe it wouldn't because it comes before the `move ||`.

bobbinth:
Would the solution proposed by @igamigo in https://github.com/0xPolygonMiden/miden-node/issues/440#issuecomment-2297068005 be simpler than a separate task?

bin/faucet/src/main.rs
```diff
@@ -72,6 +73,26 @@ async fn main() -> Result<(), FaucetError> {
...
+
+            info!("Initializing chain tip updater");
+            thread::spawn(move || {
+                let sys = actix_web::rt::System::new();
+                sys.block_on(async move {
+                    loop {
+                        let state_clone_inner = Arc::clone(&client_state_clone);
+                        actix_web::rt::spawn(async move {
+                            let mut state = state_clone_inner.lock().await;
+                            state.update_current_block_number().await.unwrap();
```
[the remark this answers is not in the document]
bobbinth: I don't have a strong opinion here - but I'm also thinking that maybe we could implement a solution mentioned by @igamigo in https://github.com/0xPolygonMiden/miden-node/issues/440#issuecomment-2297068005. What are the downsides of doing this?

bin/faucet/src/handlers.rs
```diff
@@ -77,7 +77,7 @@ pub async fn get_tokens(
     // Ideally, we should export these note details with a more sensible `after_block_num`
     let bytes = NoteFile::NoteDetails {
         details: note_details,
-        after_block_num: 0,
+        after_block_num: client.get_chain_tip(),
```
Mirko-von-Leipzig, line 80: The comment above needs to be changed now that we execute against non-zero block.

bin/faucet/src/main.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: I think a downside is that the block number used is always one transaction behind I think? So request `N` will use the block number as returned by transaction `N-1` unless I've misunderstood the order of operations?

This makes the block number staleness unbounded as it depends on transaction frequency. With a separate task its bounded by the poll period.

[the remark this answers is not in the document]
bobbinth: I think which block number we use for the transaction is less important here than the block number we attach to the generated note (this is the [after_block_num](https://github.com/0xPolygonMiden/miden-base/blob/next/objects/src/notes/file.rs#L23) hint which the client can then use to limit its search for the note). And since we can attach this after the note is created, I think the sequence of operations could be:

1. Create a transaction using the latest known `block_num` (e.g., $N$).
2. Submit the transaction to the node and get back the new latest `block_num` (e.g., $M$).
3. Set `after_block_num` for the note to $M$ - this way, the user gets the latest possible block number.
4. Save $M$ as the new latest `block_num`.

Yes, the next transaction would use $M$ as `block_num` and it may be "out of date" by that time, but what really matters here (I think) is step 3 above.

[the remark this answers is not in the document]
Mirko-von-Leipzig: Thanks that makes sense. 

ito implementation for the faucet this would be fairly straight-forward I think. The handler already holds a lock for the duration of the query/execution so that part should be simple.

iiuc the only other change would be adding this to the proto response.

@SantiagoPittella how do you feel about trying method this instead?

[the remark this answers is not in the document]
SantiagoPittella: I've been working on adding the `chain_tip` to the response of `submit_proven_transaction`. To do this, I needed to add an endpoint to the Store: `get_chain_tip`. Since I don't have access to `State` or `Store` from the `block-producer` component, the only way I see to make use of this new endpoint is by adding a call in [`crates/rpc/src/server/api.rs:190`](https://github.com/0xPolygonMiden/miden-node/blob/next/crates/rpc/src/server/api.rs#L190), which I understood to be just a dispatcher to the corresponding component, and modifying the `block-producer`'s response with the correct chain_tip. Something like:
```rust
        // Fetch the store for the chain_tip.
        let chain_tip = self.store.clone().get_chain_tip(GetChainTipRequest {} ).await?.into_inner().chain_tip;

        // Call the block-producer's submit_proven_transaction
        let mut response = self.block_producer.clone().submit_proven_transaction(request).await?;

        // Modify response adding the chain_tip
        response.chain_tip = chain_tip;
```
This doesn't seem like a good idea, but I don't see any other way to do it. Any help or workaround would be greatly appreciated.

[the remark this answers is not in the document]
bobbinth: I think we may be able to avoid adding a new endpoint. For every transaction, the block producer makes a `GetTransactionInputs` request to the store. I think we may be able to include the chain tip in the [GetTransactionInputsResponse](https://github.com/0xPolygonMiden/miden-node/blob/next/proto/responses.proto#L126). And then, the block producer would return it to the RPC component.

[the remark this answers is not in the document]
bobbinth: Though, if it does turn out that passing this info through creates too many complications, we can switch to the original approach.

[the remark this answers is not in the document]
SantiagoPittella: I removed the endpoint and used the new approach. I think the code quality is much better with version. Let me know what you think.

crates/store/src/server/api.rs
```diff
@@ -411,6 +411,11 @@ impl api_server::Api for StoreApi {
             .await
             .map_err(internal_error)?;
 
+        let (block_header, _) =
+            self.state.get_block_header(None, false).await.map_err(internal_error)?;
```
bobbinth, line 415: This works, but I think we'll do an extra database access here which we may be able to avoid. Specifically, we could add something like `State::get_chain_length()` which would return the current length of the chain. Internally, this can be implemented by getting the number of nodes in the MMR stored in the `InnerState`.

crates/block-producer/src/txqueue/mod.rs
```diff
@@ -149,10 +149,14 @@ where
     /// This method will validate the `tx` and ensure it is valid w.r.t. the rollup state, and the
     /// current in-flight transactions.
     #[instrument(target = "miden-block-producer", skip_all, err)]
-    pub async fn add_transaction(&self, tx: ProvenTransaction) -> Result<(), AddTransactionError> {
+    pub async fn add_transaction(
```
bobbinth, line 152: Let's update the doc comment to describe what this method returns.

crates/block-producer/src/txqueue/mod.rs
```diff
@@ -31,7 +31,7 @@ pub trait TransactionValidator: Send + Sync + 'static {
     ///   in-flight transactions.
     /// - Track the necessary state of the transaction until it is committed to the `store`, to
     ///   perform the check above.
-    async fn verify_tx(&self, tx: &ProvenTransaction) -> Result<(), VerifyTxError>;
+    async fn verify_tx(&self, tx: &ProvenTransaction) -> Result<Option<u32>, VerifyTxError>;
```
bobbinth, line 34: Same as the above comment.

bobbinth (commented):

One other thing I was wondering: maybe instead of working `current_block_num` with `current_block_num` we should be working with `chain_length`. This may make it a bit less ambiguous that we are returning the current length of the chain rather than the block number in which the transaction was recorded.

If we do that, we'd just set `after_block_num` to something like `chain_length - 1`.

crates/store/src/server/api.rs
SantiagoPittella: Turns out it was already implemented under the name of `latest_block_num()` in the State using the method that you described there. I will use that.

bobbinth: Ah - nice! I forgot about this one.

igamigo (approved):
LGTM! Left one small comment.

bin/faucet/src/client.rs
```diff
@@ -144,12 +156,18 @@ impl FaucetClient {
...
 
     pub fn get_faucet_id(&self) -> AccountId {
         self.id
     }
+
+    pub fn set_current_block_number(&mut self, new_block_height: u32) -> Result<(), FaucetError> {
+        self.current_block_number = new_block_height;
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
SantiagoPittella: Yes, this is not needed anymore! Thanks!

bobbinth (approved):
Looks good! Thank you!

crates/block-producer/src/store/mod.rs
```diff
@@ -85,6 +85,8 @@ pub struct TransactionInputs {
     pub nullifiers: BTreeMap<Nullifier, Option<NonZeroU32>>,
     /// List of unauthenticated notes that were not found in the store
     pub missing_unauthenticated_notes: Vec<NoteId>,
+    /// The current block height
+    pub current_block_height: Option<u32>,
```
Mirko-von-Leipzig, line 89: Should this be optional?

crates/block-producer/src/txqueue/mod.rs
```diff
@@ -24,14 +24,14 @@ mod tests;
...
-    /// Method to receive a `tx` for processing.
+    /// Method to receive a `tx` for processing and return the current block height.
     ///
     /// This method should:
     /// - Verify the transaction is valid, against the current's rollup state, and also against
     ///   in-flight transactions.
     /// - Track the necessary state of the transaction until it is committed to the `store`, to
     ///   perform the check above.
-    async fn verify_tx(&self, tx: &ProvenTransaction) -> Result<(), VerifyTxError>;
+    async fn verify_tx(&self, tx: &ProvenTransaction) -> Result<Option<u32>, VerifyTxError>;
```
Mirko-von-Leipzig, line 34: Why is the chain tip optional?

Mirko-von-Leipzig (approved):
Just the question on chain height being optional in various places; otherwise lgtm.

crates/block-producer/src/store/mod.rs
bobbinth: Good point! Seems like this should always be a `u32`.

bobbinth (approved):
All looks good! Thank you!

PR #453, opened 2024-08-22
Updated to the latest `next` of `miden-base`

polydez:
Needed to update after merging of https://github.com/0xPolygonMiden/miden-base/pull/822


Mirko-von-Leipzig (approved):
I know we need to update at some point, but why is this required now? Is it blocking some other issue?

PR #456, opened 2024-08-23
Setting up arm builder using the new runner

djpolygon:
Changed name of the amd64 packager

Added arm64 packager that will build on the newly added runner, this may need some iterations to iron out as the arm hosted github runners are still in beta.

### DIFF
--- a/.github/workflows/arm_deb_packager.yml
+++ b/.github/workflows/arm_deb_packager.yml
@@ -1,13 +1,8 @@
 name: arm_deb_packager
 
 on:
-  push:
-    branches: [main, next]
-    paths:
-      - '**'
-    tags:
-      - 'v*.*.*'
-      - 'v*.*.*-*'
+  release:
+    types: [released, prereleased]
 
 jobs:
   build:
--- a/.github/workflows/arm_deb_packager.yml
+++ b/.github/workflows/arm_deb_packager.yml
@@ -1,0 +1,90 @@
+name: arm_deb_packager
+
+on:
+  push:
+    branches: [main, next]
+    paths:
+      - '**'
+    tags:
+      - 'v*.*.*'
+      - 'v*.*.*-*'
+
+jobs:
+  build:
+    permissions:
+      id-token: write
+      contents: write
+    runs-on: ubuntu22-arm-4core
+    steps:
+      - name: Checkout
+        uses: actions/checkout@main
+        with:
+          fetch-depth: 0
+      ##### TAG Variable #####
+      - name: Adding TAG to ENV
+        run: echo "GIT_TAG=`echo $(git describe --tags --abbrev=0)`" >> $GITHUB_ENV
+      - name: adding version
+        run: |
+          NUMERIC_VERSION=$( echo ${{ env.GIT_TAG }} | sed 's/[^0-9.]//g' )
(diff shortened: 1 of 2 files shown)


.github/workflows/arm_deb_packager.yml
```diff
@@ -0,0 +1,90 @@
+name: arm_deb_packager
+
+on:
+  push:
+    branches: [main, next]
+    paths:
+      - '**'
+    tags:
+      - 'v*.*.*'
+      - 'v*.*.*-*'
```
bobbinth, line 10: I think we changed this slightly in the `next` branch ([here](https://github.com/0xPolygonMiden/miden-node/blob/next/.github/workflows/deb_packager.yml)).

bobbinth (commented):
Thank you! I'll defer the full review to @Mirko-von-Leipzig - but I did leave one comment inline. More generally, we may need to rebase this on the `next` branch as this is what we use for development (and this is also where all the latest code is).

.github/workflows/arm_deb_packager.yml
djpolygon: Updated  in the arm packager

.github/workflows/arm_deb_packager.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: adding version
+        run: |
+          NUMERIC_VERSION=$( echo ${{ env.GIT_TAG }} | sed 's/[^0-9.]//g' )
+          echo "VERSION=$NUMERIC_VERSION" >> $GITHUB_ENV
+
+      - name: cleaning repo
+        run: cargo clean
+
+      - name: Building for arm64
+        run: cargo build --release
```
Mirko-von-Leipzig, line 30:
```suggestion
      - name: Building for arm64
        run: |
          cargo build --release --locked --features testing --bin miden-node
          cargo build --release --locked --features testing --bin miden-faucet
```

Mirko-von-Leipzig (commented):
What's missing is an overall strategy to deal with the slight difference in devnet and testnet deploys. I assume we would want to utilize debian packages for both.

For testnet:

1. We click release, this means there is a formal version tag available `vx.y.z`
2. Generate the debian package for `vx.y.z` **and** upload these as assets to the github release
3. Deploy this package to the testnet instance

For devnet:

1. We push to `next` branch
2. Generate the debian package for `next` but **do not** upload them to any release
3. Deploy this package to the devnet instance

--------------------

Otherwise if this is just an initial attempt to see that packaging works (and should be separate from the deployment for now), then its just missing a way to differentiate between testnet and devnet.

Right now this looks decent for a proper release packaging i.e. for testnet only

Mirko-von-Leipzig:
> Thank you! I'll defer the full review to @Mirko-von-Leipzig - but I did leave one comment inline. More generally, we may need to rebase this on the `next` branch as this is what we use for development (and this is also where all the latest code is).

Yes we'll need to do that; but only after these workflows have appeared on `main`. github only acknowledges workflows that are on the default branch i.e. `main` for us. Though once that has happened its somehow happy to run any updates on other branches to these workflows.

Its just the initial "registration" that has to land on `main` first.

Mirko-von-Leipzig (approved):

PR #457, opened 2024-08-23
Update miden-base dependencies to latest

bobbinth:
This PR updates `miden-base` (and related `miden-vm` and `winterfell`) dependencies to the latest version. It also increases MSRV to 1.80.

### DIFF
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -1,7 +1,8 @@
 use std::{cell::RefCell, rc::Rc, time::Duration};
 
 use miden_lib::{
-    accounts::faucets::create_basic_fungible_faucet, notes::create_p2id_note, AuthScheme,
+    accounts::faucets::create_basic_fungible_faucet, notes::create_p2id_note,
+    transaction::TransactionKernel, AuthScheme,
 };
 use miden_node_proto::generated::{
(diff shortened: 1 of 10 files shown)


bin/faucet/src/client.rs
```diff
@@ -314,19 +297,17 @@ fn build_transaction_arguments(
...
-            .replace("{tag}", &Felt::new(tag.into()).to_string())
-            .replace("{amount}", &Felt::new(asset.amount()).to_string()),
-    )
-    .expect("shipped MASM is well-formed");
+    let script = &DISTRIBUTE_FUNGIBLE_ASSET_SCRIPT
+        .replace("{recipient}", &recipient)
+        .replace("{note_type}", &Felt::new(note_type as u64).to_string())
+        .replace("{aux}", &Felt::new(aux).to_string())
+        .replace("{tag}", &Felt::new(tag.into()).to_string())
+        .replace("{amount}", &Felt::new(asset.amount()).to_string());
```
igamigo, line 305: This is missing the execution hint but I believe they are also being implemented on #452 so it should be fine

igamigo (approved):
LGTM!

bin/faucet/src/client.rs
bobbinth: Ah - nice!

PR #461, opened 2024-08-26
Adding in faucet service, updated packagers for faucet, and cleanup

djpolygon:
WIP

Adding the faucet packages to amd64 and arm64
Some clean up

### DIFF
--- a/packaging/miden-faucet.service
+++ b/packaging/miden-faucet.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Miden faucet
+Wants=network-online.target
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Type=exec
+Environment="RUST_LOG=info"
+ExecStart=miden-faucet start --config /etc/miden/miden-faucet.toml
+WorkingDirectory=/opt/miden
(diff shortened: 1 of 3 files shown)


packaging/miden-faucet.service
```diff
@@ -0,0 +1,15 @@
...
+Wants=network-online.target
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Type=exec
+Environment="RUST_LOG=info"
+ExecStart=miden-faucet start --config /etc/miden/miden-faucet.toml
+WorkingDirectory=/opt/miden
```
Mirko-von-Leipzig, line 12: I believe the faucet and node need separate working directories as they both write to disk. So probably we need

```
/opt/miden/miden-node/
/opt/miden/miden-faucet/
```

Mirko-von-Leipzig (approved):

Mirko-von-Leipzig (approved):

packaging/miden-faucet.service
djpolygon: Updated the service files to reflect this

PR #462, opened 2024-08-26
Removed note details from sync responses

polydez:
Resolves https://github.com/0xPolygonMiden/miden-node/issues/309

Introduced `NoteSyncRecord` domain type which omits note details unnecessary for state synchronizations. Refactored code, removed excessive function (`select_notes_since_block_by_tag_and_sender` can work as `select_notes_since_block_by_tag` if we provide empty sender list).

### DIFF
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -7,7 +7,7 @@ use std::{
 use deadpool_sqlite::{Config as SqliteConfig, Hook, HookError, Pool, Runtime};
 use miden_node_proto::{
     domain::accounts::{AccountInfo, AccountSummary},
-    generated::note::Note as NotePb,
+    generated::note::{Note as NotePb, NoteSyncRecord as NoteSyncRecordPb},
 };
 use miden_objects::{
     accounts::AccountDelta,
@@ -83,7 +83,7 @@ impl From<NoteRecord> for NotePb {
 
 #[derive(Debug, PartialEq)]
 pub struct StateSyncUpdate {
-    pub notes: Vec<NoteRecord>,
+    pub notes: Vec<NoteSyncRecord>,
     pub block_header: BlockHeader,
     pub chain_tip: BlockNumber,
     pub account_updates: Vec<AccountSummary>,
@@ -93,9 +93,41 @@ pub struct StateSyncUpdate {
 
 #[derive(Debug, PartialEq)]
 pub struct NoteSyncUpdate {
-    pub notes: Vec<NoteRecord>,
+    pub notes: Vec<NoteSyncRecord>,
     pub block_header: BlockHeader,
     pub chain_tip: BlockNumber,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct NoteSyncRecord {
+    pub block_num: BlockNumber,
+    pub note_index: BlockNoteIndex,
+    pub note_id: RpoDigest,
+    pub metadata: NoteMetadata,
+    pub merkle_path: MerklePath,
+}
+
+impl From<NoteSyncRecord> for NoteSyncRecordPb {
+    fn from(note: NoteSyncRecord) -> Self {
+        Self {
+            note_index: note.note_index.to_absolute_index(),
+            note_id: Some(note.note_id.into()),
+            metadata: Some(note.metadata.into()),
+            merkle_path: Some(Into::into(&note.merkle_path)),
+        }
+    }
+}
+
+impl From<NoteRecord> for NoteSyncRecord {
+    fn from(note: NoteRecord) -> Self {
+        Self {
+            block_num: note.block_num,
+            note_index: note.note_index,
+            note_id: note.note_id,
+            metadata: note.metadata,
+            merkle_path: note.merkle_path,
+        }
+    }
 }
 
 impl Db {
--- a/crates/store/src/db/sql.rs
+++ b/crates/store/src/db/sql.rs
@@ -23,7 +23,8 @@ use rusqlite::{
 };
 
 use super::{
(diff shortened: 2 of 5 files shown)


crates/store/src/db/sql.rs
```diff
@@ -1042,19 +945,14 @@ pub fn get_state_sync(
...
-        let block_num = block_header.block_num();
-        (block_header, block_num)
+    let tip = select_block_header_by_block_num(conn, None)?
+        .ok_or(StateSyncError::EmptyBlockHeadersTable)?;
+    let chain_tip = tip.block_num();
+    let block_header = match notes.first() {
+        None => tip,
+        Some(note) if note.block_num == chain_tip => tip,
+        Some(note) => select_block_header_by_block_num(conn, Some(note.block_num))?
+            .ok_or(StateSyncError::EmptyBlockHeadersTable)?,
```
Mirko-von-Leipzig, line 955: iiuc this could be simplified:

```suggestion
    let block_header = notes.first().map(|note| note.block_num);
    let block_header = select_block_header_by_block_num(conn, block_header)?
        // This error variant should likely be missing block header {n}.
        .ok_or(StateSyncError::EmptyBlockHeadersTable)?;
```

crates/store/src/db/sql.rs
```diff
@@ -1093,21 +991,16 @@ pub fn get_note_sync(
...
+    let notes = select_notes_since_block_by_tag_and_sender(conn, note_tags, &[], block_num)?;
+
+    let tip = select_block_header_by_block_num(conn, None)?
+        .ok_or(NoteSyncError::EmptyBlockHeadersTable)?;
+    let chain_tip = tip.block_num();
+    let block_header = match notes.first() {
+        None => tip,
+        Some(note) if note.block_num == chain_tip => tip,
+        Some(note) => select_block_header_by_block_num(conn, Some(note.block_num))?
+            .ok_or(NoteSyncError::EmptyBlockHeadersTable)?,
```
Mirko-von-Leipzig, line 1003: Same simplification possible as the other one.

crates/store/src/db/mod.rs
```diff
@@ -93,11 +93,43 @@ pub struct StateSyncUpdate {
...
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct NoteSyncRecord {
+    pub block_num: BlockNumber,
+    pub note_index: BlockNoteIndex,
+    pub note_id: RpoDigest,
+    pub metadata: NoteMetadata,
+    pub merkle_path: MerklePath,
+}
+
```
Mirko-von-Leipzig, line 109: Consider moving this to the `proto` crate

Mirko-von-Leipzig (approved):
LGTM, left a couple of suggestions

crates/store/src/db/sql.rs
polydez: Thank you for the suggestion, but we need to return a chain tip from the method, thus we still need two `select_block_header_by_block_num` requests. In my solution we reuse information from the first request only in a case when block header points to the latest block, otherwise, we do second request.

crates/store/src/db/mod.rs
polydez: I was thinking about it and decided to put this struct to the store because it is such an adhoc structure by its nature, but in a case of moving it to proto we would have to implement backward conversion (from proto to domain type) and it's not enough fields in proto to do it (but we can implement something like `impl From<(BlockNumber, NoteSyncRecord)> for NoteSyncRecordPb`).

crates/store/src/db/sql.rs
Mirko-von-Leipzig: Ah right apologies. The github diff view had the return cutoff and I didn't think to look ahead.

crates/store/src/db/sql.rs
Mirko-von-Leipzig: Chain tip is always required for the output so my suggestion doesn't work.

crates/store/src/db/sql.rs
```diff
@@ -1093,21 +991,16 @@ pub fn get_note_sync(
...
-    } else {
-        let block_header = select_block_header_by_block_num(conn, None)?
-            .ok_or(NoteSyncError::EmptyBlockHeadersTable)?;
-
-        let block_num = block_header.block_num();
-        (block_header, block_num)
+    let notes = select_notes_since_block_by_tag_and_sender(conn, note_tags, &[], block_num)?;
+
+    let tip = select_block_header_by_block_num(conn, None)?
+        .ok_or(NoteSyncError::EmptyBlockHeadersTable)?;
```
bobbinth, line 997: Since we know the latest block number in the `state` - we may be able to pass `chain_tip_num` into this function and potentially avoid this database query (i.e., we can always do just 1 query here instead of sometimes doing 2 queries and sometimes 1).

bobbinth (approved):
Looks good! Thank you! I left one potential improvement suggestion inline.

crates/store/src/db/sql.rs
polydez: We don't store latest block number in the `State` in-memory by now or I can't find a place you mean. Should I update our in-memory state in order to save the current block number?

bobbinth: We do have [latest_block_num()](https://github.com/0xPolygonMiden/miden-node/blob/main/crates/store/src/state.rs#L707) function on the `State`. Let's create an issue for this optimization (could be a "good first issue").

polydez: @bobbinth ah, thank you, I forgot about this feature of chain mmr. :)

polydez: @bobbinth done: https://github.com/0xPolygonMiden/miden-node/pull/477

PR #473, opened 2024-08-28
ci: reorganize packaging and release workflows

Mirko-von-Leipzig:
This is a WIP meant to demonstrate a set of workflows which allows us to combine, re-use and separate testnet and devnet deployments.

⚠️ expect naming inconsistencies, typos and other gnarly minutia.

----------------------

The essence of this refactor is creating two actions:

#### 1. Package

This generates the `miden-node` and `miden-faucet` packages -- but notable only generates the files. No uploading or tagging etc. All meta-data like that should come in as an input.

#### 2. Deploy

This action takes in the aws ssm credentials and filepaths to the packages. It then uses `scp` to copy the packages across, and then `ssm` to do the deployment process similar to what is happening now.

----------------------

The workflows are now split into `release` and `deploy_devnet` using the above actions:

1. devnet generates the arm package and then deploys it.
2. release has 3 jobs
    1. generate and upload amd package
    2. generate and upload arm package
    3. wait for arm package to be available and deploy it to testnet

The nice thing is that the actions don't care about architecture or tags etc. That's up to the workflow to setup. And similarly we get to re-use the actions.

----------------------

⚠️ I haven't actually tested the scp via ssm but I believe it is possible according to the docs. And the actions are currently just stubs..

I'd also like an action to wrap the ssm command + polling loop nonsense but that's not super relevant here.

### DIFF
--- a/.github/actions/debian-packages/action.yml
+++ b/.github/actions/debian-packages/action.yml
@@ -1,0 +1,18 @@
+
+name: 'Build Debian packages'
+description: 'Builds miden-node and miden-faucet Debian packages'
+inputs:
+  suffix:
+    required: true
+outputs:
+  node-package:
+    description: "Path to the miden-node package"
+    value: ${{ steps.some-step.outputs.node-package }} 
+  faucet-package:
+    description: "Path to the miden-faucet package"
+    value: ${{ steps.some-step.outputs.faucet-package }} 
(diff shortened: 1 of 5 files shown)


Mirko-von-Leipzig:
Implemented in #546 and #547

PR #475, opened 2024-08-29
Refactoring of the block/batch note trees

polydez:
This is a parallel PR for the https://github.com/0xPolygonMiden/miden-base/pull/834

This fixes the block kernel in order to support updated batch/block note tree structures. I've also slightly refactored test affected by the changes.


Cargo.toml
```diff
@@ -26,7 +26,7 @@ readme = "README.md"
 
 [workspace.dependencies]
 miden-air = { version = "0.10", default-features = false }
-miden-lib = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "next"}
+miden-lib = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "polydez-note-tree-refactor"}
```
Mirko-von-Leipzig, line 29: I guess these will all revert once `miden-base` PR is merged.

crates/store/src/genesis.rs
```diff
@@ -59,7 +58,7 @@ impl GenesisState {
             MmrPeaks::new(0, Vec::new()).unwrap().hash_peaks(),
             account_smt.root(),
             Smt::default().root(),
-            *EmptySubtreeRoots::entry(NOTE_LEAF_DEPTH, 0),
+            *EmptySubtreeRoots::entry(BLOCK_NOTE_TREE_DEPTH, 0),
```
Mirko-von-Leipzig, line 61: Side note: we should consider new-typing the header types. They're all `Digest` / `u32` which is a bit dangerous.

Mirko-von-Leipzig (commented):
LGTM though I'm a bit out-of-context here so I'll wait on someone else to approve formally.

igamigo:
> I thought we were truncating note Merkle paths in the previous solution somewhere, but I'm not seeing it reverted. Though, it is possible that I'm miss-remembering something.

I think you meant [this](https://github.com/0xPolygonMiden/miden-base/blob/159d55a1e9f74dc8451b338076f87a1dfad9a179/objects/src/block/note_tree.rs#L60), which [does not seem to be the case anymore in Denis's PR](https://github.com/0xPolygonMiden/miden-base/blob/48e5f4743c4ff2e842296a963be25fc8b071d0f6/objects/src/block/note_tree.rs#L53-L56).

polydez:
@bobbinth integration test in client work good against my branches.

PR #481, opened 2024-09-03
fix: type shadowing in proto file

tomyrd:
When updating the dependencies in the client we started getting this error when generating the rust types from proto files:
```
Error:   × name 'account.AccountInfo' is shadowed
       ╭─[responses.proto:164:5]
   163 │     // Account info (with details for on-chain accounts)
   164 │     account.AccountInfo account = 1;
       ·     ─────────┬─────────
       ·              ╰── found here
   165 │ }
       ╰────
    help: 'account.AccountInfo' is is resolved to
          'responses.GetAccountDetailsResponse.account.AccountInfo',
          which is not defined. The innermost scope is searched first
          in name resolution. Consider using a leading '.'(i.e.,
          '.account.AccountInfo') to start from the outermost scope.
```

It seems the name of the field `account` is shadowing the name of the import `account`. As it is recommended in the message, this can be fixed by specifying the root.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v0.6.0 (TBD)
 
 - Optimized state synchronizations by removing unnecessary fetching and parsing of note details (#462).
+- [BREAKING] Changed `GetAccountDetailsResponse` field to `details` (#481).
 
 ## 0.5.0 (2024-08-27)
 
--- a/crates/proto/src/generated/responses.rs
+++ b/crates/proto/src/generated/responses.rs
@@ -201,7 +201,7 @@ pub struct ListNotesResponse {
 pub struct GetAccountDetailsResponse {
     /// Account info (with details for on-chain accounts)
     #[prost(message, optional, tag = "1")]
-    pub account: ::core::option::Option<super::account::AccountInfo>,
+    pub details: ::core::option::Option<super::account::AccountInfo>,
 }
 #[allow(clippy::derive_partial_eq_without_eq)]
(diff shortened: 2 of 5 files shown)


SantiagoPittella (approved):

bobbinth (approved):
Thank you! Looks good!

I am not sure I understand why this error is happening though. What is shadowing "account.AccountInfo"? As far as I can tell, there is nothing special about this reference as compared to other references in the same file.

tomyrd:
The `account` name is used in two places in this context:
1. For the name of the first field `GetAccountDetailsResponse`
2. For the name of the module that is imported that contains the type `AccountInfo`

From what I understood, the first name is overwriting (also called shadowing) the second name in this case, so the type `AccountInfo` is being searched inside the field when it should be searched inside the module. By specifying the root with the '.' we are removing this ambiguity and the type will be searched in `.account.AccountInfo`.

This name collision doesn't happen in the other messages and that's why it's only needed here.

tomyrd:
Yes, that would also work but the name should also be changed in the code in that case. I just looked into it and it's not that much different, maybe we prefer the name change instead of the explicit root.

bobbinth:
> I just looked into it and it's not that much different, maybe we prefer the name change instead of the explicit root.

I'm fine either way.

tomyrd:
At the end I decided to change the name to keep the type definition consistent.

bobbinth (approved):
Looks good! Thank you! Could you also update the changelog since now this is a breaking change?

PR #476, opened 2024-08-30
feat: update depedencies

Mirko-von-Leipzig:
Closes #471.

Unfortunately less exciting of an update than I hoped, essentially just `tonic` had any impact; and replaced the `once_cell` with `std` variants.

Some exceptions:

- `rusqlite` is blocked by `deadpool-sqlite`, I'm subscribed to the [PR updating this](https://github.com/bikeshedder/deadpool/pull/347)
- The ecosystem has moved to `http v1` but we still have `v0.2.12` from `actix-web` usage in node and faucet binaries. I recommend moving these to `axum` which is/has become more of the industry standard imo.
- `miden-miette` and various miden-xxx have some outdated deps, don't think anything serious though.

i.e. a couple of deps held back by `rusqlite`'s pool manager, a few other ones by miden internal libs, and quite a lot of somewhat major updates prevented by using `actix`.

List of dependencies that we cannot upgrade but that have updates

```
$ cargo update --verbose

   Unchanged addr2line v0.22.0 (latest: v0.24.1)
   Unchanged bindgen v0.69.4 (latest: v0.70.1)
   Unchanged bit-set v0.5.3 (latest: v0.8.0)
   Unchanged bit-vec v0.6.3 (latest: v0.8.0)
   Unchanged convert_case v0.4.0 (latest: v0.6.0)
   Unchanged cookie v0.16.2 (latest: v0.18.1)
   Unchanged derive_more v0.99.18 (latest: v1.0.0)
   Unchanged event-listener v2.5.3 (latest: v5.3.1)
   Unchanged fixedbitset v0.4.2 (latest: v0.5.7)
   Unchanged generic-array v0.14.7 (latest: v1.1.0)
   Unchanged gimli v0.29.0 (latest: v0.31.0)
   Unchanged h2 v0.3.26 (latest: v0.4.6)
   Unchanged hashbrown v0.12.3 (latest: v0.14.5)
   Unchanged hermit-abi v0.3.9 (latest: v0.4.0)
   Unchanged http v0.2.12 (latest: v1.1.0)
   Unchanged idna v0.5.0 (latest: v1.0.2)
   Unchanged indexmap v1.9.3 (latest: v2.4.0)
   Unchanged itertools v0.11.0 (latest: v0.13.0)
   Unchanged itertools v0.12.1 (latest: v0.13.0)
   Unchanged lalrpop v0.20.2 (latest: v0.21.0)
   Unchanged lalrpop-util v0.20.2 (latest: v0.21.0)
[...]

PR #490, opened 2024-09-11
fix: infinite recursion in Digest::try_into

### LINKED ISSUE #487: [Bug]: The node stops working after a transaction is sent
### Version

next

### Other packages versions

miden-client: next (more specifically on commit `2f8707faeded3725a1d732310f616e3e2170a50d`)

### What happened?

The node works correctly until a transaction from the client is sent and then it "freezes", all of the requests after that don't get a response. No error message is printed in the node console.

We noticed that after reverting the commit `[chore: update dependencies (#476)]` the node didn't have this problem. More specifically if we only revert the commit [Update tonic et al](https://github.com/0xPolygonMiden/miden-node/pull/476/commits/636c6ab1065c0bacc2bc3945eec368527936989f) the node also was suddenly fixed. It might be a problem with the new versions of prost and tonic (the client in next is also using these new versions but it seems that it doesn't cause any issues).

The node should have processed the transaction and continue responding.

With a node (from branch next) running locally. Run the command `make integration-test` in from the root of the `miden-client` repo (also in branch next).

### Relevant log output

```shell
The last log printed is always the same:


2024-09-06T16:35:45.727004Z  INFO store:get_block_inputs: miden-store: crates/store/src/server/api.rs:383: new
2024-09-06T16:35:45.727019Z  INFO store:get_block_inputs:validate_nullifiers: miden-store: crates/store/src/server/api.rs:591: new
2024-09-06T16:35:45.727028Z  INFO store:get_block_inputs:validate_nullifiers: miden-store: crates/store/src/server/api.rs:591: close, time.busy: 1.42µs, time.idle: 7.83µs
2024-09-06T16:35:45.727038Z  INFO store:get_block_inputs:validate_notes: miden-store: crates/store/src/server/api.rs:601: new
2024-09-06T16:35:45.727046Z  INFO store:get_block_inputs:validate_notes: miden-store: crates/store/src/server/api.rs:601: close, time.busy: 1.04µs, time.idle: 6.62µs
2024-09-06T16:35:45.727058Z  INFO store:get_block_inputs:select_block_header_by_block_num: miden-store: crates/store/src/db/mod.rs:249: new
2024-09-06T16:35:45.727281Z  INFO store:get_block_inputs:select_block_header_by_block_num: miden-store: crates/store/src/db/mod.rs:249: close, time.busy: 22.5µs, time.idle: 201µs
2024-09-06T16:35:45.727317Z  INFO store:get_block_inputs:select_note_inclusion_proofs: miden-store: crates/store/src/db/mod.rs:370: new
2024-09-06T16:35:45.727395Z  INFO store:get_block_inputs:select_note_inclusion_proofs: miden-store: crates/store/src/db/mod.rs:370: close, time.busy: 12.6µs, time.idle: 65.4µs
2024-09-06T16:35:45.727409Z  INFO store:get_block_inputs:select_block_headers: miden-store: crates/store/src/db/mod.rs:265: new
2024-09-06T16:35:45.727493Z  INFO store:get_block_inputs:select_block_headers: miden-store: crates/store/src/db/mod.rs:265: close, time.busy: 12.3µs, time.idle: 72.2µs
2024-09-06T16:35:45.727517Z  INFO store:get_block_inputs: miden-store: crates/store/src/server/api.rs:383: close, time.busy: 191µs, time.idle: 322µs


Then it just stops responding and no more logs are printed
```

Mirko-von-Leipzig:
`TryFrom` implementation called `T::try_into` instead of `(*T)::try_into()` causing an infinite loop as both `T` and `&T` have this implemented.

This was introduced as part of a manual clippy fix (by me) in #476 where `x.clone().try_into()` was manually changed to `x.try_into()` instead of the suggested  `(*x).try_into()`. Does not affect any existing releases.

Fixes #487.

---------------------------

Side note: I do think implementing `TryFrom` for both owned and reference variants an anti-pattern.

### DIFF
--- a/crates/proto/src/domain/digest.rs
+++ b/crates/proto/src/domain/digest.rs
@@ -187,7 +187,7 @@ impl TryFrom<&digest::Digest> for [Felt; 4] {
     type Error = ConversionError;
 
     fn try_from(value: &digest::Digest) -> Result<Self, Self::Error> {
-        value.try_into()
+        (*value).try_into()
     }
 }
 
@@ -195,7 +195,7 @@ impl TryFrom<&digest::Digest> for Digest {
     type Error = ConversionError;
 
     fn try_from(value: &digest::Digest) -> Result<Self, Self::Error> {
-        value.try_into()
+        (*value).try_into()
     }
 }
 
--- a/crates/proto/src/domain/digest.rs
+++ b/crates/proto/src/domain/digest.rs
@@ -3,7 +3,7 @@ use std::fmt::{Debug, Display, Formatter};
 use hex::{FromHex, ToHex};
 use miden_objects::{notes::NoteId, Digest, Felt, StarkField};
 
-use crate::{errors::ConversionError, generated::digest};
+use crate::{errors::ConversionError, generated::digest as proto};
 
 // CONSTANTS
 // ================================================================================================
@@ -13,19 +13,19 @@ pub const DIGEST_DATA_SIZE: usize = 32;
 // FORMATTING
 // ================================================================================================
 
-impl Display for digest::Digest {
+impl Display for proto::Digest {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         f.write_str(&self.encode_hex::<String>())
     }
 }
 
-impl Debug for digest::Digest {
+impl Debug for proto::Digest {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         Display::fmt(self, f)
     }
 }
 
-impl ToHex for &digest::Digest {
+impl ToHex for &proto::Digest {
     fn encode_hex<T: FromIterator<char>>(&self) -> T {
         (*self).encode_hex()
     }
@@ -35,7 +35,7 @@ impl ToHex for &digest::Digest {
     }
 }
 
-impl ToHex for digest::Digest {
+impl ToHex for proto::Digest {
     fn encode_hex<T: FromIterator<char>>(&self) -> T {
         let mut data: Vec<char> = Vec::with_capacity(DIGEST_DATA_SIZE);
         data.extend(format!("{:016x}", self.d0).chars());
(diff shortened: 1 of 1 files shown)


crates/proto/src/domain/digest.rs
```diff
@@ -187,15 +187,15 @@ impl TryFrom<&digest::Digest> for [Felt; 4] {
     type Error = ConversionError;
 
     fn try_from(value: &digest::Digest) -> Result<Self, Self::Error> {
```
bobbinth, line 189: Could we alias `digest` so that this would read as something like:
```rust
fn try_from(value: &proto::Digest) -> Result<Self, Self::Error> {
```

bobbinth (approved):
Looks good! Thank you for tracking it down! I left one minor comment inline.

> Side note: I do think implementing `TryFrom` for both owned and reference variants an anti-pattern.

How should we change it?

Mirko-von-Leipzig:
> Looks good! Thank you for tracking it down! I left one minor comment inline.
> 
> > Side note: I do think implementing `TryFrom` for both owned and reference variants an anti-pattern.
> 
> How should we change it?

Either

1. Only implement it for the owned variant (very rarely the reference version - only if no ownership of the data is required for some reason), or
2. named constructions/functions on the type itself

bobbinth:
> 1. Only implement it for the owned variant (very rarely the reference version - only if no ownership of the data is required for some reason), or

If we implement it only for the owned variant, wouldn't it imply cloning in cases when we have only a reference? Or would the compiler compile this away?

Mirko-von-Leipzig:
> > 1. Only implement it for the owned variant (very rarely the reference version - only if no ownership of the data is required for some reason), or
> 
> If we implement it only for the owned variant, wouldn't it imply cloning in cases when we have only a reference? Or would the compiler compile this away?

It depends on the implementation. If all the inner data requires cloning anyway then we shouldn't implement it for the `&T`. If some/most of the data can be taken by reference then we should only implement it for `&T`.

But maybe I'm just being overly picky because of this bug :) In this case we implemented both for a `Copy` type - and where the `&T` was internally "cloning" which isn't good. Generally we want `Clone` calls to be surfaced at the call site.

PR #494, opened 2024-09-12
Updated to the latest `next` branch of `miden-objects`

polydez:


crates/store/src/db/tests.rs
```diff
@@ -335,7 +333,10 @@ fn test_sql_public_account_details() {
     let non_fungible_faucet_id =
         AccountId::try_from(ACCOUNT_ID_NON_FUNGIBLE_FAUCET_ON_CHAIN).unwrap();
 
-    let mut storage = AccountStorage::new(vec![], BTreeMap::new()).unwrap();
+    let mut storage = AccountStorage::new(
+        std::iter::repeat(StorageSlot::Value(Word::default())).take(6).collect(),
```
Mirko-von-Leipzig, line 337: Why is this required now? Is the storage slot otherwise uninitialized and setting it fails now?

Mirko-von-Leipzig (approved):

crates/store/src/db/tests.rs
polydez: Yes, indeed, otherwise we get `slot out of range` error. @phklive, is it a bug or expected behavior?

phklive: Yes exactly you build your storage with it's different storage slots at initialisation. 

We cannot set to uninitialised storage slots, `set_` is only used to update existing storage slots.

Issue #145, opened 2024-01-15
Add healthchecking


hackaugusto:
There are multiple ways and purposes of doing healthchecks. I think we need two types:

- Health check to be used by a process manager, that can restart the process if it is unhealthy
- Health check for an external system, that can collect metrics and alert when the system is misbehaving

The second case should be based on metrics https://github.com/0xPolygonMiden/miden-node/issues/144, either because the system is not replying, or some important metric is not changing (e.g. the block number for the store and block producer).

The first case, should have native implementations, namely:

- [sdnotify](https://www.man7.org/linux/man-pages/man3/sd_notify.3.html) for systemd
- a control plane command for [docker](https://docs.docker.com/engine/reference/builder/#healthcheck) [related: #83].
    - IMO it makes sense to [use this strategy](https://github.com/hyperium/tonic/tree/master/examples/src/health) 
- and maybe an http interface for third-party process managers (probably doesn't have to be done now)

Mirko-von-Leipzig:
Implemented indirectly as a result of #675 - we use trace information in the monitoring backend to trigger health warnings and alerts.

Issue #95, opened 2023-12-11
Add interface to interact with the node servers


hackaugusto:
Implement a small CLI interface to interact and inspect the state of the servers. This is not the same as the miden wallet/client, it is intended to interact with all the endpoints of the internal servers, used for testing.

- Decide on the format, it could be a small CLI written in rust, [postman](https://blog.postman.com/postman-now-supports-grpc/), etc
- Implement the front end for the existing servers

hackaugusto:
Note: I kinda enjoy having these commands on the RPC server https://github.com/0xPolygonMiden/miden-node/blob/main/rpc/src/cli/mod.rs#L36-L43 . I think it would make sense to have something similar for the others services.

One advantage of using the CLI is that we can hex encode binary data, and use a `mock` tool https://github.com/0xPolygonMiden/miden-node/issues/96 to create it, something like (rough sketch, haven't thought about the API much):

```
miden-rpc account create $(mock account new --fungible)
```

bobbinth:
Re format, I'd probably try to see if we can use Postman first, and if that proves to be too limiting, write a simple CLI.

> Note: I kinda enjoy having these commands on the RPC server https://github.com/0xPolygonMiden/miden-node/blob/main/rpc/src/cli/mod.rs#L36-L43 . I think it would make sense to have something similar for the others services.

I like these too - but how would we use them in the context of a single binary?

hackaugusto:
related to: https://github.com/0xPolygonMiden/miden-node/issues/137

hackaugusto:
> Re format, I'd probably try to see if we can use Postman first, and if that proves to be too limiting, write a simple CLI.

I'm hijinking this issue to be CLI only, we have #137 for a GUI

okcan:
It may be useful to use the CLI structure here, as found in some of our projects.

hackaugusto:
@okcan are there any operations you would consider a must to be implemented in the CLI? So far I have only considered adding graceful shutdown and hot-config reloading. Other

okcan:
Actually, it would be more useful to ask this to a blockchain developer, but I can give these projects as examples that I know of.
https://github.com/maticnetwork/matic-cli
https://github.com/maticnetwork/polygon-cli

Mirko-von-Leipzig:
Closing as stale. We can revisit this at a later stage.

Issue #83, opened 2023-12-05
add control plane


hackaugusto:
Ideally services would be crash first, since that is th easiest model to work with. but unfortunately this won't serve all the servers:

- the store creates in-memory representations of the trees, which will take a bit of time to create, so for that service it is best to avoid restarts
- the rpc server works as a reverse proxy, restarting the service would close all client connections, which is not great for end-user performance 
- the current version of the block producer contains only in-memory store of transactions, so restarts would cause issues, eventually this will be lifted, but we should probably avoid losing in-flight transactions if we can

because of the above, we should have a way of managing the servers without requiring restarts. there are multiple ways this can be implemented:

- watch dog over the configuration files, so when the file is written, the server reloads the configuration and applies the changes
    - this is nice and has a neat interface, but it can become an issue for cases were the user introduces syntax errors to the file
- a control socket. this can be a unix socket (file socket), or a proper networked socket. 
    - this is a neat approach, usually the socket is hidden behind a CLI program. it can be useful for settings that should not be persisted
- system signals. like sigusr1
    - this is great to integrate the servers with process monitors, works well with systemd and other tools, but we have limited number of signals

I don't think any of the options above are ideal, usually a mix of each gives the best results

hackaugusto:
for file watching it seems the best option is [notify](https://lib.rs/crates/notify), it is wildly used (see the docs). The only other option I found was [watchman](https://lib.rs/crates/watchman_client), but that requires an external service.

the only downside is that the api seems to be sync, so we should probably nave a file watching "service" running on a separate thread, sending events to the async tasks in tokia via channels

Mirko-von-Leipzig:
No longer directly applicable since we don't use configuration files anymore.

We can open new issues if we find a need for this sort of thing.

PR #495, opened 2024-09-12
feat: improve `--version` for non-releases

### LINKED ISSUE #493: Improve version strings
Improve the output of our `miden-xxx --version` output to account for non-release builds, such as we might deploy on devnet instances.

Use `git describe --tags --dirty` which will give us the format

```
<version-tag>-<commit-count>-<commit SHA>-dirty
```

where everything but `<version-tag>` is only present if applicable.

This will also help us debug operator node's as we can now know if they're on a non-release build.

Mirko-von-Leipzig:
This PR improves `--version` outputs for the node and faucet by including extra information for non-release builds.

`--version` is now set to `git describe --tags --dirty` which is no change for formal releases (since `tag == package version`), but for non-releases indicates the previous tag, commit count since that tag, commit SHA and appends `-dirty` if it was built with uncommitted changes.

Motivation is given in #493.

Some points of contention:

1. I added this to the `miden-node-utils` crate. However, since this is a `build.rs` that depends on `git` data, this means the utils crate will be rebuilt after every `git` change i.e. this will cause a rebuild of utils and all of its dependents. It is probably better if this became its own top level crate - I decided not to do this because I didn't want to publish such a minor crate. But it might make sense to just do it. Alternatively, we can add this `build.rs` to every binary we create instead.
2. The tag found is sensitive. As an example, `git describe --tags` on the `next` branch currently returns `v0.4.0` because the `v0.5.0` tag on `main` has not been merged back yet. Just something to consider for our release "process" :). This differs significantly to the current output which is `v0.6.0` (the in-progress package version).

The `--dirty` flag currently doesn't work unfortunately. I've opened an [issue](https://github.com/rustyhorde/vergen/issues/376) for this in `vergen`.

We can also consider additional build descriptions. `vergen` has an illustrative example for verbose version info, which provides information that could be useful for debugging/reproduction purposes:

```
~/p/r/app λ app -vv
app 0.1.0

Build Timestamp:     2021-02-23T20:14:46.558472672+00:00
rustc Version:       1.52.0-nightly
rustc Channel:       nightly
rustc Host Triple:   x86_64-unknown-linux-gnu
rustc Commit SHA:    3f5aee2d5241139d808f4fdece0026603489afd1
cargo Target Triple: x86_64-unknown-linux-musl
cargo Profile:       release
```

Closes #493


Mirko-von-Leipzig:
Can also consider creating a status page for the node. Something to monitor the status of the node and the metadata like block height, node version etc.

Right now we can only "monitor" via the client or submitting a note via the faucet webpage.

bobbinth:
I'm not fully understanding this yet - but a couple of questions/comments:

The effect of this is that we add more info to the `--version` command, right? And this info now going be based on Git tags rather than `Cargo.toml`? Or is it going to combine the two somehow?

Generally, I think it is good to add more details to the version like you have with the `vargen` example, but I think what gets printed out should be consistent with `crates.io` version. That is, if `Cargo.toml` says `0.6.1` - and then we can add more info to it.

> I added this to the `miden-node-utils` crate. However, since this is a `build.rs` that depends on `git` data, this means the utils crate will be rebuilt after every `git` change i.e. this will cause a rebuild of utils and all of its dependents. It is probably better if this became its own top level crate - I decided not to do this because I didn't want to publish such a minor crate. But it might make sense to just do it. Alternatively, we can add this `build.rs` to every binary we create instead.

I also probably won't publish a separate crate just for this. Given that we have only 2 binaries, I'm wondering if putting this just in `build.rs` of the binaries is a better option?

> Can also consider creating a status page for the node. Something to monitor the status of the node and the metadata like block height, node version etc.
> 
> Right now we can only "monitor" via the client or submitting a note via the faucet webpage.

Yes, that would be great. This is probably somewhat related to #145 and maybe #83 and #95. But maybe it is worth creating a separate issue (or maybe something to superseded some of these mentioned issues).

Mirko-von-Leipzig:
> I'm not fully understanding this yet - but a couple of questions/comments:
> 
> The effect of this is that we add more info to the `--version` command, right? 

Correct, more specifically it uses the output of `git describe --tags --dirty` which is injected at compile time via the `build.rs`. The output format looks like this:

```python
# If at a tagged commit, only the tag is output
version=TAG             # e.g. v0.5.0

# At commit with <SHA> hash, N commits after the last tag
version=TAG-N-SHA.      # e.g. v0.5.0-12-ABCDEFG

# In addition if the repo contains uncommitted changes then a `-dirty` suffix is added.
version=TAG-dirty.      # e.g. v0.5.0-dirty if directly on a tagged commit, or 
version=TAG-N-SHA-dirty # e.g. v0.5.0-12-ABCDEFG-dirty.
```

> And this info now going be based on Git tags rather than `Cargo.toml`? Or is it going to combine the two somehow?

This is based entirely on `git`. If we publish `v0.7.0` but never tag it then the `--version` output will be based on whatever the last available tag is plus the deltas. Tagging the release is considered best practice by both `crates.io` and `github` so a gap here would really indicate an issue on our end. However this can be rectified post-publish/release by just tagging it.

And for the binaries in which this was wrong, we would still have access to the correct commit `SHA`.

> Generally, I think it is good to add more details to the version like you have with the `vargen` example, but I think what gets printed out should be consistent with `crates.io` version. That is, if `Cargo.toml` says `0.6.1` - and then we can add more info to it.

This should always be the case unless we screw up the tags. Basing it on the manifest won't work for non-releases as the manifest is always bumped immediately after the previous release, meaning this would be "incorrect" for the duration of `next`.

If the potential error between manifest and git tag is an issue, then we can leave things as is and instead add this information as extended metadata for the long version output. Non-releases will have the top level version output, but the correct data will still be present for [long version](https://docs.rs/clap/latest/clap/struct.Command.html#method.long_version). Though I would argue that it's confusing :)

> I also probably won't publish a separate crate just for this. Given that we have only 2 binaries, I'm wondering if putting this just in `build.rs` of the binaries is a better option?

I think that's probably the correct call. Will update first thing next week.

Mirko-von-Leipzig:
I should also add a tag filter so that only version related tags are considered. e.g. if we ever use tags for other things then those will be ignored.

bobbinth:
> This is based entirely on `git`. If we publish `v0.7.0` but never tag it then the `--version` output will be based on whatever the last available tag is plus the deltas. Tagging the release is considered best practice by both `crates.io` and `github` so a gap here would really indicate an issue on our end. However this can be rectified post-publish/release by just tagging it.

Currently, my process is a bit different - though, it can of course be changes. Specifically, I usually first publish crates to crates.io and then tag the release.

Also, sometimes I publish patches of individual crates without creating tags. For example, if there was a bug fixed in say the block producer crate, I would publish it under `v0.6.1` but keep the rest of the crates at `v0.6.0` and the tag would remain at `v0.0.6.0`. Maybe this is not a good practice though.



> If the potential error between manifest and git tag is an issue, then we can leave things as is and instead add this information as extended metadata for the long version output. Non-releases will have the top level version output, but the correct data will still be present for [long version](https://docs.rs/clap/latest/clap/struct.Command.html#method.long_version). Though I would argue that it's confusing :)

To clarify, would the short version show something like:
```
miden-node 0.6.1
```
But the long version would show something like:
```
miden-node 0.6.1

Build Timestamp:     2021-02-23T20:14:46.558472672+00:00
rustc Version:       1.52.0-nightly
rustc Channel:       nightly
rustc Host Triple:   x86_64-unknown-linux-gnu
rustc Commit SHA:    3f5aee2d5241139d808f4fdece0026603489afd1
cargo Target Triple: x86_64-unknown-linux-musl
cargo Profile:       release
```
If so, I don't think this is too confusing. But would it satisfy the original goal of this PR?

Mirko-von-Leipzig:
> Currently, my process is a bit different - though, it can of course be changes. Specifically, I usually first publish crates to crates.io and then tag the release.

Tagging afterwards is fine - `crates.io` doesn't actually store any binaries; so each subsequent install will checkout the repo and acquire the tag. There would be a small window prior to tagging where users might get this without the tag.

> Also, sometimes I publish patches of individual crates without creating tags. For example, if there was a bug fixed in say the block producer crate, I would publish it under `v0.6.1` but keep the rest of the crates at `v0.6.0` and the tag would remain at `v0.0.6.0`. Maybe this is not a good practice though.

This version output only applies to the executables. So long as the tags match the intended versions of the node/faucet we would be okay. There are other options - one can also override these env variables during the build process if required.

I've reverted this change for now; I've only added additional metadata to the long version. The actual version used is now the crate/manifest version. The metadata will give us the additional data we need to accurately debug. We just need to be aware that a user claiming `x.y.z` version might be referring either to the release version, or the in-progress version.

> To clarify, would the short version show something like:
> 
> ```
> miden-node 0.6.1
> ```
> 
> But the long version would show something like:
> 
> ```
> miden-node 0.6.1
> 
> Build Timestamp:     2021-02-23T20:14:46.558472672+00:00
> rustc Version:       1.52.0-nightly
> rustc Channel:       nightly
> rustc Host Triple:   x86_64-unknown-linux-gnu
> rustc Commit SHA:    3f5aee2d5241139d808f4fdece0026603489afd1
> cargo Target Triple: x86_64-unknown-linux-musl
> cargo Profile:       release
> ```
> 
> If so, I don't think this is too confusing. But would it satisfy the original goal of this PR?

I've updated the PR so that the short version output is:

```
$ miden-node -V

miden-node 0.6.0
```
and the long
```
$ miden-node --version

miden-node 0.6.0

SHA:          8d9602c-dirty
branch:       mirko-versionify
features:     testing
rust version: 1.80.1
target arch:  aarch64-apple-darwin
host arch:    aarch64-apple-darwin
opt-level:    3
```

Feel free to suggest a better formatting; I tried with prefixes like `git.XXX` and `build.YYY` or a more toml-like format but in the end I don't care that much :).

I've only implemented this for the `miden-node` binary for now; I'll copy it to the faucet once we have consensus.

Something else I was hoping to determine that it was built with `--locked` but unfortunately this [isn't possible](https://users.rust-lang.org/t/locked-status-at-compile-time/117629/2).

bin/node/src/main.rs
```diff
@@ -122,3 +122,41 @@ async fn main() -> anyhow::Result<()> {
         },
     }
 }
+
+/// Generates the `--version` output enriched with additional build metadata.
+fn long_version() -> String {
```
bobbinth, line 127: Could this function (or most of it) go into the utils crate? As far as I can tell, it mostly just looks up environment variables and joins them together in a formatted string.

bobbinth (approved):
The format looks great! Thank you! I left one small comment inline, and I think once we propagate this to the faucet as well, we are good to merge.

bobbinth:
I guess one thing to confirm: how would it work when the crate is pulled from crates.io? For example, how would `SHA` and `branch` fields be set in this case?

bin/node/src/main.rs
Mirko-von-Leipzig: Oh that's a great point. I was dreading copy-pasting this around and completely lost sight of the fact that its just the `build.rs` that needs to be recompiled.

Mirko-von-Leipzig:
> I guess one thing to confirm: how would it work when the crate is pulled from crates.io? For example, how would `SHA` and `branch` fields be set in this case?

I checked using `cargo publish --dry-run ...` and the git data is indeed just gone. For the current implementation it prints:

```
/miden-node --version`
miden-node 0.6.0

SHA:          VERGEN_IDEMPOTENT_OUTPUT
branch:       VERGEN_IDEMPOTENT_OUTPUT
features:     None
rust version: 1.81.0
target arch:  aarch64-apple-darwin
host arch:    aarch64-apple-darwin
opt-level:    0
debug:        true
```

which isn't ideal but I believe this should let us determine whether its a formal release/install at least? There are of course hacks around this if people want to abuse it but that is what it is.

bobbinth:
> which isn't ideal but I believe this should let us determine whether its a formal release/install at least? There are of course hacks around this if people want to abuse it but that is what it is.

Should we detect this and not include the missing fields in the output? Would still make it clear if this is a release/install, but also would avoid exposing implementation details (i.e., that we use `vergen`).

Mirko-von-Leipzig:
I've added a check for the published state (so we still get sha data) and also override the default vergen output in case it somehow slips through.

I've copy pasted this to the faucet instead of the utils crate for the reason given in this [comment](https://github.com/0xPolygonMiden/miden-node/pull/495#discussion_r1762665364).

I've tested that it works for `cargo publish` locally, though I only tested `miden-node` and not the faucet. Testing is a bit of a pita because publish/package requires all dependencies to have a version specified which of course we don't (all point to `next` for `miden-base`):

```
﻿﻿﻿error: all dependencies must have a version specified when packaging.
```

bin/faucet/build.rs
```diff
@@ -1,12 +1,104 @@
...
+
+        // File format:
+        // {
+        //   "git": {
+        //     "sha1": "9d48046e9654d93a86212e77d6c92f14c95de44b"
+        //   },
+        //   "path_in_vcs": "bin/node"
+        // }
+        let offset =
+            contents.find(r#""sha1""#).expect("Searching for sha1 property") + r#""sha1""#.len();
```
Mirko-von-Leipzig, line 57: I'm wondering if I should be more defensive about these `.expect` calls. The file does exist so we should expect its format to be okay.. However maybe it would be better to just return a string like `"Failed to parse"`.

crates/block-producer/src/skele/mod.rs
bobbinth, line 1: I think this file is probably not relevant to this PR.

bin/node/Cargo.toml
```diff
@@ -19,7 +19,7 @@ tracing-forest = ["miden-node-block-producer/tracing-forest"]
 
 [dependencies]
 anyhow = { version = "1.0" }
-clap = { version = "4.5", features = ["derive"] }
+clap = { version = "4.5", features = ["derive", "string"] }
```
bobbinth, line 22: Question: why do we need to add `string` dependency here, but we don't need to do the same for the faucet binary?

bin/node/build.rs
```diff
@@ -0,0 +1,95 @@
...
+            // Take until closing quote.
+            .take_while(|&c| c != '"')
+            // Short SHA format is 7 digits.
+            .take(7)
+            .collect();
+
+        Some(sha1)
+    } else {
+        None
+    }
```
bobbinth, line 64: Would it make sense to make this a stand alone function that lives in the utils crate? This way, at least the file parsing logic would be shared between different binaries.

bin/node/src/main.rs
```diff
@@ -122,3 +122,52 @@ async fn main() -> anyhow::Result<()> {
...
+
+SHA:          {sha}
+branch:       {branch}
+features:     {features}
+rust version: {rust_version}
+target arch:  {target}
+host arch:    {host}
+opt-level:    {opt_level}
+debug:        {debug}
+"
```
bobbinth, line 171: Maybe we should create a helper struct in the utils crate - something like `LongVersion` - that will encapsulate some of this logic? We could then still use `env!` macros here like:
```rust
let long_version = LongVersion {
    version: env!("CARGO_PKG_VERSION"),
    branch: env!("VERGEN_GIT_BRANCH"),
    sha: env!("VERGEN_GIT_SHA"),
    rust_version: env!("VERGEN_RUSTC_SEMVER"),
    host: env!("VERGEN_RUSTC_HOST_TRIPLE"),
    target: env!("VERGEN_CARGO_TARGET_TRIPLE"),
    features: env!("VERGEN_CARGO_FEATURES"),
    opt_level: env!("VERGEN_CARGO_OPT_LEVEL"),
    debug: env!("VERGEN_CARGO_DEBUG"),
};

long_version.to_string()
```

bin/faucet/build.rs
bobbinth: Yeah - I'd probably be more on the defensive side here so that we can avoid panics here for w/e unforeseen reason.

bobbinth (commented):

crates/block-producer/src/skele/mod.rs
Mirko-von-Leipzig: Eugh sorry - I should be more careful with `git squash`.

bin/node/Cargo.toml
Mirko-von-Leipzig: It is needed by both but its included automatically due to feature unification. Removing it causes both `node` and `faucet` to fail, including it in either then both work.

Adding to `faucet` as well since that's more correct.

Mirko-von-Leipzig: To be more specific, building with individual feature resolution would fail.

```
# This works as by default features are unified at workspace level.
cargo build --bin miden-faucet

# This fails
cargo build -p miden-faucet --bin miden-faucet
```

Maybe there is something to be said about building each binary in isolation in CI?

Mirko-von-Leipzig:
I've moved everything into the `utils` crate by:

- Feature gating the build script stuff behind `vergen`. We can remove the feature if we feel okay with leaking the build-script deps into normal builds 🤷. With dependency resolution this doesn't matter much in the end.
- Using `option_env!` which then defaults to an empty string if the environment variables aren't present. This means this function compiles even if the build script hasn't been run.

bin/node/src/main.rs
Mirko-von-Leipzig: I moved everything instead :)

bin/node/build.rs
Mirko-von-Leipzig: I moved everything instead.

bin/faucet/build.rs
Mirko-von-Leipzig: Done. Build scripts also just succeed with a warning on failure.

crates/utils/Cargo.toml
```diff
@@ -22,3 +22,12 @@ tonic = { workspace = true }
...
 tracing-subscriber = { workspace = true }
+
+# Optional dependencies enabled by `vergen` feature.
+# This must match the version expected by `vergen-gitcl`.
+vergen = { "version" = "9.0", optional = true }
+vergen-gitcl = { version = "1.0", features = ["cargo", "rustc"], optional = true }
+
+[features]
+# Enables depedencies intended for build script generation of version metadata.
+vergen = ["dep:vergen", "dep:vergen-gitcl"]
```
bobbinth, line 33: nit: we usually put `features` section above the `dependencies` section.

bin/node/Cargo.toml
bobbinth: Hmmm - don't have a strong opinion here.

bobbinth (approved):
Looks good! Thank you! I left just one optional nit inline.

bin/node/Cargo.toml
Mirko-von-Leipzig: Looks like this is somewhat easy to perform using [cargo hack](https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#--all---workspace).

If one just uses `check` then it shouldn't take much time to perform for each package. Not terribly important though.

Mirko-von-Leipzig: #501

PR #496, opened 2024-09-13
feat: Add kernel root to blocks

igamigo:


crates/block-producer/src/block_builder/prover/mod.rs
```diff
@@ -72,6 +72,7 @@ impl BlockProver {
             nullifier_root,
             note_root,
             tx_hash,
+            Digest::default(),
```
bobbinth: If we set it to the default digest, we won't be able to execute any transactions against this block. Instead, we should set `kernel_root` to `TransactionKernel::kernel_root()`.

igamigo: Yeah, I set it to that just a minute before.

crates/proto/src/domain/blocks.rs
bobbinth, line 1: Not from this PR, but since we are making changes here, I would alias `generated::block` import as `proto`. This way, something like:
```rust
impl From<&BlockHeader> for block::BlockHeader
```
Would become:
```rust
impl From<&BlockHeader> for proto::BlockHeader
```

proto/block.proto
```diff
@@ -23,8 +23,10 @@ message BlockHeader {
     digest.Digest tx_hash = 8;
     // a hash of a STARK proof attesting to the correct state transition.
     digest.Digest proof_hash = 9;
+    // an accumulative hash from all kernel hashes.
+    digest.Digest kernel_root = 10;
```
bobbinth, line 27: nit: I would change this to: "a commitment to all transaction kernels supported by this block."

bobbinth (approved):
Looks good! Thank you! I left a couple of minor comments inline.

proto/block.proto
igamigo: Changed. Maybe we want to change it on `miden-base` as well?

crates/proto/src/domain/blocks.rs
igamigo: Changed

proto/block.proto
bobbinth: Good idea! Done in https://github.com/0xPolygonMiden/miden-base/pull/850/commits/a9938cd0cf3480ea8371393f250c36807404d18d.

bobbinth, line 27: hmmm - this seems to have reverted itself.

bobbinth (approved):
All looks good! Thank you!

PR #498, opened 2024-09-17
chore: Update to miden-base's next

igamigo:
Follows https://github.com/0xPolygonMiden/miden-base/pull/875

### DIFF
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -29,6 +29,7 @@ use miden_tx::{
 use rand::{rngs::StdRng, thread_rng, Rng};
 use rand_chacha::{rand_core::SeedableRng, ChaCha20Rng};
 use tonic::transport::Channel;
+use tracing::{error, info};
 
 use crate::{config::FaucetConfig, errors::FaucetError};
 
@@ -113,9 +114,10 @@ impl FaucetClient {
             .executor
             .execute_transaction(self.id, 0, &[], transaction_args)
             .map_err(|err| {
+                error!("{err}");
                 FaucetError::InternalServerError(format!("Failed to execute transaction: {}", err))
             })?;
-
+        
         Ok((executed_tx, output_note))
     }
 
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -29,7 +29,6 @@ use miden_tx::{
 use rand::{rngs::StdRng, thread_rng, Rng};
 use rand_chacha::{rand_core::SeedableRng, ChaCha20Rng};
 use tonic::transport::Channel;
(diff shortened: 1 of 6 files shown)


bin/faucet/src/config.rs
```diff
@@ -39,8 +34,8 @@ impl FaucetConfig {
 impl Display for FaucetConfig {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         f.write_fmt(format_args!(
-            "{{ endpoint: \"{}\",  database_filepath: {:?}, asset_amount_options: {:?}, token_symbol: {}, decimals: {}, max_supply: {} }}",
-            self.endpoint, self.database_filepath, self.asset_amount_options, self.token_symbol, self.decimals, self.max_supply
+            "{{ endpoint: \"{}\", asset_amount_options: {:?}, token_symbol: {}, decimals: {}, max_supply: {} }}",
+            self.endpoint, self.asset_amount_options, self.token_symbol, self.decimals, self.max_supply
```
igamigo, line 38: The database path was used for the client's database so I opted to remove it since it was not needed anymore.

crates/store/src/genesis.rs
```diff
@@ -60,7 +61,7 @@ impl GenesisState {
             Smt::default().root(),
             *EmptySubtreeRoots::entry(BLOCK_NOTE_TREE_DEPTH, 0),
             Digest::default(),
-            Digest::default(),
```
igamigo, line 63: Faucet was failing to execute correctly due to this being an invalid kernel root.

bin/faucet/src/client.rs
```diff
@@ -113,6 +114,7 @@ impl FaucetClient {
             .executor
             .execute_transaction(self.id, 0, &[], transaction_args)
             .map_err(|err| {
+                error!("{err}");
                 FaucetError::InternalServerError(format!("Failed to execute transaction: {}", err))
```
bobbinth, line 118: Was logging not working correctly here? Unless there is a good reason, I'd rather not add this ad-hoc. Maybe worth creating an issue for this?

bin/faucet/src/config.rs
bobbinth: This is because we no longer rely on the client here, right?

bobbinth (approved):

bin/faucet/src/config.rs
igamigo: Correct!

bin/faucet/src/client.rs
igamigo: Created https://github.com/0xPolygonMiden/miden-node/issues/499

PR #500, opened 2024-09-18
docs: fix typos

cratiu222:
Hello
I fixed several typos.


Mirko-von-Leipzig (approved):
Thanks, appreciated :)

The proto check will fail because the comments are part of the codegen - I'll push a fix commit soonish.

PR #502, opened 2024-09-19
feat: block producer redesign

Mirko-von-Leipzig:
This is still a WIP.

Notably, this only demonstrates the transaction pool implementation and even there skips some parts like handling notes and nullifiers.

Does not include batch producer nor block producer; as I think they should be rather straight-forward but depend on getting the pool interface correct.

We should also discuss how we want to integrate this work once its complete. The work itself is fairly non-trivial so it may be best to keep the existing design in-place until this new design has sufficient confidence. aka I'm not going to replace any existing code until we are very ready - probably split across several PRs?

This work is also missing revoking/dropping transactions e.g. due to recency conditions. This is possible to add, but since we don't actually have this feature yet I've left it out.

### Design overview & intent

The pool tracks inflight transactions, batches and blocks. The definition of inflight is extended slightly to include all data that we consider not stale. This includes some number of blocks which we know to already be complete and in the store. This is done to minimize race conditions for new transactions inputs from the store (by providing several block's of overlap/grace).

Transactions, batches and blocks are all handled similarly. They're kept in a pool and are only removed once they're considered stale. Dependency graphs are bi-directional i.e. they store both children and parents (except blocks which are always sequential so no need). Once items are removed due to staleness all references to them are removed. This means all items/graph edges should always be in the pool which we enforce using `.expect(..)` as this would indicate a rather serious bug.

Transaction and batch selection works by tracking what I called "roots". A node is considered a root if all of its ancestors have already been processed e.g. a transaction root's parents are all already part of some inflight batch, and a batch root's parents are all already part of some inflight block.

### Outstanding work

I anticipate the need for a lot of tests unless we can find some way of abstracting parts of this. Though I think in general graph book keeping is just painful so we just need to be thorough.

The rest of the state needs to be added, and the block and batch producers need at least a basic implementation so we can modify the pool api accordingly (missing outputs).

We need to pay particular attention to code that modifies the graphs; and should consider more expensive assertations, or runtime audits until we have performance concerns when running live?

Improve naming and comments which I'll get to.

We need to decide what to do in case of failure. Right now we don't really have any options except abort all transactions forming part of the failure?

Given that this is quite a lot of fairly intertwined code, I'm thinking maybe we should split this PR up more once tests are written? e.g. PR 1 includes just `add_transaction` and its tests and then we build from there.


crates/block-producer/src/tx_pool/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+use miden_objects::{
+    accounts::AccountId,
+    transaction::{ProvenTransaction, TransactionId},
+    Digest,
+};
+
+use crate::store::TransactionInputs;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BatchId(u64);
```
bobbinth, line 18: For batch ID we can use a hash of all transaction IDs that go into the batch (using some stable order). This way, two batches with the same set of transactions would end up having the same ID.

crates/block-producer/src/tx_pool/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+    pub fn increment(mut self) {
+        self.0 += 1;
+    }
+}
+
+pub struct TransactionPool {
+    /// The latest inflight state of each account.
+    ///
+    /// Accounts without inflight transactions are not stored.
+    account_state: BTreeMap<AccountId, (Digest, TransactionId)>,
```
bobbinth, line 51: Should the value be a triplet `(Digest, TransactionId, Option<BlockNumber>)`? The block number would be used to track the block at the new state was inserted into the store. This way, we'd be able to easily remove all states for blocks which are too far in the past.

Also, it may make sense to create a newtype for this to expose an interface that is tailored for our needs.

crates/block-producer/src/tx_pool/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+
+    /// All transactions currently inflight.
+    ///
+    /// This includes those not yet processed, those in batches and blocks after the stale block.
+    tx_pool: BTreeMap<TransactionId, InflightTransaction>,
+
+    /// Transactions ready to be included in a batch.
+    ///
+    /// aka transactions whose parent transactions are already included in batches.
+    tx_roots: BTreeSet<TransactionId>,
```
bobbinth, line 70: I would maybe encapsulate these in a separate struct - something like `InFlightTransactions`.

crates/block-producer/src/tx_pool/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+
+    /// The next batches ID.
+    next_batch_id: BatchId,
+
+    batch_pool: BTreeMap<BatchId, InflightBatch>,
+
+    /// Batches which are ready to be included in a block.
+    ///
+    /// aka batches who's parent batches are already included in blocks.
+    batch_roots: BTreeSet<BatchId>,
```
bobbinth, line 80: Similar to the previous comment: maybe this deserves its own struct.

crates/block-producer/src/tx_pool/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+                current: inputs.account_hash.unwrap_or_default(),
+                expected: account_update.init_state_hash(),
+            });
+        }
+
+        // Transaction is valid, update inflight state.
+        // TODO: update notes and nullifiers.
+        let tx_id = transaction.id();
+        self.account_state
+            .insert(transaction.account_id(), (account_update.final_state_hash(), tx_id));
```
bobbinth, line 132: If `account_states` was a bespoke struct, we could move this logic there as something like:
```rust
impl AccountStates {
    pub fn add_update(
        &mut self,
        update: &AccountUpdate,
        tx_id: TransactionId,
        store_state: Option<Digest>,
    ) -> Result<Vec<TransactionId>, AddTransactionError> {

    }
}
```

crates/block-producer/src/tx_pool/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+
+        Ok(())
+    }
+
+    /// Returns at most `count` transactions and a batch ID.
+    ///
+    /// Transactions are returned in a valid execution ordering.
+    ///
+    /// Returns `None` if no transactions are available.
+    pub fn select_batch(&mut self, count: usize) -> Option<(BatchId, Vec<Arc<ProvenTransaction>>)> {
```
bobbinth, line 151: Why do we want to provide `count` as a parameter? I would have thought that it would be up to the logic in this method to figure out how many transactions to return (and that would be based on some config parameters of the transaction pool).

Also, maybe for the return type we should have a dedicated struct? Something like `SelectedBatch` or `ProposedBatch`.

crates/block-producer/src/tx_pool/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+    }
+
+    fn remove_parent(&mut self, parent: &BatchId) {
+        self.parents.remove(parent);
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum TransactionStatus {
+    InQueue,
```
bobbinth, line 501: We may want to track how many times a transaction made it into a batch (e.g., as `InQueue(usize)`). This way, if let's say, the transaction was batched 3 times but failed to make it into a block, we'd automatically drop it.

bobbinth (commented):
Looks good! Thank you! I didn't not review all the logic, but did go through the most of the high-level structure and left some small comments inline.

The main comment is that I'd probably prefer to have dedicated structs which encapsulate specific pieces of functionality, and then use these structs in the `TransactionPool`.

crates/block-producer/src/tx_pool/mod.rs
Mirko-von-Leipzig: I was unsure where `count` should live; could be configured as a part of the pool or as a value in the batch producer. Logistically it made sense to me that it might be a constraint in the batch producer. But one could also argue it might form part of the transaction selection strategy.

crates/block-producer/src/tx_pool/mod.rs
Mirko-von-Leipzig: Currently I use the transaction ID of the stale block to check this. This has the benefit that the account ID is part of the transaction data so we only have to check the specific state instead of iterating over all of it.

crates/block-producer/src/tx_pool/mod.rs
Mirko-von-Leipzig: This would be ideal but I have concerns over how expensive that hash would be? This has to be performed inline meaning it extends the transaction pool lock.

bobbinth: I think this depends on which hash function we use:

- If we go with something like BLAKE3, computing a hash of 16 transaction IDs should take less than 1 microsecond.
- If we use our arithmetization-friendly hash (RPO) it would be around 50 microseconds (or around 25 microseconds if/when we migrate to RPX hash function).

We could probably start with BLAKE3 and switch to RPO/RPX only if needed, but I think either way it may be OK as my expectation was that batch selection will probably take on the order of 1 millisecond anyways.

Mirko-von-Leipzig: I've left this as is still. I think we may want to separate the concept of batch ID and batch job ID for the situation where we drop a batch and re-create the exact same one. This would result in two "inflight" batches with the same ID.

Maybe this is fine; will have to think about it some more. For now both count and hashed are still around.

crates/block-producer/src/tx_pool/mod.rs
Mirko-von-Leipzig: I've create graph types to encapsulate these.

Mirko-von-Leipzig:
I've refactored quite a bit; and I've added implementations for the rpc `add_transaction` and a batch producer with a worker pool.

There is still quite a bit missing. Notably tests, checking all inputs, and the block producer implementation.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+    transactions: TransactionGraph,
+
+    /// Inflight batches.
+    batches: BatchGraph,
+
+    /// The next batches ID.
+    next_batch_id: BatchJobId,
+
+    /// Blocks which are inflight or completed but not yet considered stale.
+    block_pool: BTreeMap<BlockNumber, Vec<BatchJobId>>,
```
bobbinth, line 77: This is for keeping track of recent blocks, right? If so, I'm wondering if we actually need this. Specifically, there are 3 things from recent blocks that we need to keep track of:

- Account states - basically `(AccountId, Digest)` tuples.
- Newly created notes - `NoteId`s, maybe with inclusion paths.
- Newly created nullifiers - these would be just `Nullifier` structs stored in a set or something like that.

As you work though adding note/nullifier tracking to the mempool, it might make sense to store them together and get rid of this `block_pool` field.

crates/block-producer/src/mempool/batch_graph.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+#[derive(Clone, Debug)]
+struct Node {
+    status: Status,
+    transactions: Vec<TransactionId>,
+    parents: BTreeSet<BatchJobId>,
+    children: BTreeSet<BatchJobId>,
+}
```
bobbinth, line 172: I will need to have 2 different variants of batches: `SelectedBatch` and `ProvenBatch`. The main difference between the two is that `ProvenBatch` would also eventually have a proof of correct construction attached to it (and there may be other differences as well). But I think we should make provisions for them now (even if the underlying structs are identical for now).

Maybe we can track the difference through the `Status` field - but we could have a proper enum too. Maybe something like:
```rust
struct Node {
    data: Batch,
    parents: BTreeSet<BatchJobId>,
    children: BTreeSet<BatchJobId>,
}

enum Batch {
    InFlight(SelectedBatch),
    Proven(ProventBatch),
    InBlock(ProvenBatch),
}
```

crates/block-producer/src/mempool/batch_graph.rs
```diff
@@ -0,0 +1,179 @@
...
+            removed.push((batch_id, batch.transactions));
+        }
+
+        removed
+    }
+
+    /// Removes a set of batches from the graph without removing any descendents.
+    ///
+    /// This is intended to cull completed batches from stale blocks.
+    pub fn remove_stale(&mut self, batches: Vec<BatchJobId>) -> Vec<TransactionId> {
```
bobbinth, line 103: Similar comment as the one for transaction graph: maybe this could be `remove_committed()`.

bobbinth (commented):
Looks good! Thank you! This is not a full review, but I did leave some comments inline - mostly minor though. I think the biggest remaining thing is to incorporate note/nullifier tracking as this may affect the structure a bit.

crates/block-producer/src/tx_pool/mod.rs
bobbinth: > I think we may want to separate the concept of batch ID and batch job ID for the situation where we drop a batch and re-create the exact same one. This would result in two "inflight" batches with the same ID.

As far as I can tell, this shouldn't happen because we can't put the same transaction into two different batches at the same time (or at least we shouldn't). So, we could have two batches with the same ID - but it shouldn't be at the same time.

My preference is still to use a hash-based ID as I think it is conceptually a bit simpler and could help us catch some errors (i.e., if we do detect 2 batches with the same ID in flight, something probably went wrong).

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    /// Returns at most `count` transactions and a batch ID.
+    ///
+    /// Transactions are returned in a valid execution ordering.
+    ///
+    /// Returns `None` if no transactions are available.
+    pub fn select_batch(
+        &mut self,
+        count: usize,
+    ) -> Option<(BatchJobId, Vec<Arc<ProvenTransaction>>)> {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I think you can argue for both. The batch producer knows best about the performance capacity of its workers for example. Or how much load they're under; or whether some have gone offline or are unstable.

We can also consider `count` an upper bound or hint; mempool can still figure out an optimal number instead.

Given that we have none of the above requirements yet; I'll move it to the mempool. I am a bit concerned that its doing pretty much everything though 😅

crates/block-producer/src/mempool/transaction_graph.rs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        // This could be optimised by inlining this inside the parent loop. This would prevent the
+        // double iteration over parents, at the cost of some code duplication.
+        self.try_make_root(id);
+    }
+
+    pub fn pop_for_batching(
+        &mut self,
+    ) -> Option<(Arc<ProvenTransaction>, BTreeSet<TransactionId>)> {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Yes this is just a trivial implementation for now. Whatever we decide to go with in the future will change this and probably add some additional metadata to the transactions/batches.

crates/block-producer/src/mempool/transaction_graph.rs
```diff
@@ -0,0 +1,123 @@
...
+        // invalidated.
+        let mut potential_roots = transactions;
+        potential_roots.extend(&self.roots);
+        self.roots.clear();
+        for tx in potential_roots {
+            self.try_make_root(tx);
+        }
+    }
+
+    pub fn remove_stale(&mut self, transactions: Vec<TransactionId>) {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: The transaction data could be removed as soon as the block is finalized. This does mean we need to make changes when new transactions enter as the "parent" transaction might no longer exist.

This is fine though it does remove an opportunity for verifying the graphs correctness at runtime since we can no longer distinguish between a broken graph and parent data just being removed. The benefit is trimming the memory usage earlier.

For this initial implementation my thought process was to ensure that the graph was always correct i.e. all indices/edges always fully exist until the entire block is removed once its stale.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: These track all inflight blocks as well as recently applied blocks. The only reason we need the recent blocks is to efficiently trim the current state once we consider them beyond the "stale" threshold of incoming transactions. Otherwise the in-memory state just grows indefinitely.

This would also be used if we have multiple inflight blocks and one fails. So I'm not sure if we do ever want to remove this? Unless we enforce that there is at most exactly one block inflight.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+        self.transactions.remove_stale(stale_transations);
+    }
+
+    pub fn block_failed(&mut self, block: BlockNumber) {
+        // TBD.. not quite sure what to do here yet. Presumably the caller has already retried this
+        // block so the block is just inherently broken.
+        //
+        // Given lack of information at this stage we should probably just abort the node?
+        // In the future we might improve the situation with more fine-grained failure reasons.
+    }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Right now I don't we can actually recover from this though. Current options as I see them:

1. Panic
2. Requeue the transactions and hope for a different set of batches (we could introduce randomness to help here)
3. Requeue batches 
4. Drop all transactions from the failed block

crates/block-producer/src/mempool/mod.rs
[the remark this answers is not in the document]
bobbinth: > I think you can argue for both. The batch producer knows best about the performance capacity of its workers for example. Or how much load they're under; or whether some have gone offline or are unstable.
> 
> We can also consider `count` an upper bound or hint; mempool can still figure out an optimal number instead.

Agreed - but I think when we get to this level, the info will need to be a bit more detailed (e.g., try to give me a batch which is heavy on consumed notes, but light on created notes). And it is probably better to define the hint parameter once we actually know what we are going to use.

crates/block-producer/src/mempool/mod.rs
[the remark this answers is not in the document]
bobbinth: I think in the short term, option 4 seems like a good one. This would allow dropping the "poisoned" transaction which makes the block production fail.

In the longer term, I think option 2 would be better and here we can keep a count of how many times a transaction has been re-queued. If it was re-queued more times than some limit (e.g., 3), we would drop it then. The idea is to reduce the number of transactions that we actually drop as presumably not all transactions in the failed block were bad.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+    batches: BatchGraph,
+
+    /// The next batches ID.
+    next_batch_id: BatchJobId,
+
+    /// Blocks which are inflight or completed but not yet considered stale.
+    block_pool: BTreeMap<BlockNumber, Vec<BatchJobId>>,
+
+    /// The current block height of the chain.
+    completed_blocks: BlockNumber,
```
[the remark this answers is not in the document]
bobbinth: > For example, we might call `select_block` to start proving the next block even before submitting the current block to the store. This would result in effectively two inflight blocks from the perspective of the mempool.

I would probably disallow this for now as we can't do this anyway (i.e., starting construction of the next block requires that the previous block has already been computed). I think the assumption we should make is that we don't start construction of the next block until the currently constructed block is committed.

crates/block-producer/src/mempool/transaction_graph.rs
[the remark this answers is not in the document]
bobbinth: > The transaction data could be removed as soon as the block is finalized. This does mean we need to make changes when new transactions enter as the "parent" transaction might no longer exist.

Makes sense - though I was thinking we'd make this check before adding the transaction to the graph (i.e., we call `add_transaction()` only with a valid set of parents).

crates/block-producer/src/mempool/mod.rs
bobbinth: > This would also be used if we have multiple inflight blocks and one fails. So I'm not sure if we do ever want to remove this? Unless we enforce that there is at most exactly one block inflight.

As mentioned in another comment, I would not allow multiple in-flight blocks for now.

> The only reason we need the recent blocks is to efficiently trim the current state once we consider them beyond the "stale" threshold of incoming transactions.

I think we should be able to do the same without tracking the actual blocks. For example, as soon as a block is committed, we'd remove all relevant data from transaction and batch graph (this shouldn't require examining the entire state). Then, we'd update the "in-flight" data structure to add block number of the recently committed block to relevant pieces of data (account updates, nullifiers, notes). At this time, we'd also remove the stale data associated with some old block.

The above really depends on how we define the structure for the in-flight data though.

crates/block-producer/src/mempool/mod.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: > I would probably disallow this for now as we can't do this anyway (i.e., starting construction of the next block requires that the previous block has already been computed).

I do mean after the computation has completed, but before the block has been sent to the store. This would elide the store time.

1. select block `N` from mempool
2. compute & prove block `N`
3. select block `N+1` and begin proving
4. submit block `N` to store
5. notify mempool `N` is complete once store completes

But I'll keep it simpler for now; we must just be aware of this in the block producer then.

crates/block-producer/src/mempool/transaction_graph.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: I'm not sure if we're talking past each other here; so to clarify:

Inflight state contains the state of all in-progress transactions as well as some number of recently committed blocks. The latter is required so that new transactions have some grace/overlap period to pre-fetch input data from the store.

State also contains it's source transaction ID so that new transactions can be added as children to form the dependency graph.

State from stale blocks must be pruned otherwise the in-memory state will grow indefinitely and essentially be a copy of all the state ever. So if block `N` becomes "stale" (the wording here is bad since its referring to the transaction inputs being considered stale but... 🤷), all state still attached to block `N` should be removed.

Transactions in committed blocks can be deleted immediately, however this leaves the state transaction ID dangling. Here are the options to handle this:

1. Only delete transaction data once the block is stale (current approach). New transactions can depend on transactions already in committed blocks which is a bit of a waste ito compute/lookups, but shouldn't be much. Highest memory usage since txn data hangs around the longest. But graph edges remain consistent - all edges will always be valid.
2. Delete data as soon as it is committed to store. This makes invalid tx parent edges possible everywhere, meaning we cannot use `tx_graph.get(parent).expect(..)` to assert graph conditions at runtime. Instead we must allow these to fail as allowed.
3. Delete data as soon as it is committed to store. Remove edges everywhere, except state. Only the state edges are now allowed to fail. The reason we do the latter is because the state data must remain active until the stale block is removed.
    1. We can amend this by having the state contain an optional txn ID which can be removed once the block is committed. However this means each block will touch the state twice. Once when it is committed, and again when it is stale.

Maybe as an analogy, what we would be doing is changing from multiple `Rc<Transaction>` to a single `Rc` with parent and state pointers becoming `rc::Weak` pointers. Which is fine; but we do lose the ability to sanity check at runtime with possible memory leaks, and edges that should exist but no longer do.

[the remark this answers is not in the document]
Mirko-von-Leipzig: > (i.e., we call add_transaction() only with a valid set of parents).

I don't quite follow this point

crates/block-producer/src/tx_pool/mod.rs
Mirko-von-Leipzig: An example:

Two inflight batches `N` and `N+1` where `N+1` depends on `N`. 

Batch `N` fails to prove and we inform the mempool of this. The mempool requeues all transactions from `N` and `N+1`.

Select two new batches and we get `N` and `N+1` again. We now have three batches inflight, with a duplicate `N+1`.

Note that the batch producer has no idea of any dependencies so it cannot automatically drop the original `N+1`.

crates/block-producer/src/mempool/mod.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Makes sense. Ideally for (2) we would have more fine grained failure conditions which might pinpoint the problematic transaction(s).

crates/block-producer/src/mempool/mod.rs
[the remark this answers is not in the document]
bobbinth: Yeah - for now let's keep it simple:

1. Select block `N` from mempool.
2. Compute and prove block `N`.
3. Submit block `N` to the store.
4. Notify mempool `N` is complete once store completes.
5. Select block `N+1` from mempool
6. etc.

crates/block-producer/src/mempool/transaction_graph.rs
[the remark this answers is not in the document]
bobbinth: I was thinking of something along the lines of the 3rd option: once a block is committed we prune all transactions of this block from `tx_graph` and all batches of this block from `batch_graph`. We do keep the information about the block's created notes/nullifiers and updated accounts separately (i think this is what you refer to as `state`). These are removed at a later point in time when the block becomes "stale".

But I haven't thought about it as much as you, so - may be missing some details about this approach.

> > (i.e., we call add_transaction() only with a valid set of parents).
> 
> I don't quite follow this point

The idea here was that when we call `tx_graph.add_transaction()` the parents will contain only transactions which are currently in `tx_graph`.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: Resolving this thread in favor of this related one [here](https://github.com/0xPolygonMiden/miden-node/pull/502#discussion_r1783830150) to keep discussing the state tracking options.

crates/block-producer/src/mempool/transaction_graph.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: > The idea here was that when we call tx_graph.add_transaction() the parents will contain only transactions which are currently in tx_graph.

Got it; yes we can do that.

> I was thinking of something along the lines of the 3rd option: once a block is committed we prune all transactions of this block from tx_graph and all batches of this block from batch_graph. We do keep the information about the block's created notes/nullifiers and updated accounts separately (i think this is what you refer to as state). These are removed at a later point in time when the block becomes "stale".
>
> But I haven't thought about it as much as you, so - may be missing some details about this approach.

I think superficially I considered the transaction-only approach as simpler, as it doesn't have to also handle blocks. 

As a mockup:

```rust
/// Represents state tracking for a single account using tx only.
struct AccountState {
    /// Transactions and states from all inflight txns and non-stale blocks.
    /// Required to allow reverting transactions.
    states: VecDequeu<(Digest, TransactionId)>,
}

impl AccountState {
    /// Rewinds the state of all the given transactions i.e. pop_back.
    /// Could also simplify and just make it a # of txns to revert which means
    /// less checking required.
    fn revert(&mut self, txs: BTreeSet<TransactionId>) { ... }

    /// Removes the states from the from the front of the queue i.e. pop_front.
    /// Could also simplify to # here.
    fn remove_stale(&mut self, txs: BTreeSet<TransactionId>) { ... }

    /// The current state of the account.
    fn current(&self) -> Option<&(Digest, TransactionId)> { self.states.last() }

    fn insert(&mut self, state: Digest, tx: TransactionId) { self.states.push_back((state, tx)); }
}
```

Versus with block support:

```rust
struct AccountState {
    inflight: VecDequeu<(Digest, TransactionId)>,
    /// Latest complete but not-stale block.
    block: Option<Digest>,
}
```

impl AccountState {
    fn revert(&mut self, txs: BTreeSet<TransactionId>) { ... }

    fn remove_stale(&mut self, state: Digest) { 
        // Remove self.block if its state matches the input.
    }

    fn current(&self) -> Option<&(Digest, Option<TransactionId>)> { 
        self.states.last()
            .map(..)                 // map tx -> Some
            .or(
                 self.block.map(...) // add tx -> None
            )
    }

    /// Using count prevents safety checks but does make it easier to implement.
    fn block_complete(&mut self, tx_count: usize) {
         self.block = self.states
             .drain(0..tx_count)
             .last();
    }
}
```

I was expecting the block handling to be more complex tbh. The nice thing about the latter is that edges are always valid. I'll try it out and see how it goes.

File: crates/block-producer/src/mempool/inflight_state.rs [hunk 16]
```diff
@@ -0,0 +1,314 @@
...
+
+    /// The latest committed state.
+    ///
+    /// Only valid if the committed count is greater than zero.
+    committed_state: Digest,
+
+    /// The number of committed transactions.
+    ///
+    /// If this is zero then the committed state is meaningless.
+    committed_count: usize,
```
Mirko-von-Leipzig, line 202: Alternatively we can just move this back into a `VecDequeue`.

crates/block-producer/src/mempool/inflight_state.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// Nullifiers emitted by inflight transactions and recently committed blocks.
+    nullifiers: BTreeSet<Nullifier>,
+}
+
+/// Describes the impact that a set of transactions had on the state.
+///
+/// TODO: this is a terrible name.
+pub struct StateDiff {
+    /// The number of transactions that affected each account.
+    account_transactions: BTreeMap<AccountId, usize>,
```
Mirko-von-Leipzig, line 36: We could do additional checks here if we stored the actual state transitions.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        Ok(self.chain_tip.0)
+    }
+
+    /// Returns a set of transactions for the next batch.
+    ///
+    /// Transactions are returned in a valid execution ordering.
+    ///
+    /// Returns `None` if no transactions are available.
+    pub fn select_batch(&mut self) -> Option<(BatchJobId, Vec<TransactionId>)> {
```
Mirko-von-Leipzig, line 149: This intentionally returns IDs for now.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /// Select batches for the next block.
+    ///
+    /// May return an empty set if no batches are ready.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there is already a block in flight.
+    pub fn select_block(&mut self) -> (BlockNumber, BTreeSet<BatchJobId>) {
```
Mirko-von-Leipzig, line 208: Intentionally returns IDs (for now).

Mirko-von-Leipzig:
As discussed offline, this work will be done on feature branch `next-block-producer` until we can finalize the entire body of work.

I'll summarize what this PR contains, and what it purposefully punts for future PRs. Some of this future work results in less-than-stellar current naming/structuring as its going to change soon. I'll attempt to point these sore spots out in a self-review so we don't focus on them too hard.

### Notable changes from last review

- State tracking moved to its own module/types.
- Added "state diffs" as a way to commit/revert batches of transactions, instead of keeping transactions around until they're pruned.
- Block and batch builder examples.
- Independent batches in a block.
- Nullifier support.

### Included in this PR

- [x] dependency tracking at transaction, batch and block levels
- [x] basic batch selection
- [x] basic block selection (ensures batches within a block are independent)
- [x] tracking and reverting inflight state (account state & nullifiers)

### Excluded

- [ ] input and output note state tracking
- [ ] data propagation beyond the mempool
- [ ] batching, a simple worker pool is demonstrated on dummy data
- [ ] block production, a simple worker loop is demonstrated on dummy data
- [ ] tests
- [ ] replacing existing framework
- [ ] reverting individual transactions (should be simple)

### What's next

Adding stronger typing to transactions, notes and transaction inputs will inform the structure of things; and I believe note tracking will fall out of this naturally.

Following that, implementing the types for batches and blocks makes sense I think.

I also want to raise `info!` and `panic!` out of the `mempool`, and instead have those handled in the callers. This won't be trivially possible everywhere but I think it makes more sense that way.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+impl BatchJobId {
+    pub fn increment(mut self) {
+        self.0 += 1;
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BlockNumber(u32);
```
Mirko-von-Leipzig, line 44: This probably belongs in `base`? Something else I've been itching to do is new-type all of the different fields in the block header for example. Having all of them be `Digest`/`u32` isn't great.

bobbinth: Yes, I think moving this to `miden-base` would be a good idea. Would you like to make a PR there?

For the other fields - also agreed, but maybe we do these incrementally over time?

Mirko-von-Leipzig: Will do. And incrementally is good; no rush :) One can also over do it.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -143,3 +145,33 @@ where
...
+
+            match result {
+                Ok(_) => mempool.block_committed(block_number),
+                Err(_) => mempool.block_failed(block_number),
+            }
+        }
+    }
+
+    async fn build_and_commit_block(&self, batches: BTreeSet<BatchJobId>) -> Result<(), ()> {
+        todo!("Aggregate, prove and commit block");
```
bobbinth, line 175: This will contain building the block and saving it to the store, right?

For block building, since we don't actually prove anything yet, I am thinking we could put in an artificial delay of 3 - 5 seconds to get the conditions close to the real thing.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -206,3 +213,84 @@ where
...
+        }
+    }
+
+    fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    fn spawn(&mut self, id: BatchJobId, transactions: Vec<TransactionId>) {
+        self.0.spawn(async move {
+            todo!("Do actual work like aggregating transaction data");
```
bobbinth, line 246: Similar to the above comment: batch building could also take 3 - 5 seconds. Let's artificially simulate this for now.

bobbinth (approved):
Looks good! Thank you! Not a very thorough review from me - but I did leave a few comments inline (mostly for the future).

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: I was wondering if we should also simulate failures. So something like sleep a random period, and then randomly fail a batch every now and then.

If we do want this, the failure rate should probably be a configurable parameter. Similar for the entire block.

Though this might annoy users if this causes a user tx to expire. And could look bad ito throughput; though I don't think that's an issue right now (?).

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: Correct. I was also considering injecting failures randomly as mentioned [here](https://github.com/0xPolygonMiden/miden-node/pull/502#discussion_r1789644365).

crates/block-producer/src/server/mod.rs
```diff
@@ -105,3 +122,114 @@ impl BlockProducer {
...
+            input_notes = %format_input_notes(tx.input_notes()),
+            output_notes = %format_output_notes(tx.output_notes()),
+            block_ref = %tx.block_ref(),
+            "Deserialized transaction"
+        );
+        debug!(target: COMPONENT, proof = ?tx.proof());
+
+        // TODO: These steps should actually change the type of the transaction.
+        // TODO: How expensive is this step? Should it be spawn_blocking?
+        TransactionVerifier::new(MIN_PROOF_SECURITY_LEVEL).verify(tx.clone())?;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I was unsure if we can always implicitly trust the RPC component.

I'll remove the verification then.

crates/block-producer/src/batch_builder/mod.rs
bobbinth: If this is configurable and not too difficult to add, I don't mind doing it.

PR #503, opened 2024-09-23
Added transaction number limit and updated tests to the limits introduced in `miden-objects`

polydez:
Resolves issue https://github.com/0xPolygonMiden/miden-base/issues/398

Here we use the latest `miden-objects` with updated/new limits. Also implemented checking for limits we can't check in `miden-objects`. And also updated tests to the latest `miden-object` and added at least one asset in note mockups.


crates/block-producer/src/errors.rs
```diff
@@ -67,9 +68,22 @@ pub enum AddTransactionError {
 /// queue can re-queue them.
 #[derive(Debug, PartialEq, Eq, Error)]
 pub enum BuildBatchError {
+    #[error(
+        "Too many input notes in the batch. Got: {0}, max: {}",
```
Mirko-von-Leipzig, line 72: Nit: I think `limit` is more precise than `max`, same below.
```suggestion
        "Too many input notes in the batch. Got: {0}, limit: {}",
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Optimized state synchronizations by removing unnecessary fetching and parsing of note details (#462).
 - [BREAKING] Changed `GetAccountDetailsResponse` field to `details` (#481).
 - Improve `--version` by adding build metadata (#495).
+- [BREAKING] Introduced additional limits for note/asset/transaction number (#503).
```
Mirko-von-Leipzig, line 10: The note and transaction limits are enforced in the batch builder, but where is the `asset` limit checked?

Mirko-von-Leipzig (approved):

bobbinth:
Not a full review, but as I mentioned in the `miden-base` PR, what we want to limit is not the number of transactions per batch/block, but rather the number of updated accounts.

A nice side effect of this should be an incentive to put transactions which touch the same account into a batch/block.

bobbinth (approved):
Looks good! Thank you!

PR #505, opened 2024-09-24
docs: fix readme removing `--path` option from installing command

SantiagoPittella:
The previous command fails with the following error:
```bash
error: the argument '--path <PATH>' cannot be used with '--git <URL>'

Usage: cargo install --locked --path <PATH> --branch <BRANCH> [CRATE[@<VER>]]...

For more information, try '--help'.
```

Removing the `--path` and using an specific crate fixed the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,13 +82,13 @@ You can also use `cargo` to compile the node from the source code if for some re
 
 ```sh
 # Install from a specific branch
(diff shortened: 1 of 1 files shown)


bobbinth (approved):
Looks good! Thank you!

Mirko-von-Leipzig (approved):

PR #506, opened 2024-09-26
Implemented `GetAccountProofs` endpoint

polydez:
Resolves https://github.com/0xPolygonMiden/miden-node/issues/485

This PR adds `GetAccountProofs` endpoint which returns latest state headers for given accounts. For public accounts it also returns account and storage headers.

After discussion we considered to lock whole database and in-memory state for the processing of the whole request.


crates/store/src/state.rs
```diff
@@ -653,25 +737,101 @@ impl State {
...
+
+            // TODO: Should we return an error if one or more accounts are not found?
+
+            let mut responses = Vec::with_capacity(requests.len());
+            for info in infos {
+                // Check if the block number in account update from the DB is not newer than the
+                // latest block in in-memory structure we used for getting proofs for accounts.
+                // Otherwise, we need to reload the data and try again.
+                // This situation might happen because we release the lock for in-memory structure
+                // before the loading from DB. We do this to avoid holding the lock for too long.
```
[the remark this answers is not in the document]
bobbinth: I think for now we can just lock the state while we fulfill this request (assuming this simplifies things, of course).

As a longer-term solution, I like the first option - i.e., have `db_select_xxx` at a given block. Performance-wise, it shouldn't be too bad because it is unlikely we'll advance by more than 1 block while fulfilling this request.

crates/store/src/state.rs
```diff
@@ -653,25 +737,101 @@ impl State {
...
+
+                    proofs.insert(*account_id, proof);
+                }
+
+                (Self::compute_latest_block_number(&inner_state), proofs)
+            };
+
+            let infos = self.db.select_accounts_by_ids(requests.keys().copied().collect()).await?;
+
+            // TODO: Should we return an error if one or more accounts are not found?
```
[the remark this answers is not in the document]
bobbinth: Yes, we can tell just from the account ID if it is public or not ([see here](https://github.com/0xPolygonMiden/miden-base/blob/next/objects/src/accounts/account_id.rs#L249)). There are two things we can do about private accounts:

1. Return only the inclusion path.
2. Return an error.

I think for now, we should go with the latter option and we can change this in future if needed. So, basically, this endpoint becomes exclusively for public accounts (and maybe we should name it accordingly? - e.g., `GetPublicAccountStates`).

Also, since we can make the check based on IDs - we should do it this in the RPC component.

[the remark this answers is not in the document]
bobbinth: Another option could be to simplify this endpoint and make it work primarily to get account inclusion paths + basic information for accounts (when available). And this would include both public and private accounts. So, for example, the response could be (other naming suggestions welcome):
```proto
message AccountStateResponse {
    merkle.MerklePath account_proof = 3;
    optional AccountStateHeader = 4;
}

message AccountStateHeader {
    account.AccountHeader header = 1;
    bytes storage_header = 2;
}
```
This would cover 90% of what we need (if not more), and for additional data (e.g., storage maps and assets) we could have a different endpoint (maybe we could just adapt our current `GetAccountDetails` endpoint for that). But this would be for another PR.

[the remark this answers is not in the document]
Mirko-von-Leipzig: Personally I like the simplify suggestion. Its nice having well defined endpoints that just work without exceptions.

I think the naming is fine.

crates/store/src/state.rs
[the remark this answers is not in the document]
polydez: Thanks guys, but I don't think both solutions give us any benefit at the moment. Current DB scheme is pretty simple and we can query all latest accounts' data by a single request, as we do now. Even if we block the whole state for serving this request, the state could be updated in microseconds after we unlock it and the client will still receive outdated state. This is common problem for blockchains that we can only ensure that state is up-to-date for the given block number, but not absolutely latest for the current blockchain. We should run calls with providing account hashes we know and node must fail if hashes mismatch, so the client will need to update state and repeat (I guess, we already going to do so). Also I don't see any benefits with locking the whole state (stop the world instead of repeat requests in rare cases) and it doesn't seem to lead to significant simplification of the code (we would remove `loop` and that long comment, refactor of accessing proofs and that's it). So I'd prefer to keep the current solution, if you don't mind.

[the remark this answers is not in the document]
bobbinth: > Even if we block the whole state for serving this request, the state could be updated in microseconds after we unlock it and the client will still receive outdated state.

This is actually not a problem. The main thing we want to avoid is the client getting an inconsistent state (i.e., account data does not verify against Merkle paths). Sending a slightly outdated state to clients is fine.

> Once we change our DB scheme to more sophisticated, it would require to make several requests to get the data which we get by single request.

In one of the other comments, I proposed simplifying this request - so, hopefully, we can keep just one DB request for this endpoint.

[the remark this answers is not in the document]
bobbinth: I think with the endpoint simplification we don't even need approach 1. We just get basic account data from the DB in a single request, get the Merkle paths for these accounts, and we are done.

We could keep the loop, but I don't really see a strong need for it (and it is something that we'll likely need to remove in the future).

Separately, we should probably impose some limit on how many accounts can sent with the request. In the transaction kernel, we limit the number of "foreign accounts" for a single transaction to 64. We can probably do the same here.

crates/rpc/src/server/api.rs
```diff
@@ -1,16 +1,18 @@
+use miden_lib::transaction::memory::MAX_NUM_FOREIGN_ACCOUNTS;
```
bobbinth, line 1: Maybe we should move this constant into `miden-base`?

proto/account.proto
```diff
@@ -20,3 +20,14 @@ message AccountInfo {
...
+
+message AccountHeader {
+    // Vault root hash.
+    digest.Digest vault_root = 1;
+    // Storage root hash.
+    digest.Digest storage_commitment = 2;
+    // Code root hash.
+    digest.Digest code_commitment = 3;
+    // Account nonce.
+    uint32 nonce = 4;
```
bobbinth, line 32: Shouldn't `nonce` be `uint64`?

proto/requests.proto
```diff
@@ -131,3 +131,8 @@ message GetAccountStateDeltaRequest {
     // Block number up to which the delta is requested (inclusive).
     fixed32 to_block_num = 3;
 }
+
+message GetAccountStatesRequest {
+    // List of account IDs to get states.
+    repeated account.AccountId account_ids = 1;
+}
```
bobbinth, line 138: Let's add something like `iinclude_headers` flag (false by default). When set to false, `AccountStateResponse.state_header` would be omitted.

crates/store/src/state.rs
```diff
@@ -703,7 +746,12 @@ impl State {
...
     /// Returns the latest block number.
     pub async fn latest_block_num(&self) -> BlockNumber {
-        (self.inner.read().await.chain_mmr.forest() + 1) as BlockNumber
+        Self::compute_latest_block_number(self.inner.read().await.deref())
+    }
+
+    /// Returns the latest block number for the given inner state.
+    fn compute_latest_block_number(inner_state: &InnerState) -> BlockNumber {
+        (inner_state.chain_mmr.forest() + 1).try_into().expect("block number overflow")
     }
```
bobbinth, line 755: Should this be a method on the `InnerState` instead? (e.g., something like `InnerState::latest_block_num()`).

bobbinth (commented):

One thing I'm still not sure of is whether `GetAccountStates` is the right name for this endpoint since we not actually getting account states from it. Maybe `GetAccountProofs` or `GetAccountHeaders` would be better?

crates/rpc/Cargo.toml
```diff
@@ -15,6 +15,7 @@ repository.workspace = true
 directories = { version = "5.0" }
 figment = { version = "0.10", features = ["toml", "env"] }
 hex = { version = "0.4" }
+miden-lib = { workspace = true}
```
bobbinth, line 18: Is this dependency still needed?

crates/store/src/state.rs
```diff
@@ -653,25 +660,72 @@ impl State {
...
+        &self,
+        account_ids: HashSet<AccountId>,
+        include_headers: bool,
+    ) -> Result<(BlockNumber, Vec<AccountStateResponse>), DatabaseError> {
+        // Lock inner state for the whole operation. We need to hold this lock to prevent the
+        // database, account tree and latest block number from changing during the operation,
+        // because changing one of them would lead to inconsistent state.
+        let inner_state = self.inner.read().await;
+
+        let infos = self.db.select_accounts_by_ids(account_ids.iter().copied().collect()).await?;
```
bobbinth, line 693: We need to go to the database only if `include_headers` is true - right? Otherwise, all the info we need should be already in memory.

bobbinth (commented):
Thank you! Looks good! I left one comment inline. Also, let's rename the endpoint to `GetAccountProofs`.

crates/store/src/state.rs
polydez: Good catch, thank you!

proto/responses.proto
```diff
@@ -173,3 +173,28 @@ message GetAccountStateDeltaResponse {
...
+    merkle.MerklePath account_proof = 3;
+    /// State header for public accounts. Filled only if `include_headers` flag is set to `true`.
+    optional AccountStateHeader state_header = 4;
+}
+
+message AccountStateHeader {
+    // Account header.
+    account.AccountHeader header = 1;
+    /// Values of all account storage slots (max 255).
+    bytes storage_header = 2;
```
bobbinth, line 199: nit: I think this should be just `//` rather than `///`.

proto/responses.proto
```diff
@@ -173,3 +173,28 @@ message GetAccountStateDeltaResponse {
...
+
+message AccountProofsResponse {
+    // Account ID.
+    account.AccountId account_id = 1;
+    // Account hash.
+    digest.Digest account_hash = 2;
+    // Authentication path from the `account_root` of the block header to the account.
+    merkle.MerklePath account_proof = 3;
+    /// State header for public accounts. Filled only if `include_headers` flag is set to `true`.
+    optional AccountStateHeader state_header = 4;
```
bobbinth, line 192: Similar nit as the one above.

proto/responses.proto
```diff
@@ -173,3 +173,28 @@ message GetAccountStateDeltaResponse {
     // The calculated `AccountStateDelta` encoded using miden native format
     optional bytes delta = 1;
 }
+
+message GetAccountProofsResponse {
+    // Block number at which the state of the account was returned.
+    fixed32 block_num = 1;
+    // List of account state infos for the requested account keys.
+    repeated AccountProofsResponse account_state_infos = 2;
```
bobbinth, line 181: Should the name of the field be `account_proofs` or something like that? (not sure if `state_infos` has much meaning in this context).

crates/rpc/src/server/api.rs
```diff
@@ -250,4 +251,29 @@ impl api_server::Api for RpcApi {
 
         self.store.clone().get_account_state_delta(request).await
     }
+
+    #[instrument(
+        target = "miden-rpc",
+        name = "rpc:get_account_states",
```
bobbinth, line 257: This should also be changed to `rpc:get_account_proofs`.

crates/store/src/server/api.rs
```diff
@@ -475,6 +477,35 @@ impl api_server::Api for StoreApi {
         Ok(Response::new(GetBlockByNumberResponse { block }))
     }
 
+    #[instrument(
+        target = "miden-store",
+        name = "store:get_account_state",
```
bobbinth, line 482: This should also be changed to `store:get_account_proofs`.

crates/store/src/state.rs
```diff
@@ -653,25 +660,83 @@ impl State {
...
     pub async fn get_account_details(&self, id: AccountId) -> Result<AccountInfo, DatabaseError> {
         self.db.select_account(id).await
     }
 
+    /// Returns account states with details for public accounts.
+    pub async fn get_account_states(
+        &self,
+        account_ids: HashSet<AccountId>,
+        include_headers: bool,
+    ) -> Result<(BlockNumber, Vec<AccountProofsResponse>), DatabaseError> {
```
bobbinth, line 687: Question: why pass `account_ids` as a `HashSet` rather than a simple `Vec`?

crates/store/src/state.rs
```diff
@@ -653,25 +660,83 @@ impl State {
...
+        // Lock inner state for the whole operation. We need to hold this lock to prevent the
+        // database, account tree and latest block number from changing during the operation,
+        // because changing one of them would lead to inconsistent state.
+        let inner_state = self.inner.read().await;
+
+        let state_headers = if !include_headers {
+            HashMap::<AccountId, AccountStateHeader>::default()
```
bobbinth, line 694: Is there a reason to use `HashMap` rather than `BTreeMap` here? I don't really have a strong preference, but we usually use `BTreeMpas` everywhere.

bobbinth (approved):
Looks good! Thank you! I left some small comments inline - once these are addressed, we can merge.

Also, our docs for various endpoints (in `README` files) are getting out of date. Let's create an issue to update them. In the same issue we can discuss if there is some automated way to generate endpoint docs for rpc/store/block producer components.

crates/store/src/state.rs
polydez: I wanted to reduce number of conversions but now I think it was better to pass vector here. Thank you!

crates/store/src/state.rs
polydez: I also don't have strong preference here. Taking that this map will be pretty small, it shouldn't be any significant difference between them, so I will use `BTreeMap` here, thanks!

PR #507, opened 2024-09-30
test: Add expiration number to proven tx builder

igamigo:
Follows https://github.com/0xPolygonMiden/miden-base/pull/897


Cargo.toml
igamigo, line 1: These are to be changed back to `next` once https://github.com/0xPolygonMiden/miden-base/pull/897 is merged

Mirko-von-Leipzig (approved):

PR #510, opened 2024-10-02
Dropped basic wallet account from genesis file

polydez:
Resolves: https://github.com/0xPolygonMiden/miden-node/issues/446

In this PR we remove support of basic wallet creation in genesis config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - [BREAKING] Changed `GetAccountDetailsResponse` field to `details` (#481).
 - Improve `--version` by adding build metadata (#495).
 - [BREAKING] Introduced additional limits for note/account number (#503). 
+- [BREAKING] Removed support of basic walled in genesis creation (#510).
 
 ## 0.5.1 (2024-09-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - [BREAKING] Changed `GetAccountDetailsResponse` field to `details` (#481).
 - Improve `--version` by adding build metadata (#495).
 - [BREAKING] Introduced additional limits for note/account number (#503). 
-- [BREAKING] Removed support of basic walled in genesis creation (#510).
+- [BREAKING] Removed support for basic wallets in genesis creation (#510).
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - [BREAKING] Changed `GetAccountDetailsResponse` field to `details` (#481).
 - Improve `--version` by adding build metadata (#495).
 - [BREAKING] Introduced additional limits for note/account number (#503). 
+- [BREAKING] Removed support of basic walled in genesis creation (#510).
```
Mirko-von-Leipzig, line 11:
```suggestion
- [BREAKING] Drop basic wallet from default genesis configuration (#510).
```

bin/node/src/commands/genesis/inputs.rs
```diff
@@ -18,18 +18,9 @@ pub struct GenesisInput {
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(tag = "type")]
 pub enum AccountInput {
-    BasicWallet(BasicWalletInputs),
     BasicFungibleFaucet(BasicFungibleFaucetInputs),
 }
```
Mirko-von-Leipzig, line 22: Should we remove the possibility entirely like this? i.e. is there zero value in having a wallet account in the genesis file?

(I think yes, but just checking)

Mirko-von-Leipzig (approved):
LGTM just the one question

bin/node/src/commands/genesis/inputs.rs
polydez: I was thinking about it and decided that for mainnet we will need to predeploy different accounts and contracts (like DAO for initial funds management), but I don't think that predeployment of simple wallets has any sense for us except testing (but for testing purposes we can deploy wallets in tests).

bobbinth:
Added @igamigo to reviewers as this will likely affect integration tests on the client.

igamigo (approved):
LGTM! I'll follow up with a small PR on the client (should be very small and quick to address so feel free to merge this anyway).

polydez:
@bobbinth should we merge this PR or you would also like to review?

PR #511, opened 2024-10-03
feat(faucet): migrate from actix to axum

polydez:
Resolves: https://github.com/0xPolygonMiden/miden-node/issues/479

### DIFF
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -124,16 +124,21 @@ impl FaucetClient {
         &mut self,
         executed_tx: ExecutedTransaction,
     ) -> Result<u32, FaucetError> {
-        let transaction_prover = LocalTransactionProver::new(ProvingOptions::default());
+        let (request, delta) = {
+            let transaction_prover = LocalTransactionProver::new(ProvingOptions::default());
 
-        let delta = executed_tx.account_delta().clone();
+            let delta = executed_tx.account_delta().clone();
 
-        let proven_transaction = transaction_prover.prove(executed_tx).map_err(|err| {
-            FaucetError::InternalServerError(format!("Failed to prove transaction: {}", err))
-        })?;
+            let proven_transaction = transaction_prover.prove(executed_tx).map_err(|err| {
+                FaucetError::InternalServerError(format!("Failed to prove transaction: {}", err))
+            })?;
 
-        let request = SubmitProvenTransactionRequest {
-            transaction: proven_transaction.to_bytes(),
+            (
+                SubmitProvenTransactionRequest {
+                    transaction: proven_transaction.to_bytes(),
+                },
+                delta,
+            )
         };
 
         let response = self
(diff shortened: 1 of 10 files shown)


bin/faucet/src/client.rs
```diff
@@ -124,16 +124,21 @@ impl FaucetClient {
         &mut self,
         executed_tx: ExecutedTransaction,
     ) -> Result<u32, FaucetError> {
-        let transaction_prover = LocalTransactionProver::new(ProvingOptions::default());
+        let (request, delta) = {
```
Mirko-von-Leipzig, line 127: Could you add a comment explaining the reasoning for the code block?

bin/faucet/src/errors.rs
```diff
@@ -21,27 +20,34 @@ pub enum FaucetError {
...
+    fn into_response(self) -> Response {
+        let status_code = self.status_code();
+        let message = match self {
+            FaucetError::StartError(msg) => msg,
+            FaucetError::BadRequest(msg) => msg,
+            FaucetError::ConfigurationError(msg) => msg,
+            FaucetError::InternalServerError(msg) => msg,
+            FaucetError::AccountCreationError(msg) => msg,
+            FaucetError::NotFound(msg) => msg,
+        };
```
Mirko-von-Leipzig, line 48: While we're here - are there any errors we shouldn't be exposing to the caller?

`StartError` and `ConfigurationError` sound like they should be impossible to return to the caller. `InternalError` is also exposing the error string which is probably a bad idea.

bin/faucet/src/main.rs
```diff
@@ -72,36 +74,42 @@ async fn main() -> Result<(), FaucetError> {
...
+                )
+                .with_state(faucet_state);
+
+            let endpoint_url = config.endpoint_url();
+
+            let listener = TcpListener::bind((config.endpoint.host, config.endpoint.port))
+                .await
+                .map_err(|err| FaucetError::StartError(err.to_string()))?;
+
+            info!("Server is now running on: {}", endpoint_url);
```
Mirko-von-Leipzig, line 103: We should prefer structured logs where possible. There are quite a few interpolated logs in the faucet which we should redo.

Might need to add a display implementation to endpoint so that it actually prints something nice like `http://localhost:8080`.

```suggestion
            info!(endpoint=%endpoint, "Server started");
```

Mirko-von-Leipzig (approved):
I'm assuming you've tested this locally :)

Code LGTM, just a couple of comments that aren't really PR related.

bin/faucet/src/errors.rs
polydez: Yes, I agree with you. I'd rather separate this enum to two different. Will do it, thanks!

bobbinth (approved):
Looks good! Thank you! Not a very deep review from me - but once @Mirko-von-Leipzig comments are addressed, we can merge.

Issue #514, opened 2024-10-07
[block-producer]: redesign tracking issue


Mirko-von-Leipzig:
The block-producer redesign feature is being developed against the `next-block-producer` branch until it is ready to merge into `next` proper.

This issue tracks the remaining work after #502, initially derived from [this comment](https://github.com/0xPolygonMiden/miden-node/pull/502#issuecomment-2393925630).

- [x] input and output note state tracking
- [x] data propagation from mempool to batch and block builders
- [x] batch builder with [configurable failure/work rate](https://github.com/0xPolygonMiden/miden-node/pull/502#discussion_r1789644365)
- [x] block builder with configurable failure/work rate
- [x] tests
- [x] replacing existing framework
- [x] #508 
- [x] #519 
- [x] #531

bobbinth:
@Mirko-von-Leipzig - should we close this issue?

Issue #508, opened 2024-09-30
Support transaction expiration


Mirko-von-Leipzig:
Transactions now have an expiration block. The node should

- reject incoming transactions which are already expired, and
- drop transactions from the mempool once they expire

This is blocked by #491 being implemented in PR #502.

bobbinth:
Closed by #582.

PR #515, opened 2024-10-08
feat(block-producer): batch builder

Mirko-von-Leipzig:
This PR is intended as a review of the in-progress `block-producer` redesign. The initial body of work was done on the `next-block-producer` work, however that initial PR received little scrutiny and thus it makes more sense (to me) to review this one in the context of `next` proper.

In that context, this PR adds the following pieces:

- A `mempool` consisting of
  - `InflightState` to track state
  - `TransactionGraph` to form a dependency graph between transactions
  - `BatchGraph` to form a dependency graph of inflight batches
- A simple block builder implementation using this new `mempool`
- A simple batch builder implementation using this new `mempool`

The builders reuse the existing `DefaultXXX` types to minimize the footprint of this PR. Once we've switched over we could refactor these further as needed.

### Outstanding work

The work items of #514:

- [x] input and output note state tracking
- [x] data propagation from mempool to batch and block builders
- [x] batch builder with [configurable failure/work rate](https://github.com/0xPolygonMiden/miden-node/pull/502#discussion_r1789644365)
- [x] block builder with configurable failure/work rate
- [ ] extensive tests/docs
  - [x] `InflightState`
  - [x] `InflightAccountState`
  - [x] `TransactionGraph`
  - [ ] `BatchGraph`
  - [ ] `Mempool`
- [ ] replacing existing framework
- [ ] #508 

I'll continue expanding test coverage; though I'd appreciate a review already. In particular, anything that is unclear or looks _bad_ in the components I've marked as tested.

I'm also uncertain about the amount of `# Panics` involved. I could return more errors and raise the panics up, but there are complications where certain actions just can't (trivially) be made atomic and leave the graphs/state in a broken state. And a broken `mempool` means that we're dead in the water anyway. Thoughts welcome.


crates/block-producer/Cargo.toml
```diff
@@ -18,6 +18,7 @@ tracing-forest = ["miden-node-utils/tracing-forest"]
 async-trait = { version = "0.1" }
 figment = { version = "0.10", features = ["toml", "env"] }
 itertools = { version = "0.13" }
+miden-air = { workspace = true }
```
Mirko-von-Leipzig, line 21: Required to access transaction proof type for `VerifiedTransaction`.

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -34,6 +56,40 @@ pub struct TransactionBatch {
...
+pub struct ProvenBatch {
+    id: BatchId,
+    updated_accounts: BTreeMap<AccountId, AccountUpdate>,
+    input_notes: InputNotes,
+
+    /// Tracks the nullifiers consumed by this batch.
+    ///
+    /// This is tracked separately from input notes to allow for ephemeral notes
+    /// as these will be pruned from both input and output notes.
+    nullifiers: BTreeSet<Nullifier>,
```
Mirko-von-Leipzig, line 72: I'm unsure if this is required or not?

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -223,16 +284,144 @@ impl TransactionBatch {
...
+        self.update_account(&tx)?;
+        self.merge_input_notes(&tx)?;
+        self.merge_output_notes(&tx)?;
+
+        Ok(())
+    }
+
+    pub fn build(mut self) -> Result<ProvenBatch, BuildBatchErrorRework> {
+        // Remove ephemeral notes prior to asserting batch constraints.
+        let ephemeral = self.remove_ephemeral_notes();
```
Mirko-von-Leipzig, line 309: I was thinking of logging the IDs here, but maybe that's not really important?

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -223,16 +284,144 @@ impl TransactionBatch {
...
+        }
+
+        Ok(())
+    }
+
+    /// Removes all ephemeral notes within the batch.
+    ///
+    /// These are notes which are both produced and consumed within this batch.
+    ///
+    /// Their nullifiers are retained.
```
Mirko-von-Leipzig, line 387: What happens with ephemeral nullifiers? Do they also vanish?

crates/block-producer/src/errors.rs
```diff
@@ -163,6 +165,45 @@ impl BuildBatchError {
...
+
+impl BuildBatchErrorRework {
+    /// Returns a [Self::LimitExceeded] error if `actual` exceeds the `limit`.
+    pub fn check_limit(kind: &'static str, actual: usize, limit: usize) -> Result<(), Self> {
+        if actual > limit {
+            Err(Self::LimitExceeded { kind, actual, limit })
+        } else {
+            Ok(())
+        }
+    }
```
Mirko-von-Leipzig, line 204: I initially liked this idea, but now I'm less sure.

crates/block-producer/src/transaction.rs
```diff
@@ -0,0 +1,211 @@
...
+            if self.witnessed.insert(id, note).is_some() || self.unauthenticated.contains_key(&id) {
+                return Err(InputNotesError::DuplicateUnauthenticatedNote(id));
+            }
+        }
+
+        for note in other.proven {
+            if !self.proven.insert(note.clone()) {
+                return Err(InputNotesError::DuplicateProvenNote(note.0));
+            }
+        }
```
Mirko-von-Leipzig, line 135: Should this instead track and compare nullifiers instead? That might make more sense looking at it again.

So an additional `BTreeSet<Nullifier>` property.

Mirko-von-Leipzig: I've decided to do it this way instead.

crates/block-producer/src/errors.rs
Mirko-von-Leipzig: Removed in favor of more explicit fashion.

crates/block-producer/src/batch_builder/batch.rs
Mirko-von-Leipzig: This is now tracked in `InputNotes`

Mirko-von-Leipzig:
I think with hindsight I could/should have split this work into at least two separate ideas.

1. Make input note / transaction type changes either in `miden-base` or as new types here. Introduce them to the existing block-producer. This could be merged into `next` without much fear.
2. Create `mempool`.

We could still attempt doing it the above way but it will probably just take longer.

crates/block-producer/src/server/mod.rs
```diff
@@ -193,30 +198,34 @@ impl Server {
...
-                .nullifiers
-                .remove(nullifier)
-                .ok_or(AddTransactionErrorRework::AuthenticatedNoteNotFound(*nullifier))?;
-
-            if nullifier_state.is_some() {
-                return Err(AddTransactionErrorRework::NotesAlreadyConsumed([*nullifier].into()));
+        // Upgrade unauthenticated notes for which we now have witnesses.
+        //
+        // This prevents having to re-witness them later, saving on database IO.
+        let auth = self.store.get_note_authentication_info(unauthenticated_notes.iter()).await?;
```
Mirko-von-Leipzig, line 225: This should get shuffled into the `store.get_tx_inputs` method.

Mirko-von-Leipzig:
One issue that @polydez has made me aware of is that notes __must__ be ordered by transaction order. See [here](https://github.com/0xPolygonMiden/miden-node/pull/396#discussion_r1663271312).

I have some refactoring to do to support this properly.

bobbinth:
> One issue that @polydez has made me aware of is that notes **must** be ordered by transaction order. See [here](https://github.com/0xPolygonMiden/miden-node/pull/396#discussion_r1663271312).

Yes, this is how I'm currently thinking about it. But one thing we should do to confirm this is sketch out batch kernel - or at least what we think it'll take as public inputs and what public outputs it'll produce. For example:

- Public inputs could be a commitment to a list of transaction IDs (+ some additional data needed to verify transaction proofs) for the transactions included in the batch. All detailed transaction could be provided via the advice provider.
- Public outputs could be a commitment to a list of updated account IDs, a root of the batch note tree, and a commitment to list of created nullifiers.

crates/block-producer/src/mempool/inflight_state.rs
Mirko-von-Leipzig, line 1: I'm hoping I can avoid mocking `ProvenTransaction` in these tests, and instead using and extending the `crate::test_utils::Random`. It should be possible since we only care about `NoteId`, `Nullifier`, `TransactionId`, `AccountId` and `Digest` (account state).

I'll see if I can figure out something.

Mirko-von-Leipzig:
If anyone can think of a design so that we only have a single graph type; that would be great. They're always a pain to test and maintain so having just single variant would be much better.

crates/block-producer/src/mempool/batch_graph.rs
Mirko-von-Leipzig, line 1: If anyone can think of a design so that we only have a single graph type; that would be great. They're always a pain to test and maintain so having just single variant would be much better.

Currently we have `TransactionGraph` and `BatchGraph`. The former tracks transaction dependencies and each node/tx transitions from `InQueue -> Processed` once it is placed in a batch. The latter tracks batch dependencies and each node/batch transitions from `InFlight -> Proven -> InBlock`.

What I was hoping to find was a single graph type which has no different states, and could be generic over the data type it holds. i.e. the graph would just be vertices containing data `T` and parent/child edges.

Initially this might seem plausible; just have a graph for transactions `InQueue`, then as they are removed for batch proving move them to another graph, then another for `InQueue` batches, and then finally they're part of the in-progress block. This works fine for the happy/forward path; but I'm struggling with the failure cases i.e. how to re-insert transactions all the way backwards if a batch/block fails to prove. This must be possible; however tracking the parent/child relationships across all these graphs is.. painful I think especially since we need to prune these edges once a block is committed.

Still this might be worth pursuing harder; graphs remain a pain and creating a single simple and robust implementation would likely pay off.

Mirko-von-Leipzig: An alternate approach could be using a tri-state graph and adapting the transaction graph to it. e.g. a graph with `InQueue -> Processed -> Consumed` states already matches the batch states. Transactions would then be inserted as already processed i.e. you have two insert methods, `insert_inqueue`, `insert_processed`.

I'd prefer a two state solution since that's even simpler, but the above should work at least.

Mirko-von-Leipzig: We could also conceivably have a single state graph, but then txs/graphs must be removed from the graph as they're "consumed" which makes tracking added/removed dependencies after the fact much more difficult.

Mirko-von-Leipzig: I believe we can using the two-state graphs if we define them as holding dependent tasks to perform. This matches the transaction graph, and matches batches waiting to be processed into a block.

The odd duck out is batches which are inflight, being proven. This should be simple to model as a set of jobs waiting to return. And a completed job is only moved into the proper graph if its parents were already moved.

```rust
struct Mempool {
    state: InflightState,
    inflight_txs: Graph<TransactionId, ProvenTransaction>,
    batches_being_proven: BTreeMap<JobId, BTreeSet<JobId>>,
    proven_batches: Graph<BatchId, TransactionBatch>,    
}

File: crates/block-producer/src/mempool/batch_graph.rs [hunk 8]
```diff
@@ -0,0 +1,209 @@
...
+    parents: BTreeSet<BatchJobId>,
+    children: BTreeSet<BatchJobId>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum Status {
+    InFlight,
+    Proven(TransactionBatch),
+    InBlock,
+}
```
bobbinth, line 209: Would be good to add comments explaining what these enum variants mean. Also, to keep them somewhat parallel with transaction status, maybe the names could be `Queued`, and `Proven`, and `Processed`?

crates/block-producer/src/mempool/batch_graph.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /// Mark a batch as proven if it exists.
+    pub fn mark_proven(&mut self, id: BatchJobId, batch: TransactionBatch) {
+        // Its possible for inflight batches to have been removed as part of another batches
+        // failure.
+        if let Some(node) = self.nodes.get_mut(&id) {
+            node.status = Status::Proven(batch);
+            self.try_make_root(id);
+        }
```
bobbinth, line 142: Should we add an `assert!` here to make sure that status is `InFlight`?

crates/block-producer/src/mempool/batch_graph.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /// Returns at most `count` __indepedent__ batches which are ready for inclusion in a block.
+    pub fn select_block(&mut self, count: usize) -> BTreeMap<BatchJobId, TransactionBatch> {
```
bobbinth, line 146: Would be good to clarify what "independent" means here? I'm assuming that it means no batch depends on another. If so, why do we need this restriction? (i.e., why not return batches with internal dependencies?)

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -143,3 +148,31 @@ where
...
+struct BlockProducer<BB> {
+    pub mempool: Arc<Mutex<Mempool>>,
+    pub block_interval: tokio::time::Duration,
+    pub block_builder: BB,
+}
+
+impl<BB: BlockBuilder> BlockProducer<BB> {
+    pub async fn run(self) {
+        let mut interval = tokio::time::interval(self.block_interval);
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
```
bobbinth, line 161: Should the delay behavior here not be `Burst`? I'm thinking, if building a block took longer than expected, we should try to build the next block as soon as possible rather than wait until the next multiple of the interval.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -143,3 +148,31 @@ where
         Ok(())
     }
 }
+
+struct BlockProducer<BB> {
+    pub mempool: Arc<Mutex<Mempool>>,
+    pub block_interval: tokio::time::Duration,
+    pub block_builder: BB,
+}
```
bobbinth, line 156: Is the plan to eventually get rid of the `BlockBuilder` trait and merge the functionality into this struct?

Also, should we add `simulated_proof_time` here too?

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -206,3 +221,106 @@ where
...
+    fn spawn(&mut self, id: BatchJobId, transactions: Vec<ProvenTransaction>) {
+        self.in_progress.spawn({
+            let simulated_proof_time = self.simulated_proof_time;
+            async move {
+                tracing::debug!("Begin proving batch.");
+
+                let batch = TransactionBatch::new(transactions, Default::default())
+                    .map_err(|err| (id, err))?;
+
+                tokio::time::sleep(simulated_proof_time).await;
```
bobbinth, line 272: Question: is this going to sleep for the exact amount of time? I'm wondering if we should add some variance to it.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -206,3 +221,106 @@ where
...
+                tracing::debug!("Batch proof completed.");
+
+                Ok((id, batch))
+            }
+        });
+    }
+}
+
+impl BatchProducer {
+    pub async fn run(self) {
```
bobbinth, line 282: Could we add a high-level description of the logic in this function?

bobbinth (commented):
Looks good! Thank you! I left some comments inline - most are minor, but there are a few where I think transaction validation logic needs to be improved.

crates/block-producer/src/mempool/batch_graph.rs
bobbinth: If we go with this approach, how would we make the selection logic different between transaction and batch graphs?

Mirko-von-Leipzig: Ideal would be specialization but of course that's but a distant dream in `rust` at the moment.

We could try a `trait Strategy` which is given enough access to fulfill its requirements; though I think its a bit of a pita since likely it will want full access to everything.

The simplest and also what I suggest is that we fake specialization using `pub(super)` and new-types:

```rust
/// Note: TransactionGraph gets full pub access to `Graph` fields.
pub TransactionGraph(Graph<TransactionId, ProvenTransaction>);

impl TransactionGraph {
    pub fn select_batch(&mut self, ...) { ... }

    // Re-expose all Graph<K, V> methods or use Deref to shortcut it.
}
```
This conveniently also lets us store additional metadata that we may want to track without it being an explicit property of the graph nodes.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    ) -> Result<u32, AddTransactionError> {
+        // Ensure inputs aren't stale.
+        if let Some(stale_block) = self.stale_block() {
+            if inputs.current_block_height <= stale_block.0 {
+                return Err(AddTransactionError::StaleInputs {
+                    input_block: BlockNumber(inputs.current_block_height),
+                    stale_limit: stale_block,
+                });
+            }
+        }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Correct; this is only possible if the staleness limit is too constrictive for some combination of block rate, transaction throughput and/or database IO.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        // Add transaction to inflight state.
+        let parents = self.state.add_transaction(&transaction, inputs.account_hash)?;
+
+        self.transactions.insert(transaction.id(), transaction, parents);
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Correct; though if we use the newtype wrapper to side-step specialization then we could improve this api.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+    ///
+    /// May return an empty set if no batches are ready.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there is already a block in flight.
+    pub fn select_block(&mut self) -> (BlockNumber, BTreeMap<BatchJobId, TransactionBatch>) {
+        assert!(self.block_in_progress.is_none(), "Cannot have two blocks inflight.");
+
+        let batches = self.batches.select_block(self.block_batch_limit);
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: No reason aside from the code having evolved in the one implementation, but not the other. I'll align them in the final iteration.

crates/block-producer/src/mempool/batch_graph.rs
Mirko-von-Leipzig: I thought you'd specified this 😅 Maybe I misunderstood.

But yes your interpretation is correct; I meant that these batches are not coupled to each other at all.

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: Correct. And yes we should have some simulated time as well; thanks.

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: We can define some distribution function if you like :) I assume just a uniform distribution between a min and a max would be good enough?

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: It depends on how we want to handle missed blocks.

`Burst` if we believe this is a temporary overload and want to catch up the missed blocks. This will result in the most amount of non-periodic blocks, but once caught up we will be back on the correct average block time.

`Skip` if we want the timing to be perfect, but are okay with skipping a block every now and then.

`Delay` if we are okay with a single imperfect block time, and the subsequent blocks should realign to the time. `Delay` here just means that the interval should realign with the last tick. So if we had block time `2` then good blocks `[0,2,4,6]` and a slow block at `9` then the next blocks would be out at `[11, 13, 15]` instead of trying to squeeze a quicker block at `10` (`burst`).

This is less important for missing a single block, but in the case where we miss multiple blocks then things are a bit weird I think.

crates/block-producer/src/mempool/inflight_state/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+        }
+
+        // Ensure nullifiers aren't already present.
+        let double_spend = tx
+            .get_nullifiers()
+            .filter(|nullifier| self.nullifiers.contains(nullifier))
+            .collect::<Vec<_>>();
+        if !double_spend.is_empty() {
+            return Err(VerifyTxError::InputNotesAlreadyConsumed(double_spend));
+        }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I perform this check in the server handler, as a pre-condition to submitting it to the mempool.

crates/block-producer/src/mempool/inflight_state/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+        // We don't need to worry about double spending them since we already checked for
+        // that using the nullifiers.
+        let missing = tx
+            .get_unauthenticated_notes()
+            .map(|note| note.id())
+            .filter(|note_id| !self.output_notes.contains_key(note_id))
+            .collect::<Vec<_>>();
+        if !missing.is_empty() {
+            return Err(VerifyTxError::UnauthenticatedNotesNotFound(missing));
+        }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Oh right. I think I had this check/change of state as part of the pre-mempool checks in the server handler function.

Which I (mistakenly) removed when I reverted my newtype changes to simplify the PR.

I'll add these checks here; though ideally they would be pre-mempool checks and changes. This will require something like:

```rust
pub struct Transaction {
    tx: ProvenTransaction,
    /// Notes found in the store
    found_notes: BTreeSet<NoteId>,
}
```
or even better, having the found notes be the note inclusion proofs. We can make this change now already; it would just mean some adjustments and a second store call to get `note_authenticate_info` in the handler. Best would ofcourse be to fold this store call into tx inputs.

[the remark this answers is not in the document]
Mirko-von-Leipzig: I think I'll punt this work to a new PR. I'll add the checks to `InflightState`, including moving the current nullifier precheck in here.

The PR to add this should be much easier to review without all this noise then.

crates/block-producer/src/mempool/inflight_state/mod.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: I've decided to move the check into inflight state for this PR. See [here](https://github.com/0xPolygonMiden/miden-node/pull/515#discussion_r1808753514) for context.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// Blocks which have been committed but are not yet considered stale.
+    committed_diffs: VecDeque<StateDelta>,
+
+    /// The current block height of the chain.
+    chain_tip: BlockNumber,
+
+    block_in_progress: Option<BTreeSet<BatchJobId>>,
+
+    /// Number of blocks before transaction input is considered stale.
+    staleness: BlockNumber,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Agreed; I just find the `num_xxx` style naming clumsy :/ Was hoping I'd find inspiration at some point :D

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+    transactions: TransactionGraph<ProvenTransaction>,
+
+    /// Inflight batches.
+    batches: BatchGraph,
+
+    /// The next batches ID.
+    next_batch_id: BatchJobId,
+
+    /// Blocks which have been committed but are not yet considered stale.
+    committed_diffs: VecDeque<StateDelta>,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I like that idea. That way `StateDelta` becomes hidden as well.

crates/block-producer/src/mempool/inflight_state/mod.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Actually scratch that. Creating the test data is a pain this way. I'll add back some new types which should make intent more clear.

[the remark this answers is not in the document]
Mirko-von-Leipzig: I've pushed a commit with some new types which make things more explicit, and correct I think. lmkwyt

crates/block-producer/src/mempool/batch_graph.rs
bobbinth: Makes sense - and I think using a newtype is the way to go. I would do this refactoring in a separate PR though.

crates/block-producer/src/mempool/mod.rs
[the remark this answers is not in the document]
bobbinth: Let's add a short comment about this.

crates/block-producer/src/transaction.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Clone, PartialEq)]
+pub struct VerifiedTransaction(ProvenTransaction);
+
+/// A transaction who's proof has been verified, and which has been authenticated against the store.
+#[derive(Clone, PartialEq)]
+pub struct AuthenticatedTransaction {
+    inner: VerifiedTransaction,
+    store_account_state: Option<Digest>,
+    authenticated_notes: BTreeSet<NoteId>,
+    authentication_height: u32,
```
bobbinth, line 22: Let's add doc comments to these fields - otherwise it is not immediately clear what they supposed to contain. For example, `authenticated_notes` is a set of notes that was unauthenticated in the transaction but is currently present in the store, right? (i.e., it is not a set of all authenticated notes in the transaction). Maybe it should be called `newly_authenticated_notes`?

crates/block-producer/src/transaction.rs
```diff
@@ -0,0 +1,131 @@
...
+            return Err(VerifyTxError::InputNotesAlreadyConsumed(nullifiers_already_spent));
+        }
+
+        // Invert the missing notes; i.e. we now know the rest were actually found.
+        let authenticated_notes = self
+            .0
+            .get_unauthenticated_notes()
+            .map(|header| header.id())
+            .filter(|note_id| !inputs.missing_unauthenticated_notes.contains(note_id))
+            .collect();
```
bobbinth, line 122: We should probably change how `GetTransactionInputs` endpoint works so that it returns newly authenticated notes rather than missing notes. Let's create an issue for this.

crates/block-producer/src/transaction.rs
bobbinth, line 1: nit: we should probably put this into some sub-module - maybe something like `domain`.

crates/block-producer/src/mempool/inflight_state/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+
+        // Ensure nullifiers aren't already present.
+        let double_spend = tx
+            .nullifiers()
+            .filter(|nullifier| self.nullifiers.contains(nullifier))
+            .collect::<Vec<_>>();
+        if !double_spend.is_empty() {
+            return Err(VerifyTxError::InputNotesAlreadyConsumed(double_spend));
+        }
```
bobbinth, line 120: I would add a comment here explaining that we've already checked nullifiers against the store.

bobbinth (commented):

Once these and the comments from the previous review are addressed, I think we should merge this. If you'd prefer not to merge it into `next` yet, we can merge into a separate branch. Then we can make more PRs with additional changes against that branch (e.g., the ones discussed [here](https://github.com/0xPolygonMiden/miden-node/pull/515#discussion_r1808662435).

crates/block-producer/src/transaction.rs
Mirko-von-Leipzig: I was thinking we should actually return the inclusion proofs, and carry those with.

With the goal being less store IO at the time of batch/block proving as we already have some of the things. Though maybe this isn't worth it.

crates/block-producer/src/transaction.rs
```diff
@@ -0,0 +1,131 @@
...
+    notes::{NoteId, Nullifier},
+    transaction::{ProvenTransaction, TransactionId, TxAccountUpdate},
+    Digest,
+};
+
+use crate::{errors::VerifyTxError, store::TransactionInputs};
+
+/// A transaction whose proof has been verified.
+#[derive(Clone, PartialEq)]
+pub struct VerifiedTransaction(ProvenTransaction);
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I thought we could use it in the `rpc` component as well to enforce that we verified the proof before sending it onward.

Of course this currently doesn't mean much because the `gRPC` client/server don't enforce this type.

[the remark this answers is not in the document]
Mirko-von-Leipzig: I'll remove it for now. If we do want to re-use it in the `rpc` then just it should move to `proto/domain` but that can be part of a greater `proto` type safety effort if/when we do that.

crates/block-producer/Cargo.toml
```diff
@@ -41,6 +41,7 @@ miden-lib = { workspace = true, features = ["testing"] }
 miden-node-test-macro = { path = "../test-macro" }
 miden-objects = { workspace = true, features = ["testing"] }
 miden-tx = { workspace = true, features = ["testing"] }
+rand = "0.8.5"
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Will do. I used `cargo add` which defaults to the inline version.

crates/block-producer/src/mempool/transaction_graph.rs
```diff
@@ -0,0 +1,595 @@
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    sync::Arc,
+};
+
+use miden_objects::transaction::{ProvenTransaction, TransactionId};
+
+/// Tracks the dependency graph and status of transactions.
+#[derive(Clone, Debug, PartialEq)]
+pub struct TransactionGraph<T> {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Leaving this in; since the intention is to refactor towards a single graph type in a follow-up PR.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        Ok(self.chain_tip.0)
+    }
+
+    /// Returns a set of transactions for the next batch.
+    ///
+    /// Transactions are returned in a valid execution ordering.
+    ///
+    /// Returns `None` if no transactions are available.
+    pub fn select_batch(&mut self) -> Option<(BatchJobId, Vec<ProvenTransaction>)> {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: #519

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        // Update the dependency graph to reflect parents at the batch level by removing all edges
+        // within this batch.
+        for tx in &batch {
+            parents.remove(&tx.id());
+        }
+
+        let batch_id = self.next_batch_id;
+        self.next_batch_id.increment();
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: #519

crates/block-producer/src/mempool/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /// Select batches for the next block.
+    ///
+    /// May return an empty set if no batches are ready.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there is already a block in flight.
+    pub fn select_block(&mut self) -> (BlockNumber, BTreeMap<BatchJobId, TransactionBatch>) {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: #519

crates/block-producer/src/mempool/batch_graph.rs
Mirko-von-Leipzig: #519

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: Follow-up task in #519

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: #519

crates/block-producer/src/transaction.rs
Mirko-von-Leipzig: #519

Mirko-von-Leipzig:
I've addressed the review comments; I'll merge this PR into `next-block-producer` and continue with smaller, more revieawable PRs.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -219,41 +228,74 @@ pub struct BatchProducer {
...
+                tracing::debug!("Batch proof completed.");
+
+                Ok((id, batch))
+            }
         });
     }
 }
 
 impl BatchProducer {
+    /// Starts the [BlockProducer], infinitely producing blocks at the configured interval.
```
bobbinth, line 286: This should say `BatchProducer` - right?

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -219,41 +228,74 @@ pub struct BatchProducer {
...
+    ///
+    /// Block production is sequential and consists of
+    ///
+    ///   1. Pulling the next set of batches from the [Mempool]
+    ///   2. Compiling these batches into the next block
+    ///   3. Proving the block (this is not yet implemented)
+    ///   4. Committing the block to the store
     pub async fn run(self) {
```
bobbinth, line 294: Are these the right comments here? I thought this was supposed to be about batch production (not block production).

crates/block-producer/src/transaction.rs
bobbinth: > I was thinking we should actually return the inclusion proofs, and carry those with.
> 
> With the goal being less store IO at the time of batch/block proving as we already have some of the things. Though maybe this isn't worth it.

My thinking here was that since we'd need to go to the store anyways at the time of batch/block building, it's probably not worth it to get the paths immediately - especially since we don't know if the transaction is valid yet (i.e., it hasn't been checked against the in-flight state).

crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -22,7 +22,8 @@ pub type BatchId = Blake3Digest<32>;
 // TRANSACTION BATCH
 // ================================================================================================
 
-/// A batch of transactions that share a common proof.
+/// A batch of transactions that share a common proof. For any given account, at most 1 transaction
+/// in the batch must be addressing that account (issue: #186).
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Ah. This diff only showed up once I reset the target branch to `next-block-producer`.

The comment is this way on `next`, and I'd previously fixed it in the first `next-block-producer` PR. When I reset the transaction batch changes I reset it this file to `next` status.

Will refix this comment.

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: Yes apparently I cannot read :) Thanks; I'll fix these.

Issue #78, opened 2023-12-01
`BlockWitness`: pass inputs to the VM through the advice provider


plafer:
See https://github.com/0xPolygonMiden/miden-node/pull/71#discussion_r1411786328

bobbinth:
Closed by #516.

PR #516, opened 2024-10-10
`BlockWitness`: pass inputs to the VM through the advice provider

Fumuran:
This PR refactors the `BlockWitness`: it no longer uses the operand stack to provide the inputs to the VM, it uses only advice inputs instead. 

Closes: #78.

### DIFF
--- a/crates/block-producer/src/block_builder/prover/asm/block_kernel.masm
+++ b/crates/block-producer/src/block_builder/prover/asm/block_kernel.masm
@@ -13,154 +13,228 @@ const.CHAIN_MMR_PTR=1000
 
 #! Compute the account root
 #!
-#! Stack: [num_accounts_updated, OLD_ACCOUNT_ROOT,
-#!         NEW_ACCOUNT_HASH_0, account_id_0, ... , NEW_ACCOUNT_HASH_n, account_id_n]
-#! Output: [NEW_ACCOUNT_ROOT]
+#! Inputs:
+#!   Operand stack: []
+#!   Advice stack:  [num_accounts_updated, OLD_ACCOUNT_ROOT, [NEW_ACCOUNT_HASH_i, account_id_i]]
+#! Outputs:
+#!   Operand stack: [NEW_ACCOUNT_ROOT]
 proc.compute_account_root
-    dup neq.0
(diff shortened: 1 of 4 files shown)


Mirko-von-Leipzig (commented):
Code LGTM but I can't really comment on the `masm` changes. Time for me to learn some `masm` I suppose :D

crates/block-producer/src/block_builder/prover/asm/block_kernel.masm
```diff
@@ -13,154 +13,228 @@ const.CHAIN_MMR_PTR=1000
...
-# Stack: [<account root inputs>, <note root inputs>, <nullifier root inputs>, <chain mmr root inputs>]
+#! Inputs:
+#!   Operand stack: []
+#!   Advice stack:  [<account root inputs>, <note root inputs>, <nullifier root inputs>, <chain mmr root inputs>]
+#!   Advice map: {
+#!     PREV_CHAIN_MMR_HASH: [NUM_LEAVES, [peak_i], <maybe padding>]
+#!   }
+#! Outputs:
+#!   Operand stack: [ACCOUNT_ROOT, NOTE_ROOT, NULLIFIER_ROOT, CHAIN_MMR_ROOT]
 begin
```
bobbinth, line 224: This is fine for now, but we should be providing some inputs via the operand stack - otherwise this can be used to provide any arbitrary batch inputs.

Let's create 2 issues in `miden-base` for this: one issue for defining batch kernel inputs/outputs, and the other one for defining block kernel inputs/outputs.

bobbinth (approved):
Looks good! Thank you! Not a super deep review from me - but I did leave a couple of comments inline. The main one is about crating issues in `miden-base` to flash out inputs/outputs of batch and block kernels. If you create some basic issues, i'll try to fill in the details a bit later.

crates/block-producer/src/block_builder/prover/asm/block_kernel.masm
Fumuran: @bobbinth I created two issue templates: https://github.com/0xPolygonMiden/miden-base/issues/919 and https://github.com/0xPolygonMiden/miden-base/issues/920, although didn't manage to fill them properly, I'm not sure I understand what should be done.

PR #517, opened 2024-10-11
Save/restore faucet state and keypair to/from files

polydez:
Resolves https://github.com/0xPolygonMiden/miden-node/issues/504

Save secret key and faucet account state to files and restore them on restart.

Actually we keep current state until the end of the calling `submit_proven_transaction`. This will help to restore even if faucet breaks in the middle of request. But there is still be possible for block producer to fail building block with this transaction. Need to investigate it more and decide whether to make more complex solution for that case.

### DIFF
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -1,30 +1,39 @@
-use std::{cell::RefCell, rc::Rc, time::Duration};
+use std::{
+    path::Path,
+    rc::Rc,
+    sync::{Arc, RwLock},
+    time::Duration,
+};
 
+use anyhow::{anyhow, Context};
 use miden_lib::{
     accounts::faucets::create_basic_fungible_faucet, notes::create_p2id_note,
     transaction::TransactionKernel, AuthScheme,
 };
 use miden_node_proto::generated::{
-    requests::{GetBlockHeaderByNumberRequest, SubmitProvenTransactionRequest},
+    requests::{
+        GetAccountProofsRequest, GetBlockHeaderByNumberRequest, SubmitProvenTransactionRequest,
+    },
     rpc::api_client::ApiClient,
 };
 use miden_objects::{
-    accounts::{Account, AccountDelta, AccountId, AccountStorageMode, AuthSecretKey},
+    accounts::{Account, AccountId, AccountStorageMode, AuthSecretKey},
     assets::{FungibleAsset, TokenSymbol},
     crypto::{
         dsa::rpo_falcon512::SecretKey,
+        hash::rpo::RpoDigest,
         merkle::{MmrPeaks, PartialMmr},
         rand::RpoRandomCoin,
     },
-    notes::{Note, NoteId, NoteType},
-    transaction::{ChainMmr, ExecutedTransaction, InputNotes, TransactionArgs, TransactionScript},
+    notes::{Note, NoteType},
+    transaction::{ChainMmr, ExecutedTransaction, TransactionArgs, TransactionScript},
+    utils::Deserializable,
     vm::AdviceMap,
     BlockHeader, Felt, Word,
 };
 use miden_tx::{
-    auth::BasicAuthenticator, utils::Serializable, DataStore, DataStoreError,
-    LocalTransactionProver, ProvingOptions, TransactionExecutor, TransactionInputs,
-    TransactionProver,
+    auth::BasicAuthenticator, utils::Serializable, LocalTransactionProver, ProvingOptions,
+    TransactionExecutor, TransactionProver,
 };
 use rand::{rngs::StdRng, thread_rng, Rng};
 use rand_chacha::{rand_core::SeedableRng, ChaCha20Rng};
@@ -32,7 +41,8 @@ use tonic::transport::Channel;
 
 use crate::{
     config::FaucetConfig,
-    errors::{InitError, ProcessError},
+    errors::{ErrorHelper, HandlerError},
+    store::{resolve_faucet_state, FaucetDataStore},
 };
 
 pub const DISTRIBUTE_FUNGIBLE_ASSET_SCRIPT: &str =
@@ -52,27 +62,51 @@ pub struct FaucetClient {
 }
 
 unsafe impl Send for FaucetClient {}
-unsafe impl Sync for FaucetClient {}
 
 impl FaucetClient {
-    pub async fn new(config: FaucetConfig) -> Result<Self, InitError> {
-        let (rpc_api, root_block_header, root_chain_mmr) =
-            initialize_faucet_client(config.clone()).await?;
+    /// Creates a new faucet client.
+    pub async fn new(config: &FaucetConfig) -> anyhow::Result<Self> {
+        let (mut rpc_api, root_block_header, root_chain_mmr) =
+            initialize_faucet_client(config).await?;
+        let init_seed: [u8; 32] = [0; 32];
+        let (auth_scheme, authenticator) = init_authenticator(init_seed, &config.secret_key_path)
+            .context("Failed to initialize authentication scheme")?;
 
-        let (faucet_account, account_seed, secret) = build_account(config.clone())?;
-        let faucet_account = Rc::new(RefCell::new(faucet_account));
-        let id = faucet_account.borrow().id();
+        let (mut faucet_account, account_seed) = build_account(config, init_seed, auth_scheme)?;
+        let id = faucet_account.id();
+
+        let faucet_actual_hash: RpoDigest = rpc_api
+            .get_account_proofs(GetAccountProofsRequest {
+                account_ids: vec![id.into()],
+                include_headers: None,
+            })
+            .await
+            .context("Failed to get faucet account proof")?
+            .into_inner()
+            .account_proofs
+            .first()
+            .context("No account proofs returned by RPC")?
+            .account_hash
+            .context("Account hash field is empty")?
+            .try_into()
+            .context("Failed to convert hash from protobuf to `RpoDigest`")?;
+
+        if faucet_account.hash() != faucet_actual_hash {
+            faucet_account = resolve_faucet_state(&config.storage_path, faucet_actual_hash)
+                .await
+                .context("Failed to restore faucet state from files")?;
+        }
+
+        let faucet_account = Arc::new(RwLock::new(faucet_account));
 
         let data_store = FaucetDataStore::new(
-            faucet_account.clone(),
+            config.storage_path.clone(),
+            Arc::clone(&faucet_account),
             account_seed,
             root_block_header,
             root_chain_mmr,
         );
-        let authenticator = BasicAuthenticator::<StdRng>::new(&[(
-            secret.public_key().into(),
-            AuthSecretKey::RpoFalcon512(secret),
-        )]);
+
         let executor = TransactionExecutor::new(data_store.clone(), Some(Rc::new(authenticator)));
 
         let mut rng = thread_rng();
@@ -90,9 +124,9 @@ impl FaucetClient {
         target_account_id: AccountId,
         is_private_note: bool,
         asset_amount: u64,
-    ) -> Result<(ExecutedTransaction, Note), ProcessError> {
-        let asset = FungibleAsset::new(self.id, asset_amount)
-            .map_err(|err| ProcessError::InternalServerError(err.to_string()))?;
+    ) -> Result<(ExecutedTransaction, Note), HandlerError> {
+        let asset =
+            FungibleAsset::new(self.id, asset_amount).or_fail("Failed to create fungible asset")?;
 
         let note_type = if is_private_note {
             NoteType::Private
@@ -108,16 +142,14 @@ impl FaucetClient {
             Default::default(),
             &mut self.rng,
         )
-        .map_err(|err| ProcessError::InternalServerError(err.to_string()))?;
+        .or_fail("Failed to create P2ID note")?;
 
         let transaction_args = build_transaction_arguments(&output_note, note_type, asset)?;
 
         let executed_tx = self
             .executor
             .execute_transaction(self.id, 0, &[], transaction_args)
-            .map_err(|err| {
-                ProcessError::InternalServerError(format!("Failed to execute transaction: {err}"))
-            })?;
+            .or_fail("Failed to execute transaction")?;
 
         Ok((executed_tx, output_note))
     }
@@ -126,18 +158,15 @@ impl FaucetClient {
     pub async fn prove_and_submit_transaction(
         &mut self,
         executed_tx: ExecutedTransaction,
-    ) -> Result<u32, ProcessError> {
-        let delta = executed_tx.account_delta().clone();
-
+    ) -> Result<u32, HandlerError> {
         // Prepare request with proven transaction.
         // This is needed to be in a separated code block in order to release reference to avoid
         // borrow checker error.
         let request = {
             let transaction_prover = LocalTransactionProver::new(ProvingOptions::default());
 
-            let proven_transaction = transaction_prover.prove(executed_tx).map_err(|err| {
-                ProcessError::InternalServerError(format!("Failed to prove transaction: {err}"))
-            })?;
+            let proven_transaction =
+                transaction_prover.prove(executed_tx).or_fail("Failed to prove transaction")?;
 
             SubmitProvenTransactionRequest {
                 transaction: proven_transaction.to_bytes(),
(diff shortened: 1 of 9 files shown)


bobbinth:
Not a review yet (and I have barely looked at the actual code), but a couple of thoughts:

If the faucet is public (i.e., it is a public account), the only thing we need to save locally is the secret key, right? The thought here is that we can always retrieve the faucet state from the chain - and so, no need to save it to disk (we still want to keep it in memory as it is being done now, I think).

There are two cases when we may need to retrieve the faucet state from chain:

1. During faucet restart.
2. If trying to send a transaction to the node fails.

The latter case is meant to protect against situations where the faucet's in-memory state somehow gets out of sync with the node. If this happens, we try to get the most recent state from the chain and hopefully, everything will work fine after that. Failing to send a transaction may happen for other reasons though (e.g., network failures) - so, if we can handle these gracefully, that would be ideal.

If the faucet is private, then we do need to state some number of states locally - but is there a downside to making it public?

Mirko-von-Leipzig:
> If the faucet is private, then we do need to state some number of states locally - but is there a downside to making it public?

No downside afaik. One can make an argument that we should provide examples of both public and private implementations but that can wait (if we care to at all).

polydez:
As long as transactions referring incorrect faucet account's state are rejected before putting to block producer's transaction queue, it seems to be the easiest solution for us.

bobbinth:
> As long as transactions referring incorrect faucet account's state are rejected before putting to the block producer's transaction queue, it seems to be the easiest solution for us.

Let's do it this way then as this is how the block producer will work for the next couple of years.

bin/faucet/src/handlers.rs
```diff
@@ -55,33 +59,80 @@ pub async fn get_tokens(
...
+                break (created_note, block_height);
+            },
+            Err(err) => {
+                error!(
+                    target: COMPONENT,
+                    %err,
+                    "Failed to prove and submit transaction",
+                );
+
+                info!(target: COMPONENT, "Trying to request account state from the node...");
```
Mirko-von-Leipzig, line 103:
```suggestion
                info!(target: COMPONENT, "Requesting account state from the node...");
```

bin/faucet/src/handlers.rs
```diff
@@ -55,33 +59,80 @@ pub async fn get_tokens(
...
+
+        let new_hash = faucet_account.hash();
+
+        // Run transaction prover & send transaction to node
+        info!(target: COMPONENT, "Proving and submitting transaction.");
+        match client.prove_and_submit_transaction(executed_tx.clone()).await {
+            Ok(block_height) => {
+                break (created_note, block_height);
+            },
+            Err(err) => {
```
Mirko-von-Leipzig, line 96: Is there an error variant we could match here that would specific a state mismatch? Instead of trying on all errors?

bin/faucet/src/handlers.rs
```diff
@@ -55,33 +59,80 @@ pub async fn get_tokens(
...
-        req.is_private_note,
-        req.asset_amount,
-    )?;
-
-    // Run transaction prover & send transaction to node
-    info!(target: COMPONENT, "Proving and submitting transaction.");
-    let block_height = client.prove_and_submit_transaction(executed_tx).await?;
+        .map_err(|err| HandlerError::BadRequest(err.to_string()))?;
+
+    let (created_note, block_height) = loop {
```
Mirko-von-Leipzig, line 71: The loop scares me a bit. We don't actually expect the state to desync -- and something like this happening would constitute a bug, or having spawned multiple faucets with the same configuration and secret keys -- which itself is an issue.

I suggest we do something like:

On faucet startup, fetch the account state. If the account is not yet on chain, then submit a transaction to declare it. This means that the non-startup txns should always have an account seed of `None` (iiuc). In other words, for all `get_tokens` calls the account is known to be on-chain and public etc since these conditions were ensured at startup. This fulfills the requirement to be restartable by itself already.

We then need to decide what to do if `get_tokens` fails due to state mismatch. imo we should crash the faucet 😅 This is really an unexpected condition and means something is really off. However we can decide to at least make this somewhat robust against this condition (though I still argue we don't need to). If we do want then I would rather do something like:

```rust
match client.prove_and_submit_transaction(...).await {
    Err(err) if err.is_state_mismatch() => {
        // update account state and resubmit txn
    }, 
    other => other
}
```
which would only retry once. The current `loop` achieves the same thing ofc, but it does invite trouble if we have a bug elsewhere in the logic causing an infinite retry loop. I would also suggest factoring out the state update code, and the txn proving & submission code to make this easy to read.

bin/faucet/src/store.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+impl DataStore for FaucetDataStore {
+    fn get_transaction_inputs(
+        &self,
+        account_id: AccountId,
+        _block_ref: u32,
+        _notes: &[NoteId],
+    ) -> Result<TransactionInputs, DataStoreError> {
+        let account = self.faucet_account.read().expect("Poisoned lock");
```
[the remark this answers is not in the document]
polydez: I don't think we should introduce new dependency only for elimination of these `expect`'s. :)

bin/faucet/src/main.rs
```diff
@@ -59,16 +60,15 @@ pub enum Command {
 // =================================================================================================
 
 #[tokio::main]
-async fn main() -> Result<(), InitError> {
-    miden_node_utils::logging::setup_logging()
-        .map_err(|err| InitError::FaucetFailedToStart(err.to_string()))?;
+async fn main() -> anyhow::Result<()> {
+    miden_node_utils::logging::setup_logging().context("Failed to initialize logging")?;
```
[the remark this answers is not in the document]
polydez: I think, it's more to personal perception. For me contexts are like error chain. But I think we should consider solid approach of using error context messages and follow this approach everywhere in our code.

By the way, examples in `anyhow`'s documentation provide very similar context messages: https://docs.rs/anyhow/latest/anyhow/trait.Context.html

bin/faucet/src/client.rs
```diff
@@ -148,137 +150,132 @@ impl FaucetClient {
...
 
     let (faucet_account, account_seed) = create_basic_fungible_faucet(
-        seed,
+        init_seed,
         token_symbol,
         config.decimals,
         Felt::try_from(config.max_supply)
-            .map_err(|err| InitError::AccountCreationError(err.to_string()))?,
-        AccountStorageMode::Private,
+            .map_err(|err| anyhow!("Error converting max supply to Felt: {err}"))?,
```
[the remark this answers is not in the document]
polydez: This doesn't work, unfortunately:
```
error[E0599]: the method `context` exists for enum `Result<Felt, String>`, but its trait bounds were not satisfied
   --> bin/faucet/src/client.rs:244:43
    |
244 |         Felt::try_from(config.max_supply).context("Error converting max supply to Felt")?,
    |                                           ^^^^^^^ method cannot be called on `Result<Felt, String>` due to unsatisfied trait bounds
    |
   ::: /Users/polydez/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/string.rs:365:1
    |
365 | pub struct String {
    | ----------------- doesn't satisfy `_: StdError`
    |
   ::: /Users/polydez/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:527:1
```

bin/faucet/src/errors.rs
```diff
@@ -60,3 +62,13 @@ impl IntoResponse for ProcessError {
...
+
+pub trait ErrorHelper<T, E: std::error::Error> {
+    fn or_fail(self, message: impl Display) -> Result<T, E>;
+}
+
+impl<T, E: std::error::Error> ErrorHelper<T, HandlerError> for Result<T, E> {
+    fn or_fail(self, message: impl Display) -> Result<T, HandlerError> {
+        self.map_err(|err| HandlerError::InternalServerError(format!("{message}: {err}")))
+    }
+}
```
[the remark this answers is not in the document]
polydez: Nice idea, thanks!

bin/faucet/src/client.rs
```diff
@@ -148,137 +150,132 @@ impl FaucetClient {
...
+    Ok((auth_scheme, authenticator))
+}
 
 /// Builds a new faucet account with the provided configuration.
 ///
 /// Returns the created account, its seed, and the secret key used to sign transactions.
-fn build_account(config: FaucetConfig) -> Result<(Account, Word, SecretKey), InitError> {
+fn build_account(
+    config: &FaucetConfig,
+    init_seed: [u8; 32],
```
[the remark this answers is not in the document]
polydez: Good catch, thank you! This is not the only issue here I can see so far. One of the most important things here is a vulnerability with a secret key generation. Since we initialize random number generator with always the same seed, we get the same secret for each faucet. This means, everybody can generate it by themselves and interact with the faucet. We should fix it before merging of this PR.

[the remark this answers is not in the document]
igamigo: `SecretKey` generation yields the public key which is stored on the account's storage which means it will affect the storage commitment, and thus the account seed (3 in your list) as well.

bin/faucet/src/handlers.rs
polydez: Currently we have only "invalid argument" error for any error in this endpoint. We should improve errors returned from the block producer, but I didn't make it now in order to not conflict with your refactoring of the block producer. But we should improve it in the future, I will put `TODO` here.

bin/faucet/src/handlers.rs
polydez: Yes, it makes sense, thank you. I will rewrite my solution to get initial state from the node on startup and fail if the state was changed unexpectedly.

bin/faucet/src/client.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Ah then `with_context(|err| format!(...))` is the equivalent I think. Though its much the same.

polydez:
We need to decide, what to do with faucets.

Currently we have 2 faucets:
1. Created in genesis, but not used, as I can see.
2. Created in the faucet. 

This is unusual comparing to other blockchains where we have native asset (like ETH, SOL, etc.) which initialized in the genesis and tokens, provided by token contracts (like ERC20 in Etheretum, SPL Token in Solana and so on). Currently it seem that we don't have "native" asset in Miden, but some number of faucets, each minting their own token. Every user can create their own faucet(s) and asset(s). 

This design looks good in terms of creation of different decentralized tokens, but what about native cryptocurrency? I think, we will need native Miden asset at least for implementing fees.

I propose to use precreated (predeployed, in other words) faucet from the genesis in our faucet web app. That means, our faucet will be the only place to mint native Miden cryptocurrency. Users will still be able to create their own faucets for tokens they wish, but they will have to use our faucet in testnet to get native cryptocurrency to pay fees and other tasks.

polydez:
By the way, each asset has link to a faucet id which has minted this asset. But we won't use faucets in mainnet. We will probably have governance contract(s) or some other form of initial native cryptocurrency distribution. Will it correct to call them as faucets? Maybe mint id will better address the meaning of that field?

bobbinth:
> This is unusual comparing to other blockchains where we have native asset (like ETH, SOL, etc.) which initialized in the genesis and tokens, provided by token contracts (like ERC20 in Etheretum, SPL Token in Solana and so on). Currently it seem that we don't have "native" asset in Miden, but some number of faucets, each minting their own token. Every user can create their own faucet(s) and asset(s).

Technically, there are no "native" assets on Miden - or conversely, all assets are native. That is, there is nothing that currently sets one asset apart from another.

> This design looks good in terms of creation of different decentralized tokens, but what about native cryptocurrency? I think, we will need native Miden asset at least for implementing fees.

It is possible to do fees without native assets though it would probably complicate the design somewhat. This is something we'll need to think through over the next couple of months.

> I propose to use precreated (predeployed, in other words) faucet from the genesis in our faucet web app. That means, our faucet will be the only place to mint native Miden cryptocurrency. Users will still be able to create their own faucets for tokens they wish, but they will have to use our faucet in testnet to get native cryptocurrency to pay fees and other tasks.

I agree with this. It would be great to have the faucet webapp use the faucet that was generated in the genesis block. One question is how would this affect the deployment process.

> By the way, each asset has link to a faucet id which has minted this asset. But we won't use faucets in mainnet. We will probably have governance contract(s) or some other form of initial native cryptocurrency distribution. Will it be correctly to call them as faucets? Maybe mint id will better address the meaning of that field?

I think many of these things are still to be defined. Re terminology, In our context "faucet" means "an account which can issue assets." It's definitely not ideal as "faucet" is used differently in other system - so, if we come up with a better term we can switch to it.

bin/faucet/src/handlers.rs
bobbinth: I think I'm OK with the "retry once" strategy in case of state mismatch. Overall, here is my current understanding of what we are doing (correct me if I missed anything):

1. For now we are dealing with public faucets only (we can add support for private faucets later if needed).
2. On startup, we assume that our faucet is already on-chain and try to fetch its most recent state. If that fails, we crush the faucet.
3. During minting, we update the local state and submit the transaction to the node. If the transaction fails, we try to get the latest state from the node and try to submit the transaction again. If anything fails here we crush.

One question is how are the faucets created? For our faucet we can use the faucet created in the genesis block as discussed in https://github.com/0xPolygonMiden/miden-node/pull/517#issuecomment-2430532332. But what about other faucets (this could be an improvement we add later on)?

Mirko-von-Leipzig:
> I agree with this. It would be great to have the faucet webapp use the faucet that was generated in the genesis block. One question is how would this affect the deployment process.

I think its just the secret key that causes issues correct? This could be handled by either

1. storing the secret key in github secrets for automatic deployment, or
2. having some alternate custody scheme.

(1) would be good enough for now, but in the long run we'd probably want something else though I have no idea what that looks like.

bobbinth:
> I think its just the secret key that causes issues correct? This could be handled by either
> 
> 1. storing the secret key in github secrets for automatic deployment, or
> 2. having some alternate custody scheme.
> 
> (1) would be good enough for now, but in the long run we'd probably want something else though I have no idea what that looks like.

In the long run, the scheme would probably not be based on secret keys - but how exactly we'll do it still needs to be figured out.

For the short run, I'd be somewhat hesitant of putting the secret key into Github as then some can pretty easily break the testnet (e.g., by draining the faucet).

bin/faucet/src/handlers.rs
Mirko-von-Leipzig: > One question is how are the faucets created?

This actually creates the account if its not there as part of `get_tokens` afaict. Or at least I think that's what the data store's transaction inputs is doing?

Though I'm finding it difficult to keep track 😓 Not this PRs fault; things are quite muddled in the faucet code.

Mirko-von-Leipzig: I think our end goal should be something like:

Move faucet account creation to a separate subcommand. This should create a random keypair and create the account on chain in `miden-faucet init` (but maybe a better name e.g. `miden-faucet create`).

`miden-faucet start` then loads account state on startup (crashes if missing), and just updates state in `get_tokens`. I don't think there is much point in retrying if the state is different - that would constitute a pretty bad internal failure.

Mirko-von-Leipzig: > I'm fine with not doing it now if it complicates things, but I think retrying once could be useful. For example, what if there are two parallel instances of the faucet running and they make updates independently to the same account. Unlikely to happen, but also not sure if this should be a "catastrophic failure".

My counter argument would be that exactly one of the two would crash which to me is optimal solution 😬 And we would know about it instead of both just chugging along.

polydez: I like the solution proposed by @Mirko-von-Leipzig (and developed further by @bobbinth): explicitly create faucet account by special command or import of existing faucet by using account ID and secret. And I don't think that we really need to support two faucet web apps working in parallel, it's better to fail one of them (this significantly simplifies `get_tokens` method). This solution is close to what I'm implementing now, but not exactly the same, so I'm going to update my solution.

Thank you for help!

bin/faucet/src/store.rs
```diff
@@ -0,0 +1,76 @@
...
+        if account_id != account.id() {
+            return Err(DataStoreError::AccountNotFound(account_id));
+        }
+
+        let empty_input_notes =
+            InputNotes::new(Vec::new()).map_err(DataStoreError::InvalidTransactionInput)?;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: They already have `Default` implemented so we should just use that :)

crates/utils/src/crypto.rs
```diff
@@ -0,0 +1,13 @@
...
+    Felt,
+};
+use rand::{Rng, RngCore};
+
+/// Creates a new RPO Random Coin with random seed
+pub fn get_rpo_random_coin<T: RngCore>(rng: &mut T) -> RpoRandomCoin {
+    let auth_seed: [u64; 4] = rng.gen();
+    let rng_seed = RpoDigest::from(auth_seed.map(Felt::new));
+
+    RpoRandomCoin::new(rng_seed.into())
```
Mirko-von-Leipzig, line 12: I'm surprised that `RpoRandomCoin` doesn't implement https://docs.rs/rand/latest/rand/trait.SeedableRng.html.

Feels like this is a very natural use case.

bin/faucet/src/client.rs
```diff
@@ -54,32 +52,64 @@ pub struct FaucetClient {
...
+
+                account
+            },
+
+            Err(err) => match err {
+                ClientError::RequestError(status) if status.code() == tonic::Code::NotFound => {
+                    info!(target: COMPONENT, "Faucet account not found in the node");
+
+                    faucet_account_data.account
+                },
```
Mirko-von-Leipzig, line 87: Is the idea that the mint transaction will generate the account in this scenario?

If so, could we not move that into the `CreateFaucetAccount` command? And then fail on all errors here instead.

Mirko-von-Leipzig (approved):
Should `init` and `create-faucet-account` be one command?

crates/utils/src/crypto.rs
polydez: Yes, this would simplify usage a bit. But this also would require a bit more complex conversion (from `[u8; 32]` to `RpoDigest`).

bobbinth (commented):
Thank you! Looks good! I left a few comments inline. Also, we should update the README file with new instructions on how to operate the faucet.

bin/node/src/commands/genesis/mod.rs
```diff
@@ -121,21 +112,19 @@ fn create_accounts(
     fs::create_dir_all(&accounts_path).context("Failed to create accounts directory")?;
 
     let mut final_accounts = Vec::new();
+    let mut faucet_count = 0;
+    let mut rng = ChaCha20Rng::from_seed(rand::random());
 
     for account in accounts {
         // build offchain account data from account inputs
-        let mut account_data = match account {
+        let (mut account_data, name) = match account {
```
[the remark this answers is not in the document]
polydez: I also was thinking about it and decided to keep it as is because we create account data but don't store it to genesis block yet. We will store it later after returning from the function. But I have no strong preference about it, and ready to update the comment if you disagree with my point.

bin/node/src/commands/genesis/mod.rs
```diff
@@ -144,45 +133,48 @@ fn create_accounts(
                     auth_scheme,
                 )?;
 
-                AccountData::new(account, Some(account_seed), auth_secret_key)
+                let name = if faucet_count == 0 {
+                    "faucet".to_string()
+                } else {
+                    format!("faucet{faucet_count}")
+                };
```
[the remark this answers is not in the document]
polydez: I also don't really like this part of code. :) I wanted to name the file as `faucet.mac` since in the wast majority of cases there will be only one single faucet account in the genesis. But naming it as `faucet0.mac` also may be fine.

bin/faucet/src/main.rs
```diff
@@ -48,27 +59,40 @@ pub enum Command {
         config: PathBuf,
     },
 
-    /// Generates default configuration file for the faucet
+    /// Create a new public faucet account and save to the file
+    CreateFaucetAccount {
```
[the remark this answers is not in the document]
polydez: Thanks! No, this doesn't create the account on-chain, only writes it to the *.mac file. Account will be created on the first mint transaction submission.

bin/faucet/src/client.rs
```diff
@@ -54,31 +52,64 @@ pub struct FaucetClient {
...
 
 impl FaucetClient {
-    pub async fn new(config: FaucetConfig) -> Result<Self, InitError> {
-        let (rpc_api, root_block_header, root_chain_mmr) =
-            initialize_faucet_client(config.clone()).await?;
-
-        let (faucet_account, account_seed, secret) = build_account(config.clone())?;
-        let id = faucet_account.id();
+    /// Creates a new faucet client.
+    pub async fn new(config: &FaucetConfig) -> Result<Self, ClientError> {
```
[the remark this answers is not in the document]
polydez: Yes, we are fetching the faucet state from the chain, but don't create it here if it wasn't found. We will create it later (in the first minting transaction).

bin/node/src/commands/genesis/mod.rs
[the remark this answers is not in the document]
polydez: I've tried to improve the expression here, please take a look. :)

bin/faucet/src/client.rs
Mirko-von-Leipzig: > It was my first try, but then I realized that it's not a straightforward thing. In order to submit transaction, we need to instantiate client with executor, storage and so on, so it will require at least refactoring of the client and all the stuff it uses for the transaction execution and submission. I'm not sure this is worth such refactoring, but I'm open to discussion.

That's a fair point. I think overall the faucet code could do with a larger refactoring but definitely as a separate issue when we have some extra bandwidth again.

> The second challenging thing for me here is that we would need to create faucet account separately from minting transaction and I'm not sure if I understand how to do it correctly. :) Will it be enough to just omit execution script from transaction args and keep the remaining the same?

Good question. I have no idea :D

bin/faucet/README.md
bobbinth, line 1: Should we update step 1 in this file as well? I'm actually not sure if docker images work any more.

Also, is there some disconnect between step 1 and step 2? That is, if we run faucet in the testing mode, we should run node in the testing mode as well (maybe that's what's happening already, but I'd make it more explicit).

bobbinth (approved):
Looks good! Thank you! I left one more comment for the readme file. Once it is addressed, we can merge.

bin/faucet/README.md
Mirko-von-Leipzig: Yes both should be in the same mode. I do wish we could make the difficulty configurable and not gated behind a compile time feature. Then we could decouple things by having an rpc endpoint to fetch the current difficulty of the node for example.

polydez: @bobbinth,

> Should we update step 1 in this file as well? I'm actually not sure if docker images work any more.

I've checked, it works.

> Also, is there some disconnect between step 1 and step 2? That is, if we run faucet in the testing mode, we should run node in the testing mode as well (maybe that's what's happening already, but I'd make it more explicit).

If we run node by using Docker, it's created in "testing" mode. All instructions in faucet's readme build node and faucet in "testing" mode. I've improved documentation in order to make it more descriptive about that.

PR #518, opened 2024-10-18
feat(store): improve errors

polydez:
In this PR we improve error statuses returned by the some store API.

Initially it was needed by https://github.com/0xPolygonMiden/miden-node/issues/504 (I wanted to be able to process "account not in DB" errors), but finally it became unused due to changed solution. But it might make our API better, so I propose to keep these changes.

### DIFF
--- a/crates/store/src/server/api.rs
+++ b/crates/store/src/server/api.rs
@@ -70,7 +70,7 @@ impl api_server::Api for StoreApi {
     )]
     async fn get_block_header_by_number(
         &self,
-        request: tonic::Request<GetBlockHeaderByNumberRequest>,
+        request: Request<GetBlockHeaderByNumberRequest>,
     ) -> Result<Response<GetBlockHeaderByNumberResponse>, Status> {
         info!(target: COMPONENT, ?request);
         let request = request.into_inner();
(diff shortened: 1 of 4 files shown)


crates/store/src/server/api.rs
```diff
@@ -593,30 +578,24 @@ impl api_server::Api for StoreApi {
...
 }
 
 // UTILITIES
 // ================================================================================================
 
-/// Formats an error
-fn internal_error<E: core::fmt::Debug>(err: E) -> Status {
-    Status::internal(format!("{:?}", err))
+/// Formats an "Internal error" error
+fn internal_error<E: core::fmt::Display>(err: E) -> Status {
```
Mirko-von-Leipzig, line 592: Do we still need this one?

Mirko-von-Leipzig (approved):
I think this is good improvement.

There are still possibilities where the mapping might be the wrong default for some situations; e.g. if we have bad internal state and attempt to retrieve an unknown block then that shouldn't be exposed.

Though I think this will always remain the case; and this PR at least restrains the default non-internal mappings to a smaller set.

crates/store/src/server/api.rs
polydez: Yes, it's still used in 5 places. Actually we don't really need this at all, but it is a convenient method which allows us just to write `.map_err(internal_error)` instead of `.map_err(|err| Status::internal(err.to_string()))`.

polydez:
> There are still possibilities where the mapping might be the wrong default for some situations; e.g. if we have bad internal state and attempt to retrieve an unknown block then that shouldn't be exposed.

Yes, for such cases we can map it to whatever we need. :)

crates/store/src/server/api.rs
Mirko-von-Leipzig: I thought maybe those would be covered by just using `.context(..)` 👍

bobbinth (approved):
Looks good! Thank you!

PR #521, opened 2024-10-22
feat: Return code for FPI

### LINKED ISSUE #520: Add `AccountCode` to `GetAccountProof` endpoint
For FPI, we need to load an `AccountCode`'s commitment |-> `AccountCode` mapping into the advice provider. 
However, the `GetAccountProof` endpoint does not return it. 
We could make it so that the API request includes an  optional`AccountCode` commitment, and the endpoint returns an `Option<AccountCode>`, where the value is `Some(AccountCode)` if the current code version is different than the provided one and `None` otherwise.

igamigo:
Closes #520

### DIFF
--- a/crates/store/src/server/api.rs
+++ b/crates/store/src/server/api.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{collections::BTreeSet, sync::Arc};
 
 use miden_node_proto::{
     convert,
@@ -488,10 +488,19 @@ impl api_server::Api for StoreApi {
 
         debug!(target: COMPONENT, ?request);
 
-        let account_ids = convert(request.account_ids);
         let include_headers = request.include_headers.unwrap_or_default();
-        let request_code_commitments = try_convert(request.code_commitments)
-            .map_err(|err| Status::invalid_argument(format!("Invalid code commitment: {}", err)))?;
+        let account_ids: Vec<u64> = convert(request.account_ids);
+        let request_code_commitments: BTreeSet<RpoDigest> = try_convert(request.code_commitments)
+            .map_err(|err| {
+            Status::invalid_argument(format!("Invalid code commitment: {}", err))
+        })?;
+
+        if account_ids.len() != request_code_commitments.len() {
+            return Err(Status::invalid_argument(
+                "Amount of code commitments should be the same as the requested Account IDs",
+            ));
+        }
+
         let (block_num, infos) = self
             .state
             .get_account_proofs(account_ids, request_code_commitments, include_headers)
--- a/crates/store/src/server/api.rs
+++ b/crates/store/src/server/api.rs
@@ -485,9 +485,9 @@ impl api_server::Api for StoreApi {
         request: Request<GetAccountProofsRequest>,
     ) -> Result<Response<GetAccountProofsResponse>, Status> {
         let request = request.into_inner();
-        if request.account_ids.len() != request.code_commitments.len() {
+        if request.account_ids.len() >= request.code_commitments.len() {
             return Err(Status::invalid_argument(
-                "Amount of code commitments should be the same as the requested Account IDs",
+                "Amount of code commitments should not be larger than amount of requested accounts.",
             ));
         }
 
--- a/crates/store/src/server/api.rs
+++ b/crates/store/src/server/api.rs
@@ -485,6 +485,11 @@ impl api_server::Api for StoreApi {
         request: Request<GetAccountProofsRequest>,
     ) -> Result<Response<GetAccountProofsResponse>, Status> {
         let request = request.into_inner();
+        if request.account_ids.len() != request.code_commitments.len() {
+            return Err(Status::invalid_argument(
+                "Amount of code commitments should be the same as the requested Account IDs",
+            ));
+        }
 
         debug!(target: COMPONENT, ?request);
 
@@ -494,12 +499,6 @@ impl api_server::Api for StoreApi {
             .map_err(|err| {
             Status::invalid_argument(format!("Invalid code commitment: {}", err))
         })?;
-
-        if account_ids.len() != request_code_commitments.len() {
-            return Err(Status::invalid_argument(
-                "Amount of code commitments should be the same as the requested Account IDs",
-            ));
-        }
 
         let (block_num, infos) = self
             .state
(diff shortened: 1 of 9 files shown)


proto/requests.proto
```diff
@@ -135,6 +135,9 @@ message GetAccountStateDeltaRequest {
 message GetAccountProofsRequest {
     // List of account IDs to get states.
     repeated account.AccountId account_ids = 1;
+    // Account code commitments corresponding to the last-known `AccountCode` for requested 
+    // accounts. Responses will include only the ones that are not known to the caller.
+    repeated digest.Digest code_commitments = 2;
```
bobbinth, line 140: I would probably mention that the number of code commitments must be exactly the same as the number of account IDs.

Also, I think we should return the code only if `include_headers` was set to true.

crates/store/src/server/api.rs
```diff
@@ -490,8 +490,12 @@ impl api_server::Api for StoreApi {
 
         let account_ids = convert(request.account_ids);
         let include_headers = request.include_headers.unwrap_or_default();
-        let (block_num, infos) =
-            self.state.get_account_proofs(account_ids, include_headers).await?;
+        let request_code_commitments = try_convert(request.code_commitments)
+            .map_err(|err| Status::invalid_argument(format!("Invalid code commitment: {}", err)))?;
```
bobbinth, line 494: Should we also check here that the number of code commitments is the same as the number of account IDs?

proto/responses.proto
```diff
@@ -188,8 +188,11 @@ message AccountProofsResponse {
     digest.Digest account_hash = 2;
     // Authentication path from the `account_root` of the block header to the account.
     merkle.MerklePath account_proof = 3;
+    // Account code, returned only in the case where the request code commitment does not match
+    // with the current one.
+    optional bytes account_code = 4;
     // State header for public accounts. Filled only if `include_headers` flag is set to `true`.
-    optional AccountStateHeader state_header = 4;
+    optional AccountStateHeader state_header = 5;
```
bobbinth, line 195: I would probably put `account_code` inside `AccountStateHeader` (maybe we should rename it to something else).

The idea is that we'd include the code only if we also return the headers as getting just the code is probably not very useful.

bobbinth (commented):
Thank you! Looks good! I left a few comments inline. The main one is about returning account code only if `include_headers` is set to true.

proto/requests.proto
igamigo: I made this change, but I'm not sure it improves the API too much. Basically I was seeing this parameters as a set of known commitments that the node could check against without it necessarily being the same length as the requested accounts (maybe the caller knows 1 out of 3 requested accounts and so sending 1 commitment would be fine). Although having the check is nice to not receive an unnecessarily-long parameter.

proto/responses.proto
igamigo: Changed, though couldn't think of a good name for this

proto/requests.proto
bobbinth: But how would we be able to map which account ID corresponds to which commitment? We could refactor the request to use `(AccountID, CodeRoot)` tuples - but I think this may be complicating it even more.

One thing I don't love about the current approach is that now the user has to provide a digest for every account ID (even if they are not interested in account code). But I'm not seeing an easy way around this.

crates/store/src/state.rs
```diff
@@ -691,7 +693,7 @@ impl State {
         let inner_state = self.inner.read().await;
 
         let state_headers = if !include_headers {
-            BTreeMap::<AccountId, AccountStateHeader>::default()
+            BTreeMap::<u64, AccountStateHeader>::new()
```
bobbinth, line 696: Why was this change necessary?

proto/requests.proto
bobbinth: Ah - I get it now: you were thinking of the list of code commitments as not mapped to a specific ID but for a request in general. I think your view here is better. But I think we should still check to make sure the number of commitments is not greater than the number of IDs.

crates/store/src/server/api.rs
```diff
@@ -485,13 +485,25 @@ impl api_server::Api for StoreApi {
         request: Request<GetAccountProofsRequest>,
     ) -> Result<Response<GetAccountProofsResponse>, Status> {
         let request = request.into_inner();
+        if request.account_ids.len() != request.code_commitments.len() {
```
bobbinth, line 488: Based on the recent discussion, I think this should be an inequality check (i.e., number of code commitments should be smaller than or equal to the number of IDs).

proto/requests.proto
igamigo: Do we need to correlate request account IDs with request code commitments? The response account headers contain the code root already, so the response for a single account ID would either:

- Contain `None` for `account_code`, but the `AccountHeader` would have a `code_commitment` that the user knows (ie, which was on the request)
- Contain `Some(code)`, where the `AccountHeader` would have a `code_commitment` the user does not know (ie, it was not on the request's list of commitments)

Doesn't really matter for which account we thought the code commitment corresponded to on the request, because with the response we should have all the pieces.
So if `account_1` and `account_2` had the same code commitment (according to the API user's perspective), the list they would send would only contain one code commitment, but maybe the response comes back with each account having a new `AccountCode` which could even be different from each other. 
But very possible I'm missing something basic here.

proto/requests.proto
```diff
@@ -135,6 +135,9 @@ message GetAccountStateDeltaRequest {
...
     // Optional flag to include header in the response. `false` by default.
-    optional bool include_headers = 2;
+    optional bool include_headers = 3;
```
bobbinth, line 142: nit: I would move `include_headers` above `code_commitments`.

Also, we should update comments to make it clear that account code only is included if `include_headers` is true, and also that it is not associated with a specific account ID (i.e., we check code roots against all accounts).

bobbinth (approved):
Looks good! Thank you! I left a couple more comments inline. Once they are addressed, we should be good to merge.

crates/store/src/server/api.rs
```diff
@@ -485,13 +485,25 @@ impl api_server::Api for StoreApi {
         request: Request<GetAccountProofsRequest>,
     ) -> Result<Response<GetAccountProofsResponse>, Status> {
         let request = request.into_inner();
+        if request.account_ids.len() >= request.code_commitments.len() {
+            return Err(Status::invalid_argument(
+                "Amount of code commitments should not be larger than amount of requested accounts.",
```
bobbinth, line 490: nit: I would use "number" instead of "amount" here.

crates/store/src/state.rs
igamigo: It wasn't, I thought I was rolling it back correctly but I wasn't (`AccountId` here is a newtype for `u64`)

PR #528, opened 2024-10-24
Migrate to `miden-vm` v0.11

Fumuran:
This PR updates the code of the `miden-node` to work with a next v0.11 version of the `miden-vm` and next version of `miden-base`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ readme = "README.md"
 
 [workspace.dependencies]
 miden-air = { git = "https://github.com/0xPolygonMiden/miden-vm.git", branch = "next" }
-miden-lib = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "andrew-migrate-to-new-padding-rule" }
+miden-lib = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "next" }
 miden-node-block-producer = { path = "crates/block-producer", version = "0.6" }
 miden-node-faucet = { path = "bin/faucet", version = "0.6" }
 miden-node-proto = { path = "crates/proto", version = "0.6" }
@@ -35,10 +35,10 @@ miden-node-rpc-proto = { path = "crates/rpc-proto", version = "0.6" }
 miden-node-store = { path = "crates/store", version = "0.6" }
 miden-node-test-macro = { path = "crates/test-macro" }
 miden-node-utils = { path = "crates/utils", version = "0.6" }
-miden-objects = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "andrew-migrate-to-new-padding-rule" }
+miden-objects = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "next" }
 miden-processor = { git = "https://github.com/0xPolygonMiden/miden-vm.git", branch = "next" }
 miden-stdlib = { git = "https://github.com/0xPolygonMiden/miden-vm.git", branch = "next", default-features = false }
-miden-tx = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "andrew-migrate-to-new-padding-rule" }
+miden-tx = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "next" }
 prost = { version = "0.13" }
 thiserror = { version = "1.0" }
 tokio = { version = "1.40", features = ["rt-multi-thread"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ exclude = [".github/"]
 readme = "README.md"
 
 [workspace.dependencies]
-miden-air = { git = "https://github.com/0xPolygonMiden/miden-vm.git", branch = "al-migrate-new-padding-rule" }
+miden-air = { git = "https://github.com/0xPolygonMiden/miden-vm.git", branch = "next" }
 miden-lib = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "andrew-migrate-to-new-padding-rule" }
 miden-node-block-producer = { path = "crates/block-producer", version = "0.6" }
 miden-node-faucet = { path = "bin/faucet", version = "0.6" }
@@ -36,8 +36,8 @@ miden-node-store = { path = "crates/store", version = "0.6" }
 miden-node-test-macro = { path = "crates/test-macro" }
 miden-node-utils = { path = "crates/utils", version = "0.6" }
 miden-objects = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "andrew-migrate-to-new-padding-rule" }
-miden-processor = { git = "https://github.com/0xPolygonMiden/miden-vm.git", branch = "al-migrate-new-padding-rule" }
-miden-stdlib = { git = "https://github.com/0xPolygonMiden/miden-vm.git", branch = "al-migrate-new-padding-rule", default-features = false }
+miden-processor = { git = "https://github.com/0xPolygonMiden/miden-vm.git", branch = "next" }
+miden-stdlib = { git = "https://github.com/0xPolygonMiden/miden-vm.git", branch = "next", default-features = false }
 miden-tx = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "andrew-migrate-to-new-padding-rule" }
 prost = { version = "0.13" }
 thiserror = { version = "1.0" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ resolver = "2"
 
 [workspace.package]
 edition = "2021"
-rust-version = "1.80"
+rust-version = "1.82"
 version = "0.6.0"
 license = "MIT"
 authors = ["Miden contributors"]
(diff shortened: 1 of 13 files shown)


bin/faucet/src/client.rs
```diff
@@ -197,6 +200,9 @@ impl FaucetDataStore {
     }
 }
 
+unsafe impl Send for FaucetDataStore {}
+unsafe impl Sync for FaucetDataStore {}
```
Fumuran, line 204: I've implemented these two traits for the `FaucetDataStore` as it was done for other structures which were implementing the `DataStore` trait (for example, [here](https://github.com/0xPolygonMiden/miden-base/blob/ea1de499461aa27a951309bd7a725dc7ddc631dc/miden-tx/src/testing/tx_context/mod.rs#L124)), but I'm not sure whether it is needed here. With these implementations we remove one of two warnings about using a structure which is not implementing these traits in `Arc`, but we allow this warning anyway.

Mirko-von-Leipzig: These have always bothered me but I don't know the history/context behind why we need them.

Do you have an understanding of why the `Send` and `Sync` aren't automatically implemented for these types? If so we should definitely add `/// SAFETY` comments explaining this.

Ideally these shouldn't be required at all. And I'm confused why implementing another trait causes these auto-traits to disappear :D

Cargo.toml
```diff
@@ -15,7 +15,7 @@ resolver = "2"
 
 [workspace.package]
 edition = "2021"
-rust-version = "1.80"
+rust-version = "1.82"
```
Mirko-von-Leipzig, line 18: What's causing this bump?

Makefile
```diff
@@ -13,7 +13,7 @@ BUILD_PROTO=BUILD_PROTO=1
 
 .PHONY: clippy
 clippy: ## Runs Clippy with configs
-	cargo clippy --locked --workspace --all-targets --all-features -- -D warnings
+	cargo clippy --locked --workspace --all-targets --all-features -- -D warnings --allow clippy::arc_with_non_send_sync
```
Mirko-von-Leipzig, line 16: This seems really unsafe 😅 But I don't understand these unsafe impl context's so I'll withhold judgement until its been explained.

I would suggest manually `#[allow(arc_with_non_send_sync)]` on the actual instances instead of blanket allowing this.

crates/block-producer/src/block_builder/prover/asm/block_kernel.masm
Mirko-von-Leipzig, line 1: Can't comment on this; `masm` is not in my wheel house.

bin/faucet/src/client.rs
polydez: While this is working here due to serial nature of our faucet web app, I think, the right solution would be to update storage itself in order to implement Send/Sync automatically: https://github.com/0xPolygonMiden/miden-node/pull/517/files#diff-1d5b7b3cf77eb4b236ed8fec8e25c2ea541e3591c719f9b45fae328c4e1ada26R13-R18

If `Sync` is also required for `FaucetDataStore`, we will probably need to change `RwLock` to `Mutex`.

Mirko-von-Leipzig: > While this will work here due to a serial nature of our faucet web app, I do think, the proper solution would be to update storage itself in order to implement `Send` automatically: https://github.com/0xPolygonMiden/miden-node/pull/517/files#diff-1d5b7b3cf77eb4b236ed8fec8e25c2ea541e3591c719f9b45fae328c4e1ada26R13-R18
> 
> If `Sync` is also required for `FaucetDataStore`, we will probably need to change `RwLock` to `Mutex`.

This sounds good to me; and I think we may as well make it `Sync`. I don't think it will impact the faucet throughput.

Mirko-von-Leipzig: What prevents storage from implementing `Send` automatically? And `RwLock` should be `Sync` so I'm not following 🥲

polydez: Maybe there are borrowed references which require it to implement `Sync` in order to be `Send`, but `RefCell` doesn't implement `Sync`. `RwLock` implements `Sync` only it its content implement `Sync` as well, so it's likely that `Arc<RwLock<...>>` should be enough here.

polydez: You can see differences in `Send`/`Sync` implementation for mentioned structs here:

RefCell: https://doc.rust-lang.org/src/core/cell.rs.html#1280-1284
RwLock: https://doc.rust-lang.org/src/std/sync/rwlock.rs.html#89-92
Mutex: https://doc.rust-lang.org/src/std/sync/mutex.rs.html#186-189

Makefile
polydez: No, it's safe, it's just a suppression of annoying clippy warnings introduced in the latest Rust versions. :) It warns you, if you use `Arc` where it is sufficient to use just `Rc`. I think, we should update our code and switch this lint on, but in separate PR.

bin/faucet/src/client.rs
```diff
@@ -69,11 +70,12 @@ impl FaucetClient {
...
         );
-        let authenticator = BasicAuthenticator::<StdRng>::new(&[(
+        let authenticator = Arc::new(BasicAuthenticator::<StdRng>::new(&[(
             secret.public_key().into(),
             AuthSecretKey::RpoFalcon512(secret),
-        )]);
-        let executor = TransactionExecutor::new(data_store.clone(), Some(Rc::new(authenticator)));
+        )])) as Arc<dyn TransactionAuthenticator>;
+        let executor =
+            TransactionExecutor::new(Arc::new(data_store.clone()), Some(authenticator.clone()));
```
polydez, line 78: Nit: here is better to refactor a bit (return `Arc<FaucetDataStore>` from the function and just clone `Arc`'s, not data store itself.

polydez (changes requested):
Overall looks good to me, thanks! The only concern I have is stepping into `unsafe` field where it's not really needed.

Cargo.toml
Fumuran: `miden-crypto` now require rust 1.82

bin/faucet/src/client.rs
Fumuran: But we need to provide the `FaucetDataStore` to the constructor at the end of the function, so would it be efficient to wrap the  `FaucetDataStore` into `Arc` just to provide it to the `TransactionExecutor` and then somehow unwrap it back (with `into_inner`?) to provide to the constructor? I understand the downsides of cloning the actual value instead of `Arc`, but here it feels like it brings more complexity, no?

polydez: No-no, we can't unwrap it from `Arc` after cloning. I meant, we can refactor this in order to use `Arc<FaucetDataStore>` everywhere. :)

polydez: But okay, just leave `TODO` here, I'm refactoring faucet and will do this by myself later.

bin/faucet/src/client.rs
```diff
@@ -45,7 +46,7 @@ pub const DISTRIBUTE_FUNGIBLE_ASSET_SCRIPT: &str =
 /// for the faucet.
 pub struct FaucetClient {
     rpc_api: ApiClient<Channel>,
-    executor: TransactionExecutor<FaucetDataStore, BasicAuthenticator<StdRng>>,
+    executor: TransactionExecutor,
     data_store: FaucetDataStore,
```
bobbinth, line 50: I think we can make `data_store` here `Arc<FaucetDataStore>` and avoid all the extra cloning discussed in https://github.com/0xPolygonMiden/miden-node/pull/528/files#r1816146982.

This will require changing `FaucetDataStore::update_faucet_account()` to take `self` as immutable, but that shouldn't be a problem since we store `faucet_account` as `RefCell` anyway.

Also, on line 181 below, we don't need to wrap `faucet_account` in `Rc` (just `RefCell` should work fine).

bin/faucet/src/client.rs
bobbinth: Using `Mutex` for the `faucet_account` should fix this.

Makefile
bobbinth: Let's create an issue to remove this once https://github.com/0xPolygonMiden/miden-base/issues/909 is fixed.

bobbinth (commented):
Looks good! Thank you! I left a few comments inline.

bobbinth (approved):
Actually, I pushed a commit which should address https://github.com/0xPolygonMiden/miden-node/pull/528#discussion_r1817394386, https://github.com/0xPolygonMiden/miden-node/pull/528#discussion_r1816146982, and  https://github.com/0xPolygonMiden/miden-node/pull/528#discussion_r1815812815.

@Mirko-von-Leipzig, @polydez - take a look and see if this is good to merge.

Mirko-von-Leipzig (approved):
@bobbinth LGTM thank you.

bin/faucet/src/client.rs
```diff
@@ -175,23 +178,24 @@ pub struct FaucetDataStore {
...
         }
     }
 
     /// Updates the stored faucet account with the provided delta.
-    fn update_faucet_account(&mut self, delta: &AccountDelta) -> Result<(), ProcessError> {
+    fn update_faucet_account(&self, delta: &AccountDelta) -> Result<(), ProcessError> {
         self.faucet_account
-            .borrow_mut()
+            .lock()
+            .unwrap()
```
polydez, line 198: I think we should avoid using `unwrap` in favor of `expect` everywhere except tests if panic is highly unlikely to happen (otherwise, an error should be returned). This helps us to understand why method might fail or why we don't expect it to fail. I usually add `.expect("Poisoned lock")` for every `lock` or `read`/`write` in case of using `RwLock`.

bin/faucet/src/client.rs
```diff
@@ -60,20 +63,20 @@ impl FaucetClient {
             initialize_faucet_client(config.clone()).await?;
 
         let (faucet_account, account_seed, secret) = build_account(config.clone())?;
-        let faucet_account = Rc::new(RefCell::new(faucet_account));
-        let id = faucet_account.borrow().id();
+        let faucet_account = faucet_account;
```
polydez, line 66: This line seems to be useless.

bobbinth:
@Fumuran - could you fix up the remaining comments in this PR?

After this, let's merge https://github.com/0xPolygonMiden/miden-base/pull/929, and then update dependencies and merge this PR as well.

bobbinth:
https://github.com/0xPolygonMiden/miden-base/pull/929 has now been merged. Let's update the `Cargo.toml` file, fix the remaining `unwrap()` [here](https://github.com/0xPolygonMiden/miden-node/blob/997b2e77fd645e82eb23e4ab872ab576de91f8fa/bin/faucet/src/client.rs#L210), and then merge.

PR #530, opened 2024-10-25
feat(block-producer): arc transaction data

Mirko-von-Leipzig:
This PR reduces deep cloning of transaction data in the new block-producer by wrapping the core data in `Arc`.

In addition, this PR also changes the dependency graphs key bound from `Clone -> Copy`. This isn't strictly necessary, but does improve code readability as discussed [here](https://github.com/0xPolygonMiden/miden-node/pull/525#discussion_r1815699031). Since our existing keys are `Copy` this doesn't actually have a further impact.

### DIFF
--- a/crates/block-producer/src/batch_builder/mod.rs
+++ b/crates/block-producer/src/batch_builder/mod.rs
@@ -9,6 +9,7 @@ use std::{
 
 use miden_objects::{
     accounts::AccountId,
+    assembly::SourceManager,
     notes::NoteId,
     transaction::{OutputNote, TransactionId},
     Digest,
@@ -284,8 +285,13 @@ impl WorkerPool {
             async move {
                 tracing::debug!("Begin proving batch.");
 
-                let transactions =
-                    transactions.into_iter().map(AuthenticatedTransaction::into_raw).collect();
+                // TODO: This is a deep clone which can be avoided by change batch building to using
+                // refs or arcs.
+                let transactions = transactions
+                    .iter()
+                    .map(AuthenticatedTransaction::raw_proven_transaction)
+                    .cloned()
+                    .collect();
 
                 tokio::time::sleep(simulated_proof_time).await;
                 if failed {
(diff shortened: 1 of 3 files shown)


crates/block-producer/src/domain/transaction.rs
```diff
@@ -17,7 +17,7 @@ use crate::{errors::VerifyTxError, mempool::BlockNumber, store::TransactionInput
 /// Note that this is of course only valid for the chain height of the authentication.
 #[derive(Clone, Debug, PartialEq)]
 pub struct AuthenticatedTransaction {
-    inner: ProvenTransaction,
+    inner: Arc<ProvenTransaction>,
```
bobbinth, line 20: Should we mention in doc-comments that cloning this struct is relatively cheap?

crates/block-producer/src/domain/transaction.rs
```diff
@@ -117,18 +117,24 @@ impl AuthenticatedTransaction {
     //! Builder methods intended for easier test setup.
 
     /// Short-hand for `Self::new` where the input's are setup to match the transaction's initial
-    /// account state.
+    /// account state. This covers the account's initial state and nullifiers beting set to unspent.
```
bobbinth, line 120: typo: should `beting` be `being`?

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -284,8 +285,13 @@ impl WorkerPool {
...
 
-                let transactions =
-                    transactions.into_iter().map(AuthenticatedTransaction::into_raw).collect();
+                // TODO: This is a deep clone which can be avoided by change batch building to using
+                // refs or arcs.
+                let transactions = transactions
+                    .iter()
+                    .map(AuthenticatedTransaction::raw_proven_transaction)
+                    .cloned()
+                    .collect();
```
bobbinth, line 294: Let's create an issue for this.

bobbinth (approved):
Looks good! Thank you! I left a couple of small comments inline. After these, we should be good to merge.

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: #531

PR #532, opened 2024-10-29
feat(node): avoid cloning in-memory data structures on `apply_block`

polydez:
Resolves https://github.com/0xPolygonMiden/miden-node/issues/149

In this PR we utilize new `compute_mutations` and `apply_mutations` methods in `Smt` in order to get rid of cloning in-memory account and nullifier trees. We also don't clone chain MMR anymore, because it wasn't necessary for adding block hash. Also we put additional checks before applying changes and reorder applying sequence due to possible inconsistent state on applying error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Changed the `BlockWitness` to pass the inputs to the VM using only advice provider (#516).
 - [BREAKING] Improved store API errors (return "not found" instead of "internal error" status if requested account(s) not found) (#518).
 - [BREAKING] Migrated to v0.11 version of Miden VM (#528).
+- Got rid of cloning in-memory structures on `apply_block` (#532).
 
 ## 0.5.1 (2024-09-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 - Changed the `BlockWitness` to pass the inputs to the VM using only advice provider (#516).
 - [BREAKING] Improved store API errors (return "not found" instead of "internal error" status if requested account(s) not found) (#518).
 - [BREAKING] Migrated to v0.11 version of Miden VM (#528).
-- Got rid of cloning in-memory structures on `apply_block` (#532).
+- Reduce cloning in the store's `apply_block` (#532).
 
 ## 0.5.1 (2024-09-12)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1653,7 +1653,6 @@ dependencies = [
  "directories",
  "figment",
  "hex",
- "miden-crypto",
  "miden-lib",
  "miden-node-proto",
  "miden-node-utils",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1435,31 +1435,13 @@ dependencies = [
  "lock_api",
  "loom",
  "memchr",
- "miden-crypto 0.11.0",
+ "miden-crypto",
  "miden-formatting",
  "miden-miette",
  "miden-thiserror",
  "num-derive",
  "num-traits",
  "parking_lot",
- "winter-math",
- "winter-utils",
-]
-
-[[package]]
-name = "miden-crypto"
-version = "0.10.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "825fc5d2e4c951f45da54da2d0d23071918d966133f85dbc219694e0b9a1e141"
-dependencies = [
- "blake3",
- "glob",
- "num",
- "num-complex",
- "rand",
- "rand_core",
- "sha3",
- "winter-crypto",
  "winter-math",
  "winter-utils",
 ]
@@ -1671,7 +1653,7 @@ dependencies = [
  "directories",
  "figment",
  "hex",
- "miden-crypto 0.10.3",
+ "miden-crypto",
  "miden-lib",
  "miden-node-proto",
  "miden-node-utils",
@@ -1722,7 +1704,7 @@ source = "git+https://github.com/0xPolygonMiden/miden-base.git?branch=next#b624d
 dependencies = [
  "miden-assembly",
  "miden-core",
- "miden-crypto 0.11.0",
+ "miden-crypto",
  "miden-processor",
  "miden-verifier",
  "rand",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1435,13 +1435,31 @@ dependencies = [
  "lock_api",
  "loom",
  "memchr",
- "miden-crypto",
+ "miden-crypto 0.11.0",
  "miden-formatting",
  "miden-miette",
  "miden-thiserror",
  "num-derive",
  "num-traits",
  "parking_lot",
+ "winter-math",
+ "winter-utils",
+]
+
+[[package]]
+name = "miden-crypto"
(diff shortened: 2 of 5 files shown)


Cargo.lock
Mirko-von-Leipzig, line 1: I believe you should be using `miden-crypto v0.11.0` to match the one used by `miden-core`. Otherwise we have two versions of it floating about.

crates/store/Cargo.toml
```diff
@@ -16,6 +16,7 @@ deadpool-sqlite = { version = "0.8", features = ["rt_tokio_1"] }
 directories = { version = "5.0" }
 figment = { version = "0.10", features = ["toml", "env"] }
 hex = { version = "0.4" }
+miden-crypto = { version = "0.10", default-features = false }
```
Mirko-von-Leipzig, line 19: See my comment in the lock file.

```suggestion
miden-crypto = { version = "0.11", default-features = false }
```

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Changed the `BlockWitness` to pass the inputs to the VM using only advice provider (#516).
 - [BREAKING] Improved store API errors (return "not found" instead of "internal error" status if requested account(s) not found) (#518).
 - [BREAKING] Migrated to v0.11 version of Miden VM (#528).
+- Got rid of cloning in-memory structures on `apply_block` (#532).
```
Mirko-von-Leipzig, line 18:
```suggestion
- Reduce cloning in the store's `apply_block` (#532).
```

Cargo.lock
polydez: Yes, my bad, I looked into incorrect version of `miden-base` repo while deciding which version of `miden-crypto` to use. :)

Mirko-von-Leipzig (approved):

crates/store/src/state.rs
```diff
@@ -203,15 +203,15 @@ impl State {
...
         // finalized blocks. So we should check for the latest block when getting block from
         // the store.
-        let store = self.block_store.clone();
+        let store = Arc::clone(&self.block_store);
         let block_save_task =
             tokio::spawn(async move { store.save_block(block_num, &block_data).await });
 
         // scope to read in-memory data, validate the request, and compute intermediary values
-        let (account_tree, chain_mmr, nullifier_tree, notes) = {
+        let (account_tree_update, nullifier_tree_update, account_tree_old_root) = {
```
bobbinth, line 211: nit: the order of variables is a bit odd here. I'd probably do this as:
```rust
let (account_tree_old_root, account_tree_update, nullifier_tree_update) = {
```
Also, I'd probably update the comment to be something like: "scope to read in-memory data, compute mutations required for updating account and nullifier trees, and validate the request"

crates/store/src/state.rs
```diff
@@ -332,6 +316,13 @@ impl State {
             // successfully.
             let mut inner = self.inner.write().await;
 
+            // We need to check that the account tree root hasn't changed while we were waiting for
+            // the DB update task to complete. If it has changed, we mustn't proceed with in-memory
+            // updates, since it might lead to inconsistent state.
+            if inner.account_tree.root() != account_tree_old_root {
+                return Err(ApplyBlockError::ConcurrentWrite);
+            }
```
bobbinth, line 324: Why do we need this check here? If the account tree has changed somehow, wouldn't we get an error on line 344 below? Or more specifically: which scenario are we trying to prevent with this?

If we do need this check, why are we checking only the account tree and not nullifier tree as well?

crates/store/src/nullifier_tree.rs
```diff
@@ -1,7 +1,8 @@
+use miden_crypto::merkle::SMT_DEPTH;
```
bobbinth, line 1: We should be able to import this from `miden_objects::crypto::merkle` on line 5 below.

crates/store/Cargo.toml
```diff
@@ -16,6 +16,7 @@ deadpool-sqlite = { version = "0.8", features = ["rt_tokio_1"] }
 directories = { version = "5.0" }
 figment = { version = "0.10", features = ["toml", "env"] }
 hex = { version = "0.4" }
+miden-crypto = { version = "0.11", default-features = false }
```
bobbinth, line 19: Related to the last comment, we probably don't need to introduce this dependency.

bobbinth (commented):
Looks good! Thank you! I left a few comments and a question inline.

crates/store/src/state.rs
polydez: The motivation for adding this check was in the possible situation, when nullifier tree was successfully updated, but update of accounts tree raised an error, this could lead to inconsistent in-memory state. Additional check before updating of nullifier tree can prevent this.

bobbinth: Makes sense. Could we move this check closer to the update site (e.g., to line 336 below) - or is there a reason it needs to be here?

Also, I'd expand the comment to explain the motivation more (like you've done in the comment). For example, there seems to be a specific order in which we need to do updates: first nullifier tree, then account tree, and lastly MMR - so, we should add a comment that this order is meaningful and shouldn't be changed.

polydez: > Could we move this check closer to the update site (e.g., to line 336 below) - or is there a reason it needs to be here?

In this case we would allow database to update, but fail to update in-memory state. This is also not a good choice. By the way, we should add checking for nullifier tree root here as well in order to prevent this inconsistent between in-memory and db state. I will add it now.

polydez: Done!

crates/store/src/state.rs
```diff
@@ -203,15 +203,21 @@ impl State {
...
-        // scope to read in-memory data, validate the request, and compute intermediary values
-        let (account_tree, chain_mmr, nullifier_tree, notes) = {
+        // scope to read in-memory data, compute mutations required for updating account
+        // and nullifier trees, and validate the request
+        let (
+            nullifier_tree_old_root,
+            account_tree_old_root,
+            account_tree_update,
+            nullifier_tree_update,
+        ) = {
```
bobbinth, line 217: nit: this arrangement looks a bit odd again. Maybe something like this will work better:
```rust
let (
    account_tree_old_root,
    account_tree_update,
    nullifier_tree_old_root,
    nullifier_tree_update,
) = {
```

crates/store/src/state.rs
```diff
@@ -332,6 +327,16 @@ impl State {
             // successfully.
             let mut inner = self.inner.write().await;
 
+            // We need to check that neither the nullifier tree root, nor the account tree root
+            // haven't changed while we were waiting for the DB preparation task to complete.
+            // If it has changed, we mustn't proceed with both in-memory and database updates,
+            // since it might lead to inconsistent state.
```
bobbinth, line 333: nit: I would write this as:
```rust
// We need to check that neither the nullifier tree nor the account tree have changed
// while we were waiting for the DB preparation task to complete. If either of them 
// did changed, we do not proceed with in-memory and database updates, since it may
// lead to an inconsistent state.
```

bobbinth (approved):
Looks good! Thank you! I left a couple of minor comments inline. Once these are addressed, we can merge.

bobbinth (approved):
All looks good! Thank you! Once CI is green, we are good to merge.

PR #533, opened 2024-10-29
feat(block-producer): refactor batch graph to use dependency graph

Mirko-von-Leipzig:
This PR ports `BatchGraph` to use `DependencyGraph` as its foundation.

I implemented this by extending `DependencyGraph` to have a new `pending` stage, where the node exists but does not yet have an associated value. For batches this value is its proof i.e. `TransactionBatch`.

Effectively nodes now have a pending state, and must be promoted by passing in a value for it. I'm not 100% happy with the naming of things. Pending, promotions, processed, committed.

This model does not quite match the `TransactionGraph` lifecycle, however this was easy to adjust by having `TransacionGraph::insert` invoke both actions.

The main benefit of this is reducing the test surface.

### DIFF
--- a/crates/block-producer/src/mempool/batch_graph.rs
+++ b/crates/block-producer/src/mempool/batch_graph.rs
@@ -14,9 +14,9 @@ use crate::batch_builder::batch::TransactionBatch;
 
 /// Tracks the dependencies between batches, transactions and their parents.
 ///
-/// Batches are inserted with their transaction, and parent transaction sets which form the edges of
-/// the dependency graph. Batches are initially inserted in a pending state while we wait on their
-/// proofs to be generated. The dependencies are still tracked in this state.
+/// Batches are inserted with their transaction and parent transaction sets which defines the edges
+/// of the dependency graph. Batches are initially inserted in a pending state while we wait on
+/// their proofs to be generated. The dependencies are still tracked in this state.
 ///
 /// Batches can then be promoted to ready by [submitting their proofs](Self::submit_proof) once
 /// available. Proven batches are considered for inclusion in blocks once _all_ parent batches have
@@ -135,11 +135,14 @@ impl BatchGraph {
         &mut self,
         batches: BTreeSet<BatchJobId>,
     ) -> Result<BTreeMap<BatchJobId, Vec<TransactionId>>, GraphError<BatchJobId>> {
+        // This returns all descendent batches as well.
         let batches = self.inner.purge_subgraphs(batches)?;
 
+        // SAFETY: These batches must all have been inserted since they are emitted from the inner
+        // dependency graph, and therefore must all be in the batches mapping.
         let batches = batches
             .into_iter()
-            .map(|batch| (batch, self.batches.remove(&batch).expect("Malformed graph")))
+            .map(|batch| (batch, self.batches.remove(&batch).unwrap()))
             .collect::<BTreeMap<_, _>>();
 
         for tx in batches.values().flatten() {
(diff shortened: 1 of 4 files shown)


crates/block-producer/src/mempool/mod.rs
```diff
@@ -231,6 +237,11 @@ impl Mempool {
             .expect("Transaction graph is malformed");
 
         // Rollback state.
+        let transactions = transactions
+            .into_iter()
+            // FIXME
+            .map(|tx_id| todo!("Inflight state should remember diffs"))
+            .collect::<Vec<_>>();
```
Mirko-von-Leipzig, line 244: I'm intending to change `InflightState` to track transaction ID -> state delta internally in a follow-up PR.

crates/block-producer/src/mempool/dependency_graph.rs
```diff
@@ -280,7 +307,17 @@ impl<K: Ord + Copy, V: Clone> DependencyGraph<K, V> {
     ///
     /// This method assumes the node exists. Caller is responsible for ensuring this is true.
     fn try_make_root(&mut self, key: K) {
-        debug_assert!(self.vertices.contains_key(&key), "Potential root must exist in the graph");
+        if self.pending.contains(&key) {
+            return;
+        }
```
bobbinth, line 312: Why is it not an error to try to make a pending node into a root?

bobbinth (approved):
Looks good! Thank you! I left some comments inline - but most are minor.

crates/block-producer/src/mempool/batch_graph.rs
```diff
@@ -3,212 +3,307 @@ use std::collections::{BTreeMap, BTreeSet};
...
+///               select_block│          │                
+///                     ┌─────▼─────┐    │                
+///                     │ committed ┼────┤                
+///                     └─────┬─────┘    │                
+///                           │          │                
+///            prune_committed│          │purge_subgraphs
+///                     ┌─────▼─────┐    │                
+///                     │  <null>   ◄────┘                
+///                     └───────────┘                     
+/// ```
```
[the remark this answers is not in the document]
bobbinth: I think this is helpful. So, if not difficult, I'd add it in other relevant places too.

crates/block-producer/src/mempool/dependency_graph.rs
Mirko-von-Leipzig: This method is private and called from a few locations. Effectively either callers have to perform this check, or this method could enforce it. I thought it safer to do it here in one place.

More specifically, this is called when

1. A pending node is promoted (so this check isn't necessary)
2. A root node is processed, so we check all of its children. We would have to filter children here.
3. Subgraphs are reverted, where we have to check and filter a bunch of nodes

crates/block-producer/src/mempool/dependency_graph.rs
```diff
@@ -9,8 +12,15 @@ use miden_tx::utils::collections::KvMap;
...
 ///
 /// Forms the basis of our transaction and batch dependency graphs.
+///
+/// # Node lifecycle
+/// ```
+/// ```
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct DependencyGraph<K, V> {
+    /// Node's who's data is still pending.
+    pending: BTreeSet<K>,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Agreed. I think a final pass after the work is done to align the naming would be a good idea. I'll add the lifecycle diagram at least for now.

PR #534, opened 2024-10-31
feat(block-producer): inflight state custody of transaction data

Mirko-von-Leipzig:
This PR upgrades `InflightState` so that it tracks each transaction's state delta internally. We differentiate between committed and uncommitted transactions. The former cannot be reverted.

This means reverting or committing transactions now only requires passing in the transaction IDs, instead of requiring the full transaction data.

This enables making the inflight state more robust as it now differentiates between committed and uncommitted transaction IDs. Additional work is required to change the current panics into errors instead, though this is left to a future PR as it isn't high priority. Tracked in #537.

Essentially this works towards hardening the different components of the mempool individually so that we can pick up corruption/bugs as soon as possible.

### DIFF
--- a/crates/block-producer/src/mempool/inflight_state/mod.rs
+++ b/crates/block-producer/src/mempool/inflight_state/mod.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeMap, BTreeSet, VecDeque},
+    collections::{btree_map::Entry, BTreeMap, BTreeSet, VecDeque},
     sync::Arc,
 };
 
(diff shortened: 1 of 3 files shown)


crates/block-producer/src/mempool/inflight_state/mod.rs
```diff
@@ -61,35 +64,19 @@ pub struct InflightState {
...
-
-    /// The nullifiers consumed by the transactions.
+#[derive(Clone, Debug, PartialEq)]
+struct Delta {
+    account: AccountId,
     nullifiers: BTreeSet<Nullifier>,
-
-    /// The notes produced by the transactions.
     output_notes: BTreeSet<NoteId>,
 }
```
bobbinth, line 72: nit: I'd add brief doc comments for the struct and its fields.

crates/block-producer/src/mempool/inflight_state/mod.rs
```diff
@@ -61,35 +64,19 @@ pub struct InflightState {
...
-        let mut output_notes = BTreeSet::new();
-
-        for tx in txs {
-            *account_transactions.entry(tx.account_id()).or_default() += 1;
-            nullifiers.extend(tx.nullifiers());
-            output_notes.extend(tx.output_notes());
-        }
-
+impl Delta {
+    fn new(tx: &AuthenticatedTransaction) -> Self {
```
bobbinth, line 75: nit: maybe this should be `From<&AuthenticatedTransaction> for Delta`?

bobbinth (approved):
Looks good! Thank you! I left a couple of comments inline, but they are all minor nits.

crates/block-producer/src/mempool/inflight_state/mod.rs
Mirko-von-Leipzig: I'm not a fan of `From/Into` for references tbh, they're pretty hard to discover naturally.

PR #535, opened 2024-10-31
refactor(node): improve `ApplyBlockError` variants in the store

polydez:
Resolves https://github.com/0xPolygonMiden/miden-node/issues/214

In this PR we improve `ApplyBlockError` variants in the store. Now store API returns "invalid argument" for errors which signal on incorrect/malformed blocks. Sorted variants, shortened some names, some useless error granularity was substituted by just `String`. The most of errors in `ApplyBlock` endpoint were swallowed by unclear reason, now they will be thrown.

### DIFF
--- a/crates/store/src/errors.rs
+++ b/crates/store/src/errors.rs
@@ -169,28 +169,8 @@ pub enum GenesisError {
 
 // ENDPOINT ERRORS
 // =================================================================================================
-
 #[derive(Error, Debug)]
-pub enum ApplyBlockError {
-    // ERRORS WITH AUTOMATIC CONVERSIONS FROM NESTED ERROR TYPES
-    // ---------------------------------------------------------------------------------------------
-    #[error("Database error: {0}")]
-    DatabaseError(#[from] DatabaseError),
-    #[error("I/O error: {0}")]
-    IoError(#[from] io::Error),
-    #[error("Task join error: {0}")]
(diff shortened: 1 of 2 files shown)


crates/store/src/errors.rs
```diff
@@ -161,42 +172,61 @@ pub enum GenesisError {
...
+            | ApplyBlockError::NewBlockInvalidAccountRoot
+            | ApplyBlockError::NewBlockInvalidNoteRoot
+            | ApplyBlockError::NewBlockInvalidNullifierRoot
+            | ApplyBlockError::DuplicatedNullifiers(_)
+            | ApplyBlockError::InvalidOutputNoteType(_)
+            | ApplyBlockError::InvalidTxHash { .. } => Status::invalid_argument(err.to_string()),
+
+            _ => Status::internal(err.to_string()),
+        }
+    }
```
bobbinth, line 229: This works, but I wonder if we should create a dedicated enum for these errors. For example, something like `InvalidBlockError`. Then, we could add `ApplyBlockError::InvalidBlockError` variant. I think the separation may be a bit cleaner.

bobbinth (commented):
Looks good! Thank you! I left one comment inline about potentially encapsulating errors which are due to invalid block into a separate enum.

bobbinth (approved):
Looks good! Thank you!

PR #538, opened 2024-11-02
feat(block-producer): merge next

Mirko-von-Leipzig:
Merge `next` into `next-block-producer` in anticipation of touching more common code.

To date the `block-producer` rework has been done in disjoint files from `next`, meaning the parallel branch has had close to zero merge conflicts. The next steps involve swopping out the `next` block-producer for the one on `next-block-producer` which likely means merge conflicts will occur more commonly.

This PR pre-emptively merges `next` in order to minimize merge conflicts going forward.

This merge had no conflicts; just an auto-merge aka it should be safe as is.

I'd suggest auditing the `block-producer` crate's diff specifically and the rest can be ignored. Within that crate the changes are also in locations not touched by the rework and should be safe.


Makefile
```diff
@@ -13,7 +13,7 @@ BUILD_PROTO=BUILD_PROTO=1
 
 .PHONY: clippy
 clippy: ## Runs Clippy with configs
-	cargo clippy --locked --workspace --all-targets --all-features -- -D warnings
+	cargo clippy --locked --workspace --all-targets --all-features -- -D warnings --allow clippy::arc_with_non_send_sync
```
bobbinth, line 16: If we update `Cargo.lock` file (i.e., via `cargo update`), this change would no longer be necessary.

CHANGELOG.md
```diff
@@ -2,14 +2,21 @@
 
 ## v0.6.0 (TBD)
 
+- Added `AccountCode` as part of `GetAccountProofs` endpoint response (#521).
```
bobbinth, line 5: nit: we should move this to the end of the list as we usually list changes with earliest on top.

bobbinth (approved):

bobbinth:
Actually, I pushed a small commit that addresses my comments. Feel free to revert it if you think that's not the right place for it.

CHANGELOG.md
Mirko-von-Leipzig: Hmm. This is a bit concerning tbh - I would have thought post-merge the result should match `next` exactly. However even with your commit there are still changes between this branch and `next's changelog. The ordering is different:

`next`:
```
## v0.6.0 (TBD)

- Added `AccountCode` as part of `GetAccountProofs` endpoint response (#521).
- [BREAKING] Added `kernel_root` to block header's protobuf message definitions (#496).
- [BREAKING] Renamed `off-chain` and `on-chain` to `private` and `public` respectively for the account storage modes (#489).
- Optimized state synchronizations by removing unnecessary fetching and parsing of note details (#462).
- [BREAKING] Introduced additional limits for note/account number (#503).
- [BREAKING] Removed support for basic wallets in genesis creation (#510).
- Added `GetAccountProofs` endpoint (#506).
- Migrated faucet from actix-web to axum (#511).
- Changed the `BlockWitness` to pass the inputs to the VM using only advice provider (#516).
- [BREAKING] Improved store API errors (return "not found" instead of "internal error" status if requested account(s) not found) (#518).
- [BREAKING] Migrated to v0.11 version of Miden VM (#528).
- Reduce cloning in the store's `apply_block` (#532).
- [BREAKING] Changed faucet storage type in the genesis to public. Using faucet from the genesis for faucet web app. Added support for faucet restarting without blockchain restarting (#517).
```

this branch:
```
## v0.6.0 (TBD)

- Optimized state synchronizations by removing unnecessary fetching and parsing of note details (#462).
- [BREAKING] Added `kernel_root` to block header's protobuf message definitions (#496).
- [BREAKING] Renamed `off-chain` and `on-chain` to `private` and `public` respectively for the account storage modes (#489).
- [BREAKING] Introduced additional limits for note/account number (#503).
- [BREAKING] Removed support for basic wallets in genesis creation (#510).
- Added `GetAccountProofs` endpoint (#506).
- Migrated faucet from actix-web to axum (#511).
- Changed the `BlockWitness` to pass the inputs to the VM using only advice provider (#516).
- [BREAKING] Changed faucet storage type in the genesis to public. Using faucet from the genesis for faucet web app. Added support for faucet restarting without blockchain restarting (#517).
- [BREAKING] Improved store API errors (return "not found" instead of "internal error" status if requested account(s) not found) (#518).
- Added `AccountCode` as part of `GetAccountProofs` endpoint response (#521).
- [BREAKING] Migrated to v0.11 version of Miden VM (#528).
- Reduce cloning in the store's `apply_block` (#532).
```

Mirko-von-Leipzig: I think I'll try a rebase instead.

bobbinth: I think it was my commit that introduced these changes. Before it, they should have been the same?

Mirko-von-Leipzig: Ah. Right; I confused myself with the github preview of your comment it seems.

Mirko-von-Leipzig:
> Actually, I pushed a small commit that addresses my comments. Feel free to revert it if you think that's not the right place for it.

I think its fine since they're small changes; but in a technical sense probably more correct to change those directly on `next`. Will keep them here though.

PR #539, opened 2024-11-02
chore: re-enable blocks_in_conditions lint

Mirko-von-Leipzig:
Clippy issue https://github.com/rust-lang/rust-clippy/issues/12281 was fixed in v1.81. We now have MSRV of 1.82 allowing us to remove our workaround.

More specifically, `#[async_trait]` was triggering this lint, and as of 1.81 no longer does.

I did grep the repo for `blocks_in_conditions` and can confirm we have no other instances of it.


bobbinth (approved):
Thank you! PRs that add 0 new lines are always nice :)

PR #540, opened 2024-11-03
feat(block-producer): rebase next

Mirko-von-Leipzig:


Mirko-von-Leipzig:
An experiment to see what a rebased branch PR'd back looks like ito diff.

Issue #191, opened 2024-01-26
Batch producer design and features


hackaugusto:
The batch builder is currently two pieces:

1. The trait [`BatchBuilder`](https://github.com/0xPolygonMiden/miden-node/blob/cf361489d8a8305203a0226bfca6440e0fd637a7/block-producer/src/batch_builder/mod.rs#L20-L27)
2. The [`DefaultBatchBuilder`](https://github.com/0xPolygonMiden/miden-node/blob/cf361489d8a8305203a0226bfca6440e0fd637a7/block-producer/src/batch_builder/mod.rs#L109-L127) implementation

The trait has a single method:

```rust
async fn build_batch(&self, txs: Vec<SharedProvenTx>) -> Result<(), BuildBatchError>;
```

Which is called by [`TransactionQueue`](https://github.com/0xPolygonMiden/miden-node/blob/cf361489d8a8305203a0226bfca6440e0fd637a7/block-producer/src/txqueue/mod.rs#L178-L188). The current implementation pushes transactions batches on regular intervals.

Things to discuss and improve on the current design:

- Back pressure from the batch builder. Currently the tx queue pushes batches to the builder on regular intervals without back pressure.
    - Back pressure here has an unconventional reason. Ideally the batch builder will be distributed, meaning the batch proofs are produced by multiple machines, and we could have elastic scaling of the batch provers, such that back pressure would not be necessary.
    - Even though it is possible to implement auto scaling, e.g. every batch spawns a new machine to perform the proving, making back pressure unnecessary, this would not be a good implementation. There are two issues: 1. it takes time for machines to spawn and be ready to receive work 2. it costs money to run machines. Because of the previous points, instead there should be a pool of provers with a conservative auto scaling policy, meaning there are situations the batch provers may lag behind, causing back pressure.
    - The unconventional thing is that back pressure is not completely bad, because transactions have to be pooled to be ranked. So there is a balancing act going on allowing some transactions to be pooled, and providing good batching times.
- As mentioned above, batch proving will be implemented with a cluster of machines. Currently it is unclear if `BatchBuilder` is an abstraction over the cluster, or a single machine in said cluster.

hackaugusto:
> As mentioned above, batch proving will be implemented with a cluster of machines. Currently it is unclear if BlockProducer is an abstraction over the cluster, or a single machine in said cluster.

[...]

Issue #185, opened 2024-01-25
Expand the `TransactionVerifier` trait


hackaugusto:
https://github.com/0xPolygonMiden/miden-node/blob/1462d45aacd052295b0dffa568387b96c8cdc456/block-producer/src/txqueue/mod.rs#L53-L59

Implementations of the above trait are responsible to validate transactions, this has to account for transactions is the store, as well as transactions currently in flight. Namely, when two transactions try to consume the same nullifier, only one can succeed.

The issue with the above traits are the following:

- The trait doesn't expose the complete API necessary to track in-flight transactions. This needs to be done with another trait, currently `ApplyBlock`, to callback this object and signal that an in-flight transaction has been committed to store, allowing the object to clean its internal state. While this works, it makes it harder to follow the code, since the interface is not "self sufficient".
- The trait doesn't have an intermediary state, between this transaction has been added to the queue, to this transaction has been added to a batch. It would be interesting to model such a state, because once fees are implemented, we would like to prioritize transactions not by arrival order, but by amount of fees paid.

hackaugusto:
A sketch of a possible design is:

```rust
#[async_trait]
pub trait TransactionPool: Send + Sync + 'static {
    async fn tx_added(&self, tx: SharedProvenTx) -> Result<(), VerifyTxError>;
    async fn batch_created(&self, tx: Vec<SharedProvenTx>) -> Result<(), VerifyTxError>;
    async fn block_commited(&self, tx: Block) -> Result<(), VerifyTxError>;
}
```

We may want the above as a trait because of https://github.com/0xPolygonMiden/miden-node/issues/23, so we can implement an in-memory and DB implementations (although we sqlite we can have an in-memory DB, so that is not super necessary, and we could have a plain implementation instead).

The design above makes sense to me since the lifecycle of the transaction changes after it touches multiple components. Which is kinda of modeled in the example above.

hackaugusto:
A proposed flow is at https://github.com/0xPolygonMiden/miden-node/issues/191#issuecomment-1911855544

Mirko-von-Leipzig:
Rendered obsolete as the mempool now handles inflight tracking.

Issue #531, opened 2024-10-28
Batch building should use arcs


Mirko-von-Leipzig:
The batch builder in the block-producer rework should take in `Arc<ProvenTransaction>` to prevent cloning them.

https://github.com/0xPolygonMiden/miden-node/pull/530#discussion_r1819509446

Mirko-von-Leipzig:
> I agree. I don't think there is a good reason why `TransactionBatch::new()` cannot take something like `IntoIterator<Item = AsRef<ProvenTransaction>` as input.

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-node/pull/541#discussion_r1828488445_

Mirko-von-Leipzig:
It appears I managed to rebase out these changes somehow. Will need to re-do #544 once #562 lands on `next`.

Mirko-von-Leipzig:
Closed by #570

PR #541, opened 2024-11-03
feat(block-producer): promote redesign

Mirko-von-Leipzig:
This PR replaces the current FIFO block-producer with the mempool/graph based redesign.

I attempted to keep the diff manageable by not refactoring unnecessary parts i.e. I've largely only replaced and then refactored/cleaned up until clippy stopped yelling. This means that the code is not in a normalized state.

The rework attempts to avoid generics and traits where possible. The current implementation had a lot of these to allow for dependency injection and testing, however these are generally leaky abstractions and I'd prefer to avoid them where possible. In that vein, this PR _removes_ a lot of these since they're no longer used. `txqueue` and `state_view` modules are entirely gone.

I've also simply commented out the current block and batch test suites. Some of these no longer apply; some of these will need to be refactored or adjusted.

Note that this compiles and the (included) tests still pass; however this has not yet been run at all. I'll update the status here if I get it to run for a bit, however  I think this PR is valuable even without it working properly. Any (larger) changes will make this PR even more unreviewable imo. And follow-up fixes/improvement PRs should be preferred once the major issues have been identified.

This PR renders #185 and #191 obsolete.

This PR also completes some tasks in #514.


crates/block-producer/src/lib.rs
```diff
@@ -46,3 +31,9 @@ const SERVER_BUILD_BATCH_FREQUENCY: Duration = Duration::from_secs(2);
 
 /// Maximum number of batches per block
 const SERVER_MAX_BATCHES_PER_BLOCK: usize = 4;
+
+/// The number of blocks of committed state that the mempool retains.
+///
+/// This determines the grace period incoming transactions have between fetching their input from
+/// the store and verification in the mempool.
+const SERVER_MEMPOOL_STATE_RETENTION: usize = 2;
```
bobbinth, line 39: This is probably fine, but should we set it to a higher value just in case? (e.g., to something like 4).

crates/block-producer/src/store/mod.rs
```diff
@@ -53,15 +50,6 @@ pub trait Store: ApplyBlock {
...
     ) -> Result<BlockInputs, BlockInputsError>;
-
-    /// Returns note authentication information for the set of specified notes.
-    ///
-    /// If authentication info for a note does not exist in the store, the note is omitted
-    /// from the returned set of notes.
-    async fn get_note_authentication_info(
-        &self,
-        notes: impl Iterator<Item = &NoteId> + Send,
-    ) -> Result<NoteAuthenticationInfo, NotePathsError>;
```
bobbinth, line 64: Is this no longer needed? I thought we'd still need to use it during batch building - no?

crates/block-producer/src/store/mod.rs
bobbinth, line 1: Probably not for this PR, but we should combine `Store` and `ApplyBlock` traits into a single trait (or at least I don't see a good reason for keeping them separate).

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -19,61 +14,89 @@ use tracing::{debug, info, instrument};
...
 
-// BLOCK BUILDER
+// BLOCK PROVER
 // =================================================================================================
 
-#[async_trait]
-pub trait BlockBuilder: Send + Sync + 'static {
-    /// Receive batches to be included in a block. An empty vector indicates that no batches were
-    /// ready, and that an empty block should be created.
+pub struct BlockProver {
```
bobbinth, line 33: Why not call this `BlockBuilder`? In my mind, `BlockProver` would be a service that a block builder would use to prove blocks.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -19,61 +14,89 @@ use tracing::{debug, info, instrument};
...
+            simulated_proof_time: Duration::ZERO..Duration::ZERO,
+            failure_rate: 0.0,
+            block_kernel: BlockProverKernel::new(),
             store,
-            state_view,
-            block_kernel: BlockProver::new(),
         }
     }
-}
+    /// Starts the [BlockProducer], infinitely producing blocks at the configured interval.
```
bobbinth, line 57: nit: is the reference to `[BlockProducer]` here correct?

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -1,243 +1,111 @@
...
-pub struct DefaultBatchBuilder<S, BB> {
-    store: Arc<S>,
-
-    block_builder: Arc<BB>,
-
-    options: DefaultBatchBuilderOptions,
-
-    /// Batches ready to be included in a block
-    ready_batches: SharedRwVec<TransactionBatch>,
+pub struct BatchProver {
```
bobbinth, line 27: Similar question as in one of the previous comments: why not name this `BatchBuilder`? "Batch building" in my mind is a more general activity with "batch proving" as one of its subtasks.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -110,6 +99,25 @@ pub struct Mempool {
 }
 
 impl Mempool {
+    /// Creates a new [Mempool] with the provided configuration.
+    pub fn new(
+        chain_tip: BlockNumber,
+        batch_limit: usize,
+        block_limit: usize,
+        state_retention: usize,
```
bobbinth, line 107: I wonder if we should have something like `MempoolConfig` or `MempoolOptions` struct to encapsulate these. But this can also be done later on.

crates/block-producer/src/mempool/dependency_graph.rs
bobbinth, line 1: Not for this PR, but I'd probably split the tests out of this file and have `src/mempool/dependency_graph/mod.rs` and `src/mempool/dependency_graph/tests.rs`.

crates/block-producer/src/server/mod.rs
```diff
@@ -1,62 +1,45 @@
...
 /// but not yet actively responding to requests.
 ///
 /// Separating the connection binding from the server spawning allows the caller to connect other
 /// components to the store without resorting to sleeps or other mechanisms to spawn dependent
 /// components.
 pub struct BlockProducer {
-    api_service: api_server::ApiServer<Api>,
-    listener: TcpListener,
+    batch_config: BatchProver,
+    block_config: BlockProver,
```
bobbinth, line 36: The `_config` suffix here is a bit confusing. These fields are not just configs, right?

bobbinth (commented):
Looks good! Thank you! I left some comments inline. The main one is about removing the code that use to check if any of the notes became authenticated before batch construction. It seems to me that we might still need this code. But also, we can bring it back in a follow-up PR.

Once this is merged, we should try to run client integration tests against `next-block-producer` branch. cc @igamigo.

crates/block-producer/src/server/mod.rs
Mirko-von-Leipzig: Initially they were; I was hoping to have a single struct which contains just the configurable values, and then runtime specific things get passed into the `run` method e.g.

```rust
struct BlockProver {
    pub block_interval: Duration,
    pub simulated_proof_time: Range<Duration>,
    pub failure_rate: f32,
}

impl BlockProver {
    async fn run(&self, mempool, store)
}
```

which had the benefit that we could implement `Default` for it with the `const SERVER_XXX` values.

It didn't mesh well with the current flow in the node start command so I undid it.

I'll maybe streamline this again in a future PR; will fixup the naming thanks.

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: I was hoping to remove the traits entirely. Though this might change once I attempt to test things.

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: Iiuc we have several opportunities to fetch note inclusion proofs:

1. Transaction proving (client side)
2. Transaction inputs (currently we just fetch whether or not they're present I think)
3. Batch proving
4. Block proving

Of these, only (4) absolutely has to happen to complete the final proof?

And of these, only (3) has no other database IO at all aside from this so I thought maybe its somewhat pointless. As in, the rest all have to reach out to the database for other data as well, so may as well amortize the network/disk IO.

I'll add it back though; no solid argument against it :)

crates/block-producer/src/mempool/dependency_graph.rs
Mirko-von-Leipzig: Added to #519

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: Re-adding this I did run into [this](https://github.com/0xPolygonMiden/miden-node/blob/573f37b0fd7cd572889cdf2ce20623495e421f3f/crates/block-producer/src/errors.rs#L103) error description, which makes it sound like I need an accurate list to request from the store?

Previously this wasn't an issue because we effectively had the entire batch set to query against, whereas now we only have this one batch's data. i.e. we don't know which notes are inflight still, and which are committed in the store?

If we want to do this check here then we need to either allow the store lookup to be partial, or to perform some additional bookkeeping somehow.

Mirko-von-Leipzig: What I wanted to do was change `GetTransactionInputs` to include the note inclusion proofs. And then batch building doesn't reach out the store at all, and finally only the block builder gets the outstanding since it can remove ephemeral notes.

Mirko-von-Leipzig:
I managed to run the node successfully; though this is without traffic so its really just exercising empty blocks.

crates/block-producer/src/store/mod.rs
bobbinth: The scenario that we want to handle here is roughly like this:

1. Transaction `A` which consumes an unauthenticated note `x` arrives to the block producer. At this point, `x` is in-flight.
2. During transaction validation, we go to the store but since `x` is in-flight, there is nothing new the store can tell us about `x`.
3. Transaction which creates note `x` gets included in the block and persisted to the store.
4. Then, transaction `A` gets selected for batching.

At this point, we want to "upgrade" `x` to an authenticated note. For this, we need to get its authentication path and there are 2 ways to do it:

- We can get the path from the store. This is how it worked before.
- We can attach authentication path to unauthenticated notes as a part of step 3 above (i.e., when the new block gets built).

The latter approach could be more efficient as it saves us a trip to the store, but I think it would also require quite a bit more bookkeeping and has potential for more bugs. So, for now, I think the former approach (i.e., going to the store for authentication info) is probably the simplest solution.

And if we need to update how the store endpoint works for this, I think that's fine.

.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+[build]
+# Enable tokio's unstable feature set: https://docs.rs/tokio/latest/tokio/#unstable-features
+rustflags = ["--cfg", "tokio_unstable"]
```
bobbinth, line 3: What do we use this for?

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -178,18 +168,27 @@ impl WorkerPool {
...
         info!(target: COMPONENT, num_txs, "Building a transaction batch");
         debug!(target: COMPONENT, txs = %format_array(txs.iter().map(|tx| tx.id().to_hex())));
 
+        // TODO: This is a deep clone which can be avoided by change batch building to using
+        // refs or arcs.
+        let txs = txs
+            .iter()
+            .map(AuthenticatedTransaction::raw_proven_transaction)
+            .cloned()
+            .collect();
```
bobbinth, line 186: I agree. I don't think there is a good reason why `TransactionBatch::new()` cannot take something like `IntoIterator<Item = AsRef<ProvenTransaction>` as input.

bobbinth (approved):
Looks good! Thank you! I left a couple a question and a comment for the future inline.

The only other comment is https://github.com/0xPolygonMiden/miden-node/pull/541#discussion_r1827633829 - but we should probably merge this PR and then address this comment in a follow-up PR.

.cargo/config.toml
Mirko-von-Leipzig: I thought we needed this to get the `JoinSet` task IDs, but I was wrong. Removing, thanks!

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: Tracked in #531

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: Okay I'll leave this PR as is and re-introduce this in another PR with store changes as needed.

PR #542, opened 2024-11-05
docs: generate RPC documentation

polydez:
Resolves https://github.com/0xPolygonMiden/miden-node/issues/512

Added generation of API reference documentation (resulting file is docs/api.md). Regrouped documentation in order to utilize node docs in docs/index.md.

The most challenging here is to automatically generate corresponding messages right after each endpoint and this seems to be impossible to do using `protoc-gen-doc` or any other tools I had chance to try. So if it is crucial, we will probably need to regroup them manually. But for the first attempt I tried to do fully automated API reference docs generation.


proto/transaction.proto
```diff
@@ -4,13 +4,21 @@ package transaction;
...
     digest.Digest id = 1;
 }
 
+// Represents a transaction summary.
 message TransactionSummary {
+    // The transaction ID.
     TransactionId transaction_id = 1;
+
+    // The block number.
     fixed32 block_num = 2;
```
bobbinth, line 19: This comment is not very helpful (i.e., we can see from the name that this is a block number). Sometimes, such things cannot be avoided, but in this case, we should actually provide more detail on what this block number means.

proto/transaction.proto
```diff
@@ -4,13 +4,21 @@ package transaction;
...
+
+    // The account ID.
     account.AccountId account_id = 3;
```
bobbinth, line 22: Same comment as above.

proto/account.proto
```diff
@@ -3,31 +3,46 @@ package account;
...
+// A summary of an account.
 message AccountSummary {
+    // The account ID.
     AccountId account_id = 1;
+
+    // The latest account hash, zero hash if the account doesn't exist.
     digest.Digest account_hash = 2;
+
+    // Merkle path to verify the account's inclusion in the MMR.
     uint32 block_num = 3;
```
bobbinth, line 23: This comment doesn't seem correct.

proto/account.proto
```diff
@@ -3,31 +3,46 @@ package account;
...
+// An account ID.
 message AccountId {
     // A miden account is defined with a little bit of proof-of-work, the id itself is defined as
     // the first word of a hash digest. For this reason account ids can be considered as random
     // values, because of that the encoding below uses fixed 64 bits, instead of zig-zag encoding.
     fixed64 id = 1;
 }
 
+// A summary of an account.
 message AccountSummary {
```
bobbinth, line 15: "A summary of an account" comment is not very helpful here. I would expand it to a couple of sentences which describe what it is and what it is used for.

proto/account.proto
```diff
@@ -3,31 +3,46 @@ package account;
...
     uint32 block_num = 3;
 }
 
+// An account info.
 message AccountInfo {
+    // Account summary.
     AccountSummary summary = 1;
+
+    // Account details encoded using Miden native format.
     optional bytes details = 2;
```
bobbinth, line 32: I would describe what the details actually contain (i think they contain full account state description, right?). Also, we should say why the are optional.

proto/mmr.proto
```diff
@@ -3,7 +3,11 @@ package mmr;
 
 import "digest.proto";
 
+// Represents an MMR delta.
 message MmrDelta {
```
bobbinth, line 7: Let's add some detail on what is the MMR delta.

proto/merkle.proto
```diff
@@ -3,6 +3,8 @@ package merkle;
 
 import "digest.proto";
 
+// Represents a Merkle path.
 message MerklePath {
```
bobbinth, line 7: nit: let's say "Represents a Merkle authentication path."

bobbinth (commented):
Thank you! Not a full review, but I left some comments inline. These comments are not exhaustive - similar issues apply in other places which I haven't commented on. Basically, in may cases the added comments are not very useful. We should go a layer deeper and explain things more so that even an unfamiliar person can understand them.

Regarding the new doc format. I'm not sure I like it. It combines everything together, and IMO, makes docs more difficult to navigate. Is there a way to generate docs separately for RPC, store, and block producers components?

polydez:
> Regarding the new doc format. I'm not sure I like it. It combines everything together, and IMO, makes docs more difficult to navigate. Is there a way to generate docs separately for RPC, store, and block producers components?

I tried, but didn't find any way to do it by just using tools for automatic docs generation. We still have options, for example, we can write script for post-processing of generated docs in order to regroup their contents. Or regroup them manually.

Mirko-von-Leipzig:
> > Regarding the new doc format. I'm not sure I like it. It combines everything together, and IMO, makes docs more difficult to navigate. Is there a way to generate docs separately for RPC, store, and block producers components?
> 
> I tried, but didn't find any way to do it by just using tools for automatic docs generation. We still have options, for example, we can write script for post-processing of generated docs in order to regroup their contents. Or regroup them manually.

I imagine we would have to more concretely separate the various RPC service files. Which is something we want to do at some point in any case I believe?

Makefile
```diff
@@ -39,6 +39,10 @@ lint: format fix clippy ## Runs all linting tasks at once (Clippy, fixing, forma
 .PHONY: doc
 doc: ## Generates & checks documentation
 	$(WARNINGS) cargo doc --all-features --keep-going --release --locked
+	docker run --rm \
+      -v $(PWD)/docs:/out \
+      -v $(PWD)/proto:/protos \
+      pseudomuto/protoc-gen-doc --doc_opt=/out/api.tmpl,api.md
```
Mirko-von-Leipzig, line 45: We probably need a separate CI check to ensure the generated docs match this?

I'm also unsure if this should be part of `make doc` like this; though maybe its fine.

I would:

- add a `README.md` with instructions to the `docs` folder with these instructions
- optionally remove these instructions from here (only manual invocation)
- add ci check to ensure we stay in sync

Mirko-von-Leipzig (approved):
LGTM; some questions, suggestions but nothing major.

Probably we should have an actual documentation person go over the wording of things, but that can be a separate pass once we're happy.

proto/rpc.proto
```diff
@@ -6,15 +6,55 @@ import "requests.proto";
 import "responses.proto";
 
 service Api {
+    // Gets a list of proofs for given nullifier hashes, each proof as a sparse Merkle Tree.
     rpc CheckNullifiers(requests.CheckNullifiersRequest) returns (responses.CheckNullifiersResponse) {}
```
bobbinth, line 10: @igamigo - are we using this endpoint anywhere in the client?

proto/rpc.proto
```diff
@@ -6,15 +6,55 @@ import "requests.proto";
...
+
+    // Returns the latest state of an account with the specified ID.
     rpc GetAccountDetails(requests.GetAccountDetailsRequest) returns (responses.GetAccountDetailsResponse) {}
+
+    // Returns the latest state proofs of accounts with the specified IDs.
     rpc GetAccountProofs(requests.GetAccountProofsRequest) returns (responses.GetAccountProofsResponse) {}
+
+    // Returns delta of the account states in the range from `from_block_num` (exclusive) to
+    // `to_block_num` (inclusive).
     rpc GetAccountStateDelta(requests.GetAccountStateDeltaRequest) returns (responses.GetAccountStateDeltaResponse) {}
```
bobbinth, line 23: Not for this PR, but it feels like we should normalize these endpoints somehow. For example:

- I'm not sure if we are actually using `GetAccountStateDelta` in the client - @igamigo, could you check?
- I think the current `GetAccountDetails` endpoint returns the entire account, which would probably not work for large accounts.
- `GetAccountProofsRequest` is used primarily for FPI, but maybe it can be used for other things as well?

Generally, we should probably list the use case we have for retrieving account data from the node and see if these 3 methods are still the right way to get this data. @polydez - could you create an issue for this?

bobbinth (commented):
Thank you! I left some comments inline - mostly not related to this PR.

For this PR, I think the updates in the `.proto` files look fine, but I think auto-generated documentation is not great. I think it actually makes things a bit worse as compared to now.

So, maybe we split the PR so that we can commit `.proto` changes and maybe update the existing documentation to match the current state.

Then, in a separate PR, we can try to refactor how the `.proto` files are organized as @Mirko-von-Leipzig suggested or maybe do something else that will improve the overall readability of the documentation.

proto/rpc.proto
igamigo: [Responding here for completeness as well](https://github.com/0xPolygonMiden/miden-node/issues/616#issuecomment-2587786809)

proto/rpc.proto
igamigo: We don't use this one. We do use the one that checks by prefix, when we import notes to check whether it was consumed already or not.

bobbinth:
Closing as stale.

PR #545, opened 2024-11-05
fix: Return correct latest block number

igamigo:


crates/store/src/state.rs
```diff
@@ -99,7 +99,7 @@ struct InnerState {
 impl InnerState {
     /// Returns the latest block number.
     fn latest_block_num(&self) -> BlockNumber {
-        (self.chain_mmr.forest() + 1).try_into().expect("block number overflow")
+        (self.chain_mmr.forest() - 1).try_into().expect("block number overflow")
```
bobbinth, line 102: I would maybe add a safety comment here saying that `chain_mmr` always has at least one block (the genesis block) - and so this should be fine.

bobbinth (approved):
Great catch! Thank you!

crates/store/src/state.rs
igamigo: Yeah, I thought of the potential underflow but realized it was not an issue but did not write a comment. Will add now.

PR #546, opened 2024-11-06
ci: improve release devops

Mirko-von-Leipzig:
This PR changes the deploy workflow to a more controlled manual flow.

It also distinguishes between creating a package, and bundling it into a release which simplifies the code overall.

Currently, deployment is automatic and occurs on each push for `devnet` and on each release for `testnet`. This doesn't quite suit our needs as often we want to release, test on devnet and only then deploy on testnet. Further, currently its impossible to create release artifacts without doing a deployment as well. This PR separates these concerns and allows for manually deploying.

This PR also fixes some (as yet unknown) issues with the deployment scripts, causing deployments to not fully take (known to occur on at least devnet).

__This PR is not yet complete__.

### DIFF
--- a/.github/actions/build_package/action.yml
+++ b/.github/actions/build_package/action.yml
@@ -1,0 +1,93 @@
+name: build-package
+description: Builds miden-node and miden-faucet debian packages for the given git reference
+inputs:
+  gitref:
+    required: true
+    description: The git ref to build the packages from.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Identify target git SHA
+      id: git-sha
+      shell: bash
+      run: |
+        if git show-ref -q --verify "refs/remotes/origin/$gitref" 2>/dev/null; then
+          echo "sha=$(git show-ref --hash --verify "refs/remotes/origin/$gitref")" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,154 @@
...
+            sudo apt install awscli -y
+
+      # Move packages to instance using S3.
+      # TODO: does this fail if the files already exist?
+      - name: Upload packages to S3
+        run: |
+          aws s3 cp ${{ env.node-package }} ${{ env.s3-path }}/${{ env.node-package }}
+          aws s3 cp ${{ env.node-faucet }} ${{ env.s3-path }}/${{ env.node-faucet }}
+
+      # TODO: does this fail if the files already exist?
```
Dominik1999, line 102: Do we need the still as a ToDo?

Dominik1999 (approved):
looks good to me, but I have little exp in this DevOps wizardry

.github/workflows/deploy.yml
Mirko-von-Leipzig: Yeah its something I'll have to try out at some stage - can't try it locally since its a command on a remote aws server.

PR #549, opened 2024-11-13
fix(mempool): allow internal batch dependencies

Mirko-von-Leipzig:
This PR fixes a crash when a batch contained both a parent and child transaction.

In addition, this PR improves the error presentation when the node crashes.

---------------------

Background: client integration tests were stalling when run against the new mempool based block-producer. This PR fixes the bug that causes this.

The stall occurred because the block-producer component exited, but the node continued on as it only crashes if a component errors. This PR rectifies this behaviour; we now crash the node if any component ends. In addition, this PR improves the errors by identifying the failing component.

This PR also merges in `next`. This was required to properly compare and debug the client integration tests. It might be best to review the last few commits instead.


crates/block-producer/src/server/mod.rs
```diff
@@ -125,38 +124,33 @@ impl BlockProducer {
...
+        //
         // SAFETY: The JoinSet is definitely not empty.
         let task_result = tasks.join_next_with_id().await.unwrap();
+
         let task_id = match &task_result {
-            Ok((id, _)) => *id,
+            Ok((id, ())) => *id,
             Err(err) => err.id(),
         };
+        let task = task_ids.get(&task_id).unwrap_or(&"unknown");
```
igamigo, line 146: Same comment about panicking instead of defaulting to "unknown"

crates/block-producer/src/mempool/batch_graph.rs
```diff
@@ -278,6 +282,17 @@ mod tests {
         assert_eq!(err, expected);
     }
 
+    #[test]
+    fn insert_with_internal_parent_succeeds() {
+        //! Ensure that a batch with internal dependencies can be inserted.
```
bobbinth, line 287: nit: this should be a simple comment (i.e., `//`), right?

crates/block-producer/src/mempool/batch_graph.rs
```diff
@@ -91,7 +91,7 @@ impl BatchGraph {
         &mut self,
         id: BatchJobId,
         transactions: Vec<TransactionId>,
-        parents: BTreeSet<TransactionId>,
+        mut parents: BTreeSet<TransactionId>,
```
bobbinth, line 94: Could we add an explanation of what `parents` is to the doc comment for this function? (e.g., we should mention that parents may contain transactions also in `transactions`).

bobbinth (approved):

bin/node/src/commands/start.rs
```diff
@@ -16,22 +18,41 @@ pub async fn start_node(config: NodeConfig) -> Result<()> {
...
+    ]);
+
+    // SAFETY: The joinset is definitely not empty.
+    let component_result = join_set.join_next_with_id().await.unwrap();
+
+    // We expect components to run indefinitely, so we treat any return as fatal.
+    //
+    // Map all outcomes to an error, and provide component context.
+    let (id, err) = match component_result {
+        Ok((id, Ok(_))) => (id, Err(anyhow::anyhow!("Component completed unexpectedly"))),
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: This context gets added lower down, since we first have to map the ID to component name.

crates/block-producer/src/mempool/batch_graph.rs
Mirko-von-Leipzig: I was hoping we could somehow generate docs for these, allowing us to browse tests and their intentions via `cargo doc`. Seemed like a nice way to get an overview of what's been tested and how.

Unfortunately doesn't work this way; no way to generate these that I could find.

bin/node/src/commands/start.rs
```diff
@@ -16,22 +18,41 @@ pub async fn start_node(config: NodeConfig) -> Result<()> {
...
+        Ok((id, Err(err))) => (id, Err(err)),
+        Err(join_err) => (join_err.id(), Err(join_err).context("Joining component task")),
+    };
+    let component = component_ids.get(&id).unwrap_or(&"unknown");
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: It could be; but I thought this would result in a better error message than a panic in the case that we introduce a bug or forget to add a new component to the list.

With the panic we only get a `"unknown node component task ID"` message, whereas this way we get something more structured and critically includes the actual error:

```
Error: Component unknown failed

Caused by:
    0: <inner error>
```

PR #550, opened 2024-11-15
Slightly simplified solution for avoiding TX cloning for transaction batch

polydez:
My proposal to simplify https://github.com/0xPolygonMiden/miden-node/pull/544


Mirko-von-Leipzig (approved):
Perfect thank you!

When I tried this I failed to understand that one would need to restrict/define the IntoIter associated item as well. TIL!

PR #551, opened 2024-11-21
Spec 474

BrucePolygon:
secret added and updated in code.


djpolygon:
Updated branch to use next, so that Miden team can have it prepped for their next release.

djpolygon (approved):

Mirko-von-Leipzig (approved):

PR #554, opened 2024-11-26
Refactor and store account deltas by separate fields in the database

polydez:
Related to https://github.com/0xPolygonMiden/miden-node/issues/527

In this PR we redesigned delta storage, now it is written to the DB in separate tables/fields. This also affects the way we build account's delta for `GetAccountStateDelta` endpoint.

In current design, we write "initial" account state as delta and keep all the changed from the very beginning.


crates/store/src/db/migrations/001-init.sql
Mirko-von-Leipzig, line 1: Some suggestions for table names

```
account_nonce_updates
account_storage_slot_updates
account_storage_map_updates
account_non_fungible_asset_updates
```

We can also consider dropping the `account_` prefix?

Mirko-von-Leipzig: The `account_deltas` table only adds the `nonce` field - should this maybe get merged into the `accounts` table, which can then just become `account_updates`?

Some of the existing account tables also don't have `WITHOUT ROWID` - should we add that?

We're also duplicating the composite primary key everywhere (`account_id, block_num`), I'm unsure how expensive that is versus creating a `account_update_id` as PK and referencing that.

crates/store/src/db/sql.rs
```diff
@@ -184,41 +189,125 @@ pub fn select_accounts_by_ids(
...
+
+    #[derive(num_derive::FromPrimitive)]
+    enum RecordType {
+        StorageScalars = 0,
+        StorageMapValues,
+        FungibleAssets,
+        NonFungibleAssets,
+    }
+    let mut select_merged_deltas_stmt =
+        conn.prepare_cached(include_str!("sql-queries/select_merged_deltas.sql"))?;
```
Mirko-von-Leipzig, line 239: Do we know if this `UNION ALL` approach is better than just having four separate queries?

How do the `explain query plan` compare?

crates/store/src/db/sql.rs
```diff
@@ -184,41 +189,125 @@ pub fn select_accounts_by_ids(
...
+
+                fungible.insert(faucet_id.try_into()?, value);
+            },
+            RecordType::NonFungibleAssets => {
+                let vault_key_data = row.get_ref(FieldIndex::Key as usize)?.as_blob()?;
+                let vault_key = Word::read_from_bytes(vault_key_data)?;
+
+                // SAFETY: This conversion is safe because we previously wrote this key from
+                //   internal value of `NonFungibleAsset`.
+                let asset = unsafe { NonFungibleAsset::new_unchecked(vault_key) };
```
Mirko-von-Leipzig, line 291: Note that this could occur in practice due to disk corruption. I'm not sure we should allow `unsafe` here.

crates/store/src/db/migrations/001-init.sql
polydez: > We can also consider dropping the `account_` prefix?

I think, we shouldn't drop the prefix. It helps to understand that these tables are related to account states.

polydez: > The `account_deltas` table only adds the `nonce` field - should this maybe get merged into the `accounts` table, which can then just become `account_updates`?

No, these tables have different meanings. `accounts` table stores the latest account state, but `account_deltas` (`account_nonce_updates`) stores each update of the nonce.

> Some of the existing account tables also don't have `WITHOUT ROWID` - should we add that?

Ah, yes, thank you!

> We're also duplicating the composite primary key everywhere (`account_id, block_num`), I'm unsure how expensive that is versus creating a `account_update_id` as PK and referencing that.

Yes, you are right. But I'm also not sure yet, if we could benefit from this.

crates/store/src/db/sql.rs
```diff
@@ -1077,6 +1240,37 @@ pub(crate) fn schema_version(conn: &Connection) -> rusqlite::Result<usize> {
     conn.query_row("SELECT * FROM pragma_schema_version", [], |row| row.get(0))
 }
 
+/// Generates a simple insert SQL statement with values for the provided table, fields, and record
+/// number.
+fn insert_sql(table: &str, fields: &[&str], record_count: usize) -> String {
```
bobbinth, line 1245: nit: this file is pretty big already, I wonder if we should move out functions like this into something like `src/db/sql/utils.rs` (and rename the current `sql.rs` into `src/db/sql/mod.rs`). We could also move the `sql-queries` directory under the `src/db/sql/queries` directory.

bobbinth (commented):
Looks good! Thank you! This is not a super-deep review, but I did leave a few comments inline.

One thing that this PR made me think of: would be good to have a framework for benchmarking performance of the store. Right now, we are kind of guessing what would be efficient and what won't be - but ideally we should try to run simulations to test our hypothesis.

Let's create an issue to build such a benchmarking framework. The things we'd want to benchmark could be startup times and response times for various endpoints under some DB sizes (e.g., DB with 1M accounts, notes, nullifiers etc.).

bobbinth:
Oh - one other question: should this branch be based on `next`?

crates/store/src/db/sql.rs
polydez: I missed that we have another way for `NonFungibleAsset` construction, thank you for noticing!

crates/store/src/db/sql.rs
polydez: We can expect just very little benefit in performance here, but I haven't done any benchmarks yet.

crates/store/src/db/mod.rs
```diff
@@ -457,20 +457,20 @@ impl Db {
         Ok(())
     }
 
-    /// Loads account deltas from the DB for given account ID and block range.
+    /// Merges all account deltas from the DB for given account ID and block range.
     /// Note, that `from_block` is exclusive and `to_block` is inclusive.
-    pub(crate) async fn select_account_state_deltas(
+    pub(crate) async fn select_account_state_delta(
```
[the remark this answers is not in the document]
polydez: Yes, you're right. It returns `Ok(None)` if there is no deltas found for the specified account within the specified block range.

crates/store/src/db/sql-queries/select_merged_deltas.sql
```diff
@@ -0,0 +1,84 @@
...
+    block_num <= ?3 AND
+    NOT EXISTS(
+        SELECT 1
+        FROM account_storage_slot_updates AS b
+        WHERE
+            b.account_id = ?1 AND
+            a.slot = b.slot AND
+            a.block_num < b.block_num AND
+            b.block_num <= ?3
+    )
```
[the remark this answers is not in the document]
polydez: This is good point, thank you! My approach is more old-school, it forces optimizer to filter by existence, but I think, yours gives it more room for optimizations which might result in better performance. But to answer the question, it will require benchmarking. I put this query to the issue: https://github.com/0xPolygonMiden/miden-node/issues/558

polydez:
> Oh - one other question: should this branch be based on `next`?

Once https://github.com/0xPolygonMiden/miden-node/pull/559 is merged, we can rebase/merge this PR to `next`.

crates/store/src/db/migrations/001-init.sql
```diff
@@ -57,15 +57,75 @@ CREATE TABLE
 ) STRICT;
 
 CREATE TABLE
-    account_deltas
+    account_nonce_updates
```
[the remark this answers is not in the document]
bobbinth: I would probably still use the name `account_updates` (rather than `account_deltas`) to be more consistent with the naming of other tables.

bobbinth (approved):
Looks good! Thank you! Not a very deep review from me, but I left just one small comment inline. Would be good for @Mirko-von-Leipzig to take another look.

Also, another thing that came to mind is that maybe it would be good split up the `src/db/sql/mod.rs` into smaller files - but let's do it in a future PR.

crates/store/src/db/sql/utils.rs
```diff
@@ -0,0 +1,154 @@
...
+///
+/// Sqlite uses `i64` as its internal representation format, and so when retrieving
+/// we need to make sure we cast as `u64` to get the original value
+pub fn column_value_as_u64<I: rusqlite::RowIndex>(
+    row: &rusqlite::Row<'_>,
+    index: I,
+) -> rusqlite::Result<u64> {
+    let value: i64 = row.get(index)?;
+    Ok(value as u64)
+}
```
Mirko-von-Leipzig, line 99: Are these really necessary? rusqlite already [provides](https://docs.rs/rusqlite/latest/rusqlite/types/trait.ToSql.html) these out the box I think. Not really part of this PR though, so feel free to ignore.

crates/store/src/db/sql/queries/select_merged_deltas.sql
Mirko-von-Leipzig, line 1: I would prefer four simpler, separate queries over the `union all` approach unless profiling shows its significantly superior. However this can wait for the db benchmarking work.

crates/store/src/db/sql/utils.rs
```diff
@@ -0,0 +1,154 @@
...
+    values: impl IntoIterator<Item: ToSql>,
+) -> rusqlite::Result<usize> {
+    if record_count == 0 {
+        return Ok(0);
+    }
+
+    let sql = insert_sql(table, fields, record_count);
+
+    transaction.execute(&sql, params_from_iter(values))
+}
```
Mirko-von-Leipzig, line 71: I have some concerns about these abstractions:

- Note that using these makes it impossible to cache the sql statements which can make this quite expensive if used repeatedly.
- I'm unsure if there is performance value in a single batch insert versus a single statement in a loop. In my experience bulk inserts had no performance gain.
- The sql statement str size is also proportional to the delta size which is potentially an issue? i.e. this should be const/bounded?

Mirko-von-Leipzig (commented):
Some doubts in particular around the `bulk_insert`; the rest are minor.

crates/store/src/db/sql/utils.rs
polydez: I think, we can get rid of `u32_to_value`, but we still need the rest, because `rusqlite` can convert values only if they fit to destination type. It will return errors, if source type under-/overflows destination one.

crates/store/src/db/migrations/001-init.sql
```diff
@@ -61,11 +61,71 @@ CREATE TABLE
...
+    account_id  INTEGER NOT NULL,
+    block_num   INTEGER NOT NULL,
+    slot        INTEGER NOT NULL,
+    value       BLOB    NOT NULL,
+
+    PRIMARY KEY (account_id, block_num, slot),
+    FOREIGN KEY (account_id, block_num) REFERENCES account_deltas (account_id, block_num)
+) STRICT, WITHOUT ROWID;
+
+CREATE INDEX idx_account_storage_slot_updates_slot ON account_storage_slot_updates (slot);
```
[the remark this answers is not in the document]
polydez: I expected that we would have sorting by that fields, but let's remove them for now and add once it's needed.

crates/store/src/db/sql/utils.rs
polydez: These are good points, I expected that compilation and query plan building are less expensive than inserts itself, but I think we need to benchmark/profile it to have right answer. According to comments on Stack Overflow, some people got up to 2x performance gain in bulk inserts comparing to multiple inserts in single transaction, but this might be different in our workflow. 

According to SQLite docs: 
> The maximum number of bytes in the text of an SQL statement is limited to SQLITE_MAX_SQL_LENGTH which defaults to 1,000,000,000.

So, we won't hit the limit, but I can't say if linear growth of SQL statement size might lead to non-linear computational cost.

Nevertheless, I think, simplicity of code is better than premature optimization, so I agree with your suggestions of simpler approaches before we have performance measures which show better approaches in terms of performance.

crates/store/src/db/sql/utils.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// Generates a simple insert SQL statement with parameters for the provided table and fields
+pub fn insert_sql(table: &str, fields: &[&str]) -> String {
+    format!(
+        "INSERT INTO {table} ({}) VALUES ({})",
+        fields.join(", "),
+        "?, ".repeat(fields.len()).trim_end_matches(", ")
+    )
+}
```
Mirko-von-Leipzig, line 45: Final thought: could we make this const/static somehow? This will allocate each time despite the sql being a constant str per query.

Probably requires a `macro_rules!` though. Can definitely leave that sort of thing to the benchmarking time though.

Mirko-von-Leipzig (approved):

crates/store/src/db/sql/utils.rs
polydez: This is nice idea, thank you! But I have no idea, how to implement it on stable Rust. I have only solution for unstable: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=4b7136a128a7e57e44c5af8c10072dbd

Mirko-von-Leipzig: If you're up for a challenge: [tt-muncher macro](https://veykril.github.io/tlborm/decl-macros/patterns/tt-muncher.html) would be one of achieving it (I believe). Its a fair bit of complication though; so likely simply writing out the sql by hand would be better all round 😆

polydez: Thank you! Finally I've found a solution that works. Please, take a look. If it's overkilling, we still may revert back to hand-written SQLs. :)

crates/store/src/db/sql/utils.rs
```diff
@@ -35,13 +35,27 @@ pub fn schema_version(conn: &Connection) -> rusqlite::Result<usize> {
...
+            stringify!(INSERT INTO $table),
+            " (",
+            stringify!($first_field),
+            $($(concat!(", ", stringify!($field))),+ ,)?
+            ") VALUES (",
+            subst!($first_field, "?"),
+            $($(subst!($field, ", ?")),+ ,)?
+            ")"
+        )
+    };
```
Mirko-von-Leipzig, line 58: Nice. I think maybe document usage, e.g.

```rust
/// # Usage
///
/// ```
/// insert_sql!(<table_name> {
///     <column_0>,
///     <column_1>,
///     ...
/// })
/// ```
```

Mirko-von-Leipzig (approved):

Issue #403, opened 2024-07-11
Config assumes endpoints are http protocol


Mirko-von-Leipzig:
Endpoints are defined as `(host, port)` tuples and then [make the assumption](https://github.com/0xPolygonMiden/miden-node/blob/7f1909926caa481e3b3e117b44255dbf68ab738c/crates/utils/src/config.rs#L44) that the protocol is always `http`.

We probably want to at least support `https` as well.

varun-doshi:
I can take this up with some more information about how to proceed

Mirko-von-Leipzig:
@varun-doshi I think the least intrusive approach would be to add an enum to the endpoint like:

```rust
enum Protocol {
    Http,
    Https,
}
```

PR #556, opened 2024-11-29
feat: added https support in endpoint config

varun-doshi:
Ref #403 

Added https support in the endpoints struct.
By default, all default calls are to `http` type only as expected for localhost.

### DIFF
--- a/bin/node/src/config.rs
+++ b/bin/node/src/config.rs
@@ -93,10 +93,10 @@ mod tests {
                     verify_tx_proofs = true
 
                     [rpc]
-                    endpoint = { host = "127.0.0.1",  port = 8080 }
+                    endpoint = { host = "127.0.0.1",  port = 8080, protocol = "Http" }
 
                     [store]
-                    endpoint = { host = "127.0.0.1",  port = 8080 }
+                    endpoint = { host = "127.0.0.1",  port = 8080, protocol = "Https" }
                     database_filepath = "local.sqlite3"
                     genesis_filepath = "genesis.dat"
                     blockstore_dir = "blocks"
--- a/bin/node/src/config.rs
+++ b/bin/node/src/config.rs
@@ -112,7 +112,7 @@ mod tests {
                         endpoint: Endpoint {
                             host: "127.0.0.1".to_string(),
                             port: 8080,
-                            protocol: Protocol::Http
+                            protocol: Protocol::default()
                         },
                         verify_tx_proofs: true
                     },
@@ -127,7 +127,7 @@ mod tests {
                         endpoint: Endpoint {
                             host: "127.0.0.1".to_string(),
                             port: 8080,
-                            protocol: Protocol::Http
+                            protocol: Protocol::Https
                         },
                         database_filepath: "local.sqlite3".into(),
                         genesis_filepath: "genesis.dat".into(),
--- a/bin/node/src/config.rs
+++ b/bin/node/src/config.rs
@@ -74,7 +74,7 @@ impl NodeConfig {
 mod tests {
     use figment::Jail;
     use miden_node_store::config::StoreConfig;
-    use miden_node_utils::config::{load_config, Endpoint};
+    use miden_node_utils::config::{load_config, Endpoint, Protocol};
 
     use super::NodeConfig;
     use crate::{
@@ -112,6 +112,7 @@ mod tests {
                         endpoint: Endpoint {
                             host: "127.0.0.1".to_string(),
                             port: 8080,
+                            protocol: Protocol::Http
                         },
                         verify_tx_proofs: true
                     },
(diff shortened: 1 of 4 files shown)


crates/utils/src/config.rs
```diff
@@ -41,7 +52,10 @@ impl ToSocketAddrs for Endpoint {
 
 impl Display for Endpoint {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        f.write_fmt(format_args!("http://{}:{}", self.host, self.port))
+        match self.protocol {
+            Protocol::Http => f.write_fmt(format_args!("http://{}:{}", self.host, self.port)),
+            Protocol::Https => f.write_fmt(format_args!("https://{}:{}", self.host, self.port)),
+        }
```
Mirko-von-Leipzig, line 58: This works, but I think we improve readibility by implementing `Display for Protocol` and then do something like:

```suggestion
        let Endpoint { protocol, host, port } = self;
        f.write_fmt(format_args!("{protocol}://{host}:{port}"))
```

crates/utils/src/config.rs
```diff
@@ -17,18 +17,29 @@ pub const DEFAULT_BLOCK_PRODUCER_PORT: u16 = 48046;
 pub const DEFAULT_STORE_PORT: u16 = 28943;
 pub const DEFAULT_FAUCET_SERVER_PORT: u16 = 8080;
 
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
+pub enum Protocol {
+    Http,
```
Mirko-von-Leipzig, line 22: I think impl `Default` here. And then replace the `protocol: Protocol::Http` with `Default::default()`. This way we can change `Http` to `Https` default by only changing the enum's default.

```suggestion
    #[default]
    Http,
```

bin/node/src/config.rs
```diff
@@ -112,19 +112,22 @@ mod tests {
                         endpoint: Endpoint {
                             host: "127.0.0.1".to_string(),
                             port: 8080,
+                            protocol: Protocol::Http
```
Mirko-von-Leipzig, line 115: Could you update this test's endpoint inputs so that one of them remains the default (i.e. isn't specified), one is set to `"Http"`, and the other to `"Https"`. This should cover all cases.

Mirko-von-Leipzig (changes requested):
Thank you! Just some minor changes :)

Mirko-von-Leipzig (commented):
Could you add an entry into the `CHANGELOG.md` for this? 

Something like `"Support Https in endpoint configuration"` to the Enhancements section. Note that we sort these chronologically, so yours should come after the `Added GetAccountProofs endpoint (#506).`

crates/utils/src/config.rs
```diff
@@ -52,9 +52,16 @@ impl ToSocketAddrs for Endpoint {
...
+        let Endpoint { protocol, host, port } = self;
+        f.write_fmt(format_args!("{protocol}://{host}:{port}"))
+    }
+}
+
+impl Display for Protocol{
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self{
+            Protocol::Http => f.write_fmt(format_args!("http://")),
+            Protocol::Https => f.write_fmt(format_args!("https://")),
```
Mirko-von-Leipzig, line 64:
```suggestion
            Protocol::Http => f.write_str("http"),
            Protocol::Https => f.write_str("https")),
```

bin/node/src/config.rs
```diff
@@ -112,19 +112,22 @@ mod tests {
...
                         },
                     },
                     store: StoreConfig {
                         endpoint: Endpoint {
                             host: "127.0.0.1".to_string(),
                             port: 8080,
+                            protocol: Protocol::Https
```
Mirko-von-Leipzig, line 130: You'll need to update the input string from above as well:

```toml
                    [block_producer]
                    endpoint = { host = "127.0.0.1",  port = 8080 }
                    verify_tx_proofs = true

                    [rpc]
                    endpoint = { host = "127.0.0.1",  port = 8080, protocol = "http" }

                    [store]
                    endpoint = { host = "127.0.0.1",  port = 8080, protocol = "https" }
                    database_filepath = "local.sqlite3"
                    blockstore_dir = "blocks"
```

Mirko-von-Leipzig (changes requested):
Thanks a couple more minor fixups :)

crates/utils/src/config.rs
```diff
@@ -17,18 +17,29 @@ pub const DEFAULT_BLOCK_PRODUCER_PORT: u16 = 48046;
...
+}
 /// The `(host, port)` pair for the server's listening socket.
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
 pub struct Endpoint {
     /// Host used by the store.
     pub host: String,
     /// Port number used by the store.
     pub port: u16,
+    /// Protocol type: http or https.
+    pub protocol: Protocol,
```
Mirko-von-Leipzig, line 33: Required so that the config deserialization gets the default value is missing.

```suggestion
    /// Protocol type: http or https.
    #[serde(default)]
    pub protocol: Protocol,
```

Mirko-von-Leipzig (approved):
Fantastic, thank you very much!

PR #557, opened 2024-12-02
fix(block-producer): handle reverted batches

Mirko-von-Leipzig:
This PR corrects the mempool's behaviour when batches are reverted.

More concretely, it fixes an issue where a child batch jobs weren't invalidated when their parent batch was reverted. This is fixed by actively tracking inflight batch jobs, and ignoring submitted jobs not part of this list.

### DIFF
--- a/crates/block-producer/src/mempool/dependency_graph.rs
+++ b/crates/block-producer/src/mempool/dependency_graph.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::{BTreeMap, BTreeSet},
-    fmt::Display,
+    fmt::{Debug, Display},
 };
 
 // DEPENDENCY GRAPH
@@ -35,7 +35,7 @@ use std::{
 ///                              │  <null>   ◄────┘                
 ///                              └───────────┘                     
 /// ```
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct DependencyGraph<K, V> {
     /// Node's who's data is still pending.
     pending: BTreeSet<K>,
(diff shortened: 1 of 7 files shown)


crates/block-producer/src/mempool/dependency_graph.rs
```diff
@@ -61,6 +61,20 @@ pub struct DependencyGraph<K, V> {
...
+where
+    K: Debug,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DependencyGraph")
+            .field("pending", &self.pending)
+            .field("vertices", &self.vertices.keys())
+            .field("processed", &self.processed)
+            .field("roots", &self.roots)
+            .finish()
```
Mirko-von-Leipzig, line 74: This debug impl elides data and graph edges. Data would include transaction data which makes this debug impossible to navigate.

I'm unsure if we shouldn't add it all anyway.

igamigo: My two cents: I think I'd add it for completeness since it seems like potentially valuable debug information unless it interrupts an existing workflow somehow. Alternatives could be to have a dedicated method that hides that data or viceversa (and maybe if it's decided that it's important enough, a feature or env variable could help too).

igamigo (approved):
LGTM!

crates/block-producer/src/mempool/mod.rs
```diff
@@ -198,28 +199,45 @@ pub struct Mempool {
     /// The current inflight block, if any.
     block_in_progress: Option<BTreeSet<BatchJobId>>,
 
-    batch_budget: BatchBudget,
+    /// Batches which are currently being proven.
+    ///
+    /// This is used to identify jobs which have been cancelled by the mempool but might still be
+    /// submitted by the batch prover. This is achieved by ignoring all batch proofs which are not
+    /// in this set.
+    batches_in_progress: BTreeSet<BatchJobId>,
```
bobbinth, line 207: Would it make sense to track which batches are in-progress inside of the `BatchGraph`? If we need to access this info here, we could also add `BatchGraph::isInProgress(BatchJobId)` method. Or would this not work for some reason?

bobbinth (approved):
Looks good! Thank you! I left one comment inline: it feels like `BatchGraph` is potentially a more natural place to track this info - but I also didn't think it through completely to see what kind of issues this may cause.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: I've removed this entirely, and instead check that the batch is contained within the graph beforehand instead - this is equivalent.

crates/block-producer/src/mempool/dependency_graph.rs
Mirko-von-Leipzig: I've added the graph edges, but kept the data itself out under the assumption that debugging the graph should just show the graph structure and not the node contents.

PR #559, opened 2024-12-02
chore: migrate code to the latest `miden-base`

polydez:
Removed unused dependencies, updated errors, fixed compilation errors in order to make node compilable against latest `miden-base`.

### DIFF
--- a/crates/store/src/db/tests.rs
+++ b/crates/store/src/db/tests.rs
@@ -4,18 +4,17 @@ use miden_objects::{
     accounts::{
         account_id::testing::{
             ACCOUNT_ID_FUNGIBLE_FAUCET_ON_CHAIN, ACCOUNT_ID_NON_FUNGIBLE_FAUCET_ON_CHAIN,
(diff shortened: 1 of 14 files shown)


Mirko-von-Leipzig (approved):

crates/store/src/db/tests.rs
```diff
@@ -427,12 +416,14 @@ fn test_sql_public_account_details() {
     assert_eq!(account_read.nonce(), account.nonce());
 
     // Cleared item was not serialized, check it and apply delta only with clear item second time:
-    assert_eq!(account_read.storage().get_item(3), Ok(RpoDigest::default()));
+    assert!(
+        matches!(account_read.storage().get_item(3), Ok(digest) if digest == RpoDigest::default())
+    );
```
PhilippGackstatter, line 421: Nit: Could be `assert_matches!` instead.

PhilippGackstatter (approved):
Not really familiar with the node, but the error and test changes look good!

PR #562, opened 2024-12-04
feat(block-producer): switch to mempool version

Mirko-von-Leipzig:
This PR replaces the FIFO based `block-producer` with a dependency graph mempool variant.

This upgrade has largely already been reviewed via PRs into the `next-block-producer` feature branch, however this PR will provide the first holistic overview of the completed feature. As such, it would be helpful if those unfamiliar (I guess everyone excluding @bobbinth) would at minimum do a single somewhat thorough pass.

This dependency graph is a pre-requisite for adding fee support, and creating transaction/batch selection strategies.

## Outstanding work

This feature is not 100% complete, outstanding tasks are tracked in #514. The new `block-producer` should however be functionally equivalent to the FIFO queue. Getting this work back into the development branch will make future additions much less painful.

## Overview

We introduce a `Mempool` which acts as the co-ordination point between incoming transactions, batch and block producers.

### Inflight state

Incoming transactions are validated against the inflight state i.e. is account state correct, and are the notes valid. The inflight state also adds transaction dependency information e.g. the new transaction becomes the child of the last transaction to update the account, or produce the note.

### Dependency graphs

These parent - child dependencies between transactions are tracked in `TransactionGraph`. In addition we have a batch graph which acts as a sort of overlay on-top of the transaction graph. This complicates matters a bit because batch dependencies connect transactions which previously were not connected (which is why its a separate graph).

## Concerns

I am not 100% happy with the mempool implementation. It feels a bit brittle in the sense that one must be careful when making changes. It easy to forget to cleanup a graph node, or remove state in one of the methods.

Its also possible that this shape is a poor match for various selection strategies i.e. one might want access to all meta-data at all stages. We'll have to see and adapt when we get there.

## Meta

I'm unsure if I should squash merge this; or keep the various commits as is (which are effectively the various PRs already). Currently I'm leaning towards the latter..


igamigo (commented):
Great work! I've left some minor comments on the parts I've already checked out. I still need to review the server and some of the mempool code.

crates/block-producer/Cargo.toml
```diff
@@ -38,6 +39,7 @@ miden-lib = { workspace = true, features = ["testing"] }
 miden-node-test-macro = { path = "../test-macro" }
 miden-objects = { workspace = true, features = ["testing"] }
 miden-tx = { workspace = true, features = ["testing"] }
+pretty_assertions = "1.4.1"
```
bobbinth, line 42: nit: we usually omit the patch version.

crates/block-producer/Cargo.toml
```diff
@@ -24,6 +24,7 @@ miden-objects = { workspace = true }
 miden-processor = { workspace = true }
 miden-stdlib = { workspace = true }
 miden-tx = { workspace = true }
+rand = { version = "0.8.5" }
```
bobbinth, line 27: Same nit as above re patch version.

crates/block-producer/src/lib.rs
```diff
@@ -1,37 +1,27 @@
...
 
 /// The name of the block producer component
 pub const COMPONENT: &str = "miden-block-producer";
 
 /// The number of transactions per batch
-const SERVER_BATCH_SIZE: usize = 2;
+const SERVER_MAX_TXS_PER_BATCH: usize = 2;
 
 /// The frequency at which blocks are produced
 const SERVER_BLOCK_FREQUENCY: Duration = Duration::from_secs(10);
```
bobbinth, line 27: Not for this PR, but let's reduce block times to 5 seconds.

crates/block-producer/src/store/mod.rs
```diff
@@ -53,15 +50,6 @@ pub trait Store: ApplyBlock {
...
     ) -> Result<BlockInputs, BlockInputsError>;
-
-    /// Returns note authentication information for the set of specified notes.
-    ///
-    /// If authentication info for a note does not exist in the store, the note is omitted
-    /// from the returned set of notes.
-    async fn get_note_authentication_info(
-        &self,
-        notes: impl Iterator<Item = &NoteId> + Send,
-    ) -> Result<NoteAuthenticationInfo, NotePathsError>;
```
bobbinth, line 64: I think we might need to bring this back in the future. Do we have an issue for this?

Separately (and also for a different PR), I'm sure we need to have `ApplyBlock` as a separate trait. Maybe worth moving `apply_block()` method to the `Store` trait?

crates/block-producer/src/store/mod.rs
```diff
@@ -163,6 +152,20 @@ impl DefaultStore {
     pub fn new(store: store_client::ApiClient<Channel>) -> Self {
         Self { store }
     }
+
+    /// Returns the latest block's header from the store.
+    pub async fn latest_header(&self) -> Result<BlockHeader, String> {
+        // TODO: fixup the errors types.
+        let response = self
```
bobbinth, line 159: What is wrong with error types here? I'd either fix it or make the `TODO` more specific.

crates/block-producer/src/server/mod.rs
```diff
@@ -45,63 +49,191 @@ impl BlockProducer {
...
+        // Spawn rpc server and batch and block provers.
+        //
+        // These communicate indirectly via a shared mempool.
+        //
+        // These should run forever, so we combine them into a joinset so that if
+        // any complete or fail, we can shutdown the rest (somewhat) gracefully.
+        let mut tasks = tokio::task::JoinSet::new();
+
+        // TODO: improve the error situationship.
+        let batch_builder_id = tasks
```
bobbinth, line 110: Could we make this `TODO` more detailed?

crates/block-producer/src/mempool/dependency_graph.rs
```diff
@@ -0,0 +1,1000 @@
...
+    fn contains(&self, key: &K) -> bool {
+        self.pending.contains(key) || self.vertices.contains_key(key)
+    }
+}
+
+// TESTS
+// ================================================================================================
+
+#[cfg(test)]
+mod tests {
```
bobbinth, line 422: nit: I would consider breaking this file up to move tests into a separate file - e.g., `src/mempool/graph/mod.rs` and `src/mempool/graph/tests.rs`.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -1,69 +1,103 @@
...
 
 pub(crate) mod prover;
 
 use self::prover::{block_witness::BlockWitness, BlockProver};
 
-#[cfg(test)]
-mod tests;
+// FIXME: reimplement the tests.
+// #[cfg(test)]
+// mod tests;
```
bobbinth, line 28: Should this be deleted or re-enabled?

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -1,202 +1,195 @@
...
+    domain::transaction::AuthenticatedTransaction,
+    mempool::{BatchJobId, SharedMempool},
+    COMPONENT, SERVER_BUILD_BATCH_FREQUENCY,
+};
 
-#[cfg(test)]
-mod tests;
+// FIXME: fix the batch builder tests.
+// #[cfg(test)]
+// mod tests;
```
bobbinth, line 15: Similar question as above: should we delete this or re-enable it?

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -1,202 +1,195 @@
...
-        let batch = TransactionBatch::new(txs, found_unauthenticated_notes)?;
+        // TODO: This is a deep clone which can be avoided by change batch building to using
+        // refs or arcs.
+        let txs = txs
+            .iter()
+            .map(AuthenticatedTransaction::raw_proven_transaction)
+            .cloned()
+            .collect();
+        // TODO: Found unauthenticated notes are no longer required.. potentially?
+        let batch = TransactionBatch::new(txs, Default::default())?;
```
bobbinth, line 188: Related to one of the previous comments, I think this may need to be reverted in a future PR - do we have an issue for this?

bobbinth (approved):
Looks good! Thank you! Not a very deep review from my end (as I've reviewed most of this code before), but I left some small comments/questions throughout.

bobbinth:
> I'm unsure if I should squash merge this; or keep the various commits as is (which are effectively the various PRs already). Currently I'm leaning towards the latter..

I would probably merge as is (i.e., without squashing).

crates/block-producer/src/lib.rs
Mirko-von-Leipzig: #564

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: Previous comment [here](https://github.com/0xPolygonMiden/miden-node/pull/562#discussion_r1870441700) and added #566.

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: #565 and #566

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: I think the various store methods/trait could likely just use a single error enum instead of multiple for each. I've updated the comment to match this.

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: I've opted to remove this.

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: I've opted to remove this.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -1,202 +1,195 @@
...
 
-        info!(target: COMPONENT, period_ms = interval.period().as_millis(), "Batch builder started");
+type BatchResult = Result<(BatchJobId, TransactionBatch), (BatchJobId, BuildBatchError)>;
 
-        loop {
-            interval.tick().await;
-            self.try_build_block().await;
+/// Wrapper around tokio's JoinSet that remains pending if the set is empty,
+/// instead of returning None.
+struct WorkerPool {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Its possible to do it that way sure. My thinking was that `BatchBuilder` contains the configuration options which includes the worker count. Right now these always just default, but I imagine we would want to expose all these options as cli args e.g.
```
miden-node start block-producer --batch.interval ... --batch.workers 10 --batch.failure-rate 0.1
```
which would then be passed in here. If we move this around then its no longer pure configuration values, and instead also contains state which means you can no longer do things like trivially overriding defaults and instead need a bunch of `fn new`-like functions.

I'll move the worker count into the worker pool regardless; since that still works and probably makes more sense.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -1,69 +1,103 @@
...
+    /// Note: this _must_ be sign positive and less than 1.0.
+    pub failure_rate: f32,
 
-#[derive(Debug)]
-pub struct DefaultBlockBuilder<S, A> {
-    store: Arc<S>,
-    state_view: Arc<A>,
-    block_kernel: BlockProver,
+    pub store: DefaultStore,
+    pub block_kernel: BlockProver,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: No, but similarly to [this comment ](https://github.com/0xPolygonMiden/miden-node/pull/562#discussion_r1871123362), the intention is for this to be easily configurable.

tbh in general I favor making things `pub` unless there is a compelling reason not to.

crates/block-producer/src/mempool/batch_graph.rs
```diff
@@ -0,0 +1,340 @@
...
+    pub fn select_block(
+        &mut self,
+        mut budget: BlockBudget,
+    ) -> BTreeMap<BatchJobId, TransactionBatch> {
+        let mut batches = BTreeMap::new();
+
+        while let Some(batch_id) = self.inner.roots().first().copied() {
+            // SAFETY: Since it was a root batch, it must definitely have a processed batch
+            // associated with it.
+            let batch = self.inner.get(&batch_id).expect("root should be in graph").clone();
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Nice catch

Mirko-von-Leipzig:
Note: `Review: xxx` commits will all be squashed into a single commit before merging, but I'll keep the original PR commits.

tomyrd (approved):
I'm still a little unfamiliar with the node's codebase but I looked through the whole PR and I haven't found any big issues. Left some suggestions/comments, but overall it looks good. It's well explained, I like the graph figures showing the different states/transitions.

I tested this version of the node with the miden-client in the `next` branch (integration tests and manual tests with the CLI) and I haven't found any issues.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -1,202 +1,205 @@
...
-            options,
-            ready_batches: Default::default(),
+                    worker_pool.spawn(batch_id, transactions);
+                },
+                result = worker_pool.join_next() => {
+                    let mut mempool = mempool.lock().await;
+                    match result {
+                        Err(err) => {
+                            tracing::warn!(%err, "Batch job panic'd.")
+                            // TODO: somehow embed the batch ID into the join error, though this doesn't seem possible?
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Thanks, I've added this. I somehow missed that the `AbortHandle` contains the task ID. Much better :D

crates/block-producer/src/mempool/batch_graph.rs
```diff
@@ -0,0 +1,344 @@
...
+    ///   - any batch was not previously selected for inclusion in a block
+    ///   - any batch is unknown
+    ///   - any parent batch would be left dangling in the graph
+    ///
+    /// The last point implies that batches should be removed in block order.
+    pub fn prune_committed(
+        &mut self,
+        batch_ids: BTreeSet<BatchJobId>,
+    ) -> Result<Vec<TransactionId>, GraphError<BatchJobId>> {
+        self.inner.prune_processed(batch_ids.clone())?;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Sort of. The pruning is atomic, however the overall method loses this property if we move this further down as we'll then have modified internal state.

Though to be honest I'm not sure this property is worth anything to us. In general if we get into this "corrupt" graph state we're probably just screwed.

I'll think about this some more; my earliest iterations had no errors and just panic'd everywhere since there isn't anything we can actually do if the graph is corrupt..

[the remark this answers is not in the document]
Mirko-von-Leipzig: I've added a comment at least for this.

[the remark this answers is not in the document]
Mirko-von-Leipzig: @bobbinth any thoughts?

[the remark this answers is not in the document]
bobbinth: I'm not too concered about the `clone()` here because the number of batches will always be relatively small (it is capped at 64, but also, it'll be a while till we get to 64). So, I'd probably prefer keeping the method atomic rather than avoid cloning.

PR #563, opened 2024-12-04
Implemented function for computing of old account states based on deltas

polydez:
Related to https://github.com/0xPolygonMiden/miden-node/issues/527

This function (`compute_old_account_states`) computes state for given accounts for the specified block number based on deltas for accounts stored in the database.

In current solution we also remove latest account details from `accounts` table (we calculate it the same way as for old account state). Fields related to public accounts were moved to special `public_accounts` table.

Calculation of account state was made in the way of returning cursor for particular data requested (basic fields, storage slot values, storage map values, fungible and non-fungible assets). This approach can help to calculate and deliver state incrementally.

But there is still a problem, how to return states of big accounts.

I can suggest several solutions:
1. Stream requested account state. Communicate between async context of gRPC streaming and sync context of deadpool interaction by using channels. Currently we use pool of SQL connections (pool size is 4x times of CPU number). This will require increasing the number of connections in the pool and using one connection for each stream.
3. Refactor more in order to query data in paging manner (by using LIMIT and OFFSET in SELECT queries), implement paging in endpoint(s) and query data by small portions. But this will require to run the same query on each page. I'm not sure, how expensive this could be.
4. Run query on first paging request and cache rows queried in-memory. Use cached data for building responses for remaining pages.

Each of the solutions has its own weakness for DoS attacks. In 1st case we might run out of free connections in the pool and stop serving new connections. 2nd just slower and will slow down the whole database access. 3rd could drain all free RAM.


crates/store/src/db/sql/mod.rs
```diff
@@ -190,6 +193,317 @@ pub fn select_accounts_by_ids(
...
+/// # Returns
+///
+/// Account states vector, or an error.
+#[allow(dead_code)] // TODO: remove once the function is used
+pub fn compute_old_account_states(
+    conn: &mut Connection,
+    account_ids: &[AccountId],
+    block_number: BlockNumber,
+) -> Result<Vec<Account>> {
+    fn find_account(
```
Mirko-von-Leipzig, line 207: Short doc comment here would be good.

crates/store/src/db/sql/mod.rs
```diff
@@ -190,6 +193,317 @@ pub fn select_accounts_by_ids(
...
+        match find_account(&mut accounts, account_id)?.get_mut().storage.entry(slot) {
+            Entry::Vacant(entry) => {
+                entry.insert(StorageSlot::Value(value));
+            },
+            Entry::Occupied(_) => {
+                return Err(DatabaseError::DataCorrupted(format!(
+                    "Duplicate storage slot: {slot}"
+                )));
+            },
+        }
```
Mirko-von-Leipzig, line 401: Could use `if <..>.insert(..).is_some() { return Err(..) }`. Unclear if its more legible though.

```suggestion
        let mut account = find_account(&mut accounts, account_id)?;
        if account.get_mut().storage.insert(slot, value).is_some() {
            return Err(DatabaseError::DataCorrupted(format!(
                "Duplicate storage slot: {slot}"
            )));
        }
```

crates/store/src/db/sql/mod.rs
```diff
@@ -190,6 +193,317 @@ pub fn select_accounts_by_ids(
...
+    }
+
+    let mut rows =
+        select_latest_nonces_stmt.query(params![Rc::clone(&account_ids), block_number])?;
+    while let Some(row) = rows.next()? {
+        let account_id: AccountId = row.get(0)?;
+        let nonce: u64 = row.get(1)?;
+        find_account(&mut accounts, account_id)?.get_mut().nonce =
+            Some(nonce.try_into().map_err(DatabaseError::DataCorrupted)?);
```
Mirko-von-Leipzig, line 381: nit: multi-line chaining from a stand-alone function is sometimes hard to read/grok:
```suggestion
        let nonce = row.get::<_, u64>(1)?.try_into().map_err(DatabaseError::DataCorrupted)?;
        let mut account = find_account(&mut accounts, account_id)?;
        account.get_mut().nonce = Some(nonce);

```

crates/store/src/db/sql/mod.rs
```diff
@@ -190,6 +193,317 @@ pub fn select_accounts_by_ids(
...
+            account_id
+        ",
+    )?;
+
+    struct AccountRecord {
+        code: AccountCode,
+        nonce: Option<Felt>,
+        storage: BTreeMap<u8, StorageSlot>,
+        assets: Vec<Asset>,
+    }
```
Mirko-von-Leipzig, line 350: Might be worth adding an `impl` to this to separate it from the record iteration?

```rust
impl AccountRecord {
    // We should start by reading the `nonce` then it doesn't have to be optional?
    fn new(id: AccountId, nonce: Nonce) -> Self {..}

    fn push_asset(&mut self, ...)
}
```

crates/store/src/db/sql/mod.rs
```diff
@@ -190,6 +193,317 @@ pub fn select_accounts_by_ids(
...
+        account_id: AccountId,
+    ) -> Result<OccupiedEntry<AccountId, AccountRecord>> {
+        let Entry::Occupied(found_account) = accounts.entry(account_id) else {
+            return Err(DatabaseError::DataCorrupted(format!(
+                "Account not found in DB: {account_id:x}"
+            )));
+        };
+
+        Ok(found_account)
+    }
```
Mirko-von-Leipzig, line 218: Maybe we should new-type accounts? Though I'm bias against free standing functions, so take this with a big pinch of 🧂 

```rust
#[derive(Default)]
struct Accounts(BTreeMap<AccountId, AccountRecord>);

impl Accounts {
    fn insert(...)

    // or maybe `get_or_err`
    fn try_get(&mut self) -> Result<&mut AccountRecord> {...}
}
```

crates/store/src/db/sql/mod.rs
```diff
@@ -190,6 +193,317 @@ pub fn select_accounts_by_ids(
...
+        })?;
+
+        let account = Account::from_parts(
+            account_id.try_into()?,
+            vault,
+            storage,
+            record.code,
+            record.nonce.ok_or(DatabaseError::DataCorrupted(format!(
+                "Missing nonce for account: {account_id:x}"
+            )))?,
```
Mirko-von-Leipzig, line 498: If we start by iterating over the nonces the they don't need to be optional.

crates/store/src/db/sql/mod.rs
```diff
@@ -190,6 +193,317 @@ pub fn select_accounts_by_ids(
...
+    for (account_id, record) in accounts {
+        let slots: Vec<_> = record
+            .storage
+            .into_iter()
+            .enumerate()
+            .map(|(expected_slot, (slot, value))| {
+                if expected_slot != slot as usize {
+                    return Err(DatabaseError::DataCorrupted(format!(
+                        "Missing value for storage slot {expected_slot}, got {slot}"
+                    )));
```
Mirko-von-Leipzig, line 477: Is there value in expanding `DataCorrupted` with variants for each possible corruption? Probably not 🤔

Mirko-von-Leipzig (approved):
Mostly just style/nit suggestions :) LGTM.

crates/store/src/db/sql/mod.rs
```diff
@@ -130,7 +133,7 @@ pub fn select_accounts_by_block_range(
     Ok(result)
 }
 
-/// Select the latest account details by account id from the DB using the given [Connection].
+/// Select the latest account details by account ID from the DB using the given [Connection].
```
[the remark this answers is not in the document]
polydez: I have no previous experience with spell-checking on CI. This should work, if it's easy to add new words to a dictionary. And if it won't require disabling it for particular comments.

crates/store/src/db/sql/utils.rs
```diff
@@ -42,13 +42,14 @@ macro_rules! subst {
...
 ///
 /// # Usage:
 ///
-/// ```
-/// insert_sql!(users { id, first_name, last_name, age });
-/// ```
+/// `insert_sql!(users { id, first_name, last_name, age });`
+///
 /// which generates:
 /// "INSERT INTO users (id, first_name, last_name, age) VALUES (?, ?, ?, ?)"
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: TIL; okay yeah probably not worth fiddling around just for that :)

crates/store/src/db/sql/mod.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Me neither; [this](https://github.com/drahnr/cargo-spellcheck/tree/master) looks promising? But not really important at this stage :)

crates/store/src/db/tests.rs
```diff
@@ -471,6 +474,121 @@ fn test_sql_public_account_details() {
     assert_eq!(read_delta, Some(delta2));
 }
 
+#[test]
+fn test_sql_public_account_details_for_old_block() {
```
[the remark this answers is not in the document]
bobbinth: There is no good reason for this, and we are not always being consistent here (most of the time we do use `test_` prefix but not always).

I'm totally fine with removing `test_` prefixes if someone wants to rename them.

bobbinth:
> This function (`compute_old_account_states`) computes state for given accounts for the specified block number based on deltas for accounts stored in the database. Since it takes into account all deltas from the creation of account up to the current block, if account has a lot of changes, it might be more expensive than applying deltas backward from the latest account state to the specified block, so it might be a subject for future improvement.

Are you thinking of this as a temporary solution? Because I agree that computing account states based on deltas from account creation is probably not a good solution.

And if it is a temporary solution - is it actually worth it? Maybe we should go for a longer-terms solution.

In general, as I mentioned in one other comment, let's look at the whole account data storage/retrieval area more holistically. Let's list the use-cases we are trying to solve for and then see what solution would address them well.

crates/store/src/db/sql/mod.rs
polydez: No, we actually start by getting codes from accounts and crate `AccountRecord`s with only code field filled.

crates/store/src/db/sql/mod.rs
polydez: No, I don't think we will process any data corruption cases differently from any other. :)

crates/rpc/src/server/api.rs
```diff
@@ -188,7 +188,7 @@ impl api_server::Api for RpcApi {
         self.block_producer.clone().submit_proven_transaction(request).await
     }
 
-    /// Returns details for public (public) account by id.
+    /// Returns details for public account by ID.
```
igamigo, line 191: I think this is a bit misleading. This method still works for non-public accounts, right? It just doesn't return details for them (ie, it just returns the stored hash). There are some other related functions that have the same problem IIRC.

crates/store/src/db/sql/mod.rs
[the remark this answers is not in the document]
igamigo: I'd agree with this as well. There have already been quite a few of those types of PRs.

crates/store/src/db/sql/mod.rs
```diff
@@ -190,6 +191,322 @@ pub fn select_accounts_by_ids(
...
+///
+/// # Returns
+///
+/// Account states vector, or an error.
+#[allow(dead_code)] // TODO: remove once the function is used
+pub fn compute_old_account_states(
+    conn: &mut Connection,
+    account_ids: &[AccountId],
+    block_number: BlockNumber,
+) -> Result<Vec<Account>> {
```
igamigo, line 204: Should this check that account IDs are public only?

igamigo (commented):
This might have already been discussed (apologies if so), but considering the "usability" of account states decreases as they move further and further from the chain tip, would it make sense (if it is even possible) to retrieve the latest state and subtract deltas instead of building them up from 0?

crates/store/src/db/sql/mod.rs
```diff
@@ -200,6 +198,239 @@ pub fn select_accounts_by_ids(
     Ok(result)
 }
 
+/// Computes public account states for the particular block number using the given [Connection].
+///
+/// # Returns
+///
+/// Account states vector, or an error.
+#[allow(dead_code)] // TODO: remove once the function is used
```
Mirko-von-Leipzig, line 206: Or remove if no longer unused.

```suggestion
#[expect(dead_code)] // TODO: remove once the function is used
```

crates/store/src/db/sql/mod.rs
```diff
@@ -1185,6 +1388,10 @@ pub fn get_state_sync(
         select_block_header_by_block_num(conn, notes.first().map(|note| note.block_num))?
             .ok_or(StateSyncError::EmptyBlockHeadersTable)?;
 
+    // TODO: What if account was also changed after requested block number? In this case we will
+    //       miss the update (since `accounts` table stores only latest account states,
+    //       the corresponding record in the table will be updated to the state beyond the requested
+    //       block range).
```
Mirko-von-Leipzig, line 1394: Sorry, vacation brain rot occurred. What is this saying?

If the accounts table only stores the latest state then how/why do we allow querying by a block range?

polydez: It's might be confusing, but `accounts` table does store the latest state of an account. If the account is updated, its state is replaced in the `accounts` table. I re-read the code now and now I'm thinking that I probably misunderstood the idea of state sync and this behavior is expected: https://github.com/0xPolygonMiden/miden-node/blob/b36a0fdba23edcac654c7a45aaf528f26e3efde3/crates/store/src/state.rs#L500-L501

@bobbinth what do you think?

crates/store/src/db/sql/mod.rs
```diff
@@ -200,6 +198,239 @@ pub fn select_accounts_by_ids(
...
+        WHERE
+            new.account_id IN rarray(?1) AND
+            new.block_num = (
+                SELECT MIN(block_num)
+                FROM account_storage_slot_updates
+                WHERE
+                    (block_num BETWEEN ?2 + 1 AND ?3) AND
+                    new.account_id = account_id AND
+                    new.slot = slot
+            )",
```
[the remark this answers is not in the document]
polydez: The goal of this query is to return old values for the ones updated after the given block number. We don't want to return all the latest values here (but this could work for our previous approach, when we constracted storage from scratch, filling it with the most recent values up to the given block).

crates/store/src/db/sql/mod.rs
```diff
@@ -200,6 +198,239 @@ pub fn select_accounts_by_ids(
...
 }
 
+/// Computes public account states for the particular block number using the given [Connection].
+///
+/// # Returns
+///
+/// Account states vector, or an error.
+#[cfg_attr(not(test), expect(dead_code))] // TODO: remove once the function is used
+pub fn compute_old_account_states(
+    conn: &mut Connection,
```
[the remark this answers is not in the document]
polydez: Good point, thank you!

bobbinth (commented):
Thank you! Looks good. I left a couple of comments inline - but I do wonder if the current approach is going to be very computationally intensive.

Taking a step back, I wonder if we should break this up into a couple of endpoints. What the client is interested in is getting to a state of a public account not too far from the chain tip and there are really 3 parts to it:

1. Get account header + storage header.
2. Get account code.
3. Get account vault.
4. Get account storage maps.

The first one we can probably just cache in the database for every update of a public account for some number of recent blocks (e.g., 100 - 200). This will make it very fast to retrieve and will probably not result in too much extra storage pressure.

The second one could be handled pretty easily by having a separate table for storing account code keyed by code commitment. This table could be periodically pruned to get rid of obsolete code entries - but I would imaging that'll happen rather rarely.

For the vault, if the vault is small enough, we could return the entire vault. The size of the vault should be pretty easy to track for each of the account updates (it is just the number of assets in the vault). If the vault is big, we'd need to come up with a different strategy - e.g., either sending vault deltas or somehow paginating through assets.

For storage maps, we could try to apply the same strategy as for the vault - i.e., if the map is small enough, just send all key-value pairs, otherwise either send deltas or do some kind of pagination over key-value pairs.

Sending vault/storage map deltas should be pretty simple - and we kind of already do that. How to do pagination is not super clear to me yet.

Overall, before knowing if the above approach makes sense, we'll need to sketch out how would it affect the existing endpoints. I think these endpoints are:

- `SyncState` - maybe we should return account header for all public accounts?
- `GetAccountStateDelta` - would we still need this endpoint?
- `GetAccountDetails` - should we change this to be about getting vault and storage maps (either via deltas or pagination)?
- `GetAccountProofs` - should we modify the map responses to send just key-value pairs if the maps are small enough? Or maybe we should transform this into a more general `GetAccountHeader` endpoint which could also optionally return account proofs?

polydez:
@bobbinth thank you so much for in-deep review! Yes, this idea might help to make things simpler as well, as the the solution might work faster. 

> Overall, before knowing if the above approach makes sense, we'll need to sketch out how would it affect the existing endpoints. I think these endpoints are:
> 
> * `SyncState` - maybe we should return account header for all public accounts?
> * `GetAccountStateDelta` - would we still need this endpoint?
> * `GetAccountDetails` - should we change this to be about getting vault and storage maps (either via deltas or pagination)?
> * `GetAccountProofs` - should we modify the map responses to send just key-value pairs if the maps are small enough? Or maybe we should transform this into a more general `GetAccountHeader` endpoint which could also optionally return account proofs?

It's unclear for me, which endpoints work good for consumers and which ones should be improved or removed. I was wondering, if @igamigo could give us more context regarding these endpoints and how useful (or not) they are from a client's perspective and what we should improve.

igamigo:
> It's unclear for me, which endpoints work good for consumers and which ones should be improved or removed. I was wondering, if @igamigo could give us more context regarding these endpoints and how useful (or not) they are from a client's perspective and what we should improve.

I think in general [this](https://github.com/0xPolygonMiden/miden-node/issues/527#issuecomment-2529201687) is still the case. In terms of whether the endpoints that Bobbin mentions are useful or not, we use all of them except `GetAccountStateDelta`, as mentioned in the linked comment. As for the others, all of them work fine right now, except for the limitation placed on `GetAccountProofs` since we cannot get past states.

The way syncing works right now on the client is that we get the response, which includes the `AccountSummary` for any account. If the account is public, we end up using the `GetAccountDetails` endpoint. This is probably not future proof because public accounts can obviously get pretty big. I think we always care about the latest state of the public account here, but Bobbin suggested refactoring this to allow for non-current states. I don't think this is necessary now from the client's standpoint, but also he might have some other use case in mind.

So, in terms of what we don't have now but would be nice to have, I guess it boils down to allowing for non-current states for `GetAccountProofs` and also having a way to (incrementally) get a public account's state because we do want to keep public accounts in sync as well (though maybe this process should be reworked entirely). If any of this is not clear enough feel free to let me know.

polydez:
Finally, I end up with solution for getting public account state for any recent block number, which should be easy to implement (and we already have implemented the most of it in previous commits in this branch).

Currently, on block application, we save public account's delta to the database, load last public account details (deserialize `Account` from binary), apply delta to the account details and save resulting account back to the DB.

We can get rid of keeping full public account details and use just deltas we already store in order to construct account details for any recent block number including latest block.

We don't store account code changes yet, and `AccountDelta` structure in miden-base doesn't include such field as well. I think, we don't need it for now, because currently we don't support account code updates. But in future, solution proposed by @bobbinth (storing of account's code as `code_hash → code` key-value), should work good.

Constructing of account parts from deltas has advantage of good support for gRPC streaming of returning data. We will be able to query, iterate and drop unneeded data from the memory during streaming.

We also reduce disk space needed for storing accounts due to storing only changes, not full latest state. And block applying will work a bit faster.

Construction of account from deltas might work slower and more resource-consuming than just loading BLOB from the database. This is the only disadvantage of this approach I can see now. But it might partly be eliminated by network bandwidth.

In order to save space and reduce computations needed for constructing of frequently changed accounts, we can add background task for merging all updates in old deltas (older than last $N$ blocks) stored in the database into "consolidated" updates: for fungible assets it will sum changes for the whole range, for non-fungible ones and storage it will just remove obsolete operations/data.

In this case, we won't be able to return account state or deltas for block number older than last $N$ blocks so clients will need to download full account details instead of just downloading and applying deltas if they need to synchronize accounts from states older than $tip - N + 1$. But we probably won't need to do it. At least, at nearest future.

@bobbinth, @Mirko-von-Leipzig, what do you think?

> * SyncState - maybe we should return account header for all public accounts?

Yes, this is good idea! It will be useful, if we provide separate endpoints or field map for choosing what to get: code/vault/storage.

> * GetAccountStateDelta - would we still need this endpoint?
> * GetAccountDetails - should we change this to be about getting vault and storage maps (either via deltas or pagination)?

We can get rid of `GetAccountStateDelta` endpoint, if we will be able to return deltas in `GetAccountDetails`.

If number of updated entities is smaller than final number of entities, and the block is recent (we store delta for it), it might be better to return just delta instead of final state. Previously I had opinion that it would be better to have different endpoints for this, but now I think that the store has more information of what would take less space/traffic in particular case, comparing to the client. This might require more complex processing of server's response, but we can probably implement it under the hood of our SDK.

bobbinth:
@polydez - thank you! Could you provide a bit more detail about the structure (e.g., how the tables and endpoints could look like)?

polydez:
> @polydez - thank you! Could you provide a bit more detail about the structure (e.g., how the tables and endpoints could look like)?

I'm not going to change database structure significantly, generally it will be the same: https://github.com/0xPolygonMiden/miden-node/blob/polydez-compute-old-acc-states/crates/store/src/db/migrations/001-init.sql#L46-L118

Several changes we should do IMO:
1. Remove `details` field from `accounts` table since we will be able to construct details from deltas.
2. Probably add `account_codes` table:
```sql
CREATE TABLE
    account_codes
(
    code_hash BLOB NOT NULL,
    code BLOB NOT NULL
) STRICT, WITHOUT ROWID
```
3. Add field `code_hash` to `account_deltas` table referring current code from `account_codes` table.

Regarding endpoints I don't have design yet. We can't return several streams in one response due to limitations of gRPC, but we still have two options how we can implement gRPC streaming in our endpoints:
1. Make separate endpoints (for getting fungible assets, non-fungible assets, storage slots, storage maps), some of them could be easily merged into single one.
2. Stream data by using single endpoint, but encode different records differently. I'm not sure protobuf is good for this, but we still can encode as enums by winter-utils:
```rust
enum Record {
    Storage {
        // storage fields
    },
    StorageMap {
        // storage map fields
    },
    FungibleAsset {
        // fungible asset fields
    },
    NonFungibleAsset {
        // non-fungible asset fields
    },
}
```

Mirko-von-Leipzig:
> @Mirko-von-Leipzig, what do you think?

I'm also interested in more details about the structure. The way I've imagined this is something like:

We guarantee that one can always sync from within the latest `N` blocks from the tip. This means we can throw away deltas older than `N` blocks. From a high level design, this implies we should:

1. Somehow store `N` deltas, and
2. all older deltas are aggregated as the state at `block = tip - N`

So far I think this aligns with your suggestion.

For implementation, I think you're suggesting we essentially change the account state (2) from `BLOB` to simply the same as `delta` - since they're essentially the same thing in a way. This makes sense to me, so long as deltas can be partially updated? But we can iterate on this design, I think so long as the top level picture makes sense first.

Would we then have two tables? One for state (2), and another for storing deltas (1)? These then get pulled in as needed?

--------------------

The way I thought about this for implementation is something like:

1. Store all deltas to disk or as blobs by block number.
2. On startup load the last `N` deltas from disk/db into in-memory FIFO cache.
3. On each `apply_block`, pop the oldest delta, merge its diff into database and push in the new delta.
4. Also write the delta to disk as per (1)

An accounts latest state can then be retrieved by fetching the state from db and merging in the deltas from memory.

No background task required - database and FIFO queue are always in sync, and we only merge one delta per block to db (same as now).

If we make the FIFO deltas `Arc<Delta>`, then current gRPC streams can grab a cheap copy, so that even if the delta gets dropped from queue, its still got the data to stream out. This prevents (most) issues with slow gRPC readers (of course at the risk of memory balooning DoS - but we have to cater for that anyway).

polydez:
> We guarantee that one can always sync from within the latest `N` blocks from the tip. This means we can throw away deltas older than `N` blocks. From a high level design, this implies we should:
> 
> 1. Somehow store `N` deltas, and
> 2. all older deltas are aggregated as the state at `block = tip - N`
> 
> So far I think this aligns with your suggestion.

Yes, this is right.

> For implementation, I think you're suggesting we essentially change the account state (2) from `BLOB` to simply the same as `delta` - since they're essentially the same thing in a way. This makes sense to me, so long as deltas can be partially updated? 
> Would we then have two tables? One for state (2), and another for storing deltas (1)? These then get pulled in as needed?

I suggest to remove `details` (BLOB) field from `accounts` and use only deltas which we already store in tables: `account_deltas`, `account_storage_slot_updates`, `account_storage_map_updates`, `account_fungible_asset_deltas` and `account_non_fungible_asset_updates`.

> The way I thought about this for implementation is something like:
> 
> 1. Store all deltas to disk or as blobs by block number.
> 2. On startup load the last `N` deltas from disk/db into in-memory FIFO cache.
> 3. On each `apply_block`, pop the oldest delta, merge its diff into database and push in the new delta.
> 4. Also write the delta to disk as per (1)

This is a good alternative to background process! The only thing I would like to improve here, is that we could merge deltas _in-cito_ in the database. 

> An accounts latest state can then be retrieved by fetching the state from db and merging in the deltas from memory.

I think, it would be easier to just construct it from the DB, as we did it before (my first attempt to implement computing of old account states based on deltas).

Mirko-von-Leipzig:
@polydez could you please give an overview of your new solution? I'm finding it pretty hard to follow without knowing the end goal.

polydez:
@Mirko-von-Leipzig I've updated description of this PR, feel free to ask if something is unclear from it.

Mirko-von-Leipzig:
Closing as our direction is unclear and depends on the outcome of #722.

PR #568, opened 2024-12-09
chore: remove `test_` prefix stutter

Mirko-von-Leipzig:
Removes the `test_` prefix applied to many of our test cases.

Tests are already annotated with `#[test]` and are usually inside a `mod tests` so also prefixing each test function is largely redundant.

Briefly discussed [here](https://github.com/0xPolygonMiden/miden-node/pull/563#discussion_r1872856703).


bobbinth (approved):
Looks good! Thank you!

PR #570, opened 2024-12-09
feat(block-producer): remove deep clone of tx data

### LINKED ISSUE #531: Batch building should use arcs

Mirko-von-Leipzig:
This was previously added in #544 but was mistakenly dropped by a rebase.

Closes #531 (again).

### DIFF
--- a/crates/block-producer/src/batch_builder/batch.rs
+++ b/crates/block-producer/src/batch_builder/batch.rs
@@ -70,9 +70,13 @@ impl TransactionBatch {
     // CONSTRUCTORS
     // --------------------------------------------------------------------------------------------
 
-    /// Returns a new [TransactionBatch] instantiated from the provided vector of proven
-    /// transactions. If a map of unauthenticated notes found in the store is provided, it is used
-    /// for transforming unauthenticated notes into authenticated notes.
+    /// Returns a new [TransactionBatch] built from the provided transactions. If a map of
+    /// unauthenticated notes found in the store is provided, it is used for transforming
+    /// unauthenticated notes into authenticated notes.
+    ///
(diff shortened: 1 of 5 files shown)


crates/block-producer/src/batch_builder/batch.rs
```diff
@@ -81,17 +81,21 @@ impl TransactionBatch {
...
-    pub fn new(
-        txs: Vec<ProvenTransaction>,
+    pub fn new<'a, I>(
+        txs: impl IntoIterator<Item = &'a ProvenTransaction, IntoIter = I>,
         found_unauthenticated_notes: NoteAuthenticationInfo,
-    ) -> Result<Self, BuildBatchError> {
-        let id = Self::compute_id(&txs);
+    ) -> Result<Self, BuildBatchError>
+    where
+        I: Iterator<Item = &'a ProvenTransaction> + Clone,
```
bobbinth, line 89: Doc comments for this function need to be updated (e.g., line 73 has a reference to a vector of proven transactions). I would also probably write a brief explanation why the types are defined as they are (my understanding is that we can cheaply clone the iterator we get from `txs.into_iter()`, right?).

bobbinth (approved):

crates/block-producer/src/batch_builder/batch.rs
Mirko-von-Leipzig: Correct; we're essentially looking to get `T: Iterator<Item = ProvenTransaction> + Clone` which happens to be fulfilled by the above.

PR #571, opened 2024-12-09
refactor(block-producer): remove Store traits

### LINKED ISSUE #196: Remove block producer's Store trait
The `Store` trait abstracts over the gRPC server, the main utility of this trait is to implement tests. However, this is not necessary, since tonic already has a strategy to mock the servers for testing, here is an example:

https://github.com/hyperium/tonic/blob/master/examples/src/mock/mock.rs

### LINKED ISSUE #565: [block-producer]: merge store traits
> I'm not sure we need to have `ApplyBlock` as a separate trait. Maybe worth moving `apply_block()` method to the `Store` trait?

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-node/pull/562#discussion_r1870441700_

Mirko-von-Leipzig:
This PR removes the block producer's `Store` and `ApplyBlock` traits.

Instead the methods are moved directly into the `DefaultStore` struct, which has also been renamed to a more appropriate `StoreClient`.

The traits were required to facilitate mocking as part of tests. Most of these usages were removed as part of the mempool work, which allowed these traits to largely just disappear.

There are still some remnants lurking (`MockStoreSuccess` et al), but these require larger refactors to the underlying tests. 

In general, we should avoid mocks where possible as its very easy for things to desync with reality. And indeed some of the above tests aren't actually testing the unit-under-test but rather the test setup code. Ideally we would have an actual in-memory store running for each test however this is quite a lot of work. A potentially simpler option is to implement the gRPC server for tests (skipping the database). 

Closes #196 and closes #565.

### DIFF
--- a/crates/block-producer/src/store/mod.rs
+++ b/crates/block-producer/src/store/mod.rs
@@ -113,15 +113,18 @@ impl TryFrom<GetTransactionInputsResponse> for TransactionInputs {
     }
 }
 
-// DEFAULT STORE IMPLEMENTATION
+// STORE CLIENT
 // ================================================================================================
 
+/// Interface to the store's gRPC API.
+///
+/// Essentially just a thin wrapper around the generated gRPC client which improves type safety.
 #[derive(Clone)]
-pub struct DefaultStore {
+pub struct StoreClient {
     store: store_client::ApiClient<Channel>,
 }
 
-impl DefaultStore {
+impl StoreClient {
     /// TODO: this should probably take store connection string and create a connection internally
     pub fn new(store: store_client::ApiClient<Channel>) -> Self {
         Self { store }
--- a/crates/block-producer/src/store/mod.rs
+++ b/crates/block-producer/src/store/mod.rs
@@ -36,7 +36,7 @@ use crate::{block::BlockInputs, COMPONENT};
 // ================================================================================================
 
 #[async_trait]
-pub trait Store: ApplyBlock {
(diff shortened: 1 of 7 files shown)


crates/block-producer/src/store/mod.rs
```diff
@@ -230,7 +187,7 @@ impl Store for DefaultStore {
         Ok(tx_inputs)
     }
 
-    async fn get_block_inputs(
+    pub async fn get_block_inputs(
```
bobbinth, line 190: Should we add `instrument` to this method as well? (not sure why it wasn't instrumented before).

bobbinth (approved):
Looks good! Thank you! I left just one small comment inline.

Also, should we create an issue for a bigger refactoring of the tests?

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: I was thinking we'd do a pass over all instrumentation as part of the monitoring work. But I may as well instrument this while I'm here :) Thanks.

Mirko-von-Leipzig:
> Also, should we create an issue for a bigger refactoring of the tests?

#573

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: Tackled in #574

PR #572, opened 2024-12-09
refactor(block-producer): order batches within a block

### LINKED ISSUE #397: Sort batches within a block by dependency order
@bobbinth:

At the block level, output notes must come before corresponding unauthenticated input notes. That is, if `batch1` has an unauthenticated input note `A`, this note must be an output note of `batch0` (e.g., having it be an output note of `batch2` should not be allowed).

Mirko-von-Leipzig:
As mentioned in this [comment](https://github.com/0xPolygonMiden/miden-node/issues/397#issuecomment-2529077185), this PR makes it clear that batches within a block have a given order. This intention signaling is improved by using a vector instead of a map (with accidental ordering). As alluded to in [this comment](https://github.com/0xPolygonMiden/miden-node/issues/397) from #397.

> The [current](https://github.com/0xPolygonMiden/miden-node/blob/0f6b69af58189755d0ec9b13f994d3ef75317485/crates/block-producer/src/mempool/batch_graph.rs#L220-L247) batch selection strategy just selects an arbitrary available root batch. A root batch by definition is one who's parents have already been selected previously; or in other words dependency order is enforced within the selection process.
>
> The set of batches is then [consumed in order](https://github.com/0xPolygonMiden/miden-node/blob/0f6b69af58189755d0ec9b13f994d3ef75317485/crates/block-producer/src/block_builder/mod.rs#L74-L75) by the block builder. Though it does use a BTreeMap which makes this a bit opaque (the keys form a sort of FIFO so ordering is kept), and I guess more or less implies that the ordering is accidental and not necessarily intentional.
>
> This issue is technically complete but I'll create a PR changing the BTreeMap to a Vec so that this ordering is made more clear.

Closes #397.

### DIFF
--- a/crates/block-producer/src/block_builder/mod.rs
+++ b/crates/block-producer/src/block_builder/mod.rs
@@ -72,7 +72,7 @@ impl BlockBuilder {
             interval.tick().await;
 
             let (block_number, batches) = mempool.lock().await.select_block();
-            let batches = batches.into_values().collect::<Vec<_>>();
+            let batches = batches.into_iter().map(|(_, batch)| batch).collect::<Vec<_>>();
 
             let mut result = self.build_block(&batches).await;
             let proving_duration = rand::thread_rng().gen_range(self.simulated_proof_time.clone());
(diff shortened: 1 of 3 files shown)


bobbinth (approved):

PR #574, opened 2024-12-10
refactor(block-producer): cleanup StoreClient

Mirko-von-Leipzig:
This PR instruments the `StoreClient` API (follow-up task from #571), and also consolidates the errors into a single `StoreError` variant.

We should also create an issue for adapting the error guidelines discussed in https://github.com/0xPolygonMiden/miden-base/discussions/966.

### DIFF
--- a/crates/block-producer/src/block.rs
+++ b/crates/block-producer/src/block.rs
@@ -13,8 +13,6 @@ use miden_objects::{
     BlockHeader, Digest,
 };
 
-use crate::store::BlockInputsError;
-
 // BLOCK INPUTS
 // ================================================================================================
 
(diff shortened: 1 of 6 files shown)


Mirko-von-Leipzig:
Meta: what constitutes a changelog entry? Or rather, what do we consider our public API? 

From my current perspective its the gRPC API and the `miden-node` and `miden-faucet` binaries. However if we consider all crates then this becomes a much broader surface area.

bobbinth (approved):
Looks good! Thank you!

> We should also create an issue for adapting the error guidelines discussed in https://github.com/0xPolygonMiden/miden-base/discussions/966.

Yes, let's do it.

bobbinth:
> Meta: what constitutes a changelog entry? Or rather, what do we consider our public API?
> 
> From my current perspective its the gRPC API and the `miden-node` and `miden-faucet` binaries. However if we consider all crates then this becomes a much broader surface area.

I would probably add some non-trivial things to the changelog as well (under "fixes" or "changes" sections). For example, mempool refactoring should be in the changelog even though it is not really "public".

PR #576, opened 2024-12-10
chore: remove miette dependency

Mirko-von-Leipzig:
Remove the final remnants of `miette` by replacing its usage in `proto::build.rs` with `anyhow`.

Motivation is to minimize our direct dependencies and there is not really an argument for `miette` as we don't have any source code errors to display.


bobbinth (approved):
Looks good! Thank you!

Issue #519, opened 2024-10-22
[block-producer]: follow-up tasks


Mirko-von-Leipzig:
Follow-up tasks from #515. Some of these require work in `miden-base`.

- [x] ~Create and use `ProposedBatch` in `miden-base` ([source](https://github.com/0xPolygonMiden/miden-node/pull/515#discussion_r1807946778))~ Tracked in #583 
- [x] Use a more correct `BatchId` ([source](https://github.com/0xPolygonMiden/miden-node/pull/515#discussion_r1807947418))
- [x] ~Create and use `ProposedBlock` in `miden-base` ([source](https://github.com/0xPolygonMiden/miden-node/pull/515#discussion_r1807948590))~ Tracked in #583 
- [x] Loosen block selection strategy ([source](https://github.com/0xPolygonMiden/miden-node/pull/515#discussion_r1807964589))
- [x] #523
- [x] Improve block/batch simulation times ([source](https://github.com/0xPolygonMiden/miden-node/pull/515#discussion_r1807966960) and [source](https://github.com/0xPolygonMiden/miden-node/pull/515#discussion_r1807967683))
- [x] Amend `GetTransactionInputs` to found instead of missing notes ([source](https://github.com/0xPolygonMiden/miden-node/pull/515#discussion_r1809596677))
- [x] Instrument mempool
- [x] Wrap `AuthenticatedTransaction::inner` in an `Arc` for cheaper copies.
- [x] Move graph tests into a separate `tests.rs` file.
- [x] Move batch limits from `TransactionBatch` into the mempool's batch selection strategy
- [x] Move block limits from the block builder into the mempool's block selection strategy
- [x] Remove transactions from the error variant of batch builder error and in general revisit all errors there.
- [x] #564 
- [x] #565
- [x] #566

bobbinth:
Nice! I added this issue to the current milestone, though probably not everything will end up there (e.g., `ProposedBatch` and `ProposedBlock` is definitely for the next milestone).

PR #578, opened 2024-12-10
feat(block-producer): instrument mempool

Mirko-von-Leipzig:
This PR adds tracing instrumentation to the mempool.

In addition, this PR also bumps `tracing` to v0.14 which now allows using constants for instrument targets. Before this we've had the unfortunate situation of using `target = COMPONENT` when emitting events, but having to manually set `target = miden-node-<xxx>` when using the `#[instrument]` macro. This PR fixes this; though I think our behaviour here could be improved.

Blocked by #577 as it refactors batch ID, and by #574 which added instrumentation which will require a minor tweak or two in this PR.

Closes a task in #519


bobbinth (approved):
Very nice! Thank you!

PR #579, opened 2024-12-10
chore: move new changelog entries to unreleased

Mirko-von-Leipzig:
I've been appending them to the wrong section like a dummy :)

Wondering if I should add a CI check just to ensure there is always an unreleased section at the top as an enhancement to the changelog checker so I don't confuse myself.


bobbinth (approved):
Looks good! Thank you!

PR #580, opened 2024-12-11
perf: When using rand::random(), it's recommended to explicitly ensure that its seed source is sufficiently secure. In production environments, it may be necessary to switch to rand::rngs::OsRng to provide stronger randomness.

RiceChuan:
When using rand::random(), it's recommended to explicitly ensure that its seed source is sufficiently secure. In production environments, it may be necessary to switch to rand::rngs::OsRng to provide stronger randomness.

### DIFF
--- a/bin/faucet/src/main.rs
+++ b/bin/faucet/src/main.rs
@@ -138,7 +138,7 @@ async fn main() -> anyhow::Result<()> {
             let current_dir =
                 std::env::current_dir().context("Failed to open current directory")?;
 
-            let mut rng = ChaCha20Rng::from_seed(rand::random());
+            let mut rng = ChaCha20Rng::from_seed(rand::rngs::OsRng);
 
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig:
Thanks, could you make this change from and into our `next` branch instead? That's our development branch.

Could you provide a link for this recommendation? I understand both to be equivalent:

From [rand::random](https://docs.rs/rand/latest/rand/fn.random.html):
> This is simply a shortcut for thread_rng().gen()

and then following [ThreadRng](https://docs.rs/rand/latest/rand/rngs/struct.ThreadRng.html):
> ThreadRng uses the same PRNG as [StdRng](https://docs.rs/rand/latest/rand/rngs/struct.StdRng.html) for security and performance and is automatically seeded from [OsRng](https://docs.rs/rand/latest/rand/rngs/struct.OsRng.html).

Or are you simply suggesting we make it more explicit and don't rely on `rand::random` using `OsRng`?

PR #588, opened 2024-12-18
feat(block-producer): catch up missed blocks instead of skipping

### LINKED ISSUE #523: [block-producer]: missed block strategy
Options: https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html

`Burst` will let us catch up lost blocks, but will have the least predictable block times while this is happening. Essentially blocks are produced as fast as possible.

`Skip` will miss a block but keep the block publishing moment predictable. I'm not sure this matters.

`Delay` means we only have a single slow block, and that block times are kept consistent.

This is assuming the problem was temporary.

Mirko-von-Leipzig:
This PR changes the behaviour when a block tick is missed from `Delay` to `Burst`.

#523 discusses this is slightly more detail:

> Burst will let us catch up lost blocks, but will have the least predictable block times while this is happening. Essentially blocks are produced as fast as possible.
>
> Skip will miss a block but keep the block publishing moment predictable. I'm not sure this matters.
>
> Delay means we only have a single slow block, and that block times are kept consistent.

`Burst` seems fine, and means we make the best attempt at the desired block interval on average.

Closed #523

### DIFF
--- a/crates/block-producer/src/block_builder/mod.rs
+++ b/crates/block-producer/src/block_builder/mod.rs
@@ -63,7 +63,7 @@ impl BlockBuilder {
         );
 
         let mut interval = tokio::time::interval(self.block_interval);
-        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Burst);
 
         loop {
             interval.tick().await;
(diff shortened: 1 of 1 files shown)


crates/block-producer/src/block_builder/mod.rs
```diff
@@ -63,7 +63,7 @@ impl BlockBuilder {
         );
 
         let mut interval = tokio::time::interval(self.block_interval);
-        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Burst);
```
bobbinth, line 66: Could we add a comment explaining what behavior this results in?

bobbinth (approved):
Looks good! Thank you! I left one doc-related comment inline.

PR #589, opened 2024-12-18
feat(block-producer): re-add batch inputs

### LINKED ISSUE #566: [block-producer]: authenticate notes in batch builder
Related to one of the previous comments, I think this may need to be reverted in a future PR - do we have an issue for this?

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-node/pull/562#discussion_r1870454811_

Mirko-von-Leipzig:
This was removed as part of the mempool feature, but should not have been.

I've renamed the function `get_batch_inputs` and I'm wondering if we shouldn't rename the proto definition and domain object similarly?

I'm assuming there might be more input fields coming in the future so having it be more specific is ok?

Closes #566

### DIFF
--- a/crates/block-producer/src/batch_builder/mod.rs
+++ b/crates/block-producer/src/batch_builder/mod.rs
@@ -1,13 +1,14 @@
 use std::{num::NonZeroUsize, ops::Range, time::Duration};
 
 use batch::BatchId;
+use miden_node_proto::domain::notes::NoteAuthenticationInfo;
 use rand::Rng;
 use tokio::{task::JoinSet, time};
 use tracing::{debug, info, instrument, Span};
 
 use crate::{
-    domain::transaction::AuthenticatedTransaction, mempool::SharedMempool, COMPONENT,
-    SERVER_BUILD_BATCH_FREQUENCY,
+    domain::transaction::AuthenticatedTransaction, mempool::SharedMempool, store::StoreClient,
+    COMPONENT, SERVER_BUILD_BATCH_FREQUENCY,
 };
 
 pub mod batch;
@@ -54,7 +55,7 @@ impl BatchBuilder {
     /// A pool of batch-proving workers is spawned, which are fed new batch jobs periodically.
     /// A batch is skipped if there are no available workers, or if there are no transactions
     /// available to batch.
-    pub async fn run(self, mempool: SharedMempool) {
+    pub async fn run(self, mempool: SharedMempool, store: StoreClient) {
         assert!(
             self.failure_rate < 1.0 && self.failure_rate.is_sign_positive(),
             "Failure rate must be a percentage"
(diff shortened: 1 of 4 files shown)


crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -210,11 +214,18 @@ impl WorkerPool {
...
                     tracing::debug!("Begin proving batch.");
 
-                    let batch = Self::build_batch(transactions).map_err(|err| (id, err))?;
+                    let inputs = store
+                        .get_batch_inputs(
+                            transactions.iter().flat_map(|tx| tx.unauthenticated_notes()),
+                        )
+                        .await
+                        .map_err(|err| (id, err.into()))?;
+                    let batch = Self::build_batch(transactions, inputs).map_err(|err| (id, err))?;
```
bobbinth, line 228: I forget now, but do we do something similar when constructing a block? Basically, curious what happens if we while we are building a batch one of the unauthenticated notes becomes authenticated (i.e., because the block in which it gets recorded gets committed right after this code is executed).

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -210,11 +214,18 @@ impl WorkerPool {
                 //
                 // Note: Rng::gen rolls between [0, 1.0) for f32, so this works as expected.
                 let failed = rand::thread_rng().gen::<f32>() < self.failure_rate;
+                let store = self.store.clone();
```
bobbinth, line 217: Cloning the store is pretty light-weight, right?

bobbinth (approved):
Looks good! Thank you! I left a couple of questions inline.

> I've renamed the function `get_batch_inputs` and I'm wondering if we shouldn't rename the proto definition and domain object similarly?
> 
> I'm assuming there might be more input fields coming in the future so having it be more specific is ok?

I think this make sense. The only question is whether we'd need to use the same objects (e.g., the note authentication info) in multiple places - e.g., for transactions, batches, blocks.

Mirko-von-Leipzig:
> Looks good! Thank you! I left a couple of questions inline.
> 
> > I've renamed the function `get_batch_inputs` and I'm wondering if we shouldn't rename the proto definition and domain object similarly?
> > I'm assuming there might be more input fields coming in the future so having it be more specific is ok?
> 
> I think this make sense. The only question is whether we'd need to use the same objects (e.g., the note authentication info) in multiple places - e.g., for transactions, batches, blocks.

Okay lets minimize the diff, and revisit this once we know the proof requirements.

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: According to [this](https://docs.rs/tonic/latest/tonic/client/index.html#:~:text=The%20answer%20is%20simply%20to,the%20same%20channel%20for%20communication.) yes:

> Upon using the your generated client, you will discover all the functions corresponding to your rpc methods take &mut self, making concurrent usage of the client difficult. **The answer is simply to clone the client, which is cheap** as all client instances will share the same channel for communication.

PR #590, opened 2024-12-18
refactor(rpc): invert tx inputs missing notes to found

Mirko-von-Leipzig:
This PR inverts `TransactionInputs.missing_unauthenticated_notes` to `found_authenticated_ntoes`.

Closes a task from #519.

### DIFF
--- a/crates/block-producer/src/store/mod.rs
+++ b/crates/block-producer/src/store/mod.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::BTreeMap,
+    collections::{BTreeMap, BTreeSet},
     fmt::{Display, Formatter},
     num::NonZeroU32,
 };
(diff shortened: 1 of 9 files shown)


crates/block-producer/src/store/mod.rs
```diff
@@ -44,9 +44,11 @@ pub struct TransactionInputs {
     ///
     /// We use NonZeroU32 as the wire format uses 0 to encode none.
     pub nullifiers: BTreeMap<Nullifier, Option<NonZeroU32>>,
-    /// List of unauthenticated notes that were not found in the store
-    pub missing_unauthenticated_notes: Vec<NoteId>,
-    /// The current block height
+    /// Unauthenticated notes which are present in the store.
+    ///
+    /// These are notes which were committed _after_ the transaction was created.
+    pub found_unauthenticated_notes: BTreeSet<NoteId>,
```
bobbinth, line 50: This is fine for now, but in the future, we may also need to get the authentication paths for these notes.

bobbinth (approved):
Looks good! Thank you!

PR #591, opened 2024-12-18
fix: support new two `Felt` account ID in node

tomyrd:
This PR implements the changes necessary to support the new two Felt account ID from https://github.com/0xPolygonMiden/miden-base/pull/982.

The main changes needed were:
- Change testing account ID creation.
- Account ID in proto files is changed from `fixed64` to `bytes`. Similarly, the account IDs in the database were changed from `INTEGER` to `BLOB`.
- Genesis accounts have an anchor block with block number 0 and hash `EMPTY_WORD`.


crates/block-producer/src/block_builder/prover/tests.rs
```diff
@@ -38,9 +36,9 @@ use crate::{
 /// The store will contain accounts 1 & 2, while the transaction batches will contain 2 & 3.
 #[test]
 fn block_witness_validation_inconsistent_account_ids() {
-    let account_id_1 = AccountId::new_unchecked(Felt::new(ACCOUNT_ID_OFF_CHAIN_SENDER));
-    let account_id_2 = AccountId::new_unchecked(Felt::new(ACCOUNT_ID_OFF_CHAIN_SENDER + 1));
-    let account_id_3 = AccountId::new_unchecked(Felt::new(ACCOUNT_ID_OFF_CHAIN_SENDER + 2));
+    let account_id_1 = AccountId::try_from(ACCOUNT_ID_OFF_CHAIN_SENDER).unwrap();
+    let account_id_2 = AccountId::try_from(ACCOUNT_ID_OFF_CHAIN_SENDER + (1 << 80)).unwrap();
+    let account_id_3 = AccountId::try_from(ACCOUNT_ID_OFF_CHAIN_SENDER + (2 << 80)).unwrap();
```
tomyrd, line 42: The least significant bits of the new account ID must be zero, so we can't add low amounts. Plus, now for each account ID, the prefix (most significant 64 bits) must be modified. The 80 bits shift is so that the changed number ends up on the random bits and doesn't change the metadata.

crates/block-producer/src/block_builder/prover/tests.rs
```diff
@@ -373,11 +371,11 @@ async fn compute_account_root_empty_batches() {
...
-        AccountId::new_unchecked(Felt::new(0b0000_0000_0000_0000u64)),
-        AccountId::new_unchecked(Felt::new(0b1111_0000_0000_0000u64)),
-        AccountId::new_unchecked(Felt::new(0b1111_1111_0000_0000u64)),
-        AccountId::new_unchecked(Felt::new(0b1111_1111_1111_0000u64)),
-        AccountId::new_unchecked(Felt::new(0b1111_1111_1111_1111u64)),
+        AccountId::new_unchecked([Felt::new(0b0000_0000_0000_0000u64), Felt::ZERO]),
+        AccountId::new_unchecked([Felt::new(0b1000_0000_0000_0000u64), Felt::ZERO]),
+        AccountId::new_unchecked([Felt::new(0b1100_0000_0000_0000u64), Felt::ZERO]),
+        AccountId::new_unchecked([Felt::new(0b1110_0000_0000_0000u64), Felt::ZERO]),
+        AccountId::new_unchecked([Felt::new(0b1111_0000_0000_0000u64), Felt::ZERO]),
```
tomyrd, line 378: Similarly to the comment above, the first felt contains metadata that can be invalid if we just fill it with ones (the `new_unchecked` method does some small validations despite its name).

crates/store/src/db/sql/mod.rs
```diff
@@ -350,8 +360,7 @@ pub fn select_account_delta(
     ])?;
     while let Some(row) = rows.next()? {
         let vault_key_data = row.get_ref(1)?.as_blob()?;
-        let vault_key = Word::read_from_bytes(vault_key_data)?;
-        let asset = NonFungibleAsset::try_from(vault_key)
+        let asset = NonFungibleAsset::read_from_bytes(vault_key_data)
             .map_err(|err| DatabaseError::DataCorrupted(err.to_string()))?;
```
tomyrd, line 364: This is a separate issue but while testing I realized that we were using the `account_non_fungible_asset_updates::vault_key` to store a serialized word that is used to build the non fungible asset.
Before the ID changes, the vault key and the asset word were the same but this is no longer the case. Maybe we should change the name

crates/block-producer/src/block_builder/prover/tests.rs
Mirko-von-Leipzig: There must be a better way of generating arbitrary values for test purposes. Can we not add `rng` support for these types in `miden-base`?

This applies to many of the other types as well; where tests are awkward because just generating the fluff data is painful e.g. `TransactionId`.

tomyrd: I double-checked and found a `new_dummy` method that is only available for testing so I changed the tests to use this instead.

bin/node/src/commands/genesis/mod.rs
```diff
@@ -125,6 +125,7 @@ fn create_accounts(
                 let storage_mode = inputs.storage_mode.as_str().try_into()?;
                 let (account, account_seed) = create_basic_fungible_faucet(
                     rng.gen(),
+                    AccountIdAnchor::new_unchecked(0, EMPTY_WORD.into()),
```
Mirko-von-Leipzig, line 128: Could you add `SAFETY` comment for the `new_unchecked`.

crates/block-producer/src/block_builder/prover/block_witness.rs
```diff
@@ -188,7 +188,7 @@ impl BlockWitness {
                 for (idx, (account_id, account_update)) in self.updated_accounts.iter().enumerate()
                 {
                     account_data.extend(account_update.final_state_hash);
-                    account_data.push((*account_id).into());
+                    account_data.push(account_id.first_felt());
```
Mirko-von-Leipzig, line 191: I don't really understand this function yet; why can we truncate the account IDs?

crates/rpc-proto/proto/account.proto
```diff
@@ -7,7 +7,7 @@ message AccountId {
     // A miden account is defined with a little bit of proof-of-work, the id itself is defined as
     // the first word of a hash digest. For this reason account ids can be considered as random
     // values, because of that the encoding below uses fixed 64 bits, instead of zig-zag encoding.
-    fixed64 id = 1;
+    bytes id = 1;
```
Mirko-von-Leipzig, line 10: The comment above is outdated wrt to this change

crates/store/src/db/sql/utils.rs
```diff
@@ -86,20 +86,32 @@ pub fn column_value_as_u64<I: rusqlite::RowIndex>(
...
+        )
+    })
+}
+
 /// Constructs `AccountSummary` from the row of `accounts` table.
 ///
 /// Note: field ordering must be the same, as in `accounts` table!
 pub fn account_summary_from_row(row: &rusqlite::Row<'_>) -> crate::db::Result<AccountSummary> {
-    let account_id = column_value_as_u64(row, 0)?;
+    let account_id = AccountId::read_from_bytes(row.get_ref(0)?.as_blob()?)?;
```
Mirko-von-Leipzig, line 109: This could use the `read_from_blob_column` I think

crates/store/src/errors.rs
Mirko-von-Leipzig, line 1: This use removal could be reverted to a no-op

Mirko-von-Leipzig (approved):
I don't have any blockers; I don't fully understand why we're truncating the account ID in places, but that seems to be how the tree values are built still?

I don't like how the proto responses need to validated but that's not this PRs fault; more just our API definitions leaking into code too much.

crates/block-producer/src/block_builder/prover/block_witness.rs
tomyrd: Originally I had the same question, after reading the discussion I found my answer [here](https://github.com/0xPolygonMiden/miden-base/issues/140#issuecomment-2419854053).

TLDR: One of the ideas behind the Account ID refactor is that the first felt of the ID is still a unique identifier for an account, this is done so that the prefix can still be used as a key to the SMT.

I think it would be more correct if I used `prefix()` instead of `first_felt()`, just to better drive this idea home.

tomyrd: Additionally, now that I think about it. This "enforcement" should be done in the node, right? Would it be better to add it to this PR or a separate one?

crates/store/src/db/sql/utils.rs
```diff
@@ -86,20 +86,32 @@ pub fn column_value_as_u64<I: rusqlite::RowIndex>(
     Ok(value as u64)
 }
 
+pub fn read_from_blob_column<I, T>(row: &rusqlite::Row<'_>, index: I) -> rusqlite::Result<T>
```
igamigo, line 89: I know I added this, but we should add a brief doc comment mentioning what this does.

crates/store/src/server/api.rs
```diff
@@ -607,6 +630,18 @@ fn invalid_argument<E: core::fmt::Display>(err: E) -> Status {
...
+#[instrument(target = COMPONENT, skip_all, err)]
+fn validate_account_ids(
+    account_ids: &[generated::account::AccountId],
+) -> Result<Vec<AccountId>, Status> {
+    account_ids
+        .iter()
+        .cloned()
+        .map(AccountId::try_from)
+        .collect::<Result<_, ConversionError>>()
+        .map_err(|_| invalid_argument("Digest field is not in the modulus range"))
```
igamigo, line 642: I think I am also at fault here for copying this function but forgetting to change this `map_err()` error message.

igamigo (approved):
Looks good but would be good to have @bobbinth or Philipp look at this (although he's offline today AFAIK). 
Also, one comment: I believe sending the account ID as `bytes` on protobuf is maybe a byte more efficient than sending the high and low segments as `fixed64` each (although not entirely sure), but it would be probably better to save the account IDs as two different `u64` in the database for more efficient lookups an indices, considering prefixes are unique. We might want to open an issue to track this.

crates/store/src/server/api.rs
tomyrd: Missed this while reviewing draft

crates/store/src/db/sql/utils.rs
Mirko-von-Leipzig: We did something [similar in pathfinder](https://github.com/eqlabs/pathfinder/blob/79a26781e85b03fac548c7da387a5863df95104a/crates/storage/src/params.rs#L154-L395), but we did this as a trait extension for all our types.

I'll probably try something similar here at some point, it makes things quite ergonomic and type safe because you can just go

```rust
row.get_account_id(0)
```

crates/block-producer/src/block_builder/prover/block_witness.rs
Mirko-von-Leipzig: Ah I see, thanks. I'll have to read through that thread I guess 😅

Maybe `unique_prefix` would be ideal from an understanding p.o.v.? Though then maybe its not obvious that the call does no work.

bobbinth: Yes, for now the first 64 bits (i.e., the prefix) should uniquely identify an account. Some time down the road we may relax this, but for now let's assume that that's how it is. The enforcement should happen in 2 places:

- When a transaction to create an account is submitted by the node, the node should make sure no other account with the same 64-bit prefix already exists.
- In the block kernel, when we update the account SMT, we should make sure that there was no account with the same prefix.

Also, I like the idea of using `prefix()` instead of `first_felt()`. Maybe we should get rid of `first/second_felt()` and instead use `prefix()` and `suffix()` functions? cc @PhilippGackstatter.

bobbinth (approved):
Looks good! Thank you! I made a couple of minor commits to clean things up a bit and also left a couple of comments for the future.

crates/block-producer/src/block_builder/prover/block_witness.rs
PhilippGackstatter: I would definitely like to make the naming consistent between MASM (hi, lo) and Rust (first, second). It would be generally good to have a single name instead of two for the same thing. Especially places like layout descriptions in Rust would profit from this where some layout may be described as `[id_hi, id_lo]` but Rust code itself uses first and second felts which is not as clear as it could be.

As you mentioned here (https://github.com/0xPolygonMiden/miden-base/pull/982#discussion_r1897314843) the first option is to rename the Rust side to high and low.

The second option is to rename everything to prefix and suffix, and use something like `id_pre` and `id_suf` as abbreviations in MASM (or written out?).

We already have `AccountIdPrefix` and unless we rename it, the most consistent option would be the second one. Everything would be prefix/suffix then. I would be fine with that.

Because the uniqueness property of the first felt is only relevant in a limited set of circumstances (like here) from my point of view it's fine if it's not included in the name but spelled out in the docs for sure. I will add this.

bobbinth: Yes, let's go with `prefix()` and `suffix()` naming. For MASM, I don't have a strong opinion but spelling out the words is probably fine (they are not that long).

PR #598, opened 2024-12-23
feat: Enable storage map querying

### LINKED ISSUE #596: Enable requesting account storage map information
To enable FPI, `GetAccountProofs` was introduced. This endpoint returns a proof of the account's existence, alongside state headers: `AccountHeader`, `StorageHeader` and `AccountCode`. There is an implicit limitation to this workflow: `StorageHeader` only contains the top-level storage slot items (eg, for storage maps, only the root is returned, without any actual map information), so any transaction that attempts to get data related to a foreign account's storage map would fail.

A mechanism should be provided to allow querying for a specific item/key, in a way that enables FPI with procedures that look into any item, not just top-level values. I'm not sure if we would prefer to retrieve individual storage slots, or if it could be made more granular in order to request specific values of storage maps.

As a side note, if a workaround is needed with more urgency, maybe we could (in the client) get the full account object with the `GetAccountDetails` endpoint for now to work around this limitation.

igamigo:
Closes #596


crates/store/src/state.rs
```diff
@@ -698,45 +703,80 @@ impl State {
...
-                                    .then_some(details.code().to_bytes()),
-                            },
-                        )
-                    })
-                })
-                .collect()
+            let mut headers_map = BTreeMap::new();
+
+            for account_info in infos {
+                if let Some(details) = account_info.details {
```
Mirko-von-Leipzig, line 723: Under what conditions would we expect this to be `None`?

crates/store/src/state.rs
```diff
@@ -698,45 +703,80 @@ impl State {
...
-                                    .not()
-                                    .then_some(details.code().to_bytes()),
-                            },
-                        )
-                    })
-                })
-                .collect()
+            let mut headers_map = BTreeMap::new();
+
+            for account_info in infos {
```
Mirko-von-Leipzig, line 722: I would probably iterate over the requests instead? Since you also do that in the storage map loop.

crates/store/src/state.rs
```diff
@@ -698,45 +703,80 @@ impl State {
...
+                                let proof = storage_map.open(map_key);
+
+                                let slot_map_key = StorageSlotMapProof {
+                                    storage_slot: *storage_index as u32,
+                                    smt_proof: proof.to_bytes(),
+                                };
+                                storage_slot_map_keys.push(slot_map_key);
+                            }
+                        }
+                        // TODO: else error?
```
Mirko-von-Leipzig, line 748: This occurs if the user specifies an unknown slot key? Do we distinguish between `None` and zero in Miden, or is an empty slot equivalent to zero?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 - [BREAKING] Added support for new two `Felt` account ID (#591).
 - [BREAKING] Inverted `TransactionInputs.missing_unauthenticated_notes` to `found_missing_notes` (#509).
+- [BREAKING] Added way of querying account storage map state on `GetAccountProofs` endpoint (#598).
```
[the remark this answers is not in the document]
igamigo: To be consistent with other entries (past tense, etc.), I used this instead:

```
- [BREAKING] Added support for storage maps on `GetAccountProofs` endpoint (#598).
```

Let me know if you think there's a better alternative

proto/requests.proto
```diff
@@ -132,11 +132,33 @@ message GetAccountStateDeltaRequest {
...
 }
 
+// Request message to get account proofs.
 message GetAccountProofsRequest {
-    // List of account IDs to get states.
-    repeated account.AccountId account_ids = 1;
-    // Optional flag to include header and account code in the response. `false` by default.
+    // Represents per-account requests where each account ID has its own list of 
+    // (storage_slot_index, map_keys) pairs.
+    message AccountRequest {
```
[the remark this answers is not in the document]
igamigo: Correct, the original idea was to not make known code commitments be 1:1 with account IDs because they could share the same commitment. I don't mind either way but the current approach seems marginally better.

crates/store/src/server/api.rs
```diff
@@ -477,25 +480,25 @@ impl api_server::Api for StoreApi {
         &self,
         request: Request<GetAccountProofsRequest>,
     ) -> Result<Response<GetAccountProofsResponse>, Status> {
-        let request = request.into_inner();
-        if request.account_ids.len() < request.code_commitments.len() {
-            return Err(Status::invalid_argument(
-                "The number of code commitments should not exceed the number of requested accounts.",
-            ));
-        }
-
```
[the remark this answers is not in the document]
igamigo: I added a brief note on the store's readme, but any suggestion is welcome.

crates/store/src/state.rs
igamigo: Any non-public account

crates/store/src/state.rs
igamigo: Changed it!

crates/store/README.md
```diff
@@ -2,6 +2,8 @@
 
 The **Store** maintains the state of the chain. It serves as the "source of truth" for the chain - i.e., if it is not in
 the store, the node does not consider it to be part of the chain.
+Incoming requests to the store are trusted because they are validated at the RPC level.
+
```
Mirko-von-Leipzig, line 6: RPC level could be ambiguous imo

```suggestion
Incoming requests to the store are trusted because they are validated in the RPC component.

```

proto/requests.proto
```diff
@@ -132,11 +132,33 @@ message GetAccountStateDeltaRequest {
...
+
+        // A list of map keys (Digests) associated with this storage slot.
+        repeated digest.Digest map_keys = 2;
+    }
+
+    // A list of account requests, including map keys + values.
+    repeated AccountRequest account_requests = 1;
+
+    // Optional flag to include header and account code in the response. False by default.
     optional bool include_headers = 2;
```
bobbinth, line 160: We should probably clarify how this flag affects storage requests (I'm assuming that if it is false, storage requests are also ignored).

proto/responses.proto
```diff
@@ -195,9 +195,23 @@ message AccountProofsResponse {
...
     bytes storage_header = 2;
-    // Account code, returned only when none of the request's code commitments match with the
-    // current one.
+
+    // Account code, returned only when none of the request's code commitments match
+    // the current one.
     optional bytes account_code = 3;
+
+    // Storage slots information for this account
+    repeated StorageSlotMapProof storage_slots = 4;
```
bobbinth, line 207: Should the name be `storage_maps` since this is only for storage map slots?

Also, this is probably fine for now, but ideally we'd return a single batch proof per map. In cases when more than one key is requested per map, this could save quite a bit on bandwidth. Let's create an issue for this.

crates/proto/src/domain/accounts.rs
```diff
@@ -8,10 +8,12 @@ use miden_objects::{
...
 };
 
+use super::try_convert;
 use crate::{
     errors::{ConversionError, MissingFieldHelper},
     generated::{
         account as proto,
+        requests::get_account_proofs_request,
         responses::{AccountBlockInputRecord, AccountTransactionInputRecord},
     },
```
bobbinth, line 18: To follow the convention introduced in #610, I would change this to:
```rust
use crate::{
    generated as proto,
};
```
And update the rest of the file accordingly.

crates/store/src/state.rs
```diff
@@ -689,47 +695,78 @@ impl State {
...
```
bobbinth, line 749: I'm thinking this should probably be an error: we are asking for some values from a storage map, but the storage slot is not a map.

An alternative could be to convert the fact that values for some requested keys were not returned in to an error on the client side, but I'm not seeing too much benefit in doing the work if the client is going to error out anyway.

crates/store/src/state.rs
```diff
@@ -689,47 +695,78 @@ impl State {
...
+            let mut headers_map = BTreeMap::new();
+
+            // Iterate and build state headers for public accounts
+            for request in account_requests {
+                let account_info = infos
+                    .iter()
+                    .find(|info| info.summary.account_id == request.account_id)
+                    .expect("retrieved accounts were validated against request");
+
+                if let Some(details) = &account_info.details {
```
bobbinth, line 730: Not for this PR: this approach works fine for now, but it won't work with large accounts (i.e., we are loading the full account state into memory and then selecting key-value pairs from it).

bobbinth (approved):
Looks good! Thank you! I left a few small comments inline. Once these are addressed we can merge.

As I mentioned in one of the comments, this approach is fine for now, but we should refactor both how we store public account data and how we provide the data to the client.

proto/requests.proto
igamigo: Clarified this

proto/responses.proto
igamigo: Changed!

crates/proto/src/domain/accounts.rs
igamigo: Agreed, changed

crates/store/src/state.rs
igamigo: Yes, if this is an error scenario it makes more sense to error as soon as possible. Made this an error.

bobbinth (approved):
Looks good! Thank you!

PR #599, opened 2024-12-25
Tracking PR for v0.7 release

bobbinth:
This is a tracking PR for v0.7 release.


bobbinth:
@Mirko-von-Leipzig - could you make a PR to resolve merge conflicts here?

Mirko-von-Leipzig:
Hmm. I'm struggling to convince `git` to use the readme version from `next`.. Merging that change into `next` is a no-op since its already at that state; but then we just end at the same merge conflict.

bobbinth:
> Hmm. I'm struggling to convince `git` to use the readme version from `next`.. Merging that change into `next` is a no-op since its already at that state; but then we just end at the same merge conflict.

Let me try to resolve this via Github.

PR #603, opened 2024-12-27
refactor: change account ID SQL's backing type

### LINKED ISSUE #602: Consider changing account ID's SQL backing type
Following #591, the node supports the new account ID format, which fits into two felts instead of just one. As a consequence, 64-bit long types do not work as a backing type for account IDs anymore, which means we can't use `fixed64` on protobuf or `u64` for SQL. 

Currently, the SQL type for `account_id` is `BLOB` which is fine because we can deserialize and serialize as needed, but since the account ID prefix (which fits into 64 bits) is a unique identifier of the account itself, we can separate the ID into two columns (first felt or prefix, and second felt or suffix) in order to speed up lookups and improve indexing. This should also reduce allocations because it means less (de)serialization (would also leverage the fact that `AccountId` only contains the two felts so instantiation would still be simple). One thing to note is that the prefix being a unique identifier might be relaxed in the future, so this only really applies while this is still the case.

tomyrd:
closes #602 

In this PR we change the account ID sql columns from blobs to two integers (u64). This allows us to index only with the prefix (as it uniquely identifies accounts) which is faster than indexing with the whole ID.

In this implementation I used two integers in tables where the full account ID was needed and only prefix when it was used as a foreign key.

### DIFF
--- a/crates/store/src/db/sql/utils.rs
+++ b/crates/store/src/db/sql/utils.rs
@@ -4,6 +4,7 @@ use miden_objects::{
     crypto::hash::rpo::RpoDigest,
     notes::Nullifier,
     utils::Deserializable,
+    Felt,
 };
 use rusqlite::{
     params,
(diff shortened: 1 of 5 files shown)


tomyrd:
I wanted to ask about this part:
> used two integers in tables where the full account ID was needed

This is not technically necessary as the suffix may be stored only in the `accounts` table and retreived using a JOIN but I wasn't sure how this would affect the performance. If we prefer, I can use only prefix and get the suffix with joins if needed.

crates/store/src/db/sql/utils.rs
```diff
@@ -103,14 +104,29 @@ where
     })
 }
 
+pub fn read_account_id_from_columns(
```
igamigo, line 107: We should add some brief doc comments to this function.

igamigo (approved):
LGTM! I think this approach should work well. I left a small comment, but everything else looks good. Also, not sure but it might be worth mentioning (maybe in a comment on the migration file) why account IDs are broken up into 2, and why the prefix works as a primary key, etc.

bobbinth:
I would probably put this PR on-hold for now. Ideally, we'd have some sort of a framework to measure performance impact of optimizations like this before implementing them.

bobbinth:
Closing this as stale. We can probably still re-use some of this code later on when we come back to #602.

PR #604, opened 2024-12-30
feat: enforce unique account prefix

### LINKED ISSUE #601: feat: enforce account ID prefix uniqueness
> The enforcement should happen in 2 places:
> - When a transaction to create an account is submitted by the node, the node should make sure no other account with the same 64-bit prefix already exists.
> - In the block kernel, when we update the account SMT, we should make sure that there was no account with the same prefix.

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-node/pull/591#discussion_r1897463991_

tomyrd:
closes #601

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Changes
 
+- Added check for unique account ID prefixes (#604).
 - [BREAKING] Added support for new two `Felt` account ID (#591).
 - [BREAKING] Inverted `TransactionInputs.missing_unauthenticated_notes` to `found_missing_notes` (#509).
 
--- a/crates/block-producer/src/errors.rs
+++ b/crates/block-producer/src/errors.rs
@@ -1,7 +1,7 @@
 use miden_node_proto::errors::ConversionError;
 use miden_node_utils::formatting::format_opt;
 use miden_objects::{
-    accounts::AccountId,
+    accounts::{AccountId, AccountIdPrefix},
     crypto::merkle::MerkleError,
     notes::{NoteId, Nullifier},
     transaction::TransactionId,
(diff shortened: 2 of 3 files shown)


bobbinth:
Not a review yet - but one thing I'm wondering is whether we should tackle this issue in a broader context. Specifically, currently account tree uses a `SimpleSmt` and so the methodology for checking uniqueness relies on this. But it is probably a good idea to migrate from `SimpleSmt` to a regular `Smt` - and in this case, we probably would need to modify the approach somewhat.

We should still weigh all pros and cons of migrating from `SimpleSmt` to `Smt` - but it may make sense to put this PR on hold until we make a decision here.

Mirko-von-Leipzig:
@PhilippGackstatter iirc you're planning (or already have?) to implement this change to `Smt`?

PhilippGackstatter:
> @PhilippGackstatter iirc you're planning (or already have?) to implement this change to `Smt`?

There is just an issue (https://github.com/0xPolygonMiden/miden-base/issues/1157) for this but it's not implemented yet.

> We should still weigh all pros and cons of migrating from SimpleSmt to Smt - but it may make sense to put this PR on hold until we make a decision here.

I think using `Smt` allows us to eventually lift the restriction of account ID prefix uniqueness, if we can figure out a solution for non-fungible assets (which are the original cause for this restriction in the first place). `Smt` can support multiple entries with the same 64-bit prefix while `SimpleSmt` can't. `Smt` also stores the full 120 bits of the account ID while the `SimpleSmt` would only store the first 64 bits, making a later migration to `Smt` very hard, maybe even impossible.

bobbinth:
Closing this as stale. We can probably still use some of this code when we come back to #601.

Issue #581, opened 2024-12-11
Refactor errors according to guidelines


Mirko-von-Leipzig:
Our errors should be updated according to the standards discussed https://github.com/0xPolygonMiden/miden-base/discussions/966.

bobbinth:
Closed by #611.

Issue #499, opened 2024-09-17
[Task]: Correctly emit traces on faucet


igamigo:
### Node's components affected by this task

Faucet

### What should be done?

On the faucet, when any error occurs on the server's actions, it is not logged correctly to the console. Instead, errors are bubbled up as `Result<_,_>`, and absorbed in the top level. Instead, it should be logged as an error trace.

### How should it be done?

On the endpoint functions, make sure that if any error is returned, it is logged and shown pertinently.

### When is this task done?

When, for example, we can see internal server errors on the console.

### Additional context

https://github.com/0xPolygonMiden/miden-node/pull/498#discussion_r1764050281

Mirko-von-Leipzig:
This has likely been fixed in the migration from `actix` to `axum` in #479.

It now uses tower's [http-trace](https://docs.rs/tower-http/latest/tower_http/trace/index.html) which has sensible defaults and I presume logs failures.

I would keep this issue open as we may want to improve the traces further by customizing them.

The `axum` tracing [example](https://github.com/tokio-rs/axum/blob/main/examples/tracing-aka-logging/src/main.rs) shows some things we may want to include, like the rejection of extractors (we use the `Json<T>` extractor for example.

bobbinth:
Closed by #611.

PR #611, opened 2025-01-13
refactor: standardize errors

### LINKED ISSUE #484: Improve the node & faucet logs
As evidenced in #483, our logs are missing useful context.

A lot of this context is going missing because we are using structured error enums which aren't encapsulating the inner error.

Common suggestion is to prefer `anyhow` for errors in an application as usually we don't care about the what, only that an error happened. For libs instead `thiserror`/structure errors are suggested but these are just guidelines.

In general, we should use `anyhow` if we don't care to match on the exact error variant.

Mirko-von-Leipzig:
Updates errors to follow the guidelines in https://github.com/0xPolygonMiden/miden-base/discussions/966.

This PR also removes the wildcard route option from the faucet. Currently the faucet allows visiting arbitrary paths (e.g. https://testnet.miden.io/asdas). Clearly this isn't behaving as intended and since we only have the single static file I've removed the wildcard entirely.

I looked into the axum defaults relating to https://github.com/0xPolygonMiden/miden-node/issues/499#issuecomment-2399190012 and decided the defaults of `Error` for failures, `Debug` for success are fine - though maybe we want to log the latter as info? I've also enabled the built-in axum rejects by setting `axum::rejections=trace` by default.

Closes #484, #499 and #581.

I do feel we could simplify our error handling strategy but that can wait for another day. In general we have many error variants, but we never actually do anything with them except bubble them up. Or put differently, aside from differentiating between internal/user-facing errors, we don't usually care about the errors.

### DIFF
--- a/crates/block-producer/src/errors.rs
+++ b/crates/block-producer/src/errors.rs
@@ -5,7 +5,7 @@ use miden_objects::{
     crypto::merkle::MerkleError,
     notes::{NoteId, Nullifier},
     transaction::TransactionId,
-    AccountDeltaError, Digest, TransactionInputError,
+    AccountDeltaError, Digest,
 };
 use miden_processor::ExecutionError;
 use thiserror::Error;
@@ -69,9 +69,6 @@ pub enum VerifyTxError {
     #[error("failed to retrieve transaction inputs from the store")]
     StoreConnectionFailed(#[from] StoreError),
 
-    #[error("transaction input error")]
-    TransactionInputError(#[from] TransactionInputError),
-
     /// Failed to verify the transaction execution proof
     #[error("invalid transaction proof error for transaction: {0}")]
     InvalidTransactionProof(TransactionId),
@@ -85,16 +82,18 @@ pub enum AddTransactionError {
     #[error("transaction verification failed")]
     VerificationFailed(#[from] VerifyTxError),
 
-    #[error("transaction input data is stale. Required data from {stale_limit} or newer, but inputs are from {input_block}")]
+    #[error("transaction input data from block {input_block} is rejected as stale because it is older than the limit of {stale_limit}")]
     StaleInputs {
         input_block: BlockNumber,
         stale_limit: BlockNumber,
     },
 
-    #[error("deserialization failed: {0}")]
-    DeserializationError(String),
+    #[error("transaction deserialization failed")]
+    TransactionDeserializationFailed(#[source] miden_objects::utils::DeserializationError),
 
-    #[error("transaction expired at {expired_at} but the limit was {limit}")]
+    #[error(
+        "transaction expired at block height {expired_at} but the block height limit was {limit}"
+    )]
     Expired {
         expired_at: BlockNumber,
         limit: BlockNumber,
@@ -111,12 +110,12 @@ impl From<AddTransactionError> for tonic::Status {
             | VerificationFailed(VerifyTxError::IncorrectAccountInitialHash { .. })
             | VerificationFailed(VerifyTxError::InvalidTransactionProof(_))
             | Expired { .. }
-            | DeserializationError(_) => Self::invalid_argument(value.to_string()),
+            | TransactionDeserializationFailed(_) => Self::invalid_argument(value.to_string()),
 
             // Internal errors which should not be communicated to the user.
-            VerificationFailed(VerifyTxError::TransactionInputError(_))
-            | VerificationFailed(VerifyTxError::StoreConnectionFailed(_))
-            | StaleInputs { .. } => Self::internal("Internal error"),
+            VerificationFailed(VerifyTxError::StoreConnectionFailed(_)) | StaleInputs { .. } => {
+                Self::internal("Internal error")
+            },
         }
     }
 }
@@ -140,20 +139,22 @@ pub enum BuildBatchError {
         output_hash: Digest,
     },
 
-    #[error("failed to merge transaction delta into account {account_id}: {error}")]
+    #[error("failed to merge transaction delta into account {account_id}")]
     AccountUpdateError {
         account_id: AccountId,
-        error: AccountDeltaError,
+        source: AccountDeltaError,
     },
 
+    /// We sometimes randomly inject errors into the batch building process to test our failure
+    /// responses.
     #[error("nothing actually went wrong, failure was injected on purpose")]
     InjectedFailure,
 
     #[error("batch proving task panic'd")]
     JoinError(#[from] tokio::task::JoinError),
 
-    #[error("fetching inputs from store failed")]
-    StoreError(#[from] StoreError),
+    #[error("failed to fetch batch inputs from store")]
+    FetchBatchInputsFailed(#[source] StoreError),
 }
 
 // Block prover errors
@@ -162,9 +163,9 @@ pub enum BuildBatchError {
 #[derive(Debug, Error)]
 pub enum BlockProverError {
     #[error("received invalid merkle path")]
-    InvalidMerklePaths(MerkleError),
+    InvalidMerklePaths(#[source] MerkleError),
     #[error("program execution failed")]
-    ProgramExecutionFailed(ExecutionError),
+    ProgramExecutionFailed(#[source] ExecutionError),
     #[error("failed to retrieve {0} root from stack outputs")]
     InvalidRootOutput(&'static str),
 }
@@ -176,25 +177,29 @@ pub enum BlockProverError {
 pub enum BuildBlockError {
     #[error("failed to compute new block")]
     BlockProverFailed(#[from] BlockProverError),
-    #[error("failed to apply block")]
-    ApplyBlockFailed(#[source] StoreError),
+    #[error("failed to apply block to store")]
+    StoreApplyBlockFailed(#[source] StoreError),
     #[error("failed to get block inputs from store")]
     GetBlockInputsFailed(#[source] StoreError),
-    #[error("store did not produce data for account: {0}")]
+    #[error("block inputs from store did not contain data for account {0}")]
     MissingAccountInput(AccountId),
-    #[error("store produced extra account data. Offending accounts: {0:?}")]
+    #[error("block inputs from store contained extra data for accounts {0:?}")]
     ExtraStoreData(Vec<AccountId>),
-    #[error("no matching state transition found for account {0}. Current account state is {1}, remaining updates: {2:?}")]
+    #[error("account {0} with state {1} cannot transaction to remaining states {2:?}")]
     InconsistentAccountStateTransition(AccountId, Digest, Vec<Digest>),
-    #[error("transaction batches and store don't produce the same nullifiers. Offending nullifiers: {0:?}")]
+    #[error(
+        "block inputs from store and transaction batches produced different nullifiers: {0:?}"
+    )]
     InconsistentNullifiers(Vec<Nullifier>),
     #[error("unauthenticated transaction notes not found in the store or in outputs of other transactions in the block: {0:?}")]
     UnauthenticatedNotesNotFound(Vec<NoteId>),
-    #[error("failed to merge transaction delta into account {account_id}: {error}")]
+    #[error("failed to merge transaction delta into account {account_id}")]
     AccountUpdateError {
         account_id: AccountId,
-        error: AccountDeltaError,
+        source: AccountDeltaError,
     },
+    /// We sometimes randomly inject errors into the batch building process to test our failure
+    /// responses.
     #[error("nothing actually went wrong, failure was injected on purpose")]
     InjectedFailure,
 }
--- a/crates/block-producer/src/errors.rs
+++ b/crates/block-producer/src/errors.rs
@@ -40,21 +40,23 @@ pub enum BlockProducerError {
 #[derive(Debug, Error)]
 pub enum VerifyTxError {
     /// Another transaction already consumed the notes with given nullifiers
-    #[error("Input notes with given nullifiers were already consumed by another transaction")]
+    #[error(
+        "input notes with given nullifiers were already consumed by another transaction: {0:?}"
+    )]
     InputNotesAlreadyConsumed(Vec<Nullifier>),
 
     /// Unauthenticated transaction notes were not found in the store or in outputs of in-flight
     /// transactions
     #[error(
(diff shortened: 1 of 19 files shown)


crates/utils/src/logging.rs
```diff
@@ -22,11 +19,11 @@ pub fn subscriber() -> impl Subscriber + core::fmt::Debug {
...
-        .with_env_filter(
-            EnvFilter::builder()
-                .with_default_directive(LevelFilter::INFO.into())
-                .from_env_lossy(),
-        )
+        .with_env_filter(EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+            // axum logs rejections from built-in extracts on the trace level, so we enable this
+            // manually.
+            format!("{}=info,axum::rejection=trace", env!("CARGO_CRATE_NAME")).into()
+        }))
```
Mirko-von-Leipzig, line 26: Actually this is incorrect because we use `target=COMPONENT`. Will fix this.

igamigo (commented):
Looks good overall, although I left a comment with a question. Basically not sure why we took out the `#[from]` _and_ the error messages from the display messages.

Also, regarding this:

> I looked into the axum defaults relating to https://github.com/0xPolygonMiden/miden-node/issues/499#issuecomment-2399190012 and decided the defaults of Error for failures, Debug for success are fine - though maybe we want to log the latter as info? I've also enabled the built-in axum rejects by setting axum::rejections=trace by default.

These defaults sound sensible to me

crates/store/src/errors.rs
```diff
@@ -40,65 +40,59 @@ pub enum NullifierTreeError {
...
     },
-    #[error("Account {0} not found in the database")]
+    #[error("account {0} not found")]
     AccountNotFoundInDb(AccountId),
-    #[error("Accounts {0:?} not found in the database")]
+    #[error("accounts {0:?} not found")]
     AccountsNotFoundInDb(Vec<AccountId>),
-    #[error("Account {0} is not on the chain")]
+    #[error("account {0} is not on the chain")]
     AccountNotOnChain(AccountId),
```
igamigo, line 80: This should likely be renamed to `AccountNotPublic`

crates/block-producer/src/errors.rs
```diff
@@ -64,14 +66,14 @@ pub enum VerifyTxError {
     ///
     /// TODO: Make this an "internal error". Q: Should we have a single `InternalError` enum for
     /// all internal errors that can occur across the system?
-    #[error("Failed to retrieve transaction inputs from the store: {0}")]
+    #[error("failed to retrieve transaction inputs from the store")]
     StoreConnectionFailed(#[from] StoreError),
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: `#[from]` is a superset of `#[source]`, quoting from the `thiserror` docs (emphasis mine):

> The Error trait’s source() method is implemented to return whichever field has a #[source] attribute or is named source, if any. This is for identifying the underlying lower level error that caused your error.
>
> **_The #[from] attribute always implies that the same field is #[source], so you don’t ever need to specify both attributes._**
>
> Any error type that implements std::error::Error or dereferences to dyn std::error::Error will work as a source.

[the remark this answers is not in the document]
Mirko-von-Leipzig: This does mean that `Display` won't show the full backtrace; as described in the `XOR Source` section of the discussion in https://github.com/0xPolygonMiden/miden-base/discussions/966.

We should go through all the "exit" routes of our tasks and main's to ensure they're doing the correct thing and returning a full error report.

crates/store/src/errors.rs
Mirko-von-Leipzig: Thank you that makes much more sense :)

Mirko-von-Leipzig: That said it doesn't appear to be used anywhere?

bobbinth (approved):
Looks good! Thank you! This is not a super deep review from my end, but upon a quick scan, I didn't see any issues.

> I do feel we could simplify our error handling strategy but that can wait for another day. In general we have many error variants, but we never actually do anything with them except bubble them up. Or put differently, aside from differentiating between internal/user-facing errors, we don't usually care about the errors.

This could be a good idea, but I'd keep for another PR. Let's create an issue for it?

crates/block-producer/src/errors.rs
```diff
@@ -125,32 +127,32 @@ impl From<AddTransactionError> for tonic::Status {
...
+    #[error("nothing actually went wrong, failure was injected on purpose")]
     InjectedFailure,
 
-    #[error("Batch proving task panic'd")]
+    #[error("batch proving task panic'd")]
     JoinError(#[from] tokio::task::JoinError),
 
-    #[error("Fetching inputs from store failed")]
+    #[error("fetching inputs from store failed")]
     StoreError(#[from] StoreError),
```
PhilippGackstatter, line 156:
```suggestion
    #[error("failed to fetch transaction batch inputs from store")]
    FailedToFetchBatchInputsFromStore(#[from] StoreError),
```

The error message is concrete, but the variant name is generic. Can we rename this? Feel free to pick something less unwiedly.

Also I think we should include in the error message what kinds of inputs we mean (there are also block inputs, and so it's not clear if "inputs" are those or tx inputs).

I would also prefer using `#[source]` on this one, so we are forced to map the error explicitly. Where we use this, we already use `map_err` anyway. Plus, this makes analyzing where variants are used possible in IDEs (i.e. find usages currently returns zero results for this because of the implicit nature of `#[from]`).

crates/store/src/errors.rs
```diff
@@ -40,71 +40,65 @@ pub enum NullifierTreeError {
 pub enum DatabaseError {
```
PhilippGackstatter, line 40: I think this error should be refactored to not have generic variants for each sub error, because they add very little context. Similar to a previous comment about the NullifierTree, if we apply a delta to an account and it fails, we would maybe get "account error: storage slot index is 10 but the slots length is 5", which could be more helpful by making it "failed to apply account delta to account 0x123: storage slot index is 10 but the slots length is 5". Especially if this error is part of a larger error chain, this intermediate context is important for locating what actually went wrong.

Not for this PR, but a larger question is whether this error should contain all of these variants in the first place (see also "kitchen-sink enum" in https://matklad.github.io/2020/10/15/study-of-std-io-error.html). I don't have a perfect solution for this, but perhaps we could think about whether DB connection errors like `MissingDbConnection` or `SqliteError` should be part of this error or if they are actually unrecoverable and should lead to a panic at the call site. But I don't know a lot about how the node works, so consider this just a question and not a request.

crates/store/src/errors.rs
```diff
@@ -117,105 +111,105 @@ impl From<DatabaseError> for Status {
...
     // ---------------------------------------------------------------------------------------------
-    #[error("Database error: {0}")]
+    #[error("database error")]
     DatabaseError(#[from] DatabaseError),
-    #[error("Block error: {0}")]
+    #[error("block error")]
     BlockError(#[from] BlockError),
-    #[error("Merkle error: {0}")]
+    #[error("merkle error")]
     MerkleError(#[from] MerkleError),
```
PhilippGackstatter, line 145: Might be worthwhile to check if any of these could be made more concrete as well.

crates/store/src/errors.rs
```diff
@@ -117,105 +111,105 @@ impl From<DatabaseError> for Status {
...
+    #[error("failed to deserialize genesis file")]
+    GenesisFileDeserializationError(#[from] DeserializationError),
+    #[error("retrieving genesis block header failed")]
+    SelectBlockHeaderByBlockNumError(#[from] Box<DatabaseError>),
 
     // OTHER ERRORS
     // ---------------------------------------------------------------------------------------------
-    #[error("Apply block failed: {0}")]
+    #[error("apply block failed: {0}")]
     ApplyBlockFailed(String),
```
PhilippGackstatter, line 154: Should we make this a `#[source]` error instead and make the error message say "genesis block"?

crates/store/src/errors.rs
```diff
@@ -117,105 +111,105 @@ impl From<DatabaseError> for Status {
...
-    #[error("Failed to read genesis file \"{genesis_filepath}\": {error}")]
+    #[error("failed to read genesis file \"{genesis_filepath}\": {error}")]
     FailedToReadGenesisFile {
         genesis_filepath: String,
         error: io::Error,
     },
```
PhilippGackstatter, line 159:
```suggestion
    #[error("failed to read genesis file \"{genesis_filepath}\"")]
    FailedToReadGenesisFile {
        genesis_filepath: String,
        source: io::Error,
    },
```

crates/store/src/errors.rs
```diff
@@ -117,105 +111,105 @@ impl From<DatabaseError> for Status {
...
 pub enum InvalidBlockError {
-    #[error("Duplicated nullifiers {0:?}")]
+    #[error("duplicated nullifiers {0:?}")]
     DuplicatedNullifiers(Vec<Nullifier>),
-    #[error("Invalid output note type: {0:?}")]
+    #[error("invalid output note type: {0:?}")]
     InvalidOutputNoteType(Box<OutputNote>),
-    #[error("Invalid tx hash: expected {expected}, but got {actual}")]
+    #[error("invalid tx hash: expected {expected}, but got {actual}")]
     InvalidTxHash { expected: RpoDigest, actual: RpoDigest },
```
PhilippGackstatter, line 176:
```suggestion
    #[error("invalid block tx hash: expected {expected}, but got {actual}")]
    InvalidBlockTxHash { expected: RpoDigest, actual: RpoDigest },
```

Nit: This seems to be specific to the block tx hash.

PhilippGackstatter (approved):
Looks good to me. Thanks for working on this! If I understand correctly, the goal of this PR was to make errors fall in line with the guidelines and it does that so I will approve.

I left comments about where I think errors could be more concrete or add more context in their messages by refactoring them. Generally, I'm not a fan of using `#[from]` as it usually leads to generic error variants which cover multiple concrete cases, which, imho, should have their own concrete variants to ultimately result in better errors. I think `#[from]` is appropriate in places like `ConversionError`, where no more context can be added by introducing concrete variants so having to `map_err` is actually unnecessary.

This is also a matter of opinion whether applying my suggestions is necessary, so feel free to apply or not apply what you think makes sense for this codebase.

crates/block-producer/src/errors.rs
```diff
@@ -80,19 +82,19 @@ pub enum VerifyTxError {
...
 pub enum AddTransactionError {
-    #[error("Transaction verification failed: {0}")]
+    #[error("transaction verification failed")]
     VerificationFailed(#[from] VerifyTxError),
 
-    #[error("Transaction input data is stale. Required data from {stale_limit} or newer, but inputs are from {input_block}.")]
+    #[error("transaction input data is stale. Required data from {stale_limit} or newer, but inputs are from {input_block}")]
     StaleInputs {
         input_block: BlockNumber,
         stale_limit: BlockNumber,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Thanks I'll add a review

crates/block-producer/src/errors.rs
```diff
@@ -80,19 +82,19 @@ pub enum VerifyTxError {
 
 #[derive(Debug, Error)]
 pub enum AddTransactionError {
-    #[error("Transaction verification failed: {0}")]
+    #[error("transaction verification failed")]
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: This is returned specifically by the add transaction RPC method, as in there is only a single transaction in play - the one submitted to the rpc.

crates/block-producer/src/errors.rs
```diff
@@ -172,11 +174,11 @@ pub enum BlockProverError {
...
 pub enum BuildBlockError {
-    #[error("failed to compute new block: {0}")]
+    #[error("failed to compute new block")]
     BlockProverFailed(#[from] BlockProverError),
-    #[error("failed to apply block: {0}")]
+    #[error("failed to apply block")]
     ApplyBlockFailed(#[source] StoreError),
-    #[error("failed to get block inputs from store: {0}")]
+    #[error("failed to get block inputs from store")]
     GetBlockInputsFailed(#[source] StoreError),
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I've implemented the first two points, but I think I'll leave the last two as is. tbh if anything I'd prefer having far fewer error variants. As you note most/all of these are internal errors and they're not really actionable - so they should likely just be `anyhow::Error` with context added.

crates/store/src/errors.rs
Mirko-von-Leipzig: They probably can; but I think I'll yank the entire error enum instead. There isn't really any recourse to failure of database setup.

Mirko-von-Leipzig:
@PhilippGackstatter I've addressed _most_ of your comments; those I haven't I've left unresolved.

I agree with your suggestions given the current error model we have; though I'm tending towards leaning more into `anyhow` where possible in future refactorings. I'm not sure we can maintain all of these trivially - it feels pretty easy to miss something in this stack of error enums.

PhilippGackstatter (approved):
@Mirko-von-Leipzig Thanks for addressing the comments! Looks great!

I agree that using `anyhow` in the node might be a good choice for internal errors. 

Though I'm guessing that internal errors propagated up high enough would at some point have to lead to a panic as the node would not be able to recover from those (like a database no longer responding). So one question might be how the components higher in the call chain would identify those scenarios reliably for which concrete errors would be better than type-erased errors. Unless the idea is "internal errors = anyhow errors = crash", so any `anyhow::Error` would necessarily be an unrecoverable error leading to a panic eventually.

Mirko-von-Leipzig:
> @Mirko-von-Leipzig Thanks for addressing the comments! Looks great!
> 
> I agree that using `anyhow` in the node might be a good choice for internal errors.
> 
> Though I'm guessing that internal errors propagated up high enough would at some point have to lead to a panic as the node would not be able to recover from those (like a database no longer responding). So one question might be how the components higher in the call chain would identify those scenarios reliably for which concrete errors would be better than type-erased errors. Unless the idea is "internal errors = anyhow errors = crash", so any `anyhow::Error` would necessarily be an unrecoverable error leading to a panic eventually.

I suspect the answer will be that we need to perform this migration piecemeal to determine what the actual usage is. I'll elaborate further in #618 if you care to subscribe there :)

For now let's just get this PR merged in at least :)

PR #620, opened 2025-01-15
Improved RPC documentation

polydez:
We decided that we need only part with improved doc comments from the PR: https://github.com/0xPolygonMiden/miden-node/pull/542

This PR contains only needed part of the one behind.


crates/rpc-proto/proto/requests.proto
```diff
@@ -83,24 +85,31 @@ message GetBlockInputsRequest {
     repeated digest.Digest unauthenticated_notes = 3;
 }
 
+// Returns the data needed by the block producer to check validity of an incoming transaction.
```
Mirko-von-Leipzig, line 88:
```suggestion
// Request data required to validate a new transaction.
```

crates/rpc-proto/proto/requests.proto
```diff
@@ -83,24 +85,31 @@ message GetBlockInputsRequest {
...
 message GetTransactionInputsRequest {
+    // ID of the account against which a transaction is executed.
     account.AccountId account_id = 1;
+    // Array of nullifiers for all notes consumed by a transaction.
```
Mirko-von-Leipzig, line 92:
```suggestion
    // Set of nullifiers consumed by this transaction.
```

crates/rpc-proto/proto/requests.proto
```diff
@@ -83,24 +85,31 @@ message GetBlockInputsRequest {
...
     repeated digest.Digest nullifiers = 2;
+    // Array of unauthenticated note IDs to be checked for existence in the database.
```
Mirko-von-Leipzig, line 94: Clumsy attempt below.

We should strongly document what this is somewhere; even if its only for internal documentation.

```suggestion
    // Set of unauthenticated notes to check for existence onchain.
    //
    // These are notes which were not onchain at the state the transaction was proven, 
    // but could by now be present.
```

crates/rpc-proto/proto/requests.proto
```diff
@@ -83,24 +85,31 @@ message GetBlockInputsRequest {
...
     repeated digest.Digest unauthenticated_notes = 3;
 }
 
+// Submits proven transaction to the Miden network.
 message SubmitProvenTransactionRequest {
-    // Transaction encoded using miden's native format
+    // Transaction encoded using Miden's native format.
```
Mirko-von-Leipzig, line 100: Link to the format.

crates/rpc-proto/proto/requests.proto
```diff
@@ -83,24 +85,31 @@ message GetBlockInputsRequest {
...
     bytes transaction = 1;
 }
 
+// Returns a list of notes matching the provided note IDs.
 message GetNotesByIdRequest {
-    // List of NoteId's to be queried from the database
+    // List of NoteId's to be queried from the database.
```
Mirko-von-Leipzig, line 106:
```suggestion
    // List of notes to be queried from the database.
```

crates/rpc-proto/proto/requests.proto
```diff
@@ -83,24 +85,31 @@ message GetBlockInputsRequest {
...
     repeated digest.Digest note_ids = 1;
 }
 
+// Returns a list of Note inclusion proofs for the specified Note IDs.
 message GetNoteAuthenticationInfoRequest {
-    // List of NoteId's to be queried from the database
+    // List of NoteId's to be queried from the database.
```
Mirko-von-Leipzig, line 112:
```suggestion
    // List of notes to be queried from the database.
```

crates/rpc-proto/proto/requests.proto
```diff
@@ -126,9 +136,9 @@ message GetAccountStateDeltaRequest {
     fixed32 to_block_num = 3;
 }
 
-// Request message to get account proofs.
+// Returns the latest state proofs of accounts with the specified IDs.
```
Mirko-von-Leipzig, line 139:
```suggestion
// Returns the latest state proofs of the specified accounts.
```

Mirko-von-Leipzig (approved):
I've left a whole bunch of nits/suggestions. The vast majority are just stylistic/optional - feel free to disregard.

I do think we should get someone with a more technical/user documentation background to standardize these and to ensure they make sense from a more outside perspective.

crates/rpc-proto/proto/block.proto
```diff
@@ -4,34 +4,50 @@ package block;
 import "digest.proto";
 import "merkle.proto";
 
+// Represents a block header.
 message BlockHeader {
-    // specifies the version of the protocol.
+    // Specifies the version of the protocol.
```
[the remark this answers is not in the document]
polydez: Yes, indeed. We set version as `1` on genesis creation and all blocks for now use just the version from previous block.

crates/rpc-proto/proto/block.proto
```diff
@@ -4,34 +4,50 @@ package block;
...
+    // A commitment to account database.
     digest.Digest account_root = 5;
-    // a commitment to the nullifier database.
+
+    // A commitment to the nullifier database.
     digest.Digest nullifier_root = 6;
-    // a commitment to all notes created in the current block.
+
+    // A commitment to all notes created in the current block.
     digest.Digest note_root = 7;
```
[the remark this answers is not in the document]
bobbinth: Good question! I think we should definitely rename things like `tx_hash` into `tx_commitment ` (the only reason we used `hash` here is for brevity is I think is probably not the best motivation, or maybe we can say something like `tx_comm`).

There are also things like `chain_root` and `kernel_root` which in fact are not roots but commitments. So, these should probably be changed as well.

For things like `account_root`, `nullifier_root`, `note_root` the case is weaker as these are actually roots of trees. But maybe we should be forward-looking here.

If we do go through these changes, they should probably happen in `miden-base` first. Would you mind creating an issue there?

crates/rpc-proto/proto/requests.proto
```diff
@@ -14,11 +15,13 @@ message CheckNullifiersByPrefixRequest {
     // Number of bits used for nullifier prefix. Currently the only supported value is 16.
```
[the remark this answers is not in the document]
bobbinth: I think we have it here primarily for future compatibility reasons - i.e., if we enable querying for other prefix lengths (e.g., 8 or 24), we could introduce this without it being a breaking change.

proto/mmr.proto
```diff
@@ -3,7 +3,11 @@ package mmr;
 
 import "digest.proto";
 
+// Represents an MMR delta.
 message MmrDelta {
+    // The number of trees in the forest (latest block number + 1).
     uint64 forest = 1;
```
bobbinth, line 9: I think this should be: "The number of leaf nodes in the MMR". It also does contain information about the number of trees. This can be computed by counting the number of 1's in the `forest` value.

crates/rpc-proto/proto/block.proto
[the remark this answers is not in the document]
Mirko-von-Leipzig: https://github.com/0xPolygonMiden/miden-base/issues/1069

crates/rpc-proto/proto/note.proto
```diff
@@ -6,42 +6,83 @@ import "block.proto";
...
     digest.Digest note_id = 3;
+
+    // The note metadata.
     NoteMetadata metadata = 4;
+
+    // The note inclusion proof in the block.
     merkle.MerklePath merkle_path = 5;
+
     // This field will be present when the note is public.
     // details contain the `Note` in a serialized format.
```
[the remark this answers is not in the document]
bobbinth: This is basically the whole note serialized note (script, assets, inputs, serial number) - i.e., the data that hashes to `note_id`. This is only present for public notes because for private notes we don't know the underlying data.

crates/rpc-proto/proto/note.proto
```diff
@@ -6,42 +6,83 @@ import "block.proto";
...
+// Represents a note's metadata.
 message NoteMetadata {
+    // The account which sent the note.
     account.AccountId sender = 1;
+
+    // The type of the note (0b01 = public, 0b10 = private, 0b11 = encrypted).
     uint32 note_type = 2;
+
+    // A value which can be used by the recipient(s) to identify notes intended for them.
     fixed32 tag = 3;
```
bobbinth, line 18: Maybe worth adding a reference here to `miden_objects::notes::note_tag`.

crates/rpc-proto/proto/note.proto
```diff
@@ -6,42 +6,83 @@ import "block.proto";
...
+
     // This field will be present when the note is public.
     // details contain the `Note` in a serialized format.
     optional bytes details = 6;
 }
 
+// Represents proof of a note's inclusion in a block.
 message NoteInclusionInBlockProof {
+    // The ID of the note.
     digest.Digest note_id = 1;
```
bobbinth, line 54: Should we mention that this is a commitment to the underlying note data - i.e., "A unique identifier of the note which is a 32-byte commitment to the underlying note data."

crates/rpc-proto/proto/note.proto
```diff
@@ -6,42 +6,83 @@ import "block.proto";
...
     merkle.MerklePath merkle_path = 4;
 }
 
+// Represents proof of a note inclusion in the block.
 message NoteSyncRecord {
+    // The index of the note.
     uint32 note_index = 1;
+
+    // The ID of the note.
     digest.Digest note_id = 2;
```
bobbinth, line 72: Same comment as above to give a bit more detail about what note ID is.

proto/account.proto
```diff
@@ -3,30 +3,45 @@ package account;
 
 import "digest.proto";
 
+// An account ID.
 message AccountId {
     // A Miden account ID is a 120-bit value derived from the commitments to account code and
     // storage, and a random user-provided seed.
     bytes id = 1;
```
bobbinth, line 10: Should this comment be for the entire message rather than for the individual field?

crates/rpc-proto/proto/note.proto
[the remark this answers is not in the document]
polydez: I rewrote it like:
```rust
// Serialized details of the public note (empty for private notes).
```

crates/rpc-proto/proto/note.proto
```diff
@@ -6,42 +6,83 @@ import "block.proto";
...
+
+    // The note inclusion proof in the block.
     merkle.MerklePath merkle_path = 5;
+
     // This field will be present when the note is public.
     // details contain the `Note` in a serialized format.
     optional bytes details = 6;
 }
 
+// Represents proof of a note's inclusion in a block.
```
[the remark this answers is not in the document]
polydez: Rewrote it like:

```rust
// Represents a proof of note's inclusion in a block.
//
// Does not include proof of the block's inclusion in the chain.
```

I think, "does not" is clear enough, we don't need to emphasize this.

proto/account.proto
```diff
@@ -3,30 +3,45 @@ package account;
...
 
+// A summary of an account.
 message AccountSummary {
+    // The account ID.
     AccountId account_id = 1;
+
+    // The latest account hash, zero hash if the account doesn't exist.
     digest.Digest account_hash = 2;
+
+    // Merkle path to verify the account's inclusion in the MMR.
```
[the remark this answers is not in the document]
polydez: Good catch, thank you!

proto/account.proto
```diff
@@ -3,30 +3,45 @@ package account;
...
     uint32 block_num = 3;
 }
 
+// An account info.
 message AccountInfo {
+    // Account summary.
     AccountSummary summary = 1;
+
+    // Account details encoded using Miden native format.
```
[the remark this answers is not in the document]
polydez: We don't standardize format for serialized data yet, we have only `Serializable` implementations in Rust. Currently we can only refer to our implementation as description of the format. But I agree, we could add additional context here.

What do you think about this?

```rust
    // Account details encoded using [winter_utils::Serializable] implementation for
    // [miden_objects::accounts::Account].
```

crates/rpc-proto/proto/requests.proto
polydez: Looks good to me, thank you!

crates/rpc-proto/proto/note.proto
```diff
@@ -6,42 +6,83 @@ import "block.proto";
 import "digest.proto";
 import "merkle.proto";
 
+// Represents a note metadata.
 message NoteMetadata {
+    // The sender of the note.
     account.AccountId sender = 1;
+
+    // The type of the note (0b01 = public, 0b10 = private, 0b11 = encrypted).
```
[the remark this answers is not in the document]
polydez: We can do this, let's create an issue for this: https://github.com/0xPolygonMiden/miden-node/issues/624

crates/rpc-proto/proto/requests.proto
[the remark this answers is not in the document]
Mirko-von-Leipzig: Could/should we make this an enum then? As in, if we're only ever going to accept a limited set of values.

crates/rpc-proto/proto/account.proto
```diff
@@ -3,30 +3,48 @@ package account;
...
 import "digest.proto";
 
+// An account ID.
+//
+// A Miden account ID is a 120-bit value derived from the commitments to account code and storage,
+// and a random user-provided seed.
 message AccountId {
-    // A Miden account ID is a 120-bit value derived from the commitments to account code and
-    // storage, and a random user-provided seed.
+    // An account ID.
```
Mirko-von-Leipzig, line 11: We should also state the encoding i.e. how this should be interpreted into the two felts? Or is this somehow different to the prefix/suffix felt?

```suggestion
    // 15 bytes (120 bits) <..how to decode..>
```

crates/rpc-proto/proto/account.proto
```diff
@@ -3,30 +3,48 @@ package account;
 
 import "digest.proto";
 
+// An account ID.
+//
+// A Miden account ID is a 120-bit value derived from the commitments to account code and storage,
+// and a random user-provided seed.
```
Mirko-von-Leipzig, line 9: In the long run we should probably link to official documentation site which would specify the account ID:

```suggestion
// Uniquely identifies a specific account.
//
// A Miden account ID is a 120-bit value derived from the commitments to account code and storage,
// and a random user-provided seed.
```

crates/rpc-proto/proto/account.proto
```diff
@@ -3,30 +3,48 @@ package account;
...
 message AccountId {
-    // A Miden account ID is a 120-bit value derived from the commitments to account code and
-    // storage, and a random user-provided seed.
+    // An account ID.
     bytes id = 1;
 }
 
+// A summary of an account.
```
Mirko-von-Leipzig, line 15:
```suggestion
// The state of an account at a specific block height.
```

Mirko-von-Leipzig (approved):
Just a couple more suggestions; though I feel like we should just get these improvements in at this stage.

We could iterate on these for a long time still; and I feel we would be better served treating the RPC docs as semi-complete for now and instead focus on getting proper user facing documentation going. The proto definitions could then easily link to the actual documentation, with only additional details for things like serialization format.

crates/rpc-proto/proto/requests.proto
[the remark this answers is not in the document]
bobbinth: As long as adding more variants to an enum doesn't break backwards compatibility, I don't see why not. But also, I think this is probably fine as is. Either way, not a strong opinion.

bobbinth (approved):
Looks good! Thank you!

PR #621, opened 2025-01-15
test: add store seed for stress test

TomasArrachea:
This PR adds a binary program that seeds the store. The program creates new accounts, creates notes to consume assets on those accounts, builds the resulting blocks (using the BlockBuilder) and sends them to the store. Each block is built with 16 batches, where each batch contains 16 transactions.

The program also measures and prints the average insertion time of blocks into the store. For inserting 100k accounts, the resulting avg insertion time was ~1000 ms and finished in 7 minutes.


bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+    // Start store
+    let store = Store::init(store_config).await.context("Loading store").unwrap();
+    let _ = join_set.spawn(async move { store.serve().await.context("Serving store") }).id();
+
+    // Start block-producer
+    // TODO: is the full the BlockProducer needed? we should instantiate only a BlockBuilder
+    let block_producer = BlockProducer::init(block_producer_config)
+        .await
+        .context("Loading block-producer")
+        .unwrap();
```
bobbinth, line 38: I was actually thinking that we'd be constructing `Block` struct directly (we might need to do some extra bookkeeping to do this, specifically to keep data structures for updating account and nullifier tree roots) and that should be much faster than going through the block producer.

TomasArrachea:
This current implementation is mimicking the default values of `BlockBudget` (with `SERVER_MAX_BATCHES_PER_BLOCK = 4` and `SERVER_MAX_TXS_PER_BATCH = 2`) to build the transactions into the blocks. This means that to generate and store 1M accounts, 250k blocks are needed, which seems like a lot. Maybe using 16 tx per batch and 16 batches per block would make more sense? cc @bobbinth

bobbinth:
> This means that to generate and store 1M accounts, 250k blocks are needed, which seems like a lot. Maybe using 16 tx per batch and 16 batches per block would make more sense?

Yes, this is how I was thinking about it as well 16 batches per block and 16 transactions per batch. Assuming we can insert ~5 blocks/second (just a guess, not sure how close this is to reality), we should be able to "fill" 1M blocks in about 15 mins.

To do this though, we can't go via the block producer (but should be fine if we build `Block` structs directly).

TomasArrachea:
I just ran it locally with those settings (using the BlockBuilder) and got ~2.5 blocks/second, so storing the 7814 blocks would take about 50 mins. Will try to implement the Block building manually and check the difference.

bobbinth:
> Edit: actually inserting the blocks is much faster, main delay is grinding the accounts

Are you using the latest `next` branch (which relies on the latest `next` from `miden-base`). Account grinding should be pretty minimal there.

bobbinth:
> Are you using the latest `next` branch (which relies on the latest `next` from `miden-base`). Account grinding should be pretty minimal there.

btw, if account creation is a bottleneck with the `next` branch as well, we can always parallelize it (e.g., using `rayon`) to get a decent speed up. not sure if possible, but would be great to get to the point where insertions into the store are the bottleneck rather than data generation.

TomasArrachea:
Yes, even with latest version of `miden-base`, most of the binary runtime is spend on account grinding:
image

Using `rayon` sound good!

TomasArrachea:
Just pushed an PoC of a simple test: send a sync request to the loaded store and print the elapsed time. We were discussing with @igamigo about how much we would want this stress test to be configurable. We could load a config json file, specifying the db dump file (created with the binary) and the requests to send and measure. How were you thinking about it @bobbinth ?

bobbinth:
> implemented this and got 40s to load all 1M accounts (storing 7814 blocks).

Is 40 second correct here? it sounds unbelievably fast (about 200 blocks/second, right?).

> Just pushed an PoC of a simple test: send a sync request to the loaded store and print the elapsed time. We were discussing with @igamigo about how much we would want this stress test to be configurable. We could load a config json file, specifying the db dump file (created with the binary) and the requests to send and measure. How were you thinking about it @bobbinth ?

I think as the first step, would be good to be able to set the following:

- Number of accounts (or blocks) to be created - e.g., we may want to run this at some point with 1M, 10M, 100M etc. accounts.
- Location of the DB dump file so that we could pick it up for further stress testing.

In the future, we could add more config options (e.g., mix of public vs. private accounts and notes) - but that would be in follow-up PRs.

In terms of metrics, let's also start simple - I'm fine with tracking just block insertion times for now. We can also expand this in the future PRs.

TomasArrachea:
> Is 40 second correct here? it sounds unbelievably fast (about 200 blocks/second, right?).

Sorry, I don't think that number was correct. I recorded measurements again and got:
- 7814 block insertions in 700 seconds (about 10 blocks/second)
- avg insertion time (block building and storing): 26 ms

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,188 @@
...
+                MockProvenTxBuilder::with_account(faucet_id, Digest::default(), Digest::default())
+                    .output_notes(vec![OutputNote::Full(note.clone())])
+                    .build();
+
+            let consume_notes_txs =
+                MockProvenTxBuilder::with_account(account_id, Digest::default(), Digest::default())
+                    .unauthenticated_notes(vec![note])
+                    .build();
+            [create_notes_tx, consume_notes_txs]
+        })
```
bobbinth, line 176: This code is responsible for building a set of transactions, right? I wonder if it would make sense to move it into a stand-alone helper function as well.

Also, IIUC, we are consuming notes as "unauthenticated". While this work, it also will get the notes "erased" (i.e., this notes will never get recorded in the store and by extension, will also not be recorded in the nullifier database). So, we won't be putting as much stress on the system as we could be.

What we should do is issue notes in block $n$ and then consume this notes in block $n+1$. This way, they'll get recorded in block $n$ and their nullifiers will get recorded in block $n+1$ (and accounts will be created in block $n+1$ as well.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,188 @@
...
+    // Parallel account grinding and batch generation
+    (0..accounts_number)
+        .into_par_iter()
+        .map(|_| {
+            let (new_account, _) = AccountBuilder::new(init_seed)
+                .anchor(AccountIdAnchor::PRE_GENESIS)
+                .account_type(AccountType::RegularAccountImmutableCode)
+                .storage_mode(AccountStorageMode::Private)
+                .with_component(RpoFalcon512::new(key_pair.public_key()))
+                .with_component(BasicWallet)
```
bobbinth, line 152: I'd make a few changes here:

1. Use genesis block as the anchor (rather than pre-genesis).
2. Use a different `init_seed` for each account.

We could also use a different signature key for each account - but Falcon key generation is expensive (would take almost 0.5 seconds). So, re-using the keys as long as the seeds are different should be fine.

Also, separate from this PR, I'm curious why inserting identical accounts into the store didn't trigger any errors. Maybe something for us to take a look at in the future. cc @Mirko-von-Leipzig.

bobbinth (commented):
Looks good! Thank you. Not a full review yet - but I did leave some comments inline.

One general comment is that I'd probably try to break up the code more into smaller functions to make it a bit easier to follow.

bin/stress-test/src/main.rs
Mirko-von-Leipzig: Created #629 to look into why.

SantiagoPittella (commented):
I think we need a target in the Makefile to install this new CLI

crates/block-producer/Cargo.toml
```diff
@@ -16,33 +16,31 @@ workspace = true
...
 tokio = { workspace = true, features = ["rt-multi-thread", "net", "macros", "sync", "time"] }
 tokio-stream = { workspace = true, features = ["net"] }
 tonic = { workspace = true }
 tracing = { workspace = true }
-
-[dev-dependencies]
-assert_matches = { workspace = true}
 miden-air = { workspace = true }
+winterfell = { version = "0.10" }
+miden-objects = { workspace = true, features = ["testing"] }
```
SantiagoPittella, line 37: I think that we should not be using "testing" in the dependency, can we use the feature only the `dev-dependencies`?

crates/block-producer/src/store/mod.rs
```diff
@@ -126,7 +126,7 @@ impl TryFrom<GetTransactionInputsResponse> for TransactionInputs {
 /// Essentially just a thin wrapper around the generated gRPC client which improves type safety.
 #[derive(Clone)]
 pub struct StoreClient {
-    inner: store_client::ApiClient<Channel>,
+    pub inner: store_client::ApiClient<Channel>,
```
SantiagoPittella, line 129: If we are introducing the "testing" feature, we can create a method under that feature to get a mut ref of `inner`.

SantiagoPittella (commented):
Saw that you re-introduced the `testing` feature for the block-producer crate. In that case, the crate's documentation should be updated.

crates/block-producer/src/store/mod.rs
TomasArrachea: Good idea, although in the end the inner attribute is not needed to be public, so will just restore it as it was before.

crates/block-producer/Cargo.toml
TomasArrachea: yes!

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+#[derive(Subcommand)]
+pub enum Command {
+    SeedStore {
+        #[arg(short, long, value_name = "DUMP_FILE", default_value = "./miden-store.sqlite3")]
+        dump_file: PathBuf,
+
+        #[arg(short, long, value_name = "ACCOUNTS_NUMBER")]
+        accounts_number: usize,
```
bobbinth, line 49: nit: I would rename this to `num_accounts`.

bobbinth (commented):
Left a few more small comments inline.

bin/stress-test/src/main.rs
TomasArrachea: Right! Updated the binary to consume authenticated notes (in [f009a8a](https://github.com/0xPolygonMiden/miden-node/pull/621/commits/f009a8a3dc1f217ecff84f80b55bad53b3780f6d)) and resulted in significantly lower insertion times.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,283 @@
...
+
+const BATCHES_PER_BLOCK: usize = 16;
+const TRANSACTIONS_PER_BATCH: usize = 16;
+const NOTES_PER_TRANSACTION: usize = MAX_OUTPUT_NOTES_PER_BATCH / TRANSACTIONS_PER_BATCH;
+
+/// Create and store blocks into the store. Create a given number of accounts, where each account
+/// consumes a note created from a faucet. The cli accepts the following parameters:
+/// - `dump_file`: Path to the store database file.
+/// - `num_accounts`: Number of accounts to create.
+/// - `genesis_file`: Path to the genesis file of the store.
```
igamigo, line 65: nit: I think these could also be doc comments in the `SeedStore` command as well to make Clap show help accordingly

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,283 @@
...
+        .with_component(
+            BasicFungibleFaucet::new(TokenSymbol::new("TEST").unwrap(), 2, Felt::new(100_000))
+                .unwrap(),
+        )
+        .build()
+        .unwrap();
+    new_faucet.id()
+}
+
+/// Create a new note with a single fungible asset.
```
igamigo, line 98: nit:
```suggestion
/// Create a new note containing 10 tokens of the fungible asset associated with the specified `faucet_id`.
```

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,283 @@
...
+    let rng = RpoRandomCoin::new(coin_seed.map(Felt::new));
+    NoteBuilder::new(faucet_id, rng)
+        .add_assets(vec![asset])
+        .build(&TransactionKernel::assembler())
+        .unwrap()
+}
+
+/// Create a new account with a given public key and anchor block. Generates the seed from the given
+/// index.
+fn create_account(anchor_block: &BlockHeader, public_key: PublicKey, index: usize) -> AccountId {
```
igamigo, line 111: Instead of `usize`, I think this could be `u64` directly, unless there's something I'm missing

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,283 @@
...
+        .chunks(TRANSACTIONS_PER_BATCH)
+        .for_each_with(batch_sender, |sender, txs| {
+            let batch =
+                TransactionBatch::new(txs.iter().collect::<Vec<_>>(), Default::default()).unwrap();
+            sender.blocking_send(batch).unwrap();
+        });
+}
+
+/// Seed the store with a given number of accounts.
+async fn seed_store(dump_file: &Path, num_accounts: usize, genesis_file: &Path) {
```
igamigo, line 222: nit: I think this could be higher in the file to make the code a bit easier to follow

TomasArrachea: Thanks! done in [e545d92](https://github.com/0xPolygonMiden/miden-node/pull/621/commits/e545d92e1f8630daf52319b7b98ca1802cbac7da)

Makefile
```diff
@@ -73,6 +73,10 @@ install-node: ## Installs node
 install-faucet: ## Installs faucet
 	${BUILD_PROTO} cargo install --path bin/faucet --locked
 
+.PHONY: install-stress-test
+install-stress-test: ## Installs faucet
```
SantiagoPittella, line 77:
```suggestion
install-stress-test: ## Installs stress test binary
```

Makefile
```diff
@@ -73,6 +73,10 @@ install-node: ## Installs node
...
+	${BUILD_PROTO} cargo install --path bin/stress-test --locked
```
SantiagoPittella, line 78: Since the crate doesn't have a `build.rs` I'm not sure that we need `${BUILD_PROTO}` here.

crates/block-producer/Cargo.toml
```diff
@@ -16,33 +16,31 @@ workspace = true
...
-[dev-dependencies]
-assert_matches = { workspace = true}
 miden-air = { workspace = true }
+winterfell = { version = "0.11" }
+miden-objects = { workspace = true }
+rand_chacha = { version = "0.3", default-features = false }
 miden-lib = { workspace = true, features = ["testing"] }
 miden-node-test-macro = { path = "../test-macro" }
-miden-objects = { workspace = true, features = ["testing"] }
 miden-tx = { workspace = true, features = ["testing"] }
```
SantiagoPittella, line 41: We should alphabetize the deps.

Also, we are using `"testing"` for some dependencies, maybe we need to use it only in `[dev-dependencies]`?

TomasArrachea: yes! didn't realize this was no longer needed

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,287 @@
...
+    (insertion_time, num_insertions)
+}
+
+/// Create a given number of notes and group them into transactions and batches.
+/// The batches are sent to the block builder.
+fn generate_note_batches(
+    num_notes: usize,
+    faucet_id: AccountId,
+    batch_sender: UnboundedSender<TransactionBatch>,
+) -> Vec<Note> {
```
bobbinth, line 230: I'm not sure I fully understood how this works. Wouldn't we need to create just 1 transaction per block to create notes? That is, this transaction would be executed against the faucet account and would output 256 notes directed to 256 accounts that were created in the previous block. Or does this cause issues somehow?

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,287 @@
...
+        generate_account_batches(num_accounts, &notes, batch_sender, &genesis_header);
+    })
+    .await
+    .unwrap();
+    let (insertion_time, num_insertions) = db_task.await.unwrap();
+    println!(
+        "Consumed notes: inserted {} blocks with avg insertion time {} ms",
+        num_insertions,
+        (insertion_time / num_insertions).as_millis()
+    );
```
bobbinth, line 139: Would be cool to print out average block insertion time per some number of blocks (i.e., for every 1000 blocks). This way, we'd see how the insertion time changes the the number of blocks in the store increases.

Also, would be cool to print out the store database file size, ideally for each block interval as well.

bobbinth (commented):
Thank you! Looks good. I left some comments inline - but a more general comment is that I find the current structure a bit difficult to follow. Would a more "linear" approach not work? For example, something like this:

1. Create a genesis block with the faucet account.
2. Create two sets of accounts $s_0$ and $s_1$, initially empty.
3. Create two sets of notes $p_0$ and $p_1$, initially empty.
4. Start a loop to create $n$ blocks (the loop would have $n$ iterations). In this loop:
  a. Create accounts to populate $s_0$ but do not record them on chain.
  b. Create a transaction that outputs a set of notes $p_0$ directed from the faucet to the accounts in $s_0$.
  c. Create transactions to consume notes in $p_1$ against accounts in $s_1$ (this will be a noop on the first iteration).
  d. Build a block from the above transactions and insert it into the store.
  e. Assing $s_1 \leftarrow s_0$ and $p_1 \leftarrow p_0$.

TomasArrachea:
Closing this in favor of https://github.com/0xPolygonMiden/miden-node/pull/657

PR #625, opened 2025-01-16
ci: check for unused dependencies

Mirko-von-Leipzig:
This PR adds CI job to reject unused dependencies using [cargo machete](https://github.com/bnjbvr/cargo-machete).

In addition this PR also removes two unused dependencies.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -68,6 +68,14 @@ jobs:
       - name: check rust versions
         run: ./scripts/check-rust-version.sh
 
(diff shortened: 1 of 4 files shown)


Mirko-von-Leipzig:
Note that there is also `cargo-udeps` which is a similar tool. Neither is 100% correct, but `machete` at least has a github action.

bobbinth (approved):
Looks good! Thank you!

We should probably propagate this to other repos: cc @PhilippGackstatter, @plafer, @igamigo.

PR #631, opened 2025-01-20
docs: align readme rpc docs

### LINKED ISSUE #626: Update RPC docs in readme
These have drifted from the actual proto definitions.

Mirko-von-Leipzig:
This PR re-aligns the various RPC docs with the proto files.

I've opted to only provide a summary and not a full parameter doxygen style documentation as there was before as I don't think its trivial to keep aligned with reality, nor sufficient -- users will always have to use and/or read the proto files.

Closes #626

### DIFF
--- a/crates/rpc/README.md
+++ b/crates/rpc/README.md
@@ -1,129 +1,88 @@
 # Miden node RPC
 
-The **RPC** is an externally-facing component through which clients can interact with the node. It receives client requests
-(e.g., to synchronize with the latest state of the chain, or to submit transactions), performs basic validation,
-and forwards the requests to the appropriate components.
-**RPC** is one of components of the [Miden node](..).
+Contains the code defining the [Miden node's RPC component](/README.md#architecture). This component serves the
+user-facing [gRPC](https://grpc.io) methods used to submit transactions and sync with the state of the network.
 
-## Architecture
+This is the **only** set of node RPC methods intended to be publicly available.
 
-`TODO`
+For more information on the installation and operation of this component, please see the [node's readme](/README.md).
 
-## Usage
+## API overview
 
-### Installing the RPC
+The full gRPC method definitions can be found in the [rpc-proto](../rpc-proto/README.md) crate.
 
-The RPC can be installed and run as part of [Miden node](../README.md#installing-the-node).
-
-## API
-
-The **RPC** serves connections using the [gRPC protocol](https://grpc.io) on a port, set in the previously mentioned configuration file.
-Here is a brief description of supported methods.
+<!--toc:start-->
+- [CheckNullifiers](#checknullifiers)
+- [CheckNullifiersByPrefix](#checknullifiersbyprefix)
+- [GetAccountDetails](#getaccountdetails)
+- [GetAccountProofs](#getaccountproofs)
+- [GetAccountStateDelta](#getaccountstatedelta)
+- [GetBlockByNumber](#getblockbynumber)
+- [GetBlockHeaderByNumber](#getblockheaderbynumber)
+- [GetNotesById](#getnotesbyid)
+- [SubmitProvenTransaction](#submitproventransaction)
+- [SyncNotes](#syncnotes)
+- [SyncState](#syncstate)
+<!--toc:end-->
 
 ### CheckNullifiers
 
 Gets a list of proofs for given nullifier hashes, each proof as a sparse Merkle Tree.
 
-**Parameters:**
+### CheckNullifiersByPrefix
 
-- `nullifiers`: `[Digest]` – array of nullifier hashes.
-
-**Returns:**
-
-- `proofs`: `[NullifierProof]` – array of nullifier proofs, positions correspond to the ones in request.
-
-### GetBlockHeaderByNumber
-
-Retrieves block header by given block number, optionally alongside a Merkle path and the current chain length to validate its inclusion.
-
-**Parameters**
-
-- `block_num`: `uint32` _(optional)_ – the block number of the target block. If not provided, the latest known block will be returned.
-
-**Returns:**
-
-- `block_header`: `BlockHeader` – block header.
-
-### GetBlockByNumber
-
-Retrieves block data by given block number.
-
-**Parameters**
-
-- `block_num`: `uint32` – the block number of the target block.
-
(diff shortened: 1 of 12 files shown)


crates/rpc/README.md
```diff
@@ -1,129 +1,112 @@
...
-**Parameters**
-
-- `note_ids`: `[NoteId]` - list of IDs of the notes we want to query.
+---
 
-**Returns**
+### SubmitProvenTransaction
 
-- `notes`: `[Note]` - List of notes matching the list of requested NoteIds.
+Submits proven transaction to the Miden network.
```
igamigo, line 81: I know this might have been deliberately left aside (and/or it might not be worth mentioning here) but I think it could be good to mention that this returns the current block height

Mirko-von-Leipzig: We can add it in; tbh I find these docs low-value as I sort of mention in [this discussion](https://github.com/0xPolygonMiden/miden-node/discussions/630). They're neither automatically sync'd, nor fully descriptive so they end up just defaulting to "meh".

crates/rpc/README.md
```diff
@@ -1,129 +1,112 @@
...
 
-Returns the latest state of an account with the specified ID.
+### SyncNotes
 
-**Parameters**
+Note synchronization request.
 
-- `account_id`: `AccountId` – account ID.
+Specifies note tags that client is interested in. The server will return the first block which contains a note matching
+`note_tags` or the chain tip.
```
bobbinth, line 90: I think the second sentence should be more like: "The server will return the matching notes from the first block which contains a note matching `note_tags`."

I would also maybe make this description more detailed - similar to how we have it for `SyncState` endpoint (and probably propagate these details to the protobuf files as well).

crates/store/README.md
```diff
@@ -1,163 +1,133 @@
...
+- [GetTransactionInputs](#gettransactioninputs)
+- [SyncNotes](#syncnotes)
+- [SyncState](#syncstate)
+<!--toc:end-->
+
+---
 
 ### ApplyBlock
 
 Applies changes of a new block to the DB and in-memory data structures.
```
bobbinth, line 35: I would also maybe mention that this also saves the raw block data as a flat file.

crates/store/README.md
```diff
@@ -1,163 +1,133 @@
...
+---
 
-- `block`: `BlockHeader` – block header ([src](../proto/proto/block_header.proto)).
-- `accounts`: `[AccountUpdate]` – a list of account updates.
-- `nullifiers`: `[Digest]` – a list of nullifier hashes.
-- `notes`: `[NoteCreated]` – a list of notes created.
+### CheckNullifiers
 
-**Returns**
+Gets a list of proofs for given nullifier hashes, each proof as a sparse Merkle Tree.
```
bobbinth, line 41: Same comment as for the analogous endpoint for the RPC.

crates/store/README.md
```diff
@@ -1,163 +1,133 @@
...
-**Returns:**
+Retrieves block header by given block number. Optionally, it also returns the MMR path and current chain length to
+authenticate the block's inclusion.
 
-- `block`: `Block` – block data encoded using [winter_utils::Serializable](https://github.com/facebook/winterfell/blob/main/utils/core/src/serde/mod.rs#L26) implementation for [miden_objects::block::Block](https://github.com/0xPolygonMiden/miden-base/blob/main/objects/src/block/mod.rs#L43).
+---
 
 ### GetBlockInputs
 
 Returns data required to prove the next block.
```
bobbinth, line 84: For this and similar endpoints, should we mention that they are meant to be used by the block producer?

crates/store/README.md
```diff
@@ -1,163 +1,133 @@
...
-
-This request returns the next block containing requested data. It also returns `chain_tip` which is the latest block number in the chain.
-Client is expected to repeat these requests in a loop until `response.block_header.block_num == response.chain_tip`, at which point the client is fully synchronized with the chain.
-
-Each request also returns info about new notes, nullifiers etc. created. It also returns Chain MMR delta that can be used to update the state of Chain MMR.
-This includes both chain MMR peaks and chain MMR nodes.
-
-For preserving some degree of privacy, note tags and nullifiers filters contain only high part of hashes. Thus, returned data
-contains excessive notes and nullifiers, client can make additional filtering of that data on its side.
+Returns info which can be used by the client to sync up to the latest state of the chain for the objects (accounts,
```
bobbinth, line 117: Same comment as for the analogous endpoint for the RPC.

crates/store/README.md
```diff
@@ -1,163 +1,133 @@
...
+Gets a list of proofs for given nullifier hashes, each proof as a sparse Merkle Tree.
 
-This method doesn't return any data.
+---
 
-### CheckNullifiers
+### CheckNullifiersByPrefix
 
-Get a list of proofs for given nullifier hashes, each proof as a sparse Merkle Tree
+Returns a list of nullifiers that match the specified prefixes and are recorded in the node.
```
bobbinth, line 47: Same comment as for the analogous endpoint for the RPC.

crates/store/README.md
```diff
@@ -1,163 +1,133 @@
...
 
-**Returns:**
+Returns delta of the account states in the range from `from_block_num` (exclusive) to `to_block_num` (inclusive).
 
-- `block_header`: `BlockHeader` – block header.
+---
 
 ### GetBlockByNumber
 
 Retrieves block data by given block number.
```
bobbinth, line 71: Same comment as for the analogous endpoint for the RPC.

bobbinth (approved):
Looks good! Thank you! I left some small comments inline.

Mirko-von-Leipzig:
I've addressed all comments I believe. In general there are probably still discrepancies between the proto service/request/response/type comments and the readme's.

imo we should only "publicly" document the RPC API as its the only real external one; and leave the internal APIs undocumented, at least for now. Devs can use the proto files and code docs.

bobbinth (approved):
Looks good! Thank you!

> imo we should only "publicly" document the RPC API as its the only real external one; and leave the internal APIs undocumented, at least for now. Devs can use the proto files and code docs.

Yes, agree with this. We should move toward this state.

PR #632, opened 2025-01-20
refactor: Use miden-objects's `BlockNumber`

igamigo:


igamigo:
Ran the client integration tests from [this branch](https://github.com/0xPolygonMiden/miden-client/pull/677), can confirm they are all passing.

crates/store/src/genesis.rs
```diff
@@ -54,7 +54,7 @@ impl GenesisState {
         let header = BlockHeader::new(
             self.version,
             Digest::default(),
-            GENESIS_BLOCK,
+            GENESIS_BLOCK.into(),
```
bobbinth, line 57: Not for this PR: but we probably should have made `GENESIS_BLOCK` into `BlockNumber` in `miden-base`.

bobbinth (approved):
Looks good! Thank you! I left one non-blocking comment inline.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -149,14 +149,14 @@ impl BlockBuilder {
...
         let new_block_header = self.block_kernel.prove(block_header_witness)?;
-        let block_num = new_block_header.block_num();
 
         // TODO: return an error?
         let block =
             Block::new(new_block_header, updated_accounts, output_notes, produced_nullifiers)
                 .expect("invalid block components");
 
         let block_hash = block.hash();
+        let block_num = new_block_header.block_num().as_u32();
```
Mirko-von-Leipzig, line 159: Does `BlockNumber` not implement `Display`? I think that's all this was used for or no?

crates/block-producer/src/mempool/tests.rs
```diff
@@ -7,11 +8,11 @@ use crate::test_utils::MockProvenTxBuilder;
 impl Mempool {
     fn for_tests() -> Self {
         Self::new(
-            BlockNumber::new(0),
+            BlockNumber::from(0),
```
Mirko-von-Leipzig, line 11: I believe there is also `BlockNumber::default()` or `BlockNumber::GENESIS`

crates/store/src/genesis.rs
Mirko-von-Leipzig: It is there as `BlockNumber::GENESIS`

Mirko-von-Leipzig (approved):

crates/store/src/genesis.rs
bobbinth: Actually, I created https://github.com/0xPolygonMiden/miden-base/pull/1088 to get rid of `GENESIS_BLOCK` constant. So, let's use `BlockNumber::GENESIS` here.

crates/block-producer/src/block_builder/mod.rs
igamigo: It does, updated the calls.

crates/block-producer/src/mempool/tests.rs
igamigo: I had missed `BlockNumber::GENESIS`, but default looked a bit opaque which is why I changed it to this originally.

crates/store/src/genesis.rs
igamigo: I had thought of removing this on `miden-base` and was planning on mentioning it. I had also missed `BlockNumber::GENESIS` originally when doing this PR. Updated.

PR #636, opened 2025-01-22
reformat: Use latest next from miden-base; rename modules to singular

igamigo:
Also updates `winterfell` to `0.11`
Follows [these conventions](https://github.com/0xPolygonMiden/miden-base/pull/1091#issue-2802244144)


bin/faucet/src/error.rs
bobbinth, line 1: I would keep the name the same here.

bobbinth (approved):
Looks good! Thank you! I left just one small comment inline.

PR #640, opened 2025-01-23
test: add sync request benchmark

TomasArrachea:
This PR adds a cli command to execute a sync request and measure the performance. Using the load-state command introduced in [621](https://github.com/0xPolygonMiden/miden-node/pull/621), the idea is to load the store and then use that dump file to test the sync request under different load conditions.

Also, this PR updates the LoadState command to generate both Public and Private accounts. Also, it now generates two additional dump files:
- insertion times measured on loading
- accounts ids of the public accounts loaded


bobbinth:
Not a review yet, but I would probably split this up into 2 PRs: one to add more functionality for generating the test data and the other one for benchmarking state sync endpoint performance.

bobbinth:
Is this superseded by #661?

PR #643, opened 2025-01-24
Refactored account tree in order to store last `N` reverse updates, compute opening for recent blocks

polydez:
In this PR we refactored account's in-memory storage in order to store last $N$ reverse updates alongside with the latest account's SMT. Updates are persistent (in files), old updates are deleted. We also reorganize storage files by putting them into the `storage` directory. Also implemented method `compute_opening` which constructs opening for account for any recent block.

This solution is not expected to be significantly slower than constructing opening for the SMT itself, but we will measure this in benchmarks.

We will need to reimplement calculation of reverse mutation set in order to keep transactional approach of applying changes (the current solution might cause inconsistent DB and in-memory states if writing of update failed).


Mirko-von-Leipzig:
@polydez is this ready for review? Still marked as a `draft` so just checking.

crates/store/Cargo.toml
```diff
@@ -17,6 +17,7 @@ workspace = true
 [dependencies]
 deadpool-sqlite = { version = "0.9.0", features = ["rt_tokio_1"] }
 hex = { version = "0.4" }
+miden-crypto = { version = "0.13" }
```
Mirko-von-Leipzig, line 20: Should this not stay the workspace version so it gets the patch?

crates/store/src/account_tree.rs
```diff
@@ -0,0 +1,441 @@
...
+    fn update_index(&self, block_num: BlockNumber) -> Option<usize> {
+        if block_num >= self.latest_block_num {
+            return None;
+        }
+
+        Some(self.latest_block_num.as_usize() - block_num.as_usize() - 1)
+    }
+}
+
+#[allow(async_fn_in_trait)]
```
Mirko-von-Leipzig, line 303: Are we sure we want to suppress this?

crates/store/src/account_tree.rs
```diff
@@ -0,0 +1,441 @@
...
+pub trait PersistentUpdatesStorage {
+    async fn load(&self, block_num: BlockNumber) -> Result<Option<Update>, DatabaseError>;
+    async fn save(&self, block_num: BlockNumber, update: &Update) -> Result<(), DatabaseError>;
+    async fn remove(&self, block_num: BlockNumber) -> Result<(), DatabaseError>;
+}
```
Mirko-von-Leipzig, line 308: Why do we need a trait? If its only for test purposes then I would prefer using something like `tempfs` for testing over having trait/generics

Mirko-von-Leipzig (commented):
Some initial comments; still need to review the actual tree.

Do we need persistent storage for this? Could we not just store this in-memory only in a ring-buffer? Or are these huge? If they're huge then we will also have disk IO issues I think.

crates/store/Cargo.toml
polydez: It gets the patch while the crate has the same version (`version.minor.*` by semver) as the patched one. We usually specify dependencies in `Cargo.toml` of specific crate, if we use them only there. But I'm not sure this is the best approach, I would rather specify all versions in BoM (workspace's `Cargo.toml`).

Mirko-von-Leipzig: Yeah it does get a bit weird.

crates/store/src/account_tree.rs
polydez: Yes, we need this trait for testing and benchmarking. This approach also allows us to mockup implementations of storage in tests.

crates/store/Cargo.toml
polydez: As far as I know, @bobbinth has different opinion on this, we maybe need to discuss which approach is better to use. I have motivation to specify all versions in workspace's `Cargo.toml`: I expect that this will help us to reduce possible incompatibilities between different crates' versions in our project (when the same structure/trait from different dependency versions is used across different crates of the same workspace). This approach is also called as "bill of materials" (BoM) and helps developers to manage dependency versions in subprojects of some complex projects. This won't help to eliminate all incompatibilities (when dependencies outside of the workspace depend on different dependency versions), but I do think, we should use it at least to simplify dependency management within our workspace(s).

polydez:
I made benchmarking for my solution of computation of opening for one of recent blocks, comparing with opening in the latest account SMT. 

For 10k accounts and 100 blocks:

```
SMT opening elapsed (for 10000 accounts, average): 108 ms
AccountTree `compute_opening` elapsed (for 10000 accounts, average): 321 ms
```

So the `compute_opening` over reverse updates works 3 times slower than opening in single SMT. This is slower than I expected, but I think it is pretty good since it takes just ~0.03 ms for computing opening of a single account.

polydez:
> Do we need persistent storage for this? Could we not just store this in-memory only in a ring-buffer? Or are these huge? If they're huge then we will also have disk IO issues I think.

I apologize, I forgot to answer this question. We need persistent storage in order to reconstruct in-memory updates on node restart. Otherwise, we would need to wait for $N$ blocks to be applied until will be able to serve our future endpoint for up to $N$ recent proofs.

Mirko-von-Leipzig:
> > Do we need persistent storage for this? Could we not just store this in-memory only in a ring-buffer? Or are these huge? If they're huge then we will also have disk IO issues I think.
> 
> I apologize, I forgot to answer this question. We need persistent storage in order to reconstruct in-memory updates on node restart. Otherwise, we would need to wait for N blocks to be applied until will be able to serve our future endpoint for up to N recent proofs.

Is the alternative of rebuilding them on restart not possible? I'd prefer to minimize disk IO as much as possible - every new file is something that can lead to strange failures.

polydez:
> Is the alternative of rebuilding them on restart not possible? I'd prefer to minimize disk IO as much as possible - every new file is something that can lead to strange failures.

I was thinking about this and it seems to be too expensive: we need to construct reverse mutations sets and it's possible only when we know previous account's hash for each recent block. Blocks consist of all required information, but this might require to process blocks from the genesis up to the latest block because of accounts which might be changed only in first blocks. In the past we had previous account's hash in update `BlockAccountUpdate` structure (if I remember correctly), but now we haven't.

PR #646, opened 2025-01-27
Update itertools and axum versions

### LINKED ISSUE #642: Update dependencies to latest versions
We should update dependencies to the latest versions. At least the following dependencies are currently out of date: `itertools`, `rsqlite`.

sergerad:
Closes #642

* Updates itertools and axum dependancy versions
* Adds itertools as a workspace dependancy

Noting that rusqlite must remain at 0.32.1 due to deadpool-sqlite's version requirements [here](https://crates.io/crates/deadpool-sqlite/0.9.0/dependencies).
```
╰─➤  cargo upgrades
miden-node-store
dependency  current  upgrade
rusqlite    0.32.1   0.33.0
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -197,8 +197,35 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
 dependencies = [
  "async-trait",
- "axum-core",
+ "axum-core 0.4.5",
  "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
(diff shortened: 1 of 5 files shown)


Mirko-von-Leipzig (commented):
Changes itself are good. Some downsides are that we have multiple versions of `axum` et al in our lockfile now because `tonic` is still using the previous version.

Some extra compile time or we could wait for `tonic` to upgrade. Though I don't mind much either way.

bobbinth (approved):
Looks good! Thank you!

> Some downsides are that we have multiple versions of `axum` et al in our lockfile now because `tonic` is still using the previous version.

I think this is OK (or at least I don't see big downsides to this).

PR #648, opened 2025-01-28
ci: add workspace-lints to Makefile and lint.yml

### LINKED ISSUE #635: [ci] Check for packages for lint compliance
Add a CI job for https://crates.io/crates/cargo-workspace-lints.

This could be optimized by only running for PRs touching the root `Cargo.toml` to restrict it to only running for new crates/binaries.

sergerad:
Closes #635

Adds a make target for performing workspace lint check in all packages and runs this in PR workflow. See [crate](https://crates.io/crates/cargo-workspace-lints).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -59,11 +59,10 @@ jobs:
     timeout-minutes: 5
     steps:
       - uses: actions/checkout@v4
-      - uses: Swatinem/rust-cache@v2
+      - uses: taiki-e/install-action@v2
         with:
-          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
+          tool: cargo-workspace-lints
       - run: |
-          cargo install cargo-workspace-lints --locked
           make workspace-check
 
   doc:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -54,6 +54,18 @@ jobs:
           tool: taplo-cli
       - run: make toml-check
 
+  workspace-lints:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
+      - run: |
+          cargo install cargo-workspace-lints --locked
+          make workspace-check
+
   doc:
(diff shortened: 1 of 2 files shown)


sergerad:
@Mirko-von-Leipzig I haven't done this optimization:
> This could be optimized by only running for PRs touching the root Cargo.toml to restrict it to only running for new crates/binaries.

Because AFAIK there is no tidy way to do this at the job-level (just workflow level). See [this](https://stackoverflow.com/questions/70708306/github-actions-run-step-job-in-a-workflow-if-changes-happen-in-specific-folde) (would suggest not to do what the answer does there).

Tested locally for negative by editing one of the packages:
```
╰─➤  make workspace-check
cargo workspace-lints
Failed to validate:
Failing packages:
* Package path+file:///Users/sergeradinovich/Source/sergerad/miden-node/bin/node#miden-node@0.8.0:
     No `workspace.lints` field found
make: *** [workspace-check] Error 1
```

Mirko-von-Leipzig (approved):
LGTM thank you.

We could consider https://github.com/taiki-e/install-action to install the binary; but probably the caching is actually better?

Though maybe we don't want to keep increasing cache usage.

Mirko-von-Leipzig:
> @Mirko-von-Leipzig I haven't done this optimization:
> 
> > This could be optimized by only running for PRs touching the root Cargo.toml to restrict it to only running for new crates/binaries.
> 
> Because AFAIK there is no tidy way to do this at the job-level (just workflow level). See [this](https://stackoverflow.com/questions/70708306/github-actions-run-step-job-in-a-workflow-if-changes-happen-in-specific-folde) (would suggest not to do what the answer does there).
> 
> Tested locally for negative by editing one of the packages:
> 
> ```
> ╰─➤  make workspace-check
> cargo workspace-lints
> Failed to validate:
> Failing packages:
> * Package path+file:///Users/sergeradinovich/Source/sergerad/miden-node/bin/node#miden-node@0.8.0:
>      No `workspace.lints` field found
> make: *** [workspace-check] Error 1
> ```

Ack; yeah makes sense. No point in complicating things then :) This should be cheap enough.

sergerad:
> LGTM thank you.
> 
> We could consider https://github.com/taiki-e/install-action to install the binary; but probably the caching is actually better?
> 
> Though maybe we don't want to keep increasing cache usage.

Yea sorry should have done this in the first place at least for consistency with the toml job from the other day. Have pushed that change now

PR #654, opened 2025-01-29
chore: remove Endpoint and Protocol types

### LINKED ISSUE #651: Remove `Endpoint`
This type is a leaky abstraction. The `http`/`url` specification is long and many parts are optional.

I propose removing the type entirely, and letting our `http` server/client frameworks deal with parsing whatever configuration was provided. aka just use `String` or equivalent until it hits the library we use.

Briefly discussed in #647.

sergerad:
Closes #651

* Remove Endpoint and Protocol types from util crate
* Use url::Url for handling endpoints both in TOML configuration and runtime

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changes
 
 - [BREAKING] Updated minimum Rust version to 1.84.
+- [BREAKING] Replace `Endpoint` and `Protocol` types with `url::Url` and update node TOML configuration to treat endpoints as a single string (#654).
 
 ## v0.7.2 (2025-01-29)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Changes
 
 - [BREAKING] Updated minimum Rust version to 1.84.
-- [BREAKING] Replace `Endpoint` and `Protocol` types with `url::Url` and update node TOML configuration to treat endpoints as a single string (#654).
+- [BREAKING] `Endpoint` configuration simplified to a single string (#654).
 
 ## v0.7.2 (2025-01-29)
 
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -200,7 +200,7 @@ impl FaucetClient {
 pub async fn initialize_faucet_client(
     config: &FaucetConfig,
 ) -> Result<(ApiClient<Channel>, BlockHeader, ChainMmr), ClientError> {
-    let endpoint = tonic::transport::Endpoint::try_from(config.node_url.clone())
+    let endpoint = tonic::transport::Endpoint::try_from(config.node_url.to_string())
         .context("Failed to parse node URL from configuration file")?
         .timeout(Duration::from_millis(config.timeout_ms));
 
--- a/bin/faucet/src/config.rs
+++ b/bin/faucet/src/config.rs
@@ -56,8 +56,9 @@ impl Default for FaucetConfig {
 
 #[cfg(test)]
 mod tests {
-    use super::FaucetConfig;
     use tokio::net::TcpListener;
+
+    use super::FaucetConfig;
 
     #[tokio::test]
     async fn default_faucet_config() {
--- a/bin/faucet/src/config.rs
+++ b/bin/faucet/src/config.rs
@@ -43,12 +43,23 @@ impl Display for FaucetConfig {
 impl Default for FaucetConfig {
     fn default() -> Self {
         Self {
-            endpoint: Url::parse(format!("0.0.0.0:{DEFAULT_FAUCET_SERVER_PORT}").as_str()).unwrap(),
+            endpoint: Url::parse(format!("http://0.0.0.0:{DEFAULT_FAUCET_SERVER_PORT}").as_str())
+                .unwrap(),
             node_url: Url::parse(format!("http://127.0.0.1:{DEFAULT_NODE_RPC_PORT}").as_str())
                 .unwrap(),
             timeout_ms: DEFAULT_RPC_TIMEOUT_MS,
             asset_amount_options: vec![100, 500, 1000],
             faucet_account_path: DEFAULT_FAUCET_ACCOUNT_PATH.into(),
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::FaucetConfig;
+
+    #[test]
+    fn default_faucet_config() {
+        let _config = FaucetConfig::default();
     }
 }
--- a/bin/faucet/src/config.rs
+++ b/bin/faucet/src/config.rs
@@ -3,9 +3,7 @@ use std::{
     path::PathBuf,
 };
 
-use miden_node_utils::config::{
-    Endpoint, Protocol, DEFAULT_FAUCET_SERVER_PORT, DEFAULT_NODE_RPC_PORT,
-};
+use miden_node_utils::config::{DEFAULT_FAUCET_SERVER_PORT, DEFAULT_NODE_RPC_PORT};
 use serde::{Deserialize, Serialize};
 
 // Faucet config
@@ -20,8 +18,8 @@ pub const DEFAULT_RPC_TIMEOUT_MS: u64 = 10000;
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct FaucetConfig {
-    /// Endpoint of the faucet
-    pub endpoint: Endpoint,
+    /// Endpoint of the faucet in the format `<ip>:<port>`
+    pub endpoint: String,
     /// Node RPC gRPC endpoint in the format `http://<host>[:<port>]`
     pub node_url: String,
     /// Timeout for RPC requests in milliseconds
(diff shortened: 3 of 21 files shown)


crates/store/src/config.rs
```diff
@@ -41,7 +35,7 @@ impl Default for StoreConfig {
     fn default() -> Self {
         const NODE_STORE_DIR: &str = "./";
         Self {
-            endpoint: Endpoint::localhost(DEFAULT_STORE_PORT),
+            endpoint: format!("127.0.0.1:{DEFAULT_STORE_PORT}"),
```
Mirko-von-Leipzig, line 38:
```suggestion
            endpoint: format!("http://127.0.0.1:{DEFAULT_STORE_PORT}"),
```

crates/rpc/src/config.rs
```diff
@@ -36,13 +36,9 @@ impl Display for RpcConfig {
...
     fn default() -> Self {
         Self {
-            endpoint: Endpoint {
-                host: "0.0.0.0".to_string(),
-                port: DEFAULT_NODE_RPC_PORT,
-                protocol: Protocol::default(),
-            },
-            store_url: Endpoint::localhost(DEFAULT_STORE_PORT).to_string(),
-            block_producer_url: Endpoint::localhost(DEFAULT_BLOCK_PRODUCER_PORT).to_string(),
+            endpoint: format!("0.0.0.0:{DEFAULT_NODE_RPC_PORT}"),
```
Mirko-von-Leipzig, line 39:
```suggestion
            endpoint: format!("http://0.0.0.0:{DEFAULT_NODE_RPC_PORT}"),
```

crates/block-producer/src/config.rs
```diff
@@ -42,8 +42,8 @@ impl Display for BlockProducerConfig {
 impl Default for BlockProducerConfig {
     fn default() -> Self {
         Self {
-            endpoint: Endpoint::localhost(DEFAULT_BLOCK_PRODUCER_PORT),
-            store_url: Endpoint::localhost(DEFAULT_STORE_PORT).to_string(),
+            endpoint: format!("127.0.0.1:{DEFAULT_BLOCK_PRODUCER_PORT}"),
```
Mirko-von-Leipzig, line 45:
```suggestion
            endpoint: format!("http://127.0.0.1:{DEFAULT_BLOCK_PRODUCER_PORT}"),
```

bin/faucet/src/config.rs
```diff
@@ -44,12 +42,8 @@ impl Display for FaucetConfig {
 impl Default for FaucetConfig {
     fn default() -> Self {
         Self {
-            endpoint: Endpoint {
-                host: "0.0.0.0".to_string(),
-                port: DEFAULT_FAUCET_SERVER_PORT,
-                protocol: Protocol::Http,
-            },
-            node_url: Endpoint::localhost(DEFAULT_NODE_RPC_PORT).to_string(),
+            endpoint: format!("0.0.0.0:{DEFAULT_FAUCET_SERVER_PORT}"),
```
Mirko-von-Leipzig, line 45:
```suggestion
            endpoint: format!("http://0.0.0.0:{DEFAULT_FAUCET_SERVER_PORT}"),
```

config/miden-node.toml
```diff
@@ -2,18 +2,18 @@
 
 [block_producer]
 # port defined as: sum(ord(c)**p for (p, c) in enumerate('miden-block-producer', 1)) % 2**16
-endpoint = { host = "localhost", port = 48046 }
+endpoint = "127.0.0.1:48046"
```
Mirko-von-Leipzig, line 5:
```suggestion
endpoint = "http://127.0.0.1:48046"
```

config/miden-node.toml
```diff
@@ -2,18 +2,18 @@
...
 # enables or disables the verification of transaction proofs before they are accepted into the
 # transaction queue.
 verify_tx_proofs = true
 
 [rpc]
 # port defined as: sum(ord(c)**p for (p, c) in enumerate('miden-rpc', 1)) % 2**16
-endpoint = { host = "0.0.0.0", port = 57291 }
+endpoint = "0.0.0.0:57291"
```
Mirko-von-Leipzig, line 12:
```suggestion
endpoint = "http://0.0.0.0:57291"
```

config/miden-node.toml
```diff
@@ -2,18 +2,18 @@
...
 
 [store]
 # port defined as: sum(ord(c)**p for (p, c) in enumerate('miden-store', 1)) % 2**16
 blockstore_dir    = "/opt/miden/blocks"
 database_filepath = "/opt/miden/miden-store.sqlite3"
-endpoint          = { host = "localhost", port = 28943 }
+endpoint          = "127.0.0.1:28943"
```
Mirko-von-Leipzig, line 18:
```suggestion
endpoint          = "http://127.0.0.1:28943"
```

Mirko-von-Leipzig (commented):
I'm unsure when/whether we need the protocol `http` prefix or not?

Mirko-von-Leipzig:
You'll also need to add a `[BREAKING]` changelog entry for this because it impacts configuration

sergerad:
> I'm unsure when/whether we need the protocol `http` prefix or not?

Yea the faucet config might be the only one that should omit it:
```
pub struct FaucetConfig {
    pub endpoint: String,
...
            let listener = TcpListener::bind(&config.endpoint)
...
```

I think the rest should have the protocol specified (as per your suggested fixes). I was hoping some tests would prove me right / wrong but I haven't checked yet.

bin/faucet/src/config.rs
sergerad: If I'm not mistaken then the http:// prefix would cause failure here
```
let listener = TcpListener::bind(&config.endpoint)
```

Mirko-von-Leipzig:
> I think the rest should have the protocol specified (as per your suggested fixes). I was hoping some tests would prove me right / wrong but I haven't checked yet.

I think there are no tests that actually validate the urls. We should at minimum add tests for the default settings.

This does make me wonder - maybe instead of `String` we should make them [Url](https://docs.rs/url/latest/url/struct.Url.html)?

Then our tests for the config files and defauls would pass/fail deserialization.

sergerad:
> This does make me wonder - maybe instead of String we should make them [Url](https://docs.rs/url/latest/url/struct.Url.html)?

This would be my goto approach. I can do that in the next commit.

Alternative to that (I prefer the Url type) is to have the protocol prefix in every instance of the String and do this in that socket addr flow:
```rust
let endpoint = config.endpoint.split_once("://").unwrap_or(("", &config.endpoint)).1;
```

bin/faucet/src/client.rs
```diff
@@ -200,7 +200,7 @@ impl FaucetClient {
 pub async fn initialize_faucet_client(
     config: &FaucetConfig,
 ) -> Result<(ApiClient<Channel>, BlockHeader, ChainMmr), ClientError> {
-    let endpoint = tonic::transport::Endpoint::try_from(config.node_url.clone())
+    let endpoint = tonic::transport::Endpoint::try_from(config.node_url.to_string())
```
sergerad, line 203: If we use `http::uri::Uri` instead of `url::Url`, we could rely on `From<Uri> for Endpoint` here

https://docs.rs/http/latest/http/uri/index.html

bin/faucet/src/main.rs
```diff
@@ -123,9 +123,11 @@ async fn main() -> anyhow::Result<()> {
                 )
                 .with_state(faucet_state);
 
-            let listener = TcpListener::bind((config.endpoint.host.as_str(), config.endpoint.port))
-                .await
-                .context("Failed to bind TCP listener")?;
+            let socket_addr = config.endpoint.socket_addrs(|| None)?.into_iter().next().ok_or(
+                anyhow::anyhow!("Couldn't get any socket addrs for endpoint: {}", config.endpoint),
+            )?;
```
sergerad, line 128: Would note this as a risk to fail if we don't e2e test this flow in CI before deploy

crates/block-producer/src/server.rs
```diff
@@ -66,8 +66,9 @@ impl BlockProducer {
 
         let rpc_listener = config
             .endpoint
-            .to_socket_addrs()
+            .socket_addrs(|| None)
```
sergerad, line 69: Same comment as the faucet main.rs one - this should work but not certain at this stage

bin/faucet/src/client.rs
Mirko-von-Leipzig: Ah. And `tonic` actually [re-exports](https://docs.rs/tonic/latest/tonic/transport/struct.Uri.html) this I think?

Okay so probably we should go with that. Feels a bit weird because URIs allow for more. But I guess if the user wants to type in wrong things that's on him :)

Mirko-von-Leipzig: Would we use `Uri` for all endpoints then? This is only really useful for the client side of things though. The server side still wants to `TcpListener::bind` which for `Uri` implies casting it back to `str` afaiu 🤔

bin/faucet/src/main.rs
Mirko-von-Leipzig: Would it be enough to test that the default endpoints can be bound? This could be added to each of the config tests that you've created.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changes
 
 - [BREAKING] Updated minimum Rust version to 1.84.
+- [BREAKING] Replace `Endpoint` and `Protocol` types with `url::Url` and update node TOML configuration to treat endpoints as a single string (#654).
```
Mirko-von-Leipzig, line 8: Nit: changelog should be from the user/operator perspective (i.e. they don't know `url::Url` etc).

```suggestion
- [BREAKING] `Endpoint` configuration simplified to a single string (#654).
```

crates/block-producer/src/config.rs
```diff
@@ -42,9 +37,30 @@ impl Display for BlockProducerConfig {
...
     fn default() -> Self {
         Self {
-            endpoint: Endpoint::localhost(DEFAULT_BLOCK_PRODUCER_PORT),
-            store_url: Endpoint::localhost(DEFAULT_STORE_PORT).to_string(),
+            endpoint: Url::parse(
+                format!("http://127.0.0.1:{DEFAULT_BLOCK_PRODUCER_PORT}").as_str(),
+            )
+            .unwrap(),
+            store_url: Url::parse(format!("http://127.0.0.1:{DEFAULT_STORE_PORT}").as_str())
+                .unwrap(),
```
Mirko-von-Leipzig, line 45: I thought it was weird that `Url` doesn't have a const constructor: https://github.com/servo/rust-url/issues/652. We could consider using the `url-macro` crate towards the bottom of the thread, or we could add it to our own set of macros since its rather trivial.

Though in the end I guess we get back to the issue that not all `Url` are bindable so we end up requiring the tests in any case.

PR #655, opened 2025-01-29
fix: faucet webpage rejects valid account IDs

Mirko-von-Leipzig:
We forgot to update the faucet's script file when upgrading account ID from 1 to 2 felts.

I've also added in a changelog entry for the previous version's bugfix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## v0.7.2 (2025-01-29)
+
 ### Fixes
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,8 +2,16 @@
 
 ## Unreleased
```
bobbinth, line 3: Could you change this to `## v0.7.2 (2025-01-29)` - so that we don't have to update in another PR?

Also, we could update crate versions to v0.7.2 in this PR as well.

bobbinth (approved):

PR #657, opened 2025-01-29
feat: store seed command in stress test CLI

SantiagoPittella:
@bobbinth i'm applying your suggestion of a new structure (your [comment here](https://github.com/0xPolygonMiden/miden-node/pull/621#pullrequestreview-2574388219)). The only thing I changed is that I create all the accounts before hand.

I added some queries to check the size of each table, and the avg size of each entry in every table. This is the current output of the binary:

For example, this is the current output for 100k accounts:

```
Creating new faucet account...
Generated 100000 accounts in 19.377 seconds
Creating notes...
Created notes: inserted 393 blocks with avg insertion time 641 ms
Store file size every 50 blocks:
Block 0: 4096 bytes
Block 50: 2645048 bytes
Block 100: 72965512 bytes
Block 150: 137780152 bytes
Block 200: 201362360 bytes
Block 250: 264911800 bytes
Block 300: 328420280 bytes
Block 350: 392027064 bytes
Block 400: 455551928 bytes
Average growth rate: 1159154.7888040713 bytes per blocks
Total time: 274.344 seconds
DB Stats for account_deltas: 4096 bytes, 4096.0 bytes/entry
DB Stats for block_headers: 122880 bytes, 311.088607594937 bytes/entry
DB Stats for account_fungible_asset_deltas: 4096 bytes,  bytes/entry
DB Stats for notes: 460816384 bytes, 4608.16384 bytes/entry
DB Stats for account_non_fungible_asset_updates: 4096 bytes,  bytes/entry
DB Stats for nullifiers: 4370432 bytes, 43.7218087234894 bytes/entry
DB Stats for account_storage_map_updates: 4096 bytes,  bytes/entry
DB Stats for settings: 4096 bytes, 2048.0 bytes/entry
DB Stats for account_storage_slot_updates: 4096 bytes, 1365.33333333333 bytes/entry
DB Stats for transactions: 5644288 bytes, 56.244337488665 bytes/entry
DB Stats for accounts: 6062080 bytes, 60.6438446609712 bytes/entry
```

This PR was built on top of https://github.com/0xPolygonMiden/miden-node/pull/621 .


bobbinth:
> I'm getting like 1.56 blocks/second. I'm now researching what can be the bottleneck.

Do you know where the bottlenecks are? Taking a brief look at the code it seems like we are instantiating a lot of random number generators - these could be quite expensive (especially the RPO ones). So, I'd switch to lighter versions and also try to use the same instance as much as possible.

> ```
> Store file size every 1k batches:
> 0: 4096 bytes
> 1000: 4096 bytes
> 2000: 78749696 bytes
> 3000: 160059392 bytes
> 4000: 238931968 bytes
> 5000: 318918656 bytes
> 6000: 397672448 bytes
> 7000: 477810688 bytes
> Average growth rate: 1215792.854961832 bytes per batch
> ```

Something doesn't seem right here:
- Why does the first batch of 1000 blocks does not affect storage size? I guess after that we get consistent growth of about 75MB per 1000 blocks.
- 7000 blocks with 256 accounts created per block should result in 1.8M accounts. Where does the 100K accounts number come from?
- Related to the above, if there are only 100K accounts, database size of almost 500MB doesn't really make a lot of sense (this would imply almost for 5MB per account). If it is more like 1.8M then it seems a bit low (about 270 bytes per account) - though, maybe it is possible.

SantiagoPittella:
> Why does the first batch of 1000 blocks does not affect storage size? I guess after that we get consistent growth of about 75MB per 1000 blocks.

I was checking only the size of `miden-store.sqlite3`; I'm now running it again for 1M accounts using the size of both files (as Mirko mentioned) combined as the total size.

Though I want to clarify that it is each 1000 batches, each block is 16 batches in the implementation so we keep track of the increase of size every 62.5 blocks.

> 7000 blocks with 256 accounts created per block should result in 1.8M accounts. Where does the 100K accounts number come from?

The 7k are batches, so it is like ~440 blocks. We are using 255 accounts per block + 1 tx to mint assets to each one of the accounts.

> Related to the above, if there are only 100K accounts, database size of almost 500MB doesn't really make a lot of sense (this would imply almost for 5MB per account). If it is more like 1.8M then it seems a bit low (about 270 bytes per account) - though, maybe it is possible.

I will come back with the results of this new run with 1M accounts and with the store size fixed. Currently it is taking like ~1 hour to run for 1M accounts (for the total process).

SantiagoPittella:
This is the flamegraph that I got this morning for 1M accounts:

flamegraph_1m

SantiagoPittella:
flamegraph_1m

SantiagoPittella:
here you can visualize a flamegraph for 1M accounts (removed the preview because it was too big):
https://github.com/user-attachments/assets/f7639a8c-63cc-4384-99cb-fedc50e8cc58

SantiagoPittella:
I added a couple more of metrics, and re run a couple times with different number of accounts.

I'm consistently getting 4550~ish bytes/account

SantiagoPittella:
>I'm consistently getting 4550~ish bytes/account

It is worth mention that this number is the result of doing `total_db_size / total_number_account` so it includes blocks and any other stuff that we store in the DB.

I ran the following query in the DB and got this results:
```sql
SELECT
     name,
     SUM(pgsize) AS size_bytes,
     (SUM(pgsize) * 1.0) / (SELECT COUNT(*) FROM accounts) AS bytes_per_row
 FROM dbstat
 WHERE name = 'accounts';
 ```
 
 And got this results:
`accounts| 6025216 bytes | 60.2750645245193 bytes per account`

**Comment by bobbinth** [comment]
> I added a couple more of metrics, and re run a couple times with different number of accounts.
> 
> I'm consistently getting 4550~ish bytes/account

4.5KB is better than 5MB - but still looks pretty high. Part of this is nullifiers and notes - but I don't see how these contribute more than 1KB (about 40 bytes for nullifiers + 80 bytes for notes + 500 bytes for note authentication paths). But it is also possible I'm missing something.

Another possibility is that SQLite doesn't do compaction, and maybe there is a lot of "slack" in the file.

**Comment by igamigo** [comment]
There's also the overhead of block headers and indices as well, but those are also probably too small to consider.
One test that you can do is run a [VACUUM command](https://www.sqlite.org/lang_vacuum.html) to see if the file sizes change considerably after the initial seeding finishes.

**Comment by SantiagoPittella** [comment]
I added some queries to check the size of each table, and the avg size of each entry in every table. This is the current output of the binary:
```
Creating new faucet account...
Generated 100000 accounts in 19.377 seconds
Creating notes...
Created notes: inserted 393 blocks with avg insertion time 641 ms
Store file size every 50 blocks:
Block 0: 4096 bytes
Block 50: 2645048 bytes
Block 100: 72965512 bytes
Block 150: 137780152 bytes
Block 200: 201362360 bytes
Block 250: 264911800 bytes
Block 300: 328420280 bytes
Block 350: 392027064 bytes
Block 400: 455551928 bytes
Average growth rate: 1159154.7888040713 bytes per blocks
Average space used per account: 4545.705054133613 bytes
Total time: 274.344 seconds
DB Stats for account_deltas: 4096 bytes, 4096.0 bytes/entry
DB Stats for block_headers: 122880 bytes, 311.088607594937 bytes/entry
DB Stats for account_fungible_asset_deltas: 4096 bytes,  bytes/entry
DB Stats for notes: 460816384 bytes, 4608.16384 bytes/entry
DB Stats for account_non_fungible_asset_updates: 4096 bytes,  bytes/entry
DB Stats for nullifiers: 4370432 bytes, 43.7218087234894 bytes/entry
DB Stats for account_storage_map_updates: 4096 bytes,  bytes/entry
DB Stats for settings: 4096 bytes, 2048.0 bytes/entry
DB Stats for account_storage_slot_updates: 4096 bytes, 1365.33333333333 bytes/entry
DB Stats for transactions: 5644288 bytes, 56.244337488665 bytes/entry
DB Stats for accounts: 6062080 bytes, 60.6438446609712 bytes/entry
```

**Comment by SantiagoPittella** [comment]
I could not reduce the time that takes for each block. It is currently at 1.6 blocks/sec

**Comment by bobbinth** [comment]
> I could not reduce the time that takes for each block. It is currently at 1.6 blocks/sec

Do you have the breakdown of different components of this? Specifically, how much, out of 1.6 seconds is taken up by block construction vs. block insertion?

**Comment by bobbinth** [comment]
> ```
> DB Stats for account_deltas: 4096 bytes, 4096.0 bytes/entry
> DB Stats for block_headers: 122880 bytes, 311.088607594937 bytes/entry
> DB Stats for account_fungible_asset_deltas: 4096 bytes,  bytes/entry
> DB Stats for notes: 460816384 bytes, 4608.16384 bytes/entry
> DB Stats for account_non_fungible_asset_updates: 4096 bytes,  bytes/entry
> DB Stats for nullifiers: 4370432 bytes, 43.7218087234894 bytes/entry
> DB Stats for account_storage_map_updates: 4096 bytes,  bytes/entry
> DB Stats for settings: 4096 bytes, 2048.0 bytes/entry
> DB Stats for account_storage_slot_updates: 4096 bytes, 1365.33333333333 bytes/entry
> DB Stats for transactions: 5644288 bytes, 56.244337488665 bytes/entry
> DB Stats for accounts: 6062080 bytes, 60.6438446609712 bytes/entry
> ```

Interesting! It seems like 97% of the database is in the `notes` table (we are adding 4.5KB of data per note). Let's create an issue to optimize note storage. Let's create an issue to optimize note storage. I think we should be able to reduce the size by about 10x.

**Comment by SantiagoPittella** [comment]
@bobbinth I created https://github.com/0xPolygonMiden/miden-node/issues/662 to address the notes issue.

> Do you have the breakdown of different components of this? Specifically, how much, out of 1.6 seconds is taken up by block construction vs. block insertion?

According to the flamegraph, a big part of this process is in `miden_processor::Process::execute_mast_node`. Though I'm making some changes to get more insights of this.

**Comment by bobbinth** [comment]
> > Do you have the breakdown of different components of this? Specifically, how much, out of 1.6 seconds is taken up by block construction vs. block insertion?
> 
> According to the flamegraph, a big part of this process is in `miden_processor::Process::execute_mast_node`. Though I'm making some changes to get more insights of this.

Looking at the code, I think what we are measuring as insertion time is actually `BlockBuilder::build_block()` time. This will actually covers much more than just inserting block into the store (e.g., it also executes the block kernel to build the block). So, this will be much longer than what we want. What I really mean by insertion time is the time it takes the store to process `apply_block()` request. My hope is that this is less than 100ms - but we'll need to confirm.

What may make sense to do is build blocks manually (and probably the same for batches as well), without using block and batch builders. This is somewhat difficult now, but should become much easier after https://github.com/0xPolygonMiden/miden-node/pull/659 and subsequent work is done.

So, I guess we have 3 paths forward:

1. Review & merge this work roughly as is and then re-factor it once manual block/batch building becomes easier.
2. Try to implement manual block/batch building now, and then update it with the new structure later.
3. Put this on hold and then re-work once manual block/batch building becomes easier.

Assuming this is pretty close to being done, path 1 probably makes the most sense - but let me know what you think.

File: crates/block-producer/Cargo.toml [hunk 1]
```diff
@@ -15,35 +15,37 @@ version.workspace      = true
 workspace = true
 
 [features]
+testing        = ["dep:miden-air", "dep:miden-node-test-macro", "dep:rand_chacha", "dep:winterfell"]
```
bobbinth, line 18: With a more manual approach to batch/block building, we probably wouldn't need to introduce the `testing` feature. So, maybe that's an argument for putting the work on hold for now.

SantiagoPittella:
@bobbinth I think that 1) and 3) are the best options.

In a way I think that having this command anyways, building blocks like it is currently doing, is fine for now. We are not testing block insertion time but we accomplish the goal of seeding the DB to further benchmark the endpoints.

I agree with about 3) though, I don't really like re-introducing the `testing` feature, and if we can choose to replace it in the near future, I would go for that.

bobbinth:
> I agree with about 3) though, I don't really like re-introducing the `testing` feature, and if we can choose to replace it in the near future, I would go for that.

Let's go with option (3) for now so that we don't have to introduce and then again remove the `testing` feature. Batch building should be simplified even today, and block building may be in the same state by the end of the week.

SantiagoPittella:
Ok, sounds good! I will put this and the StateSync benchmark one as drafts then.

TomasArrachea:
Updated this PR to use the new `ProvenBatch`, pausing this until the new block building is implemented.

crates/block-producer/Cargo.toml
TomasArrachea: Actually, we are still using the `MockProvenTxBuilder` to build the transactions and the `TransactionBatchConstructor` to build the proven batches. Perhaps we could change it to build the complete structs, but I'm not sure how straightforward that would be

TomasArrachea:
Just updated the code to build the blocks separately and measure insertions times only on the `apply_block()` operation. This is the current output:
```
Creating new faucet account...
Generated 100000 accounts in 17.850 seconds
Creating notes...
Created notes: inserted 393 blocks with avg insertion time 123 ms
Store file size every 50 blocks:
Block 0: 4096 bytes
Block 50: 1330768 bytes
Block 100: 38184960 bytes
Block 150: 70172344 bytes
Block 200: 100568760 bytes
Block 250: 131014328 bytes
Block 300: 161394360 bytes
Block 350: 191835832 bytes
Block 400: 222232248 bytes
Average growth rate: 565466.0356234097 bytes per blocks
Total time: 148.443 seconds
DB Stats for account_deltas: 4096 bytes, 4096.0 bytes/entry
DB Stats for block_headers: 122880 bytes, 311.088607594937 bytes/entry
DB Stats for account_fungible_asset_deltas: 4096 bytes,  bytes/entry
DB Stats for notes: 205312000 bytes, 2053.12 bytes/entry
DB Stats for account_non_fungible_asset_updates: 4096 bytes,  bytes/entry
DB Stats for nullifiers: 4370432 bytes, 43.7218087234894 bytes/entry
DB Stats for account_storage_map_updates: 4096 bytes,  bytes/entry
DB Stats for settings: 4096 bytes, 2048.0 bytes/entry
DB Stats for account_storage_slot_updates: 4096 bytes, 1365.33333333333 bytes/entry
DB Stats for transactions: 5644288 bytes, 56.2437770293162 bytes/entry
DB Stats for accounts: 6062080 bytes, 60.6438446609712 bytes/entry
```

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,471 @@
...
+        miden_objects::note::NoteType::Public,
+        Felt::default(),
+        rng,
+    )
+    .expect("note creation failed")
+}
+
+/// Create a new account with a given public key and anchor block. Generates the seed from the given
+/// index.
+fn create_account(anchor_block: &BlockHeader, public_key: PublicKey, index: u64) -> AccountId {
```
bobbinth, line 329: Why are we returning only account ID from here? Do we not need the actual account?

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,471 @@
...
+    // Create first block with the faucet
+    let txs =
+        vec![
+            MockProvenTxBuilder::with_account(faucet_id, Digest::default(), Digest::default())
+                .output_notes(vec![])
+                .build(),
+        ];
+    let batch = ProvenBatch::mocked_from_transactions(txs.iter());
+
+    apply_block(vec![batch], &store_client).await;
```
bobbinth, line 133: Does the "first block" mean the genesis block here? If so, isn't the genesis block created on startup (i.e., we don't need to insert it into the store explicitly)?

If this is not a genesis block, then why do we need to insert the faucet into it (the faucet should be present in the genesis block)?

bobbinth (commented):
Thank you! Looks good. Not a full review yet, but I left comments inline - some of which may result in significant refactoring.

Also, a couple of comments regarding the output format:
- I would probably print out table sizes in KB rather than bytes and have at most 1 decimal point.
- It would be interesting to see how block insertion time changes with the number of blocks. So, the printed out blocks (50, 100, 150 etc.), I would print out:
  - Block insertion time (in ms).
  - Block size (in bytes).
  - Database size (in KB or maybe even MB).

TomasArrachea:
Refactored the output, so now it's looking like this:
```
Generated 100000 accounts in 16.962 seconds
Creating notes...
Total time: 142.380 seconds
Inserted 393 blocks with avg insertion time 124 ms
Initial DB size: 4.0 KB
Average DB growth rate: 548.9 KB per block

Block metrics:
Block #    Insert Time (ms)     Block Size (B)       DB Size (KB)
---------------------------------------------------------------------------
0          99                   690155               1267.4
50         102                  360558               37241.7
100        106                  360558               68455.5
150        117                  360558               98163.5
200        127                  360558               127939.5
250        131                  360558               157559.5
300        140                  360558               187319.5
350        144                  360558               216987.5

Database stats:
Table                               Size (KB)       KB/Entry
----------------------------------------------------------------------
account_deltas                      4.0             -
block_headers                       120.0           0.3
account_fungible_asset_deltas       4.0             -
notes                               200500.0        2.0
account_non_fungible_asset_updates  4.0             -
nullifiers                          4268.0          0.0
account_storage_map_updates         4.0             -
settings                            4.0             2.0
account_storage_slot_updates        4.0             -
transactions                        5512.0          0.1
accounts                            5920.0          0.1
```

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,471 @@
...
+        let path = MerklePath::new(vec![]);
+        let inclusion_proof = NoteInclusionProof::new(0.into(), 0, path).unwrap();
+
+        let consume_tx =
+            MockProvenTxBuilder::with_account(account, Digest::default(), Digest::default())
+                .authenticated_notes(vec![(note.clone(), inclusion_proof)])
+                .build();
+
+        accounts_and_notes.lock().unwrap().push((account, note, consume_tx));
+    });
```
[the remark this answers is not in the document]
TomasArrachea: Ok! Implemented this and the performance was not affected, so I agree it's better.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,471 @@
...
+    task::spawn_blocking(move || {
+        generate_batches(
+            num_accounts,
+            faucet_id,
+            accounts_and_notes.lock().unwrap().as_slice(),
+            &batch_sender,
+        );
+    })
+    .await
+    .unwrap();
```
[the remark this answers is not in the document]
TomasArrachea: This channel approach was taken to improve the test performance. Using the channel allows for generating the accounts, notes and batches, while concurrently consuming those batches to create and store the blocks. But since the block building was refactored, the situation now is a bit different.

I just refactored the test to see how this changes would affect the performance, and I got (for 100k accounts) 317 seconds, versus the 142 seconds I got with the previous channels implementation. Could be an acceptable delay to pay for the simpler solution.

bin/stress-test/src/main.rs
TomasArrachea: No, only accounts ids are needed to create the transactions

bobbinth: hmmm - I may need to look at it a bit more, but how do we create a transaction which consumes a note by an account without knowing the account state?

TomasArrachea: I think it's probably because we are mocking the transactions and using dummy hashes for the initial and final states.

crates/block-producer/Cargo.toml
TomasArrachea: Did this in [049f82d](https://github.com/0xPolygonMiden/miden-node/pull/657/commits/049f82d96bd9da1419c9bd7911289ed75ea18e21)

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,375 @@
...
+        let consume_tx = create_tx(
+            account,
+            Digest::default(),
+            Digest::default(),
+            vec![],
+            vec![(note.clone(), inclusion_proof)],
+        );
+        accounts_and_notes.push((account, note, consume_tx));
+    }
+    accounts_and_notes
```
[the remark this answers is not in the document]
TomasArrachea: Yes! Doing this reduced execution time greatly. Now execution took about the same time as before, 140 seconds. 🚀

bin/stress-test/src/main.rs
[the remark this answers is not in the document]
TomasArrachea: Update: parallelizing account creation took back the execution time to about 140 secs!

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,377 @@
...
+    final_account_hash: Digest,
+    output_notes: Vec<Note>,
+    input_notes: Vec<(Note, NoteInclusionProof)>,
+) -> ProvenTransaction {
+    ProvenTransactionBuilder::new(
+        account_id,
+        initial_account_hash,
+        final_account_hash,
+        BlockNumber::from(0),
+        Digest::default(),
```
bobbinth, line 364: I would try to use the actual block number and block hash. It shouldn't be too much extra work, but would create a more consistent database which should be better for subsequent tests.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,377 @@
...
+                create_note(faucet_id, account, &mut rng)
+            };
+
+            let path = MerklePath::new(vec![]);
+            let inclusion_proof = NoteInclusionProof::new(0.into(), 0, path).unwrap();
+
+            let consume_tx = create_tx(
+                account,
+                Digest::default(),
+                Digest::default(),
```
bobbinth, line 266: It may be good to use actual initial/final hashes here - this way, the database would be more correct. To do this, we would need to return the actual account from `create_account()`. Also, it may make sense to move the account update logic into `create_tx()` function. It could look something like this:
```rust
fn create_consume_note_tx(
    account: Account,
    input_note: InputNote,
) -> ProvenTransaction
```
This would also mean that we'd need a different function for creating the transaction which emits nots from the faucet. It could look something like:
```rust
fn create_emit_notes_tx(
    faucet: Account,
    output_notes: Vec<Note>,
) -> ProvenTransaction
```

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,377 @@
...
+        .map(|account_index| {
+            let account =
+                create_account(genesis_header, key_pair.public_key(), account_index as u64);
+            let note = {
+                let mut rng = rng.lock().unwrap();
+                create_note(faucet_id, account, &mut rng)
+            };
+
+            let path = MerklePath::new(vec![]);
+            let inclusion_proof = NoteInclusionProof::new(0.into(), 0, path).unwrap();
```
bobbinth, line 261: It would be good to figure use the actual Merkle paths - this would come in handy later on when we try to run other tests against the generated database.

bobbinth (commented):
Looks good! Thank you. Still not a full review, but I left some more comments inline. The main ones are about trying to make the inserted data to be as real as possible. We may not be able to get 100% there, but the more real the data is, the better tests against this data could be run later on.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,377 @@
...
+    path::{Path, PathBuf},
+    sync::{Arc, Mutex},
+    time::{Duration, Instant},
+};
+
+mod metrics;
+use anyhow::Context;
+use clap::{Parser, Subcommand};
+use metrics::Metrics;
+use miden_air::HashFunction;
```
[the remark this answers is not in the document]
TomasArrachea: I don't think so, I see only ExecutionProof is exposed on [miden-objects vm](https://github.com/0xPolygonMiden/miden-base/blob/228f8325de0de931f3738d1118665de597b3c5c4/crates/miden-objects/src/lib.rs#L190-L194)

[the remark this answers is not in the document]
TomasArrachea: It's used as a parameter in the ExecutionProof constructor. One way to remove it could be using `Default::default()` instead.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+}
+
+/// Seed the store with a given number of accounts.
+async fn seed_store(dump_file: &Path, num_accounts: usize) {
+    let start = Instant::now();
+
+    // Generate the faucet account and the genesis state
+    let faucet = create_faucet();
+    let genesis_state = GenesisState::new(vec![faucet.clone()], 1, 1_672_531_200);
```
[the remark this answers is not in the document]
TomasArrachea: Actually I took this timestamp from the `config/genesis.toml`, I'm not sure if its important for this, for now changed it to 1.

bin/stress-test/src/metrics.rs
```diff
@@ -0,0 +1,176 @@
...
+        self.get_batch_inputs_time_per_block.push(query_time);
+    }
+
+    /// Get the size of the store file and its WAL file.
+    fn get_store_size(dump_file: &Path) -> u64 {
+        let store_file_size = std::fs::metadata(dump_file).unwrap().len();
+        let wal_file_size =
+            std::fs::metadata(format!("{}-wal", dump_file.to_str().unwrap())).unwrap().len();
+        store_file_size + wal_file_size
+    }
```
bobbinth, line 78: nit: since this does not use `self` - it could probably be just a "pure" helper function.

bin/stress-test/src/metrics.rs
```diff
@@ -0,0 +1,176 @@
...
+    bytes_per_block: Vec<usize>,
+    num_insertions: u32,
+    store_file_sizes: Vec<u64>,
+    initial_store_size: u64,
+    store_file: PathBuf,
+}
+
+impl Metrics {
+    /// Create a new Metrics instance.
+    pub fn new(store_file: PathBuf) -> Self {
```
bobbinth, line 36: nit: we usually use 3-rd person verbs in doc comments - e.g., `Create` -> `Creates`, `Track` - > `Tracks` etc.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,444 @@
...
+use clap::{Parser, Subcommand};
+use metrics::Metrics;
+use miden_air::{FieldElement, HashFunction};
+use miden_block_prover::LocalBlockProver;
+use miden_lib::{
+    account::{auth::RpoFalcon512, faucets::BasicFungibleFaucet, wallets::BasicWallet},
+    note::create_p2id_note,
+    utils::Serializable,
+};
+use miden_node_block_producer::store::StoreClient;
```
bobbinth, line 20: Is there a way to get the store client w/o going through the block producer? I wonder if #723 changes this. cc @Mirko-von-Leipzig.

crates/block-producer/src/lib.rs
```diff
@@ -8,7 +8,7 @@ mod block_builder;
 mod domain;
 mod errors;
 mod mempool;
-mod store;
+pub mod store;
```
bobbinth, line 11: Related to the previous comment: we need this only to get the `StoreClient`, right? If so, there may be a better way of getting it.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,444 @@
...
+
+    metrics.add_insertion(start.elapsed(), block_size);
+
+    proven_block.header()
+}
+
+// HELPERS
+// ================================================================================================
+
+/// Create `num_accounts` accounts, and for each one create a note that mint assets.
```
bobbinth, line 222: nit: `Create` -> `Creates`. Also relevant for most functions in this module.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,444 @@
...
+    store_client: &StoreClient,
+    block_ref: &BlockHeader,
+    notes: &[Note],
+    metrics: &mut Metrics,
+) -> BatchInputs {
+    let start = Instant::now();
+    let batch_inputs = store_client
+        .get_batch_inputs(
+            vec![(block_ref.block_num(), block_ref.hash())].into_iter(),
+            notes.iter().map(Note::id),
```
bobbinth, line 412: This basically marks every not as unauthenticated, right? And I think we do this so that we can get Merkle paths for every note - correct? If so, I would add some comments here explaining this.

bobbinth (commented):
Looks good! Thank you! I left some more comments inline - but they are pretty minor.

Also, would be great to see how the latest output looks like.

bin/stress-test/src/main.rs
Mirko-von-Leipzig: The block producer's `StoreClient` is a somewhat thin wrapper around the generated store client from `miden-node-proto`. Notably it does some additional domain type translations.

At present #723 doesn't change this - its primarily just about changing how we distribute the raw proto files at the moment.

In the short term exposing this `StoreClient` is the simplest route unfortunately, or alternatively just using the generated store client - which would essentially duplicate the code I imagine.

Perhaps in the medium term we should focus on moving all client/store wrappers into the crate with the domain types. So one crate for accessing the raw proto definitions, and another for internal use with nicer domain types. 

The issue with the latter is how do we allow for extensions, e.g. enabling open-telemetry, TLS etc. Its possible I'm sure, but I find the `tonic` docs and naming to be confusing so its not immediately obvious to me how this works.

TomasArrachea:
This is how the latest output looks:
```
Total time: 224.065 seconds
Inserted 393 blocks with avg insertion time 186 ms
Initial DB size: 100.6 KB
Average DB growth rate: 550.4 KB per block

Block metrics:
Block #    Insert Time (ms)     Get Block Inputs Time (ms)     Get Batch Inputs Time (ms)     Block Size (B)       DB Size (KB)
---------------------------------------------------------------------------------------------------------------------------------------
0          23                   1                              1                              331726               655.8
50         161                  15                             6                              360558               36633.7
100        169                  19                             11                             360558               67867.1
150        177                  24                             16                             360558               97527.1
200        190                  29                             22                             360558               127275.1
250        194                  38                             28                             360558               156923.1
300        203                  37                             33                             360558               186671.1
350        211                  41                             38                             360558               216427.1

Database stats:
Table                               Size (KB)       KB/Entry
----------------------------------------------------------------------
account_deltas                      4.0             -
block_headers                       120.0           0.3
account_fungible_asset_deltas       4.0             -
notes                               200932.0        2.0
account_non_fungible_asset_updates  4.0             -
nullifiers                          4268.0          0.0
account_storage_map_updates         4.0             -
settings                            4.0             2.0
account_storage_slot_updates        4.0             -
transactions                        5512.0          0.1
accounts                            5920.0          0.1
```

bin/stress-test/src/main.rs
TomasArrachea: Yes, as for now this is the simplest way to interact with the Store. I agree it would make sense to, at some point, have the `StoreClient` declared as a domain type, but it's probably better to do that it in a separate PR.

TomasArrachea: > Another issue with moving the wrappers is that all domain types would need to live in this proto crate i.e. we cannot use types from miden-node-store, miden-node-block-producer etc, so those all have to move across. Which may or may not be fine. A lot of them are already there so perhaps its not such an issue.

At least for the store, this would require moving the `StoreErrors`. Besides that, I think most of these types are shared from miden-objects and it wouldn't be a problem.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,456 @@
...
+        /// Number of accounts to create.
+        #[arg(short, long, value_name = "NUM_ACCOUNTS")]
+        num_accounts: usize,
+    },
+}
+
+/// Creates and stores blocks into the store. Creates a given number of accounts, where each account
+/// consumes a note created from a faucet. The cli accepts the following parameters:
+/// - `data_directory`: Directory in which to store the database and raw block data.
+/// - `num_accounts`: Number of accounts to create.
```
SantiagoPittella, line 73: This doc looks unnecessary, it is from the command defined above.

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,456 @@
...
+        },
+    }
+}
+
+/// Seeds the store with a given number of accounts.
+async fn seed_store(data_directory: PathBuf, num_accounts: usize) {
+    let start = Instant::now();
+
+    let genesis_filepath = data_directory.join(GENESIS_STATE_FILENAME);
+    let database_filepath = data_directory.join("miden-store.sqlite3");
```
SantiagoPittella, line 90: This might be extracted into a constant.

SantiagoPittella (commented):
Found some `.unwrap`s throughout the CLI, those might be replaced with `.expect`.

Can not give an approve because I opened up this PR, but in general looks good 👌🏼 .

bin/stress-test/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+
+        // collect all the txs
+        block_txs.push(emit_note_tx);
+        block_txs.extend(consume_notes_txs);
+
+        // create the batches with [TRANSACTIONS_PER_BATCH] txs each
+        let batches: Vec<ProvenBatch> = block_txs
+            .chunks(TRANSACTIONS_PER_BATCH)
+            .map(|txs| create_batch(txs, &prev_header))
+            .collect();
```
bobbinth, line 208: Should we do this using a parallel iterator?

crates/block-producer/README.md
```diff
@@ -49,5 +49,10 @@ can know for certain it was evicted.
 
 ---
 
+## Crate Features
+
+- `tracing-forest`: sets logging using tracing-forest. Disabled by default.
+- `testing`: includes testing util functions to mock block-producer behaviour, meant to be used during development and not in production. Disabled by default.
```
bobbinth, line 55: Is this still correct? I don't think there is a `testing` feature currently for the block producer.

bin/stress-test/src/main.rs
bobbinth: I'd love to make to so that `stress-test` does not depend on the block producer, but we can do this in the future. Let's create an issue for this.

bobbinth (approved):

Before merging, we should rebase it from the latest `next` and make sure things still work. Also, could we run account generation for 1M accounts? Should take a bit less then an hour, right?

igamigo:
I tested this while I did a bunch of things (so probably not very representative), and I got ~2.5 hours:
<details>
<summary><b>cargo run --release -- seed-store --data-directory ./data --num-accounts 1000000</b></summary>

```

Total time: 9388.023 seconds
Inserted 3922 blocks with avg insertion time 1400 ms
Initial DB size: 136.8 KB
Average DB growth rate: 648.8 KB per block

Block metrics:
Block #    Insert Time (ms)     Get Block Inputs Time (ms)     Get Batch Inputs Time (ms)     Block Size (B)       DB Size (KB)
---------------------------------------------------------------------------------------------------------------------------------------
0          26                   1                              1                              331726               736.3
50         182                  12                             1                              360558               40818.9
100        196                  13                             1                              360558               72946.9
150        212                  19                             1                              360558               105565.3
200        210                  14                             1                              360558               138601.4
250        226                  18                             1                              360558               171476.8
300        241                  16                             1                              360558               204150.5
350        232                  25                             1                              360558               236788.5
400        247                  19                             1                              360558               269112.9
450        260                  25                             1                              360558               301694.4
500        247                  20                             1                              360558               333963.0
550        282                  61                             1                              360558               366640.2
600        327                  26                             3                              360558               398888.4
650        390                  69                             4                              360558               431577.9
700        367                  58                             2                              360558               463869.9
750        321                  83                             2                              360558               495969.9
800        358                  148                            2                              360558               528587.3
850        408                  68                             5                              360558               561151.3
900        306                  76                             2                              360558               593479.3
950        389                  77                             3                              360558               625611.7
1000       391                  104                            2                              360558               658067.7
1050       537                  246                            4                              360558               690191.8
1100       862                  87                             13                             360558               722636.1
1150       399                  142                            2                              360558               754956.4
1200       3572                 610                            26                             360558               787328.6
1250       425                  64                             2                              360558               819772.6
1300       552                  37                             5                              360558               852409.0
1350       704                  160                            17                             360558               884429.0
1400       693                  180                            3                              360558               916637.1
1450       1195                 141                            3                              360558               949362.4
1500       713                  357                            3                              360558               981558.4
1550       1054                 262                            3                              360558               1014234.9
1600       1001                 265                            7                              360558               1046402.9
1650       779                  136                            3                              360558               1078842.9
1700       1621                 303                            4                              360558               1111078.9
1750       635                  185                            3                              360558               1143454.9
1800       1032                 311                            2                              360558               1186669.9
1850       1850                 234                            7                              360558               1218749.9
1900       751                  553                            3                              360558               1251109.9
1950       1031                 225                            4                              360558               1284057.8
2000       4169                 571                            96                             360558               1316557.8
2050       2421                 7993                           7                              360558               1348785.8
2100       1138                 474                            4                              360558               1381137.8
2150       2897                 2288                           26                             360558               1413249.8
2200       1568                 315                            5                              360558               1445657.8
2250       3536                 1854                           26                             360558               1477837.8
2300       2877                 661                            14                             360558               1510201.8
2350       1772                 292                            3                              360558               1542533.8
2400       2573                 1282                           3                              360558               1574725.8
2450       791                  423                            5                              360558               1606921.8
2500       1464                 519                            3                              360558               1639453.8
2550       1793                 413                            9                              360558               1671921.8
2600       1140                 240                            11                             360558               1704269.8
2650       1242                 1703                           4                              360558               1736517.8
2700       961                  267                            6                              360558               1769033.8
2750       700                  278                            4                              360558               1801333.8
2800       1421                 416                            4                              360558               1833517.8
2850       1500                 332                            4                              360558               1865749.8
2900       2161                 633                            5                              360558               1898270.9
2950       2352                 896                            31                             360558               1930650.9
3000       1159                 886                            11                             360558               1963162.9
3050       1374                 775                            3                              360558               1995362.9
3100       1245                 571                            2                              360558               2027706.9
3150       4165                 1145                           5                              360558               2060118.9
3200       1314                 584                            4                              360558               2092322.9
3250       1059                 338                            4                              360558               2124538.9
3300       1065                 436                            3                              360558               2156938.9
3350       1022                 269                            3                              360558               2189070.9
3400       1073                 302                            3                              360558               2221126.9
3450       3353                 1024                           4                              360558               2253486.9
3500       1084                 292                            2                              360558               2286010.9
3550       924                  551                            5                              360558               2318558.9
3600       1240                 1196                           5                              360558               2350758.9
3650       2413                 713                            3                              360558               2382966.9
3700       2367                 377                            12                             360558               2415382.9
3750       1066                 283                            5                              360558               2447758.9
3800       1161                 849                            3                              360558               2480006.9
3850       2081                 710                            7                              360558               2512418.9
3900       2258                 1234                           5                              360558               2544758.9



Database stats:
Table                               Size (KB)       KB/Entry
----------------------------------------------------------------------
account_deltas                      4.0             -
block_headers                       1128.0          0.3
account_fungible_asset_deltas       4.0             -
notes                               2005264.0       2.0
account_non_fungible_asset_updates  4.0             -
nullifiers                          43040.0         0.0
account_storage_map_updates         4.0             -
settings                            4.0             2.0
account_storage_slot_updates        4.0             -
transactions                        55832.0         0.1
accounts                            56400.0         0.1
```

</details>

bobbinth (approved):
All looks good! Thank you! I'm going to merge as is, but let's do the following thing:

- Create an issue for investigating how to make the stress test not depend on `block-producer` for the store API.
- Run the account generation again first for 100K accounts (to compare with prior results) and then for 1M accounts to see if the results @igamigo got were representative.

Regarding the results, if these were in fact representative, we should create an issue to investigate what exactly is driving the performance. The database is not that huge (i.e., 2.5GB) and so, it shouldn't be taking ~800ms to retrieve block inputs or 1 - 2 seconds to insert a block.

Mirko-von-Leipzig:
I've created #763 and #762.

PR #660, opened 2025-01-31
feat: open-telemetry exporter

Mirko-von-Leipzig:
This PR adds a `--open-telemetry` flag to the node's `start` command. This exports traces to an open-telemetry backend.

The various settings of this exporter can be controlled by the various environment variables specified by the standard e.g.

```
export OTEL_EXPORTER_OTLP_ENDPOINT="https://api.honeycomb.io"
```
sets the target endpoint.

This is just initial support for this; our actual traces are still _very_ non-conforming to the standard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Enhancements
+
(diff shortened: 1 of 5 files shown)


bobbinth (approved):

Issue #664, opened 2025-02-03
[Task]: Consider taking iterator in `select_block_headers`


PhilippGackstatter:
### Node's components affected by this task

- [ ] RPC
- [ ] Block producer
- [x] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

Consider using an iterator in `miden_node_store::db::Db::select_block_headers` to avoid the need for conversions to `Vec` before calling this function.

### How should it be done?

This could perhaps take an `impl Iterator<Item = BlockNumber>` since the inner `miden_node_store::db::sql::select_block_headers` function needs to convert to `Vec<Value>` anyway.

### When is this task done?

If the conversion is no longer necessary or we decided that it's not useful to do this.

varun-doshi:
I'd like to take this

Mirko-von-Leipzig:
Completed in #667

PR #667, opened 2025-02-04
feat: taking iterator in select_block_headers

varun-doshi:
Ref #664 

Replaces taking a `Vec` in `select_block_headers` with taking an `impl Iterator`

### DIFF
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -250,11 +250,12 @@ impl Db {
 
     /// Loads multiple block headers from the DB.
     #[instrument(target = COMPONENT, skip_all, ret(level = "debug"), err)]
-    pub async fn select_block_headers(&self, blocks: Vec<BlockNumber>) -> Result<Vec<BlockHeader>> {
+    pub async fn select_block_headers(&self, blocks: impl Iterator<Item = BlockNumber>) -> Result<Vec<BlockHeader>> {
+        let block_vec: Vec<BlockNumber> = blocks.collect();
         self.pool
             .get()
             .await?
-            .interact(move |conn| sql::select_block_headers(conn, &blocks))
+            .interact(move |conn| sql::select_block_headers(conn, &block_vec))
             .await
             .map_err(|err| {
                 DatabaseError::InteractError(format!(
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -250,12 +250,14 @@ impl Db {
 
     /// Loads multiple block headers from the DB.
     #[instrument(target = COMPONENT, skip_all, ret(level = "debug"), err)]
-    pub async fn select_block_headers(&self, blocks: impl Iterator<Item = BlockNumber>) -> Result<Vec<BlockHeader>> {
-        let block_vec: Vec<BlockNumber> = blocks.collect();
+    pub async fn select_block_headers(
+        &self,
+        blocks: impl Iterator<Item = BlockNumber> + Send + 'static,
+    ) -> Result<Vec<BlockHeader>> {
         self.pool
             .get()
             .await?
-            .interact(move |conn| sql::select_block_headers(conn, &block_vec))
+            .interact(move |conn| sql::select_block_headers(conn, blocks))
             .await
             .map_err(|err| {
                 DatabaseError::InteractError(format!(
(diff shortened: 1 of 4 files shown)


crates/store/src/db/mod.rs
```diff
@@ -250,11 +250,12 @@ impl Db {
...
     /// Loads multiple block headers from the DB.
     #[instrument(target = COMPONENT, skip_all, ret(level = "debug"), err)]
-    pub async fn select_block_headers(&self, blocks: Vec<BlockNumber>) -> Result<Vec<BlockHeader>> {
+    pub async fn select_block_headers(&self, blocks: impl Iterator<Item = BlockNumber>) -> Result<Vec<BlockHeader>> {
+        let block_vec: Vec<BlockNumber> = blocks.collect();
         self.pool
             .get()
             .await?
-            .interact(move |conn| sql::select_block_headers(conn, &blocks))
+            .interact(move |conn| sql::select_block_headers(conn, &block_vec))
```
PhilippGackstatter, line 258: This `sql::select_block_headers` should also be able to take an `impl Iterator` right? In that function we collect into a `Vec<Value>` anyway, so we don't have to collect here.

varun-doshi: The error that pops up here is:

```
impl Iterator<Item = BlockNumber>` cannot be sent between threads safelyrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B0%5D?0#file%3A%2F%2F%2FUsers%2Fvarundoshi%2FDocuments%2Fvscode-github%2Fmiden-node%2Fcrates%2Fstore%2Fsrc%2Fdb%2Fmod.rs)
mod.rs(258, 23): within this `{closure@crates/store/src/db/mod.rs:258:23: 258:34}`
mod.rs(258, 14): required by a bound introduced by this call
mod.rs(258, 23): required because it's used within this closure
lib.rs(115, 34): required by a bound in `deadpool_sync::SyncWrapper::<T>::interact`
mod.rs(253, 87): consider further restricting this bound: ` + std::marker::Send`
```

varun-doshi: one solution is to pass the blocks variable in the `state.rs` in the `db.select_block_headers`  as 
```
let headers = self.db.select_block_headers(<BTreeSet<BlockNumber> as Clone>::clone(&blocks.clone()).into_iter()).await?;
```

and then the `select_block_headers ` is:
```
pub async fn select_block_headers(&self, blocks: impl Iterator<Item = BlockNumber>+Send+'static) -> Result<Vec<BlockHeader>>
```

Mirko-von-Leipzig: The issue is that the data needs to be sent across async bounds. Its one of the downsides of trying to squash async into a sync database.

There might be some way using something like `Arc<[T]>` but that likely requires further changes elsewhere.

varun-doshi: @Mirko-von-Leipzig Can you check the last commit on this PR? Does that work?

Mirko-von-Leipzig: It does, though it still requires an extra clone in the one case which I think was the hope to avoid.

Overall this is still an improvement though, so I'm happy with this. Will leave it to @PhilippGackstatter to decide though :D

crates/store/src/state.rs
```diff
@@ -479,7 +477,13 @@ impl State {
             (chain_length.into(), paths)
         };
 
-        let headers = self.db.select_block_headers(blocks).await?;
+        let headers = self
+            .db
+            .select_block_headers(
+                blocks.clone().into_iter(),
+            )
```
bobbinth, line 484: Question: why do we need `.clone()` here? It doesn't seem like `blocks` is used after this line - so, shouldn't we be able to consume it here? (i.e., does just `blocks.into_iter()` not work for some reason?)

varun-doshi: Oversight on my end
Should be fixed now

PhilippGackstatter (approved):
Looks good to me! Thanks for figuring out the details on the trait bounds.

varun-doshi:
Added Changelog and corrected the formatting

PhilippGackstatter:
@varun-doshi Can you fix the merge conflict with the changelog? Thanks!

@Mirko-von-Leipzig Should we merge this afterwards?

Issue #663, opened 2025-02-03
[Task]: Wrap `Mmr` into a `Blockchain` struct


PhilippGackstatter:
### Node's components affected by this task

- [ ] RPC
- [ ] Block producer
- [x] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

Wrap the `Mmr` in the `InnerState` into a `Blockchain` struct.

### How should it be done?

This `Blockchain` can provide methods for:
- getting the latest block number (what `InnerState::latest_block_num` does now),
- getting the current chain length (i.e. latest block number + 1), which should probably be a `usize` as it can be used as the `forest` parameter for various MMR operations,
- converting to a `ChainMmr` which is a partial view of a certain number of block headers and their proofs of inclusion into the chain.

### When is this task done?

When the `Blockchain` struct was added and the existing construction of `ChainMmr` in `State::get_batch_inputs` is using this method instead (https://github.com/0xPolygonMiden/miden-node/pull/659/files#diff-339991026f9c6728dbde8578e8d896f0c44e4234a63e25113190fd1701d343b5R551-R597).

### Additional context

See also: https://github.com/0xPolygonMiden/miden-node/pull/659#discussion_r1938318921

varun-doshi:
I can take this

PhilippGackstatter:
Closed by #668.

PR #668, opened 2025-02-05
feat: added wrapper for Mmr

varun-doshi:
Ref #663 

Ideally should be merged after #667 is merged so that we can directly send in the `Iterator` instead of collecting into a `Vec`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,15 @@
 
 ## Unreleased
 
+### Fixes
+
+- Faucet webpage is missing `background.png` and `favicon.ico` (#672).
+- Add wrapper for `Mmr` (#668).
+
 ### Enhancements
 
 - Add an optional open-telemetry trace exporter (#659).
+- Support tracing across gRPC boundaries using remote tracing context (#669).
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@
 ### Fixes
 
 - Faucet webpage is missing `background.png` and `favicon.ico` (#672).
-- Add wrapper for `Mmr` (#668).
 
 ### Enhancements
 
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -96,12 +96,17 @@ pub struct TransactionInputs {
     pub found_unauthenticated_notes: BTreeSet<NoteId>,
 }
 
-/// Blockchain wrapper for `Mmr`
+/// Blockchain wrapper for `Mmr` .c
 #[derive(Debug, Clone)]
 pub struct Blockchain(Mmr);
 
 impl Blockchain {
-    /// Returns the latest block number
+    /// Returns a new Blockchain.
+    pub fn new(chain_mmr: Mmr) -> Self {
+        Self(chain_mmr)
+    }
+
+    /// Returns the latest block number.
     pub fn chain_tip(&self) -> BlockNumber {
         let block_number: u32 = (self.0.forest() - 1)
             .try_into()
@@ -110,7 +115,7 @@ impl Blockchain {
         block_number.into()
     }
 
-    /// Returns the chain length
+    /// Returns the chain length.
     pub fn chain_length(&self) -> BlockNumber {
         let latest = self.chain_tip();
         latest + 1
@@ -129,28 +134,24 @@ impl Blockchain {
         self.0.peaks_at(forest)
     }
 
-    /// Adds a new element to the MMR.
-    pub fn add(&mut self, el: RpoDigest) {
-        self.0.add(el);
+    /// Adds a block commitment to the MMR. The caller must ensure that this commitent is the one
+    /// for the next block in the chain.
+    pub fn add(&mut self, block_commitment: RpoDigest) {
+        self.0.add(block_commitment);
     }
 
-    /// Returns an [`MmrProof`] for the leaf at the specified position
+    /// Returns an [`MmrProof`] for the leaf at the specified position.
     pub fn open(&self, pos: usize) -> Result<MmrProof, MmrError> {
         self.0.open_at(pos, self.0.forest())
     }
 
-    /// Returns the latest block number and partial mmr
-    pub fn chain_mmr(
+    /// Returns the latest block number and partial mmr.
+    pub fn latest_block_number(
         &self,
         blocks: &BTreeSet<BlockNumber>,
         highest_block_number: &BlockNumber,
     ) -> Result<(BlockNumber, PartialMmr), GetBatchInputsError> {
         // Grab the block merkle paths from the inner state.
-        //
-        // NOTE: Scoped block to automatically drop the mutex guard asap.
-        //
-        // We also avoid accessing the db in the block as this would delay
-        // dropping the guard.
         let (batch_reference_block, partial_mmr) = {
             let latest_block_num = highest_block_number;
 
@@ -196,7 +197,7 @@ struct InnerState {
 impl InnerState {
     /// Returns the latest block number.
     fn latest_block_num(&self) -> BlockNumber {
-        let block_number: u32 = (self.blockchain.0.forest() - 1)
+        let block_number: u32 = (self.blockchain.chain_tip().as_usize() - 1)
             .try_into()
             .expect("chain_mmr always has, at least, the genesis block");
 
@@ -236,7 +237,7 @@ impl State {
 
         let inner = RwLock::new(InnerState {
             nullifier_tree,
-            blockchain: Blockchain(chain_mmr),
+            blockchain: Blockchain::new(chain_mmr),
             account_tree,
         });
 
@@ -339,7 +340,7 @@ impl State {
             // compute updates for the in-memory data structures
 
             // new_block.chain_root must be equal to the chain MMR root prior to the update
-            let peaks = inner.blockchain.0.peaks();
+            let peaks = inner.blockchain.peaks();
             if peaks.hash_peaks() != header.chain_root() {
                 return Err(InvalidBlockError::NewBlockInvalidChainRoot.into());
             }
@@ -468,7 +469,7 @@ impl State {
                 .account_tree
                 .apply_mutations(account_tree_update)
                 .expect("Unreachable: old account tree root must be checked before this step");
-            inner.blockchain.0.add(block_hash);
+            inner.blockchain.add(block_hash);
         }
 
         info!(%block_hash, block_num = block_num.as_u32(), COMPONENT, "apply_block successful");
@@ -490,7 +491,7 @@ impl State {
         if let Some(header) = block_header {
             let mmr_proof = if include_mmr_proof {
                 let inner = self.inner.read().await;
-                let mmr_proof = inner.blockchain.0.open(header.block_num().as_usize())?;
+                let mmr_proof = inner.blockchain.open(header.block_num().as_usize())?;
                 Some(mmr_proof)
             } else {
                 None
@@ -554,12 +555,12 @@ impl State {
         // dropping the guard.
         let (chain_length, merkle_paths) = {
             let state = self.inner.read().await;
-            let chain_length = state.blockchain.0.forest();
+            let chain_length = state.blockchain.chain_tip().as_usize();
 
             let paths = blocks
                 .iter()
                 .map(|&block_num| {
-                    let proof = state.blockchain.0.open(block_num.as_usize())?.merkle_path;
+                    let proof = state.blockchain.open(block_num.as_usize())?.merkle_path;
 
                     Ok::<_, MmrError>((block_num, proof))
                 })
@@ -650,8 +651,29 @@ impl State {
         // there is no need to prove its inclusion.
         blocks.remove(&latest_block_num);
 
-        let (batch_reference_block, partial_mmr) =
-            self.inner.blocking_read().blockchain.chain_mmr(&blocks, &highest_block_num)?;
+        // Scoped block to automatically drop the read lock guard as soon as we're done.
+        // We also avoid accessing the db in the block as this would delay dropping the guard.
+        let (batch_reference_block, partial_mmr) = {
+            let inner_state = self.inner.blocking_read();
+
+            let latest_block_num = inner_state.blockchain.chain_tip();
+
+            let highest_block_num =
+                *blocks.last().expect("we should have checked for empty block references");
+            if highest_block_num > latest_block_num {
+                return Err(GetBatchInputsError::TransactionBlockReferenceNewerThanLatestBlock {
+                    highest_block_num,
+                    latest_block_num,
+                });
+            }
+
+            // Remove the latest block from the to-be-tracked blocks as it will be the reference
+            // block for the batch itself and thus added to the MMR within the batch kernel, so
+            // there is no need to prove its inclusion.
+            blocks.remove(&latest_block_num);
+
+            inner_state.blockchain.latest_block_number(&blocks, &latest_block_num)?
+        };
 
         // Fetch the reference block of the batch as part of this query, so we can avoid looking it
         // up in a separate DB access.
@@ -668,7 +690,7 @@ impl State {
             .find_map(|(index, header)| {
                 (header.block_num() == batch_reference_block).then_some(index)
             })
-            .expect("DB should have returned the header of the batch reference block");
+            .ok_or(GetBatchInputsError::BatchRefernceBlockError)?;
 
         // The order doesn't matter for ChainMmr::new, so swap remove is fine.
         let batch_reference_block_header = headers.swap_remove(header_index);
@@ -767,7 +789,7 @@ impl State {
 
         let note_sync = self.db.get_note_sync(block_num, note_tags).await?;
 
-        let mmr_proof = inner.blockchain.0.open(note_sync.block_header.block_num().as_usize())?;
+        let mmr_proof = inner.blockchain.open(note_sync.block_header.block_num().as_usize())?;
 
         Ok((note_sync, mmr_proof))
     }
@@ -788,9 +810,9 @@ impl State {
             .ok_or(GetBlockInputsError::DbBlockHeaderEmpty)?;
 
         // sanity check
-        if inner.blockchain.0.forest() != latest.block_num().as_usize() + 1 {
+        if inner.blockchain.chain_tip().as_usize() != latest.block_num().as_usize() + 1 {
             return Err(GetBlockInputsError::IncorrectChainMmrForestNumber {
-                forest: inner.blockchain.0.forest(),
+                forest: inner.blockchain.chain_tip().as_usize(),
                 block_num: latest.block_num(),
             });
         }
@@ -798,7 +820,7 @@ impl State {
         // using current block number gets us the peaks of the chain MMR as of one block ago;
         // this is done so that latest.chain_root matches the returned peaks
         let chain_peaks =
-            inner.blockchain.0.peaks_at(latest.block_num().as_usize()).map_err(|error| {
+            inner.blockchain.peaks_at(latest.block_num().as_usize()).map_err(|error| {
                 GetBlockInputsError::FailedToGetMmrPeaksForForest {
                     forest: latest.block_num().as_usize(),
                     error,
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -96,17 +96,114 @@ pub struct TransactionInputs {
     pub found_unauthenticated_notes: BTreeSet<NoteId>,
 }
 
+#[derive(Debug)]
+pub struct Blockchain(Mmr);
+
+impl Blockchain {
+    pub fn get(&self) -> &Mmr {
+        &self.0
+    }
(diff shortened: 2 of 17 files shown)


crates/store/src/state.rs
```diff
@@ -96,17 +96,114 @@ pub struct TransactionInputs {
...
+#[derive(Debug)]
+pub struct Blockchain(Mmr);
+
+impl Blockchain {
+    pub fn get(&self) -> &Mmr {
+        &self.0
+    }
+    pub fn get_mut(&mut self) -> &mut Mmr {
+        &mut self.0
+    }
```
PhilippGackstatter, line 108: We shouldn't need these `get` or `get_mut` methods. We should expose anything that is needed from the `Blockchain` as dedicated methods and with the chain-specific types, e.g.:
- `Blockchain::peaks(&self) -> ...`
- `Blockchain::peaks_at(&self, block_num: BlockNumber) -> ...`
- `Blockchain::add(&mut self, block_hash: Digest) -> ...`
- `Blockchain::open(&self, block_num: BlockNumber) -> ...`

crates/store/src/state.rs
```diff
@@ -96,17 +96,114 @@ pub struct TransactionInputs {
...
+            .try_into()
+            .expect("chain_mmr always has, at least, the genesis block");
+
+        block_number.into()
+    }
+
+    pub fn get_chain_length(&self) -> usize {
+        let latest = self.get_latest_block_number();
+        latest.as_usize() + 1
+    }
```
PhilippGackstatter, line 121: Nit: In Rust the `get_` prefix is usually omittted, so `latest_block_number` and `chain_length`.

crates/store/src/state.rs
```diff
@@ -96,17 +96,114 @@ pub struct TransactionInputs {
...
+
+    pub fn get_latest_block_number(&self) -> BlockNumber {
+        let block_number: u32 = (self.0.forest() - 1)
+            .try_into()
+            .expect("chain_mmr always has, at least, the genesis block");
+
+        block_number.into()
+    }
+
+    pub fn get_chain_length(&self) -> usize {
```
PhilippGackstatter, line 118: After looking at how this would be used, I think returning a `BlockNumber` is better after all, even though I find it is a bit odd to represent the _length_ of the chain as such.
But this would allow doing something like `block_chain.peaks_at(block_chain.chain_length())` which is better than requiring conversion to `usize` or some other integer type.

crates/store/src/state.rs
```diff
@@ -96,17 +96,114 @@ pub struct TransactionInputs {
...
+        let chain_mmr =
+            ChainMmr::new(partial_mmr, headers.clone()).expect("Unable to build ChainMmr");
+        Ok((chain_mmr, headers, batch_reference_block))
+    }
+}
 /// Container for state that needs to be updated atomically.
 struct InnerState {
     nullifier_tree: NullifierTree,
-    chain_mmr: Mmr,
+    chain_mmr: Blockchain,
```
PhilippGackstatter, line 199: Can we rename this field to `blockchain` instead? `chain_mmr` suggests it is a `ChainMmr` but it's not.

crates/store/src/state.rs
```diff
@@ -96,17 +96,114 @@ pub struct TransactionInputs {
...
+
+            (latest_block_num, partial_mmr)
+        };
+
+        // TODO: Unnecessary conversion. We should change the select_block_headers function to take
+        // an impl Iterator instead to avoid this allocation.
+        let mut blocks: Vec<_> = blocks.clone().into_iter().collect();
+        // Fetch the reference block of the batch as part of this query, so we can avoid looking it
+        // up in a separate DB access.
+        blocks.push(batch_reference_block);
```
PhilippGackstatter, line 185: Can you merge in `next` and update this function with the changes from next?

crates/store/src/state.rs
```diff
@@ -96,17 +96,114 @@ pub struct TransactionInputs {
...
+        let headers = db
+            .select_block_headers(blocks)
+            .await
+            .map_err(GetBatchInputsError::SelectBlockHeaderError)?;
+
+        let chain_mmr =
+            ChainMmr::new(partial_mmr, headers.clone()).expect("Unable to build ChainMmr");
```
PhilippGackstatter, line 192: When you move this back to `get_batch_inputs`, please also re-add the `SAFETY` comment and revert to the previous `expect` message.

PhilippGackstatter (commented):
Thanks for working on this. I left some comments inline.

crates/store/src/state.rs
```diff
@@ -96,17 +96,114 @@ pub struct TransactionInputs {
...
+
+impl Blockchain {
+    pub fn get(&self) -> &Mmr {
+        &self.0
+    }
+    pub fn get_mut(&mut self) -> &mut Mmr {
+        &mut self.0
+    }
+
+    pub fn get_latest_block_number(&self) -> BlockNumber {
```
igamigo, line 110: nit: I think `chain_tip()` be a more concrete name here (aligns more with other structures AFAICT)

crates/store/src/state.rs
```diff
@@ -96,17 +96,114 @@ pub struct TransactionInputs {
...
+    pub fn get_chain_length(&self) -> usize {
+        let latest = self.get_latest_block_number();
+        latest.as_usize() + 1
+    }
+
+    pub async fn get_chain_mmr(
+        &self,
+        blocks: &mut BTreeSet<BlockNumber>,
+        db: &Arc<Db>,
+    ) -> Result<(ChainMmr, Vec<miden_objects::block::BlockHeader>, BlockNumber), GetBatchInputsError>
```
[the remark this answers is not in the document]
varun-doshi: Please correct me if im wrong but 
`blocks.remove(&latest_block_num);`

this requires a mutable reference no?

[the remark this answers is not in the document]
varun-doshi: Small doubt; you mean to call `get_batch_inputs` from `chain_mmr` right? But to do that i need access to a `State` from inside this function right?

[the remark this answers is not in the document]
PhilippGackstatter: It does if we want to do the logic as it is now, but I think we can compute the `highest_block_num` in `get_batch_inputs`, remove it from `blocks` and then pass the blocks and the highest block num, i.e.: `get_chain_mmr(blocks: &BTreeSet<BlockNumber>, highest_block_num: BlockNumber)`. Does that make sense?

crates/store/src/state.rs
```diff
@@ -96,17 +96,107 @@ pub struct TransactionInputs {
     pub found_unauthenticated_notes: BTreeSet<NoteId>,
 }
 
+/// Blockchain wrapper for `Mmr`
+#[derive(Debug, Clone)]
+pub struct Blockchain(Mmr);
+
+impl Blockchain {
+    /// Returns the latest block number
```
PhilippGackstatter, line 104: Nit: Can you add `.` at the end of all doc comments?

crates/store/src/state.rs
```diff
@@ -96,17 +96,107 @@ pub struct TransactionInputs {
...
+    /// # Errors
+    /// Returns an error if the specified `forest` value is not valid for this MMR.
+    pub fn peaks_at(&self, forest: usize) -> Result<MmrPeaks, MmrError> {
+        self.0.peaks_at(forest)
+    }
+
+    /// Adds a new element to the MMR.
+    pub fn add(&mut self, el: RpoDigest) {
+        self.0.add(el);
+    }
```
PhilippGackstatter, line 135:
```suggestion
    /// Adds a block commitment to the MMR. The caller must ensure that this commitent is the one for the next block in the chain.
    pub fn add(&mut self, block_commitment: RpoDigest) {
        self.0.add(block_commitment);
    }
```

crates/store/src/state.rs
```diff
@@ -96,17 +96,107 @@ pub struct TransactionInputs {
...
-    chain_mmr: Mmr,
+    blockchain: Blockchain,
     account_tree: SimpleSmt<ACCOUNT_TREE_DEPTH>,
 }
 
 impl InnerState {
     /// Returns the latest block number.
     fn latest_block_num(&self) -> BlockNumber {
-        let block_number: u32 = (self.chain_mmr.forest() - 1)
+        let block_number: u32 = (self.blockchain.0.forest() - 1)
```
PhilippGackstatter, line 199: The idea of adding methods like `Blockchain::chain_tip` is that we can call them here, so we can avoid accessing the inner `Mmr` directly. So all `blockchain.0.xyz()` methods should be replaced by calling methods directly, e.g. `blockchain.xyz()` (or with blockchain-specific names, so `blockchain.chain_tip()` instead of `blockchain.0.forest()`.

Here we should be able to call `chain_tip`, right?

crates/store/src/state.rs
```diff
@@ -144,7 +234,11 @@ impl State {
         let chain_mmr = load_mmr(&mut db).await?;
         let account_tree = load_accounts(&mut db).await?;
 
-        let inner = RwLock::new(InnerState { nullifier_tree, chain_mmr, account_tree });
+        let inner = RwLock::new(InnerState {
+            nullifier_tree,
+            blockchain: Blockchain(chain_mmr),
```
PhilippGackstatter, line 239: Can we create a proper constructor for this, so `Blockchain::new(chain_mmr)`?

crates/store/src/state.rs
```diff
@@ -540,68 +634,32 @@ impl State {
...
+            self.inner.blocking_read().blockchain.chain_mmr(&blocks, &highest_block_num)?;
 
         // Fetch the reference block of the batch as part of this query, so we can avoid looking it
         // up in a separate DB access.
         let mut headers = self
             .db
             .select_block_headers(blocks.into_iter().chain(std::iter::once(batch_reference_block)))
             .await
-            .map_err(GetBatchInputsError::SelectBlockHeaderError)?;
+            .expect("partial mmr and block headers should be consistent");
```
PhilippGackstatter, line 662: I think we still want to error here instead of panic.

crates/store/src/state.rs
```diff
@@ -96,17 +96,107 @@ pub struct TransactionInputs {
...
+        &self,
+        blocks: &BTreeSet<BlockNumber>,
+        highest_block_number: &BlockNumber,
+    ) -> Result<(BlockNumber, PartialMmr), GetBatchInputsError> {
+        // Grab the block merkle paths from the inner state.
+        //
+        // NOTE: Scoped block to automatically drop the mutex guard asap.
+        //
+        // We also avoid accessing the db in the block as this would delay
+        // dropping the guard.
```
PhilippGackstatter, line 153: This part of the comment is no longer relevant, so we can remove it.

crates/store/src/state.rs
```diff
@@ -96,17 +96,107 @@ pub struct TransactionInputs {
...
+        self.0.add(el);
+    }
+
+    /// Returns an [`MmrProof`] for the leaf at the specified position
+    pub fn open(&self, pos: usize) -> Result<MmrProof, MmrError> {
+        self.0.open_at(pos, self.0.forest())
+    }
+
+    /// Returns the latest block number and partial mmr
+    pub fn chain_mmr(
```
PhilippGackstatter, line 143: This actually only constructs a partial MMR now, so we should name it accordingly, e.g. `partial_mmr_from_blocks` or something.

crates/store/src/state.rs
```diff
@@ -96,17 +96,107 @@ pub struct TransactionInputs {
...
+        &self,
+        blocks: &BTreeSet<BlockNumber>,
+        highest_block_number: &BlockNumber,
```
PhilippGackstatter, line 146: My bad, this should actually be called `latest_block_number`.

crates/store/src/state.rs
```diff
@@ -540,68 +634,32 @@ impl State {
...
-            }
+        // Remove the latest block from the to-be-tracked blocks as it will be the reference
+        // block for the batch itself and thus added to the MMR within the batch kernel, so
+        // there is no need to prove its inclusion.
+        blocks.remove(&latest_block_num);
 
-            (latest_block_num, partial_mmr)
-        };
+        let (batch_reference_block, partial_mmr) =
+            self.inner.blocking_read().blockchain.chain_mmr(&blocks, &highest_block_num)?;
```
PhilippGackstatter, line 654: We need to pass `latest_block_num` here instead, and we should probably acquire the `RwLock` just once and drop it as soon as we can, so everything together:

```rust
// Scoped block to automatically drop the read lock guard as soon as we're done.
// We also avoid accessing the db in the block as this would delay dropping the guard.
let (batch_reference_block, partial_mmr) = {
    let inner_state = self.inner.blocking_read();

    let latest_block_num = inner_state.blockchain.chain_tip();

    let highest_block_num =
        *blocks.last().expect("we should have checked for empty block references");
    if highest_block_num > latest_block_num {
        return Err(GetBatchInputsError::TransactionBlockReferenceNewerThanLatestBlock {
            highest_block_num,
            latest_block_num,
        });
    }

    // Remove the latest block from the to-be-tracked blocks as it will be the reference
    // block for the batch itself and thus added to the MMR within the batch kernel, so
    // there is no need to prove its inclusion.
    blocks.remove(&latest_block_num);

    inner_state.blockchain.chain_mmr(&blocks, &latest_block_num)?
};
```

PhilippGackstatter (commented):
Thanks for the updates! I left some comments.

crates/store/src/state.rs
```diff
@@ -610,7 +690,7 @@ impl State {
             .find_map(|(index, header)| {
                 (header.block_num() == batch_reference_block).then_some(index)
             })
-            .expect("DB should have returned the header of the batch reference block");
+            .ok_or(GetBatchInputsError::BatchRefernceBlockError)?;
```
PhilippGackstatter, line 693:
```suggestion
            .expect("DB should have returned the header of the batch reference block");
```

This should still be the expect and we can remove the `BatchRefernceBlockError` variant.

PhilippGackstatter (commented):
Thanks for the updates! I think we're close to finishing this. I left another round of comments.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Faucet webpage is missing `background.png` and `favicon.ico` (#672).
+- Add wrapper for `Mmr` (#668).
```
Mirko-von-Leipzig, line 8: We can leave this out of the changelog; this is essentially an internal change only.

crates/store/src/state.rs
```diff
@@ -96,17 +96,113 @@ pub struct TransactionInputs {
     pub found_unauthenticated_notes: BTreeSet<NoteId>,
 }
 
+/// Blockchain wrapper for `Mmr` .c
```
Mirko-von-Leipzig, line 99:
```suggestion
/// Blockchain wrapper for `Mmr`.
```
or maybe
```suggestion
/// A [Merkle Mountain Range](Mmr) defining a chain of blocks.
```

crates/store/src/state.rs
```diff
@@ -96,17 +96,113 @@ pub struct TransactionInputs {
...
+
+    /// Returns the current peaks of the MMR.
+    pub fn peaks(&self) -> MmrPeaks {
+        self.0.peaks()
+    }
+
+    /// Returns the peaks of the MMR at the state specified by `forest`.
+    ///
+    /// # Errors
+    /// Returns an error if the specified `forest` value is not valid for this MMR.
```
Mirko-von-Leipzig, line 132:
```suggestion
    /// # Errors
    ///
    /// Returns an error if the specified `forest` value is not valid for this MMR.
```

crates/store/src/state.rs
```diff
@@ -96,17 +96,113 @@ pub struct TransactionInputs {
...
+    pub fn peaks_at(&self, forest: usize) -> Result<MmrPeaks, MmrError> {
+        self.0.peaks_at(forest)
+    }
+
+    /// Adds a block commitment to the MMR. The caller must ensure that this commitent is the one
+    /// for the next block in the chain.
+    pub fn add(&mut self, block_commitment: RpoDigest) {
+        self.0.add(block_commitment);
```
Mirko-von-Leipzig, line 140: Side note: I'm actually surprised this isn't called `push`

crates/store/src/state.rs
```diff
@@ -96,17 +96,113 @@ pub struct TransactionInputs {
...
+                partial_mmr
+                    .track(block_num, leaf, &path)
+                    .expect("filling partial mmr with data from mmr should succeed");
+            }
+
+            partial_mmr
+        };
+        Ok(partial_mmr)
+    }
+}
```
Mirko-von-Leipzig, line 194:
```suggestion
}

```

crates/store/src/state.rs
```diff
@@ -96,17 +96,113 @@ pub struct TransactionInputs {
...
+        latest_block_number: BlockNumber,
+    ) -> Result<PartialMmr, GetBatchInputsError> {
+        // Grab the block merkle paths from the inner state.
+        let partial_mmr = {
+            let latest_block_num = latest_block_number;
+
+            // Using latest block as the target forest means we take the state of the MMR one before
+            // the latest block. This is because the latest block will be used as the reference
+            // block of the batch and will be added to the MMR by the batch kernel.
+            let target_forest = latest_block_num.as_usize();
```
Mirko-von-Leipzig, line 166:
```suggestion
            // Using latest block as the target forest means we take the state of the MMR one before
            // the latest block. This is because the latest block will be used as the reference
            // block of the batch and will be added to the MMR by the batch kernel.
            let target_forest = latest_block_number.as_usize();
```

crates/store/src/state.rs
```diff
@@ -96,17 +96,113 @@ pub struct TransactionInputs {
...
+            .try_into()
+            .expect("chain_mmr always has, at least, the genesis block");
+
+        block_number.into()
+    }
+
+    /// Returns the chain length.
+    pub fn chain_length(&self) -> BlockNumber {
+        let latest = self.chain_tip();
+        latest + 1
```
Mirko-von-Leipzig, line 121: I believe `BlockNumber::child()` exists. 

```suggestion
        self.chain_tip().child()
```

Though I'm unsure if we should have both `chain_tip` and `chain_length`? I guarantee someone's going to confuse the two. 

----------

--edit-- in fact further down we have `let chain_length = state.blockchain.chain_tip() as usize;`

crates/store/src/state.rs
```diff
@@ -96,17 +96,113 @@ pub struct TransactionInputs {
...
+    }
+
+    /// Returns the latest block number and partial mmr.
+    pub fn partial_mmr_from_blocks(
+        &self,
+        blocks: &BTreeSet<BlockNumber>,
+        latest_block_number: BlockNumber,
+    ) -> Result<PartialMmr, GetBatchInputsError> {
+        // Grab the block merkle paths from the inner state.
+        let partial_mmr = {
```
Mirko-von-Leipzig, line 160: I think we no longer need this nested block

crates/store/src/state.rs
```diff
@@ -729,17 +802,17 @@ impl State {
             .ok_or(GetBlockInputsError::DbBlockHeaderEmpty)?;
 
         // sanity check
-        if inner.chain_mmr.forest() != latest.block_num().as_usize() + 1 {
+        if inner.blockchain.chain_tip() != latest.block_num() + 1 {
```
Mirko-von-Leipzig, line 805: Is this correct? I would assume that these should be equal and not off-by-one?

varun-doshi: removing `+1`

varun-doshi:
@Mirko-von-Leipzig did not notice your commit there; apologies
I assume the only change was removing from the changelog?

Mirko-von-Leipzig:
> @Mirko-von-Leipzig did not notice your commit there; apologies I assume the only change was removing from the changelog?

That's my bad for inserting a commit without warning :) That was the only change, no harm done.

Mirko-von-Leipzig (approved):
Thanks for persevering @varun-doshi, I know its been a lengthy back-and-forth. Really appreciate it!

crates/store/src/state.rs
```diff
@@ -564,35 +656,10 @@ impl State {
...
-                partial_mmr
-                    .track(block_num, leaf, &path)
-                    .expect("filling partial mmr with data from mmr should succeed");
-            }
-
-            (latest_block_num, partial_mmr)
+            (
+                latest_block_num,
+                inner_state.blockchain.partial_mmr_from_blocks(&blocks, latest_block_num)?,
+            )
```
bobbinth, line 662: Related to the above comment, this could be:
```rust
let partial_mmr = inner_state.blockchain.partial_mmr_from_blocks(&blocks)?;
blocks.remove(&latest_block_num);
(latest_block_num, partial_mmr)
```

varun-doshi: I can open a separate PR to get this changes in, if you'd like

bobbinth: That would be awesome! Thank you!

PhilippGackstatter: I did pretty much that change in #709 since I used that function there anyway, so no need for the PR I think, @varun-doshi, but thanks!

PR #669, opened 2025-02-05
feat: remote tracing context

Mirko-von-Leipzig:
This PR enables tracing across RPC boundaries using open-telemetry.

Open-telemetry supports smuggling trace ID context across HTTP/RPC calls by embedding this context in various headers. In our case we are using gRPC metadata headers (which in  turn are just binary encoded http headers I believe). Notably, this context includes the sender's trace ID which allows us to connect the dots on the receiver end.

The general idea is that a trace should represent one cohesive item of work. Most examples are of course web-focussed and essentially boil down to a single user request being represented by one trace. In our world this becomes one trace per:

- RPC component request
- Single batch building
- Single block building
- Any other tasks we might have in the future

All of these cross our _internal_ RPC boundaries and this PR allows us to continue tracing across it.

## Implementation details

Open-telemetry tracing context is injected on the client side using `tonic`'s `Interceptor` trait. On the receiver we extract this context using `tonic`'s `trace_fn`. Both of these are implemented as middleware and are therefore RPC method agnostic.

Note that the RPC server is missing this -- the tracing is restricted to internal only APIs. We _could_ consider expanding this to include other services e.g. the faucet. We would have to authenticate these requests somehow, maybe with an authentication header or API key. Though this might be of questionable value at this stage.

## Concerns

This remote context is open-telemetry specific. I was somewhat hoping that `tracing-forest` would also magically connect the dots but apparently not :) This does sort of mean our stdout implementation will drift ito feature/quality compared to the otel exporter.

The open-telemetry specification also has a mountain of naming standards both [general](https://opentelemetry.io/docs/specs/semconv/general/) and specific to [RPC](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/) and [gRPC](https://opentelemetry.io/docs/specs/semconv/rpc/grpc/), http etc.. We will want to identify which of these are important to us. This may also mean that the current implementation can't cater to them e.g. client IP and server IP are not available to the middleware I used.

### DIFF
--- a/crates/block-producer/src/server.rs
+++ b/crates/block-producer/src/server.rs
@@ -54,7 +54,7 @@ impl BlockProducer {
         info!(target: COMPONENT, %config, "Initializing server");
 
         let channel = tonic::transport::Endpoint::try_from(config.store_url.to_string())
-            .map_err(|err| ApiError::DatabaseConnectionFailed(err.to_string()))?
+            .map_err(|err| ApiError::InvalidStoreUrl(err.to_string()))?
             .connect()
             .await
             .map_err(|err| ApiError::DatabaseConnectionFailed(err.to_string()))?;
--- a/crates/block-producer/src/server.rs
+++ b/crates/block-producer/src/server.rs
@@ -7,6 +7,7 @@ use miden_node_proto::generated::{
 use miden_node_utils::{
     errors::ApiError,
     formatting::{format_input_notes, format_output_notes},
+    tracing::grpc::OtelInterceptor,
 };
 use miden_objects::{
     block::BlockNumber, transaction::ProvenTransaction, utils::serde::Deserializable,
@@ -52,11 +53,14 @@ impl BlockProducer {
     pub async fn init(config: BlockProducerConfig) -> Result<Self, ApiError> {
         info!(target: COMPONENT, %config, "Initializing server");
 
-        let store = StoreClient::new(
-            store_client::ApiClient::connect(config.store_url.to_string())
-                .await
-                .map_err(|err| ApiError::DatabaseConnectionFailed(err.to_string()))?,
-        );
+        let channel = tonic::transport::Endpoint::try_from(config.store_url.to_string())
+            .map_err(|err| ApiError::DatabaseConnectionFailed(err.to_string()))?
+            .connect()
+            .await
+            .map_err(|err| ApiError::DatabaseConnectionFailed(err.to_string()))?;
+
+        let store = store_client::ApiClient::with_interceptor(channel, OtelInterceptor);
+        let store = StoreClient::new(store);
 
         let latest_header = store
             .latest_header()
@@ -208,6 +212,7 @@ impl BlockProducerRpcServer {
 
     async fn serve(self, listener: TcpListener) -> Result<(), tonic::transport::Error> {
         tonic::transport::Server::builder()
+            .trace_fn(miden_node_utils::tracing::grpc::block_producer_trace_fn)
             .add_service(api_server::ApiServer::new(self))
             .serve_with_incoming(TcpListenerStream::new(listener))
             .await
--- a/crates/block-producer/src/store/mod.rs
+++ b/crates/block-producer/src/store/mod.rs
@@ -19,7 +19,7 @@ use miden_node_proto::{
     },
     AccountState,
 };
-use miden_node_utils::formatting::format_opt;
+use miden_node_utils::{formatting::format_opt, tracing::grpc::OtelInterceptor};
 use miden_objects::{
     account::AccountId,
     block::{Block, BlockHeader, BlockNumber},
@@ -29,7 +29,7 @@ use miden_objects::{
     Digest,
 };
 use miden_processor::crypto::RpoDigest;
-use tonic::transport::Channel;
+use tonic::{service::interceptor::InterceptedService, transport::Channel};
 use tracing::{debug, info, instrument};
 
 use crate::{block::BlockInputs, errors::StoreError, COMPONENT};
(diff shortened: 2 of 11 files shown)


crates/utils/src/tracing/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod grpc;
```
Mirko-von-Leipzig, line 1: I've added this in anticipation of splitting the `logging.rs` into here with `otel` and `stdout` modules.

crates/utils/src/tracing/grpc.rs
```diff
@@ -0,0 +1,116 @@
...
+        Some("GetBlockByNumber") => tracing::info_span!("store.rpc/GetBlockByNumber"),
+        Some("GetBlockHeaderByNumber") => tracing::info_span!("store.rpc/GetBlockHeaderByNumber"),
+        Some("GetBlockInputs") => tracing::info_span!("store.rpc/GetBlockInputs"),
+        Some("GetBatchInputs") => tracing::info_span!("store.rpc/GetBatchInputs"),
+        Some("GetNotesById") => tracing::info_span!("store.rpc/GetNotesById"),
+        Some("GetTransactionInputs") => tracing::info_span!("store.rpc/GetTransactionInputs"),
+        Some("SyncNotes") => tracing::info_span!("store.rpc/SyncNotes"),
+        Some("SyncState") => tracing::info_span!("store.rpc/SyncState"),
+        _ => tracing::info_span!("store.rpc/Unknown"),
+    };
```
Mirko-von-Leipzig, line 42: `tracing::span!` requires a compile-time constant string.

A more maintenance free alternative is to allow the open-telemetry and stdout tracing to drift apart some more. We could set the `otel.name` property which overrides the span name for open-telemetry (currently it just defaults to the span name).

This would mean stdout would get a more generic `name = store.rpc` and a separate `rpc.method = {method}` attribute.

Might be fine/better than this manual matching?

crates/utils/src/tracing/grpc.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// A [`trace_fn`](tonic::transport::server::Server) implementation for the store which adds
+/// open-telemetry information to the span.
+///
+/// Creates an `info` span following the open-telemetry standard: `store.rpc/{method}`. Additionally
+/// also pulls in remote tracing context which allows the server trace to be connected to the
+/// client's origin trace.
+pub fn store_trace_fn(request: &http::Request<()>) -> tracing::Span {
+    let span = match request.uri().path().rsplit('/').next() {
+        Some("ApplyBlock") => tracing::info_span!("store.rpc/ApplyBlock"),
```
Mirko-von-Leipzig, line 27: Open-telemetry standard specifies `{package}.{service}/{method}` for RPC span names. Technically our protobuf says the service is called `Api`.

crates/utils/src/tracing/grpc.rs
bobbinth: Ah, I think this addresses the m question above. I think we can leave it for a future PR.

crates/block-producer/src/store/mod.rs
```diff
@@ -121,17 +121,19 @@ impl TryFrom<GetTransactionInputsResponse> for TransactionInputs {
...
 #[derive(Clone)]
 pub struct StoreClient {
-    inner: store_client::ApiClient<Channel>,
+    inner: InnerClient,
 }
 
 impl StoreClient {
     /// TODO: this should probably take store connection string and create a connection internally
-    pub fn new(store: store_client::ApiClient<Channel>) -> Self {
+    pub fn new(store: InnerClient) -> Self {
```
bobbinth, line 136: Is the `TODO` still relevant here?

bobbinth (approved):
Looks good! Thank you! A very light review from me, but I did leave a few comments inline.

bobbinth:
> Note that the RPC server is missing this -- the tracing is restricted to internal only APIs. We _could_ consider expanding this to include other services e.g. the faucet. We would have to authenticate these requests somehow, maybe with an authentication header or API key. Though this might be of questionable value at this stage.

I would leave this for the future (and not sure it is needed either).

> This may also mean that the current implementation can't cater to them e.g. client IP and server IP are not available to the middleware I used.

This means that, for example, the RPC component won't assign IP addresses to the request traces?

crates/block-producer/src/store/mod.rs
Mirko-von-Leipzig: I think so. Or at least I think the api/usage could be improved. It feels a bit clunky to use and maybe needs a revisit at some stage, though its relatively low priority.

I imagine this would be annoying to use in e.g. test setups - where we don't care about otel etc.

Mirko-von-Leipzig:
> > This may also mean that the current implementation can't cater to them e.g. client IP and server IP are not available to the middleware I used.
> 
> This means that, for example, the RPC component won't assign IP addresses to the request traces?

Correct. Right now the implementation does only the bare minimum to connect traces together. From the `tonic` docs on `Interceptor`:

> gRPC interceptors are similar to middleware but have less flexibility. An interceptor allows you to do two main things, one is to add/remove/check items in the MetadataMap of each request. Two, cancel a request with a Status.
>
> [snip]
>
> Additionally, interceptors is not the recommended way to add logging to your service. For that a [tower](https://crates.io/crates/tower) middleware is more appropriate since it can also act on the response. For example tower-http’s [Trace](https://docs.rs/tower-http/latest/tower_http/trace/index.html) middleware supports gRPC out of the box.

I just wanted to get it working for now; and this actually had examples on how to achieve it.

crates/block-producer/src/server.rs
```diff
@@ -52,11 +53,14 @@ impl BlockProducer {
     pub async fn init(config: BlockProducerConfig) -> Result<Self, ApiError> {
         info!(target: COMPONENT, %config, "Initializing server");
 
-        let store = StoreClient::new(
-            store_client::ApiClient::connect(config.store_url.to_string())
-                .await
-                .map_err(|err| ApiError::DatabaseConnectionFailed(err.to_string()))?,
-        );
+        let channel = tonic::transport::Endpoint::try_from(config.store_url.to_string())
+            .map_err(|err| ApiError::DatabaseConnectionFailed(err.to_string()))?
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I'll make it more specific.

Technically this is the same behaviour as before - where `Client::connect(&str)` would perform both the url parsing as well attempt to connect.

PhilippGackstatter (approved):
Looks good to me!

igamigo (approved):
LGTM!

PR #672, opened 2025-02-07
fix(faucet): missing background and favicon routes

### LINKED ISSUE #671: Faucet website's background is not served
> Related to the faucet but not directly to this PR: The background does not seem to be showing. It tries to fetch it from https://faucet.testnet.miden.io/background.png, but it is not there. You can still get it from https://testnet.miden.io/background.png, though

@Mirko-von-Leipzig - could you take a look at this?

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-client/issues/721#issuecomment-2641070480_

Mirko-von-Leipzig:
Closes #671.

Created #673 to create tests for this.

--------------

I've deployed this to devnet and confirmed that 

- both background and favicon load as expected, and
- no console errors

Did require clearing my cache first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Faucet webpage is missing `background.png` and `favicon.ico` (#672).
+
 ### Enhancements
 
 - Add an optional open-telemetry trace exporter (#659).
--- a/bin/faucet/src/handlers.rs
+++ b/bin/faucet/src/handlers.rs
@@ -128,6 +128,14 @@ pub async fn get_index_css(state: State<FaucetState>) -> Result<impl IntoRespons
     get_static_file(state, "index.css")
 }
 
+pub async fn get_background(state: State<FaucetState>) -> Result<impl IntoResponse, HandlerError> {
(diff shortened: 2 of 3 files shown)


bobbinth (approved):
Looks good! Thank you! Question: should this go into `main`?

bobbinth:
> If we care to make a new release with it, then `main` probably.

Maybe we release just the `faucet` crate? And for a change like this, we may even skip tagging a release.

Mirko-von-Leipzig:
> > If we care to make a new release with it, then `main` probably.
> 
> Maybe we release just the `faucet` crate? And for a change like this, we may even skip tagging a release.

I'll rebase to `main`.

Its a bit awkward with the formal deployment, options:

1. We use `Deploy` from the main branch - no release required. But this will currently redeploy everything. Would have to add an option to only deploy the faucet.
2. We manually deploy...
    1. using the faucet debian package - but this requires a github release to build it
    2. by manually building on the instance

(2ii) is the easiest and probably also the safest - but not popular with devops of course :D

bobbinth:
eh - seems like all of these are a bit of a hassle for just adding a background image. let's just keep things as they are now.

PR #685, opened 2025-02-11
WIP: stream SyncState response

TomasArrachea:


crates/store/src/server/api.rs
```diff
@@ -147,68 +152,97 @@ impl api_server::Api for StoreApi {
...
-        Ok(Response::new(SyncStateResponse {
-            chain_tip: self.state.latest_block_num().await.as_u32(),
-            block_header: Some(state.block_header.into()),
-            mmr_delta: Some(delta.into()),
-            accounts,
-            transactions,
-            notes,
-            nullifiers,
-        }))
+        let (tx, rx) = mpsc::channel(128); // TODO: check bound of the channel
```
TomasArrachea, line 163: This is still TBD, any ideas of what should be the channel size for the stream?

tomyrd:
I have a question regarding a small edge case that I had to deal with in the [sync state component refactor](https://github.com/0xPolygonMiden/miden-client/pull/650) in the client.

How do we deal with nullifiers for notes that we don't know that they exist that got created and nullified between the request block num and the chain tip?

The client can't add the nullifier to the request because it doesn't know of the note's existance so will it miss the nullifier update. Moreover, now that the client is synced to a block after the nullifier, it won't get the update in future requests (which would be the requests containing the new nullifier).

Here's a small diagram of the case:
image

Before, with separate requests, the client could update subsquent requests with the new information it got from the node.

Mirko-von-Leipzig:
> I have a question regarding a small edge case that I had to deal with in the [sync state component refactor](https://github.com/0xPolygonMiden/miden-client/pull/650) in the client.
> 
> How do we deal with nullifiers for notes that we don't know that they exist that got created and nullified between the request block num and the chain tip?
> 
> The client can't add the nullifier to the request because it doesn't know of the note's existance so will it miss the nullifier update. Moreover, now that the client is synced to a block after the nullifier, it won't get the update in future requests (which would be the requests containing the new nullifier).

To me this sounds like a case for more separate queries. As in, maybe `SyncState` shouldn't include nullifiers - but rather they should be requested separately?

igamigo:
> To me this sounds like a case for more separate queries. As in, maybe `SyncState` shouldn't include nullifiers - but rather they should be requested separately?

I think this is probably the way to go, use `CheckNullifiersByPrefix` after a sync. It would force most syncs to imply two separate requests, which is probably fine since right now a sync requires potentially multiple requests until you get to the chain tip anyway.

For public notes the node could also track notes and add nullifiers, but for private notes it would require a separate request anyway.

Mirko-von-Leipzig:
@igamigo @TomasArrachea can I review this PR as is? Or will there be changes towards separating nullifiers etc out? The latter probably needs at least some broader discussion, e.g. maybe we should have more focussed streams?

bobbinth:
So, I'd probably break this down as follows:

- in a separate PR, update `CheckNullifiersByPrefix` as described above.
- Then, in a separate PR, remove nullifiers from the `SyncState` request/response, and update the client accordingly.
- Then, finish this PR.

crates/rpc/README.md
```diff
@@ -37,7 +37,7 @@ Returns a nullifier proof for each of the requested nullifiers.
 
 ### CheckNullifiersByPrefix
 
-Returns a list of nullifiers that match the specified prefixes and are recorded in the node.
+Returns a list of nullifiers recorded in the node that match the specified prefixes and block creation height.
```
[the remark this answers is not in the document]
bobbinth: Yes, agreed - let's make it more precise - i.e., the response will contain nullifiers created in or after the specified block.

TomasArrachea:
> Could we cherry-pick https://github.com/0xPolygonMiden/miden-node/commit/eb2de00d00005206551b672093506b7ce24e386e and changes from https://github.com/0xPolygonMiden/miden-node/pull/708 once they are merged into next? This way, they could be merged and propagated to the client before this PR is merged (we can then rebase this PR from next).

Actually #707 was merged into this branch, not into `next`. So I should create a new PR to merge the changes from #707 into `next`

crates/store/src/server/api.rs
```diff
@@ -151,53 +156,83 @@ impl api_server::Api for StoreApi {
...
+                }
+                .await;
+
+                if let Ok(response) = &result {
+                    // SAFETY: SyncStateResponse always has a block header
+                    last_block_num = response.block_header.unwrap().block_num;
+                }
+                let is_error = result.is_err();
+
+                tx.send(result).await.expect("error sending sync state response to the client");
```
Mirko-von-Leipzig, line 226: This can happen if the client closes its connection. In other words, this can be caused by an external force - and shouldn't cause a panic for us. Rather `if let Err(err)`, log the closure (probably at `info`) and break the loop.

aka this isn't actually an error, but rather just indicates the connection was lost which is entirely valid. I think.

crates/store/src/server/api.rs
Mirko-von-Leipzig: Good question. Each element is essentially at most one block's worth of delta's iiuc.

I would make it much smaller - like 5? @bobbinth

This also interacts with the timeout on the sender from my other comment.

crates/store/src/server/api.rs
Mirko-von-Leipzig: This also block's correct? We should wrap it in a timeout to prevent a client from slow-rolling us.

.gitignore
```diff
@@ -37,4 +37,4 @@ env/
 *.out
 node_modules/
 *DS_Store
-*.iml
\ No newline at end of file
+*.iml
```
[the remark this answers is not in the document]
TomasArrachea: Just the formatter adding the EOF

crates/store/src/server/api.rs
TomasArrachea: It only blocks when there is no more space on the channel, so it's related to the bound size discussion. Anyway I added a 10 secs timeout in [c3b242b](https://github.com/0xPolygonMiden/miden-node/pull/685/commits/c3b242b9302f9107abbe49ba51cdf93285ed392e), along with some error handling

PR #690, opened 2025-02-13
chore: upgrade otel crates and apply filters to exporter

### LINKED ISSUE #682: Fix: Otel subscriber does not filter traces
It currently appears to emit all traces.

### LINKED ISSUE #688: Upgrade open-telemetry to v0.28
We're currently on v0.27 (and somehow `tracing_opentelemetry` is v0.28 already).

There is an [upgrade guide](https://github.com/open-telemetry/opentelemetry-rust/blob/main/docs/migration_0.28.md) available, and doing this sooner is probably better than later.

Mirko-von-Leipzig:
This PR upgrades the `otel` family of crates from v0.27 to v0.28. The `tracing_opentelemetry` crate is off-by-one.

I've added filtering to the open-telemetry trace exporter. Previously it was exporting everything at a rate of ~150 spans/events per second, which has now dropped to ~3.

I've also refactored the tracing/logging setup to get rid of the redundancy and make setup more obvious.

I'm not adding this to the changelog since no release contains otel exporting yet.

Closes #688 and closes #682

### DIFF
--- a/bin/faucet/src/main.rs
+++ b/bin/faucet/src/main.rs
@@ -17,7 +17,9 @@ use client::initialize_faucet_client;
 use handlers::{get_background, get_favicon, get_index_css, get_index_html, get_index_js};
 use http::HeaderValue;
 use miden_lib::{account::faucets::create_basic_fungible_faucet, AuthScheme};
-use miden_node_utils::{config::load_config, crypto::get_rpo_random_coin, version::LongVersion};
+use miden_node_utils::{
+    config::load_config, crypto::get_rpo_random_coin, logging::OpenTelemetry, version::LongVersion,
+};
 use miden_objects::{
     account::{AccountFile, AccountStorageMode, AuthSecretKey},
     asset::TokenSymbol,
@@ -89,7 +91,8 @@ pub enum Command {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
-    miden_node_utils::logging::setup_logging().context("Failed to initialize logging")?;
+    miden_node_utils::logging::setup_tracing(OpenTelemetry::Disabled)
+        .context("Failed to initialize logging")?;
 
     let cli = Cli::parse();
 
--- a/bin/faucet/src/main.rs
+++ b/bin/faucet/src/main.rs
@@ -92,14 +92,14 @@ pub enum Command {
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     miden_node_utils::logging::setup_tracing(OpenTelemetry::Disabled)
-        .context("Failed to initialize logging")?;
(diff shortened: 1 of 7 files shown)


bin/faucet/src/main.rs
```diff
@@ -89,7 +91,8 @@ pub enum Command {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
-    miden_node_utils::logging::setup_logging().context("Failed to initialize logging")?;
+    miden_node_utils::logging::setup_tracing(OpenTelemetry::Disabled)
+        .context("Failed to initialize logging")?;
```
bobbinth, line 95: nit: should we lower-case messages? i.e., "Failed" -> "failed"

bobbinth (approved):
Thank you! Not a very deep review from me, but looks good!

crates/utils/src/logging.rs
```diff
@@ -1,29 +1,46 @@
...
+    Disabled,
+}
+
+impl OpenTelemetry {
+    fn is_enabled(self) -> bool {
+        matches!(self, OpenTelemetry::Enabled)
+    }
+}
 
 /// Configures tracing and optionally enables an open-telemetry OTLP exporter.
```
igamigo, line 26: You can easily see this in the code (and the `env_or_default_filter` function) but this could also explain that there's an stdout layer and that filters are based on `RUST_LOG`.

bin/node/src/main.rs
```diff
@@ -89,8 +89,8 @@ async fn main() -> anyhow::Result<()> {
 
     // Open telemetry exporting is only valid for running the node.
     let open_telemetry = match &cli.command {
-        Command::Start { open_telemetry, .. } => *open_telemetry,
-        _ => false,
+        Command::Start { open_telemetry, .. } if *open_telemetry => OpenTelemetry::Enabled,
+        _ => OpenTelemetry::Disabled,
     };
```
igamigo, line 94: nitL This looks marginally clearer to me, feel free to disregard
```
let open_telemetry = if let Command::Start { open_telemetry: &true, .. } = &cli.command {
    OpenTelemetry::Enabled
} else {
    OpenTelemetry::Disabled
};
```

igamigo (approved):
LGTM. Skimmed through the upgrade guide and didn't see anything of note with what I know of the current setup.

> I'm not adding this to the changelog since no release contains otel exporting yet.

Maybe this PR number can be added to the same log line for the original otel exporter notice? Since the setup changed a bit.

bin/node/src/main.rs
Mirko-von-Leipzig: Ah I never even considered pattern matching inside like that. I do prefer your variant, thank you.

PR #693, opened 2025-02-13
feat: use remote batch prover

SantiagoPittella:
Currently, it is sending batches to prove, the proof generation succeeds but an error is occurring because it is finally mark as fail in node's side.


SantiagoPittella:
Closing it in favor of a newer, less noisy PR.

Issue #681, opened 2025-02-11
Instrument HTTP requests & responses


Mirko-von-Leipzig:
Currently remote context is handled via gRPC middleware which does _not_ have access to the http request, only the gRPC metadata. This means that we cannot trace client/server IP or other nice to have http data.

See [here](https://github.com/0xPolygonMiden/miden-node/pull/669#issuecomment-2638967432) for context, but it will likely include switching to a tower layer instead. Consider simplifying the RPC span creation as well, as mentioned [here](https://github.com/0xPolygonMiden/miden-node/pull/669#discussion_r1943366272).

Also see the open-telemetry standard on what to consider including as additional metadata.

sergerad:
Hey @Mirko-von-Leipzig just checking my understanding that the main unlock here would be to leverage an impl like [this](https://github.com/hyperium/tonic/blob/master/examples/src/tower/server.rs#L83) on the server so as to get HTTP metadata via `http::Request` type and potentially the body via the generic `ReqBody: Send + 'static,` (e.g. `+ Display` or `+ Debug`)? TY

Mirko-von-Leipzig:
@sergerad I haven't dug in too deeply, my current knowledge is mostly based on `tonic`'s [Interceptor docs](https://docs.rs/tonic/latest/tonic/service/trait.Interceptor.html):

> If you need more powerful middleware, [tower](https://crates.io/crates/tower) is the recommended approach. You can find examples of how to use tower with tonic [here](https://github.com/hyperium/tonic/tree/master/examples/src/tower).
>
> Additionally, interceptors is not the recommended way to add logging to your service. For that a [tower](https://crates.io/crates/tower) middleware is more appropriate since it can also act on the response. For example tower-http’s [Trace](https://docs.rs/tower-http/latest/tower_http/trace/index.html) middleware supports gRPC out of the box.

I know `tower` has built-in [grpc-tracing ](https://docs.rs/tower-http/latest/tower_http/trait.ServiceBuilderExt.html#tymethod.trace_for_grpc) support but that is just some default implementation based on gRPC headers.

We probably need to customise things, in which case I believe the [trace](https://docs.rs/tower-http/latest/tower_http/trace/index.html) is the next level of abstraction. Hopefully this is enough, though maybe we have to drop down to the example you've linked.

Mirko-von-Leipzig:
Closed by #698

PR #698, opened 2025-02-16
chore: add Open Telemetry attributes to grpc spans

sergerad:
Relates to #681.

- Use `tower::TraceLayer` in the tonic server builder for block producer rpc and store rpc
- Add attributes to rpc spans based on HTTP request values and Open Telemetry standards


crates/block-producer/src/server.rs
```diff
@@ -211,8 +212,17 @@ impl BlockProducerRpcServer {
...
     async fn serve(self, listener: TcpListener) -> Result<(), tonic::transport::Error> {
+        let trace_layer = TraceLayer::new_for_grpc()
+            .make_span_with(miden_node_utils::tracing::grpc::block_producer_trace_fn)
+            .on_request(|_request: &http::Request<_>, _span: &tracing::Span| todo!())
+            .on_response(
+                |_response: &http::Response<_>, _latency: Duration, _span: &tracing::Span| todo!(),
+            )
+            .on_failure(
+                |_error: GrpcFailureClass, _latency: Duration, _span: &tracing::Span| todo!(),
+            );
```
sergerad, line 223: @Mirko-von-Leipzig I agree that this is probably the correct abstraction to be using here (as you mentioned in [this](https://github.com/0xPolygonMiden/miden-node/issues/681#issuecomment-2660824531) comment).

Any suggestions on how to test this locally? TY

Mirko-von-Leipzig: Its definitely a bit awkward to test. There is the [text exporter](https://docs.rs/opentelemetry-stdout/0.28.0/opentelemetry_stdout/) which at least lets you debug/inspect the manually.

If we do want to uni/integration test these things, we'll probably have to write our own exporter - e.g. something to aggregate the data which we can then assert against. I'm unsure why this isn't already available tbh.

--------------------

What I've been doing in the meantime is using https://www.honeycomb.io/ with a free account to test things out. There's some setup info in the wip guide [here](https://github.com/0xPolygonMiden/miden-node/pull/699).

I essentially start a local node, configure otel to use the honeycomb endpoint and then generate traffic using the `miden-client` [integration test](https://github.com/0xPolygonMiden/miden-client/blob/f4d3688c26499d02103706f7b1897fc43c5e8e24/Makefile#L96-L98) against my local node.

This might be a pita though - let me know if you run into issues, or if you can think of something smarter :D

Mirko-von-Leipzig: Actually here we go: https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/testing/trace/index.html

They do actually have test infra :)

sergerad: Thanks just getting back to this now. Will have a crack at integrating those trace test capabilities in the unit tests.

sergerad: > unit tests

Or integration tests rather. Which would involve allowing the node to be configured with different SpanExporters to allow the test one to be used.

Mirko-von-Leipzig: Yeah I imagine this may take some more consideration, given that the exporters are (always?) global.

Might be good to simply explore the options and then discuss what _is_ actually testable without much fuss. 

Integration tests are a sore point at the moment - and likely will get worse before getting better. e.g. this [thread](https://github.com/0xPolygonMiden/miden-node/pull/701#discussion_r1959083239). Its not quite trivial to just spin up a node at the moment, but depending on what you discover here maybe that's something we should aim at.

sergerad: @Mirko-von-Leipzig This is how I think the testing would have to work:
1. Update a few functions to take in an `impl SpanExporter`.
2. Update node and faucet main.rs to construct either a test `SpanExporter` or an actual one based on flags or toml.
3. Implement tests that call the main functionality and assert on the `SpanData` coming out of the test `SpanExporter`. These would probably be the existing integration tests but I haven't had a look yet.

But I don't think it is worthwhile. I have seen Rust projects do similar things where they run unit or integration tests and assert against log lines produced by the stack. I have generally avoided this kind of thing in the past, but I can imagine situations where its worth the maintenance cost and brittleness (tests being coupled to log lines).

I think it would become practical to just test trace output changes on the dev deployment/cluster once the trace scaffolding/impl stabilizes for the node. Rather than testing every change locally or in CI.

---

I have connected to honeycomb successfully and eye-balled the `info!` output coming from my latest changes, e.g.:
```
2025-02-21T06:22:59.610587Z  INFO block-producer.rpc/SubmitProvenTransaction: miden_node_block_producer::server: crates/block-producer/src/server.rs:218: request: POST /block_p
roducer.Api/SubmitProvenTransaction {"te": "trailers", "content-type": "application/grpc", "traceparent": "00-20616281d08ea7970f0701e71dd2ef80-453f6f8340af0947-01", "tracestate": "",
 "user-agent": "tonic/0.12.3"}
```

Still getting my head around exactly what we want to achieve and whether the approach I've added so far fulfils that. The example above only has grpc related headers so I think I'll need to swap
```rust
        let trace_layer = TraceLayer::new_for_grpc()
```
to 
```rust
        let trace_layer = TraceLayer::new_for_http()
```

Mirko-von-Leipzig: > But I don't think it is worthwhile. I have seen Rust projects do similar things where they run unit or integration tests and assert against log lines produced by the stack. I have generally avoided this kind of thing in the past, but I can imagine situations where its worth the maintenance cost and brittleness (tests being coupled to log lines).

I agree; isn't worth it especially not at our current project maturity level. I think I was hoping we could do something like:

```rust
#[test]
async fn block_builder_trace() {
    let store = mock_store();
    let exporter = TestSpanExporter::new()...;
    exporter.register_for_this_test_only();

    BlockBuilder::build_block(...).await;

    assert_eq!(exporter.spans, expected_spans);
}
```

But I suspect its not trivially possible. Though maybe there is a way to have the global exporter registered as `TestSpanExporter` and somehow associate the spans we get from this test only.

--------------------

> The example above only has grpc related headers so I think I'll need to swap

Yeah. I think the important things to get in are the server/client IP addresses which will (probably) only be available for http I imagine? Long term we'll also want to add interesting headers and/or CORS information maybe, but I'm unsure.

sergerad: The http fn only differs from the grpc one w.r.t the error type handled by error callback:
```rust
    /// Create a new [`TraceLayer`] using [`ServerErrorsAsFailures`] which supports classifying
    /// regular HTTP responses based on the status code.
    pub fn new_for_http() -> Self {
        Self {
            make_classifier: SharedClassifier::new(ServerErrorsAsFailures::default()),
// vs:      make_classifier: SharedClassifier::new(GrpcErrorsAsFailures::default()),
```
Think the best we can do is `uri.host`. We won't be able to see source/client IP unless its put into a header by a proxy or some other part of this stack.

I suppose there is still value in the `tower_http::trace::TraceLayer` because we can register callbacks here:
```rust
            .on_request(|request: &http::Request<_>, _span: &tracing::Span| {
                tracing::info!(
                    "request: {} {} {} {:?}",
                    request.method(),
                    request.uri.host
                    request.uri().host().unwrap_or("NOHOST"),
                    request.uri().path(),
                    request.headers()
                );
            })
            .on_response(
                |response: &http::Response<_>, latency: Duration, _span: &tracing::Span| {
                    tracing::info!("response: {} {:?}", response.status(), latency);
                },
            )
            .on_failure(|error: GrpcFailureClass, _latency: Duration, _span: &tracing::Span| {
                tracing::error!("error: {}", error);

```

Here is the diff in error type for the on failure callback:
```rust
/// The failure class for [`ServerErrorsAsFailures`].
#[derive(Debug)]
pub enum ServerErrorsFailureClass {
    /// A response was classified as a failure with the corresponding status.
    StatusCode(StatusCode),
    /// A response was classified as an error with the corresponding error description.
    Error(String),
}
...
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct StatusCode(NonZeroU16);
...
```
vs
```rust
/// The failure class for [`GrpcErrorsAsFailures`].
#[derive(Debug)]
pub enum GrpcFailureClass {
    /// A gRPC response was classified as a failure with the corresponding status.
    Code(std::num::NonZeroI32),
    /// A gRPC response was classified as an error with the corresponding error description.
    Error(String),
}
```
So might as well go with the http one I guess.

I'll look at the unit test approach you mentioned above and see if its possible.

crates/block-producer/src/mempool/tests.rs
```diff
@@ -16,6 +16,29 @@ impl Mempool {
...
+            .attributes
+            .iter()
+            .filter(|kv| {
+                kv.key == "code.namespace".into()
+                    && kv.value == "miden_node_block_producer::mempool".into()
+            })
+            .count(),
+        1,
+    );
+}
```
sergerad, line 40: @Mirko-von-Leipzig I added this unit test to illustrate the otel trace test capability. It doesn't relate to the functionality added in this PR. This was just the easiest unit test I could find to impl. The block producer rpc stack doesn't have mocks etc atm.

LMK if you want to keep/rm this test or move it to another PR with other tests etc.

crates/block-producer/src/server.rs
```diff
@@ -211,8 +212,27 @@ impl BlockProducerRpcServer {
...
+                    request.uri().path(),
+                    request.headers()
+                );
+            })
+            .on_response(|response: &http::Response<_>, latency: Duration, _span: &Span| {
+                info!("response: {} {:?}", response.status(), latency);
+            })
+            .on_failure(|error: ServerErrorsFailureClass, latency: Duration, _span: &Span| {
+                error!("error: {} {:?}", error, latency);
+            });
```
sergerad, line 232: @Mirko-von-Leipzig I have checked this locally and on honeycomb. LMK any changes you would make here.
Screenshot 2025-02-22 at 10 45 51 AM

crates/block-producer/src/mempool/tests.rs
Mirko-von-Leipzig: Let's leave it for a separate PR.

Is it possible to separate by test itself - or do we need to run these kind of tests sequentially to ensure we get the spans recorded that we expect?

As in, running tests in parallel probably muddles the span exporter unless we somehow mark the spans with a test ID?

But this looks promising, we can figure out the specifics in the other PR/issue.

crates/block-producer/src/server.rs
Mirko-von-Leipzig: Right now you're emitting events, but I'd like to attach whatever we can as attributes on the span.

For context, the open-telemetry specification has a bunch of (experimental) suggestions:

- https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#server-attributes
- https://opentelemetry.io/docs/specs/semconv/rpc/grpc/

We have access to only some of them here I think.

crates/block-producer/src/server.rs
```diff
@@ -211,8 +212,27 @@ impl BlockProducerRpcServer {
     }
 
     async fn serve(self, listener: TcpListener) -> Result<(), tonic::transport::Error> {
+        // Configure the trace layer with callbacks.
+        let trace_layer = TraceLayer::new_for_grpc()
+            .make_span_with(miden_node_utils::tracing::grpc::block_producer_trace_fn)
```
Mirko-von-Leipzig, line 217: We can consider moving this entire layer into the `utils` crate since its likely going to be identical.

crates/block-producer/src/server.rs
sergerad: I just can't find a way to surface lower level details like client addr to the relevant stacks here.

The `tonic::Request` struct has `::remote_addr() -> Option<SocketAddr>` but we can't get that type wired through into our trace stack here AFAICT, only `http::Request`. And I can't find a way to get headers to contain the remote IP address via `http::Request` in any way.

Relevant links for reference
https://github.com/hyperium/tonic/issues/430
https://github.com/tower-rs/tower-http/blob/main/examples/tonic-key-value-store/src/main.rs#L191

sergerad: Should be able to get socket addr via this when building the tonic server:
https://docs.rs/tonic/0.2.1/tonic/transport/server/trait.Connected.html

Will give it a go

crates/block-producer/src/server.rs
sergerad: Potentially the way to do it:
https://stackoverflow.com/questions/71178212/how-to-configure-tower-http-tracelayer-in-a-separate-function

Mirko-von-Leipzig: Oh damn I see. Little corner cases everywhere.

Thanks, yeah I think lets go with the stackoverflow answer - and just document on the function why we're doing it this way.

crates/block-producer/src/server.rs
Mirko-von-Leipzig: Wow this was crazy hard to find: https://github.com/tower-rs/tower-http/discussions/428.

Only by chance did I go check in the discussions - no other google/search-kungfu found it.

I believe this is what we need? It may also mean that we have to just do our own layer to extract some of the info before handling the routes/request/responses.

sergerad: Ah, well done! This looks to have worked. e.g.:
```rust
.on_request(|request: &http::Request<_>, _span: &Span| {
    info!(
        "request: {} {} {} {} {:?}",
        request
            .extensions()
            .get::<tonic::transport::server::TcpConnectInfo>() // as per axum example above
            .unwrap()
            .remote_addr()
            .unwrap(),
```

I'll put together all the changes from your comments and put this in review next.

crates/store/src/server/mod.rs
```diff
@@ -61,8 +62,9 @@ impl Store {
     ///
     /// Note: this blocks until the server dies.
     pub async fn serve(self) -> Result<(), ApiError> {
+        // Build the gRPC server with the API service and trace layer.
         tonic::transport::Server::builder()
-            .trace_fn(miden_node_utils::tracing::grpc::store_trace_fn)
+            .layer(TraceLayer::new_for_grpc().make_span_with(store_trace_fn))
```
sergerad, line 67: Think the default callbacks might suffice for this given we are adding request-related attributes to the span. Note that the defaults ones are debug, not info.

crates/utils/src/tracing/span_ext.rs
```diff
@@ -48,21 +59,53 @@ impl ToValue for i64 {
 ///
 /// This is a sealed trait. It and cannot be implemented outside of this module.
 pub trait OpenTelemetrySpanExt: private::Sealed {
+    fn set_parent<T>(&self, request: &http::Request<T>);
     fn set_attribute(&self, key: impl Into<Key>, value: impl ToValue);
+    fn set_http_attributes<T>(&self, request: &http::Request<T>);
     fn set_error(&self, err: &dyn std::error::Error);
+    fn context(&self) -> opentelemetry::Context;
```
Mirko-von-Leipzig, line 66: I wouldn't add these to the trait. Bear in mind that this trait is used by all other crates - and they have no use for setting the parent from a request, nor http attributes nor context. In fact doing so would be an error / mistake as its only useful for tracing gRPC entry-points/requests.

In other words they should remain private / internal, which probably means they should just revert to free standing functions.

crates/utils/src/tracing/span_ext.rs
```diff
@@ -48,21 +59,53 @@ impl ToValue for i64 {
...
         tracing_opentelemetry::OpenTelemetrySpanExt::set_attribute(self, key, value.to_value());
     }
 
+    /// Sets standard attributes to the `Span` based on an associated HTTP request.
+    fn set_http_attributes<T>(&self, request: &http::Request<T>) {
+        let remote_addr = request
+            .extensions()
+            .get::<tonic::transport::server::TcpConnectInfo>()
+            .and_then(tonic::transport::server::TcpConnectInfo::remote_addr);
+        OpenTelemetrySpanExt::set_attribute(self, "remote_addr", remote_addr);
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#server-attributes

crates/store/src/server/mod.rs
Mirko-von-Leipzig: I think that's fine yeah

crates/block-producer/Cargo.toml
```diff
@@ -24,7 +24,7 @@ itertools             = { workspace = true }
 miden-block-prover    = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "next" }
 miden-lib             = { workspace = true }
 miden-node-proto      = { workspace = true }
-miden-node-utils      = { workspace = true }
+miden-node-utils      = { workspace = true, features = ["testing"] }
```
Mirko-von-Leipzig, line 27: The `testing` feature should only be set for `development` and not in the general case. 

Though I suggest we move all the test code out into a separate PR, i.e. remove the feature entirely for now.

crates/utils/src/tracing/grpc.rs
```diff
@@ -1,16 +1,25 @@
-use tracing_opentelemetry::OpenTelemetrySpanExt;
+/// Creates a [`tracing::Span`] based on RPC service and method name.
+macro_rules! rpc_span {
```
sergerad, line 2: This seemed like the best way of feeding in the rpc.* attributes into the span. Please double check my usage of `info_span!` w.r.t the attributes.

Mirko-von-Leipzig: I'm actually wondering - what does the full rpc request path look like? 

Maybe we can avoid the path splitting and macro entirely if the path is good enough?

As in, the path might be `xxxx/store.api/{method}`

crates/utils/src/tracing/grpc.rs
```diff
@@ -26,25 +33,68 @@ pub fn block_producer_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
...
+    tracing_opentelemetry::OpenTelemetrySpanExt::set_parent(&span, otel_ctx);
+
+    // Set HTTP attributes.
+    // See https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#server-attributes.
+    span.set_attribute("rpc.system", "grpc");
+    if let Some(host) = request.uri().host() {
+        span.set_attribute("server.address", host);
+    }
+    if let Some(host_port) = request.uri().port() {
+        span.set_attribute("server.port", host_port.as_str());
```
Mirko-von-Leipzig, line 80: I think this one must be an integer (according to spec).

crates/utils/src/tracing/grpc.rs
```diff
@@ -26,25 +33,68 @@ pub fn block_producer_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
...
+    }
+    let remote_addr = request
+        .extensions()
+        .get::<tonic::transport::server::TcpConnectInfo>()
+        .and_then(tonic::transport::server::TcpConnectInfo::remote_addr);
+    if let Some(addr) = remote_addr {
+        span.set_attribute("client.address", addr.ip());
```
Mirko-von-Leipzig, line 87: This is the one that the spec says _may_ be the DNS name if no DNS lookup is required.

> Client address - domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name.

I'm wondering if we shouldn't use the `text_map_propagator` to cheat the system a bit, and on the client side inject `block-producer` and `rpc` as an additional property.

We could then use that here, and otherwise default to the IP if its missing.

Could also be done as a follow-up PR - I imagine you might be very much over this back-and-forth :D

crates/utils/src/tracing/span_ext.rs
```diff
@@ -20,29 +21,53 @@ impl ToValue for Digest {
...
+macro_rules! impl_to_value {
+    ($($t:ty),*) => {
+        $(
+            impl ToValue for $t {
+                fn to_value(&self) -> Value {
+                    (*self).into()
+                }
+            }
+        )*
+    };
```
Mirko-von-Leipzig, line 68: Might be more idiomatic as a blanket implementation:

```rust
impl<T> ToValue for T 
where T: Into<Value> + Clone {
    fn to_value(&self) -> Value {
        self.clone().into()
    }
}
```

The downside is this blanket impl doesn't let you choose specific types.

Mirko-von-Leipzig (commented):
This is looking really good imo.

Couple more questions but really just minor things at this stage.

crates/utils/src/tracing/grpc.rs
sergerad: > Could also be done as a follow-up PR - I imagine you might be very much over this back-and-forth :D

Not at all - happy to iterate on this PR as much as required.

>I'm wondering if we shouldn't use the text_map_propagator to cheat the system a bit, and on the client side inject block-producer and rpc as an additional property.
We could then use that here, and otherwise default to the IP if its missing.

I'm a bit dubious as to the utility of that functionality but happy to add that to this PR if you think it is worthwhile.

crates/utils/src/tracing/span_ext.rs
sergerad: I think at that point you lose the benefit of having the separate `ToValue` trait which is to be able to pass in types without first converting them to `f64` or `i64` etc.

crates/utils/src/tracing/grpc.rs
Mirko-von-Leipzig: The main reason I'm considering it is because our components can (and currently do) run on the same server instance (and same process). Which means for the inter-component comms, the client and server IP address will just be the same since they're the same for all three components.

So I think it would help having the "caller" identifiable at least. Though maybe instead we should be hosting different api endpoints for this internal communication in any case.

I'm happy to just punt this to a separate issue instead.

crates/utils/src/tracing/span_ext.rs
Mirko-von-Leipzig: Yeah it does encourage poor practices. Let's leave it as is 👍

crates/utils/src/tracing/grpc.rs
Mirko-von-Leipzig: Ah wait. The issue is that `tracing::span_info!` requires static strings right?

Hmm. Okay so there isn't really a way to do this properly for the span name itself since you cannot do it dynamically.

crates/utils/src/tracing/grpc.rs
sergerad: Ah ok yea that makes sense. Thanks for elaborating

sergerad:
> I'm actually happy to merge this now, if you're happy?

All good from my end 👍

PR #700, opened 2025-02-17
Add custom OpenTelemetrySpanExt trait

### LINKED ISSUE #678: Custom span extension trait
`tracing_opentelemetry` has the [`OpenTelemetrySpanExt`](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/trait.OpenTelemetrySpanExt.html) extension trait, of which we currently use `set_attribute` and `set_status`. We use the former to add arbitrary fields to the span, and the latter to mark a span as an error and add the error information. This trait is currently re-exported from our `utils` crate.

We can simplify matters by creating our own extension trait, and no longer re-exporting the otel one. This has the following advantages:

- We can specialize `set_status` to `set_error` (removes the need to re-export `Status`).
- We can customize the error reporting by doing something like `set_error(AsRef<dyn std::error::Error>)` and compiling a proper error report.
- We can make `set_attribute` more ergonomic by adding a secondary trait `ToValue` which we can then implement for foreign types. i.e. instead of `set_attribute("hash", digest.to_hex())`, we can just do `set_attribute("hash", &digest)`.

Extension trait should also be sealed, and only implemented for `tracing::Span`.

sergerad:
Closes #678

* Replaces re-export of `tracing_opentelemetry::OpenTelemetrySpanExt` with a custom trait of the same name
* Removes re-export of `opentelemetry::trace::Status` in favour of a trait API that takes `&dyn Error` instead
* Adds `ToValue` trait and impls for various Miden and core types


crates/utils/src/tracing/mod.rs
```diff
@@ -1,6 +1,27 @@
 pub mod grpc;
 
-// Re-export useful traits for open-telemetry traces. This avoids requiring other crates from
-// importing that family of crates directly.
-pub use opentelemetry::trace::Status as OtelStatus;
-pub use tracing_opentelemetry::OpenTelemetrySpanExt;
+use opentelemetry::trace::Status as OtelStatus;
+use opentelemetry::{Key, Value};
```
sergerad, line 4: @Mirko-von-Leipzig in the issue you mentioned:
```
We can make set_attribute more ergonomic by adding a secondary trait ToValue which we can then implement for foreign types. i.e. instead of set_attribute("hash", digest.to_hex()), we can just do set_attribute("hash", &digest).
```

I was wondering if this is worthwhile considering that crates can already `impl From<...> for opentelemetry::Value` for their types and rely on the trait as is `, value: impl Into<Value>);`.

WDYT?

Mirko-von-Leipzig: Most of our core types come from the crates from [`miden-base`](https://github.com/0xPolygonMiden/miden-base/) - and I don't think introducing `open-telemetry` there is a reasonable expectation. Though it could make sense if we really want to standardize on it everywhere.

Mirko-von-Leipzig: Actually here we go: https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/testing/trace/index.html

They do actually have test infra :)

crates/utils/src/tracing/mod.rs
```diff
@@ -1,6 +1,73 @@
...
+    fn to_value(&self) -> Value {
+        i64::from(*self).into()
+    }
+}
+
+impl ToValue for i64 {
+    fn to_value(&self) -> Value {
+        (*self).into()
+    }
+}
```
sergerad, line 48: @Mirko-von-Leipzig LMK impls to refactor or delete here. Also LMK where you would move / modularize any of this. I have just lumped it in mod.rs for now.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Add an optional open-telemetry trace exporter (#659, #690).
 - Support tracing across gRPC boundaries using remote tracing context (#669).
 - Instrument the block-producer's block building process (#676).
+- Add `ToValue` and custom extension trait for `tracing::Span` instead of re-exporting `tracing_opentelemetry::OpenTelemetrySpanExt` and `opentelemetry::trace::Status` (#700).
```
Mirko-von-Leipzig, line 14: I'd remove the changelog entry, this is more of an internal change.

Changelog should only be for user facing changes

crates/utils/src/tracing/mod.rs
```diff
@@ -1,6 +1,73 @@
...
+
+/// Utility functions based on [`tracing_opentelemetry::OpenTelemetrySpanExt`].
+#[sealed]
+pub trait OpenTelemetrySpanExt {
+    fn set_attribute(&self, key: impl Into<Key>, value: impl ToValue);
+    fn set_error(&self, err: &dyn std::error::Error);
+}
```
Mirko-von-Leipzig, line 55: I don't think we need a crate dependency for sealing. You can do sealing as follows:

```suggestion
pub trait OpenTelemetrySpanExt: Sealed {
    fn set_attribute(&self, key: impl Into<Key>, value: impl ToValue);
    fn set_error(&self, err: &dyn std::error::Error);
}

mod private {
    pub trait Sealed {}
}
```

Since we're only implementing the trait for on type (`tracing::Span`), the proc macro doesn't really buy us much.

crates/utils/src/tracing/mod.rs
Mirko-von-Leipzig: I think maybe move all this into a `span_ext.rs` module/file.

We could also consider doing a blanket implementation, although maybe being explicit is fine too.

```rust
impl<T> ToValue for T where T: Into<Value>
```

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -282,8 +282,8 @@ struct ProvenBlock {
 impl SelectedBlock {
     fn inject_telemetry(&self) {
         let span = Span::current();
-        span.set_attribute("block.number", i64::from(self.block_number.as_u32()));
-        span.set_attribute("block.batches.count", i64::from(self.batches.len() as u32));
+        span.set_attribute("block.number", self.block_number.as_u32());
```
Mirko-von-Leipzig, line 285:
```suggestion
        span.set_attribute("block.number", self.block_number);
```

crates/utils/src/tracing/mod.rs
```diff
@@ -1,6 +1,73 @@
...
+    /// Implementations for `ToValue` should be added to this crate (miden-node-utils).
+    fn set_attribute(&self, key: impl Into<Key>, value: impl ToValue) {
+        tracing_opentelemetry::OpenTelemetrySpanExt::set_attribute(self, key, value.to_value());
+    }
+
+    /// Sets a status on `Span` based on an error.
+    fn set_error(&self, err: &dyn std::error::Error) {
+        tracing_opentelemetry::OpenTelemetrySpanExt::set_status(
+            self,
+            OtelStatus::Error { description: format!("{err:?}").into() },
```
Mirko-von-Leipzig, line 70: I think we may want to make this an error report (some context [here](https://github.com/0xPolygonMiden/miden-node/pull/675#discussion_r1949479717)).

Something like:

```rust
let mut description = format!("{err}.");
let mut current = err.source();
while let Some(cause) = current {
    description.join(format!("Caused by: {cause}");
    current = cause.source();
}
```

sergerad:
> We could also consider doing a blanket implementation, although maybe being explicit is fine too.
> `impl<T> ToValue for T where T: Into<Value>`

I did try this at first but it causes implementation conflicts for the types that don't impl `Into<Value>` so don't think it works here.

crates/utils/src/tracing/mod.rs
sergerad: I did try this at first but it causes implementation conflicts for the types that don't impl Into<Value> so don't think it works here.

crates/utils/src/tracing/mod.rs
Mirko-von-Leipzig: Ah right; yeah this was intended for their 😅 Will cross-post just for traceability.

Issue #676, opened 2025-02-11
[Config] Remove `toml` support


Mirko-von-Leipzig:
As discussed in #650.

Additional considerations:

- Namespace/prefix the environment variables - this includes both to `miden` as well as `component`.
- Environment variables should consistent between `miden-node start node` and `miden-node start <component>`.

## Side refactoring

This is also a good opportunity to refactor the various component crates. They really only expose the component entry-point and nothing else. A simple way to ensure this is to keep everything in `lib.rs` private and only re-export the entry-point e.g. `pub use BlockProducer`.

At the moment things are a bit more complex because configuration also lives there. However with the proposed change all configuration should live in the node binary, which is also responsible for converting from config to component structs.

bobbinth:
> Namespace/prefix the environment variables - this includes both to `miden` as well as `component`.

Do you mean this should be something like `MIDEN_STORE_XXX` and `MIDEN_BLOCK_PRODUCER_YYY`?

Mirko-von-Leipzig:
> > Namespace/prefix the environment variables - this includes both to `miden` as well as `component`.
> 
> Do you mean this should be something like `MIDEN_STORE_XXX` and `MIDEN_BLOCK_PRODUCER_YYY`?

Yeah. Because we will up with multiple `XXX_ENDPOINT` for example. Tbh the `MIDEN_` prefix is unlikely to be useful but I do worry about collisions.

PR #701, opened 2025-02-17
feat: use remote batch prover

SantiagoPittella:
This PR replaces the local batch prover with the remote one.

It adds the `remote_batch_prover` endpoint to the configuration file, that will be used at node startup to create a `RemoteBatchProver`. 

At the moment, I removed the `miden-faucet` package from certain makefile commands because of an error regarding the mix of async and non async usage of `miden-base` dependencies.

The formatting of the `Cargo.toml` is a bit odd after adding some dependencies, but it is compliant with `taplo`. I'm investigating what can be causing the issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1706,7 +1706,7 @@ dependencies = [
 [[package]]
 name = "miden-block-prover"
 version = "0.8.0"
-source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#f2d50bfa4a83841875570d1301adccbe164ea111"
+source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#0aa58b4ad81d7970edcfd32a08eb90234294ecbf"
 dependencies = [
  "miden-crypto",
  "miden-lib",
@@ -1795,7 +1795,7 @@ dependencies = [
 [[package]]
 name = "miden-lib"
 version = "0.8.0"
-source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#f2d50bfa4a83841875570d1301adccbe164ea111"
+source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#0aa58b4ad81d7970edcfd32a08eb90234294ecbf"
 dependencies = [
  "miden-assembly",
  "miden-objects",
@@ -1829,7 +1829,7 @@ dependencies = [
  "supports-hyperlinks",
  "supports-unicode",
  "syn",
- "terminal_size",
+ "terminal_size 0.3.0",
  "textwrap",
  "thiserror 2.0.11",
  "trybuild",
@@ -1885,6 +1885,7 @@ dependencies = [
  "miden-node-utils",
  "miden-objects",
  "miden-processor",
+ "miden-proving-service-client",
  "miden-tx",
  "miden-tx-batch-prover",
  "pretty_assertions",
@@ -1991,7 +1992,7 @@ dependencies = [
 [[package]]
 name = "miden-objects"
 version = "0.8.0"
-source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#f2d50bfa4a83841875570d1301adccbe164ea111"
+source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#0aa58b4ad81d7970edcfd32a08eb90234294ecbf"
 dependencies = [
  "getrandom 0.2.15",
  "miden-assembly",
@@ -2035,6 +2036,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "miden-proving-service-client"
+version = "0.8.0"
+source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#0aa58b4ad81d7970edcfd32a08eb90234294ecbf"
+dependencies = [
+ "async-trait",
+ "miden-objects",
+ "miden-tx",
+ "miette",
+ "prost",
+ "prost-build",
+ "protox",
+ "thiserror 2.0.11",
+ "tokio",
+ "tonic",
+ "tonic-build",
+ "tonic-web-wasm-client",
+]
+
+[[package]]
 name = "miden-rpc-proto"
 version = "0.8.0"
 
@@ -2050,7 +2070,7 @@ dependencies = [
 [[package]]
 name = "miden-tx"
 version = "0.8.0"
-source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#f2d50bfa4a83841875570d1301adccbe164ea111"
+source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#0aa58b4ad81d7970edcfd32a08eb90234294ecbf"
 dependencies = [
  "async-trait",
  "miden-lib",
@@ -2067,7 +2087,7 @@ dependencies = [
 [[package]]
 name = "miden-tx-batch-prover"
 version = "0.8.0"
-source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#f2d50bfa4a83841875570d1301adccbe164ea111"
+source = "git+https://github.com/0xPolygonMiden/miden-base?branch=next#0aa58b4ad81d7970edcfd32a08eb90234294ecbf"
 dependencies = [
  "miden-core",
  "miden-crypto",
@@ -2096,8 +2116,16 @@ version = "7.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1a955165f87b37fd1862df2a59547ac542c77ef6d17c666f619d1ad22dd89484"
 dependencies = [
+ "backtrace",
+ "backtrace-ext",
  "cfg-if",
  "miette-derive",
+ "owo-colors",
+ "supports-color",
+ "supports-hyperlinks",
+ "supports-unicode",
+ "terminal_size 0.4.1",
+ "textwrap",
  "thiserror 1.0.69",
  "unicode-width 0.1.14",
 ]
@@ -3367,6 +3395,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "terminal_size"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5352447f921fda68cf61b4101566c0bdb5104eff6804d0678e5227580ab6a4e9"
+dependencies = [
+ "rustix",
+ "windows-sys 0.59.0",
+]
+
+[[package]]
 name = "textwrap"
 version = "0.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3639,6 +3677,31 @@ dependencies = [
  "tower-layer",
  "tower-service",
  "tracing",
+]
+
+[[package]]
+name = "tonic-web-wasm-client"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c191f6613df48874158b6af303313eadf25d1b7a534216b62a1f049d77cd2711"
+dependencies = [
+ "base64",
+ "byteorder",
+ "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "httparse",
+ "js-sys",
+ "pin-project",
+ "thiserror 1.0.69",
+ "tonic",
+ "tower-service",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "wasm-streams",
+ "web-sys",
 ]
 
 [[package]]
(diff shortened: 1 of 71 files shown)


.github/workflows/test.yml
```diff
@@ -25,6 +25,11 @@ jobs:
...
         with:
           save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
+      - run: make prover
+      - run: make start-prover > /dev/null &
       - uses: taiki-e/install-action@nextest 
       - name: Run tests
         run: make test
+      - name: Kill prover
+        if: always()
+        run: make kill-prover
```
Mirko-von-Leipzig, line 35: This will lead to race conditions or no?

If we're going with an out-of-process prover we would have to wait for it to acknowledge it is ready somehow.

I think we would much prefer an in-process prover. I imagine the issue is that the prover is currently a binary, and is not exposed as a library crate?

I was imagining starting it like any other component. For tests we could have a dedicated `BlockProducer::for_tests()` constructor that lazily starts a single prover worker and returns a client to it for example.

Cargo.lock
Mirko-von-Leipzig, line 1: Are the patch version bumps from the dependency resolver, or did you run `cargo update`?

.github/workflows/test.yml
Mirko-von-Leipzig: And this could be overridden by an environment variable, e.g. if `MIDEN_PROVER_SERVICE_URL` is set then it should use the given url for the client instead. This would let us still use an actual remote prover for example - though I'm unsure if we actually care about this at the moment.

Makefile
```diff
@@ -110,3 +115,32 @@ docker-run-node: ## Runs the Miden node as a Docker container
...
+build-prover: update-prover-branch ## Update dependencies and build the prover binary with specified features
+	cd $(PROVER_DIR) && cargo update && cargo build --bin miden-proving-service --locked $(PROVER_FEATURES_TESTING) --release
+
+.PHONY: start-prover
+start-prover: ## Run prover. This requires the base repo to be present at `miden-base`
+	cd $(PROVER_DIR) && RUST_LOG=info cargo run --bin miden-proving-service $(PROVER_FEATURES_TESTING) --release --locked -- start-worker -p $(PROVER_PORT) --batch-prover
+
+.PHONY: kill-prover
+kill-prover: ## Kill prover process
+	pkill miden-tx-prover || echo 'process not running'
```
Mirko-von-Leipzig, line 146: I'll be honest I'm not a big fan of this. I don't really want to do spooky things with other repos at random locations.

Once again, I would much prefer this service be in-process for convenience.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -275,9 +289,12 @@ impl WorkerPool {
         Span::current().record("batch_id", proposed_batch.id().to_string());
         info!(target: COMPONENT, "Proposed Batch built");
 
-        let proven_batch = LocalBatchProver::new(MIN_PROOF_SECURITY_LEVEL)
+        let proven_batch = remote_batch_prover
+            .lock()
+            .await
```
Mirko-von-Leipzig, line 294: Is this not a gRPC client? Should be cloneable without the extra mutex or no?

Mirko-von-Leipzig (commented):
The code seems good; I'm just unhappy with requiring an external service out-of-process by default. Feels pretty unergonomic

.github/workflows/test.yml
SantiagoPittella: I agree with you, this is not very good.

To expose the worker binary we might consider moving it to an external repository. We removed the library option due to conflicting features usage between the crates.

Another option can be having remote proving as optional and enabling local proving by default, though I think that @bobbinth prefer not to.

Makefile
SantiagoPittella: I don't like this neither. I would even prefer to use a deployed prover somewhere instead, though it would stall CI test jobs if it is down, so I don't think that this is a great idea.

Having the worker as a library and basically starting up as a component here looks like the best shot

.github/workflows/test.yml
bobbinth: I guess one options is to define a `BatchProver` struct (and in the future `BlockProver` as well). This struct could look something like this:
```rust
pub enum BatchProver {
    Local(LocalBatchProver),
    Remote(RemoteBatchProver),
}

impl BatchProver {

    pub fn new_local() -> Self {
        ...
    }

    pub fn new_remote(url: Url) -> Result<Self, BatchProverError> {
        ...
    }

    pub async fn prove(
        &self,
        proposed_batch: ProposedBatch,
    ) -> Result<ProvenBatch, BatchProverError> {
        ...
    }
}
```

By default, this struct would be instantiated with a local prover, but if configured (e.g., via env variables) it could be instantiated with a remote prover as well.

A couple of point:
- In real-world settings (e.g.,testnet, and probably for devnet as well), we will be using remote provers because prover doing proving in-process will likely slow the node down to a crawl.
- Once we have real proving, even to testing we'll need to be very careful with how we run tests which require proving. For example, doing proving with `debug` build or with `debug-assertions` enabled result in very long-running tests. So, we'd either need to figure out how to do "mock proving" or make sure we run such tests with `release` code and no `debug-assertions`.

Mirko-von-Leipzig: > - Once we have real proving, even to testing we'll need to be very careful with how we run tests which require proving. For example, doing proving with debug build or with debug-assertions enabled result in very long-running tests. So, we'd either need to figure out how to do "mock proving" or make sure we run such tests with release code and no debug-assertions.

Its also possible to optimise certain dependencies using [profiles](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides). So we could have that always set just for the prover implementations (if they're separate). 

Alternatively, we could use snapshots for testing which also makes good sense to me given we're likely reaching out over the "network" in any case. If we go for this then the prover _could_ always be remote-looking for tests - we just need a real prover at some point to generate the correct responses. This works for tests, but doesn't help much for simplifying local node operation.

SantiagoPittella: Since everything was in place for, I started with the [remote block prover](https://github.com/0xPolygonMiden/miden-base/pull/1169) in `miden-base`.

I'm also researching about how to implement the snapshots approach.

bobbinth: > I'm also researching about how to implement the snapshots approach.

For now, I'd probably go with Local/Remote enum approach and we can do the snapshot approach later on.

Mirko-von-Leipzig: For snapshots I would recommend something like `insta`.

crates/block-producer/src/server.rs
```diff
@@ -90,6 +97,7 @@ impl BlockProducer {
             store,
             rpc_listener,
             chain_tip,
+            remote_batch_prover,
```
bobbinth, line 100: I wonder whether instead of making `remote_batch_prover` a field in `BlockProducer`, we should just pass it to the `BatchBuilder`. Specifically, we could have `BatchBuilder::new()` constructor which looks something like this:
```rust
impl BatchBuilder {
    pub fn new(prover_url: Option<Url>) -> Self {
        ...
    }
}
```
Then, how to construct the prover would be the internal concern of the `BatchBuilder`, and here, we'd just do something like:
```rust
BatchBuilder::new(config.batch_prover_url)
```

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -54,7 +52,12 @@ impl BatchBuilder {
     /// A pool of batch-proving workers is spawned, which are fed new batch jobs periodically.
     /// A batch is skipped if there are no available workers, or if there are no transactions
     /// available to batch.
-    pub async fn run(self, mempool: SharedMempool, store: StoreClient) {
+    pub async fn run(
+        self,
+        mempool: SharedMempool,
+        store: StoreClient,
+        batch_prover: BatchProver,
+    ) {
```
bobbinth, line 60: Related to the previous comment, if we make `batch_prover` a field of `BatchBuilder`, we won't need to pass it here.

Also, not from this PR, but we have a slight inconsistency here between batch and block builders: `BlockBuilder` takes the store as a construction parameter, but `BatchBuilder` takes it via the `run` method. Should we also move store into the constructor for `BatchBuilder`? @Mirko-von-Leipzig - what do you think?

bobbinth (commented):
Looks good! Thank you. Not a full review, but I left some comments inline.

crates/block-producer/src/config.rs
```diff
@@ -23,13 +25,20 @@ pub struct BlockProducerConfig {
     /// verification may take ~15ms/proof. This is OK when all transactions are forwarded to the
     /// block producer from the RPC component as transaction proofs are also verified there.
     pub verify_tx_proofs: bool,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: #718

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: I've been meaning to relook at these as part of #676.

Its also semi-related to properly decoupling components, i.e. a block-producer should be able to startup independently from the store.

I think we can align them now; but we should take a closer look in the medium term. Bit of a non-answer I know 🙃

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -275,13 +289,48 @@ impl WorkerPool {
...
+impl BatchProver {
+    pub fn new_local(security_level: u32) -> Self {
+        warn!(target: COMPONENT, "Using local batch prover");
+        Self::Local(LocalBatchProver::new(security_level))
+    }
+
+    pub fn new_remote(endpoint: impl Into<String>) -> Self {
+        warn!(target: COMPONENT, "Using remote batch prover");
+        Self::Remote(RemoteBatchProver::new(endpoint))
+    }
```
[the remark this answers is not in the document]
SantiagoPittella: I was using it with debug purposes (I using Ghostty and not having search scrollback is not playing out nice), to distinguish these logs easily. I'm replacing it with `info`.

crates/utils/src/config.rs
```diff
@@ -10,6 +10,7 @@ pub const DEFAULT_NODE_RPC_PORT: u16 = 57291;
 pub const DEFAULT_BLOCK_PRODUCER_PORT: u16 = 48046;
 pub const DEFAULT_STORE_PORT: u16 = 28943;
 pub const DEFAULT_FAUCET_SERVER_PORT: u16 = 8080;
+pub const DEFAULT_BATCH_PROVER_PORT: u16 = 8082;
```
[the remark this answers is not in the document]
SantiagoPittella: I was using it to match the value that i was using in the `Makefile` target that was initializing the worker. Since now we have a local prover coexisting with the remote, this default value is not needed anymore since we are defaulting to local. I'm removing the constant.

crates/block-producer/Cargo.toml
SantiagoPittella, line 1: Not loving how I break the formatting, but since I introduced a multi line dependency taplo is formatting this way.

Mirko-von-Leipzig:
@SantiagoPittella I think you need to rebase to `next` because that PR was merged.

crates/block-producer/Cargo.toml
Mirko-von-Leipzig: You could try fiddle with some of these: https://taplo.tamasfe.dev/configuration/formatter-options.html.

I imagine it would look nicer as a field per line once it exceeds the column limit.

But in the end if nothing looks good then it is what it is :) At least its consistent.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -33,22 +35,33 @@ pub struct BatchBuilder {
     ///
     /// Note: this _must_ be sign positive and less than 1.0.
     pub failure_rate: f32,
+    /// The batch prover to use.
+    ///
+    /// If not provided, a local batch prover is used.
+    batch_prover: BatchProver,
```
Mirko-von-Leipzig, line 41: The optional conversion happens in `new` so lets document it there only.

```suggestion
    /// The batch prover to use.
    batch_prover: BatchProver,
```

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -33,22 +35,33 @@ pub struct BatchBuilder {
...
 }
 
-impl Default for BatchBuilder {
-    fn default() -> Self {
+impl BatchBuilder {
+    /// Creates a new [`BatchBuilder`] with the given batch prover URL.
+    ///
+    /// If no URL is provided, a local batch prover is used.
+    pub fn new(batch_prover_url: Option<Url>) -> Self {
```
Mirko-von-Leipzig, line 48:
```suggestion
    /// Creates a new [`BatchBuilder`] with the given batch prover URL.
    ///
    /// Defaults to [`BatchProver::Local`] is no URL is provided.
    pub fn new(batch_prover_url: Option<Url>) -> Self {
```

Mirko-von-Leipzig (approved):
Some minor doc nits, but LGTM

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -275,13 +297,48 @@ impl WorkerPool {
...
+impl BatchProver {
+    pub fn new_local(security_level: u32) -> Self {
+        info!(target: COMPONENT, "Using local batch prover");
+        Self::Local(LocalBatchProver::new(security_level))
+    }
+
+    pub fn new_remote(endpoint: impl Into<String>) -> Self {
+        info!(target: COMPONENT, "Using remote batch prover");
+        Self::Remote(RemoteBatchProver::new(endpoint))
+    }
```
bobbinth, line 328: Not for this PR, but here we'll also need to provide required security level and then check if the proofs returned from the remote prover match the expected security level.

crates/block-producer/Cargo.toml
bobbinth, line 1: Somehow this file didn't get formatted properly. But we can fix this in a subsequent PR.

bobbinth (approved):
Looks good! Thank you! I left a couple of comments inline, but these can be addressed in subsequent PRs.

PR #702, opened 2025-02-18
feat: test faucet website

### LINKED ISSUE #673: Faucet website testing
There have been several faucet website bugs recently (caused by me) e.g. #671, where various parts of the website aren't served.

It would be nice to test that the website loads correctly. At the moment this isn't trivially possible without actually running the full faucet, which in turn requires a full node.

Ideally the test would be self-contained, i.e.:

- serve the website (using a single method to add the route(s), which is then also used in prod)
- load the website and ensure it loads without errors

---------------------

A better alternative might be templating/DSL #652, which could ensure at compile time that all files/data is present and bundled.

TomasArrachea:
Closes #673 

This PR adds a test for the faucet website. It uses `fantoccini` as a browser client to test that all components are loading correctly.

To run `fantoccini` it needs a webdriver running. I'm using chromedriver here, which requires to have Chrome installed.

Also, this PR introduces a stub for the node RPC API. This allows to run the faucet and test it using the stub, instead of having to run the full node.


bin/faucet/src/main.rs
```diff
@@ -229,3 +196,133 @@ fn long_version() -> LongVersion {
...
+        let website_url = config.endpoint.clone();
+        tokio::spawn(async move { serve_faucet(config).await.unwrap() });
+
+        // Start chromedriver. This requires having chromedriver and chrome installed
+        let chromedriver_port = "57709";
+        #[allow(clippy::zombie_processes)]
+        let mut chromedriver = Command::new("chromedriver")
+            .arg(format!("--port={chromedriver_port}"))
+            .spawn()
+            .expect("failed to start chromedriver");
```
TomasArrachea, line 284: I had to install Chromedriver and Chrome to run this locally, but I am not sure on why it's running fine in the CI

Mirko-von-Leipzig: The github CI images have a bunch of pre-installed stuff (I believe the image is like 50GB). This means less cpu wasted installing things etc. And given the amount of webdev, probably almost any semi-popular tooling for it will be pre-installed.

bin/faucet/src/stub_rpc_api.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    async fn check_nullifiers_by_prefix(
+        &self,
+        _request: Request<CheckNullifiersByPrefixRequest>,
+    ) -> Result<Response<CheckNullifiersByPrefixResponse>, Status> {
+        unimplemented!();
+    }
+
+    #[allow(clippy::unreadable_literal)]
```
Mirko-von-Leipzig, line 43:
```suggestion
```
should be able to remove this one by adding `_` to the values. Fairly sure rust-analyzer will do this for you. I'd also recommend using hex values instead `0xABC` (also something rust-analyzer can do).

bin/faucet/src/stub_rpc_api.rs
```diff
@@ -0,0 +1,166 @@
...
+                    d2: 9233115969432897632,
+                    d3: 1437907447409278328,
+                }),
+                tx_hash: Some(Digest { d0: 0, d1: 0, d2: 0, d3: 0 }),
+                kernel_root: Some(Digest {
+                    d0: 8894402440595556547,
+                    d1: 11075240337243789177,
+                    d2: 12654662110212372673,
+                    d3: 12816653122390928829,
+                }),
```
Mirko-von-Leipzig, line 83: Where are these values from?

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,133 @@ fn long_version() -> LongVersion {
...
+    let socket_addr =
+        config
+            .endpoint
+            .socket_addrs(|| None)?
+            .into_iter()
+            .next()
+            .ok_or(anyhow::anyhow!(
+                "Couldn't get any socket addrs for endpoint: {}",
+                config.endpoint
+            ))?;
```
Mirko-von-Leipzig, line 241:
```suggestion
            .with_context(|| format!("no sockets available on {}, config.endpoint))?;
```

Mirko-von-Leipzig: Though I'm unsure under what circumstances that would even occur.

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,133 @@ fn long_version() -> LongVersion {
         debug: option_env!("VERGEN_CARGO_DEBUG").unwrap_or_default(),
     }
 }
+
+fn create_new_router(faucet_state: FaucetState) -> Router {
```
Mirko-von-Leipzig, line 200:
```suggestion
fn create_router(faucet_state: FaucetState) -> Router {
```

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,133 @@ fn long_version() -> LongVersion {
...
+
+    use fantoccini::ClientBuilder;
+    use serde_json::{json, Map};
+    use tokio::time::sleep;
+    use url::Url;
+
+    use crate::{config::FaucetConfig, serve_faucet, stub_rpc_api::serve_stub};
+
+    #[tokio::test]
+    async fn test_website() {
```
Mirko-von-Leipzig, line 262: Add a doc comment that explains what this test is, and what it does. Its fairly unusual as far as tests go.

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,133 @@ fn long_version() -> LongVersion {
...
+        let stub_node_url = Url::from_str("http://localhost:50051").unwrap();
+
+        // Start the stub node
+        let stub_url_clone = stub_node_url.clone();
+        tokio::spawn(async move { serve_stub(&stub_url_clone).await.unwrap() });
```
Mirko-von-Leipzig, line 267: This avoids the awkward `_clone` name at least.
```suggestion
        tokio::spawn({
            let stub_node_url = stub_node_url.clone();
            async move { serve_stub(&stub_url_clone).await.unwrap() }
        });
```

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,133 @@ fn long_version() -> LongVersion {
...
+            node_url: stub_node_url,
+            faucet_account_path: "src/test_data/faucet.mac".into(),
+            ..FaucetConfig::default()
+        };
+        let website_url = config.endpoint.clone();
+        tokio::spawn(async move { serve_faucet(config).await.unwrap() });
+
+        // Start chromedriver. This requires having chromedriver and chrome installed
+        let chromedriver_port = "57709";
+        #[allow(clippy::zombie_processes)]
```
Mirko-von-Leipzig, line 280:
```suggestion
        #[expect(clippy::zombie_processes)]
```

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,133 @@ fn long_version() -> LongVersion {
...
+        sleep(Duration::from_secs(1)).await;
+
+        // Start fantoccini client
+        let mut caps = Map::new();
+        caps.insert(
+            "goog:chromeOptions".to_string(),
+            json!({"args": ["--headless", "--disable-gpu", "--no-sandbox"]}),
+        );
+        let client = ClientBuilder::native()
+            .capabilities(caps)
```
Mirko-von-Leipzig, line 296:
```suggestion
        let client = ClientBuilder::native()
            .capabilities([(
                "goog:chromeOptions".to_string(),
                json!({"args": ["--headless", "--disable-gpu", "--no-sandbox"]})
            )].into()),
```
Might be able to do something like this - and do they really abbreviate it to `goog:`?

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,133 @@ fn long_version() -> LongVersion {
...
+
+        for request in requests.as_array().unwrap() {
+            let uri = request.as_str().unwrap();
+            let status = reqwest::get(uri).await.unwrap().status();
+            assert_eq!(status, 200);
+        }
+
+        // Close the client and kill chromedriver
+        client.close().await.unwrap();
+        chromedriver.kill().unwrap();
```
Mirko-von-Leipzig, line 326: Is there no safe way to end this process?

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,133 @@ fn long_version() -> LongVersion {
...
+            return requests;
+        ";
+        let requests = client.execute(script, vec![]).await.unwrap();
+        assert!(!requests.as_array().unwrap().is_empty());
+
+        for request in requests.as_array().unwrap() {
+            let uri = request.as_str().unwrap();
+            let status = reqwest::get(uri).await.unwrap().status();
+            assert_eq!(status, 200);
+        }
```
Mirko-von-Leipzig, line 322: Do we need to repeat the requests? Is it not possible to tell if one failed? As in - if we remove the `.css` route, can we not tell without manually requesting again?

Mirko-von-Leipzig (commented):
Looks good & interesting (not something I'm familiar with), thank you.

I've left some initial questions and suggestions; I think we may also want to organise the test files a bit somehow but we can address that after everything else.

Right now this tests that the website loads and has the correct title. Would we then extend this with requesting funds from the faucet as well? (separate PR etc maybe).

bin/faucet/src/main.rs
SantiagoPittella: Should we create a target in the makefile to install Chromedriver and Chrome?

Mirko-von-Leipzig: I don't think you can unless you assume the OS? Its a pity the dep doesn't have a bundling option, makes this sort of thing much easier.

bin/faucet/src/stub_rpc_api.rs
TomasArrachea: I ran the faucet with a full node and just copied the response of the first block

bin/faucet/src/main.rs
TomasArrachea: I actually haven't found any other way of terminating it

bin/faucet/src/main.rs
TomasArrachea: I was able to do this changing the js script in [e39e1f9](https://github.com/0xPolygonMiden/miden-node/pull/702/commits/e39e1f9f59a5d5f1f6e01eab615f4456cf3f86cd)

bin/faucet/src/stub_rpc_api.rs
Mirko-von-Leipzig: Could you add a comment to the code to that affect, e.g. 

```
// Values are taken from the default genesis block as at v0.7
```

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,131 @@ fn long_version() -> LongVersion {
...
+        // Start chromedriver. This requires having chromedriver and chrome installed
+        let chromedriver_port = "57709";
+        #[expect(clippy::zombie_processes)]
+        let mut chromedriver = Command::new("chromedriver")
+            .arg(format!("--port={chromedriver_port}"))
+            .spawn()
+            .expect("failed to start chromedriver");
+
+        // Wait for the server to be running
+        sleep(Duration::from_secs(1)).await;
```
igamigo, line 287: Is this consistent? I guess 1 second doesn't hurt but otherwise maybe there could be some loop that retries connections a couple of times or something to make it a bit more robust.

igamigo (approved):
LGTM! I am not too familiar with the setup but it seems great for actually testing the page within a browser context as opposed to something like running a get request through `reqwest` and then retrieving further resources, etc.

bin/faucet/src/main.rs
TomasArrachea: turns out this sleep is no longer needed, so I just deleted it

TomasArrachea: Removed this by reading stdout to wait until chromedriver is running: [102b761](https://github.com/0xPolygonMiden/miden-node/pull/702/commits/102b761cb176b39fa549d71185b918f45575b499)

bin/faucet/src/main.rs
```diff
@@ -229,3 +196,139 @@ fn long_version() -> LongVersion {
...
+
+        // Start the stub node
+        tokio::spawn({
+            let stub_node_url = stub_node_url.clone();
+            async move { serve_stub(&stub_node_url).await.unwrap() }
+        });
+        // Start the faucet connected to the stub
+        let config = FaucetConfig {
+            node_url: stub_node_url,
+            faucet_account_path: "test_data/faucet.mac".into(),
```
Mirko-von-Leipzig, line 275: Would it be possible to generate this as part of the test instead?

I mainly ask because I dislike random binary files floating around the codebase.

This does bring up a slight refactoring though - if we factor out the `match &cli.command` from `main.rs` into a separate function, then we could:

- Revert the `serve_faucet` changes (move them back to `Command::start`. This isn't overly important, just a smaller diff.
- Run the full faucet init, account creation and start commands in order.

This would also require using `tempfs` to host the files.

I rather like this idea. Though maybe it would be best to merge this PR and upgrade it in another PR.

@TomasArrachea WDYT? cc @igamigo

Mirko-von-Leipzig (approved):
Approving, code LGTM. Merging will depend on whether we want to upgrade the test in this PR, or in a follow-up.

tomyrd (approved):
LGTM, tested locally and it worked (after installing Chrome and chromedriver)

bin/faucet/src/main.rs
```diff
@@ -229,3 +238,122 @@ fn long_version() -> LongVersion {
...
+        let stub_node_url = Url::from_str("http://localhost:50051").unwrap();
+
+        // Start the stub node
+        tokio::spawn({
+            let stub_node_url = stub_node_url.clone();
+            async move { serve_stub(&stub_node_url).await.unwrap() }
+        });
+
+        let config_path = temp_dir().join(PathBuf::from("faucet.toml"));
+        let faucet_account_path = temp_dir().join(PathBuf::from("account.mac"));
```
Mirko-von-Leipzig, line 272: From memory `join` takes `Into<Path>` or no?

```suggestion
        let config_path = temp_dir().join("faucet.toml");
        let faucet_account_path = temp_dir().join("account.mac");
```

Mirko-von-Leipzig (approved):

PR #705, opened 2025-02-19
fix(store): blocking in async fn `get_block_inputs`

### LINKED ISSUE #703: [Bug]: Node's store is crashing when running integration tests
### Version

next (commit d57c8dd4df7a)

### Other packages versions

miden-client: next (commit 7988cb26a5b0)

### What happened?

When running the `miden-client` integration tests in `next` they get stuck because [they are waiting for a transaction to get committed](https://github.com/0xPolygonMiden/miden-client/blob/7988cb26a5b0c0baf01558af1fae23260ad17094/tests/integration/common.rs#L191-L211) and it never does.

The tests should have finished successfully. In the previous `next` commit (2a10b47f5a2e) they worked correctly.

1. Clone the `miden-client` repo and checkout to the `next` branch.
2. Run `make integration-tests` with a localhost instance of the node running.

### Relevant log output

```shell
thread 'tokio-runtime-worker' panicked at crates/store/src/state.rs:641:42:
Cannot block the current thread from within a runtime. This happens because a function attempted to block the current thread while the thread is being used to drive asynchronous tasks.
2025-02-18T18:40:22.076463Z  INFO store.rpc/GetBatchInputs:store.server.get_batch_inputs: miden-store: crates/store/src/server/api.rs:368: close, time.busy: 161µs, time.idle: 138µs
2025-02-18T18:40:22.076524Z  INFO store.rpc/GetBatchInputs: miden_node_utils::tracing::grpc: crates/utils/src/tracing/grpc.rs:36: close, time.busy: 254µs, time.idle: 165µs
2025-02-18T18:40:22.076774Z ERROR store.client.get_batch_inputs: miden-block-producer: crates/block-producer/src/store/mod.rs:214: error: gRPC client error
2025-02-18T18:40:22.076852Z  INFO store.client.get_batch_inputs: miden-block-producer: crates/block-producer/src/store/mod.rs:214: close, time.busy: 165µs, time.idle: 880µs
2025-02-18T18:40:22.076874Z  INFO mempool.lock: miden-block-producer: crates/block-producer/src/mempool/mod.rs:134: new
2025-02-18T18:40:22.076883Z  INFO mempool.lock: miden-block-producer: crates/block-producer/src/mempool/mod.rs:134: close, time.busy: 1.21µs, time.idle: 8.71µs
2025-02-18T18:40:22.076889Z  WARN miden_node_block_producer::batch_builder: crates/block-producer/src/batch_builder/mod.rs:91: Batch job failed., batch_id: 0x129a345f108a003656acbe7dfd6ad7f5390be052d503d95a3bab8054895c0470, err: failed to fetch batch inputs from store
```

Mirko-von-Leipzig:
Bug was introduced in #668.

We are missing store tests.

Fixes #703.

### DIFF
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -197,11 +197,7 @@ struct InnerState {
 impl InnerState {
     /// Returns the latest block number.
     fn latest_block_num(&self) -> BlockNumber {
-        let block_number: u32 = (self.blockchain.chain_tip().as_usize() - 1)
-            .try_into()
-            .expect("chain_mmr always has, at least, the genesis block");
-
-        block_number.into()
+        self.blockchain.chain_tip()
     }
 }
 
@@ -638,7 +634,7 @@ impl State {
         // Scoped block to automatically drop the read lock guard as soon as we're done.
         // We also avoid accessing the db in the block as this would delay dropping the guard.
         let (batch_reference_block, partial_mmr) = {
-            let inner_state = self.inner.blocking_read();
+            let inner_state = self.inner.read().await;
 
             let latest_block_num = inner_state.blockchain.chain_tip();
 


crates/store/src/state.rs
```diff
@@ -197,11 +197,7 @@ struct InnerState {
 impl InnerState {
     /// Returns the latest block number.
     fn latest_block_num(&self) -> BlockNumber {
-        let block_number: u32 = (self.blockchain.chain_tip().as_usize() - 1)
-            .try_into()
-            .expect("chain_mmr always has, at least, the genesis block");
-
-        block_number.into()
+        self.blockchain.chain_tip()
```
Mirko-von-Leipzig, line 200: I believe this is also a bugfix.

These issues creep in due to the `forest` / `chain_tip` offset by one which I don't actually fully understand.

PhilippGackstatter (approved):
Looks good, thanks for the fix!

crates/store/src/state.rs
Mirko-von-Leipzig: If someone could explain to me why `forest` is offset, and for example why https://github.com/0xPolygonMiden/miden-node/pull/705/files#diff-339991026f9c6728dbde8578e8d896f0c44e4234a63e25113190fd1701d343b5R803-R811 isn't simply the latest peaks? Its extremely confusing that things are offset.

Mirko-von-Leipzig: @PhilippGackstatter, @bobbinth some insight maybe? 😓

PhilippGackstatter: Yeah, the chain MMR also confused me for a long time and I still find it confusing, with the `peaks_at` and all. 

Basically, if we're currently building block 5, then it references the previous block 4. And the state of the chain MMR at block 4, is all blocks up to, but excluding that block, so the blocks in range `0..=3`. This is so because block 5 can't add itself to the chain MMR, because it can't compute its own block hash yet, because for that, the chain MMR root is required, so there's a circular dependency. So the chain MMR therefore "lags" behind by one additional block relative to the other data structures like account or nullifier tree.
So while building block 5 we're only adding block 4 to the chain MMR, and block 6 will add block 5, etc.

To take your example for block inputs: If we fetch block inputs for block 5, then we need the peaks at the latest block (block 4), where forest is also 4 (i.e. the chain is of length 4 and the latest block is 3). That's why we're using `forest = latest_block_num`.

Does that help?

Mirko-von-Leipzig: That helps with half of the puzzle. I don't understand what `forest` actually means as an index. A forest in the colloquial sense is a collection of trees. Is `forest: usize` the `i'th` tree? I don't know what it means 😓 

Quoting from the various bits of documentation in the inner `Mmr`:

```rust
    /// Returns the peaks of the MMR at the state specified by `forest`.
    ///
    /// # Errors
    /// Returns an error if the specified `forest` value is not valid for this MMR.
    pub fn peaks_at(&self, forest: usize) -> Result<MmrPeaks, MmrError> {
```
doesn't really help, okay let's checkout the definition:
```rust
pub struct Mmr {
    /// Refer to the `forest` method documentation for details of the semantics of this value.
    pub(super) forest: usize,
```
okay another layer, let's check out the method:
```rust
    /// Returns the MMR forest representation.
    ///
    /// The forest value has the following interpretations:
    /// - its value is the number of elements in the forest
    /// - bit count corresponds to the number of trees in the forest
    /// - each true bit position determines the depth of a tree in the forest
    pub const fn forest(&self) -> usize {
```
Does number of elements mean the number of leaves? If so, how does this now relate back to `peaks_at(forest: usize)` -- since that's multiple peaks aka multiple leaves?

Mirko-von-Leipzig: I think partially its also confusing because we're mixing metaphors; we have mountains and peaks, along with trees, forests and leaves.

Is it more along the lines of: give me the peaks of the mountain range as it was back when it contained `N` elements?

bobbinth: Ah - good point! Some of these are equivalent (at least in my brain :) ):

- `peak` == `root`
- `mountain` = `tree`
- `range` = `forest`

So, MMR with 2 peaks contains 2 trees :)

> Is it more along the lines of: give me the peaks of the mountain range as it was back when it contained `N` elements?

Yes, that's exactly it. Assuming our chain has 5 blocks:

- `peaks_at(5)` gives us the current peaks (i.e., the 2 peaks representing trees of depth 0 and 2).
- `peaks_at(4)` gives us the peaks when the chain had 4 blocks (this would be a single peak for a tree of depth 2).
- `peaks_at(3)` gives us the peaks when the chain had 3 blocks (this would be two peaks for trees of depth 0 and 1).

Mirko-von-Leipzig: > Ah - good point! Some of these are equivalent (at least in my brain :) ):
>
> peak == root
> mountain = tree
> range = forest
> So, MMR with 2 peaks contains 2 trees :)

That's also my model; but also why I found it confusing to ask to ask the forest (aka MMR aka range) for the peaks at forest.

Alright, so we're getting the peaks to prove the next block. However the chain MMR lags behind by one (for circular reasons), and therefore we're off by one. Though somehow we already have an additional element in there (or otherwise we could just ask for latest).

PhilippGackstatter: > Though somehow we already have an additional element in there (or otherwise we could just ask for latest).

We could ask for the latest in the block inputs here, but I think its done that way for explicitness reasons and I would probably keep it that way. Just to express that we fetch the latest block header from the DB and then get the chain MMR for exactly that block. If we called `peaks()` we'd get the same, because we're holding on to the `RwLockReadGuard`, so nothing can change under our feet, but that's a bit too subtle here to change it I think.

A better comment explaining it would help though.

PR #706, opened 2025-02-19
feat(store): track network notes

### LINKED ISSUE #696: Track network notes in store
Network notes are identifiable by their tag.

Ensure that the store can query for network notes that have _not yet_ been consumed.

Mirko-von-Leipzig:
Adds `execution_mode` and `consumed` columns to the `notes` table.

Both are written when notes are inserted; however the `consumed` column is currently never set to `true` as we don't have that information available when applying a block (yet).

I've also added a pagination method to iterate over unconsumed network notes. This is intended to initialize the network transaction builder, probably via a streaming gRPC method. The implementation is copied from another note selection query so it should be correct.

Closes #696

### DIFF
--- a/crates/store/src/db/migrations/001-init.sql
+++ b/crates/store/src/db/migrations/001-init.sql
@@ -35,6 +35,7 @@ CREATE TABLE
     execution_hint INTEGER NOT NULL,
     merkle_path    BLOB    NOT NULL,
     consumed       INTEGER NOT NULL, -- boolean
+    nullifier      BLOB,             -- Only known for public notes, null for private notes
     details        BLOB,
 
     PRIMARY KEY (block_num, batch_index, note_index),
--- a/crates/store/src/db/migrations/001-init.sql
+++ b/crates/store/src/db/migrations/001-init.sql
@@ -149,4 +149,4 @@ CREATE TABLE
 
 CREATE INDEX idx_transactions_account_id ON transactions(account_id);
 CREATE INDEX idx_transactions_block_num ON transactions(block_num);
-CREATE INDEX note_execution_mode ON notes(execution_mode);
+CREATE INDEX unconsumed_network_notes ON notes(execution_mode, consumed);
--- a/crates/store/src/db/migrations/001-init.sql
+++ b/crates/store/src/db/migrations/001-init.sql
@@ -28,16 +28,20 @@ CREATE TABLE
     note_index     INTEGER NOT NULL, -- Index of note in batch, starting from 0
     note_id        BLOB    NOT NULL,
     note_type      INTEGER NOT NULL, -- 1-Public (0b01), 2-Private (0b10), 3-Encrypted (0b11)
-    sender         BLOB NOT NULL,
+    sender         BLOB    NOT NULL,
     tag            INTEGER NOT NULL,
+    execution_mode INTEGER NOT NULL, -- 0-Network, 1-Local
     aux            INTEGER NOT NULL,
     execution_hint INTEGER NOT NULL,
     merkle_path    BLOB    NOT NULL,
+    consumed       INTEGER NOT NULL, -- boolean
     details        BLOB,
 
     PRIMARY KEY (block_num, batch_index, note_index),
     FOREIGN KEY (block_num) REFERENCES block_headers(block_num),
     CONSTRAINT notes_type_in_enum CHECK (note_type BETWEEN 1 AND 3),
+    CONSTRAINT notes_execution_mode_in_enum CHECK (execution_mode BETWEEN 0 AND 1),
+    CONSTRAINT notes_consumed_is_bool CHECK (execution_mode BETWEEN 0 AND 1),
     CONSTRAINT notes_block_num_is_u32 CHECK (block_num BETWEEN 0 AND 0xFFFFFFFF),
     CONSTRAINT notes_batch_index_is_u32 CHECK (batch_index BETWEEN 0 AND 0xFFFFFFFF),
     CONSTRAINT notes_note_index_is_u32 CHECK (note_index BETWEEN 0 AND 0xFFFFFFFF)
@@ -145,3 +149,4 @@ CREATE TABLE
 
 CREATE INDEX idx_transactions_account_id ON transactions(account_id);
 CREATE INDEX idx_transactions_block_num ON transactions(block_num);
+CREATE INDEX note_execution_mode ON notes(execution_mode);
--- a/crates/store/src/db/migrations/001-init.sql
+++ b/crates/store/src/db/migrations/001-init.sql
@@ -46,7 +46,7 @@ CREATE TABLE
     CONSTRAINT notes_block_num_is_u32 CHECK (block_num BETWEEN 0 AND 0xFFFFFFFF),
     CONSTRAINT notes_batch_index_is_u32 CHECK (batch_index BETWEEN 0 AND 0xFFFFFFFF),
     CONSTRAINT notes_note_index_is_u32 CHECK (note_index BETWEEN 0 AND 0xFFFFFFFF)
-) STRICT, WITHOUT ROWID;
+) STRICT;
 
 CREATE TABLE
     accounts
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -64,7 +64,6 @@ pub struct NoteRecord {
     pub note_id: RpoDigest,
     pub metadata: NoteMetadata,
     pub details: Option<Vec<u8>>,
-    pub nullifier: Option<Nullifier>,
     pub merkle_path: MerklePath,
 }
 
@@ -81,8 +80,7 @@ impl NoteRecord {
             aux,
             execution_hint,
             merkle_path,
-            details,
-            nullifier
+            details
     ";
 
     /// Parses a row from the `notes` table. The sql selection must use [`Self::SELECT_COLUMNS`] to
@@ -102,11 +100,6 @@ impl NoteRecord {
         let merkle_path = MerklePath::read_from_bytes(merkle_path_data)?;
         let details_data = row.get_ref(10)?.as_blob_or_null()?;
         let details = details_data.map(<Vec<u8>>::read_from_bytes).transpose()?;
-        let nullifier = row
-            .get_ref(11)?
-            .as_blob_or_null()?
-            .map(Nullifier::read_from_bytes)
-            .transpose()?;
 
         let metadata =
             NoteMetadata::new(sender, note_type, tag.into(), execution_hint.try_into()?, aux)?;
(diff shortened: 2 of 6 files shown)


crates/store/src/db/migrations/001-init.sql
```diff
@@ -145,3 +149,4 @@ CREATE TABLE
 
 CREATE INDEX idx_transactions_account_id ON transactions(account_id);
 CREATE INDEX idx_transactions_block_num ON transactions(block_num);
+CREATE INDEX unconsumed_network_notes ON notes(execution_mode, consumed);
```
Mirko-von-Leipzig, line 152: I believe this is correct; however we should check the query plan at some stage to see if its using this index at all.

bobbinth:
> However if we want the option to prune consumed notes in the future, then we may as well keep `execution_mode`.

Yeah - let's have both `nullifier` and `execution_mode` in the table.

crates/store/src/db/sql/mod.rs
```diff
@@ -1002,6 +1029,91 @@ pub fn select_note_inclusion_proofs(
...
+        let details_data = row.get_ref(10)?.as_blob_or_null()?;
+        let details = details_data.map(<Vec<u8>>::read_from_bytes).transpose()?;
+        let nullifier = row
+            .get_ref(11)?
+            .as_blob_or_null()?
+            .map(Nullifier::read_from_bytes)
+            .transpose()?;
+
+        let metadata =
+            NoteMetadata::new(sender, note_type, tag.into(), execution_hint.try_into()?, aux)?;
```
bobbinth, line 1101: This code seems to be identical to the code on lines 758 - 779 and 943 - 964 above. Should we create a helper function which encapsulates it? Or maybe have something like `NoteRecord::from_row()` function?

Mirko-von-Leipzig: The concern I have is that it depends on the sql query ordering. So while they look the same now, thats just a consequence of them having the same field ordering in the string.

crates/store/src/db/tests.rs
```diff
@@ -187,6 +187,7 @@ fn sql_select_notes() {
             .unwrap(),
             details: Some(vec![1, 2, 3]),
             merkle_path: MerklePath::new(vec![]),
+            nullifier: None,
```
Mirko-von-Leipzig, line 190: How do I create a dummy `Nullifier`?

I think our database tests could use some/a-lot of improving, half the problem seems to be that's its a pain to generate reasonable data to input/expect on.

crates/store/src/db/mod.rs
```diff
@@ -63,6 +63,7 @@ pub struct NoteRecord {
     pub note_id: RpoDigest,
     pub metadata: NoteMetadata,
     pub details: Option<Vec<u8>>,
+    pub nullifier: Option<Nullifier>,
```
bobbinth, line 66: I'm not sure we actually need to have `nullifier` field here. The main reason why we want it in the table so that we could run a query that sets `consumed` for records which match a set of nullifiers. But other than that, we shouldn't really care about it (and the nullifier should be implicitly stored in the `details` field anyway).

crates/store/src/db/sql/mod.rs
```diff
@@ -582,17 +584,22 @@ pub fn insert_nullifiers_for_block(
     nullifiers: &[Nullifier],
     block_num: BlockNumber,
 ) -> Result<usize> {
+    let serialized_nullifiers: Vec<Value> =
+        nullifiers.iter().map(Nullifier::to_bytes).map(Into::into).collect();
+    let serialized_nullifiers = Rc::new(serialized_nullifiers);
+
+    let mut stmt = transaction
+        .prepare_cached("UPDATE notes SET consumed = TRUE WHERE nullifier IN rarray(?1)")?;
+    stmt.execute(params![serialized_nullifiers])?;
```
Mirko-von-Leipzig, line 593: I'm unsure if lumping in the nullifier set here is a good idea or not. We're now updating `nullifers` and `notes` table here. 

The benefit is that we can re-use the nullifier serialization.

crates/store/src/db/mod.rs
Mirko-von-Leipzig: Right but how will we find out which notes to consume? We need some way of linking the nullifier to note ID.

crates/store/src/db/sql/mod.rs
bobbinth: Ah - I see. Maybe we could make the order of fields explicit too? - e.g. something like:
```rust
impl NoteRecord {
    pub const SELECT_FIELDS: &str = "
        block_num,
        batch_index,
        note_index,
        note_id,
        note_type,
        sender,
        tag,
        aux,
        execution_hint,
        merkle_path,
        details,
        nullifier";
}
```

crates/store/src/db/tests.rs
bobbinth: Would `Digest::default().into()` not work? We could also add `default()` or `dummy()` to `Nullifier` in `miden-base`.

crates/store/src/db/mod.rs
bobbinth: Do you mean how to identify network notes which haven't been consumed yet? Wouldn't this be something like:
```SQL
Select ... from notes
where
    execution_mode = 0 AND consumed = 0
```

crates/store/src/db/tests.rs
Mirko-von-Leipzig: I found

```rust
fn num_to_nullifier(n: u64) -> Nullifier {
    Nullifier::from(num_to_rpo_digest(n))
}
```

which will work for now. The issue with these free standing functions (and `into`) is that they're impossible to find naturally. Ideally one types `Nullifier::` and let auto-complete show you the options on what's possible.

crates/store/src/db/mod.rs
Mirko-von-Leipzig: I mean when nullifiers are applied as part of a block, how do I mark their notes as being consumed. I need something where I can the equivalent of

```sql
UPDATE notes SET consumed = TRUE WHERE nullifier in rarray(?block.nullifiers)
```

crates/store/src/db/tests.rs
bobbinth: Yeah - ideally this would be something like:
```rust
impl Nullifier {

    #[cfg(any(feature = "testing", test))]
    pub fn dummy(n: u64) -> Self {
        Self::from(num_to_rpo_digest(n))
   }
}
```
Want to make a small PR in `miden-base`?

crates/store/src/db/mod.rs
bobbinth: Yes, this looks correct. But why does this mean that `NoteRecord` needs a `nullifier` field? (we need the field in the table, but can probably skip it for the struct).

Mirko-von-Leipzig: Oh my bad; misunderstood.

Mostly the issue is that `NoteRecord` is used for both inserting and reading from the db, and it felt pretty difficult to separate. And for inserting we of course need the nullifier (unless we want to recalculate it for each one?).

crates/store/src/db/sql/mod.rs
Mirko-von-Leipzig: I'll try this out, but this does mean the sql str will allocate because you need to do

```rust
format!("SELECT {} FROM notes ...", NoteRecord::SELECT_FIELDS)
```

I think there *may* be some work-arounds with macros from other crates. Though probably this won't show up ito performance anyway.

crates/store/src/db/mod.rs
Mirko-von-Leipzig: I've adapted it so the nullifier is passed along in addition to the `NoteRecord` for insertion. This feels a little disjointed to me; but I think we can leave it as is until we organise things a bit more in general.

crates/store/src/db/sql/mod.rs
```diff
@@ -1002,6 +936,50 @@ pub fn select_note_inclusion_proofs(
     Ok(result)
 }
 
+/// Returns a paginated batch of network notes that have not yet been consumed.
+///
+/// Pagination is controlled with the `offset` and `limit` values. If less than `limit` values are
+/// returned, then this is the final page.
```
bobbinth, line 942: How "stable" is this approach to pagination? Meaning, if new notes are added to the database between requests, would we pick up these new notes? (I think yes, but want to confirm).

Also, what happens if some notes are deleted? Not a concern now as we don't delete data from this table yet, but if that happens, i'm assuming we could end up skipping some notes.

bobbinth (approved):
Looks good! Thank you! I left just one question (for the future) inline.

crates/store/src/db/sql/mod.rs
Mirko-von-Leipzig: So long as new notes are appended and not inserted into some random block in the middle it would continue working. Notes are output chronologically so this would be stable for that.

This would not be stable for deletions, for that we would need a better "cursor", e.g. instead of `offset`, we use a pair `(block_num, note_index)` and continue on from there.

My assumption (for now at least) is that this would be called at block-producer/tx-builder startup and therefore there is no store mutation possible.

If we do allow store mutation then I think this all becomes much more complicated.

bobbinth: Yep, makes sense and is roughly how I think about it as well. One potential thing that can simplify this in SQLite context is using `rowid` for pagination - though, I've never done this myself.

Mirko-von-Leipzig: Well that led me down a rabbit hole. Somehow I imagined `offset` would be optimised but apparently most query engines do not do well with it.

I'll look into this a bit more, may update the PR again.

crates/store/src/db/migrations/001-init.sql
Mirko-von-Leipzig: I've checked and its using this index.

bobbinth (approved):
Looks good! Thank you!

crates/store/src/db/sql/mod.rs
```diff
@@ -1002,6 +936,54 @@ pub fn select_note_inclusion_proofs(
...
+    // Select the rowid column so that we can return a pagination token.
+    //
+    // rowid column _must_ come after the note fields so that we don't mess up the
+    // `NoteRecord::from_row` call.
+    let mut stmt = transaction.prepare_cached(&format!(
+        "
+        SELECT {}, rowid FROM notes
+        WHERE
+            execution_mode = 0 AND consumed = FALSE AND rowid >= ?
+        LIMIT ?
```
bobbinth, line 966: Question: are the rows automatically sorted by `rowid`? If not, we probably need the `ORDERY BY` clause.

Mirko-von-Leipzig: Ah. Right.

Thanks, its "undefined". I've added ordering.

PR #707, opened 2025-02-19
feat: add block_num parameter to check_nullifiers_by_prefix

TomasArrachea:
This PR introduces a `block_num` parameter to the `CheckNullifiersByPrefix` endpoint. Only nullifiers from blocks equal or bigger than `block_num` will be included in the response. This change allows using the endpoint to request nullifiers reducing the size of the response.

This is part of the changes needed for https://github.com/0xPolygonMiden/miden-node/pull/685, in order to remove the nullifiers from the SyncState response. 

TODO: This requires an additional PR in miden-client to update the endpoint request.


proto/requests.proto
```diff
@@ -18,6 +18,8 @@ message CheckNullifiersByPrefixRequest {
     // List of nullifiers to check. Each nullifier is specified by its prefix with length equal
     // to `prefix_len`.
     repeated uint32 nullifiers = 2;
+    // Block number from which the nullifiers are requested (exclusive).
+    optional fixed32 block_num = 3;
```
bobbinth, line 22: Question: do we need to make this optional? If the user wants to get all possible nullifiers, they could just set this to `0` (assuming we make the parameter inclusive).

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 
 - [BREAKING] Updated minimum Rust version to 1.84.
 - [BREAKING] `Endpoint` configuration simplified to a single string (#654).
+- Added `block_num` parameter to `CheckNullifiersByPrefix` endpoint (#707).
```
Mirko-von-Leipzig, line 19:
```suggestion
- [BREAKING] `CheckNullifiersByPrefix` now takes a starting block number (#707).
```

crates/store/src/db/sql/mod.rs
```diff
@@ -672,6 +672,9 @@ pub fn select_nullifiers_by_block_range(
 /// of the nullifier of interest to the client. This hides the details of the specific
 /// nullifier being requested. Currently the only supported prefix length is 16 bits.
 ///
+/// Additionally, the nullifiers are filtered by the block number, so only nullifiers created
+/// at or after the given block number are returned.
+///
```
Mirko-von-Leipzig, line 677: I think we can refactor the first paragraph to simplify:

```
Returns nullifiers filtered by prefix and block creation height.
```

Mirko-von-Leipzig (approved):

bobbinth (approved):
Looks good! Thank you! One last thing that's I'd probably do is update the description of the endpoint in the README: https://github.com/0xPolygonMiden/miden-node/tree/next/crates/rpc#checknullifiersbyprefix.

PR #709, opened 2025-02-19
Use `LocalBlockProver` to build blocks

PhilippGackstatter:
Use the `LocalBlockProver` from https://github.com/0xPolygonMiden/miden-base/pull/1152 to build blocks.

Builds on this PR in miden-base: https://github.com/0xPolygonMiden/miden-base/pull/1172.

I ran the client integration tests against this branch and they all passed.

I checked that all cases of the deleted tests in the block builder are covered by tests in miden-base.


crates/block-producer/src/block_builder/mod.rs
```diff
@@ -287,34 +259,30 @@ impl SelectedBlock {
...
         // SAFETY: We do not expect to have more than u32::MAX of any count per block.
         span.set_attribute(
             "block.updated_accounts.count",
-            i64::try_from(self.summary.updated_accounts.len())
+            i64::try_from(self.inputs.account_witnesses().len())
                 .expect("less than u32::MAX account updates"),
         );
-        span.set_attribute(
-            "block.output_notes.count",
-            i64::try_from(self.summary.output_notes.iter().fold(0, |acc, x| acc.add(x.len())))
```
PhilippGackstatter, line 302: Not sure what the best approach for telemetry is. We could:
- split it in three sections (see below)
- or keep it as two sections (just block batches and input + proven block, which can include the proposed block stuff)
- or as just one on the proven block, if we don't care about the number of unauthenticated notes.

```
// Block Batches and Input

// This info is only available here because the proposed block doesn't track what notes were erased explicitly.
"block.unauthenticated_notes.count",
// This can be here or somewhere else.
"block.updated_accounts.count",

// Proposed Block:

// Nullifiers and output notes should be emitted when note erasure was done which is here.
"block.nullifiers.count",
"block.output_notes.count",
// We could also emit the timestamp here, looks like we don't do that anywhere yet?

// Proven Block:

// BlockHeader is only available on a proven block.
"block.hash"
"block.sub_hash"
"block.parent_hash"
"block.protocol.version"
"block.commitments.kernel"
"block.commitments.nullifier"
"block.commitments.account"
"block.commitments.chain"
"block.commitments.note"
"block.commitments.transaction"
```

crates/store/src/state.rs
```diff
@@ -783,61 +751,110 @@ impl State {
...
+        reference_blocks: BTreeSet<BlockNumber>,
     ) -> Result<BlockInputs, GetBlockInputsError> {
+        // Get the note inclusion proofs from the DB.
+        // We do this first so we have to acquire the lock to the state just once. There we need the
+        // reference blocks of the note proofs to get their authentication paths in the chain MMR.
+        let unauthenticated_note_proofs = self
+            .db
+            .select_note_inclusion_proofs(unauthenticated_notes)
+            .await
+            .map_err(GetBlockInputsError::SelectNoteInclusionProofError)?;
```
PhilippGackstatter, line 763: This selects notes from the DB by note ID, but the notes table does not have the note id as a primary key, so this has to scan the entire table, if I'm not mistaken.
I also wonder if we need batch index and note index as primary keys.

crates/store/src/state.rs
```diff
@@ -783,61 +751,110 @@ impl State {
...
+        // The order doesn't matter for ChainMmr::new, so swap remove is fine.
+        let latest_block_header = headers.swap_remove(latest_block_header_index);
+
+        // SAFETY: This should not error because:
+        // - we're passing exactly the block headers that we've added to the partial MMR,
+        // - so none of the block header's block numbers should exceed the chain length of the
+        //   partial MMR,
+        // - and we've added blocks to a BTreeSet, so there can be no duplicates.
+        let chain_mmr = ChainMmr::new(partial_mmr, headers)
+            .expect("partial mmr and block headers should be consistent");
```
PhilippGackstatter, line 849: If for some reason, the DB does not return all requested block headers this would panic. Is that consistent with error handling in the node or should we actually return an error here and for similar `expect`s and let higher level components deal with it?

crates/block-producer/src/test_utils/block.rs
```diff
@@ -71,30 +68,31 @@ pub async fn build_expected_block_header(
...
-            iter::empty(),
-        )
-        .await
-        .unwrap();
-
-    let (block_witness, _) = BlockWitness::new(block_inputs_from_store, &batches).unwrap();
-
-    BlockProver::new().prove(block_witness).unwrap()
-}
+// Note: Commented as it is unused atm. Is it worth fixing it?
```
Mirko-von-Leipzig, line 71: If its unused we should nuke it. We can always resurrect it.

Probably a remnant of when I yanked the block building tests with the mempool rewrite.

I assume this should be better done using `MockChain` in the future

crates/store/src/db/sql/mod.rs
```diff
@@ -747,6 +747,11 @@ pub fn select_all_notes(conn: &mut Connection) -> Result<Vec<NoteRecord>> {
 
     let mut notes = vec![];
     while let Some(row) = rows.next()? {
+        let batch_idx = row.get(1)?;
+        let note_idx_in_batch = row.get(2)?;
+        // SAFETY: We can assume the batch and note indices stored in the DB are valid so this
+        // should never panic.
+        let note_index = BlockNoteIndex::new(batch_idx, note_idx_in_batch);
```
Mirko-von-Leipzig, line 754: Is the `new` constructor now panic'y?

Might want a more panic-inducing name then, instead of `new` 😅

crates/block-producer/src/test_utils/store.rs
```diff
@@ -290,65 +291,66 @@ impl MockStoreSuccess {
...
-
-        Ok(BlockInputs {
-            block_header: latest_header,
-            chain_peaks,
-            accounts,
-            nullifiers,
-            found_unauthenticated_notes,
-        })
-    }
+    // Note: Commented as it is unused atm. Is it worth fixing it?
```
Mirko-von-Leipzig, line 294: Nuke it :)

crates/store/src/state.rs
Mirko-von-Leipzig: That's correct; and we are missing quite a lot of indices. I think we only have like three indices explicitly created - so almost any non-primary/composite key lookup will be a scan at the moment.

There is #712 where we intend to go through these, maybe even figure out a way to test for lack of `scan` in the query plans.

I've also been eyeing `diesel` because I'm curious, and having compile-time checked queries would reduce the need for a lot of tests that only serve to check the sql statements.

crates/store/src/state.rs
```diff
@@ -783,61 +690,110 @@ impl State {
...
+            .select_note_inclusion_proofs(unauthenticated_notes)
+            .await
+            .map_err(GetBlockInputsError::SelectNoteInclusionProofError)?;
+
+        // The set of blocks that the notes are included in.
+        let note_proof_reference_blocks =
+            unauthenticated_note_proofs.values().map(|proof| proof.location().block_num());
+
+        // Collect all blocks we need to prove inclusion for, without duplicates.
+        let mut blocks = reference_blocks;
```
Mirko-von-Leipzig, line 709: Might be good to enforce the `Set` here, in case someone "optimises" the input type to a vector

```suggestion
        // Collect all blocks we need to prove inclusion for, without duplicates.
        let mut blocks: BTreeSet<_> = reference_blocks;
```

crates/store/src/state.rs
```diff
@@ -783,61 +690,110 @@ impl State {
...
-                NullifierWitness { nullifier: *nullifier, proof }
+                let proof = inner.nullifier_tree.open(&nullifier);
+                (nullifier, NullifierWitness::new(proof))
             })
             .collect();
 
-        let found_unauthenticated_notes =
-            self.get_note_authentication_info(unauthenticated_notes).await?;
+        // Release the lock.
+        std::mem::drop(inner);
```
Mirko-von-Leipzig, line 759: Would it make sense to have the above in a separate method? iiuc we avoid actual db access while we are locked, so we may as well?

As in, including from `.lock` until here.

crates/store/src/state.rs
Mirko-von-Leipzig: This would imply db corruption, or another writer somehow correct?

I think I'm happy with a panic here - given that we're essentially screwed if this occurs right?

crates/store/src/db/sql/mod.rs
PhilippGackstatter: Yeah it is. I mainly made it that way because it's not obvious which error type it should return and making a separate one seems verbose on the API. Maybe returning `Option<Self>` instead is a reasonable compromise, as we don't expect other error conditions to be added.

crates/store/src/errors.rs
```diff
@@ -230,27 +226,15 @@ pub enum GetBlockHeaderError {
...
-            GetNoteAuthenticationInfoError::DatabaseError(db_err) => db_err.into(),
-            GetNoteAuthenticationInfoError::MmrError(mmr_err) => Self::NoteInclusionMmr(mmr_err),
-        }
-    }
+    #[error("failed to select note inclusion proofs")]
+    SelectNoteInclusionProofError(#[source] DatabaseError),
+    #[error("failed to select block headers")]
+    SelectBlockHeaderError(#[source] DatabaseError),
+    #[error("highest block number {highest_block_number} referenced by a batch is newer than the latest block {latest_block_number}")]
+    BatchBlockReferenceNewerThanLatestBlock {
```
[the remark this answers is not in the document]
PhilippGackstatter: Went with `UnknownTransactionBlockReference` and `UnknownBatchBlockReference` for the batch and block cases, respectively (for consistency).

crates/store/src/genesis.rs
```diff
@@ -58,14 +58,17 @@ impl GenesisState {
...
             TransactionKernel::kernel_root(),
             Digest::default(),
             self.timestamp,
         );
 
-        Block::new(header, accounts, vec![], vec![]).map_err(Into::into)
+        // SAFETY: Header and accounts should be valid by construction.
+        // No notes or nullifiers are created at genesis, which is consistent with the above empty
+        // block note tree root and empty nullifier tree root.
+        Ok(ProvenBlock::new_unchecked(header, accounts, vec![], vec![]))
```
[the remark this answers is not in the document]
PhilippGackstatter: I like the idea in principle, but honestly can't make it work in a clean way. No matter if you define the tree or vec first, you have to map it somehow to connect them.

```rust
let empty_nullifiers: Vec<Nullifier> = Vec::new();
let empty_nullifier_tree = Smt::with_entries(
    empty_nullifiers
        .clone()
        .into_iter()
        .map(|nullifier| (nullifier.inner(), EMPTY_WORD)),
)
.unwrap();
```

Unless you have another idea, I'll go with defining them next to each other.

```rust
let empty_nullifiers: Vec<Nullifier> = Vec::new();
let empty_nullifier_tree = Smt::new();

let empty_output_notes = Vec::new();
let empty_block_note_tree = BlockNoteTree::empty();
```

crates/block-producer/src/block_builder/mod.rs
PhilippGackstatter: Sounds good. I pushed updates now. I also remove the previous `ProvenBlock` (called `BuiltBlock` in the meantime) in favor of a local `TelemetryInjector` trait. That reads cleaner and we don't have to come up with new names for proposed and proven block. Main reason for using it was because I couldn't come up with a wrapper name for proposed block :smile:.

Also added erased notes as an explicit metric. Renamed `block.unauthenticated_notes.count` to `block.unauthenticated_note_proofs.count` because that's more accurate and doesn't conflict with erased notes. This is a bit inconsistent now in regards to naming and there is an issue for it, so we might figure it out eventually (https://github.com/0xPolygonMiden/miden-base/issues/1144).

Mirko-von-Leipzig (approved):

crates/store/src/state.rs
```diff
@@ -777,64 +684,135 @@ impl State {
...
-                }
-            })?;
-        let account_states = account_ids
+        // The latest block is not yet in the chain MMR, so we can't (and don't need to) prove its
+        // inclusion in the chain.
+        blocks.remove(&latest_block_number);
+
+        // Fetch the partial MMR at the state of the latest block with authentication paths for the
+        // provided set of blocks.
+        let partial_mmr = inner.blockchain.partial_mmr_from_blocks(blocks, latest_block_number);
```
bobbinth, line 789: I had a related comment in another PR: we could probably just pass the `block` into `partial_mmr_from_blocks()` and remove the `lated_block_number` after - e.g., something like this:
```rust
let partial_mmr = inner.blockchain.partial_mmr_from_blocks(blocks);
blocks.remove(&latest_block_number);
```

bobbinth (approved):
Looks good! Thank you! Not a very deep review from me (especially around telemetry-related code), but I left some comments inline.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -287,41 +264,76 @@ impl SelectedBlock {
...
+            "block.erased_note_proofs.count",
+            i64::try_from(self.inputs.unauthenticated_note_proofs().len())
+                .expect("less than u32::MAX unauthenticated notes"),
         );
+    }
+}
+
+trait TelemetryInjector {
+    fn inject_telemetry(&self);
+}
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Do we need it to be a trait at all? I mean sure the methods all look the same, but they're not used generically/dynamically anywhere.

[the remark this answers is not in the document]
PhilippGackstatter: I mentioned it here (https://github.com/0xPolygonMiden/miden-node/pull/709#discussion_r1965695526) that I just introduced this as an extension trait so that we can implement `inject_telemtry` for `ProvenBlock` and `ProposedBlock`. So it's intended to be a local-only trait, so I wouldn't make it more general or even public. The main point is that without this, we have to wrap these into some new structs for which we have to come up with names and the way it is now its reads nicer where its used:

```rust
//...
.and_then(|inputs| self.propose_block(inputs))
.inspect_ok(ProposedBlock::inject_telemetry)
```

But if the trait is not preferred, we can also go with `{Proposed,Proven}BlockTelemetry` wrappers or something.

crates/store/src/state.rs
PhilippGackstatter: `latest_block_number` is used as the target forest of the partial MMR, so if the `blocks` do contain the latest block number and it is not removed, we will try to get a proof for block `latest_block_number` even though the MMR we're getting it from only has length `latest_block_number`, which will result in an error.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -119,67 +116,98 @@ impl BlockBuilder {
...
+            // error in BlockProducer::serve. Not sure if there's a better fix. Error:
+            // implementation of `FnOnce` is not general enough
+            // closure with signature `fn(&InputNoteCommitment) -> miden_objects::note::NoteId` must
+            // implement `FnOnce<(&InputNoteCommitment,)>` ...but it actually implements
+            // `FnOnce<(&InputNoteCommitment,)>`
+            batch
+                .input_notes()
+                .iter()
+                .cloned()
+                .filter_map(|note| note.header().map(NoteHeader::id))
```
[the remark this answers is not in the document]
PhilippGackstatter: It's not wrapped in an `Arc` or anything, so cloning is a deep-clone. I don't think it's particularly useful here, since we only clone an `InputNoteCommitment` but it seems it could be an efficiency gain in `BatchGraph::select_block` where we clone all batches that are selected.

It would also get more useful when the batch gets bigger when with inclusion of `VerifiedTransaction`s in it (https://github.com/0xPolygonMiden/miden-base/issues/1121).

crates/store/src/server/api.rs
```diff
@@ -569,3 +570,8 @@ fn validate_notes(notes: &[generated::digest::Digest]) -> Result<Vec<NoteId>, St
         .collect::<Result<_, ConversionError>>()
         .map_err(|_| invalid_argument("Digest field is not in the modulus range"))
 }
+
+#[instrument(target = COMPONENT, skip_all)]
+fn read_block_numbers(block_numbers: &[u32]) -> BTreeSet<BlockNumber> {
+    block_numbers.iter().map(|raw_number| BlockNumber::from(*raw_number)).collect()
+}
```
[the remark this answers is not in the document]
PhilippGackstatter: I put it there for consistency with the other similar functions above it.

igamigo:
BTW [this PR](https://github.com/0xPolygonMiden/miden-client/pull/758) fixes the sync (we didn't get to merge it but it's basically ready). I tested integration tests against this PR and they seem to work well!

PhilippGackstatter:
> BTW [this PR](https://github.com/0xPolygonMiden/miden-client/pull/758) fixes the sync (we didn't get to merge it but it's basically ready). We tested integration tests against this PR and they seem to work well!

Thank you very much! I also did a run of the integration tests and they all passed as well, so I'll merge :+1:.

Issue #674, opened 2025-02-07
Migrate to Edition 2024


Mirko-von-Leipzig:
This will become available with Rust 1.85 (~20 Feb 2025).

This _should_ be mostly trivial with `cargo fix --edition`; this [post](https://codeandbitters.com/rust-2024-upgrade/) describes a more piecemeal approach if we run into issues.

We also need to check for new lints we may want to enable.

Mirko-von-Leipzig:
Consider reverting #1181 once complete to let `machete` auto update.

Mirko-von-Leipzig:
Closed by #727

PR #720, opened 2025-02-25
Pin cargo machete to a specific version

PhilippGackstatter:
Pin cargo machete to a specific version since the latest one requires Edition 2024 which causes the action to fail (see current next).

Related: https://github.com/0xPolygonMiden/miden-base/pull/1181

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -94,8 +94,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@main
-      - name: machete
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):
I wonder if we shouldn't let this `machete` auto-upgrade again once #674 is complete. I imagine it improves over time and catches more dead deps..

PhilippGackstatter:
> I wonder if we shouldn't let this `machete` auto-upgrade again once #674 is complete. I imagine it improves over time and catches more dead deps..

Definitely, my intention is for this to be temporary just to make the CI work until we have upgrded :)

Issue #712, opened 2025-02-20
Optimise store indexing


Mirko-von-Leipzig:
Our sql queries are (likely) missing a bunch of indices.

The main way this will show up is (eventually) with slow queries once the db is populated. While #609 may help identify these bottlenecks, one can already identify most candidates using [`explain query plan`](https://www.sqlite.org/eqp.html).

Queries should in general be using indices (`search`) and avoiding full row `scan`s. Missing indices can therefore be identified, added and rechecked to ensure they're being used.

A more robust solution would involve running this in tests somehow, and ensuring there are no `scan` involved. There may of course be exceptions, but these can be whitelisted.

bobbinth:
Related: #57.

PR #721, opened 2025-02-26
Implemented database optimization routine

polydez:
Implemented database optimization routine:
1. Full optimization with shrinking and analysis on node start.
2. Limited indexes analysis in background each day (default value, configurable).

This is a recommended way according to docs: https://www.sqlite.org/pragma.html#pragma_optimize

Such optimization routine should help query planner to better make decision which index to use when it has choice.

### DIFF
--- a/crates/store/src/db/migrations.rs
+++ b/crates/store/src/db/migrations.rs
@@ -77,8 +77,8 @@ pub fn apply_migrations(conn: &mut Connection) -> super::Result<()> {
 
     info!(target: COMPONENT, "Starting database optimization");
 
-    // Run full database optimization. This will shrink the database and run indexes analysis for
-    // the query planner. This will also increase the `schema_version` value.
+    // Run full database optimization. This will run indexes analysis for the query planner.
+    // This will also increase the `schema_version` value.
     //
     // We should run full database optimization in following cases:
     // 1. Once schema was changed, especially new indexes were created.
--- a/crates/store/src/db/migrations.rs
+++ b/crates/store/src/db/migrations.rs
@@ -75,6 +75,20 @@ pub fn apply_migrations(conn: &mut Connection) -> super::Result<()> {
         Settings::set_value(conn, DB_MIGRATION_HASH_FIELD, &new_hash)?;
     }
 
+    info!(target: COMPONENT, "Starting database optimization");
+
+    // Run full database optimization. This will shrink the database and run indexes analysis for
+    // the query planner. This will also increase the `schema_version` value.
+    //
+    // We should run full database optimization in following cases:
+    // 1. Once schema was changed, especially new indexes were created.
+    // 2. After restarting of the node, on first connection established.
+    //
+    // More info: https://www.sqlite.org/pragma.html#pragma_optimize
+    conn.execute("PRAGMA optimize = 0x10002;", ())?;
+
+    info!(target: COMPONENT, "Finished database optimization");
+
     let new_schema_version = schema_version(conn)?;
     debug!(target: COMPONENT, new_schema_version, "Updating schema version in settings table");
     Settings::set_value(conn, DB_SCHEMA_VERSION_FIELD, &new_schema_version)?;
--- a/crates/store/src/server/db_maintenance.rs
+++ b/crates/store/src/server/db_maintenance.rs
@@ -1,8 +1,8 @@
 use std::{sync::Arc, time::Duration};
 
-use tracing::{error, info};
+use miden_node_utils::tracing::OpenTelemetrySpanExt;
 
-use crate::{state::State, COMPONENT};
+use crate::state::State;
 
 pub struct DbMaintenance {
     state: Arc<State>,
@@ -19,11 +19,14 @@ impl DbMaintenance {
         loop {
             tokio::time::sleep(self.optimization_interval).await;
 
-            info!(target: COMPONENT, "Starting database optimization");
+            let root_span = tracing::info_span!(
+                "optimize_database",
+                interval = self.optimization_interval.as_secs_f32()
+            );
 
-            match self.state.optimize_db().await {
-                Ok(_) => info!(target: COMPONENT, "Finished database optimization"),
-                Err(err) => error!(target: COMPONENT, %err, "Database optimization failed"),
+            {
+                let _enter = root_span.enter();
+                self.state.optimize_db().await.unwrap_or_else(|err| root_span.set_error(&err));
             }
         }
     }
--- a/crates/store/src/server/db_maintenance.rs
+++ b/crates/store/src/server/db_maintenance.rs
@@ -1,6 +1,7 @@
 use std::{sync::Arc, time::Duration};
 
 use miden_node_utils::tracing::OpenTelemetrySpanExt;
+use tracing::{Instrument, Span};
 
 use crate::state::State;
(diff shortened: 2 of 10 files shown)


crates/store/src/db/migrations.rs
```diff
@@ -75,6 +75,20 @@ pub fn apply_migrations(conn: &mut Connection) -> super::Result<()> {
         Settings::set_value(conn, DB_MIGRATION_HASH_FIELD, &new_hash)?;
     }
 
+    info!(target: COMPONENT, "Starting database optimization");
+
+    // Run full database optimization. This will shrink the database and run indexes analysis for
+    // the query planner. This will also increase the `schema_version` value.
```
Mirko-von-Leipzig, line 81: This does _not_ shrink the database, it just marks them for re-analysis for optimisation.

```
0x00002 | Run ANALYZE on tables that might benefit. On by default.
          
0x10000 | Check the size of all tables, not just tables that have not been recently
        | used, to see if any have grown and shrunk significantly and hence might 
        | benefit from being re-analyzed. Off by default.
```

crates/store/src/server/mod.rs
```diff
@@ -54,18 +63,33 @@ impl Store {
...
-            .map_err(ApiError::ApiServeFailed)
+            .serve_with_incoming(TcpListenerStream::new(self.listener));
+
+        tokio::select! {
+            api_service = api_service => {
+                api_service.map_err(ApiError::ApiServeFailed)
+            },
+
+            _ = db_maintenance_service => {
+                info!(target: COMPONENT, "Database maintenance service shutdown");
```
Mirko-von-Leipzig, line 89: This indicates an error/crash?
```suggestion
                error!(target: COMPONENT, "Database maintenance service crashed");
```

Mirko-von-Leipzig: Should probably also return an error. Or consider not crashing the node at all.

crates/store/src/server/mod.rs
```diff
@@ -54,18 +63,33 @@ impl Store {
...
+
+                Ok(())
```
Mirko-von-Leipzig, line 91: Are you sure this works as expected?

To me this means the maintenance service will block the api service while its running as only one of these can make progress at once?

I would expect the maintenance service to be `spawn`.

Mirko-von-Leipzig (commented):
Some questions.

This optimises the query planner i.e. index selection and ordering. Do you think our queries are complex enough for this to be beneficial?

There is also the issue of reproducibility:

> Some developers prefer that once the design of an application is frozen, SQLite will always pick the same query plans as it did during development and testing. Then if a millions of copies of the application are shipped to customers, the developers are assured that all of those millions of copies are running the same query plans regardless of what data the individual customers insert into their particular databases. This can help in reproducing complaints of performance problems coming back from the field.
>
> To achieve this objection, never run a full ANALYZE nor the "PRAGMA optimize" command in the application. Rather, only run ANALYZE during development, manually using ...

---------

This also does not shrink the database. That is done using `VACUUM` but it will take forever on large databases and should not be done automatically. Database compaction should really only be done manually on demand as the node will have significant downtime during it.

polydez:
> This optimises the query planner i.e. index selection and ordering. Do you think our queries are complex enough for this to be beneficial?

I think, yes, we run pretty complex queries for account states computing with lots of `AND`s which require query analyzer to choose which index to use.

> Or put differently - we still need to create indices for it to choose from, but more often than not we'll only have a single index for it to choose from per query. This doesn't automatically create new indices.

I didn't expect from SQLite to create indices, I'm adding new indices in another PR in terms of #712.

> There is also the issue of reproducibility:

Yes, and we should probably discuss, what is better: performance over reproducibility of performance issues or reproducibility over performance. To be honest, I can't imagine, how `PRAGMA optimize` could ruin performance.

> This also does not shrink the database. That is done using `VACUUM` but it will take forever on large databases and should not be done automatically. Database compaction should really only be done manually on demand as the node will have significant downtime during it.

Ah, my bad! I was confused by documentation and though that bit `0x10000` will cause database to shrink. I have checked this on small PoC code and now I agree, it doesn't shrink it.

crates/store/src/db/migrations.rs
```diff
@@ -75,6 +75,20 @@ pub fn apply_migrations(conn: &mut Connection) -> super::Result<()> {
         Settings::set_value(conn, DB_MIGRATION_HASH_FIELD, &new_hash)?;
     }
 
+    info!(target: COMPONENT, "Starting database optimization");
+
+    // Run full database optimization. This will run indexes analysis for the query planner.
+    // This will also increase the `schema_version` value.
```
Mirko-von-Leipzig, line 81: How does this schema version get incremented?

Mirko-von-Leipzig (commented):
Code LGTM - just need to decide whether we want determinism.

Mirko-von-Leipzig:
> > There is also the issue of reproducibility:
> 
> Yes, and we should probably discuss, what is better: performance over reproducibility of performance issues or reproducibility over performance. To be honest, I can't imagine, how `PRAGMA optimize` could ruin performance.

I've had it make a bad decision once or twice where it picked the wrong ordering. In a way its like any optimiser/compiler - sometimes it does go the wrong way. Mostly these issues were related to changes in data composition over time, e.g. more output notes per tx and block

crates/store/src/db/migrations.rs
polydez: It's incremented by running `PRAGMA optimize = 0x10002` after migration. I guess, some prepared statements might exist after migration and SQLite invalidates them by increasing schema version after statistics recalculation, but I haven't found any explanation of that behavior anywhere.

Mirko-von-Leipzig (approved):

crates/store/src/server/db_maintenance.rs
```diff
@@ -19,11 +19,14 @@ impl DbMaintenance {
...
+                "optimize_database",
+                interval = self.optimization_interval.as_secs_f32()
+            );
 
-            match self.state.optimize_db().await {
-                Ok(_) => info!(target: COMPONENT, "Finished database optimization"),
-                Err(err) => error!(target: COMPONENT, %err, "Database optimization failed"),
+            {
+                let _enter = root_span.enter();
+                self.state.optimize_db().await.unwrap_or_else(|err| root_span.set_error(&err));
```
Mirko-von-Leipzig, line 29: Spans unfortunately don't work with async because the future may be moved to a separate thread of execution. https://docs.rs/tracing/latest/tracing/struct.Span.html#in-asynchronous-code

Instead one needs to instrument it.

```suggestion
            self.state.optimize_db().instrument(root_span).await.inspact_err(|err| root_span.set_error(&err));
```

polydez: Good point, thank you! Unfortunately, suggested code can't be compiled because `root_span` is consumed by `instrument` method, but used later in the `inspect_err` method.

Mirko-von-Leipzig: They're cheap to clone iirc; or at least that's how I've been using them https://github.com/0xPolygonMiden/miden-node/blob/7546dc77036fcf8acf4a339cde5b5cb91051877e/crates/block-producer/src/block_builder/mod.rs#L103-L120

PR #723, opened 2025-02-27
WIP: remove rpc client

### LINKED ISSUE #658: Figure out RPC client and TLS
Our public testnet and devnet RPC endpoints now use https and therefore require clients with TLS support. In the `tonic` crate this means enabling one the `tls` related features, and the client automatically gains TLS support e.g. `cargo add tonic --features tls-native-roots`.

Some users are using the gRPC client from our `rpc` crate -- however this crate does not enable TLS because internally our infrastructure is still http only. This means users of this client need to additionally override `tonic` with the TLS features. This is obscure and the client errors aren't very helpful in pointing out the issue since its just a connection rejection.

### RPC client

We have a few options to improve this. Firstly though we should decide what the public interface of the `rpc` crate should be. The primary goal of the crate is to provide the RPC component of the node -- and despite the name, this is not meant as the canonical RPC client. Its intended to be the node's RPC server component.

Our options here:

1. Remove the RPC client from the rpc crate. Users should generate their own client using `rpc-proto`.
2. (1) but also add the client generation to the `rpc-proto` crate.
3. Keep as is.

### TLS features

If we go with (2) or (3) from above then we should also improve the TLS situation to make it easier to get right. Some options:

1. Only document the `cargo add tonic ..` trick.
2. Add proxy features and choose what the default should be to minimize surprises.

There are three `tonic` TLS features:

- `tls`: Enables the [rustls](https://docs.rs/rustls) based TLS options for the transport feature. Not enabled by default.
- `tls-native-roots`: Adds system trust roots to [rustls](https://docs.rs/rustls)-based gRPC clients using the [rustls-native-certs](https://docs.rs/rustls-native-certs) crate. Not enabled by default.
- `tls-webpki-roots`: Add the standard trust roots from the [webpki-roots](https://docs.rs/webpki-roots) crate to rustls-based gRPC clients. Not enabled

The latter two automatically configure TLS using the system or the mozilla certificate stores respectively. I'm unsure how these features combine, if at all. The former allows users to configure their own certificate (and is enabled as part of the other two as well).

TomasArrachea:
Closes #658


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - [BREAKING] Added `block_prover_url` to block producer configuration (#719).
 - [BREAKING] Updated to Rust Edition 2024 (#727).
 - [BREAKING] MSRV bumped to 1.85 (#727).
+- Removed `miden-rpc-proto` and introduced `miden-node-proto-build`. Added "rpc-client" feature flag on `miden-node-proto`. (#723).
```
bobbinth, line 32: This is a breaking change, right?

TomasArrachea: Yes, changed it

crates/proto_build/src/lib.rs
Mirko-von-Leipzig, line 1: Thinking about the API some more I'm wondering if we shouldn't simply aggregate the builders.

Looking at how we want to use this internally:

- We want to enrich the bindings with our domain types in the `proto` crate.
- This includes _all_ clients and servers.

And for third parties:

- they're primarily interested in the RPC client, but
- might want to generate the RPC server for
  - testing
  - proxy servers

So taking this into consideration lets instead do:

- Add `internal` feature to `proto_build`, off by default.
  - This adds `store.proto` and `block_producer.proto` to the Builder.
- Single `Builder(tonic_build::Builder), and the only differences are the above proto list changes based on `internal` feature.
- Re-expose most `tonic_build::Builder` options, including server and client as fns.
- In `proto` we then activate the `internal` feature and generate both client and server.

WDYT?

Mirko-von-Leipzig: A different approach might be to simply expose the `FileDescriptorSet` as constants and then the caller can do whatever they want?

This might be even simpler.. then no need to fiddle with the builder etc, the entire crate becomes (excluding the `build.rs` it will require to compile the protos.

```rust
pub const RPC_FILE_DESCRIPTOR: tonic_build::FileDescriptor = ...;

#[cfg(internal)]
pub const RPC_FILE_DESCRIPTOR: tonic_build::FileDescriptor = ...;
#[cfg(internal)]
pub const RPC_FILE_DESCRIPTOR: tonic_build::FileDescriptor = ...;
```

TomasArrachea: I think having a Builder still makes sense to simplify codegen for external users. And adding the flags for internal and external use makes sense.
On the contrary, the approach of exposing file descriptors would not be that simple to implement as constants. I imagine it would require generating the encoded file descriptors at build time.

bobbinth (commented):
Thank you! Not a review yet, but I wanted to understand the overall approach. The way I see it is as follows:

- We still keep protobuf files in the `proto` directory in the root of the repo as the "source of truth".
- These files are then copied to the `miden-node-proto-build` crate during the build process and are used to set the protobuf files for the tonic builder (which is passed-in externally).
  - I actually didn't see the code that copies the file from the root into the crate - but maybe I missed it.
- Then, in `miden-node-proto` we import `miden-node-proto-build` as a dependency and use it to actually generate the bindings.
- If someone wanted to use the protobuf files (e.g., `miden-client`) they would add `miden-node-proto-build` as a build dependency and this would allow them to generate the desired bindings.

Assuming this is roughly correct, I think it works (and is probably a cleaner design than what we had before) - but I don't think it addresses the original issue. That is, we still need to either remove the client code from the PRC crate or somehow simplify the support for TLS.

I also think that people will still probably try to use `miden-node-rpc` to access the RPC component as the client and maybe pointing them to `miden-node-proto-build` may be too much of a burden.

One solution is to add `miden-node-rpc-client` crate (which would basically replace the old `miden-rpc-proto` crate). This crate would expose bindings with TLS and other features configured. This crate can then be directly used in `miden-client` (i.e., no need to build anything).

Another option is to use a single `miden-node-rpc` crate but have `client` and `server` features. The `server` feature would be used by the node internally, and the `client` feature would be used externally (e.g., by the `miden-client`). This may be a bit more complicated to implement though.

bobbinth:
> I was hoping to _only_ have the proto files directly in the `miden-node-proto-build`. No more copies, only that single set. We would highlight their location in the readme and the docs if we're concerned about discoverability, or we could move this particular crate to the top level (instead of within `./crates`) and could even keep the folder named `proto`.

Yeah, I was mostly concerned about discoverability. Adding this to docs could work - but it'll need to be pretty prominent otherwise those searching for proto files won't find them. Moving the crate to the root level may be better.

> My hope is really to push users to the builder only. They may need both client and server for RPC - the latter for testing, or to create a proxy. In this case I think having both exposed by default, along with a solid example, should suffice as a jumping off point for users.

For testing, just RPC is not enough - right? They'd have to run the full node.

I think the builders could be for more sophisticated users and if we can provide the required functionality via a simple import (assuming it doesn't complicate things too much for us), that would be better. But maybe this can be a separate issue.

-----

To summarize:

- We move `miden-node-proto-build` to the root level (into the `proto` directory). I guess the actual protobuf files will end up in `proto/proto` directory - right?
  - We add a good example of how to set up a build script using the `miden-node-proto` crate.
- We remove the client code from `miden-node-rpc` and add relevant sections to the readme to point to `miden-node-proto-build`.

TomasArrachea:
> We remove the client code from miden-node-rpc and add relevant sections to the readme to point to miden-node-proto-build.

I am not sure if I am completely on the same page on this. In the rpc crate there is only a [RpcApi](https://github.com/0xPolygonMiden/miden-node/blob/67eb223069506795f123396b77545f91a459aa95/crates/rpc/src/server/api.rs#L37-L40) server component, that is used on the node binary. Which client code do you mean?

Mirko-von-Leipzig:
> > We remove the client code from miden-node-rpc and add relevant sections to the readme to point to miden-node-proto-build.
> 
> @bobbinth I am not sure if I am completely on the same page on this. In the rpc crate there is only a [RpcApi](https://github.com/0xPolygonMiden/miden-node/blob/67eb223069506795f123396b77545f91a459aa95/crates/rpc/src/server/api.rs#L37-L40) server component, that is used on the node binary. Which client code do you mean?

He meant the client exposed from `proto` which is prebuilt. Or rather I guess I did because  I wrote `rpc` in the issue description, confusing the issue.

crates/proto_build/src/lib.rs
Mirko-von-Leipzig: > On the contrary, the approach of exposing file descriptors would not be that simple to implement as constants. I imagine it would require generating the encoded file descriptors at build time and exposing that path.

It could also just be exposed via functions then. I much prefer this option over the builder so that users (and ourselves) can fully customise the codegen as they require. For users that require hand-holding a minimal example would be enough imo.

Mirko-von-Leipzig:
@TomasArrachea apologies I think we're pulling you this way and that. Mostly my fault for now having a clear vision of how this would play out.

I think we're fairly close ito the large picture, but imo we should do the following:

1. Top level `proto` folder containing the `.proto` definitions, and the `miden-node-proto-build` crate. This is correct as you have it now.
    1. Change from outputting builders to outputting the three different file descriptors, either via functions or constants or whatever you can make work.
    2. These would have to be generated using `build.rs` I imagine, unless we're including the `.proto` files in the crate as well. Which could be possible?
    3. We use an `internal` feature (off by default) to hide the `store` and `block-producer` file descriptors.
2. Our internal crate becomes `miden-node-proto` but it doesn't need any feature flags. This is intended purely for internal use. If someone wants to use the clients/servers from here then that's their problem. We could consider renaming it `miden-node-proto-internal` as well.
3. We should then remove the various proto build scripts and the generated code from being checked in ideally.

TomasArrachea:
> - Change from outputting builders to outputting the three different file descriptors, either via functions or constants or  whatever you can make work.
> - These would have to be generated using build.rs I imagine, unless we're including the .proto files in the crate as well. Which could be possible?

The tonic_build `compile_protos_with_config` function receives both the file descriptors with compiled proto files, and the raw proto files. So we would need to distribute the proto files anyway. For this, I think e could just keep the old way of copying the files.

TomasArrachea:
> 3. We should then remove the various proto build scripts and the generated code from being checked in ideally.

By this you mean to have the generated code stored in "OUT_DIR", instead of being tracked and commited with the source code on `miden-node-proto`? In that case, I think I would be a good idea to do that in a separate PR and discuss it there.

igamigo:
> > 3. We should then remove the various proto build scripts and the generated code from being checked in ideally.
> 
> By this you mean to have the generated code stored in "OUT_DIR", instead of being tracked and commited with the source code on `miden-node-proto`? In that case, I think I would be a good idea to do that in a separate PR and discuss it there.

This discussion is also about this approach: https://github.com/0xPolygonMiden/miden-base/discussions/994

Mirko-von-Leipzig:
> The tonic_build `compile_protos_with_config` function receives both the file descriptors with compiled proto files (I think this is used for server reflexion), and the raw proto files. So we would need to distribute the proto files anyway. For this, I think we could just keep the old way of copying the files and let the users compile them.

There is the https://docs.rs/tonic-build/latest/tonic_build/struct.Builder.html#method.compile_fds_with_config which compiles using the file descriptors alone. I _think_ the file descriptors fully describe the `.proto` files and are essentially just the parsed version of them used by `tonic_build`, but I could be wrong.

Mirko-von-Leipzig:
> > > 3. We should then remove the various proto build scripts and the generated code from being checked in ideally.
> > 
> > 
> > By this you mean to have the generated code stored in "OUT_DIR", instead of being tracked and commited with the source code on `miden-node-proto`? In that case, I think I would be a good idea to do that in a separate PR and discuss it there.
> 
> This discussion is also about this approach: [0xPolygonMiden/miden-base#994](https://github.com/0xPolygonMiden/miden-base/discussions/994)

I'll revive that discussion and we can discuss it there. Let's leave that suggestion out of this PR :)

proto/build.rs
Mirko-von-Leipzig, line 1: I believe you should be able to fold this into the `lib.rs` file.

Since this crate is itself intended for consumption in a `build.rs`, it shouldn't need it itself.

Essentially each `fn xxx_file_descriptor` would:

1. Emit `println!("cargo::rerun-if-changed") for its set of files. Though maybe this is a pain because you'll sort of just do all of them? That might be fine though.
2. `println!("cargo::rerun-if-env-changed=BUILD_PROTO");` should move to our internal caller `build.rs`
3. File descriptor gets created in the function itself from the proto file. We'll need to check if these files get included automatically in the uploaded crate - I believe so?

proto/Cargo.toml
```diff
@@ -1,20 +1,27 @@
...
 license.workspace      = true
-name                   = "miden-rpc-proto"
+name                   = "miden-node-proto-build"
 readme                 = "README.md"
 repository.workspace   = true
 rust-version.workspace = true
 version.workspace      = true
 
+[features]
+internal = []
```
Mirko-von-Leipzig, line 15:
```suggestion
# Enables the gRPC file descriptors for the internal component APIs,
# i.e. the `store` and `block-producer` gRPC services.
internal = []
```

proto/build.rs
Mirko-von-Leipzig: Thinking about this more, maybe a build step is still required here just for the `rerun-if-changed` bit. Though its unclear to me.

proto/src/lib.rs
```diff
@@ -1,60 +1,28 @@
...
-
-    let includes = &[proto_dir];
-    let file_descriptors = protox::compile(proto_files, includes)?;
-    fs::write(&file_descriptor_path, file_descriptors.encode_to_vec())
-        .context("writing file descriptors")?;
-
-    let mut prost_config = prost_build::Config::new();
-    prost_config.skip_debug(["AccountId", "Digest"]);
+#[cfg(feature = "internal")]
+pub fn store_file_descriptor() -> Result<FileDescriptorSet, protox::Error> {
```
Mirko-von-Leipzig, line 19: I think these shouldn't return a `Result` - we can just `.expect` since we know the files are good.

proto/README.md
```diff
@@ -0,0 +1,37 @@
+# Miden proto build
+
+This crate contains the proto definitions for the gRPC endpoints at `proto/proto`.
+
+Also exposes the compiled proto `FileDescriptorSet` to generate the Rust bindings. See next how to use it.
```
bobbinth, line 5: I'd change it to something like:

--------------

This crate contains Protobuf files defining the Miden node gRPC API. The files are exposed via `FileDescriptorSet`s to simplify generation of Rust bindings.

It also contains the raw Protobuf files in the `proto` directory to be used for binding generation in other languages.

proto/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Usage
+Likely, you will need to set up a `build.rs` file that generates the proto bindings at build time. How use the proto builder:
```
bobbinth, line 8: I'd make "Usage" a lower-level section (i.e., `## Usage`). Also, I'd write the text as follows:

---------

To generate Miden node gRPC bindings in Rust, you'll need to add a `build.rs` file to your project. The example below generates the RPC component bindings and writes them into the `src/generated` directory of your project.

proto/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```rust
+use std::{fs, path::PathBuf};
+use miden_node_proto_build::rpc_file_descriptor;
+
+fn main() {
+    let dst_dir = PathBuf::from("./generated");
```
bobbinth, line 15: This will go to `src/generated`, right? If so, I'd add a simple comment explaining this.

proto/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+    let file_descriptors = rpc_file_descriptor().unwrap();
+
+    let mut prost_config = prost_build::Config::new();
+    prost_config.skip_debug(["AccountId", "Digest"]);
+
+    tonic_build::configure()
+        .out_dir(dst_dir)
+        .compile_fds_with_config(prost_config, file_descriptors)
+        .context("compiling protobufs")?;
```
bobbinth, line 28: Would this generate both server and client parts? If so, I'd limit the example to generating only the client part (and add a comment about this).

proto/README.md
```diff
@@ -0,0 +1,37 @@
...
+}
+```
+
+# Features
```
bobbinth, line 32: Let's change this to `## Crate features`.

proto/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+- "internal": when set, the functions `store_file_descriptor()` and `block_producer_file_descriptor()` are exposed, that can be used to generate all the bindings needed for the node.
```
bobbinth, line 34: Let's write this as:

-------

- `internal`: exposes Protobuf file descriptors for the internal components of the node. These can be acceded via `store_file_descriptor()` and `block_producer_file_descriptor()`.

proto/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+
+const RPC_PROTO: &str = "rpc.proto";
+
+#[cfg(feature = "internal")]
+const STORE_PROTO: &str = "store.proto";
+
+#[cfg(feature = "internal")]
+const BLOCK_PRODUCER_PROTO: &str = "block_producer.proto";
+
+pub fn rpc_file_descriptor() -> Result<FileDescriptorSet, protox::Error> {
```
bobbinth, line 13: Should we rename this to `rpc_api_descriptor()`? (and similar comment for the other two components).

proto/build.rs
bobbinth: I think the difference would be that with `build.rs`, we can compile the `.proto` files into file descriptors and then in `lib.rs` we can just decode them (i.e., no compilation needed). So, I think it still might be better to have the `build.rs` file.

proto/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+    let proto_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("proto");
+    protox::compile([RPC_PROTO], &[proto_dir])
+}
```
bobbinth, line 16: It seems like here we compile the files again (similar to what we've done in the `build.rs` file. I thought the intent was to compile the files in `build.rs` and then just deserialize them here. In this scenario, we could do the following:

- Compile the protobuf files to their descriptors for each component separately in `build.rs`.
- Write these descriptors to disk.
- Here, just read the descriptors from disk using something like `include_bytes!`, decode and return them.

bobbinth (commented):
Looks good! Thank you. I left some comments inline.

One other thing: we should still remove the `ApiClient` from the `rpc` crate, right?

proto/build.rs
bobbinth: I would probably go with the simple route for this PR and then follow up with a custom file resolver approach in a subsequent PR. This way, we could merge this PR in the next day or so.

TomasArrachea:
@bobbinth thanks! I addressed the comments. 

> One other thing: we should still remove the ApiClient from the rpc crate, right?

About this, I am still not sure if I understand which code you mean. Maybe you mean the ApiClient from the `miden-node-proto` crate?

TomasArrachea:
@bobbinth There is no ApiClient in the `miden-node-rpc` crate. Assuming you mean the rpc ApiClient from the `proto` crate, the problem with deleting it is that it's used by the faucet, as we discussed in the [original issue](https://github.com/0xPolygonMiden/miden-node/issues/658#issuecomment-2691469363). I understood that Mirko's view on this was to just push users to use the new `proto-build` crate, and leave the `proto` crate for internal use only. I added a warning in the crate's readme for this ([ac049e3](https://github.com/0xPolygonMiden/miden-node/pull/723/commits/ac049e367d47d0794a83b1823fcd355ef03c2478)). I should also update this PR's title.

An alternative could be to add again a feature flag to the `proto` crate ("rpc-client"), that would allow the faucet to access it, but not be included by default. This would mean that we wouldn't be deleting the ApiClient, the generated files will stay as they are now. 

To actually delete the ApiClient from the crate we need a different solution. It could be to add bindings generation for the faucet, though I think that would add too much unnecessary boilerplate. Or perhaps to move the rpc client into a separate new crate.

crates/proto/README.md
```diff
@@ -2,5 +2,9 @@
 
 This crate contains protobuf bindings for the APIs exposed by the components of the Miden node and domain types.
 
+> ⚠️ **Warning**\
+> This crate is intended for internal use only. For gRPC bindings please see the [miden-node-proto-build](../../proto/README.md) crate.\
+> This crate does not support TLS and cannot be used as a client for the official RPC endpoints.
```
Mirko-von-Leipzig, line 7: github has builtin support for warnings and notes called [alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts).

```suggestion
> [!WARNING]
> This crate is intended for internal use only. For gRPC bindings please see the [miden-node-proto-build](../../proto/README.md) crate.\
> This crate does not support TLS and cannot be used as a client for the official RPC endpoints.
```

bobbinth (approved):
Looks good! Thank you! I left a few more minor comments line.

> There is no ApiClient in the `miden-node-rpc` crate.

I was going more from this statement from the original issue:

> Some users are using the gRPC client from our `rpc` crate -- however this crate does not enable TLS because internally our infrastructure is still http only. This means users of this client need to additionally override `tonic` with the TLS features. This is obscure and the client errors aren't very helpful in pointing out the issue since its just a connection rejection.

But maybe this meant `rpc-client`, which we now have deleted? If so, then probably nothing else to do here.

proto/README.md
```diff
@@ -0,0 +1,39 @@
+# Miden proto build
+
+This crate contains Protobuf files defining the Miden node gRPC API. The files are exposed via FileDescriptorSets to simplify generation of Rust bindings.
+
+It also contains the raw Protobuf files in the proto directory to be used for binding generation in other languages.
+
+## Usage
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I was hoping we could create an actual example but unfortunately `build.rs` is per package and therefore one cannot have one for a simple example.

One could embed an additional crate inside there though which would work. It could do something simple like query and print the latest `testnet` block header. This would of course only work if they've got the correct version/release but that would need to be the case regardless :)

proto/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+use protox::prost::Message;
+use tonic_build::FileDescriptorSet;
+
+/// Returns the Protobuf file descriptor for the RPC API.
+pub fn rpc_api_descriptor() -> FileDescriptorSet {
+    let bytes = include_bytes!(concat!(env!("OUT_DIR"), "/", "rpc_file_descriptor.bin"));
```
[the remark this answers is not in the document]
TomasArrachea: It's possible to share but not directly, we would have to keep the constants in a separate file and use `include!` on build.rs

proto/README.md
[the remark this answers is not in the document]
TomasArrachea: Added a "Enabling TLS for the RPC Client" section to the README in [3e08847](https://github.com/0xPolygonMiden/miden-node/pull/723/commits/3e08847412efb017860d535f7c5acfedf291a17e)

Mirko-von-Leipzig (approved):
Thanks! I know this was quite the back and forth :)

There will likely be some follow-up items like creating an example, and refactoring our internal RPC crate, but that can probably be done in a separate PR.

proto/src/lib.rs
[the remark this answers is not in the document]
TomasArrachea: And also since this string literal is used inside a macro, a constant would not be useful

bobbinth (approved):
All looks good! Thank you!

PR #724, opened 2025-03-01
fix: Implement correct nullifier root calculation in build_expected_b…

VolodymyrBg:
…lock_header

This commit fixes the FIXME in the build_expected_block_header function by properly calculating the nullifier root based on produced nullifiers from transaction batches.

The implementation now:
1. Collects all produced nullifiers from transaction batches
2. Updates the store's nullifier SMT with these nullifiers
3. Uses the updated SMT root as the nullifier root in the block header

This ensures that the expected block header matches the actual block header produced by the BlockProver, making tests more accurate and reliable.


Mirko-von-Leipzig:
Closing due to suspected AI/bot activity.

PR #728, opened 2025-03-06
Implemented SQL query plan explainer, added necessary indexes

polydez:
Resolves https://github.com/0xPolygonMiden/miden-node/issues/712 and https://github.com/0xPolygonMiden/miden-node/issues/57

In this PR we implemented SQL query plan explainer which writes to log something like a query plan below for each running query:

```
>> EXPLAIN QUERY PLAN -- Selects new notes matching the tags and account IDs search criteria.
SELECT
    block_num,
    batch_index,
    note_index,
    note_id,
    note_type,
    sender,
    tag,
    aux,
    execution_hint,
    merkle_path
FROM
    notes
WHERE
    -- find the next block which contains at least one note with a matching tag or sender
    block_num = (
        SELECT
            block_num
        FROM
            notes
        WHERE
            (tag IN rarray(NULL) OR sender IN rarray(NULL)) AND
            block_num > 0
        ORDER BY
            block_num ASC
    LIMIT 1) AND
    -- filter the block's notes and return only the ones matching the requested tags or senders
    (tag IN rarray(NULL) OR sender IN rarray(NULL))


QUERY PLAN
├── SEARCH notes USING INDEX sqlite_autoindex_notes_1 (block_num=?)
├── SCALAR SUBQUERY 3
│   ├── MULTI-INDEX OR
│   │   ├── INDEX 1
│   │   │   ├── LIST SUBQUERY 1
│   │   │   │   └── SCAN rarray VIRTUAL TABLE INDEX 1:
│   │   │   └── SEARCH notes USING INDEX idx_notes_tag (tag=? AND block_num>?)
│   │   └── INDEX 2
│   │       ├── LIST SUBQUERY 2
│   │       │   └── SCAN rarray VIRTUAL TABLE INDEX 1:
│   │       └── SEARCH notes USING INDEX idx_notes_sender (sender=? AND block_num>?)
│   └── USE TEMP B-TREE FOR ORDER BY
├── LIST SUBQUERY 4
│   └── SCAN rarray VIRTUAL TABLE INDEX 1:
└── LIST SUBQUERY 5
    └── SCAN rarray VIRTUAL TABLE INDEX 1:
```

This works for both tests and running node, if feature `explain-query-plans` is enabled.

In order to achieve this, we implemented our own SQLite connection pool which constructs special wrapper for each connection, returning special wrapper for each cached prepared statement. The latter is able to explain a query once it is being run. The same was done for transactions as well. This solution also helped us to get rid of deadpool-sqlite (all necessary functionality from it we have now). And configuring of connection now done in SQLite connection pool instead of hooking connection creation, it looks more elegant to me.

Also implemented some simple checking for problems in query plans during test run (test will fail, if any problem found in any query plan for queries run in tests).

### DIFF
--- a/crates/store/src/db/connection.rs
+++ b/crates/store/src/db/connection.rs
@@ -1,0 +1,64 @@
+use std::path::Path;
+
+use rusqlite::vtab::array;
+
+use crate::db::transaction::Transaction;
+
+pub struct Connection {
+    inner: rusqlite::Connection,
+}
+
+impl Connection {
+    pub fn open(path: impl AsRef<Path>) -> rusqlite::Result<Self> {
+        rusqlite::Connection::open(path).and_then(Self::new)
+    }
+
+    #[cfg(test)]
+    pub fn open_in_memory() -> rusqlite::Result<Self> {
+        rusqlite::Connection::open_in_memory().and_then(Self::new)
+    }
+
+    fn new(inner: rusqlite::Connection) -> rusqlite::Result<Self> {
+        // Feature used to support `IN` and `NOT IN` queries. We need to load
+        // this module for every connection we create to the DB to support the
+        // queries we want to run
+        array::load_module(&inner)?;
+
+        Ok(Self { inner })
+    }
+
+    pub(crate) fn inner(&self) -> &rusqlite::Connection {
+        &self.inner
+    }
+
+    pub(crate) fn inner_mut(&mut self) -> &mut rusqlite::Connection {
+        &mut self.inner
+    }
+
+    #[inline]
+    pub fn transaction(&mut self) -> rusqlite::Result<Transaction<'_>> {
+        self.inner.transaction().map(Transaction::new)
+    }
+}
+
+#[cfg(not(feature = "explain-query-plans"))]
+impl Connection {
+    #[inline]
+    pub fn prepare_cached(&self, sql: &str) -> rusqlite::Result<rusqlite::CachedStatement<'_>> {
+        self.inner.prepare_cached(sql)
+    }
+
+    #[inline]
+    pub fn execute<P: rusqlite::Params>(&self, sql: &str, params: P) -> rusqlite::Result<usize> {
+        self.inner.execute(sql, params)
+    }
+
+    #[inline]
+    pub fn query_row<T, P, F>(&self, sql: &str, params: P, f: F) -> rusqlite::Result<T>
+    where
+        P: rusqlite::Params,
+        F: FnOnce(&rusqlite::Row<'_>) -> rusqlite::Result<T>,
+    {
+        self.inner.query_row(sql, params, f)
+    }
+}
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -4,7 +4,6 @@ use std::{
     sync::Arc,
 };
 
-use deadpool_sqlite::{Config as SqliteConfig, Hook, HookError, Pool, Runtime};
 use miden_node_proto::{
     domain::account::{AccountInfo, AccountSummary},
     generated::note as proto,
@@ -17,7 +16,6 @@ use miden_objects::{
     transaction::TransactionId,
     utils::Serializable,
 };
-use rusqlite::vtab::array;
 use sql::utils::{column_value_as_u64, read_block_number};
 use tokio::sync::oneshot;
 use tracing::{info, info_span, instrument};
@@ -25,18 +23,25 @@ use tracing::{info, info_span, instrument};
 use crate::{
     blocks::BlockStore,
     config::StoreConfig,
-    db::migrations::apply_migrations,
+    db::{
+        migrations::apply_migrations,
+        pool_manager::{Pool, SqlitePoolManager},
+    },
     errors::{DatabaseError, DatabaseSetupError, GenesisError, NoteSyncError, StateSyncError},
     genesis::GenesisState,
-    COMPONENT, SQL_STATEMENT_CACHE_CAPACITY,
+    COMPONENT,
 };
 
 mod migrations;
+#[macro_use]
 mod sql;
 
+mod connection;
+mod pool_manager;
 mod settings;
 #[cfg(test)]
 mod tests;
+mod transaction;
 
 pub type Result<T, E = DatabaseError> = std::result::Result<T, E>;
 
@@ -194,41 +199,8 @@ impl Db {
             create_dir_all(p).map_err(DatabaseError::IoError)?;
         }
 
-        let pool = SqliteConfig::new(config.database_filepath.clone())
-            .builder(Runtime::Tokio1)
-            .expect("Infallible")
-            .post_create(Hook::async_fn(move |conn, _| {
-                Box::pin(async move {
-                    let _ = conn
-                        .interact(|conn| {
-                            // Feature used to support `IN` and `NOT IN` queries. We need to load
-                            // this module for every connection we create to the DB to support the
-                            // queries we want to run
-                            array::load_module(conn)?;
-
-                            // Increase the statement cache size.
-                            conn.set_prepared_statement_cache_capacity(
-                                SQL_STATEMENT_CACHE_CAPACITY,
-                            );
-
-                            // Enable the WAL mode. This allows concurrent reads while the
-                            // transaction is being written, this is required for proper
-                            // synchronization of the servers in-memory and on-disk representations
-                            // (see [State::apply_block])
-                            conn.execute("PRAGMA journal_mode = WAL;", ())?;
-
-                            // Enable foreign key checks.
-                            conn.execute("PRAGMA foreign_keys = ON;", ())
-                        })
-                        .await
-                        .map_err(|e| {
-                            HookError::Message(format!("Loading carray module failed: {e}").into())
-                        })?;
-
-                    Ok(())
-                })
-            }))
-            .build()?;
+        let sqlite_pool_manager = SqlitePoolManager::new(config.database_filepath.clone());
+        let pool = Pool::builder(sqlite_pool_manager).build()?;
 
         info!(
             target: COMPONENT,
--- a/crates/store/src/db/pool_manager.rs
+++ b/crates/store/src/db/pool_manager.rs
@@ -1,0 +1,64 @@
+use std::path::PathBuf;
+
+use deadpool::Runtime;
+
+use crate::errors::DatabaseError;
+
+deadpool::managed_reexports!(
+    "miden-node-store",
+    SqlitePoolManager,
+    deadpool::managed::Object<SqlitePoolManager>,
+    rusqlite::Error,
+    DatabaseError
+);
+
+use crate::SQL_STATEMENT_CACHE_CAPACITY;
+
+const RUNTIME: Runtime = Runtime::Tokio1;
+
+pub struct SqlitePoolManager {
+    database_path: PathBuf,
+}
+
+impl SqlitePoolManager {
+    pub fn new(database_path: PathBuf) -> Self {
+        Self { database_path }
+    }
+
+    pub fn new_connection(&self) -> rusqlite::Result<Connection> {
+        let conn = Connection::open(&self.database_path)?;
+        let inner = conn.inner();
+
+        // Increase the statement cache size.
+        inner.set_prepared_statement_cache_capacity(SQL_STATEMENT_CACHE_CAPACITY);
+
+        // Enable the WAL mode. This allows concurrent reads while the
+        // transaction is being written, this is required for proper
+        // synchronization of the servers in-memory and on-disk representations
+        // (see [State::apply_block])
+        inner.execute("PRAGMA journal_mode = WAL;", ())?;
+
+        // Enable foreign key checks.
+        inner.execute("PRAGMA foreign_keys = ON;", ())?;
+
+        Ok(conn)
+    }
+}
+
+use deadpool::managed::{Manager, Metrics, RecycleResult};
+
+use crate::db::connection::Connection;
+
+impl Manager for SqlitePoolManager {
+    type Type = deadpool_sync::SyncWrapper<Connection>;
+    type Error = rusqlite::Error;
+
+    async fn create(&self) -> Result<Self::Type, Self::Error> {
+        let conn = self.new_connection();
+        deadpool_sync::SyncWrapper::new(RUNTIME, move || conn).await
+    }
+
+    async fn recycle(&self, _: &mut Self::Type, _: &Metrics) -> RecycleResult<Self::Error> {
+        Ok(())
+    }
+}
--- a/crates/store/src/db/sql/plan_explainer.rs
+++ b/crates/store/src/db/sql/plan_explainer.rs
@@ -106,10 +106,7 @@ impl CachedStatementWithQueryPlan<'_> {
         };
 
         let mut rows = explain_stmt.query(params)?;
-
-        if let Some(sql) = rows.as_ref().and_then(Statement::expanded_sql) {
-            println!("\n>> {sql}");
-        }
+        let expanded_sql = rows.as_ref().and_then(Statement::expanded_sql).unwrap_or_default();
 
         // Build tree from the returned table
         let mut path = OpenedPath {
@@ -129,7 +126,7 @@ impl CachedStatementWithQueryPlan<'_> {
         }
         path.fold_up_to(0);
 
-        println!("{}", path.pop());
+        println!("\n>> {expanded_sql}\n{}", path.pop());
 
         Ok(())
     }
--- a/crates/store/src/db/sql/plan_explainer.rs
+++ b/crates/store/src/db/sql/plan_explainer.rs
@@ -1,5 +1,6 @@
 use rusqlite::{CachedStatement, Params, Row, Rows, Statement};
 use termtree::Tree;
+use tracing::info;
 
 use crate::db::{connection::Connection, transaction::Transaction};
 
@@ -128,7 +129,7 @@ impl CachedStatementWithQueryPlan<'_> {
 
         let query_plan = path.pop().to_string();
 
-        println!("\n>> {expanded_sql}\n\n{query_plan}");
+        info!("\n>> {expanded_sql}\n\n{query_plan}");
 
         #[cfg(test)]
         Self::fail_if_has_problems(&explain_sql, &query_plan);
--- a/crates/store/src/db/sql/plan_explainer.rs
+++ b/crates/store/src/db/sql/plan_explainer.rs
@@ -19,8 +19,6 @@ mod internals {
 mod internals {
     use rusqlite::{CachedStatement, Connection, Params, Row, Rows, Statement};
 
-    use crate::db::sql::utils::print_to_stdout;
-
     pub trait ConnectionHelper {
         fn prepare_cached_opt_explain(
             &self,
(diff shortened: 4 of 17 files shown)


crates/store/src/db/connection.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    #[inline]
+    pub fn query_row<T, P, F>(&self, sql: &str, params: P, f: F) -> rusqlite::Result<T>
+    where
+        P: rusqlite::Params,
+        F: FnOnce(&rusqlite::Row<'_>) -> rusqlite::Result<T>,
+    {
+        self.inner.query_row(sql, params, f)
+    }
+}
```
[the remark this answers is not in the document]
polydez: I just tried to keep the same interface, but your point makes sense, I will use only transactions, thank you!

crates/store/src/db/tests.rs
```diff
@@ -20,18 +20,23 @@ use miden_objects::{
...
-use crate::db::{migrations::apply_migrations, sql::PaginationToken, TransactionSummary};
+use crate::db::{
+    connection::Connection, migrations::apply_migrations, sql::PaginationToken, TransactionSummary,
+};
+
+#[ctor::ctor]
+fn initialize() {
+    miden_node_utils::logging::setup_tracing(miden_node_utils::logging::OpenTelemetry::Disabled)
+        .expect("Failed to setup logging for tests");
+}
```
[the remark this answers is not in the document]
polydez: This is needed for initializing logging in tests. If it was implemented somewhere else, please let me know, I haven't found it by myself.

crates/store/src/db/sql/plan_explainer.rs
```diff
@@ -0,0 +1,161 @@
...
+        self.explain(params.clone())?;
+        self.statement.query_row(params, f)
+    }
+
+    pub fn execute<P: Params + Clone>(&mut self, params: P) -> rusqlite::Result<usize> {
+        self.explain(params.clone())?;
+        self.statement.execute(params)
+    }
+
+    fn explain<P: Params>(&mut self, params: P) -> rusqlite::Result<()> {
```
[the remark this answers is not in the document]
polydez: This method generates query plan, formats it as tree and writes it to log along with an SQL query. We could simplify this by analyzing just rows of query plan returned by `EXPLAIN QUERY PLAN ...` request, without formatting, but this would make logs less human-friendly.

crates/store/src/db/tests.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Our `test-macro` crate I think. But you would need to annotate with

```rust
#[miden_node_test_macro::enable_logging]
```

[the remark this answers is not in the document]
polydez: Thank you! I compared both solutions and must admit that even though mine is simpler and more convenient to use, it doesn't create span for each test as `#[miden_node_test_macro::enable_logging]` does. And our `enable_logging` ruins syntax coloring in my IDE which I would try to avoid. We could use https://github.com/d-e-s-o/test-log instead, but it's also not perfect.

crates/store/src/db/sql/plan_explainer.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Can you provide a before/after example?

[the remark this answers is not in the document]
polydez: It depends on formatting. Current implementation generates tree like:
```
QUERY PLAN
├── SEARCH notes USING INDEX sqlite_autoindex_notes_1 (block_num=?)
├── SCALAR SUBQUERY 3
│   ├── MULTI-INDEX OR
│   │   ├── INDEX 1
│   │   │   ├── LIST SUBQUERY 1
│   │   │   │   └── SCAN rarray VIRTUAL TABLE INDEX 1:
│   │   │   └── SEARCH notes USING INDEX idx_notes_tag (tag=? AND block_num>?)
│   │   └── INDEX 2
│   │       ├── LIST SUBQUERY 2
│   │       │   └── SCAN rarray VIRTUAL TABLE INDEX 1:
│   │       └── SEARCH notes USING INDEX idx_notes_sender (sender=? AND block_num>?)
│   └── USE TEMP B-TREE FOR ORDER BY
├── LIST SUBQUERY 4
│   └── SCAN rarray VIRTUAL TABLE INDEX 1:
└── LIST SUBQUERY 5
    └── SCAN rarray VIRTUAL TABLE INDEX 1:
```

If we get rid of formatting, it would look like:
```
┌────┬────────┬─────────┬─────────────────────────────────────────────────────────────────┐
│ id ┆ parent ┆ notused ┆ detail                                                          │
╞════╪════════╪═════════╪═════════════════════════════════════════════════════════════════╡
│ 3  ┆ 0      ┆ 0       ┆ SEARCH notes USING INDEX sqlite_autoindex_notes_1 (block_num=?) │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 6  ┆ 0      ┆ 0       ┆ SCALAR SUBQUERY 3                                               │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 16 ┆ 6      ┆ 0       ┆ SEARCH notes USING INDEX sqlite_autoindex_notes_1 (block_num>?) │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 25 ┆ 6      ┆ 0       ┆ LIST SUBQUERY 1                                                 │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 27 ┆ 25     ┆ 0       ┆ SCAN rarray VIRTUAL TABLE INDEX 1:                              │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 45 ┆ 6      ┆ 0       ┆ LIST SUBQUERY 2                                                 │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 47 ┆ 45     ┆ 0       ┆ SCAN rarray VIRTUAL TABLE INDEX 1:                              │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 73 ┆ 0      ┆ 0       ┆ LIST SUBQUERY 4                                                 │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 75 ┆ 73     ┆ 0       ┆ SCAN rarray VIRTUAL TABLE INDEX 1:                              │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 93 ┆ 0      ┆ 0       ┆ LIST SUBQUERY 5                                                 │
├╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 95 ┆ 93     ┆ 0       ┆ SCAN rarray VIRTUAL TABLE INDEX 1:                              │
└────┴────────┴─────────┴─────────────────────────────────────────────────────────────────┘
```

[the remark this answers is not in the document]
Mirko-von-Leipzig: I see okay so its mimicking the format of `sqlite3` by building a query tree and formatting it.

Its a bit confusing because the `Tree` type is recursive and we're assuming that `id` is strictly incrementing I think? Would be good to either comment this extensively, or find a simpler abstraction.

I know its not optimal (but this doesn't need to be), but what about something like:

```rust
// Only required so we can implement `Display` for it.
// Also a bad name but meh.
struct Element(u64, String)

struct QueryTree(Tree<Element>)

impl QueryTree {
    fn new() -> Self { Self(Tree::new(Element(0, "QUERY PLAN")) }

    fn insert(&mut self, parent: u64, element: Element) {
        // Recursively search through all elements in tree already
        // and insert new element as a child there.
    }
}
```

crates/store/src/db/sql/plan_explainer.rs
```diff
@@ -0,0 +1,161 @@
...
+enum Executor<'conn> {
+    Connection(&'conn rusqlite::Connection),
+    Transaction(&'conn rusqlite::Transaction<'conn>),
+}
+
+pub struct CachedStatementWithQueryPlan<'conn> {
+    executor: Executor<'conn>,
+    sql: Box<str>,
+    statement: CachedStatement<'conn>,
+}
```
Mirko-von-Leipzig: We should be able to get rid of this if we only support `Transaction` right?

I'm thinking it just becomes `Transaction::check_query_plan(&self, sql: &str)` which is selectively enabled by `#[cfg(test)]`.

crates/store/src/db/sql/plan_explainer.rs
```diff
@@ -0,0 +1,161 @@
...
+            }
+        }
+
+        let explain_sql = format!("EXPLAIN QUERY PLAN {}", self.sql);
+        let mut explain_stmt = match self.executor {
+            Executor::Connection(conn) => conn.prepare(&explain_sql)?,
+            Executor::Transaction(transaction) => transaction.prepare(&explain_sql)?,
+        };
+
+        let mut rows = explain_stmt.query(params)?;
```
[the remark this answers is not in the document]
polydez: Good idea, this is not supported by ordinary `query` (we have to fill all params or get "invalid parameter count" error), but it could work for `raw_query`, I will check it.

crates/store/src/db/sql/plan_explainer.rs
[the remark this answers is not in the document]
polydez: I think, it's not that bad to rely on the current row ordering (we will explain this in comments), since it's not a critical feature, but needed only for debugging, and such implementation is simple and fast. And the official SQLIite CLI relies on this ordering as well, it's natural in terms of query plan generation and unlikely to be changed.

clippy.toml
Mirko-von-Leipzig, line 1: Do you know if this can be added to the workspace toml lints section instead?

crates/store/src/db/transaction.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+impl Transaction<'_> {
+    #[inline]
+    pub fn prepare_cached(&self, sql: &str) -> rusqlite::Result<rusqlite::CachedStatement<'_>> {
+        #[cfg(feature = "explain-query-plans")]
+        self.explain_query_plan(sql)?;
+
+        self.inner.prepare_cached(sql)
+    }
```
Mirko-von-Leipzig, line 40: I don't think there is a need to enable this on a production node. Or rather I don't think there will ever be a valid time to do so.

I think we can do this purely for tests which should cover all queries/statements.

I also think the check _should not_ print anything by default.

```suggestion
    pub fn prepare_cached(&self, sql: &str) -> rusqlite::Result<rusqlite::CachedStatement<'_>> {
       // We do this first so we get invalid sql errors here
       // instead of as a panic in the query plan check for tests.
       let stmt = self.inner.prepare_cached(sql)?;
    
        #[cfg(test)]
        self.check_query_plan(sql);

        Ok(stmt)
    }
```

where we only print the query plan tree on failure:

```rust
#[cfg(test)]
impl Transaction<'_> {
    /// Panics if the query plan contains a slow table scan.
    fn check_query_plan(&self, sql: &str) {
        let query_plan = self.query_plan(sql).expect("valid sql");

        assert!(query_plan.contains_scan().not(), "Query plan contains a table scan: \n{query_plan});
    }
}
```

crates/store/src/db/query_plan/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+        Ok(())
+    }
+}
+
+/// Checks if the given query plan contains unnecessary table scans not utilizing indices.
+#[cfg(test)]
+fn query_plan_contains_unnecessary_scans(sql: &str, query_plan: &str) -> bool {
+    use std::sync::LazyLock;
+
+    static RE_IDX_SCAN: LazyLock<regex::Regex> = LazyLock::new(|| {
```
Mirko-von-Leipzig, line 29: Could you document which cases this is ignoring?

clippy.toml
Mirko-von-Leipzig: Hah. I see. Yeah I just found a discussion where the cargo team just says "we'll ignore it".

Thanks.

polydez:
@Mirko-von-Leipzig I think, I've addressed all review comments for this PR here: https://github.com/0xPolygonMiden/miden-node/pull/733
Please, take a look!

Mirko-von-Leipzig (approved):
I believe this is ready now, though let's wait for @bobbinth to take a look as well.

As a summary of where we are.

- Tests panic if a sql query plan contains a full table scan i.e. doesn't use an index. This is automatically injected for all queries under store's `#[cfg(test)]`.
- Added missing indexes this turned up.
- This infra required our own transaction type which in turn meant our own connection pool.
- Lower sql fn's now take transaction instead of connection.

This last point we should take a step further by changing `sql::xxx` from being a free function, to a method on `Transaction`. This is now possible because we own the type. This is out of scope for this PR though, I'll create a follow-up issue once this is ok'd.

crates/store/src/db/mod.rs
```diff
@@ -406,7 +407,10 @@ impl Db {
             .get()
             .await
             .map_err(DatabaseError::MissingDbConnection)?
-            .interact(move |conn| sql::get_note_sync(conn, block_num, &note_tags))
+            .interact(move |conn| {
+                let transaction = conn.transaction().map_err(DatabaseError::SqliteError)?;
```
SantiagoPittella, line 411: This `.map_err` breaks a bit the pattern used in the previous methods (`conn.transaction()?`). Is there a reason why this methods returns an explicit error?

crates/store/src/db/pool_manager.rs
SantiagoPittella, line 1: The imports are scattered throughout the file, it would be more consistent with our formatting to group them all at the begging of the file.

crates/store/src/db/query_plan/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    /// Renders query plan as tree using ASCII graphics.
+    fn query_plan(&self, sql: &str) -> rusqlite::Result<QueryPlan> {
+        let explain_sql = format!("EXPLAIN QUERY PLAN {sql}");
+        let explain_stmt = self.prepare(&explain_sql)?;
+        let query_plan = QueryPlanRenderer::new().render_tree(explain_stmt)?;
+
+        Ok(QueryPlan { explain_sql, query_plan })
+    }
+}
```
SantiagoPittella, line 29: This impl might be better to have on the same file as the definition of the struct.

crates/store/src/db/query_plan/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+
+struct QueryPlan {
+    explain_sql: String,
+    query_plan: String,
+}
```
SantiagoPittella, line 34: This struct is missing documentation.

crates/store/src/db/query_plan/renderer.rs
```diff
@@ -0,0 +1,60 @@
+use rusqlite::Statement;
+use termtree::Tree;
+
+pub struct QueryPlanRenderer {
+    path: Vec<(u64, Tree<String>)>,
+}
```
SantiagoPittella, line 6: Missing docs.

crates/store/src/db/query_plan/renderer.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// Pushes tree element to the top of the path stack.
+    fn push(&mut self, id: u64, element: Tree<String>) {
+        self.path.push((id, element));
+    }
+
+    /// Pops element from the top of the path stack.
+    fn pop(&mut self) -> Tree<String> {
+        self.path.pop().expect("Stack must contain at least root node").1
+    }
```
SantiagoPittella, line 47: This methods looks a bit unnecessary.

crates/store/src/db/query_plan/mod.rs
polydez: These methods implemented conditionally for `#[cfg(test)]` as well, as `QueryPlan`, so it was more convenient and clean to put them into separate module. But I agree that sometimes it easier to find all implementations in a single module. We don't have coding guidelines for this case yet, so we probably need to discuss it.

polydez:
@SantiagoPittella I've addressed review comments in different PR, please take a look: https://github.com/0xPolygonMiden/miden-node/pull/747

crates/store/src/db/query_plan/mod.rs
Mirko-von-Leipzig: I'm happy to have them split by functionality where sensible.

We can revisit this when `Transaction` gets more of the sql methods moved into it.

SantiagoPittella (approved):
LGTM 👍🏼

PR #730, opened 2025-03-06
chore: upgrade rand version

### LINKED ISSUE #729: Upgrade `rand` version
Since upgrading to Rust 2024 edition in https://github.com/0xPolygonMiden/miden-node/pull/727, the `rand::Rng::gen()` calls conflicts with the new rust `gen` keyword, so we should upgrade rand version and replace these calls with [random()](https://docs.rs/rand/latest/rand/trait.Rng.html#method.random).

TomasArrachea:
Closes #729 

Updates rand version to 0.9. Also run `cargo update` and updated the `Cargo.lock.`

Also includes this PR https://github.com/0xPolygonMiden/miden-node/pull/757


bobbinth:
@TomasArrachea @igamigo - once you figure out how to make it work here, would be cool to make a similar upgrade in [miden-crypto](https://github.com/0xPolygonMiden/crypto). Probably would involve migrating `miden-crypto` to Rust 2024 as well.

bobbinth (approved):
Looks good! Thank you!

PR #738, opened 2025-03-14
Mirko/batch instrumentation

### LINKED ISSUE #679: Instrument batch building
Instrument batch building in a similar fashion to the block building in #675.

I recommend starting by defining the stages of the pipeline as you would want to explore them. As in, what timing information are you interested in - this defines your top-level spans. Likely this won't quite fit with the code; iterate until you are at a reasonable compromise between code clarity and trace generation.

Mirko-von-Leipzig:
This PR instruments batch production.

The span tree (excluding the gRPC switch to store):

```sh
batch_builder.build_batch
┝━ batch_builder.wait_for_available_worker
┝━ batch_builder.select_batch
│  ┝━ mempool.lock
│  ┕━ mempool.select_batch
┝━ batch_builder.get_batch_inputs
│  ┕━ store.client.get_batch_inputs
┝━ batch_builder.propose_batch
┝━ batch_builder.prove_batch
┝━ batch_builder.inject_failure
┕━ batch_builder.commit_batch
   ┝━ mempool.lock
   ┕━ mempool.batch_proved
```

Closes #679.

### DIFF
--- a/crates/block-producer/src/batch_builder/mod.rs
+++ b/crates/block-producer/src/batch_builder/mod.rs
@@ -29,6 +29,12 @@ use crate::{
 /// of provers for proof generation. Proving is currently unimplemented and is instead simulated via
 /// the given proof time and failure rate.
 pub struct BatchBuilder {
+    /// Represents all batch building workers.
+    ///
+    /// This pool is always at maximum capacity. Idle workers will be in a [`std::future::Ready`]
+    /// state and are immedietely available for a new batch building job.
+    ///
+    /// See also: [`BatchBuilder::wait_for_available_worker`].
     worker_pool: JoinSet<()>,
     batch_interval: Duration,
     /// The batch prover to use.
@@ -43,14 +49,21 @@ pub struct BatchBuilder {
 }
 
 impl BatchBuilder {
-    /// Creates a new [`BatchBuilder`] with the given batch prover URL.
+    /// Creates a new [`BatchBuilder`] with the given batch prover URL and maximum concurrent batch
+    /// building workers.
     ///
-    /// If no URL is provided, a local batch prover is used.
-    pub fn new(store: StoreClient, workers: NonZeroUsize, batch_prover_url: Option<Url>) -> Self {
+    /// If no batch prover URL is provided, a local batch prover is used instead.
+    pub fn new(
+        store: StoreClient,
+        num_workers: NonZeroUsize,
+        batch_prover_url: Option<Url>,
+    ) -> Self {
         let batch_prover = batch_prover_url
             .map_or(BatchProver::local(MIN_PROOF_SECURITY_LEVEL), BatchProver::remote);
 
-        let worker_pool = std::iter::repeat_n(std::future::ready(()), workers.get()).collect();
+        // It is important that the worker pool is filled to capacity with ready workers. See
+        // `Self::worker_pool` and `Self::wait_for_available_worker` for more context.
+        let worker_pool = std::iter::repeat_n(std::future::ready(()), num_workers.get()).collect();
 
         Self {
             batch_interval: SERVER_BUILD_BATCH_FREQUENCY,
@@ -101,8 +114,21 @@ impl BatchBuilder {
             .spawn(async move { job.build_batch().await }.instrument(tracing::Span::current()));
     }
 
+    /// Waits for a new batch building worker to become available.
+    ///
+    /// The worker pool is _always_ at full capacity because:
+    ///   - It is instantiated with a full cohort of [`std::future::ready()`].
+    ///   - This function removes a worker but it's _always_ added afterwards again in
+    ///     `build_batch`, keeping the pool at capacity.
+    ///
+    /// An alternate implementation might instead check the currently active jobs, but this would
+    /// require the same logic as here to handle the case when the pool is at capacity. This
+    /// design was chosen instead as it removes this branching logic by "always" having the pool
+    /// at max capacity. Instead completed workers wait to be culled by this function.
     #[instrument(target = COMPONENT, name = "batch_builder.wait_for_available_worker", skip_all)]
     async fn wait_for_available_worker(&mut self) {
+        // We must crash here because otherwise we have a batch that has been selected from the
+        // mempool, but which is now in limbo. This effectively corrupts the mempool.
         if let Err(crash) = self.worker_pool.join_next().await.expect("worker pool is never empty")
         {
             tracing::error!(message=%crash, "Batch worker pool panic'd");
@@ -111,6 +137,15 @@ impl BatchBuilder {
     }
 }
 
+// BATCH JOB
+// ================================================================================================
+
+/// Represents a single batch building job.
+///
+/// It is entirely self-contained and performs the full batch creation flow, from selecting the
+/// batch from the [`Mempool`] up to and including submitting the results back to the [`Mempool`].
+///
+/// Errors are also handled internally and are not propagated up.
 struct BatchJob {
     /// Simulated block failure rate as a percentage.
     ///
--- a/crates/block-producer/src/batch_builder/mod.rs
+++ b/crates/block-producer/src/batch_builder/mod.rs
@@ -11,7 +11,7 @@ use miden_proving_service_client::proving_service::batch_prover::RemoteBatchProv
 use miden_tx_batch_prover::LocalBatchProver;
 use rand::Rng;
 use tokio::{task::JoinSet, time};
-use tracing::{info, instrument, Instrument, Span};
+use tracing::{instrument, Instrument, Span};
 use url::Url;
 
 use crate::{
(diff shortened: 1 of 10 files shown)


crates/block-producer/src/lib.rs
```diff
@@ -16,21 +16,24 @@ pub mod server;
...
-/// The frequency at which batches are built
+/// The frequency at which batches are built.
 const SERVER_BUILD_BATCH_FREQUENCY: Duration = Duration::from_secs(2);
 
-/// Maximum number of batches per block
+/// Maximum number of batches per block.
 const SERVER_MAX_BATCHES_PER_BLOCK: usize = 4;
 
+/// Size of the batch building worker pool.
+const SERVER_BATCH_BUILDERS: NonZeroUsize = NonZeroUsize::new(2).unwrap();
```
bobbinth, line 35: nit: I would name this something like `NUM_SERVER_BATCH_BUILDERS` to make it clear that this is a number.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -67,280 +66,246 @@ impl BatchBuilder {
...
-                    .task_map
-                    .iter()
-                    .find(|(task_id, _)| &join_err.id() == task_id)
-                    .expect("Task ID should be in the task map")
-                    .1;
-
-                (batch_id, join_err.into())
-            })
-            .and_then(|x| x);
+struct BatchJob {
```
bobbinth, line 114: nit: let's add a section separator above this line.

Also, I would probably add some doc comments explaining the purpose/lifecycle of a `BatchJob`.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -67,280 +66,246 @@ impl BatchBuilder {
...
+            BatchProver::Remote(prover) => {
+                prover.prove(proposed_batch).await.map_err(BuildBatchError::RemoteProverError)
+            },
+            BatchProver::Local(prover) => tokio::task::spawn_blocking({
+                let prover = prover.clone();
+                move || prover.prove(proposed_batch).map_err(BuildBatchError::ProveBatchError)
+            })
+            .await
+            .map_err(BuildBatchError::JoinError)?,
+        }
```
bobbinth, line 213: Not for this PR, but we should make sure the returned proof matches the required security level. This is not an issue for local provers (since we set the security level explicitly), but does need to be checked for remote provers (even if only to avoid misconfiguration errors).

Let's add an issue for this.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -27,38 +29,35 @@ use crate::{
...
 impl BatchBuilder {
     /// Creates a new [`BatchBuilder`] with the given batch prover URL.
     ///
     /// If no URL is provided, a local batch prover is used.
-    pub fn new(batch_prover_url: Option<Url>) -> Self {
-        let batch_prover = match batch_prover_url {
-            Some(url) => BatchProver::new_remote(url),
-            None => BatchProver::new_local(MIN_PROOF_SECURITY_LEVEL),
-        };
+    pub fn new(store: StoreClient, workers: NonZeroUsize, batch_prover_url: Option<Url>) -> Self {
```
bobbinth, line 49: nit: I would rename `workers` to `num_workers`.

Also, maybe it is worth updating the doc comments for this function.

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -67,280 +66,246 @@ impl BatchBuilder {
...
-}
 
-// BATCH WORKER
-// ================================================================================================
+    #[instrument(parent = None, target = COMPONENT, name = "batch_builder.build_batch", skip_all)]
+    async fn build_batch(&mut self, mempool: SharedMempool) {
+        Span::current().set_attribute("workers.count", self.worker_pool.len());
 
-type BatchResult = Result<ProvenBatch, (BatchId, BuildBatchError)>;
+        self.wait_for_available_worker().await;
```
bobbinth, line 91: Could we add a brief comment (either here or as doc comments for the function) explaining how this pattern works? My understanding is that we have a set of dummy tasks in `self.worker_pool`. Here, we check if any of these tasks is "free", and the on lines 100 - 101 below we use the free task run the new batch job?

I'm assuming `wait_for_available_worker()` removes a task from the `JoinSet` while `worker_pool.spawn()` adds a new task to it. Is this correct?

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -27,38 +29,35 @@ use crate::{
...
 pub struct BatchBuilder {
-    pub batch_interval: Duration,
-    pub workers: NonZeroUsize,
-    /// Used to simulate batch proving by sleeping for a random duration selected from this range.
-    pub simulated_proof_time: Range<Duration>,
-    /// Simulated block failure rate as a percentage.
-    ///
-    /// Note: this _must_ be sign positive and less than 1.0.
-    pub failure_rate: f32,
+    worker_pool: JoinSet<()>,
```
bobbinth, line 32: Related to the previous comment, it may be good to add a brief explanation (either here or as doc comments for the struct) of the pattern we use with the `worker_pool`.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - they are mostly doc-related.

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: Correct. I'll add a comment.

A more conventional pool impl might do something like:

```rust
if pool.workers.len() < pool.max_workers() {
    pool.spawn(...);
} else {
    pool.wait_for_job_completion().await;
    pool.spawn(...);
}
```

So I figured instead of branching we instead just do the `else` branch always. By initializing the pool with `max` "jobs" which are immediately `future::ready()`.

crates/block-producer/src/lib.rs
bobbinth: It is, but that's how we usually name "countable" variables in most other places.

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: #744

Mirko-von-Leipzig:
@igamigo do you still want to review this one?

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -27,38 +29,48 @@ use crate::{
...
-    pub workers: NonZeroUsize,
-    /// Used to simulate batch proving by sleeping for a random duration selected from this range.
-    pub simulated_proof_time: Range<Duration>,
-    /// Simulated block failure rate as a percentage.
+    /// Represents all batch building workers.
     ///
-    /// Note: this _must_ be sign positive and less than 1.0.
-    pub failure_rate: f32,
+    /// This pool is always at maximum capacity. Idle workers will be in a [`std::future::Ready`]
+    /// state and are immedietely available for a new batch building job.
```
igamigo, line 35: typo: 
```suggestion
    /// state and are immediately available for a new batch building job.
```

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -27,38 +29,48 @@ use crate::{
...
+    ///
+    /// See also: [`BatchBuilder::wait_for_available_worker`].
+    worker_pool: JoinSet<()>,
+    batch_interval: Duration,
     /// The batch prover to use.
     ///
     /// If not provided, a local batch prover is used.
     batch_prover: BatchProver,
+    /// Simulated block failure rate as a percentage.
```
igamigo, line 44: nit: I know it was already like this (assertion message mentioned this already) but it wouldn't be a percentage if it's within 0 and 1, no?

docs/src/operator/monitoring.md
```diff
@@ -74,7 +74,26 @@ block_builder.build_block
...
+│  ┝━ mempool.lock
+│  ┕━ mempool.select_batch
+┝━ batch_builder.get_batch_inputs
+│  ┕━ store.client.get_batch_inputs
+┝━ batch_builder.propose_batch
+┝━ batch_builder.prove_batch
+┝━ batch_builder.inject_failure
+┕━ batch_builder.commit_batch
+   ┝━ mempool.lock
+   ┕━ mempool.batch_proved
```
igamigo, line 95: Should this be `mempool.commit_batch` (and should it mention the option of `rollback_batch`)?

igamigo (approved):
LGTM! Sorry, had taken a look yesterday but did not finish reviewing. I left three quite irrelevant comments, feel free to disregard. Instrumentation-wise it looks great.

crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig: You can have more than 100% though in this context it makes no sense. I agree its a bit overkill :)

docs/src/operator/monitoring.md
Mirko-von-Leipzig: Ah it should, I updated it afterwards - thank you.

Re: `rollback_batch` - I'm not sure how to really document the failure paths since it can happen at any stage.

PR #740, opened 2025-03-14
feat: endpoints for network txs

### LINKED ISSUE #695: gRPC methods for network transactions
See #692 for full context.

Define a new `network-transaction` service with two methods:

1. Receive new network note (can just be a plain note for now)
2. Receive status update for network transaction (`tx_id` + `Status::Committed | Status::Failed`). Consider including block height and error reason respectively.

The new component also needs access to the store and block-producer APIs but only requires one additional `store` method to retrieve existing network notes on startup.

Consider making this new method a stream - could be good to see how it fits in, considering we want to use it elsewhere potentially.

Somewhat blocked by
- #696

TomasArrachea:
Closes #695.

This PR adds the proto definitions for a new network-transaction-builder service (not the actual implementation). This API has two endpoints:
- `CreateNetworkTransaction`
- `UpdateNetworkTransaction`

Aditionally, a new method was implemented in the RPC Store API:
- `GetUnconsumedNetworkNotes`

Roughly, the flow for network notes execution is:
1. The mempool component will detect when a transaction produces a new network note. 
2. Then, it will send a `SubmitNetworkNote` request to the ntx-builder. It passes as parameters the new network notes and the id of the tx that created them.
3. The ntx-builder will trigger the transactions execution.
4. Eventually, when the mempool generates a block, it will update multiple transactions status to "commited" using the `UpdateTransactionStatus` endpoint. It passes as parameter a list of the transaction ids to update, with it's new status. The ntx-builder will filter the transactions to update.


proto/requests.proto
```diff
@@ -188,3 +189,29 @@ message GetAccountProofsRequest {
...
+    digest.Digest transaction_id = 2;
+}
+
+message UpdateNetworkNoteRequest {
+    // Transaction id of the note to update.
+    digest.Digest transaction_id = 1;
+
+    // New status of the note.
+    note.NetworkNoteStatus status = 2;
+}
```
bobbinth, line 217: I may have missed something - but it is not clear to me how we plan to use these requests.

crates/store/src/state.rs
```diff
@@ -985,6 +986,15 @@ impl State {
     pub async fn optimize_db(&self) -> Result<(), DatabaseError> {
         self.db.optimize().await
     }
+
+    /// Returns the unprocessed network notes, along with the next pagination token.
+    pub async fn get_unconsumed_network_notes(
+        &self,
+        page: PaginationToken,
+        limit: NonZero<usize>,
```
bobbinth, line 994: Do we ever use `page` separately from `limit`? If not, it may make sense to combine them into one struct.

crates/store/src/db/sql/mod.rs
```diff
@@ -944,14 +943,24 @@ pub fn unconsumed_network_notes(
...
-        token.0 = row.get::<_, i64>(11)? + 1;
+        token.0 = row.get::<_, u64>(11)? + 1;
     }
 
     Ok((notes, token))
 }
 
 #[derive(Default, Debug, Copy, Clone)]
-pub struct PaginationToken(i64);
+pub struct PaginationToken(u64);
```
bobbinth, line 953: nit: this seems like a pretty general struct. I'd probably move it to the top of the file and add doc comments for it.

bobbinth (commented):
Thank you! Not a review yet, but I left some comments inline.

Also, it would be helpful to add a PR description explaining what we are aiming to do in this PR - otherwise, it is not clear what the intent of some changes is (e.g., what is the purpose of `network_transaction.proto`).

proto/network_transaction.proto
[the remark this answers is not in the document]
TomasArrachea: Yes, this would define the service's API with those two methods: create network tx and update network tx status. I renamed it from `network-transaction` to `network-transaction-builder`.

bobbinth:
> Roughly, the flow for network notes execution is:
> 
> 1. The mempool component will detect when a transaction consumes a new network note.
> 2. Then, it will send a `CreateNetworkTransaction` request to the network-transaction-builder to create the transaction. It passes as parameters the network note and the id of the tx that created it.
> 3. The network-transaction-builder will trigger the transaction execution.
> 4. Eventually, when the mempool generates a batch, it will update multiple network-transactions status to "commited" using the `UpdateNetworkTransaction` endpoint. It passes as parameter a list of the network tx ids to update, with it's new status.

I haven't looked at the code too much, but I don't think this flow is correct and so the some of the endpoints introduced in this PRs don't seem to be correct either. More specific comments:

- The mempool component needs to detect when a transaction produces a new network note (not consumes it).
- The mempool does not ask the network-transaction-builder to create new transactions - the transaction builder will need to decide when to create transactions based on its own logic.
- Once a block (not a batch) is committed, the block producer needs to notify the transaction builder of all committed transactions (not just network transactions).

So, instead of `CreateNetworkTransaction` we need `SubmitNetworkNotes` or (`ReceiveNetworkNotes`). This endpoint will receive a list of network notes together with transaction ID that created them - so, the request part will be close to what you have for `CreateNetworkTransaction`:

```proto
message SubmitNetworkNotesRequest {
    digest.Digest transaction_id = 1;
    repeated note.Note notes = 2;
}
```
This endpoint shouldn't really return anything (except for errors if they were encountered).

Instead of `UpdateNetworkTransaction` we need something like `UpdateTransactionStatus`. The request for this endpoint will be basically identical to `UpdateNetworkTransaction`:
```proto
message UpdateTransactionStatusRequest {

    message TransactionStatusUpdate {
        digest.Digest transaction_id = 1;
        transaction.TransactionStatus status = 2;
    }

    repeated TransactionStatusUpdate updates = 1;
}
```
But again, the response here wouldn't return anything (except for errors, when relevant).

cc @Mirko-von-Leipzig in case I'm missing anything.

proto/network_transaction_builder.proto
bobbinth, line 1: `network-transaction-builder` is a bit of a mouthful to write. Should we abbreviate it somehow? Brainstorming some options:

- `ntb` - short, but unclear (IMHO).
- `network-tx-builder`
- `ntx-builder` - I kind of like this one.

Mirko-von-Leipzig: - `netx-builder` - looks too confusing :D 

I personally like `ntx-builder` best so far, though `network-tx-builder` is probably the best of the lot ito clarity.

Given that this will (likely) be its own crate, the `miden-node-` prefix is really getting painful. Though we could rename them internally at least to remove the prefix in the root toml.

TomasArrachea: Ok! `ntx-builder` sounds good for me.

proto/proto/ntx_builder.proto
```diff
@@ -0,0 +1,17 @@
+// Specification of the ntx-builder RPC.
+//
+// This provided access to the rollup data to the other nodes.
```
bobbinth, line 3: Not sure if the comment on line 3 here is correct. Also, on line 1 we could probably write out "Network Transaction Builder RPC".

proto/proto/ntx_builder.proto
```diff
@@ -0,0 +1,17 @@
...
+syntax = "proto3";
+package ntx_builder;
+
+import "requests.proto";
+import "responses.proto";
+import "google/protobuf/empty.proto";
+
+service Api {
+    // Submit a list of network notes to create network transactions.
+    rpc SubmitNetworkNotes(requests.SubmitNetworkNotesRequest) returns (google.protobuf.Empty) {}
```
bobbinth, line 13: I would write the comment as "Submit a list of network notes to the network transaction builder."

proto/proto/ntx_builder.proto
```diff
@@ -0,0 +1,17 @@
...
+
+    // Update the status of multiple transactions.
+    rpc UpdateTransactionStatus(requests.UpdateTransactionStatusRequest) returns (google.protobuf.Empty) {}
```
bobbinth, line 16: I would write the comment as "Update network transaction builder with transaction status changes."

proto/proto/requests.proto
```diff
@@ -188,3 +190,36 @@ message GetAccountProofsRequest {
...
+message GetUnconsumedNetworkNotesRequest {
+    // Page number to retrieve.
+    uint64 page = 1;
+
+    // Number of notes to retrieve per page.
+    uint64 limit = 2;
+}
+
+// Submit multiple network notes to create network transactions.
+message SubmitNetworkNotesRequest {
```
bobbinth, line 204: Similar to one of the above comments.

proto/proto/requests.proto
```diff
@@ -188,3 +190,36 @@ message GetAccountProofsRequest {
...
+message SubmitNetworkNotesRequest {    
+    // Id of the transaction that created the notes.
+    digest.Digest transaction_id = 1;
+
+    // The network notes to submit.
+    repeated note.Note note = 2;
+}
+
+// Updates the status of multiple transactions.
+message UpdateTransactionStatusRequest {
```
bobbinth, line 213: Similar to one of the above comments.

crates/store/src/db/sql/mod.rs
```diff
@@ -42,6 +42,36 @@ use crate::{
...
+
+    /// Returns the page size.
+    pub fn limit(&self) -> NonZeroUsize {
+        self.limit
+    }
+
+    /// Sets the next page number.
+    pub fn set_next(&mut self, next: u64) {
+        self.number = next;
+    }
```
bobbinth, line 72: Should this be more like:
```rust
pub fn next(&self) -> Self {
    Self {
        number: self.number + self.limit,
        limit: self.limit,
    }
}
```

crates/store/src/db/sql/mod.rs
```diff
@@ -42,6 +42,36 @@ use crate::{
     },
     errors::{DatabaseError, NoteSyncError, StateSyncError},
 };
+
+/// The page number and limit to query from the DB.
+#[derive(Debug, Copy, Clone)]
+pub struct Page {
+    number: u64,
+    limit: NonZeroUsize,
+}
```
bobbinth, line 51: nit: maybe `size` would be a better name than `limit`?

crates/store/src/server/api.rs
```diff
@@ -519,6 +524,35 @@ impl api_server::Api for StoreApi {
...
+    ) -> Result<Response<GetUnconsumedNetworkNotesResponse>, Status> {
+        let request = request.into_inner();
+        let state = self.state.clone();
+
+        let (notes, next_page) = state
+            .get_unconsumed_network_notes(Page::new(
+                request.page,
+                NonZero::try_from(request.limit as usize).map_err(|err: TryFromIntError| {
+                    invalid_argument(format!("Invalid limit: {err}"))
+                })?,
```
bobbinth, line 546: nit: I'd maybe move this code above line 541 to make things a bit more readable.

bobbinth (commented):
Looks good! Thank you! I left some comments inline - most are pretty minor.

crates/store/src/db/sql/mod.rs
TomasArrachea: This would make sense if we were making the query with OFFSET, but we are using the rowid in a WHERE clause (I think it's faster this way). For some reason the rowids are being incremented by 2 so this wouldn't work. I left the original logic where the last rowid is used to index the next page.

bobbinth: Ah yes - good point!

proto/proto/responses.proto
```diff
@@ -252,3 +252,12 @@ message StorageSlotMapProof {
...
 }
+
+// Represents the result of getting the unconsumed network notes.
+message GetUnconsumedNetworkNotesResponse {
+    // The next page number.
+    uint64 next_page = 1;
+
+    // The list of unconsumed network notes.
+    repeated note.Note notes = 2;
+}
```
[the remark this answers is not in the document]
TomasArrachea: Yes, I think using this cursor-based pagination with an opaque token makes more sense since it's actually the way we are querying the db, i changed it in [fd12eec](https://github.com/0xPolygonMiden/miden-node/pull/740/commits/fd12eec24feb8efb261a3441e910bdef0b1b8c43)

crates/store/src/db/sql/mod.rs
```diff
@@ -947,22 +953,19 @@ pub fn unconsumed_network_notes(
...
+    let mut rows = stmt.query(params![page.token.unwrap_or(0), page.size])?;
 
-    let mut notes = Vec::with_capacity(limit.into());
+    let mut notes = Vec::with_capacity(page.size.into());
     while let Some(row) = rows.next()? {
         notes.push(NoteRecord::from_row(row)?);
         // Increment by 1 because we are using rowid >=, and otherwise we would include the last
         // element in the next page as well.
-        token.0 = row.get::<_, i64>(11)? + 1;
+        page.token = Some(row.get::<_, u64>(11)? + 1);
```
Mirko-von-Leipzig, line 963: I believe you should restructure this slightly for the now optional return token. Currently the client would have to make one final call that will return no data.

There are a few options to achieve this. The simplest is to set `LIMIT - page.size + 1`, limiting the loop to `page.size`, and then checking whether there is another row to determine the return token's Someness.

The alternative is to do an additional `EXISTS` query using the last row ID but I think that's more convoluted for no good reason.

proto/proto/requests.proto
```diff
@@ -188,3 +190,36 @@ message GetAccountProofsRequest {
     // all requested accounts.
     repeated digest.Digest code_commitments = 3;
 }
+
+// Returns a list of unconsumed network notes using pagination.
+message GetUnconsumedNetworkNotesRequest {
+    // Page token to retrieve.
+    optional uint64 page_token = 1;
```
Mirko-von-Leipzig, line 197:
```suggestion
    // An opaque token used to paginate through the notes.
    //
    // This should be null on the first call, and set to the response
    // token until the response token is null, at which point all data
    // has been fetched.
    optional uint64 continuation_token = 1;
```

proto/proto/responses.proto
```diff
@@ -252,3 +252,12 @@ message StorageSlotMapProof {
     // Merkle proof of the map value
     bytes smt_proof = 2;
 }
+
+// Represents the result of getting the unconsumed network notes.
+message GetUnconsumedNetworkNotesResponse {
+    // The next page token.
+    optional uint64 next_token = 1;
```
Mirko-von-Leipzig, line 259:
```suggestion
    // An opaque pagination token.
    //
    // Use this in your next request to get the next
    // set of data.
    //
    // Will be null once there is no more data remaining.
    optional uint64 next_token = 1;
```

crates/store/src/db/sql/mod.rs
TomasArrachea: Right!

bobbinth (approved):
Looks good! Thank you!

PR #750, opened 2025-03-19
feat: block and batch frequency can now be configured

### LINKED ISSUE #735: feat: allow to configure block and batch frequency
One of the ways we found to speed up the integration tests in the client was to increase the frequency of blocks in the instantiated local node. We had to add a command to edit these values but we could have a better way via a config file or env vars.  

> see if we could add a way to configure this with env vars or something similar?

_Originally posted by @igamigo in https://github.com/0xPolygonMiden/miden-client/pull/763#discussion_r1989876677_

Mirko-von-Leipzig:
Closes #735


bin/node/Cargo.toml
```diff
@@ -20,6 +20,7 @@ tracing-forest = ["miden-node-block-producer/tracing-forest"]
 [dependencies]
 anyhow                    = { version = "1.0" }
 clap                      = { version = "4.5", features = ["derive", "env", "string"] }
+humantime                 = { version = "2.2" }
```
Mirko-von-Leipzig, line 23: I'm unsure if its worth adding this. It lets you use `10h 5m 3s`. The alternative is just keeping normal `Duration` and documenting that its an integer in milliseconds.

igamigo: Not a strong opinion at all but given the choice I'd keep milliseconds as a fixed unit instead

bobbinth: Yes, I'd probably just keep it as milliseconds.

crates/block-producer/src/server.rs
```diff
@@ -57,6 +57,8 @@ impl BlockProducer {
         store_address: SocketAddr,
         batch_prover: Option<Url>,
         block_prover: Option<Url>,
+        batch_interval: Duration,
+        block_interval: Duration,
```
bobbinth, line 61: Not for this PR, but we may need to create some kind of struct(s) for these. Otherwise, as we keep adding configuration options, we'll end up with quite a long list of parameters.

bobbinth (commented):
Looks good! Thank you! I left a coupe of small comments inline.

crates/block-producer/src/server.rs
Mirko-von-Leipzig: Agreed. My plan/aim is that we can actually just use the `BlockProducer` as that struct. This would be possible once component startup is possible without this two-stage `init -> serve`.

Mirko-von-Leipzig:
I've removed `humantime` in favor milliseconds.

bobbinth (approved):
Looks good! Thank you!

PR #752, opened 2025-03-19
fix(store): database connections fail due to pragma

Mirko-von-Leipzig:
For whatever reason the changes in the sqlite pool management introduced in #728 no longer works with manually updating the pragma values. This caused connection creation to fail.

This PR updates it to use a builtin method instead.

This PR _does not_ fix the bad errors messages, those are left for a separate issue.


bobbinth (approved):
Looks good! Thank you!

PR #755, opened 2025-03-20
fix: note submission causes database to break

### LINKED ISSUE #753: [Bug]: After sending the first transaction RAM is filled up
### Version

`next`: `7fe907609cc0bb791cb97aa3db499f0c7a33d622`

### Other packages versions

rust 1.85

### What happened?

The node seems to start up properly, but after sending 1 transaction (I was using the faucet, in particular, but looks like it is generated by transaction) starts to consume a lot of memory until it runs out. At some point it start increasing by 2GB per second.

Image

I paste an error that I got in the output a couple fields down.

The node should not run out of memory (Macbook M3 Max with 36 GB RAM).

```bash
mkdir data accounts
miden-node store dump-genesis > genesis.toml
miden-node bundled bootstrap --data-directory data --accounts-directory accounts
miden-node bundled start --rpc.url http://0.0.0.0:57291 --data-directory data
miden-faucet init
miden-faucet create-faucet-account -t POL -d 10 -m 1000000
miden-faucet start --config miden-faucet.toml
```
Then create a mint through the faucet and it should start happening after a couple seconds.

### Relevant log output

```shell
2025-03-19T18:32:59.728929Z  INFO write_block_to_db: miden-store: crates/store/src/db/mod.rs:483: new
2025-03-19T18:32:59.729036Z  INFO write_block_to_db: miden-store: crates/store/src/db/mod.rs:483: close, time.busy: 97.3µs, time.idle: 10.2µs
2025-03-19T18:32:59.729092Z ERROR store.rpc/ApplyBlock:store.server.apply_block:apply_block: miden-store: crates/store/src/state.rs:236: error: block applying was cancelled because of closed channel on database side rpc.service: "store.rpc", rpc.method: "ApplyBlock"
2025-03-19T18:32:59.729117Z  INFO store.rpc/ApplyBlock:store.server.apply_block:apply_block: miden-store: crates/store/src/state.rs:236: close, time.busy: 1.21ms, time.idle: 358µs rpc.service: "store.rpc", rpc.method: "ApplyBlock"
2025-03-19T18:32:59.729099Z ERROR apply_block: miden-store: crates/store/src/db/mod.rs:470: error: SQLite error
2025-03-19T18:32:59.729129Z ERROR store.rpc/ApplyBlock:store.server.apply_block: miden-store: crates/store/src/server/api.rs:292: error: status: Internal, message: "block applying was cancelled because of closed channel on database side", details: [], metadata: MetadataMap { headers: {} } rpc.service: "store.rpc", rpc.method: "ApplyBlock"
2025-03-19T18:32:59.729134Z  INFO apply_block: miden-store: crates/store/src/db/mod.rs:470: close, time.busy: 47.4µs, time.idle: 193µs
2025-03-19T18:32:59.729142Z  INFO store.rpc/ApplyBlock:store.server.apply_block: miden-store: crates/store/src/server/api.rs:292: close, time.busy: 1.72ms, time.idle: 373µs rpc.service: "store.rpc", rpc.method: "ApplyBlock"
2025-03-19T18:32:59.729203Z ERROR store.rpc/ApplyBlock: tower_http::trace::on_failure: /Users/santiagopittella/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tower-http-0.6.2/src/trace/on_failure.rs:93: response failed, classification: Code: 13, latency: 2 ms rpc.service: "store.rpc", rpc.method: "ApplyBlock"
2025-03-19T18:32:59.729231Z  INFO store.rpc/ApplyBlock: miden_node_utils::tracing::grpc: crates/utils/src/tracing/grpc.rs:36: close, time.busy: 1.86ms, time.idle: 396µs rpc.service: "store.rpc", rpc.method: "ApplyBlock"
2025-03-19T18:32:59.729431Z ERROR block_builder.build_block:block_builder.commit_block:store.client.apply_block: miden-block-producer: crates/block-producer/src/store/mod.rs:232: error: gRPC client error
2025-03-19T18:32:59.729449Z  INFO block_builder.build_block:block_builder.commit_block:store.client.apply_block: miden-block-producer: crates/block-producer/src/store/mod.rs:232: close, time.busy: 122µs, time.idle: 2.81ms
2025-03-19T18:32:59.729459Z ERROR block_builder.build_block:block_builder.commit_block: miden-block-producer: crates/block-producer/src/block_builder/mod.rs:210: error: failed to apply block to store
```

Mirko-von-Leipzig:
#728 introduced a [correction](https://github.com/0xPolygonMiden/miden-node/commit/26dbee7d8e5302f3eb8e9b23df3121b78ce163b6#diff-80fe25b196f42f66acaa4e63c5a23552c239be2ea7d0ea6a9050226ab63169a3R47) to `notes` table where a note now has the associated account ID as a foreign key.

Block application however first inserts notes, before inserting new accounts. This PR simply flips the order to the correct one.

Fixes #753.

### Errors

- Our store error messages are opaque with several `err.to_string()` which hides the actual errors.
- For the error in question, we also have no idea which sql query it is that is failing since we lump them all into the same boat.
- Since the `store` is essentially just a gRPC server, there is nowhere where the errors actually get logged properly. The only error logs are from tracing's instrumentation which _does not_ produce error reports.
- Additionally, it is unclear whether the store should really care about the errors either. We currently log an account state query as an error if the account does not exist. This is not a store error, but rather a user query so logging this as an error is bad.
- The block producer doesn't really get a relevant error back over the wire either.
- The store has many "unique" error enums and variants, but they're still not enough to actually cover the specifics.

We definitely need to rethink how we handle errors in the `store`, and probably just in general in the `node`. We rely on error variants but this is proving brittle and simply too little despite having a lot of them. We also don't actually do anything with the variants and they're effectively only used internally. I think part of the issue is that we treat our components as if they're libraries but really they're just an entry-point into a binary - and should probably be largely using `anyhow` for almost all errors. Exceptions would be errors we want to match on and actually do something with.

### Memory explosion

Without the fix, the node's memory explodes and I haven't determined why yet. This should be investigated further as it likely means there is a bug elsewhere e.g. mempool or block-builder maybe.

I'll do that separately to unblock `next` at least.

### Testing

We clearly don't have enough tests around the store. Inserting a basic block into the database should _probably_ be present.

Between this and the (lack of) error reporting, this would be impossible to debug remotely.

Additionally, we have far too many logs/prints that don't help. I'll have to think about this a bit more, but likely having `open` and `close` events for every span is a bit overkill for `info` level.

### New account flow

How are new accounts created? Are they only populated on first use? I was sort of expecting the faucet to create its account with a transaction the moment it launches, however it seems accounts are only created once their state changes the first time?

As in, the faucet account is only created when the first note is submitted. I find this.. weird. The client wallet appears to behave similarly?

### DIFF
--- a/crates/store/src/db/sql/mod.rs
+++ b/crates/store/src/db/sql/mod.rs
@@ -1223,9 +1223,10 @@ pub fn apply_block(
     accounts: &[BlockAccountUpdate],
 ) -> Result<usize> {
     let mut count = 0;
+    // Note: ordering here is important as the relevant tables have FK dependencies.
     count += insert_block_header(transaction, block_header)?;
-    count += insert_notes(transaction, notes)?;
     count += upsert_accounts(transaction, accounts, block_header.block_num())?;
+    count += insert_notes(transaction, notes)?;
     count += insert_transactions(transaction, block_header.block_num(), accounts)?;
     count += insert_nullifiers_for_block(transaction, nullifiers, block_header.block_num())?;
     Ok(count)


SantiagoPittella:
Regarding the high memory usage, I first thought that [this](https://github.com/0xPolygonMiden/miden-node/blob/67eb223069506795f123396b77545f91a459aa95/crates/block-producer/src/block_builder/mod.rs#L66) was related, but this is not supposed to increase memory usage since it should be freed by then.

SantiagoPittella (approved):

bobbinth (approved):
Looks good! Thank you!

Agreed on the error/testing points - let's create an issue to address this in the next milestone (unless we already have one).

> How are new accounts created? Are they only populated on first use? I was sort of expecting the faucet to create its account with a transaction the moment it launches, however it seems accounts are only created once their state changes the first time?

Yes, this is roughly how it works. A client creates an account locally, and the account gets recorded on chain only when there is a first transaction involving this account. The main reason for this is that, once we have fees, creating an account would require a fee - and so, we can't just create an account on-chain without consuming any notes, which requires a transaction.

> As in, the faucet account is only created when the first note is submitted. I find this.. weird. The client wallet appears to behave similarly?

For the faucet it should work differently. The faucet is a part of the genesis block and so should exist even before any transactions involving it are executed.

Mirko-von-Leipzig:
> > As in, the faucet account is only created when the first note is submitted. I find this.. weird. The client wallet appears to behave similarly?
> 
> For the faucet it should work differently. The faucet is a part of the genesis block and so should exist even before any transactions involving it are executed.

Ah I see, so the official faucet can have a pre-generated account in genesis, but in the general case (non-official faucet), this won't be the case and the account is only created on the first note submission.

bobbinth:
> Ah I see, so the official faucet can have a pre-generated account in genesis, but in the general case (non-official faucet), this won't be the case and the account is only created on the first note submission.

Correct - or at least that's how it should work.

PR #757, opened 2025-03-20
fix: renames after cargo update

tomyrd:
This PR deals with the new renames from miden-base. The most prevalent in the node is the change from "account hash" to "account commitment". I searched through all the instances and updated the relevant ones in variable names, docs, proto files, etc.

The branch uses `tomasarrachea-upgrade-rand` as a base as it fixes some issues with the new rand version in miden-base.

Some crate versions also needed to be updated to point to next.

This PR should bring the node up to date with base. This is needed because the client is also up to date with base and the integration tests are failing because the node is outdated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ version      = "0.8.0"
 assert_matches            = { version = "1.5" }
 http                      = { version = "1.2" }
 itertools                 = { version = "0.14" }
-miden-air                 = { version = "0.12" }
+miden-air                 = { package = "miden-air", git = "https://github.com/0xPolygonMiden/miden-vm", branch = "next", default-features = false }
 miden-lib                 = { git = "https://github.com/0xPolygonMiden/miden-base", branch = "next" }
 miden-node-block-producer = { path = "crates/block-producer", version = "0.8" }
 miden-node-proto          = { path = "crates/proto", version = "0.8" }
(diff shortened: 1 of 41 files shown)


Cargo.toml
```diff
@@ -28,7 +28,7 @@ version      = "0.8.0"
 assert_matches            = { version = "1.5" }
 http                      = { version = "1.2" }
 itertools                 = { version = "0.14" }
-miden-air                 = { version = "0.12" }
+miden-air                 = { package = "miden-air", git = "https://github.com/0xPolygonMiden/miden-vm", branch = "next", default-features = false }
```
Mirko-von-Leipzig, line 31: Did the feature set change, or why `default-features = false`? I'm also unsure if the features specified here impact anything since its the workspace toml - does this actually filter down into our packages?

Same question for the `miden-processor`.

crates/block-producer/src/state_view/mod.rs
Mirko-von-Leipzig, line 1: I'm not understanding why these got re-ordered. It seems to me the original ordering was the correct one.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 - [BREAKING] Updated to Rust Edition 2024 (#727).
 - [BREAKING] MSRV bumped to 1.85 (#727).
 - [BREAKING] Replaced `toml` configuration with CLI (#732).
+- [BREAKING] Renamed all instances of "account hash" to "account commitment" (#757).
```
Mirko-von-Leipzig, line 38:
```suggestion
- [BREAKING] Renamed multiple `xxx_hash` to `xxx_commitment` in RPC API (#757).
```

Cargo.toml
tomyrd: I think the default feature option wasn't needed. In any case, it seems the `0.13` version just got released so I'll be changing it to that.

crates/block-producer/src/state_view/mod.rs
tomyrd: I'm also not sure, it seems the vscode formatter automatically flips them. I'll fix it.

Follow up question, it seems the state view module isn't used. Am I missing something?

PR #758, opened 2025-03-21
Fix deployment of mdbook.

phklive:
Add `/html` to fix deployment of mdbook.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -57,7 +57,7 @@ jobs:
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
         uses: actions/upload-pages-artifact@v3
         with:
-          path: ./docs/book
+          path: ./docs/book/html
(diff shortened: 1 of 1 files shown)


.github/workflows/book.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
         uses: actions/upload-pages-artifact@v3
         with:
-          path: ./docs/book
+          path: ./docs/book/html
```
Mirko-von-Leipzig, line 60: Could you add a comment explaining _why_ this is "non-default".

```suggestion
          # We specify multiple [output] sections in our book.toml which causes mdbook to create separate folders for each. This moves the generated `html` into its own `html` subdirectory.
          path: ./docs/book/html
```

Mirko-von-Leipzig (approved):

PR #760, opened 2025-03-21
fix: block and batch default intervals

SantiagoPittella:
Currently, on `next`, using the default values when starting up is causing an error:
```bash
error: invalid value '5_000' for '--block.interval <MILLISECONDS>': invalid digit found in string

For more information, try '--help'.
```

I was running:
```
miden-node bundled start --rpc.url http://0.0.0.0:57291 --data-directory data
```

Removing the underscore from the default values fixed the issue.


igamigo (approved):
Tested the fix, works as expected

bobbinth (approved):
Looks good! Thank you!

Issue #512, opened 2024-10-04
Update docs for endpoints


polydez:
Our docs for various endpoints (in `README` files) are getting out of date. We should update them. In this issue we can also discuss if there is some automated way to generate endpoint docs for rpc/store/block producer components.

polydez:
@bobbinth, I've done a small research of available options for documentation generation based on comments in *.proto files. It's not a full list, but there are the most interesting tools I found so far:

1. Simple, but seems to be enough for us: [protoc-gen-doc](https://github.com/pseudomuto/protoc-gen-doc). Simple to use and and open-source (licensed under MIT), supports for custom templates. Generates markdown, as well, as HTML and DocBook. Runs on docker, so it's safe to use in our infrastructure. Example of generated markup: https://github.com/pseudomuto/protoc-gen-doc/blob/master/examples/doc/example.md
2. Generates static HTML: [sable-docs](https://github.com/markvincze/sabledocs/tree/main). Open-source (MIT). Requires compilation of `*.proto` files to binary descriptors (`*.pb`) in order to use. Have Docker images. Example: https://markvincze.github.io/sabledocs/demo/google.datastore.v1.html#service-google.datastore.v1.Datastore
3. Generates MDX docs, which I'm not sure we can fully use in GitBooks: [Docusaurus Protobuffet](https://github.com/protobuffet/docusaurus-protobuffet). Open-source, licensed under ISC (MIT equivalent). Doesn't have Docker image, or I haven't found. Requires to first generate JSON files by protobuf, consumes them for building formatted documentation. Seems to not being supported since the last year. Example: https://protobuffet.com/protodocs/protobuffet/example/ad/v1/ads.proto/
4. "Like Swagger but for gRPC" (interactive documentation): [gRPC Docs](github.com/gendocu-com/grpc-docs). Open-source (MIT). Requires to first generate JSON files by `protoc-gen-doc`, uses them for accessing to documentation. Has Docker for hosting the interactive web app. Seems to also not being supported since last year. Live demo doesn't work by now, but there is a [screenshot](https://raw.githubusercontent.com/gendocu-com/grpc-docs/master/docs/images/grpc-docs-screenshot.png). 

Another approach might be to make OpenAPI proxy, which will expose interactive documentation (Swagger).
More info in these articles:
1. https://blainsmith.com/articles/go-grpc-gateway-openapi/
2. https://bbengfort.github.io/2021/01/grpc-openapi-docs/

[...]

PR #764, opened 2025-03-25
docs: gRPC documentation

Mirko-von-Leipzig:
This PR adds bare bones gRPC documentation to the mdbook.

Its essentially just a slightly paraphrased copy-paste of the RPC method comments.

Initial implementation for #512 - probably not good enough to close the issue?


docs/src/EXPORTED.md
```diff
@@ -2,9 +2,10 @@
 
 # Summary
 
-- [Node](./operator/index.md)
+- [Node gRPC Reference](./user/rpc.md)
+- [Node Operator Guide](./operator/index.md)
```
Mirko-von-Leipzig, line 6: @phklive I'm unsure of how to properly format this for your intentions.

docs/src/user/rpc.md
```diff
@@ -0,0 +1,87 @@
+# gRPC Reference
+
+This is a reference of the Node's public RPC interface. It consists of a gRPC API which may be used to submit
+transactions and query the state of the blockchain.
+
+The gRPC service definition can be found in the Miden node's `proto`
+[directory](https://github.com/0xPolygonMiden/miden-node/tree/main/proto) in the `rpc.proto` file.
```
Mirko-von-Leipzig, line 7: This should probably point to the README instead, where readers can be directed to the appropriate version / file. Currently this links to `main` which is either the right or wrong thing depending on the reader.

docs/src/EXPORTED.md
phklive: Pushed an update to the file, let me know what you think.

docs/src/user/rpc.md
phklive: Is that a question to me? I think that I would tend to agree with you on this one.

phklive (approved):
Pushed a commit to update the structure of `EXPORTED.md`, overall LGTM.

docs/src/user/rpc.md
```diff
@@ -0,0 +1,87 @@
...
+
+## CheckNullifiers
+
+Request proofs for a set of nullifiers.
+
+## CheckNullifiersByPrefix
+
+Request all nullifiers that match the specified prefixes. Only 16-bit prefixes are supported at this time.
+
+The prefix is used to obscure the callers interest in a specific nullifier.
```
bobbinth, line 33: I think here we also can specify a block to define the lower limit for the query.

bobbinth (approved):
Looks good! Thank you. I think this is fine for now, but in the future, would be good to make the endpoint descriptions "richer" - i.e., most endpoints could probably be a few paragraphs explaining intended usage, caveats etc. But I think we don't need to do this until we stabilize things a bit more.

Mirko-von-Leipzig:
> Looks good! Thank you. I think this is fine for now, but in the future, would be good to make the endpoint descriptions "richer" - i.e., most endpoints could probably be a few paragraphs explaining intended usage, caveats etc. But I think we don't need to do this until we stabilize things a bit more.

Agreed. I didn't want to create examples which _are currently already_ poor abstractions for our own usage :)

PR #768, opened 2025-03-25
ci: update deploy scripts for CLI changes

### LINKED ISSUE #765: update deployment scripts

Mirko-von-Leipzig:
Closes #765 

This PR updates our deployment scripts to account for the change from `toml` config files to a CLI based approach.

- [ ] Test with devnet deployment

### OTel

This currently leaves OTel disabled. We likely want to enable this, but at present we don't have a sensible way of injecting the required environment variables via github deployment.

I recommend we simply do this manually by inserting some extra environment variables.

One alternative is to edit the service file manually and inserting the required values using github secrets. I don't have access myself so at minimum the secrets would require someone else to add.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -141,14 +141,14 @@ jobs:
       # The faucet uses the public faucet generated in the genesis block.
       - name: Configure environment
         uses: ./.github/actions/ssm_execute
+        env:
+          node_bin:   /usr/bin/miden-node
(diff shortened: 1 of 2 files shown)


bobbinth (approved):
Looks good! Thank you! A very superficial review from me, but since devnet deployment worked, I'm assuming that's OK.

> I recommend we simply do this manually by inserting some extra environment variables.

Would this need to be done on the box that is running the devnet?

PR #769, opened 2025-03-25
docs: operator usage and readme update

### LINKED ISSUE #766: update documentation (readme & book)

Mirko-von-Leipzig:
Closes #766 

This strips back the main `readme` rather dramatically. I'm unsure if this is a good idea or not.

Waiting on #723 to land for a proper `proto/README.md` (see currently failing lint).


README.md
```diff
@@ -5,223 +5,43 @@
...
-rm -rf <STORAGE>
-```
+Documentation, tutorials and guides for the current Miden version (aka testnet) can be found
+[here](https://0xpolygonmiden.github.io/miden-docs/), including an operator manual and gRPC reference guide. This is
+your one-stop-shop for all things Miden.
 
 > [!WARNING]
-> Failure to remove existing node data could result in strange behaviour.
+> For node operators living on the development edge, we also host the latest documentation
+> [here](https://0xpolygonmiden.github.io/miden-node/index.html).
```
bobbinth, line 28: nit: I'm not sure this should be in a "warning" callout. If we wanted to call more attention to it, maybe "note" would be better? But also, regular text is probably fine.

docs/src/operator/usage.md
```diff
@@ -1,2 +1,119 @@
...
+mkdir accounts
+
+# Bootstrap the node.
+#
+# This generates the genesis block data and stores it in `<data-directory>/genesis.dat`. This is used by the node to
+# create and verify the database during node startup.
+#
+# Account secrets are stored as `<accounts-directory>/account_xx.mac` where `xx` is the index of the account in the
+# configuration file. These account files are not used by the node and should instead be used whereever you intend to operate these accounts,
+# e.g. to run the `miden-faucet` (see Faucet section).
```
bobbinth, line 39: I would break comment lines here so that they fit onto the page without causing horizontal scroll bars.

bobbinth (approved):

> This strips back the main readme rather dramatically. I'm unsure if this is a good idea or not.

I think this is fine.

PR #770, opened 2025-03-26
ci: optimise test run time

Mirko-von-Leipzig:
Our faucet website tests takes ~1 minute to run which is two orders of magnitude slower than the rest of our test suite. This PR reduces this to a few seconds by promoting `miden-crypto` to a more optimised compilation.

This works because the primary factor slowing the website test down was faucet account creation.

Additionally, this PR also removes the `make test` logging output. It will now print only for failing tests, which is a lot less noisy imo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,10 @@ repository   = "https://github.com/0xPolygonMiden/miden-node"
 rust-version = "1.85"
 version      = "0.8.0"
 
+# Optimise the cryptography for faster tests involving account creation.
+[profile.test.package.miden-crypto]
+opt-level = 2
+
 [workspace.dependencies]
 anyhow                    = { version = "1.0" }
 assert_matches            = { version = "1.5" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -25,6 +25,10 @@ repository   = "https://github.com/0xPolygonMiden/miden-node"
 rust-version = "1.85"
 version      = "0.8.0"
 
+# Optimise the cryptography for faster tests involving account creation.
+[profile.test.package.miden-crypto]
+opt-level = 2
+
```
Mirko-von-Leipzig, line 31: I used 

```
cargo flamegraph miden-faucet --unit-test -- test_website
```

to identify the cause. I also tried several other `miden-xxx` and `winter-xxx` deps, but maybe there are some I missed, or there is some combination that should be used?

PhilippGackstatter: The slow part is `SecretKey::with_rng`, right? On my machine it takes 51s on current next, while actual account creation takes 81ms, so I think promoting miden-crypto to a higher opt level is the right thing to do.

PhilippGackstatter (approved):

PR #771, opened 2025-03-26
fix: OOM on block building error

### LINKED ISSUE #756: Investigate memory explosion
The bug discovered in #753 caused the node's memory to balloon.

The bug itself was fixed in #755 but the memory issue still needs to be investigated.

The root cause is that the block was rejected by the store. A likely candidate is the block rollback flow in the block-producer/mempool code as this is rarely (never?) run.

Mirko-von-Leipzig:
Our OTel error report injector contains a logic bug. This PR fixes that.

The root cause was failing to update the loop variable to the child, causing an infinite loop and infinitely growing the error report.

Since this is error related it never exhibits on the happy path.

Fixes #756

### DIFF
--- a/crates/utils/src/tracing/span_ext.rs
+++ b/crates/utils/src/tracing/span_ext.rs
@@ -104,9 +104,10 @@ where
     fn set_error(&self, err: &dyn std::error::Error) {
         // Coalesce all sources into one string.
         let mut description = format!("{err}");
-        let current = err;
+        let mut current = err;
         while let Some(cause) = current.source() {
             description.push_str(format!("\nCaused by: {cause}").as_str());
+            current = cause;
         }
         tracing_opentelemetry::OpenTelemetrySpanExt::set_status(
             self,


igamigo (approved):
I haven't tested inducing a problem yet but the code looks good to me!
It seems like [`iter::succesors()`](https://doc.rust-lang.org/std/iter/fn.successors.html) could be used here to avoid having to manually update the var. Probably not a net positive tradeoff though because resulting code is likely less readable

Mirko-von-Leipzig:
> It seems like [`iter::succesors()`](https://doc.rust-lang.org/std/iter/fn.successors.html) could be used here to avoid having to manually update the var. Probably not a net positive tradeoff though because resulting code is likely less readable

I really wanted to use this (thanks I didn't know what it was called!), but I can't seem to make it work due to lifetime issues.

> due to  current limitations in the borrow checkers, this implies a `'static` lifetime

SantiagoPittella (approved):

PR #776, opened 2025-03-27
feat: remove `genesis.dat`

### LINKED ISSUE #767: Make `genesis.dat` ephemeral
Node bootstrapping currently creates a `genesis.dat` file which is then used to initialize the database on node startup, as well as checked against the database on every subsequent run.

It would be better if `bootstrap` created the database, and that the `start` command always assumes the database is already present (no creation). This removes the need for having a `genesis.dat` file entirely.

Effectively, `bootstrap` would convert a `genesis.toml` configuration directly into the node's database containing the genesis block data.

We will need to figure out how/if we can make this genesis block repeatable from only the configuration file. This will be needed once we need to instantiate multiple nodes for the same network i.e. they must have the same genesis data. This is not strictly required at present though.

Mirko-von-Leipzig:
This PR moves the database initialization procedure from node/store startup, to the bootstrap command.

This feels more natural and also removes the need for the temporary `genesis.dat` file that was floating about. 

Closes #767.

### DIFF
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -7,7 +7,7 @@ use std::{
 
 use anyhow::Context;
 use miden_lib::{AuthScheme, account::faucets::create_basic_fungible_faucet, utils::Serializable};
-use miden_node_store::{genesis::GenesisState, server::Store};
+use miden_node_store::{GenesisState, Store};
 use miden_node_utils::{crypto::get_rpo_random_coin, grpc::UrlExt};
 use miden_objects::{
     Felt, ONE,
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -80,7 +80,7 @@ impl StoreCommand {
                 config,
                 data_directory,
                 accounts_directory,
-            } => Self::bootstrap(config, &data_directory, &accounts_directory),
+            } => Self::bootstrap(config, data_directory, &accounts_directory),
             // Note: open-telemetry is handled in main.
             StoreCommand::Start { url, data_directory, open_telemetry: _ } => {
                 Self::start(url, data_directory).await
@@ -121,7 +121,7 @@ impl StoreCommand {
 
     fn bootstrap(
         genesis_input: Option<PathBuf>,
-        data_directory: &Path,
+        data_directory: PathBuf,
         accounts_directory: &Path,
     ) -> anyhow::Result<()> {
         // Parse the genesis configuration input.
@@ -162,15 +162,10 @@ impl StoreCommand {
                 })?;
         }
 
-        // Write the genesis state to disk. This is used to seed the database's genesis block.
+        // Bootstrap the store database.
         let accounts = accounts.into_iter().map(|account| account.account).collect();
         let genesis_state = GenesisState::new(accounts, version, timestamp);
-        let genesis_output = data_directory.join(miden_node_store::GENESIS_STATE_FILENAME);
-        File::create_new(&genesis_output)
-            .and_then(|mut file| file.write_all(&genesis_state.to_bytes()))
-            .with_context(|| {
-                format!("failed to write genesis data to file {}", genesis_output.display())
-            })
+        Store::bootstrap(genesis_state, data_directory)
     }
 
     fn generate_account(input: &AccountInput, rng: &mut ChaChaRng) -> anyhow::Result<AccountFile> {
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -137,10 +137,9 @@ impl StoreCommand {
 
         // Generate the accounts.
         let mut rng = ChaCha20Rng::from_seed(rand::random());
-        let n_accounts = accounts.as_ref().map(Vec::len).unwrap_or_default();
+        let n_accounts = accounts.len();
         let accounts = accounts
-            .into_iter()
-            .flatten()
+            .iter()
             .enumerate()
             .inspect(|(idx, _)| tracing::info!(index=%idx, total=n_accounts, "Generating account"))
             .map(|(idx, input)| {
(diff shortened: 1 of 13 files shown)


bin/stress-test/src/main.rs
```diff
@@ -141,7 +130,8 @@ async fn seed_store(data_directory: PathBuf, num_accounts: usize) {
     };
 
     // start generating blocks
-    let genesis_header = genesis_state.into_block().unwrap();
+    let database_filepath = data_directory.join(STORE_FILENAME);
```
Mirko-von-Leipzig, line 133: One alternative is to expose the `DataDirectory` structure I added in the store.

bin/node/src/commands/store.rs
```diff
@@ -163,18 +162,13 @@ impl StoreCommand {
...
-        File::create_new(&genesis_output)
-            .and_then(|mut file| file.write_all(&genesis_state.to_bytes()))
-            .with_context(|| {
-                format!("failed to write genesis data to file {}", genesis_output.display())
-            })
+        Store::bootstrap(genesis_state, data_directory)
     }
 
-    fn generate_account(input: AccountInput, rng: &mut ChaChaRng) -> anyhow::Result<AccountFile> {
+    fn generate_account(input: &AccountInput, rng: &mut ChaChaRng) -> anyhow::Result<AccountFile> {
```
bobbinth, line 171: There is a `TODO` comment on line 189 below. We can replace it with a brief explanation. Basically, we need to set nonce to a non-zero value because `ZERO` nonce means that an account is not yet on-chain. For example, when we first create an account locally, it gets nonce `ZERO`. Once we execute a transaction against it (which is required to record the account on-chain), the nonce will be incremented. So, no account recorded in a block has nonce `ZERO`.

Accounts created in the genesis block are different since they don't require transactions executed against them to get recorded. But we still want to uphold the invariant of non-zero nonce because in several places we treat going from `ZERO` to non-zero nonce differently than incrementing a non-zero nonce.

crates/store/src/db/mod.rs
```diff
@@ -187,21 +184,29 @@ impl From<NoteRecord> for NoteSyncRecord {
...
-        database_filepath: PathBuf,
-        genesis_filepath: &str,
-        block_store: Arc<BlockStore>,
-    ) -> Result<Self, DatabaseSetupError> {
-        info!(target: COMPONENT, ?database_filepath, "Connecting to the database");
-
-        if let Some(p) = database_filepath.parent() {
-            create_dir_all(p).map_err(DatabaseError::IoError)?;
-        }
+    pub fn bootstrap(database_filepath: PathBuf, genesis: &GenesisBlock) -> anyhow::Result<()> {
```
bobbinth, line 187: I would add brief doc comments here.

Also, do we not want to instrument this function?

bobbinth (commented):
Looks good! Thank you! I left a some small comments inline.

> We could keep the `genesis.dat` file for this purpose, and essentially allow users to `bootstrap` from either the `toml` or the `dat` variations. Or maybe better would be to bring back the `make-genesis` command to translate from `toml -> dat`.
> 
> In general though, this requirement is _very_ far away still, and we should prefer not to solve problems before they actually arrive.

I agree with delaying this feature until later.

crates/store/src/blocks.rs
```diff
@@ -1,15 +1,53 @@
...
+    ///
+    /// Uses [`std::fs::create_dir`] and therefore has the same error conditions.
+    pub fn bootstrap(store_dir: PathBuf, genesis_block: &GenesisBlock) -> std::io::Result<Self> {
+        std::fs::create_dir(&store_dir)?;
+
+        let block_store = Self { store_dir };
+        block_store.save_block_blocking(BlockNumber::GENESIS, &genesis_block.inner().to_bytes())?;
+
+        Ok(block_store)
+    }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: To me that's an invariant upheld by the `GenesisBlock` wrapper type?

I can add the assert though.

bin/node/src/commands/store.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: We won't use it ourselves for deployments, but I thought it would be a good way for other users to get the format on disk to play with.

I'm happy to remove it though if we think there isn't a need for such configuration (yet).

bin/node/src/commands/store.rs
Mirko-von-Leipzig: Ah. I see, thanks. I added that `TODO` in the last PR and forgot to follow-up in the review :)

crates/store/src/db/mod.rs
Mirko-von-Leipzig: I'm on the fence tbh. Mostly because this is a once-off so its not like there are statistics or metrics we can glean from this.

On the other hand there isn't really a large downside besides polluting the logs/traces a tiny bit.

bin/node/src/commands/store.rs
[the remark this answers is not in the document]
bobbinth: I'd probably remove it. It's possible by the time we need it, the format could change as well.

crates/store/src/db/mod.rs
Mirko-von-Leipzig: That said, we probably _do_ want to instrument the database migrations. So maybe we should just do it :)

Mirko-von-Leipzig: Something that could be nice is actually using these traces during development. You can have a separate backend environment per developer, and they can use this to test/inspect/debug slowness and flows etc.

bobbinth (approved):
Looks good! Thank you! Do we also need to update deployment scripts because of this?

Mirko-von-Leipzig:
> Looks good! Thank you! Do we also need to update deployment scripts because of this?

Ah yes, thanks - the account file name changed slightly since there is only the single account.

Though maybe I should keep it named `account_1.mac` in case we add more accounts? I changed it to `account.mac` in this PR.

PR #780, opened 2025-03-31
chore: check matching security level on proofs

SantiagoPittella:
Adds a check for the proof security level to be greater than the minimum in batches and blocks.

This PR also uses the `next` branch of miden-base.

Part of https://github.com/0xPolygonMiden/miden-node/issues/744

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## v0.9.0 (TBD)
 
+### Changes
+
+- [BREAKING] Simplified node bootstrapping (#776).
+  - Database is now created during bootstrap process instead of on first startup.
+  - Data directory is no longer created but is instead expected to exist.
+  - The genesis block can no longer be configured which also removes the `store dump-genesis` command.
 
 ## v0.8.0 (2025-03-26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.9.0 (TBD)
 
-
+- Added a check for batch and block proofs to be greater than the minimum security level (#780).
 
 ## v0.8.0 (2025-03-26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## v0.9.0 (TBD)
 
-- Added a check for batch and block proofs to be greater than the minimum security level (#780).
 
 ## v0.8.0 (2025-03-26)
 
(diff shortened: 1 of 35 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## v0.9.0 (TBD)
 
-
+- Added a check for batch and block proofs to be greater than the minimum security level (#780).
```
Mirko-von-Leipzig, line 5: This doesn't affect the user(s) and is a no-op, so no need for a changelog item.

```suggestion
```

Mirko-von-Leipzig (approved):

crates/block-producer/src/batch_builder/mod.rs
```diff
@@ -245,6 +245,15 @@ impl BatchJob {
...
+        }?;
+
+        if proven_batch.proof_security_level() < MIN_PROOF_SECURITY_LEVEL {
+            Err(BuildBatchError::SecurityLevelTooLow(
+                proven_batch.proof_security_level(),
+                MIN_PROOF_SECURITY_LEVEL,
+            ))
+        } else {
+            Ok(proven_batch)
         }
```
bobbinth, line 257: Should we make this a bit more flexible and not hard-code the minimum security level? Basically, should we make `min_proof_security` be a property of `BatchBuilder` or maybe even `BatchProver`?

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -206,6 +206,13 @@ impl BlockBuilder {
     ) -> Result<ProvenBlock, BuildBlockError> {
         let proven_block = self.block_prover.prove(proposed_block).await?;
 
+        if proven_block.proof_security_level() < MIN_PROOF_SECURITY_LEVEL {
+            return Err(BuildBlockError::SecurityLevelTooLow(
+                proven_block.proof_security_level(),
+                MIN_PROOF_SECURITY_LEVEL,
+            ));
+        }
```
bobbinth, line 214: Same comment as above.

crates/block-producer/src/batch_builder/mod.rs
SantiagoPittella: I'm not sure. It would make sense to do that if the security level needs to be higher than the configured prover (assuming it can change). But if we're only interested in it being higher than the minimum (which is what we're currently doing), the current check would be fine.

bobbinth: I think for now this is fine as is. We may want to make security level a config parameter of the node - but that would be a separate task. Let's create an issue for this.

bobbinth (approved):
All looks good! Thank you!

PR #781, opened 2025-04-01
Update to use `TransactionHeader`

PhilippGackstatter:
Update usages of account updates and batch and block to use the newly introduced `TransactionHeader` from https://github.com/0xPolygonMiden/miden-base/pull/1247.

These are relatively minor updates. I think the only thing worth calling out is the insertion of the transactions into the DB. Currently the table has these fields:

```sql
CREATE TABLE transactions (
    transaction_id BLOB    NOT NULL,
    account_id     BLOB    NOT NULL,
    block_num      INTEGER NOT NULL,
```

We could insert more data here from the `TransactionHeader`, but how exactly depends on some details I'm not familiar with, like should we insert notes into this table and reference them via a foreign key or is that a bad idea because we remove notes from the notes table when they're consumed?.

If we want to support the explorer to get the transactions of a block, then we probably have to insert the full headers?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.9.0 (TBD)
 
-
+- Adapt to the use of `TransactionHeader` in batches and blocks (#781).
 
 ## v0.8.0 (2025-03-26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## v0.9.0 (TBD)
 
-- Adapt to the use of `TransactionHeader` in batches and blocks (#781).
 
 ## v0.8.0 (2025-03-26)
 
--- a/crates/block-producer/Cargo.toml
+++ b/crates/block-producer/Cargo.toml
@@ -21,15 +21,15 @@ tracing-forest = ["miden-node-utils/tracing-forest"]
 async-trait                  = { version = "0.1" }
 futures                      = { version = "0.3" }
 itertools                    = { workspace = true }
-miden-block-prover           = { version = "0.8" }
+miden-block-prover           = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "pgackst-verified-tx" }
 miden-lib                    = { workspace = true }
 miden-node-proto             = { workspace = true }
 miden-node-utils             = { workspace = true }
 miden-objects                = { workspace = true }
 miden-processor              = { workspace = true }
-miden-proving-service-client = { version = "0.8", features = ["batch-prover", "block-prover"] }
+miden-proving-service-client = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "pgackst-verified-tx", features = ["batch-prover", "block-prover"] }
 miden-tx                     = { workspace = true }
-miden-tx-batch-prover        = { version = "0.8" }
+miden-tx-batch-prover        = { git = "https://github.com/0xPolygonMiden/miden-base.git", branch = "pgackst-verified-tx" }
 rand                         = { version = "0.9" }
 thiserror                    = { workspace = true }
 tokio                        = { workspace = true, features = ["macros", "net", "rt-multi-thread", "sync", "time"] }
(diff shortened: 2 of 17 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## v0.9.0 (TBD)
 
-
+- Adapt to the use of `TransactionHeader` in batches and blocks (#781).
```
Mirko-von-Leipzig, line 5: Our changelog is intended for the end users / node operator so I don't think this PR is "visible" to them.

```suggestion
```

crates/block-producer/Cargo.toml
Mirko-von-Leipzig, line 1: Hmm. How are these `toml` changes passing the `taplo` lint check?

Are the lines too long and it gave up?

crates/store/src/db/sql/mod.rs
Mirko-von-Leipzig, line 1: > We could insert more data here from the TransactionHeader, but how exactly depends on some details I'm not familiar with, like should we insert notes into this table and reference them via a foreign key or is that a bad idea because we remove notes from the notes table when they're consumed?.
>
> If we want to support the explorer to get the transactions of a block, then we probably have to insert the full headers?

The full raw block data is technically available already via the `BlockStore` which I believe should include this new data? https://github.com/0xPolygonMiden/miden-node/blob/a0fdf2b45600ec5ac26af7c5693e9a1759aace38/proto/proto/responses.proto#L198-L202

If it does include it, then I would ignore changing the DB format until we need to refer to some part of it, or to get it separately from the block body somehow.

Mirko-von-Leipzig: Actually looking into `ProvenBlock`, I can't seem to make the actual connection so maybe they're not serialized..

CHANGELOG.md
PhilippGackstatter: Makes sense, I removed it. I can't seem to add labels to this PR, so maybe the no changelog label has to be added still.

crates/block-producer/Cargo.toml
PhilippGackstatter: I was wondering about that too but I just ran `make lint`. I guess they're too long and that's what it fell back to.

crates/store/src/db/sql/mod.rs
PhilippGackstatter: They are part of the serialization of the `ProvenBlock`, so I think what you said in your original comment should be true:
https://github.com/0xPolygonMiden/miden-base/blob/fb028b95182116e9dcfc404485f4c3c53713cc2b/crates/miden-objects/src/block/proven_block.rs#L147-L155

Mirko-von-Leipzig (approved):

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -319,8 +319,11 @@ impl TelemetryInjectorExt for ProposedBlock {
                 .expect("should have less than u32::MAX output notes"),
         );
 
-        let num_batch_created_notes =
-            self.batches().iter().fold(0, |acc, batch| acc + batch.output_notes().len());
+        let num_batch_created_notes = self
+            .batches()
+            .as_slice()
+            .iter()
+            .fold(0, |acc, batch| acc + batch.output_notes().len());
```
bobbinth, line 326: I would have maybe put this logic into the `OrderedBatches` so that we could do something like this here:
```rust
let num_batch_created_notes = self.batches.num_output_notes();
```

crates/store/src/state.rs
```diff
@@ -234,7 +234,9 @@ impl State {
 
         let header = block.header();
 
-        let tx_commitment = BlockHeader::compute_tx_commitment(block.transactions());
+        let tx_commitment = BlockHeader::compute_tx_commitment(
+            block.transactions().as_slice().iter().map(|tx| (tx.id(), tx.account_id())),
+        );
```
bobbinth, line 239: Similar to the above comments. I would have probably added this logic to the `OrderedTransactions` so that here we could say:
```rust
let tx_commitment = block.transactions.commitment();
```

crates/block-producer/src/block_builder/mod.rs
PhilippGackstatter: Nice suggestion, I've sneaked those in with https://github.com/0xPolygonMiden/miden-node/pull/783.

PR #782, opened 2025-04-03
feat: load test Store endpoints

TomasArrachea:
Addresses https://github.com/0xPolygonMiden/miden-node/issues/772.

This PR introduces a new command to the stress-test CLI: `BenchmarkStore`, which load tests some of the store endpoints.

### Example outputs

Using a seeded store with 100k accounts:

<details>
<summary><code>benchmark-store sync-state</code></summary>

```bash
miden-node-stress-test benchmark-store sync-state --data-directory ./data --iterations 1000 --concurrency 10
```

```bash
Average request latency: 7.306088ms
P95 request latency: 15.403583ms
Average notes per response: 2.606
```
</details>

<details>
<summary><code>benchmark-store sync-notes</code></summary>

```bash
miden-node-stress-test benchmark-store sync-notes --data-directory ./data --iterations 1000 --concurrency 10
```

```bash
Average request took: 6.488028ms
P95 request latency: 14.965917ms
```
</details>

<details>
<summary><code>benchmark-store check-nullifiers-by-prefix</code></summary>

```bash
miden-node-stress-test benchmark-store check-nullifiers-by-prefix --data-directory ./data --iterations 1000 --concurrency 10
```

```bash
Average request latency: 566.787µs
P95 request latency: 849.958µs
Average nullifiers per response: 11.466
```
</details>

### New command parameters
- --data-directory: used to load the store and the account IDs dump file
- --iterations: number of iterations to perform the request
- --concurrency: level of concurrency for parallel requests

### `SeedStore` command changes
- Added a new `public_accounts_percentage` parameter to allow seeding the store with both public and private accounts
- Now creates an `accounts.txt` dump file containing all newly created account IDs when seeding the store


bin/stress-test/src/endpoints_metrics.rs
```diff
@@ -0,0 +1,22 @@
+use std::time::Duration;
+
+/// Prints a summary of the benchmark results.
```
igamigo, line 3: nit: We could explain that the results summary involves average time and p95 percentile

bin/stress-test/src/endpoints_metrics.rs
```diff
@@ -0,0 +1,22 @@
...
+pub fn print_summary(timers_accumulator: Vec<Duration>) {
```
igamigo, line 4: nit: Does this need to be `pub`?

bin/stress-test/src/main.rs
```diff
@@ -1,60 +1,12 @@
...
-    time::Instant,
-};
+use std::path::PathBuf;
 
-mod metrics;
-use anyhow::Context;
+mod endpoints_metrics;
+mod seeding;
+mod seeding_metrics;
+mod store_tests;
```
igamigo, line 6: Let's put these mod declarations below the `use` statements

bin/stress-test/src/store_tests.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use crate::{
+    endpoints_metrics::print_summary,
+    seeding::{ACCOUNTS_FILENAME, start_store},
+};
+
+// SYNC STATE
+// ================================================================================================
+
+/// Sends multiple `sync_state` requests to the store and prints the performance.
```
igamigo, line 33: We should mention here (and in similar functions) what each of the parameters mean since these functions are the most central ones in relation to the benchmarks

bin/stress-test/src/store_tests.rs
```diff
@@ -0,0 +1,213 @@
...
+            break;
+        }
+    }
+    let mut nullifiers = nullifier_prefixes.into_iter().cycle();
+
+    // each request will have 10 prefixes and block number 0
+    let request = |_| {
+        let mut client = store_client.clone();
+
+        let nullifiers_batch: Vec<u32> = nullifiers.by_ref().take(10).collect();
```
igamigo, line 187: The number of sent nullifier prefixes is something that we probably want to modify easily so we may want to make it a parameter

bin/stress-test/src/store_tests.rs
```diff
@@ -0,0 +1,213 @@
...
+        .collect::<Vec<_>>()
+        .await;
+
+    print_summary(timers_accumulator);
+}
+
+/// Sends a single `check_nullifiers_by_prefix` request to the store and returns the elapsed time.
+async fn check_nullifiers_by_prefix(
+    api_client: &mut ApiClient<InterceptedService<Channel, OtelInterceptor>>,
+    nullifiers: Vec<u32>,
```
igamigo, line 206: nit: These are nullifier prefixes, right? If so I'd change the name accordingly

bin/stress-test/src/store_tests.rs
```diff
@@ -0,0 +1,213 @@
...
+        .map(request)
+        .buffer_unordered(concurrency)
+        .map(|res| res.unwrap())
+        .collect::<Vec<_>>()
+        .await;
+
+    print_summary(timers_accumulator);
+}
+
+/// Sends a single `sync_notes` request to the store and returns the elapsed time.
```
igamigo, line 115: Let's describe what the requests look like (ie, for each passed account ID you also send a related note tag)

bin/stress-test/src/store_tests.rs
```diff
@@ -0,0 +1,213 @@
...
+) -> Duration {
+    let note_tags = account_ids
+        .iter()
+        .map(|id| u32::from(NoteTag::from_account_id(*id, NoteExecutionMode::Local).unwrap()))
+        .collect::<Vec<_>>();
+    let sync_request = SyncNoteRequest { block_num: 0, note_tags };
+
+    let start = Instant::now();
+    api_client.sync_notes(sync_request).await.unwrap();
+    start.elapsed()
```
igamigo, line 128: I wonder if we want to introduce a couple of random note tags to account for non-standard use cases

bin/stress-test/src/store_tests.rs
```diff
@@ -0,0 +1,213 @@
...
+
+/// Sends a single `check_nullifiers_by_prefix` request to the store and returns the elapsed time.
+async fn check_nullifiers_by_prefix(
+    api_client: &mut ApiClient<InterceptedService<Channel, OtelInterceptor>>,
+    nullifiers: Vec<u32>,
+) -> Duration {
+    let sync_request = CheckNullifiersByPrefixRequest { nullifiers, prefix_len: 16, block_num: 0 };
+
+    let start = Instant::now();
+    api_client.check_nullifiers_by_prefix(sync_request).await.unwrap();
```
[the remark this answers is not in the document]
igamigo: This also applies for the sync state: note tags will match with more notes due to them being matched based on partial account IDs

igamigo (commented):
Overall logic looks good! Left some comments mostly related to documentation, and some things I think could be interesting to add.

bin/stress-test/src/endpoints_metrics.rs
TomasArrachea: Yes, it's used on each test to print the results so it needs to be public

bin/stress-test/src/store_tests.rs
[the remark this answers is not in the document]
TomasArrachea: Ok! I added the print for the received nullifiers and got an average of `11.466`. Also for the sync_state the average received notes per response and got `2.606`. Updated the PRs description with the current output.

Mirko-von-Leipzig (approved):
I think this looks like a very decent starting point.

What would be nice is:

- able to target non-local store i.e. one that already exists, and
- including `OTel` support

This enables a general workflow/investigation style of:

1. Launch a `store` (or even full node) on some representative instance.
2. Seed it (if required)
3. Create a new honeycomb backend to aggregate otel stats
4. Launch stress test, and configure it to output otel spans to the backend from (3).
5. You can now investigate/dig into the otel span data

Though maybe this isn't as nice as I think it is, but it could be easier to use since you get automatic span timings etc from everywhere in the `store` (once its properly setup for otel).

We aren't there yet, so as is this tool should definitely include its own metrics.

bin/stress-test/src/endpoints_metrics.rs
bobbinth, line 1: nit: I would create a `store` directory with `mod.rs` file (which would be the same as `store_tests.rs` file) and `metrics.rs` file (which would be the same as this file).

bin/stress-test/src/seeding_metrics.rs
bobbinth, line 1: nit: I would probably create a `seeding` directory with two files: `mod.rs` (with contents of the current `seeding.rs`) and `metrics.rs` (with the contents of this file).

bobbinth (approved):
Looks good! Thank you. I left a couple of small comments inline. I would also do two things:

- Update the `README.md` file for this crate with detailed instructions on how to run the stress tests.
- Add something like the "Results" section to the readme listing the results you are currently getting.

PR #784, opened 2025-04-07
feat: remove `ApiError`

### LINKED ISSUE #751: API Error is a poor abstraction
We currently have a generic API error type in the utils crate. Unfortunately it still uses `String` formatting which means it doesn't provide an error report at all.

We need to rethink this; the overarching issue is that several errors here are only applicable to the `store`, but cannot access the `store` errors because the dependency goes the wrong way around.

Likely we shouldn't have a general `ApiError` type, or at least it needs to be extensible somehow.

TomasArrachea:
Closes #751.

This PR removes the `ApiError` from the `miden-node-utils` crate. Instead, the components return an `anyhow::Result`, since it is only used to crash and display the error report.


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
   - Database is now created during bootstrap process instead of on first startup.
   - Data directory is no longer created but is instead expected to exist.
   - The genesis block can no longer be configured which also removes the `store dump-genesis` command.
+- [BREAKING] Remove `ApiError` from `miden-node-utils` and use anyhow errors (#784).
```
Mirko-von-Leipzig, line 11: Doesn't impact the node user

```suggestion
```

crates/utils/src/grpc.rs
```diff
@@ -1,20 +1,17 @@
...
 impl UrlExt for url::Url {
-    fn to_socket(&self) -> Result<SocketAddr, ApiError> {
-        self.socket_addrs(|| None)
-            .map_err(ApiError::EndpointToSocketFailed)?
+    fn to_socket(&self) -> anyhow::Result<SocketAddr> {
+        self.socket_addrs(|| None)?
             .into_iter()
             .next()
-            .ok_or_else(|| ApiError::AddressResolutionFailed(self.to_string()))
+            .ok_or_else(|| anyhow::anyhow!("address resolution failed for {}", self.to_string()))
```
Mirko-von-Leipzig, line 14: I believe you can `context` on options

Mirko-von-Leipzig (approved):
Just the couple of minor comments

bobbinth (approved):
Looks good! Thank you!

PR #785, opened 2025-04-08
feat: component startups with lazy connection

TomasArrachea:
Closes https://github.com/0xPolygonMiden/miden-node/issues/73.

### Summary
This PR updates the Rpc and Block Producer components to connect lazily. Also, the Block Producer now retries the first connection to the Store on startup and binds the `TpcListener` only after connecting to the Store.

### Motivation
Previously, the gRPC clients were connected eagerly during component init. This made it difficult to start components independently, as each required its dependencies to be serving beforehand. The new approach defers the connection until the first request is made, allowing independent startup of components.

### Implementation details
- Uses tonic's `connect_lazy` to defer the client's connection.
- Implements exponential backoff for the block producer startup.
- Removes the `init()` method for all components initialization, now only the `serve()` method is needed.

Reconnections in case of a component being restarted are handled automatically by the tonic client.

### Testing
Added tests for the Block Producer and Rpc components startup.


TomasArrachea:
I run into some problems when testing the components startup, aborting the store task did not stop the server from serving requests. I added a `serve_with_shutdown` method on the store component for now but I don't think it's a good solution, there is probably a simpler way of implementing the test.

bin/node/src/commands/block_producer.rs
```diff
@@ -92,10 +91,7 @@ impl BlockProducerCommand {
             block_interval,
         )
         .await
-        .context("Loading store")?
-        .serve()
-        .await
-        .context("Serving store")
+        .context("Failed to start serving block-producer")
```
Mirko-von-Leipzig, line 94:
```suggestion
        .context("failed while serving block-producer component")
```

bin/node/src/commands/bundled.rs
```diff
@@ -150,30 +147,38 @@ impl BundledCommand {
...
 
         // Start store. The store endpoint is available after loading completes.
-        let store = Store::init(grpc_store, data_directory).await.context("Loading store")?;
-        let store_id =
-            join_set.spawn(async move { store.serve().await.context("Serving store") }).id();
+        let store_id = join_set
+            .spawn(async move {
+                miden_node_store::serve(grpc_store, data_directory)
+                    .await
+                    .context("Serving store")
```
Mirko-von-Leipzig, line 154:
```suggestion
                    .context("failed while serving store component")
```

bin/node/src/commands/bundled.rs
```diff
@@ -150,30 +147,38 @@ impl BundledCommand {
...
+                miden_node_block_producer::serve(
+                    grpc_block_producer,
+                    store_address,
+                    batch_prover_url,
+                    block_prover_url,
+                    batch_interval,
+                    block_interval,
+                )
+                .await
+                .context("Serving block-producer")
```
Mirko-von-Leipzig, line 170:
```suggestion
                .context("failed while serving block-producer component")
```

bin/node/src/commands/bundled.rs
```diff
@@ -150,30 +147,38 @@ impl BundledCommand {
...
         // Start RPC component.
-        let rpc = Rpc::init(grpc_rpc, store_address, block_producer_address)
-            .await
-            .context("Loading RPC")?;
-        let rpc_id = join_set.spawn(async move { rpc.serve().await.context("Serving RPC") }).id();
+        let rpc_id = join_set
+            .spawn(async move {
+                miden_node_rpc::serve(grpc_rpc, store_address, block_producer_address)
+                    .await
+                    .context("Serving RPC")
```
Mirko-von-Leipzig, line 179:
```suggestion
                    .context("failed while serving RPC component")
```

bin/node/src/commands/store.rs
```diff
@@ -86,12 +86,7 @@ impl StoreCommand {
             .await
             .context("Failed to bind to store's gRPC URL")?;
 
-        Store::init(listener, data_directory)
-            .await
-            .context("Loading store")?
-            .serve()
-            .await
-            .context("Serving store")
+        miden_node_store::serve(listener, data_directory).await.context("Serving store")
```
Mirko-von-Leipzig, line 89:
```suggestion
        miden_node_store::serve(listener, data_directory).await.context("failed while serving store component")
```

tests/integration/main.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    // in order to kill the store, we need to spawn a new runtime and run the store on it. That
+    // allows us to kill all the tasks spawned by the store when we kill the runtime.
+    let store_runtime =
+        runtime::Builder::new_multi_thread().enable_time().enable_io().build().unwrap();
+    store_runtime.spawn(async move {
+        miden_node_store::serve(store_listener, dir)
+            .await
+            .expect("store should start serving");
+    });
```
Mirko-von-Leipzig, line 33: Could we not use the join handle from spawn for this purpose? By calling `abort` and then `.await` until it ends.

crates/block-producer/src/server.rs
```diff
@@ -30,153 +28,94 @@ use crate::{
...
-            chain_tip,
+/// Note: this blocks until one of the servers die.
+pub async fn serve(
+    listener: TcpListener,
+    store_address: SocketAddr,
+    batch_prover: Option<Url>,
+    block_prover: Option<Url>,
+    batch_interval: Duration,
+    block_interval: Duration,
+) -> anyhow::Result<()> {
```
Mirko-von-Leipzig, line 41: I think maybe we should keep the struct so that its more obvious what values are being assigned when reading the code.

e.g. 

```rust
BlockProducer {
    listener: ..,
    ...
    batch_interval:
}.serve().await
```

The structs effectively become configuration objects with all values `pub`.

In particular this also lets us keep `Store::bootstrap` and `Store::serve` more logically together, though this is more a personal preference. It also let's us document the "input" parameters using doc-comments on the fields which is nice imo.

wdyt?

crates/rpc/src/server/api.rs
```diff
@@ -294,6 +330,11 @@ impl api_server::Api for RpcService {
             ));
         }
 
-        self.store.clone().get_account_proofs(request).await
+        self.store
+            .inner()
+            .await
+            .map_err(|_| Status::internal("connection error"))?
+            .get_account_proofs(request)
```
Mirko-von-Leipzig, line 337: Would be good to have some way of handling this better. Maybe we shouldn't be returning `Status` as the error variant, but instead our own error that implements `IntoResponse`?

Could be done in a separate PR to keep the noise down.

crates/utils/src/grpc.rs
```diff
@@ -21,3 +25,78 @@ mod private {
...
+        }
+
+        let url = format!("http://{}", self.address);
+        let endpoint = Endpoint::try_from(url.clone())?;
+
+        let client = retry(ExponentialBackoff::default(), || async {
+            match endpoint.connect().await {
+                Ok(channel) => Ok((self.constructor)(channel)),
+                Err(e) => {
+                    error!("failed to connect to {}: {}", url, e);
```
Mirko-von-Leipzig, line 88:
```suggestion
                    error!(%url, %err, "gRPC connection failed, retrying");
```

Is there a way to access the current backoff time? If so, adding it as a structured field would be nice.

crates/utils/src/grpc.rs
```diff
@@ -21,3 +25,78 @@ mod private {
...
+                    Err(BackoffError::transient(e))
+                },
+            }
+        })
+        .await?;
+
+        *lock = Some(client.clone());
+        drop(lock);
+
+        info!("connection established to {}", url);
```
Mirko-von-Leipzig, line 98:
```suggestion
        info!(%url, "gRPC connection established");
```

crates/utils/src/grpc.rs
Mirko-von-Leipzig, line 1: You've mentioned that `tonic` already performs a retry when using the `lazy_connect` method. Do we still need this implementation then?

Something we should also think about is whether this behavior is something we _always_ desire. For example, maybe the RPC component shouldn't retry forever when a user makes an rpc request - they probably want an error at some point, or we should actually only do a single attempt.

^^ we can also handle with a timeout for the gRPC requests overall I think, so maybe this isn't such an issue..? Thoughts?

tests/integration/main.rs
```diff
@@ -0,0 +1,86 @@
+use miden_node_proto::generated::{
+    requests::GetBlockHeaderByNumberRequest, responses::GetBlockHeaderByNumberResponse,
+    rpc::api_client as rpc_client,
+};
+use miden_node_store::GenesisState;
+use tokio::{net::TcpListener, runtime, task};
+use tonic::transport::Endpoint;
+
+#[tokio::test]
+async fn test_components_startup() {
```
Mirko-von-Leipzig, line 10:
```suggestion
async fn component_startup_is_robust_to_network_failures() {
```

Mirko-von-Leipzig (commented):
Thanks, this is looking promising.

Some minor nits around error messages - these aren't really from your PR for the most part (messages should begin with lowercase).

And then a couple of questions.

crates/utils/src/grpc.rs
```diff
@@ -21,3 +25,78 @@ mod private {
...
+    /// Returns a clone of the inner client, connecting if needed.
+    ///
+    /// This method checks if a client is already available, and if not, attempts to establish
+    /// a connection using exponential backoff.
+    pub async fn inner(&self) -> Result<T, Error> {
+        if let Some(client) = self.inner.read().await.clone() {
+            return Ok(client);
+        }
+        self.connect().await
+    }
```
bobbinth, line 68: How would this work, for example, for RPC when we get multiple incoming requests? For example, let's say:

- Someone makes a call to the RPC component which will then try to access the store, but the store is unavailable, so we block until the connection becomes available - right?
- Someone else makes a call to RPC, because `self.inner` is locked, they will also block until the connection becomes available.

I guess in this case we'd need to rely on the timeouts as @Mirko-von-Leipzig mentioned https://github.com/0xPolygonMiden/miden-node/pull/785/files#r2048771843. Though, then we'd probably need to set the timeouts pretty aggressively (i.e., to under 5 seconds, which is probably fine either way).

The other option could be to return an error if the connection could not be re-established within some limit. This limit could be configurable (i.e., a field in `LazyClient`). For RPC to store calls, we could set the limit at 5 seconds. Then, after 5 seconds, we'd be able to respond to the RPC call with something like "store unavailable" error (rather than a generic timeout). For block producer to store connection the limit could be different (not sure what makes sense yet).

There could be some other options too. For example, `LazyClient` will immediately return an error on `LazyClient::inner()` call if the connection is broken, but will internally try to connect using some strategy. Not sure if this is worth the complexity though.

bobbinth (commented):
Thank you! Looks good. I left a couple of comments inline - the main one is about how to return errors to clients in case a connection is broken for a long period of time.

crates/utils/src/grpc.rs
TomasArrachea: > You've mentioned that tonic already performs a retry when using the lazy_connect method. Do we still need this implementation then?

When using the `lazy_connect`, the `tonic` client will try to establish a connection and return an error if the first attempt fails. It will then internally keep trying to reconnect. Instead, out LazyClient implementation blocks the caller until the connection is established, as this is the behaviour we need for initializing the clients.

I agree that we should probably return an error at some point if the connection is keeps failing. We could use some `backoff` setting for this, so that the `inner()` call returns a timeout error, or just handle timouts for the requests overall like you said.

A possible alternative would be to remove the `LazyClient` and just use the `tonic` client. This would require to add some retry logic when initializing the components, so that when a request to the store/block-producer fails, we retry until the tonic client reconnects. This way, all the following requests would just return an error at first attempt if some component is offline, and tonic would handle the reconnection in the backround.
Wdyt?

tests/integration/main.rs
TomasArrachea: I initially tried that but the server was still online. I think that `tonic` spawns internally another task for the server, so aborting my join handle did not actually aborted the second task and it was still running. This way, when killing the runtime all tasks are terminated.

crates/block-producer/src/server.rs
TomasArrachea: I agree that a config struct sound good. I added it in a1a83092e4af5f32f65db36e26ed83842a4309ca

crates/rpc/src/server/api.rs
TomasArrachea: Ok! I'll create the follow up once this is finished

crates/utils/src/grpc.rs
bobbinth: I think I might have misunderstood how it works. Here is my current understanding of how `tonic` works:

- When we try to establish a connection, if the connection fails, `tonic` would return an error and abort. But if we use `lazy_connect`, it would return an error but would still try to connect. Any attempt to use the connection during this time will fail.
- Once a connection is established, if the connection is dropped, `tonic` will automatically try to reconnect under the hood - but until the connection is restored, any attempt to use it, will result in an error.

If this is correct, then I think we can still keep using the `LazyClient` struct, but we may want to update it. I'm thinking the behavior we want is roughly as follows:

- We should be able to start the RPC component even before the store or block producer components are up and running. The RPC shouldn't block until these components are up. Instead, if user requests come in, it should immediately respond with errors. As soon as the downstream components are up, the store should start handling user requests correctly.
  - if the store or the block producer components go down, the store should start returning errors to the users, and then once the downstream components are back up, it should again start handling the requests.
  - The ideal scenario would be to handle store and block producer failures separately. So, for example, if the block producer goes down, `SubmitProvenTransaction` endpoint on the PRC should stop working, but all other endpoints that rely on the store only should still work.
- We should be able to start the block producer before the store is up and running. Here, I think it is OK to block until the store is up. This is because even to accept a transaction we need to go to the store. So, basically, the block producer would wait until the store is up to start accepting transactions and processing blocks.
  - If the connection to the store goes down, we try to reconnect - but should be responding with errors to any new requests that come in (and ideally shouldn't try to build batches/blocks - but if batch/block building just fails, that may be OK too).

TomasArrachea: Thanks for the detail!
> When we try to establish a connection, if the connection fails, tonic would return an error and abort. But if we use lazy_connect, it would return an error but would still try to connect. Any attempt to use the connection during this time will fail.

Sorry, I think I may have gotten confused at some point with this. The tonic `connect` and `lazy_connect` both block when trying to connect (the lazy_connect does it when the first request is made), they don't return an error right away. So I think that we should adapt the `LazyClient` to do so.

> Once a connection is established, if the connection is dropped, tonic will automatically try to reconnect under the hood - but until the connection is restored, any attempt to use it, will result in an error.

Correct.

> The RPC shouldn't block until these components are up. Instead, if user requests come in, it should immediately respond with errors

The `LazyClient` should attempt to connect with a small timeout (like 5 seconds). Upon failure, it will return the error and spawn a tokio task that keeps trying to connect the client.

> We should be able to start the block producer before the store is up and running. Here, I think it is OK to block until the store is up.

In this case, we could just use a bigger timeout on the LazyClient. I think this is similar to what you originally suggested.

TomasArrachea:
I made some changes taking into consideration to not block the consumers of the RPC api. Now, the RPC component connects to the Store and Block Producer using the `tonic` native client, with the `connect_lazy` method. This works as follows:
- Connections are lazy, they will be established when the RPC access those components for the first time.
- Reconnections are handled with exponential backoff internally by `tonic`.
- During reconnection, requests will return an error immediately. The client will still keep trying to connect in the backround. 

This way, the RPC will never be blocked waiting for connections with the other components. This will likely change if, at some point, the RPC requires startup calls to the other components.

On the other hand, the Block Producer connects to the Store using the `LazyClient`. 
- This wrapper blocks when the client tries to connect for the first time. This allows the block producer to be blocked on startup if the store is not online yet. Until that, the block producer will not start serving requests.
- After the first connection, all reconnections will be handled by `tonic` and will not be blocking, errors will be returned immediately.

TomasArrachea:
I think we will likely want to improve how the errors are returned to the RPC consumers, since right now we are just forwarding the tonic failure (e.g. "tcp connect error: Connection refused (os error 61)"), but we can leave that for later or even discuss it in https://github.com/0xPolygonMiden/miden-node/discussions/795

crates/block-producer/src/server.rs
```diff
@@ -30,153 +28,115 @@ use crate::{
...
+///
+/// Note: this blocks until one of the servers die.
+pub async fn serve(listener: TcpListener, config: BlockProducerConfig) -> anyhow::Result<()> {
+    info!(target: COMPONENT, endpoint=?listener, store=%config.store_address, "Initializing server");
+
+    let store = StoreClient::new(config.store_address);
+
+    // we drop the tpc listener so that the block producer cannot accept connections until the store is responding
+    let block_producer_address = listener.local_addr().context("failed to get address")?;
+    drop(listener);
```
TomasArrachea, line 55: This is needed to avoid the `submit_proven_transaction` endpoint of the RPC api to block in case that the block producer is blocked waiting for the store.

Another option would be to add a timeout on that request but I think this works fine.

crates/block-producer/src/server.rs
```diff
@@ -30,153 +28,113 @@ use crate::{
...
-            batch_budget: BatchBudget::default(),
-            block_budget: BlockBudget::default(),
-            state_retention: SERVER_MEMPOOL_STATE_RETENTION,
-            expiration_slack: SERVER_MEMPOOL_EXPIRATION_SLACK,
-            store,
-            rpc_listener: listener,
-            chain_tip,
+/// Serves the block-producer RPC API, the batch-builder and the block-builder.
+///
+/// Note: this blocks until one of the servers die.
```
Mirko-von-Leipzig, line 49: Nit: blocking means something specific and slightly different, i.e. blocking calls should never be run on async.

```suggestion
/// Note: Executes in place (i.e. not spawned) and will run indefinitely until
///       a fatal error is encountered.
```

crates/block-producer/src/server.rs
```diff
@@ -30,153 +28,113 @@ use crate::{
...
+pub async fn serve(config: BlockProducerConfig) -> anyhow::Result<()> {
+    info!(target: COMPONENT, endpoint=?config.block_producer_address, store=%config.store_address, "Initializing server");
+
+    let store = StoreClient::new(config.store_address);
+
+    // this call blocks initialization until the store responds
```
Mirko-von-Leipzig, line 55: > On the other hand, the Block Producer connects to the Store using the LazyClient.
>
> This wrapper blocks when the client tries to connect for the first time. This allows the block producer to be blocked on startup if the store is not online yet. Until that, the block producer will not start serving requests.
> After the first connection, all reconnections will be handled by tonic and will not be blocking, errors will be returned immediately.

I don't think is quite true. The `LazyClient` will always handle the reconnections at the moment - `tonic` is never in charge of this? Or at least its unclear to me where/when this would happen.

I think this is the _only_ place we currently need the auto-retry iiuc. What I suggest is that we drop the `LazyClient` entirely, and instead create a manual loop here, something like:

```rust
let chain_tip = loop {
    match store.latest_header().await {
        Err(ConnectionFailureErr) => {
            error!(%store.url, %retry.backoff, %retry.counter, "store connection failed while fetching chain tip, retrying");
            // backoff related code. can probably do our own simple math here with `.max` and doubling it.
            tokio::sleep().await
        }
        result => break result,
    }
}.context("failed to fetch chain tip from store")?;
```

tests/integration/main.rs
Mirko-von-Leipzig, line 1: wdyt of moving this to the RPC crate? Currently its only testing its startup. This avoids the extra crate and I think this could just be a normal test there.

It would also be nice if it could similarly test the `block-producer` connection via `submit_transaction` endpoint.

Similarly testing the block-producer startup is a bit trickier because it automatically produces blocks etc, so might not be trivially feasible.

crates/block-producer/src/server.rs
TomasArrachea: Yes, I think this is fine and it's probably simpler to just do it this way for now.

tests/integration/main.rs
TomasArrachea: Ok, I think it makes sense as an RPC tests.

> Similarly testing the block-producer startup is a bit trickier because it automatically produces blocks etc, so might not be trivially feasible.

I think there is no problem with starting the component, just that testing the `submit_proven_transaction` endpoint requires building a correct proven transaction

TomasArrachea: Added this test in [7a42eaa](https://github.com/0xPolygonMiden/miden-node/pull/785/commits/7a42eaa6ed2f87b3d5a72bb54b4086aa7b590ec4)

crates/block-producer/src/store/mod.rs
```diff
@@ -126,15 +130,21 @@ type InnerClient = store_client::ApiClient<InterceptedService<Channel, OtelInter
...
 }
 
 impl StoreClient {
-    /// TODO: this should probably take store connection string and create a connection internally
-    pub fn new(store: InnerClient) -> Self {
-        Self { inner: store }
+    pub fn new(store_address: SocketAddr) -> Result<Self, Error> {
+        let store = {
+            let store_url = format!("http://{store_address}");
+            let channel = tonic::transport::Endpoint::try_from(store_url)?.connect_lazy();
```
Mirko-von-Leipzig, line 142: imo we can unwrap on the error variant here because `store_url` is now guaranteed to be a correct url endpoint from the `format!` construction above

crates/block-producer/src/server.rs
```diff
@@ -260,3 +246,143 @@ impl BlockProducerRpcServer {
...
+        sleep(Duration::from_secs(1)).await;
+
+        let block_producer_client =
+            block_producer_client::ApiClient::connect(block_producer_endpoint)
+                .await
+                .expect("block producer client should connect");
+
+        // test: request against block-producer api should succeed
+        let response = send_request(block_producer_client.clone(), 0).await;
+        assert!(response.is_ok());
```
Mirko-von-Leipzig, line 343: I'm somewhat surprised the dummy transaction isn't rejected? Or is the dummy valid enough?

Mirko-von-Leipzig (commented):
Looking good; some minor comments/suggestions

crates/block-producer/src/server.rs
```diff
@@ -25,95 +23,86 @@ use crate::{
...
+                Err(StoreError::GrpcClientError(e)) => {
+                    // exponential backoff with base 500ms and max 30s
+                    let backoff = Duration::from_millis(500)
+                        .saturating_mul(1 << retries_counter)
+                        .min(Duration::from_secs(30));
 
-        info!(target: COMPONENT, "Server initialized");
+                    error!(
+                        store = %self.store_address,
+                        ?backoff,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Oh awesome, yeah that's fine.

crates/block-producer/src/server.rs
TomasArrachea: I think it's actually valid enough, though I guess it would eventually fail when creating the block because of invalid account updates

Mirko-von-Leipzig (approved):

crates/block-producer/src/server.rs
```diff
@@ -25,95 +23,86 @@ use crate::{
...
     store::StoreClient,
 };
 
-/// Represents an initialized block-producer component where the RPC connection is open,
-/// but not yet actively responding to requests.
-///
-/// Separating the connection binding from the server spawning allows the caller to connect other
-/// components to the store without resorting to sleeps or other mechanisms to spawn dependent
-/// components.
+/// Block producer's configuration.
```
igamigo, line 31: This could explain how the struct works and how it expects to be configured for communication with other components (similar to how it was before)

crates/rpc/src/server/mod.rs
```diff
@@ -11,44 +10,149 @@ use crate::COMPONENT;
 
 mod api;
 
-/// Represents an initialized rpc component where the RPC connection is open, but not yet actively
-/// responding to requests.
-///
-/// Separating the connection binding from the server spawning allows the caller to connect other
-/// components to the store without resorting to sleeps or other mechanisms to spawn dependent
-/// components.
```
igamigo, line 19: Would be good to keep a general overview of how it is initialized, what are the expectations, etc.

bin/node/src/commands/block_producer.rs
```diff
@@ -73,29 +73,25 @@ impl BlockProducerCommand {
...
-        BlockProducer::init(
-            listener,
-            store_url,
-            batch_prover_url,
-            block_prover_url,
+        BlockProducer {
+            block_producer_address,
+            store_address,
+            batch_prover: batch_prover_url,
+            block_prover: block_prover_url,
```
igamigo, line 88: nit: I think these fields could be named `BlockProducer::batch_prover_url` and `BlockProducer::block_prover_url` to make them a bit more consistent

igamigo (approved):
Looks good to me. Left a couple of minor comments; main ones being related to documenting the behavior of components related to the startup process. Is there any doc file that should be updated? In any case it could probably be done separately as well.

igamigo (approved):

crates/rpc/src/server/mod.rs
```diff
@@ -11,44 +10,154 @@ use crate::COMPONENT;
...
+            // Enables gRPC-web support.
+            .add_service(tonic_web::enable(api_service))
             .serve_with_incoming(TcpListenerStream::new(self.listener))
             .await
             .context("failed to serve RPC API")
     }
 }
+
+#[cfg(test)]
+mod test {
```
bobbinth, line 46: nit for the future: we usually name such modules `tests` and also put `TESTS` section separator above them.

crates/block-producer/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![recursion_limit = "256"]
```
bobbinth, line 1: I'd add a comment here explaining why this is needed.

crates/block-producer/src/server.rs
```diff
@@ -260,3 +251,153 @@ impl BlockProducerRpcServer {
         })
     }
 }
+
+#[cfg(test)]
+mod test {
```
bobbinth, line 256: Similar nit re naming and section headers as above.

PR #779, opened 2025-03-31
chore: update dependencies

TomasArrachea:
Addresses #773.

Updates:
- ~`tonic-web` -> v0.13~
- `opentelemetry` -> v0.29
- `opentelemetry-otlp` -> v0.29
- `opentelemetry_sdk` -> v0.29
- `tracing-opentelemetry` -> v0.30

Remains unupdated:
- `tonic`: needs next `opentelemetry` release.
- `tonic-build`: needs `tonic` update.
- `rusqlite`:  needs next `rusqlite_migration` release.

PR #786, opened 2025-04-10
feat: re-add `tonic-web` dependency

tomyrd:
In #779 we removed `tonic-web` but this was used to give support to web grcp calls from the web client. This PR re-adds this dependency (in version 0.12, same as `tonic`) and also updates the node to use the latest version of `miden-base`.

Just to keep in mind, it seems that `tonic_web::enable` is removed in version 0.13 so we'll have to look for its replacement once we decide to update.

### DIFF
--- a/bin/faucet/src/stub_rpc_api.rs
+++ b/bin/faucet/src/stub_rpc_api.rs
@@ -159,7 +159,7 @@ pub async fn serve_stub(endpoint: &Url) -> Result<(), ApiError> {
 
     tonic::transport::Server::builder()
         .accept_http1(true)
-        .add_service(api_service)
+        .add_service(tonic_web::enable(api_service))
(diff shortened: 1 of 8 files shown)


bin/faucet/src/stub_rpc_api.rs
tomyrd, line 1: I added the `tonic_web::enable` call here because it is the way it was before merging #779. But I'm not sure if the faucet will be receiving web grcp calls, maybe it isn't needed here.

igamigo: Not sure either but I think it should be fine to leave given [this issue](https://github.com/0xPolygonMiden/miden-node/issues/736).

igamigo (approved):

bobbinth (approved):

It would be great to add a test to make sure everything works as expected. But we could do it also in a follow-up PR.

PR #789, opened 2025-04-16
feat: crate to start node

SantiagoPittella:
On the client side, we used node to run integration tests, and when it came time to add it to the CI (and considering that we always rely on using the latest version available in the `next` branch), we decided to clone the repository and compile the binary. The latter caused long compilation times in the CI, and we had to add a sketchy cache. To resolve this, I created the `miden-node-builder` crate, which contains two main methods: starting the node with all its components and stopping it.


SantiagoPittella:
Closing it as i'm moving it to the client's repository.

Issue #662, opened 2025-01-31
Optimize note storage


SantiagoPittella:
### Node's components affected by this task

- [ ] RPC
- [ ] Block producer
- [x] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

With the new `miden-node-stress-testing` CLI, and in particular running the command `miden-node-stress-testing seed-store`, we saw that each note takes too much space in the DB (4.5KB). For ~100k notes means ~460.816384 MB. And, in our use case, represented more than 90% of the DB.

### How should it be done?

Investigate and improve the notes tables to shrink.
According to Bobbin (https://github.com/0xPolygonMiden/miden-node/pull/657#issuecomment-2628315344), it should be possible to reduce ~10x the size.

### When is this task done?

The task can be considered done when the size of each note in the store is reduced.

bobbinth:
By my estimate, all fields in the `notes` table, except for the `merkle_path` and `details`, should required about 72 bytes of storage. `merkle_path` should be about 0.5KB, and details should be about 2.5KB. This gives us about 3.1KB - so, not sure what else is there to get it to 4.5KB on average - but we should find this out.

For the 3.1KB, optimization opportunities are:

- Move `details` field into a separate table. In this table, don't record the metadata (we already have it in the `notes` table) and don't store note script separately for every note. Instead, we can have `note_scripts` table, and store every unique script there. This way, for things like P2ID notes, we should reduce the size of note details from 2.5KB to about 100 bytes.
- Use compact representation of Merkle paths. This should cut down storage from 0.5KB to about 200 bytes per path.

All in all, with these optimizations we should be able to go from 3.1KB to about 400 bytes. So, the big question will be is where the 1.4KB (i.e., 4.5 - 3.1) is coming from.

SantiagoPittella:
@bobbinth Should we add this issue to the milestone or it isn't planned for the current one?

bobbinth:
I think we can leave it for the next milestone. One of the tasks, btw, is actually in the `miden-crypto` repo (specifically, https://github.com/0xPolygonMiden/crypto/issues/215 for introducing `SparseMerklePath` struct).

TomasArrachea:
> So, the big question will be is where the 1.4KB (i.e., 4.5 - 3.1) is coming from.

Actually the latest outputs on the seed-store gave an avg of 2 kb/entry for the notes table. Querying the db I got this avg sizes per column (in bytes):
```
[...]

PR #790, opened 2025-04-16
feat: optimize note scripts storage

TomasArrachea:
Addresses #662 

This PR updates the db schemas to have the note scripts normalized. The `details` column in the notes table was removed. Instead, the columns `inputs`, `assets`, `serial_num`, and `script_root` were added. The `script_root` acts as FK to a new `scripts` table.

Also, the NoteRecord was updated. Instead of holding the notes' `details` as an array of bytes, it now holds a `NoteDetails` struct.


crates/store/src/state.rs
```diff
@@ -345,7 +345,9 @@ impl State {
             .output_notes()
             .map(|(note_index, note)| {
                 let (details, nullifier) = match note {
-                    OutputNote::Full(note) => (Some(note.to_bytes()), Some(note.nullifier())),
```
TomasArrachea, line 348: I'm not sure why the whole note was serialized and stored as `details` here, maybe it was outdated code. Maybe that caused the entries on the notes table having bigger size than expected.

crates/store/src/db/migrations/001-init.sql
```diff
@@ -58,6 +62,13 @@ CREATE INDEX idx_notes_tag ON notes(tag, block_num);
 CREATE INDEX idx_notes_nullifier ON notes(nullifier);
 CREATE INDEX idx_unconsumed_network_notes ON notes(execution_mode, consumed);
 
+CREATE TABLE scripts (
+    script_root BLOB NOT NULL,
+    script      BLOB NOT NULL,
+  
+    PRIMARY KEY (script_root)
+) STRICT, WITHOUT ROWID;
```
bobbinth, line 70: nit: I would probably call this table `note_scripts`.

crates/store/src/state.rs
bobbinth: Yeah - I don't think this was the intended behavior.

igamigo: IIRC, it was a quick way to support public notes because it was before having the current set of note structs, though I could definitely be completely off ([this PR](https://github.com/0xPolygonMiden/miden-node/pull/300) came to mind)

crates/store/src/db/sql/mod.rs
```diff
@@ -1234,6 +1275,7 @@ pub fn apply_block(
     // Note: ordering here is important as the relevant tables have FK dependencies.
     count += insert_block_header(transaction, block_header)?;
     count += upsert_accounts(transaction, accounts, block_header.block_num())?;
+    count += insert_scripts(transaction, &notes.iter().map(|(note, _)| note).collect::<Vec<_>>())?;
```
Mirko-von-Leipzig, line 1278: We can avoid the `collect` call here if `insert_scripts` takes an iterator instead

crates/store/src/db/mod.rs
```diff
@@ -108,8 +116,26 @@ impl NoteRecord {
...
         let merkle_path = MerklePath::read_from_bytes(merkle_path_data)?;
-        let details_data = row.get_ref(10)?.as_blob_or_null()?;
-        let details = details_data.map(<Vec<u8>>::read_from_bytes).transpose()?;
+
+        let assets = row.get_ref(10)?.as_blob_or_null()?;
+        let inputs = row.get_ref(11)?.as_blob_or_null()?;
+        let script = row.get_ref(12)?.as_blob_or_null()?;
+        let serial_num = row.get_ref(13)?.as_blob_or_null()?;
+
+        let details = if let (Some(assets), Some(inputs), Some(script), Some(serial_num)) =
```
igamigo, line 125: Maybe unnecessary but I wonder if there should be a `debug_assert` here that either all of these are `Some` or all of these are `None`

igamigo (approved):
LGTM!

crates/store/src/db/mod.rs
```diff
@@ -122,6 +122,10 @@ impl NoteRecord {
         let script = row.get_ref(12)?.as_blob_or_null()?;
         let serial_num = row.get_ref(13)?.as_blob_or_null()?;
 
+        debug_assert!(
+            !(assets.is_some() ^ inputs.is_some() ^ script.is_some() ^ serial_num.is_some()),
```
Mirko-von-Leipzig, line 126: This doesn't work - xor is unfortunately not chainable in this manner, e.g.

`true ^ false ^ true ^ false` would pass the above assert.

TomasArrachea: oh right! thanks

PR #796, opened 2025-04-22
refactor: Update faucet data store

igamigo:
As part of the migration to the newest revision of `next`:

- updated the test utils to implement the new `NetworkAccount` parameters
- updated the faucet `DataStore`

### DIFF
--- a/bin/faucet/src/store.rs
+++ b/bin/faucet/src/store.rs
@@ -1,13 +1,12 @@
-use std::sync::Mutex;
+use std::{collections::BTreeSet, sync::Mutex};
 
 use miden_objects::{
-    Word,
+    MastForest, Word,
     account::{Account, AccountId},
     block::{BlockHeader, BlockNumber},
-    note::NoteId,
-    transaction::{ChainMmr, InputNotes, TransactionInputs},
+    transaction::{ChainMmr, TransactionScript},
 };
-use miden_tx::{DataStore, DataStoreError};
+use miden_tx::{DataStore, DataStoreError, MastForestStore, TransactionMastStore};
 
 pub struct FaucetDataStore {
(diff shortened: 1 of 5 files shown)


bin/faucet/src/store.rs
```diff
@@ -44,27 +48,38 @@ impl FaucetDataStore {
     pub fn update_faucet_state(&self, new_faucet_state: Account) {
         *self.faucet_account.lock().expect("Poisoned lock") = new_faucet_state;
     }
+
+    /// Updates the stored faucet account with the new one.
+    pub fn load_transaction_script(&self, tx_script: &TransactionScript) {
+        // TODO: because the script string is interpolated, the MAST is different and needs to be
+        // loaded each time. Maybe it should be compiled once and inputs could be passed some other
+        // way
+        self.mast_store.insert(tx_script.mast().clone());
```
bobbinth, line 57: Yes, there is an issue in `miden-base` to enable providing inputs to transaction scripts (https://github.com/0xPolygonMiden/miden-base/issues/1262) - but this is not yet implemented.

bin/faucet/src/store.rs
```diff
@@ -44,27 +48,38 @@ impl FaucetDataStore {
...
 }
 
 impl DataStore for FaucetDataStore {
     fn get_transaction_inputs(
         &self,
         account_id: AccountId,
-        _block_ref: BlockNumber,
-        _notes: &[NoteId],
-    ) -> Result<TransactionInputs, DataStoreError> {
+        _ref_blocks: BTreeSet<BlockNumber>,
```
bobbinth, line 65: We don't care about reference blocks because we always execute against the genesis block?

bobbinth (approved):
Looks good! Thank you!

Issue #725, opened 2025-03-03
Otel testing


Mirko-von-Leipzig:
Figure out a reasonable way to test our open-telemetry spans and attributes.

Some investigation was done in #678 (but was reverted as out-of-scope):

- https://github.com/0xPolygonMiden/miden-node/pull/698#discussion_r1957237032
- https://github.com/0xPolygonMiden/miden-node/pull/698#discussion_r1966444689

## Concerns

Ideally we would:

1. Attach a test span exporter (which collects the spans in memory)
2. Run some code
3. Assert on the resulting spans

The main issue iiuc is that the exporter is registered _globally_ which makes running such tests.. a bit difficult.

We also don't want these otel tests to interfere with other tests which don't care about otel (but do create traces).

We could separate out these tests entirely, and force them to run sequentially. This might be a good use case for `cargo xtask`. Or we treat them as pure integration tests and create a full setup. The downside of this is complexity and inability to specify very niche scenarios.

sergerad:
@Mirko-von-Leipzig my vote is to do separate integration tests for this. Seems simpler to have the tests isolated to the single concern of trace content.

Linking relevant (reverted) commits from the previous PR:
- [Fixture](https://github.com/0xPolygonMiden/miden-node/pull/698/commits/8c195dbbc6900ddb429efc7c8760fe79e19baf75)
- [Setup](https://github.com/0xPolygonMiden/miden-node/pull/698/commits/099764ea656d18b07ed07befe6b591ac55331643)

PR #799, opened 2025-04-23
Add trace unit tests for block-producer mempool

sergerad:
Relates to #725

### Changes
- Adds support for testing trace content via `miden_node_utils::logging::setup_test_tracing()`.
- Imports the [serial_test](https://crates.io/crates/serial_test) crate for testing purposes.
- Adds tests to block-producer crate's mempool module and fixes instrumentation typo.

### DIFF
--- a/crates/block-producer/src/mempool/tests.rs
+++ b/crates/block-producer/src/mempool/tests.rs
@@ -1,5 +1,6 @@
 use miden_objects::block::BlockNumber;
 use pretty_assertions::assert_eq;
+use serial_test::serial;
 
 use super::*;
 use crate::test_utils::{MockProvenTxBuilder, batch::TransactionBatchConstructor};
@@ -14,6 +15,55 @@ impl Mempool {
             u32::default(),
         )
     }
+}
+
+// OTEL TRACE TESTS
+// ================================================================================================
+
+#[tokio::test]
+#[serial(open_telemetry_tracing)]
+async fn add_transaction_traces_are_correct() {
+    let (mut rx_export, _rx_shutdown) = miden_node_utils::logging::setup_test_tracing().unwrap();
+
+    let mut uut = Mempool::for_tests();
+    let txs = MockProvenTxBuilder::sequential();
+    uut.add_transaction(txs[0].clone()).unwrap();
+
+    let span_data = rx_export.recv().await.unwrap();
+    assert_eq!(span_data.name, "mempool.add_transaction");
+    assert!(span_data.attributes.iter().any(|kv| kv.key == "code.namespace".into()
+        && kv.value == "miden_node_block_producer::mempool".into()));
+    assert!(
+        span_data
+            .attributes
+            .iter()
+            .any(|kv| kv.key == "tx".into() && kv.value.to_string().starts_with("0x"))
+    );
+}
+
+#[tokio::test]
+#[serial(open_telemetry_tracing)]
+async fn revert_transactions_traces_are_correct() {
+    let (mut rx_export, _rx_shutdown) = miden_node_utils::logging::setup_test_tracing().unwrap();
+
+    let mut uut = Mempool::for_tests();
+    let txs = MockProvenTxBuilder::sequential();
+    uut.add_transaction(txs[0].clone()).unwrap();
+    let span_data = rx_export.recv().await.unwrap();
+    assert_eq!(span_data.name, "mempool.add_transaction");
+
+    uut.revert_transactions(vec![txs[0].id()]).unwrap();
+    let span_data = rx_export.recv().await.unwrap();
+    assert_eq!(span_data.name, "mempool.revert_transactions");
+    assert!(span_data.attributes.iter().any(|kv| kv.key == "code.namespace".into()
+        && kv.value == "miden_node_block_producer::mempool".into()));
+
+    assert!(
+        span_data
+            .attributes
+            .iter()
+            .any(|kv| kv.key == "transactions.expired.ids".into())
+    );
 }
 
 // BATCH FAILED TESTS
--- a/crates/utils/src/logging.rs
+++ b/crates/utils/src/logging.rs
@@ -57,6 +57,30 @@ pub fn setup_tracing(otel: OpenTelemetry) -> Result<()> {
     tracing::subscriber::set_global_default(subscriber).map_err(Into::into)
 }
 
+/// Initializes tracing to a test exporter.
+///
+/// Allows trace content to be inspected via the returned receiver.
+///
+/// All tests that use this function must be annotated with `#[serial(open_telemetry_tracing)]`.
+/// This forces serialization of all such tests. Otherwise, the tested spans could
+/// be interleaved during runtime. Also, the global exporter could be re-initialized in
+/// the middle of a concurrently running test.
+#[cfg(feature = "testing")]
+pub fn setup_test_tracing() -> Result<(
+    tokio::sync::mpsc::UnboundedReceiver<opentelemetry_sdk::trace::SpanData>,
+    tokio::sync::mpsc::UnboundedReceiver<()>,
+)> {
+    let (exporter, rx_export, rx_shutdown) =
+        opentelemetry_sdk::testing::trace::new_tokio_test_exporter();
+
+    let otel_layer = open_telemetry_layer(exporter);
+    let subscriber = Registry::default()
+        .with(stdout_layer().with_filter(env_or_default_filter()))
+        .with(otel_layer.with_filter(env_or_default_filter()));
+    tracing::subscriber::set_global_default(subscriber)?;
+    Ok((rx_export, rx_shutdown))
+}
+
 fn open_telemetry_layer<S>(
     exporter: impl SpanExporter + 'static,
 ) -> Box<dyn tracing_subscriber::Layer<S> + Send + Sync + 'static>
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2000,6 +2000,7 @@ dependencies = [
  "pretty_assertions",
  "rand 0.9.1",
  "rand_chacha 0.9.0",
+ "serial_test",
  "tempfile",
  "thiserror 2.0.12",
  "tokio",
@@ -2128,6 +2129,7 @@ dependencies = [
  "opentelemetry_sdk",
  "rand 0.9.1",
  "serde",
+ "tokio",
  "tonic",
  "tracing",
  "tracing-forest",
@@ -3296,6 +3298,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "scc"
+version = "2.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "22b2d775fb28f245817589471dd49c5edf64237f4a19d10ce9a92ff4651a27f4"
+dependencies = [
+ "sdd",
+]
+
+[[package]]
 name = "schannel"
 version = "0.1.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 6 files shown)


crates/utils/src/logging.rs
```diff
@@ -57,6 +57,27 @@ pub fn setup_tracing(otel: OpenTelemetry) -> Result<()> {
...
 
+#[cfg(feature = "testing")]
+pub fn setup_test_tracing() -> Result<(
+    tokio::sync::mpsc::UnboundedReceiver<opentelemetry_sdk::trace::SpanData>,
+    tokio::sync::mpsc::UnboundedReceiver<()>,
+)> {
+    static INIT: std::sync::Once = std::sync::Once::new();
+    INIT.call_once(|| {
+        opentelemetry::global::set_text_map_propagator(TraceContextPropagator::new());
+    });
```
sergerad, line 68: @Mirko-von-Leipzig my understanding of your point from the issue around global state leads me to believe this is an appropriate solution. Every test that wants to validate traces needs to call this function in any case.

> The main issue iiuc is that the exporter is registered globally which makes running such tests.. a bit difficult.

The only problem would be if we had parallel tests that ran `setup_tracing()`. But I don't think that will ever be the case.

LMK if I misunderstood anything.

crates/block-producer/src/mempool/tests.rs
```diff
@@ -16,6 +16,55 @@ impl Mempool {
...
+        && kv.value == "miden_node_block_producer::mempool".into()));
+
+    // TODO: Why do these fail? Am I misunderstanding something?
+    assert!(span_data.attributes.iter().any(|kv| kv.key == "transactions.ids".into()));
+    assert!(
+        span_data
+            .attributes
+            .iter()
+            .any(|kv| kv.key == "transactions.expired.ids".into())
+    );
```
sergerad, line 65: @Mirko-von-Leipzig I was expecting these to be here as per the above tests, but something must be different - do you know what it is?

```rust
    #[instrument(target = COMPONENT, name = "mempool.revert_transactions", skip_all, fields(transactions.ids))]
    fn revert_transactions(
        &mut self,
        txs: Vec<TransactionId>,
    ) -> Result<(), GraphError<TransactionId>> {
        tracing::Span::current().record("transactions.expired.ids", tracing::field::debug(&txs));
```

sergerad:
@Mirko-von-Leipzig LMK which modules you want me to cover with these kinds of tests. I think we should treat them as smoke tests to some extent - rather than testing every single span/field/attribute, validate key ones for some set of modules.

crates/block-producer/src/mempool/tests.rs
Mirko-von-Leipzig: Its because `tracing` doesn't let you dynamically add new properties. We've added `transactions.ids` as the field name, but are using a different one here.

Probably a copy-paste mistake/bug :)

Tests putting in some work already :partying_face:

crates/utils/src/logging.rs
Mirko-von-Leipzig: > The only problem would be if we had parallel tests that ran setup_tracing(). But I don't think that will ever be the case.

What prevents them from running tests concurrently? By default `cargo test` would run tests concurrently, so I imagine this function call would somehow have to block until the previous test is completed its usage of the span exporter?

Or maybe I'm missing something :)

crates/block-producer/src/mempool/tests.rs
Mirko-von-Leipzig, line 1: Replying here so that we can thread the convo.

> @Mirko-von-Leipzig LMK which modules you want me to cover with these kinds of tests.

Ideally we would test the spans we want to formally stabilize and document/support in the [operator guide](https://0xpolygonmiden.github.io/miden-node/operator/monitoring.html#what-gets-traced). Right now that's batch and block build events but would also include the RPC gRPC methods eventually.

> I think we should treat them as smoke tests to some extent - rather than testing every single span/field/attribute, validate key ones for some set of modules.

Broadly agreed. Though I have some ideas on how we might grow coverage. I'll start a separate comment thread to discuss these, if it gets more involved we can move it elsewhere.

For now lets I'd like to just know that we get the right outputs for a batch and block build success event. The hardest part here might be generating the right data, e.g. to guard against bugs like https://github.com/0xPolygonMiden/miden-node/issues/793 which would require checking actual counts etc. Additionally confirming that an error event gets logged as an error span at the root and origin spans (or whatever that means in OTel land).

crates/utils/src/logging.rs
sergerad: Yea there is another global lower in that function that I didn't notice. Looks like the tests will indeed clobber. Think best thing is to force serialization of the trace tests. Used this in the past https://crates.io/crates/serial_test

crates/block-producer/src/mempool/tests.rs
sergerad: h u g e

crates/utils/src/logging.rs
Mirko-von-Leipzig: Ah nice yeah that looks like a good solution! I like that it even let's you group serial tests together.

crates/utils/src/logging.rs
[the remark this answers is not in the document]
sergerad: Just had a look at insta's documentation and struggled to get any context on what a snapshot actually is :D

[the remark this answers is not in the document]
Mirko-von-Leipzig: That's fair, did you check their [quickstart](https://insta.rs/docs/quickstart/)? I found their crate documentation to be vague, but their website made things more clear.

crates/block-producer/src/mempool/tests.rs
sergerad: > For now lets I'd like to just know that we get the right outputs for a batch and block build success event.

Then need to figure out how to make the relevant structs unit testable. I can make a separate issue for this but might be worth sparring on some diffs on this PR before doing so. Will @ you when I push some diffs to consider.

sergerad: Yea there doesn't seem to be a way to unit test the batch / block builder stack. Lets discuss as to how we want to test this kind of thing. Easiest thing would be to make integration tests that involve in-process component instances - unsure that is something we do today?

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -395,3 +395,41 @@ impl BlockProver {
...
+            BatchBudget::default(),
+            BlockBudget::default(),
+            5,
+            u32::default(),
+        );
+        let _block = block_builder.build_block(&mempool).await;
+        let span_data = rx_export.recv().await.unwrap();
+        assert_eq!(span_data.name, "block_builder.build_block");
+    }
+}
```
sergerad, line 435: @Mirko-von-Leipzig here is an example test which fails because certain components are either not mocked or dependancies are not running in the background.

E.G. the above will get connection refused due to the store client not being mocked or a store not running.

LMK if you think its better to try mock or do integration tests for this.

crates/utils/src/logging.rs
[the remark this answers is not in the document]
sergerad: Yea I did read that. Only context it seems to give is that snapshots are strings haha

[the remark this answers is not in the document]
Mirko-von-Leipzig: Heh okay let me try summarize my understanding.

Snapshots in this context are the expected values serialized in some way in order to persist them. Often this means strings because json/yaml/toml are convenient ways to provide persistence that is human auditable.

The main upside is that it makes it simpler to generate expected values which might otherwise be tedious, or prone to changing arbitrarily. The development loop might look something like

```rust
#[test]
fn snapshot_test_example() {
    /// Initially this will be empty or contain an invalid value.
    /// This can be stored inline like here, or in a file if too verbose/large.
    const snapshot: Json = json!{}

    // Drive your test setup etc

    // Serialize your test output to match the snapshot type.
    // This feels bad, but it makes creating the snapshot easier
    // than manually constructing actual rust structs for complex things.
    let test_output = test_output.into_json().unwrap();

    // This will fail on first run, but it will also show 
    // you what json to copy-paste in the json! macro.
    //
    // Whenever the test fails, we can inspect the new json
    // and update the snapshot value if it makes sense.
    assert_eq!(test_output, snapshot);
}
```

`insta` provides tooling and convenient APIs that make maintaining and running snapshot based tests easier.

- handles file IO
- handles serialization
- supports filtering/redacting parts of the data that are prone to change e.g. if the data includes a timestamp
- cli to inspect, diff and update the snapshots

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: My gut reaction was to suggest integration tests. This is more upfront work but we would benefit from being able to easily create integration tests imo.

However one thing that would be difficult to integration test is failure conditions (since this would require bugs). Currently we have this simulated failure rate, but I don't really like that its part of our code here :D I initially added it to enable a bit of chaos-monkey style testing but we've never actually used it.

I suppose we could force failures into the store by manually corrupting the database as part of the test. Other parts are less.. forceable but maybe that's good enough?

------------

Alternatives are mocking, either by implementing a fake store, or by adding more specific traits that the tests could implement more readily.

Or even creating a real in-memory store and inserting some blocks as needed. This is really more integration testing though, so it sort of falls into the top category. The pain here is really generating correct blocks that are useful though maybe I'm overthinking it.

In general I don't like mocks or adding in traits just for the sake of testing, but I'm probably a bit overly dogmatic here.

------------

Maybe take a look into the `MockChain` from `miden-base` and see if it could help us generate real data to seed the store with. If it looks easyish then I think lets try a more integration style test. If its overly painful then either we should switch focus to improving the integration test situation or we do the quick "fix" with mocks.

The other reason integration tests would be preferred is that it would add in the OTel context spans from the store which we otherwise lose.

crates/block-producer/src/mempool/tests.rs
Mirko-von-Leipzig: > Easiest thing would be to make integration tests that involve in-process component instances - unsure that is something we do today?

Nope, we currently have no real integration tests. The closest is the out-of-process client tests. Enabling this would be valuable imo. Question is just _when_ do we do this.

sergerad:
@Mirko-von-Leipzig just noting this [commit](https://github.com/0xPolygonMiden/miden-node/commit/6b156f14dd6bec2a985060e71db90e3946caca84) which adds the `BlockProducerStore` trace and feeds it through the stack.

Next I'm going to remove the trait, keep the initialization of the store before the stack its used in, and try using fakes instead of proceeding with trait + mocks. This way we don't introduce a trait purely for testing purposes. Going to try [faux](https://github.com/nrxus/faux).

sergerad:
I tried both [faux](https://docs.rs/faux/latest/faux/) and [mry](https://docs.rs/mry/latest/mry/) which are mocking frameworks that work by faking structs rather than mocking traits.

Unfortunately they are only so far along and both fall apart due to the arguments to this function (and probably others):
```rust
    pub async fn get_block_inputs(
        &self,
        updated_accounts: impl Iterator<Item = AccountId> + Send,
        created_nullifiers: impl Iterator<Item = Nullifier> + Send,
        unauthenticated_notes: impl Iterator<Item = NoteId> + Send,
        reference_blocks: impl Iterator<Item = BlockNumber> + Send,
    ) -> Result<BlockInputs, StoreError> {
```

mry's macro falls apart due to those iterators. And faux requires the iterators to implement `Debug` which I thought was a rabbit hole we wouldn't want to go down.

sergerad:
Tried using mockall with the trait back in place. Found that the mockall macro required:
1. Methods to not use `impl Iterator<...`; and
2. The `Iterator` types to be `'static`.

Screenshot 2025-04-28 at 12 30 17 PM

The former made the code more verbose while the latter broke it altogether, e.g.:
```
rustc: `batches` does not live long enough borrowed value does not live long enough
hint: argument requires that `batches` is borrowed for `'static`
hint: `batches` dropped here while still borrowed
```

Given this, I would suggest we focus this PR to implement whatever trace tests are possible without mocks. And then we move on to working on the `MockChain` capabilities and integration tests for traces and more generally. @Mirko-von-Leipzig  LMK what you think about that. TY!

crates/block-producer/src/mempool/tests.rs
```diff
@@ -16,6 +17,55 @@ impl Mempool {
     }
 }
 
+// OTEL TRACE TESTS
+// ================================================================================================
+
+#[tokio::test]
+#[serial]
```
Mirko-von-Leipzig, line 24: Should we use a specific key for this? If yes, could you add it to the `setup_test_tracing` docs as well.

```suggestion
#[serial(otel)]
```

Mirko-von-Leipzig (approved):

PR #805, opened 2025-04-25
feat: emit network note information

### LINKED ISSUE #697: Emit network notes and transaction status
The `mempool` should monitor committed blocks for created network notes. These notes should be submitted via gRPC (#695).

Similarly, the `mempool` should also monitor for network transactions that have been dropped/reverted from the pool, or committed. These status updates should also be sent via gRPC. 

These can currently only identified if they _consume_ a network note - however this information would have been erased since network transactions would register these as unauthenticated..

An initial implementation could simply assume that such a method exists by adding an `is_network` to `AuthenticatedTransaction` and default to `false` such that nothing happens in prod yet.

One alternative is `add_network_transaction` to `block-producer`'s RPC methods, and to simply trust the builder component. However this can wait until the broader question of how to identify network accounts is resolved.

Mirko-von-Leipzig:
Closes #697.

This PR adds a network transaction client which is used by the block-producer to submit network notes and transaction status changes whenever a block is built (or fails to build - transactions get reverted on block failure for example).

I'm not entirely happy with the shape of things, and I'm unsure if the gRPC definitions are what we actually want.

I've decided to open it up for an initial review pass at least, but it definitely still needs some work.


crates/proto/src/ntx_builder/mod.rs
Mirko-von-Leipzig, line 1: I really like how this client turned out, and that its part of the proto crate.

If we repeat something similar for other clients, this would mean less generated proto getting leaked outside of this crate. Though this is a lot more work for the others.

The client methods are still missing instrumentation.

crates/proto/src/domain/note.rs
```diff
@@ -8,6 +10,18 @@ use crate::{
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct CommittedNote {
+    pub block_num: BlockNumber,
+    // TODO: can this be BlockNoteIndex? If so, how to convert to an absolute index.
+    pub note_index: usize,
+    pub note_id: Digest,
+    pub metadata: NoteMetadata,
+    pub details: Option<Vec<u8>>,
+    pub merkle_path: MerklePath,
+}
```
Mirko-von-Leipzig, line 23: Because I haven't implemented the actual network tx builder yet, I'm unsure if this is what is needed on that end. I also don't know how to get some of these values, e.g. the merkle path and note index.

crates/block-producer/src/server.rs
```diff
@@ -55,30 +55,40 @@ impl BlockProducer {
...
-
-        let store = store_client::ApiClient::with_interceptor(channel, OtelInterceptor);
+            .with_context(|| format!("failed to connect to store on {store}"))?;
+        let store = store_client::ApiClient::with_interceptor(store, OtelInterceptor);
         let store = StoreClient::new(store);
 
+        let ntx_client = miden_node_proto::ntx_builder::Client::lazy_with_interceptor(
+            network_tx_builder_address,
+            OtelInterceptor,
+        );
```
Mirko-von-Leipzig, line 78: This feels much better than the above wrangling.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -85,14 +100,26 @@ impl BlockBuilder {
...
             interval.tick().await;
 
-            self.build_block(&mempool).await;
+            // Non-fatal errors are handled internally by the block building process.
+            //
+            // As such, any error returned here should be treated as fatal.
+            if let Err(err) = self.build_block(&mempool).await {
+                tracing::error!(%err, "fatal error while building a block, aborting block production");
+                return;
+            }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: A more robust medium term solution could be to include a counter of some kind, so that the ntx builder knows if its missed something.

This becomes a lot harder to handle _if_ we decide to emit new network notes as they arrive via txs (this impl emits them only once they've been committed to the store).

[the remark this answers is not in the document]
Mirko-von-Leipzig: A further complication is if/when outsiders can submit transactions to consume network notes. The block-producer has no idea where the networks txs arrive from so this isn't something we can prevent.

This does mean our gRPC methods aren't adequate, and we should also be communicating the network note IDs that are consumed (currently we only emit committed tx IDs).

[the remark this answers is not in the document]
bobbinth: > This becomes a lot harder to handle _if_ we decide to emit new network notes as they arrive via txs (this impl emits them only once they've been committed to the store).

I think we'll need to do this, but probably further down the road.

> A further complication is if/when outsiders can submit transactions to consume network notes.

I think we'll need to do this sooner rather than later. And I agree, that this will require some refactoring of the NTX builder's API. Maybe it could work as follows:

- When a transaction enters mempool, block producer makes a request to NTX builder. This request would be similar to our current `SubmitNetworkNotesRequest` request, but would also include all nullifiers consumed in the transaction.
- When a transaction is either committed or reverted, block producer would make a `UpdateTransactionStatusRequest` (nothing changes here).

To handle the case of NTX builder crushing and then coming back online, we'll probably need to add a way for the NTX builder to query the block producer. I would do it in a separate PR, but it could work something like this:

- When NTX builder restarts, it first initializes itself with the data from the store. Once this is done, NTX builder will know that it's data is current up to block $x$.
- Next, NTX builder would make a request to the block producer to get in-flight info passing $x$ as the parameter. Since block producer keeps info about recently committed blocks, it should be able to reply with all relevant data (assuming $x$ is not too far in the past).

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -223,20 +252,93 @@ impl BlockBuilder {
...
+                        return None;
+                    }
+                    // TODO: ensure it is a network note.
+                    CommittedNote {
+                        block_num: built_block.header().block_num(),
+                        note_index: todo!("Do I really have to iterate over the batches?"),
+                        note_id: note.id().inner(),
+                        metadata: *note.metadata(),
+                        details: NoteDetails::from(note).to_bytes().into(),
+                        merkle_path: todo!("How do this?"),
```
bobbinth, line 283: We can probably simplify this for now (and maybe forever) and not include Merkle path with notes. This way, NTX builder would always create transactions with unauthenticated notes, and then these would be authenticated during batch building.

This also implies that we probably don't need to include most fields here and can just send a simple `Note` (which would consist of `NoteDetails` and `NoteMetadata`).

crates/proto/src/domain/note.rs
bobbinth: As mentioned in another comment, I think we may be able to get away with just sending notes (rather than including things like Merkle path, note index etc.).

bobbinth (commented):
Looks good! Thank you! Not a full review, but I left some comments inline.

Mirko-von-Leipzig:
@igamigo I have some additional changes in `mirko/ntb`, but they're really minor.

Mirko-von-Leipzig:
As mentioned in the sync call, I think building the ntb as a separate worker thread in the block-producer would be a good first iteration.

This side-steps gRPC definitions and restart/desync problems entirely. Communication can happen via channel, and if this breaks just crash the process, restarting both.

Once this is more settled we can move it out as a separate process, with more info on how the gRPC should look etc.

PR #807, opened 2025-04-28
Use `NullifierTree` and `BlockChain` from base

PhilippGackstatter:
Use `NullifierTree` and `BlockChain` from base.

Addresses the breaking changes from https://github.com/0xPolygonMiden/miden-base/pull/1301, https://github.com/0xPolygonMiden/miden-base/pull/1304 and https://github.com/0xPolygonMiden/miden-base/pull/1305 in one go.

Already addresses the `AccountTree` changes mentioned in https://github.com/0xPolygonMiden/miden-node/issues/806, fyi @TomasArrachea.

Marked as breaking change in changelog due to renames in proto files.


crates/proto/src/generated/responses.rs
```diff
@@ -125,7 +125,7 @@ pub struct GetBlockInputsResponse {
     /// above note inclusion proofs as well as proofs for inclusion of the requested blocks
     /// referenced by the batches in the block.
     #[prost(bytes = "vec", tag = "3")]
-    pub chain_mmr: ::prost::alloc::vec::Vec<u8>,
+    pub partial_block_chain: ::prost::alloc::vec::Vec<u8>,
```
Mirko-von-Leipzig, line 128: I wonder if we shouldn't just call these `block_inclusion_proofs` or something similar.

Mirko-von-Leipzig (approved):

crates/proto/src/generated/responses.rs
PhilippGackstatter: We don't store individual proofs here but rather an entire data structure, so I think the way it is now is more accurate. It also wouldn't be incorrect to call it block inclusion proofs, but it would suggest it contains individual proofs like for nullifier or account witnesses, which is not quite accurate.

PR #808, opened 2025-04-28
chore: remove unused code

### LINKED ISSUE #806: Remove unused files
~https://github.com/0xPolygonMiden/miden-base/pull/1301 added some minor changes to the account tree. This apparently breaks some test utilities here:~

> ~In miden-node, this breaks test code in crates/block-producer/src/test_utils/store.rs but the fix should be relatively easy~

^^ Already handled by #807 

-----

I've also noticed we still have some pre-mempool vestiges floating around in the block-producer i.e. `state-view` and some other src files that are in-tree but not part of the module system. We should remove those.

Similarly, we should check if this `test_utils/store.rs` is actively used - I suspect not.

TomasArrachea:
Closes #806

This PR removes unused code: 
- the `state-view` directory that holded old code.
- unused public functions
- unused block producer testing utils


Mirko-von-Leipzig (approved):
Awesome thank you, always great to have a negative line-count PR :D

Did you end up using that [cargo-workspace-unused-pub](https://github.com/cpg314/cargo-workspace-unused-pub) tool?

TomasArrachea:
> Did you end up using that [cargo-workspace-unused-pub](https://github.com/cpg314/cargo-workspace-unused-pub) tool?

Yes! It was useful.

PR #809, opened 2025-04-29
Use unchecked `PartialBlockChain` constructor

PhilippGackstatter:
Use the unchecked `PartialBlockChain` constructor. See inline comments for a short explanation.

Should probably only be merged after https://github.com/0xPolygonMiden/miden-base/pull/1308.


Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
> Use the unchecked `PartialBlockChain` constructor. See inline comments for a short explanation.
> 
> Should probably only be merged after [0xPolygonMiden/miden-base#1308](https://github.com/0xPolygonMiden/miden-base/pull/1308).

Feel free to merge whenever you think is good then :)

PR #817, opened 2025-04-30
feat: add `/status` endpoint

### LINKED ISSUE #794: Add status page for miden node
Recently, in `miden-base`, an status endpoint was introduced for the `miden-proving-service` (issue: https://github.com/0xPolygonMiden/miden-base/issues/1264, PR: https://github.com/0xPolygonMiden/miden-base/pull/1287). This new endpoint returns the status of the proxy and its assigned workers formatted as a JSON:

```json
{
  "version": "0.8.0",
  "prover_type": "Transaction",
  "workers": [
    {
      "address": "0.0.0.0:50051",
      "version": "0.8.0",
      "status": {
        "Unhealthy": {
          "failed_attempts": 3,
          "reason": "Unsupported prover type: Batch"
        }
      }
    },
    {
      "address": "0.0.0.0:50052",
      "version": "0.8.0",
      "status": "Healthy"
    },
  ]
}
```

The current use-case is semi-internal, meaning that this endpoint returns information useful for the developers and leaks some of the internal implementation of the components. It is not supposed to show any metrics.

As @bobbinth  proposed in https://github.com/0xPolygonMiden/miden-node/issues/788#issuecomment-2816995142, it could have this form:

```json
{
    "version": "0.8.0",
    "chain_tip": 123456,
    "block_producer": {
        "version": "0.8.0",
        "status": "connected",
    },
    "store": {
        "version": "0.8.0",
        "status": "connected",
    },
}
```

TomasArrachea:
Closes #794.

This PR adds a new `/status` endpoint on the store, block-producer and RPC components. The RPC endpoint hits the store and block-producer and aggregates their status into a single status response:

```
{
    "version": "0.9.0",
    "store_status": {
        "version": "0.9.0",
        "status": "connected",
        "chain_tip": 684
    },
    "block_producer_status": {
        "version": "0.9.0",
        "status": "connected"
    }
}
```

This endpoint can be easily accesed using Postman or grpcurl.

To run using grpcurl: 
```
grpcurl -import-path ./proto/proto -proto rpc.proto -plaintext localhost:57291 rpc.Api.Status
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - [BREAKING] Use `AccountTree` and update account witness proto definitions (#783).
 - Enabled running RPC component in `read-only` mode (#802).
 - [BREAKING] Update name of `ChainMmr` to `PartialBlockchain` (#807).
+- Added gRPC `/status` endpoint on all components (#817).
 
 ## v0.8.0 (2025-03-26)
 
--- a/bin/faucet/src/stub_rpc_api.rs
+++ b/bin/faucet/src/stub_rpc_api.rs
@@ -11,8 +11,8 @@ use miden_node_proto::generated::{
     responses::{
         CheckNullifiersByPrefixResponse, CheckNullifiersResponse, GetAccountDetailsResponse,
         GetAccountProofsResponse, GetAccountStateDeltaResponse, GetBlockByNumberResponse,
-        GetBlockHeaderByNumberResponse, GetNotesByIdResponse, SubmitProvenTransactionResponse,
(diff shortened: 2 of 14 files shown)


sergerad (approved):
Nice work. Thanks for including the `grpcurl` command. Branch works as expected 👍

bobbinth (approved):
Looks good! Thank you!

We should also update documentation (at least [here](https://github.com/0xMiden/miden-node/blob/next/docs/src/user/rpc.md) but maybe also worth mentioning it in other places).

PR #818, opened 2025-05-04
docs: update old GitHub and GitHub Pages URLs to new 0xMiden naming

Himess:
## Describe your changes

This PR updates outdated URLs across all Markdown files to reflect the repository renaming from `0xPolygonMiden` to `0xMiden`.  
It also updates GitHub Pages links from `0xpolygonmiden.github.io` to `0xMiden.github.io`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Miden node
 
-[![LICENSE](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/0xPolygonMiden/miden-node/blob/main/LICENSE)
-[![test](https://github.com/0xPolygonMiden/miden-node/actions/workflows/test.yml/badge.svg)](https://github.com/0xPolygonMiden/miden-node/actions/workflows/test.yml)
+[![LICENSE](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/0xMiden/miden-node/blob/main/LICENSE)
+[![test](https://github.com/0xMiden/miden-node/actions/workflows/test.yml/badge.svg)](https://github.com/0xMiden/miden-node/actions/workflows/test.yml)
 [![RUST_VERSION](https://img.shields.io/badge/rustc-1.85+-lightgray.svg)](https://www.rust-lang.org/tools/install)
 [![crates.io](https://img.shields.io/crates/v/miden-node)](https://crates.io/crates/miden-node)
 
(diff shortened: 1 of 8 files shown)


bobbinth (approved):
Looks good! Thank you!

I have a slight preference for using `0xMiden` rather than `0xmiden`. Would you make these change?

Himess:
I will update it as requested shortly, thank you.

Himess:
@sergerad @bobbinth n I’ve made the changes you suggested, could you please review again?

README.md
```diff
@@ -20,16 +20,16 @@ Access to the network is provided via a gRPC interface which can be found [here]
 ## Documentation
 
 Documentation, tutorials and guides for the current Miden version (aka testnet) can be found
-[here](https://0xpolygonmiden.github.io/miden-docs/), including an operator manual and gRPC reference guide. This is
+[here](https://0xMiden.github.io/miden-docs/), including an operator manual and gRPC reference guide. This is
```
sergerad, line 23: @Himess sorry I was equivocal - the domain / hostname should be lowercase because DNS is case insensitive. Its the path in the URL that can be case sensitive depending on the backend.

So
```
https://0xmiden.github.io/...
https://github.com/0xMiden
```

Himess: Hi @sergerad , I’ve made the changes you suggested, could you please review again?

sergerad (approved):

README.md
```diff
@@ -20,16 +20,16 @@ Access to the network is provided via a gRPC interface which can be found [here]
...
 
 For node operators living on the development edge, we also host the latest unreleased documentation
-[here](https://0xpolygonmiden.github.io/miden-node/index.html).
+[here](https://0xmiden.github.io/miden-node/index.html).
 
 ## Contributing
 
 Developer documentation and onboarding guide is available
-[here](https://0xpolygonmiden.github.io/miden-node/developer/index.html).
+[here](https://0xmiden.github.io/miden-node/developer/index.html).
```
bobbinth, line 32: Could we use `0xMiden` here instead of `0xmiden`?

README.md
bobbinth: @sergerad - `0xMiden` works here just as well as `0xmiden` - or am I missing something?

README.md
Himess: https://github.com/0xMiden/miden-node/pull/818#discussion_r2072674596
He said to me this. So i changed.

README.md
sergerad: It does work but DNS names are technically all lower case - the capitalization is ignored.

Which is comparison to the URL's path, which requires capitalization depending on the backend involved.

I would keep the DNS name lowercase for those reasons. Capitalized works too, its just a bit misleading.

Himess:
@sergerad @bobbinth Should i do something ? Or just wait to merge ? Or add Changelog ?

README.md
bobbinth: @Himess - @sergerad and I discussed this offline and let's always go with `0xMiden`. This may technically not be correct in case of hostnames, but it simplifies things a bit without having any negative consequences.

Himess: @bobbinth I fixed now ser. https://github.com/0xMiden/miden-node/pull/818/commits/bc06014a53f9900d5cdebc1dbb1c9e3d74e50661

bobbinth (approved):
All looks good! Thank you!

PR #819, opened 2025-05-04
feat: major faucet refactor

Mirko-von-Leipzig:
This is a near re-write of the faucet to (hopefully) address the issues experienced during the post-announcement influx of users. Its difficult to be certain because we don't actually get useful logging at present.

For a synopsis of the issues, please see #815.

This PR is against main since this is a new release.

----------

### Updated branding

Branding now reflects the new logo etc.

### TLS support

This PR enables TLS support for the gRPC client to the node RPC.

As part of infra migration, the faucet now attempts to use the node RPC via https. However I failed to realise that the faucet doesn't actually support TLS.

We either need to fix the infra with direct http access, or _need_ this PR.

### Faucet thread

The faucet client now runs in its own thread. This addresses a (potential) source of desync, wherein a user could cancel their request midway through. This could result in the chain receiving the request, but the local state now being updated to reflect this.

Instead now the request runs to completion entirely in its own thread, without user cancellation.

### Desync detection

The faucet can now detect and repair descync'd state. This could occur even now if the node drops a faucet transaction e.g. due to tx expiration, or node restart.

The faucet now stores the last ~1000 states and rolls back to one of those when a desync is detected.

Detection is implemented by inspecting the gRPC error code & message. This required a change on the node side - we weren't actually sending a useful error message. **This means we need to update the node to take advantage of this**.

### Restructure

File structure is now more sensible imo. Code is also more isolated.

<details><summary>tree bin/faucet</summary>

```
bin/faucet/src
├── config.rs
├── faucet
│   ├── distribute_fungible_asset.masm
│   ├── mod.rs
│   └── store.rs
├── main.rs
├── rpc_client.rs
├── server
│   ├── frontend.rs
│   ├── get_tokens.rs
│   ├── mod.rs
│   └── resources
│       ├── background.png
│       ├── favicon.ico
│       ├── index.css
│       ├── index.html
│       └── index.js
├── stub_rpc_api.rs
└── types.rs
```

</details>

### Type safety

Several new types have been added to ensure invariants are upheld. This is mostly for peace of mind. As an example, the faucet thread takes an `AssetAmount`, which can only created by validating the user's requested amount through an `AssetOptions`. Which in turn is constructed from the config so the faucet is now guaranteed that this is an allowed amount, and that it can be safety cast to a fungible asset.

Similarly we now differentiate between faucet and general account IDs.

### Logs

I've both reduced and improved the logs (imo). Request's are now logged only once they complete or fail.

Proving and execution was also spamming a lot, I've removed those.

This is _not_ instrumented for OTel. That can be a separate issue.

### Next steps 

Quite a few, but probably best to list them in the discussion.

There are also a few clean up items here imo, e.g. probably some `pub` and types that are unnecessary or in the wrong places.

----------

> [!WARNING]
>
> This requires a node update to be fully functional.
> 
> I can't really test this under load.


CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Changelog
 
+## Unreleased
+
+## v0.8.2 (2025-05-04)
```
Mirko-von-Leipzig, line 5: Bit awkward with release versioning where we have both faucet and node.

Probably we can just.. keep them in sync regardless of changes for now.

We would need to change all version numbers if we really care about this now I guess.

crates/block-producer/src/errors.rs
```diff
@@ -113,7 +113,7 @@ impl From<AddTransactionError> for tonic::Status {
             )
             | AddTransactionError::Expired { .. }
             | AddTransactionError::TransactionDeserializationFailed(_) => {
-                Self::invalid_argument(value.to_string())
+                Self::invalid_argument(value.as_report())
```
Mirko-von-Leipzig, line 116: This is the magic sauce to return an more indepth error via gRPC. `to_string` just provides the top level error, whereas now we return a full error report.

I haven't really tested this in-depth. So need to keep an eye out. The faucet should still error out now even if this is wrong, so it won't go undetected.

bin/faucet/src/faucet/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+
+        Ok((account_delta, block_height, p2id_note.into_inner()))
+    }
+
+    fn execute_transaction(&self, tx_args: TransactionArgs) -> MintResult<ExecutedTransaction> {
+        // TODO: Is this cheap? Do we need to carry this around with us, or can we just construct
+        //       when needed?
+        TransactionExecutor::new(self.data_store.clone(), Some(self.authenticator.clone()))
+            .execute_transaction(self.id.inner(), BlockNumber::GENESIS, &[], tx_args)
+            .map_err(MintError::Execution)
```
Mirko-von-Leipzig, line 322: This thing isn't `Send` nor `Sync`. Which made using it in async really impossible. For now I just reconstruct it, but is there some better way to actually use it?

Is construction expensive, or is this fine? The data store is also.. really annoying and requires `Arc` for reasons unknown to me?

bin/faucet/src/faucet/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    fn prove_transaction(tx: ExecutedTransaction) -> MintResult<ProvenTransaction> {
+        LocalTransactionProver::new(ProvingOptions::default())
+            .prove(tx.into())
+            .map_err(MintError::Proving)
+    }
```
Mirko-von-Leipzig, line 329: I  didn't find time for remote prover support.

We should add a general client for it. Right now the proto definitions are in `base` and I didn't feel motivated to try bring them in here myself.

bin/faucet/src/faucet/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+    BuildingP2IdNote(#[source] NoteError),
+    #[error("compiling the tx script failed")]
+    ScriptCompilation(#[source] TransactionScriptError),
+    #[error("execution of the tx script failed")]
+    Execution(#[source] TransactionExecutorError),
+    #[error("proving the tx failed")]
+    Proving(#[source] TransactionProverError),
+    #[error("submitting the tx to the node failed")]
+    Submission(#[source] RpcError),
+}
```
Mirko-von-Leipzig, line 97: I'm unsure how many of these can actually error?

For example, I feel like the note build and compilation should always succeed. But I was unsure so I err'd on the side of caution.

bin/faucet/src/faucet/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+            .context("fetching genesis header from the node")?;
+
+        // SAFETY: An empty chain MMR should be valid.
+        let genesis_chain_mmr = ChainMmr::new(
+            PartialMmr::from_peaks(
+                MmrPeaks::new(0, Vec::new()).expect("Empty MmrPeak should be valid"),
+            ),
+            Vec::new(),
+        )
+        .expect("Empty ChainMmr should be valid");
```
sergerad, line 159: Unrelated to this PR - should those crypto types provide a default impl or zero() -like constructors so we don't have to unwrap/expect in these situations?

bin/faucet/src/faucet/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+            prior_state: VecDeque::new(),
+        })
+    }
+
+    /// Runs the faucet minting process until the request source is closed, or it encounters a fatal
+    /// error.
+    pub async fn run(
+        mut self,
+        mut rpc_client: RpcClient,
+        mut requests: mpsc::Receiver<(MintRequest, oneshot::Sender<(BlockNumber, Note)>)>,
```
sergerad, line 190: nit: turn the tuple into a struct at some point.

bin/faucet/src/faucet/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+            }
+
+            match self.handle_request(request, rng, &mut rpc_client).await {
+                // Update local state on success.
+                Ok((delta, block_number, note)) => {
+                    // We ignore the channel closure here as the user may have cancelled the
+                    // request.
+                    let _ = response_sender.send((block_number, note));
+                    // SAFETY: Delta must be valid since it comes from a tx accepted by the node.
+                    self.update_state(&delta).unwrap();
```
sergerad, line 209: Would the faucet recover from this failure on restart?

bin/faucet/src/faucet/mod.rs
sergerad: Yea thats a shame. I'm hoping its purely due to the trait objects in `TransactionExecutor` that don't specify `Send` and `Sync`. Unsure if this stack depends on generated gRPC code which seems to be an issue [here](https://github.com/0xMiden/miden-client/pull/868#issuecomment-2848811369).

bin/faucet/src/main.rs
```diff
@@ -107,46 +93,36 @@ async fn run_faucet_command(cli: Cli) -> anyhow::Result<()> {
...
-            axum::serve(listener, app).await.unwrap();
+            let mut rpc_client =
+                RpcClient::connect_lazy(&config.node_url).context("failed to create RPC client")?;
+            let account_file = AccountFile::read(&config.faucet_account_path)
+                .context("failed to load faucet account from file")?;
+
+            let faucet = Faucet::load(account_file, &mut rpc_client).await?;
+
+            // Maximum of 100 requests in-queue at once. Overflow is rejected for faster feedback.
+            let (tx_requests, rx_requests) = mpsc::channel(100);
```
sergerad, line 105: In future might be something we want configurable so as to update without creating a release

sergerad:
Looks good. Appreciate the explainer/safety comments.

Do you think there are any unit tests that we can implement today (without mocks) to cover code paths relating to desync?

sergerad (approved):

bin/faucet/src/faucet/mod.rs
igamigo: Yeah, adding bounds for `Send` and `Sync` in the traits or trait objects should do the trick here. We did this a couple times on branches we didn't end up merging for other reasons but [should be a simple fix](https://github.com/0xMiden/miden-base/compare/main...tomyrd-send-sync-traits#diff-cdc577a9fd28ef742a84be42bb71bbe8193f7a3d6b0d9f4773affba863334159) if we want to add that separately.

bin/faucet/src/faucet/mod.rs
Mirko-von-Leipzig: That would be lovely - @PhilippGackstatter any chance of an empty/default maybe?

bin/faucet/src/faucet/mod.rs
Mirko-von-Leipzig: On restart the faucet reloads the latest committed state from the node.

This will work in the majority of cases, however this could lead to a mismatch due to transactions still in the mempool. In this case, the faucet would again restart on the first submitted tx (which would fail due to desync). This repeats until all txs have cleared the mempool.

There isn't really a way around that without direct mempool access. And I think this would be a rare case not really worth the bother. At least not yet.

bin/faucet/src/main.rs
Mirko-von-Leipzig: Yeah I was lazy :/ I want to replace the config file with cli args, but that would be a breaking change. And then its easier to add this in.

bin/faucet/src/faucet/mod.rs
Mirko-von-Leipzig: That would be great; they're really rather awkward :D 

Along with the `Arc` requirement in there for the datastore.

bin/faucet/src/faucet/mod.rs
sergerad: I'm happy to make this change. Keep me sane while I do the terraform stuff 😃

sergerad: https://github.com/0xMiden/crypto/pull/409

Then need to update base for `ChainMmr` I suppose.

bin/faucet/src/faucet/mod.rs
igamigo: I'm not sure the `Arc` requirement is too easy to get rid of without extra changes. Or did you mean making it a generic instead of a trait object? I have mentioned before that maybe we could make the executor just take a reference to the datastore on `execute_transaction()` instead of owning it. Initially the datastore was used to load the account code before actually executing, but as this is not needed anymore maybe this can change. Or maybe we could do something like [alternative 2 from this comment.](https://github.com/0xMiden/miden-base/issues/938#issue-2614938080)

bobbinth: One thing that is not clear to me is why `handle_request()` function needs to be `async`. All requests to `TransactionExecutor::execute_transaction()` should be serialized, and so, by extension, all requests to `handle_request()` should be serialized. Wouldn't this imply that `handle_request()` should be a synchronous function?

sergerad: `handle_request()` calls `async fn submit_transaction()` which I presume involves network I/O. Wouldn't want to block on that.

bin/faucet/src/faucet/mod.rs
Mirko-von-Leipzig: Thanks!

We need to be careful about changing base (or any further deps). This can cause issues with the client if we pull in any unforseen changes.

I'd rather do it as part of trunk for the next release.

bin/faucet/src/faucet/mod.rs
Mirko-von-Leipzig: Yeah its only the node IO that is async.

We could move that out to a separate async thread/task again via a channel which would change that. And that might be a good thing since a lot of the work is blocking.

I'll try that out.

I don't really understand why any of these structs need to be stateful so I can't really way in on it.

Mirko-von-Leipzig:
I've tried channels and a new runtime to separate the blocking and async IO in the faucet, but without success. There will be a way to do this, but I'm out of time :)

Leaving as is now.

bin/faucet/src/server/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        let account_rate_limiter = GovernorConfigBuilder::default()
+            .const_burst_size(1)
+            .const_per_second(10)
+            // The default extractor uses the peer address which is incorrect
+            // if used behind a proxy.
+            .key_extractor(AccountKeyExtractor)
+            .finish()
+            .unwrap();
+        let account_rate_limiter = Arc::new(account_rate_limiter);
```
Mirko-von-Leipzig, line 93: These rate limit by

- IP address, 8 requests which refill by 1 per second
- account ID, 1 requests, refills after 10s

The latter is only applied to the `get_tokens` endpoint.

To enable this I needed to change how the `get_tokens` works. It no longer takes in json, but rather inputs are via query params.

bin/faucet/src/faucet/mod.rs
igamigo: @bobbinth would you have any reservations about doing something like what @tomyrd did on the [PR I linked above](https://github.com/0xMiden/miden-base/compare/main...tomyrd-send-sync-traits#diff-cdc577a9fd28ef742a84be42bb71bbe8193f7a3d6b0d9f4773affba863334159)? Otherwise, I believe we could make this change.

Issue #686, opened 2025-02-12
Add system stats to telemetry


Mirko-von-Leipzig:
We probably want to track at least the following:

- storage disk space used and available
- memory use and remaining
- cpu usage and info

Ideal would also be to list our own process's usage specifically, including

- database size
- database WAL size
- block storage used
- memory used
- cpu times etc

Remember to check out the open-telemetry naming specification for [system stats](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/) (yes, the page is about metrics, but we're injecting metrics into wide-events so let's stick with the conventions).

## Implementation

Two options I can think of:

1. Background thread that periodically posts these stats as an event
2. Inject them into every span we emit

I'm assuming getting this information requires syscalls, and as such its not reasonable to do (2) in the actual span processor. Instead we likely still want a background thread to periodically update the values.

https://docs.rs/sysinfo/latest/sysinfo/ is a popular option I found, but maybe do some more research first. We want something low-cost. I also see that macOs could be problematic due to security permissions - we'll have to check that out as well.

Mirko-von-Leipzig:
@TomasArrachea I'm assigning you this one, but let me know if its not suitable - I'm fairly uncertain on how to actually approach this correctly, so maybe its underspecified/unclear.

You can find some Miden node open-telemetry information [here](https://0xpolygonmiden.github.io/miden-node/developer/monitoring.html).

PR #828, opened 2025-05-07
feat: add system monitor

TomasArrachea:
Addresses https://github.com/0xMiden/miden-node/issues/686.

This PR introduces a `SystemMonitor` struct as a module of the node binary. This struct runs a loop where it collects system metrics and posts those stats in a tracing event. To collect the system metrics, it uses [sysinfo](https://docs.rs/sysinfo/latest/sysinfo/). The monitor is run in a separate thread; it spawns when running the bundled node, or on the start command of each component independently.

If a block-producer or rpc component is started independently, the store stats will be not be shown.

### Metrics
(open to suggestions on more useful metrics to add)
```
info!(
    target: COMPONENT,
    // System memory
    system_memory_limit,
    system_memory_available,
    system_memory_usage,
    system_memory_utilization,
    // System CPU
    system_cpu_physical_count,
    system_cpu_utilization,
    // Disks
    system_disk_limit,
    system_disk_state_available,
    system_disk_utilization,
    // Process
    process_memory_usage,
    process_cpu_utilization,
    process_disk_written,
    process_disk_read,
    // Node store data
    db_file_size,
    db_wal_size,
    block_storage_size,
);
```

### Example output:
```
2025-05-07T18:56:40.237769Z  INFO system-monitor: bin/node/src/system_monitor.rs:75: system_memory_limit: 25769803776, system_memory_available: 9130770432, system_memory_usage: 14384185344, system_memory_utilization: 0.5581798553466797, system_cpu_physical_count: 14, system_cpu_utilization: 0.025455639, system_disk_limit: 1989325168640, system_disk_state_available: 1197322598932, system_disk_utilization: 39.8126250144139, process_memory_usage: 23003136, process_cpu_utilization: 0.0053307414, process_disk_written: 57344, process_disk_read: 0, db_file_size: 110592, db_wal_size: 3802792, block_storage_size: 96
```

### DIFF
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -213,7 +213,9 @@ impl BundledCommand {
         // Start system monitor.
         let data_dir =
             DataDirectory::load(data_directory.clone()).context("failed to load data directory")?;
-        std::thread::spawn(move || SystemMonitor::new(Some(data_dir), monitor_interval).run());
+        std::thread::spawn(move || {
+            SystemMonitor::new(monitor_interval).with_store_metrics(data_dir).run();
+        });
 
         // Lookup table so we can identify the failed component.
         let component_ids = HashMap::from([
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -9,8 +9,9 @@ use tokio::{net::TcpListener, task::JoinSet};
 use url::Url;
 
 use super::{
-    DEFAULT_BATCH_INTERVAL_MS, DEFAULT_BLOCK_INTERVAL_MS, ENV_BATCH_PROVER_URL,
-    ENV_BLOCK_PROVER_URL, ENV_DATA_DIRECTORY, ENV_ENABLE_OTEL, ENV_RPC_URL, parse_duration_ms,
+    DEFAULT_BATCH_INTERVAL_MS, DEFAULT_BLOCK_INTERVAL_MS, DEFAULT_MONITOR_INTERVAL_MS,
+    ENV_BATCH_PROVER_URL, ENV_BLOCK_PROVER_URL, ENV_DATA_DIRECTORY, ENV_ENABLE_OTEL, ENV_RPC_URL,
+    parse_duration_ms,
 };
 use crate::system_monitor::SystemMonitor;
 
@@ -79,6 +80,15 @@ pub enum BundledCommand {
             value_name = "MILLISECONDS"
         )]
         batch_interval: Duration,
+
+        /// Interval at which to monitor the system in milliseconds.
+        #[arg(
+            long = "monitor.interval",
+            default_value = DEFAULT_MONITOR_INTERVAL_MS,
+            value_parser = parse_duration_ms,
+            value_name = "MILLISECONDS"
+        )]
+        monitor_interval: Duration,
     },
 }
 
@@ -104,6 +114,7 @@ impl BundledCommand {
                 open_telemetry: _,
                 block_interval,
                 batch_interval,
+                monitor_interval,
             } => {
                 Self::start(
                     rpc_url,
@@ -112,6 +123,7 @@ impl BundledCommand {
                     block_prover_url,
                     batch_interval,
                     block_interval,
+                    monitor_interval,
                 )
                 .await
             },
@@ -125,6 +137,7 @@ impl BundledCommand {
         block_prover_url: Option<Url>,
         batch_interval: Duration,
         block_interval: Duration,
+        monitor_interval: Duration,
     ) -> anyhow::Result<()> {
         // Start listening on all gRPC urls so that inter-component connections can be created
         // before each component is fully started up.
@@ -200,7 +213,7 @@ impl BundledCommand {
         // Start system monitor.
         let data_dir =
             DataDirectory::load(data_directory.clone()).context("failed to load data directory")?;
-        std::thread::spawn(move || SystemMonitor::new(Some(data_dir)).run());
+        std::thread::spawn(move || SystemMonitor::new(Some(data_dir), monitor_interval).run());
 
         // Lookup table so we can identify the failed component.
         let component_ids = HashMap::from([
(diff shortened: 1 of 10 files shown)


bin/node/src/system_monitor.rs
```diff
@@ -0,0 +1,116 @@
+use std::time::Duration;
+
+use miden_node_store::DataDirectory;
+use sysinfo::{Disk, Disks, Pid, System};
+use tracing::{error, info};
+
+const COMPONENT: &str = "system-monitor";
+const MONITOR_INTERVAL: Duration = Duration::from_secs(10);
```
sergerad, line 8: We will likely want to control this number. Do we want to make it configurable?

bin/node/src/system_monitor.rs
```diff
@@ -0,0 +1,116 @@
...
+            db_file_size,
+            db_wal_size,
+            block_storage_size,
+        );
+
+        Ok(())
+    }
+
+    /// Collects the store metrics.
+    fn collect_store_metrics(&self) -> anyhow::Result<(u64, u64, u64)> {
```
sergerad, line 104: nit: Prefer to return a struct rather than a tuple here. Mainly because of the way it would facilitate adding more store-related metrics in the future.

bin/node/src/system_monitor.rs
TomasArrachea: I added it as a command line argument

bin/node/src/system_monitor.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// Collects the system metrics.
+    fn collect_system_metrics(&mut self) -> anyhow::Result<()> {
+        self.sys.refresh_all();
+
+        let disks = Disks::new_with_refreshed_list();
+        let system_disk_limit: u64 = disks.iter().map(Disk::total_space).sum();
+        let system_disk_state_available: u64 = disks.iter().map(Disk::available_space).sum();
+        let system_disk_used = system_disk_limit - system_disk_state_available;
+        #[allow(clippy::cast_precision_loss)]
+        let system_disk_utilization = (system_disk_used as f64 / system_disk_limit as f64) * 100.0;
```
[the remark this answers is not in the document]
TomasArrachea: Right, I changed it to show only the metrics of the disk where the current working dir is mounted, I think makes more sense this way

bin/node/src/commands/bundled.rs
```diff
@@ -192,6 +210,13 @@ impl BundledCommand {
             })
             .id();
 
+        // Start system monitor.
+        let data_dir =
+            DataDirectory::load(data_directory.clone()).context("failed to load data directory")?;
+        std::thread::spawn(move || {
+            SystemMonitor::new(monitor_interval).with_store_metrics(data_dir).run();
```
igamigo, line 217: Should this task be monitored somehow? As it is right now, if it fails for any reason we will never know and this could be misleading when troubleshooting.

bin/node/src/commands/rpc.rs
```diff
@@ -56,6 +72,9 @@ impl RpcCommand {
             .await
             .context("Failed to bind to RPC's gRPC URL")?;
 
+        // Start system monitor.
+        std::thread::spawn(move || SystemMonitor::new(monitor_interval).run());
```
igamigo, line 76: If this is always going to run on its own thread would it make sense to include the dispatch within `SystemMonitor`

bin/node/src/commands/bundled.rs
TomasArrachea: Added a supervisor task that runs the monitor in [ddcf447](https://github.com/0xMiden/miden-node/pull/828/commits/ddcf4474feec35d7b126cfd1e03f8097272010be), wdyt?

bin/node/src/system_monitor.rs
```diff
@@ -41,22 +41,33 @@ impl SystemMonitor {
...
-    /// Runs the system monitor loop on a separate thread.
-    pub fn run(mut self) {
-        std::thread::spawn(move || {
+    /// Runs a supervisor task that re-runs the system monitor if it panics.
+    pub fn run_with_supervisor(self) -> task::JoinHandle<()> {
+        task::spawn(async move {
             loop {
-                std::thread::sleep(self.monitor_interval);
-                if let Err(err) = self.log_system_metrics() {
+                if let Err(err) = self.clone().run().await {
```
igamigo, line 48: I think I'd add a sleep for some seconds here on the loop so that if for some reason it fails to restart, it does not retry as fast as possible.

igamigo (approved):
LGTM! Left just a small comment

bin/node/src/system_monitor.rs
```diff
@@ -0,0 +1,154 @@
...
+/// - Disk usage
+/// - Process memory, CPU and disk usage
+/// - Database size
+/// - Database WAL size
+/// - Block storage size
+#[derive(Clone)]
+pub struct SystemMonitor {
+    pid: Pid,
+    monitor_interval: Duration,
+    data_directory: Option<DataDirectory>,
```
tomyrd, line 23: We should add some documentation on how the monitor behaves if this is `None`

bin/node/src/system_monitor.rs
```diff
@@ -0,0 +1,154 @@
...
+            %system_disk_utilization,
+            // Process
+            process_memory_usage,
+            %process_cpu_utilization,
+            process_disk_written,
+            process_disk_read,
+            // Store
+            db_file_size=store_metrics.db_file,
+            db_wal_size=store_metrics.db_wal,
+            block_storage_size=store_metrics.block_storage,
```
tomyrd, line 127: Should this be shown even if no store is set? If not, is there a way to show a `-` instead of `0` for this case?

tomyrd (approved):

bin/node/src/system_monitor.rs
TomasArrachea: I changed the struct to hold the metrics as options, so when set to None, they will not be shown in the log.

bobbinth (approved):
Looks good! Thank you!

sergerad (approved):

Issue #822, opened 2025-05-06
Improve faucet throughput


Mirko-von-Leipzig:
The faucet currently creates one tx per token request.

We could instead bundle multiple requests into one tx. #819 already sets us up for this, all we need to do is

- take multiple requests from the input channel, and
- augment the tx script to include multiple p2id notes

bobbinth:
We should still probably limit the number of output notes per request. Maybe at something like 16 or 32.

igamigo:
Closed by #829

PR #829, opened 2025-05-07
feat: batch faucet requests

TomasArrachea:
Closes https://github.com/0xMiden/miden-node/issues/822

This PR refactors the faucet to take mint requests from the channel using [recv_many](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv_many). This way, the mint requests are batched and multiple notes are created in one single transaction.

Also increased the queue size of the faucet from 100 to 1000.


sergerad:
Hey @TomasArrachea I'm just wondering if we should merge https://github.com/0xMiden/miden-node/commit/39c90a0cd85c4cb5c4de3d0969ddeea588e2cc71 into `next` before doing any more on the fuacet. Otherwise merge will just get harder. I think a lot of restructuring occurred there.

#819 is the PR for reference.

igamigo:
> Hey @TomasArrachea I'm just wondering if we should merge [39c90a0](https://github.com/0xMiden/miden-node/commit/39c90a0cd85c4cb5c4de3d0969ddeea588e2cc71) into `next` before doing any more on the fuacet. Otherwise merge will just get harder. I think a lot of restructuring occurred there.
> 
> #819 is the PR for reference.

Is the plan not to re-deploy the faucet after the whole refactor (#815) is finished? If so I'd say we want to keep merging to `main` and after it we could merge to `next`. Since the faucet is very self-contained it should be simple to bring the changes anyway but maybe I'm missing something.

sergerad:
> Is the plan not to re-deploy the faucet after the whole refactor (#815) is finished? If so I'd say we want to keep merging to `main` and after it we could merge to `next`.

AFAIK the only reason that was merged straight to `main` was due to time constraints to get it out. Which is why I would like to merge ASAP. I might be mistaken on that assumption, however.

> Since the faucet is very self-contained it should be simple to bring the changes anyway but maybe I'm missing something.

Happy to defer to you guys on this, I'm not familiar enough to weigh in on that.

TomasArrachea:
> Happy to defer to you guys on this, I'm not familiar enough to weigh in on that.

Sure. I agree that it makes sense to finish the refactors on main and when it's done we'll merge it altogether into next.

bin/faucet/src/faucet/mod.rs
```diff
@@ -187,24 +190,37 @@ impl Faucet {
...
-        while let Some((request, response_sender)) = requests.recv().await {
-            // Skip doing work if the user no longer cares about the result.
-            if response_sender.is_closed() {
-                tracing::info!(request.account_id=%request.account_id, "request cancelled");
-                continue;
-            }
+        let mut rng = RpoRandomCoin::new(coin_seed.map(Felt::new));
+
+        let mut buffer = Vec::new();
+        let limit = 100; // we could include 256 notes per tx, but requests channel is limited to 100 atm
```
igamigo, line 199: Could you post the numbers for executing and proving comparing the different amount of notes that we discussed?

bin/faucet/src/faucet/mod.rs
```diff
@@ -292,26 +308,61 @@ impl Faucet {
...
+        let (executed_transaction, notes) = self.create_batch_transaction(requests, rng)?;
+        let account_delta = executed_transaction.account_delta().clone();
+        let tx = Self::prove_transaction(executed_transaction)?;
+        let block_number = self.submit_transaction(tx, rpc_client).await?;
+
+        Ok((account_delta, block_number, notes))
+    }
+
+    /// Creates and executes a transaction that outputs the requested notes.
+    fn create_batch_transaction(
```
igamigo, line 329: nit: I'd name this `execute_batch_mint_transaction()` or `execute_batch_mint_tx()`

bin/faucet/src/faucet/mod.rs
```diff
@@ -292,26 +308,61 @@ impl Faucet {
...
+            .filter_map(|request| {
+                let note = P2IdNote::build(self.faucet_id(), request, rng);
+                match note {
+                    Ok(note) => Some(note),
+                    Err(err) => {
+                        tracing::error!(request.account_id=%request.account_id, ?err, "failed to build note");
+                        None
+                    }
+                }
+        }).collect();
```
igamigo, line 346: I wonder if `P2IdNote` should be renmaed to `P2IdNotes` and take a list of requests instead, and then this piece of can be part of the building process. It could also have a single `into_inner()` that returns the `Vec<Note>` directly and then you can avoid collecting into a different vector on line 347.

igamigo (approved):
LGTM! I left some minor comments. I think if we could address the one related to allocations we could merge and maybe even try deploying as well to see how this performs (maybe after the remote prover PR as well).

TomasArrachea:
I ran some tests to see how much does the amount of output notes affect on the execution and proving times of the transaction. Results:
With 1 output note:
```
Transaction execution: 42.742875ms
Transaction proving: 723.695708ms
```
With 100 output notes:
```
Transaction execution: 305.307666ms
Transaction proving: 2.879666709s
```
With 255 output notes:
```
Transaction execution: 800.670417ms
Transaction proving: 16.271348167s
```

bin/faucet/src/faucet/mod.rs
TomasArrachea: I deleted this function instead, I think its more clear this way

bin/faucet/src/faucet/mod.rs
```diff
@@ -187,24 +188,37 @@ impl Faucet {
...
-                continue;
-            }
+        let mut rng = RpoRandomCoin::new(coin_seed.map(Felt::new));
+
+        let mut buffer = Vec::new();
+        let limit = 100; // we could include 256 notes per tx, but requests channel is limited to 100 atm
+
+        while requests.recv_many(&mut buffer, limit).await > 0 {
+            // Skip requests where the user no longer cares about the result.
+            let (requests, response_senders): (Vec<MintRequest>, Vec<Sender<(BlockNumber, Note)>>) = buffer
```
sergerad, line 201: I think its at the point where we should make a struct for the messages being sent around the channels rather than using tuples. Complex line to grok.

sergerad: Although that might not help here given the unzip?

sergerad (approved):

bin/faucet/src/faucet/mod.rs
```diff
@@ -187,24 +188,37 @@ impl Faucet {
...
+                .filter(|(request, response_sender)| {
+                    if response_sender.is_closed() {
+                        tracing::info!(request.account_id=%request.account_id, "request cancelled");
+                        false
+                    } else {
+                        true
+                    }
+                })
+                .unzip();
+            println!("received {} requests", requests.len());
```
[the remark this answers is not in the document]
TomasArrachea: Ups removed it

bin/faucet/src/faucet/mod.rs
TomasArrachea: Yes, but we could improve this. Anyway it's going to change in https://github.com/0xMiden/miden-node/pull/835 so we can tackle it there

PR #830, opened 2025-05-08
feat: add remote prover support for faucet

SantiagoPittella:
Adds support for using the remote transaction prover in the faucet.

To use a remote transaction prover, the prover url needs to be set in the configuration file like this:
```toml
remote_tx_prover_url = "http://0.0.0.0:50053"
```


bin/faucet/src/main.rs
```diff
@@ -99,30 +99,28 @@ async fn run_faucet_command(cli: Cli) -> anyhow::Result<()> {
...
-                .unwrap()
-                .context("failed to join task")?
-                .context("a server task ended unexpectedly")?;
+            // Run the server in a separate task
+            let server_handle = tokio::spawn(async move {
+                server.serve(config.endpoint).await.context("server failed")
+            });
+
+            // Run the faucet directly on the current thread
+            let faucet_result = faucet.run(rpc_client, rx_requests).await;
```
igamigo, line 117: I could be wrong but I think because this blocks here, `server_handle` is never awaited nor polled so I think it could error without us knowing. Maybe `server_handle` and `faucet.run()` could be polled with `tokio::select!`.

SantiagoPittella: I've just pushed a commit fixing this 👌🏼

bin/faucet/src/faucet/mod.rs
```diff
@@ -305,29 +336,35 @@ impl Faucet {
...
         let tx_args = p2id_note.compile()?;
 
-        let tx = self.execute_transaction(tx_args)?;
+        let tx = self.execute_transaction(tx_args).await?;
         let account_delta = tx.account_delta().clone();
 
-        let tx = Self::prove_transaction(tx)?;
+        let tx = match &self.tx_prover.as_ref() {
+            TransactionProverWrapper::Remote(prover) => {
+                prover.prove(tx.into()).await.map_err(MintError::Proving)
```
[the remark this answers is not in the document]
SantiagoPittella: I implemented the following approach:

First try with the remote prover and if it returns an error, try again with a new local prover.

bin/faucet/src/faucet/mod.rs
```diff
@@ -38,6 +40,54 @@ mod store;
...
+        tx: impl Into<TransactionWitness> + Clone,
+    ) -> Result<ProvenTransaction, MintError> {
+        match self {
+            Self::Local(prover) => prover.prove(tx.into()).await,
+            Self::Remote(prover) => {
+                let proven_tx = prover.prove(tx.clone().into()).await;
+                match proven_tx {
+                    Ok(proven_tx) => Ok(proven_tx),
+                    Err(err) => {
+                        error!("failed to prove transaction with remote prover, falling back to local prover: {}", err);
```
igamigo, line 81: I think we want to `warn` here instead of `error` since it's gracefully degrading

bin/faucet/src/faucet/mod.rs
```diff
@@ -38,6 +40,54 @@ mod store;
...
+    fn local() -> Self {
+        Self::Local(LocalTransactionProver::new(ProvingOptions::default()))
+    }
+
+    /// Creates a new remote prover.
+    ///
+    /// # Arguments
+    ///
+    /// * `endpoint` - The endpoint to connect to the remote prover.
+    fn remote(endpoint: impl Into<String>) -> Self {
```
sergerad, line 66: nit: Tend to prefer to carry the type-safety as deep into the stack as possible. So would expect this fn to take a `Url` which has been parsed by config/flags prior to this call.

If that's impractical for any reason here feel free to resolve this comment.

bin/faucet/src/faucet/mod.rs
```diff
@@ -111,6 +162,7 @@ impl Faucet {
     pub async fn load(
         account_file: AccountFile,
         rpc_client: &mut RpcClient,
+        remote_tx_prover_url: Option<impl Into<String>>,
```
sergerad, line 165: Same comment as above re `Url` type

bin/faucet/src/main.rs
```diff
@@ -305,12 +314,23 @@ mod test {
...
         // Start the faucet connected to the stub
         let website_url = config.endpoint.clone();
-        tokio::spawn(async move {
-            run_faucet_command(Cli {
-                command: crate::Command::Start { config: config_path },
-            })
-            .await
-            .unwrap();
+
+        // Use std::thread to launch faucet - avoids Send requirements
```
sergerad, line 318: What was not Send?

SantiagoPittella: I had issue mostly related to the `TransactionExecutor`, and with `DataStore` in particular, because with the addition of the remote prover, the `async` feature was triggered and now the `DataStore` trait is async.

sergerad: Yea its the trait objects again. Can't recall if `Datastore` and `TransactionAuthenticator` are implicated by gRPC generated code or not. If they aren't, we could maybe fix them. Anyway, just noting for the record, thanks.
```
rustc: `(dyn miden_tx::DataStore + 'static)` cannot be sent between threads safely
the trait `std::marker::Send` is not implemented for
`(dyn miden_tx::DataStore + 'static)` required for 
`std::sync::Arc<(dyn miden_tx::DataStore + 'static)>` to implement `std::marker::Sync`
required for `&miden_tx::TransactionExecutor` to implement `std::marker::Send
```

bin/faucet/src/faucet/store.rs
```diff
@@ -46,7 +47,9 @@ impl FaucetDataStore {
     }
 }
 
+#[maybe_async_trait(?Send)]
 impl DataStore for FaucetDataStore {
+    #[maybe_async]
     fn get_transaction_inputs(
```
bobbinth, line 53: Why do we need to mark these as `maybe_async`? Should we always have them as `async`?

SantiagoPittella: My bad, it can be just `#[async_trait::async_trait(?Send)]` and the method just be async. I will address it in one of my open PRs.

Issue #821, opened 2025-05-06
Faucet rate-limiting


Mirko-von-Leipzig:
#819 introduced rate-limiting by restricting based on requests by IP and token requests by account ID.

It would be more appropriate (and interesting) to rate limit the token requests using proof-of-work.

We could then also reduce the IP rate-limiting to more generous amounts.

igamigo:
One thing I'm not entirely sure about is whether this invalidates some flow for the wallet team since they opened an issue for [allowing calls from other origins](https://github.com/0xMiden/miden-node/issues/736) some time ago.

A draft flow could be something like this:

- The client hits a `/pow` endpoint; the server would return a random `prefix` plus the current difficulty target. This response could come signed with a server secret
- The client then grinds `hash(prefix || nonce)` until a nonce is found where the result has a `difficulty` amount of trailing zeroes or similar
- The client posts the result (with the server signature for validation) alongside the target account ID and amount of tokens

It could also be based on timeframes or account IDs but without a server-provided key it would be pre-mineable.

bobbinth:
> * The client hits a `/pow` endpoint; the server would return a random `prefix` plus the current difficulty target. This response could come signed with a server secret
> * The client then grinds `hash(prefix || nonce)` until a nonce is found where the result has a `difficulty` amount of trailing zeroes or similar
> * The client posts the result (with the server-provided signature for validation) alongside the target account ID and amount of tokens

This works. I thinking using server-signed prefix makes sense here (prefix could be a 32-byte value and it could be the same value as would be used as the resulting note serial number).

> One thing I'm not entirely sure about is whether this invalidates some flow for the wallet team since they opened an issue for [allowing calls from other origins](https://github.com/0xMiden/miden-node/issues/736) some time ago.

One option is to have a separate flow which could bypass PoW if the request provides some API key. This way, claiming tokens through the wallet could be simplified. Though, this does open an attack vector if someone can get the wallet's API key.

tomyrd:
Just added a new TODO related to this, specifically about moving all the rate limiting parameters to the config file or to env vars. For more context see [this comment](https://github.com/0xMiden/miden-node/pull/839#discussion_r2087525508).

SantiagoPittella:
With the PoW in place plus the API keys, and the IP and account rate limits, can we consider this issue as completed? For the PoW related follow ups we already have issues in place.
cc @igamigo @bobbinth

bobbinth:
[...]

PR #831, opened 2025-05-09
feat: add PoW to faucet endpoint

SantiagoPittella:
Part of #821 

This PR adds a Proof-of-Work mechanism that ensures that the client spends at least ~2 seconds before being able to mint tokens.

It uses sha3-256 as hashing function and works by:

1. Before requesting token, the client sends a request to `/pow` and gets a random seed, a difficulty and a server signature.
 - The seed is a random generated hex string with 64 digits.
 - The difficulty is a number that represents the amount of leading zeros that the client needs to look for.
 - The current timestamp of the server.
 - The server signature is the hash of a secret server salt + the random seed + the timestamp.
2. The client uses the random seed to get a nonce such that the resultant digest has at least `difficulty` leading zeroes.
3. The server, in the `/get_tokens` endpoint, receives the random seed, the server signature, the server timestamp and the nonce use to obtain the resultant digest.
4. The server checks that the timestamp in a given range of time.
5. The server checks that the server signature is valid, by re-hashing the server salt + the just received seed + the timestamp, to ensure that the received seed was sent by the server.
6. The server checks that the received seed + the nonce does in fact has the necessary leading zeroes.

### DIFF
--- a/bin/faucet/Cargo.toml
+++ b/bin/faucet/Cargo.toml
@@ -31,8 +31,8 @@ miden-tx                     = { workspace = true, features = ["concurrent"] }
 rand                         = { workspace = true }
 rand_chacha                  = "0.9"
 serde                        = { version = "1.0", features = ["derive"] }
-sha3                         = { version = "0.10" }
 serde_json                   = { version = "1.0" }
+sha3                         = { version = "0.10" }
 thiserror                    = { workspace = true }
 tokio                        = { workspace = true, features = ["fs"] }
 tokio-stream                 = { workspace = true, features = ["net"] }
--- a/bin/faucet/Cargo.toml
+++ b/bin/faucet/Cargo.toml
@@ -31,6 +31,7 @@ miden-tx                     = { workspace = true, features = ["concurrent"] }
 rand                         = { workspace = true }
 rand_chacha                  = "0.9"
 serde                        = { version = "1.0", features = ["derive"] }
+sha3                         = { version = "0.10" }
 serde_json                   = { version = "1.0" }
 thiserror                    = { workspace = true }
 tokio                        = { workspace = true, features = ["fs"] }
--- a/bin/faucet/src/config.rs
+++ b/bin/faucet/src/config.rs
@@ -34,13 +34,15 @@ pub struct FaucetConfig {
     /// Optional: Endpoint of the remote transaction prover in the format
     /// `<protocol>://<host>[:<port>]`
     pub remote_tx_prover_url: Option<Url>,
+    /// The salt to be used by the server to generate the `PoW` seed
+    pub pow_salt: String,
 }
 
 impl Display for FaucetConfig {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         f.write_fmt(format_args!(
-            "{{ endpoint: \"{}\", node_url: \"{}\", timeout_ms: \"{}\", asset_amount_options: {:?}, faucet_account_path: \"{}\", remote_tx_prover_url: \"{:?}\" }}",
-            self.endpoint, self.node_url, self.timeout_ms, self.asset_amount_options, self.faucet_account_path.display(), self.remote_tx_prover_url
+            "{{ endpoint: \"{}\", node_url: \"{}\", timeout_ms: \"{}\", asset_amount_options: {:?}, faucet_account_path: \"{}\", remote_tx_prover_url: \"{:?}\", pow_salt: \"{}\" }}",
+            self.endpoint, self.node_url, self.timeout_ms, self.asset_amount_options, self.faucet_account_path.display(), self.remote_tx_prover_url, self.pow_salt
         ))
     }
 }
@@ -57,6 +59,7 @@ impl Default for FaucetConfig {
             asset_amount_options: AssetOptions::new(vec![100, 500, 1_000]).unwrap(),
             faucet_account_path: DEFAULT_FAUCET_ACCOUNT_PATH.into(),
             remote_tx_prover_url: None,
+            pow_salt: "miden-faucet".to_string(),
         }
     }
 }
--- a/bin/faucet/src/config.rs
+++ b/bin/faucet/src/config.rs
@@ -59,7 +59,7 @@ impl Default for FaucetConfig {
             asset_amount_options: AssetOptions::new(vec![100, 500, 1_000]).unwrap(),
             faucet_account_path: DEFAULT_FAUCET_ACCOUNT_PATH.into(),
             remote_tx_prover_url: None,
-            pow_salt: "miden-faucet".to_string(),
+            pow_salt: rand::random::<[u8; 32]>().into_iter().map(|b| b as char).collect(),
         }
     }
 }
--- a/bin/faucet/src/server/get_tokens.rs
+++ b/bin/faucet/src/server/get_tokens.rs
@@ -15,7 +15,7 @@ use tokio::sync::mpsc::{self, error::TrySendError};
 use tokio_stream::{Stream, wrappers::ReceiverStream};
 use tracing::error;
 
-use super::pow::{check_pow_solution, check_server_signature};
+use super::pow::{check_pow_solution, check_server_signature, check_server_timestamp};
 use crate::{
     faucet::MintRequest,
     types::{AssetOptions, NoteType},
@@ -46,6 +46,7 @@ pub struct RawMintRequest {
     pub pow_seed: String,
     pub pow_solution: u64,
     pub server_signature: String,
+    pub server_timestamp: u64,
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -58,6 +59,8 @@ pub enum InvalidRequest {
     InvalidPoW,
     #[error("invalid server signature")]
     InvalidServerSignature,
+    #[error("server timestamp expired")]
+    ExpiredServerTimestamp,
 }
 
 pub enum GetTokenError {
@@ -142,12 +145,19 @@ impl RawMintRequest {
             .validate(self.asset_amount)
             .ok_or(InvalidRequest::AssetAmount(self.asset_amount))?;
 
-        if !check_server_signature(self.server_signature, self.pow_seed.clone()) {
+        // Check the server timestamp
+        if !check_server_timestamp(self.server_timestamp) {
+            return Err(InvalidRequest::ExpiredServerTimestamp);
+        }
+
+        // Check the server signature
+        if !check_server_signature(&self.server_signature, &self.pow_seed, self.server_timestamp) {
             error!("[get_tokens] invalid server signature");
             return Err(InvalidRequest::InvalidServerSignature);
         }
 
-        if !check_pow_solution(self.pow_seed, self.pow_solution) {
+        // Check the PoW solution
+        if !check_pow_solution(&self.pow_seed, self.pow_solution) {
             return Err(InvalidRequest::InvalidPoW);
         }
 
--- a/bin/faucet/src/server/get_tokens.rs
+++ b/bin/faucet/src/server/get_tokens.rs
@@ -177,24 +177,24 @@ pub async fn get_tokens(
     // Response channel with buffer size 5 since there are currently 5 possible updates
     let (tx_result, rx_result) = mpsc::channel(5);
 
-    let mint_error = request
-        .validate(&state.asset_options, &server.pow_salt)
-        .map_err(GetTokenError::InvalidRequest)
-        .and_then(|request| {
-            let span = tracing::Span::current();
-            span.record("account", request.account_id.to_hex());
-            span.record("amount", request.asset_amount.inner());
-            span.record("note_type", request.note_type.to_string());
+    let mint_error =
+        request
+            .validate(&server.mint_state.asset_options, &server.pow_salt)
+            .map_err(GetTokenError::InvalidRequest)
+            .and_then(|request| {
+                let span = tracing::Span::current();
+                span.record("account", request.account_id.to_hex());
+                span.record("amount", request.asset_amount.inner());
+                span.record("note_type", request.note_type.to_string());
 
-            state
-                .request_sender
-                .try_send((request, tx_result.clone()))
-                .map_err(|err| match err {
-                    TrySendError::Full(_) => GetTokenError::FaucetOverloaded,
-                    TrySendError::Closed(_) => GetTokenError::FaucetClosed,
-                })
-        })
-        .err();
+                server.mint_state.request_sender.try_send((request, tx_result.clone())).map_err(
+                    |err| match err {
+                        TrySendError::Full(_) => GetTokenError::FaucetOverloaded,
+                        TrySendError::Closed(_) => GetTokenError::FaucetClosed,
+                    },
+                )
+            })
+            .err();
 
     if let Some(error) = mint_error {
         tx_result.send(Ok(error.into_event())).await.unwrap();
--- a/bin/faucet/src/server/get_tokens.rs
+++ b/bin/faucet/src/server/get_tokens.rs
@@ -13,7 +13,9 @@ use miden_objects::{AccountIdError, account::AccountId};
 use serde::Deserialize;
 use tokio::sync::mpsc::{self, error::TrySendError};
 use tokio_stream::{Stream, wrappers::ReceiverStream};
+use tracing::error;
 
+use super::pow::{check_pow_solution, check_server_signature};
 use crate::{
     faucet::MintRequest,
     types::{AssetOptions, NoteType},
@@ -41,6 +43,9 @@ pub struct RawMintRequest {
     pub account_id: String,
     pub is_private_note: bool,
     pub asset_amount: u64,
+    pub pow_seed: String,
+    pub pow_solution: u64,
+    pub server_signature: String,
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -49,6 +54,10 @@ pub enum InvalidRequest {
     AccountId(#[source] AccountIdError),
     #[error("asset amount {0} is not one of the provided options")]
     AssetAmount(u64),
+    #[error("invalid POW solution")]
+    InvalidPoW,
+    #[error("invalid server signature")]
+    InvalidServerSignature,
 }
 
 pub enum GetTokenError {
@@ -132,6 +141,15 @@ impl RawMintRequest {
         let asset_amount = options
             .validate(self.asset_amount)
             .ok_or(InvalidRequest::AssetAmount(self.asset_amount))?;
+
+        if !check_server_signature(self.server_signature, self.pow_seed.clone()) {
+            error!("[get_tokens] invalid server signature");
+            return Err(InvalidRequest::InvalidServerSignature);
+        }
+
+        if !check_pow_solution(self.pow_seed, self.pow_solution) {
+            return Err(InvalidRequest::InvalidPoW);
+        }
 
         Ok(MintRequest { account_id, note_type, asset_amount })
     }
--- a/bin/faucet/src/server/get_tokens.rs
+++ b/bin/faucet/src/server/get_tokens.rs
@@ -164,7 +164,6 @@ impl RawMintRequest {
             &self.pow_seed,
             self.server_timestamp,
         ) {
-            error!("[get_tokens] invalid server signature");
             return Err(InvalidRequest::InvalidServerSignature);
         }
 
--- a/bin/faucet/src/server/get_tokens.rs
+++ b/bin/faucet/src/server/get_tokens.rs
@@ -60,10 +60,10 @@ pub enum InvalidRequest {
     AssetAmount(u64),
     #[error("invalid POW solution")]
     InvalidPoW,
-    #[error("invalid server signature")]
-    InvalidServerSignature,
-    #[error("server timestamp expired")]
-    ExpiredServerTimestamp,
+    #[error("server signatures do not match")]
+    ServerSignaturesDoNotMatch,
+    #[error("server timestamp expired, received: {0}, current time: {1}")]
+    ExpiredServerTimestamp(u64, u64),
 }
 
 pub enum GetTokenError {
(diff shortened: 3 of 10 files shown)


SantiagoPittella:
Just realized that we need some sort of timestamp to make sure that the seed is not expired, because otherwise the user will be able to send the same seed, nonce and signature multiple times and will keep being valid.

sergerad:
> Before requesting token, the client sends a request to /pow and gets a random seed, a difficulty and a server signature.
...
The server signature is the hash of a secret server salt + the random seed + the timestamp.

Would we consider using an actual pubkey signature for this? At the least it seems confusing to be calling a checksum a signature.

Instead of the server sharing a checksum, would it make more sense for it to share its pubkey? The clients sign with the pubkey and the server verifies that signature with its privkey.

bin/faucet/src/server/pow.rs
```diff
@@ -0,0 +1,107 @@
...
+use miden_tx::utils::ToHex;
+use rand::{Rng, rng};
+use serde::Serialize;
+use sha3::{Digest, Sha3_256};
+
+const DIFFICULTY: u64 = 5;
+const SERVER_TIMESTAMP_TOLERANCE_SECONDS: u64 = 30;
+
+static SERVER_SALT: LazyLock<String> = LazyLock::new(|| {
+    std::env::var("MIDEN_FAUCET_SERVER_SALT").expect("MIDEN_FAUCET_SERVER_SALT must be set")
```
sergerad, line 16: What are our guidelines around configurations like this? Do most things come through a config file or flags (and env vars via flags)? Wondering about this from a consistency perspective.

bin/faucet/src/server/pow.rs
```diff
@@ -0,0 +1,107 @@
...
+#[derive(Serialize)]
+struct PoWResponse {
+    seed: String,
+    difficulty: u64,
+    server_signature: String,
+    timestamp: u64,
+}
+
+/// Generate a random hex string of specified length
+fn random_hex_string(len: usize) -> String {
```
sergerad, line 28: nit: Do we use anything like https://crates.io/crates/random-string elsewhere in the codebase? Makes sense not to add an import for a single function like this. I'm just not familiar enough with the codebase - perhaps we have something to this effect elsewhere.

bin/faucet/src/server/resources/index.js
```diff
@@ -84,6 +113,52 @@ document.addEventListener('DOMContentLoaded', function () {
         publicButton.disabled = false;
     }
 
+    // Function to find a valid nonce for proof of work
+    async function findValidNonce(seed, difficulty) {
```
sergerad, line 117: Similar to comment above about the hex string - would we consider using an existing crate like this https://docs.rs/pow/latest/pow/

It has a helper function to calculate expected duration based on difficulty which might help.

I haven't used that crate or looked into it. Just wondering.

bin/faucet/src/server/pow.rs
SantiagoPittella: I never saw the usage of this particular, or similar, crate in the miden repos. Though maybe I can change it to something like the following, let me know what do you think:

```rust
fn random_hex_string() -> String {
    let mut rng = rng();
    let random_bytes: [u8; 32] = rng.random();

    let hex_string = String::from_utf8(random_bytes.to_vec()).unwrap();
    format!("0x{hex_string}")
}
```

bin/faucet/src/server/resources/index.js
SantiagoPittella: I'm not quite sure that I'm fully understanding this comment, you meant in the server side, right?

sergerad: Ah yea sorry I had too brief a look. I'll find some quiet time later today and look again.

In general I'm interested in how we would unit test the logic given the frontend is performing the pow algorithm.

bin/faucet/src/server/get_tokens.rs
```diff
@@ -141,6 +153,22 @@ impl RawMintRequest {
...
             .ok_or(InvalidRequest::AssetAmount(self.asset_amount))?;
 
+        // Check the server timestamp
+        if !check_server_timestamp(self.server_timestamp) {
+            return Err(InvalidRequest::ExpiredServerTimestamp);
+        }
+
+        // Check the server signature
+        if !check_server_signature(&self.server_signature, &self.pow_seed, self.server_timestamp) {
+            error!("[get_tokens] invalid server signature");
```
sergerad, line 163: Correct me if I'm wrong but this could be caused by bad requests right? In which case we wouldn't want to log error.

bin/faucet/src/server/pow.rs
sergerad: Would you not need to process nibbles rather than bytes for that to be valid hexadecimal? As in the 8-bit encoding will produce letters beyond `f`.

bin/faucet/src/server/pow.rs
```diff
@@ -0,0 +1,107 @@
...
+///
+/// Returns `true` if the solution is valid, `false` otherwise.
+pub(crate) fn check_pow_solution(seed: &str, solution: u64) -> bool {
+    let mut hasher = Sha3_256::new();
+    hasher.update(seed);
+    hasher.update(solution.to_string().as_bytes());
+    let hash = &hasher.finalize().to_hex()[2..];
+
+    let leading_zeros = hash.chars().take(DIFFICULTY as usize).filter(|&c| c == '0').count();
+    leading_zeros >= DIFFICULTY as usize
```
sergerad, line 93: nit: the `take(DIFFICULTY..` means `leading_zeros` will never be greater than it.

This might be a bit more readable. Feel free to ignore and resolve if you disagree.
```rust
    let leading_zeros = hash.chars().take_while(|&c| c == '0').count();
    leading_zeros >= DIFFICULTY as usize
```

bin/faucet/src/server/resources/index.js
sergerad: Yea I guess if there isn't a js library to replace this pow logic with we just need to unit test this function?

bin/faucet/src/server/pow.rs
sergerad: Can just do `format_args!("{:#x}"..` and should work

bin/faucet/Cargo.toml
```diff
@@ -29,6 +29,7 @@ miden-tx           = { workspace = true, features = ["concurrent"] }
 rand               = { workspace = true }
 rand_chacha        = "0.9"
 serde              = { version = "1.0", features = ["derive"] }
+sha3               = { version = "0.10.8" }
```
bobbinth, line 32: Question: what's the motivation for using SHA3? For non-arithmetizaton-friendly hashing we usually use BLAKE3 (which is also available through `miden-crypto`).

bin/faucet/src/server/resources/index.html
```diff
@@ -6,6 +6,10 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Miden Faucet</title>
     <link rel="stylesheet" href="./index.css">
+    <!-- Load SHA3 library directly -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/js-sha3/0.8.0/sha3.min.js" integrity="sha384-NKbaJKqJPrFeb86qfU2iNy4KO3oyIe6F5xm1PBYMpAkiVTxHjjafUKlebp0mAeNC" crossorigin="anonymous"></script>
```
bobbinth, line 10: I guess maybe this is the reason for using SHA3? I'm assuming BLAKE3 option is not as easily available?

bin/faucet/src/server/get_tokens.rs
SantiagoPittella: Yes, basically we are checking that the user is not sending a solution of a seed that wasn;'t created by the server.

bin/faucet/src/server/pow.rs
SantiagoPittella: Sounds good 👌🏼 I'm replacing it.

bin/faucet/Cargo.toml
SantiagoPittella: I defaulted to SHA3 since it is pretty standard and found its slowness helpful in this this case (also found a standardize JS version was really easy).
I dig a bit more after your comment and saw that using blake3 in the JS is possible two. If we have some preference with blake, I can replace it 👍🏼

bin/faucet/src/server/resources/index.html
SantiagoPittella: Replied in the previous comment: https://github.com/0xMiden/miden-node/pull/831#discussion_r2085212818

bin/faucet/Cargo.toml
SantiagoPittella: I'm trying to replace it with blake, getting a functional version isn't resulting as easy as i thought i was going to be. Rust side is done, but in the JS side is a bit harder that I initially thought (though I'm not really a JS savvy, will ask someone with more knowledge)

sergerad:
@SantiagoPittella @igamigo just wanted to carry on the conversation about the signature here.

I would expect that a symmetric key would be inappropriate here unless it was a session key (per request). Otherwise the FE could share / reuse the key with requests/challenges not made by the backend?

So a pubkey signature would be more appropriate? As in the BE keeps the private key, FEs have the pub key for decrypt or encrypt (depending on the flow that makes sense here, haven't had a think).

igamigo:
> @SantiagoPittella @igamigo just wanted to carry on the conversation about the signature here.
> 
> I would expect that a symmetric key would be inappropriate here unless it was a session key (per request). Otherwise the FE could share / reuse the key with requests/challenges not made by the backend?
> 
> So a pubkey signature would be more appropriate? As in the BE keeps the private key, FEs have the pub key for decrypt or encrypt (depending on the flow that makes sense here, haven't had a think).

My thinking when I wrote the draft in the issue comment was that the server just signs a message (essentially the same message that the client receives) and sends the challenge and signed message (opaque to the client). The client then does not do anything with it except hand it back to the server with the original challenge and solution. On the server you can just resign with the server secret and verify that it matches what was sent by the client. If it does not, discard. Here I hadn't thought about the client actually submitting the same `(challenge, server_signature, solution)` tuple multiple times but @SantiagoPittella included a timestamp in the challenge so as to be able to discard expired requests as well. 

The rate-limiting (based on client IP) that is in place would do the rest of the work.

Alternatively I think the server can also symmetrically encrypt the message that the client receives and decrypt it when the client hands it back opaquely, so that the challenge does not need to be sent and re-signed, but I wasn't sure it would actually be better in significant ways (not entirely sure the payload would be smaller and the overhead of decrypting is probably higher than re-signing).

SantiagoPittella:
what you're basically saying is generating a random seed, a timestamp, and passing my public key to the client. The client then looks up the solution to that seed and signs it with my public key. Finally, we decrypt on the server and verify that the solution is valid.

What I don't like about this approach is that, as Serge said, I can't verify that the client isn't simply sending the same precomputed seed with its response. With the approach implemented, the client could send a seed that the server itself had previously sent (otherwise, the server_signature would be invalid), but it's only valid for a certain amount of time.

cc @sergerad @igamigo

sergerad:
```
Client ----------------------/pow----------------> Server [ sign(hash(salt, seed, timestamp)) = sig ]

Client <----------(seed, timestamp, sig)---------- Server

Client --------(nonce, seed, timestamp, sig)-----> Server ...
```

This is how I understand @SantiagoPittella's description in the PR. The private salt prevents replay. What is the purpose of the seed?

SantiagoPittella:
@sergerad your diagram is right:

image

We cannot prevent replay because we aren't storing any state in the server side (to keep it simple). The server signature is to ensure that the seed was sent by the server and that is not created by the client.

The server signature is the hash of: the random seed, the timestamp and a private salt of the server. The client receives the signature and timestamps of the server and is expected to do nothing with this parameters, just to send it back when hitting the `/getTokens` endpoint. In the server side, when we receive a request on `/getTokens`, a validation happens, where we hash again the seed, the timestamp of the server (the one provided by the client, so it should be the same that was sent), and the server;s secret salt, and then we compare the result of the hash to the signature forwarded by the client. If the hash mismatch the signature, that means that one of the parameters was changed, ergo, we cannot trust the client.

Regarding the prevention of replays, as I mentioned above, without keeping state this is a complicated task. This is why I added a timestamp. We cannot ensure that the seed was used only one, but with the timestamp we can ensure that the seed is valid only during a certain period of time. In particular, I set it to 30 seconds. The idea is to keep it low enough for the rate limiting (requests per account, and requests per IP) to, jointly with the timestamp, act as a form of uniqueness of the seed.

bin/faucet/src/server/resources/index.js
SantiagoPittella: Not sure how to add a unit test for this function

sergerad:
> Regarding the prevention of replays, as I mentioned above, without keeping state this is a complicated task. This is why I added a timestamp. We cannot ensure that the seed was used only one, but with the timestamp we can ensure that the seed is valid only during a certain period of time. In particular, I set it to 30 seconds. The idea is to keep it low enough for the rate limiting (requests per account, and requests per IP) to, jointly with the timestamp, act as a form of uniqueness of the seed.

Yea I worry that our DDOS prevention system hinges on this rate limiting. Ideally we could decouple those systems a bit so that, for example, if the rate limiting logic had a bug in it, you couldn't bypass POW by spamming replays within a 30 second window.

What if the faucet maintained a map (in-memory, capacity limited) of challenge checksums that have been successfully solved? Replay would be impossible. If rate limiting had a bug in it, now users have to perform multiple POW solutions to attack us.

So we go from this situation where rate limiting has a bug:
- User does 1 POW solution and spams replays

to this:
- User spams challenge requests and needs to do N POW solutions

WDYT?

igamigo:
It would be nice to avoid maintaining state, but I think that sounds reasonable. Maybe it could also be done on a different PR.

Another way to provide more safety could be to make the timeout more aggressive. Additionally, if the difficulty were to adjust based on the request queue, we could also validate that the difficulty is equal to (or within a threshold of) the current difficulty.

bin/faucet/src/server/pow.rs
```diff
@@ -0,0 +1,108 @@
...
+    timestamp: u64,
+) -> bool {
+    let mut hasher = Sha3_256::new();
+    hasher.update(server_salt);
+    hasher.update(seed);
+    hasher.update(timestamp.to_string().as_bytes());
+    let hash = &hasher.finalize().to_hex();
+
+    hash == server_signature
+}
```
igamigo, line 76: Instead of this returning a `bool`, maybe there could be a function that returned the hashed message directly. Then you could do the comparison in place, and also replace the code from lines 42-50 to use the function as well.

bin/faucet/src/server/get_tokens.rs
```diff
@@ -52,6 +61,12 @@ pub enum InvalidRequest {
     AccountId(#[source] AccountIdError),
     #[error("asset amount {0} is not one of the provided options")]
     AssetAmount(u64),
+    #[error("invalid POW solution")]
+    InvalidPoW,
+    #[error("invalid server signature")]
+    InvalidServerSignature,
+    #[error("server timestamp expired")]
+    ExpiredServerTimestamp,
```
igamigo, line 69: Maybe this could include the received timestamp and the current timestamp? Should make it easier to troubleshoot/debug when reading messages.

bin/faucet/src/server/get_tokens.rs
```diff
@@ -52,6 +61,12 @@ pub enum InvalidRequest {
     AccountId(#[source] AccountIdError),
     #[error("asset amount {0} is not one of the provided options")]
     AssetAmount(u64),
+    #[error("invalid POW solution")]
+    InvalidPoW,
+    #[error("invalid server signature")]
+    InvalidServerSignature,
```
igamigo, line 67: Would `ServerSignaturesDoNotMatch` be a better fit?

bin/faucet/src/server/get_tokens.rs
```diff
@@ -52,6 +61,12 @@ pub enum InvalidRequest {
     AccountId(#[source] AccountIdError),
     #[error("asset amount {0} is not one of the provided options")]
     AssetAmount(u64),
+    #[error("invalid POW solution")]
+    InvalidPoW,
```
igamigo, line 65: Maybe this could include the expected difficulty (and/or received solution, but probably less useful for general debugging since it's arbitrary bytes.)

bin/faucet/src/config.rs
```diff
@@ -39,8 +41,8 @@ pub struct FaucetConfig {
 impl Display for FaucetConfig {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         f.write_fmt(format_args!(
-            "{{ endpoint: \"{}\", node_url: \"{}\", timeout_ms: \"{}\", asset_amount_options: {:?}, faucet_account_path: \"{}\", remote_tx_prover_url: \"{:?}\" }}",
-            self.endpoint, self.node_url, self.timeout_ms, self.asset_amount_options, self.faucet_account_path.display(), self.remote_tx_prover_url
+            "{{ endpoint: \"{}\", node_url: \"{}\", timeout_ms: \"{}\", asset_amount_options: {:?}, faucet_account_path: \"{}\", pow_salt: \"{}\", remote_tx_prover_url: \"{:?}\" }}",
+            self.endpoint, self.node_url, self.timeout_ms, self.asset_amount_options, self.faucet_account_path.display(), self.pow_salt, self.remote_tx_prover_url
```
igamigo, line 45: Does this have to be manually input? It would be nice if `miden-faucet init` already generated a random secret.

bin/faucet/src/server/pow.rs
```diff
@@ -0,0 +1,108 @@
...
+/// Check a `PoW` solution.
+///
+/// * `seed` - The seed to be used by the client as the `PoW` seed.
+/// * `solution` - The solution to be checked.
+///
+/// The solution is valid if the hash of the seed and the solution has at least `DIFFICULTY`
+/// leading zeros.
+///
+/// Returns `true` if the solution is valid, `false` otherwise.
+pub(crate) fn check_pow_solution(seed: &str, solution: u64) -> bool {
```
igamigo, line 87: This function may be a good target for a couple of unit tests.

SantiagoPittella:
@sergerad @igamigo regarding adding state:

In my opinion the current approach is shippable, and helps reducing spams. I think the best way to go is to open a new issue to talk about how this stateful approach should work, and the required changes (we can probably reduced the parameters sent to the client). Though, if any of you think that is a *must* to implement it with state I can start now with those changes.

bin/faucet/src/server/get_tokens.rs
SantiagoPittella: Hm, I think that the only way for it to be useful is to include all of those values, that way we can re-construct the check. But i'm not convinced for it to be that useful really. what do you think?

bin/faucet/src/config.rs
SantiagoPittella: Sounds good 👌🏼

sergerad:
> @sergerad @igamigo regarding adding state:
> 
> In my opinion the current approach is shippable, and helps reducing spams. I think the best way to go is to open a new issue to talk about how this stateful approach should work, and the required changes (we can probably reduced the parameters sent to the client). Though, if any of you think that is a _must_ to implement it with state I can start now with those changes.

Happy to discuss the replay prevention etc separately.

Perhaps more importantly before we ship we need to make sure the UX of the POW procedure in the FE is good enough to avoid confusion. Have you guys discussed this at all? Are we intending to implement the updated UX in this PR or separate ones?

SantiagoPittella:
> Perhaps more importantly before we ship we need to make sure the UX of the POW procedure in the FE is good enough to avoid confusion. Have you guys discussed this at all? Are we intending to implement the updated UX in this PR or separate ones?

I was thinking of adding it as one of the MintUpdate events that @TomasArrachea is creating in his UX PR ( #835 ). So, if he's merges first, I can add it in this very same PR. If this one is merged, I can open another one with this change using his as base.

cc @sergerad

bin/faucet/src/server/resources/index.js
sergerad: In terms of the js project or the function specifically? I'm pretty dusty with js so if no one else has any ideas we can come back to this. Think its worth having a unit test for this b/c of the impact it could have if faulty.

bin/faucet/src/server/get_tokens.rs
sergerad: Should we change `error!` to warn or debug then?

SantiagoPittella: It was already removed, since this is not an application error

sergerad: Ah gotcha - was confused!

sergerad (approved):

bin/faucet/src/server/resources/index.js
SantiagoPittella: I saw that we need to integrate a test framework to test this. I think the best will be to create end-to-end tests for the faucet, created: https://github.com/0xMiden/miden-node/issues/848

igamigo:
Before merging, let's also make sure there's an issue (or comment in the tracking issue) to track the work for adjusting the difficulty based on the current request queue size.

bin/faucet/src/server/get_tokens.rs
```diff
@@ -41,6 +46,10 @@ pub struct RawMintRequest {
     pub account_id: String,
     pub is_private_note: bool,
     pub asset_amount: u64,
+    pub pow_seed: String,
+    pub pow_solution: u64,
+    pub server_signature: String,
+    pub server_timestamp: u64,
```
bobbinth, line 52: One potential way to simplify this is to encode some of these things into the `pow_seed` (I'll call this "challenge"). For example, the structure of the challenge could be something like:
```
[8 difficulty bytes | 16 IP address bytes | 8 timestamp bytes | 32 signature bytes]
```
The signature here could be computed as:
```
hash(difficulty | IP address bytes | timestamp | server key)
```
The interaction between the user and the server would then be:

- The user calls the `/pow` endpoint and gets the challenge.
  - The server computes this challenge based on the users IP address and current timestamp and difficultly level.
- The user extracts the `difficulty` from the challenge, and then computes `hash(challenge | nonce)` for various nonces until the required difficulty level is met.
- The user makes a request to `get_tokens` passing the `challenge` and the `nonce`.
- The server decodes the challenge and checks that:
  - The signature is valid.
  - The difficulty level is met.
  - The challenge hasn't expired (e.g., a challenge could expire 30 seconds after it has been issued).
  - The IP address of the challenge matches the IP address of the request.

In the future, after all the checks have passed, the server could also save the challenge into a data structure which keeps track of "used" challenges (to make sure someone won't be able to use the same challenge twice). We could even index this per IP address which could organically provide rate-limiting (i.e., we won't need a separate rate-limiter).

SantiagoPittella: I like this approach, it extends the checks performed to the hash. Wdyt about implementing this as part of #849 ?

bobbinth: Yep - that sounds good.

PR #833, opened 2025-05-12
feat(ntx): Emit network notes

igamigo:


crates/block-producer/src/block_builder/mod.rs
```diff
@@ -85,14 +100,27 @@ impl BlockBuilder {
...
     /// Run the block building stages and add open-telemetry trace information where applicable.
     ///
     /// A failure in any stage will result in that block being rolled back.
     ///
+    /// ## Returns
+    ///
+    /// The transaction and note IDs committed in this block, as well as the transaction IDs that
+    /// were dropped (e.g. due to expiring).
+    ///
+    /// Returns [`None`] if the block building failed.
```
tomyrd, line 122: Could this be outdated? The function below returns `Result<()>`

Edit: Maybe this was meant for `commit_block`?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -361,15 +369,17 @@ impl Mempool {
...
     }
 
     #[instrument(target = COMPONENT, name = "mempool.revert_expired_transactions", skip_all)]
-    fn revert_expired_transactions(&mut self) {
+    fn revert_expired_transactions(&mut self) -> BTreeSet<TransactionId> {
         let expired = self.expirations.get(self.chain_tip);
 
-        self.revert_transactions(expired.into_iter().collect())
+        self.revert_transactions(expired.iter().copied().collect())
             .expect("expired transactions must be part of the mempool");
```
tomyrd, line 380: Should we chain the returned tx ids here with `expired`? Maybe there were descendant transactions that were also discarded.

crates/block-producer/src/server.rs
```diff
@@ -51,13 +64,13 @@ pub struct BlockProducer {
 }
 
 impl BlockProducer {
+    /// Serves the block-producer RPC API, the batch-builder and the block-builder.
     /// Serves the block-producer RPC API, the batch-builder and the block-builder.
```
tomyrd, line 68: Seems like this comment was duplicated

crates/block-producer/src/server.rs
```diff
@@ -261,11 +291,34 @@ impl BlockProducerRpcServer {
         let inputs = self.store.get_tx_inputs(&tx).await.map_err(VerifyTxError::from)?;
 
         // SAFETY: we assume that the rpc component has verified the transaction proof already.
```
tomyrd, line 293: Is this safety comment here correct? It doesn't seem to reference any unwrap/expect.

crates/block-producer/src/block_builder/mod.rs
tomyrd: Seems to be fixed in https://github.com/0xMiden/miden-node/pull/840

crates/block-producer/src/server.rs
tomyrd: Seems to be fixed in https://github.com/0xMiden/miden-node/pull/840

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -223,20 +253,88 @@ impl BlockBuilder {
...
+                .await
+                .context("submitting transaction status updates to network transaction builder")?;
+        }
+
+        if !delta.committed_network_notes.is_empty() {
+            self.ntx_client
+                .clone()
+                .submit_network_notes(
+                    // TODO: using default here until there is a good reason not to (is tx id even
+                    // needed?)
```
TomasArrachea, line 322: It's not needed at the moment, but the id was added in this endpoint for tracking progress through the mempool (proposed in [this comment](https://github.com/0xMiden/miden-node/issues/695#issuecomment-2660821018))

crates/block-producer/src/server.rs
igamigo: I think the main point there is that we don't verify the transaction before instantiating an `AuthenticatedTransaction` because we assume it has been verified before.

proto/proto/ntx_builder.proto
```diff
@@ -12,4 +12,7 @@ service Api {
 
     // Update network transaction builder with transaction status changes.
     rpc UpdateTransactionStatus(requests.UpdateTransactionStatusRequest) returns (google.protobuf.Empty) {}
+
+    // Update the status of network notes that were consumed externally.
+    rpc UpdateNetworkNotes(requests.UpdateNetworkNotesRequest) returns (google.protobuf.Empty) {}
```
bobbinth, line 17: Either here or somewhere else, would be great to add a bit more explanation for how these 3 endpoints are used. It took me some time to mentally visualize the flow. My understanding is as follows (the way it works currently):

- When a block containing created network notes is committed, we use `SubmitNetworkNotes` endpoint to notify the NTB about these network notes. Right now, we don't associate these notes with transactions because the transactions that create these notes are already committed and no further status updates are needed for them.
- When a transaction enters the mempool, we notify NTB of the nullifiers this transaction consumes via the `UpdateNetworkNotes` endpoint.
- Subsequently, we use `UpdateTransactionStatus` to update the status of these transactions (to know whether a transaction consuming some nullifiers was committed or rejected).

Is this correct?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -388,7 +398,7 @@ impl Mempool {
     fn revert_transactions(
         &mut self,
         txs: Vec<TransactionId>,
-    ) -> Result<(), GraphError<TransactionId>> {
+    ) -> Result<BTreeSet<TransactionId>, GraphError<TransactionId>> {
```
bobbinth, line 401: Could we update doc comments to describe what this function returns?

crates/block-producer/src/mempool/mod.rs
bobbinth: Or, rather than chaining, should we return the set returned from `self.revert_transactions()` from here?

In either case, we should add doc comments explaining what this function does and what it returns.

crates/block-producer/src/server.rs
```diff
@@ -1,34 +1,45 @@
...
     mempool::{BatchBudget, BlockBudget, Mempool, SharedMempool},
     store::StoreClient,
 };
 
+/// The block producer server.
+///
+/// Specifies how to connect to the store, batch prover, and block prover components.
+/// The connection to the store is established at startup and retried with exponential backoff
+/// until the store becomes available. Once the connection is established, the block producer
+/// will start serving requests.
```
bobbinth, line 42: I think this comment is duplicated here.

crates/block-producer/src/server.rs
```diff
@@ -93,10 +105,17 @@ impl BlockProducer {
             .await
             .context("failed to bind to block producer address")?;
 
+        let ntx_client =
+            ntx_builder::Client::lazy_with_interceptor(self.ntx_builder_address, OtelInterceptor);
```
bobbinth, line 109: nit: `ntx_client` -> `ntx_builder`.

crates/block-producer/src/server.rs
```diff
@@ -182,6 +205,8 @@ struct BlockProducerRpcServer {
     mempool: Mutex<SharedMempool>,
 
     store: StoreClient,
+
+    ntx_client: NtxClient,
```
bobbinth, line 209: nit: `ntx_client` -> `ntx_builder`.

crates/block-producer/src/server.rs
```diff
@@ -261,11 +290,34 @@ impl BlockProducerRpcServer {
...
+                    ntx_client.update_network_notes(tx_id, tx_nullifiers.into_iter()).await
+                {
+                    warn!(
+                        target: COMPONENT,
+                        message = %err,
+                        "error submitting network notes updates to ntx builder"
+                    );
+                }
+            },
+        );
```
bobbinth, line 318: This tries to update the mempool and send the transaction to the NTB at the same time - right? If so, I wonder if we should actually do this sequentially. That is, add transaction to the mempool and if it was added without errors, only then send it to NTB.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -223,20 +246,88 @@ impl BlockBuilder {
...
+            .output_notes()
+            .filter_map(|(_idx, note)| match note {
+                OutputNote::Full(inner)
+                    if inner.metadata().tag().execution_mode() == NoteExecutionMode::Network =>
+                {
+                    Some(inner.clone())
+                },
+                _ => None,
+            })
+            .collect();
```
bobbinth, line 272: nit: I would maybe create a small helper function which takes a reference to a block and returns a vector of network notes from that block.

bobbinth (commented):

crates/block-producer/src/server.rs
igamigo: I changed this to work sequentially on the execution PR, but forgot to change it back here. Will do so now.

crates/proto/src/ntx_builder/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+    transaction::TransactionStatus,
+};
+
+type StdError = Box<dyn std::error::Error + Send + Sync + 'static>;
+type GeneratedClient<T> = crate::generated::ntx_builder::api_client::ApiClient<T>;
+
+#[derive(Clone)]
+pub struct Client<T> {
+    inner: GeneratedClient<T>,
+}
```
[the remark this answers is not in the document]
igamigo: > Why do we define RPC ApiClient in the rpc crate, but NTX builder client in the proto crate?

I've questioned this myself but I decided to take Mirko's approach (from #805) as the source of truth since maybe it was decided that this would be a good general mid-term direction for the client (and to not waste too much time on this). I agree it should be more consistent, but I didn't want to make too many decisions that could be overridden (or that could override) other ongoing choices.

> Should we try to align the interfaces a bit more? For example, ApiClient uses connect() method for constructor, but Client uses lazy_with_interceptor().

I think #860 re-introduces the `connect_lazy()` approach. I've renamed `lazy_with_interceptor` to `connect_lazy()` to make it more consistent and a bit more to the point

crates/block-producer/src/mempool/mod.rs
igamigo: Updated the docs!

proto/proto/note.proto
```diff
@@ -49,6 +49,12 @@ message Note {
     optional bytes details = 6;
 }
 
+// Represents a network note.
+message NetworkNote {
+    // Serialized network note.
+    bytes note = 1; 
+}
```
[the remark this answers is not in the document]
igamigo: Yeah, agreed. I think I tried something like this at first but couldn't find a way to convert cleanly without cloning (which should be the case ideally imo). I've implemented this approach.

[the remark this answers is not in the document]
igamigo: I think ideally the `Note` mesage would be renamed to something like `CommittedNote` and then this could become a generic `Note` message.

crates/block-producer/src/mempool/mod.rs
igamigo: Good point, changed this and added some docs

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -223,20 +246,88 @@ impl BlockBuilder {
...
+                        Digest::default(),
+                        Digest::default(),
+                        Digest::default(),
+                    ),
+                    delta.committed_network_notes.into_iter(),
+                )
+                .await
+                .context(
+                    "failed to submit newly committed notes to the network transaction builder",
+                )?;
```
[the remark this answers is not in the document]
igamigo: Yes, this is a good point. It might be worth making this a non-fatal error.

proto/proto/ntx_builder.proto
```diff
@@ -8,8 +8,21 @@ import "google/protobuf/empty.proto";
...
     // Submit a list of network notes to the network transaction builder.
+    //
+    // When a block gets committed in the blockchain, the list of new network notes needs to be 
+    // submitted to the NTB through this endpoint in order for it to track them and eventually 
+    // consume them.
     rpc SubmitNetworkNotes(requests.SubmitNetworkNotesRequest) returns (google.protobuf.Empty) {}
 
     // Update network transaction builder with transaction status changes.
+    //
+
```
bobbinth, line 19: I think a part of the comment may be missing here.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -309,11 +309,15 @@ impl Mempool {
     /// The pool will mark the associated batches and transactions as committed, and prune stale
     /// committed data, and purge transactions that are now considered expired.
     ///
+    /// # Returns
+    ///
+    /// Returns the list of transactions that expired and were purged.
```
bobbinth, line 314: nit: I would maybe make this a bit more general. For example: "Returns a set of transactions that were purged from the mempool because they can no longer be included in in the chain (e.g., expired transactions and their descendants)."

crates/block-producer/src/mempool/mod.rs
```diff
@@ -330,18 +334,22 @@ impl Mempool {
...
+        self.revert_expired_transactions()
     }
 
     /// Notify the pool that construction of the in flight block failed.
     ///
     /// The pool will purge the block and all of its contents from the pool.
     ///
+    /// # Returns
+    ///
+    /// Returns the set of transactions that was purged.
```
bobbinth, line 346: Same nit as above.

crates/block-producer/src/server.rs
```diff
@@ -261,11 +284,27 @@ impl BlockProducerRpcServer {
...
+                SubmitProvenTransactionResponse { block_height: block_height.as_u32() }
+            });
+
+        let mut ntx_builder = self.ntx_builder.clone();
+        if let Err(err) = ntx_builder.update_network_notes(tx_id, tx_nullifiers.into_iter()).await {
+            debug!(
+                target: COMPONENT,
+                message = %err,
+                "error submitting network notes updates to ntx builder"
+            );
```
bobbinth, line 304: Should we log this as error (rather than debug)?

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -395,3 +483,20 @@ impl BlockProver {
         }
     }
 }
+
+// HELPER
+// ================================================================================================
+
+pub fn get_network_notes(proven_block: &ProvenBlock) -> Vec<Note> {
```
bobbinth, line 490: Does this function need to be `pub`?

bobbinth (approved):
Looks good! Thank you! I left some more comments inline - pretty much all nits.

crates/proto/src/ntx_builder/mod.rs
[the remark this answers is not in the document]
bobbinth: Let's create an issue to figure out how/if we want to make these structs more consistent.

crates/proto/src/ntx_builder/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+    T::Error: Into<StdError>,
+    T::ResponseBody: Body<Data = Bytes> + std::marker::Send + 'static,
+    <T::ResponseBody as Body>::Error: Into<StdError> + std::marker::Send,
+{
+    // TODO: this should probably be an entire block instead of just a single tx.
+    pub async fn submit_network_notes(
+        &mut self,
+        tx_id: TransactionId,
+        notes: impl Iterator<Item = Note>,
+    ) -> Result<(), tonic::Status> {
```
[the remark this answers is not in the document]
bobbinth: Should we create an issue for this?

proto/proto/note.proto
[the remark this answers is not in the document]
bobbinth: > I think ideally the `Note` mesage would be renamed to something like `CommittedNote` and then this could become a generic `Note` message.

crates/block-producer/src/server.rs
```diff
@@ -304,17 +343,25 @@ mod test {
...
             block_producer_listener
                 .local_addr()
                 .expect("Failed to get block-producer address")
         };
 
+        let ntx_builder_address = {
+            let ntx_builder_address = TcpListener::bind("127.0.0.1:0")
+                .await
+                .expect("failed to bind the ntx builder address");
+            ntx_builder_address.local_addr().expect("Failed to get ntx builder address")
```
SantiagoPittella, line 356:
```suggestion
            ntx_builder_address.local_addr().expect("failed to get ntx builder address")
```

SantiagoPittella (commented):
looking good 👍🏼

proto/proto/note.proto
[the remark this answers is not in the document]
igamigo: Created https://github.com/0xMiden/miden-node/issues/868

bobbinth (approved):
All looks good! Thank you!

SantiagoPittella (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - [BUGFIX] Prevents duplicated note IDs (#842).
 - Added a required Proof-of-Work in the faucet to request tokens (#831).
 - Added an optional API key request parameter to skip PoW in faucet (#839).
+- Block producer now emits network note information (#805).
```
SantiagoPittella, line 21: nit:
```suggestion
- Block producer now emits network note information (#833).
```

igamigo: Oops, good catch, will address on #840

PR #834, opened 2025-05-12
Add instrumentation to faucet

### LINKED ISSUE #811: Instrument the faucet
The faucet currently emits events/logs but no open-telemetry spans.

- [ ] Add an `enable-otel` cli arg in the same fashion as the node.
- [ ] Instrument the various endpoints.

I would advise reading the node documentation chapter on open-telemetry.

Especially the tokens endpoint should be split into logical spans for the duration of the request so that one can make sense of which parts are slow.

This includes spans for locking any shared state etc.

For inspiration, take a look at the block and batch building in the block-producer. As a general rule, prefer spans over events.

sergerad:
Relates to #811.

### Changes
- Adds `--enable-otel` flag and `MIDEN_FAUCET_ENABLE_OTEL` to faucet.
- Adds instrumentation to functions in the faucet that pertain to RPC endpoints.

#### Endpoints with instrumentation:
Screenshot 2025-05-14 at 3 52 30 PM

#### Endpoints without instrumentation:
Screenshot 2025-05-14 at 3 52 19 PM

### DIFF
--- a/bin/faucet/src/client.rs
+++ b/bin/faucet/src/client.rs
@@ -31,7 +31,7 @@ use miden_tx::{
 };
 use rand::{random, rngs::StdRng};
 use tonic::transport::Channel;
-use tracing::info;
+use tracing::{info, instrument};
 
 use crate::{COMPONENT, config::FaucetConfig, errors::ClientError, store::FaucetDataStore};
 
@@ -121,6 +121,7 @@ impl FaucetClient {
     /// Executes a mint transaction for the target account.
     ///
     /// Returns the executed transaction and the expected output note.
+    #[instrument(target = COMPONENT, name = "faucet.client.execute_mint_transaction", skip_all, err)]
     pub fn execute_mint_transaction(
         &mut self,
         target_account_id: AccountId,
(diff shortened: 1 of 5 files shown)


bin/faucet/src/client.rs
```diff
@@ -121,6 +121,7 @@ impl FaucetClient {
     /// Executes a mint transaction for the target account.
     ///
     /// Returns the executed transaction and the expected output note.
+    #[instrument(target = COMPONENT, name = "faucet.client.execute_mint_transaction", skip_all, err)]
```
TomasArrachea, line 124: I imagine that it would be nice to be able to trace the execution and submission logs to a request id or account id that triggered it, for debugging. Maybe its worth trying [follows_from](https://docs.rs/tracing/latest/tracing/struct.Span.html#method.follows_from) or adding a field?

sergerad: That should be unnecessary as the parent span has the account_id field already. In the image in the description you can see the parent span's fields in the logs of the `execute_mint_transaction()` span. The request has no further fields to add:
```rust
pub struct FaucetRequest {
    account_id: String,
    is_private_note: bool,
    asset_amount: u64,
}
```

LMK if I misunderstood or missed anything!

TomasArrachea: Perfect, thanks!

TomasArrachea (approved):

Issue #824, opened 2025-05-06
Faucet UX


Mirko-von-Leipzig:
The faucet website is minimal at the moment (which is good), however the user doesn't get much info beyond a spinner, waiting and then a success or error dialog.

What would be nice, is if the user could see and understand the progress of the request.

e.g. after selecting private/public the UI updates with their position in the queue, followed by the actual (lengthy steps) of the request handling once they're in front:

- building transaction
- executing transaction
- proving transaction
- submitting transaction

Further, the error dialog could use a lot of improvement. #819 added showing the error message at least, as well as the http error code.

I'm a fan of using server sent events for this, instead of having the user poll for status (which complicates rate-limiting etc). Server sent events are effectively a one-way socket aka client opens the connection on first request and then the server is free to send messages back as it wishes.

bobbinth:
If we implement PoW-based rate-limiting, I'd also add PoW computation to the list of steps which gets communicated to the user.

TomasArrachea:
Done in https://github.com/0xMiden/miden-node/pull/835

PR #835, opened 2025-05-12
feat: improve faucet ux

TomasArrachea:
Closes https://github.com/0xMiden/miden-node/issues/824.

### Summary 
This PR updates the faucet to track the status of the mint transaction and show it's progress on the frontend. This is done using server-sent events (SSE); the `get_tokens` endpoint now returns a stream of updates. This caused that the error handling also had to be refactored.

Also, now accounts are showed as bech32 addresses. The input address can be either hex or bech32.

### Demo:

https://github.com/user-attachments/assets/5f3bebeb-3767-4ed7-8ea5-64b7c4393a7d


bin/faucet/src/server/resources/index.js
```diff
@@ -31,57 +32,103 @@ document.addEventListener('DOMContentLoaded', function () {
...
 
-            if (!response.ok) {
-                throw response;
-            }
+        evtSource.onerror = function (_) {
+            // Either rate limit exceeded or invalid account id. The error event does not contain the reason.
+            evtSource.close();
+            showError('Failed to receive tokens.');
+            setLoadingState(false);
+        };
```
TomasArrachea, line 94: Previously, the faucet showed the error status and message if the request returned with an error. Now, using SSE, errors are returned as opaque events with no context or error status, so the message shown is just a generic "Failed to receive tokens.". Couldn't find a way of handling it better.

This only happens for the rate-limiter errors and connection errors. Instead if an error occurs while minting, that error is sent as an event and showed correctly.

sergerad: > This only happens for the rate-limiter errors and connection errors.

Are these logged in the FE at all? Will be difficult handling bug reports if we can't get any info from the FE when these things happen.

TomasArrachea: Yes, from the browser console you can actually see the error response, just that it won't be showed to the user in a friendly way

sergerad: Lets expand `Failed to receive tokens.` to say something like `Please try again soon`. Or something along those lines, wdyt?

TomasArrachea: Done!

TomasArrachea:
Sorry for the late changes, updated the frontend to show the explorer link on the created transaction

bin/faucet/src/faucet/updates.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Converts the mint update into an sse event.
+    /// Event types:
+    /// - `MintUpdate::Built`: event type "update"
+    /// - `MintUpdate::Executed`: event type "update"
+    /// - `MintUpdate::Proven`: event type "update"
+    /// - `MintUpdate::Submitted`: event type "update"
+    /// - `MintUpdate::Minted`: event type "note". Contains the note encoded in base64.
+    pub fn into_event(self) -> Event {
+        match self {
+            MintUpdate::Minted(note, block_height, tx_id) => {
```
igamigo, line 78: Does this mean that the notes will be sent to the frontend even if it's a public note? If so, it would be nice to make this optional to avoid unnecessary data transfers.

igamigo (commented):
Overall looks good. Not a big deal but I wonder if we could make it so that the divs don't move as the transaction progresses and finalizes. In the video you uploaded you can see that the UI moves a bit.

sergerad (approved):
Looks good. +1 re igamigo's comment about the divs.

TomasArrachea:
@igamigo Fixed the moving divs, updated the description video so you can check it there!

bin/faucet/src/faucet/mod.rs
```diff
@@ -445,24 +462,37 @@ fn parse_desync_error(err: &str) -> Result<Digest, anyhow::Error> {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if creating any p2id note fails.
+    fn build(
+        source: FaucetId,
+        requests: &[MintRequest],
+        rng: &mut impl FeltRng,
+    ) -> Result<Self, MintError> {
+        // TODO: if building a note fails, we discard the whole batch. Is this error possible?
```
igamigo, line 479: I think it's fine to assume `create_p2id_note` should not fail here, but I don't mind if you prefer keeping the error either

igamigo (approved):
LGTM, left just a minor comment, feel free to disregard.

PR #839, opened 2025-05-13
feat: add api key support to faucet

tomyrd:
This PR deals with the API key requirement from #821 

This lets users skip the PoW if they know a valid API key. API keys are stored in the `miden-faucet.toml` file as a comma separated string. The keys are randomly generated in the `init` command with the `--generated-api-keys-count` flag. A new rate limiter was added for requests with the same api key.


bin/faucet/src/server/mod.rs
```diff
@@ -215,3 +240,22 @@ impl KeyExtractor for AccountKeyExtractor {
...
+
+    fn extract<T>(&self, req: &Request<T>) -> Result<Self::Key, GovernorError> {
+        let params = Query::<RawMintRequest>::try_from_uri(req.uri())
+            .map_err(|_| GovernorError::UnableToExtractKey)?;
+
+        if let Some(api_key) = params.api_key.as_ref() {
+            Ok(api_key.to_string())
+        } else {
+            Ok(params.account_id.clone()
+                + &params.server_timestamp.ok_or(GovernorError::UnableToExtractKey)?.to_string())
```
tomyrd, line 258: This is a kind of hacky way to skip the API key rate limiter for requests without API key. I wasn't able to find a good way to do this in the `tower` or `tower_governor` crate.

The idea is that requests without the `api_key` param get a unique `Key` for that rate limiter. The "unique" `Key` is built from the account ID + timestamp so normal PoW requests will never get limited by this limiter.

bin/faucet/src/server/mod.rs
```diff
@@ -65,9 +67,12 @@ impl Server {
         // SAFETY: Leaking is okay because we want it to live as long as the application.
         let metadata = Box::leak(Box::new(metadata));
 
-        Server { mint_state, metadata, pow_salt }
+        Server { mint_state, metadata, pow_salt, api_keys }
     }
 
+    // TODO: Cannot move the rate limiter creation to its own function because it requires
+    // defining types that the `governor` crate does not export.
+    #[allow(clippy::too_many_lines)]
```
tomyrd, line 75: Basically what the TODO explains. I wanted to create a function that built the rate limiters to reduce the line count. But the `GovernorConfigBuilder` type requires generics for traits/objects that are not exposed.

bin/faucet/src/server/mod.rs
```diff
@@ -48,6 +48,7 @@ pub struct Server {
     mint_state: GetTokensState,
     metadata: &'static Metadata,
     pow_salt: String,
+    api_keys: BTreeSet<String>,
```
tomyrd, line 51: API keys are stored in the config file for persistence and in memory for fast checking. Neither should be accessible by outside users.

bin/faucet/src/server/mod.rs
```diff
@@ -95,15 +100,25 @@ impl Server {
             .unwrap();
         let account_rate_limiter = Arc::new(account_rate_limiter);
 
+        let api_key_rate_limiter = GovernorConfigBuilder::default()
+            .const_burst_size(3)
+            .const_per_second(10)
```
tomyrd, line 105: These values are totally arbitrary and can be changed. Right now it allows for 3 API key requests every 10 seconds

bin/faucet/src/config.rs
```diff
@@ -33,13 +33,15 @@ pub struct FaucetConfig {
     pub faucet_account_path: PathBuf,
     /// The salt to be used by the server to generate the `PoW` seed
     pub pow_salt: String,
+    /// List of comma separated API keys
+    pub api_keys: String,
```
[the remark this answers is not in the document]
sergerad: +1 that would be preferable to the `;` separated list

bin/faucet/src/main.rs
```diff
@@ -237,6 +254,15 @@ fn setup_tracing() {
     tracing_subscriber::registry().with(fmt.with_filter(filter)).init();
 }
 
+/// Generates a random API key for the faucet.
+/// The API key is a base64 encoded string with the prefix `miden_faucet_`.
+fn generate_api_key() -> String {
```
sergerad, line 259: Worth a simple unit test maybe

bin/faucet/src/server/get_tokens.rs
```diff
@@ -160,24 +166,39 @@ impl RawMintRequest {
             .validate(self.asset_amount)
             .ok_or(InvalidRequest::AssetAmount(self.asset_amount))?;
 
+        // Check the API key, if provideds
```
sergerad, line 169: `if provideds` typo

bin/faucet/src/server/get_tokens.rs
```diff
@@ -160,24 +166,39 @@ impl RawMintRequest {
...
+        if let Some(api_key) = &self.api_key {
+            if api_keys.contains(api_key) {
+                // If the API key is valid, we skip the PoW check
+                return Ok(MintRequest { account_id, note_type, asset_amount });
+            }
+            return Err(InvalidRequest::InvalidApiKey);
```
sergerad, line 175: nit: `if` `else` to avoid `return`

bin/faucet/src/server/mod.rs
sergerad: How does the unique key prevent limiting?

This will need an explainer comment above 👍

tomyrd: From my understanding, the key returned by the key extractor is a kind of identifier which is used by the limiter. Requests with the same key extracted will belong to the same group and will be limited together.
If we use the api key as the extracted key then different requests with the same api key will be grouped together in this specific rate limiter and will expend from the same pool of "available requests".
But for the case where they don't set the api key, if we just use `""` as the extracted key then all non-api key requests will be grouped together and the limit will be reached almost instantly. By making the extracted key `"[account_id]+[timestamp]"` we make it so that every request has a different extracted key and they won't get grouped (therefore they won't get limited).

bin/faucet/src/server/get_tokens.rs
tomyrd: I don't think we can avoid the `return` here as this `if` is returning from the whole function (an early return instead of a big if statement). But I might be missing something.

bin/faucet/src/server/mod.rs
sergerad: Thanks, great explanation. I found this comment more understandable than the explainer comment you added in the code however. This part in particular I thought was important to spell out:

> But for the case where they don't set the api key, if we just use "" as the extracted key then all non-api key requests will be grouped together and the limit will be reached almost instantly.

bin/faucet/src/server/get_tokens.rs
sergerad: Ah too right. The if above and being an early return both require return statement.

sergerad (approved):
Looks good. Minor comments left.

bin/faucet/src/server/mod.rs
tomyrd: I'll add it to the comment, was trying to summarize the idea so that the comment wasn't too long.

bin/faucet/src/server/mod.rs
tomyrd: Should this be done with all rate limiter parameters?

sergerad: Ah just noticed the other rate limit code is in the same boat. Lets just leave a TODO for now and we can do them all in a separate issue/PR.

tomyrd: Just added the TODO and rebased, will also comment about it on the tracking issue.

SantiagoPittella (approved):
LGTM, would be nice to have the issues of the TODOs before merging.

SantiagoPittella:
@tomyrd rebasing now should fix your CI errors.

igamigo (approved):

bobbinth:
I wonder if in the future we should add PoW requirements for API keys as well (i.e., each API key would have an independently-adjusted PoW requirement). And then, general access (i.e., `null` API key) would be just a special case of access via API keys. Let's create an issue for this.

PR #840, opened 2025-05-14
feat(ntx): Introduce network transaction builder

igamigo:
- Adds crate with `NetworkTransactionBuilder`, which mainly serves the gRPC service for the ntx builder and keeps an in-memory state. For now, this component is only started as part of the `bundled` command.
- Also adds some scaffolding for the next PR, that introduces the datastore for actually running and submitting transactions


crates/ntx-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+
+//mod data_store;
+mod store;
+
+/// Network Transaction Builder
+pub struct NetworkTransactionBuilder {
+    /// The address of the network transaction builder gRPC server.
+    pub address: SocketAddr,
+    /// Address of the store gRPC server.
+    pub store_address: SocketAddr,
```
sergerad, line 27: Irrelevant to this PR but is this an example where `SocketAddr` is the wrong abstraction because it means we can't use domains, only IPs? I have seen this in a few parts of the code base and made me wonder.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+use tokio_stream::wrappers::TcpListenerStream;
+use tower_http::trace::TraceLayer;
+use tracing::{debug, info, warn};
+
+use crate::COMPONENT;
+
+mod server;
+
+/// Interval for checking pending notes
+const NOTE_CHECK_INTERVAL_MS: u64 = 100;
```
sergerad, line 17: Probably something we want configurable

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+
+            loop {
+                interval.tick().await;
+
+                {
+                    let state_guard = match api_state.lock() {
+                        Ok(guard) => guard,
+                        Err(e) => {
+                            warn!(target: COMPONENT, error=%e, "Failed to lock API state - poisoned lock");
+                            continue;
```
sergerad, line 63: I think I have seen us `expect()` on locks like this in the code base. I think its OK to panic in that instance - but it depends on how important graceful shutdown (or the absence of it) is.

crates/ntx-builder/src/builder/server/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+        let tx_id = req.transaction_id.ok_or_else(|| {
+            Status::invalid_argument("transaction_id is required in SubmitNetworkNotesRequest")
+        })?;
+
+        info!(
+            target: COMPONENT,
+            tx_id = %tx_id,
+            note_count = req.note.len(),
+            "Received network notes"
+        );
```
sergerad, line 57: Might want to instrument this function. In which case could probably remove this event given its informational value is in the fields.

crates/ntx-builder/src/builder/server/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+        &self,
+        request: Request<UpdateNetworkNotesRequest>,
+    ) -> Result<Response<()>, Status> {
+        let request = request.into_inner();
+
+        info!(
+            target: COMPONENT,
+            update_count = request.nullifiers.len(),
+            "Received nullifier updates"
+        );
```
sergerad, line 82: Same idea a bout instrumenting.

crates/ntx-builder/src/builder/server/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+        &self,
+        request: Request<UpdateTransactionStatusRequest>,
+    ) -> Result<Response<()>, Status> {
+        let request = request.into_inner();
+
+        info!(
+            target: COMPONENT,
+            update_count = request.updates.len(),
+            "Received transaction status updates"
+        );
```
sergerad, line 114: Instrumenting comment as above

crates/ntx-builder/src/builder/mod.rs
igamigo: Yeah, I didn't think too much about this when writing the code but I agree it'd probably be better to comprehensively use `Url`

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -0,0 +1,133 @@
...
+
+            let tag = note.metadata().tag();
+            self.pending_notes_by_tag.entry(tag).or_default().push(note.clone());
+
+            self.unconsumed_note_queue.push_back(note);
+        }
+    }
+
+    /// Move notes from pending to inflight for a specific transaction
+    pub fn _move_notes_to_inflight(&mut self, tx_id: TransactionId, note_ids: &[NoteId]) {
```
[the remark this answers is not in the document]
igamigo: Yeah, this is not used here. I might remove this altogether because it gets refactored quite heavily on #841

crates/ntx-builder/src/builder/store.rs
```diff
@@ -0,0 +1,134 @@
...
+        let request = GetAccountDetailsRequest { account_id: Some(account_id.into()) };
+
+        let store_response = self
+            .inner
+            .clone()
+            .get_account_details(request)
+            .await?
+            .into_inner()
+            .details
+            .unwrap(); // TODO
```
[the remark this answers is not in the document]
igamigo: This gets addressed on #841 as well, but I might bring it here to make it clearer.

crates/ntx-builder/src/builder/server/mod.rs
igamigo: Instrumented function and removed the log here and in the other server functions

crates/ntx-builder/src/builder/store.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        Ok(Account::read_from_bytes(&faucet_account_state_bytes).unwrap())
+    }
+}
+
+// Store errors
+// =================================================================================================
+
+#[derive(Debug, Error)]
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
igamigo: Removed this!

crates/ntx-builder/src/builder/store.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /// Returns the latest block's header from the store.
+    #[instrument(target = COMPONENT, name = "store.client.get_unconsumed_network_notes", skip_all, err)]
+    pub async fn get_unconsumed_network_notes(&self) -> Result<Vec<Note>, StoreError> {
+        let mut all_notes = Vec::new();
+        let mut page_token: Option<u64> = None;
+
+        loop {
+            let req = GetUnconsumedNetworkNotesRequest { page_token, page_size: 128 };
```
TomasArrachea, line 77: is there a reason to use that specific page size?

igamigo: Not really, I think it could probably go way up before it being problematic but didn't think too hard about it

igamigo:
I moved some changes from #841 to here. Specifically, the `GetNetworkAccountDetailsByPrefix` store endpoint, because I had a placeholder method here which did not make sense and would be misleading when reviewing.

proto/proto/store.proto
```diff
@@ -23,6 +23,11 @@ service Api {
     // Returns the latest state of an account with the specified ID.
     rpc GetAccountDetails(requests.GetAccountDetailsRequest) returns (responses.GetAccountDetailsResponse) {}
 
+    
+    // Returns the latest state of an account with the specified ID.
+    // TODO: would probably be cleaner to do a specific response type instead of reusing GetAccountDetailsResponse
+    rpc GetNetworkAccountDetailsByPrefix(requests.GetNetworkAccountDetailsByPrefixRequest) returns (responses.GetAccountDetailsResponse) {}
```
bobbinth, line 29: Agreed re specific response type - I'd add this to the list of follow-ups for future PRs.

crates/store/src/db/sql/mod.rs
```diff
@@ -178,6 +178,32 @@ pub fn select_account(transaction: &Transaction, account_id: AccountId) -> Resul
...
+        "
+        SELECT
+            account_id,
+            account_commitment,
+            block_num,
+            details
+        FROM
+            accounts
+        WHERE
+            id_prefix = ?1;
```
bobbinth, line 197: As far as I can tell `id_prefix` field does not exist in the `accounts` table. Is there a commit missing from this PR?

crates/ntx-builder/src/builder/store.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// Essentially just a thin wrapper around the generated gRPC client which improves type safety.
+#[derive(Clone, Debug)]
+pub struct StoreClient {
+    inner: InnerClient,
+}
+
+impl StoreClient {
+    /// Creates a new store client with a lazy connection.
+    pub fn new(store_address: SocketAddr) -> Self {
```
bobbinth, line 41: Should this take `Url` as a parameter instead of `SocketAddr`?

crates/store/src/db/sql/mod.rs
```diff
@@ -178,6 +178,32 @@ pub fn select_account(transaction: &Transaction, account_id: AccountId) -> Resul
     account_info_from_row(row)
 }
 
+/// Select the latest account details by account id from the DB using the given [Connection].
+///
+/// # Returns
+///
+/// The latest account details, or an error.
+pub fn select_account_by_prefix(transaction: &Transaction, id_prefix: u32) -> Result<AccountInfo> {
```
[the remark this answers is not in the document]
igamigo: Actually because I had added a `UNIQUE` constraint (in the next PR, but will add it here as well; as you mentioned I had forgottten to add the table creation commit here), there could not be more than one account with a single prefix. This will prevent collisions, but will also do so for general accounts which might not be desirable.

An alternative could be:

- remove the unique constraint
- add the note/`TODO` as you mention here
- only store the prefix for network accounts

This reduces the probabilities of collisions but is less robust in the sense that there could be multiple network accounts with the same prefix

crates/store/src/db/sql/mod.rs
igamigo: Yes, I failed to add it here. Just added those changes

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -0,0 +1,113 @@
...
+#[derive(Debug)]
+pub struct NtxBuilderState {
+    unconsumed_note_queue: VecDeque<Note>,
+    /// A map of nullifiers mapped to their note IDs
+    notes_by_nullifiers: BTreeMap<Nullifier, NoteId>,
+    /// Pending network notes that have not been consumed as part of a committed transaction.
+    pending_notes_by_tag: BTreeMap<NoteTag, Vec<Note>>,
+    /// Inflight network notes with their associated transaction IDs
+    inflight_notes: BTreeMap<TransactionId, Vec<Note>>,
+}
```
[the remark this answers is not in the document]
bobbinth: > Yes, the main thing that I was keeping in mind is that if we requeue notes because somehow the transaction failed I'd want to make sure they get requeued last. But indeed it does not make sense to have the full note object twice.

We should be able to do the same with `NoteTag`-based queue, right?

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -0,0 +1,133 @@
...
+        self.inflight_notes.retain(|_, notes| {
+            notes.retain(|note| !to_remove.contains(&note.id()));
+            !notes.is_empty()
+        });
+
+        self.unconsumed_note_queue.retain(|note| !to_remove.contains(&note.id()));
+    }
+}
+
+// TODO: Add tests
```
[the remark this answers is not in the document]
igamigo: I'm adding tests in the #841, but I'm still missing a few

crates/ntx-builder/src/builder/server/state.rs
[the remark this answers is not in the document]
igamigo: Yeah, mostly. For instance, when discarding notes by nullifiers you cannot as easily remove one entry from the account queue. Although it's fine to leave it in the account queue and then if no notes are found, continue with the next ones.

crates/ntx-builder/src/builder/store.rs
igamigo: Changed this to take `Url`

crates/ntx-builder/src/builder/store.rs
```diff
@@ -0,0 +1,133 @@
...
+            .into_inner()
+            .block_header
+            .ok_or(miden_node_proto::generated::block::BlockHeader::missing_field(
+                "block_header",
+            ))?;
+
+        BlockHeader::try_from(response).map_err(Into::into)
+    }
+
+    /// Returns the latest block's header from the store.
```
SantiagoPittella, line 69: This doc is out of place

crates/store/src/db/sql/mod.rs
SantiagoPittella: `id_prefix` isn't indexed in the DB, this might end up being a really expensive query

crates/ntx-builder/README.md
```diff
@@ -0,0 +1,6 @@
+# Miden Network Transaction Builder
+
+This crate implements the Network Transaction Builder service for the Miden Node.
```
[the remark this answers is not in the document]
igamigo: Yeah, was keeping this in mind but did not want to lose much time before building it. Created https://github.com/0xMiden/miden-node/issues/871

crates/utils/src/network_note.rs
```diff
@@ -0,0 +1,49 @@
+use miden_objects::note::{Note, NoteExecutionMode, NoteId, NoteMetadata, NoteTag, Nullifier};
+use thiserror::Error;
+
+/// A newtype that wraps around notes targeting a single account to be used in a network mode.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct NetworkNote(Note);
```
igamigo, line 6: Wasn't too sure where to place this since it could eventually end up being used in various different crates, and this seemed like the less problematic one, but I could move it to the NTB crate if preferred

proto/proto/store.proto
igamigo: Ended up doing it here: https://github.com/0xMiden/miden-node/pull/840/commits/5069409e9b47678d5a6cd385b384600abd4c251d

crates/store/src/db/sql/mod.rs
```diff
@@ -438,6 +470,15 @@ pub fn upsert_accounts(
...
     for update in accounts {
         let account_id = update.account_id();
+        // Extract the 30-bit prefix to provide easy look ups for NTB
+        // TODO: this is hacky:
+        // Do not store prefix for accounts that are not network
+        let account_id_prefix = if account_id.is_network() {
+            NetworkAccountPrefix::try_from(account_id)?.inner()
+        } else {
+            0
+        };
```
bobbinth, line 480: Related to one of the previous comments: if we make `id_prefix` track only prefixes of network accounts, then it could be a nullable filed and we could set it to `NULL` for non-network accounts.

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Max number of notes taken for executing a single network transaction
+const MAX_BATCH: usize = 50;
+
+/// Maintains state of the network notes for the transaction builder.
+///
+/// Note tags are mainly kept in a [`VecDeque`], but also indexed by nullifiers and note tags in
+/// order to enable simple lookups when discarding by nullifiers and deciding which notes to
+/// execute against a unique account, respectively.
+/// Additionally, a list of inflight notes is kept to track their lifecycle.
```
bobbinth, line 17: I think this description needs to be updated (e.g., we don't index note tags by nullifier).

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Add network notes to the pending notes queue.
+    pub fn queue_unconsumed_notes(&mut self, notes: impl IntoIterator<Item = NetworkNote>) {
+        for n in notes {
+            let tag = n.metadata().tag();
+
+            self.by_nullifier.insert(n.nullifier(), n.id());
+            self.by_tag.entry(tag).or_default().push(n.clone());
+
+            self.account_queue.push_back(tag);
```
bobbinth, line 50: In case there are multiple notes per tag, we'll get duplicates in the `account_queue`, right? My thinking was that `account_queue` would contain unique tags. So, here, before adding the tag we'd check if the tags already in the queue, and if so, won't add it again.

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -0,0 +1,123 @@
...
+        self.by_tag.retain(|_, bucket| {
+            bucket.retain(|n| !ids.contains(&n.id()));
+            !bucket.is_empty()
+        });
+
+        self.inflight_txs.retain(|_, notes| {
+            notes.retain(|n| !ids.contains(&n.id()));
+            !notes.is_empty()
+        });
+    }
```
bobbinth, line 91: Maybe I'm missing something, but instead of this method, should we not have something like `insert_inflight()` which could look something like this:
```rust
pub fn insert_inflight(&mut self, tx: TransactionId, nullifiers: &[Nullifier]) {
    ...
}
```
The idea of this method would be to add an entry to the `inflight_txs` fields containing the notes which match the set of provided nullifiers. The overall idea is:

- When `update_network_notes()` endpoint is called, we call `insert_inflight()` to mark relevant pending note as being consumed by a transaction in the mempool. These relevant notes are removed from the `by_tag` map.
- when `update_transaction_status()` endpoint is called, we call either `commit_inflight()` or `rollback_inflight()` methods to either discard the now consumed notes or to return the notes to `by_tag` (in case the tx was rolled back).

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    pub fn rollback_inflight(&mut self, tx_id: TransactionId) -> usize {
+        if let Some(notes) = self.inflight_txs.remove(&tx_id) {
+            let count = notes.len();
+            // SAFETY: All transactions contain at least a note
+            self.account_queue.push_back(notes.first().unwrap().metadata().tag());
+            for n in notes {
+                self.by_tag.entry(n.metadata().tag()).or_default().push(n.clone());
+                self.by_nullifier.insert(n.nullifier(), n.id());
```
bobbinth, line 113: If we swap these lines, we won't need to clone the note, right?

bobbinth (commented):

crates/store/src/db/sql/mod.rs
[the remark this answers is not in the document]
bobbinth: I think for now (in the next PR, or maybe even in a follow-up), the solution could be to disallow two network accounts with the same prefix (but still allow two regular accounts with the same prefix). This could be checked before accepting a transaction that would create a network account into the mempool.

crates/utils/src/account.rs
```diff
@@ -0,0 +1,75 @@
...
+    account::AccountId,
+    note::{NoteExecutionMode, NoteTag},
+};
+use thiserror::Error;
+
+pub type AccountPrefix = u32;
+
+/// Wrapper for network account prefix
+/// Provides type safety for accounts that are meant for network execution
+pub struct NetworkAccountPrefix(u32);
```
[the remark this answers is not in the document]
igamigo: Agreed, while it doesn't feel directly connected to protobuf code it does make more sense.

crates/store/src/server/api.rs
```diff
@@ -291,6 +292,27 @@ impl api_server::Api for StoreApi {
...
+        ret(level = "debug"),
+        err
+    )]
+    async fn get_network_account_details_by_prefix(
+        &self,
+        request: Request<GetNetworkAccountDetailsByPrefixRequest>,
+    ) -> Result<Response<GetNetworkAccountDetailsByPrefixResponse>, Status> {
+        let request = request.into_inner();
+        let prefix = request.account_id_prefix;
+        assert!(prefix >> 30 == 0, "account_id_prefix must be 30 bits");
```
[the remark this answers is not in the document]
igamigo: Makes sense, changed it

crates/store/src/db/sql/mod.rs
igamigo: Yes, I made it nullable and created an index for it, should make for fast lookups and less bloated row size since most accounts will not be network accounts. A better alternative would be for it to have its own table but for now this should do OK.

crates/ntx-builder/src/builder/server/state.rs
igamigo: My thought was that having duplicate tags would be fine since you could take until one had notes remaining, but I modified it to only have one copy

crates/ntx-builder/src/builder/server/state.rs
igamigo: Yeah, makes sense. I understand better now. I refactored the struct around this. I also added a note ID -> note map since otherwise scanning for pending notes to remove in `insert_inflight` would be fairly complex (unnecessarily so, although this alternative incurs just a bit more cloning)

crates/ntx-builder/src/builder/server/state.rs
igamigo: Good call, thanks

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// set for the given `tx_id`, removing them from the indices and pruning empty tag buckets and
+    /// queue entries.
+    pub fn insert_inflight(&mut self, tx_id: TransactionId, nullifiers: &[Nullifier]) {
+        if nullifiers.is_empty() {
+            return;
+        }
+        let mut moved = Vec::new();
+
+        for nullifier in nullifiers {
+            if let Some(id) = self.by_nullifier.remove(nullifier) {
```
bobbinth, line 91: Should we add a comment here saying that if we didn't find the ID for the note, it means this is not a network note?

Though, thinking a bit more about it, I think we may have a race condition:

- Let's say a transaction creating a network note is accepted to the mempool. The NTB doesn't know about this transaction yet (we only communicate with NTB when transaction gets committed).
- Then, before the transaction gets committed another transaction that consumes the network note (as an unauthenticated note) is accepted to the mempool. The `insert_inflight()` will get called, but we don't know about this network note - so, the info will be ignored.
- Then, the first transaction gets committed but the second transaction still stays in the mempool (to be committed in a future block). At this point `queue_unconsumed_notes()` will be called, and we'll start tracking this network note.
- Finally, the second transaction gets committed. The note will be consumed, but NTB won't be notified of this - and so, will keep thinking that the note is available for consumption.

This would get fixed once we start calling `queue_unconsumed_notes()` when transactions creating the notes are accepted into the mempool - so, not sure we need to fix this now. But we should make sure we don't forget this edge case.

crates/ntx-builder/src/builder/server/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+                    Status::invalid_argument(format!(
+                        "transaction ID from request is not valid: {err}"
+                    ))
+                })?;
+
+            if TransactionStatus::Commited == tx.status() {
+                state.commit_inflight(tx_id);
+            } else {
+                state.rollback_inflight(tx_id);
+            }
```
bobbinth, line 127: nit: I would write this as a `match` statement:
```rust
match tx.status() {
    TransactionStatus::Commited => state.commit_inflight(tx_id);
    TransactionStatus::Failed => state.rollback_inflight(tx_id);
}
```
Also:
- Typo: `Commited` -> `Committed`.
- I would probably rename `Failed` into `Reverted`.

bobbinth (approved):
Looks good! Thank you! I left some more comments inline. Most of them are pretty small, and the bigger one we can leave for the future.

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -0,0 +1,278 @@
...
+        Some((next_tag, notes))
+    }
+
+    /// Move the notes whose nullifiers belong to the input `nullifiers` into the in-flight
+    /// set for the given `tx_id`, removing them from the indices and pruning empty tag buckets and
+    /// queue entries.
+    pub fn insert_inflight(&mut self, tx_id: TransactionId, nullifiers: &[Nullifier]) {
+        if nullifiers.is_empty() {
+            return;
+        }
```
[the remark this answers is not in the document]
igamigo: I left it mostly as a sanity check, but it should not happen

bobbinth (approved):
Looks good! Thank you! I noticed just one more typo, and also I think https://github.com/0xMiden/miden-node/pull/840#discussion_r2106117858 still needs to be addressed.

PR #841, opened 2025-05-14
feat(ntx): Execute network transactions

igamigo:
Last PR in the string of changes related to network transactions.
Tracing and error work is what's mainly missing, as well as forcing delegated proving to be used.


TomasArrachea (commented):
just some minor details

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
 members = [
-  "bin/faucet",
+  #"bin/faucet",
```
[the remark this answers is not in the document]
igamigo: Thanks, I can now enable this again after the refactor merges.

crates/block-producer/src/server.rs
```diff
@@ -41,7 +41,7 @@ use crate::{
 /// until the store becomes available. Once the connection is established, the block producer
 /// will start serving requests.
 pub struct BlockProducer {
-    /// The address of the block producer component.
+    /// The listener for the block producer component.
     pub block_producer_address: SocketAddr,
```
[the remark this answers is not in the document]
igamigo: I had changed this to take a listener but it was not necessary for this PR so I rolled it back

crates/ntx-builder/src/builder/data_store/account_cache.rs
```diff
@@ -0,0 +1,179 @@
...
+                generation: current_generation,
+                account: account.clone(),
+            },
+        ) {
+            Some(entry.generation)
+        } else {
+            None
+        };
+
+        if let Some(old) = removed_generation {
```
sergerad, line 81: nit: could delete this `if let Some` and update line 69 to be `if let Some`

crates/ntx-builder/src/builder/data_store/account_cache.rs
```diff
@@ -0,0 +1,179 @@
...
+            state.ordering.remove(&old);
+        }
+        state.ordering.insert(current_generation, account_id);
+
+        if state.accounts.len() > self.capacity {
```
sergerad, line 86: nit: tend to put explainer comments above even obvious operations like this to make it easier for people to scan through

sergerad: e.g.
```rust
// Remove oldest element from the cache if capacity full.
if state.accounts.len() > self.capacity {
...
```

igamigo: I was just pushing some changes which should address some of this, but tbh I think the `AccountCache` could still use a decent amount of polishing (left a `TODO`, maybe we should even replace it for a more robust/performant crate)

crates/ntx-builder/src/builder/data_store/account_cache.rs
```diff
@@ -0,0 +1,288 @@
...
+    fn insert_get() {
+        let account = create_account(10);
+        let cache = AccountCache::new(2);
+
+        cache.put(&account.clone());
+        assert_eq!(cache.get(account.id().into()).unwrap(), account);
+    }
+
+    #[test]
+    fn lru_evicts_least_recently_used_account() {
```
sergerad, line 162: Any reason we wouldn't use something like https://docs.rs/lru/latest/lru/ instead of writing our own here?

crates/ntx-builder/src/builder/data_store/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+    mast_forest_store: TransactionMastStore,
+    store_client: StoreClient,
+    account_cache: AccountCache,
+    block_ref: Mutex<BlockHeader>,
+    partial_mmr: Mutex<PartialMmr>,
+}
+
+impl NtxBuilderDataStore {
+    pub async fn new(store_client: StoreClient) -> Result<Self, StoreError> {
+        let account_cache = AccountCache::new(128);
```
sergerad, line 34: nit: Might want to make it a const just so its clearer for us to reconsider that value down the line

crates/ntx-builder/src/builder/data_store/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+        let mast_forest_store = TransactionMastStore::new();
+
+        // SAFETY: ok to unwrap because passing `None` should return the latest data everytime
```
sergerad, line 37: nit: how do we feel about using `expect()` instead of safety comments for unwraps?

crates/store/src/db/migrations/001-init.sql
```diff
@@ -149,4 +149,4 @@ CREATE TABLE transactions (
 ) STRICT, WITHOUT ROWID;
 
 CREATE INDEX idx_transactions_account_id ON transactions(account_id);
-CREATE INDEX idx_transactions_block_num ON transactions(block_num);
+CREATE INDEX idx_transactions_block_num ON transactions(block_num);
```
bobbinth, line 152: We can probably revert this change.

bin/node/src/commands/bundled.rs
```diff
@@ -57,6 +57,11 @@ pub enum BundledCommand {
         #[arg(long = "block-prover.url", env = ENV_BLOCK_PROVER_URL, value_name = "URL")]
         block_prover_url: Option<Url>,
 
+        /// The remote transaction prover's gRPC url, used for the ntx builder. If unset,
+        /// will default to running a prover in-process which is expensive.
+        #[arg(long = "tx-prover.url", env = ENV_NTX_PROVER_URL, value_name = "URL")]
+        tx_prover_url: Option<Url>,
```
bobbinth, line 63: nit: I would move this above `batch_prover_url` (i.e., above line 50).

bin/node/src/commands/bundled.rs
```diff
@@ -131,11 +138,13 @@ impl BundledCommand {
         }
     }
 
+    #[allow(clippy::too_many_arguments)]
     async fn start(
         rpc_url: Url,
         data_directory: PathBuf,
         batch_prover_url: Option<Url>,
         block_prover_url: Option<Url>,
+        ntx_proving_service_address: Option<Url>,
```
bobbinth, line 147: nit: I think we can keep this named `tx_prover_url` (or at least change `_address` to `_url` for consistency).

Also, not for this PR, but it may make sense to group prover configs into a dedicated struct - e.g.,:
```rust
pub struct ProverConfig {
    tx_prover_url: Option<Url>,
    batch_prover_url: Option<Url>,
    block_prover_url: Option<Url>,
}
```

bin/node/src/commands/bundled.rs
```diff
@@ -191,10 +199,15 @@ impl BundledCommand {
...
-                NetworkTransactionBuilder { address: ntx_builder_address, store_url }
-                    .serve()
-                    .await
-                    .context("failed while serving store component")
+                NetworkTransactionBuilder {
+                    ntx_builder_address,
+                    store_url,
+                    block_producer_address,
+                    proving_service_address: ntx_proving_service_address,
+                }
```
bobbinth, line 207: A couple of consistency questions:

- Why do we use `Url` for the store component but `SocketAddr` for `ntx_builder_address` and `block_producer_address`?
- Also, why is `proving_service_address` has `_address` suffix while it is actually a `Url`? Shouldn't it be `proving_service_url`? Or maybe even just `tx_prover_url`?

crates/block-producer/src/server.rs
```diff
@@ -361,11 +362,11 @@ mod test {
             BlockProducer {
                 block_producer_address: block_producer_addr,
                 store_address: store_addr,
-                ntx_builder_address,
                 batch_prover_url: None,
                 block_prover_url: None,
                 batch_interval: Duration::from_millis(500),
                 block_interval: Duration::from_millis(500),
+                ntx_builder_address,
```
bobbinth, line 369: nit: should we use `_addr` suffix (sims like that's what we use in other cases)? Or maybe we should just use `_address` suffix eveywhere?

Also, I'd probably move this up to be after `store_address`.

crates/rpc/src/server/api.rs
```diff
@@ -208,6 +211,18 @@ impl api_server::Api for RpcService {
         let tx = ProvenTransaction::read_from_bytes(&request.transaction)
             .map_err(|err| Status::invalid_argument(format!("Invalid transaction: {err}")))?;
 
+        // Only allow deployment transactions for new network accounts
+        if tx.account_id().is_network() {
+            match tx.account_update().details() {
+                AccountUpdateDetails::New(_) => {},
```
bobbinth, line 217: It feels like we should have something like `AccountUpdateDetails::is_new_account()` (let's add it in `miden-base`?). But also, could we not use `matches!` macro here for brevity?

bobbinth (commented):
Thank you! Looks good! Not a review yet, but I left a few small comments inline.

crates/ntx-builder/src/builder/data_store/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        };
+
+        if let Some(acc) = &account {
+            self.mast_forest_store.insert(acc.code().mast());
+        }
+
+        Ok(account)
+    }
+
+    pub async fn update_blockchain_data(&self) -> Result<BlockNumber, StoreError> {
```
bobbinth, line 79: As mentioned in one of the other comments, in the future, we could probably just add block headers to the state once they are committed to the store. So, this could become something like:
```rust
pub fn add_block_header(&mut self, block_header: BlockHeader) -> Result<BlockNumber, SomeError> {
    ...
}
```
Internally, it would update the `partial_mmr` and `block_ref` fields (similar to how does now).

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -1,95 +1,305 @@
...
+    executed_tx: &ExecutedTransaction,
+    remote_tx_prover: Option<&RemoteTransactionProver>,
+) -> anyhow::Result<()> {
+    api_state.lock().await.insert_inflight(
+        executed_tx.id(),
+        executed_tx
+            .input_notes()
+            .iter()
+            .map(transaction::ToInputNoteCommitments::nullifier),
+    );
```
bobbinth, line 289: Why do we need to do this here? Wouldn't it be done implicitly because once the block producers accepts the transaction, it would call `UpdateNetworkNotes` which should have the same effect?

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -1,95 +1,305 @@
...
+
+    let tx_witness = TransactionWitness::from(executed_tx.clone());
+    let proven_tx = if let Some(tx_prover) = remote_tx_prover {
+        tx_prover.prove(tx_witness).await
+        // TODO: should we fall back to a local tx prover if proving fails?
```
bobbinth, line 294: In production (or event testnet) scenarios, local proving would not be a feasible solution - so, we shouldn't fall back to it if remote proving fails.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -1,95 +1,305 @@
...
+    }
+}
+
+/// Proves and submit an executed transaction
+async fn prove_and_submit(
+    api_state: &SharedPendingNotes,
+    block_producer_client: &BlockProducerClient,
+    executed_tx: &ExecutedTransaction,
+    remote_tx_prover: Option<&RemoteTransactionProver>,
+) -> anyhow::Result<()> {
```
bobbinth, line 282: I would maybe create an enum - something like `TransactionProver` which could look like so:
```rust
enum TransactionProver {
    Local(LocalTransactionProver),
    Remote(RemoteTransactionProver),
}
```
And then this function would be a method on this struct.

crates/ntx-builder/src/builder/data_store/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        // SAFETY: datastore impl checks that the account ID is a valid network account
+        let account_id_prefix = transaction.account_id().try_into().unwrap();
+        let Some(mut account) = self.account_cache.get(account_id_prefix) else {
+            warn!(target:COMPONENT, "account was expected to be found in the cache");
+            return Ok(());
+        };
+        account.apply_delta(transaction.account_delta())?;
+        self.account_cache.put(&account).unwrap();
+        Ok(())
+    }
```
bobbinth, line 114: Not for this PR, but what happens if a network transaction fails after it has been accepted by the block producer and we need to revert the state of the account to what it was before the transaction? (it's also possible I missed this, but if not, let's create an issue for it).

bobbinth (commented):

Also, a question: it seems like we put pretty much everything into `src/builder` directory - but isn't the whole crate for the builder? I wouldn't change this now, but we could consider reshuffling things in a follow-up PR.

bin/node/src/commands/bundled.rs
bobbinth: > I originally used address for everything for consistency, since this is how it was done with all other components. As I mentioned in one of the other comments, I didn't want to change that too much in these PRs. I think `Url` is a better approach overall, but since we are running everything with the `bundled` command anyway for now, it didn't make sense to change it here. I changed the store component here to `Url` to make it consistent with the change from [this comment](https://github.com/0xMiden/miden-node/pull/840#discussion_r2103995425), since otherwise it seems like it doesn't make too much sense to transform a socket to URL internally. Although in this `bundled` command we basically do that, the NTB will have its standalone command.

Makes sense! Should we create an issue to make things a bit more consistent in the future?

crates/ntx-builder/src/builder/data_store/mod.rs
igamigo: What would happen is that the next execution with this account would fail due to an incorrect state and then the account would get manually evicted (but not retried).

Cargo.toml
```diff
@@ -31,36 +32,36 @@ version      = "0.9.0"
...
+miden-node-block-producer    = { path = "crates/block-producer", version = "0.9" }
+miden-node-ntx-builder       = { path = "crates/ntx-builder", version = "0.9" }
+miden-node-proto             = { path = "crates/proto", version = "0.9" }
+miden-node-proto-build       = { path = "proto", version = "0.9" }
+miden-node-rpc               = { path = "crates/rpc", version = "0.9" }
+miden-node-store             = { path = "crates/store", version = "0.9" }
+miden-node-test-macro        = { path = "crates/test-macro" }
+miden-node-utils             = { path = "crates/utils", version = "0.9" }
+miden-objects                = { version = "0.9" }
+miden-processor              = { version = "0.13" }
```
SantiagoPittella, line 51: version 0.14 of miden-processor is available, if its compatible we should update it.

crates/ntx-builder/src/builder/block_producer.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct BlockProducerClient {
+    inner: InnerClient,
+}
+
+impl BlockProducerClient {
+    /// Creates a new block producer client with a lazy connection.
+    pub fn new(block_producer_address: SocketAddr) -> Self {
+        let block_producer_url = format!("http://{block_producer_address}");
+        let channel =
+            tonic::transport::Endpoint::try_from(block_producer_url).unwrap().connect_lazy();
```
SantiagoPittella, line 29: Should this error be handled with a `Result`?

crates/ntx-builder/src/builder/data_store/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+        };
+
+        if let Some(acc) = &account {
+            self.mast_forest_store.insert(acc.code().mast());
+        }
+
+        Ok(account)
+    }
+
+    /// Updates the blockchain-related data by
```
SantiagoPittella, line 93: This doc looks unfinished.

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -81,8 +81,13 @@ impl PendingNotes {
     /// Move the notes whose nullifiers belong to the input `nullifiers` into the in-flight
     /// set for the given `tx_id`, removing them from the indices and pruning empty tag buckets and
     /// queue entries.
-    pub fn insert_inflight(&mut self, tx_id: TransactionId, nullifiers: &[Nullifier]) {
-        if nullifiers.is_empty() {
+    pub fn insert_inflight(
+        &mut self,
+        tx_id: TransactionId,
+        nullifiers: impl IntoIterator<Item = Nullifier>,
```
SantiagoPittella, line 87: Not sure why we are using an iterator here instead of the slice.

crates/ntx-builder/src/builder/data_store/account_cache.rs
```diff
@@ -0,0 +1,278 @@
...
+    ordering: BTreeMap<usize, NetworkAccountPrefix>,
+}
+
+/// A capacity-limited network account cache.
+///
+/// The cache works in an LRU fashion and evicts accounts once capacity starts getting exceeded.
+pub struct NetworkAccountCache {
+    capacity: usize,
+    state: Mutex<CacheState>,
+}
```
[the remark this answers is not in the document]
igamigo: Made it a wrapper to that crate's struct. Wasn't too familiar with any option and I wanted to get something out quickly at first but was able to test `Lru` and it seems to work just fine for this.

igamigo:
@bobbinth I have mainly 2 remaining questions:

- I run the consumption check and only execute successful notes against the account. This is not very complete because I understand the checker stops at the first failed note, so there may be a bunch of usable notes that I am not consuming. I can  loop until we test all notes (not sure if this was the intended direction), but that seems unclean and also potentially problematic in the sense that it could incur a bunch of overhead from malicious users? In any case, what should we do with the failing note and the ones we did not try? One option is to discard the failing note and requeue the ones I didn't consume. The other option (currently implemented), is to requeue anything we don't successfully consume. I think they could get requeued maybe in reverse order to not complicate things in the next iteration of the same account.

- I'm not sure how we want to handle failures. Right now, the NTB runs 2 main tasks: the RPC server and the executor task. If the executor task fails, I'm capturing all errors and continuing. If it panics, I'm restarting the task. So in that sense the NTB can never exit due to the executor task. For the RPC task, if it somehow fails it seems like it should crash the node (as do the store, RPC, and block producer tasks). Does that sound right?

SantiagoPittella (commented):
Looking good, I'm doing some tests.

I recall some talks about making the faucet a network account and cannot remember if a decision was already made. We should create an issue to track that if we want to change it.

bobbinth (approved):
Looks good! Thank you! I left just one small comment inline. To answer the questions:

> I run the consumption check and only execute successful notes against the account. This is not very complete because I understand the checker stops at the first failed note, so there may be a bunch of usable notes that I am not consuming. I can loop until we test all notes (not sure if this was the intended direction), but that seems unclean and also potentially problematic in the sense that it could incur a bunch of overhead from malicious users? In any case, what should we do with the failing note and the ones we did not try? One option is to discard the failing note and requeue the ones I didn't consume. The other option (currently implemented), is to requeue anything we don't successfully consume. I think they could get requeued maybe in reverse order to not complicate things in the next iteration of the same account.

I think this is fine for now. In the future, we'll need to do two things:

- Make the note consumption checks more robust - but this would need to be done in `miden-base` (i.e., in `NoteConsumptionChecker:: check_notes_consumability()`). Could you create an issue for this?
- Develop a heuristic that (a) uses note execution hints to determine if we should even try to execute the note at this time, and (b) keeps track of note failures to figure out at which point we should try to stope executing notes.

> I'm not sure how we want to handle failures. Right now, the NTB runs 2 main tasks: the RPC server and the executor task. If the executor task fails, I'm capturing all errors and continuing. If it panics, I'm restarting the task. So in that sense the NTB can never exit due to the executor task. For the RPC task, if it somehow fails it seems like it should crash the node (as do the store, RPC, and block producer tasks). Does that sound right?

This is also probably fine for now (though, may lead to some unexpected results because after restart NTB would loose its inflight state). Could you create an issue to make this more robust in the future (i.e., once NTB restarts it would need to ask the block producer for the info it needs to re-build the inflight state - this may require some mempool introspection endpoints).

crates/ntx-builder/src/builder/server/state.rs
igamigo: I think this is more flexible, but . Specifically, RPC endpoints and executed transactions both yield owned collections that can be mapped to a nullifier iterator.

crates/ntx-builder/src/builder/data_store/mod.rs
igamigo: Good point, I think I copied it incorrectly. Changed!

crates/ntx-builder/src/builder/block_producer.rs
igamigo: Should be fine to unwrap here because we are converting from a `SocketAddr`, but forgot the safety comment. Added.

igamigo:
> This is also probably fine for now (though, may lead to some unexpected results because after restart NTB would loose its inflight state). Could you create an issue to make this more robust in the future (i.e., once NTB restarts it would need to ask the block producer for the info it needs to re-build the inflight state - this may require some mempool introspection endpoints).

Couple of things here:

- While the RPC is alive, the inflight state is maintained. The ticker gets restarted if it somehow crashes (all errors should be handled gracefully within it) and clones a handle to the inflight state.
- If the RPC service somehow crashes as well __and__ we want to restart it (right now if the RPC handle aborts we crash the node entirely as we do with other RPC handles), it gets initialized by asking the store for any unconsumed network notes.

PR #842, opened 2025-05-14
fix: duplicated note ids

### LINKED ISSUE #823: Faucet generates duplicate output notes
Faucet transactions get rejected by the node every now and then, apparently due to duplicate output note IDs.

Truncated log:

```
status: InvalidArgument, message: "transaction verification failed\nCaused by: output note IDs already used: [NoteId(RpoDigest([7260004778636565067, 12512392632442945940, 10077979699473102117, 13867741917139306314]))]"
```

This repeats fairly regularly.

## Initial investigation

Output notes IDs are randomised as part of the p2id note script creation. The faucet correctly passes in rng here which seems like it should work.

The note IDs from the error logs are also all different, which would indicate that they are at least somewhat random.

Barring other bugs or mistakes, the error originates from https://github.com/0xMiden/miden-node/blob/a0fdf2b45600ec5ac26af7c5693e9a1759aace38/crates/block-producer/src/mempool/inflight_state/mod.rs#L189-L195 in the node.

This indicates the node _already_ has this note in its mempool - which at first glance implies

- faucet notes aren't random enough
- faucet is sending 2x txs sometimes (unsure how this would be possible)
- mempool bug
- error stringification bug

Overall this is a weird one.

## Replication

So far this has only been seen on the official faucet while under load. Which makes sense since the node needs to have the note ID in its mempool for the conflict to occur.

Maybe we can replicate locally by creating a script to request multiple tokens sequentially.

If unable to replicate or figure it out, we'll need to add more logs to the testnet node and faucet to see which txs are generating which notes.

Its of course also possible that the faucet note clashes with other faucet's note IDs somehow.

SantiagoPittella:
Closes #823 

The issue of the duplicated notes ID's had its origin in always using the same serial numbers for notes. That, generally, wasn't a problem because we were usually using different receiver IDs and asset amounts, and that modifies the final `NoteId`, but when repeating those values it was generating the error described in the issue.

The serial number was always the same because we were copying the Rng generator, thus always using the same inner index for the generator and getting the same value.

### DIFF
--- a/bin/faucet/src/faucet/mod.rs
+++ b/bin/faucet/src/faucet/mod.rs
@@ -12,7 +12,7 @@ use miden_objects::{
     block::BlockNumber,
     crypto::{
         merkle::{MmrPeaks, PartialMmr},
-        rand::{FeltRng, RpoRandomCoin},
+        rand::RpoRandomCoin,
     },
     note::Note,
     transaction::{
@@ -30,7 +30,7 @@ use miden_tx::{
 use rand::{random, rngs::StdRng};
 use serde::Serialize;
 use store::FaucetDataStore;
-use tokio::sync::mpsc::Receiver;
+use tokio::sync::{Mutex, mpsc::Receiver};
 use tonic::Code;
 use tracing::{error, info, instrument, warn};
 use updates::{ClientUpdater, MintUpdate, ResponseSender};
@@ -160,6 +160,7 @@ pub struct Faucet {
     tx_prover: Arc<FaucetProver>,
     tx_executor: Rc<TransactionExecutor>,
     account_interface: AccountInterface,
+    rng: Arc<Mutex<RpoRandomCoin>>,
 }
 
 impl Faucet {
@@ -242,6 +243,9 @@ impl Faucet {
         let tx_executor =
             Rc::new(TransactionExecutor::new(data_store.clone(), Some(authenticator.clone())));
 
+        let coin_seed: [u64; 4] = random();
+        let rng = RpoRandomCoin::new(coin_seed.map(Felt::new));
+
         Ok(Self {
             data_store,
             id,
@@ -249,6 +253,7 @@ impl Faucet {
             tx_prover,
             tx_executor,
             account_interface,
+            rng: Arc::new(Mutex::new(rng)),
         })
     }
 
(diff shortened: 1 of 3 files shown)


bin/faucet/src/faucet/mod.rs
```diff
@@ -242,13 +243,17 @@ impl Faucet {
...
+        let rng = RpoRandomCoin::new(coin_seed.map(Felt::new));
+
         Ok(Self {
             data_store,
             id,
             prior_state: VecDeque::new(),
             tx_prover,
             tx_executor,
             account_interface,
+            rng: Arc::new(Mutex::new(rng)),
```
sergerad, line 256: I'm just wondering if we prefer to take the compute hit on creating a new `RpoRandomCoin` for every batch request instead of locking over the relevant section given this is part of the faucet RPC/API stack.
```
        let p2id_notes = {
            let mut rng_guard = self.rng.lock().await;
            P2IdNotes::build(self.faucet_id(), requests, &mut rng_guard)?
        };
```

Only significant if we expect that section to take like an order of magnitude or more than generating the `RpoRandomCoin`. Probably not worth spending time benchmarking etc, just wondering.

igamigo: Yeah, we briefly discussed this as well. Initializing an `RpoRandomCoin` from `thread_rng` does not look like it's trivial but it's probably less overhead than keeping a global lock (not sure by how much, though). We could also keep `thread_local` coins and take those for less locking overhead, but seems overkill. I don't have too much of a strong opinion as this depends a lot on traffic patterns (I think we can expect a relatively low number of concurrent request to hit this point) and it's hard to tell without measuring but I think I lean towards instantiating a new coin each thread.

If we keep this it might be a good idea to emit a trace of how long it took to acquire the lock maybe.

sergerad: I think if the purpose of the `Arc<Mutex<` is for performance, I would remove it. If we are concerned, we could add a TODO or NOTE about performance overhead of the relevant section for each batch request.

I prefer to not use the `Arc<Mutex<` because:
1. The use case here is not one of state sharing; and
2. The performance improvement is speculative or at least likely nominal.

I also suspect that the faucet will never be compute bound (so the random coin generation is fine to repeat).

Just my 2 cents. Feel free to proceed how you prefer.

bobbinth (approved):
Looks good! Thank you! Let's fix the lint errors and then we can merge.

Issue #804, opened 2025-04-25
RPC version header


Mirko-von-Leipzig:
Add RPC versioning via the http header.

## Background

At present each new (major) release is largely not backwards compatible and requires gRPC clients to update to the latest gRPC definitions. There is currently no mechanism for the clients to specify the version of gRPC they're expecting and this can lead to extremely confusing errors when an incompatible client connects to a node. Method signatures might even be compatible, but some fields are serialized bytes and would be gibberish when decoded. These errors are impossible to log or identify by the client and are even more confusing to the user of the client.

As an example, running the official client from its `next` development branch defaults to connecting to the official testnet node. These are usually incompatible since the development branch is working towards the _next_ release already and the official testnet is running the latest release version. This happens rather innocently and is impossible for the user to debug. cc @igamigo - imo having any network be the default is.. dangerous.

To make this less likely, and easy to identify, lets add support for clients to specify the version they're using via an http header. Alternatives are using routes or query params but these are fairly intrusive imo.

## Design

With headers we can either use a new custom header e.g. `X-MIDEN-RPC-VERSION`, or use the formal `Accept` header ([example blog post](https://jacobbednarz.com/posts/using-accept-header-for-http-api-versioning)). I recommend using the latter, we may run into issues with `tonic`/`axum` though I think its unlikely they (ab)use this header. Its also possible that proxies or other entities could fiddle with this, but let's see what happens.

```sh
# General format
Accept: application/vnd.{app_name}.{version}+{response_type}

# Our expected format
Accept: application/vnd.miden.{version}+grpc
```

I'm just assuming `{response_type}` should be `grpc` here. Maybe double-check if there is a standard for this.

We could also add the _network_ name to the header, e.g. `{app_name} = miden.testnet`. But that feels.. dangerous in other ways. And we don't currently name our networks internally but we may start once the account ID bech32 lands.

The version would match the node version, but with semver considerations. So `v0.8` currently (excluding the patch). At this point the RPC component would _reject_ requests with incompatible versions in the header, which at this stage would be any mismatch in version.

I suggest we also _require_ this header. So if not present we reject the request. This would force all clients to implement this feature. The softer alternative is to allow it when missing, which is also okay but I somewhat like that clients should be more specific.

This only affects the RPC component, internal links shouldn't care. We could consider adding it on the internal gRPC as well, which would help when components are spun up separately.

## Summary

- Add new (required) `Accept` header to (at least) RPC component
[...]

PR #844, opened 2025-05-15
feat: Add version-specific ACCEPT header to RPC

sergerad:
Relates to #804.

#### Context

We want to enforce semver compatibility between the RPC server and clients connecting to it.

The HTTP ACCEPT header can be used to communicate semver expectations / requirements between client / server.

#### Changes

rpc crate
- Add tower layer `AcceptLayer` which processes the HTTP ACCEPT header.
- Add logic to convert Cargo package version into a version requirement (e.g. "0.9.0"->">=0.9").
- Integrate the `AcceptLayer` to the RPC server stack.
- Add a module to rpc crate to create RPC clients with the request header in place.
- Add unit tests to validate that the Cargo package's version is turned into a valid version requirement.
- Add tests to validate RPC server rejects HTTP requests without the requisite header.
- Add tests to validate RPC server rejects HTTP requests without the correct version specified.

faucet crate
- Integrate new RPC client.


sergerad:
I've put in this review before doing the client changes to get initial feedback on the approach.

Actually unsure how to add custom headers to the gRPC client. Expecting / hoping it to be possible.

sergerad:
Have managed to get the interceptor approach to work in the tests there. Adds a bit of complexity and I need to dedupe a fair bit of code.

@TomasArrachea I presume the `ApiClient` in this repo is just for testing purposes and I will need to update the miden-client repo crate(s) before shipping this? 🙏

TomasArrachea:
> @TomasArrachea I presume the `ApiClient` in this repo is just for testing purposes and I will need to update the miden-client repo (or base?) crate(s) before shipping this? 🙏

You mean the `rpc::api_client::ApiClient` right? That is the client generated automatically by the proto files on the RPC interface. It's used on the faucet to connect with the RPC api, so it's not only for testing. It lacks the tls config, so that is why it should not be used from the outside to connect with the official RPC api. So yes, we should make a PR in miden-client to add this new header to the client.

crates/rpc/src/server/accept.rs
```diff
@@ -0,0 +1,188 @@
...
+use crate::COMPONENT;
+
+/// Layer responsible for handling HTTP ACCEPT headers.
+#[derive(Clone)]
+pub struct AcceptLayer {
+    version: VersionReq,
+}
+
+impl AcceptLayer {
+    /// Create a new accept layer with the specified version requirement.
```
bobbinth, line 22: nit: I think the phrasing here can be improved a bit: we are not specifying the version, we are reading from `Cargo.toml`.

crates/rpc/src/server/accept.rs
```diff
@@ -0,0 +1,188 @@
...
+    ///
+    /// # Panics:
+    ///
+    /// Panics if the version string in Cargo.toml is not valid semver.
+    /// The version string is made into an env var at compile time which means
+    /// that the unit tests prove this cannot panic in practice.
+    pub fn new() -> Self {
```
bobbinth, line 29: nit: we usually wrap comments after 100 chars.

crates/rpc/src/server/accept.rs
```diff
@@ -0,0 +1,188 @@
...
+        // Parse the full version string (e.g., "0.8.0").
+        let full_version = env!("CARGO_PKG_VERSION");
+
+        // Extract major and minor version components
+        let parts: Vec<&str> = full_version.split('.').collect();
+        let major = parts.first().expect("package semver has major");
+        let minor = parts.get(1).expect("package semver has minor");
```
bobbinth, line 36: Instead of parsing the version manually, could we not use `Version::parse()`?

crates/rpc/src/server/accept.rs
```diff
@@ -0,0 +1,188 @@
...
+
+impl<'a> TryFrom<&'a str> for AcceptHeaderValue<'a> {
+    type Error = nom::Err<nom::error::Error<&'a str>>;
+
+    fn try_from(input: &'a str) -> Result<Self, Self::Error> {
+        Self::parse(input).map(|(_, value)| value)
+    }
+}
+
+#[cfg(test)]
```
bobbinth, line 164: nit: let's add `TESTS` section separator above this line.

crates/rpc/src/server/accept.rs
```diff
@@ -0,0 +1,188 @@
...
+                },
+                Err(e) => {
+                    debug!(target: COMPONENT, "Failed to stringify accept header value: {}", e);
+                    bad_request("Invalid accept header value").boxed()
+                },
+            }
+        } else {
+            debug!(target: COMPONENT, "Request missing ACCEPT header");
+            bad_request("Missing required ACCEPT header").boxed()
+        }
```
bobbinth, line 115: nit: I would maybe reduce the level of nesting here by adding a helper function which takes a `request` and returns `AcceptHeaderValue`.

crates/rpc/src/server/accept.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// Parses the given input string into an [`AcceptHeaderValue`].
+    fn parse(input: &str) -> IResult<&str, AcceptHeaderValue> {
+        let (input, _) = tag("application/vnd.")(input)?;
+        let (input, app_name) = take_until(".")(input)?;
+        let (input, _) = tag(".")(input)?;
+        let (input, version) = take_until("+")(input)?;
+        let (response_type, _) = tag("+")(input)?;
+
+        Ok(("", AcceptHeaderValue { app_name, version, response_type }))
+    }
```
bobbinth, line 153: Why do we need to use `IResult` here as the return type and pass an empty string as the first parameter?

Or more generally, why not move this logic into the `try_from()` method below and get rid of `parse()` method?

crates/proto/src/client.rs
```diff
@@ -0,0 +1,55 @@
...
+use tonic::{
+    metadata::AsciiMetadataValue,
+    service::{Interceptor, interceptor::InterceptedService},
+    transport::Channel,
+};
+use url::Url;
+
+use crate::generated::rpc::api_client::ApiClient;
+
+pub type RpcClient = ApiClient<InterceptedService<Channel, AcceptHeaderInterceptor>>;
```
bobbinth, line 13: Should we change this into a proper newtype - i.e.,:
```rust
pub struct RpcClient(ApiClient<InterceptedService<Channel, AcceptHeaderInterceptor>>);
```
This way we could implement `connect()` like:
```rust
impl RpcClient {
    pub async fn connect(url: &Url, timeout_ms: u64) -> anyhow::Result<Self> {
        ...
    }
}
```

crates/proto/src/client.rs
```diff
@@ -0,0 +1,55 @@
...
+    let accept_value = format!("application/vnd.miden.{version}+grpc");
+    let interceptor = AcceptHeaderInterceptor::new(accept_value);
+
+    // Return the connected client.
+    Ok(ApiClient::with_interceptor(channel, interceptor))
+}
+
+/// Interceptor designed to inject the HTTP ACCEPT header
+/// into all [`ApiClient`] requests.
+pub struct AcceptHeaderInterceptor {
```
bobbinth, line 37: nit: comment wrapping.

crates/proto/src/lib.rs
```diff
@@ -7,6 +8,7 @@ pub mod generated;
 // RE-EXPORTS
 // ================================================================================================
 
+pub use client::{AcceptHeaderInterceptor, RpcClient, connect};
```
bobbinth, line 11: If we convert `RpcClient` into a newtype as described in one of the above comments, we could export just the `RpcClient` from here (unless we need to export `AcceptHeaderInterceptor` for some other reasons).

bobbinth (commented):

crates/rpc/src/server/accept.rs
sergerad: Yep, forgot to update the comment after updating impl

crates/rpc/src/server/accept.rs
sergerad: Sorry I don't follow what you mean here. The lines are all under 100 characters long

crates/rpc/src/server/accept.rs
sergerad: Have done with early returns instead. Think its a very readable approach

crates/rpc/src/server/accept.rs
sergerad: Its the idiomatic way of working with `nom` - all `nom` fns do the same.  The idea is that fns responsible for parsing return the remainder of the &str that was not consumed for the purposes of debugging / unit testing.

I have updated it as per your suggestion and removed the parse fn altogether. We have just lost access to the remainder of the parsed string for the sake of debugging. Shouldn't be an issue though.

crates/proto/src/client.rs
sergerad: I made this change and reworked the interceptor to be a bit more general so it makes sense for it to be used to inject other metadata if we ever need.

```rust
/// Alias for gRPC client that wraps the underlying client for the purposes of metadata configuration.
type InnerClient = ApiClient<InterceptedService<Channel, MetadataInterceptor>>;

/// Client for the Miden RPC API which is fully configured to communicate with a Miden node.
pub struct RpcClient(InnerClient);
```

crates/proto/src/lib.rs
sergerad: `AcceptHeaderInterceptor` still needs to be public as it is part of the public type `RpcClient`.

```
private type client::MetadataInterceptor in public interface
can’t leak private type (rustc E0446)
hint: `client::MetadataInterceptor` declared as private
```

crates/rpc/Cargo.toml
```diff
@@ -16,16 +16,22 @@ workspace = true
 
 [dependencies]
 anyhow           = { workspace = true }
+futures          = "0.3"
```
bobbinth, line 19: nit: I would write this as `{ version = "0.3" }`.

bobbinth (approved):
Looks good! Thank you! I left just one nit inline. I didn't review the tests in too much detail - so, another set of eyes on this PR would be appreciated.

sergerad:
I'm going to start working on the corresponding PR for miden-client today. We should merge them in close proximity to avoid `next` versions being incompatible (server requires accept header, client doesn't supply it).

sergerad:
I wasn't able to de-dupe any of this code because miden-node and miden-client use different versions of tonic.

This is probably for the best - if they used the same versions by chance, we could have ended up coupling them in an undesirable way here.

Note I couldn't even dedupe the `MetadataInterceptor` because the two tonic versions meant the `Service` trait was not fulfilled in miden-client when importing that type from miden-node.

sergerad:
@igamigo @tomyrd @TomasArrachea could I get 1 more review from any of you? 🙏

crates/rpc/src/client/interceptor.rs
```diff
@@ -0,0 +1,30 @@
...
+pub struct MetadataInterceptor {
+    metadata: HashMap<&'static str, AsciiMetadataValue>,
+}
+
+impl MetadataInterceptor {
+    /// Adds or overwrites HTTP ACCEPT metadata to the interceptor.
+    ///
+    /// Provided version string must be ASCII.
+    pub fn with_accept_metadata(mut self, version: &str) -> Result<Self, anyhow::Error> {
+        let accept_value = format!("application/vnd.miden.{version}+grpc");
```
igamigo, line 16: Not for this PR, but there should be no reason why we could not in the future generalize `{version}` to be the commit hash of the `miden-base` dependency to solve [this problem](https://github.com/0xMiden/miden-node/issues/804#issuecomment-2832401324:~:text=making%20it%20required.-,There%20is%20also%20a%20different%20but%20related%20problem%20that%20we%20can,wonder%20if%20we%20could%20do%20something%20for%20this%20case%20as%20well.,-igamigo), right?

crates/rpc/src/server/accept.rs
```diff
@@ -0,0 +1,213 @@
...
+                return bad_request("Invalid version specified in accept header value").boxed();
+            },
+        };
+
+        // Verify the version matches configured requirements
+        if self.version.matches(&version) {
+            self.inner.call(request).boxed()
+        } else {
+            debug!(target: COMPONENT, "Version does not match ({}/{})", version, self.version);
+            bad_request("Client / server version mismatch").boxed()
```
igamigo, line 113: Would we want to return the valid version here?

igamigo (approved):
LGTM! Left a couple of minor comments. I'm not sure how it's usually done for gRPC services, but I wonder if we could decide not to reject requests that do not contain the header (and if we should)

crates/rpc/src/client/interceptor.rs
sergerad: I think that is more of a prerequisite of our semver-based releases rather than something that would need to be enforced here. As in compatible node/client releases based on semver must have compatible base dependancies.

I may have misunderstood.

sergerad:
> LGTM! Left a couple of minor comments. I'm not sure how it's usually done for gRPC services, but I wonder if we could decide not to reject requests that do not contain the header (and if we should)

I have updated it to accept requests that don't have the header. This is a good idea and necessary while the wasm client does not provide the header.

sergerad:
> I still need to finalize [0xMiden/miden-client#912](https://github.com/0xMiden/miden-client/pull/912) before this is merged.

Can merge this now that requests without the header are accepted.

PR #850, opened 2025-05-16
chore: Fix lint errors since cargo update

sergerad:


bobbinth:
What's driving this? For example, is this because `tonic::Status` is a big struct? If so, should we put it in a `Box`?

sergerad:
> What's driving this? For example, is this because `tonic::Status` is a big struct? If so, should we put it in a `Box`?

Its a series of errors. I would be surprised to see an actual performance benefit from boxing these. Unsure which poison I would pick out of boxing or using the allow macro here. Could box the whole error or box the relevant enum variants.

Example size: `the largest variant contains at least 176 bytes`.

Errors implicated:

```
the Err-variant returned from this function is very large
try reducing the size of errors::ClientError, for example by boxing large elements or replacing it with Box<errors::ClientError>
```

```
the Err-variant returned from this function is very large
try reducing the size of errors::HandlerError, for example by boxing large elements or replacing it with Box<errors::HandlerError>
```

```
the Err-variant returned from this function is very large
try reducing the size of errors::VerifyTxError, for example by boxing large elements or replacing it with Box<errors::VerifyTxError>
```

```
the Err-variant returned from this function is very large
try reducing the size of errors::AddTransactionError, for example by boxing large elements or replacing it with Box<errors::AddTransactionError>
```

igamigo:
> #845 introduced this problem. I'm not sure we want to "fix" it this way. 

I think the CI didn't catch this on the PR because it used Rust `1.86` and on merging it, [it started using `1.87`](https://github.com/0xMiden/miden-node/actions/runs/15052233630/job/42309623428).

Because of the functions that were changed seem to be unlikely to fail often, I'd box the errors here. Otherwise the result enums are carrying those extra bytes unnecessarily which are copied around and this can increase cache pollution, etc. The overhead of that single extra allocation once one of these functions fail is probably lower, but I may not have the full picture (and I agree it's probably unlikely we can detect any sensible performance difference).

sergerad:
"Feels" more idiomatic to have boxed enum variants than boxed Errors. Unsure what you guys think about that. The former would mean we don't have future Errors to box also.

TomasArrachea:
FYI I also made a draft https://github.com/0xMiden/miden-node/pull/851 boxing the errors, but not sure if it's the right way to go.

igamigo:
> "Feels" more idiomatic to have boxed enum variants than boxed Errors. Unsure what you guys think about that. 

Agreed.

I think I'd merge this as-is for now to unblock the CI being broken and add an issue to come back and remove the `#[allow()]`s. @bobbinth what do you think?

SantiagoPittella (approved):
LGTM for a temporary fix, as Tomas and Ignacio mentioned before, a better-more stable fix is being done by Tomas.

PR #851, opened 2025-05-16
chore: fix clippy new warnings

TomasArrachea:
Closes https://github.com/0xMiden/miden-node/issues/855.

This PR removes the clippy allows that were added as a hotfix for the new clippy warnings. 
For addressing them, this PR boxes the `tonic::Status` struct when used in error variants and changes some error handling to account for it.

### DIFF
--- a/bin/faucet/src/rpc_client.rs
+++ b/bin/faucet/src/rpc_client.rs
@@ -18,7 +18,7 @@ use crate::faucet::FaucetId;
 #[derive(Debug, thiserror::Error)]
 pub enum RpcError {
     #[error("gRPC error encountered")]
-    Transport(#[source] tonic::Status),
+    Transport(#[source] Box<tonic::Status>),
     #[error("error parsing the gRPC response")]
     ResponseParsing(#[source] anyhow::Error),
 }
@@ -46,7 +46,7 @@ impl RpcClient {
             .inner
             .get_block_header_by_number(request)
             .await
-            .map_err(RpcError::Transport)?;
+            .map_err(|e| RpcError::Transport(Box::new(e)))?;
 
         let root_block_header = response
             .into_inner()
@@ -70,7 +70,7 @@ impl RpcClient {
             .inner
             .get_account_details(request)
             .await
-            .map_err(RpcError::Transport)?
+            .map_err(|e| RpcError::Transport(Box::new(e)))?
             .into_inner()
             .details
             .context("details field is missing")
@@ -97,6 +97,6 @@ impl RpcClient {
             .submit_proven_transaction(request)
             .await
             .map(|response| response.into_inner().block_height.into())
-            .map_err(RpcError::Transport)
+            .map_err(|e| RpcError::Transport(Box::new(e)))
     }
 }
--- a/bin/faucet/src/rpc_client.rs
+++ b/bin/faucet/src/rpc_client.rs
@@ -46,7 +46,7 @@ impl RpcClient {
             .inner
             .get_block_header_by_number(request)
             .await
-            .map_err(|e| RpcError::Transport(Box::new(e)))?;
+            .map_err(|e| RpcError::Transport(e.into()))?;
 
         let root_block_header = response
             .into_inner()
@@ -70,7 +70,7 @@ impl RpcClient {
             .inner
             .get_account_details(request)
             .await
-            .map_err(|e| RpcError::Transport(Box::new(e)))?
+            .map_err(|e| RpcError::Transport(e.into()))?
             .into_inner()
             .details
             .context("details field is missing")
@@ -97,6 +97,6 @@ impl RpcClient {
             .submit_proven_transaction(request)
             .await
             .map(|response| response.into_inner().block_height.into())
-            .map_err(|e| RpcError::Transport(Box::new(e)))
+            .map_err(|e| RpcError::Transport(e.into()))
     }
 }
--- a/crates/block-producer/src/errors.rs
+++ b/crates/block-producer/src/errors.rs
@@ -191,6 +191,6 @@ pub enum StoreError {
 
 impl From<tonic::Status> for StoreError {
     fn from(value: tonic::Status) -> Self {
-        StoreError::GrpcClientError(Box::new(value))
+        StoreError::GrpcClientError(value.into())
     }
 }
--- a/crates/block-producer/src/errors.rs
+++ b/crates/block-producer/src/errors.rs
@@ -182,9 +182,15 @@ pub enum BuildBlockError {
 #[derive(Debug, Error)]
 pub enum StoreError {
     #[error("gRPC client error")]
-    GrpcClientError(#[from] tonic::Status),
+    GrpcClientError(Box<tonic::Status>),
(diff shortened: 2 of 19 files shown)


bin/faucet/src/handlers.rs
```diff
@@ -117,23 +117,23 @@ pub async fn get_tokens(
 }
 
 pub async fn get_index_html(state: State<FaucetState>) -> Result<impl IntoResponse, HandlerError> {
-    get_static_file(state, "index.html")
+    get_static_file(state, "index.html").map_err(|err| *err)
```
bobbinth, line 120: Quick question: this seems a bit less ergonomic than what we had before - what is driving the need for this change?

TomasArrachea: I just realized in this particular case it could be avoided implementing IntoResponse. There is some cases like this though, they stem from boxing the errors in the internal functions, as suggested by clippy.

bin/faucet/src/client.rs
```diff
@@ -272,7 +273,7 @@ fn build_transaction_arguments(
     output_note: &Note,
     note_type: NoteType,
     asset: FungibleAsset,
-) -> Result<TransactionArgs, ClientError> {
+) -> Result<TransactionArgs, Box<ClientError>> {
```
bobbinth, line 276: Why do we need to box the error here and in other places? I'm assuming returning just `ClientError` trips up some clippy warnings. If so, shouldn't we fix `ClientError` instead?

TomasArrachea: Yes, thats a valid point. Here i am boxing the entire error but we could just instead box the enum variants that are causing problems.

TomasArrachea: Modified this in [00be4f1](https://github.com/0xMiden/miden-node/pull/851/commits/00be4f165f9b97ae1c131ed416c28d8712376266) to only box the `tonic::Status` on variants that was causing the problem.

crates/block-producer/src/errors.rs
```diff
@@ -182,9 +182,15 @@ pub enum BuildBlockError {
 #[derive(Debug, Error)]
 pub enum StoreError {
     #[error("gRPC client error")]
-    GrpcClientError(#[from] tonic::Status),
+    GrpcClientError(Box<tonic::Status>),
```
sergerad, line 185: Any reason to remove the `#[from]` here?

bin/faucet/src/rpc_client.rs
```diff
@@ -52,7 +52,7 @@ impl RpcClient {
             .inner
             .get_block_header_by_number(request)
             .await
-            .map_err(RpcError::Transport)?;
+            .map_err(|e| RpcError::Transport(Box::new(e)))?;
```
sergerad, line 55: nit: `e.into()` a bit more concise. Unsure what people prefer.

crates/block-producer/src/errors.rs
```diff
@@ -182,9 +182,15 @@ pub enum BuildBlockError {
...
     MalformedResponse(String),
     #[error("failed to parse response")]
     DeserializationError(#[from] ConversionError),
 }
+
+impl From<tonic::Status> for StoreError {
+    fn from(value: tonic::Status) -> Self {
+        StoreError::GrpcClientError(Box::new(value))
+    }
+}
```
sergerad, line 196:
```rust
    GrpcClientError(#[from] Box<tonic::Status>),
```
Compiles for me. Wondering why you needed this?

crates/store/src/server/api.rs
```diff
@@ -406,7 +406,7 @@ impl api_server::Api for StoreApi {
 
         debug!(target: COMPONENT, ?request);
 
-        let account_id = read_account_id(request.account_id)?;
+        let account_id = read_account_id(request.account_id).map_err(|err| *err)?;
```
sergerad, line 409: Why does one fn require `Box<Status>`
```rust
fn read_account_id(id: Option<generated::account::AccountId>) -> Result<AccountId, Box<Status>> {
```

while this one doesn't?

```rust
    async fn get_account_state_delta(
        &self,
        request: Request<GetAccountStateDeltaRequest>,
    ) -> Result<Response<GetAccountStateDeltaResponse>, Status> {

```

crates/store/src/server/api.rs
```diff
@@ -582,15 +582,13 @@ fn invalid_argument<E: core::fmt::Display>(err: E) -> Status {
     Status::invalid_argument(err.to_string())
 }
 
-#[allow(clippy::result_large_err)]
-fn read_account_id(id: Option<generated::account::AccountId>) -> Result<AccountId, Status> {
+fn read_account_id(id: Option<generated::account::AccountId>) -> Result<AccountId, Box<Status>> {
     id.ok_or(invalid_argument("missing account ID"))?
         .try_into()
-        .map_err(|err| invalid_argument(format!("invalid account ID: {err}")))
+        .map_err(|err| Box::new(invalid_argument(format!("invalid account ID: {err}"))))
```
sergerad, line 588: nit: (again) `.into()` instead of `Box::new()` if we prefer.

crates/utils/src/config.rs
```diff
@@ -17,7 +17,7 @@ pub const DEFAULT_FAUCET_SERVER_PORT: u16 = 8080;
 /// relative, searches in parent directories all the way to the root as well.
 ///
 /// The above configuration options are indented to support easy of packaging and deployment.
-#[allow(clippy::result_large_err)]
+#[allow(clippy::result_large_err, reason = "This error crashes the node")]
```
sergerad, line 20: Is the reason not that we don't implement the error type ourselves here? As in its from `figment`?

crates/block-producer/src/errors.rs
TomasArrachea: This actually was not working for me. I get this compilation error: "the trait bound `StoreError: From<Status>` is not satisfied"

crates/utils/src/config.rs
TomasArrachea: Yes, also because of that. I was thinking of the reason not to box the error and allow it being large

crates/store/src/server/api.rs
TomasArrachea: I am not sure why clippy does not complain, but I would guess it's because of the function being async, so the returned result is actually being boxed under the hood.

crates/block-producer/src/errors.rs
sergerad: Yea sorry I just checked again and compile error for sure. All good.

sergerad (approved):

bobbinth:
We should probably merge the latest `next` into this branch to reduce the diff size.

TomasArrachea:
Seems like `taplo` was updated so CI is failing on `next` for the new format requirements, I included here the new formatting changes

igamigo (approved):

tomyrd (approved):
LGTM, strange that `taplo` puts the `version` field at the end.

bobbinth:
> LGTM, strange that `taplo` puts the `version` field at the end.

Is there a way to disable this specific lint (e.g., via a config file)?

PR #852, opened 2025-05-16
fix(temporary): enums too large

SantiagoPittella:
Currently, all the PRs are failing in the lint step because of the new rust version. In next is being fixed with https://github.com/0xMiden/miden-node/pull/850 , and looks like it is going to be a more definitive solution, boxing large variants. I created this PR with lint `allow`s so we can unlock the faucet PRs that use main as base.


igamigo (approved):
LGTM. This should get changed later as part of #855 (and as mentioned in #850)

PR #857, opened 2025-05-16
chore: dont remove db files on deploy

### LINKED ISSUE #670: [devops] Deploy without nuking db
The deploy workflow currently nukes the database. We should allow deploys without doing this.

This could be an additional enum on the deploy inputs, or we could try be smarter about this by checking db versions.

We should double-check that the package installer doesn't automatically destroy the db.

TomasArrachea:
Closes #670.

This PR aims to allow for migrating the current database and dont delete it on deploy. Three main things were done:
- Update post installation scripts to avoid deleting the db while running the deploy github action.
- Migrate db schemas. This involved resetting the initial db schemas as they are on v0.8 and move all changes made on v0.9 as a separate sql migration. 
- Migrate the information from the db. Some rust code was needed, since some information needs to be deserialized, separated and then stored in the new db format. This was added on the same `apply_migration` function, so it runs automatically when loading the db. As part of this changes, this PR includes the account root migration done in https://github.com/0xMiden/miden-node/pull/882.


.github/workflows/deploy.yml
```diff
@@ -16,6 +16,11 @@ on:
         required: true
         type: string
 
+      reset-db:
+        description: 'Reset database and genesis account'
+        required: false
+        type: boolean
```
sergerad, line 22: I would prefer if our automation was incapable of doing this. It is OK to require manual intervention to perform destructive actions instead.

sergerad:
We might want to delete these lines from the node/postrm script too
```
sudo rm -rf /etc/opt/miden-node
sudo rm -rf /opt/miden-node
```

.github/workflows/deploy.yml
TomasArrachea: Makes sense, I'll make it so that the database is always preserved

TomasArrachea:
> We might want to delete these lines from the node/postrm script too
> 
> sudo rm -rf /etc/opt/miden-node
> sudo rm -rf /opt/miden-node

Yes, this was changed for both the node and faucet postrm scripts

.github/workflows/deploy.yml
```diff
@@ -147,7 +148,11 @@ jobs:
         with:
           instance_id: ${{ secrets[env.instance-id] }}
           command: |
-            sudo /usr/bin/miden-node bundled bootstrap --data-directory /opt/miden-node --accounts-directory /opt/miden-faucet; \
+            if [ ! -f /opt/miden-node/miden-store.sqlite3 ]; then \
+              sudo rm -rf /opt/miden-faucet/*; \
+              sudo rm -rf /opt/miden-node/*; \
```
sergerad, line 153: This still makes me nervous having the wildcard there. Could we instead specify the exact files to delete? There are probably only a handful right?

TomasArrachea: Yes, you are right.

sergerad:
@TomasArrachea I'm wondering if #670 implicitly requires that we support schema migrations. Maybe we should make a separate issue and let this PR close that one.

bin/node/src/system_monitor.rs
TomasArrachea, line 1: this is unrelated changes, just realized while debugging that context here was missing

TomasArrachea:
I added in this PR the schema migrations from 0.8 to 0.9

crates/store/src/db/migrations/001-init.sql
SantiagoPittella, line 1: If we are going to keep DBs without nuking, we should not change previous migrations (the changes would not be reflected on the DB).

TomasArrachea: Yes! here I just pasted the schema definitions as of v0.8. The changes introduced since then were moved to the migration 002

crates/store/src/db/migrations/002-release-0.9.sql
```diff
@@ -0,0 +1,65 @@
...
+INSERT INTO notes_new (
+    block_num, batch_index, note_index, note_id, note_type, sender, tag,
+    execution_mode, aux, execution_hint, merkle_path, consumed, nullifier,
+    assets, inputs, script_root, serial_num
+)
+SELECT 
+    block_num, batch_index, note_index, note_id, note_type, sender, tag,
+    execution_mode, aux, execution_hint, merkle_path, consumed, nullifier,
+    NULL, NULL, NULL, NULL
+FROM notes;
```
bobbinth, line 54: This would copy some of the data, but keep the newly expanded fields (e.g., `nullifier`) as `NULL`, right? If so, we should also come up with a way to expand the previous `details` field into the new fields.

Similarly, we should populate the `note_scripts` table with the scripts from all notes in the database.

igamigo: @TomasArrachea  I think for this we'd need to go out of SQL and do something on the Rust side, like in #882. Maybe we can merge that here and work on a migration command.

- From this SQL script I'd keep basically everything, except the `DROP TABLE notes;` and the `ALTER TABLE notes_new RENAME TO notes`).
- Additionally, to `notes_new`, you could add a temporary `details` column and migrate the values from the previous table of notes.
- On the same account tree root migration function that you are working on, we'd need to add:
  - Run a SQL query to look up any records in `notes_new` that have `NULL` on the new columns but not on `details`.
    - For every record, deserialize `details` and do an `UPDATE` to fill in the new columns. 
    - Also for every record, update the `note_scripts` table
- Run something like:
 ```sql
  DROP TABLE notes; 
  ALTER TABLE notes_new RENAME TO notes;
  ALTER TABLE notes DROP COLUMN details;
 ```

crates/store/src/db/migration.rs
bobbinth, line 1: I would move this file under `src/db/migrations/mod.rs` and rename the current `mod.rs` file to be more specific to account root migration.

bin/stress-test/src/seeding/mod.rs
```diff
@@ -66,7 +66,6 @@ pub async fn seed_store(
     public_accounts_percentage: u8,
 ) {
     let start = Instant::now();
-
```
bobbinth, line 69: nit: we can probably revert this change.

crates/store/src/db/migration.rs
```diff
@@ -92,6 +98,9 @@ pub fn apply_migrations(conn: &mut Connection) -> super::Result<()> {
     debug!(target: COMPONENT, new_schema_version, "Updating schema version in settings table");
     Settings::set_value(conn, DB_SCHEMA_VERSION_FIELD, &new_schema_version)?;
 
+    if env!("CARGO_PKG_VERSION") == "0.9.0" {
+        migrate_account_root(conn).expect("account root migration should succeed");
+    }
```
bobbinth, line 103: This may be OK for now, but we probably should come up with a more "surgical" way to run custom code after some SQL migration has been applied.

The issue is that technically this will apply all migrations and then run `migrate_account_root()`. The reason it works now is that the last migration and v0.9.0 align as expected.

crates/store/src/db/migrations/002-release-0.9.sql
TomasArrachea: Implemented this! Had to make a small tweak: instead of creating a temporary `details` column I loaded the serialized details into the `assets` column of the table, and then on the rust code load that column to deserialize the details and load it on the new columns. This was a workaround because the rust code couldn't access the new table, I suspect due to the sql migration transaction was not yet commited.

crates/store/src/db/migration.rs
bobbinth: Yes - this works!

crates/store/src/db/migration.rs
bobbinth: I think we can track it as a part of https://github.com/0xMiden/miden-node/issues/886.

TomasArrachea: Agree, in any case this rust migration is idempotent so shouldn't be a problem for now

crates/store/src/db/migrations/002-release-0.9.sql
igamigo: So the `EOF` error I was hitting was correct. We did some more tests and @TomasArrachea found that the same note serialized by `0.9` code has more data in it compared to the same note serialized by `0.8` ([link](https://gist.github.com/igamigo/4ff1ae9ef90f958798a7c4bec62bb551)). We are testing some more, but it's looking unlikely that it's worth continuing in this path.

I also briefly looked at the code and could not find where the difference could be. My suspicion was initially the account ID, but considering that rewriting the account tree root worked well it probably cannot be that. It's hard to pinpoint from the code alone because it could be in multiple places, might try bisecting the history and see if we find it.

We could not migrate old notes correctly and effectively deprecate them, which probably might not be too bad but definitely undesirable. Or not do the data-layer migration at all, only migrate the schema and start from a new genesis. Additionally I wonder if we want/it would be worth to have a test that tries deserializing a current-version note (and any other important struct) in `next` to monitor if it breaks.

igamigo: > a binary that depends both on v0.8 and v0.9, and use v0.8 structs to deserialize both accounts and notes, and the re-serialize them using v0.9 structs

Is this doable? A `NoteScript` from the 0.8 crate is effectively different from a `NoteScript` from the 0.9 crate so it won't compile but maybe I'm thinking about it incorrectly.

igamigo: > hmmm - should be doable, but may be too complicated. We'd need to write custom functions that take MastForestV08 and returns MastForestV09. This would probably require writing converters for all the types involved.

Ah yes, that's what I was thinking of basically. We'd need to have a copy of both versions. There are a lot of structs involved as well. We kind of tried this by going back to `miden-vm` and copying the deserialization function from `0.13`. This involved adding some missing functions, which means that semantically the data changed a bit more than we had thought (though not by much, [here is the diff](https://github.com/lambdaclass/miden-vm/compare/next...lambdaclass:miden-vm:tomasarrachea-notes-migration#diff-5f12e6c7da3ec6aabfdb3738c20ffe37473b8da7b7749c9b9ad1f89a2eea1e2eR190)). However, deserializing this still fails so I'm guessing there's something more that changed internally (maybe from other dependencies?). Magic bytes check passes just fine, but it fails on the `try_into_mast_node()` here:

```rust
            let basic_block_data_decoder = BasicBlockDataDecoder::new(&basic_block_data);
            for mast_node_info in mast_node_infos {
                let node =
                    mast_node_info.try_into_mast_node(node_count, &basic_block_data_decoder)?;

                mast_forest.add_node(node).map_err(|e| {
                    DeserializationError::InvalidValue(format!(
                        "failed to add node to MAST forest while deserializing: {e}",
                    ))
                })?;
            }
```

So I'm sort of ready to give up on this. I still believe there could be a third alternative to what you mentioned which is to migrate notes as private notes basically, but there might be other unintended side effects (and the account code still needs a solution).

bobbinth: Yeah, maybe we pushed it as far as we can for now.

> So I'm sort of ready to give up on this. I still believe there could be a third alternative to what you mentioned which is to migrate notes as private notes basically, but there might be other unintended side effects (and the account code still needs a solution).

But wouldn't accounts also break? And I guess if we go this way, what's the proportion between public and private notes? (if 90% are public, I think dropping them is basically the same as dropping the full DB).

crates/store/src/db/migration.rs
Mirko-von-Leipzig: I've always done the migrations as pure rust code. Each migration is a function, and applying them is as simple as running a loop over `current..last` migration versions.

igamigo:
I've pushed a couple of changes that make this work. It imports a forked `miden-core` which contains a modified `0.14` with added `0.13` deserialization (this is quite hacky and improvised but also it was the easiest way to iterate). I haven't had time to make this its own binary. 
Would be great to have a testnet backup to test this on, since for now I've only done it with a DB seeded from `0.8` client integration tests. @TomasArrachea could you review/test it out on your local environment as well?

TomasArrachea:
@igamigo I added a small fix but now it is effectively working when migrating my local db!

crates/store/src/db/migrations/mod.rs
```diff
@@ -92,6 +96,9 @@ pub fn apply_migrations(conn: &mut Connection) -> super::Result<()> {
     debug!(target: COMPONENT, new_schema_version, "Updating schema version in settings table");
     Settings::set_value(conn, DB_SCHEMA_VERSION_FIELD, &new_schema_version)?;
 
+    if env!("CARGO_PKG_VERSION") == "0.9.0" {
+        migrate_release_0_9(conn).expect("release 0.9 migration should succeed");
+    }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Over time you can deprecate old migrations, by rolling them into a `base` version. This _must_ be done at any point where migrations are no longer backwards compatible - e.g. maybe once a hard fork has occurred on all networks.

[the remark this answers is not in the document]
Mirko-von-Leipzig: Having personally blown up a few databases with a quick-hack tool I'd advocate for just _not_ doing it unless we think its simple :grimacing:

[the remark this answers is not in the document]
igamigo: Yes, I definitely feel that. It's probably not very robust, but I think it should work. And I figured if  all else fails we can just nuke it and start over

sergerad:
Do we need to close this PR now? I have update the debian packages to not wipe the db files elsewhere.

bobbinth:
Yes, I'll close it.

PR #859, opened 2025-05-19
feat: add challanges state

SantiagoPittella:


bin/faucet/src/server/get_tokens.rs
```diff
@@ -134,6 +136,7 @@ impl RawMintRequest {
         self,
         options: &AssetOptions,
         pow_salt: &str,
+        challenge_state: &ChallengeState,
```
TomasArrachea, line 139: we should update the docs of this function, it returns error also if the challenge is expired/invalid or server signature is invalid

bin/faucet/src/server/pow.rs
```diff
@@ -1,16 +1,101 @@
...
 const SERVER_TIMESTAMP_TOLERANCE_SECONDS: u64 = 30;
 
+#[derive(Clone)]
+pub struct Challenge {
+    timestamp: u64,
+    server_signature: String,
+}
+
+#[derive(Clone)]
+pub struct ChallengeState {
```
igamigo, line 26: I'd name this `ChallengeCache` or something similar.

bin/faucet/src/server/mod.rs
```diff
@@ -66,7 +67,20 @@ impl Server {
         // SAFETY: Leaking is okay because we want it to live as long as the application.
         let metadata = Box::leak(Box::new(metadata));
 
-        Server { mint_state, metadata, pow_salt }
+        let challenge_state = pow::ChallengeState::new();
+
+        // Start the cleanup task
+        let cleanup_state = challenge_state.clone();
+        tokio::spawn(async move {
+            pow::start_cleanup_task(cleanup_state).await;
```
TomasArrachea, line 75: i would rename `start_cleanup_task` -> `run_cleanup` or either move the task dispach inside the function

bin/faucet/src/server/pow.rs
```diff
@@ -1,16 +1,101 @@
...
 use serde::Serialize;
 use sha3::{Digest, Sha3_256};
+use tokio::time::{Duration, interval};
 
 use super::{Server, get_tokens::InvalidRequest};
 
 const DIFFICULTY: u64 = 5;
 const SERVER_TIMESTAMP_TOLERANCE_SECONDS: u64 = 30;
 
+#[derive(Clone)]
```
TomasArrachea, line 19: should be add comments explaining what is the challenge and how it prevents the replay?

bin/faucet/src/server/get_tokens.rs
```diff
@@ -164,7 +167,12 @@ impl RawMintRequest {
         )?;
 
         // Check the PoW solution
-        check_pow_solution(&self.pow_seed, self.pow_solution)?;
+        check_pow_solution(
+            challenge_state,
+            &self.pow_seed,
+            &self.server_signature,
+            self.pow_solution,
+        )?;
```
igamigo, line 175: nit: Because this takes various arguments from `self`, would it make sense to make it a member function of `RawMintRequest`?

TomasArrachea (commented):
Nice! The idea is that eventually this solution would replace the rate limiter? since it's not addressed here, should we create an issue for that?

bin/faucet/src/server/get_tokens.rs
igamigo: Also, should checking that the challenge exists in the cache come before the other checks? So that if it does not exist we can exit early (could be done here or even before calling `validate()`).

bin/faucet/src/server/pow.rs
```diff
@@ -1,16 +1,101 @@
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
```
TomasArrachea, line 3: I think we should be using `tokio::sync::Mutex` since this run in an async context

igamigo (approved):
LGTM. I left a couple of comments but feel free to disregard since they are mostly related to code structure. I think also that this implementation is missing the IP address inclusion, but maybe it can be tackled separately.

bin/faucet/src/server/pow.rs
```diff
@@ -1,16 +1,101 @@
...
+    }
+
+    pub fn validate_challenge(
+        &self,
+        seed: &str,
+        server_signature: &str,
+    ) -> Result<(), InvalidRequest> {
+        let challenges = self.challenges.lock().unwrap();
+
+        if let Some(challenge) = challenges.get(seed) {
```
sergerad, line 49: nit: would appreciate explainer comments in this fn

bin/faucet/src/server/pow.rs
```diff
@@ -1,16 +1,101 @@
...
+    server_signature: String,
+}
+
+#[derive(Clone)]
+pub struct ChallengeState {
+    challenges: Arc<Mutex<HashMap<String, Challenge>>>,
+}
+
+impl ChallengeState {
+    pub fn new() -> Self {
```
sergerad, line 31: This could just be provided by `#[derive(Default)]` instead. Unless you plan to modify it.

bin/faucet/src/server/pow.rs
```diff
@@ -1,16 +1,101 @@
...
+        Self {
+            challenges: Arc::new(Mutex::new(HashMap::new())),
+        }
+    }
+
+    pub fn add_challenge(&self, seed: &str, server_signature: String, timestamp: u64) {
```
sergerad, line 37: nit: If we call the struct a cache then maybe would just call this fn `put()`.

bin/faucet/src/server/pow.rs
```diff
@@ -100,6 +201,10 @@ pub(crate) fn check_pow_solution(seed: &str, solution: u64) -> Result<(), Invali
         return Err(InvalidRequest::InvalidPoW);
     }
 
+    // If we get here, the solution is valid
+    // Remove the challenge to prevent reuse
+    challenge_state.remove_challenge(seed);
```
sergerad, line 206: This confuses me as I had assumed this struct was holding an index of created (and/or completed) challenges. So I don't follow why it would be removed upon success?

bin/faucet/src/server/pow.rs
```diff
@@ -1,16 +1,101 @@
...
+        });
+    }
+
+    pub fn remove_challenge(&self, seed: &str) {
+        let mut challenges = self.challenges.lock().unwrap();
+        challenges.remove(seed);
+    }
+}
+
+pub async fn start_cleanup_task(challenge_state: ChallengeState) {
```
sergerad, line 90: Do we need timestamp-based cleanup at all? Couldn't we do the LRU cache approach that @igamigo was doing in another PR? That way we don't need a cleanup task either? We just drop the oldest challenge when we are at capacity and adding a new one?

Then again that would create an opportunity for attack if users could create enough challenges to reach our limit continuously. Maybe easy to prevent / control but still.

Just wondering. No change required if you prefer time based.

bin/faucet/src/server/pow.rs
SantiagoPittella: i don't think so, I'm not passing the lock between async functions, but just doing hashmaps ops. Though if you see any other reason to use tokio's mutex let me know.

bin/faucet/src/server/pow.rs
SantiagoPittella: I added this removal for the challenge to be invalid from this point, without needed to keep any other field or state. Since it would no longer be in the challenge cache, any new request using the challenge would be rejected.

bin/faucet/src/server/pow.rs
SantiagoPittella: As you mentioned, using an LRU cache allows attackers to keep it full and would cause that valid challenges would be rejected.

SantiagoPittella:
@sergerad @igamigo i added a `PoWParameters` struct to abstract the validation and stop passing all the parameters to every function. Let me know what do you think.

bin/faucet/src/server/pow.rs
sergerad: Ah OK this is a different model to what I had assumed. I was thinking of keeping a map of used challenges to prevent replay.

What is stopping someone from replaying a challenge within its time range?

sergerad: As in, is it possible to have multiple concurrent runs of `pub fn check_pow_solution(` for the same challenge to succeed?

bin/faucet/src/server/pow.rs
sergerad: Yes but we definitely should not use tokio mutex here as it is not used in async fns at all. Unless I have gotten confused at the relevant code

bin/faucet/src/server/pow.rs
SantiagoPittella: you're right, there is a small gap between checking the challenge validity and removing it from the cache that allows for challenges to be be duplicated. I'm moving the lock.

bin/faucet/src/server/pow.rs
sergerad: This also goes back to the comment above where I realized the impl was different to my assumptions 👍

bin/faucet/src/server/pow.rs
TomasArrachea: Seems I was confused about when to use blocking mutex, all good!

sergerad (approved):

PR #864, opened 2025-05-21
feat: add BlockHeader from MockChain

### LINKED ISSUE #862: Improve faucet's mocked data
Right now, the faucet uses a mocked [block header](https://github.com/0xMiden/miden-node/blob/next/bin/faucet/src/stub_rpc_api.rs#L48) for the minting test. If the underlying transaction dependencies get updated, the hashes consistencies may get stale and the hardcoded block has to be updated.

We could now use `miden-testing`'s `MockChain` to programmatically generate a valid and consistent block instead of having a hardcoded one instead.

varun-doshi:
Attempt to fix #862


bin/faucet/Cargo.toml
```diff
@@ -47,6 +47,7 @@ miden-node-utils = { workspace = true, features = ["vergen"] }
 static-files     = "0.2"
 
 [dev-dependencies]
-fantoccini   = { version = "0.21" }
-serde_json   = { version = "1.0" }
-tokio-stream = { workspace = true, features = ["net"] }
+fantoccini    = { version = "0.21" }
+serde_json    = { version = "1.0" }
+tokio-stream  = { workspace = true, features = ["net"] }
+miden-testing = { git = "https://github.com/0xMiden/miden-base.git", branch = "next" }
```
igamigo, line 53: This should now ideally reference [the `0.9` version of this crate](https://github.com/0xMiden/miden-base/releases)

varun-doshi:
> LGTM! Thanks, if you could resolve the merge conflicts then we are good to go. I imagine merging `next` into your branch would be the easiest, and we squash commit so its not an issue.

Done!

Looks CI is running now?

varun-doshi:
CI fixed

igamigo (approved):

bobbinth:
> I'm actually unsure why we aren't getting CI workflow.. probably some security setting we haven't toggled for forks - @bobbinth could you [take a look](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks#about-workflow-runs-from-public-forks), I don't have the permissions to access settings?

Seems like CI ran here - or am I missing something?

Mirko-von-Leipzig:
> > I'm actually unsure why we aren't getting CI workflow.. probably some security setting we haven't toggled for forks - @bobbinth could you [take a look](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks#about-workflow-runs-from-public-forks), I don't have the permissions to access settings?
> 
> Seems like CI ran here - or am I missing something?

It became available to run after the merge conflicts were resolved, so seems related to that somehow.

PR #865, opened 2025-05-22
feat: adjustable pow difficulty

### LINKED ISSUE #854: faucet(PoW): adjust difficulty
In the first iteration ( https://github.com/0xMiden/miden-node/pull/831 ), the PoW difficulty was hard coded to 5 after some tests. A better approach would be to adjust at runtime the difficulty, to be proportional to the usage  of the faucet or some other parameter.

SantiagoPittella:
closes #854


bin/faucet/src/server/pow.rs
```diff
@@ -13,9 +13,23 @@ use tokio::time::{Duration, interval};
...
 use super::{Server, get_tokens::InvalidRequest};
 
-const DIFFICULTY: u64 = 5;
+const MAX_DIFFICULTY: u64 = 6;
 const SERVER_TIMESTAMP_TOLERANCE_SECONDS: u64 = 30;
 
+#[derive(Clone)]
+pub struct PoW {
+    pub(crate) salt: String,
+    pub(crate) difficulty: Arc<Mutex<u64>>,
```
TomasArrachea, line 22: Not sure if it's worth it but we could try [AtomicU64](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html) here. Otherwise i would use tokio::sync::Mutex instead of std::sync::Mutex

TomasArrachea (commented):
LGTM!

TomasArrachea (approved):

bin/faucet/src/server/pow.rs
```diff
@@ -13,9 +13,23 @@ use tokio::time::{Duration, interval};
...
+    pub(crate) challenge_state: ChallengeState,
+}
+
+impl PoW {
+    pub fn adjust_difficulty(&mut self, sender_count: usize) {
```
TomasArrachea, line 27: we could add a doc comment on how this difficulty is adjusted to the faucet load

bin/faucet/src/server/pow.rs
```diff
@@ -16,10 +19,10 @@ use super::{
     get_tokens::{InvalidRequest, RawMintRequest},
 };
 
-/// The difficulty of the `PoW`.
+/// The maximum difficulty of the `PoW`.
 ///
 /// The difficulty is the number of leading zeros in the hash of the seed and the solution.
-const DIFFICULTY: u64 = 5;
+const MAX_DIFFICULTY: u64 = 6;
```
bobbinth, line 25: Max difficulty of 6 leading zeros seems pretty low (i.e., it would require, on average, 64 guesses to get the right nonce). A moderately-powerful machine should be able to guess 16 leading zeros in under a second, and 20 leading zeros within a few seconds). I'd probably set the maximum at something like 24.

Also, as mentioned in another comment, ideally, PoW should be based not on the number of leading zeros but on absolute values. This way, it can be adjusted more granularly - but we can do this in a separate PR.

bin/faucet/src/server/pow.rs
```diff
@@ -122,11 +131,37 @@ impl TryFrom<&RawMintRequest> for PowParameters {
...
-// CHALLENGE STATE
+// POW
+// ================================================================================================
+
+#[derive(Clone)]
+pub struct PoW {
+    pub(crate) salt: String,
+    pub(crate) difficulty: Arc<AtomicU64>,
+    pub(crate) challenge_cache: ChallengeCache,
+}
```
bobbinth, line 150: Not for this PR, but I'd encapsulate functionality a bit more around `PoW` and `Challenge` structs. Maybe something like this:

```rust
pub struct PoW {
    salt: [u8; 32],
    difficulty: Arc<AtomicU64>,
    challenge_cache: ChallengeCache,
}

impl Pow {

    /// Generates a new challenge.
    pub fn build_challenge(&self, request_ip: IpAddr) -> Challenge {
        ...
    }

    /// Verifies that the challenge is valid (e.g., timestamp is within range, difficulty is
    /// satisfied etc.), and adds the validate challenge to the cache.
    pub fn validate_challenge(&mut self, challenge: Challenge, nonce: u64) -> Result<(), SomeError> {
        ...
    }

    pub fn adjust_difficulty(&self, active_requests: usize) {
        ...
    }
}

pub struct Challenge {
    difficulty: u64,
    ip_address: IpAddr,
    timestamp: u64,
    signature: [u8; 32],
}

impl Challenge {
    /// Decodes the challenge and verifies that the signature part of the challenge is valid
    /// in the context of the specified salt.
    pub fn decode(value: &str, salt: [u8; 32]) -> Result<Self, SomeError> {
        ...
    }

    /// Encodes the challenge into a string (or maybe [u8; 64]).
    pub encode(&self) -> String {
        ...
    }

    /// Checks whether the provided nonce satisfies the difficulty requirement encoded in the
    /// challenge.
    pub fn validate_pow(&self, nonce: u64) -> bool {
        ...
    }
}
```

bin/faucet/src/server/get_tokens.rs
```diff
@@ -181,10 +188,45 @@ impl RawMintRequest {
...
 
+struct RequestCounterGuard {
+    active_count: Arc<AtomicUsize>,
+    queued_count: Arc<AtomicUsize>,
+}
+
+impl RequestCounterGuard {
+    fn new(active_count: Arc<AtomicUsize>, queued_count: Arc<AtomicUsize>) -> Self {
+        active_count.fetch_add(1, Ordering::Relaxed);
+        RequestCounterGuard { active_count, queued_count }
```
sergerad, line 199: nit: `Self {...`

bin/faucet/src/server/get_tokens.rs
```diff
@@ -181,10 +188,45 @@ impl RawMintRequest {
...
+    }
+
+    fn mark_queued(&self) {
+        self.queued_count.fetch_add(1, Ordering::Relaxed);
+    }
+
+    fn mark_processing(&self) {
+        self.queued_count.fetch_sub(1, Ordering::Relaxed);
+    }
```
sergerad, line 208: Just wondering if it would make more sense to have two guards rather than these associated fns.

A guard for queued and a guard for processing. Might be a reason why that wouldn't work. But feels like it could be less error prone / risk of edge cases that we aren't unit testing for. Unsure.

sergerad: Perhaps `mark_enqueued` and `mark_dequeued` might be a bit more obvious for the reader?

sergerad: Wondering we don't mark dequeued when error occurs too.

bin/faucet/src/server/get_tokens.rs
```diff
@@ -181,10 +188,45 @@ impl RawMintRequest {
...
+        self.active_count.fetch_sub(1, Ordering::Relaxed);
+    }
+}
+
 pub async fn get_tokens(
     State(server): State<Server>,
     Query(request): Query<RawMintRequest>,
 ) -> Sse<impl Stream<Item = Result<Event, Infallible>>> {
+    let request_guard =
+        RequestCounterGuard::new(server.active_requests.clone(), server.queued_requests.clone());
```
igamigo, line 222: nit: You can take `&Arc<_>` in `RequestCounterGuard::new()` and clone internally for a slightly more ergonomic API here

bin/faucet/src/server/pow.rs
```diff
@@ -133,10 +142,36 @@ impl TryFrom<&RawMintRequest> for PowParameters {
...
+
+impl PoW {
+    /// Adjust the difficulty of the `PoW`.
+    ///
+    /// The difficulty is adjusted based on the number of active requests.
+    /// The difficulty is increased by 1 for every 50 active requests.
+    /// The difficulty is clamped between 1 and `MAX_DIFFICULTY`.
+    pub fn adjust_difficulty(&self, active_requests: usize) {
```
igamigo, line 169: Right now if we go back (for example) to a max queue of 200 requests the effective max difficulty is 4, right? If not, disregard this comment. 

But otherwise, and maybe not for now, should this function take into account the max queue length somehow? 
For example, we could map the queue length into the difficulty range by computing the filled ration of the queue and applying that to the `[1..MAX_DIFFICULTY]` range. This way the step interval (50 in this case) is not "coupled" with the queue length in any way and we could tweak difficulty or max queue length and expect it to work linearly (or we could decide to grow it exponentially, etc.).

bin/faucet/src/server/pow.rs
```diff
@@ -16,10 +19,10 @@ use super::{
     get_tokens::{InvalidRequest, RawMintRequest},
 };
 
-/// The difficulty of the `PoW`.
+/// The maximum difficulty of the `PoW`.
 ///
 /// The difficulty is the number of leading zeros in the hash of the seed and the solution.
-const DIFFICULTY: u64 = 5;
+const MAX_DIFFICULTY: u64 = 24;
```
igamigo, line 25: Related to my previous comment, since the max queue length is 1000 (AFAIK, maybe something changed or I'm missing something else), I think the effective max difficulty would be 20 here.

bin/faucet/src/server/pow.rs
igamigo: I just saw that the difficulty is adjusted based on total requests (even if the PoW does not validate), not just minting requests in progress. I was expecting it to be over the requests that validated correctly, but I think this might be fine and maybe dismisses my comment above. Not entirely sure if there is a practical difference. 

In any case, would it make sense to make the step interval be configurable? Maybe as part of #878

igamigo (approved):
LGTM! Left some comments/questions

bin/faucet/src/server/get_tokens.rs
SantiagoPittella: I simplified the requests tracking, having only active requests since I wasn't using the actual value of queued requests anyways.

sergerad (approved):

PR #873, opened 2025-05-24
ci: test on beta toolchain

### LINKED ISSUE #691: [CI] Rust beta toolchain
The [2024 State of Rust](https://blog.rust-lang.org/2025/02/13/2024-State-Of-Rust-Survey-results.html) survey has a call to increase feedback on the beta toolchain:

> Unsurprisingly, the most popular version of Rust is latest stable, either the most recent one or whichever comes with the users' Linux distribution. Almost a third of users also use the latest nightly release, due to various reasons (see below). **However, it seems that the beta toolchain is not used much, which is a bit unfortunate. We would like to encourage Rust users to use the beta toolchain more (e.g. in CI environments) to help test soon-to-be stabilized versions of Rust.**

Perhaps we could add a cron trigger for our CI workflow to run against the beta toolchain, e.g. that runs once a day against current `next`. This doesn't cost us anything besides setting it up, and investigating failures if we feel like it :)

c410-f3r:
Fix #691

Placing into a separated file to allow an independent execution that doesn't affect other jobs. Runs every day at 06:00am UTC.

AFAICT cron jobs target the current active branch, which already is `next`.

Although nice, IMO it shouldn't be necessary to test against beta toolchains because an official team already does it in every new pre-release. For example, https://github.com/rust-lang/rust/issues/139827.

### DIFF
--- a/.github/workflows/test-beta.yml
+++ b/.github/workflows/test-beta.yml
@@ -1,0 +1,18 @@
+name: Beta toolchain
+
+on:
+  schedule:
+    - cron: "0 6 * * *" # Everyday at 06:00am UTC
+
+jobs:
+  test:
+    name: test
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@next
+      - name: Rustup
+        run: rustup install beta && rustup default beta
+      - uses: taiki-e/install-action@nextest 
+      - name: Run tests
+        run: make test


.github/workflows/test-beta.yml
```diff
@@ -0,0 +1,25 @@
+name: test-beta
+
+on:
+  schedule:
+    - cron: "0 6 * * *"
+
+concurrency:
+  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
+  cancel-in-progress: true
```
Mirko-von-Leipzig, line 9: I think we don't need to limit concurrency since we're restricting this to a daily run.

```suggestion
```

.github/workflows/test-beta.yml
```diff
@@ -0,0 +1,25 @@
+name: test-beta
+
+on:
+  schedule:
+    - cron: "0 6 * * *"
```
Mirko-von-Leipzig, line 5: Could you add a comment saying this is a daily 06h00 run?

I always have to think about the cron job syntax :)

.github/workflows/test-beta.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: test
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@main
+      - name: Rustup
+        run: rustup install beta && rustup default beta
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
```
Mirko-von-Leipzig, line 22: github cache is quite space limited, and this only runs once a day so I don't think we should cache this.

```suggestion
```

.github/workflows/test-beta.yml
```diff
@@ -0,0 +1,25 @@
+name: test-beta
```
Mirko-von-Leipzig, line 1:
```suggestion
name: Beta toolchain
```

.github/workflows/test-beta.yml
```diff
@@ -0,0 +1,25 @@
...
+  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: test
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@main
```
Mirko-von-Leipzig, line 17: Let's run the tests against the latest dev work

```suggestion
      - uses: actions/checkout@next
```

c410-f3r:
Thanks @Mirko-von-Leipzig. All changes have been submitted.

PR #880, opened 2025-05-26
feat: store used challenges

### LINKED ISSUE #876: faucet(PoW): store solved challenges instead of active ones
Instead of caching the active challenges (challenges created but not yet solved), we should cache the solved challenges. With this approach keep ensuring that the challenges aren't reused, with the plus of avoid of getting the cache spammed.

SantiagoPittella:
closes #876


bin/faucet/src/server/pow.rs
```diff
@@ -108,7 +104,7 @@ impl PowParameters {
 
         // If we get here, the solution is valid
         // Remove the challenge to prevent reuse
-        challenges.remove(&self.pow_seed);
+        challenges.insert(self.pow_seed.to_string(), self.server_timestamp);
```
bobbinth, line 107: Comments here need to be updated.

sergerad (approved):
LGTM. Just at least one comment to update.

bin/faucet/src/server/pow.rs
```diff
@@ -87,11 +87,7 @@ impl PowParameters {
         let mut challenges =
             challenge_cache.challenges.lock().expect("PoW challenge cache lock poisoned");
 
-        if let Some(challenge) = challenges.get(&self.pow_seed) {
-            if challenge.server_signature != self.server_signature {
-                return Err(InvalidRequest::ServerSignaturesDoNotMatch);
-            }
-        } else {
+        if challenges.get(&self.pow_seed).is_some() {
             return Err(InvalidRequest::InvalidChallenge);
```
igamigo, line 91: With this change maybe the variant could be renamed as well to `ChallengeAlreadyUsed` (and the error message changed as wel)

igamigo (approved):
LGTM. I re-ran the Docker build check because it looked like a transient issue and it succeeded, but the test timed out (maybe difficulty was somehow too high) so we'll have to investigate it

PR #881, opened 2025-05-27
fix(ntx): Conversions and column rename

igamigo:
After [running a test](https://github.com/0xMiden/miden-client/pull/928/files#diff-593edd05ec1650928be20f1b680c82073de82143707903d4b76d31c4fa8e8e89R529), we found some issues related to the latest refactors in the NTB. This PR fixes them and adds minor refactors. I've also tested the builder by spamming with some notes for accounts that don't exist to see how it'd work and how the traces look.

### DIFF
--- a/crates/ntx-builder/src/builder/server/state.rs
+++ b/crates/ntx-builder/src/builder/server/state.rs
@@ -186,7 +186,7 @@ mod tests {
 
     use super::*;
 
-    /// Creates a note for a netwokr account with a 30-bit prefix based on the input.
+    /// Creates a note for a network account with a 30-bit prefix based on the input.
     fn mock_note(account_id_diff: u32) -> NetworkNote {
         let metadata = NoteMetadata::new(
             ACCOUNT_ID_PRIVATE_SENDER.try_into().unwrap(),
--- a/crates/ntx-builder/src/builder/server/state.rs
+++ b/crates/ntx-builder/src/builder/server/state.rs
@@ -105,10 +105,10 @@ impl PendingNotes {
                     .expect("note must be on the map if nullifier was found")
                     .metadata()
                     .tag();
-
-                let bucket =
-                    self.by_tag.get_mut(&tag).expect("any tracked note is in the by_tag map");
-                bucket.retain(|&x| x != id);
+                // The bucket may have been pruned already, so check before
+                if let Some(bucket) = self.by_tag.get_mut(&tag) {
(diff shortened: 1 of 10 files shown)


crates/proto/src/domain/note.rs
```diff
@@ -137,9 +137,9 @@ impl TryFrom<Note> for NetworkNote {
         if !note.metadata().tag().is_single_target()
             || note.metadata().tag().execution_mode() != NoteExecutionMode::Network
         {
-            return Ok(NetworkNote(note));
+            return Err(NetworkNoteError::InvalidExecutionMode(note.metadata().tag()));
         }
-        Err(NetworkNoteError::InvalidExecutionMode(note.metadata().tag()))
+        Ok(NetworkNote(note))
```
igamigo, line 142: Fix 1: After refactoring the newtype, I mixed up the conversion scenarios

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -180,17 +180,24 @@ mod tests {
...
         let metadata = NoteMetadata::new(
             ACCOUNT_ID_PRIVATE_SENDER.try_into().unwrap(),
             miden_objects::note::NoteType::Public,
-            NoteTag::for_public_use_case(0, tag_suffix, NoteExecutionMode::Network).unwrap(),
+            NoteTag::from_account_id(
+                (ACCOUNT_ID_NETWORK_FUNGIBLE_FAUCET + (u128::from(account_id_diff) << 99))
+                    .try_into()
+                    .unwrap(),
+                NoteExecutionMode::Network,
+            )
```
igamigo, line 199: This change was introduced to easily modify the first 30-bits of the account ID

crates/ntx-builder/src/builder/server/state.rs
```diff
@@ -105,10 +105,10 @@ impl PendingNotes {
...
                     .metadata()
                     .tag();
-
-                let bucket =
-                    self.by_tag.get_mut(&tag).expect("any tracked note is in the by_tag map");
-                bucket.retain(|&x| x != id);
+                // The bucket may have been pruned already, so check before
+                if let Some(bucket) = self.by_tag.get_mut(&tag) {
+                    bucket.retain(|&x| x != id);
+                }
```
igamigo, line 111: Fix 3: As the comment mentions, the bucket may be pruned at this point so it's better to have a sanity check

PR #882, opened 2025-05-27
chore: migration tool for account root

TomasArrachea:
While implementing a db migration from v0.8 v0.9 we encountered a problem: the account root is now computed differently because of updates in the `AccountTree`. This causes the node to fail to build blocks after the migration, as the latest header contains an invalid account root. 

To fix it, this PR adds a tool that reads the latest header from the db, recomputes the account root and updates the row in the db.

TODO:
- The function is currently placed in the stress-test binary but we should define how/where to run it


igamigo:
Would a migration work like this?:

- Stop the 0.8 node
- Run this code (or something to this effect)
- Start the new 0.9 `miden-node`

If so, maybe we could make this a command in `miden-node` (eg `miden-node migrate-acct-roots` or something) and prepare it for the 0.9 release. 

cc @sergerad @bobbinth @PhilippGackstatter

bobbinth:
Very nice! Great job!

Regarding how to run this: would it make sense to integrate this into the migration process somehow? Maybe we create a module containing this functionality in the `migrations` directory (next to SQL files), and then call `migrate_account_root()` after SQL scripts are run?

We'd probably need to add some logic to check  the version of the DB to make sure we call the function after SQL migrating it to v0.9 is run, but not at other points - so, not sure how difficult that would be.

TomasArrachea:
Done! Moved the function call to the `apply_migration`, and runs only on version v0.9.0

igamigo (approved):

sergerad:
> To fix it, this PR adds a tool that reads the latest header from the db, recomputes the account root and updates the row in the db.

Is this a retroactive modification? Ideally the migrations we do work in a manner that doesn't alter historic state. Otherwise we run the risk of (future) peers or syncing nodes from failing on the modified block.

Maybe its not the right time for it but we should probably get our heads around hard fork procedures sooner rather than later. cc @bobbinth

PR #884, opened 2025-05-27
feat(NTB): CLI args on `miden-node bundled`

igamigo:
This PR adds:

- NTB ticker interval parameter to the `miden-node bundled` command (keeps the same default as before).
- Optional NTB updtaers for block producer and `no-ntb` option to `miden-node bundled` to not spawn an NTB task.
- A small refactor to reduce trace spams

### DIFF
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -84,11 +84,11 @@ pub enum BundledCommand {
 
         /// Interval at which to run the network transaction builder's ticker.
         #[arg(
-                    long = "ntb.interval",
-                    default_value = DEFAULT_NTX_TICKER_INTERVAL_MS,
-                    value_parser = parse_duration_ms,
-                    value_name = "MILLISECONDS"
-                )]
+            long = "ntb.interval",
+            default_value = DEFAULT_NTX_TICKER_INTERVAL_MS,
+            value_parser = parse_duration_ms,
+            value_name = "MILLISECONDS"
+        )]
         ntx_ticker_interval: Duration,
 
         /// Interval at which to procude batches in milliseconds.
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -11,8 +11,8 @@ use url::Url;
 
 use super::{
     DEFAULT_BATCH_INTERVAL_MS, DEFAULT_BLOCK_INTERVAL_MS, DEFAULT_MONITOR_INTERVAL_MS,
-    ENV_BATCH_PROVER_URL, ENV_BLOCK_PROVER_URL, ENV_DATA_DIRECTORY, ENV_ENABLE_OTEL,
-    ENV_NTX_PROVER_URL, ENV_RPC_URL, parse_duration_ms,
+    DEFAULT_NTX_TICKER_INTERVAL_MS, ENV_BATCH_PROVER_URL, ENV_BLOCK_PROVER_URL, ENV_DATA_DIRECTORY,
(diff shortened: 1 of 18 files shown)


bin/node/src/commands/bundled.rs
```diff
@@ -78,6 +82,15 @@ pub enum BundledCommand {
...
         block_interval: Duration,
 
+        /// Interval at which to run the network transaction builder's ticker.
+        #[arg(
+                    long = "ntb.interval",
+                    default_value = DEFAULT_NTX_TICKER_INTERVAL_MS,
+                    value_parser = parse_duration_ms,
+                    value_name = "MILLISECONDS"
+                )]
+        ntx_ticker_interval: Duration,
```
bobbinth, line 92: nit: attribute formatting is a bit off here.

bin/node/src/commands/bundled.rs
```diff
@@ -122,33 +135,41 @@ impl BundledCommand {
...
         rpc_url: Url,
         data_directory: PathBuf,
+        no_ntb: bool,
         batch_prover_url: Option<Url>,
         block_prover_url: Option<Url>,
         tx_prover_url: Option<Url>,
         batch_interval: Duration,
         block_interval: Duration,
         monitor_interval: Duration,
+        ntx_ticker_interval: Duration,
```
bobbinth, line 170: Not for this PR, but I do think we should start thinking about how to group these. For example, `no_ntb`, `tx_prover_url`, and `ntx_ticker_interval` should probably be encapsulated in something like `NtbConfig` struct.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -213,8 +213,7 @@ impl NetworkTransactionBuilder {
     /// # Errors
     ///
     /// - Returns an error only when the preflight stage errors. On the execution stage, errors are
-    /// logged and the transaction gets rolled back.
-    #[instrument(parent = None,target = COMPONENT, name = "ntx_builder.build_network_tx", skip_all)]
```
bobbinth, line 217: Why did we have to remove the `instrument` attribute here?

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -271,7 +270,6 @@ impl NetworkTransactionBuilder {
     ///
     /// If this function errors, the notes are effectively discarded because [`Self::rollback_tx()`]
     /// is not called.
-    #[instrument(target = COMPONENT, name = "ntx_builder.select_next_batch", skip_all)]
```
bobbinth, line 274: Same question as above.

bobbinth (commented):

crates/ntx-builder/src/builder/mod.rs
igamigo: I essentially moved this one down after a tag is found. Basically it was too verbose to no benefit. If no tag is found, there is a `debug!()` trace in `build_network_tx()` anyway.

crates/ntx-builder/src/builder/mod.rs
igamigo: Similar to the other, emitting a trace here was too verbose.

bin/node/src/commands/bundled.rs
igamigo: I agree. I didn't want to change too much for now but I can open an issue to track that.

bobbinth: Yes - that would be great!

bobbinth (approved):
Looks good! Thank you!

PR #887, opened 2025-05-28
chore: Add instrumentation to get_tokens and validate

sergerad:


sergerad:
Hey @SantiagoPittella just checking there wasn't a reason for removing the instrumentation macro above `get_tokens`? Maybe in relation to the `Sse` stream log stuff you did there? AFAICT still good to have the span there despite the event/log.
```rust
pub async fn get_tokens(
    State(server): State<Server>,
) -> Sse<impl Stream<Item = Result<Event, Infallible>>> {
```

PR #891, opened 2025-05-28
chore: Remove some TODOs

c410-f3r:


crates/block-producer/src/block_builder/mod.rs
```diff
@@ -304,7 +304,6 @@ impl BlockBuilder {
                 ntb_client
                     .clone()
                     .submit_network_notes(
-                        // TODO: using default here until there is a good reason not to
                         TransactionId::new(
                             Digest::default(),
                             Digest::default(),
```
c410-f3r, line 309: Let me know if another hard-coded hash is desired. AFAICT, default values are a feature in this case.

PR #904, opened 2025-05-30
bug: Specify shell for dry run check

sergerad:
Fixes the cargo publish dry run workflow.

### DIFF
--- a/.github/workflows/crates-check.yml
+++ b/.github/workflows/crates-check.yml
@@ -3,6 +3,7 @@ name: Publish Crates Dry Run
 on:
   push:
     branches: [main]
+  pull_request: # TODO: remove before merge
 
 permissions:
   contents: read
@@ -15,4 +16,5 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Cargo Publish Dry Run
+        shell: bash
         run: ./scripts/publish-crates.sh --dry-run
--- a/.github/workflows/crates-check.yml
+++ b/.github/workflows/crates-check.yml
@@ -3,7 +3,6 @@ name: Publish Crates Dry Run
 on:
   push:
     branches: [main]
-  pull_request: # TODO: remove before merge
 
 permissions:
   contents: read
--- a/.github/workflows/crates-check.yml
+++ b/.github/workflows/crates-check.yml
@@ -16,4 +16,4 @@ jobs:
 
       - name: Cargo Publish Dry Run
         shell: bash
-        run: ./scripts/publish-crates.sh --dry-run
+        run: cargo +nightly -Z package-workspace publish --workspace --dry-run
--- a/scripts/publish-crates.sh
+++ b/scripts/publish-crates.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Script to publish all miden-node crates to crates.io.
 # Usage: ./publish-crates.sh [args]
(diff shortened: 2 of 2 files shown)


Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
> Dry-run won't actually succeed when we try to publish new version (or at least I ran into this earlier). Basically, let's say we try to run dry-run on `miden-node-store` v0.10.0. It would depend on `miden-node-utils` v0.10.0, but since the utils haven't been published yet, the dry-run will fail.

Ah. I believe one has to specify the `--workspace` flag. So something like

```sh
cargo publish --workspace --dry-run
```

Which makes me wonder if we aren't going about this completely wrong :grimacing: If we just use the workspace publishing then we just need to check which binaries and tools we want to _exclude_ from this.

`--workspace` is unstable, but I think it should be fine to use?

bobbinth:
> Which makes me wonder if we aren't going about this completely wrong 😬 If we just use the workspace publishing then we just need to check which binaries and tools we want to _exclude_ from this.
> 
> `--workspace` is unstable, but I think it should be fine to use?

I think this should be fine. The only thing is that we'd need to mark some crates as "unpublishable".

sergerad:
Just had a quick look and couldn't find a way to avoid publishing all crates in the workspace.

https://github.com/rust-lang/cargo/issues/10948

When I set `package.publish = false` the command causes the following error:
```
❮ cargo +nightly -Z package-workspace publish --workspace --dry-run
error: `miden-node-test-macro`, `miden-node-stress-test` cannot be published.
`package.publish` must be set to `true` or a non-empty list in Cargo.toml to publish
```

Might need to raise that in the issue above. But I'll have a proper read of it later today.

bobbinth:
> Oddly complains about
> 
> ```
> error: failed to prepare local package for uploading
> 
> Caused by:
>   failed to select a version for the requirement `miden-node-rpc = "^0.10"`
>   candidate versions found which didn't match: 0.9.0, 0.8.0, 0.7.1, ...
>   location searched: crates.io index
>   required by package `miden-node-rpc v0.10.0 (/Users/sergeradinovich/Source/miden-node/crates/rpc)`
> ```

I fixed this issue in `main` but haven't merged it back into `next` yet (planning to do in the next hour or so). The problem here was a self-referential dev-dependency in `miden-node-rpc` crate.

bobbinth:
The changes should now be in the `next`.

For this PR, let's mark the crates that we don't want to publish and "unpublishable" (I think this can be done via `Cargo.toml`).

sergerad:
> For this PR, let's mark the crates that we don't want to publish and "unpublishable" (I think this can be done via `Cargo.toml`).

Yea I mentioned above that `package.publish = false` seems to be incompatible with the `cargo publish --workspace command`:
```
❮ cargo +nightly -Z package-workspace publish --workspace --dry-run
error: `miden-node-stress-test` cannot be published.
`package.publish` must be set to `true` or a non-empty list in Cargo.toml to publish
```

All packages in the workspaces must be publishable AFAICT. I'm hoping this doesn't matter for dry-run. If it does, it will block us using that command with a workspace with unpublishable crates.

Unless I'm missing another configuration field you are thinking about other than `package.publish` in the `Cargo.toml`.

bobbinth:
> We might have more luck (and should probably just move straight to) using release-plz

Yes, maybe we should move directly to it. I think it can do much more than just running `cargo publish` - but we can probably use it just for publishing (at least for now).

Issue #878, opened 2025-05-26
faucet(PoW): `Challenge` and `PoW` structs refactor


SantiagoPittella:
>  I'd encapsulate functionality a bit more around `PoW` and `Challenge` structs. Maybe something like this:

```rust
pub struct PoW {
    salt: [u8; 32],
    difficulty: Arc<AtomicU64>,
    challenge_cache: ChallengeCache,
}

impl Pow {

    /// Generates a new challenge.
    pub fn build_challenge(&self, request_ip: IpAddr) -> Challenge {
        ...
    }

    /// Verifies that the challenge is valid (e.g., timestamp is within range, difficulty is
    /// satisfied etc.), and adds the validate challenge to the cache.
    pub fn validate_challenge(&mut self, challenge: Challenge, nonce: u64) -> Result<(), SomeError> {
        ...
    }

    pub fn adjust_difficulty(&self, active_requests: usize) {
        ...
    }
}

pub struct Challenge {
    difficulty: u64,
    ip_address: IpAddr,
    timestamp: u64,
    signature: [u8; 32],
}

impl Challenge {
    /// Decodes the challenge and verifies that the signature part of the challenge is valid
    /// in the context of the specified salt.
    pub fn decode(value: &str, salt: [u8; 32]) -> Result<Self, SomeError> {
        ...
    }

    /// Encodes the challenge into a string (or maybe [u8; 64]).
    pub encode(&self) -> String {
        ...
    }

    /// Checks whether the provided nonce satisfies the difficulty requirement encoded in the
    /// challenge.
    pub fn validate_pow(&self, nonce: u64) -> bool {
        ...
    }
}
```

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-node/pull/865#discussion_r2106443367_

bobbinth:
My thinking was that we'd have independent PoW requirements for each API key. So, if a lot of requests come from one API key, PoW for this API key would go up, but PoW for other API keys would be unaffected. We can then treat requests that don't have an API key as a special case of this setup (i.e., assign them to "null" API key bucket).

sergerad:
Because the faucet is irrelevant for mainnet, I would treat it as much lower priority than other work streams. In that vein, my 2 cents is to minimize complexity and time spend on any related features. Would have ideally offloaded all API key work entirely to something like AWS API Gateway for example, but that is not an option for us with e2e gRPC AFAIK.

igamigo:
[...]

Issue #846, opened 2025-05-15
feat: move rate limiting parameters to a config file or env var


tomyrd:
Creating an issue to track this TODO. It will be useful for changing the rate limiter parameters without having to redeploy the faucet, as the chosen parameters are somewhat arbitrary and might need to change in the future.

> Just added a new TODO related to this, specifically about moving all the rate limiting parameters to the config file or to env vars. For more context see [this comment](https://github.com/0xMiden/miden-node/pull/839#discussion_r2087525508). 

 _Originally posted by @tomyrd in [#821](https://github.com/0xMiden/miden-node/issues/821#issuecomment-2881719068)_

TomasArrachea:
in https://github.com/0xMiden/miden-node/issues/960, the PoW was modified to work as a native rate limiter. The `governor_tower` rate limiter is no longer necessary, but we may still want to make the rate limiter configurable. 

The native rate limiter works by limiting accounts to submit new challenges while the `ChallengeCache` still holds the previous submitted challenge. It's held while it's valid, so setting the challenge expiration would modify the rate limiting.

We can also have a more complex rate limiter with this same structure, by allowing an account to submit an arbitrary configurable amount of challenges simultaneously to the `ChallengeCache`, similar to the `governor_tower`'s `burst_size` parameter.

bobbinth:
I'd go with the simple approach for now (1 challenge per account can be in cache at any point in time).

bobbinth:
Is this largely done, or is anything outstanding?

TomasArrachea:
Yes, we can close this

PR #905, opened 2025-05-30
feat: add rate limiter settings as config

TomasArrachea:


bin/faucet/src/config.rs
```diff
@@ -38,6 +48,25 @@ pub struct FaucetConfig {
...
+    pub ip_rate_limit_per_second: u64,
+    /// Max burst of requests per account
+    pub account_rate_limit_burst_size: u32,
+    /// Max requests per account per second
+    pub account_rate_limit_per_second: u64,
+    /// Max burst of requests per API key
+    pub api_key_rate_limit_burst_size: u32,
+    /// Max requests per API key per second
+    pub api_key_rate_limit_per_second: u64,
 }
```
igamigo, line 70: I think it would be worth documenting this a bit more, even if these are concepts explained within `actix`

TomasArrachea: Ok! I added some docs to the struct explaining the rate limiter params, let me know if there is anything to clarify/improve

bobbinth:
Should we hold off on this until https://github.com/0xMiden/miden-node/issues/878 is done? It may affect how we want to configure the rate limiter.

TomasArrachea:
will leave this as draft until the refactor is done

bin/faucet/.env
```diff
@@ -9,3 +9,9 @@ MIDEN_FAUCET_REMOTE_TX_PROVER_URL=
 MIDEN_FAUCET_POW_SECRET=
 MIDEN_FAUCET_API_KEYS=
 MIDEN_FAUCET_ENABLE_OTEL=false
+MIDEN_FAUCET_RATE_LIMITER_IP_BURST_SIZE=
+MIDEN_FAUCET_RATE_LIMITER_IP_PER_SECOND=
+MIDEN_FAUCET_RATE_LIMITER_ACCOUNT_BURST_SIZE=
+MIDEN_FAUCET_RATE_LIMITER_ACCOUNT_PER_SECOND=
+MIDEN_FAUCET_RATE_LIMITER_API_KEY_BURST_SIZE=
+MIDEN_FAUCET_RATE_LIMITER_API_KEY_PER_SECOND=
```
sergerad, line 17: Please note we have changed to human readable durations now that https://github.com/0xMiden/miden-node/pull/998 has been merged

sergerad: Ah these aren't durations. I'll leave the comment here in case you do add a duration in any case. Feel free to resolve once you have read it :)

bin/faucet/src/server/rate_limiter.rs
```diff
@@ -0,0 +1,127 @@
...
+use tower_governor::{GovernorError, key_extractor::KeyExtractor};
+
+use crate::server::get_tokens::RawMintRequest;
+
+const ENV_IP_BURST_SIZE: &str = "MIDEN_FAUCET_RATE_LIMITER_IP_BURST_SIZE";
+const ENV_IP_PER_SECOND: &str = "MIDEN_FAUCET_RATE_LIMITER_IP_PER_SECOND";
+const ENV_ACCOUNT_BURST_SIZE: &str = "MIDEN_FAUCET_RATE_LIMITER_ACCOUNT_BURST_SIZE";
+const ENV_ACCOUNT_PER_SECOND: &str = "MIDEN_FAUCET_RATE_LIMITER_ACCOUNT_PER_SECOND";
+const ENV_API_KEY_BURST_SIZE: &str = "MIDEN_FAUCET_RATE_LIMITER_API_KEY_BURST_SIZE";
+const ENV_API_KEY_PER_SECOND: &str = "MIDEN_FAUCET_RATE_LIMITER_API_KEY_PER_SECOND";
```
sergerad, line 16: Please add these (as empty assignments) to `bin/faucet/.env` also.

TomasArrachea: Aren't they already added here? Or is it something different?
https://github.com/0xMiden/miden-node/blob/24a1e931a7d75fa0fd6951c9de80f447d576201c/bin/faucet/.env#L12-L17

sergerad: Yes thats it. Sorry if I missed it. Feeling a bit light-headed today!

bobbinth:
With the latest updates to how we do PoW - do we actually need these rate limits? My understanding was that they are naturally handled by how we designed challenge cache.

TomasArrachea:
> With the latest updates to how we do PoW - do we actually need these rate limits? My understanding was that they are naturally handled by how we designed challenge cache.

There is definitely an overlap now, I wasn't sure if the idea was to keep both rate limiters or to replace the `governor_tower` rate limiter. I agree that for the accounts and api key we could rely on our own rate limiter. The rate limiter by IP is different, but I don't see why it should be needed.

bobbinth:
> There is definitely an overlap now, I wasn't sure if the idea was to keep both rate limiters or to replace the `governor_tower` rate limiter. I agree that for the accounts and api key we could rely on our own rate limiter. The rate limiter by IP is different, but I don't see why it should be needed.

Yeah, my thinking was that we could remove it altogether once the PoW refactoring is done.

igamigo:
Should this PR be closed?

TomasArrachea:
Yes, we can close this. I'll make a PR removing the rate limiter.

For the PoW parameters, I think we could either make the challenge expiration time configurable or add a new parameter to set how long challenges live in the cache. I'll comment this on https://github.com/0xMiden/miden-node/issues/846.

PR #908, opened 2025-05-31
Tracking PR for v0.10.0 release

bobbinth:
This is a tracking PR for v0.10.0 release.


bobbinth:
Not clear to me why mdBook fails to build - the file it complains about it is there and it builds fine locally.

PR #911, opened 2025-06-02
feat: update `miden-base` to next

tomyrd:
We need to update the source of the `miden-base` dependencies so that we can also update in the client.

### DIFF
--- a/bin/faucet/Cargo.toml
+++ b/bin/faucet/Cargo.toml
@@ -53,7 +53,7 @@ miden-node-utils = { features = ["vergen"], workspace = true }
 fantoccini                = { version = "0.21" }
 miden-node-block-producer = { features = ["testing"], workspace = true }
 miden-objects             = { features = ["testing"], workspace = true }
-miden-testing             = { branch = "next", features = ["async"], git = "https://github.com/0xMiden/miden-base" }
+miden-testing             = { features = ["async"], workspace = true }
 serde_json                = { version = "1.0" }
 tokio                     = { features = ["process"], workspace = true }
 tokio-stream              = { features = ["net"], workspace = true }
--- a/bin/faucet/Cargo.toml
+++ b/bin/faucet/Cargo.toml
@@ -53,7 +53,7 @@ miden-node-utils = { features = ["vergen"], workspace = true }
 fantoccini                = { version = "0.21" }
 miden-node-block-producer = { features = ["testing"], workspace = true }
 miden-objects             = { features = ["testing"], workspace = true }
-miden-testing             = { features = ["async"], version = "0.9.0" }
+miden-testing             = { branch = "next", features = ["async"], git = "https://github.com/0xMiden/miden-base" }
 serde_json                = { version = "1.0" }
 tokio                     = { features = ["process"], workspace = true }
 tokio-stream              = { features = ["net"], workspace = true }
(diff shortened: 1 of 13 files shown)


bin/faucet/Cargo.toml
```diff
@@ -53,7 +53,7 @@ miden-node-utils = { features = ["vergen"], workspace = true }
 fantoccini                = { version = "0.21" }
 miden-node-block-producer = { features = ["testing"], workspace = true }
 miden-objects             = { features = ["testing"], workspace = true }
-miden-testing             = { features = ["async"], version = "0.9.0" }
+miden-testing             = { branch = "next", features = ["async"], git = "https://github.com/0xMiden/miden-base" }
```
Mirko-von-Leipzig: Let's move this to a workspace dep, I imagine we'll want these objects elsewhere as well someday.

bin/stress-test/Cargo.toml
```diff
@@ -19,7 +19,7 @@ workspace = true
 clap                      = { features = ["derive"], version = "4.5" }
 futures                   = { workspace = true }
 miden-air                 = { workspace = true }
-miden-block-prover        = { features = ["testing"], version = "0.9" }
+miden-block-prover        = { branch = "next", features = ["testing"], git = "https://github.com/0xMiden/miden-base" }
```
Mirko-von-Leipzig: Let's also make this a workspace dep

bin/faucet/src/main.rs
```diff
@@ -175,25 +175,14 @@ async fn run_faucet_command(cli: Cli) -> anyhow::Result<()> {
         },
 
         Command::CreateFaucetAccount {
-            config_path,
+            config_path: _,
```
Mirko-von-Leipzig: Let's remove this from the cli then. This would be a changelog item then

bin/faucet/src/main.rs
```diff
@@ -466,6 +456,20 @@ mod test {
...
+        let timeout = Duration::from_secs(10);
+        loop {
+            match tokio::net::TcpStream::connect(addr).await {
+                Ok(_) => break,
+                Err(_) if start.elapsed() < timeout => {
+                    sleep(Duration::from_millis(200)).await;
+                },
+                Err(e) => panic!("faucet never became reachable: {e}"),
+            }
+        }
```
igamigo, line 471: I added this waiting loop because it seems like there was a race condition where the client was attempting to connect before the faucet was fully up (since the faucet gets started on a different thread)? It was causing the website test to fail ([CI run](https://github.com/0xMiden/miden-node/actions/runs/15398251754/job/43324326150)). I don't know how this didn't come up before, maybe just worse luck now. cc @SantiagoPittella @TomasArrachea

igamigo (approved):

crates/block-producer/Cargo.toml
```diff
@@ -22,14 +22,14 @@ tracing-forest = ["miden-node-utils/tracing-forest"]
...
-miden-block-prover           = { version = "0.9" }
+miden-block-prover           = { workspace = true }
 miden-lib                    = { workspace = true }
 miden-node-proto             = { workspace = true }
 miden-node-utils             = { features = ["testing"], workspace = true }
 miden-objects                = { workspace = true }
 miden-proving-service-client = { features = ["batch-prover", "block-prover"], workspace = true }
 miden-tx                     = { workspace = true }
-miden-tx-batch-prover        = { version = "0.9" }
+miden-tx-batch-prover        = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
```
Mirko-von-Leipzig: This could also be a workspace dep I think

Mirko-von-Leipzig (approved):
Just the one dep to move still, and then adding a changelog entry for the removal of the faucet account creation cli change.

PR #913, opened 2025-06-02
chore: Add rustup update for release workflow

sergerad:


.github/workflows/package.yml
```diff
@@ -37,6 +37,16 @@ jobs:
...
 
+      - name: Rust update
+        uses: Swatinem/rust-cache@v2
+        with:
+          # Only update the cache on push onto the next branch. This strikes a nice balance between
+          # cache hits and cache evictions (github has a 10GB cache limit).
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
+      - name: Clippy no_std
+        run: |
+          rustup update --no-self-update
```
bobbinth, line 48: Is the name "Clippy no_std" correct here? I thought this was more about updating Rust rather than clippy.

sergerad: Yea I know its not correct but I wanted to keep it identical to whats in miden-base for proving service. If we end up merging workflows it will just be a bit clearer that the intention is the same. Happy to change if you don't think that makes sense.

bobbinth: If it is not correct in `miden-base` either, we should probably make a small PR there too to fix it.

PR #919, opened 2025-06-03
chore: Update publish workflow

### LINKED ISSUE #897: Implement Workflow to publish Unstable Debian packages

sergerad:
### Context

Closes #897, #900, #899, #898, #901.

We want to publish release candidate and release artefacts (Debian packages and Docker images).

### Changes
- Remove stale workflows
- Add workflow to publish Debian packages and existing Docker images on release / prerelease based on tag

### Future work
- Add a Docker image for faucet and integrate with build and publish workflows


.github/actions/build_package/action.yml
```diff
@@ -67,9 +66,9 @@ runs:
         Section: base
         Priority: optional
         Architecture: $arch
-        Maintainer: Polygon Devops <devops@polygon.technology>
+        Maintainer: Miden <sergerad@miden.team>
```
Mirko-von-Leipzig, line 69: I wonder if we should create a catch-all team email, `node@miden.team` or `devops@miden.team`.

This would also be useful for security reports.

PR #923, opened 2025-06-04
chore: refactor PoW and Challenge structs

SantiagoPittella:
part of #878

### DIFF
--- a/bin/faucet/src/server/challenge.rs
+++ b/bin/faucet/src/server/challenge.rs
@@ -1,5 +1,5 @@
 use miden_tx::utils::{ToHex, hex_to_bytes};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 use sha3::{Digest, Sha3_256};
 
 use super::get_tokens::InvalidRequest;
@@ -11,11 +11,25 @@ use super::get_tokens::InvalidRequest;
 pub(crate) const SERVER_TIMESTAMP_TOLERANCE_SECONDS: u64 = 30;
 
 /// A challenge for proof-of-work validation.
-#[derive(Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
+#[derive(Debug, Clone, Deserialize, Hash, PartialEq, Eq)]
 pub struct Challenge {
     pub(crate) difficulty: usize,
     pub(crate) timestamp: u64,
     pub(crate) signature: [u8; 32],
+}
+
+impl Serialize for Challenge {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        use serde::ser::SerializeStruct;
+        let mut state = serializer.serialize_struct("Challenge", 3)?;
+        state.serialize_field("challenge", &self.encode())?;
+        state.serialize_field("difficulty", &self.difficulty)?;
+        state.serialize_field("timestamp", &self.timestamp)?;
+        state.end()
+    }
 }
 
 impl Challenge {
@@ -91,5 +105,32 @@ impl Challenge {
         hasher.update(difficulty.to_be_bytes());
         hasher.update(timestamp.to_be_bytes());
         hasher.finalize().into()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_challenge_serialization() {
+        let secret = [1u8; 32];
+        let challenge = Challenge::new(2, 1234567890, secret);
+        
+        // Test that it serializes to the expected JSON format
+        let json = serde_json::to_string(&challenge).unwrap();
+        
+        // Should contain the expected fields
+        assert!(json.contains("\"challenge\":"));
+        assert!(json.contains("\"difficulty\":2"));
+        assert!(json.contains("\"timestamp\":1234567890"));
+        
+        // Parse back to verify structure
+        let parsed: serde_json::Value = serde_json::from_str(&json).unwrap();
+        assert!(parsed.get("challenge").is_some());
+        assert!(parsed.get("difficulty").is_some());
+        assert!(parsed.get("timestamp").is_some());
+        assert_eq!(parsed["difficulty"], 2);
+        assert_eq!(parsed["timestamp"], 1234567890);
     }
 }
--- a/bin/faucet/src/server/challenge.rs
+++ b/bin/faucet/src/server/challenge.rs
@@ -1,3 +1,5 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+
 use miden_tx::utils::{ToHex, hex_to_bytes};
 use serde::{Deserialize, Serialize, Serializer};
 use sha3::{Digest, Sha3_256};
@@ -105,6 +107,29 @@ impl Challenge {
         hasher.update(difficulty.to_le_bytes());
         hasher.update(timestamp.to_le_bytes());
         hasher.finalize().into()
+    }
+
+    /// Validates the challenge and nonce.
+    ///
+    /// # Errors
+    /// Returns an error if:
+    /// * The challenge is expired.
+    /// * The challenge is invalid.
+    pub fn validate(&self, nonce: u64) -> Result<(), InvalidRequest> {
+        // Check timestamp validity
+        if self.is_expired(SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs()) {
+            return Err(InvalidRequest::ExpiredServerTimestamp(
+                self.timestamp,
+                SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs(),
+            ));
+        }
+
+        // Validate the proof of work
+        if !self.validate_pow(nonce) {
+            return Err(InvalidRequest::InvalidPoW);
+        }
+
+        Ok(())
     }
 }
 
--- a/bin/faucet/src/server/challenge.rs
+++ b/bin/faucet/src/server/challenge.rs
@@ -51,15 +51,13 @@ impl Challenge {
         let timestamp = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
         let signature: [u8; 32] = bytes[16..48].try_into().unwrap();
 
-        let challenge = Self::from_parts(difficulty, timestamp, signature);
-
         // Verify the signature
         let expected_signature = Self::compute_signature(secret, difficulty, timestamp);
-        if signature != expected_signature {
-            return Err(InvalidRequest::ServerSignaturesDoNotMatch);
+        if signature == expected_signature {
+            Ok(Self::from_parts(difficulty, timestamp, signature))
+        } else {
+            Err(InvalidRequest::ServerSignaturesDoNotMatch)
         }
-
-        Ok(challenge)
     }
 
     /// Encodes the challenge into a hex string.
--- a/bin/faucet/src/server/challenge.rs
+++ b/bin/faucet/src/server/challenge.rs
@@ -1,5 +1,3 @@
-use std::time::{SystemTime, UNIX_EPOCH};
-
 use miden_tx::utils::{ToHex, hex_to_bytes};
 use serde::{Deserialize, Serialize};
 use sha3::{Digest, Sha3_256};
@@ -82,12 +80,7 @@ impl Challenge {
     }
 
     /// Checks if the challenge timestamp is not expired.
-    pub fn is_not_expired(&self) -> bool {
-        let current_time = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("System time should always be later than UNIX epoch")
-            .as_secs();
-
+    pub fn is_not_expired(&self, current_time: u64) -> bool {
         (current_time - self.timestamp) <= SERVER_TIMESTAMP_TOLERANCE_SECONDS
     }
 
--- a/bin/faucet/src/server/challenge.rs
+++ b/bin/faucet/src/server/challenge.rs
@@ -92,9 +92,9 @@ impl Challenge {
         leading_zeros >= self.difficulty
     }
 
-    /// Checks if the challenge timestamp is not expired.
-    pub fn is_not_expired(&self, current_time: u64) -> bool {
-        (current_time - self.timestamp) <= CHALLENGE_LIFETIME_SECONDS
+    /// Checks if the challenge timestamp is expired.
+    pub fn is_expired(&self, current_time: u64) -> bool {
+        (current_time - self.timestamp) > CHALLENGE_LIFETIME_SECONDS
     }
 
     /// Computes the signature for a challenge.
--- a/bin/faucet/src/server/challenge.rs
+++ b/bin/faucet/src/server/challenge.rs
@@ -1,0 +1,97 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use miden_tx::utils::{ToHex, hex_to_bytes};
+use serde::{Deserialize, Serialize};
+use sha3::{Digest, Sha3_256};
+
+use super::get_tokens::InvalidRequest;
+
+/// The tolerance for the server timestamp.
+///
+/// The server timestamp is valid if it is within `SERVER_TIMESTAMP_TOLERANCE_SECONDS` seconds of
+/// the current time.
+pub(crate) const SERVER_TIMESTAMP_TOLERANCE_SECONDS: u64 = 30;
+
+/// A challenge for proof-of-work validation.
+#[derive(Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
+pub struct Challenge {
+    pub(crate) difficulty: usize,
+    pub(crate) timestamp: u64,
+    pub(crate) signature: [u8; 32],
+}
+
+impl Challenge {
+    /// Creates a new challenge with the given parameters.
+    pub fn new(difficulty: usize, timestamp: u64, signature: [u8; 32]) -> Self {
+        Self { difficulty, timestamp, signature }
+    }
+
+    /// Decodes the challenge and verifies that the signature part of the challenge is valid
+    /// in the context of the specified salt.
+    pub fn decode(value: &str, salt: [u8; 32]) -> Result<Self, InvalidRequest> {
+        // Add the "0x" prefix to the hex string, otherwise the hex_to_bytes function will fail.
+        let value = "0x".to_string() + value;
+        // Parse the hex-encoded challenge string
+        let bytes = hex_to_bytes::<48>(&value).map_err(|_| InvalidRequest::MissingPowParameters)?;
+
+        if bytes.len() != 48 {
+            // 8 + 8 + 32 bytes
+            return Err(InvalidRequest::MissingPowParameters);
+        }
+
+        // SAFETY: Length of 48 is enforced above.
+        let difficulty = u64::from_le_bytes(bytes[0..8].try_into().unwrap()) as usize;
+        let timestamp = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
+        let signature: [u8; 32] = bytes[16..48].try_into().unwrap();
+
+        let challenge = Self::new(difficulty, timestamp, signature);
+
+        // Verify the signature
+        let expected_signature = Self::compute_signature(salt, difficulty, timestamp);
+        if signature != expected_signature {
+            return Err(InvalidRequest::ServerSignaturesDoNotMatch);
+        }
+
+        Ok(challenge)
+    }
+
+    /// Encodes the challenge into a hex string.
+    pub fn encode(&self) -> String {
+        let mut bytes = Vec::with_capacity(48);
+        bytes.extend_from_slice(&(self.difficulty as u64).to_le_bytes());
+        bytes.extend_from_slice(&self.timestamp.to_le_bytes());
+        bytes.extend_from_slice(&self.signature);
+        bytes.to_hex()
+    }
+
+    /// Checks whether the provided nonce satisfies the difficulty requirement encoded in the
+    /// challenge.
+    pub fn validate_pow(&self, nonce: u64) -> bool {
+        let mut hasher = Sha3_256::new();
+        hasher.update(self.encode());
+        hasher.update(nonce.to_be_bytes());
+        let hash = hasher.finalize();
+
+        let leading_zeros = hash.iter().take_while(|&b| *b == 0).count();
+        leading_zeros >= self.difficulty
+    }
+
+    /// Checks if the challenge timestamp is not expired.
+    pub fn is_not_expired(&self) -> bool {
+        let current_time = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("System time should always be later than UNIX epoch")
+            .as_secs();
+
+        (current_time - self.timestamp) <= SERVER_TIMESTAMP_TOLERANCE_SECONDS
+    }
+
+    /// Computes the signature for a challenge.
+    pub fn compute_signature(salt: [u8; 32], difficulty: usize, timestamp: u64) -> [u8; 32] {
+        let mut hasher = Sha3_256::new();
+        hasher.update(salt);
+        hasher.update(difficulty.to_be_bytes());
+        hasher.update(timestamp.to_be_bytes());
+        hasher.finalize().into()
+    }
+} 
--- a/bin/faucet/src/server/challenge.rs
+++ b/bin/faucet/src/server/challenge.rs
@@ -1,5 +1,3 @@
-use std::time::{SystemTime, UNIX_EPOCH};
-
 use miden_tx::utils::{ToHex, hex_to_bytes};
 use serde::{Deserialize, Serialize, Serializer};
 use sha3::{Digest, Sha3_256};
@@ -107,29 +105,6 @@ impl Challenge {
         hasher.update(difficulty.to_le_bytes());
         hasher.update(timestamp.to_le_bytes());
         hasher.finalize().into()
-    }
-
-    /// Validates the challenge and nonce.
-    ///
-    /// # Errors
-    /// Returns an error if:
-    /// * The challenge is expired.
-    /// * The challenge is invalid.
-    pub fn validate(&self, nonce: u64) -> Result<(), InvalidRequest> {
-        // Check timestamp validity
-        if self.is_expired(SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs()) {
-            return Err(InvalidRequest::ExpiredServerTimestamp(
-                self.timestamp,
-                SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs(),
-            ));
-        }
-
-        // Validate the proof of work
-        if !self.validate_pow(nonce) {
-            return Err(InvalidRequest::InvalidPoW);
-        }
-
-        Ok(())
     }
 }
 
--- a/bin/faucet/src/server/challenge.rs
+++ b/bin/faucet/src/server/challenge.rs
@@ -85,10 +85,11 @@ impl Challenge {
     pub fn validate_pow(&self, nonce: u64) -> bool {
         let mut hasher = Sha3_256::new();
         hasher.update(self.encode());
-        hasher.update(nonce.to_be_bytes());
+        hasher.update(nonce.to_le_bytes());
         let hash = hasher.finalize();
 
         let leading_zeros = hash.iter().take_while(|&b| *b == 0).count();
+
         leading_zeros >= self.difficulty
     }
 
@@ -101,8 +102,8 @@ impl Challenge {
     fn compute_signature(secret: [u8; 32], difficulty: usize, timestamp: u64) -> [u8; 32] {
         let mut hasher = Sha3_256::new();
         hasher.update(secret);
-        hasher.update(difficulty.to_be_bytes());
-        hasher.update(timestamp.to_be_bytes());
+        hasher.update(difficulty.to_le_bytes());
+        hasher.update(timestamp.to_le_bytes());
         hasher.finalize().into()
     }
 }
--- a/bin/faucet/src/server/challenge.rs
+++ b/bin/faucet/src/server/challenge.rs
@@ -47,10 +47,8 @@ impl Challenge {
     /// Decodes the challenge and verifies that the signature part of the challenge is valid
     /// in the context of the specified secret.
     pub fn decode(value: &str, secret: [u8; 32]) -> Result<Self, InvalidRequest> {
-        // Add the "0x" prefix to the hex string, otherwise the hex_to_bytes function will fail.
-        let value = "0x".to_string() + value;
         // Parse the hex-encoded challenge string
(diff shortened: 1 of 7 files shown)


bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,117 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-        if leading_zeros < self.difficulty {
-            return Err(InvalidRequest::InvalidPoW);
+    /// Simple hex encoding function.
+    fn hex_encode(bytes: &[u8]) -> String {
+        use std::fmt::Write;
+        bytes.iter().fold(String::new(), |mut acc, b| {
+            write!(acc, "{b:02x}").unwrap();
+            acc
+        })
+    }
```
TomasArrachea: is this encoding different from `miden_formatting::ToHex`? otherwise I think we can just use `.to_hex()` and avoid this implementation

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,117 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
         }
 
-        // If we get here, the solution is valid
-        // Add the challenge to the cache to prevent reuse
-        challenges.insert(self.pow_seed.to_string(), self.server_timestamp);
+        (0..s.len())
+            .step_by(2)
+            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).map_err(|_| ()))
+            .collect()
+    }
```
TomasArrachea: similarly, I believe you may be able to use `hex_to_bytes()` from `miden-crypto` to avoid this

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,117 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-                self.server_timestamp,
-                server_timestamp,
-            ));
-        }
+    /// Checks whether the provided nonce satisfies the difficulty requirement encoded in the
+    /// challenge.
+    pub fn validate_pow(&self, nonce: u64) -> bool {
+        let mut hasher = Sha3_256::new();
+        hasher.update(self.encode());
+        hasher.update(nonce.to_string().as_bytes());
```
Mirko-von-Leipzig: This feels weird, why don't we just use the raw bytes directly?

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,117 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
+        let hash = hasher.finalize();
+        let hash_hex = hash.to_hex();
```
Mirko-von-Leipzig: This also feels unnecessary. We should be able to count the leading zeros in the bytes.

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,117 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-        // Then check the PoW solution
+    /// Computes the signature for a challenge.
+    fn compute_signature(salt: [u8; 32], difficulty: usize, timestamp: u64) -> [u8; 32] {
         let mut hasher = Sha3_256::new();
-        hasher.update(&self.pow_seed);
-        hasher.update(self.pow_solution.to_string().as_bytes());
-        let hash = &hasher.finalize().to_hex();
+        hasher.update(salt);
+        hasher.update(difficulty.to_string().as_bytes());
+        hasher.update(timestamp.to_string().as_bytes());
```
Mirko-von-Leipzig: Can we not use the number's bytes? Casting numbers to strings is kinda weird.

bin/faucet/src/server/pow.rs
```diff
@@ -155,17 +149,80 @@ impl TryFrom<&RawMintRequest> for PowParameters {
...
+            difficulty: Arc::new(AtomicUsize::new(1)),
+            challenge_cache: ChallengeCache::default(),
+        }
+    }
+
+    /// Generates a new challenge.
+    pub fn build_challenge(&self) -> Challenge {
+        let timestamp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards")
```
Mirko-von-Leipzig:
```suggestion
            .expect("system time should always be later than UNIX epoch")
```

bin/faucet/src/server/pow.rs
```diff
@@ -155,17 +149,80 @@ impl TryFrom<&RawMintRequest> for PowParameters {
...
+            .expect("PoW challenge cache lock poisoned");
+
+        if challenges.contains_key(&challenge_key) {
+            return Err(InvalidRequest::ChallengeAlreadyUsed);
+        }
+
+        // Validate the proof of work
+        if !challenge.validate_pow(nonce) {
+            return Err(InvalidRequest::InvalidPoW);
+        }
```
Mirko-von-Leipzig: Should we move this above the mutex lock?

bin/faucet/src/server/pow.rs
```diff
@@ -155,17 +149,80 @@ impl TryFrom<&RawMintRequest> for PowParameters {
...
+
+        let difficulty = self.difficulty.load(Ordering::Relaxed);
+        let signature = Challenge::compute_signature(self.salt, difficulty, timestamp);
+
+        Challenge::new(difficulty, timestamp, signature)
+    }
+
+    /// Verifies that the challenge is valid (e.g., timestamp is within range, difficulty is
+    /// satisfied etc.), and adds the validated challenge to the cache.
+    pub fn validate_challenge(
```
Mirko-von-Leipzig:
```suggestion
    pub fn submit_challenge(
```

bin/faucet/src/server/pow.rs
```diff
@@ -155,17 +149,80 @@ impl TryFrom<&RawMintRequest> for PowParameters {
...
+                challenge.timestamp(),
+                SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .expect("Time went backwards")
+                    .as_secs(),
+            ));
+        }
+
+        // Check if challenge was already used
+        let challenge_key = challenge.encode();
```
Mirko-von-Leipzig: I think we can `derive(Hash)` on `Challenge`, then we can just use a `HashSet<Challenge>`. If we also change the timestamp from `u64` to `SystemTime` I think it will simplify things.

bin/faucet/src/server/pow.rs
```diff
@@ -155,17 +149,80 @@ impl TryFrom<&RawMintRequest> for PowParameters {
...
+        let mut challenges = self
+            .challenge_cache
+            .challenges
+            .lock()
+            .expect("PoW challenge cache lock poisoned");
+
+        if challenges.contains_key(&challenge_key) {
+            return Err(InvalidRequest::ChallengeAlreadyUsed);
+        }
+
```
Mirko-von-Leipzig: I think you can do this all in one go once its a hashset:

```rust
if self.challenge_cache
    .challenges
    .lock
    .unwrap()
    .insert(challenge)
    .not() 
{ return Err(...)}
```

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,117 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-    pub(crate) pow_seed: String,
-    pub(crate) server_signature: String,
-    pub(crate) server_timestamp: u64,
-    pub(crate) pow_solution: u64,
-    pub(crate) difficulty: usize,
+/// A challenge for proof-of-work validation.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Challenge {
+    difficulty: usize,
+    timestamp: u64,
```
Mirko-von-Leipzig: I would also make these all `pub(crate)` and remove the constructor/getters.

```suggestion
    timestamp: SystemTime,
```

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,117 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
+        let bytes = Self::hex_decode(value).map_err(|_| InvalidRequest::MissingPowParameters)?;
+
+        if bytes.len() != 48 {
+            // 8 + 8 + 32 bytes
+            return Err(InvalidRequest::MissingPowParameters);
+        }
+
+        let difficulty = u64::from_le_bytes(bytes[0..8].try_into().unwrap()) as usize;
+        let timestamp = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
+        let signature: [u8; 32] = bytes[16..48].try_into().unwrap();
```
Mirko-von-Leipzig:
```suggestion
        // SAFETY: Length of 48 is enforced above.
        let difficulty = u64::from_le_bytes(bytes[0..8].try_into().unwrap()) as usize;
        let timestamp = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
        let signature: [u8; 32] = bytes[16..48].try_into().unwrap();
```

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,117 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-    /// The solution is valid if the hash of the seed and the solution has at least `DIFFICULTY`
-    /// leading zeros.
-    pub fn check_pow_solution(
-        &self,
-        challenge_cache: &ChallengeCache,
-    ) -> Result<(), InvalidRequest> {
-        let mut challenges =
-            challenge_cache.challenges.lock().expect("PoW challenge cache lock poisoned");
+    /// Checks if the challenge timestamp is still valid.
+    pub fn is_timestamp_valid(&self) -> bool {
```
Mirko-von-Leipzig: Not the greatest suggestion either, but I'd like to emphasize that its an expiration check.

```suggestion
    pub fn is_not_expired(&self) -> bool {
```

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,90 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-/// It is used to validate the `PoW` challenge and to store the parameters for the `PoW` challenge.
-#[derive(Deserialize)]
-pub(crate) struct PowParameters {
-    pub(crate) pow_seed: String,
-    pub(crate) server_signature: String,
-    pub(crate) server_timestamp: u64,
-    pub(crate) pow_solution: u64,
+/// A challenge for proof-of-work validation.
+#[derive(Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
+pub struct Challenge {
```
bobbinth: nit: I would move this struct into a separate module - e.g., `src/server/pow/challenge.rs`.

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,90 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-            self.server_timestamp,
-            self.difficulty,
-        );
-        if hash != self.server_signature {
-            return Err(InvalidRequest::ServerSignaturesDoNotMatch);
-        }
-        Ok(())
+impl Challenge {
+    /// Creates a new challenge with the given parameters.
+    pub fn new(difficulty: usize, timestamp: u64, signature: [u8; 32]) -> Self {
```
bobbinth: I would probably change this so that it also computes the signature - e.g., would look like so:
```rust
pub fn new(difficulty: usize, timestamp: u64, salt: [u8; 32]) -> Self {
    ...
}
```
And to create a challenge with signature value, I'd just use `Self { difficulty, timestamp, signature }`.

Also, separately, I'd probably rename `salt` to `secret` everywhere.

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,90 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-        if (server_timestamp - self.server_timestamp) > SERVER_TIMESTAMP_TOLERANCE_SECONDS {
-            return Err(InvalidRequest::ExpiredServerTimestamp(
-                self.server_timestamp,
-                server_timestamp,
-            ));
+    /// Decodes the challenge and verifies that the signature part of the challenge is valid
+    /// in the context of the specified salt.
+    pub fn decode(value: &str, salt: [u8; 32]) -> Result<Self, InvalidRequest> {
+        // Add the "0x" prefix to the hex string, otherwise the hex_to_bytes function will fail.
+        let value = "0x".to_string() + value;
```
bobbinth: It feels a bit strange that we need to add `0x` prefix here. Should we change it so that this method expects the `value` to be prefixed with `0x`?

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,90 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-        if challenges.get(&self.pow_seed).is_some() {
-            return Err(InvalidRequest::ChallengeAlreadyUsed);
+        // Verify the signature
+        let expected_signature = Self::compute_signature(salt, difficulty, timestamp);
+        if signature != expected_signature {
+            return Err(InvalidRequest::ServerSignaturesDoNotMatch);
         }
 
-        // Then check the PoW solution
+        Ok(challenge)
```
bobbinth: nit: I would write this as:
```rust
let expected_signature = Self::compute_signature(salt, difficulty, timestamp);
if signature == expected_signature {
    Ok(Self { difficulty, timestamp, signature })
} else {
    Err(InvalidRequest::ServerSignaturesDoNotMatch)
}
```

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,90 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
+    }
+
+    /// Encodes the challenge into a hex string.
+    pub fn encode(&self) -> String {
+        let mut bytes = Vec::with_capacity(48);
+        bytes.extend_from_slice(&(self.difficulty as u64).to_le_bytes());
+        bytes.extend_from_slice(&self.timestamp.to_le_bytes());
+        bytes.extend_from_slice(&self.signature);
+        bytes.to_hex()
+    }
```
bobbinth: Related to one of the above comments: the hex string we output here should probably be prefixed with `0x`.

bin/faucet/src/server/pow.rs
```diff
@@ -34,119 +30,90 @@ const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_
...
-        }
+        let leading_zeros = hash.iter().take_while(|&b| *b == 0).count();
+        leading_zeros >= self.difficulty
+    }
 
-        // If we get here, the solution is valid
-        // Add the challenge to the cache to prevent reuse
-        challenges.insert(self.pow_seed.to_string(), self.server_timestamp);
+    /// Checks if the challenge timestamp is not expired.
+    pub fn is_not_expired(&self) -> bool {
```
bobbinth: I would probably pass in the current time as a parameter.

bin/faucet/src/server/pow.rs
```diff
@@ -222,183 +249,94 @@ impl ChallengeCache {
 
 #[derive(Serialize)]
 struct PoWResponse {
-    seed: String,
+    challenge: String,
     difficulty: usize,
-    server_signature: String,
     timestamp: u64,
 }
```
bobbinth: Why do we need this `PoWResponse` struct? Isn't everything we need already encoded in the `challenge` string?

bin/faucet/src/server/get_tokens.rs
```diff
@@ -178,13 +174,12 @@ impl RawMintRequest {
...
-            pow_parameters.check_server_signature(&server.pow.salt)?;
-        } else {
-            return Err(InvalidRequest::MissingPowParameters);
-        }
+        // Validate Challenge and nonce
+        let challenge_str = self.challenge.ok_or(InvalidRequest::MissingPowParameters)?;
+        let nonce = self.nonce.ok_or(InvalidRequest::MissingPowParameters)?;
+
+        let challenge = Challenge::decode(&challenge_str, server.pow.salt)?;
+        server.pow.submit_challenge(&challenge, nonce)?;
```
bobbinth: nit: I would maybe maybe have `submit_challenge()` method on the `server` which would take challenge string, decode it, and then call `submit_challenge()` on the internal `pow` instance.

bobbinth (commented):
Looks good! Thank you. I left some comments inline.

bin/faucet/src/server/pow.rs
SantiagoPittella: in the faucet PoW we do not use the prefix, this was added so we can use the the `hex_to_bytes::<48>` function from `miden-crypto`. I would rather prefer to not use the "0x" since it will add will need to trim it to calculate the leading zeros. I will try to use another way to convert it to bytes to avoid this concatenation.

bin/faucet/src/server/pow.rs
SantiagoPittella: It does not, to get the hex with the prefix it needs to be replaced with `.to_hex_with_prefix`. wdyt? i didn't do it on the first place I thought that wasn't needed.

bin/faucet/src/server/get_tokens.rs
SantiagoPittella: I don't think that this decoupling is right because it would cause fragmentation of the context. wdyt?

bin/faucet/src/server/pow.rs
SantiagoPittella: I ended up removing it and replacing it with `Challenge` + an impl of `Serialize` where we encode the challenge using the `encode` method.

bin/faucet/src/server/challenge.rs
```diff
@@ -0,0 +1,136 @@
...
+use serde::{Deserialize, Serialize, Serializer};
+use sha3::{Digest, Sha3_256};
+
+use super::get_tokens::InvalidRequest;
+
+/// The tolerance for the server timestamp.
+///
+/// The server timestamp is valid if it is within `SERVER_TIMESTAMP_TOLERANCE_SECONDS` seconds of
+/// the current time.
+pub(crate) const SERVER_TIMESTAMP_TOLERANCE_SECONDS: u64 = 30;
```
bobbinth: nit: this is not so much a "tolerance" as a lifetime of a challenge. So, I'd probably rename it into `CHALLENGE_LIFETIME_SECONDS`.

bin/faucet/src/server/challenge.rs
```diff
@@ -0,0 +1,136 @@
...
+        let hash = hasher.finalize();
+
+        let leading_zeros = hash.iter().take_while(|&b| *b == 0).count();
+        leading_zeros >= self.difficulty
+    }
+
+    /// Checks if the challenge timestamp is not expired.
+    pub fn is_not_expired(&self, current_time: u64) -> bool {
+        (current_time - self.timestamp) <= SERVER_TIMESTAMP_TOLERANCE_SECONDS
+    }
```
bobbinth: nit: `is_not_expired()` sounds a bit odd. I would probably change to `is_expired()` and update the logic accordingly.

bin/faucet/src/server/challenge.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Computes the signature for a challenge.
+    pub fn compute_signature(secret: [u8; 32], difficulty: usize, timestamp: u64) -> [u8; 32] {
```
bobbinth: does this need to be public?

bin/faucet/src/server/mod.rs
```diff
@@ -84,11 +85,13 @@ impl Server {
...
-        let pow = PoW {
-            salt: pow_salt,
-            difficulty: Arc::new(AtomicUsize::new(1)), // Initialize difficulty to 1
-            challenge_cache,
-        };
+        // Convert string secret to [u8; 32] for PoW
+        let mut secret_bytes = [0u8; 32];
+        let secret_str_bytes = pow_secret.as_bytes();
+        let copy_len = secret_str_bytes.len().min(32);
+        secret_bytes[..copy_len].copy_from_slice(&secret_str_bytes[..copy_len]);
```
bobbinth: This means that `pow_secret` can be any length, but we only use the first 32 bytes of it as the secret, right? For the testnet faucet this is probably fine, but what I'd do for production is hash the string and use the hash as the secret.

bin/faucet/src/server/pow.rs
```diff
@@ -28,144 +25,80 @@ const MAX_DIFFICULTY: usize = 24;
...
+    /// Returns an error if:
+    /// * The challenge is expired.
+    /// * The challenge is invalid.
+    /// * The challenge was already used.
+    pub fn submit_challenge(
         &self,
-        challenge_cache: &ChallengeCache,
+        challenge: &Challenge,
+        nonce: u64,
     ) -> Result<(), InvalidRequest> {
```
bobbinth: nit: I would rename this to `validate_challenge()`.

Also, maybe we should change this to accept the challenge string and do the decoding internally. Then, `secret` can remain encapsulated within the `PoW` struct. The signature would look like so:
```rust
pub fn validate_challenge(
    &self,
    challenge: &str,
    nonce: u64,
) -> Result<(), InvalidRequest> {
    ...
}
```

bin/faucet/src/server/pow.rs
```diff
@@ -220,185 +153,86 @@ impl ChallengeCache {
...
-        server_signature,
-        timestamp,
-    })
+/// Get a challenge to be used by a client for `PoW`.
+pub(crate) async fn get_pow_challenge(State(server): State<Server>) -> impl IntoResponse {
+    let challenge = server.pow.build_challenge();
+    Json(challenge)
 }
 
 #[cfg(test)]
```
bobbinth, line 168: nit: let's add "TESTS" section header above this line.

bin/faucet/src/server/pow.rs
```diff
@@ -220,185 +153,86 @@ impl ChallengeCache {
...
-    Json(PoWResponse {
-        seed: random_seed,
-        difficulty: server.pow.difficulty.load(Ordering::Relaxed),
-        server_signature,
-        timestamp,
-    })
+/// Get a challenge to be used by a client for `PoW`.
+pub(crate) async fn get_pow_challenge(State(server): State<Server>) -> impl IntoResponse {
+    let challenge = server.pow.build_challenge();
+    Json(challenge)
```
bobbinth: Do we actually need this function? It is pretty simple and these two lines could probably be copied to its callsite.

Also, conceptually, `PoW` doesn't need to be aware that it is being used in a server context (we could even move it to the `utils` crate later on).

bobbinth (commented):

bin/faucet/src/server/get_tokens.rs
bobbinth: What I'm trying to achieve is to make `pow` an internal concern of the `server`. That is, ideally, `pow` would be a private field of the `server` struct and we won't be accessing it directly.

bin/faucet/src/server/pow.rs
bobbinth: We could go either way, but it's pretty easy to output the string here with `0x` prefix (using `.to_hex_with_prefix()` as you mentioned), and then we can decode the challenge pretty easily without the need to append `0x` in front.

It also makes the string a bit more "self-describing" (i.e., by looking at it, we can tell that it is hex-encoded).

bin/faucet/src/server/pow.rs
SantiagoPittella: It was previously called like that, and Mirko suggested to change it to `submit_*` because we are marking the challenge as "used" and we mark it as invalid, and not just validation.

Regarding the `secret` part, I agree, i'm sending a change.

bobbinth: > It was previously called like that, and Mirko suggested to change it to `submit_*` because we are marking the challenge as "used" and we mark it as invalid, and not just validation.

Yep - fine to keep it as `submit_*`.

bin/faucet/src/server/challenge.rs
SantiagoPittella: No, it does not. I removed the `pub` tag.

bin/faucet/src/server/get_tokens.rs
SantiagoPittella: Oh, I think i understand now. Just pushed a commit changing that. lmk if that was what you meant.

bin/faucet/src/server/challenge.rs
```diff
@@ -0,0 +1,161 @@
...
+        hasher.finalize().into()
+    }
+
+    /// Validates the challenge and nonce.
+    ///
+    /// # Errors
+    /// Returns an error if:
+    /// * The challenge is expired.
+    /// * The challenge is invalid.
+    pub fn validate(&self, nonce: u64) -> Result<(), InvalidRequest> {
```
bobbinth: I probably wouldn't add this function - or if we do add it, I would:

- Pass the time in as a parameter.
- Use a different error type (not `InvalidRequest` but create a new error enum specifically for `PoW`).

Basically, I'm thinking about this through the lens of moving `PoW` components to the `utils` crate eventually.

bin/faucet/src/server/challenge.rs
```diff
@@ -0,0 +1,175 @@
...
+/// The lifetime of a challenge.
+///
+/// A challenge is valid if it is within `CHALLENGE_LIFETIME_SECONDS` seconds of the current time.
+pub(crate) const CHALLENGE_LIFETIME_SECONDS: u64 = 30;
+
+/// A challenge for proof-of-work validation.
+#[derive(Debug, Clone, Deserialize, Hash, PartialEq, Eq)]
+pub struct Challenge {
+    pub(crate) difficulty: usize,
+    pub(crate) timestamp: SystemTime,
```
bobbinth: What's the motivation for switching to `SystemTime`? I think keeping it a simple `u64` is fine and I'm also not sure we need to keep sub-second granularity.

bin/faucet/src/server/challenge.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        if bytes.len() != 52 {
+            // 8 + 8 + 4 + 32 bytes
+            return Err(InvalidRequest::MissingPowParameters);
+        }
+
+        // SAFETY: Length of 52 is enforced above.
+        let difficulty = u64::from_le_bytes(bytes[0..8].try_into().unwrap()) as usize;
+        let secs = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
+        let nanos = u32::from_le_bytes(bytes[16..20].try_into().unwrap());
```
bobbinth: Related to the previous comment: do we really need to keep track of nanoseconds? I think just seconds should be fine (unless I'm missing something).

bin/faucet/src/server/mod.rs
```diff
@@ -246,6 +289,12 @@ impl FromRef<Server> for GetTokensState {
     }
 }
 
+impl FromRef<Server> for PoW {
+    fn from_ref(input: &Server) -> Self {
+        input.pow.clone()
+    }
+}
```
bobbinth: This is OK because cloning `PoW` is cheap, right? If so, wonder if there is a way to make this more explicit somehow.

bin/faucet/src/server/mod.rs
```diff
@@ -232,6 +238,43 @@ impl Server {
...
+            return Err(InvalidRequest::InvalidPoW);
+        }
+
+        // Check if challenge was already used
+        if !self.pow.challenge_cache.challenges.lock().unwrap().insert(challenge.clone()) {
+            return Err(InvalidRequest::ChallengeAlreadyUsed);
+        }
+
+        Ok(())
+    }
```
bobbinth: I would move most of this method into `PoW` struct. The signature there could look like:
```rust
pub(crate) fn submit_challenge(
    &self,
    timestamp: u64,
    challenge: &str,
    nonce: u64,
) -> Result<(), InvalidRequest> {
    
}
```
The method here then would mostly just forward the call to the `PoW` struct and could look like so:
```rust
pub(crate) fn submit_challenge(
    &self,,
    challenge: &str,
    nonce: u64,
) -> Result<(), InvalidRequest> {
    let timestamp = /* get timestamp */
    self.pow.submit_challenge(timestamp, challenge, nonce)
}
```

bin/faucet/src/server/pow.rs
```diff
@@ -28,144 +20,40 @@ const MAX_DIFFICULTY: usize = 24;
...
-    pub(crate) pow_seed: String,
-    pub(crate) server_signature: String,
-    pub(crate) server_timestamp: u64,
-    pub(crate) pow_solution: u64,
-    pub(crate) difficulty: usize,
+#[derive(Clone)]
+pub(crate) struct PoW {
+    pub(crate) secret: [u8; 32],
+    pub(crate) difficulty: Arc<AtomicUsize>,
+    pub(crate) challenge_cache: ChallengeCache,
```
bobbinth: let's make all fields here private.

bin/faucet/src/server/pow.rs
```diff
@@ -179,36 +67,40 @@ impl PoW {
...
 /// is only used once.
-/// Challenges gets removed periodically.
+/// Challenges get removed periodically.
 #[derive(Clone, Default)]
-pub struct ChallengeCache {
+pub(crate) struct ChallengeCache {
     /// Once a challenge is added, it cannot be submitted again.
-    challenges: Arc<Mutex<HashMap<String, u64>>>,
+    pub challenges: Arc<Mutex<HashSet<Challenge>>>,
 }
```
bobbinth: Let's make the struct and the field in it private.

bobbinth (commented):
Thank you! Looks good. I left some more comments inline.

Let's try to wrap up this PR.

bin/faucet/src/server/challenge.rs
SantiagoPittella: Mirko suggested it in, to check if it simplifies things. Watching now the results, I don't think it does. I'm planning to rollback this change.

bin/faucet/src/server/mod.rs
SantiagoPittella: It is indeed a cheap clone, because it is basically the 32 bytes arrays and increasing two arc ref counters. I will add a comment to let this clear, maybe that clarifies things.

bobbinth (approved):
Looks good! Thank you!

PR #924, opened 2025-06-04
feat: make PoW difficulty more granular

TomasArrachea:


bobbinth:
I would probably hold off on this one until #923 is merged.

bin/faucet/src/server/challenge.rs
```diff
@@ -15,12 +16,12 @@ use crate::server::ApiKey;
...
 
 /// The size of the encoded challenge in bytes.
-const CHALLENGE_ENCODED_SIZE: usize = 95;
+const CHALLENGE_ENCODED_SIZE: usize = 119;
 
 /// A challenge for proof-of-work validation.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Challenge {
-    pub(crate) difficulty: usize,
+    pub(crate) target: BigUint,
```
bobbinth: Do we need to rename this? I think `difficulty` works for this approach just as well. If we do need to rename it, I would leave a comment what `target` means because it is not immediately obvious.  (actually, it would be good to leave the comment regardless).

Also, do we really need to make this a `BigUnit`? I though something like `u64` should work since difficulty that sets the topmost 8 bytes to zeros is not really that would practically be possible.

bin/faucet/src/server/challenge.rs
```diff
@@ -107,25 +111,29 @@ impl Challenge {
...
-    /// challenge.
+    /// Checks whether the provided nonce satisfies the target requirement encoded in the challenge.
+    /// The solution is valid if the hash H(challenge, nonce), interpreted as a number, is lower
+    /// than the target value.
     pub fn validate_pow(&self, nonce: u64) -> bool {
         let mut hasher = Sha3_256::new();
         hasher.update(self.encode());
         hasher.update(nonce.to_le_bytes());
         let hash = hasher.finalize();
+        let number = BigUint::from_bytes_be(&hash);
```
bobbinth: Similar to the last comment: I'm not sure we need the `BigUint`. I think we can just compare the top 8 bytes as a `u64` and that should be sufficient.

bin/faucet/src/main.rs
```diff
@@ -95,6 +98,18 @@ pub enum Command {
...
+        /// difficulty by 1.
+        #[arg(long = "pow-requests-per-difficulty-level", value_name = "NON_ZERO_USIZE", env = ENV_POW_REQUESTS_PER_DIFFICULTY_LEVEL, default_value = "8")]
+        pow_requests_per_difficulty_level: NonZeroUsize,
+
+        /// The initial target shift for the `PoW` challenges. This sets the base difficulty for
+        /// all challenges. It must be between 0 and 32. A bigger shift means higher
+        /// difficulty.
+        #[arg(value_parser = clap::value_parser!(u8).range(0..=32))]
+        #[arg(long = "pow-initial-target-shift", value_name = "U8", env = ENV_POW_INITIAL_TARGET_SHIFT, default_value = "12")]
+        pow_initial_target_shift: u8,
```
bobbinth: I would name these parameters a bit differently:

- `pow-baseline` - I would set the default to 8 bits.
- `pow-grow-rate` - I would set keep the default at 8 requests per bit.

bin/faucet/src/server/pow.rs
```diff
@@ -55,21 +56,26 @@ impl PoW {
...
-    /// requests, and it is clamped between 1 and `MAX_DIFFICULTY`.
+    /// The difficulty is adjusted based on the number of active requests per API key.
     fn get_difficulty(&self, api_key: &ApiKey) -> usize {
         let num_challenges = self
             .challenge_cache
             .lock()
             .expect("challenge cache lock should not be poisoned")
             .num_challenges_for_api_key(api_key);
-        (num_challenges / ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY).clamp(1, MAX_DIFFICULTY)
+        num_challenges / self.requests_per_difficulty_level
```
bobbinth: I'd probably compute the full difficulty here (including incorporating the baseline). I think it would be something like `roundup(num_challenges / pow_growth_rate)^baseline`.

Also, `get_difficulty()` should probably return a `u64`.

bin/faucet/src/server/challenge.rs
TomasArrachea: I find naming it `difficulty` a bit missleading. A higher target means an easier challenge, so it would be confusing if a higher difficulty meant an easier challenge.

TomasArrachea: Based on the formula you commented below, I think maybe you were thinking it the opposite way as I was. Like, the challenge would be valid if: `H(challenge, nonce) > difficulty`, right? This way I think it does make sense.

bobbinth: I actually didn't think it through too deeply. It does seems like `get_msl(H(challenge, nonce)) < difficulty` is the right way to go (where `get_msl` returns the most significant 64 bits of the digest converted to `u64`).

TomasArrachea: So we should change the attribute name to 'target', right?

TomasArrachea: Or perhaps keep the name as `difficulty` and before making the validation we do `u64::MAX - difficulty`

bobbinth: I'm fine with `target` but let's make sure we add comments to explain what it means.

bin/faucet/src/server/pow.rs
TomasArrachea: I moved calculation of the `target` into this `get_challenge_target` function. The target is computed as:
`max_target / (difficulty + 1)`
Where:
- `max_target = u64::MAX >> baseline`
- `difficulty = num_active_challenges / growth_rate`

I'm not sure where the formula you wrote comes from, but I think it differs from this.

bin/faucet/src/server/challenge.rs
```diff
@@ -17,10 +17,17 @@ const CHALLENGE_ENCODED_SIZE: usize = 95;
 /// A challenge for proof-of-work validation.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Challenge {
-    pub(crate) difficulty: usize,
+    /// The target used to validate the challenge solution. A lower target makes the challenge more
```
SantiagoPittella, line 20: It might be worth to clarify that isn't the whole hash but rather the first 8 bytes

SantiagoPittella (approved):
Looking good, just a small comment

bin/faucet/.env
```diff
@@ -9,5 +9,7 @@ MIDEN_FAUCET_ASSET_AMOUNTS=
 MIDEN_FAUCET_REMOTE_TX_PROVER_URL=
 MIDEN_FAUCET_POW_SECRET=
 MIDEN_FAUCET_POW_CHALLENGE_LIFETIME=
+MIDEN_FAUCET_POW_GROWTH_RATE=
+MIDEN_FAUCET_POW_BASELINE=
 MIDEN_FAUCET_API_KEYS=
```
bobbinth: Not related to this PR, but it feels a bit odd to have these blank entries in the config files. Is there a downside to populating them with default values?

cc @Mirko-von-Leipzig @sergerad.

bin/faucet/src/server/challenge.rs
```diff
@@ -17,10 +17,17 @@ const CHALLENGE_ENCODED_SIZE: usize = 95;
...
+    /// difficult to solve. A solution is valid if the hash `H(challenge, nonce)`, interpreted as a
+    /// little-endian u64 from the first 8 bytes, is less than this target value.
+    pub(crate) target: u64,
```
bobbinth: Not sure it matters, but shouldn't it technically be big-endian? Conceptually, if we imagine that the value is a 256-bit integer in big-endian form, then we'd take the first 8 bytes in big-endian form to get the most significant limb of the integer.

If we imagine the 256-bit integer in little-endian form, then we can take the last 8 bytes in little-endian form.

But again, not sure if this has practical significance.

bin/faucet/src/server/challenge.rs
```diff
@@ -104,25 +111,26 @@ impl Challenge {
...
         bytes.extend_from_slice(&self.signature);
         bytes.to_hex_with_prefix()
     }
 
-    /// Checks whether the provided nonce satisfies the difficulty requirement encoded in the
-    /// challenge.
+    /// Checks whether the provided nonce satisfies the target requirement encoded in the
+    /// challenge. The solution is valid if the hash `H(challenge, nonce)`, interpreted as a
+    /// little-endian u64 from the first 8 bytes, is lower than the target value.
     pub fn validate_pow(&self, nonce: u64) -> bool {
```
bobbinth: nit: let's add a line break after the intro sentence.

bin/faucet/src/server/pow.rs
```diff
@@ -58,23 +51,30 @@ impl PoW {
...
         let num_challenges = self
             .challenge_cache
             .lock()
             .expect("challenge cache lock should not be poisoned")
             .num_challenges_for_api_key(api_key);
-        (num_challenges / ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY).clamp(1, MAX_DIFFICULTY)
+
+        let difficulty = num_challenges / self.config.growth_rate.get();
+        let max_target = u64::MAX >> self.config.baseline;
+        max_target / (difficulty as u64 + 1)
```
bobbinth: I'm not sure this formula works. Assuming I did my math correctly (but please double-check), the issues is that the higher the `growth_rate` the smaller PoW requirement becomes:

- With `base_line=8` and `growth_rate=1`, the PoW at 4K requests would be ~20 bits.
- With `base_line=8` and `growth_rate=8`, the PoW at 4K requests would be ~17 bits.

There are probably many ways to fix the formula. One approach is to do it as follows:
```rust
let max_target = u64::MAX >> self.config.baseline;
let difficulty = num_requests << self.config.growth_rate.get();
max_target / (difficulty as u64 + 1)
```
Again, assuming I did my math correctly, `base_line=8` and `growth_rate=1` this would roughly mean that difficulty would double as the number of requests doubles, and at 4K requests the PoW would also be ~20 bits. But with `growth_rate=2`, PoW at 4K requests would be ~32 bits.

There are probably some improvements we can make to this methodology (e.g., make PoW requirements less sensitive to growth rate, cap them at some reasonable value - e.g., 24 bits etc.), but overall, we want that does the following:

- If there are few requests (e.g., less than 1 per second), PoW requirements should be negligible - i.e., under 1 second which I think should be around 16 - 18 bits, but let's double check.
- If there is a moderate number of requests (e.g., 2 - 3 per second), PoW should take under 5 seconds (which I think would be 20 - 21 bits).
- If there are more than 3 - 4 requests per second, the PoW should take close to 30 seconds (my guess is around 24 bits).

bobbinth (commented):
Thank you! Looks good. I left a few comments, the main one is about PoW formula. Let's make sure we sanity-check it and also test how much time different levels of PoW take  in the browser.

bin/faucet/src/server/challenge.rs
TomasArrachea: Yes, in practice both ways should work but I can see why big-endian makes a bit more sense.

bin/faucet/src/server/pow.rs
TomasArrachea: > I'm not sure this formula works. Assuming I did my math correctly (but please double-check), the issues is that the higher the growth_rate the smaller PoW requirement becomes:
> - With base_line=8 and growth_rate=1, the PoW at 4K requests would be ~20 bits.
> - With base_line=8 and growth_rate=8, the PoW at 4K requests would be ~17 bits.

I got same results. I think the missunderstanding comes from `growth_rate` naming. It is described as: 
```
/// The number of simultaneous active requests needed to increase the `PoW` challenges
/// difficulty. This sets how fast the difficulty increases with requests.
```

A higher growth rate means that more requests are needed to increase the PoW difficulty, so it makes sense that it makes the PoW easier for the same amount of requests. We should definitely change this though.

bobbinth: Ah yes, makes sense. The naming is indeed confusing (and I think I suggested it :))

Putting the naming aside, I think with this methodology it would be pretty difficult to get the targets I described in the previous comment to get PoW to 24 bits once we have 100 - 200 solved challenges would probably require setting a pretty high baseline (though I haven't done the math myself).

In either case, I think we should try to measure the actual time in takes in the browser to solve some of these PoW targets and try to go from there.

TomasArrachea: I took some measurements running on the browser and got:
Leading Zeroes | Average Duration (s)
-- | --
12 | 0.0294
13 | 0.1161
14 | 0.2428
15 | 0.1795
16 | 0.9664
17 | 1.1204
18 | 2.2548
19 | 4.0217
20 | 17.3793
21 | 21.7740
22 | 51.8881

If we set baseline at 14 and growth rate at 1, the progression (with current formula) would be as follows:
Num challenges | Leading Zeroes 
-- | -- 
0 | 14
1 | 15
30 | 18
60 | 19 
100 | 20
150 | 21 
300 | 22

Which I think fits roughly to what we want.

TomasArrachea: Regarding the `growth_rate` I think we can either rename it to something like `challenges_per_difficulty` and set it as a f64 to allow it being smaller than one for more steep increases in difficulty. Or, we can keep the naming and refactor the formula into something like what you proposed above.

bobbinth (approved):
Looks good! Thank you! I pushed a small commit that changes the formula to be similar to what I listed in https://github.com/0xMiden/miden-node/pull/924#discussion_r2188734125 and also updated the defaults to roughly get what you shared in https://github.com/0xMiden/miden-node/pull/924#discussion_r2192838502. I think this should be OK for now, and we can adjust/improve on this in the new repo.

PR #926, opened 2025-06-04
chore: Add Proving Service binary crate

### LINKED ISSUE #925: Move proving-service source to mide-node and integrate into workflows

sergerad:
WIP.

Closes #925.


sergerad:
@SantiagoPittella @igamigo do you guys have any idea why the `maybe_async` macro is causing this after moving the crate to this workspace?
```
error[E0599]: no method named `map_err` found for struct `Pin<Box<dyn Future<Output = Result<ProvenTransaction, ...>>>>` in the current scope
   --> bin/proving-service/src/api/prover.rs:76:14
    |
72  |           let proof = prover
    |  _____________________-
73  | |             .try_lock()
74  | |             .map_err(|_| Status::resource_exhausted("Server is busy handling another request"))?
75  | |             .prove(transaction_witness)
76  | |             .map_err(internal_error)?;
    | |             -^^^^^^^ method not found in `Pin<Box<dyn Future<Output = Result<ProvenTransaction, ...>>>>`
    | |_____________|
    |
    |
   ::: /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/futures-util-0.3.31/src/future/try_future/mod.rs:308:8
    |
308 |       fn map_err<E, F>(self, f: F) -> MapErr<Self, F>
    |          ------- the method is available for `Pin<Box<dyn std::future::Future<Output = std::result::Result<ProvenTransaction, TransactionProverError>>>>` here
    |
    = help: items from traits can only be used if the trait is in scope
help: trait `TryFutureExt` which provides `map_err` is implemented but not in scope; perhaps you want to import it
    |
1   + use futures_util::future::try_future::TryFutureExt;
```

tomyrd:
> do you guys have any idea why the maybe_async macro is causing this after moving the crate to this workspace?

In my experience this error occurs because of the `miden-tx` `async` feature. That feature decides whether the `TransactionProver` is `async` or not. In the faucet we have that feature enabled and in the proving service we do not.

The problem stems from the fact that when building the whole workspace, cargo will compile crates only once with all the combined features. In this case, if we have one crate with async and another without it, the `miden-tx` feature will compile with the async feature on. Normally this wouldn't be a problem because cargo features should be additive (for more context see [this section of the cargo book](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification). In our case, the `async` feature is not additive. Having it enabled or not will modify the same global trait (because the crate is compiled once).

To fix this you need to compile conflicting crates separately by using the `--exclude` flag in cargo commands (see the [Makefile changes in this PR](https://github.com/0xMiden/miden-base/pull/1424/files) or in [this client PR](https://github.com/0xMiden/miden-client/pull/950/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52) where the problem was worse). It's not as pretty but I haven't found another way to solve this.

SantiagoPittella:
Sorry, I've missed the previous comment. Let me know if you have any further question on Tomas's comment. Also, in order to start moving the PRs related to this crate from `miden-base` here, what commit was used?

.github/actions/deb_node_faucet/action.yml
bobbinth, line 1: Question: do we have to have separate directories for all actions? Could we not have just different names - e.g.:

- `.github/actions/deb_node.yml`
- `.github/actions/deb_faucet.yml`
- `.github/actions/deb_proving_service.yml`

Or we could have a single directory for all debian actions:

- `.github/actions/debian/node.yml`
- `.github/actions/debian/faucet.yml`
- `.github/actions/debian/proving_service.yml`

(I'm assuming we'll need something similar for Docker?)

bin/proving-service/proto/worker_status.proto
bobbinth, line 1: Eventually (assuming we don't split out proving service into a separate crate), we'll probably need to move `.proto` files to the root proto crate. But I wouldn't tackle this in this PR.

.github/actions/deb_node_faucet/action.yml
sergerad: Yea I think the action.yml approach is only required when you share the action to other repositories. Which we will not be doing here.

sergerad: > (I'm assuming we'll need something similar for Docker?)

There is a workflow for Docker, no action required atm.

sergerad: BTW I have mentioned under future work to do consolidate the Debian actions. Just keeping them as is for now to unblock the others from working on proving service

sergerad: I have consolidated them for this PR. I will need a subsequent PR to make sure they work properly however.

packaging/miden-prover/miden-prover.service
bobbinth, line 1: Should we name this file (and other related artifacts) `miden-prover-worker.service`? It feels a bit strange that the worker is names just "miden prover" while the proxy is named "miden prover proxy" (it may actually be a bit more natural to use "miden prover" name for the proxy).

bin/proving-service/Cargo.toml
```diff
@@ -0,0 +1,71 @@
...
+
+[features]
+concurrent = ["miden-tx/concurrent"]
+default    = ["concurrent"]
+
+[lints]
+workspace = true
+
+[dependencies]
+async-trait                        = "0.1"
```
bobbinth: nit: for this and other similar dependencies, we usually say `{ version = "0.1" }` instead of `"0.1"`.

crates/proving-service-client/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+
+[target.'cfg(not(all(target_arch = "wasm32", target_os = "unknown")))'.dependencies]
+tonic     = { default-features = false, features = ["codegen", "prost", "transport"], version = "0.12" }
+tonic-web = { optional = true, version = "0.12" }
+
+[lints]
+workspace = true
+
+[dependencies]
+async-trait   = "0.1"
```
bobbinth: Similar nit as above.

bobbinth (approved):
Looks good! Thank you! Not a very deep review from me - but I left a couple of small comments inline.

Also, I didn't really check most of the Rust files as I assume they are just a straight copy from the `miden-base` repo.

Cargo.toml
SantiagoPittella, line 1: We are still referencing `miden-base`'s `miden-proving-service-client` here.

packaging/miden-prover/miden-prover.service
sergerad: I'll make a followup task to rename miden-prover to miden-prover-worker

sergerad:
> Tried installing the proving service through the makefile and received some error coming from the version of miden-base.

I'm getting this
```
error[E0616]: field `stack` of struct `AdviceInputs` is private
   --> /Users/sergeradinovich/.cargo/git/checkouts/miden-base-44d309b3731cdbb1/03569de/crates/miden-lib/src/transaction/inputs.rs:370:16
    |
370 |         self.0.stack.extend(iter);
    |                ^^^^^ private field
    |
help: a method `stack` also exists, call it with parentheses
    |
370 |         self.0.stack().extend(iter);
    |                     ++

error[E0616]: field `store` of struct `AdviceInputs` is private
   --> /Users/sergeradinovich/.cargo/git/checkouts/miden-base-44d309b3731cdbb1/03569de/crates/miden-lib/src/transaction/inputs.rs:376:16
    |
376 |         self.0.store.extend(iter);
    |                ^^^^^ private field
    ```

Which is odd because miden-node is importing next for all base crates afaict.

File: crates/ntx-builder/src/builder/store.rs [hunk 3]
```diff
@@ -51,6 +51,7 @@ impl StoreClient {
     }
 
     /// Returns the latest block's header from the store.
+    #[allow(dead_code)]
     #[instrument(target = COMPONENT, name = "store.client.latest_header", skip_all, err)]
     pub async fn latest_header(&self) -> Result<BlockHeader, StoreError> {
```
PhilippGackstatter, line 56: Should there be a comment added why that lint is allowed?

bin/proving-service/Cargo.toml
```diff
@@ -0,0 +1,71 @@
...
+description            = "Miden blockchain proving service"
+edition.workspace      = true
+homepage.workspace     = true
+keywords               = ["miden", "proving", "service"]
+license.workspace      = true
+name                   = "miden-proving-service"
+readme                 = "README.md"
+repository.workspace   = true
+rust-version.workspace = true
+version.workspace      = true
```
PhilippGackstatter, line 12: I'm guessing it is intentional that the version is the same across the node repo?

crates/proving-service-client/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+authors.workspace      = true
+description            = "Client library for the Miden blockchain proving service"
+edition.workspace      = true
+keywords               = ["miden", "proving", "service"]
+license.workspace      = true
+name                   = "miden-proving-service-client"
+readme                 = "README.md"
+repository.workspace   = true
+rust-version.workspace = true
+version                = "0.10.0"
```
PhilippGackstatter: For the `proving-service` this is `version.workspace = true`, why is it explicit here?

docs/src/developer/codebase.md
```diff
@@ -6,17 +6,19 @@ component, a couple of gRPC-related codegen crates, and a catch-all utilities cr
...
+| Crate                     | Description                                                                                                                                              |
+| ------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `node`                    | The node executable. Configure and run the node and its components.                                                                                      |
+| `faucet`                  | A reference faucet app implementation used by the official Miden faucet.                                                                                 |
+| `proving-service`         | Delegated proving executables. Includes workers and proxies.                                                                                             |
+| `proving-service-client ` | Delegated proving client implementation.                                                                                                                 |
```
PhilippGackstatter:
```suggestion
| `proving-service-client` | Delegated proving client implementation.                                                                                                                 |
```

bin/node/.env
PhilippGackstatter, line 1: This file (and `bin/faucet/.env`) is  empty, so I guess it's needed for some tool that expects `.env` files?

crates/ntx-builder/src/builder/store.rs
bobbinth: This lint appeared only after we've run `cargo update` - but given the scope of this PR, I decided not to delete this function, but TBH, not sure why it is needed here. I know that @Mirko-von-Leipzig is working on some related functionality - so, assuming this will be addressed in one of the follow up PRs.

sergerad: I actually don't know if we want to keep this code or not. Who would know?

crates/proving-service-client/Cargo.toml
sergerad: Mistakenly undid change to make this workspace equivalent

bin/node/.env
sergerad: Added comment ty

sergerad:
> Are there any follow-ups that we'd like to address in future PRs which are not captures in issues yet? (e.g., handling of `.proto` files).

The proving service proto files are already integrated into `proto/proto`. Is that what you mean?

SantiagoPittella (approved):
Just tested it, and looks like it;s working now 👍🏼 regarding the debian package, I looked at it quickly and it seemed fine.

tomyrd (commented):
I tried using this branch [in the client](https://github.com/0xMiden/miden-client/pull/950) but it seems that the `RpcListener` and other structs aren't being exported anymore from `miden-proving-service` (as it happened in https://github.com/0xMiden/miden-base/pull/1424). Should we add this in a follow-up PR?

SantiagoPittella:
@tomyrd I think that your closed PR in base should be reopened in this repo using serge's branch as base.

SantiagoPittella:
The check of the `book` workflow is now fix in `next`.

tomyrd:
> I think that your closed PR in base should be reopened in this repo using serge's branch as base.

Oh, makes sense! I can open it once this gets merged.

PR #928, opened 2025-06-05
refactor: migrate to `diesel-rs` for query construction

### LINKED ISSUE #921: [Task]: trial run `diesel.rs` as thin ORM for `store`

- [ ] RPC
- [ ] Block producer
- [x] Store
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

Provide a demo of `diesel.rs` within the `store` as alternative to the current impl

### How should it be done?

Implement a `diesel.rs` based query for a single endpoint.

### When is this task done?

Once a PR is opened with the relevant demo. Subsequent feedback is out of scope for this task, a followup task is to be created.

### Additional context

Status quo:

The store consists of an RPC service and a wrapper on-top of SQLITE, resembling a thin ORM layer. There is a lot of manual work involved maintaining these abstractions and it will not become any easier once we enter the realm of database schema migrations and rust type versioning.

Proposal:

`diesel.rs` comes with a lot of features, i.e. schema migration support, embedding of those, typed queries and others, that would remove code from our maintenance envelope. Diesel has been around for 5+ years and is very mature, so the SBOM is not incurring any meaningful additional risk.

Predicted Effects:

* reduce boilerplate
* reduce amount of code to maintain by us
* allow for postgresql dropin if we ever run into bottlenecks
* will _not_ solve the general issue of rust type versioning

drahnr:
A example of using `diesel.rs` as an ORM for a single query.

# Learnings

* Caveats
    - types in `models.rs` need to align precisely with the schema generated from the database in `schema.rs` - otherwise one gets lengthy errors - but at least at compile time
    - filters still require type conversions to the native DB representation
    - requires an intermediate type to represent a row (it's good, but sometimes might get in the way
    - additional complexity

* Upsides
   - More structured / a little less boilerplate
   - No inline SQL (if desired)
   - Drop in for postgres possible

# Outstanding

* [x] Understand how to load records partially
* [ ] Migration integration - deferred

Closes #921

### DIFF
--- a/crates/store/src/db/schema.rs
+++ b/crates/store/src/db/schema.rs
@@ -128,6 +128,5 @@ diesel::allow_tables_to_appear_in_same_query!(
     note_scripts,
     notes,
     nullifiers,
-    settings,
     transactions,
 );


crates/store/src/db/mod.rs
```diff
@@ -436,17 +459,30 @@ impl Db {
...
+        use diesel::SelectableHelper;
+
+        let conn = self.diesel.get().await?;
+        let entries = conn
             .interact(move |conn| {
-                let transaction = conn.transaction()?;
-                sql::select_accounts_by_ids(&transaction, &account_ids)
+                let account_ids = Vec::from_iter(
+                    account_ids.iter().map(|account_id| account_id.to_bytes().clone()),
+                );
```
Mirko-von-Leipzig: Is it possible to use the `AccountId` type directly, or must one always clone data to form primitive data types?

crates/store/src/db/mod.rs
```diff
@@ -436,17 +459,30 @@ impl Db {
...
+
+                let filter = schema::accounts::account_id.eq_any(&account_ids[..]);
+                let stmt = schema::accounts::table.filter(filter);
+                let accounts_raw = stmt
+                    .select(models::AccountInfoRawRow::as_select())
+                    .load::<models::AccountInfoRawRow>(conn)?;
```
Mirko-von-Leipzig: How would one select just a subset of the columns? The original skipped the `network_account_id_prefix` whereas this model includes it.

Would that require a new model basically with just the desired fields? If so, would it make sense to return the model directly from this function in normal circumstances? Here you of course needed to keep the fn signature the same so an extra layer of DTO mapping was needed.

crates/store/src/db/models.rs
```diff
@@ -0,0 +1,44 @@
+//! Defines models for usage with the diesel API
+
+use diesel::prelude::*;
+
+use crate::{db::schema::accounts, errors::DatabaseError};
+
+#[derive(Debug, Clone, Queryable, QueryableByName, Selectable)]
+#[diesel(table_name = accounts)]
```
Mirko-von-Leipzig: How does this work for multi-table models? i.e. with joins?

crates/store/src/db/models.rs
```diff
@@ -0,0 +1,44 @@
...
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+pub struct AccountInfoRawRow {
+    pub account_id: Vec<u8>,
+    #[allow(dead_code)]
+    pub network_account_id_prefix: Option<i64>,
+    pub account_commitment: Vec<u8>,
+    pub block_num: i64,
+    pub details: Option<Vec<u8>>,
```
Mirko-von-Leipzig: Would we ever have these be strongly typed? e.g. `AccountId`, `BlockNumber`, or is that always an additional conversion step?

crates/store/src/db/mod.rs
drahnr: So there are two options, tuples with primitive types or a requirement of types to be `Queryable` (double checking). I chose the former for a clear transition from `DB types -> equiv rust types -> abstract types`. I am going to investigate a way to reduce the boilerplate.

crates/store/src/db/mod.rs
drahnr: Yes, we'd define an additional type `struct AccountInfoRawRowSub {..}` with only the desired fields or use a tuple (I'd discourage that, it becomes brittle and error prone rather quickly).
We'd still need a conversion, see https://github.com/0xMiden/miden-node/pull/928#discussion_r2132078688

crates/store/src/db/models.rs
drahnr: https://docs.diesel.rs/2.2.x/diesel/query_dsl/trait.QueryDsl.html#method.left_outer_join
https://docs.diesel.rs/2.2.x/diesel/query_dsl/trait.QueryDsl.html#method.inner_join

contains examples, I have yet to implement this with diesel 2.x.y, so my understanding from reading diesel's source code + docs, it's required to make `T::as_selectable()` is needed for custom types that are not tuples, and the annotation is a nice to have error improvement. It should work just fine without.

I'll put it on my todo to verify.

crates/store/src/db/models.rs
drahnr: [`FromSql`](https://docs.rs/diesel/latest/diesel/deserialize/trait.FromSql.html) seems to be the correct approach, for types we define within the crate. For foreign types I think we have to use an intermediate implementation and wrap `Queryable` / impl `FromSql` for the wrapping type.

crates/store/src/db/models.rs
drahnr: Addenda, the docs contain an example for the `left_join` case: https://docs.rs/diesel/latest/diesel/deserialize/trait.Queryable.html

crates/store/src/db/schema.rs
```diff
@@ -0,0 +1,140 @@
+// @generated automatically by Diesel CLI.
+
+diesel::table! {
+    account_deltas (account_id, block_num) {
+        account_id -> Binary,
+        block_num -> BigInt,
```
bobbinth, line 6: Block numbers are `u32` values. Should we use `Integer` here instead of `BigInt`?

crates/store/src/db/schema.rs
```diff
@@ -0,0 +1,140 @@
...
+    notes (block_num, batch_index, note_index) {
+        block_num -> BigInt,
+        batch_index -> Integer,
+        note_index -> Integer,
+        note_id -> Binary,
+        note_type -> Integer,
+        sender -> Binary,
+        tag -> Integer,
+        execution_mode -> Integer,
+        aux -> Integer,
```
bobbinth: `aux` can be up to 64 bits in size. Should we use `BitInt` here?

crates/store/src/db/schema.rs
drahnr: We can, the above was derived from the existing schema by `diesel-schema`. I didn't/don't have range boundary context yet.

crates/store/src/db/schema.rs
drahnr: Definitely!

crates/store/src/db/schema.rs
drahnr: I gave this another think; tl;dr the ordering would change once we hit the highest bit of `u32`, so if we decide for it, I'd still recommend to go for `i64` alias `BigInt` in the database. See also https://midengroup.slack.com/archives/C08UTTDUMAS/p1749562610858149?thread_ts=1749222553.944629&cid=C08UTTDUMAS

crates/store/src/db/mod.rs
drahnr: The tl;dr is that it is possible, but only if the conversions are failable. Right now this is an assumption made in a few places (whatever in the DB can be converted successfully to in mem representations) but I don't think it's a great thing to assume.

crates/store/src/db/models.rs
drahnr: Technically this should be feasible using only `#[derive(QueryableByName)]` with `#[diesel(embed)]` on the fields fields, which also implement `QueryableByName`. Each of them can then have different tables, while the combining type does not declare table. The downside to this, we still need to have an `Combo::as_select()` call which yields a type that's not compatible with the query you want to define. So I'll punt on this and use tuples.

bin/node/src/commands/store.rs
```diff
@@ -136,6 +136,8 @@ impl StoreCommand {
         accounts_directory: &Path,
         maybe_genesis_config: Option<&PathBuf>,
     ) -> anyhow::Result<()> {
+        fs_err::create_dir_all(accounts_directory)?;
+        fs_err::create_dir_all(data_directory)?;
```
Mirko-von-Leipzig: I'm unsure if we actually want to do this. Generally I'm upset if a CLI magically creates directories.

crates/store/build.rs
```diff
@@ -0,0 +1,11 @@
+fn main() {
+    // trigger a rebuild if any migrations change
+    println!("cargo:rerun-if-changed=./src/db/migrations");
+    // if we do one re-write, the defaults are gone
```
Mirko-von-Leipzig: What defaults? I'm not entirely following the todo's etc here.

crates/store/src/errors.rs
```diff
@@ -34,22 +34,34 @@ pub enum DatabaseError {
...
+    #[error(transparent)]
+    PoolRecycle(#[from] deadpool::managed::RecycleError<deadpool_diesel::Error>),
+    #[error("Summing over column {column} of table {table} exceeded {limit}")]
+    ColumnSumExceedsLimit {
+        table: &'static str,
+        column: &'static str,
+        limit: &'static str,
+        #[source]
+        source: Box<dyn std::error::Error + Send + Sync + 'static>,
+    },
```
Mirko-von-Leipzig: Error messages should start with lowercase to generate an error report properly.

crates/store/src/errors.rs
```diff
@@ -278,3 +304,43 @@ pub enum GetBatchInputsError {
...
+        AccountDeltaError, AccountError, DatabaseError, DatabaseSetupError, DeserializationError,
+        GenesisError, NetworkAccountError, NoteError, RecvError, StateInitializationError,
+    };
+
+    #[allow(dead_code)]
+    fn ensure_is_error<E>(_phony: PhantomData<E>)
+    where
+        E: std::error::Error + Send + Sync + 'static,
+    {
+    }
```
Mirko-von-Leipzig: Should we move this into the test fn? I doubt we'll ever want to use it elsewhere.

crates/store/src/db/manager.rs
Mirko-von-Leipzig, line 1: Comments look like they're wrapping too early. Unfortunately `rustfmt` seems to only truncate long comments, but doesn't extend short lines. Likely to allow the user flexibility to manually align things, but its pretty annoying.

crates/store/src/db/manager.rs
```diff
@@ -0,0 +1,97 @@
...
+use deadpool_sync::InteractError;
+
+use super::{Result, RunQueryDsl, SqliteConnection};
+
+#[derive(thiserror::Error, Debug)]
+pub enum WalConnManagerError {
+    #[error("failed to apply connection parameter")]
+    ConnectionParamSetup(#[source] diesel::result::Error),
+    #[error("SQLite pool interaction failed: {0}")]
+    InteractError(String),
```
Mirko-von-Leipzig, line 14: Would making it `Arc<anyhow::Error>` make sense at all?

Though in general I do wonder if we are getting any value from having many error variants.

crates/store/src/db/migrations/2025062000000_setup/down.sql
Mirko-von-Leipzig, line 1: Are `down` migrations a requirement? I'm not sure there will ever be a migration we can really revert.

crates/store/src/db/migrations/2025062000000_setup/up.sql
Mirko-von-Leipzig, line 1: Should we be worried about the `STRICT` removal?

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -1,20 +1,19 @@
 -- Table for storing different settings in run-time, which need to persist over runs.
--- Note: we can store values of different types in the same `value` field.
 CREATE TABLE settings (
     name  TEXT NOT NULL,
-    value ANY,
+    value BLOB, -- NOTE: previously was ANY, but that's not supported by diesel
```
Mirko-von-Leipzig:
```suggestion
    value BLOB,
```

crates/store/src/db/tests.rs
Mirko-von-Leipzig, line 1: I'm a bit worried that we no longer use database transactions, but instead pass in the connection? Or is the connection actually a wrapped transaction?

crates/store/src/db/migrations.rs
```diff
@@ -1,141 +1,25 @@
...
+    let migrations = conn.pending_migrations(MIGRATIONS).expect("In memory migrations never fail");
+    println!("Applying migrations: {}", migrations.len());
+
+    let Err(e) = conn.run_pending_migrations(MIGRATIONS) else {
+        return Ok(());
+    };
+    eprintln!("Failed to apply migration: {e:?}");
+    // something went wrong, MIGRATIONS contains
+    conn.revert_last_migration(MIGRATIONS)
+        .expect("Duality is maintained by the developer");
```
Mirko-von-Leipzig: This isn't as important yet because we don't have any active migrations beyond the initial schema, but:

1. We should use `tracing` instead of raw `println`.
2. I assume there is a way to do migrations that involve Rust in-the-loop to do model conversions for example?
3. Is there a way to add tracing to the process? What I've experienced with this is that a single migration will take a very long time to the point that the operator thinks the process has frozen and they force-kill. Having some way to indicate progress would be required eventually.
4. I'm skeptical about reverting migrations - if it failed its usually because something unexpected happened and the data is now broken or at least in an unknown state.

Mirko-von-Leipzig, line 25: If the migration failed to apply surely it doesn't need to be reverted as it wasn't applied?

crates/store/src/db/mod.rs
```diff
@@ -244,61 +162,68 @@ impl Db {
         // This will create the file if it does not exist, but will also happily open it if already
         // exists. In the latter case we will error out when attempting to insert the genesis
         // block so this isn't such a problem.
-        let mut conn = connection::Connection::open(database_filepath)
-            .context("failed to open a database connection")?;
+        let mut conn: SqliteConnection =
+            diesel::sqlite::SqliteConnection::establish(database_filepath.to_str().unwrap()) // TODO FIXME avoid unwrap
```
Mirko-von-Leipzig:
```suggestion
            diesel::sqlite::SqliteConnection::establish(database_filepath.to_str().context("database filepath is invalid")?) // TODO FIXME avoid unwrap
```

crates/store/src/db/models/conv.rs
```diff
@@ -0,0 +1,151 @@
+//! Central place to define conversion from and to database primitive types
+//!
+//! Eventually, all of them should have types and we can implement a trait for them
+//! rather than function pairs.
+//!
+//! Notice: All of them are infailable. The invariant is a sane content of the database
```
Mirko-von-Leipzig:
```suggestion
//! Notice: All of them are infallible. The invariant is a sane content of the database
```

crates/store/src/db/models/utils.rs
```diff
@@ -0,0 +1,117 @@
...
+/// Attention: We use `u32` as actual in-memory representation, since this will
+/// suffice well beyond all our lifetimes.
+pub(crate) fn block_number_to_raw_sql(block_num: BlockNumber) -> i64 {
+    block_num.as_u32() as i64
+}
+
+use diesel::{Connection, RunQueryDsl, SqliteConnection};
+use miden_objects::note::Nullifier;
+
+use crate::errors::DatabaseError;
```
Mirko-von-Leipzig: These should probably move to the top.

crates/store/src/db/models/utils.rs
```diff
@@ -0,0 +1,117 @@
...
+        raw.into_iter().map(<R as std::convert::TryInto<D>>::try_into),
+    )
+}
+
+#[allow(dead_code)]
+/// Deserialize an iterable container full of byte blobs `B` to types `T`
+pub(crate) fn deserialize_raw_vec<B: AsRef<[u8]>, T: Deserializable>(
+    raw: impl IntoIterator<Item = B>,
+) -> std::result::Result<Vec<T>, DeserializationError> {
+    std::result::Result::<Vec<_>, DeserializationError>::from_iter(
```
Mirko-von-Leipzig:
```suggestion
) -> Result<Vec<T>, DeserializationError> {
    Result::<Vec<_>, DeserializationError>::from_iter(
```

crates/store/src/db/settings.rs
Mirko-von-Leipzig, line 1: I think the only reason this exists is to store the schema version and hash. The version is conventionally stored in the user pragma.

I guess the hash is useful during development if you have the same version number but have changed the schema.. I'm unsure if that provides actual value.

I suppose we might have other settings to store in the future.

crates/store/src/db/schema.rs
Mirko-von-Leipzig: Just want to make sure I understand correctly. This is still an `i64` under the hood - `diesel` just calls it `BigInt` for some reason?

crates/store/src/db/models/types.rs
```diff
@@ -0,0 +1,341 @@
...
+            details,
+            merkle_path,
+        })
+    }
+}
+
+/// A type to represent a `sum(BigInt)`
+// TODO: make this a type, but it's unclear how that should work
+// See: <https://github.com/diesel-rs/diesel/discussions/4684>
+pub type BigIntSum = BigDecimal;
```
Mirko-von-Leipzig, line 320: Where do we ever sum such large numbers?

crates/store/src/db/models/types.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Queryable, QueryableByName, Selectable)]
+#[diesel(table_name = nullifiers)]
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+pub struct NullifierRawRow {
+    pub nullifier: Vec<u8>,
+    #[allow(dead_code)]
+    pub nullifier_prefix: i32, // TODO most usecases do not require this to be actually loaded
```
Mirko-von-Leipzig: I imagine this is only used to filter by and wouldn't need to be loaded.

crates/store/src/db/models/types.rs
Mirko-von-Leipzig, line 1: Quite a few comments, style and `.expect("XXX")` that need to be cleaned up

crates/store/src/db/models/queries.rs
```diff
@@ -0,0 +1,1640 @@
...
+    }
+
+    Ok(())
+}
+
+/// Builds an [`AccountDelta`] from the given [`Account`].
+///
+/// This function should only be used when inserting a new account into the DB.The returned delta
+/// could be thought of as the difference between an "empty transaction" and the it's initial state.
+fn build_insert_delta(account: &Account) -> Result<AccountDelta, DatabaseError> {
```
Mirko-von-Leipzig, line 430: This feels like something that we shouldn't have to implement ourselves.

@PhilippGackstatter there is probably some way of getting `Account -> AccountDelta`?

crates/store/src/db/models/queries.rs
```diff
@@ -0,0 +1,1640 @@
...
+/// All matching notes from the first block greater than `block_num` containing a matching note.
+/// A note is considered a match if it has any of the given tags, or if its sender is one of the
+/// given account IDs. If no matching notes are found at all, then an empty vector is returned.
+///
+/// # Note
+///
+/// This method returns notes from a single block. To fetch all notes up to the chain tip,
+/// multiple requests are necessary.
+pub(crate) fn select_notes_since_block_by_tag_and_sender(
+    conn: &mut SqliteConnection,
```
Mirko-von-Leipzig, line 74: Am I understanding correctly that this represents either a database connection, or a database transaction? And its up to the caller to ensure its the right one?

crates/store/src/db/models/queries.rs
Mirko-von-Leipzig, line 1: I'm gone through this a few times, but its a lot so I can't confidently say its a correct translation. I'll give it another pass in a day or two.

crates/store/src/db/models/queries.rs
```diff
@@ -0,0 +1,1640 @@
...
+            schema::transactions::transaction_id,
+        ),
+    )
+    .filter(schema::transactions::block_num.gt(block_start))
+    .filter(schema::transactions::block_num.le(block_end))
+    .filter(schema::transactions::account_id.eq_any(desired_account_ids))
+    .order(schema::transactions::transaction_id.asc())
+    .load::<TransactionSummaryRaw>(conn)
+    .map_err(DatabaseError::from)?;
+    Ok(vec_raw_try_into(raw).unwrap())
```
Mirko-von-Leipzig, line 1647: This should have a `SAFETY` comment

crates/store/src/db/models/queries.rs
Mirko-von-Leipzig, line 1: We probably should split this file up; but that can be a subsequent clean-up PR.

I'm also interested in a wrapper type for `SqliteConnection` so we can enforce transaction's only, as well as have them be methods on the type instead of free-standing.

bin/node/src/commands/store.rs
drahnr: I tend to be annoyed if it fails me because of a subdir structure that it expects by default, that I have to create upfront.

crates/store/build.rs
drahnr: In `build.rs`, if you place one `rerun-if-changed=` instruction, the _default_ builtin ones will be disabled.

crates/store/src/db/manager.rs
drahnr: `anyhow::Error` itself does not implement `std::error::Error + Send + Sync`, so it becomes a pain to use.

The value of error variants is adding extra context to the call. I think it's good enough to not invent another error system 😅

crates/store/src/db/migrations.rs
drahnr: 1. will adjust
2. to my knowledge, that's not in the scope of diesel, but would require custom logic, just reading all rows and converting them manually and writing it to an intermediate db and rename the table/drop the old one
3. There is only a coarse per-migration way of displaying progress with the current diesel API
4. I think that depends on the acceptable size of a migration and the playbook post-issue - do you want to ensure as much data as possible is recovered or focus on debugging the failure case?

drahnr: I am unsure, I'd hope you're correct and you don't need to revert because it's atomic. But I am not sure if that holds for i.e. long migrations with lot's of changes/multiple consecutive SQL statements.

Mirko-von-Leipzig: Something to investigate when we add migrations for real.

I _always_ ran each migration as a transaction to make them atomic. I can understand super-large migrations needing to be broken up, but I haven't really hit that limit before despite having a terabyte sized DB.

crates/store/build.rs
Mirko-von-Leipzig: afaik `build.rs` is _always_ included.

But we should definitely not be including `src` otherwise it will rebuild always while developing?

crates/store/src/db/models/queries.rs
drahnr: That is correct. Calling `transaction` just modifies internal state of the connection. The implementation also states that having nested transactions reduces it to a single transaction.

crates/store/src/db/settings.rs
drahnr: I am in favor of deleting, and re-introducing as-needed. Right now, we don't use it.

crates/store/src/db/migrations/2025062000000_setup/up.sql
drahnr: The type system encodes the columns, and the `STRICT` is specific to `SQLite` ( https://www.sqlite.org/lang_createtable.html ), at least it's not present in `postgresql` . I don't think it's an issue per se.

crates/store/src/db/migrations/2025062000000_setup/down.sql
drahnr: From memory, I don't think they are

bin/node/src/commands/store.rs
Mirko-von-Leipzig: I've gone both ways on this myself. In this case I would tend towards not creating them because the user specifies the directories so its not some unknown structure.

But I also get the convenience argument. I just dislike running a cmd locally and suddenly I have 5 new folders in my project directory.

crates/store/src/db/models/queries.rs
```diff
@@ -0,0 +1,1642 @@
...
+// Attention: uses the _implicit_ column `rowid`, which requires to use a few raw SQL nugget
+// statements
+#[allow(
+    clippy::cast_sign_loss,
+    reason = "We need custom SQL statements which has given types that we need to convert"
+)]
+pub(crate) fn unconsumed_network_notes(
+    conn: &mut SqliteConnection,
+    mut page: Page,
+) -> Result<(Vec<NoteRecord>, Page), DatabaseError> {
```
sergerad, line 980: This fn is significantly more complicated than the existing implementation. Is that because it involves pagination and a join? Is there anything we can do to avoid local struct and fns?

crates/store/src/db/models/queries.rs
drahnr: From other places, the common path is to deconstruct the `Account` into parts and construct the `AccountDelta` from that. I copied this from the previous sql module (almost) verbatim.

crates/store/src/db/models/types.rs
drahnr: Fixed

crates/store/src/db/models/types.rs
drahnr: We do sum an `i64` for fungible asset account deltas. Note that the issue here is diesel being opinionated on following the SQL spec to the t.

crates/store/src/db/models/types.rs
drahnr: I double checked, and it has become a single use type, so the field can be removed.

bin/node/src/commands/store.rs
drahnr: I tend to agree if it's too many directories, but if it's two that I have to pass explicitly, I tend to prefer not having to do it manually. 

I'll remove it, it's out of scope for this PR.

crates/store/build.rs
drahnr: > afaik build.rs is always included.

You are correct.

> But we should definitely not be including src otherwise it will rebuild always while developing?

You are also correct 🫠

crates/store/src/db/models/queries.rs
PhilippGackstatter: This came up some time ago and I think it's only used once here? I'm not sure this one-off use case warrants adding a method to `Account`. We don't _generally_ convert an account into a delta, because for the initial transaction, the account itself is the delta. So this particular case seems node-specific to me, that I don't think other users would need, but please correct me if I'm wrong. If you find it should be handled in base, feel free to open an issue and we'll add it :)

crates/store/src/db/models/queries.rs
drahnr: The root cause is the implicit row id, which is not part of the schema, is the parameter we use for pagination. We could make it part of the schema, but I'd like to make that a follow up. The intent here was migrating the schema verbatim as far as possbile (it already deviated some in the `settings` table, using `BLOB` instead of `ANY` type values, but we didn't use those in the first place other than migrations, which also weren't used)

crates/store/src/db/models/queries.rs
Mirko-von-Leipzig: We're technically just interested in the parts/items themselves. I'm mostly worried that we forget to add something here if things change.

But come to think of it that's purely because we insist on taking the entire delta by value in `apply_delta`. It could instead take iterators, and then we could easily do the mapping ourselves without forgetting.

crates/store/src/db/models/queries.rs
drahnr: > I'm also interested in a wrapper type for SqliteConnection so we can enforce transaction's only, as well as have them be methods on the type instead of free-standing.

I could append this to this PR, it shouldn't be too much work. But could also be a small refactor PR

crates/store/src/db/models/queries.rs
sergerad: Understood. Given how much we change the schema atm I would be more keen to see the impl that reflects what we would ultimately have/want alongside necessary schema changes. Thanks for explaining.

crates/store/src/db/models/queries.rs
PhilippGackstatter: Not sure, I follow on the last part. `Account::apply_delta` takes `&AccountDelta`, so are you referring to another function of the same name?

Mirko-von-Leipzig: Yes apologies we have a bunch of store functions that are similarly named :)

In one of them we take `AccountDelta` and persist storage deltas, map deltas etc. And we use this function for both new accounts and account updates which is why we need the account -> delta conversion at the moment.

Instead we should just take the storage items etc.

crates/store/src/db/tests.rs
drahnr: The transaction is a internal state to the connection type. So it's not visible if one is within a transaction or not.

crates/store/Cargo.toml
```diff
@@ -44,5 +46,7 @@ miden-lib             = { features = ["testing"], workspace = true }
 miden-node-test-macro = { workspace = true }
 miden-node-utils      = { features = ["tracing-forest"], workspace = true }
 miden-objects         = { default-features = true, features = ["testing"], workspace = true }
+pretty_assertions     = { version = "1.4" }
+rand                  = { version = "0.9" }
```
Mirko-von-Leipzig: I think we can make these workspace deps

crates/store/src/db/manager.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn interact(msg: &(impl ToString + ?Sized), e: &InteractError) -> Self {
+        let msg = msg.to_string();
+        Self::InteractError(format!("{msg} failed: {e:?}"))
+    }
+}
+
+/// Create a connection manager with per-connection setup
+///
+/// Particularly, `foreign_key` checks are enabled and using a write-append-log for journaling.
+pub(crate) struct WalConnManager {
```
Mirko-von-Leipzig: Nit: probably don't need to prefix it with the properties it happens to enable.


```suggestion
pub(crate) struct ConnectionManager {
```

crates/store/src/db/mod.rs
```diff
@@ -13,47 +14,47 @@ use miden_objects::{
...
+
+pub(crate) mod models;
+/// [diesel](https://diesel.rs) generated schema
+///
+/// ```sh
+/// cargo binstall diesel_cli
+/// sqlite3 -init ./src/db/migrations/001-init.sql ephemeral_setup.db ""
+/// diesel setup --database-url=./ephemeral_setup.db
+/// diesel print-schema > src/db/schema.rs
+/// ```
```
Mirko-von-Leipzig: Should this be a `build.rs` function?

Or does this mean we need to manually do this until we have db migrations?

crates/store/build.rs
Mirko-von-Leipzig, line 1: What exactly is being run/built here? Does diesel do some auto magic somehow? If so, would be good to add that information as a comment.

Right now it looks like we can just delete this file since it does nothing.

crates/store/src/db/mod.rs
drahnr: This only needs to be done for the initial migration from an existing database. Afterwards one has the option to modify the `schema.rs` manually and provide a migration, or setup a dummy db and apply migrations, and regenerate the file. There is no enforced path. The latter ensures integrity of `schema.rs` and actual DB tables.

crates/store/build.rs
drahnr: We want to ship the migrations files, so if any of those changes, we want to include the latest version, and hence must rerun the inclusion `embed_migrations!`. Now that macro is a proc-macro, and we cannot trigger that to re-run on file change.

drahnr: Expanded the comment a bit.

Mirko-von-Leipzig: Yes but where is this `embed_migrations!` macro?

crates/store/src/db/mod.rs
Mirko-von-Leipzig: So iiuc after this PR the comment won't be applicable anymore - should we remove it?

I'm reading it as an instruction to do this for schema changes

drahnr: Removing

crates/store/src/db/migrations.rs
```diff
@@ -1,141 +1,25 @@
...
-static MIGRATIONS: LazyLock<Migrations> = LazyLock::new(prepare_migrations);
-
-fn up(s: &'static str) -> M<'static> {
-    M::up(s).foreign_key_check()
-}
-
-const DB_MIGRATION_HASH_FIELD: &str = "db-migration-hash";
-const DB_SCHEMA_VERSION_FIELD: &str = "db-schema-version";
+use crate::COMPONENT;
+pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!("src/db/migrations");
```
drahnr, line 9: @Mirko-von-Leipzig this is the invocation of the macro, which is why we need the `build.rs`

Mirko-von-Leipzig: Ah thanks. Could you add a comment here and similarly in `build.rs` which includes a link to the diesel doc https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds.

```rust
// The rebuild is automatically triggered by `build.rs` as described in <link>
```
and in `build.rs` something like

```rust
// This build.rs is required to trigger the `embed_migrations!` proc macro in `<path>/migration` as per <link>.
```

crates/store/build.rs
drahnr: https://github.com/0xMiden/miden-node/pull/928#discussion_r2240255936

crates/store/src/db/models/queries.rs
drahnr: I'll leave this as is and we can discuss this on Thursday, it's not clear to me how this is supposed to be done differently.

crates/store/src/db/models/queries.rs
```diff
@@ -0,0 +1,1647 @@
...
+            .filter(
+                schema::notes::tag
+                .eq_any(&desired_note_tags)
+                .or(
+                    schema::notes::sender
+                    .eq_any(&desired_senders)
+                )
+            )
+            .get_results::<NoteSyncRecordRawRow>(conn)
+            .map_err(DatabaseError::from)?;
```
bobbinth, line 147: This sends two requests to the database while previously we sent only one request, right? This is because we cannot use `rarray` with `disel`, correct?

Probably not a big deal for now (and we may end up refactoring this anyway).

crates/store/src/db/models/queries.rs
```diff
@@ -0,0 +1,1647 @@
...
+            },
+            deserialize_raw_vec, get_nullifier_prefix, raw_sql_to_block_number, serialize_vec,
+            sql_sum_into, vec_raw_try_into,
+        },
+        schema,
+    },
+    errors::{NoteSyncError, StateSyncError},
+};
+
+/// Select notes matching the tags and account IDs search criteria using the given [Connection].
```
bobbinth, line 61: nit: shouldn't we use `[SqliteConnection]` here instead of `[Connection]`? (connection is not actually a type AFAICT).

Another option could be just to alias `SqliteConnection` as `Connection` and then use `Connection` everywhere.

crates/store/src/db/models/queries.rs
```diff
@@ -0,0 +1,1647 @@
...
+            insert_account_delta(conn, account_id, block_num, &delta)?;
+        }
+
+        count += inserted;
+    }
+
+    Ok(count)
+}
+
+/// Insert notes to the DB using the given [Transaction]. Public notes should also have a nullifier.
```
bobbinth: Similar comment as above but about `[Transaction]` - it should probably be `SqliteConnection` or just `Connection` if we alias it.

Also, to confirm `SqliteConnection` actually works like a transaction, right? i.e., multiple updates made against the same `SqliteConnection` will be committed atomically.

crates/store/src/errors.rs
```diff
@@ -277,3 +321,46 @@ pub enum GetBatchInputsError {
         latest_block_num: BlockNumber,
     },
 }
+
+// Do not scope for `cfg(test)` - if it the traitbounds don't suffice
+// the issue will already appear in the compilation of the library
+// or binary, which would prevent getting to compiling the
+// following code.
+mod compile_tests {
```
bobbinth: nit: comment wrapping.

crates/store/src/db/settings.rs
```diff
@@ -1,25 +1,53 @@
-use rusqlite::{OptionalExtension, Result, ToSql, params, types::FromSql};
+// TODO figure out if we want to retain the settings
+#![allow(dead_code)]
```
bobbinth: I don't remember exactly what we used the settings for - but if they are no longer used, should we just delete them?

crates/store/src/db/migrations/2025062000000_setup/up.sql
bobbinth, line 1: Question: the `schema.rs` file gets automatically generated from this file, correct? Or if not, where does the `shema.rs` come from?

crates/store/src/db/models/queries.rs
bobbinth: I think the need to store delta's (whether full account deltas or storage/vault deltas) will go away very soon (i.e., with the refactoring to the account retrieval API). So, I wouldn't spend much time on this.

crates/store/src/db/models/queries.rs
Mirko-von-Leipzig: Since sqlite is embedded there isn't much difference between performing the operation within the database, or within rust code. Usually the difference is

- database query plan optimiser might have additional information to optimise/index better than the rust code, and
- you can perform more complex or otherwise hidden from the db schema indexing/lookups on the rust end

But as you say, not so important right now

crates/store/src/db/models/queries.rs
Mirko-von-Leipzig: iiuc `SqliteConnection` can be either a transaction _or_ an actual connection. Its up to the caller to ensure they pass in the appropriate type. I _really_ dislike this but its how all the database rust impls do things because that's how the underlying C APIs work.

I've mentioned in some comment elsewhere that's lost in the stack that I'd like a follow-up that forces `Transaction` everywhere, probably by creating our own wrapper type.

crates/store/src/db/migrations/2025062000000_setup/up.sql
Mirko-von-Leipzig: See https://github.com/0xMiden/miden-node/pull/928#discussion_r2239762709 and https://github.com/0xMiden/miden-node/pull/928#discussion_r2240255936.

But as a summary, it gets generated by `embed_migrations!` proc macro which in turn is (weirdly & indirectly) rebuilt via `build.rs`.

crates/store/src/db/models/queries.rs
drahnr: `SqliteConnection` represents a connection, a transaction is an internal state to that and is done via `conn.transaction(|conn| { /* sql queries */ })` or the `begin/end`

crates/store/src/db/settings.rs
drahnr: Settings were used for versioning and migration. Both are not used anymore, happy to delete.

crates/store/src/db/migrations/2025062000000_setup/up.sql
drahnr: > But as a summary, it gets generated by embed_migrations! proc macro which in turn is (weirdly & indirectly) rebuilt via build.rs.

That's not entirely correct. That one just includes the mgirations, but it doesn't create the initial file, it's created by the `diesel_cli`

```sh
cargo binstall diesel_cli
sqlite3 -init ./src/db/migrations/001-init.sql ephemeral_setup.db ""
diesel setup --database-url=./ephemeral_setup.db
diesel print-schema > src/db/schema.rs
```

crates/store/src/db/models/queries.rs
drahnr: I pulled this apart for the ability to debug this better (both rust and SQL). The particular query doesn't benefit from the scheduler, since the first query always yields exactly one element, which must be known for the second query.

crates/store/src/db/models/queries.rs
bobbinth: > I've mentioned in some comment elsewhere that's lost in the stack that I'd like a follow-up that forces `Transaction` everywhere, probably by creating our own wrapper type.

Let's create an issue for this (unless one already exists). I think it would be great to be able to tell by looking at the code which function expect a simple connection and which ones expect a transaction (unless all of them expect a transaction?).

drahnr: We do have `fn transact` and `fn query` - if in doubt, we might just remove `query` and enforce a transaction and avoid additional complexity. We only have a handful of queries that don't have transactions.

Mirko-von-Leipzig: I think the point is more that the functions in `query.rs` should _not_ be allowing anything that isn't a transaction.

But I would also remove `query` because imo _all_ business logic should be using a transaction.

Issue #918, opened 2025-06-03
Separate Store gRPC methods


Mirko-von-Leipzig:
Split the store's gRPC service into three separate services: RPC, block-producer, ntx.

### Motivation

This makes it easier to identify and configure internal vs external methods. This enables

- Separate worker/connection pools for each. This is _required_ to prevent RPC from drowning out the block-producer requests.
- Separate read-only store replica's. This is an option to horizontally scale the RPC requests by cloning the store.
- Simplifies OTel configuration since different services can now be instrumented separately. And similarly, since the services would have different `target = xxx` values, the trace level could be controlled separately.

### Considerations

This will require additional configuration for the `store` component.

Also consider simply re-using the RPC proto service definition as the store's RPC service definition. This is (currently) a 1-1 _except_ for the transaction submission method. The store could simply leave this unimplemented instead of copying the entire service minus one.

Mirko-von-Leipzig:
> > This will require additional configuration for the store component.
> 
> [@Mirko-von-Leipzig](https://github.com/Mirko-von-Leipzig) The idea is that the three separate components are served on different addresses, right? An alternative could be to use different services of the same tonic server on one single address. This would simplify the configuration and avoid having multiple addresses for the store, but I think it may be limiting for having separate connection pools.

If on the same address, then the only shared pool would be the tokio/axum connections correct? We could still separate the database connection pool.. Hmm 🤔.

My thinking was to serve them on entirely separate endpoints; this also lets us give them different tracing levels and rate-limiting etc without having to open the request itself. But if you think its too large a change, we could also limit it to only separating the gRPC definitions for now.

PR #932, opened 2025-06-06
feat: refactor store api into separate services

TomasArrachea:
Addresses part of https://github.com/0xMiden/miden-node/issues/918

This PR refactors the Store API to separate it into three APIs: RPC, BlockProducer, NtxBuilder.

For follow up PR:
The three services are being hosted on one single tonic server, it's probably better to serve them separately on 3 different addresses. This would need to update the store config.


crates/store/src/server/api.rs
```diff
@@ -54,18 +54,17 @@ pub struct StoreApi {
...
 }
 
+// CLIENT ENDPOINTS
+// --------------------------------------------------------------------------------------------
 #[tonic::async_trait]
-impl api_server::Api for StoreApi {
-    // CLIENT ENDPOINTS
-    // --------------------------------------------------------------------------------------------
-
+impl rpc_server::Rpc for StoreApi {
```
bobbinth: Should we move this (and the other two similar impls) into separate files? For example, instead of a single `api.rs` file, we'd have something like:

- `src/server/rpc_api.rs`
- `src/server/ntx_builder.rs`
- `src/server/block_producer.rs`

crates/store/src/server/api.rs
```diff
@@ -517,101 +591,111 @@ impl api_server::Api for StoreApi {
...
         parent = None,
         target = COMPONENT,
-        name = "store.server.get_block_by_number",
+        name = "store.ntx_builder_server.get_block_header_by_number",
         skip_all,
         ret(level = "debug"),
         err
     )]
-    async fn get_block_by_number(
+    async fn get_block_header_by_number(
```
bobbinth: For endpoints that are shared, should we put them into a common impl? For example, we could have:
```rust
impl StoreApi {
    /// Implementation of this method without instrumentation.
    async fn get_block_header_by_number(
        &self,
        request: Request<GetBlockHeaderByNumberRequest>,
    ) -> Result<Response<GetBlockHeaderByNumberResponse>, Status> {
        ...
    }
}

impl ntx_builder_server::NtxBuilder for StoreApi {
    #[instrument(
        parent = None,
        target = COMPONENT,
        name = "store.ntx_builder_server.get_block_header_by_number",
        skip_all,
        ret(level = "debug"),
        err
    )]
    async fn get_block_header_by_number(
        &self,
        request: Request<GetBlockHeaderByNumberRequest>,
    ) -> Result<Response<GetBlockHeaderByNumberResponse>, Status> {
        self.get_block_header_by_number(request)
    }
}
```

crates/proto/src/generated/store.rs
bobbinth, line 1: Why did this file grow by this much?

bobbinth:
> The three services are being hosted on one single tonic server, maybe it's better to serve them separately on 3 different addresses. This would need to update the store config to take 3 addresses.

Maybe we should do this in stages? First logically separate the services and then in the next PR make them work on different addresses?

crates/proto/src/generated/store.rs
TomasArrachea: I think it's just because of the boilerplate of having three separate server/client bindings and the shared methods repeated.

TomasArrachea:
> Maybe we should do this in stages? First logically separate the services and then in the next PR make them work on different addresses?

Agree, we can do the different addresses in a follow-up PR

proto/proto/store.proto
```diff
@@ -84,9 +66,42 @@ service Api {
...
+    rpc GetBlockInputs(requests.GetBlockInputsRequest) returns (responses.GetBlockInputsResponse) {}
+
+    // Returns the inputs for a transaction batch.
+    rpc GetBatchInputs(requests.GetBatchInputsRequest) returns (responses.GetBatchInputsResponse) {}
+
+    // Returns data required to validate a new transaction.
+    rpc GetTransactionInputs(requests.GetTransactionInputsRequest) returns (responses.GetTransactionInputsResponse) {}
+}
+
+// Store API for the NtxBuilder
```
Mirko-von-Leipzig: Nit: let's spell this out
```suggestion
// Store API for the network transaction builder component
```

proto/proto/store.proto
```diff
@@ -8,10 +8,8 @@ import "requests.proto";
 import "responses.proto";
 import "google/protobuf/empty.proto";
 
-service Api {
-    // Applies changes of a new block to the DB and in-memory data structures.
-    rpc ApplyBlock(requests.ApplyBlockRequest) returns (responses.ApplyBlockResponse) {}
-
+// Store API for the RPC
```
Mirko-von-Leipzig:
```suggestion
// Store API for the RPC component
```

crates/store/src/server/rpc_api.rs
Mirko-von-Leipzig, line 1: Could you set all trace levels of the `instrument` to `level = "debug"`. At the moment the gRPC of the RPC component overwhelms our monitoring otherwise.

The other's should remain at `info`.

bobbinth (approved):
Looks good! Thank you!

PR #933, opened 2025-06-06
feat: add faucet dockerfile

TomasArrachea:
Fixes https://github.com/0xMiden/miden-node/issues/920

This PR adds a Dockerfile to run the faucet. It expects the node to be running on the same host machine, it connects to `http://host.docker.internal:57291`. Added a small change to the faucet cli to be able to set the `node_url` on the init command.

Also adds the CI check and publish workflow.


bin/faucet/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    --token-symbol POL \
+    --decimals 9 \
+    --max-supply 1000000000
+
+LABEL org.opencontainers.image.authors=miden@polygon.io \
+    org.opencontainers.image.url=https://0xpolygonmiden.github.io/ \
+    org.opencontainers.image.documentation=https://github.com/0xMiden/miden-node \
+    org.opencontainers.image.source=https://github.com/0xMiden/miden-node \
+    org.opencontainers.image.vendor=Polygon \
+    org.opencontainers.image.licenses=MIT
```
TomasArrachea: This seems oudated (I took it from the current node Dockerfile), but not sure what data should go here

sergerad:
> It expects the node to be running on the same host machine

Its not important right now but we may at some point want to address this limitation.

bin/faucet/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    --config-path miden-faucet.toml \
+    --node-url http://host.docker.internal:57291
+
+RUN miden-faucet create-faucet-account \
+    --output-path faucet.mac \
+    --token-symbol POL \
+    --decimals 9 \
+    --max-supply 1000000000
+
+LABEL org.opencontainers.image.authors=miden@polygon.io \
```
sergerad:
```suggestion
LABEL org.opencontainers.image.authors=devops@miden.team \
```

bin/faucet/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    org.opencontainers.image.url=https://0xpolygonmiden.github.io/ \
```
sergerad:
```suggestion
    org.opencontainers.image.url=https://0xMiden.github.io/ \
```

bin/faucet/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    org.opencontainers.image.documentation=https://github.com/0xMiden/miden-node \
+    org.opencontainers.image.source=https://github.com/0xMiden/miden-node \
+    org.opencontainers.image.vendor=Polygon \
```
sergerad:
```suggestion
    org.opencontainers.image.vendor=Miden \
```

sergerad (approved):
TY

bobbinth:
> > It expects the node to be running on the same host machine
> 
> Its not important right now but we may at some point want to address this limitation.

Yes, fine for now, but eventually faucet would be run on a different machine (and we will probably move it into a different repo).

bin/faucet/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN miden-faucet init \
+    --faucet-account-path faucet.mac \
+    --config-path miden-faucet.toml \
+    --node-url http://host.docker.internal:57291
+
+RUN miden-faucet create-faucet-account \
+    --output-path faucet.mac \
+    --token-symbol POL \
+    --decimals 9 \
+    --max-supply 1000000000
```
Mirko-von-Leipzig: This essentially means the faucet values are hardcoded for everyone, correct?

How it configuration usually handled for docker?

And I guess should we update the token, decimals and supply to match our new official values as per #945 or is this fine?

bin/faucet/src/main.rs
```diff
@@ -90,6 +91,8 @@ pub enum Command {
         faucet_account_path: String,
         #[arg(short, long, default_value = DEFAULT_API_KEYS_COUNT)]
         generated_api_keys_count: u8,
+        #[arg(short, long)]
+        node_url: Option<String>,
```
Mirko-von-Leipzig, line 95: I'm having a brainfart. How are we currently setting the faucet's node URL without this?

bin/faucet/Dockerfile
TomasArrachea: You are right.
I think we could use env vars for this. We set the default values and the user is able to override it with custom values if they need to. Or an alternative could be to let the user set the path to an already created _account.mac_ file.

bin/faucet/src/main.rs
TomasArrachea: We currently use the _miden-faucet.toml_ config file, where the node url is set. When you run `miden-faucet init` the default url is `http://127.0.0.1:57291` so it just works. But from the container we need to connect to host.docker.internal

Mirko-von-Leipzig: Okay we should probably revisit how the faucet is configured/controlled.

In general I'd prefer clap CLI with env var support over the toml config files.

But that's a separate issue :)

PR #938, opened 2025-06-09
chore: Bump genesis faucet supply to u32 max

sergerad:
### Context

The default token supply generated on genesis is 1e6. This can run out pretty quickly for the purposes of our testnet faucet, for example.

### Changes
- Update genesis faucet token decimals to 6 and supply to 1e11 * 1e6.
- Replace POL token symbol with MIDEN.

### DIFF
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -155,7 +155,7 @@ impl StoreCommand {
         let decimals = 6u8;
         let base_unit = 10u64.pow(u32::from(decimals));
         let max_supply = 100_000_000_000u64 * base_unit;
-        let max_supply = Felt::try_from(max_supply).expect("max supply is less than u64::MAX");
+        let max_supply = Felt::try_from(max_supply).expect("max supply is less than field modulus");
 
         // Create the faucet.
         let (mut account, account_seed) = create_basic_fungible_faucet(
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -155,8 +155,8 @@ impl StoreCommand {
             rng.random(),
             AccountIdAnchor::PRE_GENESIS,
             TokenSymbol::try_from("MIDEN").expect("MIDEN should be a valid token symbol"),
-            12,
-            Felt::from(u32::MAX),
+            6,
+            Felt::try_from(100_000_000_000u64).expect("1e11 is less than u64::MAX"),
             miden_objects::account::AccountStorageMode::Public,
             AuthScheme::RpoFalcon512 { pub_key: secret.public_key() },
         )?;
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -154,7 +154,7 @@ impl StoreCommand {
         let (mut account, account_seed) = create_basic_fungible_faucet(
             rng.random(),
             AccountIdAnchor::PRE_GENESIS,
-            TokenSymbol::try_from("POL").expect("POL should be a valid token symbol"),
+            TokenSymbol::try_from("MIDEN").expect("MIDEN should be a valid token symbol"),
             12,
             Felt::from(u32::MAX),
             miden_objects::account::AccountStorageMode::Public,
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -161,7 +161,7 @@ impl StoreCommand {
         let (mut account, account_seed) = create_basic_fungible_faucet(
             rng.random(),
             AccountIdAnchor::PRE_GENESIS,
-            TokenSymbol::try_from("MIDEN").expect("MIDEN should be a valid token symbol"),
+            TokenSymbol::try_from("MIDEN").expect("MIDEN is a valid token symbol"),
             decimals,
             max_supply,
             miden_objects::account::AccountStorageMode::Public,
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -151,12 +151,19 @@ impl StoreCommand {
         let mut rng = ChaCha20Rng::from_seed(rand::random());
         let secret = SecretKey::with_rng(&mut get_rpo_random_coin(&mut rng));
 
+        // Calculate the max supply of the token.
+        let decimals = 6u8;
+        let base_unit = 10u64.pow(u32::from(decimals));
+        let max_supply = 100_000_000_000u64 * base_unit;
+        let total_supply = Felt::try_from(max_supply).expect("total supply is less than u64::MAX");
+
+        // Create the faucet.
         let (mut account, account_seed) = create_basic_fungible_faucet(
             rng.random(),
             AccountIdAnchor::PRE_GENESIS,
             TokenSymbol::try_from("MIDEN").expect("MIDEN should be a valid token symbol"),
-            6,
-            Felt::try_from(100_000_000_000u64).expect("1e11 is less than u64::MAX"),
+            decimals,
+            total_supply,
             miden_objects::account::AccountStorageMode::Public,
             AuthScheme::RpoFalcon512 { pub_key: secret.public_key() },
         )?;
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -156,7 +156,7 @@ impl StoreCommand {
             AccountIdAnchor::PRE_GENESIS,
             TokenSymbol::try_from("POL").expect("POL should be a valid token symbol"),
             12,
-            Felt::from(1_000_000u32),
+            Felt::from(u32::MAX),
             miden_objects::account::AccountStorageMode::Public,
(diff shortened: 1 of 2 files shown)


igamigo:
> Update genesis faucet token supply to u32::MAX.

Any reason not to make it even higher?

sergerad:
> > Update genesis faucet token supply to u32::MAX.
> 
> Any reason not to make it even higher?

Whatever you guys think. Max would be this I believe
```
        /// Field modulus = 2^64 - 2^32 + 1
        const M: u64 = 0xff_ff_ff_ff_00_00_00_01;
```

bin/node/src/commands/store.rs
```diff
@@ -147,15 +147,15 @@ impl StoreCommand {
...
         let mut rng = ChaCha20Rng::from_seed(rand::random());
         let secret = SecretKey::with_rng(&mut get_rpo_random_coin(&mut rng));
 
         let (mut account, account_seed) = create_basic_fungible_faucet(
             rng.random(),
-            TokenSymbol::try_from("POL").expect("POL should be a valid token symbol"),
+            TokenSymbol::try_from("MIDEN").expect("MIDEN should be a valid token symbol"),
             12,
-            Felt::from(1_000_000u32),
+            Felt::from(u32::MAX),
```
bobbinth: I think max supply should be higher and the number of decimal places smaller. Specifically, I'm thinking:

- $6$ for decimal places (down from the current 12).
- 100B in tokens or $10^{11}$. Together with 6 decimal places this will result in total supply of $10^{17}$ base units (or under $2^{57}$).

We'll probably need to update the faucet UI to issue tokens in base units - i.e., 100 tokens would translate to 100M base units.

bobbinth (commented):
Looks good! Thank you. I left one comment inline.

> Possible max would be this I believe:
> 
> ```
> /// Field modulus = 2^64 - 2^32 + 1
> const M: u64 = 0xff_ff_ff_ff_00_00_00_01;
> ```

In the transaction kernel, we limit max supply for fungible assets at $2^{63}$, I believe.

bin/node/src/commands/store.rs
```diff
@@ -154,8 +154,8 @@ impl StoreCommand {
         let (mut account, account_seed) = create_basic_fungible_faucet(
             rng.random(),
             TokenSymbol::try_from("MIDEN").expect("MIDEN should be a valid token symbol"),
-            12,
-            Felt::from(u32::MAX),
+            6,
+            Felt::try_from(100_000_000_000u64).expect("1e11 is less than u64::MAX"),
```
bobbinth: Max supply is specified in base units - so, the value should be $10^{17}$.

bin/node/src/commands/store.rs
```diff
@@ -147,15 +147,15 @@ impl StoreCommand {
         Store::bootstrap(genesis_state, data_directory)
     }
 
-    fn generate_genesis_account() -> anyhow::Result<AccountFile> {
+    pub fn generate_genesis_account() -> anyhow::Result<AccountFile> {
```
Mirko-von-Leipzig: Probably doesn't need to be `pub`?

```suggestion
    fn generate_genesis_account() -> anyhow::Result<AccountFile> {
```

bin/node/src/commands/store.rs
```diff
@@ -177,3 +177,13 @@ impl StoreCommand {
         ))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::StoreCommand;
+
+    #[test]
+    fn test_generate_genesis_account_no_panic() {
```
Mirko-von-Leipzig: I'd like to move away from prefixing every test with `test_`.

```suggestion
    fn generate_genesis_account_succeeds() {
```

bin/node/src/commands/store.rs
```diff
@@ -147,15 +147,15 @@ impl StoreCommand {
...
         let mut rng = ChaCha20Rng::from_seed(rand::random());
         let secret = SecretKey::with_rng(&mut get_rpo_random_coin(&mut rng));
 
         let (mut account, account_seed) = create_basic_fungible_faucet(
             rng.random(),
-            TokenSymbol::try_from("POL").expect("POL should be a valid token symbol"),
-            12,
-            Felt::from(1_000_000u32),
+            TokenSymbol::try_from("MIDEN").expect("MIDEN should be a valid token symbol"),
+            6,
```
Mirko-von-Leipzig: @bobbinth could you maybe expand on why `6` decimal points? I'd like to add that in as a comment/constant description so we don't have dangling magic numbers.

@sergerad if you could maybe move this to a `const TOKEN_DECIMALS` with the answer as a doc?

bin/node/src/commands/store.rs
Mirko-von-Leipzig: Its actually possible to constrain this properly but we're using the wrong types at the top level. I believe we can use the [following](https://github.com/0xMiden/miden-base/blob/03569de6d2608a07b0344e87c35d440fb16fec80/crates/miden-objects/src/asset/fungible.rs#L26) (@PhilippGackstatter if you could re-confirm this is the correct one):
```rust
FungibleAsset::MAX_AMOUNT
```

```suggestion
            6,
            // SAFETY: FungibleAsset amount is guaranteed to fit into a `Felt`.
            FungibleAsset::MAX_AMOUNT.amount().try_into().unwrap(),
```

Mirko-von-Leipzig: I actually wonder if there might be benefit in a more strictly typed `FungibleAssetAmmount` type wrapper; but that would largely be the domain of the protocol so I'm unsure about the value proposition.

bin/node/src/commands/store.rs
bobbinth: > @bobbinth could you maybe expand on why `6` decimal points? I'd like to add that in as a comment/constant description so we don't have dangling magic numbers.

This is what some of the more popular tokens on Ethereum are using (e.g., USDC, USDT). Other considered options:

- 18 decimal places. This is standard for ERC20 tokens - but doesn't quite work for us because that would basically exhaust u64 values.
- 8 decimal places. This is used by Bitcoin, but I always found 100M number to be a bit odd. We could do 8 decimal places too, but then we'd be limited to about 90B tokens (i.e., we won't have enough space to do 100B).

So, I guess the choice could be between:

- 100B tokens with 6 decimal places.
- 10B tokens with 8 decimal places.

No strong preference between the two, TBH.

bin/node/src/commands/store.rs
sergerad: Yea forgot I left that there

bin/node/src/commands/store.rs
sergerad: Yes same, just did for consistency

bobbinth:
@sergerad - this PR needs to be rebased. Otherwise, we'll push many extra commits into `main`.

bin/node/src/commands/store.rs
```diff
@@ -151,12 +151,19 @@ impl StoreCommand {
         let mut rng = ChaCha20Rng::from_seed(rand::random());
         let secret = SecretKey::with_rng(&mut get_rpo_random_coin(&mut rng));
 
+        // Calculate the max supply of the token.
+        let decimals = 6u8;
+        let base_unit = 10u64.pow(u32::from(decimals));
+        let max_supply = 100_000_000_000u64 * base_unit;
+        let total_supply = Felt::try_from(max_supply).expect("total supply is less than u64::MAX");
```
bobbinth: nit: I would name the second variable `max_supply` as well (and also fix the error message):
```rust
let max_supply = 100_000_000_000u64 * base_unit;
let max_supply = Felt::try_from(max_supply).expect("max supply is less than field modulus");
```

sergerad: Weird, I thought I had implemented it that way. Maybe accidental undo or something. Thanks

bin/node/src/commands/store.rs
```diff
@@ -151,12 +151,19 @@ impl StoreCommand {
         let mut rng = ChaCha20Rng::from_seed(rand::random());
         let secret = SecretKey::with_rng(&mut get_rpo_random_coin(&mut rng));
 
+        // Calculate the max supply of the token.
+        let decimals = 6u8;
+        let base_unit = 10u64.pow(u32::from(decimals));
+        let max_supply = 100_000_000_000u64 * base_unit;
+        let max_supply = Felt::try_from(max_supply).expect("max supply is less than u64::MAX");
```
bobbinth: nit: the error message is not technically correct here (see https://github.com/0xMiden/miden-node/pull/938#discussion_r2136749052).

bin/node/src/commands/store.rs
PhilippGackstatter: Yes, using `FungibleAsset::MAX_AMOUNT` should work and could be a nice edge-case test for the protocol :)

TomasArrachea:
created https://github.com/0xMiden/miden-node/issues/948 to update the UI

PR #942, opened 2025-06-10
chore: Bump genesis faucet supply to u32 max

sergerad:
Cherry pick from #938 which went into main.

### DIFF
--- a/bin/faucet/README.md
+++ b/bin/faucet/README.md
@@ -15,14 +15,14 @@ make docker-run-node
 make install-faucet
 ```
 
-3. [Optional] Create faucet account (skip this step if you want to use an account from the genesis). This will generate authentication keypair and generate and write public faucet account data with its keypair into the file specified in `output-path`: 
+3. [Optional] Create faucet account (skip this step if you want to use an account from the genesis). This will generate authentication keypair and generate and write public faucet account data with its keypair into the file specified in `output-path`:
 
 ```bash
 miden-faucet create-faucet-account \
   --output-path <path to faucet.mac> \
-  --token-symbol POL \
-  --decimals 9 \
-  --max-supply 1000000000
+  --token-symbol MIDEN \
+  --decimals 6 \
+  --max-supply 100000000000
 ```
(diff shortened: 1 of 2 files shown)


bin/faucet/README.md
```diff
@@ -15,14 +15,14 @@ make docker-run-node
...
 
 ```bash
-  --token-symbol POL \
-  --decimals 9 \
-  --max-supply 1000000000
+  --token-symbol MIDEN \
+  --decimals 6 \
+  --max-supply 100000000000
```
bobbinth, line 25: I think we are missing 6 zeros here (to account for decimal places). I'll fix it in a different commit.

bobbinth (approved):
Looks good! Thank you!

TomasArrachea:
> @TomasArrachea @SantiagoPittella AFAICT there are no changes required to the faucet frontend to reflect this change of 12->6 decimals for the token. WDYT?
> 
> I'm expecting the transfer to implicitly go from 100e12 to 100e6 just by virtue of changing the token.
> 

I think we do actually need to make changes. Currently the faucet is just minting base units, so we should multiply that number for 10e6 to mint that amount of tokens. I created this issue earlier for this https://github.com/0xMiden/miden-node/issues/948, but maybe i'm missing something.

sergerad:
> Yes, we just need to multiply user-provided amount by 1000000 - so, should be a pretty straight-forward change.

If that is the case, then are we saying it was multiplying amounts by 1e12 before? Given decimals went from 12->6.

Mirko-von-Leipzig:
> > > Yes, we just need to multiply user-provided amount by 1000000 - so, should be a pretty straight-forward change.
> > 
> > 
> > If that is the case, then are we saying it was multiplying amounts by 1e12 before? Given decimals went from 12->6.
> 
> No, previously we were not multiplying at all (so, it was a bug).

Then how have we been running out of funds on the faucet :thinking: iiuc this means we had roughly a million times _more_ tokens than we thought since we were handing out cents?

PR #943, opened 2025-06-10
chore: Do not rm node bd on apt remove

sergerad:
### Context

We currently wipe network state on `apt remove miden-node`. We would prefer to require something more explicit like `apt purge` to cause such a deletion. Instead, `apt remove` will delete the configuration and service file only.

### Changes
- Update miden-node Debian package's postrm script to no longer wipe the application data / working directory.


Mirko-von-Leipzig (approved):
The faucet package has the same issue; should also fix that one.

sergerad:
> The faucet package has the same issue; should also fix that one.

Ah because it contains the account file right? Apart from that the state doesn't matter?

Issue #940, opened 2025-06-09
Proof verification fails between node <=> client


phklive:
### Version

0.9

### Other packages versions

If installing the miden-node using `--locked` on v0.9. And using the client on v0.9. When sending a transaction to the node the verification of the proof will fail (Image of the logs attached).

If the miden-node is installed without using `--locked` the issue disappears.

We should:
- Understand why this bug arises
- Fix the current Testnet node running
- Update the documentation to make sure all commands lead to working components

### What happened?

![Image](https://github.com/user-attachments/assets/819888ee-4006-4d92-a829-f235f1376d46)

### What should have happened?

Valid proof verification.

### How can this be reproduced?

1. Install the node using: `--locked`
2. Install the client.
3. Create a transaction.

### Relevant log output

```shell

```

Mirko-von-Leipzig:
This is likely related to breaking changes in the `winterfel` crate(s). Cargo update on `v0.9.0` 
identifies

```sh
cargo update --dry-run
    Updating crates.io index
     Locking 32 packages to latest compatible versions
    <..snip..>

    Updating winter-air v0.12.1 -> v0.12.3
    Updating winter-prover v0.12.2 -> v0.12.3
    Updating winter-verifier v0.12.2 -> v0.12.3
```

The client's lockfile has the exact same versions:

```sh
winter-air      v0.12.1
winter-prover   v0.12.2 
winter-verifier v0.12.2
```

So in general, the client and node will be identical with `--locked`, but contain a mismatch if either is used without. This isn't a bug in the node, nor client, nor a problem with how `--locked` vs no `--locked` works. Rather its a broken semver rule - which happens pretty easily, and is why I recommend using `--locked` where possible.

However since this impacts usage we should still try fix this for our active network(s). Our only real option is to `cargo update` the winterfel family of crates for **both** client and node. This would require a new release so that users get them via `cargo install`.

I don't think this type of problem is trivially catchable from the node since it requires mismatching random dependencies (from the node's perspective).

sergerad:
>In general, we should make running cargo update the first step in our release process. cc @sergerad.

I'm not sure thats something we could automate or even check automatically. We only ever want to release what has been committed so the update has to happen before we go to release. Meaning it would be a manual thing to remember when creating a PR for version bump.

igamigo:
[...]

PR #944, opened 2025-06-10
Refresh Cargo.lock

bobbinth:
This PR refreshes `Cargo.lock` file in the `main` branch by running `cargo update`. The most important consequence of this is that winterfell dependency crates are now locked at v0.12.3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (commented):
Was this a full `cargo update`, or just the specific winterfel packages? For a patch fix I'd prefer the latter so we aren't running with many changes at once.

I'd like to discuss the alternative I suggested [here](https://github.com/0xMiden/miden-node/issues/940#issuecomment-2957890756) before we merge this.

bobbinth:
> Was this a full `cargo update`, or just the specific winterfel packages? For a patch fix I'd prefer the latter so we aren't running with many changes at once.

Yeah, i just ran a full `cargo update`. I don't mind rolling it back (though, in addition to the winterfell crates, we'll need to pull in the latest versions of `miden-base` crates) - but also I think it's fine to keep as is. If something breaks, we should be able to catch it with client integration tests.

Mirko-von-Leipzig (approved):

PR #945, opened 2025-06-10
chore: Fix decimals in faucet readme

sergerad:


bin/faucet/README.md
```diff
@@ -22,7 +22,7 @@ miden-faucet create-faucet-account \
   --output-path <path to faucet.mac> \
   --token-symbol MIDEN \
   --decimals 6 \
-  --max-supply 100000000000
+  --max-supply 100000000000000000
```
Mirko-von-Leipzig, line 25: Do we know if clap accepts the `rust` way of specifying ints?

```suggestion
  --max-supply 100_000_000_000_000_000
```

Mirko-von-Leipzig (approved):

bin/faucet/README.md
sergerad: I would expect that to fail to parse as it would be interpreted as a string whereas the flag is expecting u64
```rust
        #[arg(short, long)]
        max_supply: u64,
```

PR #949, opened 2025-06-10
feat: support gRPC reflection in all services

### LINKED ISSUE #929: Enable gRPC server side reflection
This allows tools such as `grpcurl` and `postman` to work without requiring the actual protobuf files - the server(s) would now provide them instead.

This is only directly relevant for the RPC gRPC definition, but it probably doesn't hurt to enable it for the internal gRPC services either. This might simplify test scripts for example, and manual interactions.

tomyrd:
closes #929

You can test the reflection is working by running `grpcurl -plaintext localhost:57291 list rpc.Api`


crates/block-producer/src/server.rs
```diff
@@ -246,10 +247,16 @@ impl BlockProducerRpcServer {
     }
 
     async fn serve(self, listener: TcpListener) -> Result<(), tonic::transport::Error> {
+        let reflection_service = tonic_reflection::server::Builder::configure()
+            .register_file_descriptor_set(block_producer_api_descriptor())
+            .build_v1()
+            .unwrap();
```
igamigo: nit: Should we propagate the error here (and elsewhere in the changes from this PR)? Otherwise might be better to change the `unwrap()` into an `expect()` and/or add a safety comment

tomyrd: Oops, my bad. I forgot to remove the unwraps after testing. Will do.

Mirko-von-Leipzig (approved):

PR #950, opened 2025-06-10
chore: update dependencies

### LINKED ISSUE #883: Update dependencies
The following crates have new versions:
- `tonic`
- `tonic-web`
- `tonic-build`
- `opentelemetry`
- `opentelemetry_sdk`
- `opentelemetry-otlp`

But we can't update them until a new version of `tracing-opentelemetry` is released (with support for `opentelemetry v0.30`). There is already a PR to update it: https://github.com/tokio-rs/tracing-opentelemetry/pull/205 . We should keep track of their versions to update our owns once possible.

SantiagoPittella:
closes #883


sergerad (approved):
LGTM

bin/proving-service/src/commands/worker.rs
```diff
@@ -92,15 +93,15 @@ impl StartWorker {
...
 
         // Mark the service as serving
         health_reporter.set_serving::<ApiServer<RpcListener>>().await;
 
         tonic::transport::Server::builder()
             .accept_http1(true)
-            .add_service(tonic_web::enable(rpc.api_service))
-            .add_service(tonic_web::enable(rpc.status_service))
+            .add_service(tonic_web::GrpcWebLayer::new().layer(rpc.api_service))
+            .add_service(tonic_web::GrpcWebLayer::new().layer(rpc.status_service))
```
TomasArrachea: I am not sure if it works but according to tonic docs this should be done with:
```suggestion
            .layer(GrpcWebLayer::new())
            .add_service(rpc.api_service)
            .add_service(rpc.status_service)
```

SantiagoPittella: Oh, my way was working but if this works it would be much better. Will test it and come back with the results.

SantiagoPittella: Tests are passing ✔️

bin/proving-service/src/commands/worker.rs
```diff
@@ -92,15 +92,15 @@ impl StartWorker {
...
 
         // Mark the service as serving
         health_reporter.set_serving::<ApiServer<RpcListener>>().await;
 
         tonic::transport::Server::builder()
             .accept_http1(true)
-            .add_service(tonic_web::enable(rpc.api_service))
-            .add_service(tonic_web::enable(rpc.status_service))
+            .add_service(rpc.api_service)
+            .add_service(rpc.status_service)
```
bobbinth, line 107: We no longer need this? I'm assuming we still can make web requests, right?

Also, with this gone, do we still need `tonic-web` dependencies?

bin/proving-service/src/commands/worker.rs
TomasArrachea: I think we still need to add the `.layer(GrpcWebLayer::new())` line

crates/rpc/src/server/mod.rs
```diff
@@ -41,7 +41,7 @@ impl Rpc {
             .layer(TraceLayer::new_for_grpc().make_span_with(rpc_trace_fn))
             .layer(AcceptLayer::new()?)
             // Enables gRPC-web support.
-            .add_service(tonic_web::enable(api_service))
+            .add_service(api_service)
```
Mirko-von-Leipzig, line 53: Missing the gRPC-web layer iiuc

Mirko-von-Leipzig (approved):
lgtm sans the 2x missing gRPC-web layers

SantiagoPittella:
Update: currently investigating why the grpc-web test isn't passing (and the client web-client integration tests too) despite that re added the grpc-web layer.

SantiagoPittella:
@sergerad @bobbinth @Mirko-von-Leipzig 
update: 
- tonic-web  in the previous version was setting up the Cors layer, it now delegates that layer config to the user. I copied the setting from the previous version, with an explanation of what each header is doing. Let me know if we prefer to set something different.
- tonic's`tls` feature was replaced with `tls-ring` or `tls-aws-lc`, I went for `tls-ring` but don't have any strong opinion. In the examples it was changes to `tls-ring`:  https://github.com/hyperium/tonic/pull/2050/files#diff-a07b01c0e1bcaddea88402d286352d4fdf8fb3de57a136e1879402517366cc16L266 .

crates/utils/src/cors.rs
```diff
@@ -0,0 +1,47 @@
...
+/// - `expose_headers`: Sets the `access-control-expose-headers` header to the following headers:
+///   - `Status::GRPC_STATUS`
+///   - `Status::GRPC_MESSAGE`
+///   - `Status::GRPC_STATUS_DETAILS`
+/// - `allow_headers`: Sets the `access-control-allow-headers` header to the following headers:
+///   - `HeaderName::from_static("x-grpc-web")`
+///   - `http::header::CONTENT_TYPE`
+///   - `HeaderName::from_static("x-user-agent")`
+///   - `HeaderName::from_static("grpc-timeout")`
+pub fn cors_layer() -> CorsLayer {
```
Mirko-von-Leipzig: Should we maybe name it more specifically? 

```suggestion
pub fn cors_for_grpc_web_layer() -> CorsLayer {
```

bin/proving-service/src/commands/worker.rs
```diff
@@ -92,15 +94,17 @@ impl StartWorker {
...
             .accept_http1(true)
-            .add_service(tonic_web::enable(rpc.api_service))
-            .add_service(tonic_web::enable(rpc.status_service))
+            .layer(cors_for_grpc_web_layer())
+            .layer(GrpcWebLayer::new())
+            .add_service(rpc.api_service)
+            .add_service(rpc.status_service)
             .add_service(health_service)
             .serve_with_incoming(TcpListenerStream::new(rpc.listener))
             .await
```
bobbinth, line 110: Not for this PR, but would be great to align these parts of proving service components with the other node components. For example:

- In the other components this function is called `serve()` - here it is called `execute()`.
- In the other components we return `anyhow::Result<()>` - here we return `Result<(), String>`.
- In the other components we attach a span - here we do not.

There are a bunch of other small inconsistencies as well.

Not sure if it is possible, but maybe there is a way to move most of this code into a utility function that can be used in many crates. Let's create an issue for this.

bobbinth (approved):

Issue #788, opened 2025-04-15
Grpc-web test


tomyrd:
In #779 we accidentally removed a `tonic_web::enable` call which was needed to support grcp-web calls. We should add a test that ensures web calls are supported so that this doesn't happen in the future.

tomyrd:
Just to extend on [this comment](https://github.com/0xPolygonMiden/miden-node/pull/786?notification_referrer_id=NT_kwDOApac17QxNTgwNjExNjEwNDo0MzQyNDk4Mw&notifications_query=reason%3Aparticipating#issuecomment-2806889781). I thought the test would be able to be implemented just with `tonic_web` types but I got some problems with unsatisfied traits when trying to use `rpc::ApiClient`.

If we can't use just `tonic_web`, an alternative (which would be more cumbersome) would be to try to build the http request manually but I haven't tried this.

In `miden-client` we also have `TonicRpcClient` (with the feature `web-tonic`) which already implements every call. This could also be used as a dev dependency just in the test.

igamigo:
For this, it may be enough to create a TCP connection and sending an HTTP/1.1 request with a valid payload and the correct content type for grpc web. Since I don't think there are any endpoints that require no parameters, the simplest endpoint to hit might be `GetBlockHeaderByNum` and pass a serialized request for block number 0 to the request.

bobbinth:
> Since I don't think there are any endpoints (I think) that require no parameters, the simplest endpoint to hit might be `GetBlockHeaderByNum` passing a serialized request for block number 0.

Should we add a `/status` endpoint to the node? This could work very similar to the `/status` endpoint we implemented for the proving service and could return a response that could look something like this:

```json
{
    "version": "0.8.0",
    "chain_tip": 123456,
    "block_producer": {
        "version": "0.8.0",
        "status": "connected",
    },
    "store": {
        "version": "0.8.0",
        "status": "connected",
    },
}
```

tomyrd:
I think this would be ideal, as we could even skip instantiating the store and block producer for this specific test and use that endpoint to verify that we support grpc-web.

bobbinth:
@SantiagoPittella - could you create an issue for the status endpoint?

SantiagoPittella:
Done: https://github.com/0xPolygonMiden/miden-node/issues/794

Mirko-von-Leipzig:
[...]

PR #952, opened 2025-06-10
test: gRPC web support

TomasArrachea:
closes https://github.com/0xMiden/miden-node/issues/788

This PR adds a test to ensure that the RPC server has enabled web support. The test makes a status request and checks that the response contains the CORS headers.

Additionally updated a doc and a test to show the correct version header format.

### DIFF
--- a/crates/rpc/src/server/mod.rs
+++ b/crates/rpc/src/server/mod.rs
@@ -91,10 +91,12 @@ mod tests {
             .unwrap();
         let headers = response.headers();
 
-        // If `tonic_web` is enabled, the CORS headers should be set in the response
+        // CORS headers are usually set when `tonic_web` is enabled.
+        //
+        // This was deduced by manually checking, and isn't formally described
+        // in any documentation.
         assert!(headers.get("access-control-allow-credentials").is_some());
         assert!(headers.get("access-control-expose-headers").is_some());
         assert!(headers.get("vary").is_some());
-        assert!(response.status().is_success());
     }
 }
--- a/crates/rpc/src/server/mod.rs
+++ b/crates/rpc/src/server/mod.rs
@@ -80,7 +80,7 @@ mod tests {
 
         // An empty message with header format:
         //   - A byte indicating uncompressed (0)
-        //   - A u64 indicating the data length (0)
+        //   - A u32 indicating the data length (0)
         //
         // Originally described here:
         // https://github.com/hyperium/tonic/issues/1040#issuecomment-1191832200
--- a/crates/rpc/src/server/mod.rs
+++ b/crates/rpc/src/server/mod.rs
@@ -47,3 +47,54 @@ impl Rpc {
             .context("failed to serve RPC API")
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use http::{
+        HeaderMap, HeaderValue,
+        header::{ACCEPT, CONTENT_TYPE},
+    };
+    use tokio::net::TcpListener;
+
+    use crate::Rpc;
+
+    #[tokio::test]
+    async fn rpc_server_has_web_support() {
+        // Start server
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let rpc_addr = listener.local_addr().unwrap();
+        let rpc = Rpc {
+            listener,
+            store: "127.0.0.1:50051".parse().unwrap(),
+            block_producer: None,
+        };
+        tokio::spawn(async move { rpc.serve().await.unwrap() });
+
+        // Send a status request
+        let client = reqwest::Client::new();
+
+        let mut headers = HeaderMap::new();
+        let accept_header = concat!("application/vnd.miden.", env!("CARGO_PKG_VERSION"), "+grpc");
+        headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/grpc-web+proto"));
+        headers.insert(ACCEPT, HeaderValue::from_static(accept_header));
+
+        let mut message = Vec::new();
+        message.push(0); // uncompressed flag
+        message.extend_from_slice(&[0; 4]); // length of empty message
+
+        let response = client
+            .post(format!("http://{rpc_addr}/rpc.Api/Status"))
+            .headers(headers)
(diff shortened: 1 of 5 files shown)


crates/rpc/src/server/mod.rs
```diff
@@ -47,3 +47,54 @@ impl Rpc {
...
+        let client = reqwest::Client::new();
+
+        let mut headers = HeaderMap::new();
+        let accept_header = concat!("application/vnd.miden.", env!("CARGO_PKG_VERSION"), "+grpc");
+        headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/grpc-web+proto"));
+        headers.insert(ACCEPT, HeaderValue::from_static(accept_header));
+
+        let mut message = Vec::new();
+        message.push(0); // uncompressed flag
+        message.extend_from_slice(&[0; 4]); // length of empty message
```
Mirko-von-Leipzig: How did you figure out what this should be?

crates/rpc/src/server/mod.rs
```diff
@@ -47,3 +47,54 @@ impl Rpc {
...
+        let response = client
+            .post(format!("http://{rpc_addr}/rpc.Api/Status"))
+            .headers(headers)
+            .body(message)
+            .send()
+            .await
+            .unwrap();
+        let headers = response.headers();
+
+        // If `tonic_web` is enabled, the CORS headers should be set in the response
```
Mirko-von-Leipzig: Why is that? Is there a documentation link or is this an experimental thing you figured out?

Mirko-von-Leipzig (approved):

docs/src/developer/rpc.md
```diff
@@ -10,7 +10,7 @@ component), reducing the load in this critical component.
 
 ## RPC Versioning
 
-The RPC server enforces version requirements against connecting clients that povide the HTTP ACCEPT header. When this header is provided, its corresponding value must follow this format: `application/vnd.miden.v0.9.0+grpc`.
+The RPC server enforces version requirements against connecting clients that povide the HTTP ACCEPT header. When this header is provided, its corresponding value must follow this format: `application/vnd.miden.0.9.0+grpc`.
```
igamigo, line 13: Good catch! I think there are also a couple of tests that use the `v` prefix but probably not worth changing.

crates/rpc/src/server/mod.rs
TomasArrachea: Its the framing of tonic messages, I think I just stumbled upon this comment while searching in the tonic repo (https://github.com/hyperium/tonic/issues/1040#issuecomment-1191832200) and it just worked

crates/rpc/src/server/mod.rs
TomasArrachea: There is not much explained in the docs, really. I just checked the responses with and without the web support and the only visible difference was the presence of this headers. Also if the web support is disabled this request would return status 400 so the status is also checked.

docs/src/developer/rpc.md
TomasArrachea: I updated one test but i couldn't found it anywhere else

crates/rpc/src/server/mod.rs
Mirko-von-Leipzig: Could you add something to that effect as a comment? Otherwise at some point we'll be scratching our heads if this ever changes.

```
// CORS headers are usually set when `tonic_web` is enabled.
//
// This was deduced by manually checking, and isn't formally described
// in any documentation.
//
// More generally though, if web support is disabled then this would return Err(400)
```

> Also if the web support is disabled this request would return status 400 so the status is also checked.

Why would it do that? Is it due to a protocol mismatch aka its not pure gRPC so its rejected? Maybe we should only test that then, and not so much the headers.

Either way, I'd appreciate an explanatory comment so I don't find myself guessing next time :)

TomasArrachea: Sorry i got confused, it thought it would because of the header being set to `grpc-web` but it actually returns status 200, it just doesn't include the headers.

crates/rpc/src/server/mod.rs
```diff
@@ -47,3 +47,56 @@ impl Rpc {
...
+        let client = reqwest::Client::new();
+
+        let mut headers = HeaderMap::new();
+        let accept_header = concat!("application/vnd.miden.", env!("CARGO_PKG_VERSION"), "+grpc");
+        headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/grpc-web+proto"));
+        headers.insert(ACCEPT, HeaderValue::from_static(accept_header));
+
+        let mut message = Vec::new();
+        message.push(0); // uncompressed flag
+        message.extend_from_slice(&[0; 4]); // length of empty message
```
Mirko-von-Leipzig:
```suggestion
        // An empty message with header format:
        //   - A byte indicating uncompressed (0)
        //   - A u64 indicating the data length (0)
        //
        // Originally described here:
        // https://github.com/hyperium/tonic/issues/1040#issuecomment-1191832200
        let mut message = Vec::new();
        message.push(0);
        message.extend_from_slice(&0u64.as_be_bytes());
```

crates/rpc/src/server/mod.rs
```diff
@@ -86,7 +86,7 @@ mod tests {
         // https://github.com/hyperium/tonic/issues/1040#issuecomment-1191832200
         let mut message = Vec::new();
         message.push(0);
-        message.extend_from_slice(&0u64.to_be_bytes());
+        message.extend_from_slice(&0u32.to_be_bytes());
```
Mirko-von-Leipzig, line 89: My bad :grimacing: but now the comment is wrong :stuck_out_tongue_closed_eyes:

TomasArrachea:
> May be nice to check in the test that the same request fails without the web gRPC service, but it seems that this would require a slightly deeper refactor so maybe validating manually is enough

Actually the request does not fail without the web support, so I don't think it's necessary. This test just checks that the response contains the CORS headers. I think there are some other implications to the web support besides the CORS headers, but I think this check is enough to assert that `tonic_web` is enabled.

Issue #936, opened 2025-05-30
Consider using gRPC for proxy status endpoint


SantiagoPittella:
All the other services are using gRPC endpoint for status check, the proxy has an HTTP endpoint. We should considering replace it with an RPC.

The proxy does not implement any gRPC endpoint, so it is not as simple as just adding a new method. We should investigate how to start and add a method inside a Pingora service.

Issue #917, opened 2025-06-03
Update proving service to use same Otel / tracing as node


sergerad:

Mirko-von-Leipzig:
Something to consider is whether we want to bridge the trace ID via gRPC headers. This would let us have a continuous trace for when the faucet and/or the block-producer delegates a proof.

This would be nice, but the counter-argument is that we only want this to occur for our internal proofs aka not for user requests.

Internal proofs would be for batch, block, ntx proofs from the block-producer, and tx proofs from the faucet.

We should think about how (or if) to implement this, but some options off the top of my head:

- Create internal only proving service(s) which always have this enabled. External requests get routed to a separate proving service.
- Use authentication keys, enable this for white-listed keys.

Both add complexity, I'd probably prefer the latter. There is an argument that we should host separate internal instances though, to ensure block-production isn't halted by users overloading the proving services. This could also be semi-solved by prioritising the white listed keys, though this doesn't prevent clogging via expensive user submitted proofs.

PR #953, opened 2025-06-10
feat: implement gRPC status endpoint in proxy

SantiagoPittella:
closes [#936](https://github.com/0xMiden/miden-node/issues/936)


bin/proving-service/src/commands/proxy.rs
```diff
@@ -115,15 +115,9 @@ impl StartProxy {
...
-        let mut status_service = Service::new("status".to_string(), status_service);
-        status_service
-            .add_tcp(format!("{}:{}", PROXY_HOST, self.proxy_config.status_port).as_str());
-        info!(
-            "Status service listening on {}:{}/status",
-            PROXY_HOST, self.proxy_config.status_port
-        );
+        // Add gRPC status service
+        let status_service = ProxyStatusService::new(worker_lb, self.proxy_config.status_port);
+        info!("gRPC status service will start on port {}", self.proxy_config.status_port);
```
Mirko-von-Leipzig:
```suggestion
        info!(port = %self.proxy_config.status_port, "gRPC status service starting");
```

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,194 @@
...
+            TcpListenerStream::new(listener),
+            async move {
+                let _ = shutdown.changed().await;
+                info!("gRPC status service received shutdown signal");
+            },
+        );
+
+        // Run the server
+        if let Err(e) = server.await {
+            error!("gRPC status service error: {}", e);
```
Mirko-von-Leipzig:
```suggestion
            error!(err=e, "gRPC status service failed");
```

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,194 @@
...
+        } else {
+            info!("gRPC status service stopped gracefully");
```
Mirko-von-Leipzig: Is this an expected condition or should this be an error?

I suppose this would occur normally if the process is being stopped e.g. via `systemctl stop <..>`?

bin/proving-service/src/proxy/status.rs
Mirko-von-Leipzig, line 1: There are quite a few trace logs which should be made structured logs instead of interpolated strings.

We may want to do this as part of a general cleanup over the whole code base so if you want to merge without fixing them now that's okay(ish).

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,194 @@
...
 pub struct ProxyStatusService {
     load_balancer: Arc<LoadBalancerState>,
+    port: u16,
+    cache: Arc<RwLock<Option<CachedStatus>>>,
+    update_interval: Duration,
+}
+
+/// Internal gRPC service implementation
+pub struct ProxyStatusGrpcService {
+    cache: Arc<RwLock<Option<CachedStatus>>>,
```
Mirko-von-Leipzig: Have you looked into tokio's [watch channel](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) before? I think it might be a good fit for what you're doing here.

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,194 @@
...
+        }
+    }
+}
+
+#[async_trait]
+impl Service for ProxyStatusService {
+    async fn start_service(
+        &mut self,
+        #[cfg(unix)] _fds: Option<ListenFds>,
+        mut shutdown: Receiver<bool>,
```
Mirko-von-Leipzig: Is this a `watch::Receiver`? I'm surprised this isn't a `oneshot::Receiver<()>` - how does this work?

Nit: I think in general I assume that a `Receiver` means `mpsc::Receiver` by default, so I think let's keep the `<type>::Receiver` prefix for non-mpsc? wdyt? (happy to concede here as well, I know IDEs insert this automatically so there is some friction here).

Mirko-von-Leipzig (commented):
Some nits and comments.

I'm also getting worried about our proto files, but that can be a separate grooming task.

bin/proving-service/src/proxy/status.rs
SantiagoPittella: > I suppose this would occur normally if the process is being stopped e.g. via systemctl stop <..>?

Yes, this is the case. wdyt? should I remove the second?

bin/proving-service/src/proxy/status.rs
SantiagoPittella: I was planning to normalize this as part of https://github.com/0xMiden/miden-node/issues/917 , should I create a separate issue?

bin/proving-service/src/proxy/status.rs
SantiagoPittella: I do not, thanks! Will take a look.

bin/proving-service/src/proxy/status.rs
Mirko-von-Leipzig: No that's fine :)

bin/proving-service/src/proxy/status.rs
SantiagoPittella: > Is this a watch::Receiver? I'm surprised this isn't a oneshot::Receiver<()> - how does this work?

This is a `watch::Receiver` required by the pingora's `Service` trait, this argument will change from `false` to `true` when the server receives a signal to shutdown.

> so I think let's keep the <type>::Receiver prefix for non-mpsc?

Sure! Sounds good.

bin/proving-service/src/proxy/status.rs
Mirko-von-Leipzig: No it's fine, was more wondering we should change it to an `error!`

bin/proving-service/src/proxy/status.rs
SantiagoPittella: Looking better, thanks!

bin/proving-service/src/proxy/status.rs
```diff
@@ -34,31 +30,39 @@ type CachedStatus = ProxyStatusResponse;
...
+    status_rx: watch::Receiver<CachedStatus>,
 }
 
 impl ProxyStatusService {
     pub fn new(load_balancer: Arc<LoadBalancerState>, port: u16) -> Self {
-        let cache = Arc::new(RwLock::new(None));
+        // Create initial empty status
+        let initial_status = ProxyStatusResponse {
+            version: env!("CARGO_PKG_VERSION").to_string(),
+            supported_proof_type: 0, // Will be updated immediately
```
Mirko-von-Leipzig: Why can't this be read from the load_balancer here yet?

SantiagoPittella: I sent a commit creating the initial status from the LB.

bin/proving-service/README.md
```diff
@@ -193,34 +193,50 @@ The status check returns:
...
+- `version`: The worker's version
+- `status`: The worker's health status (UNKNOWN, HEALTHY, or UNHEALTHY)
+
+#### Example Usage
+
+You can query the status endpoint using a gRPC client. For example, using `grpcurl`:
+
+```bash
+grpcurl -plaintext -import-path ./proto -proto proxy_status.proto \
+  -d '{}' localhost:<status_port> proxy_status.ProxyStatusApi.Status
```
sergerad: Would be good to specify the default port here instead of `<status_port>`.

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,200 @@
...
+// ================================================================================================
 
-/// Service that handles status requests
+/// gRPC service that implements Pingora's Service trait
 pub struct ProxyStatusService {
     load_balancer: Arc<LoadBalancerState>,
+    port: u16,
+    status_tx: watch::Sender<CachedStatus>,
+    update_interval: Duration,
+}
```
bobbinth: Could we add a high level explanation of how this works here? I'm having a hard time following some of this.

For example, it is not clear what `status_tx` is used for (or why couldn't it be just `ProxyStatusResponse` value).

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,200 @@
...
+    status_tx: &watch::Sender<CachedStatus>,
+) {
+    let status = generate_status(load_balancer).await;
+
+    // This will notify all receivers of the new status
+    let _ = status_tx.send(status);
+}
+
+/// Generates a new status from the load balancer.
+async fn generate_status(load_balancer: &LoadBalancerState) -> ProxyStatusResponse {
```
bobbinth: A few nits:
- I would name this `build_status()`.
- This looks like a helper function - let's move it into the `HELPER FUNCTIONS` section at the end of the file.
- Could this be a method on `ProxyStatusService`?

Also, let's improve doc comments here to explain what's going on. My understanding is that this doesn't make any requests to the work instances, we just read the currently cached statuses of each worker - correct?

bin/proving-service/src/proxy/status.rs
bobbinth, line 1: I think code organization in this file could be improved quite a bit (this includes adding more comments etc.). I'm still having trouble understanding the rationale behind some decisions.

bobbinth (commented):
Thank you! Looks good. I left some comments inline. The main thing is that I think we still need to clean up the code quite a bit. Could be done in a follow-up PR if that's easier.

bin/proving-service/src/proxy/status.rs
SantiagoPittella: Sure, I'll add docs to it.

In particular, the `status_tx` is the sender of a tokio's watch channel. I use it to send status updates to the proxy status grpc service.

SantiagoPittella:
@bobbinth I ended up refactoring to be clearer. I implemented both traits: pingora's `Service`, allowing us to be added as a service in the proxy; and the gRPC API with the status endpoint. And use the other struct solely as an updater of the status.

bin/proving-service/README.md
```diff
@@ -193,34 +193,50 @@ The status check returns:
...
+The status service provides the following method:
+- `Status(ProxyStatusRequest) -> ProxyStatusResponse`: Returns the current status of the proxy and all its workers
+
+#### Response Format
+
+The gRPC response includes the following information:
 - `version`: The version of the proxy
-- `supported_proof_type`: The types of proof that the proxy supports
-- `workers`: A list of workers with their status
+- `supported_proof_type`: The type of proof that the proxy supports (TRANSACTION, BATCH, or BLOCK)
```
igamigo: nit:
```suggestion
- `supported_proof_type`: The type of proof that the proxy supports (`TRANSACTION`, `BATCH`, or `BLOCK`)
```

bin/proving-service/README.md
```diff
@@ -193,34 +193,50 @@ The status check returns:
...
+- `workers`: A list of workers with their status information
+
+Each worker status includes:
+- `address`: The worker's network address
+- `version`: The worker's version
+- `status`: The worker's health status (UNKNOWN, HEALTHY, or UNHEALTHY)
```
igamigo: nit:
```suggestion
- `status`: The worker's health status (`UNKNOWN`, `HEALTHY`, or `UNHEALTHY`)
```

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,217 @@
...
-            .collect();
-
-        let status = ProxyStatus {
-            version: env!("CARGO_PKG_VERSION").to_string(),
-            prover_type: self.load_balancer.supported_prover_type,
-            workers: worker_statuses,
+#[async_trait]
+impl ProxyStatusApi for ProxyStatusPingoraService {
+    /// Returns the current status of the proxy.
+    async fn status(
```
igamigo, line 92: nit: This doesn't seem to need the `async` marker

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,217 @@
...
+/// Generates a new status from the load balancer and returns it as a [`ProxyStatusResponse`].
+async fn generate_status(load_balancer: &LoadBalancerState) -> ProxyStatusResponse {
+    let workers = load_balancer.workers.read().await;
+    let worker_statuses: Vec<WorkerStatus> = workers
+        .iter()
+        .map(|w| WorkerStatus {
+            address: w.address(),
+            version: w.version().to_string(),
+            status: WorkerHealthStatus::from(w.health_status()).into(),
+        })
```
igamigo: nit: May be worth adding a `From` impl that you can easily map here

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,217 @@
...
+    /// This is used to receive the status updates from the updater.
+    status_rx: Option<watch::Receiver<ProxyStatusResponse>>,
 }
 
-impl ProxyStatusService {
-    pub fn new(load_balancer: Arc<LoadBalancerState>) -> Self {
-        Self { load_balancer }
+impl ProxyStatusPingoraService {
+    /// Creates a new [`ProxyStatusPingoraService`].
+    pub fn new(load_balancer: Arc<LoadBalancerState>, port: u16) -> Self {
```
igamigo: nit: Could the `watch::channel` be initialized here? Although I guess that would require a default initial value or something like that, but would be nice to enforce the invariant

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,217 @@
...
+        let initial_status = generate_status(&self.load_balancer).await;
+        let (status_tx, status_rx) = watch::channel(initial_status);
+        self.status_rx = Some(status_rx);
+
+        // Start the status updater task
+        let updater = ProxyStatusUpdater::new(self.load_balancer.clone(), status_tx);
+        let cache_updater_shutdown = shutdown.clone();
+        tokio::spawn(async move {
+            updater.start(cache_updater_shutdown).await;
+        });
```
igamigo: Because this isn't awaited/polled, I think errors and panics would not be propagated. Maybe the task should be joined with the `server` task or something like that.

proto/proto/proxy_status.proto
```diff
@@ -0,0 +1,43 @@
...
+
+// Health status of a worker.
+enum WorkerHealthStatus {
+    // The worker's health status is unknown.
+    UNKNOWN = 0;
+    // The worker is healthy.
+    HEALTHY = 1;
+    // The worker is unhealthy.
+    UNHEALTHY = 2;
+}
```
igamigo: Could be nice to expand on what it means that the worker is unhealthy in terms of functionality (eg they would not receive tasks)

igamigo (commented):
Leaving some comments, mostly stylistic details (feel free to disregard), and only a couple bigger ones.

bin/proving-service/src/proxy/status.rs
SantiagoPittella: it comes as a requirement of the generated `ProxyStatusApi`

bin/proving-service/src/proxy/status.rs
SantiagoPittella: let me try a couple things. probably can.

bin/proving-service/src/proxy/status.rs
SantiagoPittella: Done!

bin/proving-service/src/proxy/status.rs
SantiagoPittella: Let me know what do you think of the new approach.

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ### Enhancements
 
+- Add miden-proving-service and miden-proving-service-client crates (#926).
```
sergerad: nit: I know it wasn't you (was me?) but could you update "Add" to "Added" please

bin/proving-service/src/generated/proxy_status.rs
```diff
@@ -0,0 +1,368 @@
+// This file is @generated by prost-build.
+/// Request message for the status of the proxy.
+#[derive(Clone, Copy, PartialEq, ::prost::Message)]
+pub struct ProxyStatusRequest {}
```
sergerad, line 4: nit: No whitespace between declarations in this file

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,226 @@
...
+
+    let supported_proof_type: ProofType = load_balancer.supported_prover_type.into();
 
-        let _ = self.handle_request(&mut session).await.map_err(|e| {
-            error!("Failed to handle status request: {}", e);
-            None::<Stream>
-        });
-        None
+    ProxyStatusResponse {
+        version: env!("CARGO_PKG_VERSION").to_string(),
```
sergerad: nit: Could access the env var just once in `ProxyStatusUpdater::new()` or even as an argument provided to that fn.

sergerad: Same with this?
```
    let supported_proof_type: ProofType = load_balancer.supported_prover_type.into();
```

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,226 @@
...
 
-/// Status of the proxy
-#[derive(Debug, Serialize)]
-pub struct ProxyStatus {
-    version: String,
-    prover_type: ProverType,
-    workers: Vec<WorkerStatus>,
-}
+/// Update status every 5 seconds
+const STATUS_UPDATE_INTERVAL_SECS: u64 = 5;
```
sergerad: I think this should come through as a Clap arg with env var support. Fed through into the service and updater. With `Duration` type being used in the `new()` constructors. WDYT?

Would need to add this to `bin/proving-service/.env` also.

sergerad: Also maybe a default of 10 or 20s might be better? The only thing updated is the worker count / health right?

CHANGELOG.md
SantiagoPittella: Yeah, sure 👍🏼

bin/proving-service/src/proxy/status.rs
SantiagoPittella: Yep, sounds good 👍🏼 

> The only thing updated is the worker count / health right?

It is, yes.

bin/proving-service/src/proxy/status.rs
```diff
@@ -1,103 +1,257 @@
...
+use tracing::{error, info};
 
-// Status of a worker
-#[derive(Debug, Serialize)]
-pub struct WorkerStatus {
-    address: String,
-    version: String,
-    status: WorkerHealthStatus,
-}
+use super::worker::WorkerHealthStatus as RustWorkerHealthStatus;
```
sergerad: nit: Confused as to the context of the word "Rust" here. Why is the alias required?

bin/proving-service/README.md
```diff
@@ -193,34 +193,50 @@ The status check returns:
...
+#### Example Usage
+
+You can query the status endpoint using a gRPC client. For example, using `grpcurl`:
+
+```bash
+grpcurl -plaintext -import-path ./proto -proto proxy_status.proto \
+  -d '{}' localhost:8084 proxy_status.ProxyStatusApi.Status
+```
 
 Example response:
```
sergerad, line 224: When I run the example I don't see an empty `"workers"` or `"supported_proof_type...` which I would expect:
```
➜ grpcurl -plaintext -import-path ./proto -proto proxy_status.proto \
  -d '{}' localhost:8084 proxy_status.ProxyStatusApi.Status
{
  "version": "0.10.0"
}
```

bin/proving-service/src/proxy/status.rs
sergerad: @SantiagoPittella please be aware of #998. If it merges before this PR, we should update yours to use humantime for this field. If you merge before it, I'll try to remember to do a followup issue!

bin/proving-service/src/proxy/status.rs
SantiagoPittella: I wanted to differentiate it from the Proto generated one, I would rename it to NativeWorkerHealthStatus, since it explains better the difference i think.

bin/proving-service/README.md
SantiagoPittella: Hmmm, strange. I'm having the same error with that command. But when using postman I'm having the right response:
```
{
    "workers": [],
    "version": "0.10.0",
    "supported_proof_type": "TRANSACTION"
}
```

SantiagoPittella: Even if adding a new worker to the proxy, the grpcurl command doesn't show the `"supported_proof_type"` field

SantiagoPittella: TIL: `grpcurl` doesn't show default values (empty string, enums set to the first value, etc). Adding `-emit-defaults` prints all values:

```sh
$ grpcurl -plaintext -emit-defaults \
  -import-path ./proto -proto proxy_status.proto \
  -d '{}' localhost:8084 proxy_status.ProxyStatusApi.Status

{
  "version": "0.10.0",
  "supportedProofType": "TRANSACTION",
  "workers": []
}
```

bin/proving-service/src/proxy/status.rs
sergerad: Thanks for exaplaining. Personally would rather rename the generated one to `ProtoXYZ` or `GeneratedXYZ`. I'm unsure what we do elsewhere in the code and what other people think. Up to you.

bin/proving-service/README.md
sergerad: haha tyvm for looking

sergerad (approved):

bin/proving-service/src/proxy/status.rs
SantiagoPittella: Saw that #998 was merged, i saw that most of the remote prover vars weren't updated, will check other binaries too and address it in another PR. I created #1005

Mirko-von-Leipzig (approved):

PR #955, opened 2025-06-10
chore: refactor workers health check in proxy

### LINKED ISSUE #934: Worker health check error handling
The proxy's worker status check grew too large, and does have a proper error handling. A lot of the workers status update can be moved to a single call to update at the end, and better error should be added.

SantiagoPittella:
closes #934

### DIFF
--- a/bin/proving-service/src/proxy/health_check.rs
+++ b/bin/proving-service/src/proxy/health_check.rs
@@ -1,6 +1,6 @@
 use pingora::{prelude::sleep, server::ShutdownWatch, services::background::BackgroundService};
 use tonic::async_trait;
-use tracing::debug_span;
+use tracing::{debug_span, error};
 
 use super::LoadBalancerState;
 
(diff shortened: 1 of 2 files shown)


bin/proving-service/src/proxy/health_check.rs
```diff
@@ -34,7 +34,17 @@ impl BackgroundService for LoadBalancerState {
...
                     for worker in workers.iter_mut() {
-                        worker.check_status(self.supported_prover_type).await;
+                        let status_result = worker.check_status(self.supported_prover_type).await;
+
+                        if let Err(ref reason) = status_result {
+                            error!(
+                                "Health check failed for worker {}: {}",
+                                worker.address(),
+                                reason
+                            );
```
Mirko-von-Leipzig: In general we should try stick to structured logs. e.g. we could filter by worker address and or reason on the tracing backend.

```suggestion
                            error!(
                                err = %reason,
                                worker.address = worker.address(),
                                "Worker failed health check"
                            );
```

bin/proving-service/src/proxy/health_check.rs
Mirko-von-Leipzig, line 1: fyi the tracing span for the background service doesn't work the way you'd expect. It cannot cross async calls without manually instrumenting them, but maybe that isn't really an issue here at the moment.

I tried to document what I mean [here](https://0xmiden.github.io/miden-node/developer/monitoring.html#footguns-and-common-issues).

bobbinth (approved):
Looks good! Thank you!

PR #956, opened 2025-06-10
feat: expose proving service as lib

tomyrd:
This PR exposes and also refactors the internal structure of the proving service.

This is needed for https://github.com/0xMiden/miden-client/pull/950 which is also updated and works correctly with this branch.


bin/proving-service/src/api/prover.rs
```diff
@@ -5,15 +5,60 @@ use miden_objects::{
...
+impl ProverType {
+    /// Returns the corresponding `ProofType` from the generated code
+    pub fn to_proof_type(&self) -> crate::generated::ProofType {
+        match self {
+            ProverType::Transaction => crate::generated::ProofType::Transaction,
+            ProverType::Batch => crate::generated::ProofType::Batch,
+            ProverType::Block => crate::generated::ProofType::Block,
+        }
+    }
+}
```
SantiagoPittella: Hm probably use a `From` impl here would be better?

Also, it would be nice to normalize names here, we have `ProofType` and `ProverType`. I prefer `ProofType`, maybe we can tackle this in a new PR. wdyt?

bin/proving-service/src/commands/update_workers.rs
```diff
@@ -29,7 +29,6 @@ pub struct RemoveWorkers {
     /// Workers to be removed from the proxy.
     ///
     /// The workers are passed as host:port strings.
-    #[arg(value_name = "WORKERS", env = "MPS_PROXY_WORKERS_LIST", value_delimiter = ',')]
```
SantiagoPittella: This should not be removed.

bin/proving-service/src/api/mod.rs
```diff
@@ -1,14 +1,11 @@
...
-    generated::{api_server::ApiServer, status::status_api_server::StatusApiServer},
-};
+use crate::generated::{api_server::ApiServer, status::status_api_server::StatusApiServer};
 
-mod prover;
+pub(crate) mod prover;
 mod status;
 
-pub use prover::ProverRpcApi;
+pub use prover::{MIDEN_PROVING_SERVICE, ProverRpcApi, ProverType};
```
sergerad: Why is `MIDEN_PROVING_SERVICE` required externally?

tomyrd: It isn't, I added it by mistake, nice catch

CHANGELOG.md
```diff
@@ -4,8 +4,30 @@
...
+
+## v0.9.1 (2025-06-10)
+
+- Refresh Cargo.lock file (#944).
+
+## v0.9.0 (2025-05-30)
+
+### Changes
+
+- Separated the store API into three separate services (#932).
```
SantiagoPittella: This is from a merge into `next` and not main. It should be in the 0.10 section. Also, the other entries doesn't look right either.

tomyrd: Seems like the `v0.9.0` section was duplicated in a past `next` commit, I'll review de entries and move them accordingly.

SantiagoPittella (approved):

CHANGELOG.md
```diff
@@ -7,26 +7,25 @@
...
-- Refresh Cargo.lock file (#944).
-
-## v0.9.0 (2025-05-30)
-
 ### Changes
 
 - Separated the store API into three separate services (#932).
 - Added a faucet Dockerfile (#933).
 - [BREAKING] Update `RemoteProverError::ConnectionFailed` variant to contain `Error` instead of `String` (#968).
+- Exposed `miden-prover-service` as a library (#956).
```
sergerad: Probably meant `miden-proving-service`

bin/proving-service/src/api/prover.rs
```diff
@@ -5,15 +5,49 @@ use miden_objects::{
...
-    utils::MIDEN_PROVING_SERVICE,
-};
+use crate::generated::{self, ProvingRequest, ProvingResponse, api_server::Api as ProverApi};
+
+pub const MIDEN_PROVING_SERVICE: &str = "miden-proving-service";
+
+/// Specifies the type of proving supported.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Serialize, Deserialize)]
+pub enum ProofType {
+    /// Transaction proving
```
sergerad: nit: Unnecessary comments maybe. Also tend to put a full stop at the end of the sentence.

bin/proving-service/src/generated/mod.rs
```diff
@@ -14,7 +14,7 @@ pub mod status;
 
 pub use proving_service::*;
 
-use crate::commands::worker::ProverType;
+use crate::api::ProofType as DomainProofType;
```
sergerad: What does "Domain" mean in this context? Any synonyms you also considered?

tomyrd: I used "domain" because it's the same term we use in the client (in `crates/rust-client/src/rpc/domain` directory for example). I belive domain in this context means that is the type used in the "crate's domain" as oposed to the generic rpc type. But I could change the name if we prefer, some other names that come to mind are `ConvertedProofType` or `ApiProofType`. I could also use the module name to separate the two types and change every instance of `DomainProofType` to `api::ProofType`.

igamigo: `domain` is also used in the node for the same reason I believe. However, we could choose to rename the proto enum here and differentiate them this way. I believe this is the [convention in the node](https://github.com/0xMiden/miden-node/blob/next/crates/proto/src/domain/block.rs#L14) for implementing conversions (so you get something like `impl From<BlockHeader> for proto::BlockHeader`)

tomyrd: While changing the `api::ProofType as DomainProofType` as just `ProofType` I had to move the conversions to a separate `conversions.rs` file. This is because the `ProofType` name was already being used by the mod re-export `pub use proving_service::*;`. In the separate file I can rename the `generated` mod to `proto` and use the `ProofType` for the "domain" type.

bin/proving-service/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+pub mod api;
+pub mod error;
+pub mod generated;
```
igamigo, line 3: I don't think this would need to be public, right?

bin/proving-service/src/api/prover.rs
```diff
@@ -159,15 +190,15 @@ impl ProverApi for ProverRpcApi {
         request: Request<ProvingRequest>,
     ) -> Result<Response<ProvingResponse>, tonic::Status> {
         match request.get_ref().proof_type() {
-            ProofType::Transaction => {
+            generated::ProofType::Transaction => {
```
igamigo: What do you think about aliasing `generated` as `proto` in this file? I think this is what we do in other similar gRPC code

igamigo (approved):
LGTM, left some comments but overall I think it's good to merge

bin/proving-service/src/lib.rs
tomyrd: These modules are needed on the [client side](https://github.com/0xMiden/miden-client/pull/950)

bin/proving-service/src/api/prover.rs
tomyrd: Yes! I agree, this added `generated::` was done in a previous commit and I didn't change it on [0d63839](https://github.com/0xMiden/miden-node/pull/956/commits/0d63839a17809b6aaa27a7dc17da7f65bcf38aa6)

CHANGELOG.md
```diff
@@ -16,26 +16,25 @@
...
-- Refresh Cargo.lock file (#944).
-
-## v0.9.0 (2025-05-30)
-
 ### Changes
 
 - Separated the store API into three separate services (#932).
 - Added a faucet Dockerfile (#933).
 - [BREAKING] Update `RemoteProverError::ConnectionFailed` variant to contain `Error` instead of `String` (#968).
+- Exposed `miden-proving-service` as a library (#956).
```
SantiagoPittella, line 24: Looks like we now have 2 `### Changes` sections.

SantiagoPittella (approved):
Looking good, just a small comment on the changelog.

bin/proving-service/src/lib.rs
TomasArrachea: I think only the `api` mod is being used, no?

tomyrd: You are right, my first answer was partly wrong, the reason for the `generated` and `error` modules being public is that they are used in the binary part of the crate. Although both the lib and bin are part of the same crate, they function as separate. So having the modules be `pub(crate)` in the lib is not enough for the bin part to use it, they also need to be public.

It is unfortunate but I couldn't find a way to go around this.

PR #957, opened 2025-06-11
ci: Consolidate and fix Debian Publish workflows

sergerad:
## Context

We now publish all of the following Debian packages from this repository:
- Node
- Faucet
- Prover
- Prover Proxy

## Changes

Replace the current workflows with a pair of workflows that consolidate Debian package publishing.

The pair of workflows are as such:
1. Publish a single Debian package specified from user input (manual trigger only); and
2. Publish all 4 Debian packages (manual and on-release trigger).

I have used the first workflow to publish v0.9.1 artefacts today.

Screenshot 2025-06-11 at 2 02 20 PM

### Future Work

Add mapping logic so that only a single input string is required to specify which package to publish.
```
      - name: Build and Publish Prover Proxy
        uses: ./.github/actions/debian
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          gitref: ${{ env.version }}
          crate_dir: node
          service: miden-node
          package: node
          crate: miden-node
          arch: ${{ matrix.arch }}
```
should become something like
```
      - name: Build and Publish Prover Proxy
        uses: ./.github/actions/debian
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          gitref: ${{ env.version }}
          service: node
          arch: ${{ matrix.arch }}
```


.github/workflows/publish-debian.yml
bobbinth, line 1: Is there a way to add comments somewhere explaining the difference between this and `publish-debian-all` workflow and maybe documenting how they are triggered/used?

bobbinth (approved):

.github/workflows/publish-debian.yml
sergerad: I try to avoid comments in declarative files like these YAMLs. If the names and descriptions are good enough, it should be obvious to the user what is going on. When you go to trigger the workflows you will see the descriptions of all the required inputs.

As for the triggers, just need familiarity on github actions triggers
```
on:
  release:
    types: [prereleased, released]

  workflow_dispatch:
```
Means on release and manual trigger.

sergerad:
Merging this as it has already been used to publish v0.9.1. I'll do followup PRs if any other comments come up.

PR #959, opened 2025-06-11
chore: Always chown dir in postinst

sergerad:


bobbinth (approved):
Looks good! Thank you. This is needed to set the permission level of directories correctly, right?

PR #962, opened 2025-06-11
fix(faucet): multiply asset base units

TomasArrachea:


bin/faucet/src/faucet/mod.rs
```diff
@@ -217,6 +218,12 @@ impl Faucet {
         )
         .expect("Empty ChainMmr should be valid");
 
+        let asset_decimals =
+            account.storage().get_item(2).expect("faucet should store metadata on slot 2")[1]
+                .as_int()
+                .try_into()
+                .expect("decimals should be less than 256");
```
bobbinth: Ideally, we shouldn't need to hardcode this. It could work like so:

```rust
let faucet_info = BasicFungibleFaucet::try_from(&account).expect("not a fungible faucet");
let decimals = faucet_info.decimals();
```

But we are missing a couple of things in `miden-base` to make this work:
- We only have `TryFrom<Account>` implement for `BasicFungibleFaucet`, and we need `TryFrom<&Account>`.
- `decimals` is a private field of `BasicFungibleFaucet` - so, we need to add an accessor method.

@TomasArrachea - could you make a quick PR in `miden-base` to make these changes and then use them in this PR?

TomasArrachea: PR in miden-base: https://github.com/0xMiden/miden-base/pull/1446

bobbinth: Merged!

TomasArrachea: updated to use the new accessor in [9c8f795](https://github.com/0xMiden/miden-node/pull/962/commits/9c8f795959fa0aa5b813dee2f027644c639275d1), looks better this way 👍

bobbinth (approved):
Looks good! Thank you!

PR #965, opened 2025-06-11
refactor: Valid nostd build for proving service client

igamigo:
From the client, in order to make the proving service library available from `wasm32` targets, the repo was missing the configuration for `getrandom`. This PR adds it alongside a check that verifies whether the proving service client library can be built for web environments.

Additionally, because the proving service library crate relies on `miden-objects` and `miden-tx`, we need to make sure they dont' come with default features because these include `std` which makes the build fail.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,10 +51,10 @@ miden-node-rpc               = { path = "crates/rpc", version = "0.10" }
 miden-node-store             = { path = "crates/store", version = "0.10" }
 miden-node-test-macro        = { path = "crates/test-macro" }
 miden-node-utils             = { path = "crates/utils", version = "0.10" }
-miden-objects                = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-objects                = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
 miden-proving-service-client = { path = "crates/proving-service-client", version = "0.10" }
 miden-testing                = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
-miden-tx                     = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-tx                     = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
 miden-tx-batch-prover        = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
 prost                        = { version = "0.13" }
 rand                         = { version = "0.9" }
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -51,10 +51,10 @@ miden-node-rpc               = { path = "crates/rpc", version = "0.10" }
 miden-node-store             = { path = "crates/store", version = "0.10" }
 miden-node-test-macro        = { path = "crates/test-macro" }
 miden-node-utils             = { path = "crates/utils", version = "0.10" }
-miden-objects                = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-objects                = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
 miden-proving-service-client = { path = "crates/proving-service-client", version = "0.10" }
 miden-testing                = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
-miden-tx                     = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-tx                     = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
```
bobbinth, line 57: I wonder if we should set `default-features = false` in the `Cargo.toml` for the proving service client. There, it could look something like:
```toml
miden-objects = { optional = true, workspace = true, default-features = false}
```

bobbinth (approved):

Cargo.toml
igamigo: I tried this at first but it did not work:

```
make build

BUILD_PROTO=1 cargo build --locked --workspace --exclude miden-proving-service # miden-tx async feature on.
error: failed to load manifest for workspace member `/Users/igamigo/repos/miden-node/bin/faucet`
referenced by workspace at `/Users/igamigo/repos/miden-node/Cargo.toml`

Caused by:
  failed to load manifest for dependency `miden-proving-service-client`

Caused by:
  failed to parse manifest at `/Users/igamigo/repos/miden-node/crates/proving-service-client/Cargo.toml`

Caused by:
  error inheriting `miden-objects` from workspace root manifest's `workspace.dependencies.miden-objects`

Caused by:
  `default-features = false` cannot override workspace's `default-features`
make: *** [build] Error 101
```

bobbinth: Ah - interesting! I'm wondering if turning off standard features would have impact on other crates. Specifically, we have multithreading enabled only in `std` context (e.g., multi-thread SMT construction). I guess we could specify `std` feature for other crates that use these dependencies - but maybe this is already happening anyway?

igamigo: I can turn default features back on for every other crate that uses either `miden-tx` or `miden-objects`

bobbinth: Let's do that to be safe.

bobbinth: We should also add `default-features = true` for the `store` crate.

sergerad (approved):

PR #969, opened 2025-06-12
feat(proving-service): add attestation verification

AmeanAsad:
## Overview

This PR adds Trusted Execution Environment (TEE) attestation report verification from [Lunal's](https://lunal.dev/) TEEs to proving service clients. This implementation came from @phklive's request to demonstrate an end-to-end flow for confidential delegated proving. With this PR, clients can verify that their delegated prover was run inside a TEE and thus preserved the input and processing privacy of their request. 

**Note**: We tested this end-to-end on versions 0.8.1 of the miden client and the miden prover proxy. The reason behind using these versions is for compatibility with the Miden Testnet. 

### Changes

**Attestation Verification Integration**

- Add TEE attestation report verification to the proving service client.
- Verify attestation reports attached in the `Attestation-Report` HTTP response header from TEE-enabled delegated provers.

**New Dependencies**

- [attestation-rs](https://github.com/lunal-dot-dev/attestation-rs) - Open source rust attestation verification library that verifies TEE attestation reports written by lunal.

### Architecture

1. Client submits a proving request to a delegated prover running inside a TEE.
2. The delegated prover in the TEE processes the client request and generates a proof.
3. The delegated prover returned the generated proof and a cryptographic attestation report in the HTTP response.
4. The client verifies the attestation report and confirms the proof was generated in a private, trusted environment.
5. Now verified, the standard proving workflow continues from there.

### **Backwards Compatibility**

- If an attestation report is present in the `Attestation-Report` HTTP response header? → Verify the attestation.
- If there’s no `Attestation-Report` response header or no attestation report in the `Attestation-Report` response headers? → Proceed without verification (legacy behavior).
- No breaking changes to existing client configurations.

### Testing Changes

To test:

- Spin up a delegated prover inside a TEE (this can be provided by Lunal folks). This will have an endpoint (e.g [`https://miden.lunal.dev/`](https://miden-prover.lunal.dev/))
- Initialize a local miden client (https://github.com/0xMiden/miden-client)
- In the `miden-client.toml` , set the `remote_rpc_endpoint` to the delegated prover url:
    
    ```toml
    remote_prover_endpoint = "https://miden-prover.lunal.dev"
    ```
    

## Migration Strategy

**Phase 1: Parallel Deployment**

- Deploy TEE-secured delegated prover(s) alongside existing, non-TEE-secured delegated provers.
- Separate, independent endpoints will exist in parallel for testing and gradual adoption.
- Full backward compatibility maintained.

**Phase 2: Testing & Validation**

- End-to-end, production testing with TEE delegated provers.
- Validate client-side attestation verification.
- Benchmark performance and reliability.

**Phase 3: Transition (Optional)**

- Gradual migration of all traffic to TEE-secured delegated provers.
- Optional requirement and enforcement of attestations by clients.
- Graceful deprecation of non-TEE, non-private delegated provers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,6 +61,126 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
+name = "alloy-json-abi"
+version = "0.8.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe6beff64ad0aa6ad1019a3db26fef565aefeb011736150ab73ed3366c3cfd1b"
+dependencies = [
+ "alloy-primitives",
+ "alloy-sol-type-parser",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
+name = "alloy-primitives"
+version = "0.8.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8c77490fe91a0ce933a1f219029521f20fc28c2c0ca95d53fa4da9c00b8d9d4e"
+dependencies = [
+ "alloy-rlp",
+ "bytes",
+ "cfg-if",
+ "const-hex",
+ "derive_more",
+ "foldhash",
+ "hashbrown 0.15.4",
+ "indexmap 2.9.0",
+ "itoa",
+ "k256",
+ "keccak-asm",
+ "paste",
+ "proptest",
+ "rand 0.8.5",
+ "ruint",
+ "rustc-hash",
+ "serde",
+ "sha3",
+ "tiny-keccak",
+]
+
+[[package]]
+name = "alloy-rlp"
+version = "0.3.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5f70d83b765fdc080dbcd4f4db70d8d23fe4761f2f02ebfa9146b833900634b4"
+dependencies = [
+ "arrayvec",
+ "bytes",
+]
+
+[[package]]
+name = "alloy-sol-macro"
+version = "0.8.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e10ae8e9a91d328ae954c22542415303919aabe976fe7a92eb06db1b68fd59f2"
+dependencies = [
+ "alloy-sol-macro-expander",
+ "alloy-sol-macro-input",
+ "proc-macro-error2",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.102",
+]
+
+[[package]]
+name = "alloy-sol-macro-expander"
+version = "0.8.25"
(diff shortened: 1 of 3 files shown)


AmeanAsad:
@phklive the cargo.lock is indeed the reason for the large diff, this is due to adding [attestation-rs](https://github.com/lunal-dot-dev/attestation-rs) (lunal's attestation crate) as a dependency. the actual code diffs are 13 lines.

AmeanAsad:
Great question @bobbinth! 

To provide some background, there is a lunal proxy that runs inside every TEE. That proxy is used to establish a TLS session with any client communicating with the TEE to prevent a man in the middle attack. TLS ensures that user data is encrypted while it's being transmitted to the TEE. Since TLS terminates inside the TEE, the data is only decrypted inside the TEE where it is protected by the TEE boundary from any external party. The attestation report also includes a nonce generated by the verification client in order to prevent replay attacks for attestation verification.

With that in mind, based on our discussions with the Miden team, Miden currently runs the delegate/remote prover as one big instance. Therefore, our current setup assumes that the client communicates directly with a remote prover running inside the TEE directly without intermediate parties. In that setup the flow is:
* Client establishes TLS session with the remote prover running inside the TEE directly. The client can then start sending encrypted requests directly to the TEE. 
* The client can optionally send an empty request to receive an attestation report from the remote prover. This proves to the client that they are talking to a legitimate TEE before they send any data.
* The client then can start sending proof delegation requests with their data which would be encrypted via TLS and decrypted only inside the TEE.
* The TEE would then process the request and send back the attestation proof
* The client verifies the attestation proof for the request. 

In the scenario you described where there is a remote prover proxy running independently that directs requests to multiple TEEs, there are options there to ensure that the user is always talking to a TEE:
* Option 1: Our recommended solution is to also run the remote prover inside a TEE as a proxy. This is the cleanest solution because it allows TLS to terminate inside the proxy without revealing anything at all about the request to an external party. The user would receive a composite attestation that proves that the remote prover proxy and the end server that processed the proof request both ran in TEEs.
* Option 2:  if you want to use an untrusted proxy (not running inside a TEE) with end-to-end encryption where:
    * The proxy connects to an available TEE and obtains it’s attestation report.
    * The client receives and verifies the attestation report to ensure that it’s a legitimate TEE that the prover is connecting it to.
    * The client extracts the TEE’s public key from the verified attestation report
    * The client encrypts their data using the extracted public key rom the target TEE
    * The untrusted proxy can route the encrypted requests to the appropriate TEE but cannot decrypt or inspect the actual user data since it lacks access to the TEE's private keys
    * This ensures that the encrypted data can only be decrypted inside the specific TEE that holds the corresponding private key, maintaining the security boundary even when routing through an untrusted proxy
    * Note that with this option, the untrusted proxy would terminate TLS still. So even though the client payload (proof data) would be encrypted, other information from the HTTP headers would be visible to that proxy so one can argue that this option provides less privacy for the end client. 

We support both options if required. For now, if the delegated prover is running as a single instance, the current setup suffices for that and we can switch to another setup if required later on. 

Hope that answers your question!

PR #972, opened 2025-06-13
ci: Add cache for Docker workflows

### LINKED ISSUE #277: [Feature]: Add caching to improve docker build times

End user

We need to add caching to the Dockerfile for the node to enable faster build times.

### How is this feature used?

This will decrease the time needed to build the docker image by adding a caching layer over rust and potentially other dependencies

### Anything else?

_No response_

sergerad:
Closes #277

Triggered the on-release workflow manually to test: https://github.com/0xMiden/miden-node/actions/runs/15625025989/job/44017494929

Packages successfully published here: https://github.com/orgs/0xMiden/packages?repo_name=miden-node

Uses AWS S3 as cache backend for the Docker layers.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -28,15 +28,6 @@ jobs:
           role-to-assume: ${{ secrets.AWS_ROLE }}
           role-session-name: GithubActionsSession
 
-      - name: Set cache key
-        id: cache-key
-        run: |
-          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
-            echo "cache_key=miden-${{ matrix.component }}-pr-${{ github.event.number }}" >> $GITHUB_OUTPUT
-          else
-            echo "cache_key=miden-${{ matrix.component }}-main" >> $GITHUB_OUTPUT
-          fi
-
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
         with:
@@ -47,5 +38,5 @@ jobs:
         with:
           push: false
           file: ./bin/${{ matrix.component }}/Dockerfile
-          cache-from: type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},name=${{ steps.cache-key.outputs.cache_key }}
-          cache-to: type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},name=${{ steps.cache-key.outputs.cache_key }},mode=max
+          cache-from: type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},name=miden-${{ matrix.component }}
+          cache-to: type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},name=miden-${{ matrix.component }}
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -47,5 +47,5 @@ jobs:
         with:
           push: false
           file: ./bin/${{ matrix.component }}/Dockerfile
-          cache-from: type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},${{ steps.cache-key.outputs.cache_key }}
-          cache-to: type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},${{ steps.cache-key.outputs.cache_key }},mode=max
+          cache-from: type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},name=${{ steps.cache-key.outputs.cache_key }}
+          cache-to: type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},name=${{ steps.cache-key.outputs.cache_key }},mode=max
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -15,6 +15,7 @@ jobs:
       matrix:
         component: [node, faucet]
     runs-on: ubuntu-latest
+    name: Build ${{ matrix.component }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/build-docker.yml
```diff
@@ -15,12 +15,20 @@ jobs:
...
       - name: Checkout code
         uses: actions/checkout@v4
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
+        with:
+          cache-binary: true
 
-      - name: Build Docker
-        run: make docker-build-${{ matrix.component }}
```
sergerad, line 26: While I like testing our scripts in CI, this change was necessary to add caching.

.github/workflows/build-docker.yml
Mirko-von-Leipzig, line 1: Some concerns about adding this caching.

GH cache is limited to 10 GB after which it evicts items in FIFO order. We've worked around this by only caching on `push` into `next`. A single PR caches around 6GB - _excluding_ this PR.

The reasoning is that _most_ PRs work off `next` and therefore having a stable `next` cache to pull from gives the greatest effect. If all branches/PRs/commits cache, then every 2nd cache insert would cause overflow and clobber the cache.

This PR seems to add another ~6GB to the GH cache, which are also split into a million little cache files for efficiency. This makes it a bit hard to reason about, but in the image below I've snapshotted the current state (GH will likely start evicting cache soon since its overflowing).

The rust-cache are 1-per job/worflow aka the first 5 items. The other 78 cache items are all from the docker build(s) and range in size from KB to ~500MB.

I don't really have a solution; I tried looking into the docker action's settings but its a lot of documentation so I gave up :smiling_face_with_tear: 

image

Mirko-von-Leipzig: TL;DR: I think every run of this new caching action will clobber the GH cache

Mirko-von-Leipzig: Maybe its time we add an sccache to our infrastructure?

This would be useful in every miden repo, and would remove a lot of the thinking required. And would also let us just freely (ab)use the caching on _every_ build instead of being conscientious with CI.

bobbinth: > Maybe its time we add an sccache to our infrastructure?

I haven't worked with `sccache` before. How hard would it be to integrate?

Mirko-von-Leipzig: tbh I've never worked with it from within github context, so I can't speak from experience.

Utilizing it locally is very easy; install `sccache` and create a cache directory; and set one (maybe two) env variables, and now your builds are all using this cache.

This cache directory can be an S3 bucket as well, so I don't think that parts difficult either.

This caching wouldn't extend to PRs from forks though. I think those don't get access to github secrets for security reasons; and therefore wouldn't be able to access S3. Which is probably very reasonable.

Having an S3 bucket for caching would also let us use it for the docker caching iiuc. Or otherwise we just use github cache for that exclusively.

sergerad: If I understand correctly, the suggestion to use sscache would mean we would move to self-hosted runners for github actions. That is almost certainly not worth the time and pain we would trade for the current build times we have.

I would be very surprised to find that we need to engineer our own solution for this - it should be a solved problem.

Is there a reason we wouldn't just use Dockerhub + registry-backed caching? I've never needed to use anything other than gha cache in the past but this must already be a solved problem.
```
--cache-to type=registry,ref=dockerhub.io/... \
--cache-from type=registry...
```

Or alternatively use S3 as a backend directly?
```
--cache-from: type=s3,bucket...
--cache-to: type=s3,bucket...
```

Mirko-von-Leipzig: > If I understand correctly, the suggestion to use sscache would mean we would move to self-hosted runners for github actions. That is almost certainly not worth the time and pain we would trade for the current build times we have.

I was hoping one could simply use S3 as the cache location, but still just build locally. I _think_ sccache might be able to operate concurrently without a central build server/co-ordinator but maybe I'm wrong.

It would be simpler to move the docker image caching -- but essentially what you're suggesting is what I was hoping we could do with sccache as well.

sergerad: The builtin cache backends mean we don't have to host any infrastructure for the purposes of our builds. Which is something I would really like to avoid.

sergerad: s3 backend seems to work https://github.com/0xMiden/miden-node/actions/runs/15666678419/job/44131692644?pr=972

Can see the blobs and metadata in s3

sergerad: One run (2 images) seemed to upload around 1GB to s3

.github/workflows/build-docker.yml
```diff
@@ -5,22 +5,47 @@ on:
...
+        with:
+          aws-region: ${{ secrets.AWS_REGION }}
+          role-to-assume: ${{ secrets.AWS_ROLE }}
+          role-session-name: GithubActionsSession
+
+      - name: Set cache key
+        id: cache-key
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            echo "cache_key=miden-${{ matrix.component }}-pr-${{ github.event.number }}" >> $GITHUB_OUTPUT
```
Mirko-von-Leipzig: Does this mean each PR gets its own unique cache? If so, isn't that an issue since we'll never get cache re-use?

.github/workflows/build-docker.yml
Mirko-von-Leipzig: What exactly gets cached? The builder don't seem to go any faster than without the cache. I imagine its because the bulk of the build time is the rust binary compilation which it redoes from scratch each time.

I don't really understand docker and its build process so I'm unsure if that's how it works.

.github/workflows/build-docker.yml
sergerad: Yea would expect the first run of the PR to be slower than subsequent ones with that setup. Have removed the cache key logic and set it to be universal across workflows in any case now.

.github/workflows/build-docker.yml
sergerad: I believe our Dockerfiles might need a slight refactor to allow for more layers. For example, this should be split into separate `RUN` statements:
```
# Update machine & install required packages
# The installation of sqlite3 is needed for correct function of the SQLite database
RUN apt-get update && \
    apt-get -y upgrade && \
    apt-get install -y --no-install-recommends \
    sqlite3 \
    && rm -rf /var/lib/apt/lists/*
```

I will try a few things in this PR.

sergerad: Actually that is incorrect. The merged `RUN` is actually recommended there. Also the order of operations in the Dockerfiles seems fine too. Will have a deeper look tomorrow.

sergerad:
Confirmed cache does work with maximal example (rerun successful job)
https://github.com/0xMiden/miden-node/actions/runs/15676690727/job/44198234925

Issue #960, opened 2025-06-11
Improve faucet PoW


Mirko-von-Leipzig:
Improve the fidelity of the PoW by

- Rate limiting on a per account basis by incorporating the account ID within the challenge.
- Support API keys with separate difficulty "curves" for each. The `None` API key is its own bucket.
  - Difficulty should be configurable per key

### Original context

>> Each individual wallet would use the same API key? In that case, a malicious client could extract the API key from the wallet to do a lot of requests and harm all the other users.
> 
> Yes, the idea is that all users of a given wallet software would share the same API key. This does mean that if someone extracts the key, they would be able to impact other users of the same wallet software, but at lest it won't impact other software (or the website). The alternative is not to have API keys at all - but I'm not sure it is better.
> 
>>  But in the other hand, if an API is created for each wallet, we could end up with a huge state.
> 
> We can already do this based on account ID. If we incorporate account ID into the challenge, we can ensure that only one challenge per user can be submitted in a given period of time. I think I mentioned this in some other comments. But let's create an issue for this. 

 _Originally posted by @bobbinth in [#878](https://github.com/0xMiden/miden-node/issues/878#issuecomment-2960799539)_

PR #974, opened 2025-06-13
feat: improve faucet pow

TomasArrachea:
closes https://github.com/0xMiden/miden-node/issues/960.

This PR refactors the way challenges are generated and validated in the faucet.
- Challenges now contain an account id. This allows using the challenge to rate limit by account.
- Challenges now contain the api key. Each api key will have an independent difficulty level, that will be increased by the usage of that key.


bin/faucet/src/server/challenge.rs
```diff
@@ -14,6 +16,8 @@ pub(crate) const CHALLENGE_LIFETIME_SECONDS: u64 = 30;
 pub struct Challenge {
     pub(crate) difficulty: usize,
     pub(crate) timestamp: u64,
+    pub(crate) account_id: [u8; AccountId::SERIALIZED_SIZE],
```
TomasArrachea: Storing the account id as bytes in the challenge because `AccountId` does not implement serde serialization

bin/faucet/src/server/pow.rs
```diff
@@ -20,13 +21,17 @@ const MAX_DIFFICULTY: usize = 24;
 /// The number of active requests to increase the difficulty by 1.
 const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_DIFFICULTY;
 
+/// The time window in seconds for rate limiting per account ID.
+/// Must be less than [`CHALLENGE_LIFETIME_SECONDS`] to effectively rate limit.
+const ACCOUNT_ID_RATE_LIMIT_WINDOW: u64 = 5;
```
TomasArrachea: Maybe this could be added to the rate limiter config in https://github.com/0xMiden/miden-node/issues/846. Not sure if we want two rate limiters by account anyway, or how would that be set

bin/faucet/src/main.rs
```diff
@@ -123,6 +124,7 @@ async fn main() -> anyhow::Result<()> {
     run_faucet_command(cli).await
 }
 
+#[allow(clippy::too_many_lines)] // This will be shorter once config file is removed
```
Mirko-von-Leipzig:
```suggestion
#[allow(
    clippy::too_many_lines, 
    reason = "This will be shorter once config file is removed"
)]
```

bin/faucet/src/server/mod.rs
```diff
@@ -59,8 +62,8 @@ pub struct Server {
     mint_state: GetTokensState,
     metadata: &'static Metadata,
     pow: PoW,
-    api_keys: BTreeSet<String>,
-    active_requests: Arc<AtomicUsize>,
+    api_keys: BTreeSet<ApiKey>,
+    active_requests_per_key: Arc<Mutex<HashMap<ApiKey, Arc<AtomicUsize>>>>,
```
Mirko-von-Leipzig: This feels like too many sync primitives

```suggestion
    active_requests_per_key: Arc<Mutex<HashMap<ApiKey, AtomicUsize>>>,
```

Mirko-von-Leipzig: And maybe we don't also need `api_keys` since its inside the hash map now?

bin/faucet/src/server/api_key.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct ApiKey([u8; 32]);
+
+impl ApiKey {
+    /// Generates a random API key.
+    pub fn generate() -> Self {
+        let mut rng = ChaCha20Rng::from_seed(rand::random());
+        let mut api_key = [0u8; 32];
+        rng.fill(&mut api_key);
+        Self(api_key)
+    }
```
Mirko-von-Leipzig: Should we rather

```rust
impl rand::Rng for ApiKey {..}
```

and then the caller can supply the rng etc. I always worry about methods creating their own rng internally; often leads to seed re-use or lack of proper randomness.

bin/faucet/src/server/api_key.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// Creates a new API key from a byte array.
+    pub fn new(api_key: [u8; 32]) -> Self {
+        Self(api_key)
+    }
```
Mirko-von-Leipzig: When would this get used?

bin/faucet/src/server/api_key.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// Returns the inner bytes of the API key.
+    pub fn inner(&self) -> [u8; 32] {
+        self.0
+    }
+
+    /// Decodes the API key from a string option. If the string is `None`, returns a default API
+    /// key.
+    pub fn decode(api_key_str: Option<String>) -> Result<Self, InvalidRequest> {
```
Mirko-von-Leipzig: I don't think this should take an option. Let the caller decide what to do e.g.

```rust
maybe_key_str
  .map(ApiKey::decode)
  .transpose()?
  .unwrap_or_default()
```

bin/faucet/src/server/get_tokens.rs
```diff
@@ -216,17 +212,27 @@ impl Drop for ActiveRequestGuard {
...
     State(server): State<Server>,
     Query(request): Query<RawMintRequest>,
-) -> Sse<impl Stream<Item = Result<Event, Infallible>>> {
+) -> Result<Sse<impl Stream<Item = Result<Event, Infallible>>>, GetTokenError> {
+    let api_key = ApiKey::decode(request.api_key.clone()).map_err(GetTokenError::InvalidRequest)?;
+
+    let key_active_requests = server
+        .active_requests_per_key
+        .lock()
+        .expect("active requests per key lock should be released")
```
Mirko-von-Leipzig: Are we using the right type of mutex here? Should this be `async`?

bin/faucet/src/server/get_tokens.rs
```diff
@@ -167,18 +165,16 @@ impl RawMintRequest {
...
             }
-            return Err(InvalidRequest::InvalidApiKey(api_key.clone()));
         }
 
         // Validate Challenge and nonce
         let challenge_str = self.challenge.ok_or(InvalidRequest::MissingPowParameters)?;
         let nonce = self.nonce.ok_or(InvalidRequest::MissingPowParameters)?;
 
-        server.submit_challenge(&challenge_str, nonce)?;
+        server.submit_challenge(&challenge_str, nonce, account_id, self.api_key)?;
```
Mirko-von-Leipzig: Should we not be submitting the actual decoded API key?

bin/faucet/src/server/get_tokens.rs
```diff
@@ -112,21 +115,16 @@ impl GetTokenError {
             },
         }
     }
+}
 
-    /// Convert the error into an SSE event and trigger a trace log.
-    fn into_event(self) -> Event {
+impl IntoResponse for GetTokenError {
+    fn into_response(self) -> Response {
```
Mirko-von-Leipzig: Does this still work the same as an event?

bin/faucet/src/server/challenge.rs
Mirko-von-Leipzig: `serde` does support [delegating](https://serde.rs/field-attrs.html#serialize_with) serialization so this should be possible to work-around.

Another alternative is new-type wrapper and then implementing it yourself.

bin/faucet/src/server/challenge.rs
```diff
@@ -91,17 +132,36 @@ impl Challenge {
...
+    pub fn validate_account_id(&self, account_id: AccountId) -> bool {
+        let account_id_bytes: [u8; AccountId::SERIALIZED_SIZE] = account_id.into();
+        self.account_id == account_id_bytes
+    }
+
+    /// Checks that the requested api key matches the one in the challenge.
+    pub fn validate_api_key(&self, api_key: &ApiKey) -> bool {
+        self.api_key == *api_key
+    }
+
```
Mirko-von-Leipzig: Nit: I don't really like these. Let the caller do the matching if he so desires.

If I see this validate at the call site, I would expect it to be doing a lot more than just checking its the same.

bin/faucet/src/server/pow.rs
```diff
@@ -129,6 +161,8 @@ impl PoW {
 struct ChallengeCache {
     /// Once a challenge is added, it cannot be submitted again.
     challenges: Arc<Mutex<HashSet<Challenge>>>,
+    /// A map of account IDs to timestamps of their last challenge submission.
+    account_id_timestamps: Arc<Mutex<HashMap<String, u64>>>,
```
Mirko-von-Leipzig: Could we not use `AccountId` instead of the string?

bin/faucet/src/server/api_key.rs
TomasArrachea: When we encode/decode the challenge, we just append the api key bytes. In the decoding, we want to build the api key from its bytes with this function

bin/faucet/src/server/pow.rs
TomasArrachea: I used the string because `AccountId` does not implement `Hash`

Mirko-von-Leipzig: You might be able to use `BTreeMap` instead.

@PhilippGackstatter could we consider adding `Hash` to the base objects? I believe this also applies to `TransactionId` etc. `HashMap` is significantly faster than `BTreeMap` (though that isn't applicable here; but it is elsewhere in the node).

bin/faucet/src/server/get_tokens.rs
TomasArrachea: I actually forgot that we were using events to avoid opaque errors on the frontend, I restored it. Still added the `IntoResponse` implementation to allow for early return if the api key is invalid 52c4c89689a599498faa510db93e816f3b6af223

bin/faucet/src/server/pow.rs
igamigo: A better alternative to `String` may be to use `u128`, but as Mirko said `BTreeMap` should work fine here (as well as using `AccountId` if `Hash` gets implemented)

bin/faucet/src/server/challenge.rs
TomasArrachea: I was able to workaround this using a `BTreeMap` instead of a `HashMap`, I think the deserializing was not really necessary and the Hash was also a problem here.

bin/faucet/src/server/get_tokens.rs
TomasArrachea: The lock is not being held between `await`s so using a sync mutex should be fine

bin/faucet/src/server/mod.rs
TomasArrachea: > This feels like too many sync primitives

I just replaced the `Arc<AtomicUsize>` with `usize` by removing the guard that was used before. Now the request handler just locks the mutex of the hashmap at the beginning and the end of the request ([279c507](https://github.com/0xMiden/miden-node/pull/974/commits/279c507a2f2c8c5d095061476c585550e5878a66))

bin/faucet/src/server/api_key.rs
TomasArrachea: Im not really sure of how the Rng trait works but I think in this case we should instead implement 
```
impl Distribution<ApiKey> for StandardUniform { .. }
```
Or a simpler alternative maybe changing the function to receive a seed?

bin/faucet/src/server/challenge.rs
```diff
@@ -34,46 +41,72 @@ impl Serialize for Challenge {
...
+            Ok(Self::from_parts(difficulty, timestamp, account_id, api_key, signature))
         } else {
             Err(InvalidRequest::ServerSignaturesDoNotMatch)
         }
     }
 
     /// Encodes the challenge into a hex string.
     pub fn encode(&self) -> String {
-        let mut bytes = Vec::with_capacity(48);
+        let mut bytes = Vec::with_capacity(63);
```
bobbinth: Question: why do we allocate 63 bytes here, but on line 83 above we decode this into 95 bytes?

bin/faucet/src/server/mod.rs
```diff
@@ -59,8 +62,7 @@ pub struct Server {
     mint_state: GetTokensState,
     metadata: &'static Metadata,
     pow: PoW,
-    api_keys: BTreeSet<String>,
-    active_requests: Arc<AtomicUsize>,
+    active_requests_per_key: Arc<Mutex<HashMap<ApiKey, usize>>>,
```
bobbinth: Should this not be in the `PoW` struct?

bin/faucet/src/server/pow.rs
```diff
@@ -128,7 +160,9 @@ impl PoW {
 #[derive(Clone, Default)]
 struct ChallengeCache {
     /// Once a challenge is added, it cannot be submitted again.
-    challenges: Arc<Mutex<HashSet<Challenge>>>,
+    challenges: Arc<Mutex<BTreeSet<Challenge>>>,
+    /// A map of account IDs to timestamps of their last challenge submission.
+    account_id_timestamps: Arc<Mutex<BTreeMap<AccountId, u64>>>,
 }
```
bobbinth: Related to the pervious comment, I was thinking that `ChallengeCache` would actually be the struct that keeps track of API keys, accounts etc.

The core idea is that we clean up challenge cache pretty frequently (e.g., once every couple of seconds). This would allow us to naturally impose rate limits on API keys and account IDs. For example, the cache could have the following interface:
```rust
impl ChallengeCache {

    pub fn has_challange_for_account(&self, account_id: AccountId) -> bool {
        ...
    }

    pub fn num_challenges_for_api_key(&self, key: ApiKey) -> usize {
        ...
    }
}
```

Then, when a request for a new challenge arrives, `PoW` would do the following:
- Check if the cache has a challenge for the account. If it does (i.e., the account recently solved a PoW puzzle), the request is rejected.
- Get the number of challenges for the API key, and compute the PoW difficulty based on that.

The implication of this is that an account would be able to make only one request in a given time interval - but I don't think that's a problem.

This would also mean that there is no need to explicitly adjust difficulty per key (assuming computing difficulty from the number of challenges is not too difficult).

bobbinth (commented):
Looks good! Thank you. Not a full review, but I left a couple of comments inline.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 - Added support for gRPC server side reflection to all components (#949).
 - Added support for TLS to `miden-proving-service-client` (#968).
 
+### Changes
+
+- Added PoW to faucet API Keys (#974).
```
SantiagoPittella: This should be `[BREAKING]`.

bin/faucet/src/server/pow.rs
TomasArrachea: I see, using the cache cleanup for the rate limiter makes sense. For this we don't actually need to embed the account id inside the challenge, right? Neither the api key. `PoW` receives the account id and api key from the minting request.

> The implication of this is that an account would be able to make only one request in a given time interval - but I don't think that's a problem.

Yes, using the cache does limit the flexibility of the rate limiter logic but this sound ok for now.

bobbinth: > For this we don't actually need to embed the account id inside the challenge, right? Neither the api key. `PoW` receives the account id and api key from the minting request.

If we don't embed them in the challenge we'll need to keep track of the mappings from API key and account ID to challenge somewhere, right? This could be a potential attack vector (e.g., make a lot of requests to saturate these maps w/o actually solving the PoW). Putting account ID and API key into the challenge means we don't have to store this information in memory until PoW challenge is solved.

TomasArrachea: That's true. But the `ChallengeCache` only stores solved challenges so I'm not sure if it would be vulnerable to keep those mappings.

bobbinth: But solved challenges is what we need, no? Basically, there should be only 1 solved challenge per user in the last $n$ (e.g., 30) seconds, and the difficulty for API keys should be determined by the number of solved challenges in the last $n$ seconds. Or am I missing something?

bin/faucet/src/server/challenge.rs
TomasArrachea: This was wrong, updated it to 95

bin/faucet/src/server/pow.rs
TomasArrachea: Yes, correct. I pushed this changes to use the `ChallengeCache` for retrieving the challenge count per api key and  challenge per account. I realized that it's better to keep the account id and api key embedded in the challenge anyway, it's needed for verifying that they coincide on challenge creation and submission. i.e you can't use an account id to create a challenge and then use that challenge to mint tokens for a different account.

bin/faucet/src/server/pow.rs
```diff
@@ -20,13 +25,16 @@ const MAX_DIFFICULTY: usize = 24;
...
+
 // POW
 // ================================================================================================
 
 #[derive(Clone)]
 pub(crate) struct PoW {
     secret: [u8; 32],
-    difficulty: Arc<AtomicUsize>,
+    difficulty_per_key: Arc<Mutex<HashMap<ApiKey, usize>>>,
     challenge_cache: ChallengeCache,
```
bobbinth: I would probably put `difficulty_per_key` into the `ChallengeCache` and put the mutex outside of the challenge cache - e.g.:
```rust
pub(crate) struct PoW {
    secret: [u8; 32],
    challenge_cache: Arc<Mutex<ChallengeCache>>,
}
```
This way, we'd need only one mutex.

bin/faucet/src/server/pow.rs
```diff
@@ -128,10 +166,35 @@ impl PoW {
 #[derive(Clone, Default)]
 struct ChallengeCache {
     /// Once a challenge is added, it cannot be submitted again.
-    challenges: Arc<Mutex<HashSet<Challenge>>>,
+    challenges: Arc<Mutex<BTreeSet<Challenge>>>,
 }
```
bobbinth: Related to the previous comment, if we put the mutex on the outside, we could get rid of the internal mutex.

Also, we actually don't need to store full challenges. We could simplify this as follows:
```rust
struct ChallengeCache {
    /// Maps challenge timestamp to a tuple of AccountID and ApiKey.
    challenges: BTreeMap<u64, (AccountId, ApiKey)>,
    /// Maps ApiKey to the number of solved challenges.
    challenges_per_key: BTreeMap<ApiKey, usize>,
    /// Keeps track of accounts which have recently solved challenges.
    account_ids: BTreeSet<AccountId>,
}
```
This structure should make it quite a bit more efficient both to insert new challenges and periodically truncate the cache.

TomasArrachea: I don't follow why could be just use the challenges' timestamps. Shouldn't we store different challenges even though they have the same timestamp?

bin/faucet/src/server/api_key.rs
TomasArrachea: Done in [82d3adb](https://github.com/0xMiden/miden-node/pull/974/commits/82d3adbcd7a932c5257c12e9aaa90160521b782f)

bin/faucet/src/server/pow.rs
bobbinth: Ah yes! It should have been `BTreeMap<u64, Vec<(AccountId, ApiKey)>>`.

bobbinth: The nice thing about using a `BTreeMap` here is that the clean-up should be pretty efficient (i.e., no need to iterate over all challenges).

bin/faucet/src/server/pow.rs
```diff
@@ -127,37 +166,79 @@ impl PoW {
 /// Challenges get removed periodically.
 #[derive(Clone, Default)]
 struct ChallengeCache {
```
bobbinth, line 148: I would maybe update doc comments for this struct.

bin/faucet/src/server/pow.rs
```diff
@@ -127,37 +166,79 @@ impl PoW {
...
-        let mut challenges = self.challenges.lock().unwrap();
-        challenges
-            .retain(|challenge| (current_time - challenge.timestamp) <= CHALLENGE_LIFETIME_SECONDS);
+        self.challenges.retain(|challenge| {
+            let expired = (current_time - challenge.timestamp) >= CHALLENGE_LIFETIME_SECONDS;
+            if expired {
+                self.challenges_per_key
+                    .entry(challenge.api_key.clone())
+                    .and_modify(|c| *c = c.saturating_sub(1))
+                    .or_insert(0);
```
bobbinth: Is there a reason to insert `0` if the key doesn't exist? I would think the logic should be:

- If the key exists, decrement the count.
  - If the count wen't down to 0, remove the entry from the map.
- if the key doesn't exist, it is a bug - so, probably, panic.

bin/faucet/src/server/pow.rs
```diff
@@ -127,37 +166,79 @@ impl PoW {
...
+    pub fn num_challenges_for_api_key(&self, key: &ApiKey) -> usize {
+        self.challenges_per_key.get(key).copied().unwrap_or(0)
+    }
+
     /// Cleanup expired challenges.
     ///
     /// Challenges are expired if they are older than [`CHALLENGE_LIFETIME_SECONDS`]
     /// seconds.
-    pub fn cleanup_expired_challenges(&self) {
+    pub fn cleanup_expired_challenges(&mut self) {
```
bobbinth: nit: I would change this function to accept the current time as a parameter.

bin/faucet/src/server/pow.rs
```diff
@@ -127,37 +166,79 @@ impl PoW {
...
-        challenges
-            .retain(|challenge| (current_time - challenge.timestamp) <= CHALLENGE_LIFETIME_SECONDS);
+        self.challenges.retain(|challenge| {
+            let expired = (current_time - challenge.timestamp) >= CHALLENGE_LIFETIME_SECONDS;
+            if expired {
+                self.challenges_per_key
+                    .entry(challenge.api_key.clone())
+                    .and_modify(|c| *c = c.saturating_sub(1))
+                    .or_insert(0);
+                self.account_ids.remove(&challenge.account_id);
```
bobbinth: This makes me think that we should also track accounts as `BTreeMap<AccountId, usize>`. The reason is that with the `BTreeSet` it is technically possible to add a challenge for the same account twice, but there would be only one entry in the set. And so, when we do the clean-up, it'll remove the account even though there is a challenge for it in the map.

bin/faucet/src/server/pow.rs
```diff
@@ -20,66 +25,84 @@ const MAX_DIFFICULTY: usize = 24;
...
+    /// found, the difficulty is defaulted to 1.
+    pub fn build_challenge(&self, request: PowRequest) -> Result<Challenge, InvalidPowRequest> {
+        if self
+            .challenge_cache
+            .lock()
+            .expect("challenge cache lock should not be poisoned")
+            .has_challenge_for_account(request.account_id)
+        {
+            return Err(InvalidPowRequest::RateLimited);
+        }
```
bobbinth: Do we need this check? We should be able to issue as many challenges as the user wants.

bin/faucet/src/server/pow.rs
```diff
@@ -20,66 +25,84 @@ const MAX_DIFFICULTY: usize = 24;
 /// The number of active requests to increase the difficulty by 1.
 const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_DIFFICULTY;
 
+/// The interval at which the challenge cache is cleaned up.
+const CLEANUP_INTERVAL_SECONDS: u64 = 5;
```
bobbinth: I would reduce this to 2 seconds.

bin/faucet/src/server/pow.rs
```diff
@@ -88,28 +111,44 @@ impl PoW {
         timestamp: u64,
         challenge: &str,
         nonce: u64,
-    ) -> Result<(), InvalidRequest> {
+        account_id: AccountId,
+        api_key: &ApiKey,
+    ) -> Result<(), InvalidMintRequest> {
```
bobbinth: nit: I would rename and reorder the parameters as follows:
```rust
pub(crate) fn submit_challenge(
    &self,
    account_id: AccountId,
    api_key: &ApiKey,
    challenge: &str,
    nonce: u64,
    current_time: u64,
) -> Result<(), InvalidMintRequest>
```

bobbinth (commented):

bin/faucet/src/server/pow.rs
TomasArrachea: And wouldn't using this map allow for a user to potentially submit the same challenge twice? Cause we have no way of knowing if a challenge is present in the cache. Perhaps we could have a `BTreeMap<u64, (Challenge, AccountId, ApiKey)>`, and a separate `BTreeSet<Challenge>`

bobbinth: I don't think that should be a problem. `Challenge` right now is:
```rust
pub struct Challenge {
    pub(crate) difficulty: usize,
    pub(crate) timestamp: u64,
    pub(crate) account_id: AccountId,
    pub(crate) api_key: ApiKey,
    pub(crate) signature: [u8; 32],
}
```
`signature` is deterministically derived from the other fields. So, the only difference between `Challenge` and `(timestamp, account_id, api_key)` is the `difficulty` field - which is probably would be the same for challenges issued at the same time.

So, we can use `(timestamp, account_id, api_key)` as a proxy for the `Challenge`, and we should make sure it is not possible to insert more than one such combination into the cache. This would effectively be the same as making sure we cannot insert more than one `Challenge` into the cache.

TomasArrachea: Right! Thanks.

bin/faucet/src/server/pow.rs
```diff
@@ -119,45 +139,121 @@ impl PoW {
...
-            .retain(|challenge| (current_time - challenge.timestamp) <= CHALLENGE_LIFETIME_SECONDS);
+        let expired_challenges: Vec<u64> = self
+            .challenges
+            .range(..limit_timestamp)
+            .map(|(timestamp, _)| *timestamp)
+            .collect();
+
+        for timestamp in expired_challenges {
+            if let Some(issuers) = self.challenges.remove(&timestamp) {
+                for (account_id, api_key) in issuers {
```
bobbinth: Maybe a simpler way to do this could be something like:
```rust
for issuers in self.challenges.split_off(&limit_timestamp).into_values() {
    for (account_id, api_key) in issuers {
        // inner loop
    }
}
```
But we should make sure this works as expected (I think we already have a test for this).

bin/faucet/src/server/pow.rs
```diff
@@ -20,55 +21,55 @@ const MAX_DIFFICULTY: usize = 24;
 /// The number of active requests to increase the difficulty by 1.
 const ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY: usize = REQUESTS_QUEUE_SIZE / MAX_DIFFICULTY;
 
+/// The interval at which the challenge cache is cleaned up.
+const CLEANUP_INTERVAL_SECONDS: u64 = 2;
```
bobbinth, line 25: This is another thing that could be configured via env variables (for a different PR).

bin/faucet/src/server/pow.rs
```diff
@@ -80,36 +81,55 @@ impl PoW {
...
+
+        // Check if account has recently submitted a challenge.
+        if self
+            .challenge_cache
+            .lock()
+            .expect("challenge cache lock should not be poisoned")
+            .has_challenge_for_account(account_id)
+        {
+            return Err(InvalidMintRequest::RateLimited);
         }
```
bobbinth: I would try to combine this block with the next one so that we don't have to release and re-acquire the lock. Something like:
```rust
let challenge_cache = self.challenge_cache.lock().expect("...");
if challenge_cache.has_challenge_for_account(account_id) {
    return Err(InvalidMintRequest::RateLimited);
}

if !challenge_cache.insert_challenge(&challenge) {
    return Err(InvalidMintRequest::ChallengeAlreadyUsed);
}
```

bin/faucet/src/server/pow.rs
```diff
@@ -20,55 +21,55 @@ const MAX_DIFFICULTY: usize = 24;
...
-    /// The difficulty is increased by 1 for every `ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY` active
-    /// requests. The difficulty is clamped between 1 and `MAX_DIFFICULTY`.
-    pub fn adjust_difficulty(&self, active_requests: usize) {
-        let new_difficulty =
-            (active_requests / ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY).clamp(1, MAX_DIFFICULTY);
-        self.difficulty.store(new_difficulty, Ordering::Relaxed);
+    /// The difficulty is adjusted based on the number of active requests per API key. The
+    /// difficulty is increased by 1 for every `ACTIVE_REQUESTS_TO_INCREASE_DIFFICULTY` active
+    /// requests, and it is clamped between 1 and `MAX_DIFFICULTY`.
+    pub fn get_difficulty(&self, api_key: &ApiKey) -> usize {
```
bobbinth: nit: does this need to be public?

bin/faucet/src/server/challenge.rs
```diff
@@ -34,46 +41,73 @@ impl Serialize for Challenge {
...
     /// The signature is computed internally using the provided secret.
-    pub fn new(difficulty: usize, timestamp: u64, secret: [u8; 32]) -> Self {
-        let signature = Self::compute_signature(secret, difficulty, timestamp);
-        Self { difficulty, timestamp, signature }
+    pub fn new(
+        difficulty: usize,
+        timestamp: u64,
+        secret: [u8; 32],
+        account_id: AccountId,
+        api_key: ApiKey,
```
bobbinth: nit: I would re-order parameters here to put the `secret` at the end.

bin/faucet/src/server/challenge.rs
```diff
@@ -34,46 +41,73 @@ impl Serialize for Challenge {
...
         } else {
-            Err(InvalidRequest::ServerSignaturesDoNotMatch)
+            Err(InvalidMintRequest::ServerSignaturesDoNotMatch)
         }
     }
 
     /// Encodes the challenge into a hex string.
     pub fn encode(&self) -> String {
-        let mut bytes = Vec::with_capacity(48);
+        let mut bytes = Vec::with_capacity(95);
```
bobbinth: nit: since we use it in a couple of places, i'd make `95` a constant.

bin/faucet/src/server/pow.rs
TomasArrachea: not really

bin/faucet/src/server/pow.rs
TomasArrachea: this can be tackled with https://github.com/0xMiden/miden-node/issues/846

bobbinth (approved):
All looks good! Thank you! @Mirko-von-Leipzig, @igamigo - would you like to take another look?

Also, once these and some other faucet-related changes are merged, we should deploy it to the devnet and test out to make sure everything works as expected.

bin/faucet/src/server/api_key.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /// Returns the inner bytes of the API key.
+    pub fn inner(&self) -> [u8; 32] {
+        self.0
+    }
+
+    /// Decodes the API key from a string.
+    pub fn decode(api_key_str: &str) -> Result<Self, InvalidMintRequest> {
+        let api_key_str = api_key_str.trim_start_matches(API_KEY_PREFIX).to_string();
```
igamigo, line 44: If the key does not start with `API_KEY_PREFIX `, this will continue. Is this expected? Not entirely familiar but it seems we expect the prefix to always be there.

bin/faucet/src/server/api_key.rs
```diff
@@ -0,0 +1,71 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::server::get_tokens::InvalidMintRequest;
+
+// API KEY
+// ================================================================================================
+
+const API_KEY_PREFIX: &str = "miden_faucet_";
+
+/// The API key is a random 32-byte array.
```
igamigo, line 12: nit: I think this could contain a mention to how it can be encoded/decoded as/from a string.

bin/faucet/src/server/challenge.rs
```diff
@@ -92,28 +129,53 @@ impl Challenge {
     }
 
     /// Checks if the challenge timestamp is expired.
+    ///
+    /// Panics if the challenge timestamp is greater than the current time.
```
igamigo: Is this true? I think on debug builds, the subtraction would panic if it underflows, but otherwise it will wrap around (and probably return `true` as a result). 

Also, could I forge a challenge with an invalid timestamp to trigger the underflow (though maybe not problematic on debug builds)? If so, we should do a sanity check.

bin/faucet/src/server/challenge.rs
```diff
@@ -92,28 +129,53 @@ impl Challenge {
...
     pub fn is_expired(&self, current_time: u64) -> bool {
```
igamigo: This is not from this PR (and I think it could be tackled separately), but it seems using `u64` here could be replaced by something more self-documenting (like `SystemTime`, `Duration` for the challenge lifetime, etc)

bin/faucet/src/server/get_pow.rs
```diff
@@ -0,0 +1,83 @@
...
+        Ok(PowRequest { account_id, api_key })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum InvalidPowRequest {
+    #[error("account address failed to parse")]
+    InvalidAccount(#[source] AccountIdError),
+    #[error("API key failed to parse")]
+    InvalidApiKey,
```
igamigo: Does it make sense to add the API key here? I think it could be potentially valuable for debugging errors when the frontend and backend are more decoupled (for example, you may think you are sending a valid key from a frontend but the backend somehow receives it differently than you were expecting)

bin/faucet/src/server/get_tokens.rs
```diff
@@ -56,7 +51,7 @@ pub struct RawMintRequest {
 }
 
 #[derive(Debug, thiserror::Error)]
-pub enum InvalidRequest {
+pub enum InvalidMintRequest {
```
igamigo: nit: This was also from before, but I think I'd name this `MintRequestError`, and rename `InvalidPowRequest` to `PowRequestError`, but feel free to disregard.

bin/faucet/src/server/get_tokens.rs
```diff
@@ -73,10 +68,12 @@ pub enum InvalidRequest {
     ExpiredServerTimestamp(u64, u64),
     #[error("challenge already used")]
     ChallengeAlreadyUsed,
+    #[error("account is rate limited")]
+    RateLimited,
```
igamigo, line 72: For future work, something useful here could be to return the `Duration` after which the account will be good to request again.

bin/faucet/src/server/pow.rs
```diff
@@ -119,45 +134,113 @@ impl PoW {
...
+    /// account and the API key.
     ///
-    /// Challenges are expired if they are older than [`CHALLENGE_LIFETIME_SECONDS`]
-    /// seconds.
-    pub fn cleanup_expired_challenges(&self) {
-        let current_time = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("current timestamp should be greater than unix epoch")
-            .as_secs();
+    /// Returns true if the challenge was newly inserted.
```
igamigo: nit (feel free to disregard if you prefer the current version):
```suggestion
    /// Returns `false` if the challenge was already present in the cache, and `true` otherwise.
```

bin/faucet/src/server/pow.rs
```diff
@@ -119,45 +134,113 @@ impl PoW {
...
+    pub fn num_challenges_for_api_key(&self, key: &ApiKey) -> usize {
+        self.challenges_per_key.get(key).copied().unwrap_or(0)
+    }
+
+    /// Cleanup expired challenges and update the number of challenges submitted per API key and
+    /// account id.
+    ///
+    /// Challenges are expired if they are older than [`CHALLENGE_LIFETIME_SECONDS`] seconds.
+    fn cleanup_expired_challenges(&mut self, current_time: u64) {
+        let limit_timestamp = current_time - CHALLENGE_LIFETIME_SECONDS;
```
igamigo: Same comment here about `u64` being poorly-documented for time comparisons (maybe we should add what `current_time` is expected to be in the function doc comments)

igamigo (approved):
Left some mostly minor comments but overall LGTM!

bin/faucet/src/server/api_key.rs
TomasArrachea: The prefix is just a visual aid for users and it's not really needed for decoding. I think there is no reason to reject the api key if the prefix is trimmed, though I doubt anyone would need to do that.

bin/faucet/src/server/challenge.rs
TomasArrachea: You are right, it wouldn't panic in release. Such a challenge could be created manually (actually that is what I was doing in a test when bumped into this panic), but not from the PoW struct since it would need to be created with a future timestamp. I added the panic since it would mean a bug.

bin/faucet/src/server/pow.rs
TomasArrachea: I updated it to be a bit more clear. I took inspiration from the rust std `HashSet::insert()` docs

bin/faucet/src/server/challenge.rs
TomasArrachea: I added a comment on these current_time args

bin/faucet/src/server/pow.rs
TomasArrachea: Turns out this is not correct, since `split_off` returns the keys bigger than the cutoff, the oposite of what we need. The test was passing because it lacked a sleep to give the cache time to cleanup.

bin/faucet/src/server/resources/index.js
```diff
@@ -132,15 +138,14 @@ document.addEventListener('DOMContentLoaded', function () {
...
-
         const evtSource = new EventSource(window.location.href + 'get_tokens?' + new URLSearchParams(params));
 
         evtSource.onopen = function () {
             status.textContent = "Request on queue...";
         };
 
         evtSource.onerror = function (_) {
-            // Either rate limit exceeded or invalid account id. The error event does not contain the reason.
+            // Either rate limit exceeded or invalid account id. The error event does not contain the reason.
```
SantiagoPittella: nit: extra space

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - [BREAKING] Consolidate default account filepath for bundled bootstrap and faucet start commands to `account.mac` (#976).
 - [BREAKING] Remove default value account filepath for faucet commands and rename --output-path to --output (#976).
 - Exposed `miden-proving-service` as a library (#956).
+- [BREAKING] Added PoW to faucet API Keys (#974).
```
SantiagoPittella: This sounds a bit odd, we aren't adding PoW to the API Keys itself. I would phrase like:
```suggestion
- [BREAKING] Enforce `PoW` on all faucet API key-authenticated requests (#974).
```

bin/faucet/src/server/api_key.rs
```diff
@@ -0,0 +1,78 @@
+use base64::{Engine, prelude::BASE64_STANDARD};
+use rand::Rng;
+use serde::{Deserialize, Serialize};
+
+use crate::server::get_tokens::MintRequestError;
+
+// API KEY
+// ================================================================================================
+
+const API_KEY_PREFIX: &str = "miden_faucet_";
```
SantiagoPittella, line 10: I don't think that we need a prefix for the API keys, right?

bin/faucet/src/server/challenge.rs
Mirko-von-Leipzig: You definitely don't want to panic. Its entirely possible this could happen even during normal operation - you should not trust system time to behave well.

I would just `unwrap_or(u64::MAX)`.

bin/faucet/src/server/api_key.rs
TomasArrachea: It's not needed, it was added just for convenience for users.

bin/faucet/src/server/get_pow.rs
TomasArrachea: This requires a bit of a change since it needs a new method in the cache to return the timestamp of a challenge given an account and api key. We can implement this in a follow-up PR to avoid expanding this PR even more.

bin/faucet/src/server/get_tokens.rs
TomasArrachea: This would also be good for the UX, but this change requires a bit of a change since it needs a new method in the cache that returns the challenge timestamp for a given account. I think it would be better to leave this for a follow-up PR to avoid expanding this one even more.

Mirko-von-Leipzig (approved):

PR #975, opened 2025-06-13
chore(prover): standardize proto

### LINKED ISSUE #946: Standardize Proving Service proto codegen
### Context

We are adding proving service crates to miden-node in #926.

The way the proto code gen is handled by those crates is different to the rest of miden-node. We want to standardize this approach.

See https://github.com/0xMiden/miden-node/pull/926#issuecomment-2958330252

### Work to do

Update the way proto code gen is handled for proving-service and proving-service-client. Note these may be renamed to delegated-prover and delegated-prover-client by the time this issue is addressed.

The rest of miden-node uses `proto/proto/build.rs` so these crates should do the same.

SantiagoPittella:
closes #946


crates/proving-service-client/build.rs
```diff
@@ -1,81 +1,61 @@
...
+        .into_diagnostic()
+}
 
-    // Replace `std` references with `core` and `alloc` in `proving_service.rs`.
-    // (Only for nostd version)
-    let nostd_file_path = nostd_path.join("proving_service.rs");
-    let file_content = fs::read_to_string(&nostd_file_path).into_diagnostic()?;
+/// Replaces std references with core and alloc for nostd compatibility
+fn convert_to_nostd(file_path: &str) -> miette::Result<()> {
+    let file_content = fs::read_to_string(file_path).into_diagnostic()?;
```
Mirko-von-Leipzig, line 58: I'm surprised there isn't support for this within `tonic` itself?

Mirko-von-Leipzig (approved):

crates/proving-service-client/build.rs
SantiagoPittella: Sadly I didn't find anything that that performs this directly from the tonic building. In the miden-client we are doing the same for the web client

bin/proving-service/build.rs
```diff
@@ -1,93 +1,42 @@
...
-) -> miette::Result<()> {
+/// Builds tonic code from a `FileDescriptorSet`
+fn build_tonic_from_descriptor(descriptor: FileDescriptorSet) -> miette::Result<()> {
     tonic_build::configure()
-        .file_descriptor_set_path(file_descriptor_path)
-        .skip_protoc_run()
-        .out_dir(out_dir)
-        .build_server(true)
+        .out_dir(GENERATED_OUT_DIR)
+        .build_server(true) // this setting generates only the client side of the rpc api
```
bobbinth: Is the comment here accurate? It seems like this should generate only the server code - or am I missing something?

crates/proving-service-client/build.rs
```diff
@@ -1,81 +1,61 @@
...
-    build_tonic_client(&file_descriptor_path, &nostd_path, protos, includes, true)?;
+/// Builds tonic code from a `FileDescriptorSet` with specified configuration
+fn build_tonic_from_descriptor(
+    descriptor: FileDescriptorSet,
+    out_dir: String,
+    build_transport: bool,
+) -> miette::Result<()> {
+    tonic_build::configure()
+        .out_dir(out_dir)
+        .build_server(false)
```
bobbinth, line 50: I think this setting disabled building of the server files, right?

proto/src/lib.rs
```diff
@@ -31,3 +39,11 @@ pub fn ntx_builder_api_descriptor() -> FileDescriptorSet {
     FileDescriptorSet::decode(&bytes[..])
         .expect("bytes should be a valid file descriptor created by build.rs")
 }
+
+/// Returns the Protobuf file descriptor for the worker status API.
+#[cfg(feature = "internal")]
+pub fn worker_status_api_descriptor() -> FileDescriptorSet {
```
bobbinth, line 45: Not for this PR, but I think "worker status" is somewhat ambiguous. We should probably call it "prover worker" or something like that - but I think there is a different issue for this.

bobbinth (approved):

After this is merged (and #946 is closed), we should probably open another issue to discuss whether we can standardize things further.

bin/proving-service/build.rs
SantiagoPittella: My bad, this comments is wrong.

crates/proving-service-client/build.rs
SantiagoPittella: Yes, it does disables the server generation.

proto/src/lib.rs
SantiagoPittella: Should this comment be addressed with #993 ?

Issue #836, opened 2025-05-13
Configurable genesis for testing


mmagician:
### To whom is this feature for?

App developer

### Why is is this feature needed?

Aside from producing repeatable genesis blocks, a configurable genesis is useful for integration testing. E.g., to even begin testing a SWAP transaction on a locally running chain, one needs to spend almost 1 minute running the basic setup of assets (two faucets, each minting a note for Alice & Bob, then the users claiming the notes). This slows down the development and testing cycle.

It would be convenient to have a genesis config file that allows specifying the required faucets, wallet accounts, and assets.

Something like: https://github.com/mmagician/miden-node/pull/1, though I started developing against the `main` branch. Related to https://github.com/0xMiden/miden-node/pull/776, which removed the `toml` genesis config (for now?)

### How is this feature used?

Integration tests with a local node

### Anything else?

Bringing back an optional `genesis.toml` doesn't conflict with the removal of `genesis.dat` as in https://github.com/0xMiden/miden-node/issues/767, since we can still create a new `GenesisState` from the accounts list that we construct directly based on `genesis.toml`.

Bonus: This is also a useful step towards a mainnet genesis event, where typically a list of accounts with balances (and lockups? probably not a `BasicWallet` anymore) is specified alongside the (single?) faucet.

bobbinth:
Closed by #1000.

PR #976, opened 2025-06-16
refactor: Replace Faucet configuration file and Init command

sergerad:
WIP

### DIFF
--- a/bin/faucet/.env
+++ b/bin/faucet/.env
@@ -1,2 +1,9 @@
-# We will standardize towards putting configuration into env vars through these files.
-# Automation expects a .env for each service we deploy. So this empty file will remain.
+MIDEN_FAUCET_ENDPOINT=http://0.0.0.0:8080
+MIDEN_FAUCET_NODE_URL=http://127.0.0.1:57291
+MIDEN_FAUCET_TIMEOUT_MS=5000
+MIDEN_FAUCET_ACCOUNT_PATH="accounts/faucet.mac"
+MIDEN_FAUCET_ASSET_AMOUNTS=""
+MIDEN_FAUCET_REMOTE_TX_PROVER_URL=""
+MIDEN_FAUCET_POW_SECRET=""
+MIDEN_FAUCET_API_KEYS=""
+MIDEN_FAUCET_ENABLE_OTEL=false
--- a/bin/faucet/src/main.rs
+++ b/bin/faucet/src/main.rs
@@ -14,9 +14,7 @@ use base64::{Engine, prelude::BASE64_STANDARD};
 use clap::{Parser, Subcommand};
 use faucet::Faucet;
 use miden_lib::{AuthScheme, account::faucets::create_basic_fungible_faucet};
-use miden_node_utils::{
-    config::load_config, crypto::get_rpo_random_coin, logging::OpenTelemetry, version::LongVersion,
-};
+use miden_node_utils::{crypto::get_rpo_random_coin, logging::OpenTelemetry, version::LongVersion};
 use miden_objects::{
     Felt,
     account::{AccountFile, AccountStorageMode, AuthSecretKey, NetworkId},
@@ -28,18 +26,25 @@ use rand_chacha::ChaCha20Rng;
 use rpc_client::RpcClient;
 use server::Server;
 use tokio::sync::mpsc;
+use types::AssetOptions;
 use url::Url;
 
-use crate::config::{DEFAULT_FAUCET_ACCOUNT_PATH, FaucetConfig};
+use crate::config::DEFAULT_FAUCET_ACCOUNT_PATH;
 
 // CONSTANTS
 // =================================================================================================
 
 const COMPONENT: &str = "miden-faucet";
-const FAUCET_CONFIG_FILE_PATH: &str = "miden-faucet.toml";
+const ENV_ENDPOINT: &str = "MIDEN_FAUCET_ENDPOINT";
+const ENV_NODE_URL: &str = "MIDEN_FAUCET_NODE_URL";
+const ENV_TIMEOUT: &str = "MIDEN_FAUCET_TIMEOUT_MS";
+const ENV_ACCOUNT_PATH: &str = "MIDEN_FAUCET_ACCOUNT_PATH";
+const ENV_ASSET_AMOUNTS: &str = "MIDEN_FAUCET_ASSET_AMOUNTS";
+const ENV_REMOTE_TX_PROVER_URL: &str = "MIDEN_FAUCET_REMOTE_TX_PROVER_URL";
+const ENV_POW_SECRET: &str = "MIDEN_FAUCET_POW_SECRET";
+const ENV_API_KEYS: &str = "MIDEN_FAUCET_API_KEYS";
 const ENV_ENABLE_OTEL: &str = "MIDEN_FAUCET_ENABLE_OTEL";
 pub const REQUESTS_QUEUE_SIZE: usize = 1000;
-const DEFAULT_API_KEYS_COUNT: &str = "1";
 const API_KEY_PREFIX: &str = "miden_faucet_";
 
 // TODO: we should probably parse this from the config file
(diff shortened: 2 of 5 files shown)


bin/faucet/src/main.rs
```diff
@@ -56,12 +59,43 @@ pub struct Cli {
...
+        timeout_ms: u64,
+
+        /// Path to the faucet account file.
+        #[arg(long = "account", default_value = DEFAULT_FAUCET_ACCOUNT_PATH, env = ENV_ACCOUNT_PATH)]
+        faucet_account_path: PathBuf,
+
+        /// Possible options on the amount of asset that should be dispersed on each faucet
+        /// request.
+        #[arg(long = "asset-amounts", env = ENV_ASSET_AMOUNTS, num_args = 1.., value_delimiter = ',', default_value = "100,500,1000")]
+        asset_amounts: Vec<u64>,
```
sergerad, line 86: @TomasArrachea @Mirko-von-Leipzig are you guys familiar with the frontend tests? Seeing this failure, unsure how it related to the configuration in the past.

https://github.com/0xMiden/miden-node/actions/runs/15669793324/job/44138930189?pr=976
```
    thread 'test::test_website' panicked at bin/faucet/src/main.rs:376:14:
    called `Result::unwrap()` on an `Err` value: Standard(WebDriver { error: NoSuchElement, message: "no such element: Unable to locate element: {\"method\":\"css selector\",\"selector\":\"#asset-amount option\"}\n  (Session info: chrome=137.0.7151.68)", stacktrace: "#0 0x5598e3736c4a <unknown>\n#1 0x5598e31dc6e0 <unknown>\n#2 0x5598e322e117 <unknown>\n#3 0x5598e322e311 <unknown>\n#4 0x5598e327cec4 <unknown>\n#5 0x5598e3253e5d <unknown>\n#6 0x5598e327a2cc <unknown>\n#7 0x5598e3253c03 <unknown>\n#8 0x5598e322047b <unknown>\n#9 0x5598e32210e1 <unknown>\n#10 0x5598e36fb3fb <unknown>\n#11 0x5598e36ff32f <unknown>\n#12 0x5598e36e2f39 <unknown>\n#13 0x5598e36ffec8 <unknown>\n#14 0x5598e36c768f <unknown>\n#15 0x5598e37242b8 <unknown>\n#16 0x5598e3724496 <unknown>\n#17 0x5598e3735b26 <unknown>\n#18 0x7f44c229caa4 <unknown>\n#19 0x7f44c2329c3c <unknown>\n", data: None })
```

Locally the FE looks as expected:
Screenshot 2025-06-16 at 1 33 24 PM

bin/faucet/.env
Mirko-von-Leipzig, line 1: Is this intended as an example file? Or what is its role?

bin/faucet/src/main.rs
```diff
@@ -453,7 +456,14 @@ mod test {
             rt.block_on(async {
                 run_faucet_command(Cli {
                     command: crate::Command::Start {
-                        config: config_path,
+                        endpoint: endpoint_clone,
+                        node_url: stub_node_url,
+                        timeout_ms: 5000,
+                        asset_amounts: vec![],
```
Mirko-von-Leipzig: I would guess that this is the cause of the test failure. The previous config defaulted to `[50, 100, 1000]`.

These values are what is displayed on the faucet frontend, and also what the faucet app accepts in the requests.

I think the test fails now because with no values the css/ui div widget doesn't appear at all in the frontend. And its trying to find it to run each of the values as part of the test.

bin/faucet/src/main.rs
```diff
@@ -213,35 +253,9 @@ async fn run_faucet_command(cli: Cli) -> anyhow::Result<()> {
...
-            let config_as_toml_string =
-                toml::to_string(&config).context("failed to serialize default config")?;
-
-            std::fs::write(&config_file_path, config_as_toml_string)
-                .context("error writing config to file")?;
-
-            println!("Config file successfully created at: {}", config_file_path.display());
+        Command::CreateApiKeys { key_count } => {
+            let keys = (0..key_count).map(|_| generate_api_key()).collect::<Vec<_>>().join(",");
+            println!("{keys}");
```
Mirko-von-Leipzig, line 258: Would it make sense to just have this generate a single key at a time? Though maybe its too early to tell what the usage of this would be like.

docs/src/operator/usage.md
```diff
@@ -64,22 +64,23 @@ miden-faucet create-faucet-account \
...
-  --faucet-account-path accounts/account.mac 
+export MIDEN_FAUCET_ENDPOINT=http://0.0.0.0:8080
+export MIDEN_FAUCET_NODE_URL=http://127.0.0.1:57291
+export MIDEN_FAUCET_TIMEOUT_MS=5000
+export MIDEN_FAUCET_ACCOUNT_PATH="accounts/faucet.mac"
+export MIDEN_FAUCET_ASSET_AMOUNTS=""
+export MIDEN_FAUCET_REMOTE_TX_PROVER_URL=""
+export MIDEN_FAUCET_POW_SECRET=""
+export MIDEN_FAUCET_API_KEYS=""
+export MIDEN_FAUCET_ENABLE_OTEL=false
```
Mirko-von-Leipzig: Should we not push people towards the CLI args? They at least come with documentation.

The CLI can be minimal, relying on the defaults; and just `miden-faucet start --help` for the rest.

docs/src/operator/usage.md
```diff
@@ -64,22 +64,23 @@ miden-faucet create-faucet-account \
...
 ```
 
 Run the faucet:
 
 ```sh
-miden-faucet --config miden-faucet.toml
+miden-faucet
```
Mirko-von-Leipzig: I think this has been wrong for a while?

```suggestion
miden-faucet start
```

bin/faucet/src/main.rs
Mirko-von-Leipzig: See https://github.com/0xMiden/miden-node/pull/976#discussion_r2149250326

bin/faucet/.env
sergerad: Its main purpose is to be packaged and used in the service file
```
        git show ${{ steps.git-sha.outputs.sha }}:bin/$crate/.env             > packaging/deb/$svc/lib/systemd/system/$svc.env
```
Which reminds me I need to add this line in this PR:
```
EnvironmentFile=/lib/systemd/system/miden-faucet.env
```

But its also a helpful example. Previously you had to look at the code to understand how faucet was configured. Looking at the `.env` here answers a lot of questions for anyone trying to figure out how to configure the faucet.

bin/faucet/src/main.rs
sergerad: Not sure. Main thing I would like is to be able to copy paste the output of this command into the corresponding flag/env var for the start command.

Mirko-von-Leipzig: Let's keep it as is and see what happens.

We'll have to rethink this once/if we support different difficulty PoW per API key.

bin/faucet/src/main.rs
```diff
@@ -28,21 +25,27 @@ use rand_chacha::ChaCha20Rng;
...
+const ENV_ACCOUNT_PATH: &str = "MIDEN_FAUCET_ACCOUNT_PATH";
+const ENV_ASSET_AMOUNTS: &str = "MIDEN_FAUCET_ASSET_AMOUNTS";
+const ENV_REMOTE_TX_PROVER_URL: &str = "MIDEN_FAUCET_REMOTE_TX_PROVER_URL";
+const ENV_POW_SECRET: &str = "MIDEN_FAUCET_POW_SECRET";
+const ENV_API_KEYS: &str = "MIDEN_FAUCET_API_KEYS";
 const ENV_ENABLE_OTEL: &str = "MIDEN_FAUCET_ENABLE_OTEL";
 pub const REQUESTS_QUEUE_SIZE: usize = 1000;
-const DEFAULT_API_KEYS_COUNT: &str = "1";
 const API_KEY_PREFIX: &str = "miden_faucet_";
+const DEFAULT_FAUCET_ACCOUNT_PATH: &str = "faucet.mac";
```
sergerad, line 46: @Mirko-von-Leipzig I found it confusing how the faucet account file relates to the bundled bootstrap command default. My understanding now is that the two defaults are named differently intentionally.

Ultimately I think that the faucet should not default to a value that would cause an error if the user hasn't created a subdir from the $PWD from which the command is called to start the faucet for example.

I also think that the bootstrap command should not have `--accounts-directory` flag but instead a flag that exists a filepath that is ready to write to e.g. `--account-output-filepath`, or `-o`. But that is maybe a separate issue.

Here I have just removed the `accounts/` subdir prefix to avoid the issue of a user not creating the subdir before calling faucet commands.

sergerad: In `usage.md` right now I think we have an error here:

```
miden-node bundled bootstrap \
  --data-directory /opt/miden-node \
  --accounts-directory /opt/miden-faucet
```

This will create `/opt/miden-faucet/account.mac` but the default value of miden-faucet is `./accounts/faucet.mac`. So I might consolidate `account.mac` and `faucet.mac` further into just `account.mac`.

sergerad: And remove the default value for `miden-facet create-faucet-account` to avoid overwriting `account.mac` generated by bootstrap.

So:
1. bundled bootstrap outputs `./account.mac`
2. faucet start defaults to `./account.mac`
3. faucet create account no default, user must specify output file

Mirko-von-Leipzig: > I also think that the bootstrap command should not have --accounts-directory flag but instead a flag that exists a filepath that is ready to write to e.g. --account-output-filepath, or -o. But that is maybe a separate issue.

We do want a directory because #836 would allow multiple accounts to be created.

> So:
>
>  1. bundled bootstrap outputs ./account.mac
>  2. faucet start defaults to ./account.mac
>  3. faucet create account no default, user must specify output file

This sounds good, my only suggestion would be to consider removing the default from (2) entirely. Let them always specify the filepath?

packaging/faucet/miden-faucet.service
```diff
@@ -7,8 +7,8 @@ WantedBy=multi-user.target
 
 [Service]
 Type=exec
-Environment="RUST_LOG=info"
-ExecStart=/usr/bin/miden-faucet start --config /etc/opt/miden-faucet/miden-faucet.toml
+EnvironmentFile=/lib/systemd/system/miden-faucet.env
```
Mirko-von-Leipzig, line 10: How does this interact ito overriding values?

We install the debian package, which also installs this env file. How does one now override these default values?

bin/faucet/.env
bobbinth, line 1: Should we add a comment somewhere in this file to tell people where they can read about the meaning of these settings?

docs/src/operator/usage.md
```diff
@@ -59,27 +59,17 @@ Create a faucet account for the faucet app to use - or skip this step if you alr
 ```sh
 mkdir accounts
-  --token-symbol MY_TOKEN \
+  --token-symbol BTC \
   --decimals 12 \
   --max-supply 5000
```
bobbinth, line 64: nit: if we are using BTC as the example, should we use other values consistent with it? - i.e., $6$ for decimal places and $2100000000000000$ for `max-supply`

bobbinth (commented):
Not a review, but I left a couple of comments inline.

packaging/faucet/miden-faucet.service
drahnr: You can run `systemd-edit` which will create an overrides file in i.e. `/etc/systemd/system/miden-faucet.service.d/override.conf` which is used to patch the original. Any key in the unit file is overriden by that `override.conf`.

Issue #877, opened 2025-05-26
Feed new block headers to NTB


igamigo:
One thing the [original design](https://github.com/0xMiden/miden-node/issues/695) of the gRPC endpoints  for network transactions did not fully take into account is how the NTB keeps track of the blockchain MMR updates. 
For the first approach, I introduced an endpoint that specifically returns the MMR peaks and latest block header (if there was an update since the last time NTB updated its MMR), since this is all it takes to be able to execute transactions with unauthenticated notes. 

A better approach, as suggested by Bobbin ([comment](https://github.com/0xMiden/miden-node/pull/841#discussion_r2106303780)), could be to have the block producer feed new blocks to the NTB. This, by itself, would not be enough because if the NTB crashes and has to restart at an arbitrary block number, then the MMR will be invalid. A complement to this (but also an improvement on its own) could be to add an `include_mmr_peaks` to the `GetBlockHeaderByNumber` endpoint:

```protobuf
message GetBlockHeaderByNumberRequest {
    optional uint32 block_num = 1;
    optional bool include_mmr_proof = 2;
    optional bool include_mmr_peaks = 3;
}
```

Issue #874, opened 2025-05-25
Track network notes from when they enter the mempool


igamigo:
Right now, created network notes are tracked by the builder from the moment [they get committed into blocks](https://github.com/0xMiden/miden-node/blob/next/crates/block-producer/src/block_builder/mod.rs#L304) (by calling `SubmitNetworkNotes`). However, the endpoint also has a transaction ID parameter so we should track notes from the moment they enter the mempool (as proposed originally https://github.com/0xMiden/miden-node/issues/695#issuecomment-2660821018). 

With this, the following race condition scenario should be solved as well:

> Though, thinking a bit more about it, I think we may have a race condition:
> 
> * Let's say a transaction creating a network note is accepted to the mempool. The NTB doesn't know about this transaction yet (we only communicate with NTB when transaction gets committed).
> * Then, before the transaction gets committed another transaction that consumes the network note (as an unauthenticated note) is accepted to the mempool. The `insert_inflight()` will get called, but we don't know about this network note - so, the info will be ignored.
> * Then, the first transaction gets committed but the second transaction still stays in the mempool (to be committed in a future block). At this point `queue_unconsumed_notes()` will be called, and we'll start tracking this network note.
> * Finally, the second transaction gets committed. The note will be consumed, but NTB won't be notified of this - and so, will keep thinking that the note is available for consumption.
(https://github.com/0xMiden/miden-node/pull/840#discussion_r2106120837)

Related to this is also #870

Issue #870, opened 2025-05-23
Generalize `SubmitNetworkNotes`


igamigo:
Currently, the `SubmitNetworkNotes` allows for submitting a transaction and a list of notes related to that transaction. The idea here was to allow communicating transactions that have just entered the mempool to the transaction builder. We could generalize it to take a list of transactions, in such a way that we could reuse it for the case when a blocks gets committed as well.

Reference: https://github.com/0xMiden/miden-node/pull/833/files#discussion_r2102256493

bobbinth:
Once we implement #874, this may not be needed.

PR #977, opened 2025-06-16
[wip] feat(ntx): track tx as they enter the mempool

Mirko-von-Leipzig:
This PR is still a WIP, but I'm opening it so we can discuss the overall idea and (re)design.

The goal of this work is to address #877, #874 and #870 however this brought up a few issues with the current ntx design. To fix these I've come with a (hopefully) simpler and more robust design which unfortunately means changing quite a few things.

> [!note]
> For brevity I'll be using ntx to refer to the network transaction builder component.

### Problem: mempool state ordering

The ntx needs to have a view into the mempool's current state in order to create valid network transactions. More specifically, it needs to be in-sync with all mempool state that pertains to network accounts.

Currently we maintain sync by only submitting committed state to the ntx, however with #874 we want to improve this fidelity by tracking all inflight transactions from when they enter the mempool. This creates a few race conditions since currently the ntx messages are submitted from concurrent processes. Some examples:

- Tx is added and is reverted
- Tx is added and is committed
- Tx (a) creates a network note, tx (b) consumes it

For each of the above its possible for the messages to arrive out-of-order at the ntx. This would break the tracking in the ntx. The primary issue is that mempool events need to be ordered and therefore cannot happen concurrently, but need to be sync'd.

To this end, the new design emits a `MempoolEvent` from the mempool itself. Since the mempool is a central sync point already this ensures the messages are ordered.

### Problem: (de)sync

We want ordered and sequential events. However with the current gRPC server/client setup its impossible to know that all events were received. As an example, the ntx gRPC server cannot know, and also cannot be informed, that it missed an event.

We essentially want a stream. When the stream fails it ends on both sides so the failure cannot go unnoticed. I've therefore changed to a gRPC stream of events.

### Problem: resetting components

We decided to punt the problem of resetting and resyncing the mempool and ntx components but it does need addressing. I think my new design allows us to implement this in a sensible fashion.

What I propose is that the mempool event stream:

1. Can only be started if both ends agree on the same committed state i.e. both are on the same block number.
2. Event stream must contain all non-committed state.

The ntx startup procedure looks something like:

1. Fetch latest state from store.
2. Open mempool stream - this is rejected if the block number does not match the mempool's. Repeat from (1) until we match. There are more complex ways but this seems simple.
3. Continue until stream breaks, reset to (1).

The mempool side has some additional complexity because it needs to be able to send previous, uncommitted state when a new stream starts. But I think that should be fairly straight-forward (if annoying) to implement.

### Current impl

The current implementation is missing quite a few things still. Some things will also need adapting as additional features are added (e.g. currently I use a broadcast channel on the mempool side).

My suggestion, is in this PR to

- [x] Mempool event stream
- [ ] ntx receiver & state sync

and then in follow-up PRs

- [ ] Simpify ntx state management
- [ ] Store ntx state in a db
- [ ] Enable component resets
```

### CODE REVIEW

File: crates/ntx-builder/src/builder/mod.rs [hunk 1]
```diff
@@ -363,7 +287,7 @@ impl NetworkTransactionBuilder {
     /// Executes the transaction with the account described by the request.
     #[instrument(target = COMPONENT, name = "ntx_builder.execute_transaction", skip_all, err)]
     async fn execute_transaction(
-        tx_executor: &TransactionExecutor<'_, '_>,
+        state: &State,
         tx_request: NetworkTransactionRequest,
     ) -> Result<ExecutedTransaction, NtxBuilderError> {
         let input_notes = InputNotes::new(
```
bobbinth: nit (and not from this PR): it feels like this should be a method on the `NetworkTransactionRequest` - something like:
```rust
impl NetworkTransactionRequest {
    pub fn get_input_notes(&self) -> InputNotes {
        ...
    }
}
```

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -312,11 +232,15 @@ impl NetworkTransactionBuilder {
...
 
-        for note in input_notes.iter() {
-            data_store.insert_note_script_mast(note.note().script());
-        }
+        // // TODO: ensure that the incoming data store in the executor already has these inserted.
+        // // This could be done in State whenever a new note arrives, or we could
+        // // build a custom type with (State, TxMastForest) which impls DataStore instead.
+        // // for note in input_notes.iter() {
+        // //     data_store.insert_note_script_mast(note.note().script());
+        // // }
```
bobbinth: It is not clear to me if this is actually needed. The note script is provided directly the executor via the `execute_transaction()` method - so, adding it to the data store shouldn't be needed. cc @igamigo.

proto/proto/block_producer.proto
```diff
@@ -12,5 +18,57 @@ service Api {
...
+    transaction.TransactionId id = 1;
+    // Nullifiers consumed by the transaction.
+    repeated digest.Digest nullifiers = 2;
+    // Network notes created by the transaction.
+    repeated note.NetworkNote network_notes = 3;
+    // Changes to a network account, if any.
+    //
+    // The account delta is encoded using [winter_utils::Serializable] implementation
+    // for [miden_objects::account::delta::AccountDelta].
+    optional bytes network_account_delta = 4;
```
bobbinth: Is this used to signal creation of new network accounts as well? If so, I'd mention this in the comment.

crates/proto/src/domain/mempool.rs
```diff
@@ -0,0 +1,42 @@
...
+use super::note::NetworkNote;
+
+#[derive(Debug, Clone)]
+pub enum MempoolEvent {
+    TransactionAdded {
+        id: TransactionId,
+        nullifiers: Vec<Nullifier>,
+        network_notes: Vec<NetworkNote>,
+        // TODO: decide whether to create a new type to exclude private variant which isn't possible.
+        account_delta: Option<AccountUpdateDetails>,
```
bobbinth: If a transaction against a non-network (e.g., private or public) account is submitted to the mempool, would this be set to `None`?

crates/ntx-builder/src/builder/block_producer.rs
```diff
@@ -45,4 +50,19 @@ impl BlockProducerClient {
 
         Ok(())
     }
+
+    pub async fn subscribe_to_mempool(
```
bobbinth, line 54: Should we instrument this method?

bin/node/src/commands/bundled.rs
```diff
@@ -189,11 +189,6 @@ impl BundledCommand {
             .context("Failed to bind to block-producer gRPC endpoint")?
             .local_addr()
             .context("Failed to retrieve the block-producer's gRPC address")?;
-        let ntx_builder_address = TcpListener::bind("127.0.0.1:0")
-            .await
-            .context("Failed to bind to network transaction builder gRPC endpoint")?
-            .local_addr()
-            .context("Failed to retrieve the network transaction builder's gRPC address")?;
```
bobbinth, line 196: This is removed temporarily, right?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -165,6 +167,35 @@ pub struct Mempool {
 
     block_budget: BlockBudget,
     batch_budget: BatchBudget,
+
+    event_subscriber: Option<mpsc::Sender<MempoolEvent>>,
+}
+
+// We have to implement this manually since the event's channel does not implement PartialEq.
+impl PartialEq for Mempool {
```
bobbinth, line 176: Is this mostly for testing purposes? If not, under what condition do we want to test quality between two mempools?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -165,6 +167,35 @@ pub struct Mempool {
 
     block_budget: BlockBudget,
     batch_budget: BatchBudget,
+
+    event_subscriber: Option<mpsc::Sender<MempoolEvent>>,
```
bobbinth: nit: let's add brief doc comments here.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -221,12 +253,39 @@ impl Mempool {
...
+        let nullifiers = transaction.nullifiers().collect();
+        let network_notes = transaction
+            .output_notes()
+            .filter_map(|note| match note {
+                OutputNote::Full(inner)
+                    if inner.metadata().tag().execution_mode() == NoteExecutionMode::Network =>
+                {
+                    NetworkNote::try_from(inner.clone()).ok()
+                },
+                _ => None,
```
bobbinth: Should this be handled by `impl TryFrom<OutputNote> for NetworkNote`?

crates/ntx-builder/src/state.rs
```diff
@@ -0,0 +1,196 @@
...
+struct InflightTx {
+    account_delta: Option<AccountId>,
+    nullifiers: Vec<Nullifier>,
+    notes: Vec<Nullifier>,
+}
+
+struct AccountState {
+    state: Account,
+    deltas: VecDeque<AccountDelta>,
+}
```
bobbinth: Would be great to add some comments here. My understanding is that `state` is the state of the account before any of the `deltas` have been applied to it, right? Also, we are not guaranteed to have at least one delta (i.e., `deltas` could be empty).

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -112,88 +100,57 @@ impl NetworkTransactionBuilder {
...
-    fn spawn_ticker(&self, api_state: SharedPendingNotes) -> JoinHandle<anyhow::Result<()>> {
-        let store_url = self.store_url.clone();
-        let block_addr = self.block_producer_address;
-        let prover_addr = self.tx_prover_url.clone();
-        let ticker_interval = self.ticker_interval;
+        let block_prod = BlockProducerClient::new(self.block_producer_address);
+        let tx_prover = NtbTransactionProver::from(self.tx_prover_url.clone());
+        let mut interval = tokio::time::interval(self.ticker_interval);
 
+        // This extra runtime is required to work-around miden-base objects not being Send/Sync due to dyn Trait object usage.
```
bobbinth: Which objects specifically are causing issues here?

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: That would be great if I can ignore this :)

proto/proto/block_producer.proto
Mirko-von-Leipzig: Yes that's correct; I wrote the docs before I realized it needed to allow account creation as well.

crates/proto/src/domain/mempool.rs
Mirko-von-Leipzig: Correct

bin/node/src/commands/bundled.rs
Mirko-von-Leipzig: No its permanent. The ntx gRPC server allowed the block-producer to submit mempool events to the ntx. 

This relationship is now inverted and replaced entirely by the mempool event stream. This gets created by a request to the block-producer i.e. the ntx server methods moved entirely to a single stream method added to the block-producer.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: It is for testing purposes yeah. I could `#[cfg(test)]` scope it if you like; though I don't see the harm in having it more exposed.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: The downside is that it would move the clone. As in, we would clone every output note whereas I wanted to restrict the clone to just the notes we know to be network notes.

I went back and forth on this a few times; I'm unsure whether its worth it or not. But I figured there would be _many_ notes and very few network notes..

We could `TryFrom` on a reference but I really dislike that as a general concept. The std library actually has a `clone_into` method which I hoped would be this kind of thing, but unfortunately its not at all.

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: Its the `dyn Trait` objects. So `TransactionExecutor` and `TransactionAuthenticator` and `SourceMapper`.

We're chatting about this tonight; but as a short synopsis:

Using `struct T(&dyn U)` instead of `T<Y: U>(Y)` means that there is only ever a single implementation of `T` which covers all variations of `U`. Whereas the latter creates an implementation for each specific `Y: U` that is used.

What this means is that with `dyn U`, the compiler has to be restrictive and `T` is bound entirely by `U` definition. In our case, `U: DataStore` and the compiler cannot automatically add `U: Sync + Send`. We could fix this in the short term by adding the bound requirements ourselves, however this would now disallow implementations of `DataStore` that aren't `Send` or `Sync`.

When actual generics are used, the compiler creates a separate implementation for each `Y` and therefore `T` can inherit the properties of `Y`. So if `Y: Send` then `T: Send` and so on.

Or put differently, generics are superior to `dyn` trait objects. The downside is that they increase compile times and add code noise, though the latter can be reduced by setting defaults to the generics.

Mirko-von-Leipzig: There is also `ExecutedTransaction` which contains a `OnceCell` (or one of its variants), which isn't `Sync`.

Mirko-von-Leipzig: This becomes a problem because `tokio::spawn` requires the future to be `Send`, but since it contains types that aren't `Send` the future cannot be spawned by tokio directly.

crates/ntx-builder/src/builder/mod.rs
igamigo: Yes, this is not needed anymore after https://github.com/0xMiden/miden-base/pull/1438, which was recently merged

crates/ntx-builder/src/builder/mod.rs
igamigo: > We could fix this in the short term by adding the bound requirements ourselves, however this would now disallow implementations of DataStore that aren't Send or Sync.

If we were to add the bound requirements in the executor instead of the trait definition, it only disallows executors that use `DataStore` impls that are not `Send` nor `Sync`. But I think this is the only place where `DataStore` is really used so I guess these would be equivalent. I don't think it would be a problem to add these bounds in the executor anyway though.

crates/block-producer/src/txqueue/tests/mod.rs
igamigo, line 1: Was this deleted because they are not compatible for now?

crates/ntx-builder/src/state.rs
```diff
@@ -0,0 +1,196 @@
...
+                id,
+                nullifiers,
+                network_notes,
+                account_delta,
+            } => {
+                // TODO: AccountUpdateDetails is the wrong type to use.
+                //
+                // TODO: Update AccountState
+                let account_id = account_delta.map(|delta| match delta {
+                    AccountUpdateDetails::Private => unreachable!("Should never occur"),
```
igamigo: Hmm, would this never occur? I think if I have a private account and output a nullifier from a network note, we still want to make sure the ntx gets notified of such transaction, right? Or maybe I missed something

igamigo: Oh nvm, this is an option and we'd get `None` for a non-network account

crates/ntx-builder/src/state.rs
```diff
@@ -0,0 +1,196 @@
...
+
+impl AccountState {
+    fn new(state: Account) -> Self {
+        Self { state, deltas: Default::default() }
+    }
+
+    fn commit_one(&mut self) {
+        let Some(to_commit) = self.deltas.pop_front() else {
+            panic!("account state should have a delta to commit");
+        };
```
igamigo: nit: Could use `expect()` here

crates/ntx-builder/src/state.rs
```diff
@@ -0,0 +1,196 @@
...
+                        nullifiers: nullifiers.clone(),
+                        notes: network_notes.iter().map(NetworkNote::nullifier).collect(),
+                    },
+                );
+
+                self.notes
+                    .extend(network_notes.into_iter().map(|note| (note.nullifier(), note)));
+                self.nullifiers.extend(nullifiers);
+            },
+            MempoolEvent::BlockCommitted { header, txs } => {
```
igamigo, line 59: You were probably thinking of this already but refactoring these code blocks out to single functions would make it simpler to follow

crates/ntx-builder/src/state.rs
```diff
@@ -0,0 +1,196 @@
...
+                account_delta,
+            } => {
+                // TODO: AccountUpdateDetails is the wrong type to use.
+                //
+                // TODO: Update AccountState
+                let account_id = account_delta.map(|delta| match delta {
+                    AccountUpdateDetails::Private => unreachable!("Should never occur"),
+                    AccountUpdateDetails::New(account) => account.id(),
+                    AccountUpdateDetails::Delta(_account_delta) => {
+                        todo!("This is the wrong type to use here, we need an account ID")
```
igamigo: Is this outdated or do you still mean to change this to an `AccountId`? I guess not sure if the idea is to optimistically apply deltas after transactions are executed on the ntx builder or wait for them to be finalized in blocks (or doing something in the middle where you store various states for a single account and can then roll back appropriately and apply a new delta when finalized)

crates/block-producer/src/txqueue/tests/mod.rs
Mirko-von-Leipzig: No its actually a dangling file. We no longer use transaction queue, or any of these objects. They were replaced with the mempool.

crates/ntx-builder/src/state.rs
Mirko-von-Leipzig: I couldn't figure out how to get an account ID from the delta. It might not contain it? In which case I need to change this data structure to include it.

crates/block-producer/src/mempool/subscription.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Chronologically ordered transaction events which have not yet been committed or reverted.
+    ///
+    /// These need to be sent for new subscriptions since otherwise these events would be skipped.
+    ///
+    /// These events are stored as weak pointers to the actual data to allow for efficient removal
+    /// without moving the entire queue.
+    ///
+    /// The queue itself is trimmed by removing non-existent pointers whenever from the front and back
+    /// whenever txs are committed or reverted. This does mean slow or long-term txs let the queue build up in size.
+    ordered_txs: VecDeque<Weak<MempoolEvent>>,
```
Mirko-von-Leipzig: This implementation kinda sucks but I'm not sure about how to do this properly.

What we essentially want is an ordered collection of txs from which we can delete in arbitrary places efficiently.

The simplest implementation would be `VecDequeue<MempoolEvent>` but the deletion in the middle would cause all values to be moved/cloned. Open to suggestions..

crates/block-producer/src/mempool/mod.rs
bobbinth: Yeah, probably fine to keep as is.

crates/block-producer/src/mempool/mod.rs
bobbinth: Makes sense! I actually don't mind doing `TryFrom<&OutputNote> for NetworkNote` - but also, it's fine not to do it.

Maybe a better approach here would be to introduce something like `Note::is_network_note()` in `miden-base`.

crates/ntx-builder/src/builder/mod.rs
bobbinth: > If we were to add the bound requirements in the executor instead of the trait definition, it only disallows executors that use `DataStore` impls that are not `Send` nor `Sync`. But I think this is the only place where `DataStore` is really used so I guess these would be equivalent. I don't think it would be a problem to add these bounds in the executor anyway though.

To make the `TransactionExecutor` send/sync would we need to put trait bounds only on `DataStore` and `TransactionAuthenticator` or on `SourceManager` as well?

Also, are there any real downsides to doing requiring them to always be send/sync? (i'm guessing there may be some performance overhead?)

> There is also `ExecutedTransaction` which contains a `OnceCell` (or one of its variants), which isn't `Sync`.

If this is causing problems, it should be pretty easy to remove.

crates/ntx-builder/src/state.rs
bobbinth: `AccountDelta` doesn't currently contain account ID, but @PhilippGackstatter is planning to change this shortly.

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: > To make the TransactionExecutor send/sync would we need to put trait bounds only on DataStore and TransactionAuthenticator or on SourceManager as well?

We need to either add the bounds if we're keeping with `&dyn Trait` to all three. Though `SourceManager` is still `Arc` as well I think.

> Also, are there any real downsides to doing requiring them to always be send/sync? (i'm guessing there may be some performance overhead?)

The main downside is that it forces callers to implement them as `Send + Sync`. Technically we're over-restricting things now because we want to use `tokio` multi-threaded spawn. A user running single-threaded tokio, or some other async runtime might not have the same requirement. For the node's purposes there is no downside.

The usual way is to make them proper generics and let them inherit these properties instead. Then the caller can make the decision.

Mirko-von-Leipzig: > > There is also ExecutedTransaction which contains a OnceCell (or one of its variants), which isn't Sync.
>
> If this is causing problems, it should be pretty easy to remove.

I would vote for removal. Essentially it stops us from using references in async i.e.

```rust
async fn do_something(tx: &ExecutedTransaction)
```

cannot be called from tokio since `&ExecutedTransaction` isn't safe to share across threads.

crates/ntx-builder/src/state.rs
Mirko-von-Leipzig: I'll wait for that change to land then.

crates/ntx-builder/src/state.rs
Mirko-von-Leipzig: Thanks. Its funny, somehow I completely forgot about `expect`.. kept thinking that I never use `panic!` like this :thinking:

crates/ntx-builder/src/builder/mod.rs
igamigo: > The main downside is that it forces callers to implement them as Send + Sync. Technically we're over-restricting things now because we want to use tokio multi-threaded spawn. A user running single-threaded tokio, or some other async runtime might not have the same requirement. For the node's purposes there is no downside.

Yeah, I don't think there is any other downside other than this.

> The usual way is to make them proper generics and let them inherit these properties instead. Then the caller can make the decision.

I think this could be nicer in terms of ergonomics, but (I could be misremembering here) one reason we chose trait objects over generics was to keep Client from becoming something like `Client<E, A, S, R, N>`. Having half-a-dozen generic parameters can get noisy, whereas a single `Box<dyn Foo + Send + Sync>` keeps the type signature short. I don’t mind the longer generic list personally, but others preferred the cleaner look. I realize now that the executor is not long-lived it could eventually be better but unless we make the `DataStore` a parameter of `new_transaction()` instead of holding a `DataStore` in `Client`, this won't change I think. Maybe all these opinions are calling for a more holistic refactor.

crates/ntx-builder/src/builder/block_producer.rs
Mirko-von-Leipzig: I've decided not to for now. I'll do a pass over all instrumentation once this is completed.

crates/ntx-builder/src/state.rs
Mirko-von-Leipzig: I've moved this into a separate module and expanded on the comments. I noticed there were some issues with this impl.

crates/ntx-builder/src/state.rs
PhilippGackstatter: Nice coincidence! This is the PR: https://github.com/0xMiden/miden-base/pull/1493. This builds on top of 6 other PRs though, so might take a bit to get into `next` :upside_down_face:.

crates/block-producer/src/mempool/subscription.rs
PhilippGackstatter: I'm probably missing context, but would a `BTreeMap<u64, MempoolEvent>` work in which you insert new elements by taking the highest current value (e.g. `BTreeMap::last_key_value`) + 1? Or is a map too slow?

Mirko-von-Leipzig: Heh coincidentally I literally implemented your suggest last night :grin: 

I think it will work; we already have other graph structures in the mempool so one more shouldn't tip the scales.

I actually didn't know about `last_key_value` so that's saved me an extra tracking counter! Thanks.

PR #982, opened 2025-06-16
ci: Cache Docker only on non-forks

sergerad:
## Context

PRs created by forks of this repo cannot use secrets until they are approved. This means that the docker build workflow will fail for those PRs because it requires access to the S3 backend for caching.

Verified the non-fork version did use cache here https://github.com/0xMiden/miden-node/pull/979.

## Changes

- Only use AWS S3 and Docker caching for the PR Docker build workflow when the PR is created from the main repository, not a fork


bobbinth (approved):

PR #984, opened 2025-06-17
ci: Env var configuration for miden-node

### LINKED ISSUE #687: Separate telemetry by component
Open-telemetry has `service.name` and `service.namespace`. Currently these are `miden-node` and empty respectively.

The [specification](https://opentelemetry.io/docs/specs/semconv/attributes-registry/service/#service-attributes) indicates that `service.name` should be identical across horizontal scaling. Or in other words, we should probably have separate `service.name` for each component in the node i.e. `store`, `rpc`, `block-producer`. And then the namespace could simply be `miden` or `miden-node`.

Its unclear how easy this would be to implement, but it will require separately instrumenting each component somehow.

### LINKED ISSUE #971: Update node binary crate to rely on Clap env vars via .env file in Debian package
## Context

We are standardizing configuration of binary crates towards env vars based on Clap and packaging .env into the Debian packages for this purpose.

# Work to do

- Check if there is any configuration that the node crate relies on that cannot be specified via env vars
    - Update this to use env vars if any exist
- Update `bin/node/.env` to specify all the env vars and appropriate defaults
- Ensure the `miden-node.service` file configures the node service to use the env file as packaged in the Debian package

sergerad:
## Context

Closes #971.
Closes #687.

We just need to close the loop on configuring miden-node the same way we do with the other Debian packages moving forward.

## Changes

- Populate `bin/node/.env` with all env vars. Add values relevant to bundled command.
- Point miden-node service file to env var file in Debian package.
- Add `OTEL_SERVICE_NAME` to service files.


drahnr (approved):

packaging/node/miden-node.service
```diff
@@ -7,12 +7,10 @@ WantedBy=multi-user.target
 
 [Service]
 Type=exec
+Environment="OTEL_SERVICE_NAME=miden-node"
+EnvironmentFile=/lib/systemd/system/miden-node.env
```
Mirko-von-Leipzig, line 11: How do we override these when we deploy?

I'm also still partial to listing them all in this service file, or in an `override.conf` because then they show up for `systemctl cat <service>` but admittedly one can go find the env file here separately.

sergerad: > How do we override these when we deploy?

That is something our automation should handle. But doesn't currently. Need a task in infra repo for that.

> I'm also still partial to listing them all in this service file, or in an `override.conf` because then they show up for `systemctl cat <service>` but admittedly one can go find the env file here separately.

The problem with having them in the service file too is that we need to automate that, which would be ugly. Otherwise we would be duplicating env vars and opening ourselves up for misconfiguration b/c someone forgets to update the service file. The automation would have to convert from .env to service file (because we want .env for users and Docker). It would also have to check in CI that the .env and service files match.

Its just extra automation work we could do without. Would rather not do that just to avoid systemctl cat + cat env file. Its not required for fully automating our CD of these services.

Issue #930, opened 2025-06-06
Faucet background is large


Mirko-von-Leipzig:
The faucet website's [background image](https://github.com/0xMiden/miden-node/tree/next/bin/faucet/src/server/resources) is a whopping 1.2MB. This is quite unnecessary :)

We should be able to drastically lower this by replacing the image with a compressed version without much loss in quality.

PR #985, opened 2025-06-17
feat(faucet): compress background image

TomasArrachea:
Closes https://github.com/0xMiden/miden-node/issues/930

The `background.png` image was 1.14 MB. This PR replaces it with a minimal image of the repeating pattern that is 27 KB and tiles in the background.

Result:

image


TomasArrachea:
> An alternative to compressing the image itself could be to edit it or get a version that is squared (ie not rotated, not sure I used the right term) and can be tiled and then slightly rotate it with CSS (I'm guessing something like this would be doable)

Implemented this in [8e81313](https://github.com/0xMiden/miden-node/pull/985/commits/8e8131338bd844dd5b0213e16766e9403c55d05a). I think it's a good solution, image is only 27 KB now and the result is practically the same.

PR #986, opened 2025-06-17
feat: poll explorer link

TomasArrachea:
closes https://github.com/0xMiden/miden-faucet/issues/5

This PR polls the explorer link in the faucet frontend until the transaction is visible

TODO:
- currently fails because of CORS settings of the explorer


igamigo:
Should we close this now that the work has moved to the faucet repo?

PR #987, opened 2025-06-17
chore: use logs and tracing config from node

### LINKED ISSUE #917: Update proving service to use same Otel / tracing as node

SantiagoPittella:
closes #917

### DIFF
--- a/bin/proving-service/src/api/prover.rs
+++ b/bin/proving-service/src/api/prover.rs
@@ -1,4 +1,3 @@
-use std::time::Instant;
 
 use miden_block_prover::LocalBlockProver;
 use miden_objects::{
@@ -15,7 +14,6 @@ use crate::{
     COMPONENT,
     commands::worker::ProverType,
     generated::{ProofType, ProvingRequest, ProvingResponse, api_server::Api as ProverApi},
-    utils::MIDEN_PROVING_SERVICE,
 };
 
 /// The prover for the proving service.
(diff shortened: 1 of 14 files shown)


bin/proving-service/src/api/prover.rs
```diff
@@ -77,8 +84,16 @@ impl ProverRpcApi {
...
         let transaction_id = proof.id();
+        let duration = start_time.elapsed();
         tracing::Span::current().record("id", tracing::field::display(&transaction_id));
 
+        info!(target: COMPONENT,
+            proof_type = "transaction",
+            id = %transaction_id,
+            duration_ms = %duration.as_millis(),
+            "Proof generated successfully"
+        );
```
sergerad: We could just rely on the span close event for the duration instead of doing it manually. Just need to make sure this fn has `#[instrument]` annotation.

bin/proving-service/src/api/prover.rs
```diff
@@ -107,8 +123,16 @@ impl ProverRpcApi {
...
         let batch_id = proven_batch.id();
+        let duration = start_time.elapsed();
         tracing::Span::current().record("id", tracing::field::display(&batch_id));
 
+        info!(target: COMPONENT,
+            proof_type = "batch",
+            id = %batch_id,
+            duration_ms = %duration.as_millis(),
+            "Proof generated successfully"
+        );
```
sergerad: Same as above.

bin/proving-service/src/api/prover.rs
```diff
@@ -137,9 +162,17 @@ impl ProverRpcApi {
...
+        let duration = start_time.elapsed();
 
         tracing::Span::current().record("id", tracing::field::display(&block_id));
 
+        info!(target: COMPONENT,
+            proof_type = "block",
+            id = %block_id,
+            duration_ms = %duration.as_millis(),
+            "Proof generated successfully"
+        );
```
sergerad: As above.

bin/proving-service/src/api/prover.rs
SantiagoPittella: This `fn` does in fact have the `#[instrument]` annotation. I will delete this.

PR #988, opened 2025-06-18
Sergerad faucet config env

### LINKED ISSUE #958: Add faucet configuration TOML file to Debian package
The following data is not provided for at all in our Debian package for Faucet:
```rust
#[derive(Clone, Debug, Serialize, Deserialize)]
#[serde(deny_unknown_fields)]
pub struct FaucetConfig {
    pub endpoint: Url,
    /// Node RPC gRPC endpoint in the format `http://<host>[:<port>]`
    pub node_url: Url,
    /// Timeout for RPC requests in milliseconds
    pub timeout_ms: u64,
    /// Possible options on the amount of asset that should be dispersed on each faucet request
    pub asset_amount_options: AssetOptions,
    /// Path to the faucet account file
    /// Optional: Endpoint of the remote transaction prover in the format
    /// `<protocol>://<host>[:<port>]`
    pub remote_tx_prover_url: Option<Url>,
    pub pow_salt: String,
    /// List of API keys
    pub api_keys: Vec<String>,
}
```

Ultimately we should move to provide all of this via env vars through `bin/faucet/.env`. We should probably do that now.

sergerad:
## Context

Closes #958.

The Faucet currently relies on a configuration TOML file which is intended to be generated by first calling `miden-faucet init ...`.

We are standardizing all network-related binary crates to be completely configurable through environment variables.

## Changes
- The `miden-faucet init` command has been removed because generation of the TOML is no longer required.
- A new command `miden-faucet create-api-keys <num>` has been added to replace the functionality we still need from `miden-faucet init`.
- The code and documentation relating to the TOML configuration failed been removed / replaced.
- The Dockerfile has been updated to no longer bake configuration into the image.
- TLS support has been added to the RPC client by default
   - This was done to aid local testing against remote RPC on HTTPS


packaging/faucet/miden-faucet.service
```diff
@@ -7,8 +7,8 @@ WantedBy=multi-user.target
 
 [Service]
 Type=exec
-Environment="RUST_LOG=info"
-ExecStart=/usr/bin/miden-faucet start --config /etc/opt/miden-faucet/miden-faucet.toml
+EnvironmentFile=/lib/systemd/system/miden-faucet.env
```
sergerad, line 10: @Mirko-von-Leipzig comment https://github.com/0xMiden/miden-node/pull/976#discussion_r2151502664

---

Should be as simple as editing `/lib/systemd/system/miden-faucet.env` and restarting service. No need for override file. Would expect to use an override file to change things specific to systemd rather than the faucet app e.g. these settings

```
RestartSec=5
Restart=always
```

Issue #983, opened 2025-06-16
Update all hard-coded references to testnet to be configurable instead


sergerad:
## Context

We have some TODOs involve hard-coded settings for testnet, E.G.:

```rust
// TODO: Make these configurable.
const NETWORK_ID: NetworkId = NetworkId::Testnet;
const EXPLORER_URL: &str = "https://testnet.midenscan.com";
```

## Requirements
- All hard-coded references to testnet are replaced with configurable values that work for both local, devnet, and testnet.
- Update any/all components (node, faucet, proving-service).

TomasArrachea:
Part of this was done in https://github.com/0xMiden/miden-node/pull/1016. 

> Update any/all components (node, faucet, proving-service).

I think there is no other testnet reference hardcoded in the components, so we can probably close this.

bobbinth:
@sergerad, @Mirko-von-Leipzig - could you double check?

PR #989, opened 2025-06-18
refactor: Replace Faucet configuration file and Init command

### LINKED ISSUE #958: Add faucet configuration TOML file to Debian package

sergerad:
## Context

Closes #958.

The Faucet currently relies on a configuration TOML file which is intended to be generated by first calling `miden-faucet init ...`.

We are standardizing all network-related binary crates to be completely configurable through environment variables.

## Changes
- The `miden-faucet init` command has been removed because generation of the TOML is no longer required.
- A new command `miden-faucet create-api-keys <num>` has been added to replace the functionality we still need from `miden-faucet init`.
- The code and documentation relating to the TOML configuration failed been removed / replaced.
- The Dockerfile has been updated to no longer bake configuration into the image.
- TLS support has been added to the RPC client by default
   - This was done to aid local testing against remote RPC on HTTPS
```

### CODE REVIEW

File: packaging/faucet/miden-faucet.service [hunk 1]
```diff
@@ -7,8 +7,8 @@ WantedBy=multi-user.target
 
 [Service]
 Type=exec
-Environment="RUST_LOG=info"
-ExecStart=/usr/bin/miden-faucet start --config /etc/opt/miden-faucet/miden-faucet.toml
+EnvironmentFile=/lib/systemd/system/miden-faucet.env
```
sergerad, line 10: @Mirko-von-Leipzig regarding override -

Should be as simple as editing `/lib/systemd/system/miden-faucet.env` and restarting service. No need for override file. Would expect to use an override file to change things specific to systemd rather than the faucet app e.g. these settings

```
RestartSec=5
Restart=always
```

.gitignore
```diff
@@ -24,6 +24,7 @@ target/
 # Node files
 /accounts
 /data
+*.mac
```
drahnr, line 28: nit: I'd add a statement re keyfiles

bin/faucet/src/main.rs
```diff
@@ -28,21 +25,26 @@ use rand_chacha::ChaCha20Rng;
...
-const FAUCET_CONFIG_FILE_PATH: &str = "miden-faucet.toml";
+const ENV_ENDPOINT: &str = "MIDEN_FAUCET_ENDPOINT";
+const ENV_NODE_URL: &str = "MIDEN_FAUCET_NODE_URL";
+const ENV_TIMEOUT: &str = "MIDEN_FAUCET_TIMEOUT_MS";
+const ENV_ACCOUNT_PATH: &str = "MIDEN_FAUCET_ACCOUNT_PATH";
+const ENV_ASSET_AMOUNTS: &str = "MIDEN_FAUCET_ASSET_AMOUNTS";
+const ENV_REMOTE_TX_PROVER_URL: &str = "MIDEN_FAUCET_REMOTE_TX_PROVER_URL";
+const ENV_POW_SECRET: &str = "MIDEN_FAUCET_POW_SECRET";
+const ENV_API_KEYS: &str = "MIDEN_FAUCET_API_KEYS";
 const ENV_ENABLE_OTEL: &str = "MIDEN_FAUCET_ENABLE_OTEL";
```
drahnr, line 49: nit: I'd add a space around `ENV_*` vars block

bin/faucet/src/main.rs
```diff
@@ -28,21 +25,26 @@ use rand_chacha::ChaCha20Rng;
...
 pub const REQUESTS_QUEUE_SIZE: usize = 1000;
-const DEFAULT_API_KEYS_COUNT: &str = "1";
 const API_KEY_PREFIX: &str = "miden_faucet_";
 
-// TODO: we should probably parse this from the config file
+// TODO: Make these configurable.
```
drahnr: Do you intend to make these configurable as part of this PR?

bin/faucet/src/main.rs
```diff
@@ -56,50 +58,78 @@ pub struct Cli {
...
+        /// Timeout for RPC requests in milliseconds.
+        #[arg(long = "timeout", value_name = "MILLISECONDS", default_value_t = 5000, env = ENV_TIMEOUT)]
+        timeout_ms: u64,
+
+        /// Path to the faucet account file.
+        #[arg(long = "account", value_name = "FILE", env = ENV_ACCOUNT_PATH)]
+        faucet_account_path: PathBuf,
+
+        /// Comma-separated list of amounts of asset that should be dispersed on each request.
+        #[arg(long = "asset-amounts", value_name = "U64", env = ENV_ASSET_AMOUNTS, num_args = 1.., value_delimiter = ',', default_value = "100,500,1000")]
```
drahnr, line 83: nit: I tend to prefer the typed variant: `default_value_t = vec![100,500,1000]` _unless_ it's hidious as hell, which it might be in this case (ref https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes )

bin/faucet/src/main.rs
```diff
@@ -56,50 +58,78 @@ pub struct Cli {
...
+        /// Endpoint of the faucet in the format `<ip>:<port>`.
+        #[arg(long = "endpoint", value_name = "URL", env = ENV_ENDPOINT)]
+        endpoint: Url,
+
+        /// Node RPC gRPC endpoint in the format `http://<host>[:<port>]`.
+        #[arg(long = "node-url", value_name = "URL", env = ENV_NODE_URL)]
+        node_url: Url,
+
+        /// Timeout for RPC requests in milliseconds.
+        #[arg(long = "timeout", value_name = "MILLISECONDS", default_value_t = 5000, env = ENV_TIMEOUT)]
```
drahnr, line 75: nit: Can we make this `humantime`, such that the user can specify a unit? The default value suggests this might be useful. ( ref https://crates.io/crates/humantime )

bin/faucet/src/main.rs
```diff
@@ -56,50 +58,78 @@ pub struct Cli {
...
+        timeout_ms: u64,
+
+        /// Path to the faucet account file.
+        #[arg(long = "account", value_name = "FILE", env = ENV_ACCOUNT_PATH)]
```
drahnr, line 79:
```suggestion
        #[arg(long = "account", value_name = "PATH", env = ENV_ACCOUNT_PATH)]
```

bin/faucet/src/main.rs
```diff
@@ -56,50 +58,78 @@ pub struct Cli {
...
     CreateFaucetAccount {
-        #[arg(short, long, value_name = "FILE", default_value = DEFAULT_FAUCET_ACCOUNT_PATH)]
-        output_path: PathBuf,
-        #[arg(short, long)]
+        #[arg(short, long, value_name = "FILE")]
+        output: PathBuf,
+        #[arg(short, long, value_name = "STRING")]
         token_symbol: String,
-        #[arg(short, long)]
+        #[arg(short, long, value_name = "U8")]
```
drahnr, line 112: Q: This is a pattern, using `value_name = ALMOST_TYPE_NAME`, but shouldn't it describe the what is to be passed? I.e. `value_name = "PRECISION"` rather than `value_name = "U8"`

bin/faucet/src/main.rs
```diff
@@ -124,42 +154,50 @@ async fn main() -> anyhow::Result<()> {
...
             // - Run and wait for server (in a spawned task)
             let faucet_future = faucet.run(rpc_client, rx_requests);
             let server_future = async {
-                let server_handle = tokio::spawn(async move {
-                    server.serve(config_endpoint).await.context("server failed")
-                });
+                let server_handle =
+                    tokio::spawn(
+                        async move { server.serve(endpoint).await.context("server failed") },
+                    );
```
drahnr, line 200: Given we `select!` later on the `JoinHandle`, is there a particular reason we do a `tokio::spawn` inside the future?

packaging/faucet/miden-faucet.service
drahnr: My comment got lost I guess :melting_face:

Mirko-von-Leipzig: Right but how do you do this programmatically from a deployment perspective?

I think my expectation would be that the package is installed with sensible defaults, and now I want to specify some of these settings as part of my deployment.

1. I install the the deb package,
2. Now how do I specify my settings?

As a general statement, the node already comes with defaults as part of the CLI - do we want/need additional env configuration that is part of the debian package that repeats those?

bin/faucet/src/main.rs
Mirko-von-Leipzig: I think the concern might be that `select!` drives the futures one at a time and therefore we wouldn't have parallelism. But I'd have to think about it some more - might not be applicable here.

In general I prefer task join sets for this sort of thing, to avoid having to worry about `select!`

packaging/faucet/miden-faucet.service
Mirko-von-Leipzig: As an example, using the `service` file approach, one can do

```
systemctl edit <service>
```

which allows editing the config extension/override file. So I imagine one could pipe things (somehow)

```
echo <settings> | systemctl edit <service>
```

which would just do the right thing. Or some syntax like that.

bin/faucet/src/main.rs
drahnr: Took a brief tour through axum: There is a `tokio::spawn` per connection in `Serve` as returned by `axum::serve`, there is some boilerplate beforhands that's being done. If the `await`-points become contended spawning is better, assuming a threaded/thread-pool executor.

.gitignore
sergerad: Sorry do you mean add a comment explaining this line?

bin/faucet/src/main.rs
sergerad: https://github.com/0xMiden/miden-node/issues/983

bin/faucet/src/main.rs
sergerad: Yea same but I thought here the help output would be more helpful for users.
```
[default: 100,500,1000]
```

It might not be, will try with _t .

bin/faucet/src/main.rs
sergerad: Yes I have been meaning to do a sweep of this - will create a separate issue for this for whole of node.

sergerad: https://github.com/0xMiden/miden-node/issues/992

bin/faucet/src/main.rs
sergerad: Yes but in that case we just end up repeating the variable name somewhat so its not more helpful for the user. For example `token_symbol` + `SYMBOL`. Whereas `U8` and `U64` do convey more information to the user for decimals and supply.

The clap docs are really liberal about the canonical usage of value_name:
> Placeholders for the argument’s values in the help message / usage.
>
>These names are cosmetic only, used for help and usage strings only. 

Happy to change if we have opinions on it. Maybe @Mirko-von-Leipzig does too.

.gitignore
drahnr: `Node files` is a bit ambiguous, and additional detail would be helpful. Feel free to ignore, micro nit

bin/faucet/src/main.rs
drahnr: No strong opinion.

Mirko-von-Leipzig: No strong opinion either; I tend to just go with whatever I think helps a random person get info from `--help`

packaging/faucet/miden-faucet.service
sergerad: Seems a comment of mine in this thread also got lost 😭 

>As a general statement, the node already comes with defaults as part of the CLI - do we want/need additional env configuration that is part of the debian package that repeats those?

So we could do away with the `.env` and as you say
1. Rely on defaults from the source;
2. Leave users / automation to specify flags or env vars manually for Dockerfile; and
3. Use the systemd override approach for Debian packages.

I do feel like the `.env` file helps out point 2 but probably not enough to outweigh its removal?

sergerad: I think it makes more sense for our automation to overwrite the application's env file rather than overriding the service file for application-level settings.

This way we standardize env var handling through the `.env` which is also helpful to users looking at the code base (rather having to look at the source code to learn the env vars and their defaults).

Both would require a service restart and the override approach would also require a daemon-reload. So the only benefit to the override approach is being able to use `systemctl cat` to see the env vars, which is no better than doing `cat /path/to/env/file` in the end.

bin/faucet/.env
Mirko-von-Leipzig, line 1: Is it correct to have the env vars be `""` empty? Is that equivalent to being unset, or equals an empty value?

packaging/faucet/miden-faucet.service
Mirko-von-Leipzig: > (rather having to look at the source code to learn the env vars and their defaults).

tbc the user should _always_ be using `--help` to see the available env vars, and the CLI default values. This is the single source of truth.

> I think it makes more sense for our automation to overwrite the application's env file 

I still feel uneasy about this. We're expecting this directory to exist and whoever installs this package will now need to go edit this additional file they don't really know about. Basically I feel like the default package install shouldn't come with an additional env file. It should just set the values which don't have defaults in the service file. The user can then configure the service with whatever method they want.

But let's see; this will probably work just fine and I'm just overly anti :)

Mirko-von-Leipzig (approved):

packaging/faucet/miden-faucet.service
sergerad: > I still feel uneasy about this. We're expecting this directory to exist and whoever installs this package will now need to go edit this additional file they don't really know about. Basically I feel like the default package install shouldn't come with an additional env file. It should just set the values which don't have defaults in the service file.

Yea this sounds like a fair criticism. Lets keep it top of mind and come back to it if required.

bin/faucet/.env
sergerad: Removed ty

PR #991, opened 2025-06-18
fix: faucet flaky test

TomasArrachea:


bin/faucet/src/faucet/mod.rs
```diff
@@ -573,6 +574,13 @@ mod tests {
         // Start the stub node
         tokio::spawn(async move { serve_stub(&stub_node_url).await.unwrap() });
 
+        // Wait for the stub node to serve requests
+        let start = Instant::now();
+        while rpc_client.get_genesis_header().await.is_err() {
+            sleep(Duration::from_millis(100)).await;
+            assert!(start.elapsed() < Duration::from_secs(60), "stub node took too long to start");
```
Mirko-von-Leipzig: A minute seems quite long - should we make this shorter? 5s maybe?

SantiagoPittella (approved):

PR #994, opened 2025-06-19
Remove duplicate ntx-builder entry in workspace members

Bilogweb3:
Fixed duplicate workspace member entry for ntx-builder that could lead to unpredictable cargo behavior during builds and publishing. Clean, simple fix to maintain workspace integrity.

image


bobbinth:
Thank you - but we do not accept simple typo fixes or single-line correction PRs.

PR #996, opened 2025-06-22
chore: upgrade protocol

Mirko-von-Leipzig:
This PR upgrades `miden-base` dependencies to latest `next`.

Notably, this includes `TransactionExecutor` becoming a short-lived object as per https://github.com/0xMiden/miden-base/pull/1469 and the `AccountFile` containing multiple authentication keys.

This primarily affected the faucet's code. Like the faucet could now be refactored slightly but I think a larger redesign is on the horizon so I've ignored that insofar as possible :)

I've also re-organized `Cargo.toml` so that we distinguish between external, node, protocol and other miden dependencies. I've confirmed that the taplo linting/formatting of it works as expected.

I'm fairly sure the `AccountFile` change should be considered a breaking change, but I don't think its worth a changelog entry? Or rather, likely we should file this under general protocol upgrades closer to the release?

### DIFF
--- a/bin/faucet/src/faucet/mod.rs
+++ b/bin/faucet/src/faucet/mod.rs
@@ -1,4 +1,4 @@
-use std::{collections::VecDeque, rc::Rc, sync::Arc};
+use std::{collections::VecDeque, sync::Arc};
 
 use anyhow::{Context, anyhow};
 use miden_lib::{
@@ -162,7 +162,7 @@ pub struct Faucet {
     // Previous faucet account states used to perform rollbacks if a desync is detected.
     prior_state: VecDeque<Account>,
     tx_prover: Arc<FaucetProver>,
-    tx_executor: Rc<TransactionExecutor>,
+    authenticator: BasicAuthenticator<StdRng>,
     account_interface: AccountInterface,
     decimals: u8,
 }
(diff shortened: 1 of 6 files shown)


bin/faucet/src/faucet/mod.rs
```diff
@@ -231,31 +231,28 @@ impl Faucet {
...
-            account_file.auth_secret_key,
-        )]));
+        let keypairs = account_file
+            .auth_secret_keys
+            .into_iter()
+            .map(|key| match key {
+                AuthSecretKey::RpoFalcon512(ref falcon) => (falcon.public_key().into(), key),
+            })
+            .collect::<Vec<_>>();
+        let authenticator = BasicAuthenticator::<StdRng>::new(&keypairs);
```
Mirko-von-Leipzig: Should we be enforcing that there is one or at least one authentication key? I'm unsure what this implies if its empty, or multiple.

bin/faucet/src/faucet/mod.rs
```diff
@@ -437,7 +434,9 @@ impl Faucet {
         &self,
         tx_args: TransactionArgs,
     ) -> MintResult<ExecutedTransaction> {
-        self.tx_executor
+        let executor =
+            TransactionExecutor::new(self.data_store.as_ref(), Some(&self.authenticator));
```
Mirko-von-Leipzig, line 441: I've left the store as an `Arc<...>` because I don't want to create more churn than required aside from bumping the version.

bin/faucet/src/faucet/mod.rs
bobbinth: For the current usage, I think we should enforce that there is exactly one (if there is none or more, probably implies a bug somewhere).

bobbinth (approved):
Looks good! Thank you!

bobbinth:
> I'm fairly sure the `AccountFile` change should be considered a breaking change, but I don't think its worth a changelog entry? Or rather, likely we should file this under general protocol upgrades closer to the release?

Yeah - fine to keep as is.

sergerad (approved):

igamigo (approved):

PR #997, opened 2025-06-22
ci: Test Docker cache

sergerad:


bin/faucet/Dockerfile
```diff
@@ -1,7 +1,11 @@
 FROM rust:1.87-slim-bullseye AS builder
 
 WORKDIR /app
-COPY . .
+COPY ./Cargo.toml .
+COPY ./Cargo.lock .
+COPY ./bin ./bin
+COPY ./crates ./crates
+COPY ./proto ./proto
```
sergerad, line 8: @Mirko-von-Leipzig As expected and mentioned, the cache doesn't really help when there are any source changes at all.

This PR will at least stop the docker build jobs taking 6mins when a PR doesn't touch the source. We could achieve the same by updating the trigger in the workflow which I am happy to do also. But these diffs are good to add in any case.

Should we create another issue to follow this up further? You shared a link recently similar to this one I believe
https://hackmd.io/jgkoQ24YRW6i0xWd73S64A

Its just not worth the time investment at this stage but good to keep track of? LMK

Mirko-von-Leipzig: #999

PR #998, opened 2025-06-23
refactor: replace integer durations with human-readable format

farrellh1:
Replaces integer-based duration arguments with human-readable duration strings using the `humantime` crate across all binary crates.

**Changes:**
- Users can now specify `--timeout 5s` instead of `--timeout 5000`
- Affects faucet, node, and proving-service binaries
- All duration arguments now accept formats like `5s`, `500ms`, `2min`, `1h`
- Environment variables updated to use new format
- Maintains backward compatibility through Duration type conversion

**Examples:**
- `--block-interval 5s` (was `--block-interval 5000`)
- `--timeout 30s` (was `--timeout 30000`) 
- `--grace-period 20s` (was `--grace-period 20`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 - Compressed faucet background image (#985).
 - Added support for TLS to miden-proving-service-client (#968).
 - Added support for TLS to faucet's connection to node RPC (#976).
-- Improved CLI help text by replacing integer-based duration arguments with human-readable duration strings (e.g., `--timeout 5s` instead of `--timeout 5000`).
+- Replaced integer-based duration args with human-readable duration strings (#998).
 
 ## v0.9.2 (2025-06-12)
 


Cargo.toml
```diff
@@ -38,6 +38,7 @@ assert_matches               = { version = "1.5" }
 async-trait                  = { version = "0.1" }
 futures                      = { version = "0.3" }
 http                         = { version = "1.3" }
+humantime                    = { version = "2.2.0" }
```
Mirko-von-Leipzig: Nit: we try not specify the patch version

```suggestion
humantime                    = { version = "2.2" }
```

bin/node/src/commands/mod.rs
```diff
@@ -15,11 +13,7 @@ const ENV_STORE_URL: &str = "MIDEN_NODE_STORE_URL";
...
-const DEFAULT_MONITOR_INTERVAL_MS: &str = "10000";
-const DEFAULT_NTX_TICKER_INTERVAL_MS: &str = "200";
-
-fn parse_duration_ms(arg: &str) -> Result<std::time::Duration, std::num::ParseIntError> {
-    arg.parse().map(Duration::from_millis)
-}
+const DEFAULT_BLOCK_INTERVAL: &str = "5s";
+const DEFAULT_BATCH_INTERVAL: &str = "2s";
+const DEFAULT_MONITOR_INTERVAL: &str = "10s";
+const DEFAULT_NTX_TICKER_INTERVAL: &str = "200ms";
```
Mirko-von-Leipzig: Is there a way to have these be `Duration` and the clap help auto-magically show the right humantime string? I suspect probably not..

farrellh1: i think there is a way you mean like this right? https://docs.rs/humantime/latest/humantime/fn.format_duration.html @Mirko-von-Leipzig

Mirko-von-Leipzig: Thanks, I'm unsure if that was an overall improvement or not :grin:  Let's see what others think.

farrellh1: okok

Mirko-von-Leipzig (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Compressed faucet background image (#985).
 - Added support for TLS to miden-proving-service-client (#968).
 - Added support for TLS to faucet's connection to node RPC (#976).
+- Improved CLI help text by replacing integer-based duration arguments with human-readable duration strings (e.g., `--timeout 5s` instead of `--timeout 5000`).
```
sergerad:
```suggestion
- Replaced integer-based duration args with human-readable duration strings (#998).
```

sergerad (approved):
LGTM. Just a change for changelog. TY!

PR #1000, opened 2025-06-23
feat: add a genesis configuration toml file

drahnr:
Ref: https://github.com/0xMiden/miden-node/issues/836
Superseeds: https://github.com/mmagician/miden-node/pull/1


crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+                    value: false.to_string(),
+                    message: "Not supported just yet".to_owned(),
+                });
+            }
+
+            let component = BasicFungibleFaucet::new(token_symbol, decimals, max_supply)?;
+
+            let account_storage_mode = storage_mode.into();
+
+            // similar to `fn create_basic_fungible_faucet`, but we need to cover more cases
```
mmagician: I don't see why exactly we can't use `create_basic_fungible_faucet`? We always have `AccountType::FungibleFaucet`, and other values can be passed via the `create_basic_fungible_faucet` interface.

I don't know if there is many people building with `NonFungibleFaucet` now, but once there's demand we could probably add helper `create_non_fungible_faucet(...)` which can be then imported for configuring the genesis here.

crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -0,0 +1,15 @@
+timestamp = 1717344256
+version   = 1
+
+[[faucet]]
+decimals     = 12
+max_supply   = 1000000
+storage_mode = "public"
+symbol       = "MID"
+
+[[account]]
```
mmagician: both the faucet and the account are actually `Account`s and I wonder if the naming might end up confusing someone trying to bootstrap the node.
Since the two supported account types are `BasicFungibleFaucet` and `BasicWallet` maybe it's worth being more verbose and use these names?

drahnr: The intent here was to be concise. There is also a dependency between the token symbol that a faucet provides, and the account having such an asset, hence I find this duality should be represented in the accounts.

In principle, I don't think having the interior representation bleed into the configuration file is a good thing. That's also the rationale of adding separate types rather than expanding those that are used internally (as far as reasonable).

Is there any immediate usecase you have in mind that the above would prohibit?

crates/store/src/genesis/config/mod.rs
drahnr: My next step would be to allow custom code, and at that point we can't use either of those helper functions. I can undo this change if desired for the scope of this PR.

mmagician: What do you mean by custom code? Would this still go into the current release or the next?

drahnr: To my understanding every account has an associated component (at least one) and so far the configuration only allows for the `Basic*` provided variants of such. The logical next step would be to allow to load a template + initial vars.

This PR is the first step in that direction, if desired.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+use rand_chacha::ChaCha20Rng;
+
+use crate::GenesisState;
+
+#[cfg(test)]
+mod tests;
+
+/// Represents an account, either a wallet or a faucet
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub enum AccountRepr {
```
sergerad: nit: I've never seen `*Repr` naming in the past at all. Is it something I have missed in the Miden repos? Otherwise it feels inconsistent with the rest of the code. What alternative name would you call this? `AccountConfig` or something?

crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -0,0 +1,15 @@
+timestamp = 1717344256
+version   = 1
+
+[[faucet]]
+decimals     = 12
+max_supply   = 1000000
+storage_mode = "public"
+symbol       = "MID"
```
sergerad: Think we use `MIDEN` for now. e.g.
```
  --token-symbol MIDEN \
```

crates/store/src/genesis/config/tests.rs
```diff
@@ -0,0 +1,16 @@
+use super::*;
+
+#[test]
+fn simple_01_works() -> Result<(), Box<dyn std::error::Error>> {
```
sergerad: nit: Would rename this test to express what it validates about the genesis file to some extent. It could become parameterized in the future if we had multiple files to test. Having the filename in the test name doesn't feel right / help the reader.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+                .build()?;
+
+            faucets.insert(symbol, faucet_account.id());
+
+            secrets.push((account_type, faucet_account.id(), secret));
+
+            all_accounts.push(faucet_account);
+        }
+
+        // then setup all wallet accounts, which reference the faucet's for their provided assets
```
sergerad: nit: I think almost all our comments begin with capital letters. And many end with a full stop. Not sure if we have discussed comment style explicitly in the past.

crates/store/src/genesis/config/samples/01-simple.toml
mmagician: > Is there any immediate usecase you have in mind that the above would prohibit?

No, I don't think this limits functionality. Only clarity.

crates/store/src/genesis/config/mod.rs
mmagician: > To my understanding every account has an associated component (at least one)

Indeed.

So the idea would be to support completely custom components, e.g. by specifying its MASM code? Or only the components we define in miden-base (which currently are just `Basic` ones as you mention)? While the details are indeed for a future discussion, at least the direction of where we want to go with this might inform some choices for this PR.

crates/store/src/genesis/config/samples/01-simple.toml
mmagician: I'm thinking about how this would tie in to [the discussion above](https://github.com/0xMiden/miden-node/pull/1000?notification_referrer_id=NT_kwDOAIA2DrMxNzE1OTA5MTc2Nzo4NDAyNDQ2&notifications_query=is%3Aunread#discussion_r2170953259). Specifically, once we have custom components code, does this become:
```
[[faucet]]
[[account]]
[[custom]]
```
?

Another option would be to have something closer to the [configuration in my original PR](https://github.com/mmagician/miden-node/pull/1/files) (although there was no custom component support):
```
[[accounts]]
type = "BasicFungibleFaucet"
...
[[accounts]]
type = "BasicWallet"
...
[[accounts]]
type = "Custom"
component = "filepath with custom masm code"?
```

mmagician: Or maybe just renaming `account` -> `wallet` for now suffices and we leave the custom components discussion for when it comes :)

crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig: We haven't had a discussion; though I think using proper sentences makes sense.

crates/store/src/genesis/config/tests.rs
drahnr: How does `fn parsing_yields_expected_default_values` sound to you? Commonly I'd do one test case per example file, hence the name referencing component.

crates/store/src/genesis/config/mod.rs
drahnr: I oftentime use `*Repr` to disambiguate parsing types from other types. It's a representation of an account in the config.
Happy to rename for consistency.

crates/store/src/genesis/config/samples/01-simple.toml
drahnr: > Or maybe just renaming `account` -> `wallet` for now suffices and we leave the custom components discussion for when it comes :)

I think that's the clearest path without introducing more complexity.

crates/store/src/genesis/config/mod.rs
drahnr: The first part of the sentence was further above, but got lost.

Sidequest: What's your take on punctuation? Titles (=first line in a comment) with `.` or without?

crates/store/src/genesis/config/mod.rs
drahnr: > So the idea would be to support completely custom components, e.g. by specifying its MASM code? Or only the components we define in miden-base (which currently are just Basic ones as you mention)? While the details are indeed for a future discussion, at least the direction of where we want to go with this might inform some choices for this PR.

I tend to say full-custom is the final destination, but I'd anneal to that state on an as-needed basis/customer feedback/internal needs.

crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig: If they're actually titles then without `.`, if you just mean a singular sentence then use a `.`. Mostly because extending the comment then results in more of a diff; same way a postfix `,` should always be required/allowed by a programming language (shame on you `json` for doing the wrong thing).

I won't really cry about any of these though :)

crates/store/src/genesis/config/samples/01-simple.toml
drahnr: Migrated to `[[wallet]]`

crates/store/Cargo.toml
Mirko-von-Leipzig, line 1: I think some (most?) of the added deps should likely be workspace ones

crates/store/src/genesis/config/samples/01-simple.toml
Mirko-von-Leipzig, line 1: We should add this to the documentation for users. This could go.. somewhere under the operator's usage section in the `docs/` book.

Essentially just explaining what this is and why it might be useful for testing.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+#[derive(Debug, Clone)]
+pub struct TestSecrets {
+    pub secrets: Vec<(AccountType, AccountId, SecretKey)>,
+}
+
+/// Specify a set of faucets and wallets with assets for easier test depoyments
+///
+/// Notice: Any faucet must be declared _before_ it's use in a wallet/regular account.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct TestGenesisConfig {
```
Mirko-von-Leipzig: WDYT about `GenesisConfig` and then we use it ourselves as well?

Users will need some way of calling this from the node binary, so we already want to add it to the CLI I think (unless I'm misunderstanding the use case @mmagician?).

If so, then we'll want to add it to the bundled `start` command as an optional path to the toml file. And if its missing then have our current config as the default.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+            let account_type = if can_be_updated {
+                AccountType::RegularAccountUpdatableCode
+            } else {
+                AccountType::RegularAccountImmutableCode
+            };
+            let account_storage_mode = storage_mode.into();
+            let (mut account, _seed) =
+                create_basic_wallet(init_seed, auth, account_type, account_storage_mode)?;
+            // By convention, 1 is the nonce for a shared account, which genesis by definition
+            // is, so all the accounts there should have nonce 1.
```
Mirko-von-Leipzig: I think this comment might be lost?

But in general, we set the nonce of accounts created in genesis to `1` because a nonce of `0` is used to indicate the state of an account that has been created locally, but not yet submitted onchain as part of a transaction.

Mirko-von-Leipzig (approved):

crates/store/src/genesis/config/mod.rs
drahnr: Did you mean `start` or `bootstrap`? Looking at the code, it'd structurally fit better with `bootstrap`.

Mirko-von-Leipzig: Apologies, I meant `bootstrap`.

crates/store/src/genesis/config/samples/01-simple.toml
drahnr: How much should we emphasize the testing aspect of it? I added some docs, but I am not sure if this addresses the initial intention sufficiently from your prespective?

bin/stress-test/src/seeding/mod.rs
```diff
@@ -68,8 +68,8 @@ pub async fn seed_store(
     // Recreate the data directory (it should be empty for store bootstrapping).
     //
     // Ignore the error since it will also error if it does not exist.
-    let _ = std::fs::remove_dir_all(&data_directory);
-    std::fs::create_dir_all(&data_directory).expect("created data directory");
+    let _ = fs_err::remove_dir_all(&data_directory);
+    fs_err::create_dir_all(&data_directory).expect("created data directory");
```
drahnr, line 72: This is a bit of a secondary one, but helped debugging and _not_ having to resort to `strace`

docs/src/operator/usage.md
```diff
@@ -32,6 +32,34 @@ miden-node bundled bootstrap \
   --accounts-directory .
 ```
 
+Alternatively, provide an explicit toml-formatted configuration format which allows you to define
```
mmagician:
```suggestion
Alternatively, provide an explicit toml-formatted configuration which allows you to define
```

docs/src/operator/usage.md
```diff
@@ -32,6 +32,34 @@ miden-node bundled bootstrap \
...
+multiple assets and faucets easily, without having to register them by additional program invocations.
```
mmagician: "program invocations" sounds like these would be `miden-node` commands, when in fact these are onchain transactions. (And a user would do it via the client, i.e. different "program")

Maybe just:
"multiple assets and faucets at chain startup."

docs/src/operator/usage.md
```diff
@@ -32,6 +32,34 @@ miden-node bundled bootstrap \
...
+
+```sh
+miden-node bundled bootstrap \
+  --data-directory data \
+  --accounts-directory . \
+  --genesis-config-file genesis.toml
+```
+
+containing i.e.
```
mmagician:
```suggestion
The genesis configuration file should contain the faucet, and optionally, wallet definitions, for example:
```

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+            let auth = AuthScheme::RpoFalcon512 { pub_key: secret_key.public_key() };
+            let init_seed: [u8; 32] = rng.random();
+
+            let assets = Result::<Vec<_>, Error>::from_iter(assets.into_iter().map(
+                |AssetEntry { amount, symbol }: AssetEntry| {
+                    let token_symbol = TokenSymbol::new(&symbol)?;
+                    let faucet_id = faucets
+                        .get(&symbol)
+                        .ok_or_else(|| Error::MissingFaucetDefinition { symbol: token_symbol })?;
+                    // TODO add non funcgible assets support
```
mmagician: Should we remove this TODO for now? It's a whole new feature

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+            // Add fungible assets.
+            let mut fungible_assets = FungibleAssetDelta::default();
+            assets
+                .into_iter()
+                .try_for_each(|fungible_asset| fungible_assets.add(fungible_asset))?;
+
+            let delta = AccountDelta::new(
+                AccountStorageDelta::default(),
+                AccountVaultDelta::new(fungible_assets, NonFungibleAssetDelta::default()),
+                Some(Felt::ONE),
```
mmagician: Question: are pre-funded genesis accounts expected to have a nonce of 1, rather than 0? They technically haven't executed any transactions.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+pub enum Error {
+    #[error(transparent)]
+    Toml(#[from] toml::de::Error),
+    #[error("Account translation from config to state failed")]
+    Account(#[from] AccountError),
+    #[error("Asset translation from config to state failed")]
+    Asset(#[from] AssetError),
+    #[error("Applying assets to account failed")]
+    AccountDelta(#[from] miden_objects::AccountDeltaError),
+    #[error("You defined an asset {symbol:?} that has no faucet producing it")]
```
mmagician: Instead of the direct "you", maybe an indirect, more neutral tone would fit better

mmagician:
```suggestion
    #[error("The defined asset {symbol:?} has no corresponding faucet")]
```

mmagician: for example

crates/store/src/genesis/config/mod.rs
drahnr: The nonce should be one, since 0 by convention is a local account that hasn't made it on-chain yet. @Mirko-von-Leipzig is that correct?

Mirko-von-Leipzig: That's correct, see https://github.com/0xMiden/miden-node/pull/776#discussion_r2017977295.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Replaced integer-based duration args with human-readable duration strings (#998 & #1014).
 - [BREAKING] Refactor the `miden-proving-service` proxy status service to use gRPC instead of HTTP (#953).
 - Added configurable network id for the faucet (#1016).
+- Added option to specify genesis state for testnets via a toml config (#1000)
```
```suggestion
- Genesis state is now configurable during bootstrapping (#1000)
```

bin/node/src/commands/bundled.rs
```diff
@@ -28,6 +30,9 @@ pub enum BundledCommand {
         // Directory to write the account data to.
         #[arg(long, value_name = "DIR")]
         accounts_directory: PathBuf,
+        /// Use the given configuration file to construct the genesis state from.
+        #[arg(long, env = ENV_GENESIS_CONFIG_FILE, value_name = "GENESIS_CONFIG")]
```
Mirko-von-Leipzig:
```suggestion
        /// Constructs the genesis block from the given toml file.
        #[arg(long, env = ENV_GENESIS_CONFIG_FILE, value_name = "FILE")]
```

docs/src/operator/usage.md
```diff
@@ -10,7 +10,7 @@ which can be accessed by appending `--help` to any of the commands.
 
 ## Bootstrapping
 
-The first step in starting a new Miden network is to initialize the genesis block data. This is a once-off operation.
+The first step in starting a new Miden network is to initialize the genesis block data. This is a one-off operation. It can either be done through the default of generating a single account and genesis with default values:
```
Mirko-von-Leipzig: I was actually curious - both `once-off` and `one-off` are correct

Mirko-von-Leipzig:
```suggestion
The first step in starting a new Miden network is to initialize the genesis block data. This is a one-off operation using the `bootstrap` command and by default the genesis block contains a single faucet account.
```

docs/src/operator/usage.md
```diff
@@ -32,6 +32,34 @@ miden-node bundled bootstrap \
   --accounts-directory .
 ```
 
+Alternatively, provide an explicit toml-formatted configuration which allows you to define
+multiple assets and faucets easily at chain startup.
```
Mirko-von-Leipzig:
```suggestion
You can also configure the account and asset data in the genesis block by passing in toml file. This is particularly useful for setting up test scenarios without requiring multiple rounds of transactions to achieve this state.
```

docs/src/operator/usage.md
```diff
@@ -32,6 +32,34 @@ miden-node bundled bootstrap \
...
+
+```sh
+miden-node bundled bootstrap \
+  --data-directory data \
+  --accounts-directory . \
+  --genesis-config-file genesis.toml
+```
+
+The genesis configuration file should contain at least one faucet, and optionally, wallet definitions with assets, for example:
```
Mirko-von-Leipzig: @PhilippGackstatter what happens if we insert assets into the genesis block which have no associated faucet account? Would that break, and/or not be possible?

@drahnr they could technically also create an empty genesis block - they technically don't need an official faucet. Or they could just define a bunch of normal accounts, though whether they can define assets without an associated faucet account is not clear to me yet ^^.

docs/src/operator/usage.md
```diff
@@ -32,6 +32,34 @@ miden-node bundled bootstrap \
...
+[[faucet]]
+decimals     = 12
+max_supply   = 10000
+storage_mode = "public"
+symbol       = "FUZZY"
+
+[[wallet]]
+assets       = [{ amount = 999, symbol = "FUZZY" }]
+storage_mode = "private"
+can_be_updated = false # default value
```
Mirko-von-Leipzig: Does this control the mutability of the accounts code?

@igamigo what do you call this in the client? So we can maybe align on terminology.

bin/node/src/commands/store.rs
```diff
@@ -95,31 +100,53 @@ impl StoreCommand {
...
-            .duration_since(UNIX_EPOCH)
-            .expect("current timestamp should be greater than unix epoch")
-            .as_secs()
-            .try_into()
-            .expect("timestamp should fit into u32");
-        let genesis_state = GenesisState::new(vec![account_file.account], version, timestamp);
+    fn bootstrap(
+        data_directory: &Path,
+        accounts_directory: &Path,
+        maybe_genesis_config_path: Option<&PathBuf>,
```
Mirko-von-Leipzig: Nit: I would just call this `genesis_config`, feels bit like hungarian notation otherwise.

bin/node/src/commands/store.rs
```diff
@@ -95,31 +100,53 @@ impl StoreCommand {
...
+                .context(format!("Read from file: {}", genesis_config_path.display()))?;
+
+            let (state, secrets) = config.into_state()?;
+            // Write the accounts to disk
+            for account in secrets.as_account_files() {
+                let accountpath = accounts_directory.join(DEFAULT_ACCOUNT_PATH);
+                account.write(accountpath)?;
+            }
+            state
+        } else {
```
Mirko-von-Leipzig: I think we can generalize this by pulling out the config, i.e.

```suggestion
        let genesis_config = if let Some(genesis_config_path) = maybe_genesis_config_path {
            let toml_str = fs_err::read_to_string(genesis_config_path)?;
            let config = GenesisConfig::read_toml(toml_str.as_str())
                .context(format!("Read from file: {}", genesis_config_path.display()))?;
            config
        } else {
            // Create the default config here.
        };
        
        // Create the state and account files
```

Mirko-von-Leipzig (commented):
Some doc suggestions and questions :)

docs/src/operator/usage.md
drahnr: I've never heard `once-off`, but it's not my monther tongue :shrug:

docs/src/operator/usage.md
drahnr: The current API does not allow to have an asset without an associated faucet identifier, at least from the API we have today.
While I think that's technically correct, I don't think it's practically useful. So I'd rather not document that.

docs/src/operator/usage.md
drahnr: I've seen mutable, but in practice that's updatable, or do we literally mutate the bytecode?

bin/node/src/commands/store.rs
drahnr: I tend to prefix the `maybe_` for optionals, and suffix anything that's a path with a `_path`. C and C++ PTSD I guess.

drahnr: No strong opinion, helps me understand code faster, even before `rust-analyzer` is ready.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+            StorageMode::Public => AccountStorageMode::Public,
+        }
+    }
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub enum AssetKind {
+    Fungible,
+    NonFungible,
+}
```
mmagician: this looks like a leftover?

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+    MaxSupplyExceedsFieldModulus { max_supply: u64, modulus: u64 },
+    #[error("Unsupported value for key {key} : {value}")]
+    UnsupportedValue {
+        key: &'static str,
+        value: String,
+        message: String,
+    },
+    #[error("Failed to create fungible faucet account")]
+    FungibleFaucet(#[from] FungibleFaucetError),
+}
```
mmagician: Some of these variants seem unused

docs/src/operator/usage.md
mmagician: IIUC this controls whether the account type ends up `RegularAccountUpdatableCode` or `RegularAccountImmutableCode`.

Personal preference: I'd choose `is_updatable` over `can_be_updated`, the latter leaves ambiguity over *who can* update.

mmagician (commented):
I'm looking forward to using this for integration testing, thanks. I left a few minor comments but functionality-wise looks good 👍🏼

docs/src/operator/usage.md
mmagician: Also atm one of `mutable` and `can_be_updated` is redundant right?

crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -0,0 +1,15 @@
...
+max_supply   = 1000000
+storage_mode = "public"
+symbol       = "MIDEN"
+
+[[wallet]]
+assets       = [{ amount = 999, symbol = "MIDEN" }]
+mutable      = false
+storage_mode = "private"
+# can_be_updated = false # default value
+# fungible = true # default value
```
mmagician: Also wallet shouldn't have a `fungible` attribute

bin/node/src/commands/store.rs
```diff
@@ -119,7 +125,29 @@ impl StoreCommand {
...
+    ) -> anyhow::Result<()> {
+        let genesis_state = if let Some(genesis_config_path) = maybe_genesis_config_path {
+            let toml_str = fs_err::read_to_string(genesis_config_path)?;
+            let config = GenesisConfig::read_toml(toml_str.as_str())
+                .context(format!("Read from file: {}", genesis_config_path.display()))?;
+
+            let (state, secrets) = config.into_state()?;
+            // Write the accounts to disk
+            for account in secrets.as_account_files() {
+                let accountpath = accounts_directory.join(DEFAULT_ACCOUNT_PATH);
```
mmagician: this only creates a single file with secrets that each subsequent account overwrites

mmagician (changes requested):

bin/node/src/commands/store.rs
drahnr: Good catch!

docs/src/operator/usage.md
drahnr: `mutable` in the context is synonym to `updatable` to my understanding? `mutable` is obsolete and was replaced with `updatable`. The code should not have mutable anymore, at least `rg mutable` doesn't give me anything in branch `next`.

crates/store/src/genesis/config/mod.rs
drahnr: Removed

crates/store/src/genesis/config/mod.rs
drahnr: Some of them are not explicitly called, but the conversions are implicit with `?`. Which one specifically do you think is unused? Clippy _should_ have pointed that out

crates/store/src/genesis/config/samples/01-simple.toml
drahnr: Fixed

bin/node/src/commands/store.rs
drahnr: Changed to `config_path`

crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig: Its possible they're unused because they're part of the public API and so `clippy` won't point them out. You might be able to find these by switching it to `pub(crate)` and `cargo clippy -p miden-node-store`.

drahnr: I just double checked, and all variants are constructed. Also added a variant to verify unused variants are found.

crates/store/src/genesis/config/tests.rs
sergerad: LGTM

crates/store/src/genesis/config/errors.rs
```diff
@@ -0,0 +1,29 @@
...
+#[allow(missing_docs, reason = "Error variants must be descriptive by themselves")]
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error(transparent)]
+    Toml(#[from] toml::de::Error),
+    #[error("Account translation from config to state failed")]
+    Account(#[from] AccountError),
+    #[error("Asset translation from config to state failed")]
+    Asset(#[from] AssetError),
+    #[error("Applying assets to account failed")]
```
sergerad: nit: People were asking me to lowercase error messages in a PR recently.

And was shared this which was good https://github.com/0xMiden/miden-base/discussions/966

sergerad (approved):
LGTM

bin/node/src/commands/bundled.rs
```diff
@@ -28,6 +30,9 @@ pub enum BundledCommand {
         // Directory to write the account data to.
         #[arg(long, value_name = "DIR")]
         accounts_directory: PathBuf,
+        /// Constructs the genesis block from the given toml file.
+        #[arg(long, env = ENV_GENESIS_CONFIG_FILE, value_name = "FILE")]
+        genesis_config_file: Option<PathBuf>,
```
sergerad, line 35: Is the PR name erroneous because it says _**test**_ genesis configuration toml file? As in this flag could/would be used to set up any network including our devnet if we re-genesis it for example.

Just realized that my assumption of the use cases for this may not be correct.

drahnr: Rename the PR, it started of, but the purpose expanded a bit.

crates/store/src/genesis/config/errors.rs
drahnr: TIL: lowercase is grammatically acceptable after a `:` https://www.grammar-monster.com/punctuation/colons_capitalization_after_a_colon.html

bin/node/src/commands/store.rs
```diff
@@ -119,7 +128,29 @@ impl StoreCommand {
...
+            let (state, secrets) = config.into_state()?;
+            // Write the accounts to disk
+            for AccountFileWithName { account_file, name } in secrets.as_account_files() {
+                let accountpath = accounts_directory.join(name);
+                account_file.write(accountpath)?;
+            }
+            state
+        } else {
+            Self::default_genesis_state(accounts_directory)?
+        };
```
Mirko-von-Leipzig: WDYT about something more like this? 

```suggestion
        let config = genesis_config
            .map(GenesisConfig::from_toml_file)
            .transpose()?
            .unwrap_or_default();
            
            let (state, secrets) = config.into_state()?;
            // Write the accounts to disk
            for AccountFileWithName { account_file, name } in secrets.as_account_files() {
                let accountpath = accounts_directory.join(name);
                account_file.write(accountpath)?;
            }
            state
```

crates/store/Cargo.toml
```diff
@@ -16,19 +16,24 @@ workspace = true
 
 [dependencies]
 anyhow                 = { workspace = true }
+chrono                 = { version = "0.4" }
```
Mirko-von-Leipzig, line 19: I couldn't quite spot where we use this

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+                AccountType::FungibleFaucet
+            } else {
+                AccountType::NonFungibleFaucet
+            };
+
+            if !fungible {
+                return Err(Error::UnsupportedValue {
+                    key: "fungible",
+                    value: false.to_string(),
+                    message: "Not supported just yet".to_owned(),
```
```suggestion
                    message: "Non-fungible assets are not supported yet".to_owned(),
```

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+        for FaucetConfig {
+            name,
+            symbol,
+            decimals,
+            max_supply,
+            storage_mode,
+            fungible,
+        } in repr_faucets
+        {
+            let mut rng = ChaCha20Rng::from_seed(rand::random());
```
Mirko-von-Leipzig, line 105: Should we be re-using this `rng`? I'm unsure tbh.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+    /// Notice: It will generate the specified case during [`fn into_state`].
+    pub fn read_toml(toml_str: &str) -> Result<Self, Error> {
+        let me = toml::from_str::<Self>(toml_str)?;
+        Ok(me)
+    }
+
+    /// Convert the in memory representation into the new genesis state
+    ///
+    /// Notice: Generates keys and returns the secret keys, hence this is not sane to be used
+    /// for production environments. There, you want to generate the keys externally.
```
Mirko-von-Leipzig: I'm not sure this statement is true. Why shouldn't they use the keys generated here?

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+/// Represents a faucet with asset specific properties
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct FaucetConfig {
+    /// Provide a name, that will be used for the keyfile.
+    #[serde(default)]
+    name: Option<String>,
+    // TODO eventually directly parse to `TokenSymbol`
+    symbol: String,
```
Mirko-von-Leipzig: Couple of `TokenSymbol` todo's still. Are we punting that to a follow-up?

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+            StorageMode::Private => AccountStorageMode::Private,
+            StorageMode::Public => AccountStorageMode::Public,
+        }
+    }
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+struct AssetEntry {
+    symbol: String, // TODO move to a wrapper around `TokenSymbol`
+    amount: u64,    // TODO we might want to provide `humantime`-like denominations
```
Mirko-von-Leipzig: What did you have in mind by humanlike?

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+const fn ja() -> bool {
+    true
+}
+
+/// Represents a wallet, containing a set of assets
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct WalletConfig {
+    /// Provide a name, that will be used for the keyfile.
+    #[serde(default)]
+    name: Option<String>,
```
Mirko-von-Leipzig: I'm in two minds about this. Its a nice touch but do we really need it?

Should we not just generate files `account_{i}` in the order they're specified? Though I notice this won't be true as they're separate properties 🤔

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+    // name, account, private key, account seed
+    pub secrets: Vec<(Option<String>, Account, SecretKey, Word)>,
+}
+
+impl AccountSecrets {
+    /// Convert the internal tuple into an `AccountFile`
+    ///
+    /// If no name is present, a new one is generated based on the current time
+    /// and the index in
+    pub fn as_account_files(&self) -> impl Iterator<Item = AccountFileWithName> {
```
Mirko-von-Leipzig: As mentioned elsewhere I'm unsure about this.

Should we not keep it simple and name them `account_{i}` with no exceptions. The configurer can rename, move or otherwise do whatever they want with them.

Using the index is correct here; but doesn't line up with the order in which they're defined in the toml so this would be extremely confusing to the configurer.

I don't think the time adds anything - if they clobber existing keys we should either let them, or panic and avoid that. Probably the latter.

Mirko-von-Leipzig: I would 
- combine the config/toml wallets into a single enum property so they can be ordered
- name the files `account_{i}`
- avoid time and just refuse to clobber existing files

Alternatively, require a file name as part of the config. But I wouldn't allow flexibility here.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+
+        // First setup all the faucets
+        for FaucetConfig {
+            name,
+            symbol,
+            decimals,
+            max_supply,
+            storage_mode,
+            fungible,
+        } in repr_faucets
```
Mirko-von-Leipzig: Should we make these methods on the types? Though I'm unsure how much that captures from the outside, might be too much; though could only be the `rng`..

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+            })?;
+
+            let account_type = if fungible {
+                AccountType::FungibleFaucet
+            } else {
+                AccountType::NonFungibleFaucet
+            };
+
+            if !fungible {
+                return Err(Error::UnsupportedValue {
```
Mirko-von-Leipzig: Is this implementable in the short term? Or is it fairly complex?

I guess I'm wondering if we aren't able to provide this fairly soon, should we bother adding the config for. And would these have a similar config interface, or should we physically split them into `FungibleFaucet` and `NonFungibleFaucet` with very different fields..

I'm not versed enough with the details here; @mmagician @drahnr thoughts?

docs/src/operator/usage.md
```diff
@@ -32,6 +34,36 @@ miden-node bundled bootstrap \
   --accounts-directory .
 ```
 
+You can also configure the account and asset data in the genesis block by passing in toml file.
```
Mirko-von-Leipzig:
```suggestion
You can also configure the account and asset data in the genesis block by passing in a toml configuration file.
```

crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig: Another alternative is to name them `faucet_{symbol}` and then `wallet_{i}`? That could also be acceptable I think, and lets you order them separately still

crates/store/src/genesis/config/mod.rs
drahnr: I was thinking `k` and `m` denominations, but might as well just drop the comment

bin/node/src/commands/store.rs
drahnr: That is much nicer! Ty!

drahnr: Implemented

crates/store/src/genesis/config/mod.rs
drahnr: Dropped the comment

crates/store/src/genesis/config/mod.rs
drahnr: To me, these are just _better tuples_ in a sense.

Or the comments anchor isn't quite right :monocle_face:

crates/store/src/genesis/config/mod.rs
drahnr: We don't have a default implementation i.e. `miden_lib::faucet::BasicNonFungible` doesn't exist yet, so I'd push that to a follow up PR. If we want to make it configurable using custom code, I think that expands the scope significantly, not sure that's necessary _right now_.

crates/store/src/genesis/config/mod.rs
drahnr: > Another alternative is to name them faucet_{symbol} and then wallet_{i}? That could also be acceptable I think, and lets you order them separately still

I'll got this way

crates/store/src/genesis/config/mod.rs
drahnr: Discard, given that it's a different subcommand from `Start`, it's fine. Removed

crates/store/src/genesis/config/mod.rs
drahnr: For this to be meaningful, we'd require to change `miden_lib` to have a `Eq, Hash`  on `TokenSymbol`, otherwise we incur just more conversions back and forth for no good reason. We validate shortly after parsing anyways, so I'd consider this a nice to have with little payout.

The parsing hooks are the easy part, but there are - again - more ripples:

```rust
fn deserialize_token_symbol<'de, D>(deserializer: D) -> Result<TokenSymbol, D::Error>
where
    D: serde::de::Deserializer<'de>,
{
    struct TokenSymbolVisitor;

    impl<'dev> serde::de::Visitor<'dev> for TokenSymbolVisitor {
        type Value = TokenSymbol;
        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
        where
            E: serde::de::Error,
        {
            TokenSymbol::new(v).map_err(serde::de::Error::custom)
        }
        fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
            write!(formatter, "a token symbol from consisting of 3..=5 ASCII characters")
        }
    }

    deserializer.deserialize_str(TokenSymbolVisitor)
}

fn serialize_token_symbol<S>(symbol: &TokenSymbol, serializer: S) -> Result<S::Ok, S::Error>
where
    S: serde::ser::Serializer,
{
    serializer.serialize_str(symbol.to_string().map_err(serde::ser::Error::custom)?.as_str())
}
```

with 

```diff
-    // TODO eventually directly parse to `TokenSymbol`
-    symbol: String,
+    #[serde(deserialize_with = "deserialize_token_symbol")]
+    #[serde(serialize_with = "serialize_token_symbol")]
+    symbol: TokenSymbol,
```

crates/store/src/genesis/config/mod.rs
drahnr: Implemented

crates/store/src/genesis/config/mod.rs
drahnr: Removed the `name` fields

crates/store/src/genesis/config/mod.rs
drahnr: I think this is the safe approach, otherwise there'd be a relation between the generated keys with the nonce. Given the ringbuffer nature of ChaCha, I don't know how difficult it is to map the state of nonce `n` to state of nonce `n+1`, and hence once the seed of one key is leaked that _might_ be an issue for all keys generated due to that relation.
Regardless, the above is the safe bet.

drahnr: Different question: What's the rationale of using `ChaCha`-wrapped OS rng? Is it a trust thing that the OS randomness is not good enough/depleted/skewed? This particular code is taken from another place in the codebase.

Mirko-von-Leipzig: I have absolutely no idea :) Maybe @bobbinth or @igamigo know?

igamigo: Not sure either. Seems like it was added [here](https://github.com/0xMiden/miden-node/pull/517/files#diff-57465a9c81c91d1a07596204d6ed2adf7c39ce5136d3201c2370cb46f36fa67eR141), but there does not seem to be any valuable context in relation to this

crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig: Should we then call this a `FungibleFaucet` and remove the `funbigble: bool` since we don't know what shape that will require?

For example, a non-fungible faucet might not have a supply of tokens, but more likely would supply some sort of hashing function or whatever to determine the `NFT`. Essentially I don't think they'll be compatible so we shouldn't even give the option yet.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+            let delta = AccountDelta::new(
+                wallet_account.id(),
+                AccountStorageDelta::default(),
+                AccountVaultDelta::new(fungible_assets, NonFungibleAssetDelta::default()),
+                Felt::ONE,
+            )?;
+            wallet_account.apply_delta(&delta)?;
+
+            secrets.push((
+                format!("wallet_{index:02}.mac"),
```
Mirko-von-Leipzig: This might cause issues for programmatically handling this when we have more than 100 wallets.

I do like that it lets the filesystem order them correctly, but the trade-off is that users can't do more than 100 without _knowing_ that this is a special number.


```suggestion
                format!("wallet_{index}.mac"),
```

crates/store/src/genesis/config/mod.rs
bobbinth: I don't think there is any particular reason for this except that we probably do it the same way in other places where we get the seed not from OS rng, but from somewhere else. So, should be fine to change here.

docs/src/operator/usage.md
```diff
@@ -32,6 +34,36 @@ miden-node bundled bootstrap \
...
+
+```toml
+timestamp = 1717344256
+version   = 1
+
+[[faucet]]
+decimals     = 12
+max_supply   = 10000
+storage_mode = "public"
+symbol       = "FUZZY"
```
bobbinth: Where did we land on whether `max_supply` specifies the number of tokens or the number of base units? Either way though, we should:

- Reduce the number of decimal places to something like 6 or 8 (12, I think is a bit too much, that we limit the total number of base units to $2^{63}$).
- Add comments explaining how to interpret `max_supply`.
- If we do specify max supply in base units, we should make sure that it is consistent with decimal places.

docs/src/operator/usage.md
```diff
@@ -32,6 +34,36 @@ miden-node bundled bootstrap \
...
+
+[[wallet]]
+assets       = [{ amount = 999, symbol = "FUZZY" }]
+storage_mode = "private"
+can_be_updated = false # default value
```
bobbinth: I would probably use something like `updatable_code` instead of `can_be_updated` to indicate which part of the account is affected by this setting.

docs/src/operator/usage.md
```diff
@@ -32,6 +34,36 @@ miden-node bundled bootstrap \
...
+  --accounts-directory . \
+  --genesis-config-file genesis.toml
+```
+
+The genesis configuration file should contain at least one faucet, and optionally, wallet definitions
+with assets, for example:
+
+```toml
+timestamp = 1717344256
+version   = 1
```
bobbinth: Should we add simple comments to explain what `version` and `timestamp` mean?

crates/store/src/genesis/config/samples/01-simple.toml
bobbinth, line 1: This file would need to be synchronized with the example in the docs (e.g., `is_updatable` vs. `can_be_updated`).

crates/store/src/genesis/config/errors.rs
```diff
@@ -0,0 +1,32 @@
+use miden_lib::account::faucets::FungibleFaucetError;
+use miden_objects::{AccountError, AssetError, TokenSymbolError, asset::TokenSymbol};
+
+#[allow(missing_docs, reason = "Error variants must be descriptive by themselves")]
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
bobbinth: Just `Error` seems a bit too general - should this be `GenesisError`?

crates/store/src/genesis/config/errors.rs
```diff
@@ -0,0 +1,32 @@
...
+    #[error(transparent)]
+    Toml(#[from] toml::de::Error),
+    #[error("account translation from config to state failed")]
+    Account(#[from] AccountError),
+    #[error("asset translation from config to state failed")]
+    Asset(#[from] AssetError),
+    #[error("applying assets to account failed")]
+    AccountDelta(#[from] miden_objects::AccountDeltaError),
```
bobbinth: nit: I would write this as "adding asset to account failed".

crates/store/src/genesis/config/errors.rs
```diff
@@ -0,0 +1,32 @@
...
+    #[error("the defined asset {symbol:?} has no corresponding faucet")]
+    MissingFaucetDefinition { symbol: TokenSymbol },
+    #[error(transparent)]
+    TokenSymbol(#[from] TokenSymbolError),
+    #[error("the provided max supply {max_supply} exceeds the field modulus {modulus}")]
+    MaxSupplyExceedsFieldModulus { max_supply: u64, modulus: u64 },
```
bobbinth: This should probably be something like `MaxSupplyOutOfRange`. We currently limit max supply of a fungible token to at most $2^{63}$ and this is likely to be reduced in the future to $2^{63} - 2^{31}$.

bin/node/src/commands/store.rs
```diff
@@ -95,34 +86,53 @@ impl StoreCommand {
...
+        Felt, ONE,
+        account::{Account, AccountFile, AuthSecretKey},
+        asset::TokenSymbol,
+        crypto::dsa::rpo_falcon512::SecretKey,
+    };
+    use rand::{Rng, SeedableRng};
+    use rand_chacha::ChaCha20Rng;
+
+    /// Legacy implementation, now superseded by [`miden_store::GenesisConfig::default`]
     fn generate_genesis_account() -> anyhow::Result<AccountFile> {
```
bobbinth: Is there a reason to keep the legacy implementation?

docs/src/operator/usage.md
```diff
@@ -10,7 +10,9 @@ which can be accessed by appending `--help` to any of the commands.
 
 ## Bootstrapping
 
-The first step in starting a new Miden network is to initialize the genesis block data. This is a once-off operation.
+The first step in starting a new Miden network is to initialize the genesis block data. This is a
+one-off operation using the `bootstrap` command and by default the genesis block contains a single
+faucet account.
```
bobbinth: nit: we should update the comments on line 25 below. Maybe something like: "The generated genesis block will contain a single...".

docs/src/operator/usage.md
```diff
@@ -32,6 +34,36 @@ miden-node bundled bootstrap \
   --accounts-directory .
 ```
 
+You can also configure the account and asset data in the genesis block by passing in a toml configuration file.
+This is particularly useful for setting up test scenarios without requiring multiple rounds of
+transactions to achieve this state.
```
bobbinth: Should we also mention that private keys for these accounts will be written to files?

In the future (not for this PR), we could set authentication methods for these accounts (e.g., for testing purposes, most of these could probably be `NoAuth`).

crates/store/src/genesis/config/mod.rs
bobbinth: Yes, I would change the usage of just `Faucet` to `FungibleFaucet` here and in other places (e.g., in the toml file itself and in the docs).

crates/store/src/genesis/config/mod.rs
bobbinth, line 1: I pushed a small commit that re-organizes this file and adds some sections separators.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+
+#[cfg(test)]
+mod tests;
+
+/// Represents an account, either a wallet or a faucet
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub enum AccountConfig {
+    Wallet(WalletConfig),
+    Faucet(FaucetConfig),
+}
```
bobbinth: Is this enum needed? Doesn't seems like it is currently being used.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+            AccountFileWithName { name, account_file }
+        })
+    }
+}
+
+// HELPER FUNCTIONS
+// ================================================================================================
+
+/// Calculate the max supply of the token.
+fn max_supply_in_undividable_units(
```
bobbinth: nit: I would call these "base units" rather than "undividable units".

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+    let base_unit = 10u64.pow(u32::from(decimals));
+    let max_supply =
+        max_supply_in_token_units
+            .checked_mul(base_unit)
+            .ok_or_else(|| Error::OutOfRange {
+                max_supply: max_supply_in_token_units,
+                decimals,
+            })?;
+    let max_supply = Felt::try_from(max_supply)
+        .map_err(|_| Error::MaxSupplyExceedsFieldModulus { max_supply, modulus: Felt::MODULUS })?;
```
bobbinth: As mentioned in other comments, max supply of fungible faucets is limited by `FungibleAsset::MAX_AMOUNT` which is currently set to $2^{63}$.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+
+            // It's similar to `fn create_basic_fungible_faucet`, but we need to cover more cases
+            let (faucet_account, faucet_account_seed) = AccountBuilder::new(init_seed)
+                .account_type(account_type)
+                .storage_mode(account_storage_mode)
+                .with_auth_component(auth)
+                .with_component(component)
+                .build()?;
+
+            faucets.insert(symbol.clone(), faucet_account.id());
```
bobbinth: Are we checking that duplicate faucet definitions are not provided in the config file? If two faucets with the same token symbol are specified, we should probably error out.

crates/store/src/genesis/config/mod.rs
drahnr: At generation time we actually would know the number, and could use the syntax `{index:0width$}` where `let width = usize::ilog(n);`, it just seemed unrealistic to have more than 100 keys. Will add

docs/src/operator/usage.md
drahnr: For practicality and parsing simplicity, I'd say base units. I'll add a comment accordingly.

bin/node/src/commands/store.rs
drahnr: Only for the unit test, otherwise we could get rid of it.

crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig: I'm more thinking from the user's perspective. But maybe I'm over-thinking it.

Essentially you can't write user code that generates `n` files and then does something with `n` files because file `i`'s name depend on the size of `log10(n)`.

bin/node/src/commands/store.rs
bobbinth: I think we can probably get rid of it - but also defer to @Mirko-von-Leipzig.

drahnr: I'll delete it and the test, it was more useful for development, but has less value moving forward.

crates/store/src/genesis/config/errors.rs
drahnr: That's a different error though, and the artifical upper bound is checked before the `Felt::try_from` which is effectively the source of this. I'd rather keep the error variant  that is only going to bubble up on human-error, rather than using `expect`, iff we are sure the constant for max supply is never going to exceed the field modulus.

crates/store/src/genesis/config/mod.rs
drahnr: Done

crates/store/src/genesis/config/mod.rs
drahnr: tl;dr I think sorting is preferable


Usecases:
1. user wants to read/ingest all keys
2. user wants to read a specifc key
3. user wants to know the length of all keys

---
1. in this case the user already needs to know the index, which means leakage of information from the toml into the user code, so having `n` isn't a signficant requirement
2. the user wants to read all key files - in case of consecutive numbers indeed we have predictable paths and can use that to our advantage, for `n`-derived/`0`-prefixed names we have to find the first via `glob` or `regex` in the directory listing and derive the zero padding length. Then we can do the same. This is an additional step. Alternatively one can use the dirlisting and use a `regex` match `account0*[0-9]+` to find all keys. 
3. either way the user either needs the information hardcoded or use the directory listing with a regex pattern, or test all key indices consecutively (or via binary expansion search) - only relevant for large key sizes

Best case, we extract the generation as a library and the user would use the API return value, in which case it doesn't matter and things happen in memory, which is my preferred path.
For file based keys, I'd prefer to use an optional path to allow the user to name them as desired.

crates/store/src/genesis/config/mod.rs
drahnr: Removed, since everything is in base units now

crates/store/src/genesis/config/mod.rs
drahnr: Q: should I use an expect then? The "currently" is my only concern :)

crates/store/src/genesis/config/samples/01-simple.toml
drahnr: Synchronized to `has_updatable_code`

docs/src/operator/usage.md
drahnr: I added them and explained them according to my understanding of the code.

crates/store/src/genesis/config/mod.rs
drahnr: Implemented and bails

docs/src/operator/usage.md
```diff
@@ -32,6 +34,59 @@ miden-node bundled bootstrap \
...
+timestamp = 1717344256
+# Defines the format of the block protocol to use
+# for the genesis block.
+version   = 1
+
+[[fungible_faucet]]
+# The give faucet is a fungible one.
+# Defaults to `true` and is the only mode supported
+# for the time being.
+# fungible = true
```
bobbinth: Since this is already in the `fungible_faucet` table, I don't think we need the `fungible` flag.

docs/src/operator/usage.md
```diff
@@ -32,6 +34,59 @@ miden-node bundled bootstrap \
...
+The genesis configuration file should contain at least one faucet, and optionally, wallet definitions
+with assets, for example:
+
+```toml
+# The timestamp associated to the genesis block.
+# It will influence the hash of the genesis block.
+timestamp = 1717344256
+# Defines the format of the block protocol to use
+# for the genesis block.
+version   = 1
```
bobbinth: nit: why do we wrap comments so early? Should we do at least 60 or maybe even 80 chars per line?

crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -0,0 +1,15 @@
+timestamp = 1717344256
+version   = 1
+
+[[fungible_faucet]]
+decimals     = 12
+max_supply   = 1000000
+storage_mode = "public"
+symbol       = "MIDEN"
+# fungible = true # default value
```
bobbinth: Same as one of the previous comments: I don't think we need `fungible` flag because we already define this as a fungible faucet.

crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -0,0 +1,15 @@
+timestamp = 1717344256
+version   = 1
+
+[[fungible_faucet]]
+decimals     = 12
+max_supply   = 1000000
```
bobbinth: I don't think the combination of `max_supply` and `decimals` make sense here.

crates/store/src/genesis/config/errors.rs
bobbinth: I think this still needs to be addressed.

crates/store/src/genesis/config/errors.rs
bobbinth: I think we can basically guarantee that `FungibleAsset::MAX_AMOUNT` will never overflow field modulus. If we wanted to protect against this, I'd add an `assert` - because if that happens we have much bigger problems and the node should panic.

So, what we really want to check is that `max_supply` is smaller than `FungibleAsset::MAX_AMOUNT`.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+                .with_auth_component(auth)
+                .with_component(component)
+                .build()?;
+
+            faucet_account.apply_delta(&AccountDelta::new(
+                faucet_account.id(),
+                AccountStorageDelta::new(),
+                AccountVaultDelta::default(),
+                ONE,
+            )?)?;
```
bobbinth: `Account::increment_nonce()` is now public.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+            let assets = Result::<Vec<_>, GenesisConfigError>::from_iter(assets.into_iter().map(
+                |AssetEntry { amount, symbol }: AssetEntry| {
+                    let token_symbol = TokenSymbol::new(&symbol)?;
+                    let faucet_id = faucets.get(&symbol).ok_or_else(|| {
+                        GenesisConfigError::MissingFaucetDefinition { symbol: token_symbol }
+                    })?;
+
+                    Ok(FungibleAsset::new(*faucet_id, amount)?)
+                },
+            ))?;
```
bobbinth: nit: I would maybe move this out into a small helper function.

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+            };
+            let account_storage_mode = storage_mode.into();
+            let (mut wallet_account, wallet_account_seed) =
+                create_basic_wallet(init_seed, auth, account_type, account_storage_mode)?;
+            wallet_account.apply_delta(&AccountDelta::new(
+                wallet_account.id(),
+                AccountStorageDelta::new(),
+                AccountVaultDelta::default(),
+                ONE,
+            )?)?;
```
bobbinth: Since we are incrementing the nonce below (line 201), we don't need to increment it here.

bobbinth (commented):
Looks good! Thank you. I left some more comments inline.

docs/src/operator/usage.md
bobbinth: This comment still needs to be addressed.

crates/store/src/genesis/config/mod.rs
bobbinth: There are 3 things here:

- We need to make sure that `max_supply <= FungibleAsset::MAX_AMOUNT`. This should be an error.
- After we've checked this, we need to convert it to `Felt` and here using `expect()` is fine.
- To be sure that `FungibleAsset::MAX_AMOUNT` is smaller than `Felt::MODULUS` (which it should always be) we can add a compile-time check.

crates/store/src/genesis/config/mod.rs
drahnr: I checked yesterday after a rebase, and it still wasn't in the reference `miden-base`. Will double check.

docs/src/operator/usage.md
drahnr: Fixed

crates/store/src/genesis/config/samples/01-simple.toml
drahnr: Valid, but not sensible

crates/store/src/genesis/config/mod.rs
drahnr: Fixed

docs/src/operator/usage.md
```diff
@@ -32,6 +34,50 @@ miden-node bundled bootstrap \
...
+
+[[fungible_faucet]]
+# The token symbol to use for the token
+symbol       = "FUZZY"
+# Number of decimals your token will have, it effectively defines the fixed point accuracy.
+decimals     = 6
+# Total supply, in _base units_
+#
+# Use i.e. a max supply of `1e15` _base untis_ and decimals set to `6`, will yield you a total supply of
```
Mirko-von-Leipzig: `e.g.` is `for example` `i.e.` is `that is`

```suggestion
# e.g. a max supply of `1e15` _base units_ and decimals set to `6`, will yield you a total supply of
```

docs/src/operator/usage.md
```diff
@@ -32,6 +34,50 @@ miden-node bundled bootstrap \
...
+  --data-directory data \
+  --accounts-directory . \
+  --genesis-config-file genesis.toml
+```
+
+The genesis configuration file should contain at least one faucet, and optionally, wallet definitions
+with assets, for example:
+
+```toml
+# The timestamp associated to the genesis block. It will influence the hash of the genesis block.
```
Mirko-von-Leipzig:
```suggestion
# The UNIX timestamp of the genesis block. It will influence the hash of the genesis block.
```

crates/store/src/genesis/config/tests.rs
```diff
@@ -0,0 +1,35 @@
...
+    let gcfg = GenesisConfig::default();
+    let (state, secrets) = gcfg.into_state().unwrap();
+    let mut iter = secrets.as_account_files();
+    let AccountFileWithName { account_file: status_quo, .. } = iter.next().unwrap();
+    assert!(iter.next().is_none());
+
+    assert_eq!(status_quo.account.nonce(), ONE);
+
+    let block = state.into_block()?;
+    dbg!(block.inner().header());
```
Mirko-von-Leipzig:
```suggestion
```

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    wallet: Vec<WalletConfig>,
+    fungible_faucet: Vec<FungibleFaucetConfig>,
+}
+
+impl Default for GenesisConfig {
+    fn default() -> Self {
+        Self {
+            version: 1_u32,
+            timestamp: u32::try_from(chrono::Local::now().timestamp())
+                .expect("Timestamp should fit into u32"),
```
Mirko-von-Leipzig: Do we need `chrono` for this? I feel like this should just be something like

```rust
SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs()
```

crates/store/src/genesis/config/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+            // The genesis block is special in that accounts are "deplyed" without transactions and
+            // therefore we need bump the nonce manually to uphold this invariant.
+            let delta = AccountDelta::new(
+                wallet_account.id(),
+                AccountStorageDelta::default(),
+                AccountVaultDelta::new(fungible_assets, NonFungibleAssetDelta::default()),
+                Felt::ONE,
+            )?;
+
+            wallet_account.apply_delta(&delta)?;
```
Mirko-von-Leipzig: Something that just occurred to me.. Do we ever subtract these assets from the original faucets?

I don't think we do currently. I imagine this sort of book-keeping is usually enforced by virtue of the note's system. But iiuc we would break the stated max-supply invariant because the faucet would be at max supply but wallets would have additional amounts?

Mirko-von-Leipzig (commented):
This is looking good. Some minor stuff - but the one question about asset subtraction from faucets which tbh I hadn't even thought of before now.

crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig: @bobbinth maybe you have some ideas here? Or is it just a literal case of subtracting the amounts from the faucets already? Which probably means we have to re-order some of this.

bobbinth: Ah good point! I missed this entirely. Yes, we do need to subtract them otherwise we may end up have more assets circulating than the `max_supply` of the faucet. The way to do it is to add the sum of all issued assets from the number of currently issues assets stored in the first slot of the faucet account's storage.

Conceptually, it is not difficult to do, but as you said, may require some rework - so, we could do it in a follow up PR.

Mirko-von-Leipzig: Created #1059

crates/store/src/genesis/config/mod.rs
drahnr: Historically I've been bitten by the `.unwrap()` and lack of leap second handling and friends, so I default to `chrono`. I'll use `SystemTime` since there is no good reason unless we want to accept other kinds of values for `timestamp` (i.e. rfc based time specification) .

drahnr: Note: It's already part of `miden-node-utils` so it's not an additional dependency.

Mirko-von-Leipzig: Ah it gets pulled in via the `tracing-forest` dep

crates/store/src/genesis/config/errors.rs
```diff
@@ -0,0 +1,36 @@
...
+    TokenSymbol(#[from] TokenSymbolError),
+    #[error("unsupported value for key {key} : {value}")]
+    UnsupportedValue {
+        key: &'static str,
+        value: String,
+        message: String,
+    },
+    #[error("failed to create fungible faucet account")]
+    FungibleFaucet(#[from] FungibleFaucetError),
+
```
drahnr:
```suggestion
```

PR #1002, opened 2025-06-23
ci: Split debian steps into jobs

sergerad:
## Context

We forgot to split the Debian publish steps into separate jobs. They are all serially executed because of this.

## Changes
- Move Debian publish steps into separate jobs for each service


bobbinth (approved):
Looks good! Thank you!

PR #1004, opened 2025-06-23
chore(remote-prover): rename proving-service to remote-prover

### LINKED ISSUE #941: Rename proving service crates and Debian packages

- [ ] RPC
- [ ] Block producer
- [ ] Store
- [ ] Network transaction builder
- [x] Proving service
- [ ] Protobuf messages
- [ ] Testing

### What should be done?

Rename the following
- miden-proving-service -> miden-delegated-prover
- miden-proving-service-client -> miden-delegated-prover-client
- miden-prover -> miden-prover-worker
- miden-prover-proxy -> [unchanged]

### How should it be done?

Update absolutely every reference to miden-proving-service, miden-proving-service-client and miden-prover
- dir and file names
- crate names (Cargo.toml)
- Github workflows
- packaging dirs
- module names / source contents

### When is this task done?

There are no results for the following command
```
git grep proving-service
```

### Additional context

_No response_

SantiagoPittella:
closes #941


CHANGELOG.md
```diff
@@ -11,13 +11,13 @@
...
-- Added `miden-proving-service` and `miden-proving-service-client` crates (#926).
+- Added `miden-remote-prover` and `miden-remote-prover-client` crates (#926).
 - Added support for gRPC server side reflection to all components (#949).
 - Compressed faucet background image (#985).
-- Added support for TLS to `miden-proving-service-client` (#968).
+- Added support for TLS to `miden-remote-prover-client` (#968).
 - Added support for TLS to faucet's connection to node RPC (#976).
 - Replaced integer-based duration args with human-readable duration strings (#998).
-- [BREAKING] Refactor the `miden-proving-service` proxy status service to use gRPC instead of HTTP (#953).
+- [BREAKING] Refactor the `miden-remote-prover` proxy status service to use gRPC instead of HTTP (#953).
```
sergerad: The changelog shouldn't modify existing references to miden-proving-service. And should specify the change of name done by this PR (understand you are aware of this just probably overlooked these diffs given how many there are).

sergerad (changes requested):
Thanks, looks good. Just the changelog comment and the `ProvingServiceError`:
Screenshot 2025-06-25 at 11 07 12 AM

sergerad (approved):

proto/proto/proxy_status.proto
bobbinth, line 1: I find the naming here and related files a bit too "open-ended" - i.e., it is not clear that the proxy is for the remote prover. Maybe we should put all remote prover related Protobuf definitions into a single file? Basically, `remote_prover.proto` would contain what is now in `proxy_status.proto` and `worker_status.proto`?

SantiagoPittella: The problem with this approach is that the client then would generate the proxy status API too. I will check if we can decide at build time if we can select an specific API.

bobbinth: It would be still be a separate service, right? (just located in the same file). If we do end up with some extra generated code, it's not a big deal, and actually adding ability to get the status via the client may not be a bad thing (as a separate PR).

SantiagoPittella: I pushed a commit that merges the proto files. Let me know what do you think, it might be too much out of scope, but the implementation was pretty much straightforward. I can move it to another PR if we want to.

proto/proto/remote_prover.proto
```diff
@@ -0,0 +1,100 @@
...
+    // The proof type supported by this proxy.
+    remote_prover.ProofType supported_proof_type = 2;
+    // The list of workers managed by this proxy.
+    repeated WorkerStatus workers = 3;
+}
+
+service StatusApi {
+    // Returns the status of the worker.
+    rpc Status(StatusRequest) returns (StatusResponse) {}
+}
```
bobbinth: Should we rename this into `WorkerStatusApi` (and rename associated messages accordingly)?

Also, nit: I'd add a section separator above this service definition.

proto/proto/remote_prover.proto
```diff
@@ -0,0 +1,100 @@
...
+// Response message containing the generated proof.
+message ProvingResponse {
+    // Serialized proof bytes.
+    // - TRANSACTION: Returns an encoded ProvenTransaction.
+    // - BATCH: Returns an encoded ProvenBatch.
+    // - BLOCK: Returns an encoded ProvenBlock.
+    bytes payload = 1;
+}
+
+service ProxyStatusApi {
```
bobbinth, line 50: nit: I would add a section separator above this line.

proto/proto/remote_prover.proto
SantiagoPittella: I'm adding the separators. Regarding the worker status api: sounds good too. I'm about to the send the changes.

PR #1006, opened 2025-06-24
ci: Fix postrm prover dirs

sergerad:


packaging/prover-proxy/postrm
```diff
@@ -4,6 +4,6 @@
 # Remove miden-prover-proxy installs
 ##############
 sudo rm -f /lib/systemd/system/miden-prover-proxy.*
-sudo rm -rf /opt/miden/miden-prover-proxy/
+sudo rm -rf /opt/miden-prover-proxy/
```
Mirko-von-Leipzig, line 7: You previously argued we shouldn't remove the operational data. I realise this PR just fixes the existing paths, but should we drop the removal altogether? Or am I misunderstanding.

sergerad: That was specifically for node and faucet which only rm the `etc/` version:
```
sudo rm -rf /etc/opt/miden-node
```

AFAIK there is no such state to worry about in prover and proxy in `/opt/` - but we could make them consistent in case there ever is?

PR #1007, opened 2025-06-24
feat: v0.9.3

Mirko-von-Leipzig:
This PR fixes a bug in our network note handling which caused the ntx builder and block-producer to crash.

I've bundled this with a version bump to v0.9.3 under the assumption we probably want to redeploy this asap. But we can remove those commits as well.

### Bug description

We currently only support network notes which are "single target". This fact is checked in multiple locations; _except_ in the block producer which only checks that its a network note. aka it happily emits "multi target" network notes to the ntx builder and store.

This causes the ntx builder to crash. Unfortunately, the offending note is then also emitted by the store on startup again, which again causes the ntx builder to crash in a never-ending loop.

### Solution

This is a rather temporary hotfix since the ntx builder is being reworked.

The block-producer now only emits single target notes. The ntx builder now filters such notes from the store to prevent crashing on the existing notes. So in a way we're both stricter and looser in our checks, ideally this would be better but a more robust implementation can be part of the rework.

I've also prevented new notes to be rejected when submitted so that the user doesn't create these notes and then wonder why they're not being fulfilled.

### DIFF
--- a/crates/utils/src/account.rs
+++ b/crates/utils/src/account.rs
@@ -50,7 +50,7 @@ impl TryFrom<NoteTag> for NetworkAccountPrefix {
     type Error = NetworkAccountError;
 
     fn try_from(tag: NoteTag) -> Result<Self, Self::Error> {
-        if tag.execution_mode() != ExecutionMode::Network && !tag.is_single_target() {
+        if tag.execution_mode() != NoteExecutionMode::Network || !tag.is_single_target() {
             return Err(NetworkAccountError::InvalidExecutionMode(tag));
         }
 


crates/utils/src/account.rs
```diff
@@ -50,7 +50,7 @@ impl TryFrom<NoteTag> for NetworkAccountPrefix {
     type Error = NetworkAccountError;
 
     fn try_from(tag: NoteTag) -> Result<Self, Self::Error> {
-        if tag.execution_mode() != NoteExecutionMode::Network || !tag.is_single_target() {
```
igamigo: Yeah, I had seen that `is_single_target()` somewhat superseded the other check but thought to leave both just in case. I was also speculating with whether the `NoteTag` refactor would make it in in that release.

igamigo (approved):
LGTM. Thanks! I think I may have introduced the newtype after taking most of the changes from #805 and did not realize to revisit the check for the notes that got emitted.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -495,9 +495,8 @@ fn get_network_notes(proven_block: &ProvenBlock) -> Vec<Note> {
     proven_block
         .output_notes()
         .filter_map(|(_idx, note)| match note {
-            OutputNote::Full(inner)
-                if inner.metadata().tag().execution_mode() == NoteExecutionMode::Network =>
-            {
+            // is_single_target is poorly named, but this is only applicable to network notes.
+            OutputNote::Full(inner) if inner.metadata().tag().is_single_target() => {
```
bobbinth, line 499: This is fine for now, but now that i look at the implementation of `is_single_target()` I see that it is not entirely correct. Specifically, a note can be single target and not be a network note (i.e., targeted to a specific private account). So, ideally, this should check both - that execution mode is `Network` and that the note is single target. But again, with the current version of `is_single_target()` this works fine.

cc @PhilippGackstatter.

crates/ntx-builder/src/builder/store.rs
```diff
@@ -122,7 +122,11 @@ impl StoreClient {
                 .map(NetworkNote::try_from)
                 .collect::<Result<Vec<_>, _>>()?;
 
-            all_notes.extend(page);
+            // The ntx builder currently only support single target notes, so filter out the others.
+            //
+            // We unfortunately let some into our testnet database so this is the easiest place to filter.
```
bobbinth: nit: I would remove this part of the comment. I think generally, it is good to filter out notes that we know we don't support, as an extra precaution.

crates/utils/src/account.rs
```diff
@@ -50,7 +50,7 @@ impl TryFrom<NoteTag> for NetworkAccountPrefix {
...
+        if !tag.is_single_target() {
```
bobbinth: As mentioned in one of the other comment this will work now, but ideally we should do both checks. Is there a reason to remove `tag.execution_mode() != NoteExecutionMode::Network`?

Mirko-von-Leipzig: Reverted this

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: In the rework I'll ensure that filtering by single target is only done by the network tx builder, and at the RPC level to warn users such notes won't be useful yet.

crates/utils/src/account.rs
```diff
@@ -50,7 +50,7 @@ impl TryFrom<NoteTag> for NetworkAccountPrefix {
     type Error = NetworkAccountError;
 
     fn try_from(tag: NoteTag) -> Result<Self, Self::Error> {
-        if tag.execution_mode() != NoteExecutionMode::Network || !tag.is_single_target() {
+        if tag.execution_mode() != ExecutionMode::Network && !tag.is_single_target() {
```
bobbinth: Wasn't "or" the correct operator here? Maybe easier to write this as:
```rust
if !(tag.execution_mode() == ExecutionMode::Network && tag.is_single_target()) {
   // return an error
}
```
Ideally, we'd have a simpler way to check this - but this would need to be done in `miden-base`. Maybe:
```rust
if !tag.is_target_network_account() {
   // return an error
}
```

Mirko-von-Leipzig: Ah thanks; you're right. Always the simple things when one is rushing.

bobbinth:
Should we merge this PR or just close it?

Mirko-von-Leipzig:
Let's close; we can always re-open if we need to hotfix release.

PR #1009, opened 2025-06-24
chore: refactor node CLI commands

### LINKED ISSUE #885: Refactor CLI commands
Some CLI commands have become complex in terms of the amount of parameters they take and how they relate to each other. The best example of this is probably the `bundled` command. We should look into how we can reduce the complexity by grouping args. 

Reference: https://github.com/0xMiden/miden-node/pull/884#discussion_r2110481832

SantiagoPittella:
closes #885

### DIFF
--- a/bin/node/src/commands/mod.rs
+++ b/bin/node/src/commands/mod.rs
@@ -1,5 +1,7 @@
 use std::time::Duration;
 
+use url::Url;
+
 pub mod block_producer;
 pub mod bundled;
 pub mod rpc;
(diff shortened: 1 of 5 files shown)


sergerad (approved):
LGTM but face-value review from me. Others will have context around intent that I don't have.

For example, did we want to add hierarchy into the flags which corresponds to the new types added? E.G. `--block.interval` becomes `--block_producer.block_interval` (or something less verbose and unsightly).

bin/node/src/commands/mod.rs
```diff
@@ -24,3 +26,77 @@ const DEFAULT_NTX_TICKER_INTERVAL: Duration = Duration::from_millis(200);
 fn duration_to_human_readable_string(duration: Duration) -> String {
     humantime::format_duration(duration).to_string()
 }
+
+/// Configuration for the Network Transaction Builder component
+#[derive(clap::Args)]
+pub struct NtxBuilderConfig {
+    /// Disable spawning the network transaction builder.
+    #[arg(long = "no-ntb", default_value_t = false)]
+    pub disabled: bool,
```
bobbinth, line 35: Should we define this as conflicting with `tx_prover_url` and `ticker_interval`? Basically, if we set `disabled`, we shouldn't set the other two.

bobbinth (approved):

bin/node/src/commands/mod.rs
SantiagoPittella: Those are not conflicting meaning that will not panic/crash , but wouldn't be used if disabled is set. I can change it to enforce, wdyt?

SantiagoPittella:
> For example, did we want to add hierarchy into the flags which corresponds to the new types added? E.G. `--block.interval` becomes `--block_producer.block_interval` (or something less verbose and unsightly).



> LGTM but face-value review from me. Others will have context around intent that I don't have.
> 
> For example, did we want to add hierarchy into the flags which corresponds to the new types added? E.G. `--block.interval` becomes `--block_producer.block_interval` (or something less verbose and unsightly).

I didn't want to change the interface, but rather a reorg of the inner code. I can change that if we want to. lmk what do you guys think cc @sergerad @bobbinth

SantiagoPittella:
> For example, did we want to add hierarchy into the flags which corresponds to the new types added? E.G. --block.interval > becomes --block_producer.block_interval (or something less verbose and unsightly).

I'm merging this one, lmk what do you think @sergerad about this, we can create another issue+PR if we decide to go for it.

bin/node/src/commands/mod.rs
bobbinth: Yeah, maybe worth enforcing this to make it clear to the user that some combinations are meaningless. At least I don't see any real downsides to doing this.

PR #1012, opened 2025-06-25
chore: Update error handling to capture source errors

### LINKED ISSUE #618: Revisit error strategy
We currently follow the `thiserror` for libs and `anyhow` for binaries rule-of-thumb. However, our crates in this repo are more a way of separating concerns and components -- and are not really intended as libraries (excepting the RPC/proto maybe).

This can be seen in our error handling where we almost never match on error variants, but instead wrap and bubble them up. One notable exception is in RPC methods where we need to differentiate between internal and user-facing errors.

We should re-evaluate our error handling strategy. At minimum we could probably merge a bunch of the various error enums e.g. `store` errors are per method, but these could probably just be a single enum.

A more extreme approach could be to default to `anyhow::Result` unless required otherwise. RPC errors would likely just be split into external/internal wrappers of anyhow, and if a component becomes interested in a specific error condition then that variant could be added.

sergerad:
WIP

Closes #618.


crates/store/src/db/mod.rs
```diff
@@ -269,7 +270,7 @@ impl Db {
         let conn = pool.get().await.map_err(DatabaseError::MissingDbConnection)?;
 
         conn.interact(apply_migrations).await.map_err(|err| {
-            DatabaseError::InteractError(format!("Migration task failed: {err}"))
+            DatabaseError::InteractError(format!("Migration task failed: {}", err.as_report()))
```
sergerad: @Mirko-von-Leipzig Do we want to handle errors like `DatabaseError::InteractError(String)` like this or modify them to be `DatabaseError::InteractError(anyhow::Error)` and do
```rust
            DatabaseError::InteractError(anyhow!("Migration task failed: {err}"))
```

sergerad: cc @PhilippGackstatter

Mirko-von-Leipzig: > @Mirko-von-Leipzig Do we want to handle errors like DatabaseError::InteractError(String) like this or modify them to be DatabaseError::InteractError(anyhow::Error) and do

Yeah; though you'll find `anyhow::Error` isn't send. So you need to use `Arc<anyhow::Error>` which somewhat breaks the convenience of using `anyhow`.

```rust
DatabaseError::InteractError(anyhow!("Migration task failed: {err}"))
```

^^ this also doesn't work as you might expect because `{err}` just uses the display of the error and misses out on the error reporting. What one has to do is either `err.context("migration task failed")`, or use the `err.as_report()`. The former would be the `anyhow` way but now you also need to `Arc` it (I suspect).

bin/proving-service/src/proxy/worker.rs
```diff
@@ -375,14 +377,14 @@ async fn create_status_client(
...
+/// Returns true if the version has major and minor versions match that of the required version.
+/// Returns false otherwise.
 ///
 /// # Errors
 /// Returns an error if either of the versions is malformed.
-fn is_valid_version(version_req: &VersionReq, version: &str) -> Result<bool, String> {
+fn is_valid_version(version_req: &VersionReq, version: &str) -> anyhow::Result<bool> {
     let received =
-        Version::parse(version).map_err(|err| format!("Invalid worker version: {err}"))?;
+        Version::parse(version).map_err(|err| anyhow!("Invalid worker version: {err}"))?;
```
Mirko-von-Leipzig:
```suggestion
        Version::parse(version).context("parsing worker version failed")?;
```

crates/block-producer/src/mempool/inflight_state/mod.rs
```diff
@@ -630,7 +631,10 @@ mod tests {
             let mut reverted = InflightState::new(BlockNumber::default(), 1, 0u32);
             for (idx, tx) in txs.iter().enumerate() {
                 reverted.add_transaction(tx).unwrap_or_else(|err| {
-                    panic!("Inserting tx #{idx} in iteration {i} should succeed: {err}")
+                    panic!(
+                        "Inserting tx #{idx} in iteration {i} should succeed: {}",
+                        err.as_report()
+                    )
```
Mirko-von-Leipzig, line 637: An alternative could be to return `anyhow::Result<()>` from the test case, and then use 

```suggestion
with_context(|| format!("failed to insert tx #{idx} in iteration {i}"))?;
```

crates/ntx-builder/src/builder/server/mod.rs
Mirko-von-Leipzig, line 1: This is being reworked in #977 so you could revert this if you wish.

No worries either way; just saying I'm ignoring this file in my review :)

crates/store/src/server/api.rs
```diff
@@ -66,7 +67,7 @@ pub fn read_account_id(
 ) -> Result<AccountId, Box<Status>> {
     id.ok_or(invalid_argument("missing account ID"))?
         .try_into()
-        .map_err(|err| invalid_argument(format!("invalid account ID: {err}")).into())
+        .map_err(|err| invalid_argument(anyhow!("invalid account ID: {err}")).into())
```
Mirko-von-Leipzig: You may need to check what this actually generates because I think `{err}` doesn't do the report style.

crates/store/src/db/mod.rs
Mirko-von-Leipzig: I feel like we should figure out a better way to do this so that we don't rely on doing this so manually. But I'm unsure what that should look like.

crates/store/src/server/api.rs
sergerad: Ah yea my reading of this was incorrect:
>If called with a single argument whose type implements std::error::Error (in addition to Debug and Display, which are always required), then that Error impl’s source is preserved as the source of the resulting anyhow::Error

Here we don't pass in a "single argument whose type implements std::error::Error".

crates/store/src/db/mod.rs
sergerad: Less manual:
```rust
pub trait ErrorReport: std::error::Error {
    fn as_wrapped_report(&self, msg: &'static str) -> String {
        format!("{msg}: {}", self.as_report())
    }
}
...
DatabaseError::InteractError(err.as_wrapped_report("Migration task failed"))
...
```

Or it returns `Error` and you need to `.to_string()` manually.

WDYT?

Mirko-von-Leipzig: I like it a lot more than what we have currently.

This does make me wonder if we shouldn't looking more into `eyre` and `snafu`. We're sort of building our own `anyhow` but where `Display` implements the report instead of `anyhow`s top layer only.

Mirko-von-Leipzig: So I guess lets go with your `as_wrapped_report` (whichever impl you care to use). And we can punt the R&D a bit.

sergerad: Yea ideally we just rely on anyhow, context, and debug. Rather than our report trait. I think there are situations where that is hard / impossible however. I will have a go at that now.

bin/proving-service/src/api/prover.rs
```diff
@@ -181,11 +182,10 @@ impl ProverApi for ProverRpcApi {
...
-fn internal_error<E: core::fmt::Debug>(err: E) -> Status {
-    Status::internal(format!("{err:?}"))
+fn internal_error<E: ErrorReport>(err: E) -> Status {
+    Status::internal(err.as_report())
 }
 
-fn invalid_argument<E: core::fmt::Debug>(err: E) -> Status {
-    Status::invalid_argument(format!("{err:?}"))
+fn invalid_argument<E: ErrorReport>(err: E) -> Status {
+    Status::invalid_argument(err.as_report())
```
sergerad: The `<E: ErrorReport>` is too strict for anyhow errors but it seems none are used with this fn now so might prefer to leave it stricter.

Otherwise can revert to `<E: Debug>` and require `as_report()` to be used.

crates/block-producer/src/mempool/inflight_state/mod.rs
sergerad: Part of me wants to leave them as is. LMK what you prefer.

crates/rpc/src/server/api.rs
```diff
@@ -229,7 +232,11 @@ impl api_server::Api for RpcService {
         let tx_verifier = TransactionVerifier::new(MIN_PROOF_SECURITY_LEVEL);
 
         tx_verifier.verify(&tx).map_err(|err| {
-            Status::invalid_argument(format!("Invalid proof for transaction {}: {err}", tx.id()))
+            Status::invalid_argument(format!(
+                "Invalid proof for transaction {}: {}",
+                tx.id(),
+                err.as_report()
+            ))
```
sergerad, line 239: Would need to make `as_report_context()` a macro to support arbitrary format like this. Simpler to leave like this and use `as_report()`.

crates/store/src/db/mod.rs
sergerad: The anyhow approach is just tangential to what error enums + as_report is achieving. So for this PR I'm just addressing the problem of losing the chain of errors when converting errors to strings for error enumerations.

Cargo.lock
```diff
@@ -2493,6 +2493,7 @@ dependencies = [
  "opentelemetry_sdk",
  "rand 0.9.1",
  "serde",
+ "thiserror 2.0.12",
```
Mirko-von-Leipzig, line 2458: So we apparently have multiple versions of `thiserror` in our dep tree?

I've created #1024 to look into this.

bin/proving-service/src/proxy/update_workers.rs
```diff
@@ -103,16 +104,16 @@ impl HttpServerApp for LoadBalancerUpdateService {
         let update_workers = match update_workers {
             Ok(workers) => workers,
             Err(err) => {
-                let error_message = format!("Failed to parse query parameters: {err}");
+                let error_message = err.as_report_context("Failed to parse query parameters");
```
Mirko-von-Leipzig: I realize you're just keeping the same error message, but while we're here could we use lowercase for these error messages?

bin/proving-service/src/api/prover.rs
Mirko-von-Leipzig: Let's leave as is for now.

crates/utils/Cargo.toml
```diff
@@ -45,3 +45,6 @@ vergen       = { "version" = "9.0", optional = true }
 vergen-gitcl = { features = ["cargo", "rustc"], optional = true, version = "1.0" }
 # Optional dependencies enabled by `testing` feature.
 tokio = { optional = true, workspace = true }
+
+[dev-dependencies]
+thiserror = { version = "2.0" }
```
Mirko-von-Leipzig:
```suggestion
thiserror = { workspace = true }
```

Mirko-von-Leipzig (approved):
Thanks! If you could lowercase the error messages then I think this is good to go

bin/proving-service/src/proxy/update_workers.rs
sergerad: @PhilippGackstatter do either of you want to make an issue to do this kind of thing across the repo? If there are any other nits to bundle into the issue around error formatting would be good too. Easy `good first issue` I would think.

sergerad: Updated this PR to do lowercase. Wasn't sure about "Caused by:" being lowercased but have done that too.

PR #1013, opened 2025-06-25
feat: remove comments from SQL commands

c410-f3r:
Fixes a `TODO`.

Reduces the amount of bytes when hashing but also breaks existing migration hashes. Not sure how to handle code upgrades in this case.

### DIFF
--- a/crates/store/src/db/migrations.rs
+++ b/crates/store/src/db/migrations.rs
@@ -113,17 +113,29 @@ fn compute_migration_hashes() -> Vec<Hash> {
         .collect()
 }
 
+// Strips new lines, whitespaces and comments.
 fn preprocess_sql(sql: &str) -> String {
-    // TODO: We can also remove all comments here (need to analyze the SQL script in order to remove
-    //       comments in string literals).
-    remove_spaces(sql)
+    sql.lines()
+        .map(|line| line.split_once("--").unwrap_or((line, "")).0)
+        .flat_map(|line| line.chars())
+        .filter(|c| !c.is_whitespace())
+        .collect()
 }
 
-fn remove_spaces(str: &str) -> String {
-    str.chars().filter(|chr| !chr.is_whitespace()).collect()
-}
+#[cfg(test)]
+mod tests {
+    use crate::db::migrations::{MIGRATIONS, preprocess_sql};
 
-#[test]
-fn migrations_validate() {
-    assert_eq!(MIGRATIONS.validate(), Ok(()));
+    #[test]
+    fn migrations_validate() {
+        assert_eq!(MIGRATIONS.validate(), Ok(()));
+    }
+
+    #[test]
+    fn preprocess_sql_strips_unwanted_chars() {
+        let whitespaces = "SELECT 1 FROM foo";
+        assert_eq!(preprocess_sql(whitespaces), "SELECT1FROMfoo");
+        let comments_and_new_lines = "SELECT 1; -- Comment\n-- Another comment\nSELECT 2";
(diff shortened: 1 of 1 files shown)


crates/store/src/db/migrations.rs
```diff
@@ -114,16 +114,46 @@ fn compute_migration_hashes() -> Vec<Hash> {
...
+                rslt.pop();
+                break;
+            }
+            if !curr.is_ascii_whitespace() {
+                rslt.push(curr.into());
+            }
+            prev = curr;
+        }
+    }
+    rslt
```
Mirko-von-Leipzig: This works but I think we can make this more legible by using iterators:

```suggestion
/// Processes a sql migration to ensure that non-syntactic diffs don't change the hash.
///
/// This is done by removing all comments and whitespace.
fn preprocess_migration_script(sql: &str) -> String {
    sql
      .lines()
      .map(|line| line.split_once("--").unwrap_or((line, "")).0)
      .flat_map(|line| line.chars())
      .filter(|c| !c.is_whitespace())
      .collect()
}
```

I would also replace `preprocess_sql` with this.

Mirko-von-Leipzig (changes requested):
Thank you for taking a stab at this - I hope you don't mind the suggested impl :)

crates/store/src/db/migrations.rs
c410-f3r: In Rust iterators are historically known for generally not yielding optimized code. For example:

- https://github.com/rust-lang/rust/issues/47461
- https://github.com/rust-lang/rust/issues/117659
- https://github.com/rust-lang/rust/issues/113691
- https://github.com/rust-lang/rust/issues/75935

[This godbolt snippet](https://godbolt.org/z/Gqa4PjTve) also states that the `Iterator` version produces way more lines of assembly.

In regards to runtime performance the `loop` version is 2x faster.

Capture d’écran du 2025-06-27 08-23-23

```rust
#[bench]
fn bench_iter(b: &mut Bencher) {
  let string = String::from_utf8(vec![0; 1024 * 1024 * 2]).unwrap();
  b.iter(|| {
    let _processed = preprocess_iter(&string);
  });
}

#[bench]
fn bench_loop(b: &mut Bencher) {
  let string = String::from_utf8(vec![0; 1024 * 1024 * 2]).unwrap();
  b.iter(|| {
    let _processed = preprocess_loop(&string);
  });
}

fn preprocess_iter(sql: &str) -> String {
  sql
    .lines()
    .map(|line| line.split_once("--").unwrap_or((line, "")).0)
    .flat_map(|line| line.chars())
    .filter(|c| !c.is_whitespace())
    .collect()
}

fn preprocess_loop(sql: &str) -> String {
  let mut rslt = String::with_capacity(32);
  for line in sql.lines() {
    let mut iter = line.as_bytes().iter().copied();
    let Some(mut prev) = iter.next() else {
      continue;
    };
    rslt.push(prev.into());
    for curr in iter {
      if prev == b'-' && curr == b'-' {
        rslt.pop();
        break;
      }
      if !curr.is_ascii_whitespace() {
        rslt.push(curr.into());
      }
      prev = curr;
    }
  }
  rslt
}
```

Mirko-von-Leipzig: I understand thanks; but I would prefer something that is easy to read.

The code isn't performance sensitive so it doesn't need to be optimized for speed, but rather for legibility and ease of maintenance.

c410-f3r: The code has been updated to use the `Iterator` version

Mirko-von-Leipzig (approved):

PR #1015, opened 2025-06-25
feat(block-producer): increase and make max txs per batch and batches per block configurable via CLI

farrellh1:
- Adds CLI arguments to configure the maximum number of transactions per batch (`--max-txs-per-batch`) and the maximum number of batches per block (`--max-batches-per-block`) for both standalone and bundled node modes.
- Updates the default values to match previous behavior (8 per batch, 8 per block), but allows users to override them at runtime.

### DIFF
--- a/bin/node/src/commands/block_producer.rs
+++ b/bin/node/src/commands/block_producer.rs
@@ -118,10 +118,7 @@ mod tests {
                 max_txs_per_batch: 8,
                 max_batches_per_block: miden_objects::MAX_BATCHES_PER_BLOCK + 1, // Invalid value
             },
-            telemetry: TelemetryConfig {
-                open_telemetry: false,
-                monitor_interval: std::time::Duration::from_secs(1),
-            },
+            enable_otel: false,
         };
         let result = cmd.handle().await;
         assert!(result.is_err());
@@ -144,10 +141,7 @@ mod tests {
                                                                            * (should fail) */
                 max_batches_per_block: 8,
             },
-            telemetry: TelemetryConfig {
-                open_telemetry: false,
-                monitor_interval: std::time::Duration::from_secs(1),
-            },
+            enable_otel: false,
         };
         let result = cmd.handle().await;
         assert!(result.is_err());


farrellh1:
Hi, just let you know this PR ready to review @Mirko-von-Leipzig

bin/node/src/commands/block_producer.rs
```diff
@@ -105,6 +117,22 @@ impl BlockProducerCommand {
         // Start system monitor.
         SystemMonitor::new(monitor_interval).run_with_supervisor();
 
+        // Runtime validation for protocol constraints
+        if max_batches_per_block >= miden_objects::MAX_BATCHES_PER_BLOCK {
+            anyhow::bail!(
+                "max-batches-per-block ({}) cannot exceed protocol limit ({})",
+                max_batches_per_block,
+                miden_objects::MAX_BATCHES_PER_BLOCK
```
Mirko-von-Leipzig: Nit, the user should know what he configured already

```suggestion
                "max-batches-per-block cannot exceed protocol limit of {}",
                miden_objects::MAX_BATCHES_PER_BLOCK
```

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ### Changes
 
+- Made the maximum transactions per batch and batches per block configurable via CLI for block-producer and bundled node modes, and increased the transaction per batch limit to 8 (was 2) to allow higher throughput (#1015).
```
Mirko-von-Leipzig: Let's put the changes on separate lines

```suggestion
- Transaction limit per batch is now configurable (default 8) (#1015).
- Batch limit per block is now configurable (default 8) (#1015).
```

Mirko-von-Leipzig (approved):

farrellh1:
i applied the suggestion @Mirko-von-Leipzig , do i need to resolve the conflict too?

bin/node/src/commands/block_producer.rs
```diff
@@ -105,6 +117,21 @@ impl BlockProducerCommand {
...
+                "max-batches-per-block cannot exceed protocol limit of {}",
+                miden_objects::MAX_BATCHES_PER_BLOCK
+            );
+        }
+        if max_txs_per_batch >= miden_objects::MAX_ACCOUNTS_PER_BATCH {
+            anyhow::bail!(
+                "max-txs-per-batch ({}) cannot exceed protocol limit ({})",
+                max_txs_per_batch,
+                miden_objects::MAX_ACCOUNTS_PER_BATCH
+            );
```
Mirko-von-Leipzig: Apologies, I didn't mark this earlier, but could you also do the same here?

farrellh1:
i think we should i have a make script for installing some dependencies especially for the linting what do you think @Mirko-von-Leipzig ?

btw i resolved the conflict

Mirko-von-Leipzig (approved):
Thank you; and sorry for the delay!

bin/node/src/commands/mod.rs
```diff
@@ -78,4 +79,12 @@ pub struct BlockProducerConfig {
...
     #[arg(long = "block-prover.url", env = ENV_BLOCK_PROVER_URL, value_name = "URL")]
     pub block_prover_url: Option<Url>,
+
+    /// The number of transactions per batch.
+    #[arg(long = "max-txs-per-batch", default_value_t = SERVER_MAX_TXS_PER_BATCH)]
+    max_txs_per_batch: usize,
+
+    /// Maximum number of batches per block.
+    #[arg(long = "max-batches-per-block", default_value_t = SERVER_MAX_BATCHES_PER_BLOCK)]
+    max_batches_per_block: usize,
```
bobbinth, line 89: Do we not want these to be set via the env variables? If we do, shouldn't these have `env` attributes?

crates/block-producer/src/lib.rs
```diff
@@ -25,10 +25,10 @@ pub use server::BlockProducer;
 pub const COMPONENT: &str = "miden-block-producer";
 
 /// The number of transactions per batch.
-const SERVER_MAX_TXS_PER_BATCH: usize = 2;
+pub const SERVER_MAX_TXS_PER_BATCH: usize = 8;
 
 /// Maximum number of batches per block.
-const SERVER_MAX_BATCHES_PER_BLOCK: usize = 4;
+pub const SERVER_MAX_BATCHES_PER_BLOCK: usize = 8;
```
bobbinth, line 31: These are now defaults (i.e., these can be overridden) - right? If so, would `DEFAULT_MAX_TXS_PER_BATCH` and `DEFAULT_MAX_BATCHES_PER_BLOCK` be better names?

bin/node/src/commands/mod.rs
Mirko-von-Leipzig: Ah yeah. I've created a follow-up issue.

crates/block-producer/src/lib.rs
Mirko-von-Leipzig: We could also remove them entirely? The default values are really more of a node cli thing imo.

bobbinth: Yep, that makes sense. Would they just move to the `miden-node` crate?

Mirko-von-Leipzig: Yeah, in fact we already have a couple there in `commands/mod.rs`.

I'll update the follow-up issue to make this more of a standard.

Issue #964, opened 2025-06-11
Add Network ID to faucet config


TomasArrachea:
Currently the explorer is hardcoded to testnet, and also are all the account addresses:
https://github.com/0xMiden/miden-node/blob/2fe9cd83195699e6d27465289821a7d30e0beed8/bin/faucet/src/main.rs#L44-L46

This should set as part of the faucet config.

sergerad:
Noting this task to move the configuration to clap/.env #958. Want to take both into account when one is started.

PR #1016, opened 2025-06-25
feat(faucet): network id configurable

TomasArrachea:
closes https://github.com/0xMiden/miden-node/issues/964.

This PR adds a new `--network` cli parameter for the `miden-faucet start` command. This can take values testnet/devnet/localhost/custom string, and it is used to generate the bech32 addresses that are shown in the frontend and to generate the explorer URL. The explorer URL will only be showed if the network is testnet, since it is not yet accesible to any other.

Code is inspired in https://github.com/0xMiden/miden-client/blob/a4c9aed2e6e4a5486fe98e2da83b3bd80f92f256/bin/miden-cli/src/config.rs#L170-L207

### DIFF
--- a/bin/faucet/src/network.rs
+++ b/bin/faucet/src/network.rs
@@ -9,34 +9,34 @@ use serde::{Deserialize, Serialize};
 /// Represents the network where the faucet is running. It is used to show the correct bech32
 /// addresses and explorer URL in the UI.
 #[derive(Debug, Clone, Deserialize, Serialize)]
-pub enum Network {
+pub enum FaucetNetwork {
     Testnet,
     Devnet,
     Localhost,
     Custom(String),
 }
 
-impl FromStr for Network {
+impl FromStr for FaucetNetwork {
     type Err = Infallible;
 
     fn from_str(s: &str) -> Result<Self, Infallible> {
         match s.to_lowercase().as_str() {
-            "devnet" => Ok(Network::Devnet),
-            "localhost" => Ok(Network::Localhost),
-            "testnet" => Ok(Network::Testnet),
-            custom => Ok(Network::Custom(custom.to_string())),
+            "devnet" => Ok(FaucetNetwork::Devnet),
+            "localhost" => Ok(FaucetNetwork::Localhost),
+            "testnet" => Ok(FaucetNetwork::Testnet),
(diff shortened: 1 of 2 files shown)


bin/faucet/src/network.rs
```diff
@@ -0,0 +1,51 @@
...
+            Network::Testnet => NetworkId::Testnet,
+            Network::Devnet => NetworkId::Devnet,
+            Network::Localhost => NetworkId::new("mlcl")?,
+            Network::Custom(s) => NetworkId::new(s)?,
+        })
+    }
+
+    /// Returns the explorer URL for the given network ID.
+    /// Currently only testnet explorer is available.
+    pub fn explorer_url(network_id: NetworkId) -> Option<String> {
```
Mirko-von-Leipzig: Nit: could return `Option<'static &str>`

Mirko-von-Leipzig: Also feels a bit strange to have this be a static method for another type. But I totally get that doing the multiple unwraps is probably less convenient.

One alternative is to have a new `ExplorerUrl(&'static str)` instead which has the method instead. Unsure if better though.

Mirko-von-Leipzig (approved):

bin/faucet/src/network.rs
TomasArrachea: I added the ExplorerUrl type wrapper, I think it's better now.

bin/faucet/src/network.rs
```diff
@@ -0,0 +1,57 @@
...
+use miden_objects::{NetworkIdError, account::NetworkId};
+use serde::{Deserialize, Serialize};
+
+// NETWORK
+// ================================================================================================
+
+/// Represents the network where the faucet is running. It is used to show the correct bech32
+/// addresses and explorer URL in the UI.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub enum Network {
```
igamigo: nit: I think the `Network` name is a bit confusing, maybe I'd name this `FaucetNetwork` to emphasize the functionality more

igamigo: Another way could be to make the CLI parameter just a `NetworkId` and have a custom function to parse it directly (bypassing the `Network` enum).

igamigo (approved):
LGTM! I left a small comment but very minor, feel free to disregard.

PR #1019, opened 2025-06-26
chore(node): remove system metrics

### LINKED ISSUE #1008: Remove system metrics
The system metrics monitoring currently doesn't work on our production instances. The logs indicate its likely the underlying stats gatherer crate which doesn't find the expected files to monitor with.

We could spend effort on figuring this out but it is a painful thing to make robust in general. `honeycomb` also has integration for `aws` instances so we don't actually need to do this in-process ourselves.

I recommend we just remove it.

SantiagoPittella:
closes #1008


Mirko-von-Leipzig (approved):
@sergerad just want to double-check this is the correct course of action?

SantiagoPittella:
I removed the `TelemetryConfig` struct from the command and replced them with the opentelemetry field, that is just a boolean that determines if otel is enabled or not.

SantiagoPittella:
Rust 1.88 just released, with new linting rules 😄

bin/node/src/commands/block_producer.rs
```diff
@@ -28,8 +25,12 @@ pub enum BlockProducerCommand {
...
         block_producer: BlockProducerConfig,
 
-        #[command(flatten)]
-        telemetry: TelemetryConfig,
+        /// Enables the exporting of traces for OpenTelemetry.
+        ///
+        /// This can be further configured using environment variables as defined in the official
+        /// OpenTelemetry documentation. See our operator manual for further details.
+        #[arg(long = "enable-otel", default_value_t = false, env = ENV_ENABLE_OTEL, value_name = "BOOL")]
+        open_telemetry: bool,
```
bobbinth: nit: should we rename this into `enable_otel` to be consistent with the arg naming?

bin/node/src/commands/bundled.rs
```diff
@@ -51,8 +49,12 @@ pub enum BundledCommand {
...
         ntx_builder: NtxBuilderConfig,
 
-        #[command(flatten)]
-        telemetry: TelemetryConfig,
+        /// Enables the exporting of traces for OpenTelemetry.
+        ///
+        /// This can be further configured using environment variables as defined in the official
+        /// OpenTelemetry documentation. See our operator manual for further details.
+        #[arg(long = "enable-otel", default_value_t = false, env = ENV_ENABLE_OTEL, value_name = "BOOL")]
+        open_telemetry: bool,
```
bobbinth: Same comment as above.

bin/node/src/commands/rpc.rs
```diff
@@ -23,8 +23,12 @@ pub enum RpcCommand {
...
         block_producer_url: Option<Url>,
 
-        #[command(flatten)]
-        telemetry: TelemetryConfig,
+        /// Enables the exporting of traces for OpenTelemetry.
+        ///
+        /// This can be further configured using environment variables as defined in the official
+        /// OpenTelemetry documentation. See our operator manual for further details.
+        #[arg(long = "enable-otel", default_value_t = false, env = ENV_ENABLE_OTEL, value_name = "BOOL")]
+        open_telemetry: bool,
```
bobbinth: Same comment as above.

bin/node/src/commands/store.rs
```diff
@@ -52,8 +52,12 @@ pub enum StoreCommand {
...
         data_directory: PathBuf,
 
-        #[command(flatten)]
-        telemetry: TelemetryConfig,
+        /// Enables the exporting of traces for OpenTelemetry.
+        ///
+        /// This can be further configured using environment variables as defined in the official
+        /// OpenTelemetry documentation. See our operator manual for further details.
+        #[arg(long = "enable-otel", default_value_t = false, env = ENV_ENABLE_OTEL, value_name = "BOOL")]
+        open_telemetry: bool,
```
bobbinth: Same comment as above.

bobbinth (approved):
Looks good! Thank you! I left an optional comment inline.

Regarding the lint errors - if it is just a few, we could fix them in this PR - but if not, we could do it in a separate PR.

bin/node/src/commands/bundled.rs
SantiagoPittella: Sure, I renamed this and the following arguments 👍🏼 Additionally, fixed the lint error on another PR that is merged already.

PR #1020, opened 2025-06-26
fix: `cargo update` and fix build

tomyrd:
Minor fix after `cargo update`. It's needed because we are getting the error over on the `miden-client` when updating the version of `miden-base`


Cargo.lock
Mirko-von-Leipzig, line 1: Did you run a full `cargo update` or did you only `cargo update -p miden-lib`?

tomyrd: I ran a full `cargo update`. Should I rollback and only update `miden-lib`?

Mirko-von-Leipzig: I typically prefer being more specific with it, but I think its fine since we're changing many things at the moment

PR #1021, opened 2025-06-26
feat(store): sync network tx data via gRPC

Mirko-von-Leipzig:
This PR adds new gRPC methods to the store to support the ntx rework in #977.

Since #977 is already far too large, I've split it out as a smaller precursor PR. To this end, this PR leaves the current ntx-related store endpoints in place since those will be used until the full rework is complete.

### Sync summary

The idea is to have the ntx builder sync to the chain tip with all data related to the network accounts and notes. From there on the ntx builder can use the mempool events to stay in sync. In the not-to-distant future the ntx builder should also persist this data itself so that the resync time is shorter.

To this end, the ntx builder needs state deltas in order to increment local state to the store's state. There are considerations to optimise the general sync such that larger gaps can be skipped more efficiently. That doesn't really apply here since the ntx builder is intended to run permanently and should therefore always be _close_ to in sync.

This implies state needs to be transmitted as an event log of sorts. To keep things simple, this transmits every network account delta, network note created, and network nullifier within a chunk.

- [x] network account updates
- [ ] network notes created
- [ ] network note nullifiers


proto/proto/store.proto
```diff
@@ -104,4 +106,48 @@ service NtxBuilder {
...
+// Represents a block range with a start but no end.
+message InclusiveUnboundedRange {
+    // Start of the range, inclusive.
+    fixed32 start = 1;
+}
+
+// All network transaction related updates that took place between the request's starting block
+// and some later block chosen by the server.
+message NetworkUpdates {
+    // The last block in the ra
```
igamigo, line 130: nit: Seems like the sentence ends unexpectedly
```suggestion
    // The last block number of the range. 
```

proto/proto/store.proto
```diff
@@ -104,4 +106,48 @@ service NtxBuilder {
...
+    // Complete set of network account updates in the inclusive range formed by
+    // the request's starting block and `end_inclusive`.
+    //
+    // The data is encoded using [winter_utils::Serializable] implementation for
+    // [miden_objects::account::delta::AccountUpdateDetails].
+    //
+    // Note that the type allows for `Private` but this isn't applicable to network
+    // accounts which are public by definition. Servers should avoid sending `Private`
+    // since its nonsensical, however clients should filter these out.
+    repeated bytes account_updates = 2;
```
igamigo, line 142: Probably could use its own message definition, though not too important

crates/store/src/db/sql/mod.rs
```diff
@@ -1061,6 +1145,45 @@ pub fn unconsumed_network_notes(
     Ok((notes, page))
 }
 
+/// Returns a paginated batch of network notes that have not yet been consumed.
+///
+/// # Returns
+///
+/// A set of unconsumed network notes with maximum length of `size` and the page to get
```
igamigo, line 1152: I've realized now that we can't confirm these have been unconsumed until you crosscheck against the nullifier set

crates/store/src/server/ntx_builder.rs
```diff
@@ -151,4 +154,38 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
...
+        err
+    )]
+    async fn get_network_updates(
+        &self,
+        request: Request<InclusiveUnboundedRange>,
+    ) -> Result<Response<ProtoNetworkUpdates>, Status> {
+        let start: BlockNumber = request.into_inner().start.into();
+
+        let (end, updates, notes, nullifiers) = self.state.network_account_updates(start).await?;
+        let account_updates = updates.iter().map(AccountUpdateDetails::to_bytes).collect();
```
igamigo, line 172: Worth `debug_assert`ing that these are network accounts?

igamigo (approved):
Looks good in terms of the code itself, but as I do wonder about the utility of keeping network account updates on a separate table and feeding them to the NTB. As you mention, given that in the (fairly) short term we should be switching to `PartialAccount` as the main way of providing transaction inputs, and enabling lazily loading parts of the account, it seems its utility could be void against just asking for details to the store. I also think this approach could be simpler in general:

- When executing a network transaction, check if account is in cache
- If it is, use the cached account. If it's not, go to the store and fetch the basic `PartialAccount`
- Optimistically apply new delta and store in cache. If you find out the transaction gets reverted, evict from cache. This can later be optimized to caching multiple states of the same account

Though maybe your approach could be considered simple because it's nicely consistent with the other ways to start up the service (aka reach once for everything). The other nice thing about lazily reaching for the store is that it minimizes duplicate data which could also be nice considering network accounts could potentially have more state updates than the average account.

crates/store/src/db/migrations/001-init.sql
Mirko-von-Leipzig, line 1: >  As you mention, given that in the (fairly) short term we should be switching to PartialAccount as the main way of providing transaction inputs, and enabling lazily loading parts of the account, it seems its utility could be void against just asking for details to the store. I also think this approach could be simpler in general:
>
> - When executing a network transaction, check if account is in cache
> - If it is, use the cached account. If it's not, go to the store and fetch the basic PartialAccount
> - Optimistically apply new delta and store in cache. If you find out the transaction gets reverted, evict from cache. This can later be optimized to caching multiple states of the same account

The calculus here is fairly variable I think, even if we switch to lazy-loading. If the goal is to minimize redundant storage then we should always load from the `store`. If the goal is to minimize `store` IO then we should always duplicate the data on the `ntx-builder` as well. And then everything else is some middle ground.

In my mind we want to minimize store IO; and lazy-loading actually makes this worse _if_ we utilize the `store` for this. Storage is cheap and network IO should prefers large chunks over small pieces.

The other thing to consider is that once ntx is sync'd it _never_ has to talk to the store ever again. To my mind this is a massive saving in store IO (and simplifies design); but maybe my perspective is wrong here. As in, if you add up sending all network account data early on in the network history (plus some extra here and there to allow for crashes), then that is much better than lazy-loading over all of time. At the cost of some duplicate storage.

crates/store/src/db/sql/mod.rs
Mirko-von-Leipzig: You could amend the gRPC method to state that it returns all network notes that are unconsumed at the time of the request.

The client would still be required to sync and cross-check over all nullifiers received at the end. So no change to the required client impl, but we get to skip already consumed network notes.

This is probably worth doing since over time this would be _most_ network notes. This does have some minor implications e.g. this can't be used to find all network notes ever. But that's probably fine.

But initially I thought just KISS and optimize this later. Easier to reason that you're 100% in-sync at the end of this block range than you're partially in-sync but also ahead in unknown nullifiers. maybe.

Mirko-von-Leipzig: I should fix the comments though :)

crates/store/src/db/migrations/001-init.sql
Mirko-von-Leipzig: That said, at this point this is mostly a philosophical problem. So we could minimize the delta and choose to lazy load accounts on the ntx builder side.

This may even be safer since we don't have ntx persistence yet, so resets would be expensive IO-wise. And a nice property of lazy-loading is that it makes this stateless. Which is a pretty nice thing to have.

crates/store/src/db/sql/mod.rs
```diff
@@ -739,6 +786,43 @@ pub fn select_nullifiers_by_prefix(
...
+
+    let mut stmt = transaction.prepare_cached(
+        "SELECT
+            nullifier
+        FROM
+            nullifiers
+        LEFT JOIN notes ON notes.nullifier = nullifiers.nullifier
+        WHERE
+            nullifiers.block_num BETWEEN ?1 AND ?2 AND
+            notes.execution_mode = 0
```
bobbinth, line 812: Probably fine for now, but these tables will get big and performing a join may get expensive. An alternative would be to keep something like `is_network` flag in the `nullifiers` table.

crates/store/src/db/sql/mod.rs
```diff
@@ -1061,6 +1145,45 @@ pub fn unconsumed_network_notes(
     Ok((notes, page))
 }
 
+/// Returns a paginated batch of network notes that have not yet been consumed.
```
bobbinth, line 1148: From what I can tell, this method returns all network notes in the given range (not just the unconsumed ones), right?

Generally, I think it would be good to send only the unconsumed notes (and if we do this, we should be able to avoid sending nullifiers as well). For now, this would require an expensive join.

In the longer run, we probably should have a `consumed` flag in the `notes` table, and periodically prune all network and public consumed notes. Let's create an issue for this.

crates/store/src/state.rs
```diff
@@ -964,6 +964,35 @@ impl State {
...
+        DatabaseError,
+    > {
+        let chain_tip = self.latest_block_num().await;
+        if start > chain_tip {
+            return Err(DatabaseError::BlockOutOfRange { requested: start, chain_tip });
+        }
+
+        // Cap the block range at 1_000 or the chain tip.
+        let stop = (start + 1_000).min(chain_tip);
+        let range = start..=stop;
```
bobbinth, line 990: 1000 blocks is probably quite aggressive. I think gRPC responses have some relatively small size limit (maybe 10MB?). If so, 1000 blocks implies 10KB of updates per block, on average, which could be just a single account update or a few notes. I'd reduce this to 100 or maybe even 10.

crates/store/src/db/migrations/001-init.sql
```diff
@@ -124,6 +124,16 @@ CREATE TABLE account_non_fungible_asset_updates (
...
+CREATE TABLE network_account_updates (
+    block_num   INTEGER NOT NULL,
+    account_id  BLOB NOT NULL,
+    details     BLOB NOT NULL,
+
+    PRIMARY KEY (account_id, block_num),
+    FOREIGN KEY (account_id) REFERENCES accounts(account_id),
+    FOREIGN KEY (block_num) REFERENCES block_headers(block_num)
+) STRICT, WITHOUT ROWID;
+
```
bobbinth, line 136: Is the intent for this table to (eventually) be pruned periodically?

proto/proto/store.proto
```diff
@@ -104,4 +106,48 @@ service NtxBuilder {
...
+
+    // All network notes that were created in the block range.
+    //
+    // A note and its nullifier may be included in the same response.
+    repeated note.NetworkNote notes = 3;
+
+    // All network note nullifiers that were produced in the block range.
+    //
+    // A note and its nullifier may be included in the same response.
+    repeated digest.Digest nullifiers = 4;
```
bobbinth, line 152: If we can send only unconsumed notes, we don't really need to send `nullifiers` too, right?

crates/store/src/db/migrations/001-init.sql
bobbinth: I think I largely agree with the idea of reducing store IO at the expense of using extra storage. But a few comments (assuming I understood everything correctly, of course):

- Without having a separate persistence layer for the NTB, every server restart will require downloading all the data from genesis from the store. This may quite negatively affect node restart times (though, maybe not too bad for now).
- As mentioned in one of the comments, it will be tricky to figure out the range of blocks to sync over per request because of gRPC size limitations. A safe bet for now would probably be around 10 - but not sure how this would affect sync times.
- Keeping all network account deltas in the store is probably undesirable in medium/long-term perspective. Ideally, we'd land on some design where, the store does not need to keep the deltas for more than $n$ blocks for some small values of $n$ (i.e., less than 1000). This means that we'll probably need some separate API to sync NTB with the store if it falls behind by more than $n$ blocks for some reason.

So, while I think this approach is good, I wonder if now is the right time to implement it. A potentially simpler approach could be:

- On startup, perform sync it two steps:
  - First sync unconsumed network notes (this would require a new endpoint similar to the one you've added in this PR).
  - Then, use the notes to figure out a set of network accounts they target, and then get all these accounts from the store using our existing API (i.e., `GetAccountDetails`).
- If a note comes in targeting an account which NTB is unaware of, use `GetAccountDetails` endpoint to get it from the store.
- For now, all the data will be kept in NTB's memory.

Then, we can make incremental improvements to this structure:
- We can add a persistence layer so that NTB doesn't need to re-sync from genesis on startup.
- Once we refactor account sync API, NTB could use it to stay up to date with all network accounts as well.

This is just a suggestion, and it is possible that I missed some important details - so, if you think it simplifies things, the approach in this PR is fine too.

proto/proto/store.proto
Mirko-von-Leipzig: No unfortunately not.

The intention was for this to be a live sync - as in, it should be possible for the ntx builder to resync while the network is live.

We can avoid sending notes that are already consumed, but there is no way for the server to know which ones the client previously sync'd.

As an example, 

```
1. client syncs up to block T which contain note N
2. store receives block T+1 with nullifier N
3. client syncs block T+1 and therefore needs to know about nullifier N+1
```

crates/store/src/db/migrations/001-init.sql
Mirko-von-Leipzig: I think I'm going to remove this entirely and simply have the `ntx builder` fetch the account state from the store. We can then amend this with caching etc but that will require quite a bit more care.

proto/proto/store.proto
bobbinth: Ah yes, for live syncing we need them. But do we need live syncing right now? AFAICT, we need it if block producer and NTB get out of sync, right? And currently, I think if NTB crushes, block producer will crush as well. If that's the case, the should not get out of sync.

Mirko-von-Leipzig: We can keep things as they are for now if you prefer. I think originally I just added this because it made sense _if_ we were syncing network accounts as well. So if we're not doing the latter, then we can try drop this.

There is the small race condition though, because the mempool stream needs to be opened by the ntx builder, where the block-producer might have created a block before that happens, causing a "desync".

I can try make that part of the general startup requirement somehow; will see if it works out.

Mirko-von-Leipzig:
@bobbinth @igamigo I'm going to close this PR, I don't think its strictly necessary anymore.

- Account sync'ing is probably not worth it.
- Live sync isn't strictly required yet.

Issue #993, opened 2025-06-19
gRPC: refactor schemas


Mirko-von-Leipzig:
Our gRPC schema collection has become somewhat larger now and its time to relook at how we organise things.

The current flat file structure makes the actual service entry points quite difficult to find. I suggest we move the other auxilliary files into a folder, keeping only the service files at the top-level.

I also suggest we look into removing the request/response files, and move single use messages into the appropriate service file. As in `message StoreXxxRequest` could just become `message Xxx` in the store's service file.

Common types should remain in appropriately named files.

In a way, this would emulate the rust module system. e.g. start with a `types.proto` and once that grows too large, split into `primitives.proto`, `transaction.proto` etc.

-----------

The above are just some suggestions; feel free to try find a better layout/design.

PR #1022, opened 2025-06-26
chore(proto): rename note message

### LINKED ISSUE #868: Refactor Note-related protobuf messages
> I think ideally the `Note` mesage would be renamed to something like `CommittedNote` and then this could become a generic `Note` message.

_Originally posted by @igamigo in https://github.com/0xMiden/miden-node/pull/833#discussion_r2102847986_

Currently, the `Note` message contains information related to the proof of the note's inclusion in a block. This clashes a bit with the domain `Note` definition. Additionally, there are now requirements for communicating just the notes, without any extra information (specifically for network transactions). 

To refactor this, we could:

- Rename `NetworkNote` to `Note` (and remove the assumption that it's only used for network notes) 
- Rename the current `Note` message to something like `CommittedNote`, because the current definition involves the inclusion proof.
  - It feels like we could reuse `NoteInclusionInBlockProof` here and make this message something like the following:
  ```protobuf
  message CommittedNote {
     NoteInclusionInBlockProof inclusion_proof = 1;
     optional Note note_details = 2; // null for private notes
  } 
  ```
    Though this would duplicate the note ID in the payload (it would be in both `inclusion_proof` and `note_details`)

SantiagoPittella:
closes #868

@igamigo I didn't change anything of the inner struct of the message since I plan to do it in #993 . Let me know what do you think, if I should change it here anyways.

Also, should I make any other change in NTB or block-producer?


igamigo:
> I didn't change anything of the inner struct of the message since I plan to do it in https://github.com/0xMiden/miden-node/issues/993 . Let me know what do you think, if I should change it here anyways.

I think I would add those changes here because #993 seems more related to file structure and moving things around rather than refactoring internal messages (which was more the intent of #868), but I wouldn't mind it at all if you preferred otherwise.

> Also, should I make any other change in NTB or block-producer?

Not that I know of. There should be no functionality affected other than conversions once the remaining points of #868 are implemented (though not 100% sure with the NTB refactors @Mirko-von-Leipzig is working on).

proto/proto/note.proto
```diff
@@ -28,8 +28,10 @@ message NoteMetadata {
     fixed64 aux = 5;
 }
 
-// Represents a note.
-message Note {
+// Represents a committed note.
+//
+// A committed note is a note that has been included in a block.
+message CommittedNote {
```
bobbinth, line 34: I would replace `block_num`, `note_index`, `note_id`, and `merkle_path` with `NoteInclusionInBlockProof`.

We could also try to replace `metadata` and `details` with the `Note`, but this may be more tricky as `details` are optional here and mandatory there.

SantiagoPittella: I implemented this changes, and made `details` always optional so that we can use the message. I'm not quite sure if this last thing is good, wdyt?

Mirko-von-Leipzig: I would prefer stricter typing and it doesn't seem like too much redundancy given its just two fields.

Or put differently, I think WET (write everything twice) is a better starting point than DRY.

Mirko-von-Leipzig: Something to maybe consider is how this differs from a `NoteSyncRecord`?

I imagine we technically only have notes and committed notes. And if we want less types then we can collapse them all into a singular type with optional fields.

Mirko-von-Leipzig (approved):
I'm somewhat undecided on which route to take regarding more or less types. But I think either many or as close to one as we can get.

proto/proto/note.proto
SantiagoPittella: I agree, it might we might be losing more type safety than needed. In the other hand, it really helps to reduce the proto messages.

proto/proto/note.proto
```diff
@@ -28,37 +28,26 @@ message NoteMetadata {
...
+    // The note's inclusion proof and data.
+    NoteInclusionInBlockProof inclusion_proof = 1;
 
-    // Serialized details of the public note (empty for private notes).
-    optional bytes details = 6;
+    // The note.
+    Note note = 2;
```
bobbinth: nit: I would order the message like so:
```proto
message CommittedNote {
    // Either private, public, or network note.
    Note note = 1;

    // The data needed to prove that the note is present in the chain.
    NoteInclusionInBlockProof inclusion_proof = 2.
}
```

proto/proto/note.proto
bobbinth: I would prefer more type safety here. So, basically:

- Let's keep `Note` and `CommittedNote` as they are now.
- Let's add `NetworkNote` which would be like `Note` but with `details` being mandatory.

I would keep `NoteSyncRecord` as is for now. We could still consider removing but I'd do it in a separate issue/PR.

bobbinth (commented):
Thank you! Looks good. I left a couple more comments inline.

proto/proto/note.proto
bobbinth: I think this comment still needs to be addressed.

SantiagoPittella: I already re-ordered as you mention the message's inner fields 🤔 Though i will also update the comments with the ones that you mentioned.

crates/proto/src/domain/note.rs
```diff
@@ -158,7 +167,7 @@ impl TryFrom<proto::NetworkNote> for NetworkNote {
         let (assets, recipient) = details.into_parts();
         let metadata: NoteMetadata = proto_note
             .metadata
-            .ok_or_else(|| proto::NetworkNote::missing_field(stringify!(metadata)))?
+            .ok_or_else(|| proto::Note::missing_field(stringify!(metadata)))?
```
bobbinth: Why did we need to change it here from `NetworkNote` to `Note`?

bin/stress-test/src/store/mod.rs
```diff
@@ -195,7 +195,7 @@ pub async fn bench_check_nullifiers_by_prefix(
                 .iter()
                 .filter_map(|n| {
                     // private notes are filtered out because `n.details` is None
-                    Note::read_from_bytes(n.details.as_ref()?)
+                    Note::read_from_bytes(n.note.as_ref().unwrap().details.as_ref()?)
                         .ok()
                         .map(|n| u32::from(n.nullifier().prefix()))
```
bobbinth: Is this correct? Why are we reading `Note` from the `details` field of `CommittedNote::Note`? If we want to get `CommittedNote` into a `Note` - could we just convert `CommittedNote::Note` directly (e.g., maybe there should be `TryFrom` in the `domain` namespace somewhere?

Also, it is not clear where private notes are filtered out. Should we explicitly check this somewhere?

SantiagoPittella: Hmm, I think that the previously called method (`get_notes_by_id`) should now return a `Note` instead of `CommittedNote`, right? In that case we wouldn't have the note inclusion proof, and the Note is going to be able to be cast form the proto type.

Regarding the private notes filter, the `?` operator is filtering them out because it is returning None, and since it is inside a `filter_map` it gets discarded.

igamigo: > Is this correct? Why are we reading `Note` from the `details` field of `CommittedNote::Note`?

I think this is probably a leftover from an older version. `details` had the full `Note` at some point (was changed in [PR 790](https://github.com/0xMiden/miden-node/pull/790/files#r2047755521) but probably got never updated here).

> Hmm, I think that the previously called method (`get_notes_by_id`) should now return a `Note` instead of `CommittedNote`, right?

I think it should still probably return inclusion proof information

SantiagoPittella: Ran the nullifiers benchmark and looks like it is working now:

```
Average request latency: 958.534µs
P95 request latency: 1.539834ms
Average nullifiers per response: 11.3151
```

bin/stress-test/src/store/mod.rs
```diff
@@ -195,9 +195,16 @@ pub async fn bench_check_nullifiers_by_prefix(
...
+                    let inner_note = n.note.as_ref()?;
+                    let details =
+                        NoteDetails::read_from_bytes(inner_note.details.as_ref()?).unwrap();
+                    let (assets, recipient) = details.into_parts();
+                    let note = Note::new(
+                        assets,
+                        NoteMetadata::try_from(inner_note.clone().metadata?).ok()?,
+                        recipient,
+                    );
+                    Some(u32::from(note.nullifier().prefix()))
```
bobbinth: We don't need to construct the full note to get the `Nullifier` - right? (we should be able to get it from `NoteDetails`). We could write it something like:
```rust
let detail_bytes = n.note.as_ref()?.details.as_ref()?;
let details = NoteDetails::read_from_bytes(detail_bytes).unwrap();
Some(u32::from(details.nullifier().prefix()))
```

PR #1025, opened 2025-06-27
store: minor refactor to reduce boilerplate 

drahnr:
Carve out from #928 

Introduces two intermediate functions `framed` and `framed_without_transaction`, which take a closure each, to either operate on a `&mut Transaction<'_>` or `&mut Connection` respectively. It puts all error handling of the sync/async boundary into those functions rather than have it at each wrapping of a `sql::*` function.

Also uses `NullifierInfo` more, rather than using `(Nullifier, Blocknumber)`, but does NOT expose it to the user in this PR, yet!

### DIFF
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -295,7 +295,7 @@ impl Db {
         E: From<rusqlite::Error>,
         E: std::error::Error + Send + Sync + 'static,
     {
-        let conn = self.pool.get().await.unwrap(); // FIXME XXX TOODO
+        let conn = self.pool.get().await.map_err(DatabaseError::MissingDbConnection)?;
 
         conn.interact(move |conn| {
             let r = query(conn)?;
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -9,7 +9,6 @@ use miden_node_proto::{
     domain::account::{AccountInfo, AccountSummary},
     generated::note as proto,
 };
-use miden_node_utils::ErrorReport;
 use miden_objects::{
     Word,
     account::{AccountDelta, AccountId},
@@ -28,8 +27,10 @@ use tracing::{info, info_span, instrument};
 use crate::{
     COMPONENT,
     db::{
+        connection::Connection,
         migrations::apply_migrations,
         pool_manager::{Pool, SqlitePoolManager},
+        transaction::Transaction,
     },
     errors::{DatabaseError, DatabaseSetupError, NoteSyncError, StateSyncError},
     genesis::GenesisBlock,
@@ -160,21 +161,16 @@ impl NoteRecord {
     }
 }
 
-impl From<NoteRecord> for proto::CommittedNote {
+impl From<NoteRecord> for proto::Note {
     fn from(note: NoteRecord) -> Self {
-        let inclusion_proof = Some(proto::NoteInclusionInBlockProof {
-            note_id: Some(note.note_id.into()),
+        Self {
             block_num: note.block_num.as_u32(),
-            note_index_in_block: note.note_index.leaf_index_value().into(),
-            merkle_path: Some(Into::into(&note.merkle_path)),
-        });
-
-        let note = Some(proto::Note {
+            note_index: note.note_index.leaf_index_value().into(),
+            note_id: Some(note.note_id.into()),
             metadata: Some(note.metadata.into()),
-            details: note.details.map(|details| details.to_bytes()),
-        });
-
-        Self { inclusion_proof, note }
+            merkle_path: Some(Into::into(&note.merkle_path)),
+            details: note.details.as_ref().map(Serializable::to_bytes),
+        }
     }
 }
 
@@ -260,6 +256,30 @@ impl Db {
         db_tx.commit().context("failed to commit database transaction")
     }
 
+    /// Avoid repeated boilerplate, frame the query in a transaction
+    pub(crate) async fn framed<R, E, Q, M>(&self, msg: M, query: Q) -> std::result::Result<R, E>
+    where
+        Q: Send
+            + for<'a, 't> FnOnce(&'a mut Transaction<'t>) -> std::result::Result<R, E>
+            + 'static,
+        R: Send + 'static,
+        M: Send + ToString,
+        E: From<DatabaseError>,
+        E: From<rusqlite::Error>,
+        E: std::error::Error + Send + Sync + 'static,
+    {
+        let conn = self.pool.get().await.unwrap(); // FIXME XXX TOODO
+
+        conn.interact(|conn| {
+            let mut db_tx = conn.transaction()?;
+            let r = query(&mut db_tx)?;
+            db_tx.commit()?;
+            Ok(r)
+        })
+        .await
+        .map_err(|err| E::from(DatabaseError::interact(&msg.to_string(), &err)))?
+    }
+
     /// Open a connection to the DB and apply any pending migrations.
     #[instrument(target = COMPONENT, skip_all)]
     pub async fn load(database_filepath: PathBuf) -> Result<Self, DatabaseSetupError> {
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - [BREAKING] Consolidate default account filepath for bundled bootstrap and faucet start commands to `account.mac` (#976).
 - [BREAKING] Remove default value account filepath for faucet commands and rename --output-path to --output (#976).
 - Exposed `miden-proving-service` as a library (#956).
+- Refactor store aiming at boilerplate reduction (#1025).
```
Mirko-von-Leipzig: This isn't really a user facing change so I'd remove it.

```suggestion
```

crates/store/src/state.rs
```diff
@@ -170,8 +170,13 @@ impl State {
             .await?
             .ok_or(ApplyBlockError::DbBlockHeaderEmpty)?;
 
-        if block_num != prev_block.block_num() + 1 {
-            return Err(InvalidBlockError::NewBlockInvalidBlockNum.into());
+        let expected_block_num = prev_block.block_num() + 1;
```
Mirko-von-Leipzig: I believe there is now actually a `fn child` (or something like that).

```suggestion
        let expected_block_num = prev_block.block_num().child();
```

crates/store/src/db/mod.rs
```diff
@@ -254,6 +262,53 @@ impl Db {
...
+    where
+        Q: Send
+            + for<'a, 't> FnOnce(&'a mut Transaction<'t>) -> std::result::Result<R, E>
+            + 'static,
+        R: Send + 'static,
+        M: Send + ToString,
+        E: From<DatabaseError>,
+        E: std::error::Error + Send + Sync + 'static,
+    {
+        let conn = self.pool.get().await.unwrap(); // FIXME XXX TOODO
```
Mirko-von-Leipzig: Dangling todo

crates/store/src/db/mod.rs
```diff
@@ -254,6 +262,53 @@ impl Db {
...
+    ) -> std::result::Result<R, E>
+    where
+        Q: Send + FnOnce(&mut Connection) -> std::result::Result<R, E> + 'static,
+        R: Send + 'static,
+        M: Send + ToString,
+        E: From<DatabaseError>,
+        E: From<rusqlite::Error>,
+        E: std::error::Error + Send + Sync + 'static,
+    {
+        let conn = self.pool.get().await.unwrap(); // FIXME XXX TOODO
```
Mirko-von-Leipzig: Dangling todo

crates/store/src/db/mod.rs
```diff
@@ -266,29 +321,21 @@ impl Db {
...
-        let conn = pool.get().await.map_err(DatabaseError::MissingDbConnection)?;
+        let me = Db { pool };
+        me.framed_without_transaction("migrations", apply_migrations).await?;
 
-        conn.interact(apply_migrations).await.map_err(|err| {
-            DatabaseError::InteractError(format!("Migration task failed: {err}"))
-        })??;
-
-        Ok(Db { pool })
+        // TODO rationalize magic numbers, and make them
```
Mirko-von-Leipzig: Dangling todo or is this future work?

crates/store/src/db/mod.rs
```diff
@@ -254,6 +262,53 @@ impl Db {
...
+            let r = query(&mut db_tx)?;
+            db_tx.commit().map_err(DatabaseError::SqliteError)?;
+            Ok(r)
+        })
+        .await
+        .map_err(|err| E::from(DatabaseError::interact(&msg.to_string(), &err)))?
+    }
+
+    /// Avoid repeated boilerplate, frame the query in a transaction
+    pub(crate) async fn framed_without_transaction<R, E, Q, M>(
```
Mirko-von-Leipzig: I know this replaces existing code; but maybe we should force things to always operate on a transaction instead? The underlying `sqlite` essentially opens and closes a transactions so its not really any different to the transactional case iiuc.

Mirko-von-Leipzig: aka I'm suggesting we remove this function entirely.

crates/store/src/db/mod.rs
```diff
@@ -254,6 +262,53 @@ impl Db {
         db_tx.commit().context("failed to commit database transaction")
     }
 
+    /// Avoid repeated boilerplate, frame the query in a transaction
+    pub(crate) async fn framed<R, E, Q, M>(&self, msg: M, query: Q) -> std::result::Result<R, E>
```
Mirko-von-Leipzig: Naming is hard but this did confuse me a bit. Is `framed` a known term for this?

Maybe `fn transact` or `fn query` or just `transaction`.

Mirko-von-Leipzig (commented):
Sorry for the delay. This got lost in the pool of PRs/notifications.

crates/store/src/state.rs
drahnr: There is indeed

crates/store/src/db/mod.rs
drahnr: I think eventually all inline numbers should become constants. But I don't think this is the right scope.

crates/store/src/db/mod.rs
drahnr: Note to self: Nesting functions expecting transactions is fine, since diesel flattens them to the same one.

crates/store/src/db/mod.rs
drahnr: It's not, it just put a "frame" around the call and `transaction` is just to omnipresent.

crates/store/src/db/mod.rs
drahnr: I'd like to punt on that, the change ripples through a bunch of code and will get me straight into rebase hell with #928

crates/store/src/db/mod.rs
drahnr: Rename to `transact` and `query`

crates/store/src/db/mod.rs
drahnr: Looking at the code this is very out of place, deleted

crates/store/src/db/mod.rs
```diff
@@ -548,36 +492,28 @@ impl Db {
...
-            })
-            .await
-            .map_err(|err| {
-                DatabaseError::InteractError(format!("Apply block task failed: {err}"))
-            })??;
-
-        Ok(())
+        self.transact("apply block", move |conn| -> Result<()> {
+            // TODO: This span is logged in a root span, we should connect it to the parent one.
+            let _span = info_span!(target: COMPONENT, "write_block_to_db").entered();
```
sergerad, line 500: I think we would just have the transact fns span handle this instead of manually doing a span and event here. So would suggest removing this line and add an `#[instrument` line to 
```
    pub(crate) async fn transact<R, E, Q, M>(&self, msg: M, query: Q) -> std::result::Result<R, E>
```

This would be a trace or debug level span probably.

crates/store/src/db/mod.rs
Mirko-von-Leipzig: Sure no problem :)

crates/store/src/db/mod.rs
Mirko-von-Leipzig: This one is still here 
```suggestion
        let conn = self.pool.get().await.map_err(DatabaseError::MissingDbConnection)?;
```

crates/store/src/db/mod.rs
Mirko-von-Leipzig: I think we can probably just delete this one. It probably came from a time before we instrumented at the top of this function.

I'm somewhat in favor of instrumenting `transact` but at the same time its almost always just a single call nested inside another instrumented function so its probably unnecessary?

drahnr: Deleted

PR #1026, opened 2025-06-27
feat(block-producer): mempool event subscription

Mirko-von-Leipzig:
This PR adds a mempool event subscription stream via a gRPC endpoint on the block-producer. This will be used by the ntx builder to remain in sync with the mempool state.

This is a component of #977 and is a companion PR to #1021.

The even stream endpoint is intended to supersede all of the ntx builder's current gRPC interface i.e. `ntx_builder.proto` will be entirely removed once the rework is complete. However, in order to keep PRs manageable this PR is purely additive - event stream support is added but the ntx <-> block-producer interface remains as is for now.

This PR also removes a vestigial test file.

### Intended Usage

The ntx sync's to the latest committed state using #1021. Once in-sync, it subscribes to the mempool's event stream. This subscription only succeeds if both sides agree on the chain tip which ensures both are in-sync. Repeat until in-sync.

The mempool then emits all events from that committed state going forward. This implies it needs to track events which have not yet been committed in order to re-send those to new subscriptions.

If stream drops, repeat from the top.


crates/block-producer/src/mempool/mod.rs
```diff
@@ -318,7 +350,7 @@ impl Mempool {
     ///
     /// Panics if there is no block in flight.
     #[instrument(target = COMPONENT, name = "mempool.commit_block", skip_all)]
-    pub fn commit_block(&mut self) -> BTreeSet<TransactionId> {
+    pub fn commit_block(&mut self, header: BlockHeader) -> BTreeSet<TransactionId> {
```
igamigo, line 358: Docs for this function could probably get updated to reflect that it also updates subscribers with the header and transactions info

crates/block-producer/src/mempool/subscription.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// These events need to be transmitted when a subscription is started, since the subscriber
+    /// only has the committed state.
+    ///
+    /// A [`BTreeMap`] is used to maintain event ordering while allowing for efficient removals of
+    /// committed or reverted transactions.
+    ///
+    /// The key is auto-incremented on each new insert to support this event ordering.
+    ///
+    /// A reverse lookup index is maintained in `uncommitted_txs_index`.
+    uncommitted_txs: BTreeMap<u64, MempoolEvent>,
```
igamigo: nit: Maybe `inflight_txs` adjusts better?

igamigo: In the future, could be nice to group these in `InflightTxs` with an `insert()`/`remove()`

Mirko-von-Leipzig: Thanks this was a good idea

crates/block-producer/src/mempool/mod.rs
```diff
@@ -220,9 +253,8 @@ impl Mempool {
     ) -> Result<BlockNumber, AddTransactionError> {
         // Add transaction to inflight state.
         let parents = self.state.add_transaction(&transaction)?;
-
+        self.subscription.transaction_added(&transaction);
```
bobbinth, line 258: Should we also mention in the doc comment that an event is sent to the subscribers?

crates/block-producer/src/server.rs
```diff
@@ -235,6 +240,48 @@ impl api_server::Api for BlockProducerRpcServer {
             status: "connected".to_string(),
         }))
     }
+
+    type MempoolSubscriptionStream = MempoolEventSubscription;
+
+    async fn mempool_subscription(
+        &self,
+        request: tonic::Request<MempoolSubscriptionRequest>,
+    ) -> Result<tonic::Response<Self::MempoolSubscriptionStream>, tonic::Status> {
```
bobbinth, line 249: Should we not instrument this?

crates/block-producer/src/mempool/subscription.rs
```diff
@@ -0,0 +1,190 @@
...
+        let capacity = self.inflight_txs.len().mul(2).max(1024);
+        let (tx, rx) = mpsc::channel(capacity);
+        self.subscription.replace(tx);
+
+        // Send each uncommitted tx event in chronological order.
+        //
+        // The ordering is guaranteed by the tracker.
+        for tx in self.inflight_txs.iter() {
+            Self::send_event(&mut self.subscription, tx.clone());
+        }
```
bobbinth: Should we not clear the sent vents from `inflight_txs`? I'm guessing it is not necessary, but may be good to add a comment explaining why.

Mirko-von-Leipzig: The original idea was for this to allow for subscribing while things are live.

In which case all tx events need to be resent anytime a new subscription is started.

We refactor to only support the current always-in-sync methodology but I'm not sure if that helps.

bobbinth: Ah - makes sense! I think I missed that events will be cleared once the block is committed. Let's keep as is.

Mirko-von-Leipzig: I've added some extra comments to explain this a bit better here. I can see how it would be confusing in a local scope.

PR #1027, opened 2025-06-27
feat: add limits for queries

drahnr:
The very most basic level to protect the database receiving invalid values that would cause excessive IO/CPU/time use of the store.

### DIFF
--- a/crates/store/src/db/sql/mod.rs
+++ b/crates/store/src/db/sql/mod.rs
@@ -35,6 +35,8 @@ use super::{
 use crate::{
     db::{
         sql::utils::{
+            QueryParamAccountIdLimit, QueryParamBlockLimit, QueryParamLimiter,
+            QueryParamNoteIdLimit, QueryParamNoteTagLimit, QueryParamNullifierPrefixLimit,
             account_info_from_row, account_summary_from_row, apply_delta, column_value_as_u64,
             get_nullifier_prefix, u64_to_value,
         },
@@ -119,6 +121,8 @@ pub fn select_accounts_by_block_range(
     block_end: BlockNumber,
     account_ids: &[AccountId],
 ) -> Result<Vec<AccountSummary>> {
+    QueryParamAccountIdLimit::check(account_ids.len())?;
+
     let mut stmt = transaction.prepare_cached(
         "
         SELECT
@@ -224,6 +228,8 @@ pub fn select_accounts_by_ids(
     transaction: &Transaction,
     account_ids: &[AccountId],
 ) -> Result<Vec<AccountInfo>> {
+    QueryParamAccountIdLimit::check(account_ids.len())?;
+
     let mut stmt = transaction.prepare_cached(
         "
         SELECT
@@ -710,6 +716,8 @@ pub fn select_nullifiers_by_prefix(
 ) -> Result<Vec<NullifierInfo>> {
     assert_eq!(prefix_len, 16, "Only 16-bit prefixes are supported");
 
+    QueryParamNullifierPrefixLimit::check(nullifier_prefixes.len())?;
+
     let nullifier_prefixes: Vec<Value> =
         nullifier_prefixes.iter().copied().map(Into::into).collect();
 
@@ -936,6 +944,8 @@ pub fn select_notes_by_id(
     transaction: &Transaction,
     note_ids: &[NoteId],
 ) -> Result<Vec<NoteRecord>> {
+    QueryParamNoteIdLimit::check(note_ids.len())?;
+
     let note_ids: Vec<Value> = note_ids.iter().map(|id| id.to_bytes().into()).collect();
 
     let mut stmt = transaction.prepare_cached(&format!(
@@ -965,6 +975,8 @@ pub fn select_note_inclusion_proofs(
     transaction: &Transaction,
     note_ids: BTreeSet<NoteId>,
 ) -> Result<BTreeMap<NoteId, NoteInclusionProof>> {
+    QueryParamNoteIdLimit::check(note_ids.len())?;
+
     let note_ids: Vec<Value> = note_ids.into_iter().map(|id| id.to_bytes().into()).collect();
 
     let mut select_notes_stmt = transaction.prepare_cached(
@@ -1124,6 +1136,13 @@ pub fn select_block_headers(
     transaction: &Transaction,
     blocks: impl Iterator<Item = BlockNumber> + Send,
 ) -> Result<Vec<BlockHeader>> {
+    // Best effort, the iterators are all deterministic `n+1`
+    // <https://doc.rust-lang.org/src/core/slice/iter/macros.rs.html#195>
+    // <https://doc.rust-lang.org/src/core/option.rs.html#2273>
+    // And the conjunction is truthful:
+    // <https://doc.rust-lang.org/src/core/iter/adapters/chain.rs.html#184>
+    QueryParamBlockLimit::check(blocks.size_hint().0)?;
+
     let blocks: Vec<Value> = blocks.map(|b| b.as_u32().into()).collect();
 
     let mut headers = Vec::with_capacity(blocks.len());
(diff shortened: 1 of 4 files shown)


crates/store/src/db/sql/mod.rs
```diff
@@ -119,6 +121,8 @@ pub fn select_accounts_by_block_range(
     block_end: BlockNumber,
     account_ids: &[AccountId],
 ) -> Result<Vec<AccountSummary>> {
+    QueryParamAccountIdLimit::check(account_ids.len())?;
```
bobbinth, line 129: Should we check these limits at the API level (i.e., even before we get to making database requests)? Also, it may make sense to impose these limits in the RPC component as well - so that invalid requests don't make it to the store component at all.

drahnr: My intent was to keep the actual query closeby the limitation. In an error case it would mean additional unwind steps, but the commoncase wouldn't be hurt.
Are we planning to expose the store / its RPC API publicly? If so, that'd be an argument to move it as close as possible to the RPC entrypoints for DoS mitigations.

Mirko-von-Leipzig: The store basically implements the RPC components API 1 to 1. The RPC component just does some shallow checks in places (e.g. ensures the request can parse) before passing it on.

The store does also provide separate methods for the block-producer and network-tx-builder components which are trusted.

I think definitely should be checked in the RPC component. Question is just whether we also duplicate the check on the store. I don't mind much tbh.

drahnr: In that case I pull out the logic and re-use for both.

drahnr: Pulled the logic into `limiter.rs` and pushed it to the query edges just after we parse the tonic input.

bobbinth: I think the primary checks should be in the RPC component (i.e., [here](https://github.com/0xMiden/miden-node/blob/de0e4caffc779867235036361aaec6acecfd87c0/crates/rpc/src/server/api.rs)). And the secondary checks could be at the DB level. We could have a third set of checks at the store's RPC API (as you have it now) - but I'm not sure that's necessary.

drahnr: I clearly misunderstood that.

crates/store/src/limiter.rs
```diff
@@ -0,0 +1,56 @@
+#[derive(Debug, thiserror::Error)]
+#[error("Parameter {which} exceeded limit {limit}: {size}")]
```
Mirko-von-Leipzig: Nit: lowercase errors compose better in error reports

```suggestion
#[error("parameter {which} exceeded limit {limit}: {size}")]
```

Mirko-von-Leipzig: https://github.com/0xMiden/miden-base/discussions/966

Mirko-von-Leipzig (approved):

Mirko-von-Leipzig (commented):
Approved; modulo the changes suggested by Bobbin.

crates/utils/src/limiter.rs
```diff
@@ -0,0 +1,74 @@
+//! Limit the number parameters that would end up in an SQL query
+//!
+//! Used at two sites:
+//! 1. the external facing RPC
+//! 2. just before generating the SQL statement
+//!
+//! The 1st is good to terminate invalid requests as early as possible,
+//! where the second is both a fallback and a safeguard not benching
+//! pointless parameter combinations.
```
bobbinth, line 9: nit: we usually wrap comments after 100 chars.

crates/utils/src/limiter.rs
```diff
@@ -0,0 +1,74 @@
+//! Limit the number parameters that would end up in an SQL query
```
bobbinth: nit: I think this functionality is a bit more broad than just SQL query - so, maybe adjust the language a bit here (though, I don't have any concrete suggestions).

crates/utils/src/limiter.rs
```diff
@@ -0,0 +1,74 @@
...
+        }
+        Ok(())
+    }
+}
+
+pub struct QueryParamAccountIdLimit;
+impl QueryParamLimiter for QueryParamAccountIdLimit {
+    const PARAM_NAME: &str = "account_id";
+    const LIMIT: usize = 1000;
+}
```
bobbinth, line 46: nit: I would maybe add a comment with RPC endpoints where these parameters are used.

crates/rpc/src/server/api.rs
```diff
@@ -82,6 +88,17 @@ impl RpcService {
...
+/// Formats an "Out of range" error
+fn out_of_range_error<E: core::fmt::Display>(err: E) -> Status {
+    Status::out_of_range(err.to_string())
+}
+
+/// Check, but don't repeat ourselves mapping the error
+#[allow(clippy::result_large_err)]
+fn check<Q: QueryParamLimiter>(n: usize) -> Result<(), Status> {
+    <Q as QueryParamLimiter>::check(n).map_err(out_of_range_error)
+}
```
bobbinth: nit: I would move these functions to the end of the file into a `HELPER FUNCTIONS` section.

crates/utils/src/limiter.rs
Mirko-von-Leipzig: I'm surprised this didn't auto-format

drahnr: My editor says 72 chars for the first shown line.

bobbinth: Right - so, the wrapping happens a bit too early.

crates/utils/src/limiter.rs
```diff
@@ -1,8 +1,8 @@
-//! Limit the number parameters that would end up in an SQL query
+//! Limit the number of parameters for single parameter
```
drahnr:
```suggestion
//! Limit the size of a parameter list for a specific parameter
```

crates/utils/src/limiter.rs
```diff
@@ -1,8 +1,8 @@
...
 //!
-//! Used at two sites:
+//! Used for;
```
drahnr:
```suggestion
//! Used for:
```

crates/utils/src/limiter.rs
```diff
@@ -0,0 +1,87 @@
+//! Limit the size of a parameter list for a specific parameter
+//!
+//! Used for:
+//! 1. the external facing RPC
+//! 2. limiting SQL statements not exceeding parameter limits
+//!
+//! The 1st is good to terminate invalid requests as early as possible,
+//! where the second is both a fallback and a safeguard not benching
+//! pointless parameter combinations.
```
bobbinth, line 9:
```suggestion
//! Limit the size of a parameter list for a specific parameter
//!
//! Used for:
//! 1. the external facing RPC
//! 2. limiting SQL statements not exceeding parameter limits
//!
//! The 1st is good to terminate invalid requests as early as possible, where the second is
//! both a fallback and a safeguard not benching pointless parameter combinations.
```

crates/utils/src/limiter.rs
```diff
@@ -0,0 +1,87 @@
...
+impl QueryParamLimiter for QueryParamNoteIdLimit {
+    const PARAM_NAME: &str = "note_id";
+    const LIMIT: usize = 10_000;
+}
+
+/// Only used internally, not exposed via public RPC.
+pub struct QueryParamBlockLimit;
+impl QueryParamLimiter for QueryParamBlockLimit {
+    const PARAM_NAME: &str = "block_header";
+    const LIMIT: usize = 100_000;
```
bobbinth: Is 100K not a problem for Sqlite to handle? (I thought the limit was 10K).

bobbinth (approved):
Looks good! Thank you! I left one small nit suggestion and a question inline.

crates/utils/src/limiter.rs
drahnr: Fixed

PR #1030, opened 2025-06-30
chore: upgrade base

Mirko-von-Leipzig:
This PR upgrades `miden-base` to `next@76ca26` (latest).

This pulls in the changes from https://github.com/0xMiden/miden-base/pull/1493 which I need for the ntx rework.


bin/faucet/src/faucet/mod.rs
```diff
@@ -415,21 +404,6 @@ impl Faucet {
         Ok((account_delta, block_number, notes, tx_id))
     }
 
-    /// Compiles the transaction script that creates the given set of notes.
-    fn compile(&self, notes: &[Note]) -> MintResult<TransactionArgs> {
-        let partial_notes = notes.iter().map(Into::into).collect::<Vec<_>>();
-        let script = self
-            .account_interface
-            .build_send_notes_script(&partial_notes, None, false)
-            .map_err(MintError::ScriptCompilation)?;
```
PhilippGackstatter: Does the faucet still work without this tx script, or is that added somewhere else?

crates/store/src/db/sql/mod.rs
```diff
@@ -359,9 +359,9 @@ pub fn select_account_delta(
         ",
     )?;
 
-    let account_id = account_id.to_bytes();
+    let account_id_bytes = account_id.to_bytes();
     let nonce = match select_nonce_stmt
```
PhilippGackstatter, line 360: Btw. https://github.com/0xMiden/miden-base/pull/1404 was a breaking change to the way the nonce is stored, i.e. it is now the relative change instead of the absolute new value of the nonce. So this will probably require a DB reset or some more work to migrate.

Mirko-von-Leipzig: That's okay; I think there are many breaking changes already included. We've given up on DB compatibility in the short term

Mirko-von-Leipzig: Oh right wait. That is a much larger change um... Will have to think about this.

bin/faucet/src/faucet/mod.rs
Mirko-von-Leipzig: Thanks; you're right I did this far too naively :)

crates/store/src/db/sql/mod.rs
Mirko-von-Leipzig: I think I've fixed this by summing over the delta's block range.

Thanks for bringing this to my attention :)

igamigo:
> CI failures are due to some strangeness in `base` where its failing to re-compile between the cached version and this PRs one.

Happened locally to me, cleared `./target` and it worked out

Issue #1032, opened 2025-06-30
Faucet cache cleanup time configurable


TomasArrachea:
> I would probably make CLEANUP_INTERVAL also configurable, but we can do this in a follow-up PR.

Posted by @bobbinth in https://github.com/0xMiden/miden-node/pull/1017#discussion_r2170582587

PR #1033, opened 2025-06-30
feat(faucet): make `PoW` cleanup interval configurable

TomasArrachea:
closes https://github.com/0xMiden/miden-node/issues/1032


bin/faucet/src/main.rs
```diff
@@ -102,6 +103,10 @@ pub enum Command {
         #[arg(long = "pow-challenge-lifetime", value_name = "DURATION", env = ENV_POW_CHALLENGE_LIFETIME, default_value = "30s", value_parser = humantime::parse_duration)]
         pow_challenge_lifetime: Duration,
 
+        /// The interval at which the `PoW` challenge cache is cleaned up.
+        #[arg(long = "pow-cleanup-interval", value_name = "DURATION", env = ENV_POW_CLEANUP_INTERVAL, default_value = "2s", value_parser = humantime::parse_duration)]
```
Mirko-von-Leipzig, line 107: Is `2s` maybe a bit frequent? But I'm unsure; can't really judge this.

Mirko-von-Leipzig (approved):

bin/faucet/src/server/pow.rs
```diff
@@ -244,8 +241,12 @@ impl ChallengeCache {
     /// The cleanup task is responsible for removing expired challenges from the cache.
     /// It runs every minute and removes challenges that are no longer valid because of their
     /// timestamp.
-    pub async fn run_cleanup(cache: Arc<Mutex<Self>>, challenge_lifetime: Duration) {
-        let mut interval = interval(Duration::from_secs(CLEANUP_INTERVAL_SECONDS));
+    pub async fn run_cleanup(
+        cache: Arc<Mutex<Self>>,
+        challenge_lifetime: Duration,
+        cleanup_interval: Duration,
```
Mirko-von-Leipzig, line 247: This does let me wonder if there isn't a cleaner model for this.

For example, whenever a challenge is created we spawn a cleanup task which sleeps the required lifetime and returns the thing to clean-up.

Or we spawn the same task and its the sole strong owner of `Arc<Challenge>` and all other interested parties only hold `Weak<Challenge>`. Expiration is managed by dropping the primary `Arc`.

```rust
let challenge = Arc::new(challenge);
let interested_party_copy = challenge.weak();

tokio::spawn(async move {
    tokio::time::sleep(lifetime).await;
    drop(challenge);
});
```

bin/faucet/src/main.rs
TomasArrachea: Yeah, it's difficult to say for me too. The constant was set to 2s after this comment: https://github.com/0xMiden/miden-node/pull/974#discussion_r2162650623

bin/faucet/src/server/pow.rs
TomasArrachea: This would mean that all submitted challenges spawn a task for cleanup. It sounds like it would be less efficient and may cause lock contention spikes. Challenges timestamps are sensitive to seconds, so I think that having a `1s` cleanup interval would achieve the same result requiring only one cleanup task.

bin/faucet/src/server/pow.rs
```diff
@@ -36,13 +33,13 @@ pub(crate) struct PoW {
...
     /// Creates a new `PoW` instance.
-    pub fn new(secret: [u8; 32], challenge_lifetime: Duration) -> Self {
+    pub fn new(secret: [u8; 32], challenge_lifetime: Duration, cleanup_interval: Duration) -> Self {
         let challenge_cache = Arc::new(Mutex::new(ChallengeCache::default()));
 
         // Start the cleanup task
         let cleanup_state = challenge_cache.clone();
         tokio::spawn(async move {
-            ChallengeCache::run_cleanup(cleanup_state, challenge_lifetime).await;
+            ChallengeCache::run_cleanup(cleanup_state, challenge_lifetime, cleanup_interval).await;
```
igamigo, line 42: Because we don't do anything with the `JoinHandle`, the task can abort somehow (though unlikely based on the code) and we wouldn't know, right? If so, would be nice to handle this differently. We could open an issue to track this and handle it in a separate PR.

bin/faucet/src/server/pow.rs
Mirko-von-Leipzig: It would let you simplify your internal tracking and cleanup. Instead of tracking timestamps and accounts etc, instead you can do:

```rust
// You can ofc split this into proper types so its less confusing.
ChallengeCache(BTreeMap<AccountId, BTreeMap<ApiKey, Weak<Challenge>>>);

impl ChallengeCache {
    fn cleanup(&mut self) {
        // Remove all challenges where `challenge.strong_count() == 0`
        // because they have expired.
    }
}
```

Basically the expiration task doesn't interact with the cache at all. It just drops the challenge, which should be the primary copy and therefore invalidates all weak references to it. Which can be cleaned up if required.

Mirko-von-Leipzig: I'm not suggesting to do this; I'm just exploring alternatives that involve less manual book keeping. Ideally one would also find something without an `Arc<Mutex<T>>` to centralize things but that might be less viable given the current setup.

bin/faucet/src/server/pow.rs
TomasArrachea: created https://github.com/0xMiden/miden-node/issues/1043

bin/faucet/src/server/pow.rs
TomasArrachea: Hmm I see, using the reference count may be a better alternative than keeping the count manually. Should we create an issue to discuss it?

TomasArrachea: https://github.com/0xMiden/miden-node/issues/1044

bobbinth (approved):
Looks good! Thank you!

PR #1034, opened 2025-06-30
feat: store services on separate addresses

TomasArrachea:
closes https://github.com/0xMiden/miden-node/issues/918

In https://github.com/0xMiden/miden-node/pull/932 the Store API was separated it into three APIs: RPC, BlockProducer, NtxBuilder. This PR changes the startup so that each API runs in a separate address.


bin/node/src/commands/store.rs
```diff
@@ -43,10 +46,21 @@ pub enum StoreCommand {
...
-        /// Url at which to serve the gRPC API.
-        #[arg(long, env = ENV_STORE_URL, value_name = "URL")]
-        url: Url,
+        /// Url at which to serve the store's RPC API.
+        #[arg(long = "rpc.url", env = ENV_STORE_RPC_URL, value_name = "URL")]
+        rpc_url: Url,
+
+        /// Url at which to serve the store's network transaction builder API.
+        #[arg(long = "ntx-builder.url", env = ENV_STORE_NTX_BUILDER_URL, value_name = "URL")]
+        ntx_builder_url: Option<Url>,
```
Mirko-von-Leipzig: I think let's make it required.

Currently everyone only uses the `bundled` cmd so it shouldn't really bother (m)any to have to include another value.

And there is no downside to hosting this API always.

crates/store/src/server/mod.rs
```diff
@@ -26,7 +26,9 @@ mod rpc_api;
 
 /// The store server.
 pub struct Store {
-    pub listener: TcpListener,
+    pub rpc_listener: TcpListener,
+    pub ntx_builder_listener: Option<TcpListener>,
```
Mirko-von-Leipzig: As mentioned elsewhere, lets make this non-optional.

crates/store/src/server/mod.rs
```diff
@@ -117,18 +123,43 @@ impl Store {
...
+                    .layer(TraceLayer::new_for_grpc().make_span_with(store_trace_fn))
+                    .add_service(ntx_builder_service)
+                    .add_service(reflection_service.clone())
+                    .add_service(reflection_service_alpha.clone())
+                    .serve_with_incoming(TcpListenerStream::new(ntx_builder_listener)),
+            );
+        }
+        join_set.spawn(
+            tonic::transport::Server::builder()
+                .layer(TraceLayer::new_for_grpc().make_span_with(store_trace_fn))
```
Mirko-von-Leipzig: Take a look at the `store_trace_fn` - I believe its pretty incorrect at this stage, and we should either split it into three separate ones, or add the missing method names there.

I don't mind which you decide on.

bin/node/src/commands/bundled.rs
```diff
@@ -175,12 +185,21 @@ impl BundledCommand {
             (rpc_id, "rpc"),
         ]);
 
+        // Start network transaction builder. The endpoint is available after loading completes.
+        // SAFETY: socket addr yields valid URLs
+        let store_ntx_builder_url = Url::parse(&format!(
+            "http://{}:{}/",
+            store_ntx_builder_address.ip(),
+            store_ntx_builder_address.port()
+        ))
```
Mirko-von-Leipzig: I think `SocketAddr::to_string()` actually does the right thing here as well 

```suggestion
        let store_ntx_builder_url = Url::parse(&store_ntx_builder_address.to_string());
```

Mirko-von-Leipzig (approved):

SantiagoPittella (approved):
👍🏼

crates/block-producer/src/server.rs
bobbinth, line 1: nit (and not from this PR): I would probably move the tests out of this file.

bin/node/.env
```diff
@@ -5,7 +5,9 @@ MIDEN_NODE_NTX_BUILDER_URL=
 MIDEN_NODE_BATCH_PROVER_URL=
 MIDEN_NODE_BLOCK_PROVER_URL=
 MIDEN_NODE_NTX_PROVER_URL=
-MIDEN_NODE_STORE_URL=
+MIDEN_NODE_STORE_RPC_URL=
+MIDEN_NODE_STORE_NTX_BUILDER_URL=
+MIDEN_NODE_STORE_BLOCK_PRODUCER_URL=
```
bobbinth, line 10: Question: could these all be set to the same address?

Mirko-von-Leipzig: That would error out because each gRPC service gets created separately and should attempt to bind the port and fail.

Its something that is possible to enable if we wanted to support it, but I don't imagine we'd want that?

bobbinth: Aren't we binding them all to the same port [here](https://github.com/0xMiden/miden-node/blob/ce2b895c7686cea04bf81323d604d9d04e42d70a/bin/node/src/commands/bundled.rs#L112-L121)?

Mirko-von-Leipzig: No, port `0` has a special meaning which is to select a random open port. This is useful in the `bundled` mode because we only need to know the port internally but it can be any arbitrary valid port.

crates/block-producer/src/server.rs
TomasArrachea: Where should we move it to? It tests the block producer server startup so I think it actually makes some sense to be placed in this file

TomasArrachea:
Created https://github.com/0xMiden/miden-node/issues/1052 as follow-up

bin/node/.env
bobbinth: Ah - makes sense!

crates/block-producer/src/server.rs
bobbinth: I was thinking of splitting this file into `server/mod.rs` and `server/tests.rs` - but not a strong preference.

PR #1035, opened 2025-06-30
fix: stress-test assumed a -wal file exist from the get go

drahnr:
Address an assumption that doesn't hold for sqlite 3.46.1 as shipped with fedora 41

### DIFF
--- a/bin/stress-test/src/seeding/metrics.rs
+++ b/bin/stress-test/src/seeding/metrics.rs
@@ -163,16 +163,13 @@ impl Display for SeedingMetrics {
 
 /// Gets the size of the store file and its WAL file.
 fn get_store_size(dump_file: &Path) -> u64 {
-    let store_file_size = fs_err::metadata(dump_file)
-        .unwrap_or_else(|_| panic!("Dumpfile always exists: {}", dump_file.display()))
-        .len();
+    let store_file_size = fs_err::metadata(dump_file).expect("Dumpfile always exists").len();
     let wal_file = format!("{}-wal", dump_file.to_str().unwrap());
-    let wal_file_size = fs_err::metadata(&wal_file).map_or_else(
-        |_err| {
+    let wal_file_size = fs_err::metadata(&wal_file)
+        .inspect_err(|_err| {
             eprintln!("No WAL file found: {wal_file}");
-            0
-        },
-        |meta| meta.len(),
-    );
+        })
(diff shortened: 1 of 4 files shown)


SantiagoPittella (approved):

bin/stress-test/src/seeding/metrics.rs
```diff
@@ -163,8 +163,16 @@ impl Display for SeedingMetrics {
 
 /// Gets the size of the store file and its WAL file.
 fn get_store_size(dump_file: &Path) -> u64 {
-    let store_file_size = std::fs::metadata(dump_file).unwrap().len();
-    let wal_file_size =
-        std::fs::metadata(format!("{}-wal", dump_file.to_str().unwrap())).unwrap().len();
+    let store_file_size = fs_err::metadata(dump_file)
+        .unwrap_or_else(|_| panic!("Dumpfile always exists: {}", dump_file.display()))
```
Mirko-von-Leipzig: I thought `fs_err` was meant to handle the filepath errors?

So something like `.expect("dumpfile should exist")` would automatically add in the expected filepath as part of the panic dump?

bin/stress-test/src/seeding/metrics.rs
```diff
@@ -163,8 +163,16 @@ impl Display for SeedingMetrics {
...
+        .len();
+    let wal_file = format!("{}-wal", dump_file.to_str().unwrap());
+    let wal_file_size = fs_err::metadata(&wal_file).map_or_else(
+        |_err| {
+            eprintln!("No WAL file found: {wal_file}");
+            0
+        },
+        |meta| meta.len(),
+    );
```
Mirko-von-Leipzig: Maybe? 

```suggestion
    let wal_file_size = fs_err::metadata(&wal_file)
        .inspect_err(|_| eprintln!("WAL file {wal_file} not found"))
        .map(Metadata::len)
        .unwrap_or_default();

```

Mirko-von-Leipzig (approved):

bin/stress-test/src/seeding/metrics.rs
drahnr: `clippy::pedantic` had an opinion on `.map().unwrap_or_else()`, I'll give your solution a shot

TomasArrachea (approved):

PR #1036, opened 2025-07-01
feat: support gRPC server reflection v1 alpha

Mirko-von-Leipzig:
This adds support for gRPC server reflection `v1alpha` as an addition to the already supported `v1`.

This version is deprecated but not all external applications have migrated. Notably, this includes `postman` which doesn't work our RPC endpoint currently due to this.

The tracking issue for postman is https://github.com/postmanlabs/postman-app-support/issues/13120. Once this is closed and released we can consider reverting this support, though there probably isn't much harm in simply keeping it until its formally removed in `tonic` etc.

### DIFF
--- a/crates/block-producer/src/server.rs
+++ b/crates/block-producer/src/server.rs
@@ -252,11 +252,21 @@ impl BlockProducerRpcServer {
             .build_v1()
             .context("failed to build reflection service")?;
 
+        // This is currently required for postman to work properly because
+        // it doesn't support the new version yet.
+        //
+        // See: <https://github.com/postmanlabs/postman-app-support/issues/13120>.
+        let reflection_service_alpha = tonic_reflection::server::Builder::configure()
+            .register_file_descriptor_set(block_producer_api_descriptor())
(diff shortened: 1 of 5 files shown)


crates/block-producer/src/server.rs
```diff
@@ -252,11 +252,21 @@ impl BlockProducerRpcServer {
             .build_v1()
             .context("failed to build reflection service")?;
 
+        // This is currently required for postman to work properly because
+        // it doesn't support the new version yet.
+        //
+        // See: <https://github.com/postmanlabs/postman-app-support/issues/13120>.
+        let reflection_service_alpha = tonic_reflection::server::Builder::configure()
+            .register_file_descriptor_set(block_producer_api_descriptor())
+            .build_v1()
```
bobbinth: Should this say `build_v1alpha()`? (and same comments for the other instances)

Mirko-von-Leipzig: Yes. Thanks - too much or too little :coffee:

PR #1039, opened 2025-07-01
feat: network transaction state rework

Mirko-von-Leipzig:
This PR creates a new `State` tracker for the network transaction builder.

It is heavily inspired by the existing `PendingNotes` with some minor tweaks but with support for network account deltas and the ability to correlate all of this with inflight transactions.

I've also included a mockup in `NetworkTransactionBuilder::serve_new` of what the overall flow might look like - though I've tried to keep it fairly bare bones to keep the focus on the state portion. However I do recommend starting here as the entry-point since it might help inform the choices I've made.

Once again, I've left existing code in place as is.

### Architectural Overview

State tracks all current inflight txs and their impact on network account state.

State is updated purely via mempool events. The only exception is that an account is locked while it has a network tx in-progress _locally_.

The main loop consists of handling three possible events:

1. Mempool event received -> update state,
2. Interval tick -> select network tx (locking the account) and spawn a task for it,
3. A network tx task completes -> unlock the account if err (ok delegates this responsibility to the mempool event to ensure the account state is correct)

Spawning network txs allows for easy concurrency though this isn't strictly required. Its just a nice side-effect of the in-progress txs not actually mutating the `State`.

### State overview

Account deltas are stored separately to notes, and this should be fairly straight-forward code-wise I think. Notes are a fair bit more complicated to handle - and also drive which accounts are available for selection.

A specific account is locked once its been selected. This locking period is only until the result of the transaction is known locally - the transaction does not have to be committed; only either failed locally, or acknowledged by the mempool via an event. This is to prevent re-using the same notes and incomplete account state which would otherwise be incorrect i.e. this is a  temporary lock. I'm happy to hear alternatives here.

### Todo's

Main thing missing is migrating and extending the `PendingNotes` tests to the new `State`.

I'm unhappy with how complicated the `Notes` state feels, but I think there isn't really a way around this.

A separate PR would color in the `rest of the owl` for creating network transactions, and correct the startup procedures.


crates/ntx-builder/src/lib.rs
```diff
@@ -1,7 +1,14 @@
...
+pub(crate) mod state;
+pub(crate) mod store;
 
 // CONSTANTS
 // =================================================================================================
 
 pub const COMPONENT: &str = "miden-ntx-builder";
+
+/// Maximum allowed network ntoes per network transaction.
+pub(crate) const TX_NOTE_LIMIT: NonZeroUsize = NonZeroUsize::new(50).unwrap();
```
bobbinth: nit: I would maybe rename this to `MAX_NOTES_PER_TX` to make it even more self-explanatory.

crates/ntx-builder/src/state/account.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        let deltas = self.deltas.get_mut(&account).unwrap();
+        match change {
+            TxUpdate::Commit => deltas.pop_front(),
+            TxUpdate::Revert => deltas.pop_back(),
+        };
+        if deltas.is_empty() {
+            self.deltas.remove(&account);
+        }
+    }
```
bobbinth: This assumes that transactions are reverted and committed in order, right? Or at least that the order doesn't matter?

Basically, what happens if we have 2 in-flight transactions against some account and both of them are committed (or revered)?

crates/ntx-builder/src/state/account.rs
```diff
@@ -0,0 +1,97 @@
...
+    account::{Account, AccountDelta, AccountId, delta::AccountUpdateDetails},
+    transaction::TransactionId,
+};
+
+/// Tracks network account deltas for all currently inflight transactions.
+#[derive(Default)]
+pub struct AccountDeltas {
+    deltas: HashMap<NetworkAccountPrefix, VecDeque<NetworkAccountUpdate>>,
+    txs: BTreeMap<TransactionId, NetworkAccountPrefix>,
+}
```
bobbinth: Question: here, and in other places, we use `NetworkAccountPrefix` to identify an account - how difficult would it be to switch to using `AccountId`? (and would that even make sense?)

crates/ntx-builder/src/state/note.rs
```diff
@@ -0,0 +1,249 @@
...
+/// if these are committed or reverted.
+#[derive(Default)]
+pub struct Notes {
+    /// Viable candidate accounts available for selection.
+    ///
+    /// They are guaranteed to have network notes.
+    queue: VecDeque<NetworkAccountPrefix>,
+
+    /// Accounts that have been selected and not yet deselected.
+    in_progress: HashSet<NetworkAccountPrefix>,
```
bobbinth: It feels like these should be tracked at the `State` level somehow - though, I'm not sure if there is a good way to do it. The thinking is that a higher-level component should be responsible for selection strategy and it may need to incorporate other information into the process. So the `Notes` component should be able to tell which notes are available to consume by a given account (or if an account has note available for consumption at all), but the `State` will decide whether to build a transaction against this account and which notes to include.

bobbinth (commented):
Thank you! Looks good! And a very detailed explanation of the architecture is greatly appreciated.

This is not a full review yet, but I left some small questions/comments inline. At the high-level, I think the approach makes sense.

> I'm unhappy with how complicated the `Notes` state feels, but I think there isn't really a way around this.

I'm a bit concerned about this too. I think the complexity that is there is fine for now - but it will only get more complex as we try to incorporate additional information (e.g., how many times executing a note failed, should a note be executed now or after some block etc.).

I tried to think of an alternative approach, and maybe making the state more "account-centric" could help - though, I haven't thought it through far enough to know if it actually makes sense. A rough sketch could look something like this:

```rust
pub struct State {
    /// AccountState here is kind of like a "mini-state" that keeps track of
    /// everything that is account-related. This includes the "inbox" of
    /// network notes (with their consumption status), in-flight account
    /// updates and transactions, the latest account state etc.
    accounts: BTreeMap<AccountId, AccountState>,

    /// A lookup table to help processing of transaction commits and reverts.
    txs_to_accounts: BTreeMap<TransactionId, AccountId>,

    /// A lookup table to help processing of newly received transactions.
    notes_to_accounts: BTreeMap<Nullifier, AccountId>

    /// Keeps track of accounts for which we are building a transaction but
    /// the mempool does not have this transaction yet.
    in_progress: BTreeMap<AccountId>
}
```

crates/ntx-builder/src/state/account.rs
Mirko-von-Leipzig: > This assumes that transactions are reverted and committed in order, right? Or at least that the order doesn't matter?

The order shouldn't matter, it relies purely on counts. We could try make the ordering matter using commitments, but it doesn't really change much aside from having to think about what to do if the ordering is wrong.

> Basically, what happens if we have 2 in-flight transactions against some account and both of them are committed (or revered)?

Then we would expect two calls to this function, and both changes would be reverted or committed.

As an example, if we have three inflight transactions in the order `A B C`, and two revert (that would have to be `B C`), then even if `C` is submitted first we still end up with overall state `A` at the end. And similarly for committed we would have `C` remaining.

crates/ntx-builder/src/state/account.rs
Mirko-von-Leipzig: We would need to manage some sort of `Prefix -> AccountId` mapping. Which would require extra trips to the store for ones we don't have locally and an additional thing to manage in for reverts etc since it will need to have all current local network accounts.

Its possible and all the data is around but I'm not sure it adds any value since the prefix is always available directly from both note tag and account ID.

crates/ntx-builder/src/state/note.rs
Mirko-von-Leipzig: I agree, and think as you mention [here](https://github.com/0xMiden/miden-node/pull/1039#pullrequestreview-2977864079) its not entirely clear how to separate responsibilities and data out without ending up with a single point for all of it.

I suspect we'll have similar issues in the mempool as well. How does one allow good high level decision making but abstract the nitty gritty things lower down so they can be ironclad and simplified.

Mirko-von-Leipzig: I'll try out the structure you've suggested; I had something similar in my earlier experiments but found it clunky. Notably a single transaction can affect multiple multiple account notes/nullifiers, but only a single account state.

But maybe there's a nice way using an additional transaction map.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -94,6 +101,81 @@ pub struct NetworkTransactionBuilder {
...
+            .context("failed to fetch the chain tip data from the store")?
+            .context("chain tip data was None")?;
+
+        let mut mempool_events = block_prod
+            .subscribe_to_mempool(chain_tip.block_num())
+            .await
+            .context("failed to subscribe to mempool events")?;
+
+        let mut state = crate::state::State::new(unconsumed.into_iter());
+        let mut interval = tokio::time::interval(self.ticker_interval);
```
drahnr, line 126: https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variant.Burst is the default behaviour if we miss a tick. Not sure that's intended.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -94,6 +101,81 @@ pub struct NetworkTransactionBuilder {
...
+        // Tracks network transaction tasks until they are submitted to the mempool.
+        //
+        // We also map the task ID to the network account so we can mark it as failed if
+        // it doesn't get submitted.
+        let mut inflight = JoinSet::new();
+        let mut inflight_idx = HashMap::new();
+
+        loop {
+            tokio::select! {
+                _next = interval.tick() => {
```
drahnr, line 138: I just checked, `.tick(..)` is cancellation safe, although it's not in the list of `tokio::select!`.

Mirko-von-Leipzig: Thanks - I've used it in the past so I kinda just assumed :sweat_smile: Always good to confirm though.

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: I think its.. probably fine. We should probably just `skip` instead

This does make me realize though what we definitely want to limit the overall number of in-progress tasks by skipping if `tasks.len() > N`. Thanks!

crates/ntx-builder/src/state/account.rs
Mirko-von-Leipzig: The big upside to using `AccountId` would be detecting conflicting prefixes. I'm still unclear on how we should be dealing with that in the longer term.

bobbinth: I think we'll need to keep the mapping un-ambiguous. One option is to include more bits of the account ID in the note metadata (e.g., in the `aux` field) - but we still need to think through this.

I'd say it should be safe to assume that for every note we'll be able to determine the exact network account it was set to. @PhilippGackstatter, what do you think?

PhilippGackstatter: We're still in the process of coming to a conclusion on the [`NoteTag` refactor](https://github.com/0xMiden/miden-base/discussions/1449) (didn't have time for this recently) but I do think we're going in a direction where the account ID prefix should end up fitting into the note metadata. Given that the account ID prefix is globally unique, a given `NoteTag::NetworkAccount` will unambiguously target a single network account. I think that should resolve that issue?

Mirko-von-Leipzig: > Given that the account ID prefix is globally unique

It is? Why is this then not just the account ID? :sweat_smile: Hashing purposes?

But yes that would resolve it then.

PhilippGackstatter: Yes, they are unique as enforced by the `AccountTree`:

https://github.com/0xMiden/miden-base/blob/21cbcb60dd3c012d69289b5a887ed184cb076a1d/crates/miden-objects/src/block/account_tree.rs#L20-L21

> Why is this then not just the account ID?

The account ID went through a lot of changes, so not sure if I recall the full rationale, but I think a 120-bit account ID makes _highjacking_ (see https://github.com/0xMiden/miden-base/issues/140#issuecomment-2505383425 for a brief explanation) practically impossible. It only allows an attacker to deny access to an ID by frontrunning the registration of that ID if the attacker has a seed that hashes to an account ID with the same prefix (but could be a different suffix). That is more realistically possible because it only requires a match on the first 64 bits. But in order to actually steal assets, the attacker would need to match the full 120 bits.

The only reason for having the prefix uniqueness restriction so far was that the ID prefix uniquely identifies the issuer of a non-fungible asset, since its WORD-sized layout only allows for containing the prefix. We may change asset layouts to two words in the future, in which case we might want to lift the prefix uniqueness restriction. Which in turn will affect network account note tags. One way or another, I don't really see a way around making it unambiguous at the note tag level which network account is targeted, so we'll have to take that into account when considering lifting the uniqueness restriction.

Mirko-von-Leipzig:
I've refactored towards something like @bobbinth suggested https://github.com/0xMiden/miden-node/pull/1039#pullrequestreview-2977864079.

I think this has worked out much better :fire:

As a summary:
- each account gets its own state, including deltas and notes
- the top level is now responsible for managing selection strategy

I've also simplified queue management imo.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+        account_delta: Option<AccountUpdateDetails>,
+    ) {
+        // Skip transactions we already know about.
+        //
+        // This can occur since both ntx builder and the mempool might
+        // inform us of the same transaction. Once when it was submitted to
+        // the mempool, and once by the mempool event.
+        if self.inflight_txs.contains_key(&id) {
+            return;
+        }
```
Mirko-von-Leipzig: This doesn't currently occur; but I may change things so that we do both updates. I'm not sure it matters much tbh.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+                self.queue.push_back(prefix);
+                vacant_entry.insert(AccountState::default())
+            },
+        }
+    }
+
+    /// Handles [`MempoolEvent::BlockCommitted`] events.
+    fn commit_transaction(&mut self, tx: TransactionId) {
+        let Impact { account_delta, notes: _, nullifiers } =
+            self.inflight_txs.remove(&tx).unwrap_or_default();
```
igamigo: If the service somehow crashes and comes back (not sure if this is the idea), it could happen that the inflight transactions does not have the transaction here, right?

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -94,6 +101,87 @@ pub struct NetworkTransactionBuilder {
...
+                    let task_id = match &completed {
+                        Ok((task_id, _)) => *task_id,
+                        Err(join_handle) => join_handle.id(),
+                    };
+                    // SAFETY: both inflights should have the same set.
+                    let candidate = inflight_idx.remove(&task_id).unwrap();
+
+                    match completed {
+                        // Nothing to do. State will be updated by the eventual mempool event.
+                        Ok((_, Ok(_))) => {},
```
igamigo, line 176: Is there a race condition here if the builder picks the same account before the committal mempool event reaches it? Then we'd execute with the account without the delta, no? I guess this depends on how account management occurs in general

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: The initial implementation will just crash the entire node still; but long term the idea is that when this service crashes that the gRPC mempool event subscription will be reset.

And the [mempool will re-send](https://github.com/0xMiden/miden-node/blob/de0aa6bfccead97cc15c3ea918c5b4b0337a46ca/crates/block-producer/src/mempool/subscription.rs#L62-L70) all uncommitted transactions when a subscription is started.

The missing part of this is that the ntx service needs to be sync'd to the same block height as the mempool, which is difficult to guarantee at the moment unless block production is paused :) But once we figure out a good way of sync'ing in general, then we'll enable this decoupled restarting.

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: The concept is that we keep the account locked until either:

- the tx build fails, or
- the tx's mempool event arrives

Right now this is sort of implicit because the account gets unlocked when it receives a delta update. And since the account gets locked when it is selected this means that update event _must_ be that transaction.

I'm still trying to come up with something better, but its eluding me. Essentially the problem is that there are two possible things we can use to unlock the account again. Either when we know locally that the tx was a success, or when that tx's mempool event arrives. The issue is that those two things can occur in either order. Both can carry the same information.

I've chosen to just use the mempool event since we need that code in any case. Though it does feel weird to notify of failure using the local way, and of success in a.. indirect way.

We could also say that we submit both to the state, and it just ignores the 2nd one - logic is already there to ignore duplicate txs.

Mirko-von-Leipzig: The mental model I've used is

```
select account -> lock -> submit tx -> Ok  -> mempool -> event -> unlock
                                    -> Err -> unlock
```

crates/ntx-builder/src/lib.rs
```diff
@@ -1,7 +1,20 @@
...
+pub(crate) const COMPONENT: &str = "miden-ntx-builder";
+
+/// Maximum number of network notes a network transaction is allowed to consume.
+pub(crate) const MAX_NOTES_PER_TX: NonZeroUsize = NonZeroUsize::new(50).unwrap();
+
+/// Maximum number of network transactions which should be in progress concurrently.
+///
+/// This only counts transactions which are being computed locally and does not include
+/// uncommitted transactions in the mempool.
+pub(crate) const MAX_IN_PROGRESS_TXS: usize = 4;
```
bobbinth: nit: since these are defined in `lib.rs`, is `pub(crate)` really needed here? (same question for module imports on lines 5 and 6 above)

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+use miden_objects::{
+    account::delta::AccountUpdateDetails, note::Nullifier, transaction::TransactionId,
+};
+
+mod account;
+
+/// A candidate network transaction.
+///
+/// Contains the data pertaining to a specific network account
+/// which can be used to build a network transaction.
```
bobbinth: nit: comment wrapping (also in a few other places in this file).

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+                continue;
+            }
+
+            return Candidate {
+                reference: candidate,
+                _account_deltas: account.deltas().clone(),
+                _notes: notes,
+            }
+            .into();
+        }
```
bobbinth: Maybe I'm missing something, but shouldn't we add the selected account to `self.in_progress` somewhere here?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+    pub fn mempool_update(&mut self, update: MempoolEvent) {
+        match update {
+            MempoolEvent::TransactionAdded {
+                id,
+                nullifiers,
+                network_notes,
+                account_delta,
+            } => {
+                self.add_transaction(id, nullifiers, network_notes, account_delta);
+            },
```
bobbinth, line 139: I would maybe add a comment that this event is triggered for both externally-submitted transactions and the transactions built by the NTB.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+        // We don't need to prune the inflight transactions because if the account is empty, then it
+        // would have no inflight txs.
+        self.accounts.remove(&prefix);
+        self.queue.retain(|x| x != &prefix);
+    }
+}
+
+/// The impact a transaction has on the state.
+#[derive(Default)]
+struct Impact {
```
bobbinth: nit: I'd maybe name this `TransactionImpact` to improve readability a bit.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+        for nullifier in nullifiers {
+            let prefix = self.nullifier_idx.get(&nullifier).unwrap();
+            self.accounts.get_mut(prefix).unwrap().revert_nullifier(nullifier);
+        }
+    }
+
+    /// Removes the account from tracking under the assumption that it is empty.
+    fn remove_account(&mut self, prefix: NetworkAccountPrefix) {
+        // We don't need to prune the inflight transactions because if the account is empty, then it
+        // would have no inflight txs.
```
bobbinth, line 266: Should we add an `assert!` for this?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+            if self.accounts.get_mut(&prefix).unwrap().commit_nullifier(nullifier).is_empty() {
+                self.remove_account(prefix);
+            }
+        }
+    }
+
+    /// Handles [`MempoolEvent::TransactionsReverted`] events.
+    fn revert_transaction(&mut self, tx: TransactionId) {
+        let Impact { account_delta, notes, nullifiers } =
+            self.inflight_txs.remove(&tx).unwrap_or_default();
```
bobbinth: Would the `or_default()` part be unreachable here? Basically, if we are reverting a transaction, isn't it guaranteed that is is in `self.inflight_txs`?

crates/ntx-builder/src/state/mod.rs
bobbinth: Same question as above: under what conditions would we not find an inflight tx for a committed transaction?

crates/ntx-builder/src/state/account.rs
```diff
@@ -0,0 +1,172 @@
...
+use miden_objects::{
+    account::{Account, AccountDelta, AccountId, delta::AccountUpdateDetails},
+    note::Nullifier,
+};
+
+/// Tracks the state of a network account and its notes.
+#[derive(Default)]
+pub struct AccountState {
+    /// Inflight account updates in chronological order.
+    deltas: VecDeque<NetworkAccountUpdate>,
```
bobbinth, line 13: Is the reason we are tracking account deltas rather than full account states (i.e., `Account` structs) is memory efficiency? If so, I wonder if we should care about it for now. We could probably assuming that we could hold tens of thousands of account states in memory and once we go beyond that, we'd probably need a different solution anyways.

So, basically, if changing this to `VecDeque<Account>` simplifies things, I think we could consider this as a viable alternative.

bobbinth (commented):
Thank you! Looks good! And the structure feels much cleaner now. This is not a full review, but I left some comments inline - mostly nits and questions.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: The mempool sends _all_ reverted transactions but we only store txs with network-related impact. A tx with no network account delta and which produces no network notes or network nullifiers would not be in `self.inflight_txs`.

Might be clearer if I rewrite this as an early return with a comment though.

```rust
// Ignore transactions that aren't tracked.
//
// We only track transactions that have an impact on network state.
let Some(Impact { account_delta, notes, nullifier }) = self.inflight_txs.remove(&tx) else {
    return;
};
```

crates/ntx-builder/src/state/account.rs
Mirko-von-Leipzig: We don't have access to the base account at this moment without reaching out to the store.

Reaching out to the store is of course possible, but then you run into potential race-conditions between the event, mempool and store states. My initial plan to avoid making things overly complex here was to download all account state on startup, but we decided to avoid that for now.

My current plan is to have these delta's and when the ntx is being built, it first reaches out to the store for the latest committed account state. There is still a race-condition here, where the first couple of delta's _might_ be already part of this committed state so we need to filter for those using the commitments. Which also means we probably want to use `TxAccountUpdate` instead of `AccountDelta` but that can be another PR.

crates/ntx-builder/src/state/mod.rs
bobbinth: Makes sense! Yes, using early return would make things more clear here.

crates/ntx-builder/src/state/account.rs
bobbinth: My thinking was that the state could be loaded on startup in 2 steps:

- Step 1: get all network notes from the store. From these, identify a set of relevant network accounts (i.e., accounts that have outstanding network notes for them).
- Step 2: get the accounts identified in Step 1 from the store.

At this point, 3 things could affect the account states:

- An account is updated via a transaction created by NTB. This will be handled naturally.
- A new network account is created by some transaction. This too should be handled naturally.
- A note is received for an account which has not been previously loaded from the store.

The last case would require additional handling, but given that NTB is currently the only one who's allowed to update network accounts, this shouldn't have any race conditions.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+    account::delta::AccountUpdateDetails, note::Nullifier, transaction::TransactionId,
+};
+
+mod account;
+
+/// A candidate network transaction.
+///
+/// Contains the data pertaining to a specific network account
+/// which can be used to build a network transaction.
+pub struct Candidate {
```
bobbinth: nit: should this be `CandidateTx` or `CandidateTransaction`?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// The account ID prefix of this network account.
+    pub reference: NetworkAccountPrefix,
+    /// The current inflight deltas which should be applied to this account.
+    ///
+    /// Note that the first item _might_ be an account creation update.
+    pub _account_deltas: VecDeque<NetworkAccountUpdate>,
+    /// A set of notes addressed to this network account.
+    pub _notes: Vec<NetworkNote>,
```
bobbinth, line 28: nit: why do these filed start with `_`?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+    ///
+    /// This is used to select the next transaction's account.
+    ///
+    /// Note that this _always_ includes _all_ network accounts. Filtering out
+    /// accounts that aren't viable is handled within the select method itself.
+    queue: VecDeque<NetworkAccountPrefix>,
+
+    /// Network accounts which have been selected but whose transaction has not yet
+    /// completed.
+    in_progress: HashSet<NetworkAccountPrefix>,
```
bobbinth: nit: I would maybe add here that this effectively acts as an account lock. My understanding is that the NTB won't attempt to build a transaction for any account that is currently in the `in_progress` set.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: To suppress `clippy::unused_xxx` lints. They're `pub` and not yet actively used anywhere in the codebase (but will be in as part of the next PR).

crates/ntx-builder/src/state/account.rs
Mirko-von-Leipzig: > At this point, 3 things could affect the account states:
>
> - An account is updated via a transaction created by NTB. This will be handled naturally.
> - A new network account is created by some transaction. This too should be handled naturally.
> - A note is received for an account which has not been previously loaded from the store.

I see. So in effect we would _always_ load the account as soon as the first network note arrives, whether at start-up or during operation. And this might include re-loading the account if it was dropped locally once all state was committed. Thanks, I think this will work well but will leave for a follow-up.

Semi-related question: what do we do with network notes that have no account (yet)?

Do we keep them around in-case the account gets created afterwards, or do we just drop them?

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: Ah yes thank you!

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: Its non-trivial - we would have to traverse every transaction's impact and check whether it impacts this specific account.

I can add this if we're concerned?

crates/ntx-builder/src/lib.rs
Mirko-von-Leipzig: Thanks, I think I got confused while refactoring and moving them.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: Its frustrating that this isn't enforceable somehow. afaict `rustfmt` can wrap the comments, but doesn't expand the comments to fill the line.

crates/ntx-builder/src/state/account.rs
bobbinth: > And this might include re-loading the account if it was dropped locally once all state was committed.

Yes, and in the future we could also add some kind of "timeout" logic so that we don't drop the account immediately after all of its notes have been processed, but after some time.

> Semi-related question: what do we do with network notes that have no account (yet)?
> 
> Do we keep them around in-case the account gets created afterwards, or do we just drop them?

Hmmm - good question! We could keep them around, but this would have to be time-bounded - so, doesn't really solve things.

What we could do instead is when a new network account gets created (and NTB receives the corresponding transaction), NTB could make a request to the store to get all network notes associated with this account. There probably is some kind of a race condition here - but should be possible to work around it.

Mirko-von-Leipzig: I've created #1055 to discuss this

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -94,6 +101,87 @@ pub struct NetworkTransactionBuilder {
...
+        //
+        // We also map the task ID to the network account so we can mark it as failed if it doesn't
+        // get submitted.
+        let mut inflight = JoinSet::new();
+        let mut inflight_idx = HashMap::new();
+
+        loop {
+            tokio::select! {
+                _next = interval.tick() => {
+                    if inflight.len() > MAX_IN_PROGRESS_TXS {
```
drahnr, line 139: The inflight set is limited by `MAX_INPROGRESS_TXS`. In each of the m`MAX_NOTES_PER_TX` can exist. How do we ensure they're not invalidating one another? I assume this means implicitly we lock an account as described further down, and hence prevent progress of a task.

Q: How do we prevent dead-locks? I.e. a multicore systems where the tx order for block `A` is `[X Y _]` for `B` is `[Y _ X]` `A` could hold a lock onto `Y` and `B` could hold onto `X` and now both run into each others lock, being locked and block indefinitely.

crates/ntx-builder/src/state/account.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    /// Unconsumed notes of this account.
+    available_notes: BTreeMap<Nullifier, NetworkNote>,
+
+    /// Notes which have been consumed by transactions that are still inflight.
+    nullified_notes: BTreeMap<Nullifier, NetworkNote>,
```
drahnr, line 19: Any particular reason we use `BTreeMap`s over `HashMap`s on the client/node side? We never iterate but only ever do lookups

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: This follows from two facts:

1. A network note is bound to a single network account. As in, the note specifies a single network account that is allowed to consume it.
2. The `State` itself locks a network account once it has been selected until it has completed.

So each note can only be consumed by a specific account, and no account can have more than one transaction in-progress at once.

crates/ntx-builder/src/state/account.rs
Mirko-von-Leipzig: `#[derive(Hash)]` was only [recently](https://github.com/0xMiden/miden-base/issues/1472) added so it wasn't an option before.

I've created issue #1064 for this.

Issue #869, opened 2025-05-23
Align internal-facing gRPC clients and structs


igamigo:
Reference: https://github.com/0xMiden/miden-node/pull/833/files#discussion_r2102247394
Also related: https://github.com/0xMiden/miden-client/pull/912#issuecomment-2902074365

Will add more details later

SantiagoPittella:
# gRPC Client Centralization

## Summary

Currently, gRPC client wrappers are scattered across the node codebase. The goal is to eliminate code duplication, improve maintainability, and provide consistent patterns for client configuration, interceptors, and error handling.

## The Problem

### Current Issues

1. **Code Duplication**: Multiple near-identical client wrappers exist across different crates
2. **Inconsistent Patterns**: Different approaches to TLS configuration, interceptors, and error handling
3. **Maintenance Overhead**: Changes to client logic require updates in multiple locations
4. **Scattered Configuration**: Client setup logic is dispersed throughout the codebase

### Existing Wrappers

#### Store Client Wrappers (Different APIs, Similar Patterns)


- **ntx-builder StoreClient**: Wraps `store::ntx_builder_client::NtxBuilderClient`
  - Methods: `get_current_blockchain_data()`, `get_unconsumed_network_notes()`, `get_network_account_details_by_prefix()`, `get_block_header_by_number()`
  
- **block-producer StoreClient**: Wraps `store::block_producer_client::BlockProducerClient`  
  - Methods: `get_transaction_inputs()`, `get_block_inputs()`, `get_batch_inputs()`, `apply_block()`, `get_block_header_by_number()`

**What's duplicated**: The wrapper pattern (same name, similar structure, same interceptor setup, same error handling patterns)

#### Other Service Wrappers

- `crates/ntx-builder/src/builder/block_producer.rs` - BlockProducerClient
- `crates/rpc/src/client/api.rs` - RPC ApiClient with MetadataInterceptor
- `crates/remote-prover-client/src/remote_prover/` - Multiple prover clients with TLS

#### Common Patterns

- **OtelInterceptor**: Used across most internal clients for tracing
- **TLS Configuration**: Repeated patterns for secure connections
- **Type Safety Wrappers**: Converting between proto types and native types
- **Error Handling**: Similar error patterns across wrappers
- **Client Structure**: Similar wrapper patterns with different underlying clients

## Proposed Solutions

### Option 1: Centralized Client Factory in `miden-node-proto`

[...]

PR #1041, opened 2025-07-02
feat: align internal clients

SantiagoPittella:
closes: #869


SantiagoPittella:
About the remote prover clients, a big chunk of the client's new code (and by far the most complex part IMO), we may want to don't use this new builder for them and it in the `miden-remote-prover` as they were before.

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,633 @@
...
+        Self { inner: self.inner.clone() }
+    }
+}
+
+/// `BlockProducer` store client wrapper with OpenTelemetry tracing.
+///
+/// This wrapper provides access to all `BlockProducer` methods like `get_transaction_inputs`,
+/// `get_block_inputs`, `apply_block`, etc.
+#[derive(Debug)]
+pub struct BlockProducerStoreClient {
```
igamigo: Seems like these are pretty much the same (`BlockProducerStoreClient`, `BlockProducerStoreClient`, `RpcStoreClient`), is there any benefit to making these explicit instead of having a more generic `build<T>` as suggested [here](https://github.com/0xMiden/miden-node/issues/869#issuecomment-3022224233)?

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,633 @@
...
+// WASM-specific factory helper for remote prover clients
+#[cfg(target_arch = "wasm32")]
+impl<C> RemoteClientManager<C>
+where
+    C: Clone,
+{
+    /// Connects using a WASM-compatible client factory.
+    ///
+    /// This is a convenience method for WASM targets that don't use `ClientBuilder`.
+    pub async fn connect_wasm_with<F, E>(&self, factory: F) -> Result<C, E>
```
igamigo: Is there any big difference between this and `connect_with()`? It seems that the `wasm-32`-specific changes are related to what happens within the passed function (basically, defining the tonic transport/channel), right? If so maybe this can be generic over the transport directly and the user could choose what to do there

igamigo (commented):
Not the most in-depth review but leaving some comments for now!

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,633 @@
...
+/// ```
+/// Default timeout for gRPC connections (10 seconds).
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(10);
+
+#[derive(Clone)]
+pub struct ClientBuilder {
+    with_otel: bool,
+    tls: Option<ClientTlsConfig>,
+    timeout: Duration,
+    timeout_disabled: bool,
```
bobbinth: Why not use `timeout: Option<Duration>` instead of having 2 different fields for this?

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,633 @@
...
+        let endpoint = format!("http://{addr}");
+        let channel = self.create_channel(endpoint).await?;
+        Ok(BlockProducerStoreClient::new(channel))
+    }
+
+    /// Build an `NtxBuilder` store client with OpenTelemetry interceptor.
+    pub async fn build_ntx_builder_store_client(
+        &self,
+        url: &Url,
+    ) -> Result<NtxBuilderStoreClient, ClientError> {
```
bobbinth: Why do we specify `Url` here, but for other store clients (RPC, block producer), we use `SockeAddr`?

crates/rpc/src/lib.rs
```diff
@@ -3,7 +3,7 @@ mod server;
 #[cfg(test)]
 mod tests;
 
-pub use client::{ApiClient, MetadataInterceptor};
+pub use client::ApiClient;
```
bobbinth: Would the `MetadataInterceptor` not be needed by the `miden-client`?

crates/rpc/src/client/mod.rs
bobbinth, line 1: Do we still want to export the client from here? Basically, should we think of `rpc`, `store`, `block-producer` crates as pure server crates or should they include clients as well?

crates/ntx-builder/src/builder/block_producer.rs
```diff
@@ -23,15 +27,18 @@ pub struct BlockProducerClient {
...
-
-        Self { inner: block_producer }
+    pub async fn new(
+        block_producer_address: SocketAddr,
+    ) -> Result<Self, miden_node_proto::clients::ClientError> {
+        let client = ClientBuilder::new()
+            .with_otel()
+            .with_lazy_connection(true)
+            .build_block_producer_api_client(block_producer_address)
+            .await?;
```
bobbinth: Why do we need `.await` here now but we didn't need it before? (also relevant in other similar places)

crates/block-producer/src/store/mod.rs
```diff
@@ -121,26 +120,28 @@ impl TryFrom<GetTransactionInputsResponse> for TransactionInputs {
...
-        // SAFETY: The store_url is always valid as it is created from a `SocketAddr`.
-        let channel = tonic::transport::Endpoint::try_from(store_url).unwrap().connect_lazy();
-        let store = store_client::BlockProducerClient::with_interceptor(channel, OtelInterceptor);
+    pub async fn new(
+        store_address: SocketAddr,
+    ) -> Result<Self, miden_node_proto::clients::ClientError> {
+        let inner = ClientBuilder::new()
+            .with_otel()
+            .with_lazy_connection(true)
+            .build_block_producer_store_client(store_address)
```
sergerad: nit: Part of me would prefer `.build::<BlockProducerClient>(store_address)`. But I understand that would require a bit more boilerplate (e.g.`<T: From<Channel>>` or something).

crates/remote-prover-client/src/remote_prover/batch_prover.rs
```diff
@@ -90,17 +97,8 @@ impl RemoteBatchProver {
...
-        let mut client = self
-            .client
-            .lock()
-            .await
-            .as_ref()
-            .ok_or_else(|| {
-                RemoteProverClientError::ConnectionFailed("client should be connected".into())
-            })?
-            .clone();
+        let mut client = self.connect().await?;
```
sergerad, line 106: Do we ever plan on replacing this client approach with the one through the builder?

As I understand it this pathway utilizes a client type that is not wrapped (like `BlockProducerStoreClient` for example).
```rust
    async fn connect(
        &self,
    ) -> Result<ApiClient<tonic::transport::Channel>, RemoteProverClientError> {
```

crates/block-producer/src/store/mod.rs
sergerad: That would at least force us to standardize a bit more. E.G. the interceptor logic could be part of the builder, rather than being part of the RPC build fn
```rust
    /// Build an RPC API client with metadata interceptor and TLS.
    pub async fn build_rpc_api_client(&self, url: &Url) -> Result<RpcApiClient, ClientError> {
```

crates/ntx-builder/src/builder/block_producer.rs
SantiagoPittella: This is a lazy connection, which before it wasn't necessary to .await since it wasn't doing anything `async`. In this version, the `build` method checks if the connection is lazy (doesn't require to be awaited) or eager, and in this last part is where we need to introduce asynchronicity.

crates/block-producer/src/store/mod.rs
SantiagoPittella: I added a `ClientFromChannel` trait to be able to use this approach but it ended up being much more complex than expected and added much more lines of code.

crates/rpc/src/client/mod.rs
SantiagoPittella: Yes, mb. I re-added it again.

crates/rpc/src/lib.rs
SantiagoPittella: Yes, mb. I re-added it.

PR #1042, opened 2025-07-02
fix: update `miden-lib` and use `LexicographicWord` in store

tomyrd:
This PR updates the `miden-lib` dependencies and fixes the generated build errors. Mainly the use of `LexicographicWord` in the store. 

The lock file was updated with `cargo update -p miden-lib` so the changes are only related to this base update.

### DIFF
--- a/bin/stress-test/src/seeding/mod.rs
+++ b/bin/stress-test/src/seeding/mod.rs
@@ -354,11 +354,11 @@ fn create_consume_note_txs(
     accounts
         .into_iter()
         .zip(notes)
-        .map(|(mut account, note)| {
+        .map(|(account, note)| {
             let inclusion_proof = note_proofs.get(&note.id()).unwrap();
             create_consume_note_tx(
                 block_ref,
-                &mut account,
+                account,
                 InputNote::authenticated(note, inclusion_proof.clone()),
             )
         })
@@ -370,7 +370,7 @@ fn create_consume_note_txs(
 /// The account is updated with the assets from the input note, and the nonce is set to 1.
 fn create_consume_note_tx(
     block_ref: &BlockHeader,
-    account: &mut Account,
+    mut account: Account,
     input_note: InputNote,
 ) -> ProvenTransaction {
     let init_hash = account.init_commitment();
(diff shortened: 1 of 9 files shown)


igamigo (approved):
LGTM! Probably not worth the effort too much but should `LexicographicWord` implement `(De)Serializable`? cc @PhilippGackstatter

Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
I assume we're waiting on the base PR so this will get bumped again?

bin/stress-test/src/seeding/mod.rs
```diff
@@ -378,12 +378,20 @@ fn create_consume_note_tx(
...
     });
-    account.set_nonce(Felt::ONE).unwrap();
+
+    let updated_account = Account::from_parts(
+        account.id(),
+        account.vault().clone(),
+        account.storage().clone(),
+        account.code().clone(),
+        ONE,
+    );
```
igamigo: nit: I think you should be able to do `into_parts()` and then `from_parts()` with the updated nonce here in order to avoid cloning

crates/store/src/db/sql/mod.rs
```diff
@@ -493,7 +494,51 @@ pub fn upsert_accounts(
                     });
                 }
 
-                let insert_delta = AccountDelta::from(account.clone());
+                //TODO: Bring back Account to AccountDelta conversion
```
Mirko-von-Leipzig: Is this something we're still waiting on? If not, feel free to merge.

PhilippGackstatter:
> Should we make the new `increment_nonce` public?

Yes, that should not be a big deal if that is the easiest for you. Building a genesis account with a non-zero nonce is a fairly special case and probably only every used once, so again not sure if this should be a general API on the account or if this warrants a slightly more hacky solution using `into_parts` and `from_parts`. I'd lean towards the parts solution but not a strong opinion by any means.

crates/store/src/db/sql/mod.rs
tomyrd: I'm still debugging some issues in the client's integration tests. Once they are working I'll tidy up the code and then merge. I want to be sure no more changes in `miden-base` are needed so the update can be as up to date as posible.

crates/store/src/db/sql/mod.rs
```diff
@@ -534,6 +536,52 @@ pub fn upsert_accounts(
     Ok(count)
 }
 
+fn build_insert_delta(account: &Account) -> Result<AccountDelta, DatabaseError> {
```
igamigo, line 543: nit: Would be nice to comment why this workaround is needed (and when it is used)

igamigo (approved):

PR #1045, opened 2025-07-04
chore(proto): refactor messages structure

SantiagoPittella:
closes: #993 

This PR aims to refactor the structure of the proto files and some messages. I split the common messages into three files: primitives (with the merkle tree and digest messages), accounts (account headers, info, notes), and blockchain (block headers, transactions).

The services remain as top-level "modules" (as Mirko suggested in the issue), and the messages specific to them were placed inside each one of them. Some messages are shared between services, so I introduced a "shared.proto" file with those.

Also, I removed the Request/Response suffixes. In some cases, due to the same name on both messages without the suffix, I used the suffix `Result`. I don't really love this last part, any suggestion on this would be appreciated.

The next step is to use this new names from the generated code in our domain. I'll do this in this PR, but opened it anyways to start discussing the new message's structure.


SantiagoPittella:
1. In the first commit is the initial migration, where I use the `Result` suffix in some messages.
2. In the second commit I replaced the old rust code with the new generated one.
3. After that, each commit is a removal of the `Result` suffix in a message.

proto/proto/remote_prover.proto
```diff
@@ -23,7 +23,7 @@ enum ProofType {
 }
 
 // Request message for proof generation containing payload and proof type metadata.
-message ProvingRequest {
+message Prove {
```
bobbinth: I'm not sure about using action verbs for messages. I find `Prove` to be somewhat confusing and `ProofRequest` would probably have been a more clear name.

proto/proto/remote_prover.proto
```diff
@@ -49,14 +49,14 @@ message ProvingResponse {
 
 service ProxyStatusApi {
     // Returns the status of the proxy.
-    rpc Status(ProxyStatusRequest) returns (ProxyStatusResponse) {}
+    rpc Status(GetProxyStatus) returns (ProxyStatus) {}
 }
 
 // Request message for the status of the proxy.
-message ProxyStatusRequest {}
+message GetProxyStatus {}
```
bobbinth: In other places we use `google.protobuf.Empty` when there are no request parameter. Any reason not to use this here?

proto/proto/remote_prover.proto
```diff
@@ -80,28 +80,28 @@ enum WorkerHealthStatus {
...
 
 service WorkerStatusApi {
     // Returns the status of the worker.
-    rpc Status(WorkerStatusRequest) returns (WorkerStatusResponse) {}
+    rpc Status(GetWorkerStatus) returns (WorkerStatus) {}
 }
 
 // Request message for the status of the worker.
-message WorkerStatusRequest {}
+message GetWorkerStatus {}
```
bobbinth: Same question as above: could we not use `google.protobuf.Empty` instead of creating an empty request type?

proto/proto/remote_prover.proto
```diff
@@ -80,28 +80,28 @@ enum WorkerHealthStatus {
 }
 
 // Response message containing the status of the proxy.
-message ProxyStatusResponse {
+message ProxyStatus {
     // The version of the proxy.
     string version = 1;
     // The proof type supported by this proxy.
     remote_prover.ProofType supported_proof_type = 2;
```
bobbinth: nit: does this need to be `remote_prover.ProofType`? I think we should be able to say just `ProofType` here. (same comment for line 108 below).

proto/proto/block_producer.proto
```diff
@@ -29,10 +25,10 @@ service Api {
     //
     // Currently only a single active subscription is supported. Subscription requests will cancel the active
     // subscription, if any.
-    rpc MempoolSubscription(MempoolSubscriptionRequest) returns (stream MempoolEvent) {}
+    rpc MempoolSubscription(MempoolSubscription) returns (stream MempoolEvent) {}
 }
 
-message MempoolSubscriptionRequest {
+message MempoolSubscription {
```
bobbinth: I would keep this as `MempoolSubscriptionRequest`. Using `MempoolSubscription` makes it sound too much like a message containing mempool subscription.

Also, let's add `MEMPOOL SUBSCRIPTION` section separator above this line.

proto/proto/block_producer.proto
```diff
@@ -2,22 +2,18 @@
...
 import "google/protobuf/empty.proto";
 
 service Api {
     // Submits proven transaction to the Miden network
-    rpc SubmitProvenTransaction(requests.SubmitProvenTransactionRequest) returns (responses.SubmitProvenTransactionResponse) {}
+    rpc SubmitProvenTransaction(shared.SubmitProvenTransaction) returns (shared.ProvenTransaction) {}
 
     // Returns the status info.
-    rpc Status(google.protobuf.Empty) returns (responses.BlockProducerStatusResponse) {}
+    rpc Status(google.protobuf.Empty) returns (shared.BlockProducerStatus) {}
```
bobbinth: Could we define `BlockProducerStatus` in this file? I believe it needs to be shared with the `rpc.proto`, but we should be able to import `block_producer.proto` there, no?

proto/proto/store.proto
```diff
@@ -64,44 +66,232 @@ service Rpc {
     // For preserving some degree of privacy, note tags and nullifiers filters contain only high
     // part of hashes. Thus, returned data contains excessive notes and nullifiers, client can make
     // additional filtering of that data on its side.
-    rpc SyncState(requests.SyncStateRequest) returns (responses.SyncStateResponse) {}
+    rpc SyncState(shared.SyncState) returns (shared.SyncedState) {}
 
     // Returns the status info.
-    rpc Status(google.protobuf.Empty) returns (responses.StoreStatusResponse) {}
+    rpc Status(google.protobuf.Empty) returns (shared.StoreStatus) {}
```
bobbinth: Similar to the last comment, I would define `StoreStatus` here and then just import it from here wherever it is needed.

proto/proto/block_producer.proto
```diff
@@ -2,22 +2,18 @@
...
+import "types/account.proto";
+import "types/blockchain.proto";
+import "types/primitives.proto";
+import "types/shared.proto";
 import "google/protobuf/empty.proto";
 
 service Api {
     // Submits proven transaction to the Miden network
-    rpc SubmitProvenTransaction(requests.SubmitProvenTransactionRequest) returns (responses.SubmitProvenTransactionResponse) {}
+    rpc SubmitProvenTransaction(shared.SubmitProvenTransaction) returns (shared.ProvenTransaction) {}
```
bobbinth: A couple of comments here:

- I don't think `ProvenTransaction` is the right message name here. What we are returning is just the block height and calling it `ProvenTransaction` is going to be confusing. I'm open to other suggestions, but I'd strongly prefer `SubmitProvenTransactionResponse` over `ProvenTransaction` here.
  - Also, this message should probably be defined here and then imported in other files from here as needed.
- `SubmitProvenTransaction` could actually be called `ProvenTransaction` because it in fact contains a proven transaction. I would put this message into `types/transaction.proto` together with other transaction-related messages (e.g., `TransactionId`).

proto/proto/ntx_builder.proto
```diff
@@ -12,18 +13,51 @@ service Api {
...
     //
     // When a new transaction enters the mempool, the NTB needs to be notified of the nullifiers
     // that this transaction consumes through this endpoint. This way, the NTB can discard
     // nullified notes correctly.
-    rpc UpdateNetworkNotes(requests.UpdateNetworkNotesRequest) returns (google.protobuf.Empty) {}
+    rpc UpdateNetworkNotes(UpdateNetworkNotes) returns (google.protobuf.Empty) {}
+}
+
+// Submit a list of network notes to the network transaction builder.
+message SubmitNetworkNotes {
```
bobbinth: Similar to some other comments: I would not use action verbs to name messages. This could probably be `TransactionNetworkNotes` (as this contains network notes generated by a single transaction).

proto/proto/ntx_builder.proto
```diff
@@ -12,18 +13,51 @@ service Api {
...
+        // New status of the transaction.
+        blockchain.TransactionStatus status = 2;
+    }
+
+    // The list of updates for each transaction.
+    repeated TransactionUpdate updates = 1;
+}
+
+// Update the status of network notes that were consumed externally.
+message UpdateNetworkNotes {
```
bobbinth: I would rename this to something like `ConsumedNetworkNotes`.

proto/proto/ntx_builder.proto
```diff
@@ -12,18 +13,51 @@ service Api {
...
+// Update network transaction builder with transaction status changes.
+message UpdateTransactionStatus {
+    // Represents the update for a single transaction.
+    message TransactionUpdate {
+        // Id of the transaction to update.
+        primitives.Digest transaction_id = 1;
+
+        // New status of the transaction.
+        blockchain.TransactionStatus status = 2;
+    }
```
bobbinth: A couple of comments here:

- Shouldn't `TransactionStatus` be defined here? Its definition seems to be pretty specific to the NTB.
- We don't use this nested message pattern often (i.e., `TransactionUpdate` is defined in `UpdateTransactionStatus`) - should we use it more frequently?

proto/proto/ntx_builder.proto
```diff
@@ -12,18 +13,51 @@ service Api {
...
+message SubmitNetworkNotes {    
+    // Id of the transaction that created the notes.
+    primitives.Digest transaction_id = 1;
+
+    // The network notes to submit.
+    repeated account.NetworkNote note = 2;
+}
+
+// Update network transaction builder with transaction status changes.
+message UpdateTransactionStatus {
```
bobbinth: Should this be `TransactionStatusUpdates`?

proto/proto/types/account.proto
```diff
@@ -1,15 +1,65 @@
...
+
+    // Code root hash.
+    primitives.Digest code_commitment = 3;
+
+    // Account nonce.
+    uint64 nonce = 4;
+}
+
+// NOTES
+// ================================================================================================
```
bobbinth: Why do we put notes under accounts? I would create a separate `types/note.proto` file for all note-related messages.

proto/proto/store.proto
```diff
@@ -64,44 +66,232 @@ service Rpc {
...
-    rpc GetNetworkAccountDetailsByPrefix(requests.GetNetworkAccountDetailsByPrefixRequest) returns (responses.GetNetworkAccountDetailsByPrefixResponse) {}
+    rpc GetNetworkAccountDetailsByPrefix(GetNetworkAccountDetailsByPrefix) returns (NetworkAccountDetailsByPrefix) {}
+}
+
+// Applies changes of a new block to the DB and in-memory data structures.
+message ApplyBlock {
+    // Block data encoded using [winter_utils::Serializable] implementation for
+    // [miden_objects::block::Block].
+    bytes block = 1;
+}
```
bobbinth: This should probably be just called `Block` and be in the `types/blockchain.proto`.

proto/proto/types/shared.proto
```diff
@@ -0,0 +1,339 @@
...
+    optional bytes delta = 1;
+}
+
+
+// Represents the result of getting block by number.
+message BlockByNumber {
+    // The requested block data encoded using [winter_utils::Serializable] implementation for
+    // [miden_objects::block::Block].
+    optional bytes block = 1;
+}
```
bobbinth: This is very similar to `ApplyBlock` from the previous comment. I would probably name this `MaybeBlock` and also put it into the `types/blockchain.proto`.

proto/proto/store.proto
```diff
@@ -4,39 +4,41 @@
...
-import "requests.proto";
-import "responses.proto";
+import "types/shared.proto";
+import "types/account.proto";
+import "types/blockchain.proto";
+import "types/primitives.proto";
 import "google/protobuf/empty.proto";
 
 // Store API for the RPC component
 service Rpc {
```
bobbinth: I think most of the message names below need to be improved. I would also probably move most of them here and import them in the `rpc.proto` from here.

The main downside is that this file is probably going to get pretty big - but maybe there is a good way to make it readable with separators.

bobbinth (commented):
Thank you! Not a full review but I left some comments inline.

Overall, I don't think this is much of an improvement over existing layout (and in some ways more confusing). A few things I've highlighted in the comments:

- I would not use action verbs to for message naming.
- If we are removing `Request`/`Response` suffixes, we shouldn't just keep the previous names. For example, `ApplyBlockResponse` is OK as a message name but `ApplyBlock` is probably not. And I think sometimes it is OK to keep `Request` suffix to make things clear.
- I'm not sure we need the `shared.proto` file. Most of these messages could either go into other files under `types` or go into individual service files.
- I think it should be OK to import messages from other service files. For example, in `rpc.proto` we could import messages from `block_producer.proto` and `store.proto`.

proto/proto/rpc.proto
```diff
@@ -64,8 +76,20 @@ service Api {
...
+// Represents the status of the node.
+message RpcStatus {
+    // The rpc component's running version.
+    string version = 1;
+
+    // The store status.
+    store.StoreStatus store_status = 3;
+
+    // The block producer status.
+    block_producer.BlockProducerStatus block_producer_status = 4;
```
bobbinth: nit: I would probably rename field names as:

- `store_status` -> `store`.
- `block_producer_status` -> `block_producer`.

proto/proto/store.proto
```diff
@@ -64,44 +70,505 @@ service Rpc {
     // For preserving some degree of privacy, note tags and nullifiers filters contain only high
     // part of hashes. Thus, returned data contains excessive notes and nullifiers, client can make
     // additional filtering of that data on its side.
-    rpc SyncState(requests.SyncStateRequest) returns (responses.SyncStateResponse) {}
+    rpc SyncState(SyncStateRequest) returns (SyncStateResponse) {}
 
     // Returns the status info.
-    rpc Status(google.protobuf.Empty) returns (responses.StoreStatusResponse) {}
+    rpc Status(google.protobuf.Empty) returns (StoreStatus) {}
```
bobbinth: nit: I would move `Status` to the top of the list (like for other services).

proto/proto/store.proto
```diff
@@ -47,7 +53,7 @@ service Rpc {
     //
     // A basic note sync can be implemented by repeatedly requesting the previous response's block until reaching the
     // tip of the chain. 
-    rpc SyncNotes(requests.SyncNoteRequest) returns (responses.SyncNoteResponse) {}
+    rpc SyncNotes(NoteTags) returns (SyncNotesResponse) {}
```
bobbinth: I would probably keep `SyncNoteRequest` here because the underlying message contains more than just tags.

Also, maybe not for this PR, but we should document how many tags can be passed in and adjust if needed. For example, I think currently the limit may be something like 1000, but we probably need to restrict it quite a bit more. This applies to most endpoints which take lists as parameters.

proto/proto/types/note.proto
```diff
@@ -84,11 +85,23 @@ message NoteSyncRecord {
...
 
     // The note's inclusion proof in the block.
-    merkle.MerklePath merkle_path = 4;
+    primitives.MerklePath merkle_path = 4;
+}
+
+// List of note IDs.
+message NoteIds {
+    // List of note IDs to be queried from the database.
+    repeated primitives.Digest ids = 1;
```
bobbinth: Should we have a message for a single `NoteId` (similar to how we have one for `AccountId` and `TransactionId`).

proto/proto/types/note.proto
```diff
@@ -84,11 +85,23 @@ message NoteSyncRecord {
...
     // The note's metadata.
     NoteMetadata metadata = 3;
 
     // The note's inclusion proof in the block.
-    merkle.MerklePath merkle_path = 4;
+    primitives.MerklePath merkle_path = 4;
+}
+
+// List of note IDs.
+message NoteIds {
```
bobbinth: nit: adding "s" at the end may be a bit too subtle. I wonder if it would be better to say `NoteIdList`. This could also apply to a few other messages.

bobbinth (commented):
Thank you! I think this structure works much better. This is not a full review yet - but I left a few comments inline.

proto/proto/store.proto
SantiagoPittella: I created #1080 to track the second.

proto/proto/types/note.proto
SantiagoPittella: Yes, makes sense to keep that consistency

proto/proto/types/note.proto
```diff
@@ -75,20 +95,20 @@ message NoteInclusionInBlockProof {
...
 message NoteSyncRecord {
-    // The index of the note.
-    uint32 note_index = 1;
-
     // A unique identifier of the note which is a 32-byte commitment to the underlying note data.
-    digest.Digest note_id = 2;
+    NoteId note_id = 1;
+
+    // The index of the note.
+    uint32 note_index = 2;
```
bobbinth: This this index different from `NoteInclusionInBlockProof.note_index_in_block`? If not, we should use the same name for both (probably `note_index_in_block` as it is more descriptive).

proto/proto/types/note.proto
```diff
@@ -61,12 +64,29 @@ message NetworkNote {
...
+message CommittedNote {
+    // Either private, public, or network note.
+    Note note = 1;
+
+    // The data needed to prove that the note is present in the chain.
+    NoteInclusionInBlockProof inclusion_proof = 2;
+}
+
+// Represents the result of getting committed notes.
+message CommittedNotes {
```
bobbinth: This should probably be `CommittedNoteList`.

proto/proto/store.proto
bobbinth, line 1: I still feel that this file is a bit too difficult to navigate. Could we potentially split this into 3 - 4 files? For example, something like this:
```
.
└── proto
    └── store
        ├── rpc.proto
        ├── block_producer.proto
        ├── ntx_builder.proto
        └── shared.proto
```
The `shared.proto` file will probably have very few things in it, and it might make sense to roll it into the `store.proto`.

crates/proto/src/domain/account.rs
```diff
@@ -14,13 +14,16 @@ use thiserror::Error;
 use super::try_convert;
 use crate::{
     errors::{ConversionError, MissingFieldHelper},
-    generated as proto,
+    generated::{
+        account as proto_account,
+        store::{self as proto_store, get_account_proofs_request::account_request},
+    },
```
bobbinth: nit: I would probably still prefer to refer to this as `proto` so that we could do `proto::account::accountId` and `proto::store:: get_account_proofs_request:: account_request`. But also if this creates complications, it is fine as is.

bobbinth (commented):
Looks good! Thank you! I left a few comments inline - mostly nits. The main one is about trying to split up the `store.proto` file for better readability. Other than that, I think this is basically good to go - but would be great to get others to take a look at it too.

proto/proto/types/note.proto
SantiagoPittella: The `note_index` is the leaf index value of the note in the note tree, and the `note_index_in_block` is the index of the note in the note Merkle tree of the block the note was created in.

bobbinth: Maybe I'm missing something, but aren't these the same?

crates/proto/src/domain/account.rs
SantiagoPittella: I did the change for the domain conversions in the proto crate, i can do it in the components too if you prefer it that way, like changing the:
`block_producer_proto::SubmitProvenTransactionResponse` into `proto::block_producer::SubmitProvenTransactionResponse`. wdyt?

proto/proto/store/shared.proto
bobbinth, line 1: Most of the messages here are not actually shared, right? For example, the `Status` message is only relevant to the Store-RPC service, right? If so, I'd move the messages that are not shared into their respective files.

proto/proto/store/shared.proto
```diff
@@ -0,0 +1,278 @@
...
+}
+
+// Represents the result of checking nullifiers.
+message CheckNullifiersResponse {
+    // Each requested nullifier has its corresponding nullifier proof at the same position.
+    repeated primitives.SmtOpening proofs = 1;
+}
+
+// GET BLOCK HEADER BY NUMBER
+// ================================================================================================
```
bobbinth, line 13: Related to the previous comment: I think these may be the only shared messages between the stores sub-services.

proto/proto/store/block_producer.proto
```diff
@@ -0,0 +1,162 @@
...
+
+    // The state commitments of the requested accounts and their authentication paths.
+    repeated account.AccountWitness account_witnesses = 4;
+
+    // The requested nullifiers and their authentication paths.
+    repeated NullifierWitness nullifier_witnesses = 5;
+}
+
+// A nullifier returned as a response to the `GetBlockInputs`.
+message NullifierWitness {
```
bobbinth: Should this message be "nested" in the `BlockInputs`? (to follow the pattern we use in other places).

proto/proto/store/shared.proto
SantiagoPittella: There are a couple messages, like the one that you mentioned, that are shared between one of store subcomponents and other component, in this case would be Store-RPC and RPC component, so it is shared but not between two Store subcomponents. I makes sense to move it to the store-rpc proto file and have the rpc component to use it from there. wdyt?

SantiagoPittella: Just added a commit with this change, and included more renaming of the proto aliases into just proto and then qualifying like what you mentioned before.

SantiagoPittella: I was doing that just in the conversion from generated into native types (qualifying from `proto::*`), should I use this approach in the services definitions too?

bin/faucet/src/rpc_client.rs
```diff
@@ -64,9 +65,9 @@ impl RpcClient {
     ///
     /// Note that this _does not_ include any uncommitted state in the mempool.
     pub async fn get_faucet_account(&mut self, id: FaucetId) -> Result<Account, RpcError> {
-        let request = GetAccountDetailsRequest { account_id: Some(id.account_id.into()) };
+        let request = ProtoAccountId { id: id.account_id.to_bytes().clone() };
```
bobbinth: Could we do this as `proto::AccountId`?

bin/faucet/src/rpc_client.rs
```diff
@@ -38,7 +39,7 @@ impl RpcClient {
     }
 
     pub async fn get_genesis_header(&mut self) -> Result<BlockHeader, RpcError> {
-        let request = GetBlockHeaderByNumberRequest {
+        let request = store_proto::BlockHeaderByNumberRequest {
```
bobbinth: Could we do this as `proto::store::BlockHeaderByNumberRequest`?

bin/faucet/src/rpc_client.rs
```diff
@@ -91,7 +87,7 @@ impl RpcClient {
         &mut self,
         tx: ProvenTransaction,
     ) -> Result<BlockNumber, RpcError> {
-        let request = SubmitProvenTransactionRequest { transaction: tx.to_bytes() };
+        let request = ProtoProvenTransaction { transaction: tx.to_bytes() };
```
bobbinth: Could we do this as `proto::ProvenTranasaction`?

bin/remote-prover/src/api/prover.rs
```diff
@@ -100,7 +103,7 @@ impl ProverRpcApi {
     pub fn prove_tx(
         &self,
         transaction_witness: TransactionWitness,
-    ) -> Result<Response<ProvingResponse>, tonic::Status> {
+    ) -> Result<Response<Proof>, tonic::Status> {
```
bobbinth: I'd prefer this to be `proto::Proof`.

bin/remote-prover/src/api/prover.rs
```diff
@@ -192,18 +195,18 @@ impl ProverApi for ProverRpcApi {
     )]
     async fn prove(
         &self,
-        request: Request<ProvingRequest>,
-    ) -> Result<Response<ProvingResponse>, tonic::Status> {
+        request: Request<ProofRequest>,
+    ) -> Result<Response<Proof>, tonic::Status> {
         match request.get_ref().proof_type() {
-            proto::ProofType::Transaction => {
+            ProtoProofType::Transaction => {
```
bobbinth: I'd prefer this to be `proto::ProofType`.

bobbinth (commented):
Looks good! Thank you! I left some comments inline - mostly about using namespaces to prefix protobuf types (i.e., instead of `ProtoMyType` doing something like `proto::MyType`) - let's try to use this approach consistently throughout.

Also, I think https://github.com/0xMiden/miden-node/pull/1045#discussion_r2206434766 still needs to be addressed.

proto/proto/types/note.proto
SantiagoPittella: I thought that those were different, but talking with @igamigo made me realize that are in fact the same. I'll replace `note_index` with `note_index_in_block` and use the docs of the second as it is more descriptive.

proto/proto/store/rpc.proto
```diff
@@ -0,0 +1,319 @@
...
+    // Last block known by the client. The response will contain data starting from the next block,
+    // until the first block which contains a note of matching the requested tag.
+    fixed32 block_num = 1;
+
+    // Specifies the tags which the client is interested in.
+    repeated fixed32 note_tags = 2;
+}
+
+// Represents the result of syncing notes request.
+message SyncNotesResponse {
```
sergerad, line 304: Part of me wants to strive for fully replacing the request/response suffixes. Mainly because that naming tends to be less expressive than the alternative. Understood that sometimes it may be very difficult to think of an appropriate / expressive name in the context.

Taking this as an example. What would we rename `SyncNotesResponse` to? Perhaps `NoteSyncRecordsByTag` or `..ByTagAndBlock` which gets verbose but you get the idea.

LMK what you think. cc @bobbinth

proto/proto/block_producer.proto
```diff
@@ -32,43 +31,75 @@ service Api {
...
+message SubmitProvenTransactionResponse {
+    // The node's current block height.
+    fixed32 block_height = 1;
+}
+
+// MEMPOOL SUBSCRIPTION
+// ================================================================================================
+
+// Request to subscribe to mempool events.
 message MempoolSubscriptionRequest {
```
sergerad, line 59: Even just `MempoolSubscription` would make sense to me here.

proto/proto/store/block_producer.proto
```diff
@@ -0,0 +1,161 @@
...
+
+    // Returns data required to validate a new transaction.
+    rpc GetTransactionInputs(GetTransactionInputsRequest) returns (TransactionInputs) {}
+}
+
+// GET BLOCK INPUTS
+// ================================================================================================
+
+// Returns data required to prove the next block.
+message GetBlockInputsRequest {
```
sergerad: `BlockInputRequirements` or `BlockInputFields` or similar?

sergerad: Or `*Parameters` for messages that specify parameters like this?

sergerad: would make more sense in the code: 
```Request<BlockInputsParameters>```

rather than
```Request<GetBlockInputsRequest>```.

@bobbinth

File: proto/proto/block_producer.proto (hunk 33 above)
Comment (bobbinth) [reply proto/proto/block_producer.proto:59]: I'm not sure `MempoolSubscription` is a better option here - it implies that the message carries a subscription while it is actually a request for a subscription.

File: proto/proto/store/rpc.proto (hunk 32 above)
Comment (bobbinth) [reply proto/proto/store/rpc.proto:304]: We tried to do that where new names were unambiguous - but for a few, I'm not sure there are good alternatives. For example, we could rename `SyncNotesRequest` into something like `BlockNumberAndNoteTagList` - but I think this may actually be more confusing then `SyncNotesRequest`.

Similar with the `SyncNotesResponse` - though, here maybe something like `NoteSyncRecords` or maybe `BlockNoteSyncRecords` would actually be an improvement (though, I don't think by much).

File: proto/proto/store/block_producer.proto (hunk 34 above)
Comment (bobbinth) [reply proto/proto/store/block_producer.proto]: "InputsParameters" sounds somewhat redundant - but I do think we can improve the naming here by at least getting rid of the `Get` prefix - e.g.:

- `GetBlockInputsRequest` -> `BlockInputsRequest`
- `GetBatchInputsRequest` -> `BatchInputsRequest`
- `GetTransactionInputsRequest` -> `TransactionInputsRequest`

Comment (SantiagoPittella) [reply proto/proto/store/block_producer.proto]: I just pushed this changes, and removed an alias that I missed before.

Comment (sergerad) [reply proto/proto/store/block_producer.proto]: > "InputsParameters" sounds somewhat redundant

The parameters refers to the fields in the request that result the block inputs returned. Although I could be confused about these endpoints I'm not super familiar.

Removing `Get` prefix is definitely an improvement. I still have the nit of disliking `Request<SomethingRequest>` in the code (2x "request" in the type). But its not a problem for readability.

**Review by bobbinth — APPROVED** [review]
Looks good! Thank you! I'm sure we can rename/reshuffle things a bit more, but given the footprint of this PR, I'd prefer merge it sooner rather than later. If there are specific things we want to address, we can do this in follow-up PR/issues.

But would be good to get 1 or 2 more people look through it.

File: proto/proto/types/primitives.proto [hunk 35]
```diff
@@ -0,0 +1,75 @@
...
+message SmtLeafEntryList {
+    // The entries list.
+    repeated SmtLeafEntry entries = 1;
+}
+
+// A leaf in an SMT, sitting at depth 64. A leaf can contain 0, 1 or multiple leaf entries.
+message SmtLeaf {
+    oneof leaf {
+        // An empty leaf.
+        uint64 empty = 1;
```
Mirko-von-Leipzig: This could probably be the empty type `google.proto.Empty` (or whatever it actually is).

proto/build.rs
Mirko-von-Leipzig, line 1: I don't think we need to separate the file descriptors by service, especially for the store.

Mirko-von-Leipzig (approved):
Some minor comments you can ignore; this looks like a really good step forward imo. Thank you!

proto/build.rs
bobbinth: It was my suggestion to separate them. My thinking was that this file has gotten very big and it was difficult to parse through.

There may be better ways to split it up though (rather than by service) - but maybe that could be a separate task.

proto/proto/types/primitives.proto
SantiagoPittella: I think that we still want the leaf index (that's what the `empty` represents here). Probably it is better to just rename it

SantiagoPittella: done

bobbinth (approved):
Looks good! Thank you!

PR #1049, opened 2025-07-07
bug: Fix Otel output for ProposedBlock

### LINKED ISSUE #793: OTel stats are incorrect for block output notes
Ah yeah nice catch. It looks like the `block` variant is incorrect now.

`ProposedBlock::output_note_batches` should be used there it seems, though the naming is a bit confusing :)

I'll create an issue.

_Originally posted by @Mirko-von-Leipzig in https://github.com/0xPolygonMiden/miden-node/pull/783#discussion_r2050589280_

sergerad:
## Context

Closes #793.

Tracing values currently use the wrong value for  `block.batches.output_notes.count`.

## Changes

- Fix `block.batches.output_notes.count` value.
- Fix related expect message.

### DIFF
--- a/crates/block-producer/src/block_builder/mod.rs
+++ b/crates/block-producer/src/block_builder/mod.rs
@@ -411,7 +411,7 @@ impl TelemetryInjectorExt for ProposedBlock {
                 .expect("should have less than u32::MAX output notes"),
         );
 
-        let num_batch_created_notes = self.batches().num_created_notes();
+        let num_batch_created_notes = self.output_note_batches().len();
         span.set_attribute(
             "block.batches.output_notes.count",
             u32::try_from(num_batch_created_notes)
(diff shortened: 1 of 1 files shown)


sergerad:
> Given this is a relatively tiny fix, it might be good to combine this with https://github.com/0xMiden/miden-node/issues/725 and add a test demonstrating this "fix".

@Mirko-von-Leipzig the relevant stack (`ProposedBlock` and `BlockProducer::build_block()`) doesn't have a means of unit testing AFAICT atm.

I also don't think its worthwhile testing tracing values in general. We did create some unit tests a few months ago which I think work sufficiently almost as smoke tests for our telemetry setup in general. But all other meaningful trace output will be inspected by humans during daily / weekly ops which should naturally bring issues out of the woodwork.

Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
> I also don't think its worthwhile testing tracing values in general. We did create some unit tests a few months ago which I think work sufficiently almost as smoke tests for our telemetry setup in general. But all other meaningful trace output will be inspected by humans during daily / weekly ops which should naturally bring issues out of the woodwork.

This should be automatable and checkable so I would much rather have this be tested. Specifically we would want this to pick up odd issues; which we would never be able to manually inspect.

I agree this isn't exactly trivial to test at the moment, but I think this is important to _know_ is correct since this is our view of whats going on.

PR #1051, opened 2025-07-07
feat(ntx): load account state from store

Mirko-von-Leipzig:
This PR implements loading account state from the store whenever the network account receives inflight state as suggested by https://github.com/0xMiden/miden-node/pull/1039#discussion_r2189137207.

In addition, this PR also slightly alters how account state is pruned by replacing checking on each removal to checking as part of the natural iteration through the accounts during the selection process. This delays the pruning of empty accounts (which has little to no impact) but has the benefit of being located in exactly one spot.

----------------------------------

### Improvements

This has a rather nice improvement of letting us store `Account` instead of delta's, and also removes the need for ntx building to reach out to the store.

The downside is that this adds async and possible errors into state management. In general this makes reasoning about effects a bit more difficult.

Currently, we will crash if the store request fails. Which is inline with the gRPC stream syncing, however this does make me think about more robust ways of handling this in the future. I wonder if we wouldn't be better served mimic'ing the accounts on chain behaviour - that is, making them independent agents of sorts aka have each account be its own actor task. They could then fail individually without killing the rest, and could (in-theory) be revived at some point in the future.

### DIFF
--- a/crates/ntx-builder/src/state/account.rs
+++ b/crates/ntx-builder/src/state/account.rs
@@ -25,7 +25,7 @@ pub struct AccountState {
 
 impl AccountState {
     /// Creates a new account state using the given value as the committed state.
-    pub fn committed(account: Account) -> Self {
+    pub fn from_committed_account(account: Account) -> Self {
         Self {
             committed: Some(account),
             inflight: VecDeque::default(),
@@ -35,7 +35,7 @@ impl AccountState {
     }
 
     /// Creates a new account state where the creating transaction is still inflight.
-    pub fn uncommitted(account: Account) -> Self {
+    pub fn from_uncommitted_account(account: Account) -> Self {
         Self {
             inflight: VecDeque::from([account]),
             committed: None,
@@ -46,7 +46,7 @@ impl AccountState {
 
     /// Appends a delta to the set of inflight account updates.
     pub fn add_delta(&mut self, delta: &AccountDelta) {
-        let mut state = self.account();
+        let mut state = self.latest_account();
         state
             .apply_delta(delta)
             .expect("network account delta should apply since it was accepted by the mempool");
@@ -137,7 +137,7 @@ impl AccountState {
     }
 
     /// Returns the latest inflight account state.
-    pub fn account(&self) -> Account {
+    pub fn latest_account(&self) -> Account {
         self.inflight
             .back()
             .or(self.committed.as_ref())
--- a/crates/ntx-builder/src/state/mod.rs
+++ b/crates/ntx-builder/src/state/mod.rs
@@ -81,7 +81,7 @@ impl State {
             let prefix = note.account_prefix();
 
             // Ignore notes which don't target an existing account.
-            let Some(account) = state.account_or_load(prefix).await? else {
+            let Some(account) = state.fetch_account(prefix).await? else {
                 continue;
             };
             account.add_note(note);
@@ -139,7 +139,7 @@ impl State {
 
             self.in_progress.insert(candidate);
             return TransactionCandidate {
-                account: account.account(),
+                account: account.latest_account(),
                 _notes: notes,
             }
             .into();
@@ -186,6 +186,10 @@ impl State {
     ///
     /// Note that this will include our own network transactions as well as user submitted
     /// transactions.
+    ///
+    /// This updates the state of network accounts affected by this transaction. Account state
+    /// may be loaded from the store if it isn't already known locally. This would be the case if
+    /// the network account has no inflight state changes.
     async fn add_transaction(
         &mut self,
         id: TransactionId,
@@ -206,12 +210,12 @@ impl State {
             let prefix = update.prefix();
             match update {
                 NetworkAccountUpdate::New(account) => {
-                    let account_state = AccountState::uncommitted(account);
+                    let account_state = AccountState::from_uncommitted_account(account);
                     self.accounts.insert(prefix, account_state);
                     self.queue.push_back(prefix);
                 },
                 NetworkAccountUpdate::Delta(account_delta) => {
-                    self.account_or_load(prefix)
+                    self.fetch_account(prefix)
                         .await
                         .context("failed to load account")?
                         .context("account with delta not found")?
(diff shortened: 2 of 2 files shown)


crates/ntx-builder/src/state/mod.rs
```diff
@@ -148,36 +178,60 @@ impl State {
...
             self.nullifier_idx.insert(note.nullifier(), note.account_prefix());
-            self.account_or_default(note.account_prefix()).add_note(note);
+            // Skip notes which target a non-existent network account.
+            if let Some(account) = self
+                .account_or_load(note.account_prefix())
+                .await
+                .context("failed to load account")?
+            {
+                account.add_note(note);
+            }
```
Mirko-von-Leipzig: This could be an issue if many notes get created for accounts that don't exist. We probably need a blocklist which prevents re-querying the store for the same non-existent account.

crates/ntx-builder/src/store.rs
```diff
@@ -133,7 +133,7 @@ impl StoreClient {
         Ok(all_notes)
     }
 
-    #[instrument(target = COMPONENT, name = "store.client.get_network_account", skip_all, err)]
+    #[instrument(target = COMPONENT, name = "store.client.get_network_account_by_tag", skip_all, err)]
     pub async fn get_network_account_by_tag(
```
bobbinth: Do we actually need to keep this endpoint? The tag could be converted to prefix by the caller, and then we could just use the new `get_network_account()` endpoint.

crates/ntx-builder/src/state/account.rs
```diff
@@ -20,29 +24,59 @@ pub struct AccountState {
 }
 
 impl AccountState {
+    /// Creates a new account state using the given value as the committed state.
+    pub fn committed(account: Account) -> Self {
```
bobbinth: nit (feel free to ignore): I would maybe rename this to `from_committed_account()`. And the same comment for the `uncommitted()` constructor.

crates/ntx-builder/src/state/account.rs
```diff
@@ -105,39 +136,22 @@ impl AccountState {
         self.available_notes.values()
     }
 
-    pub fn deltas(&self) -> &VecDeque<NetworkAccountUpdate> {
-        &self.deltas
+    /// Returns the latest inflight account state.
+    pub fn account(&self) -> Account {
```
bobbinth: nit (feel free to ignore): maybe this could be `latest_state()`?

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -145,22 +147,20 @@ impl NetworkTransactionBuilder {
...
                         continue;
                     };
-                    let account = candidate.reference;
 
                     let task_id = inflight.spawn(async move {
-                        // TODO:
-                        //
-                        // 1. Fetch account from store.
-                        // 2. The rest of the owl.
+                        // TODO: Run the actual tx.
```
bobbinth, line 152: What needs to be done here?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -242,30 +282,51 @@ impl State {
...
-    /// Removes the account from tracking under the assumption that it is empty.
-    fn remove_account(&mut self, prefix: NetworkAccountPrefix) {
-        // We don't need to prune the inflight transactions because if the account is empty, then it
-        // would have no inflight txs.
-        self.accounts.remove(&prefix);
-        self.queue.retain(|x| x != &prefix);
+    /// Returns the current inflight account, loading it from the store if it isn't present locally.
+    ///
+    /// Returns `None` if the account is unknown.
+    async fn account_or_load(
```
bobbinth: nit (feel free to ignore): maybe use something like `fetch_account()`? I'm really into using action verbs for method/function names :)

crates/ntx-builder/src/state/mod.rs
```diff
@@ -148,36 +178,60 @@ impl State {
...
+
+        Ok(())
     }
 
     /// Handles a [`MempoolEvent::TransactionAdded`] event.
     ///
     /// Note that this will include our own network transactions as well as user submitted
     /// transactions.
-    fn add_transaction(
+    async fn add_transaction(
```
bobbinth, line 193: I would update the doc comments to describe a bit more what happens in this method (e.g., that we may reach out to the store).

crates/ntx-builder/src/store.rs
Mirko-von-Leipzig: We can; I guess I considered this dead code to be removed with the rest of the old impl so I took the shortest route.

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: This will be a separate PR, but essentially just taking the `CadidateTransaction`, adding the additional context required to execute, prove and submit the tx, and then spawn that task.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: I was inspired by the `Map::get_or_insert` style names; I'll try out your suggestion

crates/ntx-builder/src/state/mod.rs
```diff
@@ -59,17 +59,35 @@ pub struct State {
...
+        let mut state = Self {
+            store,
+            accounts: HashMap::default(),
+            queue: VecDeque::default(),
+            in_progress: HashSet::default(),
+            inflight_txs: BTreeMap::default(),
+            nullifier_idx: BTreeMap::default(),
+        };
+
+        let notes = state.store.get_unconsumed_network_notes().await?;
```
drahnr, line 79: Do we want to pull in the `store` in here rather than keeping it external? The benefit seems to be small, at least so far. Or do we intend to use it some more in the future?

Mirko-von-Leipzig: The main use is actually in the `fetch_account` a bit later. As a summary:

- We can download all unconsumed network notes, but
- we don't want to download all network accounts if they have no notes

So when we receive notes for an account that we don't have locally, then we have to reach out to the store to download it. There are some extra complications with this, and in the future we also want some form of lazy loading, so the ultimate design isn't entirely clear.

#1056 might also be of interest.

PR #1054, opened 2025-07-08
Add OnFailure impl and add to RPC trace layer

### LINKED ISSUE #1050: Endpoints return HTTP 200 responses to HTTP GET requests but also log error event
### Version

v0.10.0

### Other packages versions

_No response_

### What happened?

```
curl -vsS https://rpc.devnet.miden.io/
```

Will result in HTTP 200 response. But the backend will log `response failed` error.

This is probably intended behaviour and something we can't change. But I'm wondering if there is a way we can capture a specific HTTP request in middleware and short circuit the typical pathway that results in `response failed`.

We have loadbalancers that send various HTTP requests to the gRPC endpoints. It would be ideal to facilitate these while also not logging error. 

```
curl -vsS https://rpc.devnet.miden.io/
```

### Relevant log output

```shell
`response failed`
```

sergerad:
## Context

Closes #1050.

AWS load balancers hit our RPC server with a request expecting gRPC status code 12. This occurs as expected. But our stack also logs an error event when this happens because of the way the tracing crate's `DefaultOnFailure` type implements the `OnFailure` trait.

This means our dashboards are muddied with constant (30s interval healthcheck) error events.

## Changes

- Add an `OnFailure` struct which implements tracing crate's `OnFailure` trait.
    - Logs DEBUG events for gRPC status code 12 situations.
- Integrate the `OnFailure` struct to the RPC server tracing layer.

### DIFF
--- a/crates/utils/src/tracing/grpc.rs
+++ b/crates/utils/src/tracing/grpc.rs
@@ -1,3 +1,31 @@
+use std::{fmt::Display, time::Duration};
+
+/// A [`tower_http::trace::OnFailure`] implementation which creates events of level dependant on the
+/// failure classification.
+#[derive(Clone)]
+pub struct OnFailure {}
+
+impl<F: Display> tower_http::trace::OnFailure<F> for OnFailure {
+    fn on_failure(&mut self, failure_class: F, latency: Duration, _span: &tracing::Span) {
+        let latency = format!("{} ms", latency.as_millis());
+        if failure_class.to_string().as_str() == "Code: 12" {
+            // Unimplemented failure classifications occur for example when
+            // external load balancers send healthchecks (expecting status code 12).
+            tracing::debug!(
+                classification = %failure_class,
+                latency = %latency,
+                "unimplemented"
+            );
+        } else {
+            tracing::error!(
+                classification = %failure_class,
+                latency = %latency,
+                "response failed"
+            );
+        }
+    }
+}
+
 /// Creates a [`tracing::Span`] based on RPC service and method name.
 macro_rules! rpc_span {
     ($service:literal, $method:literal) => {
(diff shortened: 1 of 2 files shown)


crates/utils/src/tracing/grpc.rs
```diff
@@ -1,3 +1,31 @@
...
+/// failure classification.
+#[derive(Clone)]
+pub struct OnFailure {}
+
+impl<F: Display> tower_http::trace::OnFailure<F> for OnFailure {
+    fn on_failure(&mut self, failure_class: F, latency: Duration, _span: &tracing::Span) {
+        let latency = format!("{} ms", latency.as_millis());
+        if failure_class.to_string().as_str() == "Code: 12" {
+            // Unimplemented failure classifications occur for example when
+            // external load balancers send healthchecks (expecting status code 12).
```
Mirko-von-Leipzig, line 13: Are you saying the load-balancer expects to receive an error as part of its health check?

Mirko-von-Leipzig: Ah I see [here](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html). Should we not be changing these from the AWS defaults instead?

So that it hits a gRPC method that exists, and we change the code to `0` or whatever gRPC indicates as success?

Mirko-von-Leipzig: The default for gRPC is `/AWS.ALB/healthcheck` which of course fails with error code 12 because it is indeed unimplemented.

Should we change that to `/rpc.Api/Status` (or whatever that is), and then fix the code as well?

sergerad: Ah thanks I didn't have the full path correct. Hopefully that works. Let you know ASAP

Mirko-von-Leipzig: Does make me wonder if we shouldn't reformat the gRPC schema so that it becomes `/miden.Rpc` or something in that style

sergerad: That fixed it thanks. Will close this.

sergerad:
Path used in infra configuration wasn't correct. @Mirko-von-Leipzig mentioned the correct one: `/rpc.Api/Status` 🙏 

This is not required.

PR #1057, opened 2025-07-08
feat(ntx): complete the rework

### LINKED ISSUE #870: Generalize `SubmitNetworkNotes`

### LINKED ISSUE #874: Track network notes from when they enter the mempool

### LINKED ISSUE #877: Feed new block headers to NTB

Mirko-von-Leipzig:
This PR adds the final pieces of the network transaction builder rework.

This PR (re)implements:
- the transaction execution process, and
- adds this to the main system loop

There is an inherent sync problem on startup because the ntx builder wants to subscribe to the mempool which in turn requires that the block-producer gRPC server is active. However we don't want block-production to begin because this would cause a desync between mempool and ntx. This PR solves this by adding a barrier as a checkpoint. This only works in bundled mode of course, but should be sufficient until we have more robust/decoupled components.

This PR also removes everything related to the previous implementation that was no longer required.

Closes #877 and closes #874 and closes #870.


crates/ntx-builder/src/transaction.rs
```diff
@@ -0,0 +1,206 @@
...
+            None => return Err(DataStoreError::other("no reference block requested")),
+        }
+
+        Ok((
+            self.account.clone(),
+            None,
+            self.genesis_header.clone(),
+            // TODO: is this correct or should one actually construct it from the genesis header
+            PartialBlockchain::default(),
+        ))
```
Mirko-von-Leipzig: Are these `TODO` fine as is? My understanding was that we only require the genesis block because we treat all notes as unauthenticated.

crates/ntx-builder/src/block_producer.rs
```diff
@@ -53,7 +53,35 @@ impl BlockProducerClient {
...
+                    // exponential backoff with base 500ms and max 30s
+                    let backoff = Duration::from_millis(500)
+                        .saturating_mul(1 << retry_counter)
+                        .min(Duration::from_secs(30));
+
+                    tracing::warn!(
+                        ?backoff,
+                        %retry_counter,
+                        %err,
+                        "store connection failed while fetching genesis header, retrying"
```
igamigo: This message does not look right

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -135,6 +79,12 @@ impl NetworkTransactionBuilder {
         let mut inflight = JoinSet::new();
         let mut inflight_idx = HashMap::new();
 
+        let context = crate::transaction::NtxContext {
+            block_producer: block_producer.clone(),
```
igamigo, line 84: nit: Is this clone needed? Doesn't look like it at first sight

crates/ntx-builder/src/transaction.rs
```diff
@@ -0,0 +1,206 @@
...
+#[async_trait::async_trait(?Send)]
+impl DataStore for NtxDataStore {
+    async fn get_transaction_inputs(
+        &self,
+        account_id: AccountId,
+        ref_blocks: BTreeSet<BlockNumber>,
+    ) -> Result<(Account, Option<Word>, BlockHeader, PartialBlockchain), DataStoreError> {
+        if self.account.id() != account_id {
+            return Err(DataStoreError::AccountNotFound(account_id));
+        }
```
igamigo, line 232: Feels like this should never happen so I wonder if it should be an `assert!()` instead

crates/ntx-builder/src/transaction.rs
```diff
@@ -0,0 +1,206 @@
...
+
+        match ref_blocks.last().copied() {
+            Some(BlockNumber::GENESIS) => {},
+            Some(other) => return Err(DataStoreError::BlockNotFound(other)),
+            // TODO: is this fine to do?
+            None => return Err(DataStoreError::other("no reference block requested")),
```
igamigo: Same here

crates/ntx-builder/src/transaction.rs
igamigo: As for this `TODO`, I think it is indeed correct. I was under the impression we would always want to execute against the latest headers (I think I validated this with Bobbin in one of the issues/PRs), though, but not sure if this is actually beneficial somehow. My initial thought was that it might not be possible to consume the notes (even if not authenticated) against an older block, though I never validated this.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -411,12 +330,8 @@ impl TelemetryInjectorExt for ProposedBlock {
...
         );
 
-        let num_batch_created_notes = self.output_note_batches().len();
-        span.set_attribute(
-            "block.batches.output_notes.count",
-            u32::try_from(num_batch_created_notes)
-                .expect("should have less than u32::MAX batch output notes"),
-        );
+        let num_batch_created_notes: usize = self.output_note_batches().iter().map(Vec::len).sum();
+        span.set_attribute("block.batches.output_notes.count", num_batch_created_notes);
```
Mirko-von-Leipzig: This is an unrelated bug I found while running client integration tests.

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig: I considered that; but decided I shouldn't make that kind of assumption about the inner execution mechanics and lets the error bubble where it might get more context.

crates/ntx-builder/src/transaction.rs
```diff
@@ -114,12 +114,18 @@ impl NtxContext {
...
                     .map(|id| notes.iter().find(|note| note.id() == id).unwrap())
                     .cloned()
-                    .collect();
+                    .collect::<Vec<InputNote>>();
+
+                if notes.is_empty() {
+                    let err =
+                        error.map(|err| err.as_report()).unwrap_or_else(|| "None".to_string());
+                    tracing::warn!(%err, "all network notes failed");
+                }
```
igamigo: Just in case, `successful_notes` or `notes` being empty at this point does not neessarily mean all notes are not executable against the account, but rather that one note failed, which exits the check altogether. I think this prompted [this issue](https://github.com/0xMiden/miden-base/issues/1403).

Mirko-von-Leipzig: Ah wait so this would just mean the very first note failed, for example, and the rest never got evaluated?

igamigo: Correct, this could be the case AFAIU

Mirko-von-Leipzig: I guess options are:

- shuffle notes each time
- retry and rotate until we are out of rotations
- drop notes in the naughty bin (but this is effectively rotation with extra steps)

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig: @bobbinth could you confirm that we want the later block MMR?

We do have the block headers coming in from the mempool, but I thought one can't append new commitments to a partial MMR. If so, how would one update the local partial mmr with the new header? Maybe I've just missed that API.

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig: If we retry and rotate and therefore exhaust all options, we could notify the `AccountState` to invalidate/sleep the account until it gets updated again to wake it up.

Mirko-von-Leipzig: @Fumuran could you explain the current note checker behaviour? Does it iterate in order and stop at first failure?

Fumuran: Yes, right now it works like you said: we are iterating over our notes, checking their consumability. If the check succeeds, this note got pushed to the vector of the success notes (`successful_notes` here) and if not — we are returning the `Failure` with information of failed note and successfully consumed notes.

The reason why (for now) we can't just skip the failed note and continue iterating is because we can not save the VM in the intermediate state to skip the note and continue executing starting from the next one. 

Though if it is important to get the data about successful notes even if we already found the failing one, I can slightly modify the existing algorithm which will allow us to do so, but it will work only for the well known notes, because we can do the static analysis without really executing them. So we will be able to iterate over the well known notes and collect information about all passing notes and all failing notes, but we won't be able to tell anything about custom notes.

Mirko-von-Leipzig: I think that means my handling of successful notes is incorrect - I should iterate in the order of the original input notes then, and not the returned set. Which makes sense now that I think about it.

I think its fine as is for now; I just wanted clarity on the current situation. Right now we would submit the same notes, in the same order, and of course get the same failure. So I need to change things a bit.

crates/ntx-builder/src/transaction.rs
bobbinth: Since we are consuming notes as unauthenticated, this _should_ work (at least for now). But there could be some situations where this could start breaking. Fees would be one example as fee estimation would be much easier if the reference block is close to the tip. Also, expiring transactions won't work with the genesis block - though, we are not creating such transactions in the NTB. So, generally, it would be much better to execute transactions against the latest known block (though, being close to the tip is good enough).

> We do have the block headers coming in from the mempool, but I thought one can't append new commitments to a partial MMR. If so, how would one update the local partial mmr with the new header? Maybe I've just missed that API.

We should be able to add new data to the partial MMR and it would look exactly like the full MMR (this is one of the beauties of the MMR data structure - we can extend it w/o needing to know most the underling data). This can be done via the `PartialMmr::add()` method.

crates/ntx-builder/src/transaction.rs
bobbinth: I actually think that it should be the responsibility of the `NoteConsumptionChecker` to figure out what sequence of notes is consumable. Right now we use a very naive algorithm, but ideally it would work something like this:

- We send a sequence of $n$ notes to it.
- It figures out the biggest subset notes that can be executed in a specific sequences and returns them back.

In the future, it may even take fees into the account.

So, my thinking is that it doesn't need to be the concern of the node.

crates/ntx-builder/src/block_producer.rs
bobbinth: Yes, I think this needs to be updated.

crates/ntx-builder/src/store.rs
```diff
@@ -49,16 +51,40 @@ impl StoreClient {
         Self { inner: store }
     }
 
-    /// Returns the latest block's header from the store.
-    #[allow(dead_code)]
-    #[instrument(target = COMPONENT, name = "store.client.latest_header", skip_all, err)]
-    pub async fn latest_header(&self) -> Result<BlockHeader, StoreError> {
+    #[instrument(target = COMPONENT, name = "store.client.genesis_header", skip_all, err)]
+    pub async fn genesis_header_with_retry(&self) -> Result<BlockHeader, StoreError> {
```
bobbinth, line 55: What was the motivation for introducing this and the other retry endpoints? We were trying to fetch the header before the store became available?

crates/ntx-builder/src/transaction.rs
bobbinth, line 1: nit: would be great to add a few section separators to this file, and maybe some more comments.

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig: I've pushed a commit which tracks the latest MMR. Integration tests still pass.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -27,12 +28,25 @@ pub struct TransactionCandidate {
     pub account: Account,
     /// A set of notes addressed to this network account.
     pub notes: Vec<NetworkNote>,
+    /// The latest locally committed block header.
+    ///
+    /// This should be used as the reference block during transaction execution. It is guaranteed
+    /// to be tracked by `chain_mmr`.
+    pub chain_tip: BlockHeader,
+    /// The chain MMR, which includes the `chain_tip`'s block.
+    pub chain_mmr: PartialBlockchain,
```
bobbinth: This may be OK for now, but cloning `PartialBlockchain` will get expensive as we add more and more block headers into it. Ideally, we should keep only the latest $n$ block headers in it (for relatively small $n$ - i.e., less than 100), but I don't think `PartialBlockchain` exposes any methods for pruning its internal state (e.g., would be great to be able to say prune all data before block $x$). cc @PhilippGackstatter - maybe we should add something like this in `miden-base` (or might even have to do it in `miden-crypto`).

Even if we keep only the last $n$ block headers in the partial blockchain, cloning it may still get expensive (it could still be as big as dozens of KB) - so, regardless of pruning, I wonder if we should be passing it as a reference (or maybe `Arc` or `Arc<Mutex>` if we need to keep it mutable.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -27,12 +28,25 @@ pub struct TransactionCandidate {
     pub account: Account,
     /// A set of notes addressed to this network account.
     pub notes: Vec<NetworkNote>,
+    /// The latest locally committed block header.
+    ///
+    /// This should be used as the reference block during transaction execution. It is guaranteed
+    /// to be tracked by `chain_mmr`.
+    pub chain_tip: BlockHeader,
```
bobbinth: One thing worth clarifying si whether `chain_tip` is already in the `chain_mmr` or not (I think it isn't).

bobbinth (commented):
Thank you! Looks good. Not a full review yet, but I left some comments inline.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: I assumed this structure was cheap because the API for `DataStore` means you always have to clone it.

We'll have to think about this because `Arc<Mutex>` will really lock things down if you have many txs.

crates/ntx-builder/src/store.rs
Mirko-von-Leipzig: Yes, basically on startup we should always retry until the other component is up.

At some stage this would ideally be some sort of extension method that one could apply on any gRPC method e.g.

```rust
store.genesis_header().auto_retry().await
```

which is something I was hoping we could get to once we standardize our clients and servers a bit more.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: It should be possible to safely mutate and share in theory because its an append only structure.

As an example, if we design it as a graph where nodes are connected via `Arc<Child>`, then a caller can keep a subgraph open by retaining the pointer(s). Pruning the graph would consist of dropping the nodes within the `PartialBlockchain` itself, but the nodes would continue to exist and work until all users have dropped their copies.

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: Its sneaky but its required because the mempool event stream subscription above it actually holds a mutable reference to the original client instance.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: That said, the shortest route is using references internally in the interfaces, and then externally one would decide to use either clone or arc + mutex.

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig: Agreed; but what should we do right now? We can

1. ignore and this will lead to retrying accounts indefinitely, or
2. drop the notes entirely, or
3. mark the account as kaputt within the state until it gets another note

(3) sounds good, but isn't trivial to implement.

igamigo: Isn't pushing the notes back at the end of the queue for that network account viable? I think that's what I did before. Next attempts will try using all other notes before retrying the failed note.

> 2. drop the notes entirely, or

Maybe not a bad alternative to drop the failing note since it's an early iteration anyway.

> 3. mark the account as kaputt within the state until it gets another note

How would this work? Essentially for a transaction candidate, you get a set of 0 to 50 notes that you are trying to consume with an account. Out of those, any of them could fail so you end up with 3 sets of notes:

- the successful notes
- the one failing note (if any)
- the set of notes that were not tested because the check halted on the failing note 

We probably want to execute the successful ones anyway (as it's being done), but next attempt for the account should likely not involve the note that failed and instead continue with the untested notes. This was my thinking originally but I may be missing something.

crates/ntx-builder/src/state/mod.rs
bobbinth: I think if we can prune the historical data, it solves the problem for the immediate (and maybe even medium-term) future. So, if `Arc<Mutext>` causes complications now, I wouldn't worry about it. But we should implement pruning sooner rather than later.

Basically, we don't want the `PartialMmr` to contain 100K+ block headers and then clone that for every transaction.

crates/ntx-builder/src/transaction.rs
bobbinth: Yeah, I think 3 sounds too drastic (i.e., if one note failed doesn't mean we should kaputt the entire account).

I think a variation of 2 would work + what @igamigo suggested could work - but not sure how much more complexity it adds:

- If a note failed, rotate it to the back of the list so that next time it would be the last one in the list to try to execute.
- If it failed for the second time, remove it from the list entirely.

In the future, we could get more sophisticated (e.g., adding retry count, some backoff strategy etc.).

Mirko-von-Leipzig: I've cheated a bit and simply shuffle the note vector each time before execution.

Rotating would require a bit more book-keeping because the notes are currently a `BTreeMap` and not something rotatable. Its of course possible but not something I'd like to do on the fly.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: I've reverted to using the genesis block for now. I'll create an issue to revisit this once we've added pruning to `base`.

Mirko-von-Leipzig:
I've reverted to using genesis as the reference block, and to avoid blocking an account by a faulty first note I've added note shuffling.

crates/ntx-builder/src/store.rs
bobbinth: Should we create an issue about this?

crates/ntx-builder/src/transaction.rs
```diff
@@ -0,0 +1,256 @@
...
+                .expect("runtime should be built");
+
+            rt.block_on(async move {
+                let mut notes = notes
+                    .into_iter()
+                    .map(|note| InputNote::Unauthenticated { note: note.into() })
+                    .collect::<Vec<_>>();
+                // We shuffle the notes here to prevent having a failing note always in front.
+                notes.shuffle(&mut rand::rng());
+                let notes = InputNotes::new(notes).map_err(NtxError::InputNotes)?;
```
bobbinth, line 86: Not for this PR: it may be convenient to have something like `InputNotes::from_unauthenticated_notes(Vec<Note>)` constructor in `miden-base`. cc @PhilippGackstatter and @Fumuran.

crates/ntx-builder/src/transaction.rs
```diff
@@ -0,0 +1,256 @@
...
+            )
+            .await
+        {
+            Ok(NoteAccountExecution::Success) => notes,
+            Ok(NoteAccountExecution::Failure { successful_notes, error, .. }) => {
+                let notes = successful_notes
+                    .into_iter()
+                    .map(|id| notes.iter().find(|note| note.id() == id).unwrap())
+                    .cloned()
+                    .collect::<Vec<InputNote>>();
```
bobbinth, line 137: Not for this PR: I wonder if `NoteConsumptionChecker` should return successful notes as `Vec<Note>` for convenience. cc @Fumuran.

bobbinth (approved):
All looks good! Thank you! I left a couple of comments inline - but they are for potential improvements in `miden-base`.

crates/ntx-builder/src/transaction.rs
bobbinth: I think this may be covered by our list of future improvements - but I think it may be good to create a dedicated issue for this.

bobbinth:
I think I broke something telemetry-related by merging `next` into this PR. @sergerad - could you take a look?

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig: #1071 maybe?

crates/ntx-builder/src/store.rs
Mirko-von-Leipzig: #1072

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig: Please, this would be useful :)

I was actually thinking the return type could always be:

```rust
struct NoteCheckResult {
    pub successful: Vec<Note>,
    pub remainder: Vec<Note>,
    pub error: Option<Err>,
}
```

Though maybe one still needs the overall `Result` wrapper depending on the inner failure reasons..

But essentially I thought `Result::Err` would be the same as `successful.is_empty()` at the moment.

crates/ntx-builder/src/transaction.rs
Fumuran: Created an issue: https://github.com/0xMiden/miden-base/issues/1565

crates/ntx-builder/src/transaction.rs
Fumuran: Created an issue: https://github.com/0xMiden/miden-base/issues/1566

PR #1058, opened 2025-07-08
Remove unused ntx_builder initialization and related dead code from server

Galoretka:
This commit removes an unused initialization of the ntx_builder variable in the BlockProducer::serve function, as well as all related dead code in the server module. The ntx_builder client was previously created but never used in this context, which could cause confusion and reduce code clarity. The codebase is now cleaner and easier to maintain, with no impact on the server's core logic or functionality.


Mirko-von-Leipzig:
Thanks @Galoretka but this was done intentionally, the PR #1057 already addresses this.

PR #1061, opened 2025-07-09
feat: Add trace layer to remote prover and proxy

### LINKED ISSUE #1060: Add Trace Layer to Remote Prover and Proxy
## Context

We have 3 RPC servers which are configured with a gRPC trace layer that adds OTEL info to spans:
```rust
            .layer(TraceLayer::new_for_grpc().make_span_with(store_trace_fn))
            .add_service(rpc_service)
...
```

We have a number of servers that do not add this layer:
```rust
        let server = Server::builder().add_service(status_server).serve_with_incoming_shutdown(
            TcpListenerStream::new(listener),
            async move {
                let _ = server_shutdown.changed().await;
                info!("gRPC status service received shutdown signal");
            },
        );
```

These include the remote prover and proxy.

## Work

Add the trace layers to remote prover and proxy with path matching as seen in `crates/utils/src/tracing/grpc.rs`.

sergerad:
## Context

Closes #1060.

All of our components run an Otel trace layer on their tonic stack.

Some components are missing these layers.

## Changes
- Add trace layers to remote prover and proxy tonic servers.
- Refactor the traced span fns in `tracing/grpc.rs` so that the explanatory comment is not duplicated.

### DIFF
--- a/crates/utils/src/tracing/grpc.rs
+++ b/crates/utils/src/tracing/grpc.rs
@@ -9,7 +9,7 @@ macro_rules! rpc_span {
     };
 }
 
-/// Represents an Otel compatible, traced Miden compoent.
+/// Represents an Otel compatible, traced Miden component.
 ///
 /// Used to select an appropriate [`tracing::Span`] for the tonic server to use.
 #[derive(Copy, Clone)]
--- a/crates/utils/src/tracing/grpc.rs
+++ b/crates/utils/src/tracing/grpc.rs
@@ -9,13 +9,39 @@ macro_rules! rpc_span {
     };
 }
 
-/// A [`trace_fn`](tonic::transport::server::Server) implementation for the RPC which
+/// Represents an Otel compatible, traced Miden compoent.
+///
+/// Used to select an appropriate [`tracing::Span`] for the tonic server to use.
+#[derive(Copy, Clone)]
+pub enum TracedComponent {
+    Rpc,
+    BlockProducer,
+    StoreRpc,
+    StoreBlockProducer,
+    StoreNtxBuilder,
+    RemoteProver,
+    RemoteProverProxy,
+}
+
+/// Returns a [`trace_fn`](tonic::transport::server::Server) implementation for the RPC which
 /// adds open-telemetry information to the span.
 ///
-/// Creates an `info` span following the open-telemetry standard: `rpc.rpc/{method}`.
+/// Creates an `info` span following the open-telemetry standard: `{service}.rpc/{method}`.
 /// Additionally also pulls in remote tracing context which allows the server trace to be connected
 /// to the client's origin trace.
-pub fn rpc_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
+pub fn traced_span_fn<T>(svc: TracedComponent) -> fn(&http::Request<T>) -> tracing::Span {
+    match svc {
+        TracedComponent::Rpc => rpc_trace_fn,
+        TracedComponent::BlockProducer => block_producer_trace_fn,
+        TracedComponent::StoreRpc => store_rpc_trace_fn,
+        TracedComponent::StoreBlockProducer => store_block_producer_trace_fn,
+        TracedComponent::StoreNtxBuilder => store_ntx_builder_trace_fn,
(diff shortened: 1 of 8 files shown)


crates/utils/src/tracing/grpc.rs
```diff
@@ -129,6 +131,30 @@ pub fn store_ntx_builder_trace_fn<T>(request: &http::Request<T>) -> tracing::Spa
     add_network_attributes(span, request)
 }
 
+fn remote_prover_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
+    let method = request.uri().path().rsplit('/').next().unwrap_or("Unknown");
+    let span = if method == "Prove" {
+        rpc_span!("remote-prover.rpc", "Prove")
+    } else {
+        rpc_span!("remote-prover.rpc", "Unknown")
```
sergerad: This is probably wrong. I struggle to infer the correct names for these based on the proto files.

crates/utils/src/tracing/grpc.rs
```diff
@@ -129,6 +131,30 @@ pub fn store_ntx_builder_trace_fn<T>(request: &http::Request<T>) -> tracing::Spa
...
+    let span = add_otel_span_attributes(span, request);
+    add_network_attributes(span, request)
+}
+
+fn remote_prover_proxy_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
+    let method = request.uri().path().rsplit('/').next().unwrap_or("Unknown");
+    let span = if method == "Status" {
+        rpc_span!("remote-prover-proxy.rpc", "Status")
+    } else {
+        rpc_span!("remote-prover-proxy.rpc", "Unknown")
```
sergerad, line 151: Same as above, probably wrong.

crates/utils/src/tracing/grpc.rs
Mirko-von-Leipzig: I think its correct.

We could try automate this as part of the build.rs somehow because it does allow introspection of the services and methods.

The main issue is that the span name must be a `&'static str` which is why we use the match.

Maybe we are using the wrong approach, and the span shouldn't contain the method name but just have a `method` field which we could set. But we can revisit that in a separate PR.

Mirko-von-Leipzig: Ah hang on we probably also need to account for the `Status` endpoint on both.

They're separate services, but we can probably combine them here?

sergerad: Yea you mean add this here right? I only recall seeing status service on the proxy server setup. But I must've missed it for worker.
```rust
rpc_span!("remote-prover.rpc", "Status")
```

sergerad: And maybe health service too
```rust
            .add_service(rpc.api_service)
            .add_service(rpc.status_service)
            .add_service(health_service)
```
edit: maybe not, doesn't seem so based on the proto at least

sergerad: Oh wait should it then be `remote-prover.api.rpc` and `remote-prover.status.rpc`?

Mirko-von-Leipzig: Ah. Yeah technically it should be. I really am starting to think we should just make the span name generic like `grpc-server` and then `component=remote-prover`, `grpc.service=rpc`, `grpc.method=Status`, which could all be derived from the input path. Or something like that, which could all be done in a single impl for all.

Mirko-von-Leipzig: I should lookup what the otel standards say again.

sergerad: > make the span name generic like grpc-server and then component=remote-prover, grpc.service=rpc, grpc.method=Status, which could all be derived from the input path

I think this would be good

sergerad: But for the sake of observability the name of the span should be as specific as possible. Because its one of the main fields that comes up by default when looking at events on Honeycomb.

sergerad: Lets keep what we have today and see how our observability experience goes. Rack up a bunch of changes to make once we have used Honeycomb in anger.

I added the Status endpoint for the worker span now.

PR #1075, opened 2025-07-10
chore: align internal clients

SantiagoPittella:
Simplified version of the client builder ( #869 ), in this version I added a generic client builder with the addition of a trait to use the builder: `GrpcClientBuilder`. 

The trait has the `from_builder` method, which is first implemented for the store-related clients.

The idea in this version is to keep it as simple as possible. Adding custom interceptors should removed the need for the explicit `<InterceptedService<Channel, OtelInterceptor>>` in client's types. Also, custom interceptors should be enough to replce the `miden_node_rpc` client with a custom one using this changes, since it uses an interceptor to match the package's version.

cc @Mirko-von-Leipzig


crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+        self.address = address;
+        self
+    }
+
+    pub async fn build<T>(self) -> Result<T>
+    where
+        T: GrpcClientBuilder,
+    {
+        T::from_builder(self).await
+    }
```
Mirko-von-Leipzig: If you replace `build` with `connect` and `connect_lazy` then the general case doesn't need to be async.

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+/// consistent configuration options like TLS, OTEL interceptors, and connection types.
+pub trait GrpcClientBuilder: Sized {
+    /// Creates a new client instance from the builder configuration.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the client cannot be created due to invalid configuration,
+    /// network issues, or other initialization problems.
+    #[allow(async_fn_in_trait)]
+    async fn from_builder(builder: Builder) -> Result<Self>;
```
Mirko-von-Leipzig: I think we can simplify this further by

```rust
pub trait ClientWithInterceptor {
    fn with_interceptor(channel, interceptor) -> Self;
}
```

then you can move the bulk of the trait impls into the `Builder::build` (or `connect`/`connect_layy` if you use that suggestion).

You may need to add some associated types to make this work. Or maybe `Client<InterceptedService<Channel, OtelInterceptor>>` will cause issues. This should be possible somehow though.

Mirko-von-Leipzig (commented):
I think this looks good; I think we should see if we can collapse them a bit further which should be possible.

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+        if self.with_tls {
+            endpoint = endpoint
+                .tls_config(tonic::transport::ClientTlsConfig::new().with_native_roots())
+                .context("Failed to configure TLS")?;
+        }
+
+        self.endpoint = Some(endpoint);
+
+        Ok(())
+    }
```
Mirko-von-Leipzig: This should let you simplify `trait GrpcClientBuilder` to just the wrapping call, i.e. `RpcApiClient::with_interceptor(channel, OtelInterceptor)`. 

```suggestion
    pub async fn connect<T>(mut self) -> Result<T>
    where
        T: GrpcClientBuilder,
    {
        let channel = self.build_endpoint()?.connect().await?;
        Ok(T::with_interceptor(channel))
    }

    pub fn connect_lazy<T>(mut self) -> Result<T>
    where
        T: GrpcClientBuilder,
    {
        let channel = self.build_endpoint()?.co;
        Ok(T::with_interceptor(channel))
    }

    fn build_endpoint(self) -> Result<Endpoint> {
        let mut endpoint = Endpoint::from_shared(self.address.clone())
            .context("Failed to create endpoint from address")?;

        if let Some(timeout) = self.with_timeout {
            endpoint = endpoint.timeout(timeout);
        }

        if self.with_tls {
            endpoint = endpoint
                .tls_config(tonic::transport::ClientTlsConfig::new().with_native_roots())
                .context("Failed to configure TLS")?;
        }

        Ok(endpoint)
    }
```

crates/rpc/src/client/api.rs
bobbinth, line 1: Question: why do we have a client defined in the RPC component? AFAICT, no other components have clients defined in this way. Why is the RPC component special?

SantiagoPittella: It is defined in the component itself due to the MetadataInterceptor , that checks for matching versions of client and server, that is used only in with this component. We can for sure centralize it in the new builder and define it there with the other `Clients*` structs. @Mirko-von-Leipzig do you have an opinion on this?

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+// ================================================================================================
+
+pub type InstrumentedRpcApiClient = RpcApiClient<InterceptedService<Channel, OtelInterceptor>>;
+pub type InstrumentedBlockProducerApiClient =
+    BlockProducerApiClient<InterceptedService<Channel, OtelInterceptor>>;
+pub type InstrumentedStoreNtxBuilderClient =
+    StoreNtxBuilderClient<InterceptedService<Channel, OtelInterceptor>>;
+pub type InstrumentedStoreBlockProducerClient =
+    StoreBlockProducerClient<InterceptedService<Channel, OtelInterceptor>>;
+pub type InstrumentedStoreRpcClient = StoreRpcClient<InterceptedService<Channel, OtelInterceptor>>;
```
bobbinth: nit: the names here feel pretty verbose. Do we have non-instrumented versions of the clients? If not, I would consider dropping the `Instrumented` prefix - e.g., `InstrumentedRpcApiClient` -> `RpcApiClient`.

crates/rpc/src/client/api.rs
bobbinth: Can we add it as another method on `GrpcClientBuilder` - something like:

```rust
impl GrpcClientBuilder for Rpc {
    type Service = RpcApiClient;

    fn with_otel_interceptor(channel: Channel, _builder: &Builder) -> Self::Service {
        ...
    }

    fn with_metadata_interceptor(channel: Channel, _builder: &Builder) -> Self::Service {
        ...
    }
}
```
Or this wouldn't work for some reason?

SantiagoPittella: I added the metadata interceptor in the new module and removed the implementation from the rpc component, I left the re-export in case we want still use the client directly from the component's crate.

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+// TYPE ALIASES FOR INSTRUMENTED CLIENTS
+// ================================================================================================
+
+pub type RpcApiClient =
+    generated::rpc::api_client::ApiClient<InterceptedService<Channel, MetadataInterceptor>>;
+pub type BlockProducerApiClient =
+    generated::block_producer::api_client::ApiClient<InterceptedService<Channel, OtelInterceptor>>;
```
bobbinth: nit: why do `RpcApiClient` and `BlockProducerApiClient` have `Api` in them but the rest of the types do not? Should we make them just `RpcClient` and `BlockProducerClient`?

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+// BUILDER CONFIGURATION
+// ================================================================================================
+
+#[derive(Default, Clone)]
+pub struct Builder {
+    pub with_tls: bool,
+    pub address: String,
+    pub with_timeout: Option<Duration>,
+    pub metadata_version: Option<String>,
+}
```
bobbinth: Could we add some doc comments to the strut and to the fields as well?

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+        Ok(T::with_interceptor(channel, &self))
+    }
+
+    pub fn connect_lazy<T>(self) -> Result<T::Service>
+    where
+        T: GrpcClientBuilder,
+    {
+        let channel = self.build_endpoint()?.connect_lazy();
+        Ok(T::with_interceptor(channel, &self))
+    }
```
bobbinth: Could we add doc comments to these functions to explain the behavior of clients generated with them? (i.e., to make it clear why to use one vs. the other).

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+
+impl GrpcClientBuilder for Rpc {
+    type Service = RpcApiClient;
+
+    fn with_interceptor(channel: Channel, builder: &Builder) -> Self::Service {
+        // Use version from builder or default
+        let version = builder.metadata_version.as_deref().unwrap_or(env!("CARGO_PKG_VERSION"));
+        let interceptor = MetadataInterceptor::default()
+            .with_accept_metadata(version)
+            .expect("Failed to create metadata interceptor");
```
bobbinth: This will change a bit after #1084, right?

Also, we don't need to attach the `OtelInterceptor` here?

crates/block-producer/src/store/mod.rs
```diff
@@ -121,33 +120,33 @@ impl TryFrom<GetTransactionInputsResponse> for TransactionInputs {
...
     /// Creates a new store client with a lazy connection.
     pub fn new(store_address: SocketAddr) -> Self {
         let store_url = format!("http://{store_address}");
-        // SAFETY: The store_url is always valid as it is created from a `SocketAddr`.
-        let channel = tonic::transport::Endpoint::try_from(store_url).unwrap().connect_lazy();
-        let store = store_client::BlockProducerClient::with_interceptor(channel, OtelInterceptor);
+        let store = Builder::new()
+            .with_address(store_url)
+            .connect_lazy::<StoreBlockProducer>()
+            .expect("failed to connect to store"); // TODO: handle error
```
bobbinth: Do we need to address this and other similar `TODO`s?

bobbinth: nit (feel free to ignore): since `store_url` is used only once, we could probably write this as:
```rust
let store = Builder::new()
    .with_address(format!("http://{store_address}"))
    .connect_lazy::<StoreBlockProducer>()
    .expect("failed to connect to store");
```

crates/ntx-builder/src/store.rs
```diff
@@ -29,23 +25,22 @@ use crate::COMPONENT;
...
 /// Essentially just a thin wrapper around the generated gRPC client which improves type safety.
 #[derive(Clone, Debug)]
 pub struct StoreClient {
-    inner: InnerClient,
+    inner: StoreNtxBuilderClient,
 }
 
 impl StoreClient {
     /// Creates a new store client with a lazy connection.
     pub fn new(store_url: &Url) -> Self {
```
bobbinth, line 32: Why does the block producer take this parameter as `SocketAddr` but here we use `Url`? Should we make them consistent?

crates/rpc/src/client/api.rs
```diff
@@ -1,88 +1,2 @@
...
-    }
-}
-
-impl DerefMut for ApiClient {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.0
-    }
-}
+// Re-export the RPC client implementation from the proto crate
+pub use miden_node_proto::clients::{Builder, Rpc, RpcApiClient};
```
bobbinth: Who's the intended user of this re-export? i.e., is the idea to make it available to external users?

bobbinth:
After https://github.com/0xMiden/miden-node/pull/1045 got merged, we need to rebase this PR.

Also, is this PR largely ready for review - or is there something outstanding?

crates/proto/src/clients/mod.rs
SantiagoPittella: Yes, i'm changing it.

Regarding the `OtelInterceptor`, maybe @Mirko-von-Leipzig  can confirm but the current impl of the ApiClient doesn't have the OtelInterceptor.

SantiagoPittella: Just udpated the version checking format to the new one.

crates/ntx-builder/src/store.rs
SantiagoPittella: Not sure why the store in particular uses `Url` instead of `SocketAddr`, maybe we should standardize into `Url` to be more flexible and include protocol? cc @Mirko-von-Leipzig

crates/rpc/src/client/api.rs
SantiagoPittella: We were using it from the faucet, and in some places internally in the components. We can remove it if we decide to fully switch to new crate for all client things basically.

SantiagoPittella: I removed the re-exports in https://github.com/0xMiden/miden-node/pull/1075/commits/544706f3d765b652740cb10beade341e2f539818

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+
+impl MetadataInterceptor {
+    /// Adds or overwrites HTTP ACCEPT metadata to the interceptor.
+    ///
+    /// Provided version string must be ASCII.
+    pub fn with_accept_metadata(mut self, version: &str) -> Result<Self, anyhow::Error> {
+        let accept_value = format!("application/vnd.miden; version={version}");
+        self.metadata.insert("accept", AsciiMetadataValue::try_from(accept_value)?);
+        Ok(self)
+    }
```
bobbinth: Does this need to be updated to handle the `genesis` part of the accept header?

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+#[derive(Default, Clone)]
+pub struct Builder {
+    /// Whether to enable TLS encryption for the connection.
+    pub with_tls: bool,
+    /// The gRPC server address in the format `{protocol}://{hostname}:{port}`.
+    pub address: String,
+    /// Optional timeout for gRPC operations.
+    pub with_timeout: Option<Duration>,
+    /// Optional version string to include in request metadata.
+    pub metadata_version: Option<String>,
```
bobbinth: Similar question as above - does this need to handle the `genesis` part of the accept header somehow as well?

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+    pub fn with_address(mut self, address: String) -> Self {
+        self.address = address;
+        self
+    }
+
+    /// Establishes an eager connection to the gRPC server.
+    ///
+    /// This method attempts to connect to the server immediately and returns a client
+    /// only after the connection is established. If the connection fails, an error
+    /// is returned.
```
bobbinth: nit: here and in other places in this file it seems like we may be wrapping comments a bit too early (i.e., we usually do this around 100 chars.

crates/rpc/src/server/api.rs
```diff
@@ -27,50 +25,43 @@ use miden_objects::{
...
-            // SAFETY: The store_url is always valid as it is created from a `SocketAddr`.
-            let channel = tonic::transport::Endpoint::try_from(store_url).unwrap().connect_lazy();
-            let store = store_client::RpcClient::with_interceptor(channel, OtelInterceptor);
-            info!(target: COMPONENT, store_endpoint = %store_address, "Store client initialized");
+            // SAFETY: The store_url is always valid as it is a user-provided URL that has been
+            // validated.
+            let store = Builder::new()
+                .with_address(store_url.to_string())
+                .connect_lazy::<StoreRpc>()
+                .unwrap();
```
bobbinth: nit: here and in other similar places, I'd probably replace `unwrap()` with `expect()`.

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    fn with_interceptor(channel: Channel, builder: &Builder) -> Self::Service {
+        // Use version from builder or default
+        let version = builder.metadata_version.as_deref().unwrap_or(env!("CARGO_PKG_VERSION"));
+        let interceptor = MetadataInterceptor::default()
+            .with_accept_metadata(version)
+            .expect("Failed to create metadata interceptor");
+
+        generated::rpc::api_client::ApiClient::with_interceptor(channel, interceptor)
+    }
```
bobbinth: I understand that the current RPC client also doesn't have telemeter interceptor attached to it, but why is that? cc @Mirko-von-Leipzig.

bobbinth (approved):
Looks good! Thank you! I left a couple of nits and questions inline.

One other question: once we remove the faucet from this repo, what would be the primary use for the `Rpc` client? Or would it go away with the faucet as well?

crates/proto/src/clients/mod.rs
Mirko-von-Leipzig: This interceptor enables trace ID continuity across gRPC client/server boundaries. And should therefore only be used in places where we want this behaviour.

At some point we discussed this and decided that the faucet shouldn't be some special app where its trace's get connected to ours. It also wouldn't work without creating a special exemption for the faucet in the RPC server because there we would need to read this metadata and use it for the traces. But specifically only for _our_ faucet account because otherwise anyone could set the trace ID for their request in our telemetry which would be bad e.g. overwriting other traces etc.

I _think_ provers should be similarly treated as external apps for now? So those client's should also not have interceptors.

### If we do want this

I think it would be nice to have these trace's connected but it would require a bit of extra work.

- Faucet client need's to support this injection
  - This needs to be configurable so that not all user faucet's have this enabled as they probably don't want it.
- RPC needs to _know_ the faucet IP/identifier to special case it.
- Prover's should also configurably have this enabled
  - Once we get separate internal instances this is easy
  - Until then we would need them to know the IP/identifier of our components to special case them

crates/proto/src/clients/mod.rs
SantiagoPittella: I'm not sure, since this is basically used for tests, that this needs the genesis part. Maybe @Mirko-von-Leipzig can confirm.

crates/proto/src/clients/mod.rs
SantiagoPittella: ~~Since it is used in tests, the measurements aren't collected thus making the interceptor useless here~~

EDIT: Mirko's answer wasn't shown until I refreshed (after my comment).

bobbinth: My understanding is that the RPC client is currently used for two things:

- The faucet - but this will go away soon.
- The tests - for which we probably don't need to enable this functionality.

So, I think this answers my question - we don't really need this for the RPC client.

crates/proto/src/clients/mod.rs
bobbinth: I'm assuming if the genesis is not specified, we'll accept any genesis. If this is correct, I think this is probably fine for tests, and so we can merge. cc @Mirko-von-Leipzig to confirm.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 - Added environment variable support for batch and block size CLI arguments ([#1081](https://github.com/0xMiden/miden-node/pull/1081)).
 - A transaction's account delta is now checked against its commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
+- Add client builders for all components ([#1075]((https://github.com/0xMiden/miden-node/pull/1075))).
```
Mirko-von-Leipzig: Internal only change


```suggestion
```

crates/rpc/src/server/mod.rs
```diff
@@ -35,7 +34,9 @@ impl Rpc {
     /// Note: Executes in place (i.e. not spawned) and will run indefinitely until
     ///       a fatal error is encountered.
     pub async fn serve(self) -> anyhow::Result<()> {
-        let api = api::RpcService::new(self.store, self.block_producer);
+        let store_url = self.store.clone();
+        let block_producer_url = self.block_producer.clone();
+        let api = api::RpcService::new(&self.store, self.block_producer);
```
Mirko-von-Leipzig: Are the clones required? Think they're only used in the later `info!`.

crates/ntx-builder/src/store.rs
Mirko-von-Leipzig: `Url` everywhere is fine. Let's use a single method; I'm somewhat unhappy with `Url` but so far `Uri`, `Url`, `String`, `Endpoint`, `SocketAddr` all have their own warts.

So long as the builder only takes the one input at least its centralized and easy to change.

crates/proto/src/clients/mod.rs
Mirko-von-Leipzig: Yes it does; the genesis part was added after this PR was opened I think 😅

crates/proto/src/clients/mod.rs
Mirko-von-Leipzig, line 1: I think it would be good to force the caller to make a decision about every config item i.e. no defaults allowed.

wdyt about something like this

```rust
pub struct Builder<T> {
    endpoint: Endpoint,
    accept: Option<String>,
    _state: PhantomData<T>,
}

pub struct WantsTls;
pub struct WantsTimeout;
pub struct WantsVersion;
pub struct WantsGenesis;
pub struct WantsConnection;

impl<T> Builder<T> {
    // And whatever constructors we find useful.
    pub fn new(url: impl AsRef<str>) -> Result<Builder<WantsTls>>, Error> {
        Builder {
            endpoint: Endpoint::from_str(url.as_ref())?,
            accept: None,
            _state: PhantomData,
        }
    }

    /// Convenience function to cast the state type.
    fn next_state<U>(self) -> Builder<U> {
        Builder {
            endpoint: self.endpoint,
            accept: self.accept,
            _state: PhantomData::<U>
        }
    }
}

impl Builder<WantsTls> {
    pub fn without_tls(self) -> Builder<WantsTimeout> {
        self.next_state()
    }

    pub fn with_tls(self) -> Result<Builder<WantsTimeout>, Error> {
        self.endpoint.tls_config(...)?;
        self.next_state()
    }
}

impl Builder<WantsTimeout> {
    pub fn without_timeout(self) -> Builder<WantsConnection> {...}
}

// etc
```

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,347 @@
...
+// ================================================================================================
+
+/// Trait for building gRPC clients from a common [`Builder`] configuration.
+///
+/// This trait provides a standardized way to create different gRPC clients with consistent
+/// configuration options like TLS, OTEL interceptors, and connection types.
+pub trait GrpcClientBuilder {
+    type Service;
+
+    fn with_interceptor(channel: Channel, builder: &Builder) -> Self::Service;
```
Mirko-von-Leipzig: This isn't quite right, because we may want to combine interceptors 🤔 

As in, internal gRPC should support both OTel _and_ the accept header interceptor..

I'm not entirely sure how to solve this tbh. Maybe we move on for now and cross that bridge when we get there.

crates/rpc/src/server/mod.rs
SantiagoPittella: They shouldn't be required. I'm moving the `info!` before the usage of the urls (and changing a bit the message).

crates/proto/src/clients/mod.rs
SantiagoPittella: Done!

crates/proto/src/clients/mod.rs
SantiagoPittella: Maybe we can use a new interceptor that wraps the 2 interceptors (otel + metadata). Something like `Interceptor = InterceptedService<InterceptedService<Channel, Otel>, Metadata>`, will see what can be done.

SantiagoPittella: I added an approach to this in the last commit.

crates/proto/src/clients/mod.rs
```diff
@@ -365,7 +181,193 @@ impl GrpcClientBuilder for StoreBlockProducer {
...
+        self.tls_enabled = Some(false);
+        self.next_state()
+    }
+
+    /// Explicitly enable TLS.
+    pub fn with_tls(mut self) -> Result<Builder<WantsTimeout>> {
+        // We only record the decision here; the actual TLS wiring is performed by the underlying
+        // `Builder` during `connect()` to keep the logic centralized.
+        self.tls_enabled = Some(true);
+        Ok(self.next_state())
```
Mirko-von-Leipzig: I would put this into the builder now already. This removes the result from `lazy_connect` for example.

crates/ntx-builder/src/store.rs
```diff
@@ -22,23 +19,27 @@ use crate::COMPONENT;
...
-        let store = store_client::NtxBuilderClient::with_interceptor(channel, OtelInterceptor);
+        // SAFETY: The store_url is always valid as it is created from a `Url`.
+        let store = Builder::<WantsTls>::new(store_url.to_string())
+            .expect("Failed to initialize store endpoint")
+            .without_tls()
+            .without_timeout()
+            .without_metadata_version()
+            .without_metadata_genesis()
+            .connect_lazy::<StoreNtxBuilder>()
+            .expect("Failed to build client");
```
bobbinth: A superficial comment: `WantsTls` and then `.without_tls()` seems somewhat contradictory.

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,396 @@
+//! gRPC client builder utilities for Miden node.
+//!
+//! This module provides a unified type-state [`Builder`] for creating various gRPC clients with
+//! explicit configuration decisions for TLS, timeout, and metadata.
```
bobbinth: nit: should "type-state" be "type-safe"?

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+impl GrpcClientBuilder for StoreRpc {
+    type Service = StoreRpcClient;
+
+    fn with_interceptor(channel: Channel, _config: &ClientConfig) -> Self::Service {
+        generated::rpc_store::rpc_client::RpcClient::with_interceptor(channel, OtelInterceptor)
+    }
+}
+
+// STRICT TYPE-STATE BUILDER (NO DEFAULTS)
+// ================================================================================================
```
bobbinth: Same nit as above: should `TYPE-STATE` be `TYPE-SAFE`?

bobbinth (approved):
Looks good! Thank you! This is a pretty high-level review from me and I just left a couple of nits inline.

Overall, what is left to do in this PR? Should anyone else look it over?

crates/proto/src/clients/mod.rs
Mirko-von-Leipzig: I would still prefer using `Endpoint` in the builder struct, instead of rebuilding it in `connect` and `connect_lazy`.

SantiagoPittella: Just changed it!

crates/ntx-builder/src/store.rs
SantiagoPittella: I removed the turbofish and it is just `Builder::new`

bin/node/src/commands/bundled.rs
```diff
@@ -219,9 +230,13 @@ impl BundledCommand {
         if should_start_ntb {
             let id = join_set
                 .spawn(async move {
+                    // SAFETY: The block_producer_address is always valid as it is created from a
```
drahnr: nit: Since we're using `anyhow` anyways, it might be shorter to use `Url::parse(..).context("foo")?`.

crates/ntx-builder/src/store.rs
```diff
@@ -22,23 +19,26 @@ use crate::COMPONENT;
 // STORE CLIENT
 // ================================================================================================
 
-type InnerClient = store_client::NtxBuilderClient<InterceptedService<Channel, OtelInterceptor>>;
```
drahnr, line 25: Yes, 👍

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+#[derive(Copy, Clone, Debug)]
+pub struct WantsTls;
+#[derive(Copy, Clone, Debug)]
+pub struct WantsTimeout;
+#[derive(Copy, Clone, Debug)]
+pub struct WantsVersion;
+#[derive(Copy, Clone, Debug)]
+pub struct WantsGenesis;
+#[derive(Copy, Clone, Debug)]
+pub struct WantsConnection;
```
drahnr, line 255: 💟

crates/proto/src/clients/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+///     .with_timeout(Duration::from_secs(5)) // or `.without_timeout()`
+///     .with_metadata_version("1.0".into()) // or `.without_metadata_version()`
+///     .without_metadata_genesis()           // or `.with_metadata_genesis(genesis)`
+///     .connect::<Rpc>()
+///     .await?;
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Clone, Debug)]
+pub struct Builder<State> {
```
drahnr, line 239: note: we imply an ordering here, and in the associated functions, using a generic per assoticated function with a corresponding `Has*` to its `Wants*`, one could make the order arbitrary, which is what I'd expect as user. It comes with a compile time cost of extra generics and a bag of generated types. It's a trade-off.

drahnr: *At some point this becomes art (or so)

drahnr (approved):

crates/proto/src/clients/mod.rs
drahnr: I think `type-state` is correct in this context, the state is typed :)

crates/proto/src/clients/mod.rs
Mirko-von-Leipzig: I considered that, but figured from the user perspective its all opaque so let's just have less generics.

crates/rpc/src/server/mod.rs
```diff
@@ -25,8 +24,8 @@ mod api;
 /// Requests will fail if the components are not available.
 pub struct Rpc {
     pub listener: TcpListener,
-    pub store: SocketAddr,
-    pub block_producer: Option<SocketAddr>,
+    pub store: Url,
+    pub block_producer: Option<Url>,
```
igamigo: nit: Seems like the other components have a `*_url` suffix now, maybe those could be taken out or we could add it to the `Rpc` component as well

crates/rpc/src/server/mod.rs
```diff
@@ -57,8 +58,6 @@ impl Rpc {
             .build_v1alpha()
             .context("failed to build reflection service")?;
 
-        info!(target: COMPONENT, endpoint=?self.listener, store=%self.store, block_producer=?self.block_producer, "Server initialized");
```
igamigo, line 65: Should this or something similar be brought back? Feels like it would be interesting from a debugging perspective to know when it got to load the genesis, etc.

crates/ntx-builder/src/block_producer.rs
```diff
@@ -1,44 +1,51 @@
...
+        // validated.
+        let block_producer = Builder::new(block_producer_url.to_string())
+            .expect("Failed to initialize block-producer endpoint")
+            .without_tls()
+            .without_timeout()
+            .without_metadata_version()
+            .without_metadata_genesis()
+            .connect_lazy::<BlockProducer>();
+
+        info!(target: COMPONENT, block_producer_endpoint = %block_producer_url, "Block producer client initialized");
```
igamigo: nit: I think these were like this before, but I wonder how much value these add. The client was created but it connects lazily, right? So it would effectively always initialize correctly. Maybe not something to change, but I think it could look misleading in logs unless you always keep in mind that these are lazy clients.

igamigo (approved):
LGTM! Left a couple of comments mostly related to logs

crates/rpc/src/server/mod.rs
SantiagoPittella: Maybe adding the suffix here is the way to go, it is not a large change and adds meaning 👌🏼

crates/rpc/src/server/mod.rs
SantiagoPittella: Sure, i can rollback this change to re-add it.

crates/ntx-builder/src/block_producer.rs
SantiagoPittella: I've changed the message to better reflect this. What do you think?

PR #1081, opened 2025-07-12
feat(block-producer): increase and make max txs per batch and batches per block configurable via CLI

farrellh1:
- Adds CLI arguments to configure the maximum number of transactions per batch (--max-txs-per-batch) and the maximum number of batches per block (--max-batches-per-block) for both standalone and bundled node modes.
- Updates the default values to match previous behavior (8 per batch, 8 per block), but allows users to override them at runtime.


Mirko-von-Leipzig (approved):
Thank you!

Issue #1072, opened 2025-07-10
gRPC extension traits


Mirko-von-Leipzig:
We sometimes extend our gRPC client functionality to include things like automatically retrying on connection failures. It would be convenient if we could centralize these implementations as an extension method which is callable on _any_ gRPC client function call.

This isn't trivially possible right now because gRPC methods don't all implement a specific trait or type so one can't target them. Part one of this issue would therefore involve prototyping a few approaches and deciding whether its worth it or not.

I'll list a couple of ideas I've had bouncing around here, that doesn't necessarily mean they're good ones.

### Standardize the return result

If we standardize gRPC methods to return `GrpcResult<T>` then we could implement the extension trait on that. At minimum we need the error type to return whether it was a connection issue, or not, so we know when to retry.

This would be similar to tokio's `timeout` extension method, which wraps the future in a timeout.

### `trait GrpcMethod`

This would be more involved, but would let us standardize and add nicer telemetry to operations. In general, our gRPC methods all broadly follow the process of deserializing request, do some work, serialize the response.

Using a trait isn't required, but it would let us standardize the error types and process. However, there are still unknowns to me, like how would one inject the actual protobuf client, and which method to call on it. This should be possible, but I haven't fully explored this yet.

⚠️ Reminder that this may still result in bad ergonomics in the end..

```rust
enum GrpcDecodingError {..}

enum GrpcError {
    Decoding(GrpcDecodingError),
    ...
}

trait GrpcDecode<Raw, Domain> {
    fn decode(proto: Raw) -> Result<Self, GrpcDecodingError> {...}
}

trait GrpcEncode<Raw, Domain> {
    fn encode(domain: Domain) -> Raw {...}
}

trait GrpcMethod {
    // Raw protobuf inputs and outputs
    type Request;
    type Response;

    // Actual domain inputs and outputs.
    //
    // Must satisfy encoding/decoding from the protobufs.
    type Input: GrpcDecoding<Self::Request, Self::Input>;
    type Output: GrpcEncoding<Self::Response, Self::Output>;
    
    async fn call(&self, client: T, input: Self::Input) -> Result<Self::Output, GrpcError> {
        // We can add a tracing span for each step here, so we know
        // how long encoding, work, and decoding take.
        let request = input.encode();
        let response = client.call(request).await?; // Client bit is unclear.
        Self::Output::decode(response)
    }
}
```

Mirko-von-Leipzig:
Completed in #1950

PR #1084, opened 2025-07-14
feat(rpc): network verification via commitment header

### LINKED ISSUE #1066: gRPC genesis check
Add genesis commitment check to the gRPC header check in the RPC component.

# Context

We added similar support in #844 which checks that a client is using a compatible gRPC schema.

However, it also occasionally happens that a client with existing state is re-used on the wrong network. We can check the network via the genesis commitment.

Client's can fetch the commitment when they are initialized, and then add it as part of the HTTP header. On the node end, the RPC component can load it on start-up and trivially verify it there. This could function similarly for inter-component communication.

We could also use something like a network name, but the genesis commitment would be more unique. This does mean this wouldn't work for forks of networks; though if we wanted to support that sort of thing we wouldn't have to restrict it to the genesis block, but rather something like the latest epoch, or some other sort of anchoring block.

Relates to #1040.

Mirko-von-Leipzig:
This PR adds support for verifying the network identity for RPC clients.

This PR also changes the existing format for the version checking:

```
# Original
application/vnd.miden+grpc.x.y.z

# New
application/vnd.miden; version=x.y.z; genesis=0xABC
```

The now obsolete format gets rejected now which I think is okay. The new format still allows for an optional `+grpc` suffix.

The original version checking behavior is retained, though we may want to consider switching to `Caret` instead of `Exact` but that's separate from this addition.

This PR also improves the correctness of the accept header verification by:

- accepting wildcards `*/*` and `application/*` media-types
- rejecting zero-weighted media-types `q=0`

~Similar to the RPC version check added in #844, we now check the network the client thinks its on. The client can now set the genesis block's commitment in the `miden-network` header, and the RPC service will reject the request if there is a mismatch.~

~This does add an additional header, and maybe instead we should add it to the `accept` header? I'm just unsure what this format should look like; and we probably want to do additional parsing instead.~

Closes #1066.


crates/rpc/src/server/accept.rs
```diff
@@ -162,61 +193,88 @@ impl<'a> TryFrom<&'a str> for AcceptHeaderValue<'a> {
...
 
-    #[test]
-    fn same_minor_different_patch_matches() {
-        let a = AcceptLayer::new().unwrap();
-        let version = env!("CARGO_PKG_VERSION");
-        let mut version = Version::parse(version).unwrap();
-        version.patch += 1;
-        assert!(a.version_req.matches(&version));
+    impl HeaderVerificationLayer {
+        fn for_tests() -> Self {
```
sergerad, line 313: can we `#[cfg(test)]` this fn?

crates/rpc/src/server/accept.rs
```diff
@@ -162,61 +193,88 @@ impl<'a> TryFrom<&'a str> for AcceptHeaderValue<'a> {
...
-    fn valid_accept_header_is_parsed() {
-        let input = "application/vnd.miden.1.0.0+grpc";
-        let expected = AcceptHeaderValue {
-            app_name: "miden",
-            version: "1.0.0",
-            response_type: "grpc",
-        };
-        assert_eq!(AcceptHeaderValue::try_from(input), Ok(expected));
+        let uut = HeaderVerificationLayer::for_tests();
+        uut.verify(&headers).unwrap_err();
```
sergerad: Could we test the exact error variants coming out? Given you have created the error enums now.

crates/rpc/src/server/api.rs
```diff
@@ -88,6 +91,54 @@ impl RpcService {
 
         Self { store, block_producer }
     }
+
+    /// Fetches the genesis block header from the store.
+    ///
+    /// Automatically retries until the store connection becomes available.
+    pub async fn get_genesis_header_with_retry(&self) -> anyhow::Result<BlockHeader> {
+        let mut retry_counter = 0;
+        loop {
```
sergerad, line 99: Looks like we are now doing this pattern 4x in the node repo now. Wondering if we could implement a utility fn that runs a closure with retries.

crates/rpc/src/server/api.rs
```diff
@@ -88,6 +91,54 @@ impl RpcService {
...
+                        .min(Duration::from_secs(30));
+
+                    tracing::warn!(
+                        ?backoff,
+                        %retry_counter,
+                        %err,
+                        "connection failed while subscribing to the mempool, retrying"
+                    );
+
+                    retry_counter += 1;
```
sergerad, line 134: Does this loop forever until it succeeds? Should we add a limit and error out instead? This would be something handled by the utility fn I mentioned  [here](https://github.com/0xMiden/miden-node/pull/1084/files#r2205996865).

crates/rpc/src/server/accept.rs
```diff
@@ -1,61 +1,146 @@
...
 
-use crate::COMPONENT;
-
-/// Layer responsible for handling HTTP ACCEPT headers.
+/// Rejects requests whose headers indicate a mismatch in RPC version or the network.
+///
+/// The former is taken from the `accept` header with the expected format
+/// `application/vnd.miden.<major>.<minor>.<patch>+grpc`.
+///
+/// The latter expects the network's genesis block's commitment in the `miden-network` header.
```
sergerad: Wondering if we should just re-use the accept header for this. If we plan on putting more things like this in place, I don't think we should. But if this is the only other verification for client/server/network matching, then we could integrate it e.g.
```
application/vnd.miden.<network>.<major>.<minor>.<patch>+grpc

File: crates/rpc/src/server/mod.rs [hunk 6]
```diff
@@ -55,10 +61,14 @@ impl Rpc {
 
         info!(target: COMPONENT, endpoint=?self.listener, store=%self.store, block_producer=?self.block_producer, "Server initialized");
 
+        let rpc_version = env!("CARGO_PKG_VERSION");
+        let rpc_version =
+            semver::Version::parse(rpc_version).context("failed to parse crate version")?;
```
bobbinth, line 66: We also use the package version for the `status` endpoint. Should we centralize this somehow? For example, should we put the version into the `RpcService` struct?

A somewhat related thought: should we also include the `genesis` (or `genesis_commitment`) field into the response from the `status` endpoint?

crates/rpc/src/server/mod.rs
```diff
@@ -38,6 +38,12 @@ impl Rpc {
     ///       a fatal error is encountered.
     pub async fn serve(self) -> anyhow::Result<()> {
         let api = api::RpcService::new(self.store, self.block_producer);
+
+        let genesis = api
+            .get_genesis_header_with_retry()
+            .await
+            .context("Fetching genesis header from store")?;
```
bobbinth, line 45: If I'm understanding correctly, this will naturally prevent the RPC from serving requests until the store is up and running, correct?

bobbinth (commented):
Thank you! Looks good! Not a review but I did leave a couple of questions/comments inline.

> This does add an additional header, and maybe instead we should add it to the `accept` header? I'm just unsure what this format should look like; and we probably want to do additional parsing instead.

I haven't thought about it too much, but it does feel like maybe handling everything through the `accept` header is a bit more "natural" - i.e., we have a single header which tells us if the request should be accepted or rejected.

On the other hand, this probably complicates parsing a bit and makes it more difficult to handle partial scenarios (e.g., missing version info but present genesis commitment  or vise-versa). But it is also not clear if we should care about partial scenarios - and in that case, maybe channeling everything through the `accept` header is a better option.

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: It already is because its in `mod tests`

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: #1072 contains some of my thoughts on this; but yeah its something I'm hoping will fall out of all the gRPC refactoring.

crates/rpc/src/server/mod.rs
Mirko-von-Leipzig: That is correct yeah.

crates/rpc/src/server/mod.rs
Mirko-von-Leipzig: > We also use the package version for the status endpoint. Should we centralize this somehow? For example, should we put the version into the RpcService struct?

I don't think this is complicated enough to really warrant de-duping it. And in this case its required in a layer and not as part of the service itself so this wouldn't really change much.

> A somewhat related thought: should we also include the genesis (or genesis_commitment) field into the response from the status endpoint?

We can; would be less fluff than fetching the genesis block's header specifically which would be nice for manual `grpcurl` style checks. #1088

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: We can but we can't do this in a backwards compatible way, or at least not without complicating things.

This would mean we can't do either of them optionally. I'm not clued up enough on the allowed format of accept headers. Maybe if one appends it?

```
application/vnd.miden.<major>.<minor>.<patch>+grpc+network=<network>
```

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: It does loop indefinitely.

My thought was that there isn't anything we can change except retry until it works. It logs on `WARN` so it shouldn't really be ignored.

We have no idea why the store isn't available; it could take a long time after a lot of blocks for example to build the MMR on startup.

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: We can; but I'm not sure that adds any specific value?

For example, in the case where both mismatch, which error should it be? Could be either..

crates/rpc/src/server/mod.rs
bobbinth: > I don't think this is complicated enough to really warrant de-duping it. And in this case its required in a layer and not as part of the service itself so this wouldn't really change much.

My thinking here was that maybe in the future the version we'd want to use is not the same as the version of the crate (e.g., we could have "protocol versions" which are different from crate versions) and so it would be nice to have a central place where it is defined. But also, it is probably something we can do if/when we actually use something other than the crate version.

crates/rpc/src/server/accept.rs
sergerad: > We can but we can't do this in a backwards compatible way, or at least not without complicating things.

Yea the new server would need to support both formats. At this point we are more likely to opt for breaking approach.

3 different robots told me this is idiomatic (specify parameter with `;`):
```
Accept: application/vnd.miden.<major>.<minor>.<patch>+grpc; network=<network>
```

They cited some RFCs and I did see an example `Accept: text/plain; q=0.5, text/html,` which does show parameter usage with `;` and separate MIME type listing with `,`.

crates/rpc/src/server/api.rs
sergerad: Yea I've always been in the practice of crash looping with exponential backoff. Thats definitely in part due to Kubernetes behaviour of course. But it does also help alerting.

A practical issue with relying on infinite loop + WARN log line is that our alerting might need to rely on the log line content. Which we would obviously want to avoid (having to maintain consistency between logs and alerts is not good).

I imagine it would be easier / more reliable to have an alert on a crash loop. But if we are confident that we would have a reliable alert despite this infinite loop without a crash (without relying on log content), then its fine as is.

crates/rpc/src/server/accept.rs
sergerad: > For example, in the case where both mismatch, which error should it be? Could be either..

It would be well defined which error would be returned if both mismatched. Checking the error proves our logic is working precisely as expected. We could erroneously return invalid semver variant for an input with a correct semver for example. Unless I'm missing something.

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: Its a bit more complex than that iiuc. The intention of the header is that clients supply a comma separated list of media options they can handle and the server should select _one_ of these as the response.

The format is

```
<media_type>/<subtype>; q=<quality-weighting>, ...
```

`q` is intended to let the user set priorities, a higher value means they prefer this content type.

So having two separate "media type" as part of the header would be incorrect afaict, since they're `OR` and not `AND` interpreted. More likely its the `+grpc` thing we should be following somehow.

Mirko-von-Leipzig: As an example, we (at some point) should allow clients to specify:

```
# Multiple versions if the client supports that somehow. The client prefers 0.9.3 over 0.10.4.
application/vnd.miden.0.9.3+grpc; q=0.5, application/vnd.miden.0.10.4+grpc; q=0.3

# Version ranges (maybe)
application/vnd.miden.0.10.*+grpc
```

See mozilla's [example](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept#media-type) on `image/*`.

crates/rpc/src/server/api.rs
sergerad: One more thought on this - with the loop + WARN approach we would need individual alerts for each such situation. Whereas crash loops would be alerted on by my general alerts that are not so implementation-aware.

crates/rpc/src/server/accept.rs
sergerad: > So having two separate "media type" as part of the header would be incorrect afaict

Yep.

> q is intended to let the user set priorities, a higher value means they prefer this content type.

Is this to say that `; network=<..>` would be inappropriate?

bobbinth: Is `network` self-explanatory enough here? maybe it should be `genesis=0x123...`? Though, probably doesn't matter too much.

sergerad: Yea network is a bit misleading. Genesis is better.

Mirko-von-Leipzig: Ah I had to read the full specification a bit more carefully. It gives an [example](https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.1-14):

```
text/plain;format=flowed
text/plain;format=fixed;q=0.4
```

so we can use extra args so to speak. I'll change this PR to use the genesis suffix in the header instead.

Should we change the version to also be a "suffix" arg? e.g. `application/vnd.miden+grpc;version=<x.y.z>;genesis=<0xABC>`

sergerad: Yes I think that is practical for us and also idiomatic as far as I could tell from the relevant RFCs.

crates/rpc/src/server/accept.rs
```diff
@@ -75,68 +160,14 @@ where
...
-        };
-
-        // Parse the accept str
-        let accept = match AcceptHeaderValue::try_from(accept_str) {
-            Ok(value) => value,
-            Err(e) => {
-                // If the accept header value is different from expected, don't enforce the version
-                // requirement.
-                debug!(target: COMPONENT, "Failed to parse accept header value: {}", e);
-                return self.inner.call(request).boxed();
```
tomyrd, line 99: In the old version the accept header was not rejected when it failed to parse. This is not the same in the new version and caused a specific issue with the web client.
Apparently the web client sends an initial request with a `*/*` accept header which is causing the error. [We noticed this ](https://github.com/0xMiden/miden-client/pull/922#issuecomment-2910454361) when we originally added the accept header. Could we add this behavior to the new version? Or maybe add a special case for `*/*`.

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: I've still left this as is. I don't know how long one should loop for - and its a bit difficult to set this appropriately due to unknown start-up times of the other components.

I agree in principle, but I don't know how to specify this in a practical sense currently.

crates/rpc/src/server/accept.rs
```diff
@@ -1,61 +1,168 @@
...
+/// is specified as the genesis block's commitment. If the server cannot satisfy either of these
+/// constraints then the request is rejected.
+///
+/// Note that both values are optional, as is the header itself. If unset, the server considers
+/// any value acceptable.
+///
+/// As part of the accept header's standard, all media ranges are examined in quality weighting
+/// order until a matching content type is found. This means that the client can set multiple
+/// `application/vnd.miden` values and each will be tested until one passes. If none pass, the
+/// request is rejected.
```
drahnr, line 27: Q: Should we limit the number of these entries?

crates/rpc/src/server/accept.rs
```diff
@@ -1,61 +1,168 @@
...
+                }
+            }
+
+            // Verify the user's genesis commitment against ours.
+            if let Some(genesis) = range.params.get(&CaselessKey::GENESIS) {
+                let genesis = Word::try_from(*genesis).map_err(|err| {
+                    AcceptHeaderError::GenesisParsing(err, (*genesis).to_string())
+                })?;
+
+                if genesis != self.genesis_commitment {
```
drahnr: Should we reject those referencing a wrong genesis rather than ignoring, or do we anticipate someone crafting requests with a set of headers for different genesis'?

crates/rpc/src/server/accept.rs
```diff
@@ -75,148 +182,232 @@ where
...
 
+    #[rstest::rstest]
+    #[case::empty("")]
+    #[case::wildcard("*/*")]
+    #[case::wildcard_subtype("application/*")]
+    #[case::media_type_only("application/vnd.miden")]
+    #[case::with_quality("application/vnd.miden; q=0.3")]
+    #[case::version_exact("application/vnd.miden; version==1.2.3")]
+    #[case::version_major("application/vnd.miden; version=1")]
+    #[case::version_minor("application/vnd.miden; version=1.2")]
```
drahnr: Given the current implementation, I'd a variant with spaces around `=`

crates/rpc/src/tests.rs
```diff
@@ -77,14 +81,7 @@ async fn rpc_server_rejects_requests_with_accept_header_invalid_version() {
...
         assert_eq!(response.as_ref().err().unwrap().code(), tonic::Code::InvalidArgument);
-        assert!(
-            response
-                .as_ref()
-                .err()
-                .unwrap()
-                .message()
-                .contains("Client / server version mismatch"),
-        );
+        assert!(response.as_ref().err().unwrap().message().contains("server does not support"),);
```
drahnr, line 95: nit: I find the output of `assert_matches::assert_matches!` much easier to interpret if any issues are encountered

crates/rpc/src/server/accept.rs
```diff
@@ -75,148 +182,232 @@ where
...
+        uut.verify(&headers).unwrap();
     }
 
+    #[rstest::rstest]
+    #[case::backwards_incompatible_with_version("application/vnd.miden+grpc.1.2.3")]
+    #[case::backwards_incompatible("application/vnd.miden+grpc")]
+    #[case::invalid_version("application/vnd.miden; version=0x123")]
+    #[case::invalid_genesis("application/vnd.miden; genesis=aaa")]
+    #[case::version_too_old("application/vnd.miden; version=>1.2.3")]
+    #[case::version_too_new("application/vnd.miden; version=<1.2.3")]
```
drahnr: Just to rubberducky this:

The situations we have, is a client and a server in the broadest sense. Neither can know what breakage a future version will do, so the latest known version at the point of the release is the mutual cut-off point for guaranteed compatibility.
This would logically shift the responsibility of compatibility checking to the "newer"/more recently released party, since only that one can have knowledge about the "older" other party.

1. server newer than client --> which is in scope for this PR.
2. client newer than server --> server cannot decide if this would work, so this is out of scope for this PR 


In the scope of 1., I would have anticipated a single version being sent to the server, and the server can check compatibility.
The client might know about a minimum server version it requires, but cannot know about a maximum version.
If I think about use cases for `version<=` I can only see patch releases as usecase.

Is that interpretation correct?

crates/rpc/src/server/accept.rs
```diff
@@ -1,61 +1,168 @@
...
-    /// Panics if the version string in Cargo.toml is not valid semver. The version string is made
-    /// into an env var at compile time which means that the unit tests prove this cannot panic
-    /// in practice.
-    pub fn new() -> anyhow::Result<Self> {
-        // Parse the full version string (e.g. "0.8.0").
-        let version = env!("CARGO_PKG_VERSION");
-        let version = Version::parse(version)?;
+#[derive(Debug, thiserror::Error)]
+enum AcceptHeaderError {
+    #[error("Header value could not be parsed as a UTF8 string")]
```
igamigo:
```suggestion
    #[error("header value could not be parsed as a UTF8 string")]
```

crates/rpc/src/server/accept.rs
```diff
@@ -75,148 +182,232 @@ where
...
+    quality: Quality,
+}
+
+#[derive(Debug)]
+enum MediaType<'a> {
+    Wildcard,
+    Type(&'a str),
+}
+
+#[derive(Debug, Eq, Hash)]
```
igamigo: nit: `CaselessKey("q")` and `CaselessKey("Q")` would hash to different things (but comparing them would return `true`) which might not be an issue here but should probably get changed. Maybe you were keeping this in mind but just noting it just in case.

crates/rpc/src/server/accept.rs
```diff
@@ -75,148 +182,232 @@ where
...
+            MediaType::Type(media) => media.eq_ignore_ascii_case(value),
+        }
+    }
 
-    futures::future::ready(Ok(response))
+    fn starts_with(&self, prefix: &str) -> bool {
+        match self {
+            MediaType::Wildcard => false,
+            MediaType::Type(media) => {
+                media[..prefix.len().min(media.len())].eq_ignore_ascii_case(prefix)
```
igamigo, line 254: Would it be worth making `CaselessKey` transform the inner key into lowercase when instantiated? Then this comparison could read simpler (I think): `media.as_bytes().starts_with(prefix.as_bytes())`

crates/rpc/src/server/accept.rs
```diff
@@ -75,148 +182,232 @@ where
...
+    #[case::media_type_only("application/vnd.miden")]
+    #[case::with_quality("application/vnd.miden; q=0.3")]
+    #[case::version_exact("application/vnd.miden; version==1.2.3")]
+    #[case::version_major("application/vnd.miden; version=1")]
+    #[case::version_minor("application/vnd.miden; version=1.2")]
+    #[case::version_wildcard_major("application/vnd.miden; version=*")]
+    #[case::version_wildcard_minor("application/vnd.miden; version=1.*")]
+    #[case::version_wildcard_patch("application/vnd.miden; version=1.2.*")]
+    #[case::matching_network(
+        "application/vnd.miden; network=0x00000000000000000000000000000000000000000000000000000000deadbeef"
```
igamigo: Should this be `genesis=`?


```suggestion
        "application/vnd.miden; genesis=0x00000000000000000000000000000000000000000000000000000000deadbeef"
```

I guess it works anyway because it's not validated in any way, but not sure if that was the intention.

igamigo (approved):
Overall LGTM

crates/rpc/src/server/accept.rs
sergerad: I had a crack at adding the expected error variants into the cases but I think its blocked on variants like 
```rust
    VersionParsing(#[source] semver::Error, String),
```

Because `semver::Error` and its constructors etc are not public so we can't recreate them for the tests.

sergerad (approved):
LGTM

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: What I was hoping this would add is the ability to exclude a buggy patch or version. But looking at [`VersionReq`](https://docs.rs/semver/latest/semver/struct.VersionReq.html) again, it uses comma's to separate definitions so I think this won't work at all :/

e.g. I was hoping to let client's skip buggy patch x.y.z by doing:

```
application/vnd.miden; version=>x.y.0, version<x.y.z,
application/vnd.miden; version=>x.y.z
```

This would let client's be more specific, but its a bit of a narrow use case so maybe I should revert to just `Version` since the `,` kinda screw it up.

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: Thanks; definitely wasn't the intention. Really wish the `Accept` let us reject unknown headers :sweat: Like I know one shouldn't but silently accepting things feels wrong.

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: I really want to. But technically its perfectly legal to combine media ranges like that.

I can't really think of a reason anyone might do so though, so maybe its okay to do so.

@igamigo thoughts?

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: Maybe this is a problem that gets solved by adding the server's RPC version to the content-type response header? Then the client can do as it wishes with that.

And the server can do normal semver version selection/rejection.

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: Ah damn, nice catch. I've refactored this entirely so we instead only worry about the parameters we care about i.e. `quality`, `version`, and `genesis`.

Mirko-von-Leipzig:
Having looked into the specification some more, parsing gets really messy. I've since found a crate that actually adheres to the standard (or at least more so than the previous one's I'd discarded). I'll try replace my "parser" with this crate to judge if it might work.

For example, my parser doesn't handle quoted parameter values -- which notably would allow `,` inside the value. There are a bunch of these.

bobbinth:
> Having looked into the specification some more, parsing gets really messy. I've since found a crate that actually adheres to the standard (or at least more so than the previous one's I'd discarded). I'll try replace my "parser" with this crate to judge if it might work.
> 
> For example, my parser doesn't handle quoted parameter values -- which notably would allow `,` inside the value. There are a bunch of these.

We can also bring the scope of this PR back to the original issue to just add the `genesis` part to the `ACCEPT` header. For more comprehensive version compatibility checks we can create a separate issue and address it at a later point in time (IMO, it is a lower priority right now compared to checking if the genesis hashes match).

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: I've reverted this to the original implementation i.e. client sets its own version and the server validates it. We can revisit if this becomes an issue.

Mirko-von-Leipzig:
I've refactored this to use an external crate and revert the version requirements to what they were before. I've also updated the description to match the current reality.

If you could re-review please :)

crates/rpc/src/server/accept.rs
```diff
@@ -1,61 +1,186 @@
...
+    pub fn new(rpc_version: &Version, genesis_commitment: Word) -> Self {
+        let supported_versions = VersionReq {
+            comparators: vec![Comparator {
+                op: semver::Op::Exact,
+                major: rpc_version.major,
+                minor: rpc_version.minor.into(),
+                patch: None,
+                pre: semver::Prerelease::default(),
+            }],
+        };
```
Mirko-von-Leipzig, line 73: I'm a bit on the fence about this; but I've left the current behavior as is.

The issue is that its a bit back-to-front. The client should be indicating which versions they're compatible with. For example, their RPC version might be patch compatible with our server's but they may require a new feature or addition that we introduced.

Its easy to do by swapping the client and server types around. The downside is that client's need to think a bit more about what they want.

crates/rpc/src/server/accept.rs
```diff
@@ -1,61 +1,186 @@
...
+            let genesis = media_type
+                .get_param(Self::GENESIS)
+                .map(|value| Word::try_from(value.unquoted_str().as_ref()))
+                .transpose()
+                .map_err(AcceptHeaderError::InvalidGenesis)?;
+            if let Some(genesis) = genesis
+                && genesis != self.genesis_commitment
+            {
+                continue;
+            }
```
Mirko-von-Leipzig, line 164: As @drahnr pointed out, its probably nonsensical to allow specifying multiple media-types with different genesis values so we could error out here with a more specific error. Currently they'll just get a generic `NoSupportedMediaRange` error which is not that useful.

This is technically against the spirit of the `Accept` header which should allow setting multiple media-types, but 🤷

crates/rpc/src/server/accept.rs
```diff
@@ -1,61 +1,186 @@
...
+
+            // Skip those that don't match the version requirement.
+            let version = media_type
+                .get_param(Self::VERSION)
+                .map(|value| Version::parse(value.unquoted_str().as_ref()))
+                .transpose()
+                .map_err(AcceptHeaderError::InvalidVersion)?;
+            if let Some(version) = version
+                && !self.supported_versions.matches(&version)
```
drahnr, line 149: 👍

drahnr (approved):

crates/rpc/src/server/accept.rs
```diff
@@ -70,153 +195,157 @@ where
...
+impl QValue {
+    fn is_zero(&self) -> bool {
+        self.kilo == 0
+    }
+}
+
+impl FromStr for QValue {
+    type Err = QParsingError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
igamigo, line 261: Implementation of this looks correct but I'd add a test or two for some corner cases like `q=0.`

crates/rpc/src/server/accept.rs
igamigo: Not sure about the spec but I think the current implementation should be fine

igamigo (approved):

CHANGELOG.md
```diff
@@ -2,9 +2,20 @@
...
+### Changes
+
 - [BREAKING] Updated MSRV to 1.88.
 - Added environment variable support for batch and block size CLI arguments ([#1081](https://github.com/0xMiden/miden-node/pull/1081)).
 - A transaction's account delta is now checked against its commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
+- [BREAKING] RPC accept header format changed from `application/miden.vnd+grpc.<version>` to `application/vnd.miden; version=<version>` ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
+
+### Fixes
+
+- RPC requests with wildcard (`*/*`) media-type are rejected ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
```
bobbinth: This contracts the PR description a bit which states that we are accepting `*/*` media types.

crates/rpc/src/server/accept.rs
```diff
@@ -70,153 +195,195 @@ where
...
+        self.kilo == 0
+    }
 }
 
-impl<'a> TryFrom<&'a str> for AcceptHeaderValue<'a> {
-    type Error = nom::Err<nom::error::Error<&'a str>>;
+impl FromStr for QValue {
+    type Err = QParsingError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
bobbinth, line 261: Could we add a short doc comment explaining what the valid formats are? Also, looking at this, it seems like we should accept a fixed-point number (up to 3 decimal places). If so, could we not parse it as a regular float (or maybe a fixed-point number)?

bobbinth (approved):

crates/rpc/src/server/accept.rs
bobbinth: I think this is fine for now, but I would create an issue to revisit this in the future because in principle I agree: the client should be specifying the version that it is compatible with and if the node is within this specifications, the request should be accepted.

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: The format is a fixed point number with up-to 3 decimal places between zero and one, inclusive.

> (or maybe a fixed-point number)?

I assume you mean using some crate?

> If so, could we not parse it as a regular float

We should be able to, I initially dismissed this because I didn't think we could check the number of decimal places afterwards due to floating point idiosyncrasies. But I realize now we can check the length of the string instead :facepalm:

I'll try this out.

-- edit -- actually `f32` accepts a lot more formats e.g. scientific notation etc

CHANGELOG.md
Mirko-von-Leipzig: I'm used to writing `Fixes` from the bug's perspective. As in `These requests were incorrectly rejected` is now fixed.

But maybe that's confusing?

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: > Could we add a short doc comment explaining what the valid formats are?

I'll add it; the format is specified on the `QValue` type but I'll repeat it here.

Mirko-von-Leipzig: Oh hang on :facepalm: I guess you meant simply parsing the decimal places as a `u16` directly.. Will do that.

> > (or maybe a fixed-point number)?
>
> I assume you mean using some crate?

Mirko-von-Leipzig: I've changed to `u16::from_str`, its still a bit weird due to handling of decimal places. However my impl had a bug there so its definitely a positive :)

CHANGELOG.md
bobbinth: I would maybe say that "RPC requests with wildcard (`*/*`) media-type are no longer rejected" - but I'm fine to leave as is if you think that's better.

bobbinth (approved):
All looks good! Thank you!

PR #1085, opened 2025-07-14
ci: restrict workflow permissions

Mirko-von-Leipzig:
Sets the workflow permissions to the lowest setting (`read`) for those workflows without any set.

This fixes a bunch of the security complaints from the github bot.


bobbinth (approved):
Looks good! Thank you!

PR #1086, opened 2025-07-14
fix(ntx): accounts are never unlocked

Mirko-von-Leipzig:
I forgot to remove network accounts from the in-progress jail once the transaction completes.

I noticed this while debugging the instrumentation.

### Instrumentation

I've noticed that the ntx spans aren't actually chaining together. This is because we're spawning a task and not passing along the span.

This PR addresses this (I think) - its difficult to be sure because I can't trigger ntx using client integration tests at the moment.

I've added a call to the outer blocking spawn, and the inner spawn so I assume this should work.

### DIFF
--- a/crates/ntx-builder/src/state/mod.rs
+++ b/crates/ntx-builder/src/state/mod.rs
@@ -235,6 +235,9 @@ impl State {
                 },
             }
 
+            // If this account was in-progress, then it should no longer be as this update is the
+            // result of our own network transaction.
+            self.in_progress.remove(&prefix);
             tx_impact.account_delta = Some(prefix);
         }
         for note in network_notes {
--- a/crates/ntx-builder/src/transaction.rs
+++ b/crates/ntx-builder/src/transaction.rs
@@ -1,7 +1,7 @@
 use std::{collections::BTreeSet, sync::Arc};
 
 use futures::TryFutureExt;
-use miden_node_utils::{ErrorReport, FlattenResult, tracing::OpenTelemetrySpanExt};
+use miden_node_utils::{ErrorReport, tracing::OpenTelemetrySpanExt};
 use miden_objects::{
     TransactionInputError, Word,
     account::{Account, AccountId},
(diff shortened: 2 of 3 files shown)


crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig, line 1: Diff looks larger than it is due to whitespace. If using github, I'd advise toggling the whitespace filter on.

Its effectively just 2x `.in_current_span()` added.

igamigo:
> its difficult to be sure because I can't trigger ntx using client integration tests at the moment.

Is this an issue with the tests or your local environment? FWIW I was able to test against this branch and it seemed to work fine, but not entirely sure how to validate the spans are chained correctly (and that accounts are unlocked correctly without updating the tests to attempt triggering a second ntx)

crates/ntx-builder/src/state/mod.rs
```diff
@@ -235,6 +235,9 @@ impl State {
                 },
             }
 
+            // If this account was in-progress, then it should no longer be as this update is the
+            // result of our own network transaction.
+            self.in_progress.remove(&prefix);
```
bobbinth, line 240: This mechanism would need to change once we allow transactions affecting network accounts to be submitted externally, right?

bobbinth (approved):
Looks good! Thank you!

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: I _think_ it might be sufficient still, but we should definitely think about the exact mechanics here. On first blush my thoughts are that essentially any local ntx would be racing this external tx and then

- either our local ntx wins and its the same as before, or
- we lose the race and our ntx gets rejected by the block-producer (which already unlocks the account)

So at worst we do one tx's worth of work for no reason. This is already the same scenario as an external tx consuming a note, or something being reverted. The block-producer should remain the "sync" point.

Mirko-von-Leipzig:
> > its difficult to be sure because I can't trigger ntx using client integration tests at the moment.
> 
> Is this an issue with the tests or your local environment? FWIW I was able to test against this branch and it seemed to work fine, but not entirely sure how to validate the spans are chained correctly (and that accounts are unlocked correctly without updating the tests to attempt triggering a second ntx)

It was my mistake - I had latest `main` instead of the actual version which meant the RPC version mismatched. I can re-run tomorrow; the span chaining is only really visible locally when using the `tracing-forest` feature so that span relationships are visible.

sergerad (approved):

PR #1092, opened 2025-07-15
chore: cargo update and fix build

tomyrd:
This PR updates the use of the miden base libraries to their `next` version and fixes build issues. These changes are needed to do the same updates in the client and it's testing node binary.

The changes include:
- All use of `Digest` and `RpoDigest` is replaced with `Word`. I also replaced references to digest in comments and variable names.
- Forests now have their own type instead of being `usize`.
- The `miden-tx` needs to be compiled with the `testing` feature until [it is fixed in base](https://github.com/0xMiden/miden-base/pull/1589).
- Some dependency versions had to be updated to match with miden-base.
- The `NoteInclusionProof` now uses a `SparseMerklePath` instead of `MerklePath` so some conversions had to be added.


crates/block-producer/src/test_utils/mod.rs
```diff
@@ -45,7 +43,7 @@ impl Random {
         AccountIdBuilder::new().build_with_rng(&mut self.0)
     }
 
-    pub fn draw_digest(&mut self) -> Digest {
-        self.0.draw_word().into()
+    pub fn draw_word(&mut self) -> Word {
```
sergerad: Just worth checking whether we want these (and similar) fns to remain using the digest word (`draw_digest()`). Unsure myself for this one.

crates/proto/src/domain/word.rs
```diff
@@ -0,0 +1,238 @@
+use std::fmt::{Display, Formatter};
```
sergerad: Similarly worth checking whether this file should remain `digest.rs` or not.

crates/store/src/db/migrations/001-init.sql
```diff
@@ -131,7 +131,7 @@ CREATE TABLE nullifiers (
 
     PRIMARY KEY (nullifier),
     FOREIGN KEY (block_num) REFERENCES block_headers(block_num),
-    CONSTRAINT nullifiers_nullifier_is_digest CHECK (length(nullifier) = 32),
+    CONSTRAINT nullifiers_nullifier_is_word CHECK (length(nullifier) = 32),
```
sergerad: Unsure if this should have changed.

proto/proto/account.proto
```diff
@@ -19,7 +19,7 @@ message AccountSummary {
     AccountId account_id = 1;
 
     // The current account commitment or zero if the account does not exist.
-    digest.Digest account_commitment = 2;
+    word.Word account_commitment = 2;
```
sergerad: Flagging this too. Just not sure the proto needs to change. Unsure who to @ to ask.

igamigo: I think this (ie the ones related to protobuf types) should be fine. `Digest` and `Word` used to be pretty much equivalent though they were two different types. They were recently consolidated and all instances of `Digest` were changed to `Word` upstream, so probably makes sense to keep these for consistency. cc @PhilippGackstatter @bobbinth

tomyrd:
Agreed, maybe I jumped the gun on renaming instances of "digest". I can revert that part if it's decided it shouldn't be done.

proto/proto/account.proto
bobbinth: I would probably keep it as `Digest` - at least for now. I think almost all uses of this in the protobuf files are actual digests (i.e., result of hashing something).

PhilippGackstatter: I would also lean towards using `Word` since that's the type and it'd be consistent with Rust. In [miden-base](https://github.com/0xMiden/miden-base/blob/7d0e446c885bcd19b782f3f9423b60fe976aaae9/crates/miden-objects/src/transaction/transaction_id.rs#L89) I've left such instances as `digest: Word`, so the name indicates its a digest but the type has to be `Word`.

But this is not a strong opinion.

sergerad (approved):

crates/store/src/db/sql/mod.rs
```diff
@@ -1076,7 +1073,11 @@ pub fn select_note_inclusion_proofs(
         let merkle_path_data = row.get_ref(4)?.as_blob()?;
         let merkle_path = MerklePath::read_from_bytes(merkle_path_data)?;
 
-        let proof = NoteInclusionProof::new(block_num.into(), node_index_in_block, merkle_path)?;
+        let proof = NoteInclusionProof::new(
+            block_num.into(),
+            node_index_in_block,
+            merkle_path.try_into()?,
```
bobbinth, line 1079: We need to do this conversion here because we store `MerklePath` in the database, but `NoteInclusionProof` needs `SparseMerklePath`, right? If so, why not store `SparseMerklePath` in the database?

crates/store/src/state.rs
```diff
@@ -280,15 +277,15 @@ impl State {
...
+                let sparse_merkle_path = note_tree.open(note_index);
 
                 let note_record = NoteRecord {
                     block_num,
                     note_index,
                     note_id: note.id().into(),
                     metadata: *note.metadata(),
                     details,
-                    merkle_path,
+                    merkle_path: sparse_merkle_path.into(),
```
bobbinth, line 288: Similar comment as above about conversions between sparse and regular Merkle paths.

crates/proto/src/domain/note.rs
```diff
@@ -85,7 +86,7 @@ impl From<(&NoteId, &NoteInclusionProof)> for proto::NoteInclusionInBlockProof {
             note_id: Some(note_id.into()),
             block_num: proof.location().block_num().as_u32(),
             note_index_in_block: proof.location().node_index_in_block().into(),
-            merkle_path: Some(Into::into(proof.note_path())),
+            merkle_path: Some(MerklePath::from(proof.note_path().clone()).into()),
```
bobbinth, line 89: Same comment as above.

crates/proto/src/domain/account.rs
```diff
@@ -270,9 +269,9 @@ impl TryFrom<proto::responses::AccountTransactionInputRecord> for AccountState {
             )))?
             .try_into()?;
 
-        // If the commitment is equal to `Digest::default()`, it signifies that this is a new
+        // If the commitment is equal to `Word::default()`, it signifies that this is a new
         // account which is not yet present in the Store.
-        let account_commitment = if account_commitment == Digest::default() {
+        let account_commitment = if account_commitment == Word::default() {
```
bobbinth: nit: we started using `Word::empty()` instead of `Word::default()`.

crates/store/src/db/sql/mod.rs
tomyrd: Created [an issue](https://github.com/0xMiden/miden-node/issues/1096) to track this. Will tackle it in a follow up PR.

crates/proto/src/domain/account.rs
```diff
@@ -270,9 +269,9 @@ impl TryFrom<proto::responses::AccountTransactionInputRecord> for AccountState {
             )))?
             .try_into()?;
 
-        // If the commitment is equal to `Digest::default()`, it signifies that this is a new
+        // If the commitment is equal to `Word::empty()`, it signifies that this is a new
         // account which is not yet present in the Store.
-        let account_commitment = if account_commitment == Digest::default() {
+        let account_commitment = if account_commitment == Word::empty() {
```
Mirko-von-Leipzig, line 274:
```suggestion
        let account_commitment = if account_commitment.is_empty() {
```

crates/block-producer/src/domain/transaction.rs
```diff
@@ -140,8 +140,10 @@ impl AuthenticatedTransaction {
     /// Short-hand for `Self::new` where the input's are setup to match the transaction's initial
     /// account state. This covers the account's initial state and nullifiers being set to unspent.
     pub fn from_inner(inner: ProvenTransaction) -> Self {
+        use miden_objects::Word;
+
         let store_account_state = match inner.account_update().initial_state_commitment() {
-            zero if zero == Digest::default() => None,
+            zero if zero == Word::empty() => None,
```
Mirko-von-Leipzig, line 146:
```suggestion
            zero if zero.is_empty() => None,
```

PR #1093, opened 2025-07-16
chore: check matching account delta commitments

### LINKED ISSUE #306: [Feature]: Verify public accounts' details

End user

From @bobbinth:

> The main open question is how to verify that the account details included with a public account are valid. The most problematic part of this is to verify account delta for existing accounts. Ideally, this should be happening in the RPC so that a proven transaction with malformed account delta would not make it into the block producer. But I'm not seeing an efficient way of doing this. So, the current approach is fine, but let's create an issue to figure out how to validate account delta more holistically.

### How is this feature used?

Verify in RPC (in `SubmitProvenTransaction` endpoint).

### Anything else?

_No response_

SantiagoPittella:
closes #306

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - [BREAKING] Updated MSRV to 1.88.
 - Added environment variable support for batch and block size CLI arguments ([#1081](https://github.com/0xMiden/miden-node/pull/1081)).
-- Added check of matching account delta commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
+- A transaction's account delta is now checked against its commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
 
 ## v0.10.1 (2025-07-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [BREAKING] Updated MSRV to 1.88.
 - Added environment variable support for batch and block size CLI arguments ([#1081](https://github.com/0xMiden/miden-node/pull/1081)).
+- Added check of matching account delta commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
 
 ## v0.10.1 (2025-07-14)
 
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -31,11 +31,11 @@ use miden_node_utils::{
     tracing::grpc::OtelInterceptor,
 };
 use miden_objects::{
-    Digest, MAX_NUM_FOREIGN_ACCOUNTS, MIN_PROOF_SECURITY_LEVEL,
+    Digest, Hasher, MAX_NUM_FOREIGN_ACCOUNTS, MIN_PROOF_SECURITY_LEVEL,
     account::{AccountId, delta::AccountUpdateDetails},
     crypto::hash::rpo::RpoDigest,
     transaction::ProvenTransaction,
-    utils::serde::Deserializable,
+    utils::serde::{Deserializable, Serializable},
 };
 use miden_tx::TransactionVerifier;
 use tonic::{
@@ -245,6 +245,20 @@ impl api_server::Api for RpcService {
             return Err(Status::invalid_argument(
                 "Network transactions may not be submitted by users yet",
             ));
+        }
+
+        // Compare the account delta commitment of the ProvenTransaction with the actual delta
+        let delta_commitment = tx.account_update().account_delta_commitment();
+
+        // Verify that the delta commitment matches the actual delta
+        if let AccountUpdateDetails::Delta(delta) = tx.account_update().details() {
+            let computed_commitment = Hasher::hash(&delta.to_bytes());
+
+            if computed_commitment != delta_commitment {
+                return Err(Status::invalid_argument(
+                    "Account delta commitment does not match the actual account delta",
+                ));
+            }
         }
 
         let tx_verifier = TransactionVerifier::new(MIN_PROOF_SECURITY_LEVEL);
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -31,10 +31,10 @@ use miden_node_utils::{
     tracing::grpc::OtelInterceptor,
 };
 use miden_objects::{
-    Hasher, MAX_NUM_FOREIGN_ACCOUNTS, MIN_PROOF_SECURITY_LEVEL, Word,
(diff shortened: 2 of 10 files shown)


crates/rpc/src/server/api.rs
```diff
@@ -247,6 +247,20 @@ impl api_server::Api for RpcService {
             ));
         }
 
+        // Compare the account delta commitment of the ProvenTransaction with the actual delta
+        let delta_commitment = tx.account_update().account_delta_commitment();
+
+        // Verify that the delta commitment matches the actual delta
+        if let AccountUpdateDetails::Delta(delta) = tx.account_update().details() {
+            let computed_commitment = Hasher::hash(&delta.to_bytes());
```
bobbinth: We should compute the commitment using `AccountDelta::commitment()` (which I renamed to `to_commitment()` in https://github.com/0xMiden/miden-base/pull/1603). Otherwise, the commitment wouldn't match and this will always fail.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - [BREAKING] Updated MSRV to 1.88.
 - Added environment variable support for batch and block size CLI arguments ([#1081](https://github.com/0xMiden/miden-node/pull/1081)).
+- Added check of matching account delta commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
```
```suggestion
- A transaction's account delta is now checked against its commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
```

Mirko-von-Leipzig (approved):
Looks good; modulo the change from Bobbin.

crates/rpc/src/tests.rs
```diff
@@ -1,17 +1,29 @@
...
 use miden_node_store::{GenesisState, Store};
+use miden_objects::{
+    Word,
+    account::{
+        AccountDelta, AccountId, AccountIdVersion, AccountStorageDelta, AccountStorageMode,
+        AccountType, AccountVaultDelta, delta::AccountUpdateDetails,
+    },
+    transaction::ProvenTransactionBuilder,
+};
+use miden_tx::utils::Serializable;
```
bobbinth: nit: `Serializable` could come from `miden_objects::utils`.

crates/rpc/src/tests.rs
```diff
@@ -1,17 +1,29 @@
 use std::{net::SocketAddr, time::Duration};
 
+use miden_air::{ExecutionProof, Felt, HashFunction};
```
bobbinth: `Felt` could come from `miden_objects` and `ExecutionProof` could come from `miden_tx`.

crates/rpc/src/tests.rs
```diff
@@ -137,6 +149,68 @@ async fn rpc_startup_is_robust_to_network_failures() {
...
+        AccountDelta::new(
+            account_id,
+            AccountStorageDelta::new(),
+            AccountVaultDelta::default(),
+            Felt::default(),
+        )
+        .unwrap(),
+    ))
+    .build()
+    .unwrap();
```
bobbinth: I wonder if there is an easier way to create a dummy proven transaction. At the very least, would be great to have a simpler way to create dummy `ExecutionProof`. Maybe we should make a small PR in `miden-vm` that adds `ExecutionProof:dummy()`? This way, we could avoid adding `miden-air` and `winterfell` dependencies in this PR.

cc @PhilippGackstatter and @igamigo in case you have other suggestions.

bobbinth (approved):
Thank you! Looks good! I think the code itself is there, but I wonder if we should improve the construction of a dummy transaction a bit. I left a comment about this inline.

crates/rpc/src/tests.rs
SantiagoPittella: I have the changes, but I don't have permission to push in the vm repo. Will sync lambdaclass fork and send it from there.

SantiagoPittella: Just opened https://github.com/0xMiden/miden-vm/pull/2007

Mirko-von-Leipzig: Is there a world where we make it easier to create real txs? I've been thinking about a db-tx like API for a while now, but I don't actually know enough :)

```rust
// Create an account with a `Counter` component (however that works).
// This could also come from a mock chain.
let account = AccountBuilder::default().with_component(Counter).build();

// Create an AccountTx which holds a ref to `Account` plus whatever stuff it 
// wants to execute. I imagine this would be some sort of cmd/vm stack.
let mut tx = account.begin_tx();

// This exposes the counter state + API.
// SAFETY: account definitely has a counter component.
let mut counter = tx.get_component(Counter).unwrap();
counter.increment(); // Or whatever method it has.
counter.increment();

// Do other stuff. Grab other components.

// This is now an ExecutedTransaction using the original account state.
let executed_tx = tx.execute().unwrap();

// You can now apply or discard the changes of the tx to the account.
tx.commit().unwrap();
// tx.abort().unwrap();
```

Mirko-von-Leipzig: It would be nice to have something simple to create realistic transactions. Though I guess proving is still an issue ito performance for tests?

bobbinth: > Just opened [0xMiden/miden-vm#2007](https://github.com/0xMiden/miden-vm/pull/2007)

`miden-vm` v0.16.3 with these changes is now on crates.io.

> It would be nice to have something simple to create realistic transactions. Though I guess proving is still an issue ito performance for tests?

Generating a real proof is expensive, most other things shouldn't be.

@PhilippGackstatter @igamigo - anything you can recommend on how to improve creation of a dummy transaction here?

If there isn't anything too simple to do, I'd probably use what we have here and create another issue for improving creation of dummy transactions in `miden-base`.

PhilippGackstatter: For creating `ExecutedTransaction` it's easiest to use `MockChain::build_tx_context`. Once we've gotten rid of the pending note APIs in `MockChain`, there shouldn't really be a way to create transactions that couldn't exist in a real chain, so that would be optimal. This can already be done now just by not using those APIs and instead adding notes via the APIs on `MockChainBuilder` or creating them yourself.

Once you have an `ExecutedTransaction`, then you could technically use `ProvenTransactionExt::from_executed_transaction_mocked` to convert that transaction into a proven transaction with a dummy proof. But I also would like to get rid of it, because it's hacky and could easily break or do something that a real transaction could not do as the codebase evolves.

Ideally, all our prover types in miden-base would all have a `prove` and `prove_dummy` function that takes an `ExecutedTransaction` and turns it into a `ProvenTransaction` with the only difference being the use of a dummy proof on the latter. That would be much easier to maintain. The batch and block prover roughly have that functionality, but it could be exposed in a better way.

SantiagoPittella:
@bobbinth I created #1104 , and put it in the next release (v0.12).

PR #1094, opened 2025-07-17
feat: Enforce Account ID Prefix uniqueness

### LINKED ISSUE #601: feat: enforce account ID prefix uniqueness

sergerad:
WIP

Closes #601.


crates/store/src/server/block_producer.rs
```diff
@@ -176,6 +177,19 @@ impl block_producer_server::BlockProducer for StoreApi {
...
+            // If the Account does not exist, this transaction must be creating it.
+            Err(DatabaseError::AccountNotFoundInDb(_)) => {
+                // Validate the new Account's Id prefix is unique.
+                if self.state.account_exists(account_id).await {
+                    return Err(Status::already_exists("Account Id prefix already exists"));
+                }
+            },
+            Err(err) => return Err(err.into()),
+            _ => {},
+        }
```
sergerad: @bobbinth LMK what might not make sense here. My understanding is that we need to check the database for existence of the account because the `account_tree.smt` only stores prefixes.

bobbinth: I would move the logic into `State::get_transaction_inputs()` - we already working with account tree there - so, might as well centralize the logic.

I would also make the check after we get `account_commitment` because we can tell if account is a new account or not by looking at its commitment (as discussed in the issue) - so, we don't need to do an extra lookup for that.

To check if there is another account in the tree with a different ID but the same 64-bit prefix, we probably will need to modify `AccountTree` as I don't think it exposes a method to check that. Maybe even we'll need to go to `miden-crypto` and expose the required functionality from the `Smt` struct.

sergerad: > To check if there is another account in the tree with a different ID but the same 64-bit prefix, we probably will need to modify `AccountTree` as I don't think it exposes a method to check that. Maybe even we'll need to go to `miden-crypto` and expose the required functionality from the `Smt` struct.

I made an attempt to do this without modifying `AccountTree` or smt. Added this to `State`:
```rust
    /// Checks if an account exists in the state.
    pub async fn account_exists(&self, account_id: AccountId) -> bool {
        inner_state.account_tree.get(account_id) != EMPTY_WORD.into()
    }
```

Which depends on this deeper into the stack of smt:
```rust
    fn get_value(&self, key: &Self::Key) -> Self::Value {
        let leaf_pos = LeafIndex::<SMT_DEPTH>::from(*key).value();

        match self.leaves.get(&leaf_pos) {
            Some(leaf) => leaf.get_value(key).unwrap_or_default(),
            None => EMPTY_WORD,
        }
    }
```

bobbinth: I'm not sure this will work: `Smt::get_value()` returns a value for a given key - but we don't know the key we want to use here (we need a key which is built from an account ID for a different account which shares the same 64-bit prefix with the `account_id`).

Also, I'm not sure we need to create `account_exists()` method as we can check this directly by just looking at the `account_commitment` in `State::get_transaction_inputs()`.

sergerad: > I'm not sure this will work: `Smt::get_value()` returns a value for a given key - but we don't know the key we want to use here (we need a key which is built from an account ID for a different account which shares the same 64-bit prefix with the `account_id`).

Yea it seems the index into the tree is made from the entirety of the account id as a word.
```rust
    /// Returns the SMT key of the given account ID.
    pub(super) fn id_to_smt_key(account_id: AccountId) -> Digest {
        // We construct this in such a way that we're forced to use the constants, so that when
        // they're updated, the other usages of the constants are also updated.
        let mut key = [Felt::ZERO, Felt::ZERO, Felt::ZERO, Felt::ZERO];
        key[Self::KEY_SUFFIX_IDX] = account_id.suffix();
        key[Self::KEY_PREFIX_IDX] = account_id.prefix().as_felt();

        Digest::from(key)
    }
```
and
```rust
    fn get_value(&self, key: &Self::Key) -> Self::Value {
        let leaf_pos = LeafIndex::<SMT_DEPTH>::from(*key).value();
```

Which seems at odds with this comment in the code:
```rust
/// Each account ID occupies exactly one leaf in the tree, which is identified by its
/// [`AccountId::prefix`]. In other words, account ID prefixes are unique in the blockchain.
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct AccountTree {
```

sergerad: The only way the `AccountTree` checks for uniqueness of prefixes ATM is via insert:
```rust
        let key = Self::id_to_smt_key(account_id);
        let prev_value = Digest::from(self.smt.insert(key, Word::from(state_commitment)));

        // If the leaf of the account ID now has two or more entries, we've inserted a duplicate
        // prefix.
        if self.smt.get_leaf(&key).num_entries() >= 2 {
            return Err(AccountTreeError::DuplicateIdPrefix {
                duplicate_prefix: account_id.prefix(),
            });
        }
```

Only changes to SMT I can think of to support our use case are:
- Add another mapping, `prefix -> leaf`
- Replace the current mapping of the `entire accound id -> leaf` with `prefix -> leaf`

bobbinth: I think the way it works is that `AccountTree` ensures that we can only insert a single leaf per 64-bit prefix. This is done in `AccountTree::compute_mutations()` method (and if we do try to insert another account with the same prefix but different key, it'll return `AccountTreeError::DuplicateIdPrefix` error). So, "by construction" we can't have more than one leaf per account.

The issue is that we get this error only on insert - but, AFAICT, we can't currently ask the `AccountTree` if there will be an error on insert, without actually executing `AccountTree::compute_mutations()`. Ideally, we'd have some method on the account tree that would tell is whether an update will result in an error beforehand.

Also, cc @PhilippGackstatter in case he has some additional thoughts.

bobbinth: > Only changes to SMT I can think of to support our use case are:
> 
> * Add another mapping, `prefix -> leaf`
> * Replace the current mapping of the `entire account id -> leaf` with `prefix -> leaf`

I think we could add something like `AccountTree::has_leaf_for_account(AccountId) -> bool` and inside this method we could use `Smt::get_leaf()` method and check what this method returns.

sergerad: I think that won't work because `Smt::get_leaf()` doesn't just use the prefix to get the leaf index.

E.G. if you follow the stack from `AccountTree::get()`:
```rust
    pub fn get(&self, account_id: AccountId) -> Digest {
        let key = Self::id_to_smt_key(account_id);
        Digest::from(self.smt.get_value(&key))
    }
```
The above key is made up of the entire account id:
```rust
    /// Returns the SMT key of the given account ID.
    pub(super) fn id_to_smt_key(account_id: AccountId) -> Digest {
        // We construct this in such a way that we're forced to use the constants, so that when
        // they're updated, the other usages of the constants are also updated.
        let mut key = [Felt::ZERO, Felt::ZERO, Felt::ZERO, Felt::ZERO];
        key[Self::KEY_SUFFIX_IDX] = account_id.suffix();
        key[Self::KEY_PREFIX_IDX] = account_id.prefix().as_felt();

        Digest::from(key)
    }
```
And eventually you see leaf position in smt calculated using the entire key:
```rust
    fn get_value(&self, key: &Self::Key) -> Self::Value {
        let leaf_pos = LeafIndex::<SMT_DEPTH>::from(*key).value();

        match self.leaves.get(&leaf_pos) {
            Some(leaf) => leaf.get_value(key).unwrap_or_default(),
```

So I would expect `get_leaf()` to be the same - the entire account id is used (`key`), not just the prefix:
```rust
    fn get_leaf(&self, key: &RpoDigest) -> Self::Leaf {
        let leaf_pos = LeafIndex::<SMT_DEPTH>::from(*key).value();

        match self.leaves.get(&leaf_pos) {
```

bobbinth: > So I would expect `get_leaf()` to be the same - the entire account id is used (`key`), not just the prefix:
> 
> ```rust
>     fn get_leaf(&self, key: &RpoDigest) -> Self::Leaf {
>         let leaf_pos = LeafIndex::<SMT_DEPTH>::from(*key).value();
> 
>         match self.leaves.get(&leaf_pos) {
> ```

`Smt::get_leaf` works a bit differently. Here, we map `key` to `leaf_pos`. All keys sharing the same 64-bit prefix will map to the same `leaf_pos`. Then, we get the leaf from the tree based on its `leaf_pos`. In theory, the leaf itself could contain multiple key-value entries (in cases when multiple keys map to the same leaf - but for the account tree this won't happen in practice).

sergerad: Looks like that works, based on the unit test I have added. Please have a review
https://github.com/0xMiden/miden-base/pull/1610

sergerad: Noting that AFAICT this is the reason only the prefix is used here
```rust
impl From<Word> for LeafIndex<SMT_DEPTH> {
    fn from(value: Word) -> Self {
        // We use the most significant `Felt` of a `Word` as the leaf index.
        Self::new_max_depth(value[3].as_int())
    }
}
```

bin/remote-prover/src/api/prover.rs
```diff
@@ -277,8 +277,9 @@ mod test {
         let mut client_2 = ApiClient::connect("http://127.0.0.1:50052").await.unwrap();
 
         // Create a mock transaction to send to the server
-        let mut mock_chain = MockChain::new();
-        let account = mock_chain.add_pending_existing_wallet(Auth::BasicAuth, vec![]);
+        let mut mock_chain = MockChain::builder();
+        let account = mock_chain.add_existing_wallet_with_assets(Auth::BasicAuth, vec![]).unwrap();
+        let mut mock_chain = mock_chain.build().unwrap();
```
sergerad: @PhilippGackstatter @TomasArrachea any idea whether this diff, or anything else coming out of next miden-base could cause this error?

https://github.com/0xMiden/miden-node/actions/runs/16405132235/job/46349950672?pr=1094

```
thread 'api::prover::test::test_prove_transaction' panicked at bin/remote-prover/src/api/prover.rs:312:57:
called `Result::unwrap()` on an `Err` value: 
TransactionProgramExecutionFailed(FailedAssertion { 
label: SourceSpan { source_id: SourceId(4294967295), start: ByteIndex(0), end: ByteIndex(0) }, source_file: None, clk: RowIndex(63360), err_code: 6550476539395293665, err_msg: Some("account nonce can only be incremented once")
})
```

Highlighting: `account nonce can only be incremented once`.

bobbinth: This has been recently merged into `next` in `miden-base`. Basically, previously account nonce could have been incremented multiple times in the same transaction. Now, we allow incrementing nonce only once. Maybe some tests was incrementing the nonce more than once?

PhilippGackstatter: If you omit the transaction script that calls `call.::miden::contracts::auth::basic::auth__tx_rpo_falcon512` then it passes. Authentication is no longer done manually via the tx script but automatically as part of the account code.

Here are some docs if you're interested: https://github.com/0xMiden/miden-base/blob/a3b3cba3f42dd94a0669b80d6f906d209c725d88/docs/src/account/code.md#authentication

sergerad: Thanks that does get the test further down the line. But some behaviour seems to have changed because this check fails:
```rust
        assert!(response_1.is_err() || response_2.is_err());
```

```
thread 'api::prover::test::test_prove_transaction' panicked at bin/remote-prover/src/api/prover.rs:330:9:
assertion failed: response_1.is_err() || response_2.is_err()
```

I don't follow why concurrency may have been affected.

Omitting tx script:
```rust
        let tx_context = mock_chain
            .build_tx_context(account.id(), &[], &[])
            .unwrap()
            .extend_input_notes(vec![note_1])
            .build()
            .unwrap();
```

Have pushed that up for now.

PhilippGackstatter: Btw. the `note_1` that gets passed to `extend_input_notes` is completely ignored and is not part of the transaction. So the test was always executing with 0 input notes.

There is 1) a `mock_chain.prove_next_block()` missing, to add this note to the chain state and 2) the note ID should be passed directly to `build_tx_context`, otherwise it is ignored. (The call to `extend_input_notes` should be removed). 

A better way to do 1) would be to add this note directly using the `MockChainBuilder` further above, so it is directly part of the genesis state.

As for the concurrency issue, I don't really understand why two proving requests are made in the first place and why one of them would be expected to fail and the other succeed.

sergerad: Thanks I've created an issue to follow up on the test after this PR https://github.com/0xMiden/miden-node/issues/1102

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - [BREAKING] Updated MSRV to 1.88.
 - Added environment variable support for batch and block size CLI arguments ([#1081](https://github.com/0xMiden/miden-node/pull/1081)).
+- Added check for Account Id prefix uniqueness when transactions to create accounts are submitted to the store ([#1094](https://github.com/0xMiden/miden-node/pull/1094)).
```
bobbinth: Should "submitted to the store" be "submitted to the mempool"?

crates/store/src/errors.rs
```diff
@@ -74,6 +74,8 @@ pub enum DatabaseError {
         Remove all database files and try again."
     )]
     UnsupportedDatabaseVersion,
+    #[error("duplicate account ID prefix {0}")]
+    DuplicateAccountIdPrefix(AccountId),
```
bobbinth: nit: should this be inserted in alphabetical order?

crates/store/src/state.rs
```diff
@@ -800,6 +800,14 @@ impl State {
 
         let account_commitment = inner.account_tree.get(account_id);
 
+        // If account commitment is empty, this transaction must be creating a new account.
+        if account_commitment == EMPTY_WORD {
```
Mirko-von-Leipzig: I believe this now exists

```suggestion
        if account_commitment.is_empty() {
```

Mirko-von-Leipzig: I'm actually surprised `clippy` doesn't suggest merging the nested `if`.

crates/store/src/errors.rs
```diff
@@ -74,6 +74,8 @@ pub enum DatabaseError {
         Remove all database files and try again."
     )]
     UnsupportedDatabaseVersion,
+    #[error("duplicate account ID prefix {0}")]
```
Mirko-von-Leipzig: Maybe? Original is fine as well tbh. Mostly thinking from the user's perspective `duplicate` might be confusing but maybe I'm overthinking it.

```suggestion
    #[error("account ID prefix {0} already exists")]
```

crates/store/src/errors.rs
Mirko-von-Leipzig: Unfortunately this won't be enough to propagate the error to the user. At the moment, the block-producer marks these `store` errors as internal and doesn't expose them to the client. https://github.com/0xMiden/miden-node/blob/531451a813dcb4538887b82cebf0d75b98570a23/crates/block-producer/src/errors.rs#L119-L120

We also _don't_ want to expose these as a general rule. I'm unsure what the broader solution to this is because gRPC doesn't really do structured errors. 

The immediate short term solution is matching on the error message.

Some longer term options:

- (ab)use the gRPC errors code. This isn't really practical because
  - codes _must_ be a specific subset so we can't make up our own
  - store server does not know what constitutes an internal vs external error, only the block-producer should know that
- add our own `Result` message type which doesn't really sound practical
- we use gRPC metadata to encode the actual error code/value
- we switch to a different internal protocol since we don't really care about backwards compatibility internally (and gRPC is messy). This would probably mean other's can't implement their own components in their own language though its hard to know if that matters since the internal protocol has its own quirks and assumptions already.

Mirko-von-Leipzig (commented):
Implementation LGTM, however there is an issue with error propagation as @bobbinth alludes to.

crates/store/src/errors.rs
bobbinth: Another option could be to just modify the `TransactionInputsRequest` message to add a field to it indicating whether the prefix was unique. Then the block producer can process it accordingly.

Mirko-von-Leipzig: > Another option could be to just modify the TransactionInputsRequest message to add a field to it indicating whether the prefix was unique. Then the block producer can process it accordingly.

In a way this is similar to a custom result type; but with less types. I think this is probably fine; though since its just a single error variant maybe we can try making it a result-like.

One public API that might benefit from a more structure result is the `submit_transaction` endpoint e.g. faucet does some funky error message parsing to know when its desync'd.

bobbinth: > In a way this is similar to a custom result type; but with less types. I think this is probably fine; though since its just a single error variant maybe we can try making it a result-like.

The response message currently looks like this:
```proto
message TransactionInputs {

    message AccountTransactionInputRecord { ... }
    message NullifierTransactionInputRecord { ...  }

    AccountTransactionInputRecord account_state = 1;
    repeated NullifierTransactionInputRecord nullifiers = 2;
    repeated primitives.Digest found_unauthenticated_notes = 3;
    fixed32 block_height = 4;
}
```

I was thinking of changing it to something like:
```proto
message TransactionInputs {

    message AccountTransactionInputRecord { ... }
    message NullifierTransactionInputRecord { ...  }

    AccountTransactionInputRecord account_state = 1;
    repeated NullifierTransactionInputRecord nullifiers = 2;
    repeated primitives.Digest found_unauthenticated_notes = 3;
    fixed32 block_height = 4;
    optional bool is_account_prefix_unique = 5;
}
```
And I think what you are recommending is more like this:
```proto
message TransactionInputsResponse {

    // this can also be generalized to work for multiple messages
    message TransactionInputsError {
        uint32 error_code = 1;
        optional string error_message = 2;
    }

    message TransactionInputs {
    
        message AccountTransactionInputRecord { ... }
        message NullifierTransactionInputRecord { ...  }
    
        AccountTransactionInputRecord account_state = 1;
        repeated NullifierTransactionInputRecord nullifiers = 2;
        repeated primitives.Digest found_unauthenticated_notes = 3;
        fixed32 block_height = 4;
        optional bool is_account_prefix_unique = 5;
    }

    oneof payload {
        TransactionInputs inputs = 1;
        TransactionInputsError error = 2;
    }
}
```
If so, I think that maybe for this specific purpose it is a bit of an overkill - though I don't mind it either.

> One public API that might benefit from a more structure result is the `submit_transaction` endpoint e.g. faucet does some funky error message parsing to know when its desync'd.

Let's create an issue for this - this may be a good use case for explicitly returning an error as a part of the message.

crates/store/src/state.rs
sergerad: Yea true. I do prefer the nesting here as it makes for a more peaceful read with the comments.

crates/store/src/errors.rs
sergerad: Created https://github.com/0xMiden/miden-node/issues/1106

Part of me would prefer to block this PR on that. More practical to come back to it and just finish this PR with
```proto
optional bool is_account_prefix_unique = 5;
```

bobbinth: I would probably do the simpler approach here now (i.e., with `is_account_prefix_unique`) and then come back and update with the more comprehensive solution later (as a part of #1106).

sergerad: Updated the PR to involve the additional proto field for uniqueness.

crates/store/src/state.rs
```diff
@@ -800,6 +801,14 @@ impl State {
 
         let account_commitment = inner.account_tree.get(account_id);
 
+        // If account commitment is empty, this transaction must be creating a new account.
+        let new_account_id_prefix_is_unique = if account_commitment.is_empty() {
+            let is_unique = !inner.account_tree.contains_account_id_prefix(account_id.prefix());
+            Some(is_unique)
```
bobbinth: if `is_unique == false`, we don't need to process the rest of the request right? We could probably put the retrieval of nullifiers and found unauthenticationed notes in a conditional block that gets executed only if `new_account_id_prefix_is_unique` is `None` or contains `true`.

crates/block-producer/src/errors.rs
```diff
@@ -181,6 +182,8 @@ pub enum BuildBlockError {
 /// Errors returned by the [`StoreClient`](crate::store::StoreClient).
 #[derive(Debug, Error)]
 pub enum StoreError {
+    #[error("account Id prefix already exists: {0}")]
+    DuplicateAccountIdPrefix(AccountId),
```
bobbinth, line 186: I think this should go into the `VerifyTxError` enum as it signals that the transaction submitted by the user is invalid (rather than something went wrong in the store).

crates/store/src/state.rs
sergerad: And put just `::default()` values in the response for those? Either way we are left in a bit of an awkward implementation until the error approach to the proto messages is implemented. Happy to impl the way you have suggested.

bobbinth: Yeah, those are just vectors of values - so, we'll effectively be sending empty vectors (which I think is fine - the transaction is invalid anyways and so we don't need to send more data).

crates/block-producer/src/errors.rs
sergerad: The relevant fn returns `StoreError` atm which I presume we don't want to change
```rust
    pub async fn get_tx_inputs(
        &self,
        proven_tx: &ProvenTransaction,
    ) -> Result<TransactionInputs, StoreError> {
```
So do you mean to add `StoreError::SomeVariantNameUnsure(#[from] VerifyTxError)` or similar?

bobbinth: Ah! Maybe then the right place to handle this error is not in `get_tx_inputs()` but in w/e calls `get_tx_inputs()`.

sergerad: The caller already maps the `StoreError` to `VerifyTxError`.
```rust
        let inputs = self.store.get_tx_inputs(&tx).await.map_err(VerifyTxError::from)?;
```

Although it seems like an awkward variant name `StoreConnectionFailed` but the error message is fine I think:
```rust
    #[error("failed to retrieve transaction inputs from the store")]
    StoreConnectionFailed(#[from] StoreError),
```

crates/store/src/state.rs
sergerad: Updated

crates/store/src/state.rs
```diff
@@ -800,6 +801,20 @@ impl State {
...
+        };
+
+        // Non-unique account Id prefixes for new accounts are not allowed.
+        if let Some(false) = new_account_id_prefix_is_unique {
+            return Ok(TransactionInputs {
+                new_account_id_prefix_is_unique,
+                ..Default::default()
+            });
+        }
+
```
Mirko-von-Leipzig, line 817: I think its easier to grok if the early return is embedded, but up to you.

```suggestion
        if account_commitment.is_empty() && 
            !inner.account_tree.contains_account_id_prefix(account_id.prefix()))
         {
            return Ok(TransactionInputs {
                new_account_id_prefix_is_unique,
                ..Default::default()
            });
        }
```

And then set the `Ok` variant at the end to `None`/`Default`.

Mirko-von-Leipzig (approved):

crates/block-producer/src/errors.rs
Mirko-von-Leipzig: The context changed over time, but the variant name didn't. This would get refactored with #1106 so I think its fine for now.

crates/store/src/state.rs
sergerad: That way we don't create `new_account_id_prefix_is_unique` which is fine for the early return but then we would have to figure out whether to return `None` or `Ok(true)` below.

Issue #1089, opened 2025-07-15
Track _when_ a network note is consumed and expose this via gRPC


Mirko-von-Leipzig:
Add a gRPC method for the `store.NtxBuilder` service which returns the list of unconsumed network notes for a specific network account.

This will enable a network account actor to load its available committed notes from the store. In this context, actor means some task that is building txs for a specific network account. 

In order to avoid race conditions we want to pin the block height of our request aka notes consumed _after_ this block should still be returned as part of the unconsumed set. This is because the actor will be receiving mempool events concurrently, which might include things committed to the store _after_ this block.

To support this block height condition, we need to change the notes table a little bit, from `notes.consumed: bool` to `notes.consumed: Option<BlockNumber>`.

The final result would enable something like this method in the ntx's store client:

```rust
fn unconsumed_notes_for_network_account(&self, account: NetworkAccountPrefix, block: BlockNumber) -> Vec<NetworkNote>;
```

Consider paginating the actual database/store gRPC interface though, as the list of notes could be very large. Though in the short term this shouldn't be an issue.

Additional context is provided in the parent issue #1055.

Mirko-von-Leipzig:
@sergerad let me know if you have any questions. I'd recommend subscribing to the parent issue as well in case we discover or discuss flaws to this idea.

Mirko-von-Leipzig:
@PhilippGackstatter is there a potential problem with this plan? I'm thinking about the potential note tag changes wrt network accounts.

At present the idea here would be to filter notes for a specific account by the note tag; but if the schema for that is changing to something more complex maybe that won't be trivially possible anymore?

PhilippGackstatter:
> At present the idea here would be to filter notes for a specific account by the note tag; but if the schema for that is changing to something more complex maybe that won't be trivially possible anymore?

We (still) haven't continued the note tag discussion, but I would say what is possible now should also be possible in the future. Best case the note tag will contain more of the account ID prefix to make it less likely that a given note tag matches multiple accounts, but that should be a strict improvement.

Mirko-von-Leipzig:
> I'm not sure I fully understood this part. When a network account is created, don't we want to get all of its notes from the store? We don't really care if some of them are also in the "in-flight" set - we can de-duplicate these notes in the NTB. But maybe I'm missing something?

[...]

PR #1097, opened 2025-07-17
refactor: store `SparseMerkleTree` for inclusion proofs

### LINKED ISSUE #1096: Store `SparseMerklePath` natively instead of converting `MerkleStore`
> We need to do this conversion here because we store `MerklePath` in the database, but `NoteInclusionProof` needs `SparseMerklePath`, right? If so, why not store `SparseMerklePath` in the database?

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-node/pull/1092#discussion_r2212259012_

We should refactor the node so that the `SparseMerklePath` is natively stored for note inclusion proofs. This should remove most of the `MerklePath` to `SparseMerklePath` conversions we do.

tomyrd:
closes #1096

### DIFF
--- a/crates/proto/src/domain/merkle.rs
+++ b/crates/proto/src/domain/merkle.rs
@@ -46,17 +46,28 @@ impl TryFrom<proto::merkle::MerklePath> for MerklePath {
 // SPARSE MERKLE PATH
 // ================================================================================================
 
-impl From<SparseMerklePath> for proto::merkle::MerklePath {
+impl From<SparseMerklePath> for proto::merkle::SparseMerklePath {
     fn from(value: SparseMerklePath) -> Self {
-        MerklePath::from(value).into()
+        let (empty_nodes_mask, siblings) = value.into_parts();
+        proto::merkle::SparseMerklePath {
+            empty_nodes_mask,
+            siblings: siblings.into_iter().map(proto::digest::Digest::from).collect(),
+        }
     }
 }
 
-impl TryFrom<proto::merkle::MerklePath> for SparseMerklePath {
+impl TryFrom<proto::merkle::SparseMerklePath> for SparseMerklePath {
     type Error = ConversionError;
 
-    fn try_from(merkle_path: proto::merkle::MerklePath) -> Result<Self, Self::Error> {
-        Ok(MerklePath::try_from(merkle_path)?.try_into()?)
+    fn try_from(merkle_path: proto::merkle::SparseMerklePath) -> Result<Self, Self::Error> {
+        Ok(SparseMerklePath::from_parts(
+            merkle_path.empty_nodes_mask,
+            merkle_path
+                .siblings
+                .into_iter()
+                .map(Word::try_from)
+                .collect::<Result<Vec<_>, _>>()?,
+        )?)
     }
 }
 
--- a/proto/proto/merkle.proto
+++ b/proto/proto/merkle.proto
@@ -8,3 +8,14 @@ message MerklePath {
     // List of sibling node hashes, in order from the root to the leaf.
     repeated digest.Digest siblings = 1;
 }
+
+// A different representation of a Merkle path designed for memory efficiency.
+message SparseMerklePath {
+    // A bitmask representing empty nodes.
+    //
+    // The set bit corresponds to the depth of an empty node. The least significant bit (bit 0)
+    // describes depth 1 node (root's children). The `bit index + 1` is equal to node's depth.
+    fixed64 empty_nodes_mask = 1;
+    // The non-empty nodes, stored in depth-order, but not contiguous across depth.
+    repeated digest.Digest siblings = 2;
+}
--- a/proto/proto/note.proto
+++ b/proto/proto/note.proto
@@ -75,7 +75,7 @@ message NoteInclusionInBlockProof {
     uint32 note_index_in_block = 3;
 
     // The note's inclusion proof in the block.
-    merkle.MerklePath inclusion_path = 4;
+    merkle.SparseMerklePath inclusion_path = 4;
 }
 
 // Represents proof of a note inclusion in the block.
@@ -90,5 +90,5 @@ message NoteSyncRecord {
     NoteMetadata metadata = 3;
 
     // The note's inclusion proof in the block.
-    merkle.MerklePath inclusion_path = 4;
+    merkle.SparseMerklePath inclusion_path = 4;
 }
--- a/proto/proto/responses.proto
+++ b/proto/proto/responses.proto
@@ -30,7 +30,7 @@ message GetBlockHeaderByNumberResponse {
     // The requested block header.
     block.BlockHeader block_header = 1;
 
-    // Sparse Merkle path to verify the block's inclusion in the MMR at the returned `chain_length`.
+    // Merkle path to verify the block's inclusion in the MMR at the returned `chain_length`.
     optional merkle.MerklePath mmr_path = 2;
 
     // Current chain length.
@@ -76,7 +76,7 @@ message SyncNoteResponse {
     // Block header of the block with the first note matching the specified criteria.
     block.BlockHeader block_header = 2;
 
-    // Sparse Merkle path to verify the block's inclusion in the MMR at the returned `chain_tip`.
+    // Merkle path to verify the block's inclusion in the MMR at the returned `chain_tip`.
     //
     // An MMR proof can be constructed for the leaf of index `block_header.block_num` of
     // an MMR of forest `chain_tip` with this path.
@@ -99,7 +99,7 @@ message AccountWitness {
     // The state commitment whose inclusion the witness proves.
     digest.Digest commitment = 3;
 
-    // The sparse merkle path of the state commitment in the account tree.
+    // The merkle path of the state commitment in the account tree.
     merkle.MerklePath path = 4;
 }
 
--- a/proto/proto/smt.proto
+++ b/proto/proto/smt.proto
@@ -37,7 +37,7 @@ message SmtLeaf {
 
 // The opening of a leaf in an SMT.
 message SmtOpening {
-    // The sparse Merkle path to the leaf.
+    // The Merkle path to the leaf.
     merkle.MerklePath path = 1;
 
     // The leaf itself.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2230,9 +2230,9 @@ dependencies = [
 
 [[package]]
 name = "miden-crypto"
-version = "0.15.5"
+version = "0.15.8"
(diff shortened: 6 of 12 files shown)


crates/proto/src/domain/merkle.rs
```diff
@@ -41,6 +43,22 @@ impl TryFrom<proto::merkle::MerklePath> for MerklePath {
     }
 }
 
+// SPARSE MERKLE PATH
+// ================================================================================================
+
+impl From<&SparseMerklePath> for proto::merkle::MerklePath {
```
Mirko-von-Leipzig: Maybe I missed this previously, but do we need to change the proto schema as well?

Mirko-von-Leipzig (approved):

crates/proto/src/domain/merkle.rs
tomyrd: Yes, now that you mention it, I think it should also be the case. I'll change the proto files to use sparse merkle paths.

crates/proto/src/domain/merkle.rs
```diff
@@ -9,34 +11,66 @@ use crate::{
...
 
-impl TryFrom<proto::merkle::MerklePath> for MerklePath {
+impl TryFrom<proto::merkle::SparseMerklePath> for MerklePath {
     type Error = ConversionError;
 
-    fn try_from(merkle_path: proto::merkle::MerklePath) -> Result<Self, Self::Error> {
+    fn try_from(merkle_path: proto::merkle::SparseMerklePath) -> Result<Self, Self::Error> {
         (&merkle_path).try_into()
     }
 }
```
tomyrd: I had to leave these conversions to and from the normal `MerklePath` as it is still used in other objects like the `AccountWitness` and `SmtProof`

crates/store/src/db/migrations/001-init.sql
```diff
@@ -31,23 +31,23 @@ CREATE TABLE accounts (
...
+    batch_index         INTEGER NOT NULL, -- Index of batch in block, starting from 0
+    note_index          INTEGER NOT NULL, -- Index of note in batch, starting from 0
+    note_id             BLOB    NOT NULL,
+    note_type           INTEGER NOT NULL, -- 1-Public (0b01), 2-Private (0b10), 3-Encrypted (0b11)
+    sender              BLOB    NOT NULL,
+    tag                 INTEGER NOT NULL,
+    execution_mode      INTEGER NOT NULL, -- 0-Network, 1-Local
+    aux                 INTEGER NOT NULL,
+    execution_hint      INTEGER NOT NULL,
+    sparse_merkle_path  BLOB    NOT NULL,
```
bobbinth: I'm not sure I'd change the name here. Conceptually, it is still a Merkle path just stored in a compressed form. So, maybe we could keep a name and add a comment explaining that we actually store it as a Sparse Merkle path?

crates/store/src/db/migrations/001-init.sql
```diff
@@ -31,23 +31,23 @@ CREATE TABLE accounts (
...
+    consumed            INTEGER NOT NULL, -- boolean
```
bobbinth: Not related to this PR: shouldn't `consumed` be nullable? We know if a note has been consumed only for public notes, for private notes it should technically be `NULL`. Probably doesn't make a difference in practice though.

@Mirko-von-Leipzig - what do you think?

crates/store/src/db/mod.rs
```diff
@@ -82,7 +82,7 @@ pub struct NoteRecord {
     pub note_id: Word,
     pub metadata: NoteMetadata,
     pub details: Option<NoteDetails>,
-    pub merkle_path: MerklePath,
+    pub sparse_merkle_path: SparseMerklePath,
```
bobbinth: Similar to my other comment: I would keep this as `merkle_path`. If we are not comfortable with that, we could also use `authentication_path` or `inclusion_path`.

proto/proto/merkle.proto
```diff
@@ -3,8 +3,8 @@ package merkle;
 
 import "digest.proto";
 
-// Represents a Merkle path.
-message MerklePath {
+// Represents a sparse Merkle path.
+message SparseMerklePath {
     // List of sibling node hashes, in order from the root to the leaf.
     repeated digest.Digest siblings = 1;
 }
```
bobbinth: A couple of comments:

- I don't think we can get rid of `MerklePath` because we still user regular Merkle paths in several places. Specifically, sparse Merkle paths currently apply only to note inclusion proofs. Nullifier and account proofs are regular Merkle paths still.
- I'm surprised this representation works for `SparseMarklePath`. The internal structure of the path is different from regular `MerklePath`. Specifically, the message should look something like:

```proto
message SparseMerklePath {
    // A bitmask representing empty nodes.
    //
    // The set bit corresponds to the depth of an empty node. The least significant bit (bit 0)
    // describes depth 1 node (root's children). The `bit index + 1` is equal to node's depth.
    fixed64 empty_nodes_mask = 1;
    // The non-empty nodes, stored in depth-order, but not contiguous across depth.
    repeated digest.Digest siblings = 1;
}
```

proto/proto/note.proto
```diff
@@ -75,7 +75,7 @@ message NoteInclusionInBlockProof {
     uint32 note_index_in_block = 3;
 
     // The note's inclusion proof in the block.
-    merkle.MerklePath merkle_path = 4;
+    merkle.SparseMerklePath merkle_path = 4;
```
bobbinth: Maybe we should rename this to `inclusion_path`.

proto/proto/note.proto
```diff
@@ -90,5 +90,5 @@ message NoteSyncRecord {
     NoteMetadata metadata = 3;
 
     // The note's inclusion proof in the block.
-    merkle.MerklePath merkle_path = 4;
+    merkle.SparseMerklePath merkle_path = 4;
```
bobbinth: Same comment as above.

proto/proto/responses.proto
```diff
@@ -30,8 +30,8 @@ message GetBlockHeaderByNumberResponse {
     // The requested block header.
     block.BlockHeader block_header = 1;
 
-    // Merkle path to verify the block's inclusion in the MMR at the returned `chain_length`.
-    optional merkle.MerklePath mmr_path = 2;
+    // Sparse Merkle path to verify the block's inclusion in the MMR at the returned `chain_length`.
+    optional merkle.SparseMerklePath mmr_path = 2;
```
bobbinth: This should actually remain a regular `MerklePath`.

proto/proto/responses.proto
```diff
@@ -76,11 +76,11 @@ message SyncNoteResponse {
     // Block header of the block with the first note matching the specified criteria.
     block.BlockHeader block_header = 2;
 
-    // Merkle path to verify the block's inclusion in the MMR at the returned `chain_tip`.
+    // Sparse Merkle path to verify the block's inclusion in the MMR at the returned `chain_tip`.
     //
     // An MMR proof can be constructed for the leaf of index `block_header.block_num` of
     // an MMR of forest `chain_tip` with this path.
-    merkle.MerklePath mmr_path = 3;
+    merkle.SparseMerklePath mmr_path = 3;
```
bobbinth: This should should remain a regular `MerklePath`.

proto/proto/responses.proto
```diff
@@ -99,8 +99,8 @@ message AccountWitness {
     // The state commitment whose inclusion the witness proves.
     digest.Digest commitment = 3;
 
-    // The merkle path of the state commitment in the account tree.
-    merkle.MerklePath path = 4;
+    // The sparse merkle path of the state commitment in the account tree.
+    merkle.SparseMerklePath path = 4;
```
bobbinth: Same comment as above.

proto/proto/smt.proto
```diff
@@ -37,8 +37,8 @@ message SmtLeaf {
 
 // The opening of a leaf in an SMT.
 message SmtOpening {
-    // The merkle path to the leaf.
-    merkle.MerklePath path = 1;
+    // The sparse Merkle path to the leaf.
+    merkle.SparseMerklePath path = 1;
```
bobbinth: For now, this should also be a regular `MerklePath` - though, in the future, we'll probably change this to a sparse Merkle path as well (after https://github.com/0xMiden/crypto/issues/404 is done).

crates/store/src/db/migrations/001-init.sql
Mirko-von-Leipzig: If viewed in isolation then a ternary value would be more technically correct. In practice one we always check the `note_type` and/or `tag` for private/public/network so it ends up already being a ternary.

#1089 suggests we make it an `Option<BlockNumber>` instead which would make this a moot point, but if we punt on that then I'm happy to have it be `Option<bool>` to prevent accidents.

crates/store/src/db/migrations/001-init.sql
Mirko-von-Leipzig: `sparse_inclusion_path` or `sparse_inclusion_proof`? Though I'm assuming eventually every inclusion proof would be a sparse one? If so then we can drop the prefix.

bobbinth: I would drop the prefix even now. It conveys the type of data rather than the meaning of it (e.g., we don't usually prefix other fields with underlying data type names).

proto/proto/merkle.proto
tomyrd: > I'm surprised this representation works for SparseMarklePath. The internal structure of the path is different from regular MerklePath.

Yes, now that I studied the type implementation more in depth, I think the reason my implementation works is because the `iter()` function also returns the empty nodes, so I was storing the full merkle path in the proto message.

In order to use this correct proto representation, we would need to add a new constructor for the `SparseMerklePath` as the only current way to build one is via the `from_sized_iter` which expects empty nodes.

Should I create a PR in miden crypto for this?

bobbinth: > Should I create a PR in miden crypto for this?

Yes, let's do it. You can make it against the `main` branch as it should be a non-breaking change.

tomyrd: [Done](https://github.com/0xMiden/crypto/pull/457), I also had to add an `into_parts` function to take into account the creation of the proto type in itself.

tomyrd:
Just finished the use of [the new `SparseMerklePath` constructor](https://github.com/0xMiden/crypto/pull/457) in the latest `miden-crypto` release, so this PR is ready for new re-reviews.

bobbinth (approved):
Looks good! Thank you!

Would be cool to run the benchmarks to see how much space each note takes now (I remember we had a benchmark that showed it was quite a lot - but maybe now it looks more reasonable).

Issue #1082, opened 2025-07-13
Unknown RPC endpoint being hit on Devnet Prover


sergerad:
On Devnet we have this error log every ~10s:
```
response failed 448d5811be9401b72c347a14c35cfbf7 2025-07-14 11:38:24.730 UTC+12:00
ERROR remote-prover.rpc/Unknown tower_http::trace::on_failure miden-prover

remote-prover.rpc/Unknown 448d5811be9401b72c347a14c35cfbf7 2025-07-14 11:38:24.730 UTC+12:00
Unknown 10.0.35.30 miden-prover
```

The proxy is trying to hit an endpoint that is not supported by the prover. Both are v0.10.0 (latest release candidate).

I can't replicate this locally. Going to look at adding more log information to debug it on Devnet before putting in the fix.

sergerad:
I think what is happening is that the proxy is forwarding load balancer healthcheck requests to the prover. This won't occur once we point the load balancer to the status endpoint. Just noting that we need to implement it:

```
❮ grpcurl -plaintext -import-path . -proto remote_prover.proto 127.0.0.1:8082 remote_prover.ProxyStatusApi/Status  

ERROR:
  Code: Unavailable
  Message: unexpected HTTP status code received from server: 502 (Bad Gateway); malformed header: missing HTTP content-type
```

I'll see if I can fix Devnet now and then add an impl for the status endpoint.

sergerad:
Fixed load balancer against proxy. Still seeing an error from proxy->worker:
```
response failed 35279ecc64a37a04e69d83957d57c947 2025-07-14 12:29:42.950 UTC+12:00
ERROR remote-prover.rpc/Status tower_http::trace::on_failure miden-prover
remote-prover.rpc/Status 35279ecc64a37a04e69d83957d57c947 2025-07-14 12:29:42.950 UTC+12:00
Status 10.0.35.30 miden-prover
```

sergerad:
@SantiagoPittella you might have some thoughts on how this should work in practice.

Do we intend for the proxy to have the `/Status` gRPC endpoint? Or is it supposed to be the Pingora endpoint only?

sergerad:
> Best guess is that its forwarding remote_prover.ProxyStatusApi/Status from the load balancer check.

I think I have confirmed this is the problem. So the question is how do we want to resolve this?

@SantiagoPittella @Mirko-von-Leipzig should we consolidate the worker and proxy status RPC service from
[...]

PR #1098, opened 2025-07-17
chore: handle proxy status requests in regular pipeline

SantiagoPittella:
@bobbinth @sergerad this is an example of how the status endpoint could be embedded in the regular request pipeline. The downside of this is that we need to generate a gRPC response by hand. It does not add much overhead IMO.

If we decide to go down this path, I can bring here the proxy status cache from the status service and remove the service itself.

part of #1082

### DIFF
--- a/bin/remote-prover/Cargo.toml
+++ b/bin/remote-prover/Cargo.toml
@@ -28,6 +28,7 @@ async-trait           = { version = "0.1" }
 axum                  = { version = "0.8" }
 bytes                 = { version = "1.0" }
 clap                  = { features = ["derive", "env"], version = "4.5" }
+http                  = { version = "1.0" }
 humantime             = { workspace = true }
 miden-block-prover    = { workspace = true }
 miden-lib             = { workspace = true }
--- a/bin/remote-prover/src/proxy/mod.rs
+++ b/bin/remote-prover/src/proxy/mod.rs
@@ -193,6 +193,16 @@ impl LoadBalancerState {
     pub async fn num_busy_workers(&self) -> usize {
         self.workers.read().await.iter().filter(|w| !w.is_available()).count()
     }
+
+    /// Get the supported proof type.
+    pub fn supported_proof_type(&self) -> ProofType {
+        self.supported_proof_type
+    }
+
+    /// Get a read lock on the workers.
+    pub async fn workers(&self) -> tokio::sync::RwLockReadGuard<'_, Vec<Worker>> {
+        self.workers.read().await
+    }
 }
 
 /// Rate limiter
@@ -355,6 +365,14 @@ impl ProxyHttp for LoadBalancer {
         };
 
         info!("Client address: {:?}", client_addr);
+
+        // Check if the request is a grpc proxy status request by checking the path
+        if session.downstream_session.req_header().as_ref().uri.path()
+            == "/remote_prover.ProxyStatusApi/Status"
+        {
+            info!("Proxy status request");
+            return crate::utils::create_proxy_status_response(session, &self.0).await;
+        }
 
         // Increment the request count
         REQUEST_COUNT.inc();
--- a/bin/remote-prover/src/utils.rs
+++ b/bin/remote-prover/src/utils.rs
@@ -1,13 +1,78 @@
 use std::net::TcpListener;
 
-use miden_remote_prover::error::RemoteProverError;
+use http::HeaderMap;
+use miden_remote_prover::{
+    error::RemoteProverError,
+    generated::remote_prover::{ProxyStatusResponse, WorkerStatus},
+};
 use pingora::{Error, ErrorType, http::ResponseHeader, protocols::http::ServerSession};
 use pingora_proxy::Session;
+use prost::Message;
 use tracing::debug;
 
-use crate::{COMPONENT, commands::PROXY_HOST, proxy::metrics::QUEUE_DROP_COUNT};
+use crate::{
+    COMPONENT,
+    commands::PROXY_HOST,
+    proxy::{LoadBalancerState, metrics::QUEUE_DROP_COUNT},
+};
 
 const RESOURCE_EXHAUSTED_CODE: u16 = 8;
+
+/// Create a gRPC proxy status response
+///
+/// This creates a proper gRPC response with the `ProxyStatusResponse` serialized as protobuf
+pub async fn create_proxy_status_response(
+    session: &mut Session,
+    load_balancer_state: &LoadBalancerState,
+) -> pingora_core::Result<bool> {
+    // Build the proxy status response
+    let version = env!("CARGO_PKG_VERSION").to_string();
+    let supported_proof_type: i32 = load_balancer_state.supported_proof_type().into();
+
+    let workers = load_balancer_state.workers().await;
+    let worker_statuses: Vec<WorkerStatus> = workers.iter().map(WorkerStatus::from).collect();
+
+    let status_response = ProxyStatusResponse {
+        version,
+        supported_proof_type,
+        workers: worker_statuses,
+    };
+
+    // Serialize the protobuf message
+    let mut response_body = Vec::new();
+    status_response.encode(&mut response_body).map_err(|e| {
+        Error::new(ErrorType::InternalError)
+            .more_context(format!("Failed to encode proto response: {e}"))
+    })?;
+
+    let mut grpc_message = Vec::new();
+
+    // Add compression flag (1 byte, 0 = no compression)
+    grpc_message.push(0u8);
+
+    // Add message length (4 bytes, big-endian)
+    let msg_len = response_body.len() as u32;
+    grpc_message.extend_from_slice(&msg_len.to_be_bytes());
+
+    // Add the actual message
+    grpc_message.extend_from_slice(&response_body);
+
+    // Create gRPC response headers WITHOUT grpc-status (that goes in trailers)
+    let mut header = ResponseHeader::build(200, None)?;
+    header.insert_header("content-type", "application/grpc".to_string())?;
+    // Don't set grpc-status here - it must be in trailers for proper gRPC
+
+    session.set_keepalive(None);
+    session.write_response_header(Box::new(header), false).await?;
+    session.write_response_body(Some(grpc_message.into()), false).await?;
+
+    // Send trailers with gRPC status
+    let mut trailers = HeaderMap::new();
+    trailers.insert("grpc-status", "0".parse().map_err(|_| Error::new(ErrorType::InternalError))?);
+    session.write_response_trailers(trailers).await?;
+
+    Ok(true)
+}
 
 /// Create a 503 response for a full queue
 pub(crate) async fn create_queue_full_response(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2662,6 +2662,7 @@ dependencies = [
  "axum",
  "bytes",
  "clap 4.5.41",
+ "http 1.3.1",
  "humantime",
  "miden-block-prover",
  "miden-lib",


bin/remote-prover/src/utils.rs
```diff
@@ -1,14 +1,79 @@
...
+    session.write_response_header(Box::new(header), false).await?;
+    session.write_response_body(Some(grpc_message.into()), false).await?;
+
+    // Send trailers with gRPC status
+    let mut trailers = HeaderMap::new();
+    trailers.insert("grpc-status", "0".parse().map_err(|_| Error::new(ErrorType::InternalError))?);
+    session.write_response_trailers(trailers).await?;
+
+    Ok(true)
+}
```
sergerad, line 94: >If we decide to go down this path, I can bring here the proxy status cache from the status service and remove the service itself.

I guess that is a benefit of this approach? And we wouldn't run anything on 8084?

bin/remote-prover/src/utils.rs
```diff
@@ -1,14 +1,79 @@
...
+    // Don't set grpc-status here - it must be in trailers for proper gRPC
+
+    session.set_keepalive(None);
+    session.write_response_header(Box::new(header), false).await?;
+    session.write_response_body(Some(grpc_message.into()), false).await?;
+
+    // Send trailers with gRPC status
+    let mut trailers = HeaderMap::new();
+    trailers.insert("grpc-status", "0".parse().map_err(|_| Error::new(ErrorType::InternalError))?);
+    session.write_response_trailers(trailers).await?;
```
bobbinth: Is there a way to avoid building the response manually?

And do we also build gRPC responses manually for the `prove` endpoint? (it's been a while since I looked a the code).

SantiagoPittella: > Is there a way to avoid building the response manually?

I couldn't find a way to avoid this. Basically pingora knows nothing about grpc, it just forwards requests without caring about the format. So I needed to replicate the gRPC response standard from HTTP.

> And do we also build gRPC responses manually for the prove endpoint? (it's been a while since I looked a the code).

For the `Prove` method, the response is handled by worker itself so there's no need to build anything. We have 2 cases where the proxy sends an early response to the user:
  1. The queue is full.
  2. The user is rate-limited.

For this cases, we do create a response manually, but it is much simpler because it is just setting the headers:

```rust
let mut header = ResponseHeader::build(503, None)?;
header.insert_header("grpc-message", "Too many requests in the queue".to_string())?;
header.insert_header("grpc-status", RESOURCE_EXHAUSTED_CODE)?;
session.set_keepalive(None);
session.write_response_header(Box::new(header.clone()), true).await?;|
```

bin/remote-prover/src/utils.rs
SantiagoPittella: Yes, in this case we can remove the status service as a separate service.

bin/remote-prover/src/utils.rs
Mirko-von-Leipzig: I _think_ we might be able to do something with `tonic` like:

```rust
// Or use resource_exhausted - though I think that's intended for a permanent
// situation, whereas unavailable is temporary.
let response = tonic::Status::unavailable("Too many requests in queue")
  .into_http();      // This is now an http::Response

// You can create the other gRPC responses similarly.
let response = tonic::Response(...).into_response();

// And then have a function which takes an http response and sends it on the session.
// Maybe pingora already supports this natively? That would make sense to me..
```

Mirko-von-Leipzig: The other alternative would be to host the gRPC service locally as well and route it there instead? Though that does seem a bit silly :)

SantiagoPittella: > The other alternative would be to host the gRPC service locally as well and route it there instead? Though that does seem a bit silly :)

Thought about it but looked way more overkill than needed, considering that it is just one endpoint.

SantiagoPittella: > I think we might be able to do something with tonic like:

Will give it a try.

SantiagoPittella: Couldn't find anything useful in the tonic crate to convert responses into pingora's `Session`.

bobbinth: As far as I can tell, `tonic::Response::into_http()` is not a public method (see [here](https://github.com/hyperium/tonic/blob/master/tonic/src/response.rs#L82)) - though it is not that complex either.

I think maybe what we should do is write a helper function which could write a Tonic response into a Pingora session. Something like:
```rust
fn write_tonic_response_to_session<T>(session: &mut Session, response: tonic::Response<T>) {
    ...
}
```
Or would this complicate things?

Mirko-von-Leipzig: Its `pub` on docs: https://docs.rs/tonic/latest/tonic/struct.Status.html#method.into_http, maybe they re-export from elsewhere though that's weird.

SantiagoPittella: I ended up doing something like, check this commit https://github.com/0xMiden/miden-node/pull/1098/commits/9258c11fbf45e99957fc9f9c342657a4281c3c1b

bin/remote-prover/src/proxy/mod.rs
```diff
@@ -193,8 +217,32 @@ impl LoadBalancerState {
...
+        let version = env!("CARGO_PKG_VERSION").to_string();
+        let supported_proof_type: i32 = self.supported_proof_type.into();
+        let workers = self.workers.read().await;
+        let worker_statuses: Vec<WorkerStatus> = workers.iter().map(WorkerStatus::from).collect();
+
+        let new_status = ProxyStatusResponse {
+            version,
+            supported_proof_type,
+            workers: worker_statuses,
+        };
```
sergerad: nit: Consider deduplicating retrieval of status response. Bit of repetition with line 103 above.

bin/remote-prover/src/proxy/mod.rs
```diff
@@ -356,6 +404,15 @@ impl ProxyHttp for LoadBalancer {
 
         info!("Client address: {:?}", client_addr);
 
+        // Check if the request is a grpc proxy status request by checking the path
+        if session.downstream_session.req_header().as_ref().uri.path()
+            == "/remote_prover.ProxyStatusApi/Status"
+        {
+            info!("Proxy status request");
```
sergerad: We would see this log every ~10s for each target group we have pointing at the node. I would remove it as it doesn't really add value for our ops.

sergerad: Ideally we would add some fields to the span of this fn pertaining to the request instead. But we can look at that separately.

bin/remote-prover/src/proxy/status_cache.rs
```diff
@@ -0,0 +1,52 @@
...
+        Self {
+            address: worker.address(),
+            version: worker.version().to_string(),
+            status: match worker.health_status() {
+                WorkerHealthStatus::Healthy => ProtoWorkerHealthStatus::Healthy,
+                WorkerHealthStatus::Unhealthy { .. } => ProtoWorkerHealthStatus::Unhealthy,
+                WorkerHealthStatus::Unknown => ProtoWorkerHealthStatus::Unknown,
+            } as i32,
+        }
+    }
```
bobbinth: nit: is this the right place for this code? It doesn't look like we do anything either with `Worker` or with `WorkerStatus` in this module. Maybe the code should be where `Worker` or `WorkerStatus` are defined?

bin/remote-prover/src/proxy/mod.rs
bobbinth: Yeah - maybe we could have a small helper function that takes `supported_proof_type` and `workers` as an input and returns `ProxyStatusResponse`.

bin/remote-prover/src/utils.rs
```diff
@@ -1,52 +1,124 @@
...
 
-const RESOURCE_EXHAUSTED_CODE: u16 = 8;
+/// Write a protobuf message as a gRPC response to a Pingora session
+///
+/// This helper function takes a protobuf message and writes it to a Pingora session
+/// in the proper gRPC format, handling message encoding, headers, and trailers.
+pub async fn write_grpc_response_to_session<T>(
+    session: &mut Session,
+    message: T,
+    grpc_status: Option<&str>,
```
bobbinth: Why do we pass the `grpc_status` here? Doesn't seem like we actually ever use it.

Also, base don how we use it below in the function, if we do want to keep it as a parameter, should we not pass it as `Code`?

bin/remote-prover/src/utils.rs
```diff
@@ -1,52 +1,124 @@
...
+    let status_code = (grpc_status as i32).to_string();
+    trailers.insert(
+        "grpc-status",
+        status_code.parse().map_err(|_| Error::new(ErrorType::InternalError))?,
+    );
+
+    trailers.insert(
+        "grpc-message",
+        error_message.parse().map_err(|_| Error::new(ErrorType::InternalError))?,
+    );
```
bobbinth: nit: maybe we can extract this out into a small helper function - something like:
```rust
fn build_grpc_trailers(grpc_status: Code, err_message: Option<&str>) -> HeaderMap {
    ...
}
```
Then we'd be able to use it here and also instead of lines 53 - 58 above.

bobbinth (approved):
Looks good! Thank you! I left a few small nits inlines.

bin/remote-prover/Cargo.toml
```diff
@@ -28,6 +28,7 @@ async-trait           = { version = "0.1" }
 axum                  = { version = "0.8" }
 bytes                 = { version = "1.0" }
 clap                  = { features = ["derive", "env"], version = "4.5" }
+http                  = { version = "1.0" }
```
Mirko-von-Leipzig:
```suggestion
http                  = { workspace = true }
```

bin/remote-prover/src/proxy/health_check.rs
Mirko-von-Leipzig, line 1: Unrelated but shouldn't we be using the `ShutdownWatch`?

bin/remote-prover/src/proxy/status_cache.rs
Mirko-von-Leipzig, line 1: I think this could be replaced by a [watch channel](https://docs.rs/tokio/latest/tokio/sync/watch/index.html)?

bin/remote-prover/src/utils.rs
```diff
@@ -1,52 +1,124 @@
...
+    grpc_status: Option<&str>,
+) -> pingora_core::Result<bool>
+where
+    T: Message,
+{
+    // Serialize the protobuf message
+    let mut response_body = Vec::new();
+    message.encode(&mut response_body).map_err(|e| {
+        Error::new(ErrorType::InternalError)
+            .more_context(format!("Failed to encode proto response: {e}"))
```
Mirko-von-Leipzig: I _think_ [`Error::because`](https://docs.rs/pingora/latest/pingora/struct.Error.html#method.because) might be more appropriate for this chaining.

```suggestion
        Error::because(ErrorType::InternalError, "failed to encode gRPC response", e)
```

bin/remote-prover/src/utils.rs
```diff
@@ -1,52 +1,124 @@
...
+    // Add message length (4 bytes, big-endian)
+    let msg_len = response_body.len() as u32;
+    grpc_message.extend_from_slice(&msg_len.to_be_bytes());
+
+    // Add the actual message
+    grpc_message.extend_from_slice(&response_body);
+
+    // Create gRPC response headers WITHOUT grpc-status (that goes in trailers)
+    let mut header = ResponseHeader::build(200, None)?;
+    header.insert_header("content-type", "application/grpc".to_string())?;
```
Mirko-von-Leipzig: Also elsewhere similarly (move the constant to somewhere appropriate I guess).

```suggestion
    const GRPC_CONTENT: HeaderValue = HeaderValue::from_static("application/grpc");
    header.insert_header(http::header::CONTENT_TYPE, GRPC_CONTENT).unwrap();
```

bin/remote-prover/src/utils.rs
```diff
@@ -1,52 +1,124 @@
...
     QUEUE_DROP_COUNT.inc();
 
-    Err(error)
+    // Use our helper function to create a proper gRPC error response
+    write_grpc_error_to_session(session, Code::ResourceExhausted, "Too many requests in the queue")
+        .await
 }
 
-/// Create a 429 response for too many requests
+/// Create a gRPC `RESOURCE_EXHAUSTED` response for rate limiting
```
Mirko-von-Leipzig, line 132: I'm actually a bit unsure about moving rate-limiting away from http. As a general rule, 429 is well understood and _should_ usually include the backoff time i.e. when the client may retry without being rejected.

I guess I'm also surprised `pingora` doesn't include this itself somehow?

Mirko-von-Leipzig: Is this a global rate limit, or limit per client connection?

bin/remote-prover/src/proxy/mod.rs
SantiagoPittella: I removed it and added the path to the span ( https://github.com/0xMiden/miden-node/pull/1098/commits/ff89263e5bbd29d2920d08032345eb3849c4c739 ), was that correct?

bin/remote-prover/src/proxy/status_cache.rs
SantiagoPittella: moved it to `worker.rs`.

bin/remote-prover/src/proxy/health_check.rs
SantiagoPittella: yes, it should check in every iteration of the service should be terminated.

bin/remote-prover/src/proxy/status_cache.rs
SantiagoPittella: I give it a try, and was able to delete the whole status cache. It was done in this commit: https://github.com/0xMiden/miden-node/pull/1098/commits/6ff2d91bfb117f8b303b449a22058e35cf55dc8d

bin/remote-prover/src/proxy/mod.rs
```diff
@@ -342,6 +378,9 @@ impl ProxyHttp for LoadBalancer {
     where
         Self::CTX: Send + Sync,
     {
+        let path = session.downstream_session.req_header().uri.path();
+        Span::current().record("path", path);
+
```
igamigo: nit: May be better to move this down, after the client address check

bin/remote-prover/src/utils.rs
SantiagoPittella: I basically copied what pingora describes as [rate limiter](https://github.com/cloudflare/pingora/blob/main/docs/user_guide/rate_limiter.md), counting by client address.

SantiagoPittella: Pingora provides the `Rate` struct, which tracks events in a certain period of time.

bin/remote-prover/src/proxy/mod.rs
```diff
@@ -356,6 +395,12 @@ impl ProxyHttp for LoadBalancer {
 
         info!("Client address: {:?}", client_addr);
 
+        // Check if the request is a grpc proxy status request by checking the path
+        if path == "/remote_prover.ProxyStatusApi/Status" {
```
igamigo: You were probably keeping this in mind, but would be nice to explain on the doc comments that explain the flow of a request that this short-circuits on this specific gRPC request

bin/remote-prover/src/utils.rs
```diff
@@ -1,52 +1,138 @@
...
+pub async fn write_grpc_response_to_session<T>(
     session: &mut Session,
-) -> pingora_core::Result<bool> {
-    // Set grpc-message header to "Too many requests in the queue"
-    // This is meant to be used by a Tonic interceptor to return a gRPC error
-    let mut header = ResponseHeader::build(503, None)?;
-    header.insert_header("grpc-message", "Too many requests in the queue".to_string())?;
-    header.insert_header("grpc-status", RESOURCE_EXHAUSTED_CODE)?;
+    message: T,
+) -> pingora_core::Result<bool>
```
igamigo: nit: Without the context, it's not trivial to realize why this returns `bool`. Maybe this could return `Result<()>` and then the calling function returns `Ok(true)` (or alternatively we could just add a comment)

SantiagoPittella: Done! did it in the last commit, wdyt?

Mirko-von-Leipzig (approved):

PR #1100, opened 2025-07-21
feat: Add store endpoint for retrieving unconsumed network notes of a network account

### LINKED ISSUE #1089: Track _when_ a network note is consumed and expose this via gRPC

sergerad:
WIP

Closes #1089.

### DIFF
--- a/crates/store/src/db/migrations/001-init.sql
+++ b/crates/store/src/db/migrations/001-init.sql
@@ -42,7 +42,7 @@ CREATE TABLE notes (
     aux            INTEGER NOT NULL,
     execution_hint INTEGER NOT NULL,
     merkle_path    BLOB    NOT NULL,
-    consumed       INTEGER NOT NULL, -- boolean
+    consumed       INTEGER,          -- 32-bit block number
     nullifier      BLOB,             -- Only known for public notes, null for private notes
     assets         BLOB,
     inputs         BLOB,
@@ -69,7 +69,7 @@ CREATE INDEX idx_unconsumed_network_notes ON notes(execution_mode, consumed);
 CREATE TABLE note_scripts (
     script_root BLOB NOT NULL,
     script      BLOB NOT NULL,
-  
+
     PRIMARY KEY (script_root)
 ) STRICT, WITHOUT ROWID;
 
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -6,7 +6,7 @@ use std::{
 use anyhow::Context;
 use miden_lib::utils::Serializable;
 use miden_node_proto::{
-    domain::account::{AccountInfo, AccountSummary},
+    domain::account::{AccountInfo, AccountSummary, NetworkAccountPrefix},
     generated::note as proto,
 };
 use miden_objects::{
@@ -555,6 +555,25 @@ impl Db {
     ) -> Result<(Vec<NoteRecord>, Page)> {
         self.transact("unconsumed network notes", move |conn| {
             sql::unconsumed_network_notes(conn, page)
+        })
+        .await
+    }
+
+    /// Loads the network notes for a network account that have not been consumed yet, using
+    /// pagination to limit the number of notes returned.
+    pub(crate) async fn select_unconsumed_network_notes_for_network_account(
+        &self,
+        network_account_id_prefix: NetworkAccountPrefix,
+        block_num: BlockNumber,
+        page: Page,
+    ) -> Result<(Vec<NoteRecord>, Page)> {
+        self.transact("unconsumed network notes", move |conn| {
+            sql::unconsumed_network_notes_for_network_account(
+                conn,
+                network_account_id_prefix,
+                block_num,
+                page,
+            )
         })
         .await
     }
--- a/crates/store/src/db/sql/mod.rs
+++ b/crates/store/src/db/sql/mod.rs
@@ -1135,6 +1135,55 @@ pub fn unconsumed_network_notes(
     Ok((notes, page))
 }
 
+/// ...
+pub fn unconsumed_network_notes_for_network_account(
+    transaction: &Transaction,
+    network_account_id_prefix: NetworkAccountPrefix,
+    block_num: BlockNumber,
+    mut page: Page,
+) -> Result<(Vec<NoteRecord>, Page)> {
+    assert_eq!(
+        NoteExecutionMode::Network as u8,
+        0,
+        "Hardcoded execution value must match query"
+    );
+
+    // Select the rowid column so that we can return a pagination token.
+    //
+    // rowid column _must_ come after the note fields so that we don't mess up the
+    // `NoteRecord::from_row` call.
+    let mut stmt = transaction.prepare_cached(&format!(
+        "
+        SELECT {}, rowid
+        FROM notes
+        LEFT JOIN note_scripts ON notes.script_root = note_scripts.script_root
+        WHERE
+            execution_mode = 0 AND consumed = FALSE AND
+            block_num = ? AND rowid >= ?
+        ORDER BY rowid
+        LIMIT ?
+        ",
+        NoteRecord::SELECT_COLUMNS
+    ))?;
+
+    // The `page.size` is the maximum number of notes to return. We add 1 to it so that we can
+    // check if there are more notes for the next page.
+    let mut rows =
+        stmt.query(params![page.token.unwrap_or(0), block_num.as_u32(), page.size.get() + 1])?;
+
+    page.token = None;
+    let mut notes = Vec::with_capacity(page.size.into());
+    while let Some(row) = rows.next()? {
+        if notes.len() == page.size.get() {
+            page.token = Some(row.get::<_, u64>(14)?);
+            break;
+        }
+        notes.push(NoteRecord::from_row(row)?);
+    }
+
+    Ok((notes, page))
+}
+
 // BLOCK CHAIN QUERIES
 // ================================================================================================
 
--- a/crates/store/src/server/ntx_builder.rs
+++ b/crates/store/src/server/ntx_builder.rs
@@ -5,7 +5,8 @@ use miden_node_proto::{
     generated::{
         requests::{
             GetBlockHeaderByNumberRequest, GetCurrentBlockchainDataRequest,
-            GetNetworkAccountDetailsByPrefixRequest, GetUnconsumedNetworkNotesRequest,
+            GetNetworkAccountDetailsByPrefixRequest,
+            GetUnconsumedNetworkNotesForNetworkAccountRequest, GetUnconsumedNetworkNotesRequest,
         },
         responses::{
             GetBlockHeaderByNumberResponse, GetCurrentBlockchainDataResponse,
@@ -137,6 +138,57 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
         // instead
         let (notes, next_page) =
             state.get_unconsumed_network_notes(page).await.map_err(internal_error)?;
+
+        let mut network_notes = Vec::with_capacity(notes.len());
+        for note in notes {
+            // SAFETY: Network notes are filtered in the database, so they should have details;
+            // otherwise the state would be corrupted
+            let (assets, recipient) = note.details.unwrap().into_parts();
+            let note = Note::new(assets, note.metadata, recipient);
+            network_notes.push(note.into());
+        }
+
+        Ok(Response::new(GetUnconsumedNetworkNotesResponse {
+            notes: network_notes,
+            next_token: next_page.token,
+        }))
+    }
+
+    #[instrument(
+        parent = None,
+        target = COMPONENT,
+        name = "store.ntx_builder_server.get_unconsumed_network_notes_for_network_account",
+        skip_all,
+        err
+    )]
+    async fn get_unconsumed_network_notes_for_network_account(
+        &self,
+        request: Request<GetUnconsumedNetworkNotesForNetworkAccountRequest>,
+    ) -> Result<Response<GetUnconsumedNetworkNotesResponse>, Status> {
+        let request = request.into_inner();
+        let block_num = BlockNumber::from(request.block_num);
+        let network_account_id_prefix =
+            NetworkAccountPrefix::try_from(request.network_account_id_prefix).map_err(|err| {
+                invalid_argument(err.as_report_context("invalid network_account_id_prefix"))
+            })?;
+
+        let state = self.state.clone();
+
+        let size =
+            NonZero::try_from(request.page_size as usize).map_err(|err: TryFromIntError| {
+                invalid_argument(err.as_report_context("invalid page_size"))
+            })?;
+        let page = Page { token: request.page_token, size };
+        // TODO: no need to get the whole NoteRecord here, a NetworkNote wrapper should be created
+        // instead
+        let (notes, next_page) = state
+            .get_unconsumed_network_notes_for_network_account(
+                network_account_id_prefix,
+                block_num,
+                page,
+            )
+            .await
+            .map_err(internal_error)?;
 
         let mut network_notes = Vec::with_capacity(notes.len());
         for note in notes {
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -12,7 +12,7 @@ use std::{
 use miden_node_proto::{
     AccountWitnessRecord,
     domain::{
-        account::{AccountInfo, AccountProofRequest, StorageMapKeysProof},
+        account::{AccountInfo, AccountProofRequest, NetworkAccountPrefix, StorageMapKeysProof},
         batch::BatchInputs,
     },
     generated::responses::{AccountProofsResponse, AccountStateHeader, StorageSlotMapProof},
@@ -965,6 +965,23 @@ impl State {
         page: Page,
     ) -> Result<(Vec<NoteRecord>, Page), DatabaseError> {
         self.db.select_unconsumed_network_notes(page).await
+    }
+
+    /// Returns the unprocessed network notes for a network account, along with the next pagination token.
+    pub async fn get_unconsumed_network_notes_for_network_account(
+        &self,
+        network_account_id_prefix: NetworkAccountPrefix,
+        block_num: BlockNumber,
+        page: Page,
+    ) -> Result<(Vec<NoteRecord>, Page), DatabaseError> {
+        // TODO: check account exists
+        self.db
+            .select_unconsumed_network_notes_for_network_account(
+                network_account_id_prefix,
+                block_num,
+                page,
+            )
+            .await
     }
 }
 
--- a/proto/proto/requests.proto
+++ b/proto/proto/requests.proto
@@ -215,3 +215,22 @@ message GetUnconsumedNetworkNotesRequest {
     // Number of notes to retrieve per page.
     uint64 page_size = 2;
 }
+
+// Returns a list of unconsumed network notes for a network account using pagination.
+message GetUnconsumedNetworkNotesForNetworkAccountRequest {
+    // An opaque token used to paginate through the notes.
+    //
+    // This should be null on the first call, and set to the response
+    // token until the response token is null, at which point all data
+    // has been fetched.
+    optional uint64 page_token = 1;
+
+    // Number of notes to retrieve per page.
+    uint64 page_size = 2;
+
+    // ...
+    uint32 network_account_id_prefix = 3;
+
+    // ...
+    fixed32 block_num = 4;
+}
--- a/proto/proto/store.proto
+++ b/proto/proto/store.proto
@@ -46,7 +46,7 @@ service Rpc {
     // The response includes each note's metadata and inclusion proof.
     //
     // A basic note sync can be implemented by repeatedly requesting the previous response's block until reaching the
-    // tip of the chain. 
+    // tip of the chain.
     rpc SyncNotes(requests.SyncNoteRequest) returns (responses.SyncNoteResponse) {}
 
     // Returns info which can be used by the client to sync up to the latest state of the chain
@@ -98,8 +98,11 @@ service NtxBuilder {
     // Returns the list of unconsumed network notes and the next page number to query.
     rpc GetUnconsumedNetworkNotes(requests.GetUnconsumedNetworkNotesRequest) returns (responses.GetUnconsumedNetworkNotesResponse) {}
 
-    // Returns the block header at the chain tip, as well as the MMR peaks corresponding to this 
-    // header for executing network transactions. 
+    // Returns the list of unconsumed network notes for a network account and the next page number to query.
+    rpc GetUnconsumedNetworkNotesForNetworkAccount(requests.GetUnconsumedNetworkNotesForNetworkAccountRequest) returns (responses.GetUnconsumedNetworkNotesResponse) {}
+
+    // Returns the block header at the chain tip, as well as the MMR peaks corresponding to this
+    // header for executing network transactions.
     rpc GetCurrentBlockchainData(requests.GetCurrentBlockchainDataRequest) returns (responses.GetCurrentBlockchainDataResponse) {}
 
     // Returns the latest state of a network account with the specified account prefix.
--- a/crates/proto/src/generated/requests.rs
+++ b/crates/proto/src/generated/requests.rs
@@ -242,3 +242,23 @@ pub struct GetUnconsumedNetworkNotesRequest {
     #[prost(uint64, tag = "2")]
     pub page_size: u64,
 }
+/// Returns a list of unconsumed network notes for a network account using pagination.
+#[derive(Clone, Copy, PartialEq, ::prost::Message)]
+pub struct GetUnconsumedNetworkNotesForNetworkAccountRequest {
+    /// An opaque token used to paginate through the notes.
+    ///
+    /// This should be null on the first call, and set to the response
+    /// token until the response token is null, at which point all data
+    /// has been fetched.
+    #[prost(uint64, optional, tag = "1")]
+    pub page_token: ::core::option::Option<u64>,
+    /// Number of notes to retrieve per page.
+    #[prost(uint64, tag = "2")]
+    pub page_size: u64,
+    /// ...
+    #[prost(uint32, tag = "3")]
+    pub network_account_id_prefix: u32,
+    /// ...
+    #[prost(fixed32, tag = "4")]
+    pub block_num: u32,
+}
--- a/crates/proto/src/generated/store.rs
+++ b/crates/proto/src/generated/store.rs
@@ -774,6 +774,38 @@ pub mod ntx_builder_client {
                 );
             self.inner.unary(req, path, codec).await
         }
+        /// Returns the list of unconsumed network notes for a network account and the next page number to query.
+        pub async fn get_unconsumed_network_notes_for_network_account(
+            &mut self,
+            request: impl tonic::IntoRequest<
+                super::super::requests::GetUnconsumedNetworkNotesForNetworkAccountRequest,
+            >,
+        ) -> std::result::Result<
+            tonic::Response<super::super::responses::GetUnconsumedNetworkNotesResponse>,
+            tonic::Status,
+        > {
+            self.inner
+                .ready()
+                .await
+                .map_err(|e| {
+                    tonic::Status::unknown(
+                        format!("Service was not ready: {}", e.into()),
+                    )
+                })?;
+            let codec = tonic::codec::ProstCodec::default();
+            let path = http::uri::PathAndQuery::from_static(
+                "/store.NtxBuilder/GetUnconsumedNetworkNotesForNetworkAccount",
+            );
+            let mut req = request.into_request();
+            req.extensions_mut()
+                .insert(
+                    GrpcMethod::new(
+                        "store.NtxBuilder",
+                        "GetUnconsumedNetworkNotesForNetworkAccount",
+                    ),
+                );
+            self.inner.unary(req, path, codec).await
+        }
         /// Returns the block header at the chain tip, as well as the MMR peaks corresponding to this
         /// header for executing network transactions.
         pub async fn get_current_blockchain_data(
@@ -2066,6 +2098,16 @@ pub mod ntx_builder_server {
             tonic::Response<super::super::responses::GetUnconsumedNetworkNotesResponse>,
             tonic::Status,
         >;
+        /// Returns the list of unconsumed network notes for a network account and the next page number to query.
+        async fn get_unconsumed_network_notes_for_network_account(
+            &self,
+            request: tonic::Request<
+                super::super::requests::GetUnconsumedNetworkNotesForNetworkAccountRequest,
+            >,
+        ) -> std::result::Result<
+            tonic::Response<super::super::responses::GetUnconsumedNetworkNotesResponse>,
+            tonic::Status,
+        >;
         /// Returns the block header at the chain tip, as well as the MMR peaks corresponding to this
         /// header for executing network transactions.
         async fn get_current_blockchain_data(
@@ -2256,6 +2298,62 @@ pub mod ntx_builder_server {
                     let inner = self.inner.clone();
                     let fut = async move {
                         let method = GetUnconsumedNetworkNotesSvc(inner);
+                        let codec = tonic::codec::ProstCodec::default();
+                        let mut grpc = tonic::server::Grpc::new(codec)
+                            .apply_compression_config(
+                                accept_compression_encodings,
+                                send_compression_encodings,
+                            )
+                            .apply_max_message_size_config(
+                                max_decoding_message_size,
+                                max_encoding_message_size,
+                            );
+                        let res = grpc.unary(method, req).await;
+                        Ok(res)
+                    };
+                    Box::pin(fut)
+                }
+                "/store.NtxBuilder/GetUnconsumedNetworkNotesForNetworkAccount" => {
+                    #[allow(non_camel_case_types)]
+                    struct GetUnconsumedNetworkNotesForNetworkAccountSvc<T: NtxBuilder>(
+                        pub Arc<T>,
+                    );
+                    impl<
+                        T: NtxBuilder,
+                    > tonic::server::UnaryService<
+                        super::super::requests::GetUnconsumedNetworkNotesForNetworkAccountRequest,
+                    > for GetUnconsumedNetworkNotesForNetworkAccountSvc<T> {
+                        type Response = super::super::responses::GetUnconsumedNetworkNotesResponse;
+                        type Future = BoxFuture<
+                            tonic::Response<Self::Response>,
+                            tonic::Status,
+                        >;
+                        fn call(
+                            &mut self,
+                            request: tonic::Request<
+                                super::super::requests::GetUnconsumedNetworkNotesForNetworkAccountRequest,
+                            >,
+                        ) -> Self::Future {
+                            let inner = Arc::clone(&self.0);
+                            let fut = async move {
+                                <T as NtxBuilder>::get_unconsumed_network_notes_for_network_account(
+                                        &inner,
+                                        request,
+                                    )
+                                    .await
+                            };
+                            Box::pin(fut)
+                        }
+                    }
+                    let accept_compression_encodings = self.accept_compression_encodings;
+                    let send_compression_encodings = self.send_compression_encodings;
+                    let max_decoding_message_size = self.max_decoding_message_size;
+                    let max_encoding_message_size = self.max_encoding_message_size;
+                    let inner = self.inner.clone();
+                    let fut = async move {
+                        let method = GetUnconsumedNetworkNotesForNetworkAccountSvc(
+                            inner,
+                        );
                         let codec = tonic::codec::ProstCodec::default();
                         let mut grpc = tonic::server::Grpc::new(codec)
                             .apply_compression_config(


crates/store/src/db/sql/mod.rs
```diff
@@ -1135,6 +1135,55 @@ pub fn unconsumed_network_notes(
...
+        "
+        SELECT {}, rowid
+        FROM notes
+        LEFT JOIN note_scripts ON notes.script_root = note_scripts.script_root
+        WHERE
+            execution_mode = 0 AND consumed = FALSE AND
+            block_num = ? AND rowid >= ?
+        ORDER BY rowid
+        LIMIT ?
+        ",
```
sergerad: @Mirko-von-Leipzig Do we need to make a schema change in order to correlate `notes` table with network account id prefixes?

Really early draft, try ignore most of it.

Mirko-von-Leipzig: I don't think that's required. You should be able to convert the prefix into the note tag and filter for the account using that.

If you check the `TryFrom<NoteTag> for NetworkAccountPrefix` you'll notice that the note tag is a simple cast once the pre-conditions are met. So going in reverse is essentially a no-op at the moment.

So you can filter for the matching notes by adding `tag = {network_account_id_prefix}`.

crates/store/src/db/sql/mod.rs
```diff
@@ -1135,6 +1135,59 @@ pub fn unconsumed_network_notes(
     Ok((notes, page))
 }
 
+/// ...
+pub fn unconsumed_network_notes_for_network_account(
```
sergerad: @Mirko-von-Leipzig I think the intention is that this functionality is part of the original `unconsumed_network_notes()` perhaps with added arg:
```rust
params: Option<(NetworkAccountPrefix, BlockNumber)>
```

But I suppose the impl of that will change with the diesel PR merged? I'm just keeping them separate for now in anticipation of that.

Mirko-von-Leipzig: Keeping them separate makes sense. The original may eventually become unnecessary but there's a bunch of other things to do before that's an option.

crates/store/src/db/tests.rs
```diff
@@ -449,6 +450,17 @@ fn sql_unconsumed_network_notes() {
...
+        &db_tx,
+        network_account_id.try_into().unwrap(),
+        block_num,
+        Page {
+            token: None,
+            size: NonZeroUsize::new(N as usize * 10).unwrap(),
+        },
+    )
+    .unwrap();
+    assert_eq!(result, network_notes);
```
sergerad, line 452: This is probably an offensive way to test the new functionality (by just pasting it into the existing test for the similar existing fn).

LMK if you have any ideas how better to test this.

sergerad: @Mirko-von-Leipzig

crates/store/src/server/ntx_builder.rs
```diff
@@ -152,4 +153,55 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
...
+            let (assets, recipient) = note.details.unwrap().into_parts();
+            let note = Note::new(assets, note.metadata, recipient);
+            network_notes.push(note.into());
+        }
+
+        Ok(Response::new(GetUnconsumedNetworkNotesResponse {
+            notes: network_notes,
+            next_token: next_page.token,
+        }))
+    }
```
sergerad, line 192: Another basically-duplicated fn that we could potentially merge.

proto/proto/store.proto
```diff
@@ -98,8 +98,11 @@ service NtxBuilder {
     // Returns the list of unconsumed network notes and the next page number to query.
     rpc GetUnconsumedNetworkNotes(requests.GetUnconsumedNetworkNotesRequest) returns (responses.GetUnconsumedNetworkNotesResponse) {}
 
-    // Returns the block header at the chain tip, as well as the MMR peaks corresponding to this 
-    // header for executing network transactions. 
+    // Returns the list of unconsumed network notes for a network account and the next page number to query.
```
Mirko-von-Leipzig:
```suggestion
    // Returns a paginated list of a network account's unconsumed notes.
    rpc GetUnconsumedNetworkNotesForNetworkAccount(requests.GetUnconsumedNetworkNotesForNetworkAccountRequest) returns (responses.GetUnconsumedNetworkNotesResponse) {}
```

proto/proto/requests.proto
```diff
@@ -215,3 +215,24 @@ message GetUnconsumedNetworkNotesRequest {
...
     uint64 page_size = 2;
 }
+
+// Returns a list of unconsumed network notes for a network account using pagination.
+message GetUnconsumedNetworkNotesForNetworkAccountRequest {
+    // An opaque token used to paginate through the notes.
+    //
+    // This should be null on the first call, and set to the response
+    // token until the response token is null, at which point all data
+    // has been fetched.
```
Mirko-von-Leipzig:
```suggestion
    // This should be null on the first call, and set to the response
    // token until the response token is null, at which point all data
    // has been fetched.
    //
    // Note that this token is only valid if used with the same parameters.
```

proto/proto/requests.proto
```diff
@@ -215,3 +215,24 @@ message GetUnconsumedNetworkNotesRequest {
...
+    // Number of notes to retrieve per page.
+    uint64 page_size = 2;
+
+    // The network account ID prefix to filter notes by.
+    uint32 network_account_id_prefix = 3;
+
+    // The maximum (inclusive) block number to scope the query by.
+    // Notes that are not consumed before or within the specified block number
+    // are treated as unconsumed.
+    fixed32 block_num = 4;
```
Mirko-von-Leipzig: Trying to think of a better name here; currently context is only known if reading  the comment.

But I don't really have anything.. `unconsumed_at`, `state_at`,..

crates/store/src/db/tests.rs
Mirko-von-Leipzig: I definitely don't like it; but it is on par with the rest of our store tests so I think its okay (for now).

What I would love is overhauling the store tests to be more genuine. Using `MockChain` to generate realistic blocks, and then applying them to the `store`. And then running tests on that.

crates/store/src/db/sql/mod.rs
```diff
@@ -1135,6 +1135,68 @@ pub fn unconsumed_network_notes(
...
+    // rowid column _must_ come after the note fields so that we don't mess up the
+    // `NoteRecord::from_row` call.
+    let mut stmt = transaction.prepare_cached(&format!(
+        "
+        SELECT {}, rowid
+        FROM notes
+        LEFT JOIN note_scripts ON notes.script_root = note_scripts.script_root
+        WHERE
+            execution_mode = 0 AND tag = ? AND
+            (consumed_block_num IS NULL OR consumed_block_num > ?) AND rowid >= ?
```
Mirko-von-Leipzig: You also need to exclude notes that were created _after_ the block number


```suggestion
        WHERE
            execution_mode = 0 AND tag = ?1 AND
            block_num <= ?2 AND
            (consumed_block_num IS NULL OR consumed_block_num > ?2) AND rowid >= ?3
```

Mirko-von-Leipzig: If possible, could you squeeze in a test case for this?

crates/store/src/server/ntx_builder.rs
Mirko-von-Leipzig: This made me realize that we could actually remove the original now already. Feeling a bit dumb for not noticing this sooner - apologies.

The original `get_unconsumed_network_notes` is used by the `ntx-builder` under the condition that block-production is halted, and you'll see a `Barrier` sync primitive used to co-ordinate this condition between block-production and ntx builder.

This means that the ntx could use the new method by either fetching the current chain tip and using that, or just setting the block num to maximum (I think that works currently).

Given that this PR is close to complete, you can decide if you want to do that as part of this PR or not.

sergerad: Yea that sounds good! Will do in this PR.

sergerad: I'm not sure how to make the current impl work while removing the existing fn. For example if you look at the impl of:

```rust
impl State {
    /// Load's all available network notes from the store, along with the required account states.
    #[instrument(target = COMPONENT, name = "ntx.state.load", skip_all)]
    pub async fn load(store: StoreClient) -> Result<Self, StoreError> {
```

The `load()` fn relies on notes returned by the store to determine network account prefixes:
```rust
            let notes = state
                .store
                .get_unconsumed_network_notes(network_account_id_prefix, block_num)
                .await?;
```

I don't think there is a way to invert that logic and start from network account prefixes to inform which notes to retrieve via the new fn.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -82,15 +83,22 @@ impl State {
...
-        for note in notes {
-            let prefix = note.account_prefix();
-
-            // Ignore notes which don't target an existing account.
-            let Some(account) = state.fetch_account(prefix).await? else {
-                continue;
-            };
-            account.add_note(note);
+        for prefix in self.accounts.iter() {
+            let network_account_id_prefix = NetworkAccountPrefix::try_from(0u32).unwrap();
```
sergerad: @Mirko-von-Leipzig its this part where I don't know that we could invert the note->account mapping and go account->notes

crates/store/src/server/ntx_builder.rs
Mirko-von-Leipzig: Ah yeah you're right. We would at minimum need a new method to inform us of all network accounts with unconsumed notes. Let's leave as is then and tackle that when we get there.

crates/store/src/server/ntx_builder.rs
```diff
@@ -152,4 +154,55 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
...
     }
+
+    #[instrument(
+        parent = None,
+        target = COMPONENT,
+        name = "store.ntx_builder_server.get_unconsumed_notes_for_network_account",
+        skip_all,
+        err
+    )]
+    async fn get_unconsumed_notes_for_network_account(
```
sergerad: @Mirko-von-Leipzig I'm trying to understand why the state has DB access and this functionality isn't required to be accessed via store fn like `crates/ntx-builder/src/store.rs`:
```rust
    /// Returns the list of unconsumed network notes.
    #[instrument(target = COMPONENT, name = "store.client.get_unconsumed_network_notes", skip_all, err)]
    pub async fn get_unconsumed_network_notes(&self) -> Result<Vec<NetworkNote>, StoreError> {
```

As in there is no corresponding `get_unconsumed_network_notes()` version of that ^ and it seems it is not required. In my head it would go `ntx-builder server -> store -> db` rather than `ntx-builder server -> state -> db`. Thanks in advance.

crates/store/src/db/sql/mod.rs
sergerad: Added `sql_unconsumed_notes_for_network_account`

crates/store/src/server/ntx_builder.rs
Mirko-von-Leipzig: This might just be a confusing naming problem - this is the store's gRPC service for the network transaction builder. I'm assuming that `server/ntx_builder.rs` is confusing and sounds like its part of the ntx builder component? Or maybe I'm misunderstanding.

So the full set of "spans" is:

```
- ntx builder's store client
- store server
- store's ntx gRPC service
- store's state (which includes the DB)
```

proto/proto/requests.proto
sergerad: Went with
```proto
     // The maximum (inclusive) block number to scope the query by.
     //
     // Notes that are consumed after this block number are excluded from the result.
     fixed32 latest_block_num = 4;
```

crates/store/src/server/ntx_builder.rs
sergerad: > I'm assuming that server/ntx_builder.rs is confusing and sounds like its part of the ntx builder component?

Yea its this 👍

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 
 - [BREAKING] Integrated proxy status endpoint into main proxy service, removing separate status port.
 - RPC requests with wildcard (`*/*`) media-type are not longer rejected ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
+- [BREAKING] Update `notes` table schema to have a nullable `consumed_block_num` ([#1100](https://github.com/0xMiden/miden-node/pull/1100)).
+- Add gRPC endpoint for retrieving unconsumed network notes for a specific network account and block limit ([#1100](https://github.com/0xMiden/miden-node/pull/1100)).
```
Mirko-von-Leipzig: This is internal only so no need

```suggestion
```

crates/store/src/db/migrations/001-init.sql
```diff
@@ -31,31 +31,31 @@ CREATE TABLE accounts (
...
+    serial_num               BLOB,
 
     PRIMARY KEY (block_num, batch_index, note_index),
     FOREIGN KEY (block_num) REFERENCES block_headers(block_num),
     FOREIGN KEY (sender) REFERENCES accounts(account_id),
     FOREIGN KEY (script_root) REFERENCES note_scripts(script_root),
     CONSTRAINT notes_type_in_enum CHECK (note_type BETWEEN 1 AND 3),
     CONSTRAINT notes_execution_mode_in_enum CHECK (execution_mode BETWEEN 0 AND 1),
-    CONSTRAINT notes_consumed_is_bool CHECK (execution_mode BETWEEN 0 AND 1),
+    CONSTRAINT notes_consumed_block_num_is_u32 CHECK (batch_index BETWEEN 0 AND 0xFFFFFFFF),
```
Mirko-von-Leipzig:
```suggestion
    CONSTRAINT notes_consumed_block_num_is_u32 CHECK (consumed_block_num BETWEEN 0 AND 0xFFFFFFFF),
```

crates/store/src/db/mod.rs
```diff
@@ -547,15 +547,34 @@ impl Db {
         Ok(())
     }
 
-    /// Loads the network notes that have not been consumed yet, using pagination to limit the
-    /// number of notes returned.
+    /// Loads the network notes that have not been consumed yet, using
+    /// pagination to limit the number of notes returned.
     pub(crate) async fn select_unconsumed_network_notes(
```
bobbinth: nit: comment wrapping.

crates/store/src/db/sql/mod.rs
```diff
@@ -1131,6 +1131,69 @@ pub fn unconsumed_network_notes(
...
+/// # Returns
+///
+/// A set of unconsumed network notes with maximum length of `size` and the page to get
+/// the next set.
+pub fn unconsumed_notes_for_network_account(
+    transaction: &Transaction,
+    network_account_id_prefix: NetworkAccountPrefix,
+    latest_block_num: BlockNumber,
+    mut page: Page,
+) -> Result<(Vec<NoteRecord>, Page)> {
```
bobbinth: Should this maybe be `select_unconsumed_network_notes_by_tag`? The thinking is:

- For queries that retrieve data from the DB we usually use `select_` prefix.
- We are explicitly filtering out non-network notes.
- While we are passing in `network_account_id_prefix`, we are actually using it as a tag.

The signature could be:
```rust
pub fn select_unconsumed_network_notes_by_tag(
    transaction: &Transaction,
    tag: [u32],
    latest_block_num: BlockNumber,
    mut page: Page,
) -> Result<(Vec<NoteRecord>, Page)> {
    ...
}
```
And then the caller would be responsible for converting `NetworkAccountPrefix` into a `u32`.

crates/store/src/db/sql/mod.rs
```diff
@@ -1131,6 +1131,69 @@ pub fn unconsumed_network_notes(
...
+        "Hardcoded execution value must match query"
+    );
+
+    // Select the rowid column so that we can return a pagination token.
+    //
+    // rowid column _must_ come after the note fields so that we don't mess up the
+    // `NoteRecord::from_row` call.
+    let mut stmt = transaction.prepare_cached(&format!(
+        "
+        SELECT {}, rowid
```
bobbinth: Technically not related to this PR, but would be we able to use `rowid` for pagination after migration to Disel?

crates/store/src/db/sql/mod.rs
```diff
@@ -1131,6 +1131,69 @@ pub fn unconsumed_network_notes(
     Ok((notes, page))
 }
 
+/// Returns a paginated batch of network notes for a specific account that have not yet been
+/// consumed.
+///
+/// The scope of the query is limited by the specified block number. Notes that are consumed after
+/// the specified block number are excluded from the result.
```
bobbinth: My understanding of the `WHERE` clause is:

- We return notes created before the specified block number AND
- Either not consumed OR consumed after the specified block number.

This seems to contradict the second sentence here. Or am I missing something?

crates/store/src/db/sql/mod.rs
```diff
@@ -1131,6 +1131,69 @@ pub fn unconsumed_network_notes(
...
+    // `NoteRecord::from_row` call.
+    let mut stmt = transaction.prepare_cached(&format!(
+        "
+        SELECT {}, rowid
+        FROM notes
+        LEFT JOIN note_scripts ON notes.script_root = note_scripts.script_root
+        WHERE
+            execution_mode = 0 AND tag = ?1 AND
+            block_num <= ?2 AND
+            (consumed_block_num IS NULL OR consumed_block_num > ?2) AND rowid >= ?3
```
bobbinth: This part of the `WHERE` clause confuses me a little. Why do we want to get notes that have already been consumed? (even if the consumption happened after `latest_block_num`).

proto/proto/store/ntx_builder.proto
```diff
@@ -60,6 +63,27 @@ message UnconsumedNetworkNotesRequest {
...
+     // Number of notes to retrieve per page.
+     uint64 page_size = 2;
+
+     // The network account ID prefix to filter notes by.
+     uint32 network_account_id_prefix = 3;
+
+     // The maximum (inclusive) block number to scope the query by.
+     //
+     // Notes that are consumed after this block number are excluded from the result.
+     fixed32 latest_block_num = 4;
```
bobbinth: Similar to one of my other comments: it seems like we are including notes consumed after the specified block number rather than excluding them.

proto/proto/store/ntx_builder.proto
```diff
@@ -60,6 +63,27 @@ message UnconsumedNetworkNotesRequest {
     uint64 page_size = 2;
 }
 
+// Returns a list of unconsumed notes for a network account using pagination.
+message UnconsumedNotesForNetworkAccountRequest {
```
bobbinth: nit: I'd maybe rename this to `UnconsumedNetworkNotesForAccountRequest`.

crates/store/src/state.rs
```diff
@@ -977,13 +977,31 @@ impl State {
         self.db.optimize().await
     }
 
-    /// Returns the unprocessed network notes, along with the next pagination token.
+    /// Returns the unprocessed network notes, along with the next pagination
+    /// token.
     pub async fn get_unconsumed_network_notes(
```
bobbinth: nit: comment wrapping.

bobbinth (commented):

crates/store/src/db/sql/mod.rs
Mirko-von-Leipzig: Yes though it requires manually adding `rowid` to the schema.

https://github.com/0xMiden/miden-node/pull/928#discussion_r2220941689

I will also note that using `rowid` is probably not a long-term solution as it has performance and size implications when used for composite key tables and its therefore encouraged to _disable_ rowids. Though its always a trade-off and for example, tables with large rows are recommended to _enable_ rowids so :shrug:.

More pertinent for pagination though is that they aren't guaranteed to be stable. 

#### References 

https://sqlite.org/rowidtable.html
https://sqlite.org/withoutrowid.html

crates/store/src/db/sql/mod.rs
Mirko-von-Leipzig: https://github.com/0xMiden/miden-node/issues/1089#issuecomment-3077750640

Though in that comment I didn't account for using `rowid` as the ordering parameter, so its not required to keep pagination stable.

We can simplify on this end and send _all_ unconsumed notes (no block limit either), and then on the ntx side we need to allow for duplication. Or we leave this as is and then we can error out on the ntx if duplication occurs. I usually favor the latter, but there might be a performance argument for the former.

bobbinth: I guess what confuses me here is the `consumed_block_num > ?2` criteria. This means that we'll be sending notes that have actually already been consumed just not before the specified block number. So, the NTX builder would need to do more work here (i.e., it would need to filter out the notes that have been consumed after the `latest_block_num`).

Mirko-von-Leipzig: The order of operations is something like

1. Ntx builder decides to load network account `A`. This could be because it received an account creation event for `A`, or some rehydration note event targeting `A`, or the actor for `A` crashed and it should be re-loaded etc.
2. Ntx builder creates actor for `A`, and seeds it with the local chain tip `C`.
3. Actor starts by using this gRPC method to sync all state `..=C`.
4. In the meantime ntx builder is sending all inflight mempool events `> C` to the actor.
5. Actor completes store sync. It is now effectively at block `C` ito state.
6. Actor starts processing the mempool events `> C`.
7. With current impl, no extra logic here is needed, notes consumed after `C` will be naturally nullified once the mempool nullifier event is processed.
8. Removing `consumed_block_num` means `A` will receive nullifiers for notes that it never loaded from the store.

Mirko-von-Leipzig: Something similar happens if we remove the created before critera `block_num <= ?`, because if the actor decides to build a tx before the latest mempool block event arrives, then we'll be building a note created _after_ our local state which should fail in the data store.

Mirko-von-Leipzig: This is why I said

> In general I prefer not time-jumping/duplicating because that means having to think hard about what could potential go wrong. Whereas this way one doesn't have to think at all. I'm just worried we miss something, or we refactor something out.

because without timejumping we can conclusively state that the actor `A` has the correct state at block `C` and is now receiving all mempool events `> C` and that's trivially correct ito state management because we need to perform exactly that in general.

crates/store/src/db/sql/mod.rs
Mirko-von-Leipzig: > While we are passing in network_account_id_prefix, we are actually using it as a tag.

Then we require the caller to know the encoding and table schema, and that the note tag column holds the network account prefix.

crates/store/src/db/models/queries.rs
```diff
@@ -1072,7 +1071,152 @@ pub(crate) fn unconsumed_network_notes(
         ),
     )
     .filter(schema::notes::execution_mode.eq(0_i32))
-    .filter(schema::notes::consumed.eq(consumed_to_raw_sql(false)))
+    .filter(schema::notes::consumed_block_num.eq(None::<i64>))
```
sergerad: @drahnr I'm failing to retrieve any notes after changing `consumed` to `consumed_block_num` with the diesel setup. It might be to do with `None`, I know you had a warning about `None<T>` somewhere.

Test `sql_unconsumed_network_notes` fails on this branch. LMK if you have any ideas around that schema change.

crates/store/src/db/sql/mod.rs
bobbinth: > 8\. Removing `consumed_block_num` means `A` will receive nullifiers for notes that it never loaded from the store.

Yes, makes sense. I've assumed that handling such situations would be pretty straight-forward (and it probably isn't super difficult) but what you are describing seems a bit cleaner.

I do think that if we keep it this way, we should probably update the docs to make it clear that we are getting all notes which have been created before `latest_block_num` and not been consumed by `latest_block_num `.

crates/store/src/db/models/queries.rs
sergerad: fix was
```rust
.filter(schema::notes::consumed_block_num.is_null())
```

crates/store/src/db/sql/mod.rs
sergerad: Has this clarified the query? https://github.com/0xMiden/miden-node/pull/1100#discussion_r2242036846

I have updated the comments around it. LMK if anything is still unclear or potentially wrong.

proto/proto/store/ntx_builder.proto
```diff
@@ -60,6 +63,28 @@ message UnconsumedNetworkNotesRequest {
     uint64 page_size = 2;
 }
 
+// Returns a list of unconsumed network notes for an account up to a specified block number using
+// pagination.
+message UnconsumedNetworkNotesForAccountRequest {
+    // This should be null on the first call, and set to the response
+    // token until the response token is null, at which point all data
+    // has been fetched.
```
bobbinth: nit: comment wrapping.

Not from this PR, but this also applies to lines 57 - 59 above.

crates/store/src/db/sql/mod.rs
bobbinth, line 1: This shows up as a new file, which seems odd. Are we still using it?

proto/proto/store/ntx_builder.proto
```diff
@@ -60,6 +63,28 @@ message UnconsumedNetworkNotesRequest {
...
+     // Number of notes to retrieve per page.
+     uint64 page_size = 2;
+
+     // The network account ID prefix to filter notes by.
+     uint32 network_account_id_prefix = 3;
+
+     // The maximum (inclusive) block number to scope the query by.
+     //
+     // Notes that are consumed after this block are excluded from the result.
+     fixed32 latest_block_num = 4;
```
bobbinth: I still don't think the comment here is correct. We are not excluding the notes consumed after this block number in the result right? Instead, we are including them - or am I missing something. Shouldn't the comment read:

The result includes notes which: (1) created before this block and (2) haven't been consumed by this block.

crates/store/src/db/models/queries.rs
```diff
@@ -1072,7 +1071,155 @@ pub(crate) fn unconsumed_network_notes(
...
+        notes.push(TryInto::<NoteRecord>::try_into(raw_item)?);
+    }
+
+    Ok((notes, page))
+}
+
+/// Returns a paginated batch of network notes for an account that have not been consumed
+/// before, or at, a specified block number.
+///
+/// Notes that have been consumed after the specified block number are excluded from the result.
```
bobbinth: Similar to one of my previous comments: I am not sure this is accurate. If `latest_block_num == 1000` and note `A` was consumed at block 1001, the note will be included in the result - right?

crates/store/src/db/models/queries.rs
```diff
@@ -1072,7 +1071,155 @@ pub(crate) fn unconsumed_network_notes(
...
+#[allow(
+    clippy::too_many_lines,
+    reason = "Lines will be reduced when schema is updated to simplify logic"
+)]
+pub(crate) fn unconsumed_network_notes_for_account(
+    conn: &mut SqliteConnection,
+    network_account_id_prefix: NetworkAccountPrefix,
+    latest_block_num: BlockNumber,
+    mut page: Page,
+) -> Result<(Vec<NoteRecord>, Page), DatabaseError> {
```
bobbinth: I think my comment from https://github.com/0xMiden/miden-node/pull/1100#discussion_r2241701939 still applies there - but I'll reply there.

crates/store/src/db/models/queries.rs
```diff
@@ -1072,7 +1071,155 @@ pub(crate) fn unconsumed_network_notes(
...
+                execution_hint: tuple.8,
+                merkle_path: tuple.9,
+                assets: tuple.10,
+                inputs: tuple.11,
+                script: tuple.12,
+                serial_num: tuple.13,
+            },
+            tuple.14,
+        )
+    }
```
bobbinth: Probably not for this PR, but this seems like a quite a bit of boilerplate code as compared to what we had before and would probably result in quite a bit of maintenance burden. Is there a way to simplify this?

crates/store/src/db/models/queries.rs
```diff
@@ -1072,7 +1071,155 @@ pub(crate) fn unconsumed_network_notes(
...
+            schema::notes::tag,
+            schema::notes::aux,
+            schema::notes::execution_hint,
+            schema::notes::inclusion_path,
+            // details
+            schema::notes::assets,
+            schema::notes::inputs,
+            schema::note_scripts::script.nullable(),
+            schema::notes::serial_num,
+            rowid_sel.clone(),
```
bobbinth: This code seems to be repeating a few times in this file - is there a way to encapsulate it somehow?

crates/store/src/db/models/queries.rs
bobbinth, line 1: Not for this PR, but we should really try to split this file up into multiple smaller files.

bobbinth (commented):
Looks good! Thank you. I left a few more comments inline.

Not for this PR, but as I mentioned in one of the comments, the amount of boilerplate for defining a query seems to be quite high. Would be great to find a way to reduce it.

crates/store/src/db/sql/mod.rs
bobbinth: > Then we require the caller to know the encoding and table schema, and that the note tag column holds the network account prefix.

I think this is fine. The caller is in the `db.rs` file which I think is where the "business logic" is. The query file, in my opinion, is meant to be close to the database itself. In this query we are filtering by tag - so, the caller needs to know that we need to translate the account ID prefix into a tag.

crates/store/src/db/models/queries.rs
drahnr: We can define a type and should be able to use `(NotesRaw::as_select(), schema::note_scripts::script.nullable(), rowid_sel.clone())` 🤞 - `left_join` is _very_ particular what it accpets

crates/store/src/db/models/queries.rs
drahnr: The root cause for the tuple stuff is the issue around the implicit row id which we use for pagination. The path forward is to change the `schema.rs` to explicitly contain the row id, or use a different way of pagination.

drahnr: PR addressing the issue (and a few more): https://github.com/0xMiden/miden-node/pull/1126

crates/store/src/db/models/queries.rs
drahnr: Also covered in https://github.com/0xMiden/miden-node/pull/1126

crates/store/src/db/models/queries.rs
drahnr: I'll wait until this and https://github.com/0xMiden/miden-node/pull/1126 are landed, and will do so Monday~ish

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -21,31 +21,31 @@ CREATE TABLE accounts (
...
+    note_index               INTEGER NOT NULL, -- Index of note in batch, starting from 0
+    note_id                  BLOB    NOT NULL,
+    note_type                INTEGER NOT NULL, -- 1-Public (0b01), 2-Private (0b10), 3-Encrypted (0b11)
+    sender                   BLOB    NOT NULL,
+    tag                      INTEGER NOT NULL,
+    execution_mode           INTEGER NOT NULL, -- 0-Network, 1-Local
+    aux                      INTEGER NOT NULL,
+    execution_hint           INTEGER NOT NULL,
+    inclusion_path           BLOB NOT NULL,    -- Serialized sparse Merkle path of the note in the block's note tree
+    consumed_block_num       INTEGER,
```
drahnr: nit: `consumed_in(_block_num)?` might be easier to understand

crates/store/src/db/sql/mod.rs
drahnr: Shouldn't exist anymore, this is obsolete. I think this is a leftover from a rebase / older revision / merge

crates/store/src/db/models/queries.rs
```diff
@@ -1072,7 +1071,155 @@ pub(crate) fn unconsumed_network_notes(
...
+        ),
+    )
+    .filter(schema::notes::execution_mode.eq(0_i32))
+    .filter(schema::notes::tag.eq(tag_to_raw_sql(u32::from(network_account_id_prefix).into())))
+    .filter(schema::notes::block_num.le(block_number_to_raw_sql(latest_block_num)))
+    .filter(
+        schema::notes::consumed_block_num
+            .is_null()
+            .or(schema::notes::consumed_block_num.gt(block_number_to_raw_sql(latest_block_num))),
```
drahnr: We are including all blocks that fullfil the following conditions:
`x.ececuttion_mode == Local`
`x.block_num  <= latest_block_num`
`x.tag == tag` 
`consumed_block_num == NIL || x.consumed_block_num > latest_block_num` <-- could you elaborate? To my understanding this means not consumed yet or will be consumed in a block in the future (relative to `latest_block_num`)?

crates/store/src/db/tests.rs
```diff
@@ -472,6 +495,94 @@ fn sql_unconsumed_network_notes() {
...
+                &mut conn,
+                account_id.try_into().unwrap(),
+                b.into(),
+                Page {
+                    token: None,
+                    size: NonZeroUsize::new(note_count_per_block * (block_count as usize)).unwrap(),
+                },
+            )
+            .unwrap();
+            assert_eq!(result.len(), expected_count);
```
drahnr: nit/discard: I oftentimes find sorted arrays in conjunction with `pretty_assert` rather useful when something goes south, to get immediate output what differs between left and right

proto/proto/store/ntx_builder.proto
sergerad: Yea sorry you are right. I must have been thinking consumption is ignored after that block. I do find this query and endpoint quite confusing still. Will update soon.

crates/store/src/db/migrations/2025062000000_setup/up.sql
sergerad: Happy to change. Interested in what others think about it.

crates/store/src/db/models/queries.rs
sergerad: > `consumed_block_num == NIL || x.consumed_block_num > latest_block_num` <-- could you elaborate?

Not consumed or consumed after the threshold specified by the query. You can think of that block number specifying the snapshot of state that scopes the query.

proto/proto/store/ntx_builder.proto
sergerad: LMK how you think that reads now. Probably still room for improvement.

crates/store/src/db/models/queries.rs
sergerad: Updated

crates/store/src/db/migrations/2025062000000_setup/up.sql
Mirko-von-Leipzig: I would have said `consumed_at` and changed `block_num` to `created_at`, but I imagine the latter might be unpopular :)

crates/store/src/db/models/queries.rs
sergerad: > You can think of that block number specifying the snapshot of state that scopes the query.

Re-reading this and it felt misleading. The "scope" only affects whether the fact that the note is consumed is taken into account in the query.

Mirko-von-Leipzig: Is the confusion not just the naming of `latest_block_num`? It implies that its the latest block in the chain, but its really the block the caller wants the state from. As in:

> Give me all network notes for this network account which were unconsumed as of this block.

proto/proto/store/ntx_builder.proto
```diff
@@ -51,15 +54,35 @@ message MaybeAccountDetails {
...
+     // Number of notes to retrieve per page.
+     uint64 page_size = 2;
+
+     // The network account ID prefix to filter notes by.
+     uint32 network_account_id_prefix = 3;
+
+     // The block number (inclusive) that the notes must be unconsumed by.
+     //
+     // Notes that are consumed after this block are included in the result.
+     fixed32 unconsumed_by_block_num = 4;
```
bobbinth: I'm not sure the `unconsumed_by_block_num` name is appropriate here. If I'm understanding correctly, this block number is used for 2 things:

1. Filter out any note created before this block.
2. Filter out any note consumed after this block.

If we keep both criteria, then we should probably change the name to something else (e.g., just `block_num` or `latest_block_num` as used in other places).

For the comment, we could say something like this:
```proto
 // The block number to filter the returned notes by.
 //
 // Notes that are created or consumed after this block are excluded from the result.
```

proto/proto/store/ntx_builder.proto
```diff
@@ -51,15 +54,35 @@ message MaybeAccountDetails {
...
+    // is null, at which point all data has been fetched.
     optional uint64 page_token = 1;
 
     // Number of notes to retrieve per page.
     uint64 page_size = 2;
 }
 
+// Returns a paginated list of network notes for an account which are unconsumed by a specified
+// block number.
+message UnconsumedNetworkNotesForAccountRequest {
```
bobbinth: Related to the previous comment, I'd maybe say something like:
```
// Returns a paginated list of unconsumed network notes for an account.
//
// Note that notes created or consumed after the specified block are excluded from the result.
```

crates/store/src/db/mod.rs
```diff
@@ -487,4 +487,23 @@ impl Db {
...
+    ) -> Result<(Vec<NoteRecord>, Page)> {
+        self.transact("unconsumed network notes for account", move |conn| {
+            models::queries::select_unconsumed_network_notes_by_tag(
+                conn,
+                network_account_id_prefix.into(),
+                block_num,
+                page,
+            )
+        })
+        .await
```
bobbinth, line 510: nit: I'd maybe add a comment here explaining why we are converting `network_account_id_prefix` into `tag`.

crates/store/src/db/models/queries.rs
```diff
@@ -966,7 +966,113 @@ pub(crate) fn unconsumed_network_notes(
...
+    // LIMIT ?
+    #[allow(
+        clippy::items_after_statements,
+        reason = "It's only relevant for a single call function"
+    )]
+    type RawLoadedTuple = (
+        NoteRecordRaw,
+        Option<Vec<u8>>, // script
+        i64,             // rowid (from sql::<BigInt>("notes.rowid"))
+    );
```
bobbinth, line 1043: Question (and probably not for this PR): would having an explicit integer primary key (which auto-increments) in the table make it possible to avoid this type (and associated functionality)?

crates/store/src/db/models/queries.rs
sergerad: Latest iteration is to leave it as `block_num`

crates/store/src/db/mod.rs
sergerad: I'm not sure what helpful context there would be to explain here. I don't know why the schema uses tag vs prefix for example. What were you thinking?

crates/store/src/db/models/queries.rs
sergerad: @drahnr WDYT?

crates/store/src/db/mod.rs
bobbinth: Something like: "network notes sent to a specific account have their tags set to the prefix of the target account ID - so, we can convert the ID prefix into a note tag to query the notes for a given account"

crates/store/src/db/migrations/2025062000000_setup/up.sql
bobbinth: I actually don't mind `consumed_at` and `created_at` - though, maybe I'd change `created_at` to `committed_at` since this would be the block in which the note is committed into the chain (but not a strong preference).

sergerad: Will merge committed_at and consumed_at in this PR

PR #1101, opened 2025-07-21
chore: introduce `.gitattritbutes` for `CHANGELOG.md`

drahnr:
We do frequent conflict resolutions on merges and rebases. The default resolution is to unionize the lined entries. This makes `merge=union` the default for `CHANGELOG.md`. To be copied over to other repos.


Mirko-von-Leipzig (approved):

drahnr:
Post-"mortem": This works when doing local merges, which is good, however, github does not honour the user provided `.gitattributes` and hence doesn't quite work when there are conflicts between branch and target branch.

https://github.com/orgs/community/discussions/9288

PR #1107, opened 2025-07-23
chore: update miden-base to latest rev `d6dc69b8`

mmagician:
Migrations:
- `TransactionExecutor` uses generics instead of `dyn Trait`
    - miden-base reference: https://github.com/0xMiden/miden-base/pull/1580
    - commit: 8fe85f2c82219b16d65df499caf262e2004ccc9b

### DIFF
--- a/crates/ntx-builder/src/transaction.rs
+++ b/crates/ntx-builder/src/transaction.rs
@@ -16,7 +16,7 @@ use miden_remote_prover_client::remote_prover::tx_prover::RemoteTransactionProve
 use miden_tx::{
     DataStore, DataStoreError, LocalTransactionProver, MastForestStore, NoteAccountExecution,
     NoteConsumptionChecker, TransactionExecutor, TransactionExecutorError, TransactionMastStore,
-    TransactionProverError,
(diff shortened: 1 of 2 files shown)


crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig, line 1: Does this make it possible to remove the work-around in `execute_transaction`?

Basically removing the block-on runtime.

Mirko-von-Leipzig (approved):

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig, line 1: And maybe just out of curiosity, but is this just a bump to keep updated, or is this required for something you're working on?

mmagician: Just regular maintenance, also I'm updating the client so having the node migrated helps

mmagician: I haven't looked into this too deeply but I think we still need `Send` on the executor

PR #1108, opened 2025-07-24
feat: Track chain tip in ntx-builder state

### LINKED ISSUE #1070: Track chain tip in ntb
We [reverted](https://github.com/0xMiden/miden-node/pull/1057#discussion_r2194514114) this because `PartialBlockchain` currently cannot be pruned of old block headers, making this too expensive to do right now.

We should re-add this once pruning is added (https://github.com/0xMiden/miden-base/issues/1561). https://github.com/0xMiden/miden-node/pull/1057/commits/ac7485b5891ebe2592099f188a3b9d43221a1a1e should give an indication on how to do this.

sergerad:
## Context

Closes #1070.

Most of the changes are taken from https://github.com/0xMiden/miden-node/commit/ac7485b5891ebe2592099f188a3b9d43221a1a1e#diff-12c51f34bc0d99acc62d688e262b00e4c6d59fde915a23ae8a3f4ed780db04daL331.

That commit was reverted because of concerns of cloning the entirety of `PartialBlockchain` every time a `TransactionCandidate` was instantiated in `State::select_candidate()`.

## Changes
- Reinstate the changes from `ac7485b` which updates Ntx Builder state to track chain tip.
- Add chain mmr pruning logic in Ntx Builder state.

### DIFF
--- a/crates/ntx-builder/src/state/mod.rs
+++ b/crates/ntx-builder/src/state/mod.rs
@@ -29,8 +29,6 @@ mod account;
 
 /// The maximum number of blocks to keep in memory while tracking the chain tip.
 const MAX_BLOCK_COUNT: usize = 4;
-/// The number of blocks kept after pruning.
-const PRUNED_BLOCK_COUNT: usize = 2;
 
 /// A candidate network transaction.
 ///
@@ -208,12 +206,10 @@ impl State {
         // Set the new tip.
         self.chain_tip_header = tip;
 
-        // Prune MMR if necessary.
-        if self.chain_mmr.num_tracked_blocks() > MAX_BLOCK_COUNT {
-            let pruned_block_height =
-                (self.chain_mmr.chain_length().as_usize() - PRUNED_BLOCK_COUNT) as u32;
-            self.chain_mmr.prune_to(..pruned_block_height.into());
-        }
+        // Keep MMR pruned.
+        let pruned_block_height =
+            (self.chain_mmr.chain_length().as_usize() - MAX_BLOCK_COUNT) as u32;
+        self.chain_mmr.prune_to(..pruned_block_height.into());
     }
 
     /// Marks a previously selected candidate account as failed, allowing it to be available for


crates/ntx-builder/src/state/mod.rs
```diff
@@ -70,10 +87,24 @@ pub struct State {
 }
 
 impl State {
+    /// The number of blocks to keep in memory while tracking the chain tip.
+    const CHAIN_TIP_LENGTH: u32 = 128;
```
sergerad: @Mirko-von-Leipzig unsure what length we want and whether we want it configurable.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -70,10 +87,24 @@ pub struct State {
...
     #[instrument(target = COMPONENT, name = "ntx.state.load", skip_all)]
     pub async fn load(store: StoreClient) -> Result<Self, StoreError> {
+        let (chain_tip, chain_mmr) = store
+            .get_latest_blockchain_data_with_retry()
+            .await?
+            .expect("store should contain a latest block");
+
+        let mut chain_mmr = PartialBlockchain::new(chain_mmr, [])
+            .expect("PartialBlockchain should build from latest partial MMR");
+        chain_mmr.prune_to(..Self::CHAIN_TIP_LENGTH.into());
```
sergerad: It would be nice if the constructor handled a range to avoid large construction in the first place? @igamigo would that be possible?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -178,7 +220,18 @@ impl State {
...
+                anyhow::ensure!(
+                    header.prev_block_commitment() == self.chain_tip.commitment(),
+                    "New block's parent commitment {} does not match local chain tip {}",
+                    header.prev_block_commitment(),
+                    self.chain_tip.commitment()
+                );
+
+                // Chain MMR always lags by one block.
+                self.chain_mmr.add_block(self.chain_tip.clone(), true);
+                self.chain_mmr.prune_to(..Self::CHAIN_TIP_LENGTH.into());
```
sergerad: @Mirko-von-Leipzig lmk if you would like to avoid pruning after every block add. For example we could have an initial length and then prune back to initial length once we reach 2x initial.

crates/ntx-builder/src/state/mod.rs
bobbinth: I think this depends on whether we can avoid cloning the struct for every transaction. If we can avoid cloning, we can make it fairly long (e.g., 128 blocks or maybe even more), if not, I'd stick on the smaller side - e.g., 16 blocks.

crates/ntx-builder/src/state/mod.rs
bobbinth: I think pruning on every block is probably an overkill. We could also prune on every 8th block or something like that.

Separately, is `prune_to()` used correctly here? I though this would prune just the initial `CHAIN_TIP_LENGTH` from the MMR (while we want to prune to `chain_tip - CHAIN_TIP_LENGTH`, I think).

crates/ntx-builder/src/state/mod.rs
sergerad: We do currently clone here
```rust
    pub fn select_candidate(&mut self, limit: NonZeroUsize) -> Option<TransactionCandidate> {
       ...
        return TransactionCandidate {
            account: account.latest_account(),
            notes,
            chain_tip: self.chain_tip.clone(),
            chain_mmr: self.chain_mmr.clone(),
        }
        .into();
```

Have left that as is and kept the block count low.

crates/ntx-builder/src/state/mod.rs
sergerad: Yea I wasn't sure whether `self.chain_mmr.chain_length()` returned block count or length of the chain. My understanding now is that it is the latter.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -30,14 +31,33 @@ mod account;
...
 /// It tracks inflight transactions, and their impact on network-related state.
 pub struct State {
+    /// The latest committed block header.
+    chain_tip: BlockHeader,
+
+    /// The chain MMR including the latest block header.
+    chain_mmr: PartialBlockchain,
+
+    /// The number of blocks in the chain MMR after pruning.
+    chain_mmr_block_count: u32,
```
sergerad: I believe we need to track this manually because there is no way to get block count from the mmr, only total chain length (latest block number) AFAICT.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -146,13 +184,44 @@ impl State {
...
+        self.chain_tip = tip;
+
+        // Prune MMR if necessary.
+        if self.chain_mmr_block_count > Self::MAX_BLOCK_COUNT {
+            let pruned_block_height =
+                self.chain_mmr.chain_length().as_u32() - Self::PRUNED_BLOCK_COUNT;
+            self.chain_mmr.prune_to(..pruned_block_height.into());
+            self.chain_mmr_block_count = Self::PRUNED_BLOCK_COUNT;
+        }
+    }
```
sergerad, line 213: Unfortunately don't have a way of unit testing the logic in this PR. Depending on the constructors available for `PartialBlockchain` and `PartialMmr`, could look at factoring this logic out to be testable as a separate struct.

crates/ntx-builder/src/store.rs
```diff
@@ -51,11 +51,16 @@ impl StoreClient {
...
 
-    #[instrument(target = COMPONENT, name = "store.client.genesis_header", skip_all, err)]
-    pub async fn genesis_header_with_retry(&self) -> Result<BlockHeader, StoreError> {
+    /// Returns the block header and MMR peaks at the chain tip if the input `block_num` is either
+    /// `None` or an outdated block number. If the input `block_num` is equal to the current
+    /// chain tip, this function returns `None`.
+    #[instrument(target = COMPONENT, name = "store.client.get_latest_blockchain_data_with_retry", skip_all, err)]
+    pub async fn get_latest_blockchain_data_with_retry(
+        &self,
+    ) -> Result<Option<(BlockHeader, PartialMmr)>, StoreError> {
```
bobbinth: I am not sure the comment here is correct since we are not passing a `block_num` into this function.

But also, my understanding is that this will return the latest chain MMR + the latest block header, correct? If so, I wonder if we should also include additional block headers (i.e., the last 8 blocks) into the response. This could be specified by adding something like `num_latest_blocks` parameter to the endpoint.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -30,14 +31,33 @@ mod account;
...
     pub notes: Vec<NetworkNote>,
+
+    /// The latest locally committed block header.
+    ///
+    /// This should be used as the reference block during transaction execution. It is guaranteed
+    /// to be tracked by `chain_mmr`.
+    pub chain_tip: BlockHeader,
+
+    /// The chain MMR, which includes the `chain_tip`'s block.
+    pub chain_mmr: PartialBlockchain,
```
bobbinth: Is the comment on line 44 correct? That is, does chain MMR actually include the chain tip? When we create transaction inputs, one of the requirements is that `chain_tip.chain_commitment() = chain_mmr.peaks().hash_peaks()` - but this would be impossible if chain tip was contained in `chain_mmr`.

If the comment is correct, then I think transaction construction would fail (but would be good to double-check this).

crates/ntx-builder/src/state/mod.rs
```diff
@@ -70,11 +90,29 @@ pub struct State {
...
     pub async fn load(store: StoreClient) -> Result<Self, StoreError> {
+        let (chain_tip, chain_mmr) = store
+            .get_latest_blockchain_data_with_retry()
+            .await?
+            .expect("store should contain a latest block");
+
+        let mut chain_mmr = PartialBlockchain::new(chain_mmr, [])
+            .expect("PartialBlockchain should build from latest partial MMR");
+        let pruned_block_height = chain_mmr.chain_length().as_u32() - Self::PRUNED_BLOCK_COUNT;
+        chain_mmr.prune_to(..pruned_block_height.into());
```
bobbinth: Why do we need to prune the chain MMR here? My understanding that it currently arrives from the store with just one leaf. But maybe I'm missing something?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -70,11 +90,29 @@ pub struct State {
 }
 
 impl State {
+    /// The maximum number of blocks to keep in memory while tracking the chain tip.
+    const MAX_BLOCK_COUNT: u32 = 16;
+    /// The number of blocks kept after pruning.
+    const PRUNED_BLOCK_COUNT: u32 = 8;
```
bobbinth: nit: I would put these constants into the `CONSTANTS` section at the top of the file.

crates/ntx-builder/src/state/mod.rs
bobbinth: We could technically do this via something like `chain_mmr.block_headers().count()` but it may be better to add something like `num_tracked_blocks()` method to it. It would look something like this:
```rust
impl PartialBlockchain {
   pub fn num_tracked_blocks(&self) -> usize {
        self.blocks.len()
    }
}
```

crates/ntx-builder/src/state/mod.rs
```diff
@@ -146,13 +184,44 @@ impl State {
...
         }
         self.inject_telemetry();
 
         None
     }
 
+    /// The latest block number the state knows of.
+    pub fn chain_tip(&self) -> BlockNumber {
+        self.chain_tip.block_num()
+    }
```
bobbinth: It's a bit confusing that the `chain_tip` field and `chain_tip()` accessors have different types. Maybe we could name the field something like `latest_block_header` or `chain_tip_header`?

crates/ntx-builder/src/state/mod.rs
sergerad: > My understanding that it currently arrives from the store with just one leaf

I wasn't aware of this. Can remove the prune then.

crates/ntx-builder/src/state/mod.rs
sergerad: https://github.com/0xMiden/miden-base/pull/1643

crates/ntx-builder/src/store.rs
sergerad: Do mean that we could update the gRPC endpoint so that this request specifies `num_latest_blocks`:
```proto
// Returns peaks for the given block number, or the chain tip if a block number not provided.
message GetCurrentBlockchainDataRequest {
    // Last client-known block number. If not specified, the latest header and peaks will be retrieved.
    optional fixed32 block_num = 1;
}
```

And this response contains additional block headers:
```proto
// Current blockchain data based on the requested block number.
message GetCurrentBlockchainDataResponse {
    // Commitments that represent the current state according to the MMR.
    repeated digest.Digest current_peaks = 1;
    // Current block header.
    optional block.BlockHeader current_block_header = 2;
}
```

crates/ntx-builder/src/state/mod.rs
sergerad: I can't quite tell whether mmr contains the tip based on this fn:
```rust
    #[instrument(target = COMPONENT, name = "store.client.get_latest_blockchain_data", skip_all, err)]
    async fn get_latest_blockchain_data(
        &self,
    ) -> Result<Option<(BlockHeader, PartialMmr)>, StoreError> {
```

I think the comment contradicts another one and the corresponding logic where it indicates the mmr is 1 block behind:
```rust
    /// Updates the chain tip and MMR block count.
    ///
    /// Blocks in the MMR are pruned if the block count exceeds the maximum.
    fn update_chain_tip(&mut self, tip: BlockHeader) {
        // Update MMR which lags by one block.
        self.chain_mmr.add_block(self.chain_tip.clone(), true);
```

@Mirko-von-Leipzig might recall based on https://github.com/0xMiden/miden-node/commit/ac7485b5891ebe2592099f188a3b9d43221a1a1e#diff-12c51f34bc0d99acc62d688e262b00e4c6d59fde915a23ae8a3f4ed780db04daL331

crates/ntx-builder/src/store.rs
bobbinth: Yes, this is roughly what I was thinking - though the response would probably look like so:
```proto
message GetCurrentBlockchainDataResponse {
    repeated digest.Digest current_peaks = 1;
    repeated block.BlockHeader block_headers = 2;
    repeated Digest nodes = 3;
}
```
I believe we'll need the `nodes` so that we can add the authentication nodes to the `PartialMmr` which would allow us to verify that a given block header is actually in the MMR.

crates/ntx-builder/src/state/mod.rs
bobbinth: Given the rest of the code, my guess is that the MMR does not contain the chain tip (which is probably the right approach).

crates/ntx-builder/src/state/mod.rs
```diff
@@ -146,13 +180,42 @@ impl State {
...
         }
         self.inject_telemetry();
 
         None
     }
 
+    /// The latest block number the state knows of.
+    pub fn chain_tip_header(&self) -> BlockNumber {
+        self.chain_tip.block_num()
+    }
```
bobbinth: I think the naming is still a bit off here: the name is "chain_tip_header" but we are returning a block number.

I would probably do the revers - name the field `chain_tip_header` and the method name would then be just `chain_tip()`.

bobbinth (commented):
Looks good! Thank you! I left one more comment above. I think the main thing is to think through how we get the initial data from the store. I think there are two options:

1. Get the data needed to reconstruct an empty `PartialBlockchain` and the chain tip.
2. Get enough data to reconstruct the `PartialBlockchain` with $n$ latest block headers in it (currently $n = 8$, I think).

The first approach is how it works now, and it is probably the simplest way to go about it. The only thing I'm not sure is how it could affect the rest as we can't guarantee that the partial MMR has $n$ block headers in it - but also, I'm not sure if it matters.

So, maybe we just keep the existing approach and ignore my comment in https://github.com/0xMiden/miden-node/pull/1108#discussion_r2229824577.

bobbinth:
Oh - and we should re-based this from the latest `next`.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -23,21 +24,44 @@ use crate::{
 
 mod account;
 
+// CONSTANTS
+// =================================================================================================
+
+/// The maximum number of blocks to keep in memory while tracking the chain tip.
+const MAX_BLOCK_COUNT: usize = 16;
+/// The number of blocks kept after pruning.
+const PRUNED_BLOCK_COUNT: usize = 8;
```
sergerad: @Mirko-von-Leipzig do you think we want to make this configurable?

sergerad:
> Also, what's the best way to test if this approach is working (i.e., we can still build network transactions)? Maybe we already have this covered by tests?

@Mirko-von-Leipzig WDYT? Do the existing unit or integration tests cover the updated functionality?

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: I don't really see a reason to do so at the moment so I'd just keep them as constants only.

Mirko-von-Leipzig: Why do we need more than just the latest one? Is it for FPI reasons somehow?

Mirko-von-Leipzig:
> > Also, what's the best way to test if this approach is working (i.e., we can still build network transactions)? Maybe we already have this covered by tests?
> 
> @Mirko-von-Leipzig WDYT? Do the existing unit or integration tests cover the updated functionality?

The only real tests we have at the moment are the client's integration tests which do include some network notes. I'd really love for us to make the node easier to control and create integration tests with.

crates/ntx-builder/src/state/mod.rs
sergerad: Unsure about use cases for the block range. We can always set `PRUNED_BLOCK_COUNT` to 1 if we want. Main thing is not to prune every time we add a block to the `PartialBlockchain`.

bobbinth: The other reason I had in mind is if/when we transition to passing the partial blockchain by reference, we would want to avoid pruning block headers against which we may have in-progress transactions, and so, keeping a small buffer helps here.

Mirko-von-Leipzig: I'm unsure about the internals of the `PartialBlockChain` here, but in order to do something like that we would need the DAG to use `Arc<Node>` which would automatically keep copies of it alive until they die.

I see any other way in which references can be kept while mutating the core structure but I'm not very familiar with how it works.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -146,13 +180,42 @@ impl State {
...
 
             self.in_progress.insert(candidate);
-            return TransactionCandidate { account: account.latest_account(), notes }.into();
+            return TransactionCandidate {
+                account: account.latest_account(),
+                notes,
+                chain_tip_header: self.chain_tip_header.clone(),
+                chain_mmr: self.chain_mmr.clone(),
+            }
+            .into();
```
bobbinth, line 187: We end up cloning `chain_mmr` twice for each transaction: once here, and then once returning it from the `NtxDataStore`. Would be great to figure out how we can avoid one of the clones. Let's create an issue for this (this would be a fairly low-priority issue at this point).

bobbinth (approved):

Would be great to run client integration tests against this branch to make sure network transactions still work. But other than that, I think it is good to merge.

crates/ntx-builder/src/state/mod.rs
bobbinth: Maybe for now we should reduce the numbers to something like 4 and 2 to reduce the amount of cloned data?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - A transaction's account delta is now checked against its commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
 - Added check for Account Id prefix uniqueness when transactions to create accounts are submitted to the mempool ([#1094](https://github.com/0xMiden/miden-node/pull/1094)).
 - RPC accept header now supports specifying the genesis commitment in addition to the RPC version. This lets clients ensure they are on the right network ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
+- Update network transaction builder state to track chain tip ([#1108](https://github.com/0xMiden/miden-node/pull/1108)).
```
Mirko-von-Leipzig: Think we don't need to mention this as it doesn't change anything for the user

```suggestion
```

crates/ntx-builder/src/state/mod.rs
```diff
@@ -146,13 +180,42 @@ impl State {
...
+
+        // Set the new tip.
+        self.chain_tip_header = tip;
+
+        // Prune MMR if necessary.
+        if self.chain_mmr.num_tracked_blocks() > MAX_BLOCK_COUNT {
+            let pruned_block_height =
+                (self.chain_mmr.chain_length().as_usize() - PRUNED_BLOCK_COUNT) as u32;
+            self.chain_mmr.prune_to(..pruned_block_height.into());
+        }
```
Mirko-von-Leipzig: Is pruning multiple blocks cheaper than just pruning them as you go? As in, why not just `push` and `pop` to keep the amount retained constant?

Mirko-von-Leipzig (approved):
Just the one question but should be good to go whichever way :)

Maybe just lower the constants as suggested by @bobbinth

crates/ntx-builder/src/state/mod.rs
sergerad: I can't see a way to push/pop via `PartialBlockchain` or `PartialMmr` atm. We could add it and update this. I'll proceed with current impl and we can follow up if need.

sergerad:
Waiting on https://github.com/0xMiden/miden-client/pull/1045 to be merged so I can run integration tests against this branch before merge.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: I think my actual question is why not always prune to a constant length.

```suggestion
        self.chain_mmr.prune_to(..self.chain_mmr.chain_length()-PRUNED_BLOCK_COUNT)
```

sergerad: Given the `MAX_BLOCK_COUNT` is so low, this would make more sense. Will update to do that. Will simplify the logic.

sergerad:
Integration tests passed 👍

Issue #1064, opened 2025-07-09
Convert from `BTree` to `Hash` where possible


Mirko-von-Leipzig:
With https://github.com/0xMiden/miden-base/issues/1472 having landed we can now convert most of our `BTreeMap` and `BTreeSet` to `HashMap` and `HashSet` respectively. The latter are preferred as they have much better performance in the general case.

There are one or two places where the guaranteed ordering is required, so care should be taken. One known instance is in the mempool subscription where events must retain their chronological ordering.

Himess:
@Mirko-von-Leipzig 
I noticed PR #1291 was closed due to mempool refactoring. Is this issue still relevant, or has the refactoring covered all the required changes?

PR #1109, opened 2025-07-24
chore: Start the usage of hash-based structures

c410-f3r:
Works towards #1064

AFAICT, all changed elements in this PR don't require ordering although I could be wrong.


Mirko-von-Leipzig:
> Hum... I would say ~20 instances of BTreeMap/BTreeSet that could be replaced. Unfortunately, I not familiar enough with the code-base to confidently provide a full replacement at once.

No worries; thanks a lot!

Issue #1111, opened 2025-07-24
Benchmark note inclusion path changes


Mirko-von-Leipzig:
See https://github.com/0xMiden/miden-node/pull/1097#pullrequestreview-3049914011.

#1087 changed storage of note inclusion path from a full to a sparse merkle path. Use the `stress-test` benchmark tool to compare the note storage changes before and after #1087.

Unfortunately the `stress-test` is currently broken #1110. The fix will likely need to be backported to before #1087 in order to compare them.

The very original stress test benchmark is https://github.com/0xMiden/miden-node/pull/657#issuecomment-2737059481 but ideally we would get a proper before/after comparison.

PR #1113, opened 2025-07-25
fix: stress test accounts creation

### LINKED ISSUE #1110: Stress-test is broken
The `stress-test` binary is no longer compatible with the protocol and crashes when seeding the store.

```sh
cargo run --release --bin miden-node-stress-test -- -d . -n 10
```

The initial problem is that the created accounts don't have the required auth component but that's a quick fix.

After that it crashes due to a bad state transition which implies we're building things wrong.

```sh
thread 'main' panicked at bin/stress-test/src/seeding/mod.rs:233:68:
called `Result::unwrap()` on an `Err` value: InconsistentAccountStateTransition {
  account_id: V0(AccountIdV0 { prefix: 6785433524099505824, suffix: 1855316500512461568 }), 
  state_commitment:             Word([17255946689815004829, 1554740649509918628, 1779742492878787968, 17513288932995029093]), 
  remaining_state_commitments: [Word([18092564632307135786, 10840672615757279668, 12706973090722215322, 18087040674081572953])] 
}
```

SantiagoPittella:
closes #1110


SantiagoPittella:
I'm currently running the benchmarks to update the results in the crate's README. The previous results were wrong since we weren't actually creating public accounts.

Mirko-von-Leipzig (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 - [BREAKING] Integrated proxy status endpoint into main proxy service, removing separate status port.
 - RPC requests with wildcard (`*/*`) media-type are not longer rejected ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
+- Stress-test CLI account now properly sets the storage mode and increment nonce in transactions ([#????](https://github.com/0xMiden/miden-node/pull/????)).
```
bobbinth: Let's update the PR number here.

bobbinth (commented):
Looks good! Thank you! I left one small comment inline. Also, I believe we are still planning to update the benchmark results, right?

CHANGELOG.md
SantiagoPittella: Yes, sorry. I opened the PR when I made this work, and then starting to re-run all the benchmarks. Received some bad results on latency, so I'm currently checking that.

bin/stress-test/src/seeding/mod.rs
```diff
@@ -412,13 +421,12 @@ fn create_emit_note_tx(
         .set_item(0, [slot[0], slot[1], slot[2], slot[3] + Felt::new(10)].into())
         .unwrap();
 
-    let (id, vault, sorage, code, nonce) = faucet.clone().into_parts();
-    let updated_faucet = Account::from_parts(id, vault, sorage, code, nonce + ONE);
+    faucet.increment_nonce(ONE).unwrap();
```
drahnr, line 424: 👍

bobbinth:
> Update: the check_nullifiers_by_prefix was panicking in every run. the issue is that we were syncing the 100k accounts in one call but the DB is capped at 1k items per query.

Was this during data seeding or during sync tests? For sync tests, we should probably not query for more than a handful accounts.

SantiagoPittella:
> > Update: the check_nullifiers_by_prefix was panicking in every run. the issue is that we were syncing the 100k accounts in one call but the DB is capped at 1k items per query.
> 
> Was this during data seeding or during sync tests? For sync tests, we should probably not query for more than a handful accounts.

This was during sync tests. We were requesting a sync for all the accounts at once.

SantiagoPittella:
ATM I changed it to do a request every 1k elements. Should I change it to do just 1 requests? with at most 1k accounts for sync state and 1k IDs for `get_notes_by_id`? cc @bobbinth

bin/stress-test/README.md
```diff
@@ -37,35 +37,71 @@ miden-node-stress-test benchmark-store \
...
+| 0      | 22                 | 1                             | 0                              | 384593             | 341.4         |
+| 50     | 186                | 9                             | 1                              | 484994             | 24509.6       |
+| 100    | 199                | 10                            | 1                              | 484994             | 41866.6       |
+| 150    | 219                | 10                            | 1                              | 484994             | 59477.6       |
+| 200    | 218                | 11                            | 1                              | 484994             | 76849.9       |
+| 250    | 222                | 11                            | 1                              | 484994             | 94141.4       |
+| 300    | 228                | 12                            | 1                              | 484994             | 111200.0      |
+| 350    | 232                | 13                            | 1                              | 484994             | 127973.5      |
```
bobbinth: A couple of small comments:

- Could we show "Block size" in KB? Also, would be great to add a sentence or two explaining what block size here means - e.g., each block contains so many transactions and creates/consumes so many notes.
- Could we show "DB size" in MB?

bin/stress-test/README.md
```diff
@@ -37,35 +37,71 @@ miden-node-stress-test benchmark-store \
...
+| account_fungible_asset_deltas       | 2240.0         |  0.0      |
+| notes                               | 50236.0        |  0.5      |
+| account_non_fungible_asset_updates  | 4.0            |  -        |
+| nullifiers                          | 4736.0         |  0.0      |
+| account_storage_map_updates         | 4.0            |  -        |
+| settings                            | 4.0            |  2.0      |
+| account_storage_slot_updates        | 3092.0         |  0.1      |
+| transactions                        | 6168.0         |  0.1      |
+| accounts                            | 26704.0        |  0.3      |
+| note_scripts                        | 8.0            |  8.0      |
```
bobbinth: Could we add a sentence explaining what this database contains (e.g., how many accounts, how many notes)?

Also, let's show "Size" in MB (but I would still keep "KB/Entry").

bin/stress-test/README.md
```diff
@@ -37,35 +37,71 @@ miden-node-stress-test benchmark-store \
...
 ``` bash
-miden-node-stress-test benchmark-store --data-directory ./data --iterations 10000 --concurrency 16 sync-state
-```
-``` bash
-Average request latency: 4.448008ms
-P95 request latency: 6.519875ms
-Average notes per response: 2.9486
+$ miden-node-stress-test benchmark-store --data-directory ./data --iterations 10000 --concurrency 16 sync-state
+
+Average request latency: 73.391934ms
```
bobbinth: ~75ms latency is OK - but since it doesn't include network time, ideally, it should be much lower (e.g., under 10 ms). But maybe the `concurrency` parameter here makes a difference? Could we explain above (in the `Benchmark Store` section) what `concurrency` means?

bin/stress-test/README.md
```diff
@@ -37,35 +37,71 @@ miden-node-stress-test benchmark-store \
...
 ``` bash
-miden-node-stress-test benchmark-store --data-directory ./data --iterations 10000 --concurrency 16 sync-notes
-```
-``` bash
-Average request latency: 3.569797ms
-P95 request latency: 5.724ms
+$ miden-node-stress-test benchmark-store --data-directory ./data --iterations 10000 --concurrency 16 sync-notes
+
+Average request latency: 70.498732ms
+P95 request latency: 7.286208ms
```
bobbinth: Why is average 10x higher than P95? Does this mean that there are some requests that take super long?

bobbinth (commented):
Looks good! Thank you! Not a full review yet - but I left some questions/comments inline.

bin/stress-test/README.md
```diff
@@ -49,35 +51,42 @@ Average DB growth rate: 325.3 KB per block
 ```
 
 #### Block metrics
+
+> Note: Each block contains 256 transactions (16 batches * 16 transactions).
+
 | Block  | Insert Time (ms)   |  Get Block Inputs Time (ms)   |  Get Batch Inputs Time (ms)    | Block Size (B)     |  DB Size (KB) |
```
bobbinth: nit: "Block Size (B)" -> "Block Size (KB)".

bobbinth: And "DB Size (KB)" -> "DB Size (MB)".

bin/stress-test/src/seeding/metrics.rs
```diff
@@ -5,19 +5,18 @@ use std::{
...
-    "notes",
+    "note_scripts",
     "account_non_fungible_asset_updates",
-    "nullifiers",
+    "notes",
     "account_storage_map_updates",
-    "settings",
+    "nullifiers",
     "account_storage_slot_updates",
     "transactions",
```
bobbinth: nit: should we put these in the alphabetical order?

bin/stress-test/src/store/mod.rs
```diff
@@ -41,10 +41,10 @@ pub async fn bench_sync_state(data_directory: PathBuf, iterations: usize, concur
 
     let (store_client, _) = start_store(data_directory).await;
 
-    // each request will have 3 account ids, 3 note tags and will be sent with block number 0
+    // each request will have 5 account ids, 5 note tags and will be sent with block number 0
     let request = |_| {
         let mut client = store_client.clone();
-        let account_batch: Vec<AccountId> = account_ids.by_ref().take(3).collect();
+        let account_batch: Vec<AccountId> = account_ids.by_ref().take(5).collect();
         tokio::spawn(async move { sync_state(&mut client, account_batch, 0).await })
```
bobbinth: I would create constants for these numbers so that they are easy to adjust in the future.

bin/stress-test/src/store/mod.rs
```diff
@@ -109,10 +109,10 @@ pub async fn bench_sync_notes(data_directory: PathBuf, iterations: usize, concur
 
     let (store_client, _) = start_store(data_directory).await;
 
-    // each request will have 3 note tags and will be sent with block number 0.
+    // each request will have 15 note tags and will be sent with block number 0.
     let request = |_| {
         let mut client = store_client.clone();
-        let account_batch: Vec<AccountId> = account_ids.by_ref().take(3).collect();
+        let account_batch: Vec<AccountId> = account_ids.by_ref().take(15).collect();
         tokio::spawn(async move { sync_notes(&mut client, account_batch).await })
```
bobbinth: Same comment as above re constants. Also relevant in a few other places in this file.

bin/stress-test/README.md
SantiagoPittella: I'm checking this, expanded a bit the percentiles shown and this are the current results for sync state:

```
Average request latency: 68.031821ms
P50 request latency: 1.027125ms
P95 request latency: 1.615ms
P99 request latency: 2.838417ms
P99.9 request latency: 41.842534834s
Average notes per response: 1.323
```

bobbinth: Ah - interesting! So, almost all requests complete in under 3ms, but there are some requests that take 40+ seconds?

Not related to this PR, but I wonder if such requests should just time out somehow. Like if we can return within 1 second, we should just cancel the request.

SantiagoPittella: > Ah - interesting! So, almost all requests complete in under 3ms, but there are some requests that take 40+ seconds?

Yes! That's right. I didn't manage to determine which requests are the ones taking that long yet.

> Not related to this PR, but I wonder if such requests should just time out somehow. Like if we can return within 1 second, we should just cancel the request.

Sounds good, will open an issue.

Mirko-von-Leipzig: I would imagine/hope that its not actually the db request that's taking so long, but rather that the request was waiting on being allocated a database connection from the pool.

drahnr: There is https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.disable_lifo_slot

tl;dr If we have a task that does some IO / CPU in sync mode, and blocks that slot indefinitely, a re-used connection might be blocked for a very long time before every polled again. The scheduler of tokio has a heuristic for message passing patterns, bit it bytes if anyone ever does sync heavy workloads, so disabling is a good thing. The underlying issue is, that effectively introeduces a second queue, that cannot be stolen but has priority over the stealable task-queue: https://github.com/tokio-rs/tokio/issues/4941

SantiagoPittella: It is possible that the first request gets stuck waiting for the nullifier tree to be loaded from the DB. For example, this is the result of only 1 sync state:
```
2025-07-31T21:45:03.102847Z  INFO load: miden_node_store::state: crates/store/src/state.rs:1004: Loaded nullifier tree, num_of_leaves: 99960, tree_construction: 21, COMPONENT: "miden-store"
2025-07-31T21:45:24.308338Z  INFO load: miden-store: crates/store/src/state.rs:97: close, time.busy: 42.6s, time.idle: 55.5ms
2025-07-31T21:45:24.309338Z  INFO miden-store: crates/store/src/server/mod.rs:133: Database loaded
2025-07-31T21:45:24.310547Z  INFO store.rpc.rpc/SyncState: miden_node_utils::tracing::grpc: crates/utils/src/tracing/grpc.rs:90: close, time.busy: 387µs, time.idle: 559µs rpc.service: "store.rpc.rpc", rpc.method: "SyncState"
Average request latency: 42.624749833s
```

The request latency matches with the time that takes to load the nullifier tree.

Tried adding a request to the store status before starting to sync the state (and not counting the time the it takes), and the results are this for 10k iteration:

```
Average request latency: 323.768µs
P50 request latency: 297.042µs
P95 request latency: 414.209µs
P99 request latency: 1.1775ms
P99.9 request latency: 2.198292ms
Average notes per response: 1.3128
```

bin/stress-test/README.md
```diff
@@ -37,35 +39,87 @@ miden-node-stress-test benchmark-store \
...
+| account_deltas                     | 1.2             | 0.0        |
+| account_fungible_asset_deltas      | 2.2             | 0.0        |
+| account_non_fungible_asset_updates | 0.0             | -          |
+| account_storage_map_updates        | 0.0             | -          |
+| account_storage_slot_updates       | 3.1             | 0.1        |
+| block_headers                      | 0.1             | 0.3        |
+| notes                              | 49.1            | 0.5        |
+| note_scripts                       | 0.0             | 8.0        |
+| nullifiers                         | 4.6             | 0.0        |
+| transactions                       | 6.0             | 0.1        |
```
bobbinth, line 84: If I sum up the sizes here, I get about 95 MB, but the table above shows 124 MB at 350 blocks (which should have a bit fewer than 100K accounts). What accounts for this discrepancy?

bobbinth (approved):
Looks good! Thank you. I left one more question inline.

Also, I think we should still track down what's happening in https://github.com/0xMiden/miden-node/pull/1113#discussion_r2246456573 - but this could be done in a follow-up PR.

bin/stress-test/README.md
bobbinth: Ah - interesting! I would have though the that request would have timed out. Generally though, we should only start the benchmarks once the store has been fully loaded. Could we use the `Status` endpoint for this? That is, make a request to the status endpoint and only start the benchmark once we receive a response.

bin/stress-test/README.md
SantiagoPittella: Nice catch, re ran it to double check that it wasn't a copy-paste error and it is an actual mismatch. Searched inside the sqlite DB and realized that the missing MB are from the indexes:

```
sqlite> SELECT name, ROUND(SUM(pgsize) / 1024.0 / 1024.0, 2) AS size_mb
   ...> FROM dbstat
   ...> GROUP BY name
   ...> ORDER BY size_mb DESC;
notes|49.06
accounts|26.14
transactions|5.99
idx_transactions_account_id|5.71
nullifiers|4.67
idx_notes_note_id|4.35
idx_nullifiers_prefix|4.33
idx_notes_nullifier|4.33
idx_transactions_block_num|4.25
idx_nullifiers_block_num|4.23
account_storage_slot_updates|3.04
idx_notes_sender|2.88
account_fungible_asset_deltas|2.2
idx_notes_tag|1.63
sqlite_autoindex_notes_1|1.52
account_deltas|1.26
idx_unconsumed_network_notes|1.06
block_headers|0.12
sqlite_schema|0.01
note_scripts|0.01
sqlite_autoindex___diesel_schema_migrations_1|0.0
idx_accounts_network_prefix|0.0
account_storage_map_updates|0.0
account_non_fungible_asset_updates|0.0
__diesel_schema_migrations|0.0
```

The sum matches. Should I expand the table in the readme with the indexes?

bobbinth: Interesting! So indexes are about 25% of the storage. I'd include them - but not sure what's the best way to display (e.g., we could have just a single row for indexes, or list all indexes, or something in-between).

SantiagoPittella: Done! Added:

#### Index stats

| Index                              | Size (MB)       |
| ---------------------------------- | --------------- |
| idx_accounts_network_prefix        | 0.0             |
| idx_notes_note_id                  | 4.4             |
| idx_notes_sender                   | 2.9             |
| idx_notes_tag                      | 1.6             |
| idx_notes_nullifier                | 4.4             |
| idx_unconsumed_network_notes       | 1.1             |
| idx_nullifiers_prefix              | 4.3             |
| idx_nullifiers_block_num           | 4.2             |
| idx_transactions_account_id        | 5.6             |
| idx_transactions_block_num         | 4.2             |

bobbinth (approved):
All looks good! Thank you! Everything else we can do in follow-ups - i.e., in #1111 and #1124.

PR #1114, opened 2025-07-28
chore: Update default genesis max token supply

sergerad:
## Context

The default max supply is small enough to get burned through via faucet in a few days / weeks despite rate limiting.

## Changes

Update default max token supply from 100e9 to 100e15.


bobbinth (approved):
Looks good! Thank you!

PR #1116, opened 2025-07-29
feat: Handle NoteExecutionHints in NTB and manage failed notes

### LINKED ISSUE #1071: Improve note execution management
We should consider how to deal with network notes that fail to execute. If such a note is at the front of the list, the network account would no longer be able to produce any transactions. We currently hack-around this by shuffling the note list on every execution but of course this won't work well if there are many invalid notes.

We can improve matters substantially if we are more deliberate. 

- Take the note's `ExecutionHint` into account.
- Heuristics to place a note in the naughty corner. A note might _never_ be executable for example. This would be especially annoying if _all_ notes of an account are like this, which would actually imply the account should not be considered at all.
- Improvements in the note checker will also make this easier.

The difficulty here will be designing something robust but simple.

Once network transactions by external parties are supported, we could even be aggressive about dropping notes since they could be self-run by users.

sergerad:
## Context

Closes #1071.

The Network Transaction Builder currently does not have any logic to manage notes that have failed. We perform a shuffle of the note list so that we don't get blocked on a forever-failing note.

## Changes
- Add `InflightNetworkNote` struct to allow `AccountState` and NTB `State` to keep track of failed notes.
- Update NTB `State` note selection logic to consider note failure counts and relevant block numbers.


crates/ntx-builder/src/state/mod.rs
```diff
@@ -138,15 +148,49 @@ impl State {
...
+                    let backoff_threshold_passed = block_num.as_usize()
+                        - note
+                            .last_attempt()
+                            .map(|block_num| block_num.as_usize())
+                            .unwrap_or_default()
+                        > backoff_threshold;
+                    can_consume && backoff_threshold_passed
+                })
+                .cloned()
+                .collect::<Vec<_>>();
```
sergerad, line 180: Probably want to extract the backoff logic and unit test it. Might want to factor is so that you can choose how long a note hangs around for based on the block time.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -170,7 +173,28 @@ impl State {
...
             }
 
-            let notes = account.notes().take(limit.get()).cloned().collect::<Vec<_>>();
+            // Remove notes that have failed too many times.
+            account.drain_failing_notes(Self::MAX_NOTE_ATTEMPTS);
+
+            // Select notes from the account that can be consumed or are ready for a retry.
+            let notes = account
+                .notes()
+                .take(limit.get())
```
Mirko-von-Leipzig, line 178: Should the limit move to after the filters?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -208,15 +232,24 @@ impl State {
...
         self.chain_mmr.prune_to(..pruned_block_height.into());
     }
 
     /// Marks a previously selected candidate account as failed, allowing it to be available for
     /// selection again.
     #[instrument(target = COMPONENT, name = "ntx.state.candidate_failed", skip_all)]
-    pub fn candidate_failed(&mut self, candidate: NetworkAccountPrefix) {
-        self.in_progress.remove(&candidate);
+    pub fn candidate_failed(&mut self, candidate: &TransactionCandidate) {
+        let prefix = NetworkAccountPrefix::try_from(candidate.account.id()).unwrap();
```
Mirko-von-Leipzig: Could you add a `SAFETY` comment.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -208,15 +232,24 @@ impl State {
...
+
+        if let Some(account) = self.accounts.get_mut(&prefix) {
+            account.fail(&candidate.notes, candidate.chain_tip_header.block_num());
+        } else {
+            tracing::error!("failed to find network account with prefix {}", prefix);
```
Mirko-von-Leipzig:
```suggestion
            tracing::error!(account.prefix=%prefix, "failed network transaction has no local account state");
```

crates/ntx-builder/src/state/mod.rs
```diff
@@ -138,7 +142,7 @@ impl State {
     ///   - the transaction was submitted successfully, indicated by the associated mempool event
     ///     being submitted
     #[instrument(target = COMPONENT, name = "ntx.state.select_candidate", skip_all)]
-    pub fn select_candidate(&mut self, limit: NonZeroUsize) -> Option<TransactionCandidate> {
+    pub async fn select_candidate(&mut self, limit: NonZeroUsize) -> Option<TransactionCandidate> {
```
Mirko-von-Leipzig: Does this need to be `async`?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -170,7 +173,28 @@ impl State {
...
+                    let block_num = self.chain_tip_header.block_num();
+                    let can_consume = note
+                        .to_inner()
+                        .metadata()
+                        .execution_hint()
+                        .can_be_consumed(block_num)
+                        .unwrap_or(true);
+                    // Filter based on attempts also.
+                    can_consume
+                        && Self::backoff_has_passed(note.last_attempt(), block_num, note.attempts())
```
Mirko-von-Leipzig: Should we merge this all into something like

```rust
fn is_note_available(&self, note: &InflightNetworkNote) -> bool {
    // Check execution hint.
    ...
    // Check backoff
}
```

This does mess with your tests because creating notes with those conditionals might be a pain. But it might be worth trying it out with some basic `P2ID` or other simple note constructors. @PhilippGackstatter where should we go looking for a real but simple note builder? We essentially only care about manipulating the `execution_hint`.

Mirko-von-Leipzig: Alternate idea: move the `can_consume` check as a method into the `InflightNetworkNote`. This might be sensible to do regardless; I don't mind where these single use functions live tbh, but in terms of ownership it probably makes sense that its there..

Or even move the entire `is_note_available` into the note impl since all it really requires from the state is the current chain tip.

crates/ntx-builder/src/state/account.rs
```diff
@@ -153,8 +207,22 @@ impl AccountState {
...
             && self.nullified_notes.is_empty()
     }
+
+    /// Marks the specified notes as failed.
+    pub fn fail(&mut self, notes: &[InflightNetworkNote], block_num: BlockNumber) {
+        for nullifier in notes.iter().map(|note| note.to_inner().nullifier()) {
+            if let Some(note) = self.available_notes.get_mut(&nullifier) {
+                note.fail(block_num);
+            } else {
+                tracing::warn!("failed to find note with nullifier {nullifier}");
```
Mirko-von-Leipzig:
```suggestion
                tracing::warn!(%nullifier, "failed note is not in account's state");
```

We should also consider how we want to formally trace these event's. This is fine for now; but we should consider what to instrument when we refactor to actors. For example it would be good to know the account ID here which is possible similar to `latest_account`.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -170,7 +173,28 @@ impl State {
                 continue;
             }
 
-            let notes = account.notes().take(limit.get()).cloned().collect::<Vec<_>>();
+            // Remove notes that have failed too many times.
+            account.drain_failing_notes(Self::MAX_NOTE_ATTEMPTS);
```
Mirko-von-Leipzig: I would move this above the empty check since this might make the account empty.

crates/ntx-builder/src/state/account.rs
```diff
@@ -3,9 +3,59 @@ use std::collections::{HashMap, VecDeque};
...
+
+/// An unconsumed network note that may have failed to execute.
+///
+/// The block numbers at which the network note was attempted are approximate and may not
+/// reflect the exact block number for which the execution attempt failed. The actual block
+/// will likely be soon after the number that is recorded here.
+#[derive(Debug, Clone)]
+pub struct InflightNetworkNote {
+    note: NetworkNote,
+    attempted_at: Vec<BlockNumber>,
```
Mirko-von-Leipzig: I would either:

1. Replace this with a separate `count` and `last_attempt: Option<BlockNumber>`, or
2. document that this is bounded by the maximum number of failed attempts allowed per note

I would probably go with (1) to keep the reasoning local to the type.

crates/ntx-builder/src/state/mod.rs
sergerad: Opted to pass the prefix into tx context instead.

crates/ntx-builder/src/state/mod.rs
sergerad: Good catch. Must have forgotten to remove it after changing impl.

crates/ntx-builder/src/transaction.rs
```diff
@@ -85,13 +86,12 @@ impl NtxContext {
...
                                 .into_iter()
-                                .map(|note| InputNote::Unauthenticated { note: note.into() })
+                                .map(|note| InputNote::Unauthenticated {
+                                    note: note.into_inner().into(),
+                                })
                                 .collect::<Vec<_>>();
-                            // We shuffle the notes here to prevent having a failing note always in
-                            // front.
-                            notes.shuffle(&mut rand::rng());
                             let notes = InputNotes::new(notes).map_err(NtxError::InputNotes)?;
```
bobbinth: nit: we should now be able to write this as:
```rust
let notes = notes
    .into_iter()
    .map(|note| note.into_inner().into())
    .collect::<Vec<Note>>();
let notes = InputNotes::from_unauthenticated_notes(notes)
    .map_err(NtxError::InputNotes)?;
```
Also, cc @Fumuran - we should probably change `InputNotes::from_unauthenticated_notes()` to take an iterator rather then a vector to make things like this even simpler.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -81,19 +83,18 @@ impl NetworkTransactionBuilder {
                         continue;
                     }
 
-                    let Some(candidate) = state.select_candidate(crate::MAX_NOTES_PER_TX) else {
+                    let Some(candidate) = state.select_candidate(crate::MAX_NOTES_PER_TX)else {
```
bobbinth: nit: missing space before `else`.

crates/ntx-builder/src/state/account.rs
```diff
@@ -3,9 +3,92 @@ use std::collections::{HashMap, VecDeque};
...
+
+    /// Checks if the backoff block period has passed.
+    ///
+    /// The number of blocks passed since the last attempt must be greater than or equal to the
+    /// square of the attempt count.
+    fn backoff_has_passed(
+        last_attempt_block_num: Option<BlockNumber>,
+        current_block_num: BlockNumber,
+        attempt_count: usize,
+    ) -> bool {
```
bobbinth: nit: I would change the order of parameters here to be:
```rust
fn backoff_has_passed(
    current_block_num: BlockNumber,
    last_attempt_block_num: Option<BlockNumber>,
    attempt_count: usize,
) -> bool {
   ...
}
```

crates/ntx-builder/src/state/account.rs
```diff
@@ -3,9 +3,92 @@ use std::collections::{HashMap, VecDeque};
...
+            .can_be_consumed(block_num)
+            .unwrap_or(true);
+        can_consume && Self::backoff_has_passed(self.last_attempt, block_num, self.attempt_count)
+    }
+
+    /// Checks if the backoff block period has passed.
+    ///
+    /// The number of blocks passed since the last attempt must be greater than or equal to the
+    /// square of the attempt count.
+    fn backoff_has_passed(
```
bobbinth: Could we add an example to the doc comments? My understanding is that it the delays would be something like 1, 4, 9, 16 - correct? I think this works, but we could maybe make it a bit better by using exponential function. For example, it could be something like:

$$
\Delta = e^{0.25 \cdot n}
$$

Where $n$ is the number of attempts.

Assuming I did my math correctly, this would give us the following results:

- Roughly 10 attempts in the first 5 mins (5 attempts with the current approach).
- Another 10 attempts in the next hour (8 attempts with the current approach).
- Another 10 attempts in the following 10 hours (15 attempts with the current approach).
- Another 10 attempts in the following week (40 attempts with the current approach).

Basically, the idea is to back off a bit slower in the beginning, but faster later on.

Also, I would if we should impose some limit. For example, if we tried to execute a note for a day, and it didn't work, we probably shouldn't try again (this could be around 30 - 40 attempts).

crates/ntx-builder/src/state/mod.rs
```diff
@@ -92,6 +96,9 @@ pub struct State {
 }
 
 impl State {
+    /// Maximum number of attempts to execute a network note.
+    const MAX_NOTE_ATTEMPTS: usize = 10;
```
bobbinth: As mentioned in the last comment, I'd probably bump this up to something like 30.

crates/ntx-builder/src/state/account.rs
```diff
@@ -17,7 +100,7 @@ pub struct AccountState {
     inflight: VecDeque<Account>,
 
     /// Unconsumed notes of this account.
-    available_notes: HashMap<Nullifier, NetworkNote>,
+    available_notes: HashMap<Nullifier, InflightNetworkNote>,
 
     /// Notes which have been consumed by transactions that are still inflight.
     nullified_notes: HashMap<Nullifier, NetworkNote>,
```
bobbinth: Doesn't this also need to be `HashMap<Nullifier, InflightNetworkNote>`? My understanding is that:

- When a new network transaction is built, we submit it to the block producer, which in turn would call `State::add_transaction()`.
- In `State::add_transaction()`, we'll call `AccountState::add_nullifier()`, which will remove the note from `AccountState::available_notes` and move it to `AccountState::nullified_notes`.
- Then, if transaction execution fails, `AccountState::revert_nullifier()` will get called, which will move the note `AccountState::nullified_notes` to `AccountState::available_notes`.

Currently, this will result in all attempt info being wiped out on transaction failure. But maybe I'm missing something?

crates/ntx-builder/src/state/account.rs
sergerad: Have updated impl, tests, and comments.

crates/ntx-builder/src/transaction.rs
```diff
@@ -145,19 +150,25 @@ impl NtxContext {
...
-                if notes.is_empty() {
-                    let err = error.map_or_else(|| "None".to_string(), |err| err.as_report());
-                    tracing::warn!(%err, "all network notes failed");
+            Ok(NoteAccountExecution::Failure {
+                successful_notes, error, failed_note_id, ..
+            }) => {
+                if let Some(error) = error {
+                    return Err(NtxError::NoteFilter(error));
                 }
+                return Err(NtxError::ConsumptionCheck { successful_notes, failed_note_id });
```
bobbinth: If the error happens, this would not produce a transaction, right? I think the behavior we want is this:

- If there is no error, we build a transaction (like we do now, I think).
- If there is an error, we mark the failed note with the error, and
  - If there are any successful notes, still build a transaction with them.
  - The remaining notes should not be affected.

Also, it may make sense to address https://github.com/0xMiden/miden-base/issues/1566 first (should be pretty simple) - let me know if you'd like me to assign it to you.

sergerad: > let me know if you'd like me to assign it to you.

Yes please.

sergerad: > If the error happens, this would not produce a transaction, right? I think the behavior we want is this:

```rust
                            self.filter_notes(&data_store, notes)
                                .and_then(|notes| self.execute(&data_store, notes))
                                .and_then(|tx| self.prove(tx))
                                .and_then(|tx| self.submit(tx))
                                .await
```

Everything before the failed note should have succeeded here ^.

sergerad: Ah nvm, I keep forgetting that logic is batched, rather than in a loop. Need to have a think then.

sergerad: I have updated the logic now. It works OK with note id as is from base.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - A transaction's account delta is now checked against its commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
 - Added check for Account Id prefix uniqueness when transactions to create accounts are submitted to the mempool ([#1094](https://github.com/0xMiden/miden-node/pull/1094)).
 - RPC accept header now supports specifying the genesis commitment in addition to the RPC version. This lets clients ensure they are on the right network ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
+- Updated Network transaction builder to drop notes that fail excessively and to retry failed notes after some block delay ([#1116](https://github.com/0xMiden/miden-node/pull/1116)).
```
```suggestion
- Retry failed network notes with exponential backoff instead of immediately ([#1116](https://github.com/0xMiden/miden-node/pull/1116))
- Network notes are now dropped after failing 30 times ([#1116](https://github.com/0xMiden/miden-node/pull/1116))
```

Mirko-von-Leipzig: Or 

```suggestion
- Improved handling of network notes in network transaction builder ([#1116](https://github.com/0xMiden/miden-node/pull/1116))
  - Added exponential delay to retrying failed notes
  - Notes are dropped after 30 failures 
```

crates/ntx-builder/src/state/mod.rs
```diff
@@ -204,15 +220,24 @@ impl State {
 
         // Keep MMR pruned.
         let pruned_block_height =
-            (self.chain_mmr.chain_length().as_usize() - MAX_BLOCK_COUNT) as u32;
+            (self.chain_mmr.chain_length().as_usize().saturating_sub(MAX_BLOCK_COUNT)) as u32;
```
Mirko-von-Leipzig: Would changing `MAX_BLOCK_COUNT` to `BlockNumber` simplify things here?

`BlockNumber::checked_sub` does exist so one can still do that.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -204,15 +220,24 @@ impl State {
...
     /// Marks a previously selected candidate account as failed, allowing it to be available for
     /// selection again.
+    ///
+    /// All notes in the candidate will be marked as failed.
     #[instrument(target = COMPONENT, name = "ntx.state.candidate_failed", skip_all)]
-    pub fn candidate_failed(&mut self, candidate: NetworkAccountPrefix) {
-        self.in_progress.remove(&candidate);
+    pub fn candidate_failed(&mut self, candidate: &TransactionCandidate) {
+        if let Some(account) = self.accounts.get_mut(&candidate.account_id_prefix) {
+            account.fail(&candidate.notes, candidate.chain_tip_header.block_num());
```
Mirko-von-Leipzig: iiuc then we're being overly zealous here which is probably fine for now until we can be finer grained. We'll want to distinguish why the tx failed in the future and be a bit more thoughtful about which notes to hold responsible.

No change required imo, just musing out loud.

crates/ntx-builder/src/transaction.rs
```diff
@@ -76,15 +76,17 @@ pub struct NtxContext {
...
             account,
+            account_id_prefix,
             notes,
             chain_tip_header,
             chain_mmr,
         } = tx;
 
-        tracing::Span::current().set_attribute("account.id", account.id());
+        tracing::Span::current()
+            .set_attribute("account.id_prefix", account_id_prefix.to_string().as_str());
```
Mirko-von-Leipzig: I would include both ID and prefix. I imagine we'll want to debug a specific network account and will only know the ID (well at least trivially know the full ID).

I then thought maybe we would want the `bech32` format as well but unfortunately we can only convert to that if we know the network string which is a bit of a pain so let's leave that for now..

Naming is also a problem because account IDs also have a prefix, which is different to the network account prefix we use as the tag https://github.com/0xMiden/miden-node/blob/d54b2c2efa208484a57e560e35a7541c095e8252/crates/proto/src/domain/account.rs#L361-L364

Maybe something like: 
```suggestion
        tracing::Span::current().set_attribute("account.id", account.id());
        tracing::Span::current()
            .set_attribute("account.id.network_prefix", account_id_prefix.to_string().as_str());
```

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig, line 1: Not this PRs problem but I've noticed the ntx spans don't actually link together :/ Probably the `instrument` does not propagate into the block on etc.

Mirko-von-Leipzig: We could tackle that as part of the actor refactor.

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig: Additional note to self - maybe a good idea to have an extension method for an account itself to inject common properties like `id`, `commitment` and various counts.

crates/ntx-builder/src/transaction.rs
```diff
@@ -159,28 +161,44 @@ impl NtxContext {
...
                     .cloned()
-                    .collect::<Vec<InputNote>>();
+                    .collect::<Vec<_>>();
+
+                // If none are successful, abort.
+                if successful_notes.is_empty() {
+                    tracing::warn!(
+                        err = %error.as_ref().map_or_else(|| "none".to_string(), ErrorReport::as_report),
+                        "failed to check note consumability",
+                    );
```
Mirko-von-Leipzig: I think this will already be reported as part of the tracing/otel error instrumentation
```suggestion
```

crates/ntx-builder/src/transaction.rs
```diff
@@ -159,28 +161,44 @@ impl NtxContext {
...
-                    let err = error.map_or_else(|| "None".to_string(), |err| err.as_report());
-                    tracing::warn!(%err, "all network notes failed");
+                // If there were some successful notes but still an error, log it
+                let successful_count = successful_notes.len();
+                if let Some(err) = error {
+                    tracing::warn!(
+                        %err,
+                        successful_count,
+                        "failed to check note consumability",
+                    );
```
Mirko-von-Leipzig: I would expect this to happen fairly often. I'm unsure what we should do here tbh.

It feels like logging it as an event would be problematic. Maybe we could inject this as a field in the span instead so its searchable, but doesn't cause additional flooding?

Its also possible we'll want to inject this higher up but 🤷

I suspect this will also change in the future as the checker changes shape.

```suggestion
                let successful_count = successful_notes.len();
                if let Some(err) = error {
                    Span::current().set_attribute("failed_note.err", err.as_report());
                    Span::current().set_attribute("failed_note.id", failed_note_id);
```

crates/ntx-builder/src/transaction.rs
```diff
@@ -144,12 +146,12 @@ impl NtxContext {
         &self,
         data_store: &NtxDataStore,
         notes: InputNotes<InputNote>,
-    ) -> NtxResult<InputNotes<InputNote>> {
+    ) -> NtxResult<(InputNotes<InputNote>, Option<NoteId>)> {
```
Mirko-von-Leipzig: Might be worth creating (or update the doc comment)

```rust
struct FilteredNotes {
    viable: InputNotes<InputNote>,
    first_failure: Option<NoteId>,
}
```

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -92,14 +92,13 @@ impl NetworkTransactionBuilder {
                         continue;
                     };
 
-                    let prefix = NetworkAccountPrefix::try_from(candidate.account.id()).unwrap();
                     let task_id = inflight.spawn({
                         let context = context.clone();
-                        context.execute_transaction(candidate)
+                        context.execute_transaction(TransactionCandidate::clone(&candidate))
```
Mirko-von-Leipzig: The clone is a pity, but it will go a way once we move to an actor based system. Or at least we can replace it with an `Arc` then.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -114,19 +113,24 @@ impl NetworkTransactionBuilder {
...
+                    let mut candidate = inflight_idx.remove(&task_id).unwrap();
 
                     match completed {
                         // Nothing to do. State will be updated by the eventual mempool event.
-                        Ok((_, Ok(_))) => {},
+                        Ok((_, Ok(None))) => {},
+                        Ok((_, Ok(Some(failed_note_id)))) => {
+                            // Filter out successful notes from the candidate's notes and register them as failed.
+                            candidate.notes.retain(|note| note.to_inner().id() == failed_note_id);
+                            state.candidate_failed(&candidate);
```
Mirko-von-Leipzig: Ah... I think this won't work because the tx itself did go through and the state should wait for the mempool event. 

Marking the candidate tx as failed informs the state that the account should be available for selection again which is incorrect.

I think we need to separate marking a candidate tx as failed from marking notes as failed.

```suggestion
state.notes_failed(candidate.prefix, [failed_note_id]);
```

and then in the other branches you mark the tx as failed, and also mark the failed notes.

crates/ntx-builder/src/state/account.rs
```diff
@@ -4,8 +4,105 @@ use miden_node_proto::domain::account::NetworkAccountPrefix;
...
+        last_attempt_block_num: Option<BlockNumber>,
+        attempt_count: usize,
+    ) -> bool {
+        if attempt_count == 0 {
+            return true;
+        }
+        // Compute the number of blocks passed since the last attempt.
+        let last_attempt_block_num =
+            last_attempt_block_num.map(|block_num| block_num.as_usize()).unwrap_or_default();
+        let blocks_passed = current_block_num.as_usize().saturating_sub(last_attempt_block_num);
```
Mirko-von-Leipzig:
```suggestion
        let blocks_passed = last_attempt_block_num
            .map(|last| current_block_num.checked_sub(last))
            .flatten()
            .unwrap_or_default();
```

crates/ntx-builder/src/state/account.rs
```diff
@@ -4,8 +4,105 @@ use miden_node_proto::domain::account::NetworkAccountPrefix;
...
+    /// - 2 attempts: 2 blocks
+    /// - 3 attempts: 3 blocks
+    /// - 10 attempts: 13 blocks
+    /// - 20 attempts: 149 blocks
+    /// - etc...
+    #[allow(clippy::cast_precision_loss, clippy::cast_sign_loss)]
+    fn backoff_has_passed(
+        current_block_num: BlockNumber,
+        last_attempt_block_num: Option<BlockNumber>,
+        attempt_count: usize,
```
Mirko-von-Leipzig: Nit
```suggestion
        chain_tip: BlockNumber,
        last_attempt: Option<BlockNumber>,
        attempts: usize,
```

crates/ntx-builder/src/state/account.rs
Mirko-von-Leipzig: Or 

```rust
        let last_attempt_block_num = last_attempt_block.unwrap_or(current_block_num);

        let blocks_passed = current_block_num.checked_sub(last_attempt_block_num)
            .unwrap_or_default();
```

crates/ntx-builder/src/state/account.rs
```diff
@@ -132,10 +229,15 @@ impl AccountState {
...
     }
 
-    pub fn notes(&self) -> impl Iterator<Item = &NetworkNote> {
+    pub fn notes(&self) -> impl Iterator<Item = &InflightNetworkNote> {
         self.available_notes.values()
     }
 
+    /// Drains all notes that have failed to be consumed after a certain number of attempts.
+    pub fn drain_failing_notes(&mut self, max_attempts: usize) {
+        self.available_notes.retain(|_, note| note.attempt_count() <= max_attempts);
```
Mirko-von-Leipzig: 🤔 technically?

```suggestion
        self.available_notes.retain(|_, note| note.attempt_count() < max_attempts);
```

Mirko-von-Leipzig (changes requested):
Mostly minor nits and suggestions. The main concern is the `candidate_failed` in the main worker loop in the success case.

crates/ntx-builder/src/state/mod.rs
sergerad: Feels confusing to overload the semantics of `BlockNumber` to include a range of blocks here. As in `MAX_BLOCK_COUNT` is a number of blocks, whereas `BlockNumber` is a block index conceptually.

crates/ntx-builder/src/state/mod.rs
sergerad: I believe I can address this now that https://github.com/0xMiden/miden-base/pull/1680 is merged.

crates/ntx-builder/src/transaction.rs
```diff
@@ -100,23 +103,23 @@ impl NtxContext {
...
-                                .map(|note| InputNote::Unauthenticated { note: note.into() })
-                                .collect::<Vec<_>>();
-                            // We shuffle the notes here to prevent having a failing note always in
-                            // front.
-                            notes.shuffle(&mut rand::rng());
-                            let notes = InputNotes::new(notes).map_err(NtxError::InputNotes)?;
+                                .map(|note| note.into_inner().into())
+                                .collect::<Vec<Note>>();
+                            let notes = InputNotes::from_unauthenticated_notes(notes)
+                                .map_err(NtxError::InputNotes)?;
```
sergerad: @bobbinth is this an ok way of going from `Note` -> `InputNote` (unauthenticated)? I don't see another way.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -204,15 +220,37 @@ impl State {
...
+    /// All notes in the candidate will be marked as failed. Does not remove the candidate from the
+    /// in-progress pool.
+    #[instrument(target = COMPONENT, name = "ntx.state.notes_failed", skip_all)]
+    pub fn notes_failed(&mut self, candidate: &TransactionCandidate) {
+        if let Some(account) = self.accounts.get_mut(&candidate.account_id_prefix) {
+            account.fail(&candidate.notes, candidate.chain_tip_header.block_num());
+        } else {
+            tracing::error!(account.prefix=%candidate.account_id_prefix, "failed network notes have no local account state");
+        }
+    }
```
sergerad, line 238: @Mirko-von-Leipzig lmk if this is what you had in mind for this [comment](https://github.com/0xMiden/miden-node/pull/1116#discussion_r2259804342).

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -114,19 +113,27 @@ impl NetworkTransactionBuilder {
...
                     // SAFETY: both inflights should have the same set.
-                    let candidate = inflight_idx.remove(&task_id).unwrap();
+                    let mut candidate = inflight_idx.remove(&task_id).unwrap();
 
                     match completed {
-                        // Nothing to do. State will be updated by the eventual mempool event.
-                        Ok((_, Ok(_))) => {},
+                        // Inform state of failed notes.
+                        Ok((_, Ok(failed))) => {
+                            // TODO: Analyze errors provided for each failed note to determine what to do here.
```
sergerad: @Mirko-von-Leipzig I'm not sure exactly what we would do here with the errors but I think this is the place we would do it?

crates/ntx-builder/src/state/mod.rs
sergerad: I don't think it is. Or I'm at least missing additional logic but I don't quite follow your suggestion.

crates/ntx-builder/src/transaction.rs
```diff
@@ -158,42 +140,29 @@ impl NtxContext {
...
-                let successful_count = successful_notes.len();
-                if let Some(err) = error {
-                    tracing::warn!(
-                        %err,
-                        successful_count,
-                        "failed to check note consumability",
-                    );
-                }
+                // TODO: do something with error? Return errors instead of InputNotes?
+                // Map failed notes to input notes.
```
Mirko-von-Leipzig: Nit: missing plural

crates/ntx-builder/src/transaction.rs
```diff
@@ -45,7 +45,7 @@ pub enum NtxError {
     #[error("failed to filter notes")]
     NoteFilter(#[source] TransactionExecutorError),
     #[error("no viable notes")]
-    NoViableNotes,
+    NoViableNotes(),
```
Mirko-von-Leipzig:
```suggestion
    NoViableNotes,
```

crates/ntx-builder/src/transaction.rs
```diff
@@ -129,12 +128,12 @@ impl NtxContext {
...
         data_store: &NtxDataStore,
         notes: InputNotes<InputNote>,
-    ) -> NtxResult<InputNotes<InputNote>> {
+    ) -> NtxResult<(InputNotes<InputNote>, Vec<FailedNote>)> {
         let executor: TransactionExecutor<'_, '_, _, UnreachableAuth> =
             TransactionExecutor::new(data_store, None);
         let checker = NoteConsumptionChecker::new(&executor);
 
-        let notes = match Box::pin(checker.check_notes_consumability(
+        match Box::pin(checker.check_notes_consumability(
```
Mirko-von-Leipzig, line 164: Do we know if the checker actually tests the failed notes fully yet?

As in, is it checking each note in the failed set or is it aborting at the first failure and marking the rest as failed?

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: I think we can't realistically do anything with the errors just yet.

I imagine we would need to do some sort of account type introspection to be able to say anything meaningful about how the errors relate to the account.

I would ignore that until we complete the actor refactoring. At that stage we could identify types of network accounts and specialize them somehow 🤷 e.g. this is a P2ID note that fails due to lack of funds; no need to retry it until the account gets more funds of that asset.

The actor refactor makes this more manageable imo because each account will be self-contained so its easier to be more specific.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -114,19 +113,27 @@ impl NetworkTransactionBuilder {
...
+                            // Filter out successful notes from the candidate's notes and register them as failed.
+                            let failed_note_ids = failed.iter().map(|note| note.note.id()).collect::<HashSet<_>>();
```
Mirko-von-Leipzig: Should we change the output of the execution to `HashSet` from `Vec`?

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: I think this is still incorrect. The candidate tx can fail for a multitude of reasons - not necessarily that the notes are to blame.

I would revert this function, and have the main loop inspect the error to determine whether `notes_failed(..)` should be called and with what notes.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -114,19 +113,27 @@ impl NetworkTransactionBuilder {
...
+                            candidate.notes.retain(|note| {
+                                failed_note_ids.contains(&note.to_inner().id())
+                            });
+                            state.notes_failed(&candidate);
```
Mirko-von-Leipzig: I'm not sure we should be re-using the candidate type for this, the state doesn't require that much data - it just needs the account prefix and note IDs. This would remove a bunch of clones e.g. `inflight_idx` can reverted to only storing the prefix and not the full candidate data.

e.g.

```rust
state.notes_failed(prefix, failed_ids);
```

crates/ntx-builder/src/builder/mod.rs
Mirko-von-Leipzig: As mentioned https://github.com/0xMiden/miden-node/pull/1116/files#r2276251029 I think we shouldn't need this

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -114,19 +113,27 @@ impl NetworkTransactionBuilder {
...
+                        },
                         // Inform state if the tx failed.
                         Ok((_, Err(err))) => {
                             tracing::warn!(err=err.as_report(), "network transaction failed");
-                            state.candidate_failed(candidate);
+                            state.candidate_failed(&candidate);
```
Mirko-von-Leipzig: I would separate marking the candidate as failed from marking the notes as failed. The tx can fail for a variety of reasons, not only because of the notes.

We need to inspect the error to determine whether some notes should be marked as poor.

Mirko-von-Leipzig: I imagine the only error case would be `NoViableNotes` in which case we should change it to `NoViableNotes(HashSet<NoteId>)` (or something similar) so that we can know here which notes to target.

crates/ntx-builder/src/state/mod.rs
Mirko-von-Leipzig: Its close, though I would change the signature and ordering of things as suggested

https://github.com/0xMiden/miden-node/pull/1116/files#r2276251029
https://github.com/0xMiden/miden-node/pull/1116/files#r2276237648
https://github.com/0xMiden/miden-node/pull/1116/files#r2276254936

crates/ntx-builder/src/transaction.rs
```diff
@@ -143,26 +142,22 @@ impl NtxContext {
...
+                    .collect::<Vec<_>>();
+
+                // If none are successful, abort.
+                if successful.is_empty() {
+                    return Err(NtxError::NoViableNotes());
                 }
-                InputNotes::new_unchecked(notes)
+
+                // Return successful and failed note id.
+                Ok((InputNotes::new_unchecked(successful), failed))
```
Mirko-von-Leipzig: Nit: `// SAFETY` for the unchecked

crates/ntx-builder/src/transaction.rs
Mirko-von-Leipzig: Though as mentioned https://github.com/0xMiden/miden-node/pull/1116#discussion_r2276257495 maybe this should be

```suggestion
    AllNotesFailed(HashSet<NoteId>),
```

crates/ntx-builder/src/transaction.rs
sergerad: block_on is gone now

crates/ntx-builder/src/transaction.rs
sergerad: It only ever returns 1 failed note in that vector atm.

crates/ntx-builder/src/builder/mod.rs
sergerad: I don't follow why you would need a set/list of note Ids when you know all notes failed for a transaction candidate.

sergerad: Ah right because we have now removed tx candidate from the index. Hmm

sergerad: I have updated the logic. Still unsure about it. Do we only want candidate failed to be called on unexpected error this way?

crates/ntx-builder/src/transaction.rs
```diff
@@ -143,26 +142,23 @@ impl NtxContext {
...
-                    .filter_map(|success| notes.iter().find(|note| note.id() == success.id()))
+                    .filter_map(|id| notes.iter().find(|&note| note.id() == id.id()))
                     .cloned()
-                    .collect::<Vec<InputNote>>();
-                // TODO revisit verbosity once we scale up to many more notes
-                for failed in failed {
-                    let err = failed.error.as_report();
-                    let note_id = failed.note.id();
-                    tracing::debug!(%note_id, %err, "note failed");
+                    .collect::<Vec<_>>();
```
bobbinth: We probably don't need to do this as we should be able to create `InputNotes` like so:
```rust
let input_notes = InputNotes::from_unauthenticated_notes(successful)?;
```

crates/ntx-builder/src/transaction.rs
```diff
@@ -129,12 +128,12 @@ impl NtxContext {
         &self,
         data_store: &NtxDataStore,
         notes: InputNotes<InputNote>,
-    ) -> NtxResult<InputNotes<InputNote>> {
+    ) -> NtxResult<(InputNotes<InputNote>, Vec<FailedNote>)> {
```
bobbinth: Let's update doc comments of this method to reflect the latest functionality.

crates/ntx-builder/src/transaction.rs
```diff
@@ -78,36 +78,35 @@ impl NtxContext {
     pub fn execute_transaction(
         self,
         tx: TransactionCandidate,
-    ) -> impl FutureMaybeSend<NtxResult<()>> {
+    ) -> impl FutureMaybeSend<NtxResult<Vec<FailedNote>>> {
```
bobbinth, line 103: Let's add doc comments to this method to explain what it does and returns.

crates/ntx-builder/src/transaction.rs
```diff
@@ -78,36 +78,35 @@ impl NtxContext {
...
-                    .map(|note| InputNote::Unauthenticated { note: note.into() })
-                    .collect::<Vec<_>>();
-                // We shuffle the notes here to prevent having a failing note always in
-                // front.
-                notes.shuffle(&mut rand::rng());
-                let notes = InputNotes::new(notes).map_err(NtxError::InputNotes)?;
+                let notes =
+                    notes.into_iter().map(|note| note.into_inner().into()).collect::<Vec<Note>>();
+                let notes =
+                    InputNotes::from_unauthenticated_notes(notes).map_err(NtxError::InputNotes)?;
```
bobbinth: nit: I would probably move this into the `filter_notes()` function. This would require changing its signature to something like:
```rust
async fn filter_notes(
    &self,
    data_store: &NtxDataStore,
    notes: Vec<Note>,
) -> NtxResult<(InputNotes<InputNote>, Vec<FailedNote>)>
```

crates/ntx-builder/src/state/mod.rs
```diff
@@ -30,12 +30,16 @@ const MAX_BLOCK_COUNT: usize = 4;
 ///
 /// Contains the data pertaining to a specific network account which can be used to build a network
 /// transaction.
+#[derive(Clone)]
 pub struct TransactionCandidate {
     /// The current inflight state of the account.
     pub account: Account,
 
+    /// The account id prefix corresponding to the candidate account.
+    pub account_id_prefix: NetworkAccountPrefix,
```
bobbinth: Why do we need to add this here? Could we not get the prefix from the `account` field?

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -113,18 +113,29 @@ impl NetworkTransactionBuilder {
...
+                        },
+                        // Transaction execution failed.
                         Ok((_, Err(err))) => {
                             tracing::warn!(err=err.as_report(), "network transaction failed");
-                            state.candidate_failed(candidate);
+                            if let NtxError::AllNotesFailed(failed) = err {
+                                let notes = failed.into_iter().map(|note| note.note).collect::<Vec<_>>();
+                                state.notes_failed(candidate,
+                                                   notes.as_slice(),
+                                                   block_num);
```
bobbinth: Same nit as above.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -113,18 +113,29 @@ impl NetworkTransactionBuilder {
...
+                            let notes = failed.into_iter().map(|note| note.note).collect::<Vec<_>>();
+                            state.notes_failed(candidate,
+                                               notes.as_slice(),
+                                               block_num);
+                        },
+                        // Transaction execution failed.
                         Ok((_, Err(err))) => {
                             tracing::warn!(err=err.as_report(), "network transaction failed");
-                            state.candidate_failed(candidate);
+                            if let NtxError::AllNotesFailed(failed) = err {
```
bobbinth: Not for this PR, but we should probably think through what happens if we get other errors. This may require re-visiting `NtxError` enum to see if all the variants there are relevant.

crates/ntx-builder/src/state/account.rs
```diff
@@ -4,8 +4,105 @@ use miden_node_proto::domain::account::NetworkAccountPrefix;
...
+
+    /// Consumes the inflight network note and returns the inner network note.
+    pub fn into_inner(self) -> NetworkNote {
+        self.note
+    }
+
+    /// Returns a reference to the inner network note.
+    pub fn to_inner(&self) -> &NetworkNote {
+        &self.note
+    }
```
bobbinth, line 43: We could also consider adding something like `into_note()` and `to_note()` method as we frequently need the `Note` rather than `NetworkNote`.

crates/ntx-builder/src/state/account.rs
```diff
@@ -4,8 +4,105 @@ use miden_node_proto::domain::account::NetworkAccountPrefix;
...
+    ///
+    /// The note is available if it can be consumed and the backoff period has passed.
+    pub fn is_available(&self, block_num: BlockNumber) -> bool {
+        let can_consume = self
+            .to_inner()
+            .metadata()
+            .execution_hint()
+            .can_be_consumed(block_num)
+            .unwrap_or(true);
+        can_consume && Self::backoff_has_passed(block_num, self.last_attempt, self.attempt_count)
```
bobbinth: nit: `backoff_has_passed` can be just a "free" helper function - right? (i.e., it does not need to be attached to `Self`).

Also, I'd probably rename it to `has_backoff_passed()`

crates/ntx-builder/src/state/account.rs
```diff
@@ -132,10 +229,15 @@ impl AccountState {
...
 
-    pub fn notes(&self) -> impl Iterator<Item = &NetworkNote> {
+    pub fn notes(&self) -> impl Iterator<Item = &InflightNetworkNote> {
         self.available_notes.values()
     }
 
+    /// Drains all notes that have failed to be consumed after a certain number of attempts.
+    pub fn drain_failing_notes(&mut self, max_attempts: usize) {
+        self.available_notes.retain(|_, note| note.attempt_count() < max_attempts);
+    }
```
bobbinth: nit: "drain" usually implies that we return the drained data. Would `remove_failing_notes()` or `drop_failing_notes()` be better here?

crates/ntx-builder/src/state/account.rs
```diff
@@ -4,8 +4,105 @@ use miden_node_proto::domain::account::NetworkAccountPrefix;
...
+    ///
+    /// The number of blocks passed since the last attempt must be greater than or equal to
+    /// e^(0.25 * `attempt_count`).
+    ///
+    /// The expected thresholds are therefore:
+    /// - 1 attempt: 2 blocks
+    /// - 2 attempts: 2 blocks
+    /// - 3 attempts: 3 blocks
+    /// - 10 attempts: 13 blocks
+    /// - 20 attempts: 149 blocks
```
bobbinth: This is not super clear to me. Does "2 attempts: 2 blocks" mean that once there is a failure, we'll try to execute the note 2 more times in the next two blocks? Or is it something else?

crates/ntx-builder/src/state/account.rs
sergerad: `to_inner()` seemed to only be used once and not to reach `Note` type.

I added a `From<InflightNewtorkNote> for Note` rather than `into_note()` because the former did not require further exporting `InflightNetworkNote` type into the module doing the conversion. So
```rust
```

crates/ntx-builder/src/state/mod.rs
sergerad: I don't see a way to get the `NetworkAccountPrefix` from `Account` at all.

bobbinth: Could it not be something like `NetworkAccountPrefix::try_from(account.id())`?

sergerad: Yep ty

crates/ntx-builder/src/state/account.rs
sergerad: Updated

crates/ntx-builder/src/builder/mod.rs
sergerad: I have updated this to match on all variants. I'm not sure when it is appropriate to allow a candidate to be selected again. But at least we can chip away at this list explicitly now.

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -113,18 +116,39 @@ impl NetworkTransactionBuilder {
...
                     match completed {
-                        // Nothing to do. State will be updated by the eventual mempool event.
-                        Ok((_, Ok(_))) => {},
-                        // Inform state if the tx failed.
+                        // Some notes failed.
+                        Ok((_, Ok(failed))) => {
+                            let notes = failed.into_iter().map(|note| note.note).collect::<Vec<_>>();
+                            state.notes_failed(candidate,
+                                               notes.as_slice(),
+                                               block_num);
```
bobbinth: nit: the formatting here looks weird - I think this could be just a single line like:
```rust
state.notes_failed(candidate, notes.as_slice(), block_num);
```

crates/ntx-builder/src/state/mod.rs
```diff
@@ -166,7 +172,13 @@ impl State {
                 continue;
             }
 
-            let notes = account.notes().take(limit.get()).cloned().collect::<Vec<_>>();
+            // Select notes from the account that can be consumed or are ready for a retry.
+            let notes = account
+                .notes()
+                .filter(|&note| note.is_available(self.chain_tip_header.block_num()))
```
bobbinth: nit: I would maybe move the filtering logic into `AccountState`. We could do this by introducing something like `available_notes()` method. Then, here it would look something like:
```rust
let notes = account
    .available_notes(self.chain_tip_header.block_num())
    .take(limit.get())
    .cloned()
    .collect::<Vec<_>>();
```

crates/ntx-builder/src/transaction.rs
```diff
@@ -74,67 +74,94 @@ pub struct NtxContext {
...
 
-                let notes = Box::pin(self.filter_notes(&data_store, notes)).await?;
-                let executed = Box::pin(self.execute(&data_store, notes)).await?;
+                let notes = notes.into_iter().map(Note::from).collect::<Vec<_>>();
+                let (successful, failed) = self.filter_notes(&data_store, notes).await?;
+                let executed = Box::pin(self.execute(&data_store, successful)).await?;
                 let proven = Box::pin(self.prove(executed)).await?;
                 self.submit(proven).await?;
-                Ok(())
+                Ok(failed)
```
bobbinth, line 126: This is fine for now (no need to change anything) but if `self.execute()` or `self.prove()` fail for some reason, failed notes will never be marked as "failed". We can take this into consideration as we refactor this in the future to work more in the actor-model-like fashion.

PR #1120, opened 2025-07-30
store: deduplicate code blobs stored in database

### LINKED ISSUE #1083: Move account code storage into a separate table
Currently, the `accounts` table stores full account data (in its `details` field). Over the long term, we should normalize this into multiple tables (e.g., see https://github.com/0xMiden/miden-node/discussions/722#discussioncomment-13748611) but for now, we should break up the `details` field into multiple fields and move account code into a separate table as this will reduce data duplication quite a bit (multiple accounts sharing the same code would not need to keep multiple copies of this code).

The simplest approach to change the `accounts` able as follows:

```sql
CREATE TABLE accounts (
    account_id                BLOB NOT NULL,
    network_account_id_prefix INTEGER NULL
    account_commitment        BLOB NOT NULL,
    code_commitment           BLOB,
    storage                   BLOB,
    vault                     BLOB,
    nonce                     INTEGER,
    block_num                 INTEGER NOT NULL,
)

CREATE TABLE account_code (
    commitment: BLOB,
    code:       BLOB,
)
```
Then, `accounts.code_commitment` would be linked to the code via `account_code.commitment`.

drahnr:
Modifies the table

* externalizes code by code_commitement into a separate table
* expands former details into separate columns

Closes #1083

Open Issues:

* [x] cleanup some rough edges
* [x] cross check constraints on types
* [x] verify `LEFT JOIN` is the right thing, or if `INNER JOIN` might be a better fit

### DIFF
--- a/crates/store/src/db/migrations/2025062000000_setup/up.sql
+++ b/crates/store/src/db/migrations/2025062000000_setup/up.sql
@@ -6,16 +6,31 @@ CREATE TABLE block_headers (
     CONSTRAINT block_header_block_num_is_u32 CHECK (block_num BETWEEN 0 AND 0xFFFFFFFF)
 );
 
+CREATE TABLE account_codes (
+    code_commitment BLOB NOT NULL,
+    code            BLOB NOT NULL,
+    PRIMARY KEY(code_commitment)
+) WITHOUT ROWID;
 
 CREATE TABLE accounts (
     account_id                              BLOB NOT NULL,
     network_account_id_prefix               INTEGER NULL, -- 30-bit account ID prefix, only filled for network accounts
     account_commitment                      BLOB NOT NULL,
     block_num                               INTEGER NOT NULL,
-    details                                 BLOB,
+    code_commitment                         BLOB,
+    storage                                 BLOB,
+    vault                                   BLOB,
+    nonce                                   INTEGER,
 
     PRIMARY KEY (account_id),
-    FOREIGN KEY (block_num) REFERENCES block_headers(block_num)
+    FOREIGN KEY (block_num) REFERENCES block_headers(block_num),
+    FOREIGN KEY (code_commitment) REFERENCES account_codes(code_commitment),
+    CONSTRAINT all_null_or_none_null CHECK
+        (
+            (code_commitment IS NOT NULL AND storage IS NOT NULL AND vault IS NOT NULL AND nonce IS NOT NULL)
+            OR
+            (code_commitment IS NULL AND storage IS NULL AND vault IS NULL AND nonce IS NULL)
+        )
 ) WITHOUT ROWID;
 
 CREATE INDEX idx_accounts_network_prefix ON accounts(network_account_id_prefix) WHERE network_account_id_prefix IS NOT NULL;
--- a/crates/store/src/db/migrations/2025062000000_setup/up.sql
+++ b/crates/store/src/db/migrations/2025062000000_setup/up.sql
@@ -15,8 +15,8 @@ CREATE TABLE account_codes (
 CREATE TABLE accounts (
     account_id                              BLOB NOT NULL,
     network_account_id_prefix               INTEGER NULL, -- 30-bit account ID prefix, only filled for network accounts
-    account_commitment                      BLOB NOT NULL,
     block_num                               INTEGER NOT NULL,
+    account_commitment                      BLOB NOT NULL,
     code_commitment                         BLOB,
     storage                                 BLOB,
     vault                                   BLOB,
--- a/crates/store/src/db/models/queries.rs
+++ b/crates/store/src/db/models/queries.rs
@@ -481,15 +481,18 @@ pub(crate) fn upsert_accounts(
         account_id: AccountId,
     ) -> Result<Vec<Account>, DatabaseError> {
         let account_id = account_id.to_bytes();
-        let accounts =
-            SelectDsl::select(
-                schema::accounts::table.left_join(schema::account_codes::table.on(
-                    schema::accounts::code_commitment.eq(schema::account_codes::code_commitment),
-                )),
-                (AccountRaw::as_select(), schema::account_codes::code.nullable()),
-            )
-            .filter(schema::accounts::account_id.eq(account_id))
-            .get_results::<(AccountRaw, Option<Vec<u8>>)>(conn)?;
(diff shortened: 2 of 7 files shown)


crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -6,16 +6,31 @@ CREATE TABLE block_headers (
...
+    code_commitment BLOB NOT NULL,
+    code            BLOB NOT NULL,
+    PRIMARY KEY(code_commitment)
+) WITHOUT ROWID;
 
 CREATE TABLE accounts (
     account_id                              BLOB NOT NULL,
     network_account_id_prefix               INTEGER NULL, -- 30-bit account ID prefix, only filled for network accounts
     account_commitment                      BLOB NOT NULL,
     block_num                               INTEGER NOT NULL,
```
bobbinth, line 18: nit (feel free to ignore): I'd maybe move `block_num` to the end of the field list. It would feel a bit more natural if `code_commitment` can right after `account_commitment`.

crates/store/src/db/schema.rs
```diff
@@ -50,8 +50,18 @@ diesel::table! {
         account_id -> Binary,
         network_account_id_prefix -> Nullable<BigInt>,
         account_commitment -> Binary,
+        code_commitment -> Binary, // double check if this can ever be the case, I don't think so
```
bobbinth: What is the comment referring to?

Also, shouldn't it be `Nullable<Binary>` (since code commitment can be `NULL`)?

crates/store/src/db/models/types.rs
```diff
@@ -455,3 +497,23 @@ impl From<(NoteRecord, Option<Nullifier>)> for NoteInsertRowRaw {
...
+pub(crate) struct AccountRowInsert {
+    pub(crate) account_id: Vec<u8>,
+    pub(crate) network_account_id_prefix: Option<i64>,
+    pub(crate) account_commitment: Vec<u8>,
+    pub(crate) block_num: i64,
+    pub(crate) nonce: Option<i64>,
+    pub(crate) storage: Option<Vec<u8>>,
+    pub(crate) vault: Option<Vec<u8>>,
+    pub(crate) code_commitment: Option<Vec<u8>>,
+}
```
bobbinth: nit: let's put the fields here in the same order as in the table.

crates/store/src/db/models/queries.rs
```diff
@@ -481,11 +481,20 @@ pub(crate) fn upsert_accounts(
...
-                .get_results::<Vec<u8>>(conn)?;
-        let accounts = deserialize_raw_vec::<_, Account>(account_details_serialized)?;
+        let accounts =
+            SelectDsl::select(
+                schema::accounts::table.left_join(schema::account_codes::table.on(
+                    schema::accounts::code_commitment.eq(schema::account_codes::code_commitment),
+                )),
+                (AccountRaw::as_select(), schema::account_codes::code.nullable()),
+            )
+            .filter(schema::accounts::account_id.eq(account_id))
```
bobbinth: Question: should we add comments with SQL statements next to Disel code? We do this in some other places in this file, and I personally find SQL statements much easier to read.

crates/store/src/db/models/queries.rs
bobbinth, line 1: Not for this PR, but as we are refactoring this file would be great to "normalize" the functions (e.g., sometimes we pass lists of values as slices, other times as iterators), add doc comments to functions (and maybe include SQL statements into these doc comments), and also include internal comments into big functions as functions like `upsert_accounts()` are pretty difficult to follow.

crates/store/src/db/models/queries.rs
```diff
@@ -481,11 +481,20 @@ pub(crate) fn upsert_accounts(
         account_id: AccountId,
     ) -> Result<Vec<Account>, DatabaseError> {
```
bobbinth, line 482: Leaving a comment here but it is actually for the line 472 above: I'm not sure that statement is accurate. I think this function should work fine when `details` are `NULL`. Otherwise, we have a bug, I think.

bobbinth (approved):
Looks good! Thank you! I left a few comments inline - some of them are for the future.

bobbinth:
One thing would be cool to understand if this code de-duplication had a noticeable impact on the database size (as measured in our benchmarks).

crates/store/src/db/schema.rs
```diff
@@ -50,8 +50,18 @@ diesel::table! {
...
+        nonce -> Nullable<BigInt>,
         block_num -> BigInt,
-        details -> Nullable<Binary>,
+    }
+}
+
+diesel::table! {
+    account_codes (code_commitment) {
+        code_commitment -> Binary,
+        code -> Binary,
```
Mirko-von-Leipzig, line 64: More of a philosophical question, but @bobbinth why do we call it account code, and then note script, transaction script?

i.e. with less stuttering

```
    account_scripts (commitment) {
        commitment -> Binary,
        code -> Binary
    }
```

or using `masm` instead of `code`

bobbinth: In case of notes and transaction, what we have is a single executable program - so, we call these "script" (I guess "program could have worked as well, but "script" is shorter and a bit more specific).

In case of accounts, the code is a collection of procedures which note/tx scripts can call when they execute. We cold have probably called it "library" or something like that, but "account library" sounds a bit off here.

crates/store/src/db/models/queries.rs
drahnr: I can do so. The initial idea was to have a sanity check reference from the migrated SQL statements

Mirko-von-Leipzig (approved):

crates/store/src/db/migrations/2025062000000_setup/up.sql
drahnr: Re-grouped

crates/store/src/db/schema.rs
drahnr: Thank you! Fixed

bobbinth:
> If we continue doing changes, we should consider investing in benchmarks for these metrics and tracking it automatically.

The current benchmarks should cover these (e.g., [here](https://github.com/0xMiden/miden-node/tree/next/bin/stress-test#database-stats)) - but the process for updating these is manual.

Issue #1059, opened 2025-07-08
Simulate genesis bootstrap with real transactions


Mirko-von-Leipzig:
#1000 added support for a configurable genesis state, but we missed that the assets assigned to the wallets should be subtracted from the faucets available supply.

> Yes, we do need to subtract them otherwise we may end up have more assets circulating than the max_supply of the faucet. The way to do it is to add the sum of all issued assets from the number of currently issues assets stored in the first slot of the faucet account's storage.

This was then fixed as part of #1000 however we should explore making the genesis bootstrapping more robust by simulating it with real transactions as described below.

Mirko-von-Leipzig:
@PhilippGackstatter is there some way we could get proof-like guarantees to the genesis block?

I'm a bit concerned that its possible to create a genesis state that's illegal.

Maybe something dumb like dry-running this:

1. generate empty genesis
2. create faucets using transactions
3. create wallets using transactions

and if this all works then its safe to generate the final account states as the genesis state? Maybe some issues with nonces being more increases than they should be, but they could be "safely" set to 1..

PhilippGackstatter:
> * generate empty genesis
> * create faucets using transactions
> * create wallets using transactions

I think this could work and should be much safer in terms of not accidentally violating protocol rules, invariants, etc. So this would basically be a separate "bootstrap chain" to which you apply genesis transactions. After those have been executed, you take the final account state (and potentially notes?) and create the genesis block of the actual "chain proper".

That should work. I don't think you even need to set the nonces back to `1` unless you want that just so it's "clean".

Mirko-von-Leipzig:
@drahnr yes I think that makes sense. I suspect (2) will come with its own set of design decisions which we aren't really prepared to answer yet.

Mirko-von-Leipzig:
@mmagician once genesis can be bootstrapped with arbitrary accounts, is there a practical way in which we can validate that the invariants were upheld within that block itself? I suspect the answer is no?

With only basic faucet and wallet support, we could ensure this by recreating the genesis block as described in my previous comment [here](https://github.com/0xMiden/miden-node/issues/1059#issuecomment-3050277752). However with arbitrary accounts and storage, this is no longer possible.

Currently we just have some basic checks i.e. that assets stem from faucets but that doesn't really scale. On the other hand I'm unsure how we could ever vet this.

PR #1122, opened 2025-07-31
Move away from the `#[maybe_async]` usage, avoid `TransactionProvider`

drahnr:
Depends on https://github.com/0xMiden/miden-base/pull/1666

### DIFF
--- a/crates/ntx-builder/src/transaction.rs
+++ b/crates/ntx-builder/src/transaction.rs
@@ -70,43 +70,26 @@ impl NtxContext {
         tracing::Span::current().set_attribute("notes.count", notes.len());
 
         // Work-around for `TransactionExecutor` not being `Send`.
-        tokio::task::spawn_blocking(move || {
-            {
-                {
-                    let rt = tokio::runtime::Builder::new_current_thread()
-                        .enable_all()
-                        .build()
-                        .expect("runtime should be built");
+        async move {
+            let mut notes = notes
+                .into_iter()
+                .map(|note| InputNote::Unauthenticated { note: note.into() })
+                .collect::<Vec<_>>();
+            // We shuffle the notes here to prevent having a failing note always in
+            // front.
+            notes.shuffle(&mut rand::rng());
+            let notes = InputNotes::new(notes).map_err(NtxError::InputNotes)?;
 
-                    rt.block_on(
-                        async move {
-                            let mut notes = notes
-                                .into_iter()
-                                .map(|note| InputNote::Unauthenticated { note: note.into() })
-                                .collect::<Vec<_>>();
-                            // We shuffle the notes here to prevent having a failing note always in
-                            // front.
-                            notes.shuffle(&mut rand::rng());
-                            let notes = InputNotes::new(notes).map_err(NtxError::InputNotes)?;
+            let data_store = NtxDataStore::new(account, self.genesis_header.clone());
 
-                            let data_store =
-                                NtxDataStore::new(account, self.genesis_header.clone());
-
-                            self.filter_notes(&data_store, notes)
-                                .and_then(|notes| self.execute(&data_store, notes))
-                                .and_then(|tx| self.prove(tx))
-                                .and_then(|tx| self.submit(tx))
-                                .await
-                        }
-                        .in_current_span(),
-                    )
-                }
-            }
-            .in_current_span()
-        })
+            let notes = self.filter_notes(&data_store, notes).await?;
+            let executed = self.execute(&data_store, notes).await?;
+            let proven = self.prove(executed).await?;
+            self.submit(proven).await?;
+            Ok(())
+        }
+        .in_current_span()
         .await
-        .map_err(NtxError::Panic)
-        .and_then(Instrumented::into_inner)
         .inspect_err(|err| tracing::Span::current().set_error(err))
     }
 
@@ -125,7 +108,7 @@ impl NtxContext {
         data_store: &NtxDataStore,
         notes: InputNotes<InputNote>,
     ) -> NtxResult<InputNotes<InputNote>> {
-        let executor: TransactionExecutor<'_, '_, _, UnreachableAuth> =
+        let executor: TransactionExecutor<'_, '_, NtxDataStore, UnreachableAuth> =
             TransactionExecutor::new(data_store, None);
         let checker = NoteConsumptionChecker::new(&executor);
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -503,9 +503,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.29"
+version = "1.2.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c1599538de2394445747c8cf7935946e3cc27e9625f889d979bfb2aaf569362"
+checksum = "deec109607ca693028562ed836a5f1c4b8bd77755c4e132fc5ce11b0b6211ae7"
 dependencies = [
  "jobserver",
  "libc",
@@ -599,9 +599,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.41"
+version = "4.5.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "be92d32e80243a54711e5d7ce823c35c41c9d929dc4ab58e1276f625841aadf9"
+checksum = "ed87a9d530bb41a67537289bafcac159cb3ee28460e0a4571123d2a778a6a882"
 dependencies = [
  "clap_builder",
  "clap_derive 4.5.41",
@@ -609,9 +609,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.41"
+version = "4.5.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "707eab41e9622f9139419d573eca0900137718000c517d47da73045f54331c3d"
+checksum = "64f4f3f3c77c94aff3c7e9aac9a2ca1974a5adf392a8bb751e827d6d127ab966"
 dependencies = [
  "anstream",
  "anstyle",
@@ -747,9 +747,9 @@ dependencies = [
 
 [[package]]
 name = "crc32fast"
-version = "1.4.2"
+version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a97769d94ddab943e4510d138150169a2758b5ef3eb191a9ee688de3e23ef7b3"
+checksum = "9481c1c90cbf2ac953f07c8d4a58aa3945c425b7185c9154d67a65e4230da511"
 dependencies = [
  "cfg-if",
 ]
@@ -1535,9 +1535,9 @@ dependencies = [
 
 [[package]]
 name = "hyper-util"
-version = "0.1.15"
+version = "0.1.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f66d5bd4c6f02bf0542fad85d626775bab9258cf795a4256dcaf3161114d1df"
+checksum = "8d9b05277c7e8da2c93a568989bb6207bef0112e8d17df7a6eda4a3cf143bc5e"
 dependencies = [
  "base64",
  "bytes",
@@ -1551,7 +1551,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.0",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -1804,9 +1804,9 @@ checksum = "c8fae54786f62fb2918dcfae3d568594e50eb9b5c25bf04371af6fe7516452fb"
 
 [[package]]
 name = "io-uring"
-version = "0.7.8"
+version = "0.7.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b86e202f00093dcba4275d4636b93ef9dd75d025ae560d2521b45ea28ab49013"
+checksum = "d93587f37623a1a17d94ef2bc9ada592f5465fe7732084ab7beefabe5c77c0c4"
 dependencies = [
  "bitflags 2.9.1",
  "cfg-if",
@@ -1967,7 +1967,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "07033963ba89ebaf1584d767badaa2e8fcec21aedea6b8c0346d487d49c28667"
 dependencies = [
  "cfg-if",
- "windows-targets 0.53.2",
+ "windows-targets 0.53.3",
 ]
 
 [[package]]
@@ -2156,9 +2156,8 @@ dependencies = [
 
 [[package]]
 name = "miden-air"
-version = "0.16.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "77e07b5ea75948306dd6aaf341c7865a927a3fc2135bb261571435f315793bd5"
+version = "0.17.0"
+source = "git+https://github.com/0xMiden/miden-vm?branch=async-experimental#a35aa4dccef9ecc960e73e6a8b40ee824170791c"
 dependencies = [
  "miden-core",
  "thiserror 2.0.12",
@@ -2168,9 +2167,8 @@ dependencies = [
 
 [[package]]
 name = "miden-assembly"
-version = "0.16.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fd3ed10cf5894fea966161ef42e5a366779ff2ae5ea35fd45e12236115083851"
+version = "0.17.0"
+source = "git+https://github.com/0xMiden/miden-vm?branch=async-experimental#a35aa4dccef9ecc960e73e6a8b40ee824170791c"
 dependencies = [
  "log",
  "miden-assembly-syntax",
@@ -2182,9 +2180,8 @@ dependencies = [
 
 [[package]]
 name = "miden-assembly-syntax"
-version = "0.16.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "822b0c69f22346ddcaba307dbcb69d2e881fd4d074511da8f1b51c6d2a03d384"
+version = "0.17.0"
+source = "git+https://github.com/0xMiden/miden-vm?branch=async-experimental#a35aa4dccef9ecc960e73e6a8b40ee824170791c"
 dependencies = [
  "aho-corasick",
  "lalrpop",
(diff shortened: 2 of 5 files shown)


crates/proto/src/domain/block.rs
```diff
@@ -95,17 +96,9 @@ impl TryFrom<proto::blockchain::BlockHeader> for BlockHeader {
...
-        AccountId::dummy(
-            [0_u8; 15],
-            AccountIdVersion::Version0,
-            miden_objects::account::AccountType::FungibleFaucet,
-            miden_objects::account::AccountStorageMode::Network,
-        ),
-        0_u32,
-    )
-    .unwrap()
+    FeeParameters::new(AccountId::new_unchecked(<[Felt; 2]>::default()), 0_u32).unwrap()
```
igamigo: Pushed a commit that changed this, turns out `AccountId::dummy` was behind the `testing` flag on `miden-objects`

igamigo: Had to push another one because there were other occurrences

crates/ntx-builder/src/transaction.rs
```diff
@@ -75,58 +75,45 @@ pub struct NtxContext {
...
-                            let mut notes = notes
-                                .into_iter()
-                                .map(|note| InputNote::Unauthenticated { note: note.into() })
-                                .collect::<Vec<_>>();
-                            // We shuffle the notes here to prevent having a failing note always in
-                            // front.
-                            notes.shuffle(&mut rand::rng());
-                            let notes = InputNotes::new(notes).map_err(NtxError::InputNotes)?;
+        async move {
+            // Work-around for `TransactionExecutor` not being `Send`.
```
bobbinth: Is this comment still relevant?

drahnr: It doesn't hold anymore, `TransactionExecutor` does not hold `Arc<dyn _>` without further bounds anymore since some time.

crates/store/src/genesis/mod.rs
```diff
@@ -25,6 +26,7 @@ pub mod config;
 /// Represents the state at genesis, which will be used to derive the genesis block.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct GenesisState {
+    pub fee_parameters: FeeParameters,
     pub accounts: Vec<Account>,
     pub version: u32,
     pub timestamp: u32,
```
bobbinth: nit: I would reverse the order of fields - first `accounts` and then `fee_parameters`.

crates/store/src/genesis/mod.rs
```diff
@@ -45,8 +47,18 @@ impl GenesisBlock {
...
 
 impl GenesisState {
-    pub fn new(accounts: Vec<Account>, version: u32, timestamp: u32) -> Self {
-        Self { accounts, version, timestamp }
+    pub fn new(
+        fee_parameters: FeeParameters,
+        accounts: Vec<Account>,
+        version: u32,
+        timestamp: u32,
+    ) -> Self {
```
bobbinth: Same comment as above re order of parameters.

crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -1,6 +1,11 @@
...
 version   = 1
 
+
+[fee_parameters]
+symbol                = "MIDEN"
+verification_base_fee = 0
+
 [[fungible_faucet]]
 decimals     = 3
 max_supply   = 100_000_000
```
bobbinth, line 11: Let's create an issue for refactoring this to be more inline with something like what is described in https://github.com/0xMiden/miden-node/pull/1143#discussion_r2265981740.

crates/block-producer/src/test_utils/proven_tx.rs
```diff
@@ -27,6 +27,12 @@ pub struct MockProvenTxBuilder {
     output_notes: Option<Vec<OutputNote>>,
     input_notes: Option<Vec<InputNote>>,
     nullifiers: Option<Vec<Nullifier>>,
+    fee: FungibleAsset,
+}
+
+fn test_fee() -> FungibleAsset {
+    let faucet = ACCOUNT_ID_NATIVE_ASSET_FAUCET.try_into().unwrap();
+    FungibleAsset::new(faucet, 0).unwrap()
 }
```
bobbinth: We have this function in multiple places. Maybe there is a way to have just a single copy of it (i.e., in the `test-utils` create)?

bin/stress-test/src/seeding/mod.rs
```diff
@@ -396,6 +404,11 @@ fn create_consume_note_tx(
         Word::empty(),
         block_ref.block_num(),
         block_ref.commitment(),
+        FungibleAsset::new(
+            block_ref.fee_parameters().native_asset_id(),
+            u64::from(block_ref.fee_parameters().verification_base_fee()),
+        )
```
bobbinth: This is probably fine for now, but it may be better to have a helper function to create this fee (similar to how we have `test_fee()` functions elsewhere). This function could take `&FeeParameters` as an input and output `FungibleAsset`.

crates/ntx-builder/src/transaction.rs
```diff
@@ -75,58 +75,44 @@ pub struct NtxContext {
...
-                            let mut notes = notes
-                                .into_iter()
-                                .map(|note| InputNote::Unauthenticated { note: note.into() })
-                                .collect::<Vec<_>>();
-                            // We shuffle the notes here to prevent having a failing note always in
-                            // front.
-                            notes.shuffle(&mut rand::rng());
-                            let notes = InputNotes::new(notes).map_err(NtxError::InputNotes)?;
+        async move {
+            async move {
```
bobbinth, line 94: Question: why do we need nested `async move`s here?

crates/ntx-builder/src/transaction.rs
```diff
@@ -75,58 +75,44 @@ pub struct NtxContext {
...
-                            let proven = Box::pin(self.prove(executed)).await?;
-                            self.submit(proven).await?;
-                            Ok(())
-                        }
-                        .in_current_span(),
-                    )
-                }
+                let notes = Box::pin(self.filter_notes(&data_store, notes)).await?;
+                let executed = Box::pin(self.execute(&data_store, notes)).await?;
+                let proven = Box::pin(self.prove(executed)).await?;
```
bobbinth, line 108: Question: is there an alternative to these `Box::pin()` or do we have to keep using them from now on?

Makefile
Mirko-von-Leipzig, line 1: There are additional commands that should also remove the `--exclude miden-remote-prover` suffix/duplication. `make test` and `make check` I think.

bobbinth (approved):
Looks good! Thank you! I left some nits and questions inline (some of them are just a carry-over from https://github.com/0xMiden/miden-node/pull/1143)

CHANGELOG.md
```diff
@@ -18,7 +12,10 @@
 ### Changes
 
 - [BREAKING] Updated MSRV to 1.88.
+- [BREAKING] Refactor protobuf messages ([#1045](https://github.com/0xMiden/miden-node/pull/#1045)).
+- [BREAKING] De-duplicate storage of code in DB (no-migration) ([#1083](https://github.com/0xMiden/miden-node/issue/#1083)).
 - [BREAKING] RPC accept header format changed from `application/miden.vnd+grpc.<version>` to `application/vnd.miden; version=<version>` ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
+- [BREAKING] Removed `maybe_async` usage and integrated `FeeParameters` into block headers. ([#1122](https://github.com/0xMiden/miden-node/pull/1122)).
```
Mirko-von-Leipzig: The `maybe_async` is more an internal change; I don't think its visible externally.

```suggestion
- [BREAKING] Integrated `FeeParameters` into block headers. ([#1122](https://github.com/0xMiden/miden-node/pull/1122)).
```

proto/proto/types/blockchain.proto
```diff
@@ -65,6 +65,20 @@ message BlockHeader {
...
-    fixed32 timestamp = 11;
+    fixed32 timestamp = 12;
+}
+
+// FEE PARAMETERS
+// ================================================================================================
+
+// Definition of the fee parameters.
+message FeeParameters {
+    // Respective faucet `AccountId`.
```
Mirko-von-Leipzig: The faucet account ID which is used for native fee assets.

proto/proto/types/blockchain.proto
```diff
@@ -65,6 +65,20 @@ message BlockHeader {
...
+    account.AccountId native_asset_id = 1;
+    // Verification base fee.
```
Mirko-von-Leipzig: Could we expound on this definition? I don't actually know what this is from a user's perspective.

crates/store/src/genesis/config/errors.rs
```diff
@@ -42,4 +45,12 @@ pub enum GenesisConfigError {
...
     #[error("Total issuance overflowed u64")]
     IssuanceOverflow,
+    #[error("missing fee faucet for native asset {0}")]
+    MissingFeeFaucet(String),
+    #[error("fee error")]
+    FeeError(#[from] FeeError),
+    #[error("faucet account of {0} is not a fungible faucet")]
+    NativeAssetFaucetIsNotPublic(String),
+    #[error("faucet account of {0} is not public")]
+    NativeAssetFaucitIsNotAFungibleFaucet(String),
```
Mirko-von-Leipzig, line 55: I think most of these variants fall away if we make the native faucet a special section in the genesis toml file.

crates/store/src/genesis/config/samples/01-simple.toml
Mirko-von-Leipzig: #1149

crates/block-producer/src/test_utils/proven_tx.rs
Mirko-von-Leipzig: We should probably move towards replacing these mocks with the mock chain from base as well.

bin/stress-test/src/seeding/mod.rs
Mirko-von-Leipzig: It would be nicer if this was a function on `FeeParameters` itself but I suppose it wants to be more flexible than just native assets.

crates/store/src/genesis/config/mod.rs
```diff
@@ -92,8 +98,11 @@ impl GenesisConfig {
             timestamp,
             fungible_faucet: fungible_faucet_configs,
             wallet: wallet_configs,
+            fee_parameters,
         } = self;
 
+        let _ = TokenSymbol::new(&fee_parameters.symbol)?;
```
Mirko-von-Leipzig, line 104: Is the problem that this isn't implicitly checked anywhere else? Is this something that would be solved by #1059?

crates/store/src/genesis/config/mod.rs
```diff
@@ -278,6 +305,22 @@ impl GenesisConfig {
...
+
+/// Represents a the fee parameters using the given asset
+///
+/// A faucet providing the `symbol` token moste exist.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct FeeParameterConfig {
+    // TODO eventually directly parse to `TokenSymbol`
+    /// Token symbol to use for base fees.
+    symbol: String,
```
Mirko-von-Leipzig, line 319: Does `TokenSymbol` not implement deserialize? We should be able to use `serde(deserialize_with)`

crates/proto/src/domain/block.rs
```diff
@@ -173,3 +178,34 @@ impl TryFrom<proto::block_producer_store::BlockInputs> for BlockInputs {
...
+// ================================================================================================
+
+impl TryFrom<proto::blockchain::FeeParameters> for FeeParameters {
+    type Error = ConversionError;
+    fn try_from(fee_params: proto::blockchain::FeeParameters) -> Result<Self, Self::Error> {
+        let account_id = fee_params
+            .native_asset_id
+            .clone()
+            .ok_or(proto::blockchain::FeeParameters::missing_field(stringify!(native_asset_id)))?;
+        let native_asset_id = AccountId::try_from(account_id)?;
```
Mirko-von-Leipzig: Or use `take` to avoid the clone?

```suggestion
        let account_id = fee_params
            .native_asset_id
            .map(TryInto::try_into)
            .ok_or(proto::blockchain::FeeParameters::missing_field(stringify!(native_asset_id)))??;
```

crates/ntx-builder/src/transaction.rs
```diff
@@ -52,8 +52,8 @@ pub enum NtxError {
     Proving(#[source] TransactionProverError),
     #[error("failed to submit transaction")]
     Submission(#[source] tonic::Status),
-    #[error("the ntx task panic'd")]
-    Panic(#[source] JoinError),
+    #[error("failed to spawn blocking task")]
+    Spawn(#[source] JoinError),
```
Mirko-von-Leipzig: Ideally this would be rolled into `Proving` since that's where it occurs. 

The error message here is also incorrect; its not that the task failed to spawn but rather that the task itself panic'd during execution.

crates/ntx-builder/src/transaction.rs
```diff
@@ -153,23 +139,21 @@ impl NtxContext {
...
                     .collect::<Vec<InputNote>>();
-
-                if notes.is_empty() {
-                    let err = error.map_or_else(|| "None".to_string(), |err| err.as_report());
-                    tracing::warn!(%err, "all network notes failed");
+                // TODO revisit verbosity once we scale up to many more notes
+                for failed in failed {
+                    let err = failed.error.as_report();
+                    let note_id = failed.note.id();
+                    tracing::warn!(%note_id, %err, "note failed");
```
Mirko-von-Leipzig: I think the verbosity is already a problem now. A transaction will take up to 50 notes 5 times a second.

I'm not entirely sure what to do here tbh. One option would be to simply add the list of failed note IDs as a property in the span but then one loses the error information. That said, I'm unsure how valuable it is since the notes are user supplied so we should expect them to fail 🤷

crates/ntx-builder/src/transaction.rs
```diff
@@ -255,29 +239,29 @@ impl NtxDataStore {
...
-    async fn get_transaction_inputs(
+    fn get_transaction_inputs(
         &self,
         account_id: AccountId,
         ref_blocks: BTreeSet<BlockNumber>,
-    ) -> Result<(Account, Option<Word>, BlockHeader, PartialBlockchain), DataStoreError> {
-        if self.account.id() != account_id {
-            return Err(DataStoreError::AccountNotFound(account_id));
-        }
+    ) -> impl FutureMaybeSend<
```
Mirko-von-Leipzig, line 247: Ai okay so because we hack'd around the future/send problem we can no longer use `async fn` sugar?

crates/ntx-builder/src/transaction.rs
```diff
@@ -255,29 +239,29 @@ impl NtxDataStore {
...
+        Result<(Account, Option<Word>, BlockHeader, PartialBlockchain), DataStoreError>,
+    > {
+        let account = self.account.clone();
```
Mirko-von-Leipzig, line 250: Its a bummer that this is outside the check below. Is this because of the async move?

crates/ntx-builder/src/transaction.rs
drahnr: We do get clippy warnings due to future sizes, ``Box::pin` shifts pieces of it to the heap, which makes the wrapping future smaller and quiets the clippy warnings. Out future sizes exceeded 100 kBytes, which is already pretty heavy, mostly from having to carry plenty of stack residing state across await points.

crates/ntx-builder/src/transaction.rs
drahnr: Reverted

bin/stress-test/src/seeding/mod.rs
drahnr: I added the helper, if we want, we can move that into the `FeeParameters` at a later point in time

crates/block-producer/src/test_utils/proven_tx.rs
drahnr: I moved them into utils for now, to unblock the PR

crates/ntx-builder/src/transaction.rs
drahnr: Downgrading to `debug!` then

crates/ntx-builder/src/transaction.rs
drahnr: Unfortunately we cannot, because the implicit lifetime of `&self` passed does not have a representation in our `FutureMaybeSend<_>` trait.

crates/ntx-builder/src/transaction.rs
drahnr: We need the future to be `Send`, we can only do this if we guarantee `&'a` to exist as as long as `self`, but we don't want to, so cloning and moving is the cheap alternative.

crates/store/src/genesis/config/mod.rs
drahnr: I'll keep this for a follow-up, it's used in multiple places.

proto/proto/types/blockchain.proto
drahnr: From base: `    /// The base fee (in base units) capturing the cost for the verification of a transaction.` I don't fully grasp what it means for the user to supply a different value.

crates/store/src/genesis/config/mod.rs
drahnr: This would be solved by using `TokenSymbol` in conjunction with `serde(deserialize_with = "foo")` AND if `TokenSymbol` would be used in `HashMap`. Eventually I want to either use a wrapper type or implement them on the type in `base`.

Makefile
drahnr: I could have sworn I already made these exact changes, maybe a merge included them again

drahnr: Nvm, that was another repository 😶‍🌫️

drahnr: Fixed

crates/ntx-builder/src/transaction.rs
drahnr: If you use the variant with `TryFutureExt::inspect_err` rather than the `Result::inspect_err`, the instrument `proc_macro` 

```
error[E0277]: the trait bound `Result<_, _>: FutureMaybeSend<Result<(), NtxError>>` is not satisfied
  --> crates/ntx-builder/src/transaction.rs:70:10
   |
66 |     #[instrument(target = COMPONENT, name = "ntx.execute_transaction", skip_all, err)]
   |     ---------------------------------------------------------------------------------- return type was inferred to be `Result<_, _>` here
...
70 |     ) -> impl FutureMaybeSend<NtxResult<()>> {
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `futures::Future` is not implemented for `Result<_, _>`
   |
```

is unhappy, since it generates a `Result` type from it, rather than a future, which causes compile issues.

The trailing `async move {..}` keeps the heuristic happy and compiles.

PR #1126, opened 2025-08-01
refactor: queries minor cleanup

drahnr:
* Avoids excessive usage of tuple and indexing into them, but prefers types and named fields
* first step to wrapper types for all DB stored and retrieved values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,6 @@
 
 - [BREAKING] Updated MSRV to 1.88.
 - [BREAKING] Refactor protobuf messages ([#1045](https://github.com/0xMiden/miden-node/pull/#1045)).
-- [BREAKING] Consolidate to a single async interface and drop `#[maybe_async]` usage ([#1666](https://github.com/0xMiden/miden-node/pull/#1666))
-
 ### Enhancements
 
 - Added environment variable support for batch and block size CLI arguments ([#1081](https://github.com/0xMiden/miden-node/pull/1081)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - [BREAKING] Updated MSRV to 1.88.
 - [BREAKING] Refactor protobuf messages ([#1045](https://github.com/0xMiden/miden-node/pull/#1045)).
+
 ### Enhancements
 
 - Added environment variable support for batch and block size CLI arguments ([#1081](https://github.com/0xMiden/miden-node/pull/1081)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - [BREAKING] Updated MSRV to 1.88.
 - [BREAKING] Refactor protobuf messages ([#1045](https://github.com/0xMiden/miden-node/pull/#1045)).
+- [BREAKING] Consolidate to a single async interface and drop `#[maybe_async]` usage ([#1666](https://github.com/0xMiden/miden-node/pull/#1666))
+
 ### Enhancements
 
 - Added environment variable support for batch and block size CLI arguments ([#1081](https://github.com/0xMiden/miden-node/pull/1081)).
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - [BREAKING] Updated MSRV to 1.88.
 - [BREAKING] Refactor protobuf messages ([#1045](https://github.com/0xMiden/miden-node/pull/#1045)).
+- [BREAKING] Consolidate to a single async interface and drop `#[maybe_async]` usage ([#1666](https://github.com/0xMiden/miden-node/pull/#1666))
```
bobbinth: I think this is a wrong changelog entry for this PR.

crates/store/src/db/models/conv.rs
```diff
@@ -27,8 +27,71 @@
...
-use miden_node_proto::domain::account::NetworkAccountPrefix;
-use miden_objects::{Felt, note::NoteTag};
+use std::any::type_name;
+
+use miden_node_proto::domain::account::{NetworkAccountError, NetworkAccountPrefix};
+use miden_objects::{Felt, block::BlockNumber, note::NoteTag};
+
+#[derive(Debug, thiserror::Error)]
+#[error("failed to convert a database value to it's in memory type {0}")]
+pub struct DbReprConvError(&'static str);
```
bobbinth: nit: I'd prefer fewer abbreviations - maybe `DbTypeConversionError`?

crates/store/src/db/models/conv.rs
```diff
@@ -27,8 +27,71 @@
...
+
+/// Convert from and to it's database representation and back
+///
+/// We do not assume sanity of DB types.
+pub(crate) trait DbReprConv: Sized {
```
bobbinth: Similar nit as above - maybe something like `DbTypeConvert` or `SqlTypeConvert`?

sergerad (approved):
LGTM. +1 on replacing the abbreviations

crates/store/src/db/models/conv.rs
Mirko-von-Leipzig: `DatabaseType` makes sense to me, and then `DatabaseTypeError`

crates/store/src/db/models/conv.rs
```diff
@@ -27,8 +27,71 @@
...
+    type Raw: Sized;
+    type Error: std::error::Error + Send + Sync + 'static;
+    fn to_raw_sql(self) -> Self::Raw;
+    fn from_raw_sql(_raw: Self::Raw) -> ::std::result::Result<Self, Self::Error>;
```
Mirko-von-Leipzig: Can we remove the `::std::...`? I assume this is auto-complete or AI remnants

drahnr: I actually write them to avoid any ambiguity with other shortcut forms of i.e. `anyhow`

drahnr: Not necessary in this scope, removed

crates/store/src/db/models/conv.rs
drahnr: Last call, it's `DatabaseTypeConversionError` and `SqlTypeConvert`

crates/store/src/db/models/conv.rs
```diff
@@ -27,8 +27,71 @@
...
+impl SqlTypeConvert for BlockNumber {
+    type Raw = i64;
+    type Error = DatabaseTypeConversionError;
+    fn from_raw_sql(raw: Self::Raw) -> Result<Self, Self::Error> {
+        #[allow(clippy::cast_sign_loss)]
+        if raw <= u32::MAX as i64 {
+            Ok(BlockNumber::from(raw as u32))
+        } else {
+            Err(DatabaseTypeConversionError(type_name::<BlockNumber>()))
+        }
```
Mirko-von-Leipzig, line 58:
```suggestion
        u32::try_from(raw)
            .map(BlockNumber::from)
            .map_err(|_| DatabaseTypeConversionError(...))
```

Mirko-von-Leipzig:
```suggestion
        u32::try_from(raw)
            .map(BlockNumber::from)
            .map_err(|_| DatabaseTypeConversionError(...))
```

Issue #1112, opened 2025-07-25
Support submitting `ProvenBatch`


Mirko-von-Leipzig:
We'd like to allow users to submit an entire `ProvenBatch` directly. This enables transaction sets which _must_ be bundled together e.g. atomic swaps.

This will require a new RPC endpoint, some mempool improvements and some care in designing around this concept to ensure there are no untoward side-effects.

### RPC

Add a `SubmitProvenBatch` analogue to `SubmitProvenTransaction` to both RPC and block-producer gRPC schemas. 

The RPC implementation should verify the integrity of the batch before proxying it on (similar to what the `SubmitProvenTransaction` does).

### Mempool

The current implementation doesn't really allow for this because it assumes the atomic building block is a `ProvenTransaction`. 

A `ProvenBatch` contains the transaction headers, but _not_ the raw transaction data and we therefore cannot deconstruct it back to "atoms" but instead this becomes a new atomic unit of sorts.

### Open questions

1. What do we do if the submitted batch contains transactions already in the mempool?
2. What do we do if the submitted batch depends on transactions which are not already in batches? The current design operates on a DAG basis where the roots are those which are available to choose from. This should be allowed but could result in strangeness if we allow (1) as some txs could be stuck in limbo..

Its possible to brute force this and hope that we consider every edge case. Ideal would be a graph structure where this sort of property falls out naturally. This may mean consolidating the transaction and batch graphs into a more general graph - currently the batch graph is effectively an overlay graph on-top of the transaction DAG.

cc @bobbinth

bobbinth:
> 1. What do we do if the submitted batch contains transactions already in the mempool?

I think the ideal approach here would be to mark this transactions as being in a batch but not remove them from the mempool. This way, they won't be selected for any other batches, but if the proven batch gets reverted for some reason, these transactions would become available for selection again.

> 2. What do we do if the submitted batch depends on transactions which are not already in batches? The current design operates on a DAG basis where the roots are those which are available to choose from. This should be allowed but could result in strangeness if we allow (1) as some txs could be stuck in limbo..

Initially, we can probably impose a simplifying restriction that that shouldn't happen. That is, if such a batch were to arrive, we would not accept it into the mempool.

[...]

PR #1130, opened 2025-08-04
feat(rpc): `submitProvenBatch` skeleton

Mirko-von-Leipzig:
This PR adds gRPC support for user's to submit proven batches to the node.

See #1112 for more information, but as a summary this provides a batch-level analogue to submitting a transaction. This supports use-cases where multiple accounts should be updated atomically which isn't possible using txs since these operate on single accounts. Examples might be atomic swaps.

This PR _does not_ include any actual implementation for handling this, which instead will be added in separate PRs. This PR focuses on the gRPC schema (which is minor), and the pre-verification that's possible at the RPC component.

Part of #1112


crates/rpc/src/server/api.rs
Mirko-von-Leipzig, line 1: Is there any other validation we can perform on user batches at the RPC component? Proof verification isn't a thing yet..

Do I need to manually unwrap and compare the tx headers with the account deltas?

Mirko-von-Leipzig: @PhilippGackstatter maybe

PhilippGackstatter: It is on my todo list to add an issue in miden-base to validate internal consistency of batches, but haven't gotten around to it yet. For batches specifically, I think we'd want to redo the checks that `ProposedBatch::new` does as long as they're possible given that the `ProvenBatch` has only `TransactionHeader`s. Plus some additional ones, like validating that the `BatchId` derives from the `TransactionIds`, etc. Ideally, each of these error conditions should also be tested, so this takes some time to implement.

But before you add this logic here, it seems better to just do it in `ProvenBatch::new` directly (which the deserialization code uses), so we enforce these constraints at the type level.

If you want, feel free to open that issue in miden-base.

Mirko-von-Leipzig: https://github.com/0xMiden/miden-base/issues/1691

crates/proto/src/generated/block_producer.rs
```diff
@@ -16,6 +16,19 @@ pub struct SubmitProvenTransactionResponse {
...
+    /// The node's current block height.
+    #[prost(fixed32, tag = "1")]
+    pub block_height: u32,
+}
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct ProvenBatch {
+    /// Encoded using \[winter_utils::Serializable\] implementation for
+    /// \[miden_objects::transaction::proven_tx::ProvenTransaction\].
+    #[prost(bytes = "vec", tag = "1")]
+    pub encoded: ::prost::alloc::vec::Vec<u8>,
```
drahnr: Do we want an enum here to do be able to have multiple formats / encodings eventually?

drahnr (approved):
The skeleton looks good

crates/proto/src/generated/block_producer.rs
Mirko-von-Leipzig: Probably at some stage yes; but at the moment none of these are versioned so I've just kept with the standard.

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: It looks like we're not quite ready for this feature to exist publicly yet without either (see issue above)

- batch proofs, or
- training wheels

Should I instead drop this PR again until we are ready? Or simply return an error each time while we wire up the required internals?

I can still proceed with the other internally required changes in the mempool.

bobbinth: I think we can still proceed - even if this is for testing purposes now only. We just won't advertise it yet, and in the worst case (if someone tries to abuse it), we can make a patch to disable the endpoint.

proto/proto/rpc.proto
```diff
@@ -51,6 +51,18 @@ service Api {
     // Submits proven transaction to the Miden network.
     rpc SubmitProvenTransaction(transaction.ProvenTransaction) returns (block_producer.SubmitProvenTransactionResponse) {}
 
+    // Submits a proven batch to the Miden network.
```
bobbinth: nit: "Submits a proven batch of transactions to the Miden network.".

proto/proto/block_producer.proto
```diff
@@ -52,6 +64,20 @@ message SubmitProvenTransactionResponse {
...
+message SubmitProvenBatchResponse {
+    // The node's current block height.
+    fixed32 block_height = 1;
+}
+
+message ProvenBatch {
+    // Encoded using [winter_utils::Serializable] implementation for
+    // [miden_objects::transaction::proven_tx::ProvenTransaction].
+    bytes encoded = 1;
+}
```
bobbinth: nit: I would maybe name this `ProvenTransactionBatch` and move it into /types/transaction.proto` file.

bobbinth (approved):
Looks good! Thank you! I left a couple of nits inline.

PR #1131, opened 2025-08-04
chore: wait for nullifier tree to be loaded in stress test

### LINKED ISSUE #1124: Wait for Nullifier Tree to be loaded before benchmarking
In the stress tests we realized that the first request was taking too long (~46 seconds with 100k accounts in the store), and after some digging I saw that this lapse matches the times that takes for the nullifier tree to be fully loaded.

We should wait for the store component to be ready to receive requests before sending benchmarking requests. A way to achieve it is to send a status request before starting with the `sync_state`/`sync_notes`/`get_nullifiers_by_prefix`.

SantiagoPittella:
closes #1124

This PR adds a call to the `status` method of the store component before starting with the benchmark, to ensure that is ready to operate.

The benchmarks results were updated in the README.


bin/stress-test/src/store/mod.rs
```diff
@@ -281,3 +287,20 @@ async fn check_nullifiers_by_prefix(
...
+) -> Result<(), String> {
+    // Get status from the store component to confirm that it is ready.
+    let status = store_client.clone().status(()).await.unwrap().into_inner();
+
+    if status.status != "connected" {
+        return Err("Store component failed to start".to_string());
+    }
+
+    Ok(())
+}
```
Mirko-von-Leipzig: This checks, but does not wait for the store - should it be waiting for the connection?

bobbinth: If this doesn't wait for the store to load, why did the benchmarks change?

Also, separately, as discussed offline, it would probably good to add a benchmark for node startup time (probably in a separate PR).

TomasArrachea: I guess that the tonic client has an internal timeout that is being used here to wait for the server response. Maybe we can just rename this function to something like `send_warmup_request`?

SantiagoPittella: > it would probably good to add a benchmark for node startup time (probably in a separate PR).

I opened an issue to track the startup times benchmarks.

> This checks, but does not wait for the store - should it be waiting for the connection?

Added a very basic retry until the store is fully loaded, is that what you meant?

bobbinth: To clarify - the previous approach (without the loop) also worked fine, right? Basically, if the store is not loaded yet, does  the `GetStatus` request block until the store is loaded? Or does it return right away?

SantiagoPittella: Yes, without the loop works too. Since the listener is already created before initializing the state and creating the tonic::Server, the connection is received, but only after the state is loaded and the server is up and running it will be handled.

bobbinth: > Since the listener is already created before initializing the state and creating the tonic::Server, the connection is received, but only after the state is loaded and the server is up and running it will be handled.

Not for this PR, but Is this something that we'd like to change? For example, the `GetStatus` endpoint could start working right away but would just return "loading" (or something like that) status. And then once it loads, the other endpoints will start working. Or does that create unnecessary complications everywhere?

cc @Mirko-von-Leipzig

SantiagoPittella: Maybe we could delegate the load initialization to some task, and put the state behind an option or something, so we can start taking requests right away but if the state is None we reject them

Mirko-von-Leipzig: Somewhat ironically, the reason it works this way is because the components couldn't auto-retry the store connection until it was live. So instead we first open the listener and then delay responding until the service is ready.

If we have fully decoupled components then I would instead first load all the state and only then start listening.

In general client's should be robust to connection issues (retries) _and_ slow servers (request timeouts).

We can make the status more complex but I'm not sure that really helps anything except to loop and print `server not ready`.

Mirko-von-Leipzig: Kubernetes/docker use a `livez` and `readyz` to indicate health, and readiness to accept commands/requests respectively. This used to be captured in a `healthz` endpoint but they deprecated it in favor of those two. Presumably they would have used `healthz` and `readyz` if they didn't need to deprecate the original behaviour.

bobbinth:
What is left to do in this PR? Should we create any followup issues based on the https://github.com/0xMiden/miden-node/pull/1131#discussion_r2252423108? cc @Mirko-von-Leipzig and @igamigo.

PR #1132, opened 2025-08-04
feat(db): Add map values table

igamigo:
Adds a SQL table for keeping account storage map values.
There is already an existing table which is very similar (`account_storage_map_updates`), but for now I'm not touching old code util we can replace old flows completely. Although for the account deltas tables and queries we might want to delete endpoints and tables altogether.

### DIFF
--- a/crates/store/src/db/models/queries/insertions.rs
+++ b/crates/store/src/db/models/queries/insertions.rs
@@ -185,7 +185,7 @@ pub(crate) fn upsert_accounts(
 
                 for (slot_idx, slot) in account.storage().slots().iter().enumerate() {
                     match slot {
-                        StorageSlot::Value(_) => continue,
+                        StorageSlot::Value(_) => {},
                         StorageSlot::Map(storage_map) => {
                             for (key, value) in storage_map.entries() {
                                 // SAFETY: We can safely unwrap the conversion to u8 because
@@ -195,7 +195,7 @@ pub(crate) fn upsert_accounts(
                                     account_id,
                                     block_num,
                                     u8::try_from(slot_idx).unwrap(),
-                                    (*key).into(),
+                                    *key,
                                     *value,
                                 )?;
                             }


bobbinth:
@igamigo - is this PR still active? or has it been superseded by something else?

igamigo:
Failing test should be fixed with (something like) https://github.com/0xMiden/miden-node/pull/1164

Makefile
```diff
@@ -57,7 +57,7 @@ workspace-check: ## Runs a check that all packages have `lints.workspace = true`
 
 
 .PHONY: lint
-lint: typos-check format fix clippy toml workspace-check machete ## Runs all linting tasks at once (Clippy, fixing, formatting, workspace, machete)
+lint: typos-check format fix clippy toml machete ## Runs all linting tasks at once (Clippy, fixing, formatting, workspace, machete)
```
Mirko-von-Leipzig:
```suggestion
lint: typos-check format fix clippy toml machete ## Runs all linting tasks at once (Clippy, fixing, formatting, machete)
```

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -109,6 +109,18 @@ CREATE TABLE account_storage_map_updates (
...
 
+CREATE TABLE account_storage_map_values (
+    account_id      BLOB NOT NULL,
+    block_num       INTEGER NOT NULL,
+    slot            INTEGER NOT NULL,
+    key             BLOB    NOT NULL,
+    value           BLOB    NOT NULL,
+    is_latest_update   INTEGER NOT NULL,
+
+    PRIMARY KEY (account_id, block_num, slot, key, is_latest_update),
```
Mirko-von-Leipzig: Nit: otherwise we technically allow duplicate updates within a block

```suggestion
    PRIMARY KEY (account_id, block_num, slot, key),
```

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -109,6 +109,18 @@ CREATE TABLE account_storage_map_updates (
     FOREIGN KEY (account_id, block_num) REFERENCES account_deltas (account_id, block_num)
 ) WITHOUT ROWID;
 
+CREATE TABLE account_storage_map_values (
+    account_id      BLOB NOT NULL,
+    block_num       INTEGER NOT NULL,
+    slot            INTEGER NOT NULL,
+    key             BLOB    NOT NULL,
+    value           BLOB    NOT NULL,
+    is_latest_update   INTEGER NOT NULL,
```
Mirko-von-Leipzig:
```suggestion
    is_latest_update   BOOLEAN NOT NULL,
```

crates/store/src/db/schema.rs
Mirko-von-Leipzig, line 1: @drahnr dumb question, but since this is a generated file, should it be committed? Or could it just sit in `{OUT_DIR}` from `build.rs`?

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -459,6 +459,82 @@ pub(crate) fn select_slot_updates_stmt(
     Ok(results)
 }
 
+/// Select account storage map values from the DB using the given [`SqliteConnection`].
+///
+/// # Returns
+///
+/// A vector of tuples containing `(slot, key, value, is_latest_update)` for the given account and
+/// block range.
+#[cfg(test)]
```
Mirko-von-Leipzig: I assume this is a temp `#[cfg(test)]` because its _currently_ only used by tests and nothing in prod yet?

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -459,6 +459,82 @@ pub(crate) fn select_slot_updates_stmt(
     Ok(results)
 }
 
+/// Select account storage map values from the DB using the given [`SqliteConnection`].
+///
+/// # Returns
+///
+/// A vector of tuples containing `(slot, key, value, is_latest_update)` for the given account and
+/// block range.
```
Mirko-von-Leipzig: This says block range but the method seems to pick either the exact block number or all values which feels incorrect?

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -459,6 +459,82 @@ pub(crate) fn select_slot_updates_stmt(
...
+pub(crate) fn select_latest_account_storage_map_values(
+    conn: &mut SqliteConnection,
+    account_id: AccountId,
+) -> Result<Vec<(u8, Word, Word)>, DatabaseError> {
+    let results: Vec<(i32, Vec<u8>, Vec<u8>)> = SelectDsl::select(
+        schema::account_storage_map_values::table.filter(
+            schema::account_storage_map_values::account_id
+                .eq(account_id.to_bytes())
+                .and(schema::account_storage_map_values::is_latest_update.eq(1)),
+        ),
```
Mirko-von-Leipzig: I've been thinking about this some more because I feel like we're manually re-implementing an aggregate function.

Would the following not work? Because `MAX` would restrict the set to the latest `value` per `(slot, key)` pair?

```sql
SELECT slot, key, value, MAX(block_num) FROM account_storage_map_values
    WHERE account_id = ?
    GROUP BY slot, key
```

crates/store/src/db/models/queries/insertions.rs
```diff
@@ -285,6 +295,49 @@ fn build_insert_delta(account: &Account) -> Result<AccountDelta, DatabaseError>
...
+    // is_latest_update=false
+    let update_count = diesel::update(schema::account_storage_map_values::table)
+        .filter(
+            schema::account_storage_map_values::account_id
+                .eq(account_id.to_bytes())
+                .and(schema::account_storage_map_values::slot.eq(slot_to_raw_sql(slot)))
+                .and(schema::account_storage_map_values::key.eq(&key.to_bytes())),
+        )
+        .set(schema::account_storage_map_values::is_latest_update.eq(0))
+        .execute(conn)?;
```
Mirko-von-Leipzig: I think this isn't quite right because it would update _every_ existing prior value whereas it should only update the current latest one.

I do think we could get away without `is_latest_update` by using aggregate functions as per https://github.com/0xMiden/miden-node/pull/1132/files#r2284573251.

crates/store/src/db/schema.rs
drahnr: Needs to be committed, since it's generated by an external tool. The build.rs just triggers the inclusion of the SQL queries for initial setup/migrations.

crates/store/src/db/tests.rs
```diff
@@ -1196,6 +1196,71 @@ fn notes() {
...
+    let block1: BlockNumber = 1.into();
+    let block2: BlockNumber = 2.into();
+    create_block(conn, block1);
+    create_block(conn, block2);
+
+    let account_id =
+        AccountId::try_from(ACCOUNT_ID_REGULAR_PRIVATE_ACCOUNT_UPDATABLE_CODE).unwrap();
+
+    let slot = 3u8;
+    let key1 = Word::from([Felt::new(1), Felt::new(2), Felt::new(3), Felt::new(4)]);
```
drahnr: nit: do we want to create a helper function for this?

crates/store/src/db/schema.rs
```diff
@@ -45,6 +45,17 @@ diesel::table! {
...
 }
 
+diesel::table! {
+    account_storage_map_values (account_id, block_num, slot, key, is_latest_update) {
+        account_id -> Binary,
+        block_num -> BigInt,
+        slot -> Integer,
+        key -> Binary,
+        value -> Binary,
+        is_latest_update -> Integer,
```
drahnr:
```suggestion
        is_latest_update -> Boolean,
```

crates/store/src/db/schema.rs
```diff
@@ -45,6 +45,17 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    account_storage_map_values (account_id, block_num, slot, key, is_latest_update) {
+        account_id -> Binary,
+        block_num -> BigInt,
+        slot -> Integer,
+        key -> Binary,
+        value -> Binary,
```
drahnr, line 8: 👍

crates/store/src/db/tests.rs
```diff
@@ -1196,6 +1196,71 @@ fn notes() {
...
+    let v1 = queries::select_account_storage_map_values(conn, account_id, Some(block1)).unwrap();
+    assert_eq!(v1.len(), 2, "expect 2 initial rows");
+    assert!(
+        v1.iter().all(|(s, _, _, is_latest)| *s == slot && *is_latest),
+        "initial rows should be latest"
+    );
+
+    let mut map2 = StorageMapDelta::default();
+    map2.insert(key1, value3);
+    let maps2: BTreeMap<_, _> = [(slot, map2)].into_iter().collect();
```
drahnr:
```suggestion
    let maps2 = BTreeMap::from_iter([(slot, map2)]);
```

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -109,6 +109,18 @@ CREATE TABLE account_storage_map_updates (
...
+
+    PRIMARY KEY (account_id, block_num, slot, key, is_latest_update),
+    FOREIGN KEY (account_id, block_num) REFERENCES account_deltas (account_id, block_num)
```
drahnr: We should add constraints for the value ranges of i.e. `slot`.

crates/store/src/db/tests.rs
igamigo: Searching for alternatives to this I saw in another test that `From<[u32;4]>` is implemented for `Word`, so I'm switching to that instead since it is more convenient.

crates/store/src/db/models/queries/accounts.rs
igamigo: Yes, basically an alternative to `#[allow(dead_code)]`

crates/store/src/db/models/queries/accounts.rs
igamigo: I thought of this, but mainly discarded it because unless I'm missing something I don't think it would be easy to select old rows for the clean up process (and also making sure we keep the latest value for each key that is older than the block history we want to keep).

Also, in your example we would be missing an aggregate for `value`. I think this would imply either joining back to the table from the query, or something like [`DISTINCT ON`](https://www.geeksforgeeks.org/postgresql/postgresql-distinct-on-expression/) (but AFAIK this is not doable on sqlite, only postgres)

crates/store/src/db/models/queries/accounts.rs
igamigo: Updated this

crates/store/src/db/migrations/2025062000000_setup/up.sql
igamigo: I think when I started this other tables were using `INTEGER` for bools, but seems something changed recently, so updated the table.

crates/store/src/db/models/queries/accounts.rs
Mirko-von-Leipzig: As a minor nit, I prefer the latter because the lint allows a `reason = "xyz"` (not required for this PR)

igamigo:
I've removed the select query and test from this PR and added it in the followup (#1140)

crates/store/src/db/schema.rs
```diff
@@ -45,6 +45,17 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    account_storage_map_values (account_id, block_num, slot, key, is_latest_update) {
+        account_id -> Binary,
+        block_num -> BigInt,
+        slot -> Integer,
```
bobbinth, line 6: Not related to this PR but just wanted to understand: what's the range of `BigInt` vs. `Integer`? Like `block_num` is 32 bits, but i'm guessing `Integer` is not sufficient for this?

crates/store/src/db/models/queries/insertions.rs
```diff
@@ -285,6 +295,50 @@ fn build_insert_delta(account: &Account) -> Result<AccountDelta, DatabaseError>
...
+    let insert_count = diesel::insert_into(schema::account_storage_map_values::table)
+        .values(&[(
+            schema::account_storage_map_values::account_id.eq(account_id.to_bytes()),
+            schema::account_storage_map_values::block_num.eq(block_num.to_raw_sql()),
+            schema::account_storage_map_values::slot.eq(slot_to_raw_sql(slot)),
+            schema::account_storage_map_values::key.eq(key.to_bytes()),
+            schema::account_storage_map_values::value.eq(value.to_bytes()),
+            schema::account_storage_map_values::is_latest_update.eq(true),
+        )])
+        .execute(conn)?;
```
bobbinth: Not for this PR: I think this works fine because we are working with an in-memory database, but if we were working with something like Postgres running on a separate machine, we'd probably want to send both queries together to the server. And probably send a batch of updates to minimize the number of network requests.

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -109,6 +109,19 @@ CREATE TABLE account_storage_map_updates (
...
 
+CREATE TABLE account_storage_map_values (
+    account_id          BLOB NOT NULL,
+    block_num           INTEGER NOT NULL,
+    slot                INTEGER NOT NULL,
+    key                 BLOB    NOT NULL,
+    value               BLOB    NOT NULL,
+    is_latest_update    BOOLEAN NOT NULL,
+
+    PRIMARY KEY (account_id, block_num, slot, key),
```
bobbinth, line 89: This creates an index on `(account_id, block_num, slot, key)` - right? But I think we also need an index on `(account_id, block_num)` (for the select query).

bobbinth (approved):

crates/store/src/db/models/queries/accounts.rs
Mirko-von-Leipzig: > Also, in your example we would be missing an aggregate for value.

I think you're misunderstanding how the aggregate works. `SELECT x, y, MAX(z)` returns only the _row_ for which `z` is the maximum -- so `x,y,z` would all be from the same entry/record. Or maybe I'm misunderstanding why we would need an additional aggregate for value.

As an example

```sql
create table if not EXISTS test(
    k integer not null,
    v integer not NULL,
    x integer not null
);

insert into test (k,v,x) values (1,1,1), (1,2,2), (2,3,3), (2,4,4);

SELECt k,v,MAX(x) from test GROUP by k;
```

returns

```
k  v  x
--------
1  2  2
2  4  4
```

Mirko-von-Leipzig: > I don't think it would be easy to select old rows for the clean up process (and also making sure we keep the latest value for each key that is older than the block history we want to keep).

It would be less easy that's true. I wonder if one can't (ab)use the aggregate in a similar way; but one would definitely need an additional `where x in (select ...)` to check this.

bobbinth: > ```sql
> SELECt k,v,MAX(x) from test GROUP BY k;
> ```

It's been a while since I seriously programmed by SQL - but I'm not sure this is valid: if we use `GROUP BY` we need to apply an aggregate function to all column which are not a part of the `GROUP BY` clause. So, it would have to be something like:

```sql
SELECt k, v, MAX(x) from test GROUP BY k, v;
```

But I don't think this would produce the desired dataset. I think the aggregate functions are applied to each column individually, rather than across rows.

Mirko-von-Leipzig: I might be wrong but here's how I think it works.

The way I think of it is that `GROUP BY` generates a unique mini-table for every unique combination of columns provided. In my example with `GROUP BY k` this would be two tables for `k=1` and `k=2`, each with two records.

The `MAX(x)` aggregate function selects the record where `x` is the maximum. If we just said `SELECT k,v,MAX(x) FROM test` we would get `2,4,4`. If there was an additional `(1,1,5)` record we would get `1,1,5`.

When combined with `GROUP BY`, this is repeated for each mini-table generated by it. And since we want the latest value for each key this becomes the example I gave. For our use case we want the latest record for every `slot+key` combination so it should be as per my [original comment](https://github.com/0xMiden/miden-node/pull/1132#discussion_r2284573251).

[Playground link](https://sqliteonline.com/#sqltext=%23url-sqlite%3Ddb-sqlite%0D%0A%23tab-name%3DSQLite%0D%0Acreate%20table%20if%20not%20EXISTS%20test(%0A%20%20%20%20k%20integer%20not%20null%2C%0A%20%20%20%20v%20integer%20not%20NULL%2C%0A%20%20%20%20x%20integer%20not%20null%0A)%3B%0A%0Ainsert%20into%20test%20(k%2Cv%2Cx)%20values%20(1%2C1%2C1)%2C%20(1%2C2%2C2)%2C%20(2%2C3%2C3)%2C%20(2%2C4%2C4)%3B%0A%0ASELECt%20k%2Cv%2CMAX(x)%20from%20test%20GROUP%20by%20k%3B) to the example I gave above.

igamigo: PostgreSQL (and probably other engines as well) reject the same SQL due to the reason we were stating: 

```
PostgreSQL
Error 42803 column "test.v" must appear in the GROUP BY clause or be used in an aggregate function
```

If some engines don't support this I'd assume it's nondeterministic. And probably in this case, because we know there should be only one value for a specific block/slot/account, it could probably work as long as we maintain SQLite as the backend.

This script works on Postgres: 

```sql
SELECT DISTINCT ON (k) k, v, x
FROM test
ORDER BY k, x DESC;
```

so does this:

```sql
SELECT t.k, t.v, t.x
FROM test AS t
JOIN (
  SELECT k, MAX(x) AS x
  FROM test
  GROUP BY k
) AS m
ON t.k = m.k AND t.x = m.x;
```

The latter also works on SQLite. But I'm thinking this complicates more than it simplifies, and possibly performs worse as well.

EDIT: To be clear, I also did not know this was possible at all in SQLite. I would have guessed it would error.

bobbinth: > [Playground link](https://sqliteonline.com/#sqltext=%23url-sqlite%3Ddb-sqlite%0D%0A%23tab-name%3DSQLite%0D%0Acreate%20table%20if%20not%20EXISTS%20test(%0A%20%20%20%20k%20integer%20not%20null%2C%0A%20%20%20%20v%20integer%20not%20NULL%2C%0A%20%20%20%20x%20integer%20not%20null%0A)%3B%0A%0Ainsert%20into%20test%20(k%2Cv%2Cx)%20values%20(1%2C1%2C1)%2C%20(1%2C2%2C2)%2C%20(2%2C3%2C3)%2C%20(2%2C4%2C4)%3B%0A%0ASELECt%20k%2Cv%2CMAX(x)%20from%20test%20GROUP%20by%20k%3B) to the example I gave above.

oh wow - I didn't realize SQLite worked like this. But as far as I know (and as @igamigo mentioned), this is a non-standard behavior (i.e., it shouldn't work in standard SQL). So, I'd be hesitant to rely on it.

crates/store/src/db/schema.rs
igamigo: I think you essentially `i32` for `Integer` and `i64` for `BigInt`:

```rust

/// The integer SQL type.
///
/// ### [`ToSql`](crate::serialize::ToSql) impls
///
/// - [`i32`][i32]
///
/// ### [`FromSql`](crate::deserialize::FromSql) impls
///
/// - [`i32`][i32]
///
/// [i32]: https://doc.rust-lang.org/nightly/std/primitive.i32.html
#[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
#[diesel(postgres_type(oid = 23, array_oid = 1007))]
#[diesel(sqlite_type(name = "Integer"))]
#[diesel(mysql_type(name = "Long"))]
pub struct Integer;


/// The big integer SQL type.
///
/// ### [`ToSql`](crate::serialize::ToSql) impls
///
/// - [`i64`][i64]
///
/// ### [`FromSql`](crate::deserialize::FromSql) impls
///
/// - [`i64`][i64]
///
/// [i64]: https://doc.rust-lang.org/nightly/std/primitive.i64.html
#[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
#[diesel(postgres_type(oid = 20, array_oid = 1016))]
#[diesel(sqlite_type(name = "Long"))]
#[diesel(mysql_type(name = "LongLong"))]
pub struct BigInt;
```

igamigo: I mainly went with these since it seems to be the convention in other tables, but I think we could actually use `i32` for block numbers and cast them appropriately (when writing, `block_num.as_u32() as i32` and `BlockNumber::from(block_num as u32)` when reading). Unsure whether this was previously discussed somewhere else

crates/store/src/db/models/queries/insertions.rs
igamigo: I meant to add the change here, but I had modified this to use a single transaction in the follow up: https://github.com/0xMiden/miden-node/pull/1140/files#diff-1a7b5569fe361ba6496af6cef83594571e7e29590f0ced96fbeadfaf51298146R320

crates/store/src/db/migrations/2025062000000_setup/up.sql
igamigo: I created an index on #1140 to account for the select, but I think we might need to add `slot` and `key` in there as well because we want to keep ordering between calls to account for the `LIMIT` cutoff. 

EDIT: Never mind this comment, we just want to order by block number so it should be fine

crates/store/src/db/schema.rs
drahnr: [It was, the issue is with sorting a signed vs unsigned column at some point in the distant future.](https://midengroup.slack.com/archives/C08UTTDUMAS/p1749222553944629) As per usual, trade-offs

crates/store/src/db/models/queries/insertions.rs
drahnr: We should not require that, since - with a few exceptions - we use the async-sync wrapper function that also wraps the `queries::*` call in a transaction. Any nested transactions won't have an effect.

crates/store/src/db/models/queries/accounts.rs
Mirko-von-Leipzig: Ah :D I see. I've only ever done this in SQLite so I just assumed :sweat:.

Let's just go ahead with simpler and then we'll revisit if we hit performance issues. Quite a few ways to approach this.

crates/store/src/db/schema.rs
igamigo: I don't have access to that conversation but makes sense!

PR #1133, opened 2025-08-04
feat: add timeout to servers

### LINKED ISSUE #1125: Add timeouts in component servers
After watching some requests in the benchmark taking almost a full minute, we should add a timeout in the server side on all requests. This can be achieved by using the `timeout` method on the `Server` builder from tonic. For example:

```rust
    .timeout(Duration::from_secs(10)) // Limits request to 10 seconds.
    .accept_http1(true)
    .layer(TraceLayer::new_for_grpc().make_span_with(traced_span_fn(TracedComponent::Rpc)))
    .layer(AcceptHeaderLayer::new(&rpc_version, genesis.commitment()))
```

SantiagoPittella:
closes #1125

### DIFF
--- a/crates/block-producer/src/server/mod.rs
+++ b/crates/block-producer/src/server/mod.rs
@@ -11,6 +11,7 @@ use miden_node_proto::generated::{self as proto};
 use miden_node_proto_build::block_producer_api_descriptor;
 use miden_node_utils::formatting::{format_input_notes, format_output_notes};
 use miden_node_utils::tracing::grpc::{TracedComponent, traced_span_fn};
+use miden_objects::batch::ProvenBatch;
 use miden_objects::block::BlockNumber;
 use miden_objects::transaction::ProvenTransaction;
 use miden_objects::utils::serde::Deserializable;
@@ -25,7 +26,13 @@ use url::Url;
 use crate::batch_builder::BatchBuilder;
 use crate::block_builder::BlockBuilder;
 use crate::domain::transaction::AuthenticatedTransaction;
-use crate::errors::{AddTransactionError, BlockProducerError, StoreError, VerifyTxError};
+use crate::errors::{
+    AddTransactionError,
+    BlockProducerError,
+    StoreError,
+    SubmitProvenBatchError,
+    VerifyTxError,
+};
 use crate::mempool::{BatchBudget, BlockBudget, Mempool, SharedMempool};
 use crate::store::StoreClient;
 use crate::{
@@ -234,6 +241,17 @@ impl api_server::Api for BlockProducerRpcServer {
              .map_err(Into::into)
     }
 
+    async fn submit_proven_batch(
+        &self,
+        request: tonic::Request<proto::transaction::ProvenTransactionBatch>,
+    ) -> Result<tonic::Response<proto::block_producer::SubmitProvenBatchResponse>, Status> {
+        self.submit_proven_batch(request.into_inner())
+             .await
+             .map(tonic::Response::new)
+             // This Status::from mapping takes care of hiding internal errors.
+             .map_err(Into::into)
+    }
+
     #[instrument(
          target = COMPONENT,
          name = "block_producer.server.status",
@@ -368,5 +386,21 @@ impl BlockProducerRpcServer {
                 block_height: block_height.as_u32(),
             }
         })
+    }
+
+    #[instrument(
+         target = COMPONENT,
+         name = "block_producer.server.submit_proven_batch",
+         skip_all,
+         err
+     )]
+    async fn submit_proven_batch(
+        &self,
+        request: proto::transaction::ProvenTransactionBatch,
+    ) -> Result<proto::block_producer::SubmitProvenBatchResponse, SubmitProvenBatchError> {
+        let _batch = ProvenBatch::read_from_bytes(&request.encoded)
+            .map_err(SubmitProvenBatchError::Deserialization)?;
+
+        todo!();
     }
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ url                = { features = ["serde"], version = "2.5" }
 pedantic = { level = "warn", priority = -1 }
 
 cast_possible_truncation     = "allow" # Overly many instances especially regarding indices.
+collapsible-if               = "allow" # Too new to enforce.
 from_iter_instead_of_collect = "allow" # at times `FromIter` is much more readable
 ignored_unit_patterns        = "allow" # Stylistic choice.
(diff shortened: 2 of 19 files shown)


bin/node/src/commands/bundled.rs
```diff
@@ -60,6 +61,15 @@ pub enum BundledCommand {
...
         enable_otel: bool,
+
+        /// Timeout of the requests.
+        #[arg(
+            long = "timeout",
+            default_value = &duration_to_human_readable_string(DEFAULT_TIMEOUT),
+            value_parser = humantime::parse_duration,
+            value_name = "DURATION"
+        )]
+        timeout: Duration,
```
Mirko-von-Leipzig: We need to somehow communicate that this is specifically for the gRPC server(s). There are many operations that a generic timeout could apply to within the node context. I don't really have a good suggestion though.


```suggestion
        /// Maximum duration a gRPC request is allocated before being dropped by the server.
        ///
        /// This may occur if the server is overloaded or due to an internal bug.
        #[arg(
            long = "grpc.timeout",
            default_value = &duration_to_human_readable_string(DEFAULT_TIMEOUT),
            value_parser = humantime::parse_duration,
            value_name = "DURATION"
        )]
        grpc_timeout: Duration,
```

Mirko-von-Leipzig: Something similar in the others.

bin/stress-test/src/seeding/mod.rs
```diff
@@ -512,6 +512,7 @@ pub async fn start_store(
             ntx_builder_listener,
             block_producer_listener,
             data_directory: dir,
+            timeout: Duration::from_secs(30),
```
Mirko-von-Leipzig: Was this just arbitrarily chosen? Or why is it different to the `10s` defaults.

crates/block-producer/src/server/mod.rs
```diff
@@ -63,6 +63,9 @@ pub struct BlockProducer {
     /// The block-producers gRPC endpoint will be available before this point, so this lets the
     /// mempool synchronize its event stream without risking a race condition.
     pub production_checkpoint: Arc<Barrier>,
+    /// Maximum time allowed for a request to complete. Once exceeded, the request is
+    /// aborted.
+    pub timeout: Duration,
```
Mirko-von-Leipzig: I would also be more specific here. Maybe you can think of a better name.


```suggestion
    /// Maximum time allowed for a request to complete. Once exceeded, the request is
    /// aborted.
    pub server_timeout: Duration,
```

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - A transaction's account delta is now checked against its commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
 - Added check for Account Id prefix uniqueness when transactions to create accounts are submitted to the mempool ([#1094](https://github.com/0xMiden/miden-node/pull/1094)).
 - RPC accept header now supports specifying the genesis commitment in addition to the RPC version. This lets clients ensure they are on the right network ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
+- Added an optional `timeout` parameter to components ([#1133](https://github.com/0xMiden/miden-node/pull/1133))
```
```suggestion
- gRPC server timeout is now configurable (defaults to `10s`) ([#1133](https://github.com/0xMiden/miden-node/pull/1133))
```

bin/stress-test/src/seeding/mod.rs
SantiagoPittella: On monday's call Bobbin mentioned that 10s was probably a good default. Though it wasn't something definitive, wdyt?

crates/block-producer/src/server/mod.rs
SantiagoPittella: What about using grpc_timeout too?

Mirko-von-Leipzig: The issue is that we have servers _and_ clients for gRPC, so its difficult ito naming.

For example, even the comment is ambiguous and its unclear whether its referring to the gRPC server's incoming requests, or the client's outgoing requests.

Mirko-von-Leipzig: But I think this is probably fine for now, maybe just update the comment to indicate its for the server side. 

I was thinking maybe one should somehow consolidate things by having 

```rust
struct Server {
    url: Url,
    timeout: Duration,
}
```

but I think at this point the desired structure is unclear and we shouldn't worry too much.

bin/stress-test/src/seeding/mod.rs
Mirko-von-Leipzig: Either is fine; was just wondering if maybe `30s` was the maximum you'd seen in the stress tests.

crates/block-producer/src/server/mod.rs
bobbinth: Just to make sure I understand, the `timeout` here is on the server side, right? i.e., if the server cannot fulfill the request in some period of time, it'll abort.

Or is it on the client side? i.e., if the client doesn't get a response in some period of time, it'll stop waiting for it.

Would be good to clarify this in the comments.

Mirko-von-Leipzig: Its on the server side only - so its a defensive measure.

In theory we _should_ also be doing something similar on the client side but given things are internal mostly the server side should cover it. Something to consider is adding it to the prover calls.

SantiagoPittella: Regarding this approach:

```rust
struct Server {
    url: Url,
    timeout: Duration,
}
```

If we use it to share it between different commands, we will lose the different env vars for the Url fields:

```rust
const ENV_BLOCK_PRODUCER_URL: &str = "MIDEN_NODE_BLOCK_PRODUCER_URL";
const ENV_BATCH_PROVER_URL: &str = "MIDEN_NODE_BATCH_PROVER_URL";
const ENV_BLOCK_PROVER_URL: &str = "MIDEN_NODE_BLOCK_PROVER_URL";
const ENV_NTX_PROVER_URL: &str = "MIDEN_NODE_NTX_PROVER_URL";
const ENV_RPC_URL: &str = "MIDEN_NODE_RPC_URL";
```

Not quite sure how can be solve this.

SantiagoPittella: Regarding the ambiguity, for the commands it might be a problem. But for the components structs it is less ambiguous since the qualification is more obvious: `server::Store` , `server::BlockProducer` .

Mirko-von-Leipzig: I think just making the comment more specific is fine for now

bobbinth (approved):
Looks good! Thank you!

PR #1137, opened 2025-08-06
fix: Flaky test_prove_transaction

### LINKED ISSUE #1102: Address test_prove_transaction flakiness
Address the test issues mentioned by this comment. There also appears to be a flakiness around the concurrency check mentioned in the comment thread.

https://github.com/0xMiden/miden-node/pull/1094#discussion_r2218381503

>Btw. the note_1 that gets passed to extend_input_notes is completely ignored and is not part of the transaction. So the test was always executing with 0 input notes.
>
>There is 1) a mock_chain.prove_next_block() missing, to add this note to the chain state and 2) the note ID should be passed directly to build_tx_context, otherwise it is ignored. (The call to extend_input_notes should be removed).
>
>A better way to do 1) would be to add this note directly using the MockChainBuilder further above, so it is directly part of the genesis state.
>
>As for the concurrency issue, I don't really understand why two proving requests are made in the first place and why one of them would be expected to fail and the other succeed.

sergerad:
## Context

Closes #1102.

Sometimes the `test_prove_transaction` fixture will fail locally on line `330`:
```rust
        // Check the failure response
        assert!(response_1.is_err() || response_2.is_err());
```

## Changes
- Fix failing test by adding tokio join of relevant tasks.
- Use `MockBuilder` to add note for execution.

### DIFF
--- a/bin/remote-prover/src/api/prover.rs
+++ b/bin/remote-prover/src/api/prover.rs
@@ -279,14 +279,13 @@ mod test {
         // Create a mock transaction to send to the server
         let mut mock_chain_builder = MockChainBuilder::new();
         let account = mock_chain_builder.add_existing_wallet(Auth::BasicAuth).unwrap();
-        let mut mock_chain = mock_chain_builder.build().unwrap();
 
         let fungible_asset_1: Asset =
             FungibleAsset::new(ACCOUNT_ID_PUBLIC_FUNGIBLE_FAUCET.try_into().unwrap(), 100)
                 .unwrap()
                 .into();
-        let note_1 = mock_chain
-            .add_pending_p2id_note(
+        let note_1 = mock_chain_builder
+            .add_p2id_note(
                 ACCOUNT_ID_SENDER.try_into().unwrap(),
                 account.id(),
                 &[fungible_asset_1],
@@ -294,10 +293,11 @@ mod test {
             )
             .unwrap();
 
+        let mock_chain = mock_chain_builder.build().unwrap();
+
         let tx_context = mock_chain
-            .build_tx_context(account.id(), &[], &[])
+            .build_tx_context(account.id(), &[note_1.id()], &[])
             .unwrap()
-            .extend_input_notes(vec![note_1])
             .build()
             .unwrap();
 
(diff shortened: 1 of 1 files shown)


bin/remote-prover/src/api/prover.rs
```diff
@@ -277,31 +277,32 @@ mod test {
```
sergerad: @tomyrd do you have any idea as to why the test would fail sometimes on line 330?
```rust
        // Check the failure response
        assert!(response_1.is_err() || response_2.is_err());
```

I presume its due to parallelism. Is there any way we could make it deterministic?

tomyrd: I think this test may be trying to send two transactions at the same time to get one proven and another one declined because the prover is busy with the first one ([this check](https://github.com/0xMiden/miden-node/blob/d54b2c2efa208484a57e560e35a7541c095e8252/bin/remote-prover/src/api/prover.rs#L169-L170)). If the first transaction gets proven before the second task is spawned maybe both get proven, but it seems rather improbable. Does this test fail often?

cc'ing @SantiagoPittella who might know more about the remote prover itself.

sergerad: It never fails in CI so its not super important. Fails locally for me but usually only on the first run after I rebuild it.

sergerad: Sorry I think @TomasArrachea might have context too. It fails on my machine every time atm.

SantiagoPittella: Sorry for the late response. I've been checking this, and after a couple tests realized that with `tokio::join!` the error ceased:

```
-        let (t1, t2) = (
-            tokio::spawn(async move { client.prove(request_1).await }),
-            tokio::spawn(async move { client_2.prove(request_2).await }),
-        );
-
-        let (response_1, response_2) = (t1.await.unwrap(), t2.await.unwrap());
+        let (response_1, response_2) = tokio::join!(client.prove(request_1), client_2.prove(request_2));

```

sergerad: TY!

SantiagoPittella (approved):
awesome! thanks!

PR #1140, opened 2025-08-06
feat: `SyncStorageMaps` RPC endpoints

### LINKED ISSUE #1095: Implement storage map sync endpoint
We should add an endpoint which would allow the user to sync a storage map of a given account. The endpoint could be named `SyncStorageMap` and its request/response messages could look like so:

```proto
message SyncStorageMapRequest {
    // block number to start syncing from
    fixed32 block_num = 1;
    
    // Account for which we want to sync storage maps
    AccountId account_id = 2;

    // Indexes of storage slots for which to return the data
    repeated uint32 storage_slots = 3;
}

message SyncStorageMapResponse {
    fixed32 chain_tip = 1;
    repeated StorageMapUpdate updates = 2;
}

message StorageMapUpdate {
    uint32 slot_idx = 1;
    Digest key = 2;
    Digest value = 3;

    // Block number at which the above key was set to the above value.
    fixed32 block_num = 4;
}
```

The behavior of this endpoint should be such that we return all updates that happened close to the current chain tip. "Close" still needs to be defined, but we could probably start with something like within 100 blocks of the chain tip.

This endpoint would exist on both Store and RPC components. The internal implementation is not fully defined yet but there is a discussion about it in https://github.com/0xMiden/miden-node/discussions/722#discussioncomment-13749307. We will probably need to have at least `storage_map_updates` table in the database, and maybe an additional table - depending on design.

### Open questions

- What's the best design for this on the database side?
- How can we limit the size of the responses without introducing too much overhead?

igamigo:
Closes #1095 
Implements endpoints and queries for storage maps sync requests from users.

What's missing is pruning the table, but I'm leaving that for a separate set of changes.

### DIFF
--- a/docs/src/user/rpc.md
+++ b/docs/src/user/rpc.md
@@ -79,3 +79,11 @@ The low part of note tags are redacted to preserve some degree of privacy. Retur
 ## Status
 
 Request the status of the node components. The response contains the current version of the RPC component and the connection status of the other components, including their versions and the number of the most recent block in the chain (chain tip).
+
+## SyncStorageMaps
+
+Returns storage map synchronization data for a specified public account within a given block range. This method allows clients to efficiently sync the storage map state of an account by retrieving only the changes that occurred between two blocks.
+
+Caller specifies the `account_id` of the public account and the block range (`block_from`, `block_to`) for which to retrieve storage updates. The response includes all storage map key-value updates that occurred within that range, along with the last block included in the sync and the current chain tip.
+
+This endpoint enables clients to maintain an updated view of account storage.


CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
 - A transaction's account delta is now checked against its commitments in `SubmitProvenTransaction` endpoint ([#1093](https://github.com/0xMiden/miden-node/pull/1093)).
 - Added check for Account Id prefix uniqueness when transactions to create accounts are submitted to the mempool ([#1094](https://github.com/0xMiden/miden-node/pull/1094)).
 - RPC accept header now supports specifying the genesis commitment in addition to the RPC version. This lets clients ensure they are on the right network ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
-- Added tables for storing account storage maps ([#1132](https://github.com/0xMiden/miden-node/pull/1132)).
+- Added RPC endpoints for retrieving account storage maps ([#1140](https://github.com/0xMiden/miden-node/pull/1140), [#1132](https://github.com/0xMiden/miden-node/pull/1132)).
```
```suggestion
- Added gRPC method `SyncStorageMap` for retrieving account storage maps ([#1140](https://github.com/0xMiden/miden-node/pull/1140), [#1132](https://github.com/0xMiden/miden-node/pull/1132)).
```

proto/proto/store/rpc.proto
```diff
@@ -317,3 +320,54 @@ message SyncNotesResponse {
...
+
+    // Individual storage slot key-value pair.
+    message StorageSlot {
+        // The storage map key.
+        primitives.Digest key = 1;
+        
+        // The storage map value.
+        primitives.Digest value = 2;
+    }
+}
```
Mirko-von-Leipzig: Maybe its a storage map that just happens to be used as updates 🤷 

```suggestion
// Represents storage map updates for a specific slot.
message StorageMap {
    // Slot index ([0..255]).
    uint32 slot_index = 1;
    
    // Storage map key-value pairs.
    repeated Entry entries = 2;

    // A key-value entry in a storage map.
    message Entry {
        primitives.Digest key = 1;
        primitives.Digest value = 2;
    }
}
```

proto/proto/store/rpc.proto
```diff
@@ -317,3 +320,54 @@ message SyncNotesResponse {
...
+    account.AccountId account_id = 2;
+
+    // Storage slots for which to return the data.
+    repeated StorageMapPageRequest storage_slot_maps = 3;
+}
+
+// Represents a storage map page request for a specific slot.
+message StorageMapPageRequest {
+    // Block that corresponds to the last time the client synced this index.
+    fixed32 block_from = 1;
```
Mirko-von-Leipzig: Do we need the fidelity to specify this per map? Or could it just be entire request?

The comment indicates that its the last sync point, but its probably the last sync point+1? Otherwise it won't be possible to sync from genesis i.e. `0`?

```suggestion
message StorageMapPageRequest {
    // Block to start sync'ing updates from.
    fixed32 block_from = 1;
```

proto/proto/store/rpc.proto
```diff
@@ -317,3 +320,54 @@ message SyncNotesResponse {
...
+// SYNC STORAGE MAP
+// ================================================================================================
+
+// Storage map synchronization request.
+//
+// Allows clients to sync storage map values for specific public accounts and slots within a block range,
+// with support for pagination to handle large storage maps.
+message SyncStorageMapRequest {
+    // Block number up to which to sync. If not specified, syncs up to latest values.
+    optional fixed32 block_to = 1;
```
Mirko-von-Leipzig: If this is optional then the response _must_ also return the actual `block_to` otherwise the client can't know the sync point.

crates/store/src/server/rpc_api.rs
```diff
@@ -358,6 +358,58 @@ impl rpc_server::Rpc for StoreApi {
...
+
+        for slot_request in request.storage_slot_maps {
+            let slot_index = slot_request.slot_index as u8;
+            let block_from = slot_request.block_from.into();
+
+            let values = self
+                .state
+                .get_storage_map_sync_values(account_id, slot_index, block_from, block_to)
+                .await
+                .map_err(internal_error)?;
```
Mirko-von-Leipzig: Each of these calls is a new database transaction. This may need to be pushed down into the `state`/`db` call itself.

crates/store/src/db/models/queries.rs
```diff
@@ -545,6 +545,61 @@ pub(crate) fn select_latest_account_storage_map_values(
...
+    block_from: BlockNumber,
+    block_to: Option<BlockNumber>,
+) -> Result<Vec<(BlockNumber, u8, Word, Word)>, DatabaseError> {
+    let mut query = schema::account_storage_map_values::table
+        .filter(
+            schema::account_storage_map_values::account_id
+                .eq(account_id.to_bytes())
+                .and(schema::account_storage_map_values::slot.eq(slot_to_raw_sql(slot)))
+                .and(schema::account_storage_map_values::block_num.gt(block_from.to_raw_sql()))
+                .and(schema::account_storage_map_values::is_latest_update.eq(1)),
```
Mirko-von-Leipzig: I don't think this works if `block_to` is less than the latest update block?

crates/store/src/db/tests.rs
```diff
@@ -1239,6 +1239,79 @@ fn sql_account_storage_map_values_insertion() {
     );
 }
 
+#[test]
+fn test_select_storage_map_sync_values() {
```
Mirko-von-Leipzig:
```suggestion
fn select_storage_map_sync_values() {
```

crates/store/src/db/tests.rs
Mirko-von-Leipzig, line 1: Could you add a test where data is set in the genesis block? Since its possible for custom accounts etc to be injected there now, its a valid target.

proto/proto/store/rpc.proto
bobbinth: I would simplify this as well. I think the request could look like:
```proto
message SyncStorageMapsRequest {
    // block number to start sending updates from (inclusive).
    fixed32 block_from = 1;

    // Block number up to which to sync. If not specified, syncs up to the latest block.
    //
    // If specified, this block must be close to the chain tip (i.e., within 30 blocks),
    // otherwise an error will be returned.
    
    // Account for which we want to sync storage maps.
    AccountId account_id = 3;

    // Indexes of storage slots (in [0..255] range) for which to return the data.
    //
    // If any of the specified slots is not a storage map, an error is returned.
    repeated uint32 storage_slots = 4;
}
```

bobbinth: There is also an argument that we don't need to send `storage_slots` parameter and instead would sync all maps for the specified account. Selective storage map syncing may be useful, but maybe that's a more advanced feature that we can enable later if there is demand for it (and it should be a non-breaking change to add it as an optional parameter).

proto/proto/store/rpc.proto
bobbinth: I would say that even if it is not optional, we should still include the block number in the response. Otherwise we have nothing to paginate over.

proto/proto/store/rpc.proto
```diff
@@ -317,3 +320,54 @@ message SyncNotesResponse {
...
+
+    // Slot index ([0..255]).
+    uint32 slot_index = 2;
+}
+
+// Represents the result of syncing storage map request.
+message SyncStorageMapResponse {
+    // Storage map updates for the requested slots.
+    repeated StorageMapUpdate updates = 1;
+}
```
bobbinth: I think this still needs to be something like:
```proto
message SyncStorageMapsResponse {
    // The block number up to which this response progressed.
    //
    // If this is not equal to the `request.block_to`, then we are not fully synced yet
    // and the client needs to make a request with `request.block_from=response.block_num`.
    fixed32 block_num = 1;

    // Current chain tip
    fixed32 chain_tip = 1;

    // the list of updates up to `block_num`
    repeated StorageMapUpdate updates = 1;
}
```

proto/proto/store/rpc.proto
bobbinth: But nesting updates per storage map, we save a bit on bandwidth, but my guess is that building the response becomes a bit more complicated (because we need to sort entries by storage slot). I'm not sure this tradeoff is beneficial. That is, it may be simpler to define the update as:
```proto
message StorageMapUpdate {
    uint32 slot_idx = 1;
    Digest key = 2;
    Digest value = 3;
}
```
This will increase the bandwidth by something like 2% or 3% in the worst case.

proto/proto/rpc.proto
```diff
@@ -78,6 +78,10 @@ service Api {
     // part of hashes. Thus, returned data contains excessive notes, client can make
     // additional filtering of that data on its side.
     rpc SyncState(rpc_store.SyncStateRequest) returns (rpc_store.SyncStateResponse) {}
+
+    // Returns storage map updates for specified account and storage slots within a block range.
+    // Supports pagination for large storage maps using the last_key parameter.
+    rpc SyncStorageMap(rpc_store.SyncStorageMapRequest) returns (rpc_store.SyncStorageMapResponse) {}
```
bobbinth: nit: should this be `SyncStorageMaps`? (plural)

crates/store/src/db/models/queries.rs
```diff
@@ -545,6 +545,61 @@ pub(crate) fn select_latest_account_storage_map_values(
...
+        )
+        .order(schema::account_storage_map_values::key)
+        .into_boxed();
```
bobbinth: Could we add in the comments (maybe doc comment for the function) the SQL statement this is representing?

Also, a few thoughts based on my current understanding:
- We can simplify this statement if we remove filtering by slot index. In this case, we'll just need to run one query for all slots.
- Having `is_latest_update == true` doesn't work here. If `block_to` is smaller then the block of the latest update, we'll need to return an entry where `block_num <= bock` but only the latest such entry. This also makes me thing that the complexity of the query will be quite high and we may need to go back to my original proposal where the filter is pretty simple - i.e., `block_num >= block_from AND block <= block_to` (or in my very original proposal it was just `block_num >= block_from`).
- As far as I can tell, this does not limit the size of the response in any way. So, in theory, this could return many MB of data. We need to limit the size of the returned dataset (it should definitely be smaller than the limit imposed by gRPC, which I think is 10MB). We can't use a simple `LIMIT` here because we need to make sure we get all records for a given block - I don't have a solution for this yet.

bobbinth (commented):
Looks good! Thank you. Not a full review but I left some comments inline.

Overall, assuming I understood everything correctly, my main concern is that we are not limiting the sizes of responses, and so, it is not clear to me how we'll handle the cases when there is a large number of updates between `block_from` and `block_to` (this could be arbitrarily far apart).

Another thing is that it seems like we are trying to optimize response size over server complexity. I'd rather each response send a bit more data then have a very complicate query. This may require going back to my original proposal not having `block_to` parameter and also maybe getting rid of the slot indexes in the the request. But maybe I'm missing something here and there is a simple way handle the extra complexity.

crates/store/src/db/models/queries.rs
igamigo: Yeah, I meant to update this before EOD last friday, but unfortunately did not get to. I should've set the PR back to draft as well.

crates/store/src/db/models/queries.rs
igamigo: > We can't use a simple LIMIT here because we need to make sure we get all records for a given block - I don't have a solution for this yet.

I think if we had ordered map keys, we could be able to paginate over slot index and key by letting the user specify the last one they saw. I also think this simplifies the SQL query itself.

But since we want a simpler solution that selects all maps for an account, could we use `LIMIT` still? For example, if we assume that there's not going to be over 5MB worth of updates in a block, we could always also retrieve 5MB's worth of slot updates (since they are constant in size) ordered by block number and select the cutoff block in Rust. With no hard limit of updates within a block, not sure you can make any assumptions though (and also we might want the data transfers to be much smaller).

Mirko-von-Leipzig: That was a benefit of ordering by slot/key/value instead of by block. The pagination token can simply be an opaque encoding of these which we use to index the next page on without reading more than we need.

crates/store/src/db/models/queries.rs
```diff
@@ -545,6 +545,61 @@ pub(crate) fn select_latest_account_storage_map_values(
         .collect()
 }
 
+/// Select storage map values for syncing within a block range.
+///
+/// # Returns
+///
+/// A vector of tuples containing `(block_num, slot, key, value)` for the given account,
+/// slot, and block range. Results are ordered by key for consistent pagination.
+pub(crate) fn select_storage_map_sync_values(
```
drahnr: note: This change will clash with #1148

crates/store/src/db/models/queries.rs
bobbinth: > I think if we had ordered map keys, we could be able to paginate over slot index and key by letting the user specify the last one they saw. I also think this simplifies the SQL query itself.

> That was a benefit of ordering by slot/key/value instead of by block. The pagination token can simply be an opaque encoding of these which we use to index the next page on without reading more than we need.

I think I understand the approach now but I'm not sure I see how the query would look here and how the endpoints would work. For example, to support pagination we'd need to either have two separate endpoints or we'd need to optionally specify `slot_idx` and `key` of the item for the new page. And then the complexity of the query would be pretty high again. But I'm probably missing something.

> But since we want a simpler solution that selects all maps for an account, could we use `LIMIT` still? For example, if we assume that there's not going to be over 5MB worth of updates in a block, we could always also retrieve 5MB's worth of slot updates (since they are constant in size) ordered by block number and select the cutoff block in Rust. With no hard limit of updates within a block, not sure you can make any assumptions though (and also we might want the data transfers to be much smaller).

I think we can safely assume that there will be a limit on how much can change within a single block. Specifically, we could put a limit on the total size of all account deltas in the block and this limit will probably be under 5MB (cc @PhilippGackstatter as this could be a resource that we may need to include into fees).

I think using `LIMIT` and then filtering out some data in Rust could work. I believe there are two scenarios here:
- We get exactly $n$ records back where $n$ is our limit for how many records would constitute a 5MB response (or w/e other threshold we pick). In this case, we filter out the records with the highest block number because we don't know if we got all records back.
- We get fewer than $n$ records, in which case we know that we have all the data we need to send.

I think the SQL statement then would be something like:
```SQL
SELECT .. FROM account_storage_map_values
WHERE
    account_id = {account_id}
    AND block_num > {block_from}
ORDER BY
    block_num
LIMIT {n}
```

igamigo: >  For example, to support pagination we'd need to either have two separate endpoints or we'd need to optionally specify slot_idx and key of the item for the new page. And then the complexity of the query would be pretty high again. But I'm probably missing something.

Yes, my thinking was that it could be an optional page token/cursor that indicated where the user synced up to, down to the row. As we discussed on the call, if we could safely assume that there will be no more than a specific amount of storage updates in a block, we can go with the `LIMIT` query and then it becomes a very simple query, similar to the one you outlined.

crates/store/src/db/tests.rs
igamigo: Do you mean querying maps from accounts created in the genesis block?

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -459,6 +459,91 @@ pub(crate) fn select_slot_updates_stmt(
...
+    // ORDER BY block_num ASC, slot ASC, key ASC
+    // LIMIT :row_limit;
+
+    // TODO: These limits should be given by the protocol
+    pub const MAX_PAYLOAD_BYTES: usize = 5 * 1024 * 1024; // 5 MB
+    pub const ROW_OVERHEAD_BYTES: usize = size_of::<Word>() + size_of::<Word>() + size_of::<u8>(); // key + value + slot_idx
+    pub const ROW_LIMIT: usize = MAX_PAYLOAD_BYTES / ROW_OVERHEAD_BYTES;
+
+    let mut raw: Vec<(i64, i32, Vec<u8>, Vec<u8>)> =
+        SelectDsl::select(t::table, (t::block_num, t::slot, t::key, t::value))
```
drahnr, line 300: nit: there are always trade-offs between introducing a type and using the tuples. I think this is fine in this particular case, generally I'd default to a typed, non-tuple version.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -459,6 +459,91 @@ pub(crate) fn select_slot_updates_stmt(
...
+/// Each row contains one of:
+///
+/// - the historical value for a slot and key specifically on block `block_to`
+/// - the latest updated value for the slot and key combination, alongside the block number in which
+///   it was updated
+pub(crate) fn select_account_storage_map_values(
+    conn: &mut SqliteConnection,
+    account_id: AccountId,
+    block_from: BlockNumber,
+    block_to: BlockNumber,
```
drahnr, line 265: nit: we need better var names, the above suggests a range, but it's not

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -459,6 +459,91 @@ pub(crate) fn select_slot_updates_stmt(
...
+
+    // SELECT
+    //   block_num,
+    //   slot,
+    //   key,
+    //   value
+    // FROM account_storage_map_values
+    // WHERE account_id = ?1
+    //   AND block_num >= ?2
+    //   AND (block_num = ?3 OR is_latest_update = 1)
```
drahnr: nit: we should add an argument validation of `block_from <= block_to`

crates/store/src/db/mod.rs
```diff
@@ -454,6 +455,21 @@ impl Db {
         .await
     }
 
+    /// Selects storage map values for syncing within a block range, with pagination support.
```
drahnr: I don't think this is entirely correct given the actual query.

crates/store/src/db/tests.rs
Mirko-von-Leipzig: Yes correct :+1:

crates/rpc/README.md
```diff
@@ -109,6 +110,16 @@ used to update the state of Chain MMR. This includes both chain MMR peaks and ch
...
 notes, client can make additional filtering of that data on its side.
+
+---
+
+### SyncStorageMaps
+
+Returns storage map synchronization data for a specified public account within a given block range. This method allows clients to efficiently sync the storage map state of an account by retrieving only the changes that occurred between two blocks.
+
+Caller specifies the `account_id` of the public account and the block range (`block_from`, `block_to`) for which to retrieve storage updates. The response includes all storage map key-value updates that occurred within that range, along with the last block included in the sync and the current chain tip.
+This endpoint enables clients to maintain an updated view of account storage.
```
bobbinth, line 122: nit: I'd add a newline above this line.

proto/proto/store/rpc.proto
```diff
@@ -292,3 +295,48 @@ message SyncNotesResponse {
...
+    optional fixed32 block_to = 2;
+    
+    // Account for which we want to sync storage maps.
+    account.AccountId account_id = 3;
+}
+
+message SyncStorageMapsResponse {
+    // The block number of the last row included in this response.
+    // For chunked responses, this may be less than request.block_to.
+    fixed32 block_num = 1;
```
bobbinth: nit: the reference to "row" here may be confusing. Could we say "last update" or something like that instead?

Should we also mention that if this is not equal to `request.block_to` then the user would need to make another request with `request.block_from` equal to `block_num + 1`?

crates/store/README.md
```diff
@@ -131,6 +132,15 @@ data contains excessive notes and nullifiers, client can make additional filteri
 
 ---
 
+### SyncStorageMaps
+
+Returns storage map synchronization data for a specified public account within a given block range. This method allows clients to efficiently sync the storage map state of an account by retrieving only the changes that occurred between two blocks.
+
+Caller specifies the `account_id` of the public account and the block range (`block_from`, `block_to`) for which to retrieve storage updates. The response includes all storage map key-value updates that occurred within that range, along with the last block included in the sync and the current chain tip.
+This endpoint enables clients to maintain an updated view of account storage.
```
bobbinth, line 141: Same nit as in another comment: I would add a newline above this line.

crates/store/src/server/rpc_api.rs
```diff
@@ -329,6 +330,67 @@ impl rpc_server::Rpc for StoreApi {
...
+        let block_from = BlockNumber::from(request.block_from);
+        let chain_tip = self.state.latest_block_num().await;
+
+        let block_to = match request.block_to {
+            Some(block_to) => BlockNumber::from(block_to),
+            None => chain_tip,
+        };
+
+        if block_from > block_to {
+            return Err(Status::invalid_argument("block_from cannot be higher than block_to"));
```
bobbinth: nit: "cannot be higher" -> "cannot be greater".

proto/proto/store/rpc.proto
```diff
@@ -292,3 +295,48 @@ message SyncNotesResponse {
...
+// Represents a single storage map update.
+message StorageMapUpdate {
+    // Slot index ([0..255]).
+    uint32 slot_index = 1;
+    
+    // The storage map key.
+    bytes key = 2;
+        
+    // The storage map value.
+    bytes value = 3;
```
bobbinth: I would probably use `Digest` instead of `bytes` here, mostly because using `Digest` should allow us not to allocate vectors (I think).

bin/faucet/src/stub_rpc_api.rs
bobbinth, line 1: I think we should be able to delete this file.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -219,6 +220,105 @@ pub(crate) fn select_all_accounts(
...
+    //   block_num,
+    //   slot,
+    //   key,
+    //   value
+    // FROM account_storage_map_values
+    // WHERE account_id = ?1
+    // AND block_num >= ?2
+    // AND (block_num = ?3 OR (is_latest_update = 1 AND block_num <= ?3))
+    // ORDER BY block_num ASC
+    // LIMIT :row_limit;
```
bobbinth: nit: I'd format this as:
```rust
// SELECT
//     block_num,
//     slot,
//     key,
//     value
// FROM
//     account_storage_map_values
// WHERE
//     account_id = ?1
//     AND block_num >= ?2
//     AND (block_num = ?3 OR (is_latest_update = 1 AND block_num <= ?3))
// ORDER BY
//     block_num ASC
// LIMIT
//     :row_limit;
```

Also, what does `AND (block_num = ?3 OR (is_latest_update = 1 AND block_num <= ?3))` do? That is, why not replace it with just `AND block_num <= ?3`?

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -219,6 +220,105 @@ pub(crate) fn select_all_accounts(
...
+
+    // TODO: These limits should be given by the protocol.
+    // See miden-base/issues/1770 for more details
+    pub const MAX_PAYLOAD_BYTES: usize = 5 * 1024 * 1024; // 5 MB
```
bobbinth: Let's reduce this to 2MB.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -219,6 +220,105 @@ pub(crate) fn select_all_accounts(
...
+    let mut last_block_included = block_to;
+    if raw.len() == ROW_LIMIT {
+        // NOTE: If the query contains at least one more row than the amount of storage map updates
+        // allowed in a single block for an account, then the response is guaranteed to have at
+        // least two blocks
+        if let Some(last_block_num) = raw.last().map(|r| r.0) {
+            raw.retain(|(row_block_num, ..)| *row_block_num != last_block_num);
+            last_block_included = BlockNumber::from_raw_sql(last_block_num.saturating_sub(1))?;
+        }
+    }
```
bobbinth: A small optimization could be to have an `else` branch to handle the case when `raw.len() != ROW_LIMIT`. This way, we could avoid iterating over `raw` data set twice. To simplify the code we probably would need to have a helper function that maps raw rows to `StorageMapValue`. Maybe something like `StorageMapValue::from_raw_row()`.

proto/proto/store/rpc.proto
```diff
@@ -292,3 +295,48 @@ message SyncNotesResponse {
...
+message SyncStorageMapsResponse {
+    // The block number of the last row included in this response.
+    // For chunked responses, this may be less than request.block_to.
+    fixed32 block_num = 1;
+
+    // Current chain tip
+    fixed32 chain_tip = 2;
+
+    // The list of storage map updates.
+    repeated StorageMapUpdate updates = 3;
```
bobbinth: The set of returned updates here could contain duplicates, right? If so, I would make it clear in the comment that only the latest value for a given `(slot_index, key)` combination should be retained.

bobbinth (commented):

crates/store/src/db/models/queries/accounts.rs
igamigo: I agree. I'm not sure what would be better candidates though

crates/store/src/server/rpc_api.rs
igamigo: One day I'll internalize this one

crates/store/src/db/models/queries/accounts.rs
igamigo: Yeah, I think your solution is more appropriate. I drafted this as returning either the value for the exact `block_to` or  latest values if they were created in a previous block, but this has some problems.

crates/store/src/server/rpc_api.rs
```diff
@@ -329,6 +329,68 @@ impl rpc_server::Rpc for StoreApi {
...
+        let chain_tip = self.state.latest_block_num().await;
+
+        let block_to = match request.block_to {
+            Some(block_to) => BlockNumber::from(block_to),
+            None => chain_tip,
+        };
+
+        if block_from > block_to {
+            return Err(Status::invalid_argument("block_from cannot be greater than block_to"));
+        }
```
Mirko-von-Leipzig, line 368: I think this gets handled internally as well?

That said, shouldn't this be valid and just returns zero items? But okay no strong feelings here :)

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -219,6 +220,119 @@ pub(crate) fn select_all_accounts(
...
+    let (last_block_included, values) = if raw.len() == ROW_LIMIT {
+        // NOTE: If the query contains at least one more row than the amount of storage map updates
+        // allowed in a single block for an account, then the response is guaranteed to have at
+        // least two blocks
+
+        // SAFETY: we checked that the vector is not empty
+        let &(last_block_num, ..) = raw.last().unwrap();
+        let values = raw
+            .into_iter()
+            .filter(|(bn, ..)| *bn != last_block_num)
```
Mirko-von-Leipzig: Because we've ordered by block number we can use `take_while`

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -219,6 +220,119 @@ pub(crate) fn select_all_accounts(
...
+                    .eq(account_id.to_bytes())
+                    .and(t::block_num.ge(block_from.to_raw_sql()))
+                    .and(t::block_num.le(block_to.to_raw_sql())),
+            )
+            .order(t::block_num.asc())
+            .limit(i64::try_from(ROW_LIMIT).expect("limit fits within i64"))
+            .load(conn)?;
+
+    // Discard the last block in the response (assumes more than one block may be present)
+    let (last_block_included, values) = if raw.len() == ROW_LIMIT {
```
Mirko-von-Leipzig: Nit: we sort of rely on sqlite's `LIMIT` implementation here


```suggestion
    let (last_block_included, values) = if raw.len() >= ROW_LIMIT {
```

Mirko-von-Leipzig (approved):
Some minor nits

crates/store/src/db/models/queries/accounts.rs
Mirko-von-Leipzig: An even larger optimization would be doing an `rfind` since we expect this to be at the back end of a large vector, but that's probably not necessary at this stage.

crates/store/src/server/rpc_api.rs
igamigo: I wasn't sure either in regards of how we would want to approach this, but seems better to try to catch it as early as possible

crates/store/src/db/models/queries/accounts.rs
igamigo: Good point, changed

bobbinth (approved):
All looks good! Thank you!

PR #1143, opened 2025-08-07
bump miden base - integrate `FeeParameters` into `GenesisConfig`

drahnr:
bump miden base to [https://github.com/0xMiden/miden-base?branch=next#3da4d1f2dbd58092a501ba39123afc3f7a80e9f3](3da4d1f2dbd58092a501ba39123afc3f7a80e9f3)

TODOs
* [x] figure out where `Fee*` ( or `FaucetId` and amount) is supposed to be ingested from
* [x] remove all `fn fee_(.+_)?stub` stubs 

Contains changes from:
* https://github.com/0xMiden/miden-node/pull/1144


crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -1,5 +1,6 @@
-timestamp = 1717344256
-version   = 1
+native_fee_symbol = "MIDEN"
```
bobbinth: Another option to do this is to add `is_native` option to the `fungible_faucet` element:
```toml
[[fungible_faucet]]
decimals     = 3
max_supply   = 100_000_000
storage_mode = "public"
symbol       = "MIDEN"
is_native    = true
```
We would have to validate that there is exactly one native fungible faucet specified - but probably not too difficult to do this.

Other things to keep in mind: `storage_mode` of the native faucet should always be "public", and we should enforce this.

-------

Another option is to have a separate section for fees. This will give us a place to record initial base fees as well. Something like:

```toml
[fee_parameters]
fee_asset_symbol      = "MIDEN"
verification_base_fee = 0
```

crates/store/src/genesis/mod.rs
```diff
@@ -25,6 +26,7 @@ pub mod config;
 /// Represents the state at genesis, which will be used to derive the genesis block.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct GenesisState {
+    pub native_asset_account_id: AccountId,
     pub accounts: Vec<Account>,
     pub version: u32,
     pub timestamp: u32,
```
bobbinth: Related to the previous comment, this should probably look like so:
```rust
pub struct GenesisState {
    pub accounts: Vec<Account>,
    pub version: u32,
    pub timestamp: u32,
    pub fee_parameters: FeeParameters,
}
```
With `fee_parameters` being populated from the genesis TOML file.

crates/store/src/genesis/config/mod.rs
```diff
@@ -151,6 +156,11 @@ impl GenesisConfig {
             // we know the remaining supply in the faucets.
         }
 
+        let native_asset_account_id = faucet_accounts
+            .get(&native_fee_symbol)
+            .ok_or_else(|| GenesisConfigError::MissingFeeFaucet(native_fee_symbol))?
+            .id();
```
bobbinth: As mentioned in one of the comments: we should also add a check to make sure that the fee faucet is a public account.

crates/proto/src/domain/block.rs
```diff
@@ -86,11 +87,25 @@ impl TryFrom<proto::blockchain::BlockHeader> for BlockHeader {
...
+        AccountId::dummy(
+            [0_u8; 15],
+            AccountIdVersion::Version0,
+            miden_objects::account::AccountType::FungibleFaucet,
+            miden_objects::account::AccountStorageMode::Network,
+        ),
+        0_u32,
+    )
+    .unwrap()
+}
```
bobbinth: I think we should add `FeeParameters` to the protobuf messages. The message would look something like this:

```proto
message FeeParameters {
    account.AccountId native_asset_id = 1;
    uint32 verification_base_fee = 2;
}
```

And then we'd include it into the `BlockHeader` message.

bin/stress-test/src/seeding/mod.rs
```diff
@@ -367,10 +370,15 @@ fn create_consume_note_txs(
...
+fn fee_asset(faucet_id: AccountId) -> FungibleAsset {
+    FungibleAsset::new(faucet_id, 0).unwrap()
+}
+
 /// Creates a transaction that creates an account and consumes the given input note.
 ///
 /// The account is updated with the assets from the input note, and the nonce is incremented.
 fn create_consume_note_tx(
+    faucet_id: AccountId,
     block_ref: &BlockHeader,
```
bobbinth: I don't think we need an additional parameter for this. Instead we should be able to do something like `block_ref.fee_parameters().native_asset_id()`.

Same applies to a couple more functions in this file.

crates/block-producer/src/test_utils/proven_tx.rs
```diff
@@ -3,8 +3,10 @@ use std::ops::Range;
 use itertools::Itertools;
 use miden_air::HashFunction;
 use miden_objects::account::AccountId;
+use miden_objects::asset::FungibleAsset;
 use miden_objects::block::BlockNumber;
 use miden_objects::note::{Note, NoteExecutionHint, NoteHeader, NoteMetadata, NoteType, Nullifier};
+use miden_objects::testing::account_id::ACCOUNT_ID_PUBLIC_FUNGIBLE_FAUCET;
```
bobbinth: For test fees we should probably use `ACCOUNT_ID_NATIVE_ASSET_FAUCET` here.

crates/store/src/db/tests.rs
```diff
@@ -67,6 +68,11 @@ fn create_db() -> SqliteConnection {
     conn
 }
 
+fn test_fee_params() -> FeeParameters {
+    let faucet: AccountId = ACCOUNT_ID_PUBLIC_FUNGIBLE_FAUCET.try_into().unwrap();
```
bobbinth: Same comment re `ACCOUNT_ID_NATIVE_ASSET_FAUCET` as above.

crates/rpc/src/tests.rs
```diff
@@ -16,6 +17,8 @@ use miden_objects::account::{
     AccountType,
     AccountVaultDelta,
 };
+use miden_objects::asset::FungibleAsset;
+use miden_objects::testing::account_id::ACCOUNT_ID_PUBLIC_FUNGIBLE_FAUCET;
```
bobbinth: Same comment as `ACCOUNT_ID_NATIVE_ASSET_FAUCET` as above.

bobbinth (commented):
Looks good! Thank you! I left some comments inline. The main one is about putting the full `FeeParameters` data into genesis config.

crates/store/src/genesis/config/samples/01-simple.toml
Mirko-von-Leipzig: How about a specific item for it then which doesn't let you specify invalid options i.e. no storage mode etc.

```toml
[[native_faucet]]
decimals = 3
max_supply = 100_000_000
symbol = "MIDEN"
```

The deserialized struct could then enforce the requirement via serde or whatever we're using to parse things.

bobbinth: Yep - that works too.

One question: where would other fee parameters go? (e.g., `verification_base_fee`, but there will be more)

bobbinth: I guess we could still keep them separate - e.g., something like:
```toml
[[native_faucet]]
decimals = 3
max_supply = 100_000_000
symbol = "MIDEN"

[[fee_parameters]]
verification_base_fee = 0
# more in the future
```

I think that's probably what we should go with.

Mirko-von-Leipzig: I think my example was technically incorrect since it used `[[ ]]` instead of `[ ]` since we only want the one.

We could nest all these things under `fee` :shrug: Many ways to slice this up but if we think they're all related

```toml
# Flattened
fee.faucet.decimals = 3
fee.facuet.max_supply = 100_000_000
fee.faucet.symbol = "MIDEN"
fee.parameters.verification.base = 0
fee.parameters.verification.xxx = y
fee.parameters.growth = zzz

# Sectioned
[fee.faucet]
decimals = 3
max_supply = 100_000_000
symbol = "MIDEN"

[fee.parameters.verification]
base = 0
xxx = y

[fee.parameters]
growth = zZZ
```


```

Comment (bobbinth) [reply crates/store/src/genesis/config/samples/01-simple.toml]: Yes, definitely many ways :) though I do have a slight preference for keeping the native faucet and fee parameters separate (they are obviously related, but native asset could be used for other things within the protocol too). How about something like:
```toml
[native_asset]
decimals = 3
max_supply = 100_000_000
symbol = "MIDEN"

[fee_parameters]
base_fee.verification = 0
# more base fees in the future
# more other fee parameters in the future
```

Mirko-von-Leipzig: `fee_parameters` just feels weird to me because all these things are parameters/configs :D

drahnr: Implemented the latter

crates/proto/src/generated/blockchain.rs
```diff
@@ -62,20 +62,21 @@ pub struct BlockHeader {
     /// A commitment to all transaction kernels supported by this block.
     #[prost(message, optional, tag = "10")]
     pub tx_kernel_commitment: ::core::option::Option<super::primitives::Digest>,
-    /// Fee parameters for block processing.
-    #[prost(message, optional, tag = "11")]
-    pub fee_parameters: ::core::option::Option<FeeParameters>,
     /// The time when the block was created.
-    #[prost(fixed32, tag = "12")]
+    #[prost(fixed32, tag = "11")]
```
drahnr: From experience, this shouldn't change after once used. The error message used to funny.

crates/store/src/genesis/mod.rs
```diff
@@ -26,6 +26,7 @@ pub mod config;
 /// Represents the state at genesis, which will be used to derive the genesis block.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct GenesisState {
+    pub fee_parameters: FeeParameters,
     pub accounts: Vec<Account>,
```
bobbinth, line 30: nit: I would reverse the order of fields - first `accounts` and then `fee_parameters`.

crates/store/src/genesis/mod.rs
```diff
@@ -46,8 +47,18 @@ impl GenesisBlock {
...
 
 impl GenesisState {
-    pub fn new(accounts: Vec<Account>, version: u32, timestamp: u32) -> Self {
-        Self { accounts, version, timestamp }
+    pub fn new(
+        fee_parameters: FeeParameters,
+        accounts: Vec<Account>,
+        version: u32,
+        timestamp: u32,
+    ) -> Self {
```
bobbinth, line 55: Same nit as above.

crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -1,6 +1,11 @@
...
 version   = 1
 
+
+[fee_parameters]
+symbol                = "MIDEN"
+verification_base_fee = 0
+
 [[fungible_faucet]]
 decimals     = 3
 max_supply   = 100_000_000
```
bobbinth, line 11: I'm assuming we left the new approach for defining these (i.e., something like in https://github.com/0xMiden/miden-node/pull/1143#discussion_r2265981740) for the future - right? Let's create an issue for refactoring this later.

crates/ntx-builder/src/transaction.rs
```diff
@@ -91,7 +91,6 @@ impl NtxContext {
             .set_attribute("reference_block.number", chain_tip_header.block_num());
 
         async move {
-            // Work-around for `TransactionExecutor` not being `Send`.
             async move {
```
bobbinth, line 94: Question: why do we need nested `async move`s here?

bobbinth (approved):
Looks good! Thank you! I left a few small nits inline - but this is basically good to go. Would be good to get another look from @Mirko-von-Leipzig though.

crates/ntx-builder/src/transaction.rs
drahnr: For the inspection and span wrapping for futures, but the returned future would have incompatible lifetimes last time I checked

crates/store/src/genesis/config/samples/01-simple.toml
drahnr: I prefer simplicity until we have the means to make an educated choice of how it's going to be used, and didn't want to block the client any longer on that discussion specifically

bobbinth: Yep - makes sense. @Mirko-von-Leipzig already created https://github.com/0xMiden/miden-node/issues/1149 for this.

PR #1144, opened 2025-08-08
adjust to latest clippy lints and rust doc quotation

drahnr:
* Adds backticks around the argument inside `[]` ```[`Self::foo()`]```
* Disable some new lints - the purpose of them become increasingly questionable
* Apply the sane lints (i.e. future size warnings)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ url                = { features = ["serde"], version = "2.5" }
 pedantic = { level = "warn", priority = -1 }
 
 cast_possible_truncation     = "allow" # Overly many instances especially regarding indices.
+collapsible-if               = "allow" # Too new to enforce.
 from_iter_instead_of_collect = "allow" # at times `FromIter` is much more readable
 ignored_unit_patterns        = "allow" # Stylistic choice.
 large_types_passed_by_value  = "allow" # Triggered by BlockHeader being Copy + 334 bytes.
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -89,12 +89,14 @@ url                = { features = ["serde"], version = "2.5" }
 pedantic = { level = "warn", priority = -1 }
 
 cast_possible_truncation     = "allow" # Overly many instances especially regarding indices.
+collapsible-if               = "allow" # Too new to enforce.
```
Mirko-von-Leipzig, line 92: Was this buggy? Seems like a reasonable thing to do in _most_ cases?

Mirko-von-Leipzig (approved):
Where are all the `Box::pin` from? A new lint?

Cargo.toml
drahnr: `if let Foo(_) = bar && if let Some(_) = baz { .. } ` barely makes this more readable. This is opinionated, there is also just one instance.

On a broader stroke: We should opt-in to specific lints rather than the ever-changing bag of `pedantic`.

Mirko-von-Leipzig: While that's true; there is also the lack of discoverability if its not suggested at all. Personally I like the intent that it shows its a single scope and not something with potential fall-through.

But let's see how it goes; doesn't come up all that often

PR #1145, opened 2025-08-08
feat: fee params in proto messages

igamigo:


bobbinth (approved):
Looks good! Thank you!

PR #1146, opened 2025-08-10
Make use of boolean type in DB

c410-f3r:
Fixes a TODO

Sqlite accepts BOOL and BOOLEAN as aliases for INTEGER -> https://www.db-fiddle.com/f/6Y7oEY2Xkwc29iFHzQWG7J/0

Not sure how diesel handles this particular scenario but if the tests aren't failing then it should be OK.

### DIFF
--- a/crates/store/src/db/migrations/2025062000000_setup/up.sql
+++ b/crates/store/src/db/migrations/2025062000000_setup/up.sql
@@ -123,7 +123,7 @@ CREATE TABLE account_non_fungible_asset_updates (
     account_id  BLOB NOT NULL,
     block_num   INTEGER NOT NULL,
     vault_key   BLOB NOT NULL,
-    is_remove   INTEGER NOT NULL, -- 0 - add, 1 - remove
+    is_remove   BOOLEAN NOT NULL,
 
     PRIMARY KEY (account_id, block_num, vault_key),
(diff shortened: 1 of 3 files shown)


crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -123,7 +123,7 @@ CREATE TABLE account_non_fungible_asset_updates (
     account_id  BLOB NOT NULL,
     block_num   INTEGER NOT NULL,
     vault_key   BLOB NOT NULL,
-    is_remove   INTEGER NOT NULL, -- 0 - add, 1 - remove
+    is_remove   BOOLEAN NOT NULL,
```
c410-f3r, line 126: Miden is still in development so it is OK to overwrite this migration, right?

Mirko-von-Leipzig: That's correct yes :100: 

The only question mark for me is the `diesel` interaction with this but I'll leave that to @drahnr

drahnr: We'd get encoding errors, if the value range was out of bounds. It implies assumptions on the diesel level, and hence error earlier in teh diesel abstraction, which is fine.

My question is, do we anticipate this becoming an enum soon~ish? If not, good to go

Mirko-von-Leipzig (approved):
Thank you!

Issue #1134, opened 2025-08-05
store: refactor `queries.rs` and `types.rs`


drahnr:
The `queries.rs` file has grown to 1.5k loc which is far too many. There are also types that are oftentimes only relevant to a single query or insert/upsert that accumulated in `types.rs`.

The goal is to find a more coherent way of rerpresenting the queries, where all relevant types "live together".

Issue #629, opened 2025-01-18
Investigate why store seemingly allows duplicate accounts


Mirko-von-Leipzig:
> Also, separate from this PR, I'm curious why inserting identical accounts into the store didn't trigger any errors. Maybe something for us to take a look at in the future. cc @Mirko-von-Leipzig.

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-node/pull/621#discussion_r1920895223_

polydez:
Storage could check for creation of new public account and return error, if it already exists. This check is only possible for public accounts now, because block account update doesn't consist of account's previous commitment: https://github.com/0xPolygonMiden/miden-base/blob/next/crates/miden-objects/src/block/block_account_update.rs#L13-L30

Thus, it's impossible for store to distinguish account update from insertion by current design.

bobbinth:
> We should _not_ mutate the store outside of `apply_block`. Any other method requires knowing exactly what you're doing and what invariants you should uphold.

The seed data builder should be going through `apply_block`, right? I think as long as we know that we can't insert duplicate accounts via `apply_block`, we are good.

Mirko-von-Leipzig:
> The seed data builder should be going through apply_block, right?

I think it does from a cursory check; but at the time this issue was created it used the inner APIs.

> I think as long as we know that we can't insert duplicate accounts via apply_block, we are good.

So long as the [account tree](https://github.com/0xMiden/miden-node/blob/dc0b1202894c56bb3a1816d5f6ba4a11432522ac/crates/store/src/state.rs#L237-L246) does its job, I think we're safe.

@SantiagoPittella I think if you could confirm seed builder uses `apply_block` via the store client (which I think it does), then we can close this.

SantiagoPittella:
Yes, the benchmark seed uses `apply_block` 👍🏼 .

PR #1148, opened 2025-08-11
refactor: split queries into files

drahnr:
Bare minimum of splitting `queries.rs` into multiple files, and only splits them up, no behaviour is changed in this PR.

Ref #1134

### DIFF
--- a/crates/store/src/db/models/queries/accounts.rs
+++ b/crates/store/src/db/models/queries/accounts.rs
@@ -223,9 +223,10 @@ pub(crate) fn select_account_delta(
         let asset = NonFungibleAsset::read_from_bytes(&vault_key_asset)
             .map_err(|err| DatabaseError::DataCorrupted(err.to_string()))?;
 
-        match is_remove {
-            false => non_fungible_delta.add(asset)?,
-            true => non_fungible_delta.remove(asset)?,
+        if is_remove {
+            non_fungible_delta.remove(asset)?;
+        } else {
+            non_fungible_delta.add(asset)?;
         }
     }
 
--- a/crates/store/src/db/models/queries/notes.rs
+++ b/crates/store/src/db/models/queries/notes.rs
@@ -1,3 +1,8 @@
+#![allow(
+    clippy::cast_possible_wrap,
+    reason = "We will not approach the item count where i64 and usize cause issues"
+)]
+
 use std::collections::{BTreeMap, BTreeSet};
 
 use diesel::prelude::{
--- a/crates/store/src/db/models/conv.rs
+++ b/crates/store/src/db/models/conv.rs
@@ -26,6 +26,11 @@
     clippy::cast_sign_loss,
     reason = "This is the one file where we map the signed database types to the working types"
 )]
+#![allow(
+    clippy::cast_possible_wrap,
+    reason = "We will not approach the item count where i64 and usize casting will cause issues
+    on relevant platforms"
+)]
 
 use std::any::type_name;
 
@@ -57,7 +62,7 @@ impl SqlTypeConvert for BlockNumber {
             .map_err(|_| DatabaseTypeConversionError(type_name::<BlockNumber>()))
     }
     fn to_raw_sql(self) -> Self::Raw {
-        self.as_u32() as i64
+        i64::from(self.as_u32())
     }
 }
 
@@ -69,7 +74,7 @@ impl SqlTypeConvert for NetworkAccountPrefix {
             .map_err(|_e| DatabaseTypeConversionError(type_name::<NetworkAccountError>()))
     }
     fn to_raw_sql(self) -> Self::Raw {
-        self.inner() as i64
+        i64::from(self.inner())
     }
 }
 
@@ -99,7 +104,7 @@ pub(crate) fn raw_sql_to_nullifier_prefix(raw: i32) -> u16 {
 }
 #[inline(always)]
 pub(crate) fn nullifier_prefix_to_raw_sql(prefix: u16) -> i32 {
-    prefix as i32
+    i32::from(prefix)
 }
 
 #[inline(always)]
@@ -119,7 +124,7 @@ pub(crate) fn raw_sql_to_slot(raw: i32) -> u8 {
 }
 #[inline(always)]
 pub(crate) fn slot_to_raw_sql(slot: u8) -> i32 {
-    slot as i32
+    i32::from(slot)
 }
 
 #[inline(always)]
@@ -137,8 +142,8 @@ pub(crate) fn raw_sql_to_note_type(raw: i32) -> u8 {
     raw as u8
 }
 #[inline(always)]
-pub(crate) fn note_type_to_raw_sql(note_type_: u8) -> i32 {
-    note_type_ as i32
+pub(crate) fn note_type_to_raw_sql(note_type: u8) -> i32 {
+    i32::from(note_type)
 }
 
 #[inline(always)]


crates/store/src/db/models/queries/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+pub use block_headers::*;
+mod accounts;
+pub use accounts::*;
+mod nullifiers;
+pub(crate) use nullifiers::*;
+mod notes;
+pub(crate) use notes::*;
+
+mod insertions;
+pub(crate) use insertions::*;
```
Mirko-von-Leipzig, line 28: Are these required externally? I would imagine only `apply_block` would need these

drahnr: We do use them in `tests`, so we need to expose them at the crate level

Mirko-von-Leipzig: Ideally we shouldn't be using these directly since there are some assumptions I think, i.e. only `apply_block` is a safe writer interface.

This has lead to "confusion" before e.g. https://github.com/0xMiden/miden-node/issues/629#issuecomment-3130975505.

Might be good to improve tests by moving to using `MockChain` for data generation and then only using `apply_block` to get the data in. But maybe I'm over thinking this.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Queryable, Selectable, QueryableByName)]
+#[diesel(table_name = schema::transactions)]
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+pub struct TransactionSummaryRaw {
+    account_id: Vec<u8>,
+    block_num: i64,
+    transaction_id: Vec<u8>,
+}
```
bobbinth, line 67: General question about potential future improvements: is there a way somehow to use arrays instead of vectors for things like this?

Basically, we know that account ID would always be `[u8; 16]` and transaction ID would always be `[u8; 32]`, and it feels quite wasteful to allocate two vectors for every record (especially if we read a lot of rows from the database). Or maybe the complier can somehow optimize these away.

If using arrays is not possible, I'm wondering if the limitation lies with SQLite, Disel, both, or maybe something else.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -0,0 +1,78 @@
...
+    //     block_num,
+    //     transaction_id
+    // FROM
+    //     transactions
+    // WHERE
+    //     block_num > ?1 AND
+    //     block_num <= ?2 AND
+    //     account_id IN rarray(?3)
+    // ORDER BY
+    //     transaction_id ASC
```
bobbinth, line 41: Not for this PR, but maybe it would make sense to move SQL statements to doc comments of this function (and other similar such functions).

We should also note in doc comments that the query imposes a limit on the number of account IDs that could be provided.

Another thing that may be good to note in the comments is that this query could return very large amounts of data, and so we are relying on the caller to make sure the `block_start..block_end` range is relatively small.

Separately, maybe we should normalize naming across queries and use `block_from` and `block_to` instead of `block_start` and `block_end`.

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -0,0 +1,139 @@
...
+///
+/// # Returns
+///
+/// A vector of [`NullifierInfo`] with the nullifiers and the block height at which they were
+pub(crate) fn select_nullifiers_by_prefix(
+    conn: &mut SqliteConnection,
+    prefix_len: u8,
+    nullifier_prefixes: &[u16],
+    block_num: BlockNumber,
+) -> Result<Vec<NullifierInfo>, DatabaseError> {
```
bobbinth, line 40: Not for this PR, but a few comments:

- The sentence on line 34 seems to be cut off.
- We should probably mention that we impose a limit on the number of nullifier prefixes.
- `block_num` should probably be renamed to `block_from`.
- Similar to one of the previous comments, we could consider moving SQL statement from the body of the function into doc comments.

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/// Select all nullifiers from the DB
+///
+/// # Returns
+///
+/// A vector with nullifiers and the block height at which they were created, or an error.
+pub(crate) fn select_all_nullifiers(
+    conn: &mut SqliteConnection,
+) -> Result<Vec<NullifierInfo>, DatabaseError> {
```
bobbinth, line 73: Comment for the future: we use this function during store startup to build the nullifier tree. It works now, but as soon as the number of nullifiers becomes large, this strategy will become infeasible. Fixing this will involve migrating to the `LargeSmt` which may involve additional code that would let us re-build the nullifier SMT from scratch by connecting to the database directly (i.e., not as a part of the gRPC interface).

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -0,0 +1,139 @@
...
+///
+/// # Note
+///
+/// The [`SqliteConnection`] object is not consumed. It's up to the caller to commit or rollback the
+/// transaction.
+pub(crate) fn insert_nullifiers_for_block(
+    conn: &mut SqliteConnection,
+    nullifiers: &[Nullifier],
+    block_num: BlockNumber,
+) -> Result<usize, DatabaseError> {
```
bobbinth, line 96: Not for this PR, but we should mention in the comments that we impose a limit on the number of nullifiers. This is especially important because this implicitly limits how many nullifiers can be created in a single block.

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -0,0 +1,139 @@
...
+    QueryParamNullifierLimit::check(nullifiers.len())?;
+
+    // UPDATE notes SET consumed = TRUE WHERE nullifier IN rarray(?1)
+    let serialized_nullifiers =
+        Vec::<Vec<u8>>::from_iter(nullifiers.iter().map(Nullifier::to_bytes));
```
bobbinth, line 101: Similar to one of the previous questions: I wonder if there is a way to avoid allocating a vector per nullifier here (a nullifier is always `[u8; 32]`).

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -0,0 +1,139 @@
...
+            |(nullifier, bytes)| {
+                (
+                    schema::nullifiers::nullifier.eq(bytes),
+                    schema::nullifiers::nullifier_prefix
+                        .eq(nullifier_prefix_to_raw_sql(get_nullifier_prefix(nullifier))),
+                    schema::nullifiers::block_num.eq(block_num.to_raw_sql()),
+                )
+            },
+        )))
+        .execute(conn)?;
```
bobbinth, line 119: Do we need to allocate a new vector here?

Also, would be good to add a comment with the SQL statement that we are executing here.

crates/store/src/db/models/queries/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+pub(crate) fn apply_block(
+    conn: &mut SqliteConnection,
+    block_header: &BlockHeader,
+    notes: &[(NoteRecord, Option<Nullifier>)],
+    nullifiers: &[Nullifier],
+    accounts: &[BlockAccountUpdate],
+    transactions: &OrderedTransactionHeaders,
+) -> Result<usize, DatabaseError> {
```
bobbinth, line 37: Would be good to add some doc comments here - e.g., it is not immediately clear what this function returns.

crates/store/src/db/models/queries/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    Ok(count)
+}
+
+/// Loads the state necessary for a state sync.
+pub(crate) fn get_state_sync(
+    conn: &mut SqliteConnection,
+    since_block_number: BlockNumber,
+    account_ids: Vec<AccountId>,
+    note_tags: Vec<u32>,
+) -> Result<StateSyncUpdate, StateSyncError> {
```
bobbinth, line 55: nit: `since_block_number` should probably be renamed to `block_from`.

crates/store/src/db/models/queries/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    let account_updates =
+        select_accounts_by_block_range(conn, &account_ids, block_start, block_end)?;
+
+    // select transactions by accounts and block range
+    let transactions = select_transactions_by_accounts_and_block_range(
+        conn,
+        &account_ids,
+        block_start,
+        block_end,
+    )?;
```
bobbinth, line 81: Not for this PR, but this query could potentially result in a very large result set. The case when this could happen is if an account had a very large number of transactions which produced no notes over a long period of time. This could result in the `block_start..block_end` range that is very big.

crates/store/src/db/models/queries/block_headers.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// # Note
+///
+/// Only returns the block headers that are actually present.
+///
+/// # Returns
+pub fn select_block_headers(
+    conn: &mut SqliteConnection,
+    blocks: impl Iterator<Item = BlockNumber> + Send,
+) -> Result<Vec<BlockHeader>, DatabaseError> {
```
bobbinth, line 60: The comment here seems malformed (e.g., should `Note` actually be `Returns`?).

Also, would be good to mention that we impose a limit on the number of  requested block headers.

crates/store/src/db/models/queries/block_headers.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/// Select all block headers from the DB using the given [`SqliteConnection`].
+///
+/// # Returns
+///
+/// A vector of [`BlockHeader`] or an error.
+pub fn select_all_block_headers(
+    conn: &mut SqliteConnection,
+) -> Result<Vec<BlockHeader>, DatabaseError> {
```
bobbinth, line 85: I believe this function is used to load the MMR. It is fine for now, but in the future we may need to re-think it as a year worth of block headers would probably be around 2GB.

crates/store/src/db/models/queries/insertions.rs
bobbinth, line 1: I didn't do a thorough review of this file, but one question: is there a reason to have a single file for most insert statements? An alternative here would be to move inserts into relevant sections of their corresponding entities. For example, `insert_block_header()` could go into `queries/block_headers.rs`, `upsert_accounts()` would go into `queries/accounts.rs` etc.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -0,0 +1,686 @@
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Selectable, Queryable, QueryableByName)]
+#[diesel(table_name = schema::accounts)]
+#[diesel(check_for_backend(Sqlite))]
+pub struct AccountSummaryRaw {
+    account_id: Vec<u8>,         // AccountId,
+    account_commitment: Vec<u8>, //RpoDigest,
+    block_num: i64,              //BlockNumber,
```
bobbinth, line 670: nit: missing space in comments.

crates/store/src/db/models/queries/accounts.rs
bobbinth, line 1: I didn't do a through review of this file, in part because a lot of this will get refactored with the work that @igamigo is doing. But a few comments for the future:

- We should probably split this file up into multiple files (still under `accounts` namespace) - especially if we move account update/inserts functions into this namespace as well.
- Further down the road, we'll need to make sure none of the functions returns `Account` objects because these may be prohibitively expensive to load. Instead, we'd probably return something like `AccountHeader`.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -0,0 +1,712 @@
...
+use crate::errors::NoteSyncError;
+
+/// Select notes matching the tags and account IDs search criteria using the given
+/// [`SqliteConnection`].
+///
+/// # Returns
+///
+/// All matching notes from the first block greater than `block_num` containing a matching note.
+/// A note is considered a match if it has any of the given tags, or if its sender is one of the
+/// given account IDs. If no matching notes are found at all, then an empty vector is returned.
```
bobbinth, line 60: There is no `block_num` parameter any more - it is now `start_block_number` (but probably should be renamed to `block_from`).

crates/store/src/db/models/queries/notes.rs
```diff
@@ -0,0 +1,712 @@
...
+/// # Note
+///
+/// This method returns notes from a single block. To fetch all notes up to the chain tip,
+/// multiple requests are necessary.
+pub(crate) fn select_notes_since_block_by_tag_and_sender(
+    conn: &mut SqliteConnection,
+    start_block_number: BlockNumber,
+    account_ids: &[AccountId],
+    note_tags: &[u32],
+) -> Result<Vec<NoteSyncRecord>, DatabaseError> {
```
bobbinth, line 71: We should mention in doc comments that we impose the limit on the number of account IDs and note tags.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -0,0 +1,712 @@
...
+    //         FROM
+    //             notes
+    //         WHERE
+    //             (tag IN rarray(?1) OR sender IN rarray(?2)) AND
+    //             block_num > ?3
+    //         ORDER BY
+    //             block_num ASC
+    //     LIMIT 1) AND
+    //     -- filter the block's notes and return only the ones matching the requested tags or
+    // senders     (tag IN rarray(?1) OR sender IN rarray(?2))
```
bobbinth, line 101: I would probably add a comment to say that we actually execute this as two separate queries: first query to get the block number, and the second query to get the relevant data from that block (maybe worth splitting this up into two separate functions?).

Also, comment for the future: this is probably one of the more complex queries and it is also on a very hot path (i.e., it would be executed on every state sync request which each connected client may be done once every few seconds). We should think about how we can optimize it as much as possible. Maybe we should create a dedicated issue for this.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -0,0 +1,712 @@
...
+            .get_results::<NoteSyncRecordRawRow>(conn)
+            .map_err(DatabaseError::from)?;
+
+    vec_raw_try_into(notes)
+}
+
+pub(crate) fn select_notes_by_id(
+    conn: &mut SqliteConnection,
+    note_ids: &[NoteId],
+) -> Result<Vec<NoteRecord>, DatabaseError> {
```
bobbinth, line 148: We should add doc comments to this function.

Also, it doesn't seem like we impose a limit on the number of requested notes, but we probably should?

crates/store/src/db/models/queries/notes.rs
```diff
@@ -0,0 +1,712 @@
...
+}
+
+/// Select all notes from the DB using the given [`SqliteConnection`].
+///
+///
+/// # Returns
+///
+/// A vector with notes, or an error.
+#[cfg(test)]
+pub(crate) fn select_all_notes(
```
bobbinth, line 177: I would maybe move test-specific functions into a separate section in the file.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -0,0 +1,712 @@
...
+/// Select note inclusion proofs matching the `NoteId`, using the given [`SqliteConnection`].
+///
+/// # Returns
+///
+/// - Empty map if no matching `note`.
+/// - Otherwise, note inclusion proofs, which `note_id` matches the `NoteId` as bytes.
+pub(crate) fn select_note_inclusion_proofs(
+    conn: &mut SqliteConnection,
+    note_ids: &BTreeSet<NoteId>,
+) -> Result<BTreeMap<NoteId, NoteInclusionProof>, DatabaseError> {
```
bobbinth, line 207: We should mention in the doc comments that we impose a limit on the number of note IDs.

Also, we should probably start using `HashMap` and `HashSet` to improve performance.

crates/store/src/db/models/queries/notes.rs
bobbinth, line 1: I would maybe split this file up further (still under `notes` namespace). For example, we could have the definitions of raw records in the `notes/mod.rs` file, queries related to network notes `notes/network_notes.rs` file etc.

PR #1151, opened 2025-08-12
feat: Handle panics gracefully

### LINKED ISSUE #1117: Panic handler
Create and attach panic handler(s) to our runtimes.

In #1115 our gRPC method crashed during parsing. This resulted in the failure not being logged at all which in turn meant we assumed the request never reached the node during initial investigation.

The panic handler should at minimum emit an error event with a full error report attached as the message. This needs to follow OTel standards so it would make sense to check if there is an otel helper for something like this in one of the otel/trace crates.

sergerad:
## Context

Closes #1117.

## Changes

WIP


crates/rpc/src/server/api.rs
```diff
@@ -445,35 +446,39 @@ impl api_server::Api for RpcService {
...
-        Ok(Response::new(proto::rpc::RpcStatus {
-            version: env!("CARGO_PKG_VERSION").to_string(),
-            store: store_status.or(Some(proto::rpc_store::StoreStatus {
-                status: "unreachable".to_string(),
-                chain_tip: 0,
-                version: "-".to_string(),
-            })),
-            block_producer: block_producer_status.or(Some(
-                proto::block_producer::BlockProducerStatus {
+        handle_panic(async {
```
sergerad: @Mirko-von-Leipzig @bobbinth I tried turning this into a procedural macro but had issues with async in the macro fn.

Once we settle on how to handle panics I can go back to trying to make the macro work.

crates/rpc/src/server/api.rs
```diff
@@ -10,12 +10,10 @@ use miden_node_proto::generated::{self as proto};
...
 use miden_node_utils::ErrorReport;
 use miden_node_utils::limiter::{
-    QueryParamAccountIdLimit,
-    QueryParamLimiter,
-    QueryParamNoteIdLimit,
-    QueryParamNoteTagLimit,
+    QueryParamAccountIdLimit, QueryParamLimiter, QueryParamNoteIdLimit, QueryParamNoteTagLimit,
     QueryParamNullifierLimit,
 };
+use miden_node_utils::panic::handle_panic;
```
sergerad: @Mirko-von-Leipzig initial thinking is that all external-facing RPC endpoints should handle panics. Unsure if we want interal ones to. WDYT?

Mirko-von-Leipzig: I think maybe all panics so that they get logged in honeycomb at least. Otherwise we'll never be able to see the internal panic message without trying to find them on the machine's `journald` logs somehow. And finding them is going to be next to impossible imo.

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: I think maybe take a look at using middleware for this instead. For example https://docs.rs/tower-http/latest/tower_http/catch_panic/index.html.

This would require some additional work to make gRPC compliant but I think then at least its applicable at the server/layer level and not per method?

Mirko-von-Leipzig: That said we would still need a general tokio panic handler maybe? To catch non-gRPC/server panics.. and I'm unsure how they would interact. Maybe we can't use a general panic handler then.

crates/rpc/src/server/api.rs
sergerad: Makes sense. So RPC stack actually recovers from panic while the other stacks just log the panic before exiting?

crates/rpc/src/server/api.rs
sergerad: I imagine we should be able to have a `main()` level handler for logging panics before exiting alongside the RPC level one which stops unwinding well before `main()`.

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: Thinking on it some more; I think we already handle panics in most cases since we spawn tasks and then join them and then exit or restart them.

Exceptions to this would be any axum/tonic servers and anywhere we might spawn and detach a background task. I'm not aware of any of the latter, and that wouldn't be good practice.

So maybe its just
- add middleware for the servers
  - it might be good enough to hook into the error tracing middleware to log things - depends how they propagate errors/panics :shrug: 
- register a global panic handler (and for tokio?), but check that it only captures unhandled panics

sergerad: Yea for example from ntx builder:
```

Caused by:
    0: Joining component task
    1: task 20 panicked with message "attempt to subtract with overflow"
```

crates/panic-macro/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+    let original_body = func.block;
+
+    func.block = Box::new(syn::parse_quote!({
+        miden_node_utils::panic::handle_panic(async move {
+            #original_body
+        }).await?
+    }));
+
+    quote::quote!(#func).into()
+}
```
sergerad: This fn is purely from prompt so if anyone is versed in procedural macros, please critique.

crates/rpc/src/server/api.rs
sergerad: So panics captured by tokio should be handled by us as errors already. Is there any part of the stack where you think we may have not handled such cases appropriately?

crates/panic-macro/src/lib.rs
sergerad: I updated it to provide validation. If we prefer, could keep it simpler without validation:
```rust
#[proc_macro_attribute]
pub fn main(_attr: TokenStream, item: TokenStream) -> TokenStream {
    let mut func = syn::parse_macro_input!(item as syn::ItemFn);
    let original_body = func.block;

    func.block = Box::new(syn::parse_quote!({
        miden_node_utils::panic::handle_panic(async move {
            #original_body
        }).await?
    }));

    quote::quote!(#func).into()
}
```

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: The only _known_ unlogged cases would be the http/grpc servers.

All other tasks/threads _should_ be joined and handled/aborted somewhere. The only reason to consider an additional global panic handler (and tokio panic handler) is as a backup for unknowns.

bin/remote-prover/src/commands/update_workers.rs
```diff
@@ -69,35 +69,31 @@ impl UpdateWorkers {
...
+        let client = Client::builder().http2_prior_knowledge().build()?;
 
         // Make the request
-        let response = client.get(url).send().await.map_err(|err| err.to_string())?;
+        let response = client.get(url).send().await?;
 
         // Check status code
         if !response.status().is_success() {
-            return Err(format!("Request failed with status code: {}", response.status()));
+            return Err(anyhow::anyhow!("Request failed with status code: {}", response.status()));
```
Mirko-von-Leipzig:
```suggestion
            anyhow::bail!("Request failed with status code: {}", response.status());
```

bin/remote-prover/src/commands/update_workers.rs
```diff
@@ -69,35 +69,31 @@ impl UpdateWorkers {
...
         }
 
         // Read the X-Worker-Count header
         let workers_count = response
             .headers()
             .get("X-Worker-Count")
-            .ok_or("Missing X-Worker-Count header")?
-            .to_str()
-            .map_err(|err| err.to_string())?;
+            .ok_or(anyhow::anyhow!("Missing X-Worker-Count header"))?
```
Mirko-von-Leipzig: `.context` actually works on options directly as well


```suggestion
            .context("Missing X-Worker-Count header")?
```

Mirko-von-Leipzig (commented):
I'm now wondering if I haven't led us down the wrong path.

As in; we shouldn't expect to have panics and therefore a general panic handler with `set_hook` that logs the panic as an `error!` is sufficient to let us know we have something to address?

I think my question would be what happens to the gRPC panics - do those still meet the panic hook, or do we need the middleware to properly manage these? Is the connection simply dropped, or does axum/tonic send an internal error?

sergerad:
> I think my question would be what happens to the gRPC panics - do those still meet the panic hook, or do we need the middleware to properly manage these? Is the connection simply dropped, or does axum/tonic send an internal error?

If I panic in main and the RPC endpoint, we see these logs
```
2025-08-14T22:53:57.948274Z ERROR miden_node: bin/node/src/main.rs:80: panic hook
2025-08-14T22:46:27.875358Z ERROR miden_node_utils::panic: crates/utils/src/panic.rs:35: panic occurred: panic status 123
```
This means that the hook was set off and the middleware was too. The middleware caught the panic and returned a 500 response so that works fine.

Panic in main:
```
2025-08-14T22:53:57.948274Z ERROR miden_node: bin/node/src/main.rs:80: panic hook
```

So yea, the panic hook is enough for global catch all.

crates/rpc/src/server/api.rs
sergerad: Have updated to use set_hook. I don't follow what you mean by tokio panic handler or why that would be required (tokio should catch panics for us and return as errors which we handle already).

crates/utils/src/logging.rs
```diff
@@ -53,7 +54,13 @@ pub fn setup_tracing(otel: OpenTelemetry) -> Result<()> {
     let subscriber = Registry::default()
         .with(stdout_layer().with_filter(env_or_default_filter()))
         .with(otel_layer.with_filter(env_or_default_filter()));
-    tracing::subscriber::set_global_default(subscriber).map_err(Into::into)
+    tracing::subscriber::set_global_default(subscriber).map_err(Into::<anyhow::Error>::into)?;
+
+    // Register panic hook now that tracing is initialized.
+    std::panic::set_hook(Box::new(|info| {
+        tracing::error!("{info}");
+    }));
```
Mirko-von-Leipzig: Should we include something more specific so we can filter for these?

I'm not entirely sure how to do this but maybe we should include a `panic = true` field? Then we can filter by it.

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: >  I don't follow what you mean by tokio panic handler or why that would be required (tokio should catch panics for us and return as errors which we handle already).

Because that's only true if we keep and join the handle. A common pattern is to spawn and forget the task at which point it will just do... what? I hope it still triggers the panic hook?

crates/utils/src/panic.rs
Mirko-von-Leipzig, line 1: Do we still need this if the general panic hook includes these already?

I'm tempted to just drop this as well. iiuc the downside is that we can't control the user experience here i.e. it might result in a connection drop and not an error response. But given that this is supposed to be an exceptional circumstance, maybe that's good enough so long as the panic message is informative?

Mirko-von-Leipzig (approved):

crates/utils/src/panic.rs
```diff
@@ -0,0 +1,33 @@
...
+pub fn catch_panic_layer_fn(err: Box<dyn Any + Send + 'static>) -> Response<Full<bytes::Bytes>> {
+    // Log the panic error details.
+    let err = stringify_panic_error(err);
+    tracing::error!("panic: {err}");
+
+    // Return generic error response.
+    Response::builder()
+        .status(StatusCode::INTERNAL_SERVER_ERROR)
+        .header(header::CONTENT_TYPE, "application/grpc")
+        .body(Full::from(""))
```
bobbinth, line 20: Question: why not return `err` here as a part of the body?

bobbinth (approved):
Looks good! Thank you! Though this is not a very deep review from me.

Is there anything outstanding left in this PR?

crates/utils/src/panic.rs
sergerad: Yes, if there is a possibility that a panic in the RPC stack leaves it in a broken state, then we should not just catch the stack unwind and continue. Ideally we would catch, respond, and signal shutdown of the entire process so that all existing responses are completed before shutdown initiates.

The benefit of the way its implemented here is that the node could not be DDOS'd via some panic bug.

In the context of the node and the exceptional circumstances of a panic, I think its probably fine from a UX perspective to just hang up on connections when we do panic.

@bobbinth any preferences?

crates/utils/src/panic.rs
sergerad: Vaguely - security through obscurity (DDOS bug/exploit). And maybe also more confusing for client side to receive the panic message, depending on its nature. Happy to add it in the response if we keep this functionality.

crates/utils/src/panic.rs
bobbinth: No strong preferences from me - mostly because I don't fully understand pros/cons of having the general panic handlers vs. not. I think the main goal here was to make sure we log the panics (i.e., we don't have situations where the node went down but we don't know why) - but beyond that, what does the general panic hook give us?

sergerad: The panic hook achieves the logging.

The middleware that catches stack unwind from panic allows us to return responses to requests that initiated the panicking code path.

Its the latter functionality that we are considering not merging. If we don't merge it, then we will continue to just hang up on all requests and exit the process when RPC stack panics after receiving a relevant request.

bobbinth: And with the middleware we'll catch the panic and be able to return something like a 500 response (or w/e the equivalent in gRPC is)? And the panic would not bring down the RPC process?

sergerad: That's right

bobbinth: Not a strong preference, but returning 500 to the user is nice. Whether the process should continue after the panic, I'm not so sure. If a panic happens, it is likely that the process may be in an inconsistent state - though, on the RPC side, there isn't much state to talk about. So, I think both should be fine, but maybe I still have a slight preference toward crushing on panic.

crates/rpc/src/server/api.rs
sergerad: Yes the panic hook will still go off

sergerad: Tried locally quickly, see the hook
```
2025-08-18T00:04:51.097590Z ERROR miden_node_utils::logging: crates/utils/src/logging.rs:61: panicked at bin/node/src/main.rs:79:9:
task, panic: true
```

crates/utils/src/panic.rs
sergerad: I've merged it with the middleware in place for now. If we think its worthwhile we can make a followup issue to determine whether RPC panics should trigger shutdown or not, and if they do allow for graceful shutdown. Probably low priority.

sergerad: I will add alerts based on the field panic=true too.

PR #1154, opened 2025-08-13
chore: add store startup benchmark

### LINKED ISSUE #1139: Add benchmark of the store startup
A new subcommand should be added to the stress-test CLI, testing how long it takes to start up a store with a pre-seeded database.

We saw in previous benchmarks that the startup times can increase significantly as the DB is seeded, since before the service is functional we need to wait for the nullifiers and accounts trees to be loaded.

SantiagoPittella:
closes #1139


crates/store/src/state.rs
```diff
@@ -111,10 +113,25 @@ pub struct State {
...
+                    format!("failed to load block store at {}", data_directory.display())
+                })
+                .unwrap(),
+        );
+
+        let database_filepath = data_directory.database_path();
+        let mut db = Db::load(database_filepath.clone())
+            .await
+            .with_context(|| format!("failed to load database at {}", database_filepath.display()))
+            .unwrap();
```
Mirko-von-Leipzig: These errors should be propagated and not unwrapped

crates/store/src/state.rs
```diff
@@ -1054,7 +1071,21 @@ async fn load_mmr(db: &mut Db) -> Result<Mmr, StateInitializationError> {
...
+        .map_err(StateInitializationError::FailedToCreateAccountsTree)?;
+
+    let elapsed = now.elapsed().as_secs();
+
+    info!(
+        num_of_leaves = len,
+        tree_construction = elapsed,
+        COMPONENT,
+        "Loaded account tree"
+    );
```
Mirko-von-Leipzig: Do we want to keep this? `elapsed` is also unused I think so it makes me think this was used while developing only?

SantiagoPittella: We are using the `elapsed` only in the `info!`. I don't have any strong preference about this, I'll remove it now but if anyone prefer to have it back I can rollback the change.

crates/store/src/state.rs
```diff
@@ -1021,22 +1031,9 @@ impl State {
...
-    info!(
-        num_of_leaves = len,
-        tree_construction = elapsed,
-        COMPONENT,
-        "Loaded nullifier tree"
-    );
-    Ok(nullifier_tree)
+
+    NullifierTree::with_entries(nullifiers.into_iter().map(|info| (info.nullifier, info.block_num)))
+        .map_err(StateInitializationError::FailedToCreateNullifierTree)
```
bobbinth, line 1037: We can simplify this because we already instrument the entire function - right?

If so, should we change the `level` from "debug" to info"? Since this is executed only on startup, I don't think this would create too much "noise".

I'd do the same with `load_mmr` and `load_accounts` (which we should probably rename into `load_account_tree`).

bin/stress-test/src/store/mod.rs
```diff
@@ -288,6 +289,16 @@ async fn check_nullifiers_by_prefix(
...
 
+// LOAD STATE
+// ================================================================================================
+
+pub async fn load_state(data_directory: &Path) {
+    let start = Instant::now();
+    let _state = State::load(data_directory).await.unwrap();
+    let elapsed = start.elapsed();
+    println!("State loaded in {elapsed:?}");
+}
```
bobbinth: This works. Another approach would be start the store service and then wait until we get a response from the `status` endpoint - right? I'm assuming this would be more complicated though?

crates/store/src/state.rs
```diff
@@ -111,10 +111,20 @@ pub struct State {
...
+        );
+
+        let database_filepath = data_directory.database_path();
+        let mut db = Db::load(database_filepath.clone())
+            .await
+            .map_err(StateInitializationError::DatabaseLoadError)?;
+
         let nullifier_tree = load_nullifier_tree(&mut db).await?;
         let chain_mmr = load_mmr(&mut db).await?;
         let account_tree = load_accounts(&mut db).await?;
```
bobbinth: Not from this PR, but I'd probably change the order to be:

1. Load MMR.
2. Load account tree.
3. Load nullifier tree.

crates/store/src/lib.rs
```diff
@@ -1,11 +1,11 @@
 use std::time::Duration;
 
 mod blocks;
-mod db;
+pub mod db;
```
bobbinth: Why does `db` need to be `pub`?

bin/stress-test/README.md
```diff
@@ -22,6 +22,7 @@ The endpoints that you can test are:
 - `sync_state`
 - `sync_notes`
 - `check_nullifiers_by_prefix`
+- `load_state`
```
bobbinth: nit: I would move the `load_state` to the top of the list.

bin/stress-test/README.md
```diff
@@ -22,6 +22,7 @@ The endpoints that you can test are:
...
 
 Each benchmark accepts options to control the number of iterations and concurrency level.
```
bobbinth: I don't think concurrency applies to `load_store` - right? If so, I'd re-phrase this a bit.

bin/stress-test/README.md
```diff
@@ -138,5 +139,12 @@ P99.9 request latency: 2.289709ms
 Average nullifiers per response: 21.0348
 ```
 
+- load-state
+``` bash
+$ miden-node-stress-test benchmark-store --data-directory ./data load-state
+
+State loaded in 41.494029542s
+```
```
bobbinth: It would be great to make it clear how big the underlying state is. On line 71 we say that the DB has about 100K accounts and 100K notes. On line 45 we say that there are 1M accounts (which I think is incorrect). And I don't think we ever say how many nullifiers are in the store.

It may make sense to have a small paragraph at the beginning of the "Results" section to explain the dataset which will be used for all benchmarks.

------

Separately, if we turn on tracing, could we get info on how long it takes to load the MMR, account tree, and nullifier tree separately? Would be great to put this into a table.

bobbinth (commented):

bin/stress-test/src/store/mod.rs
Mirko-von-Leipzig: We could also move towards using the telemetry span information though we should think about _how_ we would want to represent that.

A benefit to this is that we could also use honeycomb or whichever backend to investigate slowdowns without having to debug directly - so long as we have the appropriate pieces instrumented.

But I would leave that for a future improvement.

bin/stress-test/README.md
Mirko-von-Leipzig: > Separately, if we turn on tracing, could we get info on how long it takes to load the MMR, account tree, and nullifier tree separately? Would be great to put this into a table.

We would have a couple of options

1. `tracing` but we need to ensure we also log the close event which would include the overall time
2. `tracing` with the forest one which creates those nicer looking span trees
3. Using OTel and collect the spans ourselves to print nicely. Or send to honeycomb on a separate environment to also allow inspecting whats happening

crates/store/src/lib.rs
SantiagoPittella: I used to need it before a refactor that I done to the `State::load` method, it should be safe to remove now.

bin/stress-test/README.md
SantiagoPittella: Answering the second part, and related to above's comment ( https://github.com/0xMiden/miden-node/pull/1154#discussion_r2282476453 ). This is the current result when running the command with the default tracing config:

```
2025-08-19T18:11:31.220824Z  INFO load:load: miden-store: crates/store/src/db/mod.rs:227: Connected to the database, sqlite: ./data/miden-store.sqlite3
2025-08-19T18:11:31.223389Z  INFO miden_node_store::db::migrations: crates/store/src/db/migrations.rs:17: Applying 0 migration(s), target: "miden-store"
2025-08-19T18:11:31.223437Z  INFO load:load: miden-store: crates/store/src/db/mod.rs:222: close, time.busy: 574µs, time.idle: 2.21ms
2025-08-19T18:11:31.239363Z  INFO load:load_mmr: miden-store: crates/store/src/state.rs:1039: close, time.busy: 2.11ms, time.idle: 13.8ms
2025-08-19T18:11:52.637285Z  INFO load:load_account_tree: miden-store: crates/store/src/state.rs:1051: close, time.busy: 21.3s, time.idle: 82.6ms
2025-08-19T18:12:14.179202Z  INFO load:load_nullifier_tree: miden-store: crates/store/src/state.rs:1031: close, time.busy: 21.5s, time.idle: 79.4ms
2025-08-19T18:12:14.179247Z  INFO load: miden-store: crates/store/src/state.rs:113: close, time.busy: 42.8s, time.idle: 178ms
State loaded in 42.959271667s
Database contains 99961 accounts and 99960 nullifiers
```

It prints the detail of each step: `load_mmr`, `load_account_tree` and `load_nullifier_tree`.

SantiagoPittella: regarding the addition of the nullifiers and account counts at the end, since the `db` field is private, and the homonym module too, I couldn't query the DB to get the counts. I added a call to the `sqlite3` command.

crates/store/src/lib.rs
SantiagoPittella: I got it out, but it's giving me a problem now that the `State` module is public, the documentation of one of its methods (`get_notes_by_id`) fails when running `make doc`, since it refers to `NoteRecord`, and it's from the `Db` module.

Not really sure which path take here, the crate doesn't have features so I don't want to create one, but moving the public visibility of the `State` under a feature fixes the issue. Making Db public also fixes it.

crates/store/src/state.rs
```diff
@@ -62,6 +61,7 @@ use crate::errors::{
     StateInitializationError,
     StateSyncError,
 };
+use crate::{COMPONENT, DataDirectory};
 // STRUCTURES
```
bobbinth: nit: I would add a linebreak between these two lines.

crates/store/src/lib.rs
bobbinth: I think we can just modify the doc comment not to link to `NoteRecord`.

bin/stress-test/README.md
bobbinth: Looking at the tracing results, we basically have:

- Account and nullifier tree take about 21 seconds to load each.
- Everything else (including the MMR loading) is basically negligible.

I would probably add a comment about this somewhere next to the benchmark.

bobbinth (approved):
Looks good! Thank you! I left a couple of small comments inline. After these are addressed (and we fix the lints), let's merge.

PR #1155, opened 2025-08-13
fix: Fix underflow

sergerad:
Bootstrapping a new node leads to error caused by panic due to arithmetic underflow:

```
   ./target/debug/miden-node bundled bootstrap --data-directory /tmp/1 --accounts-directory /tmp/1
   ./target/debug/miden-node bundled start --rpc.url http://0.0.0.0:57291 --data-directory /tmp/1
```

will cause
```

Caused by:
    0: Joining component task
    1: task 20 panicked with message "attempt to subtract with overflow"
```

This fixes the underflow.

### DIFF
--- a/crates/ntx-builder/src/state/mod.rs
+++ b/crates/ntx-builder/src/state/mod.rs
@@ -204,7 +204,7 @@ impl State {
 
         // Keep MMR pruned.
         let pruned_block_height =
-            (self.chain_mmr.chain_length().as_usize() - MAX_BLOCK_COUNT) as u32;
+            (self.chain_mmr.chain_length().as_usize().saturating_sub(MAX_BLOCK_COUNT)) as u32;
         self.chain_mmr.prune_to(..pruned_block_height.into());
     }
 


crates/ntx-builder/src/state/mod.rs
```diff
@@ -204,7 +204,7 @@ impl State {
 
         // Keep MMR pruned.
         let pruned_block_height =
-            (self.chain_mmr.chain_length().as_usize() - MAX_BLOCK_COUNT) as u32;
+            (self.chain_mmr.chain_length().as_usize().saturating_sub(MAX_BLOCK_COUNT)) as u32;
```
sergerad, line 207: @bobbinth I preferred not to put a `if  num_blocks > MAX_BLOCK_COUNT` because that logic would only be relevant for at the first few blocks in the entirety of the chain. This does mean that the ntx builder's chain_mmr won't have MAX_BLOCK_COUNT blocks until the chain height is MAX_BLOCK_COUNT*2 but I think that is fine?

Mirko-von-Leipzig: Maybe I'm misunderstanding but I think your fix is already as good as it can be?

`pruned_block_height` will remain 0 until the chain length exceeds `MAX_BLOCK_COUNT` which is correct. The chain mmr should reach `MAX_BLOCK_COUNT` normally and only then be pruned.

Mirko-von-Leipzig (approved):

crates/ntx-builder/src/state/mod.rs
sergerad: Yea you are right. I always struggle to grok this
```rust
pub fn prune_to(&mut self, to: RangeTo<BlockNumber>)

/// Drop every block header whose number is strictly less than to.end.
```

Issue #1153, opened 2025-08-13
Post `async-experimental` merge cleanup


drahnr:
* [ ] move `fn test_fee()` and `fn test_fee_params()` into the `Mock*Chain` in `miden-base`
* [x] `GenesisConfig` could make better use of `TokenSymbol` - must implement `Hash` and possibly `serde::{Ser..,De..}`
* [x] `GenesisConfig` definition of native asset faucet could be improved https://github.com/0xMiden/miden-node/issues/1149

drahnr:
>  move fn test_fee() and fn test_fee_params() into the Mock*Chain in miden-base

This doesn't quite work as anticipated, since most, calls are around the `MockTransactionBuilder`, which already sets up the default.
Now this leaves some dummy blocks being built, but we don't have mocking features, so I'll ignore that one. https://github.com/0xMiden/miden-base/pull/1737 eases any possible future pain a bit.

PR #1156, opened 2025-08-13
introduce `TokenSymbolStr` for `GenesisConfig` representation

drahnr:
Introduces a `TokenSymbolStr` to represent the `TokenSymbol` from `base`, which in turn is limited by `winter-fell`.

The implementation uses the raw representation (`String`) and the parsed version (`TokenSymbol`). Now this allows us to stay in-sync with the definition of `TokenSymbol` (have the cake) as well as have the impl of `serde::de::Deserialize` and `Hash`, needed for state construction (eat it too). 

First part of https://github.com/0xMiden/miden-node/issues/1153

### DIFF
--- a/crates/store/src/genesis/config/mod.rs
+++ b/crates/store/src/genesis/config/mod.rs
@@ -57,6 +57,7 @@ pub struct GenesisConfig {
 
 impl Default for GenesisConfig {
     fn default() -> Self {
+        let miden = TokenSymbolStr::from_str("MIDEN").unwrap();
         Self {
             version: 1_u32,
             timestamp: u32::try_from(
@@ -68,14 +69,14 @@ impl Default for GenesisConfig {
             .expect("Timestamp should fit into u32"),
             wallet: vec![],
             fee_parameters: FeeParameterConfig {
-                symbol: TokenSymbolStr::from_str("MIDEN").unwrap(),
+                symbol: miden.clone(),
                 verification_base_fee: 0u32,
             },
             fungible_faucet: vec![FungibleFaucetConfig {
                 max_supply: 100_000_000_000_000_000u64,
                 decimals: 6u8,
                 storage_mode: StorageMode::Public,
-                symbol: TokenSymbolStr::from_str("MIDEN").unwrap(),
+                symbol: miden.clone(),
             }],
         }
     }


crates/store/src/genesis/config/mod.rs
```diff
@@ -478,3 +474,91 @@ fn prepare_fungible_asset_update(
...
+
+use serde::de::Visitor;
+
+struct TokenSymbolVisitor;
+
+impl<'de> Visitor<'de> for TokenSymbolVisitor {
+    type Value = TokenSymbolStr;
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+        formatter.write_str("a 3 to 5 ascii digits")
```
```suggestion
        formatter.write_str("3 to 5 ascii digits")
```

Mirko-von-Leipzig: Looking at `TokenSymbol` comments it seems to be `1 to 6` ascii digits there?

crates/store/src/genesis/config/mod.rs
```diff
@@ -66,14 +68,14 @@ impl Default for GenesisConfig {
             .expect("Timestamp should fit into u32"),
             wallet: vec![],
             fee_parameters: FeeParameterConfig {
-                symbol: "MIDEN".to_owned(),
+                symbol: TokenSymbolStr::from_str("MIDEN").unwrap(),
```
Mirko-von-Leipzig: We could make this an associated `const` though maybe this is less useful once the native faucet becomes its own type (since the constant would likely be there instead? maybe?).

Mirko-von-Leipzig (approved):
Pity the underlying types aren't more hash friendly, even `Ord` would at least let us use a `BTree` based collection.

crates/store/src/genesis/config/mod.rs
drahnr: We'd have to either use `lazy_static!` / `once_cell` or create a `const fn` constructor for `TokenSymbolStr` _and_ `TokenSymbol`, at which point it looks a bit like overkill

Mirko-von-Leipzig: Ah right. I guess we could make `TokenString` more const friendly but that's probably unnecessary then.

cc @PhilippGackstatter though its probably a bit of work for not all thaaaat much gain. That said it would be nice as part of a const Miden native faucet declaration.

PhilippGackstatter: I looked into string validation a bit for `const fn` `SlotName::from_static_str` (for https://github.com/0xMiden/miden-base/issues/1724) and the std library enables it, but it's not quite as nice as, e.g. using a `str::chars` iterator. But since this is coming up here as well, perhaps it'd be nice to leverage a crate like [`konst`](https://docs.rs/konst/latest/konst/) or something else for both a `TokenSymbol::from_static_str` and `SlotName::from_static_str`. (Haven't looked into that crate a lot yet, so if you have any alternative suggestions for const string validation, please let me know). Also, maybe not using a crate is also fine, idk.

I guess I'd see how it goes for `SlotName` and then possibly apply that to `TokenSymbol` as a follow-up.

Mirko-von-Leipzig: I've only ever done it manually for hex strings. `konst` looks pretty nice at first glance.

PhilippGackstatter: Opened https://github.com/0xMiden/miden-base/issues/1738.

drahnr: I'd leave this for a future PR

Issue #1149, opened 2025-08-11
Specialize native faucet configuration


Mirko-von-Leipzig:
Simplify the native faucet specification in the genesis configuration.

#1122 added basic fee support to the node. As part of this it added support for specifying the native asset in the genesis configuration file. At the moment this asset's faucet is specified as a conventional basic fungible faucet, however the native faucet _must_ 

- use public storage, and
- have exactly one native faucet declared

A better configuration would be specializing the native faucet as suggested [here](https://github.com/0xMiden/miden-node/pull/1143#discussion_r2265981740), which would enforce that this native faucet exists and prevent storage from being configured.

In addition, this should also be carried through from the genesis config file to the actual state creation i.e. it should take in a native faucet in addition to the extra, arbitrary accounts.

PR #1157, opened 2025-08-14
[refactor] genesis config

drahnr:
Addresses https://github.com/0xMiden/miden-node/issues/1149

### DIFF
--- a/crates/store/src/genesis/config/mod.rs
+++ b/crates/store/src/genesis/config/mod.rs
@@ -137,7 +137,7 @@ impl GenesisConfig {
 
         let native_faucet_account_id = faucet_accounts
             .get(&symbol)
-            .expect("Parsing guarnatees the existence of a native faucet.")
+            .expect("Parsing guarantees the existence of a native faucet.")
             .id();
 
         let fee_parameters =


crates/store/src/genesis/config/mod.rs
```diff
@@ -304,6 +276,39 @@ impl GenesisConfig {
...
+    symbol: TokenSymbolStr,
+
+    decimals: u8,
+    /// Max supply in full token units
+    ///
+    /// It will be converted internally to the smallest representable unit,
+    /// using based `10.powi(decimals)` as a multiplier.
+    max_supply: u64,
+    /// Fee parameters
+    #[serde(alias = "fee")]
```
Mirko-von-Leipzig: I wouldn't introduce an alias except for backwards compatibility reasons.

crates/store/src/genesis/config/mod.rs
```diff
@@ -475,6 +478,43 @@ fn prepare_fungible_asset_update(
     Ok(wallet_asset_delta)
 }
 
+/// Create a fungible faucet from a config entry
+fn create_fungible_faucet(
+    config: FungibleFaucetConfig,
```
Mirko-von-Leipzig: Could we make this a method on the type instead of a free floating method?

```rust
fn build_account(self) -> .. {}
```

crates/store/src/genesis/config/mod.rs
```diff
@@ -475,6 +478,43 @@ fn prepare_fungible_asset_update(
...
+        .expect("The `Felt::MODULUS` is _always_ larger than the `max_supply`");
+
+    let component = BasicFungibleFaucet::new(*symbol.as_ref(), decimals, max_supply)?;
+
+    let account_storage_mode = storage_mode.into();
+
+    // It's similar to `fn create_basic_fungible_faucet`, but we need to cover more cases.
+    let (faucet_account, faucet_account_seed) = AccountBuilder::new(init_seed)
+        .account_type(account_type)
+        .storage_mode(account_storage_mode)
```
Mirko-von-Leipzig: I find the extra indirection more confusing since there is no real work being done

```suggestion
        .account_type(AccoutType::FungibleFaucet)
        .storage_mode(storage_mode.into())
```

crates/store/src/genesis/config/mod.rs
```diff
@@ -475,6 +478,43 @@ fn prepare_fungible_asset_update(
...
+fn create_fungible_faucet(
+    config: FungibleFaucetConfig,
+) -> Result<(Account, Word, SecretKey), GenesisConfigError> {
+    let FungibleFaucetConfig {
+        symbol,
+        decimals,
+        max_supply,
+        storage_mode,
+    } = config;
+    let mut rng = ChaCha20Rng::from_seed(rand::random());
```
Mirko-von-Leipzig: Is there value in having this passed in instead? Probably not at the moment and I guess if we want consistency for tests we can always change it then

crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -1,16 +1,19 @@
...
 version   = 1
 
+[native_faucet]
+symbol     = "MIDEN"
+decimals   = 3
+max_supply = 100_000_000
 
-[fee_parameters]
-symbol                = "MIDEN"
+[native_faucet.fee_parameters]
```
Mirko-von-Leipzig: I'm unsure if we want to nest the fee parameters in the native faucet specifically, or if it shouldn't be the other way around. Or not connected at all.

cc @bobbinth

crates/store/src/genesis/config/mod.rs
```diff
@@ -112,37 +111,23 @@ impl GenesisConfig {
...
+        let (native_faucet_account, native_faucet_account_seed, native_secret_key) =
+            create_fungible_faucet(native_faucet.to_faucet_config())?;
 
-            let max_supply = Felt::try_from(max_supply)
-                .expect("The `Felt::MODULUS` is _always_ larger than the `max_supply`");
+        faucet_accounts.insert(symbol.clone(), native_faucet_account.clone());
 
-            let component = BasicFungibleFaucet::new(*symbol.as_ref(), decimals, max_supply)?;
+        secrets.push((
+            format!("native_faucet_{symbol}.mac", symbol = symbol.to_string().to_lowercase()),
```
Mirko-von-Leipzig: I'm unsure if we want or need to special case the native faucet. To use it programmatically one always needs to know its symbol in which case `faucet_{symbol}.mac` is just as normal.

Mirko-von-Leipzig: Mostly I don't like the code dup but maybe its not really an issue.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - [BREAKING] De-duplicate storage of code in DB (no-migration) ([#1083](https://github.com/0xMiden/miden-node/issue/#1083)).
 - [BREAKING] RPC accept header format changed from `application/miden.vnd+grpc.<version>` to `application/vnd.miden; version=<version>` ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
 - [BREAKING] Integrated `FeeParameters` into block headers. ([#1122](https://github.com/0xMiden/miden-node/pull/1122)).
+- [BREAKING] Adjust the `GenesisConfig` toml representation ([#1157](https://github.com/0xMiden/miden-node/pull/1157)).
```
Mirko-von-Leipzig: Maybe something like

```suggestion
- [BREAKING] Genesis configuration now supports fees ([#1157](https://github.com/0xMiden/miden-node/pull/1157)).
  - Configure `NativeFaucet`, which determines the native asset used to pay fees
  - Configure the base verification fee
  - Note: fees are not yet activated, and this has no impact beyond setting these values in the block headers
```

crates/store/src/genesis/config/mod.rs
drahnr: I considered it, but we'd need one entry from `fungible_faucet_configs` (and only that) so it doesn't appear `self` is particularly useful, since we destructure it at the very top. We could slice it differently, i.e. processing all in an associated function, but then we'd have the code duplication issue again. The function is in one spot and exists to avoid repetition.

crates/store/src/genesis/config/samples/01-simple.toml
drahnr: My assumption from previous conversations were, there being
* one `native_faucet` per config file
* one `fee_params` for the `native_faucet`

crates/store/src/genesis/config/mod.rs
drahnr: No strong opinion, can also `std::iter::once(native.to_faucet_config()).chain(fungible_faucet_configs.iter())` and avoid the specialization ruckus

crates/store/src/genesis/config/mod.rs
drahnr: No explicit value, this was a question when `GenesisConfig` was initially introduced, since we use it in multiple places.
We could use `rand::thread_rng()` for simplicity, the above is a relatively safe bet though (rand seed and then CSPRNG) using independent start values for each account. The latter is the only important thing imho.

adr1anh: If I understand correctly, this would be about passing the rng to the `into_state` function to deterministically derive the account secret keys? Are we using the `GenesisConfig` outside of testing purposes though?

The only disadvantage I could see with passing the rng as argument is that it pushes the responsibility of choosing a safe randomness source to the caller, potentially making the generated accounts vulnerable to being hacked. We could instead pass in `Option<Seed>` which, if `None` would initialize a secure randomness source, while otherwise setting a deterministic source for testing purposes. 

I'll let @Al-Kindi-0 chime in though.

Mirko-von-Leipzig: > We could instead pass in Option<Seed> which, if None would initialize a secure randomness source, while otherwise setting a deterministic source for testing purposes.

I would much prefer an explicit `with_default_rng`, `with_seed`, etc instead of hiding the meaning

drahnr: > Are we using the GenesisConfig outside of testing purposes though?

We probably will, not as of today though.

Mirko-von-Leipzig: > > If I understand correctly, this would be about passing the rng to the into_state function to deterministically derive the account secret keys? Are we using the GenesisConfig outside of testing purposes though?
> >
> > The only disadvantage I could see with passing the rng as argument is that it pushes the responsibility of choosing a safe randomness source to the caller, potentially making the generated accounts vulnerable to being hacked. 

The node is an executable so we're completely in charge of the caller etc. The question is just where do we move it to. The only insecure option imo would be if we allow users to configure the rng somehow but I don't see a reason to do that.

I think we can probably just leave this as is; if we want this deterministic we can always move it one level higher.

crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig: I'm not quite following, surely these are equivalent in everything except ownership?

```rust
fn create_fungible_faucet(
    config: FungibleFaucetConfig,
) -> Result<(Account, Word, SecretKey), GenesisConfigError>

impl FungibleFaucetConfig {
    fn build_account(self) -> Result<(Account, Word, SecretKey), GenesisConfigError>
}
```

crates/store/src/genesis/config/mod.rs
Al-Kindi-0: I would prefer to pass in the RNG as input to the function and mark it with `CryptoRng` (which we should probably start doing) whenever we want the RNG to be cryptographically secure. This pattern is the one we follow in our `crypto` crate and it was motivated by the need to support `no-std`.

Regarding the question of using a seedable RNG or not, the main motivation for its use is speed and reproducibility. For example, if one needs to make 100-s of calls to the RNG (e.g., Falcon DSA signing procedure)  then a seedable RNG is the way to go, `thread_rng` should probably be the go to solution in all other situations.

crates/store/src/genesis/config/samples/01-simple.toml
bobbinth: While related, I think fee parameters are conceptually different from the native faucet and so I would probably have them as a separate item. Also, we'll have several different base fees and so maybe it makes sense to group them together. Here is a slightly different way to represent these:
```toml
[fee_parameters]
base_fees = { verification = 0 }
```
Or
```toml
[fee_parameters.base_fees]
verification = 0
```
(both of these should result in an identical underlying structure, I believe)

But also, if we want to keep it simple, I think the following is totally fine:
```toml
[fee_parameters]
verification_base_fee = 0
```

crates/store/src/genesis/config/mod.rs
bobbinth: No strong preferences either - but also think that keeping them uniform (e.g., with just `faucet_{symbol}.mac`) is probably fine.

crates/store/src/genesis/config/mod.rs
bobbinth: I'm not too concerned about security at this point in time - so, I'd go for w/e is convenient (e.g., if we need reproducibility for tests, then we should make this code deterministic).

The reason I'm not too concerned about security is that the current approach works only for devnet/tesnet and general testing purposes. Once we go to mainnet, private keys for genesis accounts would need to be generated externally and here we'll be dealing only with public keys. So, we'll need to update the genesis generation process to work with with scenarios where we want to create accounts and we already have public keys for them - so, no private keys need to be generated during genesis. But this is for a different PR (and we should probably create an issue for this).

bobbinth (commented):
Looks good! Thank you! I left a few comments inline. The main one is about treating `fee_parameters` as a stand alone section in the config file (rather than having it be a sub-section under `native_faucet`).

crates/store/src/genesis/config/mod.rs
drahnr: I managed to misread your comment.

crates/store/src/genesis/config/mod.rs
drahnr: tl;dr will leave as is, if we need it for predictability, we should change it then (provided seed, provided rng), once we actually need it

PR #1158, opened 2025-08-15
feat: use a partial SMT proof instead of a list of merkle proofs

### LINKED ISSUE #617: Return one batch proof per storage map
On the `GetAccountProofs` endpoint, we are returning a list of the requested storage map keys with an SMT proof that can be used on the client side to provide data to the VM in order to execute transactions with foreign account data. This is fine but we could optimize bandwidth and processing on the client side by returning one batch proof per map instead.

More context: https://github.com/0xPolygonMiden/miden-node/pull/598#discussion_r1912749484

drahnr:
Closes #617 with the carve-out of improving efficiency in https://github.com/0xMiden/miden-node/pull/1178

### DIFF
--- a/proto/proto/store/rpc.proto
+++ b/proto/proto/store/rpc.proto
@@ -139,7 +139,7 @@ message AccountProofs {
         // State header for public accounts.
         message AccountStateHeader {
             // Represents a single storage slot with the requested keys and their respective values.
-            message StorageSlotMapPartialSmt {
+            message StorageSlotMap {
                 // The storage slot index ([0..255]).
                 uint32 storage_slot = 1;
 
@@ -169,7 +169,7 @@ message AccountProofs {
             optional bytes account_code = 3;
 
             // A sparse merkle tree per storage slot, including all relevant merkle proofs for storage entries.
-            repeated StorageSlotMapPartialSmt partial_storage_smts = 5;
+            repeated StorageSlotMap partial_storage_smts = 5;
         }
 
         // The account witness for the current state commitment of one account ID.
--- a/crates/proto/src/generated/rpc_store.rs
+++ b/crates/proto/src/generated/rpc_store.rs
@@ -102,14 +102,14 @@ pub mod account_proofs {
             /// A sparse merkle tree per storage slot, including all relevant merkle proofs for storage entries.
             #[prost(message, repeated, tag = "5")]
             pub partial_storage_smts: ::prost::alloc::vec::Vec<
-                account_state_header::StorageSlotMapPartialSmt,
+                account_state_header::StorageSlotMap,
             >,
         }
         /// Nested message and enum types in `AccountStateHeader`.
         pub mod account_state_header {
             /// Represents a single storage slot with the requested keys and their respective values.
             #[derive(Clone, PartialEq, ::prost::Message)]
-            pub struct StorageSlotMapPartialSmt {
+            pub struct StorageSlotMap {
                 /// The storage slot index (\[0..255\]).
                 #[prost(uint32, tag = "1")]
                 pub storage_slot: u32,
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -948,7 +948,7 @@ impl State {
 
                     let partial_storage_smts = Vec::from_iter(
                         partials.into_iter()
-                            .map(|(slot, partial_smt)| proto::rpc_store::account_proofs::account_proof::account_state_header::StorageSlotMapPartialSmt {
+                            .map(|(slot, partial_smt)| proto::rpc_store::account_proofs::account_proof::account_state_header::StorageSlotMap {
                                 storage_slot: u32::from(slot),
                                 partial_smt: partial_smt.to_bytes(),
                             })


proto/proto/store/rpc.proto
```diff
@@ -136,6 +136,15 @@ message AccountProofs {
     message AccountProof {
         // State header for public accounts.
         message AccountStateHeader {
+            // Represents a single storage slot with the requested keys and their respective values.
+            message StorageSlotMapPartialSmt {
+                // The storage slot index ([0..255]).
+                uint32 storage_slot = 1;
+
+                // Merkle proofs of the map value as partial sparse merkle tree for compression.
+                bytes partial_smt = 2;
```
Mirko-von-Leipzig, line 149: I'm a bit out-of-the-loop here; how complex a data structure is this? Would it be easy enough to represent this within protobuf schema, similar to `types::primitives::SparseMerklePath`?

If kept as raw bytes I'd extend the comment to explain what actual rust type this is encoded by and how e.g. using winterfel's serde.

proto/proto/store/rpc.proto
```diff
@@ -136,6 +136,15 @@ message AccountProofs {
     message AccountProof {
         // State header for public accounts.
         message AccountStateHeader {
+            // Represents a single storage slot with the requested keys and their respective values.
+            message StorageSlotMapPartialSmt {
```
Mirko-von-Leipzig: We probably don't need both `StorageSlotMapPartialSmt` and `StorageSlotMapProof`?

proto/proto/store/rpc.proto
drahnr: Effectively that'd re-implement the `Serializable for PartialSmt` with the generated types. I am not convinced that adds much value for introspectability of messages.

proto/proto/store/rpc.proto
drahnr: The latter was supposed to be gone already :)

proto/proto/store/rpc.proto
drahnr: added a comment with the relevant info

bobbinth: I think in this case, we can probably describe most of the structure using protobuf messages. One of the reasons is that we already have `SmtLeaf` message implemented (in the `primitives.proto`). So, we'll just need `InnerNode` implemented and then `PartialSmt` could look like so:
```proto
message PartialSmt {
    Digest root = 1;
    repeated SmtLeaf leaves = 2;
    repeated InnerNode nodes = 3;
}
```
And I would put these messages into `primitives.proto`.

-----

Separately, I've just realized that the way we serialize `PartialSmt` right now is not entirely right. It is technically correct, but serializing all inner nodes is not necessary, and moreover, we don't check on deserialization if the data is actually correct (so, we assume that serialization was done by a trusted party). We have the same issue with many other structs/messages - so, it is not like `PartialSmt` is unique here - but it is something that we should start thinking about fixing.

So, if there is a simple way to make the protobuf-based serialization of `PartialSmt` work correct - that would be awesome. But also, if this requires too much work - we can come back to it later.

drahnr: `PartialSmt` is indeed serialized by adding all inner nodes, but by definition we only serialize those present. To my understanding by constructing the `PartialSmt` from `MerklePath`s, this is guaranteed to only contain relevant inner nodes.

We do check validity on the deserialization side _after_ obtaining the `PartialSmt` by doing a `::open` call which would fail at `get_path` if any element in the path would be missing.

drahnr: I've started to impl the protobuf representation of the inner nodes and the `PartialSmt`, I'll give it another shot. I am not certain we want to do the validation at that point or if we want to rely on `PatialSmt::open(key)` instead.

bobbinth: > `PartialSmt` is indeed serialized by adding all inner nodes, but by definition we only serialize those present. To my understanding by constructing the `PartialSmt` from `MerklePath`s, this is guaranteed to only contain relevant inner nodes.

We indeed only would have relevant nodes, but some of these relevant nodes are "redundant" - i.e., we can compute them from other nodes. For example, consider the following tree:

image

Let's say that this is a partial SMT where we can open nodes `7` and `13`. In the current implementation, all nodes with solid borders will be included because they are all needed to authenticate the two openings. But, we actually don't need nodes `6`, `4`, `14`, `12`, and `8` because we can compute them from the remaining nodes.

Computing these nodes also automatically makes the deserialized structure "correct" - i.e., we know for sure that node `8` would be the root of the tree because we've computed it to be so.

> We do check validity on the deserialization side _after_ obtaining the `PartialSmt` by doing a `::open` call which would fail at `get_path` if any element in the path would be missing.

I haven't looked at the code but unless we call `open()` on all leaves - we won't be able to guarantee that the structure is correct (and calling it on all leaves is probably quite expensive).

drahnr: > Let's say that this is a partial SMT where we can open nodes 7 and 13. In the current implementation, all nodes with solid borders will be included because they are all needed to authenticate the two openings. But, we actually don't need nodes 6, 4, 14, 12, and 8 because we can compute them from the remaining nodes.

I'll reduce the included inner-nodes to child-less (in the sense of there is no such node present in the _partial_ smt) nodes only.

The encoding of the minimal set of nodes is done, the reconstruction is WIP.

drahnr: Due to increased effort updating `base` to `crypto@0.16.1`, the optimization is excluded from this PR / carved out to #1178

bobbinth: I'd rather not break the API twice - so, probably makes sense to go directly to #1178.

drahnr: #1178 requires `miden-base` to bump to dependency `miden-crypto`, which is a bit more work, since it's still at `0.15.6` in `next`, and has a few more breaking changes https://github.com/0xMiden/miden-vm/issues/2113 and https://github.com/0xMiden/miden-base/issues/1831.

bobbinth: Ah yes! I'm hoping that we'll update `miden-base` to the next version of the VM some time next week.

drahnr:
Not quite sure what is left to address in the scope of this PR, the inefficiencies will be addressed in https://github.com/0xMiden/miden-node/pull/1178 (duplicate inner nodes, additional merkle path traversal on decoding for verification, ..)

SantiagoPittella (approved):
Considering https://github.com/0xMiden/miden-node/pull/1178 I think that this looks good.

proto/proto/store/rpc.proto
```diff
@@ -138,6 +138,17 @@ message AccountProofs {
...
         message AccountStateHeader {
+            // Represents a single storage slot with the requested keys and their respective values.
+            message StorageSlotMap {
+                // The storage slot index ([0..255]).
+                uint32 storage_slot = 1;
+
+                // Merkle proofs of the map value as partial sparse merkle tree for compression.
+                // The respective rust types is `SparseMerkleTree` and the transformation to and from
+                // bytes is done via the traits `Serializable::to_bytes` and `Deserializable::from_bytes`.
+                bytes partial_smt = 2;
```
bobbinth, line 149: I think we are intending to deserialize this as `PartialStorageMap` (cc @igamigo or @SantiagoPittella to confirm). If so, I would name this message `PartialStorageMap` as well.

Also, we've recently modified the definition of `PartialStorageMap` to also include original keys (see https://github.com/0xMiden/miden-base/pull/1878) - so, serializing `PartialSmt` may no-longer be sufficient. So, maybe, what we should do here is serialize `PartialStorageMap`.

proto/proto/store/rpc.proto
bobbinth, line 159: Do we still need this message?

proto/proto/store/rpc.proto
```diff
@@ -157,8 +168,8 @@ message AccountProofs {
             // the current one.
             optional bytes account_code = 3;
 
-            // Storage slots information for this account
-            repeated StorageSlotMapProof storage_maps = 4;
+            // A sparse merkle tree per storage slot, including all relevant merkle proofs for storage entries.
+            repeated StorageSlotMap partial_storage_smts = 5;
```
bobbinth, line 172: nit: I would name this field just `storage_maps`.

proto/proto/store/rpc.proto
drahnr: We should be using the serialization of `PartialSmt` _inside_ the impl of `PartialStorageMap` iiuc.

drahnr:
Marking as blocked, since there is a lot of conflict with #1214 and that takes precedence

Issue #616, opened 2025-01-13
Normalize RPC endpoints


Mirko-von-Leipzig:
> Not for this PR, but it feels like we should normalize these endpoints somehow. For example:
> 
> - I'm not sure if we are actually using `GetAccountStateDelta` in the client - @igamigo, could you check?
> - I think the current `GetAccountDetails` endpoint returns the entire account, which would probably not work for large accounts.
> - `GetAccountProofsRequest` is used primarily for FPI, but maybe it can be used for other things as well?
>
> Generally, we should probably list the use case we have for retrieving account data from the node and see if these 3 methods are still the right way to get this data. @polydez - could you create an issue for this?

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-node/pull/542#discussion_r1868531536_

Looks like we can also remove `CheckNullifiers`, but not `CheckNullifiersByPrefix` ([origin comment](https://github.com/0xPolygonMiden/miden-node/pull/542#discussion_r1868528446))?

Mirko-von-Leipzig:
@igamigo could you follow-up on the above re: `GetAccountStateDelta`

igamigo:
Sorry, I think I hadn't responded to that comment because of [this discussion](https://github.com/0xPolygonMiden/miden-node/issues/527#issuecomment-2529201687) that was also going on. In short, we do not currently use the `GetAccountStateDelta` endpoint (but maybe we should as I mention in the comment)

PR #1162, opened 2025-08-18
remove account delta API

### LINKED ISSUE #422: [Feature]: Limit number of deltas per request in `GetAccountStateDelta`

Rollup operator

Our current implementation is great in terms of space consumption and it's pretty fast to get needed deltas, but for some accounts it might be needed to get enormous number of deltas in order to sync to the latest state. Obvious solution is to limit number of deltas processed by one single call to the endpoint. For cases, when deltas number is too hight it might be better to just request account's full state.

I see two solutions for it:
1. Limit maximum number of deltas processed in current solution. This will require adding block number to the response in order for the client to be aware that not all requested deltas were returned and make next requests starting from the last returned block number until it get all needed deltas. Another solution is to just return an error, if number of deltas is too high and client will request account's full latest state.
2. Limit maximum number of account changes requested (by nonce). Return an error, if number of nonces (`to_nonce - from_nonce`) requested is higher than the limit. In this case client should request account's full latest state.

From @bobbinth:
I'm leaning toward the first approach. In general, I think we could introduce "epoch-based pagination" I described in https://github.com/0xPolygonMiden/miden-node/issues/404#issuecomment-2226192376 to a couple of endpoints and this would be one of them. We should also "stress-test" at some point to see what happens if the response is too big. The default Tonic message size is [4MB](https://github.com/hyperium/tonic/blob/1934825ff52bff26bb88b709aee9ac73d3ea51c0/tonic/src/codec/mod.rs#L32). Would be good to find out what happens if we exceed this limit (i.e., what will break and what kind of an error message we'll get).

### How is this feature used?

-

### Anything else?

_No response_

drahnr:
Addresses the first piece of #616

* removes the public store RPC endpoint
* removes the account delta table and associated tables
* removes any unused queries using those

Closes #422 by removing the underlying API to change


igamigo (approved):

proto/proto/store/rpc.proto
```diff
@@ -34,10 +34,6 @@ service Rpc {
     // Returns the latest state proofs of the specified accounts.
     rpc GetAccountProofs(AccountProofsRequest) returns (AccountProofs) {}
 
-    // Returns delta of the account states in the range from `from_block_num` (exclusive) to
-    // `to_block_num` (inclusive).
-    rpc GetAccountStateDelta(AccountStateDeltaRequest) returns (AccountStateDelta) {}
```
bobbinth, line 39: Do we still need to keep `AccountStateDeltaRequest` and `AccountStateDelta` messages around?

drahnr: No, unused types. Removed

PR #1163, opened 2025-08-18
chore: merge remote-prover crates

### LINKED ISSUE #995: Merge `miden-proving-service` and `miden-proving-service-client`
We should determine, and implement if needed, if we want to merge the server and client of the proving service in a single crate.

The RPC component does have both parts together, separated in different folders, we the client is only for internal use. In particular, in the `miden-client` we are reconstructing the RPC Client using `miden-node-proto-build.

SantiagoPittella:
closes #995


bobbinth:
> `cmake` for the faucet

The faucet will be removed shortly - so, we can ignore this. Not sure about the other ones. cc @sergerad and @Mirko-von-Leipzig.

bin/remote-prover/src/client/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+#[cfg(feature = "block-prover")]
+pub use block_prover::RemoteBlockProver;
+
+// Generated protobuf code
+pub mod generated {
+    #[cfg(feature = "std")]
+    pub use crate::generated::proto::*;
+    #[cfg(not(feature = "std"))]
+    pub use crate::generated::proto::*;
+}
```
Mirko-von-Leipzig:
```suggestion
pub use crate::generated::proto as generated;
```

bin/remote-prover/src/client/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+
+#[cfg(feature = "batch-prover")]
+pub mod batch_prover;
+#[cfg(feature = "batch-prover")]
+pub use batch_prover::RemoteBatchProver;
+
+#[cfg(feature = "block-prover")]
+pub mod block_prover;
+#[cfg(feature = "block-prover")]
+pub use block_prover::RemoteBlockProver;
```
Mirko-von-Leipzig: Are there additional items within each that are required? At least the `batch_prover` module only contains the `RemoteBatchProver` so we should be able to simplify the exports.

bin/remote-prover/src/generated/client/nostd/mod.rs
Mirko-von-Leipzig, line 1: Is this a consequence of us including the generated files in the source tree? i.e. would this be required if we didn't do that but kept them in the build directory?

bin/remote-prover/src/generated/client/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[cfg(not(feature = "std"))]
+pub mod nostd;
+
+// Re-export based on current environment
+#[cfg(feature = "std")]
+pub use std::*;
+
+#[cfg(not(feature = "std"))]
+pub use nostd::*;
```
Mirko-von-Leipzig: I don't think we should have both `pub mod x` and `pub use x::*`.

Definitely `pub mod std` is a no-no because that will trivially clash with the Rust `std::` paths.

bin/remote-prover/Cargo.toml
Mirko-von-Leipzig, line 1: Do we truely need all these different feature gates? 

I assume `nostd` is definitely required, but is there strictly a need to separate client and server out from each other?

Mirko-von-Leipzig (commented):
I'm focusing on the higher level organisation under the assumption that the implementation is mostly just moves :)

bin/remote-prover/Cargo.toml
SantiagoPittella: I will remove the specific clients features, it was useful before when we used different `miden-tx` features depending on the client type, but isn't needed now.

SantiagoPittella:
Update: I've tested both block and batch prover clients within the node, I'm currently performing tests to check the transaction prover and the respective update in the `miden-client` crate.

bin/remote-prover/src/generated/client/nostd/mod.rs
SantiagoPittella: You meant the use of `include!`? Or having a separated `nostd` generated mod?

Cargo.toml
```diff
@@ -6,7 +6,7 @@ members = [
   "crates/block-producer",
   "crates/ntx-builder",
   "crates/proto",
-  "crates/remote-prover-client",
+
```
sergerad: nit: unnecessary newline

sergerad (approved):

bin/remote-prover/src/generated/client/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[cfg(feature = "std")]
+mod std;
+
+#[cfg(not(feature = "std"))]
+mod nostd;
+
+// Re-export based on current environment
+#[cfg(feature = "std")]
+pub use std::*;
```
igamigo, line 9: nit (feel free to disregard): I'd put this after the `std` mod declaration (on line 8). Same for the `nostd` one.

bin/remote-prover/src/generated/server/mod.rs
```diff
@@ -0,0 +1,15 @@
+
+
```
igamigo: nit: Shoudl these lines be removed?

crates/remote-prover-client/README.md
tomyrd, line 1: Were these feature descriptions moved elswhere or were they completely removed? Should we explain the new added features in a separate readme?

tomyrd (approved):
Looks good! Left a small question

bin/remote-prover/src/generated/client/nostd/mod.rs
igamigo: I could be wrong but I think Mirko meant to ask whether we would need this file structure if we used `OUT_DIR` as the directory for placing the generated gRPC bindings, instead of checking in the files to the source tree (aka, putting them under `./src`)

crates/remote-prover-client/README.md
SantiagoPittella: mb, this needs an update.

Makefile
```diff
@@ -86,7 +88,7 @@ check: ## Check all targets and features for errors without code generation
 .PHONY: build
 build: ## Builds all crates and re-builds protobuf bindings for proto crates
 	${BUILD_PROTO} cargo build --locked --workspace
-	${BUILD_PROTO} cargo build --locked -p miden-remote-prover-client --target wasm32-unknown-unknown --no-default-features  # no-std compatible build
+	${BUILD_PROTO} cargo build --locked -p miden-remote-prover --target wasm32-unknown-unknown --no-default-features --features client  # no-std compatible build
```
igamigo, line 99: I think one implicit change from the new feature is that nothing tests that the crate builds well if you want to test the client in a non-wasm target, right? Although maybe I missed somethig.

If so, it would be nice to test somehow. It seems like there are 3 main scenarios that we want to ensure work well: running the service, running the client on `std` targets and running the client on `wasm` targets.

scripts/publish-crates.sh
```diff
@@ -29,7 +29,7 @@ miden-node-proto-build
 miden-node-proto
 miden-node-store
 miden-remote-prover
-miden-remote-prover-client
+
```
igamigo:
```suggestion
```

Cargo.toml
```diff
@@ -33,15 +32,15 @@ opt-level = 2
...
-miden-remote-prover-client = { path = "crates/remote-prover-client", version = "0.11" }
+miden-node-block-producer = { path = "crates/block-producer", version = "0.11" }
+miden-node-ntx-builder    = { path = "crates/ntx-builder", version = "0.11" }
+miden-node-proto          = { path = "crates/proto", version = "0.11" }
+miden-node-proto-build    = { path = "proto", version = "0.11" }
+miden-node-rpc            = { path = "crates/rpc", version = "0.11" }
+miden-node-store          = { path = "crates/store", version = "0.11" }
+miden-node-test-macro     = { path = "crates/test-macro" }
+miden-node-utils          = { path = "crates/utils", version = "0.11" }
+miden-remote-prover       = { path = "bin/remote-prover", version = "0.11" }
```
igamigo: Should this have `default-features = false`? Or maybe `server` should be taken out of the default features? So that the other crates do not include it unnecessarily

bin/remote-prover/Cargo.toml
```diff
@@ -12,48 +12,112 @@ rust-version.workspace = true
...
+  "dep:miden-block-prover",
+  "dep:miden-node-utils",
+  "dep:miden-objects",
+  "dep:miden-tx",
+  "dep:miden-tx-batch-prover",
+  "dep:pingora",
+  "dep:pingora-core",
+  "dep:pingora-limits",
+  "dep:pingora-proxy",
+  "dep:prometheus",
```
igamigo, line 48: nit: Should this be only gated by `metrics`?

bin/remote-prover/Cargo.toml
```diff
@@ -12,48 +12,112 @@ rust-version.workspace = true
...
-path = "src/main.rs"
+name              = "miden-remote-prover"
+path              = "src/main.rs"
+required-features = ["server"]
 
 [features]
-concurrent = ["miden-tx/concurrent"]
-default    = ["concurrent"]
+# Default features
+default = ["concurrent", "server"]
```
igamigo: Since the `miden-remote-prover` already requires `server`, should it be removed from the defaults?

bin/remote-prover/src/client/batch_prover.rs
```diff
@@ -1,16 +1,18 @@
-use alloc::string::{String, ToString};
-use alloc::sync::Arc;
-use alloc::vec::Vec;
+#[cfg(not(feature = "std"))]
+use alloc::{format, string::String, sync::Arc, vec::Vec};
+#[cfg(feature = "std")]
 use core::time::Duration;
```
igamigo: nit: In order to avoid the `#[cfg()]`, you could use the fully qualified name in the code:

```rust
    .timeout(core::time::Duration::from_millis(10000));
```

igamigo: But also, because the API endpoint is based on whether the target is `wasm32`, would this fail to compile if the target is an std environment but the `std` feature is not enabled?

bin/remote-prover/README.md
```diff
@@ -68,6 +70,37 @@ make install-remote-prover
 
 The CLI can be installed from the source code using specific git revisions with `cargo install` or from crates.io with `cargo install miden-remote-prover`.
 
+## Client Library
```
bobbinth: It feels like we need to re-organize this file to make two top-level sections - client and server. For example, the "Source installation" section above is not really relevant to the client - and for the client, we should have a different section for installation.

bin/remote-prover/src/shared/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! Shared types and utilities used by both client and server
+
+pub mod error;
+
+// Re-export commonly used items
+pub use error::*;
```
bobbinth: Seems like the only think under this module is `errors`. Maybe this should be just `errors` module? (and then we can get rid of `shared` module).

bin/remote-prover/src/lib.rs
```diff
@@ -1,6 +1,41 @@
...
+
+// Conditional module exports based on features
+#[cfg(feature = "client")]
+pub mod client;
+
+#[cfg(feature = "server")]
+pub mod server;
+
+pub mod generated;
+pub mod shared;
```
bobbinth: A few thoughts on the organization:

- Why do we need the `server` module to be a part of the library? Could we not just import it directly in `main.rs` file? Or is the intent for the server component to be accessible programmatically - i.e., out side of the binary?
- If we can get rid of the `server` module, the library could just export the client structs and there is no need for the separate `client` module.

I guess the only thing that we may want to share is the `generated` module - but even then, we could probably keep them separate.

bobbinth (commented):
Looks good! Thank you. Not a full review but a left a few comments inline.

The main thing is that is that I'm not sure about is having both `client` and `server` in the library. Do we actually need this? If not, we can probably separate them so that the library is only exposing client-related structs, and the binary is using binary-related structs. Basically, there would be very little sharing between the two.

But also, taking a step back, it is not clear to me that this actually improves on the current state of things. I guess the main benefit here is that we get a single `miden-remote-prover` crate - but AFAICT, it comes at the expense of more intertwined code, and potentially worse readability.

Or maybe I'm missing some other benefits?

Cargo.toml
SantiagoPittella: I added the `default-features = false`

bin/remote-prover/Cargo.toml
SantiagoPittella: The `metrics` feature should be removed, we need prometheus for the server since we use it all over the place for metrics and we already use an env var for enabling it.

bin/remote-prover/src/client/batch_prover.rs
SantiagoPittella: Yes, if the target is different than `wasm32` but the `std` feature isn't enabled this fails to compile

SantiagoPittella: Not sure how should we handle this, maybe with something similar to what we do in the client with:
`#[cfg(all(not(target_arch = "wasm32"), not(feature = "std")))]
compile_error!("std feature is required for non-wasm32 targets");`

wdyt?

bin/remote-prover/src/lib.rs
SantiagoPittella: I think that this comments answers this two questions: https://github.com/0xMiden/miden-node/pull/1163#issuecomment-3229180047 let me know what do you think

bobbinth:
Let's put this on-hold for now as we are figuring out the general direction for remote provers.

SantiagoPittella:
Closing since the issue is on hold and the code diverted way too much already.

PR #1164, opened 2025-08-19
test: fix `GenesisConfig` test

igamigo:
This test was failing since the [last PR merge](https://github.com/0xMiden/miden-node/actions/runs/17043558181/job/48313283604).
It didn't come up while the [PR was on review](https://github.com/0xMiden/miden-node/actions/runs/17046924559/job/48325685660?pr=1132) apparently though.

The main problem seemed to be that the test is based on account ordering as defined in the TOML, but this order is not guaranteed due to the accounts passing through a `HashMap` first. For this I switched to a `BTreeMap`, but there might be some other way to do this. For example, I'm not sure but it seems like `TokenSymbolStr` could have been introduced for `Hash`. Or maybe the test should not be based on the ordering itself and this behavior could be documented.

### DIFF
--- a/crates/store/src/genesis/config/mod.rs
+++ b/crates/store/src/genesis/config/mod.rs
@@ -1,6 +1,7 @@
 //! Describe a subset of the genesis manifest in easily human readable format
 
-use std::collections::HashMap;
+use std::cmp::Ordering;
+use std::collections::{BTreeMap, HashMap};
 use std::str::FromStr;
 
 use miden_lib::AuthScheme;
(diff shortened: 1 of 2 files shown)


crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig, line 1: I _think_ its probably a good idea to retain the ordering of the accounts as defined in the toml because otherwise its difficult to know which wallet is which account though one could "force" this in the test by sorting by asset value or something.

We should maybe consider special casing the native faucet all the way through to genesis state types but that doesn't really address the issue of this test.

Likely this just means we should keep the vec throughout and a separate `Token -> usize` lookup map? Just my 2c.

drahnr: We might just use `indexmap::IndexMap` to retain insertion order.

drahnr:
The tests didn't fail for me locally just yet. This is on me. Thanks @igamigo for taking care of it

crates/store/src/genesis/config/mod.rs
Mirko-von-Leipzig: Seems heavy to pull in a dependency just for that :sweat_smile: Could also just be `Vec<(Token, Account)>` and using `find`

drahnr: `indexmap` is such a common crate - it's already part of our dependency tree in 2 versions :melting_face:

drahnr:
https://github.com/0xMiden/miden-node/pull/1165

PR #1166, opened 2025-08-20
fix: Enforce single target network notes in NTB

### LINKED ISSUE #1115: Client interaction failure after creating network note
**Description**:
When creating a network note to be consumed by the "counter contract" and setting the `NoteTag` to use `NoteExecutionMode::Network`, the transaction succeeds, but any call to the rust client afterwards will fail with the following error:
```
Error: RpcError(RequestError("sync_state", "status: Unavailable, message: \"unavailable\", details: [], metadata: MetadataMap { headers: {\"date\": \"Sat, 26 Jul 2025 08:50:00 GMT\", \"content-type\": \"application/grpc\", \"content-length\": \"0\"} }"))
```

However, if the note tag is changed to use `NoteExecutionMode::Local`, the transaction succeeds and calling the client does not error out.

**Steps to Reproduce**:

1. Clone the PoC repository [here](https://github.com/partylikeits1983/masm-project-template/tree/ntx-poc-0.10.0) and checkout the `ntx-poc-0.10.0` branch.
2. Run the following test:

   ```bash
   cargo test --release --package masm-project-template --test increment_count_test -- increment_counter_with_note --exact --show-output
   ```

**Expected Behavior**:
The note should be successfully created when using `NoteExecutionMode::Network`.

**Actual Behavior**:
The transaction fails with the `RpcError` message provided above.

sergerad:
## Context

Closes #1115.

The NTB only supports single target network notes and does not currently check for this when constructing `NetworkNote` from `Note`.

## Changes
- Update `NetworkNote` to be an enum with single and multi target variants.
- Add `SingleTargetNetworkNote` and `MultiTargetNetworkNote` types.
- Use `SingleTargetNetworkNote` in NTB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - RPC requests with wildcard (`*/*`) media-type are not longer rejected ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
 - Stress-test CLI account now properly sets the storage mode and increment nonce in transactions ([#1113](https://github.com/0xMiden/miden-node/pull/1113)).
 - [BREAKING] Update `notes` table schema to have a nullable `consumed_block_num` ([#1100](https://github.com/0xMiden/miden-node/pull/1100)).
+- Network Transaction Builder checks that a `Note` is single target when validating conversion to `NetworkNote` ([#1166](https://github.com/0xMiden/miden-node/pull/1166)).
 
 ## v0.10.1 (2025-07-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@
 - RPC requests with wildcard (`*/*`) media-type are not longer rejected ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
 - Stress-test CLI account now properly sets the storage mode and increment nonce in transactions ([#1113](https://github.com/0xMiden/miden-node/pull/1113)).
 - [BREAKING] Update `notes` table schema to have a nullable `consumed_block_num` ([#1100](https://github.com/0xMiden/miden-node/pull/1100)).
-- Network Transaction Builder checks that a `Note` is single target when validating conversion to `NetworkNote` ([#1166](https://github.com/0xMiden/miden-node/pull/1166)).
+- Network Transaction Builder now correctly discards non-single-target network notes instead of panicking ([#1166](https://github.com/0xMiden/miden-node/pull/1166)).
 
 ## v0.10.1 (2025-07-14)
 
--- a/crates/ntx-builder/src/state/mod.rs
+++ b/crates/ntx-builder/src/state/mod.rs
@@ -6,7 +6,7 @@ use account::{AccountState, InflightNetworkNote, NetworkAccountUpdate};
 use anyhow::Context;
 use miden_node_proto::domain::account::NetworkAccountPrefix;
 use miden_node_proto::domain::mempool::MempoolEvent;
-use miden_node_proto::domain::note::NetworkNote;
+use miden_node_proto::domain::note::SingleTargetNetworkNote;
 use miden_node_utils::tracing::OpenTelemetrySpanExt;
 use miden_objects::account::Account;
 use miden_objects::account::delta::AccountUpdateDetails;
@@ -265,7 +265,7 @@ impl State {
             } => {
                 let network_notes = network_notes
                     .into_iter()
-                    .filter(|note| note.metadata().tag().is_single_target())
+                    .filter_map(|note| SingleTargetNetworkNote::try_from(note).ok())
                     .collect::<Vec<_>>();
                 self.add_transaction(id, nullifiers, network_notes, account_delta).await?;
             },
@@ -304,7 +304,7 @@ impl State {
         &mut self,
         id: TransactionId,
         nullifiers: Vec<Nullifier>,
-        network_notes: Vec<NetworkNote>,
+        network_notes: Vec<SingleTargetNetworkNote>,
         account_delta: Option<AccountUpdateDetails>,
     ) -> anyhow::Result<()> {
         // Skip transactions we already know about.
--- a/crates/ntx-builder/src/state/mod.rs
+++ b/crates/ntx-builder/src/state/mod.rs
@@ -116,7 +116,7 @@ impl State {
 
         let notes = state.store.get_unconsumed_network_notes().await?;
         for note in notes {
-            // Single target network notes only.
+            // TODO: support multi target network notes.
             if let Some(prefix) = note.account_prefix() {
                 // Ignore notes which don't target an existing account.
                 let Some(account) = state.fetch_account(prefix).await? else {
--- a/crates/ntx-builder/src/state/mod.rs
+++ b/crates/ntx-builder/src/state/mod.rs
@@ -263,6 +263,10 @@ impl State {
                 network_notes,
                 account_delta,
             } => {
+                let network_notes = network_notes
+                    .into_iter()
+                    .filter(|note| note.metadata().tag().is_single_target())
+                    .collect::<Vec<_>>();
                 self.add_transaction(id, nullifiers, network_notes, account_delta).await?;
             },
             MempoolEvent::BlockCommitted { header, txs } => {
--- a/crates/ntx-builder/src/state/mod.rs
+++ b/crates/ntx-builder/src/state/mod.rs
@@ -6,7 +6,7 @@ use account::{AccountState, InflightNetworkNote, NetworkAccountUpdate};
 use anyhow::Context;
 use miden_node_proto::domain::account::NetworkAccountPrefix;
 use miden_node_proto::domain::mempool::MempoolEvent;
-use miden_node_proto::domain::note::SingleTargetNetworkNote;
+use miden_node_proto::domain::note::NetworkNote;
 use miden_node_utils::tracing::OpenTelemetrySpanExt;
 use miden_objects::account::Account;
 use miden_objects::account::delta::AccountUpdateDetails;
@@ -116,13 +116,14 @@ impl State {
 
         let notes = state.store.get_unconsumed_network_notes().await?;
         for note in notes {
-            let prefix = note.account_prefix();
-
-            // Ignore notes which don't target an existing account.
-            let Some(account) = state.fetch_account(prefix).await? else {
-                continue;
-            };
-            account.add_note(note);
+            // Single target network notes only.
+            if let Some(prefix) = note.account_prefix() {
+                // Ignore notes which don't target an existing account.
+                let Some(account) = state.fetch_account(prefix).await? else {
+                    continue;
+                };
+                account.add_note(note);
+            }
         }
         state.inject_telemetry();
 
@@ -263,9 +264,10 @@ impl State {
                 network_notes,
                 account_delta,
             } => {
+                // TODO: support multi target network notes.
                 let network_notes = network_notes
                     .into_iter()
-                    .filter_map(|note| SingleTargetNetworkNote::try_from(note).ok())
+                    .filter(|note| matches!(note, NetworkNote::SingleTarget(_)))
                     .collect::<Vec<_>>();
                 self.add_transaction(id, nullifiers, network_notes, account_delta).await?;
             },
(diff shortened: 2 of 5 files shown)


crates/proto/src/domain/note.rs
```diff
@@ -217,10 +217,12 @@ impl TryFrom<Note> for NetworkNote {
...
-            return Err(NetworkNoteError::InvalidExecutionMode(note.metadata().tag()));
+        // Only support network notes targeted at a single network account.
+        if note.is_network_note() && note.metadata().tag().is_single_target() {
+            Ok(NetworkNote(note))
+        } else {
+            Err(NetworkNoteError::InvalidExecutionMode(note.metadata().tag()))
         }
-        Ok(NetworkNote(note))
     }
 }
```
sergerad, line 234: I'm not actually sure we want the `note.metadata().tag().is_single_target()` check here. Right now only the NTB cares about NetworkNotes. But if we think other parts of the stack might want to handle NetworkNotes then we might want to move this check directly in the NTB stack instead.

Mirko-von-Leipzig: At some point we will want to tackle non-single-target notes as well, however that isn't the case at the moment.

We basically have two options to handle this here

1. Rename `NetworkNote` to something like `SingleTargetNetworkNote`.
2. Keep `NetworkNote` as is, remove the check here and also remove `NetworkNote::account_prefix` which is invalid.

If we do (2) then we probably want (1) created within the ntb to enforce that invariant there. I would go with (1) and then when we want to handle the broader spectrum we can revisit the types.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - RPC requests with wildcard (`*/*`) media-type are not longer rejected ([#1084](https://github.com/0xMiden/miden-node/pull/1084)).
 - Stress-test CLI account now properly sets the storage mode and increment nonce in transactions ([#1113](https://github.com/0xMiden/miden-node/pull/1113)).
 - [BREAKING] Update `notes` table schema to have a nullable `consumed_block_num` ([#1100](https://github.com/0xMiden/miden-node/pull/1100)).
+- Network Transaction Builder checks that a `Note` is single target when validating conversion to `NetworkNote` ([#1166](https://github.com/0xMiden/miden-node/pull/1166)).
```
```suggestion
- Network Transaction Builder now correctly discards non-single-target network notes instead of panicking ([#1166](https://github.com/0xMiden/miden-node/pull/1166)).
```

crates/proto/src/domain/note.rs
Mirko-von-Leipzig: Note that the mempool's subscription currently uses this `TryFrom` to filter for network notes. It might be worth changing that filter to only use `is_network_note` which more aligns with long term direction, and then perform the `single_target` filtering at the ntb only 🤷

sergerad: I have opted to simply filter network transactions coming from mempool for ntb:

```rust
            MempoolEvent::TransactionAdded {
                id,
                nullifiers,
                network_notes,
                account_delta,
            } => {
                let network_notes = network_notes
                    .into_iter()
                    .filter(|note| note.metadata().tag().is_single_target())
                    .collect::<Vec<_>>();
                self.add_transaction(id, nullifiers, network_notes, account_delta).await?;
            },
```

Given we plan to support both single target and other network notes, and we don't know what that will look like, I think at this stage it is not worth creating another newtype `SingleTargetNetworkNote(NetworkNote)`. Given we would probably remove/replace it down the line.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -263,6 +263,10 @@ impl State {
                 network_notes,
                 account_delta,
             } => {
+                let network_notes = network_notes
+                    .into_iter()
+                    .filter(|note| note.metadata().tag().is_single_target())
+                    .collect::<Vec<_>>();
```
bobbinth: Should we also do something similar when reading notes from the store? (e.g., on startup)

sergerad: I don't see `State::add_transaction` used anywhere else and don't see NTB getting network notes any other way. Checked `NetworkTransactionBuilder::serve_new`. May have missed something @Mirko-von-Leipzig LMK if there is anywhere else to do the filter you know of.

Mirko-von-Leipzig: On startup we load all existing network notes from the store https://github.com/0xMiden/miden-node/blob/0f5378df208380add491cdcb802a86946cd035c9/crates/ntx-builder/src/state/mod.rs#L117-L126

Would have to just ensure that this checks out.

crates/proto/src/domain/note.rs
Mirko-von-Leipzig: That is true but as per https://github.com/0xMiden/miden-node/pull/1166#discussion_r2290512661 the ntb currently wants specifically `SingleTargetNetworkNote` - which would be caught by something like a more specific type.

Mirko-von-Leipzig: I'm okay with things as they are; I'm just pointing out that having better type safety would have prevented a lot of this admittedly confusing issue :D

crates/ntx-builder/src/state/mod.rs
bobbinth: I think we should probably update `NetworkNote::account_prefix()` to return `Option<NetworkAccountPrefix>`. Right now, there is an `unwrap()` there which will fail if a network note is not a single-target note.

crates/proto/src/domain/note.rs
sergerad: Decided to add the additional SingleTargetNetworkNode newtype in the end

crates/ntx-builder/src/state/mod.rs
sergerad: I ended up adding the SingleTargetNetworkNote newtype so the account_prefix() change was not required / appropriate. Instead the expect assertion is left in the member fn.

crates/ntx-builder/src/store.rs
```diff
@@ -106,7 +106,9 @@ impl StoreClient {
 
     /// Returns the list of unconsumed network notes.
     #[instrument(target = COMPONENT, name = "store.client.get_unconsumed_network_notes", skip_all, err)]
-    pub async fn get_unconsumed_network_notes(&self) -> Result<Vec<NetworkNote>, StoreError> {
+    pub async fn get_unconsumed_network_notes(
+        &self,
+    ) -> Result<Vec<SingleTargetNetworkNote>, StoreError> {
```
bobbinth: I am not sure this is correct. `get_unconsumed_network_notes()` would return all network notes from the store and so this should also do the same. Also, currently, if there are non-single-target network notes, this will result in an error, I think.

Instead, we should keep this returning all network notes, and the caller should filter out the ones that are not single-target.

crates/proto/src/domain/note.rs
```diff
@@ -202,8 +202,10 @@ impl NetworkNote {
     }
 
     pub fn account_prefix(&self) -> NetworkAccountPrefix {
-        // SAFETY: This must succeed because this is a network note.
-        self.metadata().tag().try_into().unwrap()
+        self.metadata()
+            .tag()
+            .try_into()
+            .expect("NetworkNotes have network account prefix")
     }
```
bobbinth: This method should probably go a way since now we have an explicit type for single-target network notes.

crates/proto/src/domain/note.rs
bobbinth: If we go this way, we need to make sure to handle them properly (I left a couple of comments about this). We also need to think through the full hierarch - e.g., should `NetworkNote` become an enum with `SingleTarget` and `MultiTarget` (or something like this) as variants?

Overall, I do think that introducing `SingleTargetNetworkNote` is a bit more work and will take a couple more iterations than just making `NetworkNote::account_prefix()` return an `Option`.

sergerad: If NTB will eventually support MultiTarget, then the enum approach is a good idea to do now I think.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -116,13 +116,14 @@ impl State {
...
-            };
-            account.add_note(note);
+            // TODO: support multi target network notes.
+            if let Some(prefix) = note.account_prefix() {
+                // Ignore notes which don't target an existing account.
+                let Some(account) = state.fetch_account(prefix).await? else {
+                    continue;
+                };
+                account.add_note(note);
+            }
```
bobbinth: I would probably replace the `TODO` with a simple comment explaining that we handle only single-target accounts for now.

Also, we can probably re-write this as:
```rust
if let Some(prefix) = note.account_prefix() {
    let Some(account) = state.fetch_account(prefix).await? {
        account.add_note(note);
    }
}
```

Lastly, I would expand the comment and explain under which circumstances we may get a single-target note that doesn't have a target account.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -336,14 +342,15 @@ impl State {
...
-                account.add_note(note);
+            // TODO: support multi target network notes.
+            if let Some(prefix) = note.account_prefix() {
+                self.nullifier_idx.insert(note.nullifier(), prefix);
+                // Skip notes which target a non-existent network account.
+                if let Some(account) =
+                    self.fetch_account(prefix).await.context("failed to load account")?
+                {
+                    account.add_note(note);
+                }
```
bobbinth: Here too I would replace the `TODO` with a simple comment about not handling multi-target notes yet (mostly because enabling support for multi-target notes is a much bigger refactoring than a simple `TODO` here implies).

Also, I would expand the comment and explain why there could be non-existent network accounts.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -263,6 +264,11 @@ impl State {
                 network_notes,
                 account_delta,
             } => {
+                // TODO: support multi target network notes.
+                let network_notes = network_notes
+                    .into_iter()
+                    .filter(|note| matches!(note, NetworkNote::SingleTarget(_)))
+                    .collect::<Vec<_>>();
```
bobbinth: Same comment as above re TODO.

bobbinth (approved):
Looks good! Thank you. Overall, I think this approach works fine for now. We could make it more type-safe by introducing newtype for each type of notes. `NetworkNote` then could look something like:
```rust
pub enum NetworkNote {
    SingleTarget(SingleTargetNote),
    MultiTarget(MultiTargetNote),
}

pub struct SingleTargetNote(Note);
pub struct MultiTargetNote(Note);
```
And then most of the APIs would work with `SingleTargetNote` for now.

But I'll defer to @Mirko-von-Leipzig on this.

crates/proto/src/domain/note.rs
```diff
@@ -201,26 +206,47 @@ impl NetworkNote {
...
+    /// which identifies the target account. For multi-target notes, this returns `None` as
+    /// they don't target a specific account prefix.
+    pub fn account_prefix(&self) -> Option<NetworkAccountPrefix> {
+        match self {
+            NetworkNote::SingleTarget(_) => {
+                let prefix: NetworkAccountPrefix = self
+                    .metadata()
+                    .tag()
+                    .try_into()
+                    .expect("NetworkNote has a network account prefix");
```
Mirko-von-Leipzig:
```suggestion
                    .expect("Single target network note's tag should contain an account prefix");
```

crates/ntx-builder/src/state/mod.rs
```diff
@@ -116,13 +116,14 @@ impl State {
...
         for note in notes {
-            let prefix = note.account_prefix();
-
-            // Ignore notes which don't target an existing account.
-            let Some(account) = state.fetch_account(prefix).await? else {
-                continue;
-            };
-            account.add_note(note);
+            // TODO: support multi target network notes.
+            if let Some(prefix) = note.account_prefix() {
```
Mirko-von-Leipzig: This does feel a bit weird because its effectively an enum match but we're pushing the match inside the function. I would prefer having a strong type for this instead. This would also let us be confident in the other code areas where right now we still take a `NetworkNote` and just assume its single target from external usage.

Mirko-von-Leipzig (approved):
As mentioned; I would prefer a `SingleTargetNetworkNote` so that we're certain what we're working with everywhere.

We don't know what or how we're going to manage the other variant(s); I think adding a strong type will help us refactor in the future without worrying about screwing up.

PR #1167, opened 2025-08-20
chore: Remove unnecessary safety statements by passing Url type through

sergerad:
It seems that we convert Url type to String unnecessarily and then make safety statements to explain.

This PR feeds the Url type deeper into the stack and gives ownership because the conversion to String from Url is akin to a clone.


crates/block-producer/src/store/mod.rs
```diff
@@ -125,19 +125,17 @@ pub struct StoreClient {
 
 impl StoreClient {
     /// Creates a new store client with a lazy connection.
-    pub fn new(store_url: &Url) -> Self {
-        // SAFETY: The store_url is always valid as it is a user-provided URL that has been
-        // validated.
-        let store = Builder::new(store_url.to_string())
-            .expect("Failed to initialize store endpoint")
+    pub fn new(store_url: Url) -> Self {
+        info!(target: COMPONENT, store_endpoint = %store_url, "Initializing store client");
```
bobbinth, line 129: Question: what is the motivation for moving the `info!()` statements from after initialization to before initialization? I don't mind this change, but maybe we should have these traces both before and after initialization?

This comment is also relevant for other files in changes in this PR.

crates/block-producer/src/store/mod.rs
```diff
@@ -125,19 +125,17 @@ pub struct StoreClient {
...
+
+        let store = Builder::new(store_url)
+            .expect("Store Url is valid")
```
bobbinth, line 132: Question: why do we assume that the user-provided URL is always valid? Is the thinking that since this only happens at startup, it is fine to panic?

sergerad: Its unfortunate that the `tonic::Endpoint` type doesn't support `Url` itself so it converts a `String`
```
        let endpoint = Endpoint::from_shared(String::from(url))
            .context("Failed to create endpoint from address")?;
```

Here `store_url` is a `Url` type which must be valid. So we know that the `Builder::new` won't fail.

crates/block-producer/src/store/mod.rs
sergerad: Just to avoid cloning `Url`

crates/block-producer/src/store/mod.rs
bobbinth: In this case, should this not be something like:
```rust
let endpoint = Endpoint::from_shared(String::from(url))
    .expect("Failed to create endpoint from a valid URL");
```
And then `Builder::new()` wouldn't need to return `Result` at all.

sergerad: Good point let me fix that in another PR

Mirko-von-Leipzig: > Question: why do we assume that the user-provided URL is always valid? Is the thinking that since this only happens at startup, it is fine to panic?

This was the thinking yeah. 

> Here store_url is a Url type which must be valid. So we know that the Builder::new won't fail.

I'm not sure how strictly true this is, but I think the same is true of the `Uri` constructor so eh :shrug: `Uri` also has its short-comings and in general its just a bit of mess if one wants to support DNS resolution and load-balancers, otherwise one could simply do `<host>:<port>`.

PR #1168, opened 2025-08-20
feat(store): sql conversions for note execution mode

c410-f3r:
Fixes a TODO. The function does not make sense and is not used anywhere.


drahnr:
I think it'd be preferable to make query filter `    .filter(schema::notes::execution_mode.eq(0_i32))` more expressive. This would imply transforming the `ExecutionMode` to implement `impl SqlTypeConvert for NoteExecutionMode` and use the conversions directly than today's `as i32` cast.

PR #1172, opened 2025-08-21
chore: update base version without bech32 conversion

tomyrd:
Updates the `miden-base` version to [this PR](https://github.com/0xMiden/miden-base/pull/1762) so that the client an also be updated.
These changes required updates in the `faucet` binary, which I wasn't sure if it was still being used after we moved it to a separate repository. For the moment I removed the binary, but if we want to keep it I can revert that change and update it accordingly.


Cargo.toml
```diff
@@ -45,12 +44,12 @@ miden-node-utils           = { path = "crates/utils", version = "0.11" }
...
-miden-objects         = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
-miden-testing         = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
-miden-tx              = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
-miden-tx-batch-prover = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-block-prover    = { branch = "pgackst-remove-account-id-bech32", git = "https://github.com/0xMiden/miden-base" }
+miden-lib             = { branch = "pgackst-remove-account-id-bech32", git = "https://github.com/0xMiden/miden-base" }
+miden-objects         = { branch = "pgackst-remove-account-id-bech32", default-features = false, git = "https://github.com/0xMiden/miden-base" }
+miden-testing         = { branch = "pgackst-remove-account-id-bech32", git = "https://github.com/0xMiden/miden-base" }
+miden-tx              = { branch = "pgackst-remove-account-id-bech32", default-features = false, git = "https://github.com/0xMiden/miden-base" }
+miden-tx-batch-prover = { branch = "pgackst-remove-account-id-bech32", git = "https://github.com/0xMiden/miden-base" }
```
bobbinth: This could now point to `next`, right?

bobbinth (approved):
Looks good! Thank you!

I think it is fine to remove the faucet at this point - so, I'm going to merge it.

PR #1176, opened 2025-08-21
feat: asset vault sync

### LINKED ISSUE #1099: Implement account vault sync endpoint
This is similar to https://github.com/0xMiden/miden-node/issues/1095 but for syncing account vaults. The request/response messages could look like:

```proto
message SyncAccountVaultRequest {
    // block number to start syncing from
    fixed32 block_num = 1;
    
    // Account for which we want to sync storage maps
    AccountId account_id = 2;
}

message SyncAccountVaultResponse {
    fixed32 chain_tip = 1;
    repeated AccountVaultUpdate updates = 2;
}

message AccountVaultUpdate {
    // Technically, we don't need the full digest for this - just 2 elements would be enough
    Digest vault_key = 1;

    // If not present, the asset was removed from the vault.
    optional Asset asset = 2;

    // Block number at which the above asset existed in the account vault
    fixed32 block_num = 3;
}
```
There could also be alternative ways of representing `AccountVaultUpdate` - but this is what I came up with off the top of my head.

The remaining part should probably work pretty similarly to how storage map syncing will work.

igamigo:
Closes #1099

### DIFF
--- a/crates/store/src/server/rpc_api.rs
+++ b/crates/store/src/server/rpc_api.rs
@@ -329,68 +329,6 @@ impl rpc_server::Rpc for StoreApi {
         }))
     }
 
-    /// Returns storage map updates for the specified account within a block range.
-    ///
-    /// Supports cursor-based pagination for large storage maps.
-    #[instrument(
-        parent = None,
-        target = COMPONENT,
-        name = "store.rpc_server.sync_storage_maps",
-        skip_all,
-        level = "debug",
-        ret(level = "debug"),
-        err
-    )]
-    async fn sync_storage_maps(
-        &self,
-        request: Request<proto::rpc_store::SyncStorageMapsRequest>,
-    ) -> Result<Response<proto::rpc_store::SyncStorageMapsResponse>, Status> {
-        let request = request.into_inner();
-
-        let account_id = read_account_id(request.account_id).map_err(|e| *e)?;
-
-        if !account_id.is_public() {
-            return Err(Status::invalid_argument(format!(
-                "account with ID {account_id} is not public"
-            )));
-        }
-
-        let block_from = BlockNumber::from(request.block_from);
-        let chain_tip = self.state.latest_block_num().await;
-
-        let block_to = match request.block_to {
-            Some(block_to) => BlockNumber::from(block_to),
-            None => chain_tip,
-        };
-
-        if block_from > block_to {
-            return Err(Status::invalid_argument("block_from cannot be greater than block_to"));
-        }
-
-        let storage_maps_page = self
-            .state
-            .get_storage_map_sync_values(account_id, block_from, block_to)
-            .await
-            .map_err(internal_error)?;
-
-        let updates = storage_maps_page
-            .values
-            .into_iter()
-            .map(|map_value| proto::rpc_store::StorageMapUpdate {
-                slot_index: u32::from(map_value.slot_index),
-                key: Some(map_value.key.into()),
-                value: Some(map_value.value.into()),
-                block_num: map_value.block_num.as_u32(),
-            })
-            .collect();
-
-        Ok(Response::new(proto::rpc_store::SyncStorageMapsResponse {
-            block_num: storage_maps_page.last_block_included.as_u32(),
-            chain_tip: chain_tip.as_u32(),
-            updates,
-        }))
-    }
-
     #[instrument(
         parent = None,
         target = COMPONENT,


bin/faucet/src/faucet/mod.rs
bobbinth, line 1: This file should probably be deleted.

bin/faucet/src/stub_rpc_api.rs
bobbinth, line 1: Same comment as above.

crates/utils/src/tracing/grpc.rs
```diff
@@ -46,6 +46,7 @@ fn rpc_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
         Some("CheckNullifiers") => rpc_span!("rpc.rpc", "CheckNullifiers"),
         Some("CheckNullifiersByPrefix") => rpc_span!("rpc.rpc", "CheckNullifiersByPrefix"),
         Some("GetBlockHeaderByNumber") => rpc_span!("rpc.rpc", "GetBlockHeaderByNumber"),
+        Some("SyncAccountVault") => rpc_span!("rpc.rpc", "SyncAccountVault"),
```
bobbinth, line 50: Related to https://github.com/0xMiden/miden-node/pull/1140: should we also add an entry here for `SyncAccountStorageMaps`? (I didn't see it in the other PR).

crates/store/README.md
```diff
@@ -114,6 +114,16 @@ the chain.
 
 ---
 
+### SyncAccountVault
```
bobbinth: This should probably be added to the ToC on lines 14 - 28 above.

crates/rpc/README.md
```diff
@@ -95,6 +94,16 @@ the chain.
 
 ---
 
+### SyncAccountVault
```
bobbinth: Same comment as above.

proto/proto/store/rpc.proto
```diff
@@ -259,6 +262,50 @@ message SyncStateResponse {
...
+    repeated AccountVaultUpdate updates = 3;
+}
+
+message AccountVaultUpdate {
+    // Vault key asscoiated with the asset.
+    primitives.Digest vault_key = 1;
+
+    // Serialized asset.
+    // If not present, the asset was removed from the vault.
+    optional bytes asset = 2; // TODO: see if it's worth making a message for assets
```
bobbinth: Yes, let's make a separate message for `Asset`. For now, it could be just `Digest`.

proto/proto/store/rpc.proto
```diff
@@ -259,6 +262,50 @@ message SyncStateResponse {
...
+
+    // Block number at which the above asset was updated in the account vault.
+    fixed32 block_num = 3;
```
bobbinth, line 315: Whey do we need `block_num` here, but we didn't need it for `StorageMapUpdate`?

proto/proto/store/rpc.proto
```diff
@@ -259,6 +262,50 @@ message SyncStateResponse {
...
+    optional fixed32 block_to = 2;
+    
+    // Account for which we want to sync storage maps.
+    account.AccountId account_id = 3;
+}
+
+message SyncAccountVaultResponse {
+    // The block number of the last row included in this response.
+    // For chunked responses, this may be less than request.block_to.
+    fixed32 block_num = 1;
```
bobbinth: nit: the reference to "row" here may be confusing. Could we say "last update" or something like that instead?

Should we also mention that if this is not equal to `request.block_to` then the user would need to make another request with `request.block_from` equal to `block_num + 1`?

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -91,6 +91,16 @@ CREATE TABLE account_storage_map_values (
...
 
+CREATE TABLE account_vault_assets (
+    account_id          BLOB    NOT NULL,
+    block_num           INTEGER NOT NULL,
+    vault_key           BLOB    NOT NULL,
+    asset               BLOB,
+    is_latest_update    BOOLEAN NOT NULL,
+
+    PRIMARY KEY (account_id, block_num, vault_key)
+) WITHOUT ROWID;
```
bobbinth, line 103: Do we not need an index on `(account_id, block_num)` here?

crates/store/src/db/mod.rs
```diff
@@ -40,6 +41,17 @@ pub struct Db {
     pool: deadpool_diesel::Pool<ConnectionManager, deadpool::managed::Object<ConnectionManager>>,
 }
 
+/// Describes the value of an asset for an account ID at `block_num` specifically.
+///
+/// If `asset` is `None`, the asset was removed.
+#[derive(Debug, Clone)]
+pub struct AccountVaultValue {
+    pub block_num: BlockNumber,
```
bobbinth, line 51: Same question as in one of the other comments: why do we need `block_num` here but we didn't need it for `StorageMapValue` in https://github.com/0xMiden/miden-node/pull/1140?

To me it feels right to include the `block_num` (because it makes the list of updates a less ambiguous), but then we also should make the same changes in https://github.com/0xMiden/miden-node/pull/1140.

Also, this struct should probably be defined next to `StorageMapValue` (which is defined in `src/db/models/queries/accounts.rs` - though, I'm not sure if that's the right place for it).

crates/store/src/state.rs
```diff
@@ -993,6 +1000,15 @@ impl State {
         self.db.optimize().await
     }
 
+    pub async fn sync_account_vault(
```
bobbinth, line 1015: Let's add doc comments to this function.

crates/store/src/db/models/queries/insertions.rs
```diff
@@ -56,6 +63,54 @@ pub(crate) fn insert_block_header(
     Ok(count)
 }
 
+/// Insert an account vault asset row into the DB using the given [`SqliteConnection`].
+///
+/// This function will set `is_latest_update=true` for the new row and update any existing
+/// row with the same `(account_id, asset_id)` tuple to `is_latest_update=false`.
```
bobbinth: Should `asset_id` here be actually `vault_key`?

crates/store/src/db/models/queries/insertions.rs
```diff
@@ -56,6 +63,54 @@ pub(crate) fn insert_block_header(
...
+    asset: Option<Asset>,
+) -> Result<usize, DatabaseError> {
+    let account_id = account_id.to_bytes();
+    let vault_key = vault_key.to_bytes();
+    let block_num = block_num.to_raw_sql();
+    let asset = asset.map(|a| a.to_bytes());
+
+    diesel::Connection::transaction(conn, |conn| {
+        // First, update any existing rows with the same (account_id, asset_id) to set
+        // is_latest_update=false
```
bobbinth: Same comment as above re `asset_id` -> `vault_key`.

crates/store/src/db/models/queries/insertions.rs
```diff
@@ -226,6 +281,44 @@ pub(crate) fn upsert_accounts(
...
+                    // TODO: make vault key function public on miden-base
+                    let asset_remove_me = FungibleAsset::new(*faucet_id, 0).unwrap();
+                    insert_account_vault_asset(
+                        conn,
+                        account.id(),
+                        block_num,
+                        asset_remove_me.vault_key(),
+                        asset_update,
+                    )
+                    .unwrap();
```
bobbinth: Let's make this change ASAP as I'm hoping to release `miden-base` today.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -148,6 +149,79 @@ pub(crate) fn select_accounts_by_id(
...
+) -> Result<(BlockNumber, Vec<AccountVaultValue>), DatabaseError> {
+    use schema::account_vault_assets as t;
+
+    // SELECT block_num, faucet_id, delta
+    // FROM account_fungible_asset_deltas
+    // WHERE account_id = ?
+    //   AND block_num >= ?
+    //   AND block_num <= ?
+    // ORDER BY block_num ASC, faucet_id ASC
+    // LIMIT ROW_LIMIT;
```
bobbinth: nit: I'd format this as:
```rust
// SELECT
//     block_num,
//     faucet_id,
//     delta
// FROM
//     account_fungible_asset_deltas
// WHERE
//     account_id = ?
//     AND block_num >= ?
//     AND block_num <= ?
// ORDER BY
//     block_num ASC,
//     faucet_id ASC
// LIMIT
//     ROW_LIMIT;
```
Also, I think this query is not correct - shouldn't we use `account_vault_assets` for this?

bobbinth (commented):
Looks good! Thank you. Not a full review, but I left some comments inline. Also, two general comments:

- We should probably put this PR on top of https://github.com/0xMiden/miden-node/pull/1140 - they kind of are related to the same thing and it would be simpler to see the incremental changes.
- We should make the methodology consistent across these two PRs (unless there is a good reason not to do so).

crates/store/src/db/models/queries/insertions.rs
igamigo: Created https://github.com/0xMiden/miden-base/pull/1794, let me know if it's about what you had in mind

bobbinth: Hmmm - I initially went by the comment, but looking into this in more detail now, I'm wondering if we actually need the changes in `miden-base`.

Here, I think we can just always construct the fungible asset `FungibleAsset::new(*faucet_id, current_amount)` and then use `FungibleAsset::vault_key()` to get the key.

This will require `insert_account_vault_asset()` to always take an asset as a parameter - but I think that may be fine.

igamigo: This works for fungible faucets, but not as well for sharing the API with non fungible faucets because we need a new way to specify the removals.

crates/block-producer/src/server/mod.rs
```diff
@@ -87,7 +87,7 @@ impl BlockProducer {
         info!(target: COMPONENT, endpoint=?self.block_producer_address, store=%self.store_url, "Initializing server");
         let store = StoreClient::new(self.store_url.clone());
 
-        // Retry fetching the chain tip from the store until it succeeds.
+        // retry fetching the chain tip from the store until it succeeds.
```
Mirko-von-Leipzig: Nit: bit of a random change, could probably remain upper case

crates/rpc/README.md
Mirko-von-Leipzig, line 1: Ah. We should also update the mdbook user docs for the RPC reference. For the other PR as well.

crates/store/src/server/rpc_api.rs
Mirko-von-Leipzig, line 1: Nit: `fn status` moved for some reason

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -149,6 +149,68 @@ pub(crate) fn select_accounts_by_id(
...
+                    .eq(account_id.to_bytes())
+                    .and(t::block_num.ge(block_from.to_raw_sql()))
+                    .and(t::block_num.le(block_to.to_raw_sql())),
+            )
+            .order(t::block_num.asc())
+            .limit(i64::try_from(ROW_LIMIT).expect("should fit within i64"))
+            .load::<(i64, Vec<u8>, Option<Vec<u8>>)>(conn)?;
+
+    // Discard the last block in the response (assumes more than one block may be present)
+    let mut last_block_included = block_to.checked_sub(1).unwrap_or_default();
```
Mirko-von-Leipzig: I don't understand why we're discarding the last block?

igamigo: The idea is that if we get enough rows where we know there is at minimum 2 blocks involved, then we can always discard the last one (because it would likely not contain all updates for that block) and keep all the other updates, which would contain, for all blocks there, all updates and not partial data. This is the reason for calculating the overhead based on the limit of updates in a block (which I made up, for now, but they should be defined by the protocol) and adding 1 on top of that.

crates/block-producer/src/server/mod.rs
igamigo: Yeah, not even sure why/how I did this, I think I looked at this code to review that PR related to returning the genesis commitment on the status endpoint

bobbinth (approved):
All looks good! Thank you!

crates/store/src/db/models/queries/accounts.rs
Mirko-von-Leipzig: Sorry my comment was ambiguous. What I meant was I don't understand is what `last_block_included` represents.

I think its intended as the pagination token iirc. If so, should it not be 

```rust
let last_block_num = if raw.len() >= ROW_LIMIT {
    raw.last().unwrap().0 - 1
} else {
    block_to
};
```

bobbinth: Yes, this is one of the changes I made before merging - though, it looks a bit different: https://github.com/0xMiden/miden-node/blob/next/crates/store/src/db/models/queries/accounts.rs#L202

PR #1178, opened 2025-08-22
optimize merkle node compression

drahnr:
Follow up to #1158 

Consequence #617

# What it does

Previously we used the `Serializable`/`Deserializable` implementations for `PartialSmt` which uses a more leaves than the theoretical minum required to reconstruct all intermediate nodes. 

The PR implements a DFS per sibling required, calculating all inner nodes dynamically. Recomputation is avoided by using lookup cache per node index.


crates/proto/src/domain/merkle.rs
bobbinth, line 1: Looking at the complexity of the code, I wonder if this should actually live in `miden-base`. The main reason is that the client would need to deserialize this data but the client doesn't get anything from the node except for protobuf files.

In `miden-base`, we could attach the logic to the `PartialStorageMap` struct. Basically, we need two things there:

- Given a `PartialStorageMap` we need to get `SmtLeaf`s and `InnerNode`s from it. Not sure what the name of the function would be - but getting this data shouldn't be too difficult.
- Given a set of `SmtLeaf`s and `InnerNode`s, we need a constructor that would build the underlying `PartialSmt` from this.

Then, here in `miden-node` we'll just need to convert these to/from protobuf structs - so, the logic will be pretty straight-forward.

proto/proto/store/rpc.proto
```diff
@@ -137,12 +137,12 @@ message AccountProofs {
         // State header for public accounts.
         message AccountStateHeader {
             // Represents a single storage slot with the requested keys and their respective values.
-            message StorageSlotMapProof {
+            message StorageSlotMapPartialSmt {
```
bobbinth, line 140: I think this message could be just `PartialStorageMap`.

crates/proto/src/domain/merkle.rs
```diff
@@ -206,3 +211,226 @@ impl From<SmtProof> for proto::primitives::SmtOpening {
...
+    }
+}
+impl TryFrom<proto::primitives::NodeIndex> for NodeIndex {
+    type Error = ConversionError;
+    fn try_from(index: proto::primitives::NodeIndex) -> Result<Self, Self::Error> {
+        let depth = u8::try_from(index.depth)?;
+        let value = index.value;
+        Ok(NodeIndex::new(depth, value)?)
+    }
+}
```
SantiagoPittella, line 232: Probably nothing, but we encode the depth as u32 and then cast it to u8. Should a comment be added about why is it always valid?

PR #1180, opened 2025-08-25
fix: remove deleted table reference

tomyrd:
Fixes a CI error we were getting in miden-client when initializing the genesis block with specific accounts.


igamigo:
Thanks for fixing this, was my bad for leaving this

PR #1181, opened 2025-08-25
feat: include genesis block's commitment in Status

### LINKED ISSUE #1088: Include genesis block's commitment in `Status`
Add the genesis block's commitment to the status message of the RPC component.

The value should be available after #1084.

Related comment https://github.com/0xMiden/miden-node/pull/1084#discussion_r2206007608.

varun-doshi:
Fixes #1088

### DIFF
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -77,10 +77,14 @@ impl RpcService {
         }
     }
 
+    /// Sets the genesis commitment, returning an error if it is already set.
+    ///
+    /// Required since `RpcService::new()` sets up the `store` which is used to fetch the
+    /// `genesis_commitment`.
     pub fn set_genesis_commitment(&self, commitment: Word) -> anyhow::Result<()> {
         self.genesis_commitment
             .set(commitment)
-            .map_err(|_| anyhow::anyhow!("Genesis commitment already set"))?;
+            .map_err(|_| anyhow::anyhow!("genesis commitment already set"))?;
         Ok(())
     }
 
@@ -451,7 +455,7 @@ impl api_server::Api for RpcService {
 
         let genesis_commitment = self.genesis_commitment.get().ok_or({
             tracing::warn!(target: COMPONENT, "Failed to fetch genesis header");
-            Status::internal("failed to fetch genesis header")
+            Status::failed_precondition("genesis header was not set")
         })?;
 
         Ok(Response::new(proto::rpc::RpcStatus {
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -436,6 +436,16 @@ impl api_server::Api for RpcService {
             None
         };
 
+        let genesis = self
+            .get_genesis_header_with_retry()
+            .await
+            .context("Fetching genesis header from store");
+
+        if let Err(err) = genesis {
+            tracing::warn!(target: COMPONENT, "Failed to fetch genesis header: {}", err);
+            return Err(Status::invalid_argument("Failed to fetch genesis header".to_string()));
+        }
+
         Ok(Response::new(proto::rpc::RpcStatus {
             version: env!("CARGO_PKG_VERSION").to_string(),
             store: store_status.or(Some(proto::rpc_store::StoreStatus {
@@ -449,6 +459,7 @@ impl api_server::Api for RpcService {
                     version: "-".to_string(),
                 },
             )),
+            genesis_commitment: Some(genesis.unwrap().commitment().into()),
         }))
     }
 }
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -28,7 +28,6 @@ use miden_objects::transaction::ProvenTransaction;
 use miden_objects::utils::serde::Deserializable;
 use miden_objects::{MAX_NUM_FOREIGN_ACCOUNTS, MIN_PROOF_SECURITY_LEVEL, Word};
 use miden_tx::TransactionVerifier;
-use tokio::sync::OnceCell;
 use tonic::{IntoRequest, Request, Response, Status};
 use tracing::{debug, info, instrument};
 use url::Url;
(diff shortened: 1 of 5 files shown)


crates/rpc/src/server/api.rs
```diff
@@ -436,6 +436,16 @@ impl api_server::Api for RpcService {
...
 
+        let genesis = self
+            .get_genesis_header_with_retry()
+            .await
+            .context("Fetching genesis header from store");
+
+        if let Err(err) = genesis {
+            tracing::warn!(target: COMPONENT, "Failed to fetch genesis header: {}", err);
+            return Err(Status::invalid_argument("Failed to fetch genesis header".to_string()));
+        }
```
igamigo: Not a strong opinion but you can `map_err` here so that you don't have to call `unwrap()` later. Also, `invalid_argument` is probably the incorrect error code here:

```suggestion
        let genesis = self
            .get_genesis_header_with_retry()
            .await
            .map_err(|err| {
                tracing::warn!(target: COMPONENT, "Failed to fetch genesis header: {err}");
                Status::internal("failed to fetch genesis header")
            })?;
```

crates/rpc/src/server/api.rs
```diff
@@ -68,7 +70,18 @@ impl RpcService {
...
+            block_producer,
+            genesis_commitment: OnceCell::new(),
+        }
+    }
+
+    pub fn set_genesis_commitment(&self, commitment: Word) -> anyhow::Result<()> {
+        self.genesis_commitment
+            .set(commitment)
+            .map_err(|_| anyhow::anyhow!("Genesis commitment already set"))?;
+        Ok(())
```
SantiagoPittella: Should this be set on the struct initialization? cc @igamigo

crates/rpc/src/server/api.rs
```diff
@@ -68,7 +70,18 @@ impl RpcService {
...
+            store,
+            block_producer,
+            genesis_commitment: OnceCell::new(),
+        }
+    }
+
+    pub fn set_genesis_commitment(&self, commitment: Word) -> anyhow::Result<()> {
+        self.genesis_commitment
+            .set(commitment)
+            .map_err(|_| anyhow::anyhow!("Genesis commitment already set"))?;
```
igamigo:
```suggestion
            .map_err(|_| anyhow::anyhow!("genesis commitment already set"))?;
```

crates/rpc/src/server/api.rs
```diff
@@ -436,6 +449,11 @@ impl api_server::Api for RpcService {
             None
         };
 
+        let genesis_commitment = self.genesis_commitment.get().ok_or({
+            tracing::warn!(target: COMPONENT, "Failed to fetch genesis header");
+            Status::internal("failed to fetch genesis header")
```
igamigo:
```suggestion
            Status::failed_precondition("genesis header was not set")
```

crates/rpc/src/server/api.rs
igamigo: I think ideally you would set this on construction, but I'm not sure this is doable right now without changing the semantics a bit. Basically you would need to change `new` to be `async` and also it would not return until the store was up and running (so that `get_genesis_header_with_retry` returns). 
The current way is not as clean and adds some (very manageable) complexity, but can't think of a better way without refactoring other sections too much.

crates/rpc/src/server/api.rs
```diff
@@ -68,7 +70,18 @@ impl RpcService {
...
 
-        Self { store, block_producer }
+        Self {
+            store,
+            block_producer,
+            genesis_commitment: OnceCell::new(),
+        }
+    }
+
+    pub fn set_genesis_commitment(&self, commitment: Word) -> anyhow::Result<()> {
```
igamigo: Let's add some doc comments to mention what this function does and why it's needed.

igamigo (approved):
LGTM. As mentioned in the comments, not sure if there is a better way to set this up without refactoring some other pieces of code, but overall the added complexity seems low enough that this is a good tradeoff (and we could add a `TODO` or an issue to revisit this). Would be nice to see if @sergerad has any opinions on this as well.

SantiagoPittella (approved):

proto/proto/rpc.proto
```diff
@@ -101,4 +101,7 @@ message RpcStatus {
 
     // The block producer status.
     block_producer.BlockProducerStatus block_producer = 4;
+
+    // The genesis commitment.
+    primitives.Digest genesis_commitment = 5;
```
bobbinth: nit: I'd put this filed after `version` and change it's number to `2`.

crates/rpc/src/server/api.rs
```diff
@@ -40,6 +41,7 @@ use crate::COMPONENT;
 pub struct RpcService {
     store: StoreRpcClient,
     block_producer: Option<BlockProducerClient>,
+    genesis_commitment: OnceCell<Word>,
```
bobbinth: Why does this need to be `OnceCell<Word>` rather than just a simple `Word`?

bobbinth (commented):
Looks good! Thank you! I left a couple of small comments/questions inline.

crates/rpc/src/server/api.rs
varun-doshi: oh it doesn't actually need to be a `OnceCell`
I thought since this is something that will be set only once...`OnceCell` provides guarentees out of the box that a value can only be set once.
I can change it to a Word if that's what is preferred.

bobbinth: Yeah - I'd probably go with just `Word`. It could initially be set to `Word::empty()` and if we really wanted to make sure it can be only set once, we could check before updating if it is not `Word::empty()`, and if so, return an error.

varun-doshi: Understood
Changes made!

crates/rpc/src/server/api.rs
```diff
@@ -40,6 +40,7 @@ use crate::COMPONENT;
 pub struct RpcService {
     store: StoreRpcClient,
     block_producer: Option<BlockProducerClient>,
+    genesis_commitment: Word,
```
sergerad: I made a PR to your branch to move this to option also

https://github.com/varun-doshi/miden-node/pull/1

sergerad (approved):

crates/rpc/src/server/api.rs
varun-doshi: Merged!

bobbinth (approved):
Looks good! Thank you!

PR #1184, opened 2025-08-29
fix: release-plz ordering

mmagician:
Main fix (to address the [failed CI publishing job](https://github.com/0xMiden/miden-node/actions/runs/17316443413):
- trick release-plz to release `miden-node-store` before `miden-node-block-producer`

release-plz makes some opinionated choices about release order. Specifically, it allows for cyclic (dev) dependencies, but as a result, dev-dependencies are not published first - unless the dev-dependency is mentioned in a feature (which is the trick applied here). See also [this comment](https://github.com/release-plz/release-plz/issues/1316#issuecomment-1971996730). 

---

Further improvements:
- skip publishing `stress-test` binary
- skip publishing `test-macro` crate (and remove the reference to non-existent `README.md`)

### DIFF
--- a/crates/block-producer/Cargo.toml
+++ b/crates/block-producer/Cargo.toml
@@ -15,9 +15,9 @@ version.workspace      = true
 workspace = true
 
 [features]
+release-plz    = ["miden-node-store/default"]        # needed for correct release order
 testing        = []
 tracing-forest = ["miden-node-utils/tracing-forest"]
-release-plz    = ["miden-node-store/default"] # needed for correct release order
 
 [dependencies]
 anyhow                     = { workspace = true }


crates/block-producer/Cargo.toml
Mirko-von-Leipzig, line 1: I see the problem but I really don't like the solution because it means that the store methods will now pop up as options when developing in the block-producer.

I'll have a look around for alternatives..

Mirko-von-Leipzig: This might be a silly question - but `cargo publish --workspace` now exists. Could we not just use that if we're only publishing?

It looks like they already [solved](https://github.com/rust-lang/cargo/issues/15412) this problem.

Mirko-von-Leipzig: Another option could be to replace `dev` dependency with a path

```toml
miden-node-store      = { path = "../store" }
# instead of
miden-node-store      = { workspace = true }
```

bobbinth: Yes, if there is a way to avoid adding release-specific features, I'd also prefer to avoid this.

PR #1191, opened 2025-09-01
feat: add pagination to CheckNullifiersByPrefix

### LINKED ISSUE #1186: Add pagination to the `CheckNullifiersByPrefix` endpoint
We should make sure that `CheckNullifiersByPrefix` never turns too much data in a single response. For this, we can use a similar strategy as we did with `SyncStorageMaps` and `SyncAccountVault` endpoints. Specifically:

We can impose a limit on the number of nullifiers returned from the database, and if the returned number of records is greater than or equal to this limit, we discard the last block of data. The user then will have to make a new request with the last returned block number + 1.

As with other two responses, we should try to limit the response size to ~ 2MB - though, at some point, we should make this a configurable setting.

SantiagoPittella:
closes #1186

### DIFF
--- a/crates/store/README.md
+++ b/crates/store/README.md
@@ -45,9 +45,15 @@ Returns a nullifier proof for each of the requested nullifiers.
 
 ### CheckNullifiersByPrefix
 
-Returns a list of nullifiers that match the specified prefixes and are recorded in the node.
+Returns nullifier synchronization data for a set of prefixes within a given block range. This method allows
+clients to efficiently track nullifier creation by retrieving only the nullifiers produced between two blocks.
 
-Only 16-bit prefixes are supported at this time.
+Caller specifies the `prefix_len` (currently only 16), the list of prefix values (`nullifiers`), and the block
+range (`block_from`, optional `block_to`). The response includes all matching nullifiers created within that
+range, the last block included in the response (`block_num`), and the current chain tip (`chain_tip`).
+
+If the response is chunked (i.e., `block_num < block_to`), continue by issuing another request with
+`block_from = block_num + 1` to retrieve subsequent updates.
 
 ---
 
--- a/crates/store/README.md
+++ b/crates/store/README.md
@@ -45,8 +45,7 @@ Returns a nullifier proof for each of the requested nullifiers.
 
 ### CheckNullifiersByPrefix
 
-Returns nullifier synchronization data for a set of prefixes within a given block range. This method allows
-clients to efficiently track nullifier creation by retrieving only the nullifiers produced between two blocks.
+Returns nullifier synchronization data for a set of prefixes within a given block range. This method allows clients to efficiently track nullifier creation by retrieving only the nullifiers produced within a specific range of blocks.
 
 Caller specifies the `prefix_len` (currently only 16), the list of prefix values (`nullifiers`), and the block
 range (`block_from`, optional `block_to`). The response includes all matching nullifiers created within that
--- a/crates/store/README.md
+++ b/crates/store/README.md
@@ -14,7 +14,6 @@ The full gRPC API can be found [here](../../proto/proto/store.proto).
 <!--toc:start-->
 - [ApplyBlock](#applyblock)
 - [CheckNullifiers](#checknullifiers)
-- [CheckNullifiersByPrefix](#checknullifiersbyprefix)
 - [GetAccountDetails](#getaccountdetails)
 - [GetAccountProofs](#getaccountproofs)
 - [GetBlockByNumber](#getblockbynumber)
@@ -24,6 +23,7 @@ The full gRPC API can be found [here](../../proto/proto/store.proto).
 - [GetNotesById](#getnotesbyid)
 - [GetTransactionInputs](#gettransactioninputs)
 - [GetNoteScriptByRoot](#getnotescriptbyroot)
+- [SyncNullifiers](#syncnullifiers)
 - [SyncAccountVault](#syncaccountvault)
 - [SyncNotes](#syncnotes)
 - [SyncState](#syncstate)
@@ -41,19 +41,6 @@ Applies changes of a new block to the DB and in-memory data structures. Raw bloc
 ### CheckNullifiers
 
 Returns a nullifier proof for each of the requested nullifiers.
-
----
-
-### CheckNullifiersByPrefix
-
-Returns nullifier synchronization data for a set of prefixes within a given block range. This method allows clients to efficiently track nullifier creation by retrieving only the nullifiers produced within a specific range of blocks.
-
-Caller specifies the `prefix_len` (currently only 16), the list of prefix values (`nullifiers`), and the block
-range (`block_from`, optional `block_to`). The response includes all matching nullifiers created within that
-range, the last block included in the response (`block_num`), and the current chain tip (`chain_tip`).
-
-If the response is chunked (i.e., `block_num < block_to`), continue by issuing another request with
-`block_from = block_num + 1` to retrieve subsequent updates.
 
 ---
 
@@ -111,6 +98,19 @@ Used by the `block-producer` to query state required to verify a submitted trans
 ### GetNoteScriptByRoot
 
 Returns the script for a note by its root.
+
+---
+
+### SyncNullifiers
+
+Returns nullifier synchronization data for a set of prefixes within a given block range. This method allows clients to efficiently track nullifier creation by retrieving only the nullifiers produced within a specific range of blocks.
+
+Caller specifies the `prefix_len` (currently only 16), the list of prefix values (`nullifiers`), and the block
+range (`block_from`, optional `block_to`). The response includes all matching nullifiers created within that
+range, the last block included in the response (`block_num`), and the current chain tip (`chain_tip`).
+
+If the response is chunked (i.e., `block_num < block_to`), continue by issuing another request with
+`block_from = block_num + 1` to retrieve subsequent updates.
 
 ---
 
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -276,8 +276,9 @@ impl Db {
         &self,
         prefix_len: u32,
         nullifier_prefixes: Vec<u32>,
-        block_num: BlockNumber,
-    ) -> Result<Vec<NullifierInfo>> {
+        block_from: BlockNumber,
+        block_to: BlockNumber,
+    ) -> Result<(Vec<NullifierInfo>, BlockNumber)> {
         assert_eq!(prefix_len, 16, "Only 16-bit prefixes are supported");
 
         self.transact("nullifieres by prefix", move |conn| {
@@ -287,7 +288,8 @@ impl Db {
                 conn,
                 prefix_len as u8,
                 &nullifier_prefixes[..],
-                block_num,
+                block_from,
+                block_to,
             )
         })
         .await
--- a/crates/store/src/db/models/queries/nullifiers.rs
+++ b/crates/store/src/db/models/queries/nullifiers.rs
@@ -61,7 +61,7 @@ pub(crate) fn select_nullifiers_by_prefix(
     pub const NULLIFIER_BYTES: usize = 32; // digest size (nullifier)
     pub const BLOCK_NUM_BYTES: usize = 4; // 32 bits per block number
     pub const ROW_OVERHEAD_BYTES: usize = NULLIFIER_BYTES + BLOCK_NUM_BYTES; // 36 bytes
-    pub const ROW_LIMIT: usize = (MAX_PAYLOAD_BYTES / ROW_OVERHEAD_BYTES) + 1;
+    pub const MAX_ROWS: usize = MAX_PAYLOAD_BYTES / ROW_OVERHEAD_BYTES;
 
     assert_eq!(prefix_len, 16, "Only 16-bit prefixes are supported");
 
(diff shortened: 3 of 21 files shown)


crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -23,23 +23,39 @@ use crate::db::models::conv::{SqlTypeConvert, nullifier_prefix_to_raw_sql};
 use crate::db::models::utils::{get_nullifier_prefix, vec_raw_try_into};
 use crate::db::{NullifierInfo, schema};
 
-/// Returns nullifiers filtered by prefix and block creation height.
+/// Returns nullifiers filtered by prefix within a block number range.
```
drahnr, line 26: is `block_to` inclusive or exclusive?

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -50,17 +66,37 @@ pub(crate) fn select_nullifiers_by_prefix(
...
-            .filter(schema::nullifiers::block_num.ge(block_num.to_raw_sql()))
+            .filter(schema::nullifiers::block_num.ge(block_from.to_raw_sql()))
+            .filter(schema::nullifiers::block_num.le(block_to.to_raw_sql()))
             .order(schema::nullifiers::block_num.asc())
+            .limit(i64::try_from(ROW_LIMIT).expect("limit fits within i64"))
             .load::<NullifierWithoutPrefixRawRow>(conn)?;
-    vec_raw_try_into(nullifiers_raw)
+
+    // Discard the last block in the response (assumes more than one block may be present)
+    if raw.len() >= ROW_LIMIT {
```
drahnr: nit: With newer rust versions we should be able to do `if let Some(last) = raw.last() && raw.len() >= ROW_LIMIT {`

crates/store/src/db/models/queries/nullifiers.rs
SantiagoPittella: It is inclusive. I will add it to the docs.

docs/src/user/rpc.md
```diff
@@ -30,9 +30,15 @@ Request proofs for a set of nullifiers.
 
 ## CheckNullifiersByPrefix
```
bobbinth: I'm thinking we should rename this to `SyncNullifiers` endpoint. This would be consistent with having all endpoints which may return "chunked" data to have a `Sync` prefix.

proto/proto/store/rpc.proto
```diff
@@ -186,7 +186,9 @@ message CheckNullifiersByPrefixRequest {
     // to `prefix_len`.
     repeated uint32 nullifiers = 2;
     // Block number from which the nullifiers are requested (inclusive).
-    fixed32 block_num = 3;
+    fixed32 block_from = 3;
+    // Block number up to which to check. If not specified, checks up to the latest block.
+    optional fixed32 block_to = 4;
```
bobbinth: I think we should adapt a convention where all "sync" requests specify block from/to as the first two fields. For this message this would look something like:
```proto
message CheckNullifiersByPrefixRequest {
    fixed32 block_from = 1;
    uint32 prefix_len = 3;
    repeated uint32 nullifiers = 3;
}
```
This is already the case for `SyncStorageMapsRequest` and `SyncAccountVaultRequest`, and we should make it the same for `SyncNotesRequest` (in a different PR - let's make an issue for this).

proto/proto/store/rpc.proto
```diff
@@ -201,6 +203,14 @@ message CheckNullifiersByPrefixResponse {
...
     // List of nullifiers matching the prefixes specified in the request.
     repeated NullifierUpdate nullifiers = 1;
+    // The block number of the last check included in this response.
+    //
+    // For chunked responses, this may be less than request.block_to.
+    // If it is less than request.block_to, the user is expected to make a subsequent request
+    // starting from the next block to this one (ie, request.block_from = block_num + 1).
+    fixed32 block_num = 2;
+    // Current chain tip
+    fixed32 chain_tip = 3;
```
bobbinth: Similar to the previous comment, I'd change the order of these to be:
```proto
message CheckNullifiersByPrefixResponse {
    ...
    fixed32 chain_tip = 1;
    fixed32 block_num = 2;
    repeated NullifierUpdate nullifiers = 3;
}
```
And if we could update `SyncStorageMapsResponse` and `SyncAccountVaultResponse` to follow the same convention, that would be great.

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -23,23 +23,49 @@ use crate::db::models::conv::{SqlTypeConvert, nullifier_prefix_to_raw_sql};
...
-/// A vector of [`NullifierInfo`] with the nullifiers and the block height at which they were
+/// Range and pagination semantics:
+/// - Both `block_from` and `block_to` are inclusive bounds.
+/// - To keep responses ≤ ~2MB, an internal row limit is enforced. When the limit is hit and the
+///   result spans multiple blocks, the last block in the page is dropped entirely, and
+///   `last_block_included` is set to the block number immediately before that dropped block.
+///   Clients should resume with `block_from = last_block_included + 1`.
+/// - If all rows belong to a single block and hit the limit, this function returns an empty
+///   `nullifiers` page with `last_block_included = that_block - 1`. Intra-block pagination is not
+///   supported; callers may need to narrow the range.
```
bobbinth, line 42: A block can have a maximum of $2^{16}$ nullifiers. So, if each nullifier is 36 bytes, at most a single block can take is 2.25MB (and this will return all nullifiers in a block). While hitting this limit is highly unlikely, I wonder if we should bump up the limit to something like 2.5MB to make sure this situation never happens.

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -23,23 +23,49 @@ use crate::db::models::conv::{SqlTypeConvert, nullifier_prefix_to_raw_sql};
...
     prefix_len: u8,
     nullifier_prefixes: &[u16],
-    block_num: BlockNumber,
-) -> Result<Vec<NullifierInfo>, DatabaseError> {
+    block_from: BlockNumber,
+    block_to: BlockNumber,
+) -> Result<(Vec<NullifierInfo>, BlockNumber), DatabaseError> {
+    pub const MAX_PAYLOAD_BYTES: usize = 2 * 1024 * 1024; // 2 MB
+    pub const NULLIFIER_BYTES: usize = 32; // digest size
+    pub const ROW_OVERHEAD_BYTES: usize = NULLIFIER_BYTES + size_of::<i64>(); // 40 bytes
```
bobbinth: Why do we add `size_of::<i64>()` here? The block number is 32 bits - so, shouldn't this be `size_of::<u32>()`?

drahnr: @SantiagoPittella in the database the block number is an `i64` but the actual required in-memory representation is `u32` - we do this to guarantee ordering over long periods of time.

crates/store/src/state.rs
```diff
@@ -426,10 +426,11 @@ impl State {
         &self,
         prefix_len: u32,
         nullifier_prefixes: Vec<u32>,
-        block_num: BlockNumber,
-    ) -> Result<Vec<NullifierInfo>, DatabaseError> {
+        block_from: BlockNumber,
+        block_to: BlockNumber,
```
Mirko-von-Leipzig, line 430: Could we use ranges here to make the inclusive/exclusive explicit?

crates/store/README.md
```diff
@@ -45,9 +45,15 @@ Returns a nullifier proof for each of the requested nullifiers.
 
 ### CheckNullifiersByPrefix
 
-Returns a list of nullifiers that match the specified prefixes and are recorded in the node.
+Returns nullifier synchronization data for a set of prefixes within a given block range. This method allows
+clients to efficiently track nullifier creation by retrieving only the nullifiers produced between two blocks.
```
Mirko-von-Leipzig:
```suggestion
clients to efficiently track nullifier creation by retrieving only the nullifiers produced within a specific range of blocks.
```

crates/store/src/server/rpc_api.rs
```diff
@@ -96,14 +96,19 @@ impl rpc_server::Rpc for StoreApi {
             return Err(Status::invalid_argument("Only 16-bit prefixes are supported"));
         }
 
-        let nullifiers = self
+        let chain_tip = self.state.latest_block_num().await.as_u32();
```
Mirko-von-Leipzig: Nit: I think probably makes more sense to add `.as_u32()` in the response segment.

```suggestion
        let chain_tip = self.state.latest_block_num().await;
```

proto/proto/store/rpc.proto
Mirko-von-Leipzig: Should we create a dedicated `BlockRange` message for it?

crates/store/src/db/mod.rs
```diff
@@ -276,8 +276,9 @@ impl Db {
         &self,
         prefix_len: u32,
         nullifier_prefixes: Vec<u32>,
-        block_num: BlockNumber,
-    ) -> Result<Vec<NullifierInfo>> {
+        block_from: BlockNumber,
+        block_to: BlockNumber,
```
Mirko-von-Leipzig, line 287: Similarly, I suggest we use a range to represent this

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -23,23 +23,49 @@ use crate::db::models::conv::{SqlTypeConvert, nullifier_prefix_to_raw_sql};
...
+///   that case, the caller should re-issue the query with `block_from = last_block_included + 1` to
+///   continue.
 pub(crate) fn select_nullifiers_by_prefix(
     conn: &mut SqliteConnection,
     prefix_len: u8,
     nullifier_prefixes: &[u16],
-    block_num: BlockNumber,
-) -> Result<Vec<NullifierInfo>, DatabaseError> {
+    block_from: BlockNumber,
+    block_to: BlockNumber,
```
Mirko-von-Leipzig, line 56: Also a range

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -23,23 +23,49 @@ use crate::db::models::conv::{SqlTypeConvert, nullifier_prefix_to_raw_sql};
...
+) -> Result<(Vec<NullifierInfo>, BlockNumber), DatabaseError> {
+    pub const MAX_PAYLOAD_BYTES: usize = 2 * 1024 * 1024; // 2 MB
+    pub const NULLIFIER_BYTES: usize = 32; // digest size
+    pub const ROW_OVERHEAD_BYTES: usize = NULLIFIER_BYTES + size_of::<i64>(); // 40 bytes
+    pub const ROW_LIMIT: usize = (MAX_PAYLOAD_BYTES / ROW_OVERHEAD_BYTES) + 1;
```
Mirko-von-Leipzig: I know we've done this previously, but I would prefer calling this `MAX_ROWS` and then 

```rust
// Request an additional row so we can determine whether this is the last page.
.limit(i64::try_from(MAX_ROWS + 1).expect("limit fits in i64"))
```

crates/store/src/state.rs
```diff
@@ -426,10 +426,11 @@ impl State {
...
+    ) -> Result<(Vec<NullifierInfo>, BlockNumber), DatabaseError> {
         self.db
-            .select_nullifiers_by_prefix(prefix_len, nullifier_prefixes, block_num)
+            .select_nullifiers_by_prefix(prefix_len, nullifier_prefixes, block_from, block_to)
```
drahnr, line 433: Out of scope: I propose introducing a `Range` like type to simplify our API and reduce the need to document inclusive/exclusive boundaries.

docs/src/user/rpc.md
drahnr: I'd prefer to have a `Paged`-prefix or similar, `Sync` implies what you do with the data, but imho it's more useful to describe the behavior of the function at hand.

drahnr (approved):
Bar the open comments, looks good!

@SantiagoPittella @Mirko-von-Leipzig I'd defer the `BlockRange` proto / `Range` rust type refactor to a separate follow up PR / #1193 if this lands before that.

crates/store/src/db/mod.rs
drahnr: I'll do this as part of #1193

crates/store/src/db/models/queries/nullifiers.rs
Mirko-von-Leipzig: Deferred to #1193

docs/src/user/rpc.md
SantiagoPittella: Should I opened an issue to discuss whether to use `Sync`, `Paged` or another prefix?

proto/proto/store/rpc.proto
SantiagoPittella: I like the `BlockRange` idea, i'm opening a new issue to address this and implement it for all the sync related methods.

SantiagoPittella: Opened https://github.com/0xMiden/miden-node/issues/1204 , we can do something similar to the `BlockRange` message for the responses too, since all have the `chain_tip` and block_num fields.

crates/store/src/db/models/queries/nullifiers.rs
SantiagoPittella: Reading [the docs](https://docs.rs/tonic/latest/tonic/#max-message-size), the max message size by default is 4MB so it shouldn't be a problem to increase this to 2.5MB

crates/store/src/db/models/queries/nullifiers.rs
SantiagoPittella: Updated!

crates/store/src/state.rs
SantiagoPittella: I can either do this in this PR or as part of https://github.com/0xMiden/miden-node/issues/1204 , wdyt?

docs/src/user/rpc.md
bobbinth: I'd still probably prefer with `SyncNullifiers`. One reason is that I'd prefer our endpoints to start with an action verb (though, we could of course do `GetPagedNullifiers`). The other reason I think it better describes what this endpoint is for - i.e., it is for synchronizing the state of the client with the state of the network (and it should be used together with other `Sync*` endpoints).

I would do the renaming in this PR.

SantiagoPittella: Just pushed a commit renaming to `SyncNullifiers`.

PR #1194, opened 2025-09-02
chore: remove decorators from Note's masm

SantiagoPittella:
partially address https://github.com/0xMiden/miden-base/issues/1812

> **This PR is using `main` as the base branch**

Removes decoratos from the note before inserting it in the DB.

### DIFF
--- a/crates/store/src/db/models/queries/insertions.rs
+++ b/crates/store/src/db/models/queries/insertions.rs
@@ -380,13 +380,18 @@ pub(crate) fn insert_notes(
 ) -> Result<usize, DatabaseError> {
     let count = diesel::insert_into(schema::notes::table)
         .values(Vec::from_iter(notes.iter().map(|(note, nullifier)| {
+            // Remove decorators from the script
             let stripped_details = note.details.as_ref().map(|details| {
                 let mut mast = details.script().mast().clone();
-                Arc::make_mut(&mut mast).strip_decorators();
-                let script = NoteScript::from_parts(mast, details.script().entrypoint());
-                let recipient =
-                    NoteRecipient::new(details.serial_num(), script, details.inputs().clone());
-                NoteDetails::new(details.assets().clone(), recipient)
+                if mast.decorators().is_empty() {
+                    details.clone()
+                } else {
+                    Arc::make_mut(&mut mast).strip_decorators();
+                    let script = NoteScript::from_parts(mast, details.script().entrypoint());
+                    let recipient =
+                        NoteRecipient::new(details.serial_num(), script, details.inputs().clone());
+                    NoteDetails::new(details.assets().clone(), recipient)
+                }
             });
 
             let modified_note = NoteRecord {


Mirko-von-Leipzig:
> LGTM. Not that it makes a difference, but would this cover the network transaction builder mempool events? AFAIK when a transaction gets added to the mempool, new notes are sent to the ntx builder so maybe it would receive notes with the decorators.

Yeah it would still have the decorators. Good point. Maybe this should be done in the RPC then to strip them at arrival? Bit of a waste but its the single point of entry for now..

SantiagoPittella:
Should this be moved to `submit_proven_transaction` and `submit_proven_batch` handlers?

SantiagoPittella:
Implemented this change, leaving the decorator removal before inserting in the DB too. wdyt?

@bobbinth @Mirko-von-Leipzig @igamigo

crates/rpc/src/server/api.rs
Mirko-von-Leipzig, line 1: This works but feels extremely fragile and expensive since we build everything twice. I don't think there is currently a better way though.

Maybe notes should be stripped as part of building a transaction already? I'm a bit unclear on when the decorators are desired.

As a side note `ProvenTransactionBuilder::new` feels like it really needs wrapper types since it has many words and block numbers which can be trivially swapped around. cc @PhilippGackstatter

Mirko-von-Leipzig (approved):

crates/rpc/src/server/api.rs
PhilippGackstatter: > As a side note `ProvenTransactionBuilder::new` feels like it really needs wrapper types since it has many words and block numbers which can be trivially swapped around.


Agreed, but we need https://github.com/0xMiden/crypto/issues/430 for this first.

PhilippGackstatter: > Maybe notes should be stripped as part of building a transaction already? I'm a bit unclear on when the decorators are desired.

In a mock-chain context, we often consume notes that were created by a previous transaction, so we'd ideally retain the decorators. It would be sufficient to have that on `testing` though, and strip them in non-`testing` during `ProvenTransactionBuilder::build`. I think that could work - but there may be more cases where we'd want decorators that I'm not thinking about.

crates/store/src/db/models/queries/insertions.rs
```diff
@@ -378,11 +379,22 @@ pub(crate) fn insert_notes(
...
+                    NoteRecipient::new(details.serial_num(), script, details.inputs().clone());
+                NoteDetails::new(details.assets().clone(), recipient)
+            });
+
+            let modified_note = NoteRecord {
+                details: stripped_details,
+                ..note.clone()
+            };
+            NoteInsertRowRaw::from((modified_note, *nullifier))
+        })))
```
igamigo: Would it be better to check if the notes had decorators before attempting to strip them as Bobbin suggested? Not sure what that would look like in terms of code but it does seem like we would save ourselves the note reconstruction, although maybe it doesn't make too much of a difference because we are cloning the note anyway.

igamigo (approved):
LGTM in terms of code. Although it feels redundant to do it in the DB if we are already stripping them at the only available RPC entrypoints, if we do it we might want to check if the note had decorators before attempting to do the reconstruction as suggested [here](https://github.com/0xMiden/miden-node/pull/1194#issuecomment-3246309825).

crates/store/src/db/models/queries/insertions.rs
SantiagoPittella: I added a check before striping decorators in the insertion part

crates/rpc/src/server/api.rs
```diff
@@ -325,6 +327,38 @@ impl api_server::Api for RpcService {
...
+            tx.account_update().final_state_commitment(),
+            tx.account_update().account_delta_commitment(),
+            tx.ref_block_num(),
+            tx.ref_block_commitment(),
+            tx.fee(),
+            tx.expiration_block_num(),
+            tx.proof().clone(),
+        )
+        .account_update_details(tx.account_update().details().clone())
+        .add_input_notes(tx.input_notes().iter().cloned());
```
bobbinth, line 343: This is fine for now, but ideally, we should have a less error-prone way to strip decorators from any note in a transaction. I'm not sure what this way would be - but let's create an issue about for this.

PR #1193, opened 2025-09-02
store/queries: cleanup

drahnr:
Addresses parts of

Ref #1134
Ref #1161

PR #1196, opened 2025-09-02
feat: add GetNoteScriptByRoot endpoint

### LINKED ISSUE #1187: Implement `GetNoteScriptByRoot` endpoint
This would be a pretty simple endpoint: it would take note script root as input and return the note script. It may make sense to create a `NoteScript` message for this. It should be pretty simple - something like:
```proto
message NoteScript {
    unit32 entrypoint = 1;
    bytes mast = 2;
}
```

SantiagoPittella:
closes #1187

### DIFF
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -506,6 +506,23 @@ impl api_server::Api for RpcService {
             genesis_commitment: self.genesis_commitment.map(Into::into),
         }))
     }
+
+    #[instrument(
+        parent = None,
+        target = COMPONENT,
+        name = "rpc.server.get_note_script_by_root",
+        skip_all,
+        ret(level = "debug"),
+        err
+    )]
+    async fn get_note_script_by_root(
+        &self,
+        request: Request<proto::note::NoteRoot>,
+    ) -> Result<Response<proto::rpc_store::MaybeNoteScript>, Status> {
+        debug!(target: COMPONENT, request = ?request);
+
+        self.store.clone().get_note_script_by_root(request).await
+    }
 }
 
 // LIMIT HELPERS
--- a/crates/store/src/db/models/queries/notes.rs
+++ b/crates/store/src/db/models/queries/notes.rs
@@ -17,7 +17,7 @@ use diesel::query_dsl::methods::SelectDsl;
 use diesel::query_dsl::{QueryDsl, RunQueryDsl};
 use diesel::sqlite::Sqlite;
 use diesel::{JoinOnDsl, NullableExpressionMethods, OptionalExtension, SqliteConnection};
-use miden_lib::utils::Deserializable;
+use miden_lib::utils::{Deserializable, Serializable};
 use miden_node_utils::limiter::{
     QueryParamAccountIdLimit,
     QueryParamLimiter,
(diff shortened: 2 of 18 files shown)


crates/store/src/server/rpc_api.rs
```diff
@@ -462,4 +465,34 @@ impl rpc_server::Rpc for StoreApi {
             chain_tip: self.state.latest_block_num().await.as_u32(),
         }))
     }
+
+    #[instrument(
+        parent = None,
+        target = COMPONENT,
+        name = "store.rpc_server.get_note_script_by_root",
```
bobbinth, line 472: Do we also need to add an entry for this in https://github.com/0xMiden/miden-node/blob/307b3ca0ffecc08b3db91cd49b12fd6a9978836e/crates/utils/src/tracing/grpc.rs#L44 - and maybe in other similar places?

crates/rpc/src/server/api.rs
```diff
@@ -506,6 +506,23 @@ impl api_server::Api for RpcService {
             genesis_commitment: self.genesis_commitment.map(Into::into),
         }))
     }
+
+    #[instrument(
+        parent = None,
+        target = COMPONENT,
+        name = "rpc.server.get_note_script_by_root",
```
bobbinth, line 513: Same comment as above.

SantiagoPittella: Done!

crates/store/src/db/models/queries/notes.rs
```diff
@@ -250,6 +250,29 @@ pub(crate) fn select_note_inclusion_proofs(
...
+    //     root = ?1;
+
+    let raw = SelectDsl::select(schema::note_scripts::table, schema::note_scripts::script)
+        .filter(schema::note_scripts::script_root.eq(root.to_bytes()))
+        .get_result::<Vec<u8>>(conn)
+        .optional()?;
+
+    let note_script = raw.map(|bytes| NoteScript::from_bytes(&bytes[..])).transpose()?;
+
+    Ok(note_script)
```
Mirko-von-Leipzig: Nit/style: should be able to elide some of this

```suggestion
    raw.as_ref().map(NoteScript::from_bytes).transpose().map_err(Into::into)
```

crates/store/src/server/rpc_api.rs
Mirko-von-Leipzig: I've created #1198 which should improve/remove the upkeep of this. It really is quite fragile.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -250,6 +250,29 @@ pub(crate) fn select_note_inclusion_proofs(
...
+    conn: &mut SqliteConnection,
+    root: Word,
+) -> Result<Option<NoteScript>, DatabaseError> {
+    // SELECT
+    //     root,
+    //     script
+    // FROM
+    //     note_scripts
+    // WHERE
+    //     root = ?1;
```
drahnr: Note: in #1193 these comments get moved to the doc comments

```
# Raw SQL

\`\`\`sql
SELECT ..
\`\`\`
```

SantiagoPittella: Awesome, I moved the query to doc comments already 👌🏼

PR #1197, opened 2025-09-02
chore: update `miden-base` to next

tomyrd:


Mirko-von-Leipzig:
I think let's wait until there are either more substantial changes, or a specific feature that we want.

PR #1205, opened 2025-09-03
feat: add `BlockRange` and `PaginationInfo` messages

### LINKED ISSUE #1204: Implement `BlockRange` message
As Mirko suggested in https://github.com/0xMiden/miden-node/pull/1191#discussion_r2318040279 , a new message could be added to the proto files, representing a block range to be used in queries in the DB. Additionally, methods that have pagination also have `chain_tip` and `block_number` in the response. We can abstract those fields in a message like `Page` or similar.

SantiagoPittella:
closes #1204


crates/proto/src/generated/rpc_store.rs
```diff
@@ -371,6 +341,30 @@ pub struct MaybeNoteScript {
...
+#[derive(Clone, Copy, PartialEq, ::prost::Message)]
+pub struct BlockRange {
+    /// Block number from which to start (inclusive).
+    #[prost(fixed32, tag = "1")]
+    pub block_from: u32,
+    /// Block number up to which to check (inclusive). If not specified, checks up to the latest block.
+    #[prost(fixed32, optional, tag = "2")]
+    pub block_to: ::core::option::Option<u32>,
+}
+/// Represents pagination information.
```
drahnr: nit: it is pagination across a single axis of blocks, we might have different pagination axis/multiple so we should be specific in the comment

crates/proto/src/generated/rpc_store.rs
```diff
@@ -316,36 +296,26 @@ pub struct SyncNotesResponse {
...
-    /// Block number up to which to sync. If not specified, syncs up to the latest block.
+    /// Block range from which to start synchronizing.
     ///
-    /// If specified, this block must be close to the chain tip (i.e., within 30 blocks),
+    /// If the `block_to` is specified, this block must be close to the chain tip (i.e., within 30 blocks),
     /// otherwise an error will be returned.
-    #[prost(fixed32, optional, tag = "2")]
-    pub block_to: ::core::option::Option<u32>,
+    #[prost(message, optional, tag = "1")]
+    pub block_range: ::core::option::Option<BlockRange>,
```
drahnr, line 304: Note: I am still not sure why we generate so many options when the proto definition doesn't use the `optional` field. EDIT: After reading some more, v2 to v3 is the root, lacking a `required` keyword, making everything optional by default :face_holding_back_tears:

drahnr (approved):

crates/proto/src/generated/rpc_store.rs
Mirko-von-Leipzig: Protobuf went with _everything_ is optional that isn't a literal primitive to encourage "backwards" compatibility. But imo that only ensures it on the wire and its a pain when you _do_ require something absolutely.. which is the general case.

drahnr: Isn't the requirements and ordering such a key criteria? I always preferred https://capnproto.org/ over protobuf (though optionals come only in the form of unions)

Mirko-von-Leipzig: https://protobuf.dev/best-practices/dos-donts/

Its really a pita.

crates/proto/src/generated/rpc_store.rs
SantiagoPittella: Currently it is the only axis, and I'm not aware of a plan to add another. I'm updating docs to better reflect this.

crates/proto/src/generated/rpc_store.rs
SantiagoPittella: In our domain it is mandatory, but all non-scalar types are marked as optional when building. This is why we used `missing_field` in a bunch of places. The reason behind is field presence and is explained here:
- https://github.com/tokio-rs/prost/issues/521
- https://protobuf.dev/programming-guides/field_presence/
- https://github.com/protocolbuffers/protobuf/blob/main/docs/implementing_proto3_presence.md

SantiagoPittella: Sorry, i didn't reload before sending my reply and didn't saw Mirko's replies

proto/proto/store/rpc.proto
```diff
@@ -430,3 +400,37 @@ message MaybeNoteScript {
...
+// ================================================================================================
+
+// Represents pagination information for chunked responses.
+//
+// Pagination is done using block numbers as the axis, allowing clients to request
+// data in chunks by specifying block ranges and continuing from where the previous
+// response left off.
+// To request the next chunk, the client should use the `block_num` from the previous response
+// as the `block_from` for the next request.
+message PaginationInfo {
```
bobbinth: This is a bit ambiguous because it doesn't use the `+1` - so, it is not clear whether we should use just `block_num` from the previous response or `bock_num + 1`. I would rewrite this to clarify.

Also, nit: let's add a newline after line 423.

crates/store/src/server/rpc_api.rs
```diff
@@ -361,8 +365,10 @@ impl rpc_server::Rpc for StoreApi {
             )));
         }
 
-        let block_from = request.block_from.into();
-        let block_to: BlockNumber = request.block_to.map_or(chain_tip, BlockNumber::from);
+        let block_range = request.block_range.ok_or(invalid_argument("block_range is required"))?;
+
+        let block_from = block_range.block_from.into();
+        let block_to: BlockNumber = block_range.block_to.map_or(chain_tip, BlockNumber::from);
```
Mirko-von-Leipzig: Let's add a method for this to the proto crate? Since the type is generated there, it should be possible to impl it in the domain files somewhere.

```rust
impl proto::BlockRange {
    pub fn into_inclusive_range(self, fallback: BlockNumber) -> InclusiveRange<BlockNumber>
}
```

bin/stress-test/src/store/mod.rs
```diff
@@ -279,10 +279,9 @@ async fn sync_nullifiers(
     nullifiers_prefixes: Vec<u32>,
 ) -> (Duration, proto::rpc_store::SyncNullifiersResponse) {
     let sync_request = proto::rpc_store::SyncNullifiersRequest {
+        block_range: Some(proto::rpc_store::BlockRange { block_from: 0, block_to: None }),
```
Mirko-von-Leipzig, line 282: Similarly to the other `impl` suggestion, we could add

```rust
impl proto::BlockRange {
    pub fn from_range(range: impl std::ops::RangeBound<BlockNumber>) -> Self
}
```

This would be nice to have on hand to simplify any test writing we may do. Though maybe its not necessary since there is only a single instance..

PR #1206, opened 2025-09-03
chore: improve gRPC spans

### LINKED ISSUE #1198: Improve gRPC spans for OTel integration
Replace the various gRPC `xxx_trace_fn` with a single function which utilizes the http path.

#### Context

A gRPC root span should be named `<service>/<method>`. This is convenient because this is also the http path format for gRPC requests.

We currently match on this path to select the appropriate gRPC method for each service. Unfortunately `tracing::span!(<name>)` only accepts string literals - which is why we have these matches e.g. https://github.com/0xMiden/miden-node/blob/480f01b452564711b63312d265736dada8fb3bfc/crates/utils/src/tracing/grpc.rs#L99-L112 

This is fragile as we constantly forget to update these whenever a method is added or renamed and our monitoring is actually littered with several `Unknown` requests. We could add tests for this but its not as straightforward as one might hope.

What's taken me far too long to realize is that we don't actually care about the tracing span name -- only the span name that OTel recognizes. The OTel tracing [integration crate uses the official](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/index.html#special-fields) `otel.name` for this. fit defaults to the tracing span name but we can override this instead to create a dynamic span name.

SantiagoPittella:
closes #1198

### DIFF
--- a/crates/utils/src/tracing/grpc.rs
+++ b/crates/utils/src/tracing/grpc.rs
@@ -1,115 +1,26 @@
-/// Represents an Otel compatible, traced Miden component.
-///
-/// Used to select an appropriate service name for the tonic server to use.
-#[derive(Copy, Clone)]
-pub enum TracedComponent {
-    Rpc,
-    BlockProducer,
-    StoreRpc,
-    StoreBlockProducer,
-    StoreNtxBuilder,
-    RemoteProver,
-    RemoteProverProxy,
-}
+use crate::tracing::OpenTelemetrySpanExt;
 
-impl TracedComponent {
-    /// Returns the service name for this component.
-    fn service_name(self) -> &'static str {
-        match self {
-            TracedComponent::Rpc => "rpc.rpc",
-            TracedComponent::BlockProducer => "block-producer.rpc",
-            TracedComponent::StoreRpc => "store.rpc.rpc",
-            TracedComponent::StoreBlockProducer => "store.block-producer.rpc",
-            TracedComponent::StoreNtxBuilder => "store.ntx-builder.rpc",
-            TracedComponent::RemoteProver => "remote-prover.rpc",
-            TracedComponent::RemoteProverProxy => "remote-prover-proxy.rpc",
-        }
-    }
-}
-
-/// Returns a [`trace_fn`](tonic::transport::server::Server) implementation for the RPC which
-/// adds open-telemetry information to the span.
+/// Returns a [`trace_fn`](tonic::transport::server::Server) implementation for gRPC requests
+/// which adds open-telemetry information to the span.
 ///
-/// Creates an `info` span following the open-telemetry standard: `{service}.rpc/{method}`.
+/// Creates an `info` span following the open-telemetry standard: `{service}/{method}`.
 /// The span name is dynamically set using the HTTP path via the `otel.name` field.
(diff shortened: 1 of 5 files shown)


crates/utils/src/tracing/grpc.rs
```diff
@@ -41,120 +46,59 @@ pub fn traced_span_fn<T>(component: TracedComponent) -> fn(&http::Request<T>) ->
...
+/// This function extracts the method name from the HTTP path and creates a span with
+/// the service name from the component. The actual span name used by OpenTelemetry
+/// is set dynamically via the `otel.name` field, allowing us to avoid hardcoded
+/// method name matches.
+fn dynamic_trace_fn<T>(request: &http::Request<T>, component: TracedComponent) -> tracing::Span {
+    let service_name = component.service_name();
+
+    // Extract the method name from the HTTP path (e.g., "/rpc.rpc/CheckNullifiers" ->
+    // "CheckNullifiers")
+    let method_name = request.uri().path().rsplit('/').next().unwrap_or("Unknown");
```
sergerad: Now that we are not doing an exhaustive match on the method, maybe we can just unwrap_or_default() and keep it empty?

crates/utils/src/tracing/grpc.rs
```diff
@@ -41,120 +46,59 @@ pub fn traced_span_fn<T>(component: TracedComponent) -> fn(&http::Request<T>) ->
...
-    let method = request.uri().path().rsplit('/').next().unwrap_or("Unknown");
-    let span = if method == "Status" {
-        rpc_span!("remote-prover-proxy.rpc", "Status")
-    } else {
-        rpc_span!("remote-prover-proxy.rpc", "Unknown")
-    };
-
-    let span = add_otel_span_attributes(span, request);
-    add_network_attributes(span, request)
+    dynamic_trace_fn(request, TracedComponent::RemoteProverProxy)
```
sergerad: Should we just call the dynamic fn in the services and get them to pass in the enum?

sergerad (approved):

crates/utils/src/tracing/grpc.rs
Mirko-von-Leipzig, line 1: We can simplify this further because the path format is actually `../<service>/<method>` so we can read both method _and_ service from the request itself.

I think we can have a single `grpc_trace_fn` (no generics), with everything in there using:

```rust
    // A gRPC request's path ends with `../<service>/<method>`.
    let mut path_segments = request.uri().path().rsplit('/');
    let method = path_segments.next().unwrap_or_default();
    let service = path_segments.next().unwrap_or_default();
```

I would also inline the otel span attributes and network attributes. iirc we don't use them elsewhere - but maybe I'm wrong.

crates/utils/src/tracing/grpc.rs
Mirko-von-Leipzig: We can simplify even further because the path also contains the service name, https://github.com/0xMiden/miden-node/pull/1206/files#r2320988255

crates/utils/src/tracing/grpc.rs
SantiagoPittella: Ooooh, ok. I just sent a commit changing to this

Issue #1106, opened 2025-07-22
Refactor Proto Messages for App-level Errors


sergerad:
Derived from [comment](https://github.com/0xMiden/miden-node/pull/1094#discussion_r2223846772).

We should consider handling app-level errors through a standardized means involving a generalized error message. For example:
```proto
// Generalized error
message Error {
    uint32 code = 1;
    string message = 2;
}

// Message for specific endpoint
message TransactionInputsResponse {
    message TransactionInputs {
        message AccountTransactionInputRecord { ... }
        message NullifierTransactionInputRecord { ...  }
    
        AccountTransactionInputRecord account_state = 1;
        repeated NullifierTransactionInputRecord nullifiers = 2;
        repeated primitives.Digest found_unauthenticated_notes = 3;
        fixed32 block_height = 4;
    }

    // Result may be an error
    oneof payload {
        TransactionInputs inputs = 1;
        Error error = 2;
    }
}
```

bobbinth:
This should probably apply to all RPC endpoints and also to the `GetTransactionInputs` endpoint (maybe others as well).

sergerad:
> Do we want full structured errors? As in, no message but instead structs with fields.

This is probably worth trying first. Perhaps for the specific use case you described above in faucet. I can start working on a PR. I'm hoping that we can stagger these changes rather than having to do a big PR for a bulk of endpoints at once.

sergerad:
Just as FYI for the edification of anyone interested.

An endpoint for a gRPC service is structured like so:
```rust
    #[instrument(parent = None, target = COMPONENT, name = "rpc.server.submit_proven_transaction", skip_all, err)]
    async fn submit_proven_transaction(
        &self,
        request: Request<proto::transaction::ProvenTransaction>,
    ) -> Result<Response<proto::block_producer::SubmitProvenTransactionResponse>, Status> {
```

Where `Status` is:
```rust
pub struct Status {
    /// The gRPC status code, found in the `grpc-status` header.
    code: Code,
    /// A relevant error message, found in the `grpc-message` header.
    message: String,
    /// Binary opaque details, found in the `grpc-status-details-bin` header.
    details: Bytes,
    /// Custom metadata, found in the user-defined headers.
    /// If the metadata contains any headers with names reserved either by the gRPC spec
    /// or by `Status` fields above, they will be ignored.
    metadata: MetadataMap,
    /// Optional underlying error.
[...]

PR #1208, opened 2025-09-04
wip: refactor SubmitProvenTransaction proto error

SantiagoPittella:
@Mirko-von-Leipzig @igamigo @sergerad  I'm opening this as a draft just to double check the approach before continuing, what do you guys think? I'm based on [this comment](https://github.com/0xMiden/miden-node/issues/1106#issuecomment-3162673814)


proto/proto/block_producer.proto
```diff
@@ -58,13 +58,45 @@ message BlockProducerStatus {
...
+
 // Represents the result of submitting proven transaction.
 message SubmitProvenTransactionResponse {
-    // The node's current block height.
-    fixed32 block_height = 1;
+    // Success payload containing the transaction submission result
+    message Success {
+        // The node's current block height.
+        fixed32 block_height = 1;
+    }
```
bobbinth: I thought the protobuf messages wouldn't change and we'd pass error codes though gRPC headers. Or would that not work?

SantiagoPittella: Ohh ok, awesome. I confused the approach. I now reverted the old approach and sent a commit with the errors on the message and details fields of `Status`. I think that we can reduce the code a bit further, thought if you already have a comment on this approach it would be much appreciated.
cc @Mirko-von-Leipzig

crates/block-producer/src/errors.rs
```diff
@@ -100,25 +101,112 @@ pub enum AddTransactionError {
...
+            _ => tonic::Code::InvalidArgument,
+        };
+
+        // Serialize the error details to bytes
+        let Ok(details_bytes) = bincode::serialize(&error_details) else {
+            // Fallback to simple error message if serialization fails
+            return tonic::Status::new(grpc_code, value.as_report());
+        };
+
+        tonic::Status::with_details(grpc_code, value.as_report(), details_bytes.into())
```
bobbinth: I was thinking it would work a bit differently:

- For message, we'd always use string-based messages.
- For details we'd use a numeric code assigned to each error variant.

So, basically, it would look something like:
```rust
tonic::Status::with_details(grpc_code, err_message, err_code)
```

For `err_message` we'd use something like `error.to_string()` (which means that we need to provide a string description for each error). We could also use `Report` here, but we still need to make sure error messages have human-readable descriptions.

And `err_code` could be just a `u8` converted into a byte (I think `u8` is enough because it is unlikely that we'll have more than 256 errors per endpoint). Here, the `InputNotesAlreadyConsumed` could have error code `0`, `UnauthenticatedNotesNotFound` could have error code `1` etc.

The idea is that the client would be able to look at details to programmatically determine the type of the error, and we'd complement it with an error message to show to the user (but the intent is that the client won't need to parse error messages).

crates/block-producer/src/errors.rs
```diff
@@ -100,25 +100,68 @@ pub enum AddTransactionError {
...
 
-            // Internal errors which should not be communicated to the user.
-            AddTransactionError::VerificationFailed(VerifyTxError::StoreConnectionFailed(_))
-            | AddTransactionError::StaleInputs { .. } => Self::internal("Internal error"),
-        }
+        tonic::Status::with_details(grpc_code, value.to_string(), vec![error_code].into())
+    }
+}
+
+fn get_error_code_and_grpc_code(error: &AddTransactionError) -> (u8, tonic::Code) {
```
bobbinth: nit: it is not clear to me if we gain much by splitting this out into a separate function (i.e., it could probably be just a part of the `fn from(value: AddTransactionError) -> Self` function).

crates/block-producer/src/errors.rs
```diff
@@ -100,25 +100,68 @@ pub enum AddTransactionError {
...
+                SubmitTransactionErrorCode::InvalidTransactionProof as u8,
+                tonic::Code::InvalidArgument,
+            ),
+            VerifyTxError::StoreConnectionFailed(_) => {
+                (SubmitTransactionErrorCode::Internal as u8, tonic::Code::Internal)
+            },
+        },
+        AddTransactionError::StaleInputs { .. } => {
+            (SubmitTransactionErrorCode::StaleInputs as u8, tonic::Code::InvalidArgument)
+        },
```
bobbinth: I think this is also an internal error.

crates/block-producer/src/errors.rs
```diff
@@ -100,25 +100,68 @@ pub enum AddTransactionError {
...
+    InputNotesAlreadyConsumed = 0,
+    UnauthenticatedNotesNotFound = 1,
+    OutputNotesAlreadyExist = 2,
+    IncorrectAccountInitialCommitment = 3,
+    InvalidTransactionProof = 4,
+    StaleInputs = 5,
+    Expired = 6,
+    DeserializationFailed = 7,
+    Internal = 8,
+}
```
bobbinth: As per previous comment, I'd get rid of `StaleInputs` variant, and also, arrange variants as follows:
```rust
pub enum SubmitTransactionErrorCode {
    InternalError = 0,
    DeserializationFailed = 1,
    InvalidTransactionProof = 2,
    IncorrectAccountInitialCommitment = 3,
    InputNotesAlreadyConsumed = 4,
    UnauthenticatedNotesNotFound = 5,
    OutputNotesAlreadyExist = 6,
    TransactionExpired = 7,
}
```

bobbinth (commented):

I think the bigger question is whether we want to include more info than just the error code with the message. If we do, then I think the errors should be encoded as JSON, but then question would be where do we put it - i.e., do we still put it into `details` or `maybe` the JSON can go into `message`?

Also, what do the text messages currently look like? A couple of examples would be great to see what kind of info we convey in the "human-readable" form.

Mirko-von-Leipzig:
> I think the bigger question is whether we want to include more info than just the error code with the message. If we do, then I think the errors should be encoded as JSON, but then question would be where do we put it - i.e., do we still put it into `details` or `maybe` the JSON can go into `message`?

I think broadly speaking there are two use cases:

1. Display the error to some end user without having to worry about the content or context. For this `message` seems ideal and I think we should keep it.
2. Act on the error, and for this we want the json which can then go in `details` imo.

(2) is likely the exception so having a convenient `message` would probably be useful for most use cases.

Cargo.toml
```diff
@@ -58,6 +58,7 @@ miden-air = { version = "0.17" }
 anyhow             = { version = "1.0" }
 assert_matches     = { version = "1.5" }
 async-trait        = { version = "0.1" }
+bincode            = { version = "1.3" }
```
Mirko-von-Leipzig: What's this needed for?

crates/block-producer/src/errors.rs
Mirko-von-Leipzig, line 1: Some food for thought: if we want the errors to be part of our schema, then their description should also live in `/proto` with the rest of the schemas. I propose we say `message` should _not_ be treated as stable aka is not part of the semver versioning of the schema aka for presentation only. And then we define the `details` schema for the errors.

## Schema options

If we go with numeric then its just a simple, append-only list.

For structured errors we have the choices (imo) of `json` or `gRPC`. For the former we would add a [json schema](https://json-schema.org/), for the latter a separate gRPC schema. Both make sense to me. The former is human-readible on the wire, the latter.. well they already have to be using `gRPC` to receive it so why change it? The downside of the latter is that users have to manually invoke the decoding; json is easier as its somewhat self-describing.

## Architecture

Regardless of the decisions we make above, I think its probably sensible that we enforce this a bit via types. More specifically, `crates/proto` should parse the error schema and provide Rust types from these. This would also include a conversion of the error type to `tonic::Status`.

The RPC methods in question should then adapt these new errors types, and perform the mappings as needed.

crates/block-producer/src/errors.rs
sergerad, line 137: I think this approach should be sufficient and that we wouldn't need json/proto encoding.

Would the rust client have access to this enum somehow? The web client would rely on some specification derived from this code?

Cargo.toml
SantiagoPittella: This is not needed at all, i used it to test a different approach and forgot to remove it.

crates/block-producer/src/errors.rs
bobbinth: > if we want the errors to be part of our schema, then their description should also live in `/proto` with the rest of the schemas. I propose we say `message` should _not_ be treated as stable aka is not part of the semver versioning of the schema aka for presentation only. And then we define the `details` schema for the errors.

Yes, agreed with this.

> If we go with numeric then its just a simple, append-only list.

Trying to think if numeric only approach is sufficient. Looking at the enum from https://github.com/0xMiden/miden-node/pull/1208#discussion_r2335539132, there are a couple of variants where additional data may be helpful to the client:

- `InputNotesAlreadyConsumed`: it may be good to get the list of notes that have been consumed. Though, the client could probably just query the node for all input notes in the transaction to check that.
- `UnauthenticatedNotesNotFound`: would also be good to get a list of note IDs for this. But unlike the previous variant, there isn't really an easy way for the client to figure out which un-authenticated note was missing.
- `OutputNotesAlreadyExist`: this is similar to the first variant - would be good to get, but we should be able to query the node for this data separately.

So, really, the only problematic variant is `UnauthenticatedNotesNotFound` - but even hear, I'm not sure what are the exact meaningful actions the client would take based on the extra info provided. cc @igamigo in case he has some ideas.

----
So, it seems to me that maybe going with just numeric values is OK for now. And we could still consider adding rich error info in the future (i.e., before mainnet) especially, if we identify good use cases for it.

crates/block-producer/src/errors.rs
Mirko-von-Leipzig: As mentioned https://github.com/0xMiden/miden-node/pull/1208#discussion_r2336523909 I think the list should exist (and be exported by) in the `/proto/` schema crate. Within the node we would then re-export this again via `/crates/proto` for our internal usage (and as part of the RPC exports we do there).

I believe the rust client pulls in the latter already.

I'm unaware of any schema language that just maps error codes but a basic json dictionary probably makes sense? SDKs can then parse this into whatever structure they want :shrug:

crates/block-producer/src/errors.rs
Mirko-von-Leipzig: > Trying to think if numeric only approach is sufficient. Looking at the enum from https://github.com/0xMiden/miden-node/pull/1208#discussion_r2335539132, there are a couple of variants where additional data may be helpful to the client:
>
> [..snip..]

This isn't entirely possible - because the clashing state can be inflight in the mempool and are not available via the store. We can add mempool introspection endpoints but that will get messy quickly as currently this would require locking the mempool so we would need to create duplicate read-only mempool state for RPC (possible via a mempool subscription for example).

This still leaves a bit of a race condition though because the mempool state would change between the result and the follow-up query. Either because the state has since been committed and pruned, or due to reversions.

I think in general, if you want the error to be actionable you don't want to mess around with this - just give them the data they need upfront.

igamigo: Yeah, I agree that asking for data post-query is not very reliable. I also think an enum is a good starting point. For a schema I'd try to keep gRPC because introducing JSON, while convenient in many languages (and for human readability), seems overkill just for these. As Mirko said, you'd already have to be using a gRPC stack anyway.
I'm not sure that the client itself can solve any of the mentioned potential errors, but at the user level it sounds like it could be useful to have richer data.

SantiagoPittella: I;ve moved the error enum to a proto enum

bobbinth: It seems like the commit with these changes hasn't been pushed yet?

SantiagoPittella: That is indeed what happened, sorry. Just pushed the changes.

crates/block-producer/src/errors.rs
```diff
@@ -100,25 +101,51 @@ pub enum AddTransactionError {
...
+                SubmitProvenTransactionError::TransactionExpired as u8,
+                tonic::Code::InvalidArgument,
+            ),
+            AddTransactionError::TransactionDeserializationFailed(_) => (
+                SubmitProvenTransactionError::DeserializationFailed as u8,
+                tonic::Code::InvalidArgument,
+            ),
+        };
+
+        tonic::Status::with_details(grpc_code, value.to_string(), vec![error_code].into())
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: @SantiagoPittella the bot is correct 😅

We also can't use `value.to_string()` because this will lose the error stack.

crates/block-producer/src/errors.rs
Mirko-von-Leipzig, line 1: I think we can simplify this by adding some helper functions to `crates/proto`:

```rust
impl SubmitProvenTransactionError {
    // This replaces the `as u8` - casts in random places always stand out to me
    // as potential problems so it would be best to isolate it imo.
    fn api_code(&self) -> u8 { self as u8 }

    fn tonic_code(&self) -> tonic::Code { ... }

    fn is_internal(&self) -> bool { ... }
}
```

and then in this file:

```rust
impl AddTransactionError {
    fn api_error(&self) -> SubmitProvenTransactionError { .. }
}

impl From<AddTransactionError> for tonic::Status {
    fn from(value: AddTransactionError) -> Self {
        let api_error = value.api_error();

        let message = if api_error.is_internal() {
            "Internal error".to_owned()
        } else {
            value.as_report()
        };
        
        tonic::Status::with_details(api_error.tonic_code(), message, vec![api_error.api_code()].into())
    }
}
```

crates/block-producer/src/errors_test.rs
```diff
@@ -0,0 +1,97 @@
...
+        let error = AddTransactionError::StaleInputs { input_block, stale_limit };
+
+        let status: tonic::Status = error.into();
+
+        assert_eq!(status.code(), tonic::Code::Internal);
+        assert!(status.message().contains("stale"));
+
+        let error_code = parse_error_code(&status).expect("Expected error code in details");
+        assert_eq!(error_code, SubmitProvenTransactionError::InternalError as u8);
+    }
```
Mirko-von-Leipzig: Consider using `rstest` to combine all of these, though maybe its not a good fit.

```suggestion
    #[rstest::rstest]
    #[case::stale_inputs(
        AddTransactionError::StaleInputs {
            input_block: BlockNumber::from(100),
            stale_limit: BlockNumber::from(150),
        },
        SubmitProvenTransactionError::Internal,
        "Internal error",
    )]
    #[case::expired(
        AddTransactionError::Expired {
            expired_at: BlockNumber::from(100),
            limit: BlockNumber::from(150),
        },
        SubmitProvenTransactionError::TransactionExpired,
        "expired",
    )]
    fn submit_transaction_error_mapping(
        #[case] uut: AddTransactionError,
        #[case] expected: SubmitProvenTransactionError,
        #[case] message: &str,
    ) {
        let status = tonic::Status::from(uut);

        assert_eq!(status.code(), expected.tonic_code());
        assert!(status.message().contains(message));
        
        let api_code = match status.details() {
            &[single] => single,
            other => panic!("Details should contain exactly one byte for the api error code, but got {other:?}),
        };
        assert_eq!(api_code, expected.api_code());
    }
```

SantiagoPittella: It does look better, thx!

crates/block-producer/src/errors.rs
```diff
@@ -100,28 +101,59 @@ pub enum AddTransactionError {
...
+        let message = if api_error.is_internal() {
+            "Internal error".to_owned()
+        } else {
+            value.as_report()
+        };
+
+        tonic::Status::with_details(
+            api_error.tonic_code(),
+            message,
+            vec![api_error.api_code()].into(),
```
bobbinth, line 155: Is there an there an implicit assumption here that we know how protobuf enums are serialized? Or do we not care about this and just using protobuf messages as means of specifying numeric error values, but otherwise don't care about their serialization?

bobbinth (approved):

Also, it would be good to update the RPC API docs to explain how errors are reported and maybe even add a table with error explanations fro the `SubmitProvenTransaction` endpoint.

crates/block-producer/src/errors.rs
Mirko-von-Leipzig: Ah, that's a fair point - we're only using it as a specification here.

And that's probably not great; ideally we should be using protobuf to encode the type to bytes.. But I don't see anything generated that we could use for that.

Looking the [proto3 enum docs](https://protobuf.dev/programming-guides/proto3/#enum) it doesn't really help wrt encoding, but @SantiagoPittella  we should also be adhering to:

> In proto3, the first value defined in an enum definition must have the value zero and should have the name ENUM_TYPE_NAME_UNSPECIFIED or ENUM_TYPE_NAME_UNKNOWN.

Mirko-von-Leipzig: @SantiagoPittella could you try figure out how we could instead write something like:

```rust
tonic::Status::with_details(
    api_error.tonic_code(),
    message,
    tonic::to_bytes(api_error)
)
```

I assume the `::prost::Enumeration` derived on the type enables some sort of serialization.

SantiagoPittella: I pushed a commit with the docs update and another with the enum stanrd 0 field. I'm still looking for a "native" way to serialize the errors but couldn't find anything useful yet.

SantiagoPittella: I found something but it is more like a workaround than a solution, i don't really like it. It is basically wrapping the enum in a message where it is the only field, which does implement serialization

SantiagoPittella: Not really sure how to tackle the serialization yet. Should we merge this PR with a comment on the serialization?

Mirko-von-Leipzig: I think that's a good idea :+1:

Issue #1185, opened 2025-08-31
Combine `GetAccountDetails` and `GetAccountProofs` endpoints


bobbinth:
This issue supersedes both https://github.com/0xMiden/miden-node/issues/527 and https://github.com/0xMiden/miden-node/issues/616, and probably splits out into a few sub-issues itself.

The goals are:

1. Let clients sync to an account state close to the chain tip - but not necessarily to the chain tip itself. This is needed for FPI calls where while a transaction is in progress, the chain tip may move, and so the block against which we are executing a transaction may be slightly off from the chain tip.
2. Do not require clients to download full account state (specifically all contents of storage maps and the entire vault) to be able to work with accounts. This basically means that clients should be able to download partial vault and partial maps. Another consequence of this is that all responses should have bounded size (i.e., not bigger than 2MB - 4MB).
3. Make the requests and the sync process as simple and "light-weight" as possible for the general case (i.e., small accounts), while still supporting edge cases of large accounts.

The way I think to accomplish these goals is to split the process into steps:

1. The user makes a request to a single endpoint to get basic info about an account (or a set of accounts). For small accounts, this could return all account data.
2. For larger accounts, the user would follow this up with calls to `SyncStorageMaps` and `SyncAccountVault` endpoints to download the data that was not returned from the first request.

We already have `SyncStorageMaps` and `SyncAccountVault` endpoints, and so this issue is about the first point above.

My current think is that this endpoint should work with just a single account. The main reason is that if we make it about more than one account, keeping the response size small will be very challenging. For example, if the user requests data for 10 accounts, even 200KB per account could overflow the limit. And so, we'd probably need to keep the number of requested accounts pretty small. But at that point, might as well just have it work for a single account only.

The endpoint would be more similar to the current `AccountProofs`, and so I'll call it `AccountProof` for now, but we should probably come up with a better name. The endpoint should return:

- For private and public accounts:
  - Account state commitment.
  - Account witness. We can consider making this optional as this is currently only useful for FPI calls, but I think this may be generally useful in the future.
- For public accounts only:
  - Account code, if needed. If included, this would not exceed 64KB as this is what we think to impose as a limit for account code (though, there is an argument to make this even bigger - maybe 128KB or bigger).
  - Account storage header. This would include values of all value slots and roots of all map slots (and some metadata). Since we can have at most 256 storage slots, this part cannot exceed 16KB.
[...]

PR #1211, opened 2025-09-08
Change `GetAccountProofs` -> `GetAccountProof`

drahnr:
Part one of https://github.com/0xMiden/miden-node/issues/1185


proto/proto/store/rpc.proto
```diff
@@ -122,7 +122,7 @@ message AccountProofsRequest {
     }
 
     // A list of account requests, including map keys + values.
-    repeated AccountRequest account_requests = 1;
+    AccountRequest account_request = 1;
```
igamigo: With this, the `code_commitments` field could change to only be be a single code commitment. The idea is that the user can send the last known code commitments of the accounts they were requesting, so as to minimize bandwidth by having the node not return the serialized `AccountCode` if the user was already storing it.

Since there would be one account per request now, there'd be no need to make it a list.

igamigo (commented):
Looks good overall, only one small comment for now (which you might have already been keeping in mind)

proto/proto/store/rpc.proto
bobbinth: This doesn't need to be a list, but I think this should be `optional`. The main reason is that detailed requests are only relevant for public accounts (i.e., we can't request these for private accounts).

crates/store/src/db/sql/utils.rs
bobbinth, line 1: Was this an obsolete file?

drahnr: Isolated change-set https://github.com/0xMiden/miden-node/pull/1220

proto/proto/store/rpc.proto
drahnr: If we need to make it optional, that requires additional refactor to extract the `AccountId` as a separate field. I don't think this worked for private accounts previously.

crates/store/src/server/rpc_api.rs
```diff
@@ -300,42 +298,41 @@ impl rpc_server::Rpc for StoreApi {
...
-        let (block_num, infos) = self
+        let request_code_commitment = Word::try_from(
+            &(code_commitment.ok_or(Status::invalid_argument(stringify!(code_commitment)))?),
+        )
+        .map_err(|err| Status::invalid_argument(format!("Invalid code commitment: {err}")))?;
+
+        let account_request =
+            account_request.ok_or(Status::invalid_argument(stringify!(account_request)))?;
+        let account_request = account_request.try_into().map_err(|err| {
+            Status::invalid_argument(format!("Invalid account proofs request: {err}"))
```
SantiagoPittella:
```suggestion
            Status::invalid_argument(format!("Invalid account proof request: {err}"))
```

crates/store/src/state.rs
```diff
@@ -883,101 +883,76 @@ impl State {
     }
 
     /// Returns account proofs with optional account and storage headers.
```
SantiagoPittella:
```suggestion
    /// Returns account proof with optional account and storage headers.
```

proto/proto/store/rpc.proto
```diff
@@ -101,7 +101,7 @@ message StoreStatus {
 // ================================================================================================
 
 // Returns the latest state proofs of the specified accounts.
```
SantiagoPittella:
```suggestion
// Returns the latest state proof of the specified accounts.
```

SantiagoPittella: Doing this change and re-generating should fix the comment in multiple places.

proto/proto/store/rpc.proto
```diff
@@ -122,60 +122,53 @@ message AccountProofsRequest {
...
+
+        // Account code, returned only when none of the request's code commitments match
+        // the current one.
+        optional bytes account_code = 3;
 
-        // State header for public accounts. Filled only if `include_headers` flag is set to `true`.
-        optional AccountStateHeader state_header = 2;
+        // Storage slots information for this account
+        repeated StorageSlotMapProof storage_maps = 4;
     }
```
SantiagoPittella: I might be wrong, feel free to disregard. I recall having this "internal" messages definition on top of the source message, before the message fields

drahnr: It's a good point, but this has become unwieldly at the given nesting depth

proto/proto/store/rpc.proto
bobbinth: Right, we'll need to re-shuffle some fields. The messages would look like:
```proto
message AccountProofRequest {
    message AccountRequest {
        message StorageRequest {
            uint32 storage_slot_index = 1;
            repeated primitives.Digest map_keys = 2;
        }

        optional primitives.Digest code_commitment = 1;
        repeated StorageRequest storage_requests = 2;
        optional bool include_headers = 3;
    }

    optional AccountRequest account_request = 2;
}
```
Though, it is not clear to me if `include_headers` field would actually be needed - and so, it may make sense ot get rid of it.

crates/store/src/server/rpc_api.rs
```diff
@@ -300,42 +298,41 @@ impl rpc_server::Rpc for StoreApi {
...
+        let request_code_commitment = Word::try_from(
+            &(code_commitment.ok_or(Status::invalid_argument(stringify!(code_commitment)))?),
+        )
+        .map_err(|err| Status::invalid_argument(format!("Invalid code commitment: {err}")))?;
+
+        let account_request =
+            account_request.ok_or(Status::invalid_argument(stringify!(account_request)))?;
+        let account_request = account_request.try_into().map_err(|err| {
+            Status::invalid_argument(format!("Invalid account proof request: {err}"))
+        })?;
```
bobbinth: If we update the request message as I mentioned in the other comment, I think the logic here and downstream would get simpler. Basically, we'd just need to pass account Id and an optional "account request" (maybe makes sense to rename it to "account details request").

proto/proto/store/rpc.proto
drahnr: I don't quite understand why we'd refactor this any further, while I see the changes are useful in themselves, the immediate next step is #1208 which refactors the particular query to the suggested API from #1185 ? Unless something has changed in your thinking.

bobbinth: I'm not sure #1208 is related to this (or did you mean #1214)? I was thinking of this as a simple change in the request that is self-contained enough to make further changes smaller. But we can do it in #1214 too. I'm just trying to think how to split things up into small but roughly equal chunks (without this, I think #1214 would be quite big).

drahnr: You're correct, it's #1214 . I moved some changes from #1214 into this one.

proto/proto/store/rpc.proto
```diff
@@ -97,14 +97,14 @@ message StoreStatus {
...
-// GET ACCOUNT PROOFS
+// GET ACCOUNT PROOF
 // ================================================================================================
 
-// Returns the latest state proofs of the specified accounts.
-message AccountProofsRequest {
+// Returns the latest state proof of the specified accounts.
+message AccountProofRequest {
     // Represents per-account requests where each account ID has its own list of
     // (storage_slot_index, map_keys) pairs.
```
bobbinth: This comment should probably be updated.

proto/proto/store/rpc.proto
```diff
@@ -114,68 +114,55 @@ message AccountProofsRequest {
...
-
-            // Account code, returned only when none of the request's code commitments match
-            // the current one.
-            optional bytes account_code = 3;
-
-            // Storage slots information for this account
-            repeated StorageSlotMapProof storage_maps = 4;
+message AccountProof {
+    // State header, available for public accounts only.
+    message AccountDetailsResponse {
```
bobbinth, line 134: Once we get rid of `GetAccountDetails` endpoint, we should rename `AccountDetailsResponse` into just `AccountDetails`.

proto/proto/store/rpc.proto
```diff
@@ -114,68 +114,55 @@ message AccountProofsRequest {
...
 
     // Block number at which the state of the accounts is returned.
     fixed32 block_num = 1;
 
-    // List of account state infos for the requested account keys.
-    repeated AccountProof account_proofs = 2;
+    // The account witness for the current state commitment of one account ID.
+    account.AccountWitness witness = 2;
+
+    // State header for public accounts. Filled only if the flag `AccountDetailsRequest` was present and the account was a public account/the information was available.
```
bobbinth: nit: comment wrapping.

crates/store/src/state.rs
```diff
@@ -883,101 +887,77 @@ impl State {
...
-    pub async fn get_account_proofs(
+    pub async fn get_account_proof(
```
bobbinth: The doc comment here should probably be updated a bit.

crates/store/src/state.rs
```diff
@@ -883,101 +887,77 @@ impl State {
...
 
-            headers_map
-        };
+                // Only include unknown account code blobs
+                let account_code =
+                    account_detail_request.code_commitment.and_then(|known_code_commitment| {
+                        (known_code_commitment == details.code().commitment())
+                            .not()
+                            .then(|| details.code().to_bytes())
+                    });
```
bobbinth: Would this also return the code if `account_detail_request.code_commitment` is `None`? Basically, we need to return the code if the request didn't have code commitment OR if the sent code commitment didn't match the current commitment.

crates/proto/src/domain/account.rs
```diff
@@ -149,6 +178,284 @@ impl TryInto<StorageMapKeysProof>
...
 
+// ACCOUNT DETAILS RESPONSE
+//================================================================================================
+
+pub struct AccountHeader2 {
+    pub vault_root: Word,
+    pub storage_commitment: Word,
+    pub code_commitment: Word,
+    pub nonce: Felt,
+}
```
bobbinth: We can't use `AccountHeader` from `miden-base` here because we don't have easy access to `AccountId`, right? If so, may it is worth modifying the protobuf message to include account ID? This can be done in the follow-up as well.

proto/proto/store/rpc.proto
drahnr: I think for clarity the `*Response` suffix does help, but it's also a bit weird if we'd use these message types in i.e. queries.

crates/proto/src/domain/account.rs
drahnr: Added

crates/store/src/state.rs
drahnr: Addressed

PR #1212, opened 2025-09-09
chore: Add healthcheck layer to rpc

sergerad:
## Context

We have LB and OPTIONS requests that are pure HTTP2 or HTTP1.1 requests without a path. Currently we have error logs when these occur - which is every few seconds.

## Changes

Adds a layer to the RPC service that will intercept requests with `/` path so that we can use it for LB health checks and OPTIONS requests, etc.

### DIFF
--- a/crates/rpc/src/server/health.rs
+++ b/crates/rpc/src/server/health.rs
@@ -4,7 +4,12 @@ use std::task::{Context as StdContext, Poll};
 use tonic::body::Body;
 use tower::{Layer, Service};
 
-// Simple health check layer that intercepts requests to root path.
+/// Simple health check layer that intercepts requests to root path.
+///
+/// The root path is used by load-balancers and options requests to check the health
+/// of the server. Since our gRPC server doesn't serve anything on the root
+/// these get logged as errors. This layer instead intercepts these requests
+/// and returns `Ok(200)`, preventing the errors.
 #[derive(Clone)]
 pub struct HealthCheckLayer;
 


crates/rpc/src/server/mod.rs
```diff
@@ -78,6 +80,7 @@ impl Rpc {
             .layer(TraceLayer::new_for_grpc().make_span_with(traced_span_fn(TracedComponent::Rpc)))
             .layer(AcceptHeaderLayer::new(&rpc_version, genesis.commitment()))
             .layer(cors_for_grpc_web_layer())
+            .layer(HealthCheckLayer)
```
Mirko-von-Leipzig, line 83: I think we should also swap the two layers above maybe? This is not related to this PR, but I'm worried that if the `Accept` layer rejects the request, then the CORS related headers won't be applied and a web client wouldn't accept the response at all, obscuring the error.

Is that how it works?

crates/rpc/src/server/health.rs
```diff
@@ -0,0 +1,49 @@
+use std::pin::Pin;
+use std::task::{Context as StdContext, Poll};
+
+use tonic::body::Body;
+use tower::{Layer, Service};
+
+// Simple health check layer that intercepts requests to root path.
```
Mirko-von-Leipzig:
```suggestion
/// Simple health check layer that intercepts requests to root path.
///
/// The root path is used by load-balancers and options requests to check the health
/// of the server. Since our gRPC server doesn't serve anything on the root
/// these get logged as errors. This layer instead intercepts these requests
/// and returns `Ok(200)`, preventing the errors.
```

crates/rpc/src/server/mod.rs
sergerad: Yea, done. LMK if thats what you meant.

crates/rpc/src/server/health.rs
```diff
@@ -0,0 +1,54 @@
+use std::pin::Pin;
+use std::task::{Context as StdContext, Poll};
+
+use tonic::body::Body;
+use tower::{Layer, Service};
+
+/// Simple health check layer that intercepts requests to root path.
+///
+/// The root path is used by load-balancers and options requests to check the health
```
Mirko-von-Leipzig, line 9: I would name this something else. Right now it sounds like its serving a `/health` response whereas its handling requests to the root path, which happens to be used by our infra layer for health checks. But that's out-of-scope here so its confusing.

Mirko-von-Leipzig (approved):

crates/rpc/src/server/mod.rs
Mirko-von-Leipzig: Perfect, yeah I think that's right.

crates/rpc/src/server/health.rs
sergerad: Maybe keepalive?

Mirko-von-Leipzig: `AllowRootRoute` maybe? tbh I'm unsure :D

bobbinth (approved):
Looks good! Thank you!

What kind of tests should we run to make sure these changes don't cause any issues with the client? (maybe none? but I also just want to be sure that requests don't get rejected for some reason).

PR #1214, opened 2025-09-09
feat: optimize `GetAccountProof` for small accounts

drahnr:
Part 2 of #1185 - reworks the `GetAccountProof` query with a new type.

The changeset focuses on providing account data / vault entries in the response if they are less than 100 (might be too conservative, for both). @SantiagoPittella

### DIFF
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -349,7 +349,7 @@ impl AccountVaultDetails {
     const MAX_RETURN_ENTRIES: usize = 1000;
 
     pub fn new(vault: &AssetVault) -> Self {
-        if vault.assets().skip(Self::MAX_RETURN_ENTRIES).next().is_some() {
+        if vault.assets().nth(Self::MAX_RETURN_ENTRIES).is_some() {
             Self::too_many()
         } else {
             Self {
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -163,7 +163,6 @@ impl TryFrom<proto::rpc_store::account_proof_request::AccountDetailRequest>
     }
 }
 
-/// Represents a request for an account's storage map values and its proof of existence.
 impl TryFrom<proto::account::AccountStorageHeader> for AccountStorageHeader {
     type Error = ConversionError;
 
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -348,7 +348,7 @@ impl AccountVaultDetails {
     const MAX_RETURN_ENTRIES: usize = 1000;
 
     pub fn new(vault: &AssetVault) -> Self {
-        if vault.asset_tree().num_leaves() > Self::MAX_RETURN_ENTRIES {
+        if vault.assets().skip(Self::MAX_RETURN_ENTRIES).next().is_some() {
             Self {
                 too_many_assets: true,
                 assets: Vec::new(),
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -7,6 +7,7 @@ use miden_objects::account::{
     AccountHeader,
     AccountId,
     AccountStorageHeader,
+    StorageMap,
     StorageSlotType,
 };
 use miden_objects::asset::Asset;
@@ -384,6 +385,44 @@ pub struct AccountStorageMapDetails {
     pub slot_index: u8,
     pub too_many_entries: bool,
     pub map_entries: Vec<(Word, Word)>,
+}
+
+impl AccountStorageMapDetails {
+    const MAX_RETURN_ENTRIES: usize = 1000;
+
+    pub fn new(slot_index: u8, slot_data: SlotData, storage_map: &StorageMap) -> Self {
+        let exceeds = if let SlotData::MapKeys(keys) = &slot_data {
+            keys.len() > Self::MAX_RETURN_ENTRIES
+        } else {
+            storage_map.entries().nth(Self::MAX_RETURN_ENTRIES).is_some()
+        };
+        if exceeds {
+            Self::too_many_entries(slot_index)
+        } else {
+            Self::from_slot_data(slot_index, slot_data, storage_map)
+        }
+    }
+
+    pub fn from_slot_data(slot_index: u8, slot_data: SlotData, storage_map: &StorageMap) -> Self {
+        let map_entries = match slot_data {
+            SlotData::All => Vec::from_iter(storage_map.entries().map(|(k, v)| (*k, *v))),
+            SlotData::MapKeys(keys) => {
+                Vec::from_iter(keys.iter().map(|key| (*key, storage_map.get(key))))
+            },
+        };
+        Self {
+            slot_index,
+            too_many_entries: false,
+            map_entries,
+        }
+    }
+    pub fn too_many_entries(slot_index: u8) -> Self {
+        Self {
+            slot_index,
+            too_many_entries: true,
+            map_entries: Vec::new(),
+        }
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -206,8 +206,18 @@ impl TryFrom<proto::rpc_store::account_storage_details::AccountStorageMapDetails
                 .entries
                 .into_iter()
                 .map(|entry| {
-                    let key = entry.key.ok_or(ConversionError::NotAValidFelt)?.try_into()?;
-                    let value = entry.value.ok_or(ConversionError::NotAValidFelt)?.try_into()?;
+                    let key = entry
+                        .key
+                        .ok_or(proto::rpc_store::account_storage_details::account_storage_map_details::map_entries::StorageMapEntry::missing_field(
+                            stringify!(key),
+                        ))?
+                        .try_into()?;
+                    let value = entry
+                        .value
+                        .ok_or(proto::rpc_store::account_storage_details::account_storage_map_details::map_entries::StorageMapEntry::missing_field(
+                            stringify!(value),
+                        ))?
+                        .try_into()?;
                     Ok((key, value))
                 })
                 .collect::<Result<Vec<_>, ConversionError>>()?,
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -435,7 +435,7 @@ const fn storage_slot_type_to_raw(slot_type: StorageSlotType) -> u32 {
 }
 
 /// Represents account details returned in response to an account proof request.
-pub struct AccountDetailsResponse {
+pub struct AccountDetails {
     pub account_header: AccountHeader,
     pub account_code: Option<Vec<u8>>,
     pub vault_details: AccountVaultDetails,
@@ -445,7 +445,7 @@ pub struct AccountDetailsResponse {
 /// Represents the response to an account proof request.
 pub struct AccountProofResponse {
     pub witness: AccountWitness,
-    pub details: Option<AccountDetailsResponse>,
+    pub details: Option<AccountDetails>,
 }
 
 impl TryFrom<proto::rpc_store::AccountProofResponse> for AccountProofResponse {
@@ -475,15 +475,13 @@ impl From<AccountProofResponse> for proto::rpc_store::AccountProofResponse {
     }
 }
 
-impl TryFrom<proto::rpc_store::account_proof_response::AccountDetailsResponse>
-    for AccountDetailsResponse
-{
+impl TryFrom<proto::rpc_store::account_proof_response::AccountDetails> for AccountDetails {
     type Error = ConversionError;
 
     fn try_from(
-        value: proto::rpc_store::account_proof_response::AccountDetailsResponse,
+        value: proto::rpc_store::account_proof_response::AccountDetails,
     ) -> Result<Self, Self::Error> {
-        let proto::rpc_store::account_proof_response::AccountDetailsResponse {
+        let proto::rpc_store::account_proof_response::AccountDetails {
             header,
             code,
             vault_details,
@@ -491,31 +489,25 @@ impl TryFrom<proto::rpc_store::account_proof_response::AccountDetailsResponse>
         } = value;
 
         let account_header = header
-            .ok_or(
-                proto::rpc_store::account_proof_response::AccountDetailsResponse::missing_field(
-                    stringify!(header),
-                ),
-            )?
+            .ok_or(proto::rpc_store::account_proof_response::AccountDetails::missing_field(
+                stringify!(header),
+            ))?
             .try_into()?;
 
         let vault_details = vault_details
-            .ok_or(
-                proto::rpc_store::account_proof_response::AccountDetailsResponse::missing_field(
-                    stringify!(vault_details),
-                ),
-            )?
+            .ok_or(proto::rpc_store::account_proof_response::AccountDetails::missing_field(
+                stringify!(vault_details),
+            ))?
             .try_into()?;
         let account_code = code;
 
         let storage_details = storage_details
-            .ok_or(
-                proto::rpc_store::account_proof_response::AccountDetailsResponse::missing_field(
-                    stringify!(storage_details),
-                ),
-            )?
+            .ok_or(proto::rpc_store::account_proof_response::AccountDetails::missing_field(
+                stringify!(storage_details),
+            ))?
             .try_into()?;
 
-        Ok(AccountDetailsResponse {
+        Ok(AccountDetails {
             account_header,
             account_code,
             vault_details,
@@ -524,11 +516,9 @@ impl TryFrom<proto::rpc_store::account_proof_response::AccountDetailsResponse>
     }
 }
 
-impl From<AccountDetailsResponse>
-    for proto::rpc_store::account_proof_response::AccountDetailsResponse
-{
-    fn from(value: AccountDetailsResponse) -> Self {
-        let AccountDetailsResponse {
+impl From<AccountDetails> for proto::rpc_store::account_proof_response::AccountDetails {
+    fn from(value: AccountDetails) -> Self {
+        let AccountDetails {
             account_header,
             account_code,
             storage_details,
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -556,18 +556,18 @@ impl TryFrom<proto::rpc_store::account_proof_response::AccountDetails> for Accou
             ))?
             .try_into()?;
 
+        let storage_details = storage_details
+            .ok_or(proto::rpc_store::account_proof_response::AccountDetails::missing_field(
+                stringify!(storage_details),
+            ))?
+            .try_into()?;
+
         let vault_details = vault_details
             .ok_or(proto::rpc_store::account_proof_response::AccountDetails::missing_field(
                 stringify!(vault_details),
             ))?
             .try_into()?;
         let account_code = code;
-
-        let storage_details = storage_details
-            .ok_or(proto::rpc_store::account_proof_response::AccountDetails::missing_field(
-                stringify!(storage_details),
-            ))?
-            .try_into()?;
 
         Ok(AccountDetails {
             account_header,
@@ -582,21 +582,21 @@ impl From<AccountDetails> for proto::rpc_store::account_proof_response::AccountD
     fn from(value: AccountDetails) -> Self {
         let AccountDetails {
             account_header,
-            account_code,
             storage_details,
+            account_code,
             vault_details,
         } = value;
 
         let header = Some(proto::account::AccountHeader::from(account_header));
+        let storage_details = Some(storage_details.into());
         let code = account_code;
         let vault_details = Some(vault_details.into());
-        let storage_details = Some(storage_details.into());
 
         Self {
             header,
+            storage_details,
             code,
             vault_details,
-            storage_details,
         }
     }
 }
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -105,75 +105,174 @@ impl From<&AccountInfo> for proto::account::AccountDetails {
 /// Represents a request for an account proof.
 pub struct AccountProofRequest {
     pub account_id: AccountId,
-    pub account_details: Option<AccountDetailRequest>,
+    pub block_num: BlockNumber,
+    pub details: Option<AccountDetailRequest>,
 }
 
 impl TryFrom<proto::rpc_store::AccountProofRequest> for AccountProofRequest {
     type Error = ConversionError;
 
-    fn try_from(request: proto::rpc_store::AccountProofRequest) -> Result<Self, Self::Error> {
-        let proto::rpc_store::AccountProofRequest { account_id, account_details } = request;
+    fn try_from(value: proto::rpc_store::AccountProofRequest) -> Result<Self, Self::Error> {
+        let proto::rpc_store::AccountProofRequest { account_id, block_num, details } = value;
 
         let account_id = account_id
             .ok_or(proto::rpc_store::AccountProofRequest::missing_field(stringify!(account_id)))?
             .try_into()?;
 
-        let account_details = account_details.map(TryFrom::try_from).transpose()?;
+        let details = details.map(TryFrom::try_from).transpose()?;
 
-        Ok(AccountProofRequest { account_id, account_details })
+        Ok(AccountProofRequest {
+            account_id,
+            block_num: block_num.into(),
+            details,
+        })
     }
 }
 
 /// Represents a request for account details alongside specific storage data.
 pub struct AccountDetailRequest {
     pub code_commitment: Option<Word>,
-    pub storage_requests: Vec<StorageMapKeysProof>,
+    pub storage_requests: Vec<StorageMapRequest>,
+    pub include_assets: bool,
 }
 
-impl TryFrom<proto::rpc_store::account_proof_request::AccountDetailsRequest>
+impl TryFrom<proto::rpc_store::account_proof_request::AccountDetailRequest>
     for AccountDetailRequest
 {
     type Error = ConversionError;
 
     fn try_from(
-        request: proto::rpc_store::account_proof_request::AccountDetailsRequest,
+        value: proto::rpc_store::account_proof_request::AccountDetailRequest,
     ) -> Result<Self, Self::Error> {
-        let proto::rpc_store::account_proof_request::AccountDetailsRequest {
+        let proto::rpc_store::account_proof_request::AccountDetailRequest {
+            code_commitment,
+            storage_maps,
+            include_assets,
+        } = value;
+
+        let code_commitment = code_commitment.map(TryFrom::try_from).transpose()?;
+        let storage_requests = try_convert(storage_maps).collect::<Result<_, _>>()?;
+
+        Ok(AccountDetailRequest {
             code_commitment,
             storage_requests,
-        } = request;
-
-        let storage_requests = try_convert(storage_requests).collect::<Result<_, _>>()?;
-        let code_commitment = code_commitment.map(Word::try_from).transpose()?;
-
-        Ok(AccountDetailRequest { code_commitment, storage_requests })
+            include_assets,
+        })
     }
 }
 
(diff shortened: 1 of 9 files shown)


crates/store/src/state.rs
```diff
@@ -883,101 +892,130 @@ impl State {
...
+    ) -> Result<AccountProof, DatabaseError> {
+        const RESPONSE_COUNT_LIMIT: usize = 100;
         // Lock inner state for the whole operation. We need to hold this lock to prevent the
         // database, account tree and latest block number from changing during the operation,
         // because changing one of them would lead to inconsistent state.
         let inner_state = self.inner.read().await;
 
-        let account_ids: Vec<AccountId> =
-            account_requests.iter().map(|req| req.account_id).collect();
+        let AccountProofRequest { account_id, block_num: _, details } = account_request;
```
drahnr: The `block_num` will be handled once the relevant data is stored #1185 , which is part of step 3, the next PR.

bobbinth:
@drahnr - could you rebase this from the latest `next`? Seems like some of the changes from #1211 are not here yet.

crates/store/src/state.rs
```diff
@@ -887,78 +890,123 @@ impl State {
...
-            account_id, // FIXME TODO see account.proto we need to triple check this is correct
-            witness: inner_state.account_tree.open(account_id),
-        };
-
         let response = AccountProofResponse {
-            block_num: inner_state.latest_block_num(),
-            witness,
-            account_details,
+            witness: inner_state.account_tree.open(account_id),
+            details: account_details,
```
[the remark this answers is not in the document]
drahnr: >The `AccountProofRequest` implementation ignores the `block_num` parameter, always returning current account state instead of historical state.

Correct, see https://github.com/0xMiden/miden-node/issues/1227 for actually keeping that historical data.


> Additionally, the RESPONSE_COUNT_LIMIT check for storage maps and vault assets is too broad: it prevents returning specific requested keys or assets if the total map/vault size exceeds the limit, and it has an off-by-one error, triggering for 101+ entries instead of 100+.

This is meant as broad stroke limit, and not relevant for the time.

crates/store/src/state.rs
```diff
@@ -887,78 +890,123 @@ impl State {
...
+            include_assets,
+            storage_requests,
+        }) = account_request.details
         {
             let info = self.db.select_account(account_id).await?;
 
-            if let Some(details) = &info.details {
-                let mut storage_proofs = Vec::new();
+            // if we get a query for a private account, we'll return `None`
+            if let Some(details) = info.details {
```
Mirko-von-Leipzig: Good place to use the `let Some(x) = x else { return }` imo, the `else` is so far removed that the branching is hard to parse.


```suggestion
            // if we get a query for a private account, we'll return `None`
            let Some(details) = info.details else {
                return Err(AccountError::StorageSlotNotMap(slot_index).into());
            };
```

crates/store/src/state.rs
```diff
@@ -887,80 +890,125 @@ impl State {
...
+
+                let mut storage_map_details = Vec::<AccountStorageMapDetails>::new();
 
-                for StorageMapKeysProof { storage_index, storage_keys } in
-                    &account_detail_request.storage_requests
-                {
+                for StorageMapRequest { slot_index, slot_data } in storage_requests {
                     if let Some(StorageSlot::Map(storage_map)) =
-                        details.storage().slots().get(*storage_index as usize)
+                        details.storage().slots().get(slot_index as usize)
```
Mirko-von-Leipzig: This could also benefit from `let Some(..) else { return Err(...); )}`

crates/store/src/state.rs
```diff
@@ -887,80 +890,125 @@ impl State {
...
                     {
-                        for map_key in storage_keys {
-                            let proof = storage_map.open(map_key);
-
-                            let slot_map_key = StorageSlotMapProof {
-                                storage_slot: *storage_index,
-                                proof: proof.into(),
+                        if storage_map.entries().nth(RESPONSE_COUNT_LIMIT).is_some() {
```
Mirko-von-Leipzig: Nit: could this be 
```suggestion
                        if storage_map.len() >= RESPONSE_COUNT_LIMIT {
```

crates/store/src/state.rs
```diff
@@ -887,80 +890,125 @@ impl State {
...
+                        } else {
+                            let converter =
+                                |key: &Word, value: &Word| -> (Word, Word) { (*key, *value) };
+                            let map_entries = match slot_data {
+                                SlotData::All => Vec::from_iter(
+                                    storage_map.entries().map(|(k, v)| converter(k, v)),
+                                ),
+                                SlotData::MapKeys(keys) => Vec::from_iter(
+                                    keys.iter().map(|key| converter(key, &storage_map.get(key))),
+                                ),
```
Mirko-von-Leipzig: Should we maybe move this logic into a constructor for `AccountStorageMapDetails`? Basically everything including the `if/else` would be my suggestion i.e. 

```rust
impl AccountStorageMapDetails {
    fn from_map(index: usize, entries: &Entries) -> Self
}
```

crates/store/src/state.rs
```diff
@@ -887,80 +890,125 @@ impl State {
...
-                                proof: proof.into(),
+                        if storage_map.entries().nth(RESPONSE_COUNT_LIMIT).is_some() {
+                            storage_map_details.push(AccountStorageMapDetails {
+                                slot_index,
+                                too_many_entries: true,
+                                map_entries: Vec::new(),
+                            });
+                        } else {
+                            let converter =
+                                |key: &Word, value: &Word| -> (Word, Word) { (*key, *value) };
```
Mirko-von-Leipzig: Is this not just `.cloned()` or `.copied()`?

crates/store/src/state.rs
```diff
@@ -887,80 +890,125 @@ impl State {
...
     /// Returns the respective account proof with optional details, such as asset and storage
     /// entries.
+    #[allow(clippy::too_many_lines)]
     pub async fn get_account_proof(
         &self,
         account_request: AccountProofRequest,
     ) -> Result<AccountProofResponse, DatabaseError> {
+        // If this number is exceed
+        // TODO use a total size to derive
+        const RESPONSE_COUNT_LIMIT: usize = 100;
```
Mirko-von-Leipzig: Comment is incomplete

crates/proto/src/domain/account.rs
```diff
@@ -105,75 +105,174 @@ impl From<&AccountInfo> for proto::account::AccountDetails {
...
+            .map(|slot| {
+                let slot_type = storage_slot_type_from_raw(slot.slot_type)?;
+                let commitment =
+                    slot.commitment.ok_or(ConversionError::NotAValidFelt)?.try_into()?;
+                Ok((slot_type, commitment))
+            })
+            .collect::<Result<Vec<_>, ConversionError>>()?;
+
+        // Create AccountStorageHeader from the items
+        // Note: This might need adjustment based on the actual AccountStorageHeader constructor
```
Mirko-von-Leipzig: I'm not sure what these comments mean?

crates/store/src/state.rs
drahnr: `entries()` yields an iterator, so we only get a guess on the size. Using it with `skip` allows us to get an accurate value that's not impl dependent.

Mirko-von-Leipzig: @PhilippGackstatter would it be possible to add a length/count here?

crates/store/src/state.rs
drahnr: Done, although it got a bit hairier than anticipated.

crates/store/src/state.rs
drahnr: Removed

crates/store/src/state.rs
PhilippGackstatter: Sure, opened https://github.com/0xMiden/miden-base/pull/1935.

I think this just wasn't exposed because the `StorageMap` is a relatively "old" type and `Smt::num_entries` was a relatively recent addition.

bobbinth:
@drahnr - could you merge the latest `next` into this branch. I think currently it contains some changes that are already in `next`.

proto/proto/store/rpc.proto
```diff
@@ -103,71 +103,123 @@ message StoreStatus {
...
     }
 
-    // The account ID for this request.
+    // ID of the account for which we want to get data
     account.AccountId account_id = 1;
 
-    // A account detail requests, including map keys + values.
-    optional AccountDetailsRequest account_details = 2;
+    // Block at which we'd like to get this data. Must be close to the chain tip.
+    fixed32 block_num = 2;
```
bobbinth, line 149: In the current implementation, this would be ignored, right?

proto/proto/store/rpc.proto
```diff
@@ -103,71 +103,123 @@ message StoreStatus {
...
+        // only if its commitment is different from this value or if the value is not present.
         optional primitives.Digest code_commitment = 1;
 
-        // List of storage requests for this account.
-        repeated StorageRequest storage_requests = 2;
+        // A flag indicating whether to include asset vault data.
+        //
+        // Assets will only be returned for small accounts with less than 100 asset entries,
+        // otherwise they have to be requested separately.
+        bool include_assets = 2;
```
bobbinth: Should we make this work similar to `code_commitment`? i.e., we'd have something like `optional Digest asset_vault_commitment` and send assets only if the current commitment doesn't match the provided one?

proto/proto/store/rpc.proto
```diff
@@ -103,71 +103,123 @@ message StoreStatus {
...
+                // A list of map keys associated with the given storage slot identified by `slot_index`.
+                MapKeys map_keys = 3;
+            }
         }
 
-        // Account code commitment corresponding to the last-known `AccountCode` for the requested
-        // account. The response will include only code that is known.
+        // Last known code commitment to the client. The response will include account code
+        // only if its commitment is different from this value or if the value is not present.
         optional primitives.Digest code_commitment = 1;
```
bobbinth, line 132: To clarify the logic:

- If we send the commitment and it matches the one held by the node, we don't send anything back.
- If we send the commitment but it doesn't match the one held by the node, we send the code back.
- If we don't send the commitment, would we send the code back or no?

For the last one, I'm thinking the answer should be "no". This could be used as a way for the requestor to signal that they are not interested in the code.

proto/proto/store/rpc.proto
```diff
@@ -103,71 +103,123 @@ message StoreStatus {
...
 
-        // Account code, if the current account code does not match the request provided commitment digest.
-        optional bytes account_code = 3;
+        // Account asset vault data; empty if vault commitments matched
+        optional AccountVaultDetails vault_details = 3;
 
-        // Storage slots information for this account
-        repeated StorageSlotMapProof storage_maps = 4;
+        // Account storage data; empty if storage commitments matched
+        optional AccountStorageDetails storage_details = 4;
```
bobbinth: I don't think we currently send storage commitment - so, this probably shouldn't be optional. If so, I'd move it up to be right after `header`.

proto/proto/store/rpc.proto
```diff
@@ -103,71 +103,123 @@ message StoreStatus {
...
+            // Storage slot index (`[0..255]`).
+            uint32 slot_index = 1;
+
+            oneof slot_data {
+                // Request to return all storage map data. If the number exceeds a threshold of 1000 entries,
+                // the response will not contain them but must be requested separately.
+                bool all_entries = 2;
+
+                // A list of map keys associated with the given storage slot identified by `slot_index`.
+                MapKeys map_keys = 3;
```
bobbinth, line 126: From what I can tell, the current response does not support retrieval of specific keys from storage maps. If so, a request with this property cannot be fulfilled. Should we remove it for now?

crates/store/src/state.rs
```diff
@@ -897,70 +900,77 @@ impl State {
...
-                    },
-                    _ => Some(details.code().to_bytes()),
-                };
+            // if we get a query for a private account, we'll return `None`
+            let Some(details) = info.details else {
+                return Ok(AccountProofResponse {
+                    witness: inner_state.account_tree.open(account_id),
+                    details: None,
+                });
+            };
```
bobbinth: Shouldn't this be an error? i.e., asking for details for a private account should not be allowed.

proto/proto/store/rpc.proto
drahnr: Correct, this is for a later PR, but it could spare us one extra PR on the client.

proto/proto/store/rpc.proto
drahnr: I think that's a good idea, the asymmetry seemed odd.

proto/proto/store/rpc.proto
drahnr: I think previously we discussed that we'd want to send it back, since the requester might not have any code iiuc? So we'd either have to add a bool/enum to state that or have a "magic" value (`0x000..000`) to ensure a mismatch. While I don't like the latter, I think in this case it'd be the right approach _if_ it's a frequent use-case to _not_ be interested in the code at all.

proto/proto/store/rpc.proto
drahnr: My understanding was that it might be optional if i.e. someone only wants the code, unless that's not a use-case?

proto/proto/store/rpc.proto
bobbinth: I think the `0x000..000` could work fine here (no code commitment should ever match it).

proto/proto/store/rpc.proto
drahnr: I'll be working on this today.

crates/store/src/state.rs
drahnr: Don't we still want to return the witness?

proto/proto/store/rpc.proto
bobbinth: We could do this, but it'd complicate things a bit more. The tradeoff here is that we'd always send the storage header, but in the worst case, that's 8KB (and in most cases it'll be under 1KB). I think that's a tradeoff worth making.

crates/store/src/state.rs
bobbinth: We do, but the witness is not a part of the "details" (it is a part of the response). In general, I think it is on the user to build the request correctly, and asking for details for a private account is an badly constructed request.

proto/proto/store/rpc.proto
drahnr: Actually this is implemented in 
```
    pub fn new(slot_index: u8, slot_data: SlotData, storage_map: &StorageMap) -> Self {
        let exceeds = if let SlotData::MapKeys(keys) = &slot_data {
            keys.len() > Self::MAX_RETURN_ENTRIES
        } else {
            storage_map.entries().nth(Self::MAX_RETURN_ENTRIES).is_some()
        };
        if exceeds {
            Self::too_many_entries(slot_index)
        } else {
            Self::from_slot_data(slot_index, slot_data, storage_map)
        }
    }
    ```

File: crates/store/src/state.rs (hunk 15 above)
Comment (drahnr) [reply crates/store/src/state.rs]: Took a few seconds to wrap my head around; tl;dr correct request means `never have a ``Some(details)`` in the request when it's a private account`, this would trigger the `else` branch and we should error out and not serve "however much we have".

I'll implement that and just bail.

File: proto/proto/store/rpc.proto (hunk 12 above)
Comment (drahnr) [reply proto/proto/store/rpc.proto:132]: This would be only a client specific logic, nothing to do on the server side.

File: crates/proto/src/domain/account.rs [hunk 16]
```diff
@@ -105,75 +106,171 @@ impl From<&AccountInfo> for proto::account::AccountDetails {
...
-        Ok(AccountDetailRequest { code_commitment, storage_requests })
+        Ok(AccountDetailRequest {
+            code_commitment,
+            asset_vault_commitment,
+            storage_requests,
+        })
     }
 }
 
 /// Represents a request for an account's storage map values and its proof of existence.
```
igamigo: Looks misplaced now

crates/proto/src/domain/account.rs
```diff
@@ -105,75 +106,171 @@ impl From<&AccountInfo> for proto::account::AccountDetails {
...
+        let slot_index = slot_index.try_into().map_err(ConversionError::TryFromIntError)?;
+
+        // Extract map_entries from the entries field
+        let map_entries = match entries {
+            Some(entries) => entries
+                .entries
+                .into_iter()
+                .map(|entry| {
+                    let key = entry.key.ok_or(ConversionError::NotAValidFelt)?.try_into()?;
+                    let value = entry.value.ok_or(ConversionError::NotAValidFelt)?.try_into()?;
```
igamigo: nit: should these errors be `MissingFieldInProtobufRepresentation`?

crates/proto/src/domain/account.rs
```diff
@@ -243,6 +340,147 @@ impl From<AccountStorageHeader> for proto::account::AccountStorageHeader {
...
+    const MAX_RETURN_ENTRIES: usize = 1000;
+
+    pub fn new(vault: &AssetVault) -> Self {
+        if vault.assets().nth(Self::MAX_RETURN_ENTRIES).is_some() {
+            Self::too_many()
+        } else {
+            Self {
+                too_many_assets: false,
+                assets: Vec::from_iter(vault.assets()),
+            }
```
igamigo, line 367: This iterates over the vault's assets twice, would be nice to do it in one go but not sure there's a nice way to do that either

crates/store/src/state.rs
```diff
@@ -897,70 +900,87 @@ impl State {
...
+                    .storage()
+                    .slots()
+                    .iter()
+                    .map(|storage_slot| (storage_slot.slot_type(), storage_slot.value())),
+            );
 
-                Some(account_details_response)
-            } else {
-                None
+            let storage_header = AccountStorageHeader::new(slot_headers);
```
igamigo: nit: `AccountStorage::to_header` should work here

crates/store/src/state.rs
```diff
@@ -897,70 +900,87 @@ impl State {
...
+
+            let mut storage_map_details = Vec::<AccountStorageMapDetails>::new();
```
igamigo: nit: 
```suggestion
            let mut storage_map_details = Vec::<AccountStorageMapDetails>::with_capacity(storage_requests.len());
```

crates/store/src/state.rs
```diff
@@ -897,70 +900,87 @@ impl State {
...
+                    details.storage().slots().get(slot_index as usize)
+                else {
+                    return Err(AccountError::StorageSlotNotMap(slot_index).into());
+                };
+                let details = AccountStorageMapDetails::new(slot_index, slot_data, storage_map);
+                storage_map_details.push(details);
             }
+
+            // Only include unknown account code blobs, which is equal to a account code digest
+            // mismatch. If `None` was requested, don't return any.
```
igamigo, line 932: I think this differs from the previous approach where we did respond with the account code if it was `None`, right? Because it signaled that the user did not know the code. 

Is the idea that if we want the code we send a dummy commitment now? I think that works but also I don't think there's  valid usecases for sending `None` here and not expecting the code back AFAICT.

proto/proto/store/rpc.proto
igamigo: Never mind my other comment, just read this. I think this would work fine although from the client itself I don't think there currently any usecase where we don't want to get the account code.

proto/proto/store/rpc.proto
```diff
@@ -103,71 +103,123 @@ message StoreStatus {
...
-        // account. The response will include only code that is known.
+        // Last known code commitment to the client. The response will include account code
+        // only if its commitment is different from this value or if the value is not present.
         optional primitives.Digest code_commitment = 1;
 
-        // List of storage requests for this account.
-        repeated StorageRequest storage_requests = 2;
+        // Last known asset vault commitment to the client. The response will include asset vault data
+        // only if its commitment is different from this value or if the value is not present.
+        // If the number of to-be-returned asset entries exceed a threshold, they have to be requested
```
igamigo, line 137: I think I'd make this consistent with the account code request and not return anything when the value is `None`

crates/proto/src/domain/account.rs
drahnr: We unfortunately don't get access to the underlying data structure, but only in form of an iterator

crates/store/src/state.rs
drahnr: The interpretation here changed, can you take a look @bobbinth

proto/proto/store/rpc.proto
drahnr: I think the comment has gone stale, will address

crates/store/src/state.rs
```diff
@@ -898,70 +900,80 @@ impl State {
...
-                        if known_code_commitment == details.code().commitment() =>
-                    {
-                        // the known code matches the expected commitment, which implies
-                        // the user already has that code
-                        None
-                    },
-                    _ => Some(details.code().to_bytes()),
-                };
+            // if we get a query for a private account _with_ details requested, we'll error out
+            let Some(details) = info.details else {
```
PhilippGackstatter: Nit: On first read, it is confusing that `details` if of type `Account`, should it be `account`? Same for `info.details` -> `info.account`.

crates/store/src/state.rs
```diff
@@ -898,70 +900,80 @@ impl State {
...
+                    AccountVaultDetails::empty()
+                },
+                Some(_) => {
+                    // Either no commitment provided or it doesn't match, so return vault data
+                    AccountVaultDetails::new(details.vault())
+                },
+                None => {
+                    // No commitment provided, so don't return vault data
+                    AccountVaultDetails::empty()
+                },
```
PhilippGackstatter, line 961: > Either no commitment provided

This is handled by the `None` case right?

proto/proto/store/rpc.proto
```diff
@@ -103,71 +103,124 @@ message StoreStatus {
...
+
+            repeated StorageMapEntry entries = 1;
+        }
+        // slot index of the storage map
+        uint32 slot_index = 1;
+
+        // A flag that is set to true if all_entries == true was used in the request for this
+        // storage map and the map contains too many entries. This indicates to the user
+        // that `SyncStorageMaps` endpoint should be used to get all storage map data
+        bool too_many_entries = 2;
```
PhilippGackstatter: Is this comment up to date? `AccountStorageMapDetails::new` seems to always set this to `true` if `MAX_RETURN_ENTRIES` is exceeded independent of `all_entries`.

Nit: Should `all_entries` be `AllEntries` - not sure how casing maps in gRPC though tbh.

crates/store/src/state.rs
```diff
@@ -898,70 +900,80 @@ impl State {
...
+
+            for StorageMapRequest { slot_index, slot_data } in storage_requests {
+                let Some(StorageSlot::Map(storage_map)) =
+                    details.storage().slots().get(slot_index as usize)
+                else {
+                    return Err(AccountError::StorageSlotNotMap(slot_index).into());
+                };
+                let details = AccountStorageMapDetails::new(slot_index, slot_data, storage_map);
+                storage_map_details.push(details);
             }
```
PhilippGackstatter, line 929: Question out of curiosity: Why does this not return a merkle proof that proves inclusion of the entry in the storage map? I would have thought that the client needed this, but it seems it doesn't.

crates/proto/src/domain/account.rs
PhilippGackstatter: I think you are one miden-base commit behind https://github.com/0xMiden/miden-base/commit/b6f679b2bd21f4f977f950242a0cbbfed1c6ab59 which merged https://github.com/0xMiden/miden-base/pull/1939. If you can update to that, you could use `num_assets`.

crates/store/src/state.rs
drahnr: We do have the `proto::account::AccountDetails` which binds both of them together, the representation in rust of an account with all details is just called `Account`. I don't think naming all of these `account` or `details` improves much, I'd argue we should rename `info` to `account_info` :shrug: open for others suggestions

proto/proto/store/rpc.proto
drahnr: >Nit: Should all_entries be AllEntries - not sure how casing maps in gRPC though tbh.

`all_entries` is fine, it's the field name, the type is `bool`. 

> Is this comment up to date? AccountStorageMapDetails::new seems to always set this to true if MAX_RETURN_ENTRIES is exceeded independent of all_entries.

Fixed

crates/store/src/state.rs
drahnr: I am not fully in the loop how the client uses it CC @igamigo might be able to help here if this is OK, I'll double check tomorrow

crates/store/src/state.rs
PhilippGackstatter: For me, it was mostly about the local context being unclear. `account` seems less generic than `info` or `details`, so I'm all for improving it one way or another.

crates/store/src/state.rs
drahnr: As I see it, the client currently doesn't use the account storage map details, but only the _account_ proofs. So the usage would remain the same.
Do we need the proofs for FPI calls? I'd assume so? CC @bobbinth

I don't necessarily want to block the PR any longer

proto/proto/store/rpc.proto
```diff
@@ -103,71 +103,124 @@ message StoreStatus {
...
+                // A list of map keys associated with the given storage slot identified by `slot_index`.
+                MapKeys map_keys = 3;
+            }
         }
 
-        // Account code commitment corresponding to the last-known `AccountCode` for the requested
-        // account. The response will include only code that is known.
+        // Last known code commitment to the client. The response will include account code
+        // only if its commitment is different from this value or if the value is not present.
         optional primitives.Digest code_commitment = 1;
```
bobbinth, line 132: I think the comment here should be something like:
```
// Last known code commitment to the client. The response will include account code
// only if its commitment is different from this value.
//
// If the field is omitted, the response will not include account code.
```

Also, I wonder if we should not explicitly reference the "client" here (and in other places) and use some broader term - e.g., "requestor" or something like that.

proto/proto/store/rpc.proto
```diff
@@ -103,71 +103,124 @@ message StoreStatus {
...
 
-        // Account code, if the current account code does not match the request provided commitment digest.
-        optional bytes account_code = 3;
+        // Account code; empty if code commitments matched or none was requested
+        optional bytes code = 3;
 
-        // Storage slots information for this account
-        repeated StorageSlotMapProof storage_maps = 4;
+        // Account asset vault data; empty if vault commitments matched
+        optional AccountVaultDetails vault_details = 4;
```
bobbinth, line 169: Should this be: "Account asset vault data; empty if vault commitments matched **or none was requested**".

crates/store/src/state.rs
bobbinth: I think this was covered in another comment, but yes, this is slightly different. Now, the client can do the following:

- If the client doesn't know the code send `0x0000...` as commitment. This would guarantee that the current code is returned.
- If the client has the code, send its commitment, and then the code would be returned only if there is a mismatch.
- There is now also the 3rd option that the client doesn't currently use: send `None` to indicate that the client is not interested in the code.

crates/store/src/state.rs
bobbinth: There are 3 ways to fulfill these requests and currently we are supporting two of them. The requests types are:

1. We request all entries of the storage map: if the number of entries is less than some threshold $n$ (currently, I think $n = 1000$ but maybe it should be smaller), then we return all entries and the client can reconstruct the SMT locally.
2. We request specific set of key-value pairs. Here we have two options:
  a. The total number of key-value pairs in the SMT is less than $n$, in which case we return all key-value pairs.
  b. The total number of key-value pairs in the SMT is greater than $n$, in which case we'd need to return Merkle paths for each returned key-value pair (and we can also apply path compaction here).

Case 2b is currently not implemented. Also, looking at the code, I'm not sure the case 2a handled correctly. Specifically, `AccountStorageMapDetails::from_slot_data()` for now should return all entries if the storage size under a given threshold and return `too_many_entries` otherwise.

The reason for the above is that if we don't return all entries (and don't return Merkle paths in case of partial entries), the client would have no way to reconstruct the underling SMT - and so the returned data would not be usable.

I think we should do the following:

- Open a small PR to fix handling of case 2a.
- Create a follow-up issue to implement handling of case 2b.

PR #1215, opened 2025-09-09
chore: update `miden-base` to next

tomyrd:
This PR updates the node's `miden-base` version to `next`. This is needed to suppress some compilation errors and also to advance this same version in the client.

### DIFF
--- a/crates/ntx-builder/src/transaction.rs
+++ b/crates/ntx-builder/src/transaction.rs
@@ -294,14 +294,14 @@ impl DataStore for NtxDataStore {
 
             if account.vault().root() != vault_root {
                 return Err(DataStoreError::Other {
-                    error_msg: "Vault root mismatch".into(),
+                    error_msg: "vault root mismatch".into(),
                     source: None,
                 });
             }
 
             AssetWitness::new(account.vault().asset_tree().open(&vault_key)).map_err(|err| {
                 DataStoreError::Other {
-                    error_msg: "Failed to open vault asset tree".into(),
+                    error_msg: "failed to open vault asset tree".into(),
                     source: Some(Box::new(err)),
                 }
             })


crates/ntx-builder/src/transaction.rs
```diff
@@ -277,7 +276,35 @@ impl DataStore for NtxDataStore {
...
+    ) -> impl FutureMaybeSend<Result<AssetWitness, DataStoreError>> {
+        let account = self.account.clone();
+        async move {
+            if account.id() != account_id {
+                return Err(DataStoreError::AccountNotFound(account_id));
+            }
+
+            if account.vault().root() != vault_root {
+                return Err(DataStoreError::Other {
+                    error_msg: "Vault root mismatch".into(),
```
Mirko-von-Leipzig: Nit: error messages should be lowercase so they can be chained

```suggestion
                    error_msg: "vault root mismatch".into(),
```

crates/ntx-builder/src/transaction.rs
```diff
@@ -277,7 +276,35 @@ impl DataStore for NtxDataStore {
...
+                    source: None,
+                });
+            }
+
+            AssetWitness::new(account.vault().asset_tree().open(&vault_key)).map_err(|err| {
+                DataStoreError::Other {
+                    error_msg: "Failed to open vault asset tree".into(),
```
Mirko-von-Leipzig:
```suggestion
                    error_msg: "failed to open vault asset tree".into(),
```

Mirko-von-Leipzig (approved):
@tomyrd  Could you elaborate on this?

>  This is needed to suppress some compilation errors

tomyrd:
> Could you elaborate on this?

Sure, I should've explained it better. The compilation errors were appearing in the client and were referencing code from the node. This probably happens because we import some of the node's crates.

After these changes and pointing to this branch, they no longer appear.

Issue #1190, opened 2025-09-01
Network monitoring app


Mirko-von-Leipzig:
Telemetry isn't always sufficient to determine the liveness of a system. Create a separate light-weight binary that periodically performs some tasks to determine the status of various network components.

This should include a webpage that users could use to see the current network status.

I would recommend a design where each test/task runs forever in a separate task. This makes it easy to write and add additional tasks, maybe with a central controller to handle task crashes and reset them.

Initial suggestions

1. Basic status checks using gRPC status endpoints. This would also let us communicate the current version of each component.
2. Faucet workflow. Mint some tokens from the official faucet.
3. Use the delegated prover to prove some transactions. Note that these don't need to be submitted aka could be a static transaction since we're only testing the prover.
4. Submit a transaction to the network.

You may be tempted to combine some of these e.g. submit a tx and also prove a tx but I would suggest keeping them separate to make things easier to maintain and change.

I suggest keeping things simple initially e.g. each task is just a loop

```rust
loop {
    run_task().await
        .inspect_err(|err| tracing::error!(....))
        .inspect_ok(|x| tracing::info!(...));
    tokio::sleep(Duration::from_secs(60));
}
```

## Telemetry

I would _also_ suggest we emit OTel telemetry so we can create triggers and alerts. Essentially this app would function as a third party _external_ view of our network. Internal monitoring is important, but external can pick up issues which are more likely to present to the end user.

bobbinth:
I think the overall approach works. I'd separate this into 3 tasks:

1. First, we'll need to add a command to the monitor CLI to create and deploy the "test" account. We should also make a provision to allow the monitor to use a previously created account.
2. Then, we'll need to update the monitor to periodically send transactions against this account. We can verify that this is working manually (e.g., via a block explorer or otherwise).
3. Lastly, we'll need to update the monitor to read the state of the chain (probably by making direct RPC requests) to make the account state has in fact changed as expected.

We could combine steps 2 and 3 into a single PR - but they could also be done separately.

One thing that would be interesting to track is the delay between when the transaction was submitted to the network, and when it appeared in a block.

------

Separately, I think one question is whether we want to make the test account a local account (as I think you describe in the issue) or a network account. Going with the network account would be a bit more work, but it would also let us test network transactions. Though, it may be good to test both local and network transactions separately, and in that case, starting with a local account makes sense.

PR #1217, opened 2025-09-09
feat: add network monitoring application

SantiagoPittella:
partially addresses #1190 

This is the current look of the frontned
Screenshot 2025-09-09 at 17 50 19


bin/network-monitoring/README.md
```diff
@@ -0,0 +1,32 @@
+# Miden network monitoring
+
+This crate contains a binary for running a Miden network monitor that can monitor multiple remote provers.
```
Mirko-von-Leipzig: Maybe something like

```suggestion
A monitoring app for a Miden network's infrastructure.

It serves a webpage with an overview of the current infrastructure status and emits OpenTelemetry events which can be ingested for more advanced monitoring and alerting.
```

bin/network-monitoring/README.md
Mirko-von-Leipzig, line 1: Maybe also a section on currently supported items, and future items?

.github/workflows/network-monitoring.yml
Mirko-von-Leipzig, line 1: Could we get away with just `cargo check`? But maybe the full one is better - I'm mainly concerned with keeping github cache usage low. So maybe we keep as is, and then if cache gets clobbered we revisit.

bin/network-monitoring/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+    let frontend_config = config.clone();
+    let frontend_task = tokio::spawn(run_frontend(frontend_status, frontend_config));
+
+    // Start a task to hit status endpoint periodically in the different components.
+    let status_status = shared_status.clone();
+    let status_config = config.clone();
+    let status_task = tokio::spawn(check_status(status_status, status_config));
+
+    // Wait for either task to complete or fail, then abort the other
+    let (frontend_result, status_result) = tokio::join!(frontend_task, status_task);
```
Mirko-von-Leipzig: I don't think this works - `join!` waits for _all_ tasks to complete:

> Waits on multiple concurrent branches, returning when all branches complete.

For short lists like this `tokio::select!` is probably the way to go; for longer lists I use `JoinSet` with `join_next_with_id` but that involves storing the IDs as well to identify which task completed.

bin/network-monitoring/src/frontend.rs
```diff
@@ -0,0 +1,44 @@
...
+pub async fn get_dashboard() -> Html<&'static str> {
+    Html(include_str!("../assets/index.html"))
+}
+
+pub async fn get_status(
+    axum::extract::State(shared_status): axum::extract::State<SharedStatus>,
+) -> axum::response::Json<crate::status::NetworkStatus> {
+    let status = shared_status.lock().await;
+    axum::response::Json(status.clone())
+}
```
Mirko-von-Leipzig: Do these need to be pub?

bin/network-monitoring/src/frontend.rs
```diff
@@ -0,0 +1,44 @@
...
+
+/// Runs the frontend server.
+///
+/// This function runs the frontend server that serves the dashboard and the status data.
+///
+/// # Arguments
+///
+/// * `shared_status` - The shared status of the network.
+/// * `config` - The configuration of the network.
+pub async fn run_frontend(shared_status: SharedStatus, config: MonitoringConfig) {
```
Mirko-von-Leipzig: Nit: 
```suggestion
pub async fn serve(shared_status: SharedStatus, config: MonitoringConfig) {
```

bin/network-monitoring/src/frontend.rs
```diff
@@ -0,0 +1,44 @@
...
+    // build our application with routes
+    let app = Router::new()
+        // Serve static files from assets directory
+        .nest_service("/assets", ServeDir::new("bin/network-monitoring/assets"))
```
Mirko-von-Leipzig: Does this work when installed via `cargo install`? Because I don't think the `assets` folder will be bundled with unless `ServeDir` somehow embeds the assets into the binary.

I've typically forced this by embedding the files directly in the binary by using `include_str!` and friends, but then you can't use a directory approach.

See https://github.com/0xMiden/miden-faucet/issues/59 for some additional context.

bin/network-monitoring/src/frontend.rs
```diff
@@ -0,0 +1,44 @@
...
+        // Main dashboard route
+        .route("/", get(get_dashboard))
+        // API route for status data
+        .route("/status", get(get_status))
+        .with_state(shared_status);
+
+    // run our app with hyper, listening globally on the configured port
```
Mirko-von-Leipzig: Nit: don't think we use `hyper` directly

```suggestion
```

bin/network-monitoring/src/status.rs
Mirko-von-Leipzig, line 1: This is fine for now but I think we'll run into problems if/when we start adding more checks.

An alternative I would suggest as follow-up is to split each check into its own task, using a tokio `Watch` channel to communicate its latest status.

bin/network-monitoring/assets/index.html
Mirko-von-Leipzig, line 1: @bobbinth should we be worried about exposing IP addresses in the front-end? Its useful for us to see and identify; but _could_ be a problem showing our internal infrastructure mapping?

bobbinth: For the testnet, I don't mind showing them as it may be helpful for us to identify which specific worker is down. But overall, agreed if there is a way to assign names to workers somehow, that would be better than showing IP addresses.

.github/workflows/network-monitoring.yml
SantiagoPittella: It is a relatively simple binary, using just check is probably enough. I will remove both jobs (build, install) and add a check one. I can re-add it if we change our minds

bin/network-monitoring/src/frontend.rs
SantiagoPittella: No, it does not

bin/network-monitoring/src/status.rs
SantiagoPittella: Just created https://github.com/0xMiden/miden-node/issues/1223

bin/network-monitoring/assets/index.html
SantiagoPittella: I can add names like "Worker {index}", is that ok?

bobbinth: How stable are the indexes? (i.e., would they change from time-to-time?)

SantiagoPittella: Unless that we persist them somewhere, we can persist them just in runtime. Meaning that it will make some sense as long as we don't stop the server.

bobbinth: I think that part would be OK - but would we be able to identify somehow which worker maps to which machine?

Maybe we are overthinking it a bit - and for now IP addresses are fine and we can update this part in the future.

Mirko-von-Leipzig: I think let's leave it for now. I suspect there will be more changes before we leave testnet :)

bin/network-monitoring/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+authors.workspace      = true
+description            = "Miden network monitoring"
+edition.workspace      = true
+homepage.workspace     = true
+keywords               = ["miden", "monitoring", "network"]
+license.workspace      = true
+name                   = "miden-network-monitoring"
```
bobbinth: nit: I would name this `miden-network-monitor` and update the description accordingly.

Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [workspace]
 members = [
+  "bin/network-monitoring",
```
bobbinth: Similar comment as above: I'd call this `miden-network-monitor` and this is applicable in many other places as well.

bin/network-monitoring/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Usage
+
+```bash
+# Single remote prover
+MIDEN_MONITORING_REMOTE_PROVER_URLS="http://localhost:50052" miden-network-monitoring
+
+# Multiple remote provers
+MIDEN_MONITORING_REMOTE_PROVER_URLS="http://localhost:50052,http://localhost:50053,http://localhost:50054" miden-network-monitoring
+```
```
bobbinth: Should we give an example of how this can also be used with `.env` file?

bin/network-monitoring/src/frontend.rs
```diff
@@ -0,0 +1,76 @@
...
+        .route("/assets/favicon.ico", get(serve_favicon))
+        // Main dashboard route
+        .route("/", get(get_dashboard))
+        // API route for status data
+        .route("/status", get(get_status))
+        .with_state(shared_status);
+
+    let bind_address = format!("0.0.0.0:{}", config.port);
+    println!("Starting web server on {bind_address}");
+    println!("Dashboard available at: http://localhost:{}/", config.port);
```
sergerad: Did we want to use the same tracing setup we use everywhere else?

Makefile
```diff
@@ -102,6 +102,10 @@ install-remote-prover: ## Install remote prover's CLI
 install-stress-test: ## Installs stress-test binary
 	cargo install --path bin/stress-test --locked
 
+.PHONY: install-network-monitoring
+install-network-monitoring: ## Installs network monitoring binary
+	cargo install --path bin/network-monitoring --locked
```
sergerad: We will need to add a Debian package in order to deploy this thing.

bin/network-monitoring/README.md
sergerad: We have example .envs in `bin/node/` and `bin/remote-prover/`. Would need the same for this

bin/network-monitoring/assets/background.png
TomasArrachea, line 1: Is this being used?

SantiagoPittella: No, i did not used it. Just removed it. thx!

bin/network-monitoring/src/frontend.rs
SantiagoPittella: I can do it easily by importing it from `miden_node_utils`. Should I always mark opentelemetry as disabled? Not sure if we want that here.

bin/network-monitor/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+repository.workspace   = true
+rust-version.workspace = true
+version.workspace      = true
+
+[lints]
+workspace = true
+
+[dependencies]
+anyhow           = { workspace = true }
+axum             = { version = "0.8.4" }
```
bobbinth: nit: just `0.8` for version should be enough.

bin/network-monitor/src/status.rs
bobbinth, line 1: Could we add some sections separators to this file to more clearly separate some of the logical components?

Also, would be great to add some comments.

bobbinth (approved):
Looks good! Than you! Again not a very deep review from me, but i left a couple of small comments inline.

bin/network-monitoring/src/frontend.rs
sergerad: I would want it enabled so that we can keep track of it via Honeycomb too

bobbinth: I'm actually not sure the monitor should be tied in to Honeycomb - or even have OTEL enabled at all. In my mind it is a simple binary that we run purely as an external user (and anyone can run it as well). So, the simpler we can make it, the better.

bin/network-monitor/src/status.rs
```diff
@@ -0,0 +1,327 @@
...
+async fn check_rpc_status(
+    rpc: &mut miden_node_proto::clients::RpcClient,
+    current_time: u64,
+) -> ServiceStatus {
+    match rpc.status(()).await {
+        Ok(response) => {
+            let status = response.into_inner();
+
+            // Use From implementation to convert gRPC status to domain type
+            let rpc_details = RpcStatusDetails::from(status);
```
sergerad: nit: Would just do `rpc_status: Some(status.into()),` below

bin/network-monitor/src/status.rs
```diff
@@ -0,0 +1,327 @@
...
+            let status = response.into_inner();
+
+            // Use the new method to convert gRPC status to domain type
+            let remote_prover_details =
+                RemoteProverStatusDetails::from_proxy_status(status, name.clone(), url);
+
+            // Determine overall health based on worker statuses
+            let overall_health = if remote_prover_details.workers.is_empty() {
+                "unknown"
+            } else if remote_prover_details.workers.iter().any(|w| w.status == "HEALTHY") {
```
sergerad, line 339: Should this be any or all?

bin/network-monitor/src/status.rs
```diff
@@ -0,0 +1,327 @@
...
+                .without_metadata_genesis()
+                .connect_lazy::<RemoteProverProxy>();
+            (name, url_str, client)
+        })
+        .collect();
+
+    loop {
+        let current_time = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .context("Failed to get current time")?
```
sergerad: nit: lowercase error strings

bin/network-monitor/src/status.rs
```diff
@@ -0,0 +1,327 @@
...
+        }
+
+        // Update shared status
+        {
+            let mut status = shared_status.lock().await;
+            status.services = services;
+            status.last_updated = current_time;
+        }
+
+        tokio::time::sleep(Duration::from_secs(3)).await;
```
sergerad: Might be more expressive to use an interval/ticker instead

packaging/network-monitor/miden-network-monitor.service
```diff
@@ -0,0 +1,15 @@
...
+Wants=network-online.target
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Type=exec
+EnvironmentFile=/lib/systemd/system/miden-network-monitor.env
+ExecStart=/usr/bin/miden-network-monitor
+WorkingDirectory=/opt/miden-network-monitor
```
sergerad, line 12: @bobbinth would we consider calling it just `miden-monitor`? Just for brevity's sake.

bin/network-monitor/src/status.rs
```diff
@@ -0,0 +1,327 @@
...
+            .unwrap_or_else(|_| "http://localhost:50052".to_string());
+
+        let remote_prover_urls = remote_prover_urls
+            .split(',')
+            .map(str::trim)
+            .filter(|url| !url.is_empty())
+            .map(Url::parse)
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let port = std::env::var("MIDEN_MONITOR_PORT")
```
sergerad: Same as above - const?

bin/network-monitor/src/status.rs
```diff
@@ -0,0 +1,327 @@
...
+    pub port: u16,
+}
+
+impl MonitorConfig {
+    pub fn from_env() -> Result<Self, Box<dyn std::error::Error>> {
+        let rpc_url = std::env::var("MIDEN_MONITOR_RPC_URL")
+            .unwrap_or_else(|_| "http://localhost:50051".to_string());
+
+        // Parse multiple remote prover URLs from environment variable
+        let remote_prover_urls = std::env::var("MIDEN_MONITOR_REMOTE_PROVER_URLS")
```
sergerad: Same as above - const?

bin/network-monitor/src/status.rs
```diff
@@ -0,0 +1,327 @@
...
+#[derive(Debug, Clone)]
+pub struct MonitorConfig {
+    pub rpc_url: Url,
+    pub remote_prover_urls: Vec<Url>,
+    pub port: u16,
+}
+
+impl MonitorConfig {
+    pub fn from_env() -> Result<Self, Box<dyn std::error::Error>> {
+        let rpc_url = std::env::var("MIDEN_MONITOR_RPC_URL")
```
sergerad: Should MIDEN_MONITOR_RPC_URL be a const somewhere else?

bin/network-monitor/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+            id,
+            Err::<(), anyhow::Error>(anyhow::anyhow!("Component completed unexpectedly")),
+        ), // SAFETY: The JoinSet is definitely not empty.
+        Ok((id, Err(err))) => (id, Err(err)), // SAFETY: The JoinSet is definitely not empty.
+        Err(join_err) => (join_err.id(), Err(join_err).context("Joining component task")),
+    };
+    let component = component_ids.get(&id).unwrap_or(&"unknown");
+
+    // Exit with error context
+    err.context(format!("Component {component} failed"))
```
sergerad: nit: lowercase error strings in this file

bin/network-monitor/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+    // Spawn status monitoring task
+    let status_status = shared_status.clone();
+    let status_config = config.clone();
+    let id = join_set
+        .spawn(async move { check_status(status_status, status_config).await })
+        .id();
+    component_ids.insert(id, "status");
+
+    // Wait for any task to complete or fail
+    let component_result = join_set.join_next_with_id().await.unwrap();
```
sergerad: nit: prefer to expect() over unwrap() just to give context on intent/expectations from implementor

bin/network-monitor/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    // Load configuration from environment variables
+    let config = match MonitorConfig::from_env() {
+        Ok(config) => {
+            info!("Loaded configuration: {:?}", config);
+            config
+        },
+        Err(e) => {
+            error!("Failed to load configuration: {e}");
+            std::process::exit(1);
```
sergerad: More idiomatic to `anyhow::bail!()` here?

packaging/network-monitor/miden-network-monitor.service
bobbinth: I think this could be fine too, but I do think `miden-network-monitor` is a bit better as it is more descriptive. Not a strong preference though.

bin/network-monitor/src/status.rs
SantiagoPittella: While at least one worker is up, the proxy should be able to continue processing requests

bin/network-monitor/src/main.rs
SantiagoPittella: Sure!

packaging/network-monitor/miden-network-monitor.service
SantiagoPittella: I do like `midne-network-monitor` a bit more too, but as Bobbin it is not a strong opinion either

bobbinth:
@SantiagoPittella, @Mirko-von-Leipzig, @sergerad - is this mostly done? or is there anything outstanding?

SantiagoPittella:
> @SantiagoPittella, @Mirko-von-Leipzig, @sergerad - is this mostly done? or is there anything outstanding?

Yes, there is a follow up issue that I'm already addressing and has a PR open that can be merged after this one. All comments were addressed here.

Mirko-von-Leipzig (approved):

bobbinth (approved):
Not a deep review from me - but looks good! Thank you!

@Mirko-von-Leipzig, @sergerad - we should probably deploy this on a small instance (e.g., 1 - 2 core). Maybe one deployment for testnet and one for devenet? It would be cool if this was accessible under something like `status.testnet.miden.io` and `status.devnet.miden.io`.

Another cool thing for a future PR would be to include some additional data for block producer (and maybe NTX builder). For example, something like number of tx in the mempool. Let's create an issue for this (unless we already have one).

Mirko-von-Leipzig:
> Another cool thing for a future PR would be to include some additional data for block producer (and maybe NTX builder). For example, something like number of tx in the mempool. Let's create an issue for this (unless we already have one).

We do already track this via the `mempool`'s instrumentation, so its query-able in honeycomb

image

But of course this isn't available publicly

PR #1221, opened 2025-09-10
fix(rpc): accept `application/grpc` content type

Mirko-von-Leipzig:
> [!IMPORTANT] 
> PR is against main

We didn't accept `application/grpc` as a content type because we expect the more specific `application/vnd.miden` format. 

The former is likely the default used by gRPC clients and we should therefore allow it. The counter argument is that if the clients are just using the defaults then they're also not setting the version or commitment settings - which we might want to require at some point in the future.

I've created this PR in case we do want to go ahead with this fix as it has come up by some clients.

cc @igamigo -- do you think we need/want this?


bobbinth:
So, right now if someone tries to use the gRPC APIs directly, and they don't set content type to `application/vnd.miden`, the request will fail, right?

As you said, if they are doing this, they are probably also not setting the version or the genesis commitment - which is probably not the best practice. So, my initial thinking is that the current state of things is OK - but we should make sure we have this documented thoroughly somewhere in the mdBook (and maybe we already have this).

PR #1224, opened 2025-09-10
feat: add SyncTransaction endpoint

### LINKED ISSUE #1207: Implement `SyncTransactions` endpoint
As discussed in https://github.com/0xMiden/miden-node/discussions/722#discussioncomment-14270330, it would be good to create an endpoint for syncing transaction data. The request for the endpoint could look something like this:
```proto
message SyncTransactionsRequest {
    fixed32 block_from = 1;
    repeated AccountId account_ids = 3;
}
```
And the response could look something like this:
```proto
message SyncTransactionsResponse {
    fixed32 chain_tip = 1;
    fixed32 block_num = 2;
    repeated TransactionHeader = 3;
}

// Maybe use some other name not to confuse with TransactionHeader struct from miden-base
message TranactionHeader {
    fixed32 block_num = 1;
    AccountId accountId = 2;
    Digest initial_state_commitment = 3;
    Digest final_state_commitment = 4;
    repeated Nullifier input_notes = 5;
    repeated NoteSyncRecord output_notes = 6;
}
```

One open question is how to limit the size of this response. In the worst case, a transaction header could be over 500KB (assuming a transaction consumes 1K notes and creates 1K notes) - so, only a few of these would fit into 2MB - 3MB limit. This is probably not a problem for now as we don't yet allow such large transactions, but if there is an easy way to address this, we'd be avoiding a potential breaking change in the future.

SantiagoPittella:
closes #1207

### DIFF
--- a/crates/store/src/db/models/queries/transactions.rs
+++ b/crates/store/src/db/models/queries/transactions.rs
@@ -329,11 +329,6 @@ pub fn select_transactions_records(
             .load::<TransactionRecordRaw>(conn)
             .map_err(DatabaseError::from)?;
 
-        // If no more data, we're done
-        if chunk.is_empty() {
-            break;
-        }
-
         // Add transactions from this chunk one by one until we hit the limit
         let mut added_from_chunk = 0;
         let mut last_added_tx: Option<TransactionRecordRaw> = None;


proto/proto/store/rpc.proto
```diff
@@ -435,3 +438,31 @@ message PaginationInfo {
...
+}
+
+message TransactionHeader {
+    fixed32 block_num = 1;
+    account.AccountId account_id = 2;
+    primitives.Digest initial_state_commitment = 3;
+    primitives.Digest final_state_commitment = 4;
+    repeated primitives.Digest input_notes = 5;
+    repeated note.NoteSyncRecord output_notes = 6;
+}
```
bobbinth, line 459: Let's add more doc comments here.

crates/store/src/server/rpc_api.rs
```diff
@@ -504,4 +506,112 @@ impl rpc_server::Rpc for StoreApi {
...
+
+        let transaction_summaries = self
+            .state
+            .sync_transactions(account_ids, block_range.clone())
+            .await
+            .map_err(internal_error)?;
+
+        // Convert TransactionSummary to TransactionHeader proto by loading full block data
+        // from BlockStore to get complete transaction header information
+        let mut transaction_headers = Vec::new();
```
bobbinth: I don't think loading the full block is a viable solution here. Instead, we should store this data in the `transactions` table. The table should basically be sufficient to contain `TranactionHeader` messages, except instead of full note sync records, we'd store note IDs for `output_notes`.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -137,3 +138,34 @@ impl TransactionSummaryRowInsert {
...
+/// A vector of [`TransactionSummary`] or an error.
+///
+/// # Note
+/// This function returns basic transaction information since the database only stores
+/// `transaction_id`, `account_id`, and `block_num`.
+pub fn select_transactions_headers(
+    conn: &mut SqliteConnection,
+    account_ids: &[AccountId],
+    block_range: RangeInclusive<BlockNumber>,
+) -> Result<Vec<TransactionSummary>, DatabaseError> {
```
bobbinth: Let's add the SQL statement to the doc comments (we started doing this for other queries).

Also, how are we thinking of limiting the response size here? (and for the entire response in general).

SantiagoPittella: I addressed the limitation the same way as the other paginated endpoints

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -119,9 +119,13 @@ CREATE INDEX idx_nullifiers_prefix ON nullifiers(nullifier_prefix);
...
-    transaction_id BLOB    NOT NULL,
-    account_id     BLOB    NOT NULL,
-    block_num      INTEGER NOT NULL,
+    transaction_id               BLOB    NOT NULL,
+    account_id                   BLOB    NOT NULL,
+    block_num                    INTEGER NOT NULL,
+    initial_state_commitment     BLOB    NOT NULL,
+    final_state_commitment       BLOB    NOT NULL,
+    input_notes                  BLOB    NOT NULL,
+    output_notes                 BLOB    NOT NULL,
```
bobbinth: I would add comments here explaining that input notes are a list of nullifiers and output notes are a list of note IDs.

proto/proto/store/rpc.proto
```diff
@@ -435,3 +438,47 @@ message PaginationInfo {
...
+    account.AccountId account_id = 2;
+
+    // State commitment of the account before the transaction was executed.
+    primitives.Digest initial_state_commitment = 3;
+
+    // State commitment of the account after the transaction was executed.
+    primitives.Digest final_state_commitment = 4;
+
+    // Input notes of the transaction.
+    repeated primitives.Digest input_notes = 5;
```
bobbinth: I would mention that these are nullifiers.

proto/proto/store/rpc.proto
```diff
@@ -435,3 +438,47 @@ message PaginationInfo {
...
+    repeated TransactionHeader transaction_headers = 2;
+}
+
+// Represents a transaction header.
+message TransactionHeader {
+    // Block number in which the transaction was executed.
+    fixed32 block_num = 1;
+
+    // Account ID of the transaction.
+    account.AccountId account_id = 2;
```
bobbinth: nit: I would change the comment to "ID of the account against which the transaction was executed."

crates/store/src/state.rs
```diff
@@ -1047,6 +1047,24 @@ impl State {
...
+        self.db.select_transactions_headers(account_ids, block_range).await
+    }
+
+    /// Loads block data from the block store.
+    pub async fn load_block_data(
+        &self,
+        block_num: BlockNumber,
+    ) -> Result<Option<Vec<u8>>, std::io::Error> {
+        self.block_store.load_block(block_num).await
+    }
```
bobbinth: Do we still need this method?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +167,131 @@ pub struct TransactionSummaryRowInsert {
...
+/// LIMIT ?4
+/// ```
+pub fn select_transactions_headers(
+    conn: &mut SqliteConnection,
+    account_ids: &[AccountId],
+    block_range: RangeInclusive<BlockNumber>,
+) -> Result<(BlockNumber, Vec<crate::db::TransactionHeader>), DatabaseError> {
+    const MAX_PAYLOAD_BYTES: usize = 5 * 1024 * 1024; // 5 MB
+    const ROW_OVERHEAD_BYTES: usize = 500 * 1024; // 500 KB per transaction header
+    const MAX_ROWS: usize = MAX_PAYLOAD_BYTES / ROW_OVERHEAD_BYTES; // ~10 transactions
```
bobbinth: I don't think this is correct. The size of the transaction header depends on the number of output notes. Each output note contributes at most 580 bytes - though, usually this will be much less (probably under 300 bytes). To be safe though, we could estimate 500 bytes per output note. So, in theory, a single response could include up to 10000 output notes (though, input notes also cost a little).

Basically, the methodology here should be more sophisticated and it should take both input and output notes into account.

crates/store/src/state.rs
SantiagoPittella: No, we do not. I'm deleting it.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: I followed another approach based on a talk with @igamigo , consist of storing the size in bytes of the entry in a new column. This way I can check the size of the items to use in the response by the given size limit (5MB). If the limit is reach, the whole "block" is removed. And we repeat the process until the response size is below the limit.

SantiagoPittella: Also, I'm working on the removal of the N+1 queries on the output notes NoteSyncRecords. Not very familiar with how to do so with diesel.

SantiagoPittella: About this last part, since the note ids of the output notes are stored serialized i'm not able to do a join to return the full transaction summaries in one single query. In order to do so, a larger change in the store is needed.
Another option is to keep it as it is in the `sync_state` endpoint, which is composed by a transaction summary and a vector with `NoteSyncRecord`, separately. This way you delegate the "joining" to the user, but remove the needed for N+1 queries since we can flatten all the transaction header's output notes vector and make just one query to the DB.

SantiagoPittella: Regarding the N+1 queries and the pagination. What about adding a new endpoint like `SyncOutputNotes` or similar, taking `NoteIDs` and returning the `NoteSyncRecord`? Basically, it makes the each transaction header form the `SyncTransactions` much lighter, and removes the need of N+1 queries in that same endpoint.

crates/store/src/db/mod.rs
```diff
@@ -92,6 +92,43 @@ pub struct TransactionSummary {
...
+#[derive(Debug, PartialEq)]
+pub struct TransactionHeader {
+    pub account_id: AccountId,
+    pub block_num: BlockNumber,
+    pub transaction_id: TransactionId,
+    pub initial_state_commitment: Word,
+    pub final_state_commitment: Word,
+    pub input_notes: Vec<Nullifier>, // Store nullifiers for input notes
+    pub output_notes: Vec<NoteId>,   // Store note IDs for output notes
+}
```
bobbinth: nit: I would arrange the fields as follows:
```rust
pub struct TransactionHeader {
    pub block_num: BlockNumber,
    pub transaction_id: TransactionId,
    pub account_id: AccountId,
    pub initial_state_commitment: Word,
    pub final_state_commitment: Word,
    pub input_notes: Vec<Nullifier>,
    pub output_notes: Vec<NoteId>,
}
```
Also, maybe we should rename this into something like `TransactionRecord` or at least add a doc comment explaining that this is basically the same as `TransactionHeader` from `miden-base`, except here we also have `block_num`.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,129 @@ pub struct TransactionSummaryRowInsert {
...
+///     block_num,
+///     transaction_id,
+///     initial_state_commitment,
+///     final_state_commitment,
+///     input_notes,
+///     output_notes
+/// FROM transactions
+/// WHERE block_num > ?1 AND block_num <= ?2 AND account_id IN rarray(?3)
+/// ORDER BY block_num ASC
+/// LIMIT ?4
```
bobbinth: I think this statement may need to be updated.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,129 @@ pub struct TransactionSummaryRowInsert {
...
+        // Serialize input notes using binary format (store nullifiers)
+        let input_notes_binary = transaction_header.input_notes().to_bytes();
+
+        // Serialize output notes using binary format (store note IDs)
+        let output_notes_binary = transaction_header.output_notes().to_bytes();
+
+        // Store the size required to store the row + the size that the output notes will take when
+        // casted to NoteSyncRecord (~500B per output note).
+        let size_in_bytes = (transaction_header.to_bytes().len()
+            + (transaction_header.output_notes().len() * 500)) as i64;
```
bobbinth: Converting `transaction_header` here to bytes seems a bit wasteful (and I don't think the compiler will be able to optimize this way). Unfortunately, we don't have the `get_size_hint()` function implemented on the `TransactionHeader`. What we could do is manually estimate the size of the header which is:

- 4 bytes for block number.
- 32 bytes for transaction ID.
- 16 bytes for account ID.
- 32 + 32 bytes for initial + final commitments.
- 32 * `num_input_notes` for the input notes.
- 500 * `num_output_notes` for the output notes.

As I mentioned, 500 bytes is a bit of an over-estimate, but it is fine for now. We should explain all of this in the doc comments though.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,129 @@ pub struct TransactionSummaryRowInsert {
...
+    }
+
+    let desired_account_ids = serialize_vec(account_ids);
+    let raw = SelectDsl::select(schema::transactions::table, TransactionHeaderRaw::as_select())
+        .filter(schema::transactions::block_num.gt(block_range.start().to_raw_sql()))
+        .filter(schema::transactions::block_num.le(block_range.end().to_raw_sql()))
+        .filter(schema::transactions::account_id.eq_any(desired_account_ids))
+        .order((schema::transactions::block_num.asc(),))
+        .load::<TransactionHeaderRaw>(conn)
+        .map_err(DatabaseError::from)?;
```
bobbinth: This will work, but I'm concerned that it may return a lot of extra data and may become an attack vector. At the very least, we should put a `LIMIT` here. A simple way to do it would be to put something like 10,000 as the limit as we know 10,000 transactions are unlikely to fit into the request (assuming each one of them has at least 1 output note). This still may return too much data (because some transactions may contain multiple output notes).

A better approach may be to read data from the database in "chunks" - e.g., 1000 rows at a time (or maybe even fewer) and stop once the next chunk exceeds the total accumulated limit.

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -119,9 +119,14 @@ CREATE INDEX idx_nullifiers_prefix ON nullifiers(nullifier_prefix);
...
-    account_id     BLOB    NOT NULL,
-    block_num      INTEGER NOT NULL,
+    transaction_id               BLOB    NOT NULL,
+    account_id                   BLOB    NOT NULL,
+    block_num                    INTEGER NOT NULL, -- Block number in which the transaction was included.
+    initial_state_commitment     BLOB    NOT NULL, -- State of the account before applying the transaction.
+    final_state_commitment       BLOB    NOT NULL, -- State of the account after applying the transaction.
+    input_notes                  BLOB    NOT NULL, -- Serialized vector with the Nullifier of the input notes.
+    output_notes                 BLOB    NOT NULL, -- Serialized vector with the NoteId of the output notes.
+    size_in_bytes                INTEGER NOT NULL, -- Size of the row in bytes.
```
bobbinth: I would explain that this is an estimated size taking into account size of output note records.

bobbinth (commented):
Looks good! Thank you. I left some more comments inline.

crates/store/src/db/models/queries/transactions.rs
bobbinth: I think doing a query per transaction to get `NoteSyncRecord`s is fine for now (assuming we do this only for transactions which have output notes).

> What about adding a new endpoint like `SyncOutputNotes` or similar, taking `NoteIDs` and returning the `NoteSyncRecord`? Basically, it makes the each transaction header form the `SyncTransactions` much lighter, and removes the need of N+1 queries in that same endpoint.

I'm not sure this will make things better. The client would still need to get this data, and then we'd be turning local memory lookups into network RPC calls - which will be way worse.

crates/store/src/db/mod.rs
SantiagoPittella: I think that `TransactionRecord` fits better 👍🏼

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: Done!

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,191 @@ pub struct TransactionSummaryRowInsert {
...
+///     initial_state_commitment,
+///     final_state_commitment,
+///     input_notes,
+///     output_notes,
+///     size_in_bytes
+/// FROM transactions
+/// WHERE block_num > ?1 AND block_num <= ?2 AND account_id IN rarray(?3)
+/// ORDER BY block_num ASC
+/// LIMIT ?4 OFFSET ?5
+/// ```
```
bobbinth, line 269: nit: I would format this as:
```rust
/// SELECT
///     account_id,
///     block_num,
///     transaction_id,
///     initial_state_commitment,
///     final_state_commitment,
///     input_notes,
///     output_notes,
///     size_in_bytes
/// FROM
///     transactions
/// WHERE
///     block_num > ?1
///     AND block_num <= ?2
///     AND account_id IN rarray(?3)
/// ORDER BY
///     block_num ASC
/// LIMIT
///     ?4
/// OFFSET
///     ?5
```

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,191 @@ pub struct TransactionSummaryRowInsert {
...
+/// Note: The query is executed in chunks of 1000 transactions to prevent loading
+/// excessive data and to stop as soon as the accumulated size approaches the 5MB limit.
+pub fn select_transactions_records(
+    conn: &mut SqliteConnection,
+    account_ids: &[AccountId],
+    block_range: RangeInclusive<BlockNumber>,
+) -> Result<(BlockNumber, Vec<crate::db::TransactionRecord>), DatabaseError> {
+    const MAX_PAYLOAD_BYTES: usize = 5 * 1024 * 1024; // 5 MB
```
bobbinth: Let's reduce this to 4MB to have a buffer before we get to 5MB.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,191 @@ pub struct TransactionSummaryRowInsert {
...
+    const CHUNK_SIZE: i64 = 1000; // Read 1000 transactions at a time
```
bobbinth: I would maybe mention that given the size of the note records, 1000 records are guaranteed never to return more than about 60MB of data.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,191 @@ pub struct TransactionSummaryRowInsert {
...
+            for tx in chunk {
+                if current_size + tx.size_in_bytes <= MAX_PAYLOAD_BYTES as i64 {
+                    current_size += tx.size_in_bytes;
+                    all_transactions.push(tx);
+                } else {
+                    // Can't fit this transaction, stop here
+                    break;
+                }
+            }
+            break;
```
bobbinth, line 356: We should never reach this `break`, right? Would it make sense to change it to `unreachable!()` or an error?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,191 @@ pub struct TransactionSummaryRowInsert {
...
+        }
```
bobbinth: Maybe for simplicity we always add transactions one-by-one. It will be a bit less efficient, but I think the difference should be pretty small. Basically, the code could be something like:
```rust
// Add transactions from this chunk one by one until we hit the limit
for tx in chunk {
    if current_size + tx.size_in_bytes <= MAX_PAYLOAD_BYTES as i64 {
        current_size += tx.size_in_bytes;
        all_transactions.push(tx);
    } else {
        // Can't fit this transaction, stop here
        break;
    }
}
```
And then we don't need the code on lines 321 - 330 below.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,191 @@ pub struct TransactionSummaryRowInsert {
...
+        if chunk_len < CHUNK_SIZE {
+            break;
         }
     }
+
+    // Ensure block consistency: remove complete blocks from the end until we're under the size
+    // limit
+    let mut filtered_raw = all_transactions;
+    while !filtered_raw.is_empty() {
+        let total_size: i64 = filtered_raw.iter().map(|tx| tx.size_in_bytes).sum();
```
bobbinth: Would this `total_size` be different from the one we calculated above as `current_size`?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,191 @@ pub struct TransactionSummaryRowInsert {
...
+        }
+    }
+
+    // Convert to transaction records
+    let headers: Vec<crate::db::TransactionRecord> = filtered_raw
+        .into_iter()
+        .map(std::convert::TryInto::try_into)
+        .collect::<Result<Vec<_>, _>>()?;
+
+    let last_included_block = headers.last().map_or(*block_range.start(), |tx| tx.block_num);
```
bobbinth: Can we not use the same approach for filtering out the last block as we did for other sync endpoints (e.g., sync storage map and sync asset vault - where we pass over the result set only once)?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,191 @@ pub struct TransactionSummaryRowInsert {
...
+
+    loop {
+        let chunk =
+            SelectDsl::select(schema::transactions::table, TransactionRecordRaw::as_select())
+                .filter(schema::transactions::block_num.gt(block_range.start().to_raw_sql()))
+                .filter(schema::transactions::block_num.le(block_range.end().to_raw_sql()))
+                .filter(schema::transactions::account_id.eq_any(&desired_account_ids))
+                .order((schema::transactions::block_num.asc(),))
+                .offset(offset)
+                .limit(CHUNK_SIZE)
```
bobbinth: Since we are using limit + offset here, there a few things to consider:

1. We need to make sure that all queries happen in the same transaction (I think that's already the case).
2. I think to ensure stable order, we also need to sort by transaction ID (i.e., just sorting by block number may be insufficient).
3. From what I know, `offset` may not be all that efficient in SQLite (cc @Mirko-von-Leipzig as he may know more on this) - and I wonder if there is a better way to structure the query that will rely on a more explicit condition.

bobbinth (commented):
Looks good! Thank you. I left some more comments inline.

bobbinth:
One more general comment: there is a much more efficient way to serialize note inclusion paths which would guarantee that we fit into 5MB limit even if a block contained $2^{16}$ output notes. It is quite a bit more involved though - instead of sending a Merkle path with every node, we'd need to send a single partial SMT for an entire block. So, it is not something we should do now, but let's create an issue for this.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: That `break` is reachable, after discarding the tx that exceeds the size limit we want to break the main loop and that's what that `break` is doing

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: 1. The call is wrapped in a DB transaction:
```rust
        self.transact("full transactions records", move |conn| {
            queries::select_transactions_records(conn, &account_ids, block_range)
        })
        .await
```
2. I added the order by transaction ID too.
3. For this point, maybe we can go with some cursors using the block numbers & transaction ID instead of the `OFFSET`, but I'm not entirely sure what's best on SQLite.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: The same as in requesting something like `MAX_ROWS + 1` and then checking if `result[last].block_num != result[last - 1].block_num`, I didn't take that path because MAX_ROWS is different to calculate precisely due to the notes variability.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,206 @@ pub struct TransactionSummaryRowInsert {
...
+///   stop as soon as the accumulated size approaches the 4MB limit.
+/// - Given the size of note records, 1000 records are guaranteed never to return more than about
+///   60MB of data.
+pub fn select_transactions_records(
+    conn: &mut SqliteConnection,
+    account_ids: &[AccountId],
+    block_range: RangeInclusive<BlockNumber>,
+) -> Result<(BlockNumber, Vec<crate::db::TransactionRecord>), DatabaseError> {
+    const MAX_PAYLOAD_BYTES: usize = 4 * 1024 * 1024; // 4 MB
+    const CHUNK_SIZE: i64 = 1000; // Read 1000 transactions at a time
```
bobbinth: nit: I would probably rename `CHUNK_SIZE` into `NUM_TXS_PER_CHUNK` to make it clear that this is number of transactions rather than number of bytes.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,206 @@ pub struct TransactionSummaryRowInsert {
...
+        if added_from_chunk < chunk_len {
+            break;
+        }
+
+        offset += CHUNK_SIZE;
+
+        // If we got fewer transactions than requested, we've reached the end
+        if chunk_len < CHUNK_SIZE {
+            break;
+        }
```
bobbinth, line 357: Could this code be written like so?

```rust
// if we've added fewer transactions than in the full chunk, then we've either
// hit the size limit or we've loaded all the data from the database
if added_from_chunk < CHUNK_SIZE {
    break;
}

offset += CHUNK_SIZE;
```

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,206 @@ pub struct TransactionSummaryRowInsert {
...
+                .map(|tx| tx.size_in_bytes)
+                .sum();
+
+            filtered_raw.retain(|tx| tx.block_num != last_block_num);
+            total_size -= removed_size;
+        } else {
+            // This shouldn't happen since we check !filtered_raw.is_empty() above, but be safe
+            break;
+        }
+    }
```
bobbinth, line 376: Does this need to be a loop? My understanding is that if we "overshot", we've done so only by 1 transaction - and so, we'd need to remove only the last block.

crates/store/src/db/models/queries/transactions.rs
bobbinth: I left a comment related to this, but I think we just need to remove the last block because we can never "overshoot" by more than one transaction, right? If so, we should be able to apply the same methodology here too.

crates/store/src/db/models/queries/transactions.rs
bobbinth: I think @Mirko-von-Leipzig looked into using `OFFSET` in other context and decided that it was not a good idea. We could use transaction ID for filtering - but I wonder if we can also access the row ID somehow explicitly.

Mirko-von-Leipzig: Apologies, I missed the initial ping. Using `OFFSET` to paginate usually results in poor performance because it is only applied at the end of the query. 

Put differently, it always builds the full query result set, and then truncates it. It is almost always better to use some dedicated ID to first restrict the search window using the `WHERE` clauses. As an example:

```sql
-- This would first sort the entire table (bad), then skip 10 rows and take 5.
SELECT * FROM table ORDER BY column OFFSET 10 LIMIT 5;

-- This first constrains the result set and can use an index to count, no table scan required.
SELECT * FROM table WHERE column > 123 ORDER BY column LIMIT 5;
```

In this specific case we're already applying some `WHERE` clauses, so `OFFSET` could be okay if we are certain that the search bounds will always be constrained enough that its not a large set. As in, the block range + account ID result set must be small, which I think it may be. If figuring out a better way is difficult, then it might be okay as is for now. cc @drahnr just to double check me here.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: We can avoid the loop by checking the size of the result first

SantiagoPittella: Did so in the last commit.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +173,194 @@ pub struct TransactionSummaryRowInsert {
...
+        }
+
+        offset += NUM_TXS_PER_CHUNK;
+    }
+
+    // Ensure block consistency: remove the last block if it's incomplete
+    // (we may have stopped loading mid-block due to size constraints)
+    let mut filtered_raw = all_transactions;
+
+    if filtered_raw.len() > 1 {
```
bobbinth: Is this the correct condition? We should remove the last block only if `total_size >= MAX_PAYLOAD_BYTES` - or am I missing something?

SantiagoPittella: You're right. It sohuld be done only in that scenario. I'm fixing it.

bobbinth:
@SantiagoPittella - what's outstanding in this PR? I think maybe just https://github.com/0xMiden/miden-node/pull/1224#discussion_r2365281064 is left?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,211 @@ pub struct TransactionSummaryRowInsert {
...
+    if total_size >= MAX_PAYLOAD_BYTES {
+        let last_idx = filtered_raw.len() - 1;
+        let last_block_num = filtered_raw[last_idx].block_num;
+        let prev_block_num = filtered_raw[last_idx - 1].block_num;
+
+        // If the last transaction is from the same block as the previous one,
+        // we may have stopped mid-block, so remove the entire last block
+        if last_block_num == prev_block_num {
+            filtered_raw.retain(|tx| tx.block_num != last_block_num);
+        }
```
bobbinth: I'm not quite following this logic: why do we filter out the last block only if last block number is the same as the second to last block number? Shouldn't we do this regardless?

Also, could we not use the same general approach for filtering out the last block as we did in other places (e.g., for sync nullifiers) - e.g., something like:

```rust
if total_size >= MAX_PAYLOAD_BYTES {
    // TODO: filter out the last block
} else {
    Ok((vec_raw_try_into(filtered_raw)?, *block_range.end()))
}
```

Or would this not work for some reason?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,211 @@ pub struct TransactionSummaryRowInsert {
...
+    }
+
+    // Ensure block consistency: remove the last block if it's incomplete
+    // (we may have stopped loading mid-block due to size constraints)
+    let mut filtered_raw = all_transactions;
+
+    if total_size >= MAX_PAYLOAD_BYTES {
+        let last_idx = filtered_raw.len() - 1;
+        let last_block_num = filtered_raw[last_idx].block_num;
```
bobbinth: nit: I would add a safety note saying that we are guaranteed to have at least 1 transaction in the vector because we know that total size is greater than zero.

bobbinth (commented):
Looks good! Thank you. I left a couple more comments inline. The main one is that I'm not sure the logic for filtering out the last block is correct.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: To be more clear, I ended up moving an approach more similar to the other paginated queries. With the difference than in the other we set a maximum number of rows, but here it is variable with the size. This way we ensure that the payload is always `< MAX_PAYLOAD_BYTES`.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: This was removed with the new approach mentioned in above's comment.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,203 @@ pub struct TransactionSummaryRowInsert {
...
+            break;
         }
     }
+
+    // Ensure block consistency: remove the last block if it's incomplete
+    // (we may have stopped loading mid-block due to size constraints)
+    let mut filtered_raw = all_transactions;
+
+    if total_size >= MAX_PAYLOAD_BYTES {
+        if let Some(last_tx) = filtered_raw.last() {
```
bobbinth: Could we not use the `last_block_num` variable here (the one declared on line 301 above). Also, we should be guaranteed to have at least one transaction - right? So, we could probably use `.expect()` here.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,203 @@ pub struct TransactionSummaryRowInsert {
...
+            let last_block_num = last_tx.block_num;
+            filtered_raw.retain(|tx| tx.block_num != last_block_num);
+        }
+    }
+
+    // Convert to transaction records
+    let headers: Vec<crate::db::TransactionRecord> = vec_raw_try_into(filtered_raw)?;
+
+    let last_included_block = headers.last().map_or(*block_range.end(), |tx| tx.block_num);
```
bobbinth: Same question as above: can't we use `last_block_num` here?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,203 @@ pub struct TransactionSummaryRowInsert {
...
+
+    Ok((last_included_block, headers))
```
bobbinth: Related to the previous comments - can we not simplify this block to look something like:
```rust
if total_size >= MAX_PAYLOAD_BYTES {
    let last_block_num = last_block_num.expect("todo");
    let filtered_transactions = vec_raw_try_into(
        all_transactions.into_iter().take_while(|row| row.block_num != last_block_num),
    )?;

    let last_included_block = BlockNumber::from_raw_sql(last_block_num.saturating_sub(1))?;
    Ok((last_included_block, filtered_transactions))
} else {
    let last_included_block = (last_block_num.unwrap_or(*block_range.end()) as u32).into();
    Ok((last_included_block, vec_raw_try_into(all_transactions)?))
}
```

If this works, it also has the benefit of iterating over the result set only once in both cases.

bobbinth (commented):
Looks good! Thank you. I left a couple more comments inline.

Also, not for this PR, but it would be great to test this functionality somehow - but not really sure what's the best way to do it. Let's create an issue for this.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: Yes, in fact since the limit was reach we have multiple. I'm replacing with the expect.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: Yes, this works with some minor changes 👌🏼

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,210 @@ pub struct TransactionSummaryRowInsert {
...
+        // - 4 bytes for block number
+        // - 32 bytes for transaction ID
+        // - 16 bytes for account ID
+        // - 64 bytes for initial + final state commitments (32 bytes each)
+        // - 32 bytes per input note (nullifier size)
+        // - 500 bytes per output note (estimated size when converted to NoteSyncRecord)
+        //
+        // Note: 500 bytes per output note is an over-estimate but ensures we don't
+        // exceed memory limits when these transactions are later converted to proto records.
+        let header_base_size = 4 + 32 + 16 + 64; // block_num + tx_id + account_id + commitments
```
igamigo: This could be `const`, right?

proto/proto/store/rpc.proto
```diff
@@ -422,3 +425,47 @@ message PaginationInfo {
...
+message SyncTransactionsResponse {
+    // Pagination information.
+    PaginationInfo pagination_info = 1;
+
+    // List of transaction records.
+    repeated TransactionRecord transaction_records = 2;
+}
+
+// Represents a transaction record.
+message TransactionRecord {
```
igamigo, line 453: I think this is mostly a `TransactionHeader` but I'm guessing this wasn't named that because it also includes the `block_num`, right? Not sure it's worth refactoring but it would be nice to be able to correlate the protobuf responses with the domain struct names.

proto/proto/store/rpc.proto
```diff
@@ -422,3 +425,47 @@ message PaginationInfo {
...
+    account.AccountId account_id = 2;
+
+    // State commitment of the account before the transaction was executed.
+    primitives.Digest initial_state_commitment = 3;
+
+    // State commitment of the account after the transaction was executed.
+    primitives.Digest final_state_commitment = 4;
+
+    // Nullifiers of the input notes of the transaction.
+    repeated primitives.Digest input_notes = 5;
```
igamigo: Not related to this PR (I think this is the case for `TransactionHeader` as well), but it feels confusing that this is named `input_notes` and is of type `primitives.Digest`. There are 3 digests that could be related to an input note: the commitment, the ID and the nullifier, so to be more precise I'd name the field in the header `nullifiers` instead.

igamigo: Also, looking at blocks/batches structs they have fields called `created_nullifiers` instead of "notes". But maybe I'm missing something

crates/store/src/server/rpc_api.rs
```diff
@@ -483,4 +483,61 @@ impl rpc_server::Rpc for StoreApi {
...
+        let account_ids: Vec<AccountId> = read_account_ids(&request.account_ids)?;
+
+        let (last_block_included, transaction_records_db) = self
+            .state
+            .sync_transactions(account_ids, block_range.clone())
+            .await
+            .map_err(internal_error)?;
+
+        // Convert database TransactionRecord to proto TransactionRecord
+        let mut transaction_records = Vec::new();
```
igamigo:
```suggestion
        let mut transaction_records = Vec::with_capacity(transaction_records_db.len());
```

crates/store/src/server/rpc_api.rs
```diff
@@ -483,4 +483,61 @@ impl rpc_server::Rpc for StoreApi {
...
+
+        for tx_header in transaction_records_db {
+            // Retrieve full note data for output notes from the database
+            let note_records = self
+                .state
+                .get_notes_by_id(tx_header.output_notes.clone())
+                .await
+                .map_err(internal_error)?;
```
igamigo, line 524: This might have been discussed before (maybe for a follow up) but going to the store for every single transaction seems potentially wasteful if in a single query you can fit note responses for more than one of them. I think joining the tables is also not very desirable here but that could be another alternative.

crates/store/src/server/rpc_api.rs
```diff
@@ -483,4 +483,61 @@ impl rpc_server::Rpc for StoreApi {
...
+
+            // Convert NoteRecord to NoteSyncRecord
+            let note_sync_records: Vec<crate::db::NoteSyncRecord> =
+                note_records.into_iter().map(std::convert::Into::into).collect();
```
igamigo: I think some of the collections into `Vec<_>` could be avoided. Here we collect the `Vec<NoteRecord>` into `Vec<NoteSyncRecord>` and then in `to_proto_with_note_records` we collect again into `Vec<proto::NoteSyncRecord>`. 

Instead of collecting and passing vectors, maybe `to_proto_with_note_records` could take `IntoIterator<Item=R>` for `R: Into<NoteSyncRecord>` (or just hardcode the type to `NoteRecord`) and we would avoid the intermediate `Vec<NoteSyncRecord>`

crates/store/src/db/mod.rs
```diff
@@ -92,6 +92,43 @@ pub struct TransactionSummary {
...
+    /// Convert to proto `TransactionRecord`, but requires note sync records for output notes.
+    /// For `sync_transactions` RPC, we need to fetch note sync records separately since we only
+    /// store note IDs in the database.
+    #[allow(clippy::wrong_self_convention)]
+    pub fn to_proto_with_note_records(
+        self,
+        note_records: Vec<NoteSyncRecord>,
+    ) -> proto::rpc_store::TransactionRecord {
+        let input_notes: Vec<proto::primitives::Digest> =
+            self.input_notes.iter().map(|nullifier| (*nullifier).into()).collect();
```
igamigo: minor nit: Because this function takes `self`, this could be within the response declaration and the reference/deref could be avoided, right?

```rust
proto::rpc_store::TransactionRecord {
            // ...
            input_notes: self.input_notes
                .into_iter()
                .map(From::from)
                .collect(),
            // ...
}
```

Probably does not make too much (if any) of a difference though

crates/store/src/db/mod.rs
```diff
@@ -92,6 +92,43 @@ pub struct TransactionSummary {
...
+    pub final_state_commitment: Word,
+    pub input_notes: Vec<Nullifier>, // Store nullifiers for input notes
+    pub output_notes: Vec<NoteId>,   // Store note IDs for output notes
+}
+
+impl TransactionRecord {
+    /// Convert to proto `TransactionRecord`, but requires note sync records for output notes.
+    /// For `sync_transactions` RPC, we need to fetch note sync records separately since we only
+    /// store note IDs in the database.
+    #[allow(clippy::wrong_self_convention)]
```
igamigo: nit (but feel free to disregard if you prefer the current one): I'd name this `into_proto_with_note_records` instead, or maybe `into_proto_with_notes` if the other comment related to making this more generic ends up working out

crates/store/src/db/mod.rs
```diff
@@ -538,4 +575,21 @@ impl Db {
         })
         .await
     }
+
+    /// Returns the complete transaction records for the specified accounts within the specified
+    /// block range, including state commitments and note IDs.
+    ///
+    /// Note: This method is size-limited (~5MB) and may not return all matching transactions
```
igamigo, line 596: If/once we move these limits to be more globally-scoped constants, we should remember update all these doc comments to reference the constants instead (and also in the previous syncing endpoints if there are any references there).

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,210 @@ pub struct TransactionSummaryRowInsert {
...
         Self {
-            transaction_id: transaction_id.to_bytes(),
-            account_id: account_id.to_bytes(),
+            transaction_id: transaction_header.id().to_bytes(),
+            account_id: transaction_header.account_id().to_bytes(),
             block_num: block_num.to_raw_sql(),
+            initial_state_commitment: transaction_header.initial_state_commitment().to_bytes(),
+            final_state_commitment: transaction_header.final_state_commitment().to_bytes(),
+            input_notes: input_notes_binary,
+            output_notes: output_notes_binary,
```
igamigo, line 220: Maybe there is already reasons why this is not this way (apologies if that's the case), but because we don't need to index, nor query anything with these fields I wonder if we should avoid vector allocations here in favor of just serializing the whole `TransactionHeader`. 
Maybe premature optimization (and also maybe I'm missing something) but you'd be trading keeping a duplicate copy of `account_id` and `transaction_id` in favor of removing 3 `to_bytes()` allocations (basically these 4 minus the one you'd still have to do for `TransactionHeader`). Conversely it could improve deserialization allocations.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,210 @@ pub struct TransactionSummaryRowInsert {
...
+        }
+    }
+}
+
+/// Select complete transaction records for the given accounts and block range.
+///
+/// # Parameters
+/// * `account_ids`: List of account IDs to filter by
+///     - Limit: 0 <= size <= 1000
+/// * `block_range`: Range of blocks to include inclusive
```
igamigo, line 231: nit: "Inclusive" here sounds ambiguous because the SQL query uses "> block_num"

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,210 @@ pub struct TransactionSummaryRowInsert {
...
+        }
+
+        // Add transactions from this chunk one by one until we hit the limit
+        let mut added_from_chunk = 0;
+        for tx in chunk {
+            if total_size + tx.size_in_bytes <= MAX_PAYLOAD_BYTES {
+                total_size += tx.size_in_bytes;
+                // Update cursor position for next iteration
+                last_block_num = Some(tx.block_num);
+                last_transaction_id = Some(tx.transaction_id.clone());
```
igamigo: Looks like the `clone()` (and the assignment) happen per-row but are really only needed for the last row.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,210 @@ pub struct TransactionSummaryRowInsert {
...
+    if total_size >= MAX_PAYLOAD_BYTES {
+        // SAFETY: We're guaranteed to have at least one transaction since total_size > 0
+        let last_block_num = last_block_num.expect(
+            "guaranteed to have processed at least one transaction when size limit is reached",
+        );
+        let filtered_transactions = vec_raw_try_into(
+            all_transactions
+                .into_iter()
+                .take_while(|row| row.block_num != last_block_num)
+                .collect::<Vec<_>>(),
```
igamigo: nit: I think this `collect` could be avoided as well, right?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,210 @@ pub struct TransactionSummaryRowInsert {
...
+        )?;
+
+        // SAFETY: block_num came from the database and was previously validated
+        let last_included_block = BlockNumber::from_raw_sql(last_block_num.saturating_sub(1))?;
+        Ok((last_included_block, filtered_transactions))
+    } else {
+        // SAFETY: last_block_num came from the database and was previously validated
+        let last_included_block = last_block_num.map_or(*block_range.end(), |block_num| {
+            BlockNumber::from_raw_sql(block_num).expect("valid block number from database")
+        });
```
igamigo: Since this is the case where the max payload size was not reached, would this always be equal to the end of the block range?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -1,19 +1,27 @@
+#![allow(
+    clippy::cast_possible_wrap,
+    reason = "We will not approach the item count where i64 and usize cause issues"
+)]
```
igamigo: Would this be better to put into the function itself?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -126,14 +174,210 @@ pub struct TransactionSummaryRowInsert {
...
+                schema::transactions::transaction_id.asc(),
+            ))
+            .limit(NUM_TXS_PER_CHUNK)
+            .load::<TransactionRecordRaw>(conn)
+            .map_err(DatabaseError::from)?;
+
+        // If no more data, we're done
+        if chunk.is_empty() {
+            break;
+        }
```
igamigo: nit: I think I'd remove this `if` because it's functionally very similar to the `if` after the `for` loop and it doesn't seem to be too beneficial to return before it since `chunk` would contain no transactions to loop through

igamigo:
Overall logic looking good. Left a few comments mainly related to avoiding unnecessary allocations and a few minor nits (feel free to disregard if you think they don't apply)

crates/store/src/db/models/queries/transactions.rs
bobbinth: I would probably keep this as is for now. We'll need to address the issue of allocations on deserialization in the future more comprehensively, and I think this will be covered by that.

crates/store/src/server/rpc_api.rs
bobbinth: I'd add it to the list of follow-ups. We could make a single query that brings back note data for all transactions in the response - but that's a bit more involved and shouldn't make a huge difference for an in-process database (but would be very important if if the database was on a different machine).

proto/proto/store/rpc.proto
bobbinth: I think we wanted to keep naming close to how it works with `TransactionHeader`. Agreed that there is some discrepancy here - but I probably wouldn't tackle it in this PR.

crates/store/src/db/mod.rs
bobbinth: Agreed. Not for this PR, but we should probably collect all assumptions in a single place and also make sure we set the response size consistently across different endpoints. For example, some endpoints limit it to 2MB, some to 2.5MB, and this one to 5MB (or maybe actually 4MB).

I'm not sure there is much benefit to such variability - might as well just go with 4MB as a standard max response size target for all endpoints.

proto/proto/store/rpc.proto
SantiagoPittella: Yes, it is essentially that. Makes, i'll do it. Shouldn't be much work.

proto/proto/store/rpc.proto
SantiagoPittella: Ok, I will open an issue to investigate this inconsistencies

proto/proto/store/rpc.proto
bobbinth: I'd probably keep it as `TransactionRecord`, but would just note in the docs that it is almost the same as `TransactionHeader`. Otherwise, it may be too confusing to have identically named entities that are slightly different.

crates/store/src/server/rpc_api.rs
SantiagoPittella: To make it easier than taking the impl trait or something like that, I removed the conversion to `NoteSyncRecord` and implemented `From<NoteRecord> for proto::note::NoteSyncRecord`, thus deleting one conversion.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: This shouldn't be updated, but rather the query should be inclusive too. using `ge` instead.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: I split a bit this logic to avoid the extra clones.

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: Yes, makes sense. The response, in practice, is the same but in the next iteration we avoid going to blocks that we already checked and we saw that didn't matched our criteria. I;/m moving to `block_range.end()`.

proto/proto/store/rpc.proto
SantiagoPittella: Added this and the other followups in https://github.com/0xMiden/miden-node/issues/1258

proto/proto/store/rpc.proto
```diff
@@ -422,3 +425,35 @@ message PaginationInfo {
...
+    repeated account.AccountId account_ids = 2;
+}
+
+// Represents the result of syncing transactions request.
+message SyncTransactionsResponse {
+    // Pagination information.
+    PaginationInfo pagination_info = 1;
+
+    // List of transaction records.
+    repeated TransactionRecord transaction_records = 2;
```
bobbinth, line 449: I'd name this just `transactions` - let's add it to the follow-up.

proto/proto/store/rpc.proto
```diff
@@ -422,3 +425,35 @@ message PaginationInfo {
...
+}
+
+// Represents a transaction record.
+message TransactionRecord {
+    // Block number in which the transaction was executed.
+    fixed32 block_num = 1;
+
+    // A transaction header.
+    transaction.TransactionHeader transaction_header = 2;
```
bobbinth, line 458: nit: maybe I'd change this to just `header`. Let's add this to the follow-ups as well.

PR #1228, opened 2025-09-15
refactor: one task per component in monitor

### LINKED ISSUE #1223: Use one task per component in network monitoring status check
This is fine for now but I think we'll run into problems if/when we start adding more checks.

An alternative I would suggest as follow-up is to split each check into its own task, using a tokio `Watch` channel to communicate its latest status.

_Originally posted by @Mirko-von-Leipzig in https://github.com/0xMiden/miden-node/pull/1217#discussion_r2336031470_

SantiagoPittella:
closes #1223


bin/network-monitor/src/status.rs
```diff
@@ -259,6 +259,49 @@ impl From<RpcStatus> for RpcStatusDetails {
...
+/// and sends updates through a watch channel.
+///
+/// # Arguments
+///
+/// * `rpc_url` - The URL of the RPC service.
+/// * `status_sender` - The sender for the watch channel.
+///
+/// # Returns
+///
+/// A `JoinHandle` for the spawned task.
```
sergerad: Is this comment outdated re return type?

bin/network-monitor/src/status.rs
```diff
@@ -304,6 +347,58 @@ async fn check_rpc_status(
...
+///
+/// # Arguments
+///
+/// * `prover_url` - The URL of the remote prover service.
+/// * `name` - The name of the remote prover.
+/// * `status_sender` - The sender for the watch channel.
+///
+/// # Returns
+///
+/// A `Result` indicating success or failure.
```
sergerad: Probably unnecessary comment about return type

bin/network-monitor/src/status.rs
```diff
@@ -304,6 +347,58 @@ async fn check_rpc_status(
...
+#[instrument(target = COMPONENT, name = "remote-prover-status-task", skip_all)]
+async fn run_remote_prover_status_task(
```
sergerad: If there is any way to deduplicate the logic in this and `run_rpc_status_task` that would be good. Might not be worth the generics involved, if it is practical at all.

bin/network-monitor/src/status.rs
```diff
@@ -377,80 +506,92 @@ async fn check_remote_prover_status(
...
+        tokio::select! {
+            // RPC status changed - update immediately
+            _ = rpc_status_rx.changed() => {
+                update_shared_status(&shared_status, &rpc_status_rx, &prover_status_receivers).await;
+            },
+            // Any prover status changed - update immediately
+            _ = async {
+                for rx in &mut prover_status_receivers {
+                    if rx.changed().await.is_ok() {
+                        return;
```
sergerad: Is the return here intentional?

bin/network-monitor/src/status.rs
```diff
@@ -366,9 +461,43 @@ async fn check_remote_prover_status(
...
+) {
+    let current_time = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap_or_else(|_| Duration::from_secs(0))
+        .as_secs();
+
+    let mut services = Vec::new();
+
+    // Collect RPC status
+    services.push(rpc_status_rx.borrow().clone());
```
sergerad: Would we want borrow_and_update() instead? Given
```
    /// This method does not mark the returned value as seen, so future calls to
    /// [`changed`] may return immediately even if you have already seen the
    /// value with a call to `borrow`.
```

bin/network-monitor/src/status.rs
```diff
@@ -366,9 +461,43 @@ async fn check_remote_prover_status(
 // NETWORK STATUS CHECKER
 // ================================================================================================
 
-/// Checks the status of the network.
+/// Updates the shared status with the latest statuses from all services.
 ///
-/// This function checks the status of the network.
+/// This helper function collects the current status from all watch channels
+/// and updates the shared network status atomically.
+async fn update_shared_status(
```
Mirko-von-Leipzig: You shouldn't need a shared status. Nothing should be shared.

The web server can simply hold all the watch receivers and iterate through them on demand, no locks, no aggregation or sharing required. 

Unless I'm misunderstanding something I would expect things to look roughly like this in `main`:

```rust
async fn main() -> anyhow::Result<()> {
    let config = parse_env()?;

    let mut tasks = JoinSet::new();

    // Spawn the RPC checker
    let (rpc_tx, rpc_rx) = watch::channel();
    tasks.spawn(run_rpc_status_task(rpc_tx, config.rpc_url));

    // Spawn the prover checkers..
    let mut prover_rx = Vec::new();
    for url in config.prover_urls {
        let (tx, rx) = watch::channel();
        tasks.spawn(run_prover_task(tx, url));
        prover_tx.push(rx);
    }

    // Other tasks.

    // Create http server.
    let server = ServerState {
        rpc: rpc_rx,
        provers: prover_rx,
        // etc
    };

    tasks.spawn(server.serve());

    // Abort if any tasks ends etc
    tasks.join_next().await
        .context("...")
    
}
```

bin/network-monitor/src/status.rs
SantiagoPittella: I think that we can get away with this by passing the check function as a parameter

SantiagoPittella: I ended up introducing some generics for the clients and got me some lifetimes issues. I don;t think it is worth it unless we end up adding more component types

bin/network-monitor/src/status.rs
SantiagoPittella: Done!

bin/network-monitor/src/status.rs
SantiagoPittella: This code was removed.

bin/network-monitor/src/status.rs
SantiagoPittella: This code was removed.

bin/network-monitor/src/main.rs
```diff
@@ -34,44 +40,71 @@ async fn main() -> anyhow::Result<()> {
...
+    let mut component_ids: HashMap<_, String> = HashMap::new();
+
+    // Create initial empty status for services
+    let initial_rpc_status = ServiceStatus {
+        name: "RPC".to_string(),
+        status: "unknown".to_string(),
+        last_checked: 0,
+        error: None,
+        details: None,
+    };
```
Mirko-von-Leipzig: Nit: this could probably be it `ServiceStatus::new(name: String)`

bin/network-monitor/src/main.rs
```diff
@@ -34,44 +40,71 @@ async fn main() -> anyhow::Result<()> {
...
+        prover_rxs.push(prover_rx);
+
+        let prover_url_clone = prover_url.clone();
+        let name_clone = name.clone();
+        let component_name = format!("prover-checker-{}", i + 1);
+        let id = tasks
+            .spawn(async move {
+                run_remote_prover_status_task(prover_url_clone, name_clone, prover_tx).await
+            })
+            .id();
```
Mirko-von-Leipzig: Nit: you can move the clones "inside" to prevent the naming pains


```suggestion
        let component_name = format!("prover-checker-{}", i + 1);
        let id = tasks.spawn({
                let prover_url = prover_url.clone();
                let name = name.clone();
                
                run_remote_prover_status_task(prover_url, name, prover_tx)
            })
            .id();
```

Mirko-von-Leipzig (approved):
Thank you!

PR #1230, opened 2025-09-17
feat: distributed tracing in remote prover

### LINKED ISSUE #937: followup(proxy, worker): traceability between services
### What should be done?

In the current state, the traces of the proxy and worker are shown separately. We need to investigate a way to aggregate the traces into one (at least for the normal use case of a prove request). The trace of the worker should be included as part of the trace of the proxy.

This will be a great debugging tool when needed, and will provide rich information about the insights of the system.

### How should it be done?

A way to implement it is through some sort of UUID in the traces, as @Mirko-von-Leipzig [mentioned in this comment](https://github.com/0xPolygonMiden/miden-base/pull/1014#discussion_r1889755609).

From the proxy approach we do not have an easy way to access the `transaction_id` of a given witness, so that approach might be discarded.

Probably the best shot is to use a header like `X-Request-ID`. In order to avoid a user sending duplicated IDs we can always set the header in the proxy.

### When is this task done?

The task can be considered done when in any selected frontend for the traces we have the ability see a single trace for the whole request.

### Additional context

_No response_

SantiagoPittella:
closes #937


bin/remote-prover/src/proxy/mod.rs
```diff
@@ -294,6 +294,27 @@ impl RequestQueue {
...
+struct PingoraHeaderInjector<'a>(&'a mut pingora::http::RequestHeader);
+
+impl opentelemetry::propagation::Injector for PingoraHeaderInjector<'_> {
+    /// Set a key and value in the `RequestHeader` using pingora's API
+    fn set(&mut self, key: &str, value: String) {
+        // Use pingora's insert_header method which handles the proper header insertion
+        // Convert key to owned string to satisfy lifetime requirements
+        if let Err(e) = self.0.insert_header(key.to_string(), value) {
+            // Log error but don't fail the request if header injection fails
+            tracing::warn!(target: COMPONENT, "Failed to inject OpenTelemetry header {}: {}", key, e);
```
Mirko-von-Leipzig:
```suggestion
            tracing::warn!(target: COMPONENT, header: %key, err: %e, "Failed to inject OpenTelemetry header");
```

PR #1231, opened 2025-09-17
chore: add SubmitProvenBatch error in proto

SantiagoPittella:
Did not modified the docs since the endpoint is not documented yet.

Had to add the enum's name as prefix for variants since it is best practice ( https://protobuf.dev/programming-guides/style/?#enums ). The generated code does remove the prefix from the enum.

EDIT: Related to this last part, i replaced INTERNAL_ERROR with just INTERNAL since it is more commonly used.


crates/proto/src/domain/errors.rs
```diff
@@ -7,16 +7,36 @@ impl SubmitProvenTransactionError {
...
+        *self as u8
+    }
+
+    pub fn tonic_code(&self) -> tonic::Code {
+        match self {
+            SubmitProvenBatchError::Internal | SubmitProvenBatchError::Unspecified => {
+                tonic::Code::Internal
+            },
+            SubmitProvenBatchError::DeserializationFailed => tonic::Code::InvalidArgument,
+        }
```
Mirko-von-Leipzig: We could probably combine things a bit by (and the other one)

```suggestion
        if self.is_internal() {
            tonic::Code::Internal
        } else {
            tonic::Code::InvalidArgument
        }
```

proto/proto/types/errors.proto
```diff
@@ -1,14 +1,28 @@
...
-    OUTPUT_NOTES_ALREADY_EXIST = 7;
-    TRANSACTION_EXPIRED = 8;
+    SUBMIT_PROVEN_TRANSACTION_ERROR_INTERNAL = 1;
+    SUBMIT_PROVEN_TRANSACTION_ERROR_DESERIALIZATION_FAILED = 2;
+    SUBMIT_PROVEN_TRANSACTION_ERROR_INVALID_TRANSACTION_PROOF = 3;
+    SUBMIT_PROVEN_TRANSACTION_ERROR_INCORRECT_ACCOUNT_INITIAL_COMMITMENT = 4;
+    SUBMIT_PROVEN_TRANSACTION_ERROR_INPUT_NOTES_ALREADY_CONSUMED = 5;
+    SUBMIT_PROVEN_TRANSACTION_ERROR_UNAUTHENTICATED_NOTES_NOT_FOUND = 6;
+    SUBMIT_PROVEN_TRANSACTION_ERROR_OUTPUT_NOTES_ALREADY_EXIST = 7;
+    SUBMIT_PROVEN_TRANSACTION_ERROR_TRANSACTION_EXPIRED = 8;
```
bobbinth: Question: why do we need to make them this verbose?

SantiagoPittella: Two enum variants, despite being of different `enum`s can share name and causes a collision, which makes the file to fail to build. For that reason, prefixes refering the struct are added ([styling guide here](https://protobuf.dev/programming-guides/style/#enums)). We could use some other, shorter, prefix but couldn't think of anything better than this. In the generated code, the enum (or whatever picked prefix) is removed.

bobbinth: Ah! - I didn't realize this.

I wonder if we should re-consider using gRPC enums for this. As discussed in another PR, this is currently mostly for documentation purposes, and it seems like it creates more issues than it solves.

SantiagoPittella: Despite the long variant names (maybe we can think of some better prefixes), I like the self-documenting features of having it defined in the protos and the auto-generation that can be achieved for any given language

crates/block-producer/src/errors.rs
```diff
@@ -157,6 +155,40 @@ impl From<AddTransactionError> for tonic::Status {
...
+pub(crate) enum SubmitProvenTransactionGrpcError {
+    Internal = 0,
+    DeserializationFailed = 1,
+    InvalidTransactionProof = 2,
+    IncorrectAccountInitialCommitment = 3,
+    InputNotesAlreadyConsumed = 4,
+    UnauthenticatedNotesNotFound = 5,
+    OutputNotesAlreadyExist = 6,
+    TransactionExpired = 7,
+}
```
bobbinth, line 172: Not for this PR, but I wonder if this enum (and all similar enums) should be defined in the `miden-node-proto` crate. We already have `errors.rs` file there. Maybe in addition to that we should have something like `grpc_errors.rs` file.

bobbinth (approved):
Looks good! Thank you! I left one comment inline - but we can address it in a follow-up PR.

PR #1234, opened 2025-09-18
refactor(mempool): state inferred DAG

### LINKED ISSUE #537: [block-producer]: improve `InflightState` error handling
Follow-up task to #534.

`InflightState` and its submodule `InflightAccountState` currently panic instead of passing up errors. Both are also not as strict as they could be e.g. account state handles revert/commiting via an amount `count` instead of reverting/committing a given set of transaction IDs.

Effectively, account state should check that a given set of `n` transactions should

- match the frontmost `n` transactions for commit, and
- match the backmost `n` transactions for revert

Additionally, `InflightState` should not panic if the transaction delta is missing.

Effectively we should check that the transactions exist, and that they are being handled in the correct order.

Note that we probably want to consider transactions as unordered sets here, so that the bookkeeping at the mempool level can be relaxed. e.g. once transactions are rolled into batches, and batches into blocks, that we can erase the transaction ordering. At least within this component.

### LINKED ISSUE #594: [mempool]: improve tx reverting strategy
Currently all transactions in a failed block are reverted in their entirety. Though blocks should not fail, the likely cause would be a bug in the processing of some transaction(s).

Probably most transactions in the block are fine, and we only need to weed out the couple that are causing issues. Right now our strategy just nukes the entire set.

We could improve this by allowing transactions to be part of failing proofs `x` number of times before reverting them. e.g. a transaction is only reverted entirely once it has been part of `x` batch and/or block failures.

### LINKED ISSUE #595: Investigate graph decoupling
The mempool's transaction and batch graphs are currently somewhat indirectly coupled.

As an example, transaction expiration is currently tracked separately in the mempool but ideally transaction and batch graphs should determine their expirations internally (see [comment](https://github.com/0xPolygonMiden/miden-node/pull/582#issuecomment-2542565719)).

One hurdle is that the batch graph is effectively an overlay on top of the transaction graph. Which means separating them is difficult. One possible angle is to increase the richness of the graph return values, which could help the mempool react to changes in one that should impact the other.

Mirko-von-Leipzig:
This PR replaces the existing mempool implementation in an effort to enable user submitted batches #1112.

This PR leaves the mempool in an incomplete state, and as such some follow-up PRs will be made. This is already enough (or too much) to review at once. I'll stack the follow-up PRs on this one before we merge it all.

### Status Quo

The original implementation operated on the assumption that all state could be represented as a DAG of transactions. Unfortunately this assumption no longer holds true with user submitted batches, invalidating the original design.

### New design

The mempool is represented as a DAG of nodes, where a node can represent an unbatched transaction, a batch or a block respectively. The nodes reflect the active state of the system e.g. if a batch is selected, all the transaction nodes are replaced by the batch node. The new node retains the replaced nodes to (a) know what its state impact is, and (b) to allow reverting/re-inserting.

Edges in the DAG are stored by the inflight state (unimplemented). Each piece of state stores the node ID that created and consumed it. A node can therefore lookup its children by looking up the consumers of the data it created. And vice versa for parent lookup. This layer of indirection makes it easy to replace nodes in the graph e.g. replacing a set of transaction nodes with a single batch is simply updating the node ID associated with the node.

The original goal was decoupling the current transaction and batch graphs (#595), however this new design (imo) creates a more sensible single graph layout.

### What's missing

~~The inflight state is not yet implemented; but shouldn't be overly complicated - I've implemented several similar on my way here already.~~

The user submitted batches aren't implemented yet; though I hope you can see it won't be difficult to squeeze into the current design. I've purposefully left this out to allow comparison with the current requirements.

~~Reverting isn't handled yet; this will be complicated a bit by user batches so I'm waiting until then to re-add that.~~

Mempool tests are disabled for now.

Mempool subscription will also need an overhaul to support user batches -- its still unclear how this should look.

------------------------------

Closes #595, closes #537, closes #594

Supersedes #1210.


crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+
+                let parents = self.parents(NodeId::Transaction(candidate.id()));
+                for parent in parents {
+                    match parent {
+                        NodeId::Transaction(parent) if !selected.contains(&parent) => {
+                            continue 'next_candidate;
+                        },
+                        _ => {},
+                    }
+                }
```
PhilippGackstatter: Just to double check: We do this check because the current DAG represents all unprocessed transactions/batches, and so if a parent was _not_ present in the selected batch, we would build an invalid batch with a transaction whose parent is not committed yet, correct?

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+
+                let parents = self.parents(NodeId::ProvenBatch(candidate.id()));
+                for parent in parents {
+                    match parent {
+                        NodeId::ProposedBlock(block) => panic!(
+                            "Block selection encountered an existing proposed block ID {block} while navigating the state DAG"
+                        ),
+                        NodeId::ProvenBatch(batch) if selected.contains(&batch) => {},
+                        NodeId::CommittedBlock(_) => {},
+                        _ => continue 'next_candidate,
```
PhilippGackstatter: Nit: It would be a bit more readable if the patterns were explicitly covered, e.g.:

```rust
NodeId::Transaction(_)
  | NodeId::ProposedBatch(_)
  | NodeId::ProvenBatch(_) => continue 'next_candidate,
```

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+        //
+        // The additional bookkeeping can be implemented once we have fee related strategies. KISS.
+
+        let mut selected = Vec::new();
+
+        'outer: loop {
+            'next_candidate: for candidate in self.nodes.txs.values().map(|tx| &tx.0) {
+                if selected.contains(&candidate.id()) {
+                    continue;
+                }
```
PhilippGackstatter: I'm not sure I understand why this loop works: We never modify `self.nodes.txs` in the loop and so when we `continue 'outer`, aren't we always iterating the same transactions?

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+    /// and consuming [`NodeId's`] are stored alongside the state, which allows this to act as the
+    /// edges of the DAG.
+    state: InflightState,
+
+    /// Number of committed block's whose state we retain.
+    ///
+    /// This provides an overlap with the committed state from the store, giving new data's
+    /// authentication inputs a grace period before it is submitted to the mempool. Without this,
+    /// fetching the inputs from the store would race against new block being committed.
+    state_retention: NonZeroUsize,
```
drahnr: Could we make this deterministic through means of a feedback channel from the block import / application pipeline?

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+                }
+
+                let parents = self.parents(NodeId::ProvenBatch(candidate.id()));
+                for parent in parents {
+                    match parent {
+                        NodeId::ProposedBlock(block) => panic!(
+                            "Block selection encountered an existing proposed block ID {block} while navigating the state DAG"
+                        ),
+                        NodeId::ProvenBatch(batch) if selected.contains(&batch) => {},
+                        NodeId::CommittedBlock(_) => {},
```
PhilippGackstatter: If a proven batch has a committed block as a parent, doesn't that mean it is already committed, or if not, what are the circumstances under which this could happen? If it is already commited I we would have to `continue 'next_candidate`, right?

In any case, I think a one line comment above each non-obvious case might be helpful.

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    /// Selects a set of transactions for inclusion in a batch within the constraints of the budget.
+    ///
+    /// Returns [`None`] if no transactions are available for selection.
+    pub fn propose_batch(
+        &mut self,
+        mut budget: BatchBudget,
+    ) -> Option<(BatchId, Vec<Arc<AuthenticatedTransaction>>)> {
+        // The selection algorithm is fairly neanderthal in nature.
```
drahnr: :smile:

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+                        _ => {},
+                    }
+                }
+
+                if budget.check_then_subtract(candidate.as_ref()) == BudgetStatus::Exceeded {
+                    break 'outer;
+                }
+
+                selected.push(candidate.id());
+                continue 'outer;
```
drahnr: This continue outer is a bit confusing, since it only works due to the `continue block in 189, using an `let mut ier = self.nodes.txs.values().map(..);` and re-using that might be easier to understand.

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+        // We iterate over all transaction nodes, each time selecting the first transaction which
+        // has no parent nodes that are unselected transactions. This is fairly primitive, but
+        // avoids the manual bookkeeping of which transactions are selectable.
+        //
+        // This is still reasonably performant given that we only retain unselected transactions as
+        // transaction nodes i.e. selected transactions become batch nodes.
+        //
+        // The additional bookkeeping can be implemented once we have fee related strategies. KISS.
+
+        let mut selected = Vec::new();
```
drahnr: We might want to do a `HashSet` to ensure uniqueness (yes, we also do by construction, but the logic will change, the constraint should not).

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+            return;
+        };
+
+        // Propagate the node ID change internally.
+        let batch_id = proof.id();
+        let node = ProvenBatchNode { txs, inner: proof };
+        self.nodes.proven_batches.insert(batch_id, node);
+
+        // FIXME: missing safety
+        self.insert_into_state_unchecked(NodeId::ProvenBatch(batch_id));
```
drahnr: nit: We seem to have this as a pattern

PhilippGackstatter (commented):
Looks good!

I took a first look, but will have to take another one later to wrap my head around this.

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+                self.nodes.proven_batches.values().map(|node| &node.inner)
+            {
+                if selected.contains(&candidate.id()) {
+                    continue;
+                }
+
+                let parents = self.parents(NodeId::ProvenBatch(candidate.id()));
+                for parent in parents {
+                    match parent {
+                        NodeId::ProposedBlock(block) => panic!(
```
drahnr: I am a bit confused, why is this an issue? We're checking the parent? That cannot be a proposed block because we only ever can have one, and when calling the function the assumption is there is none in the DAG?

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+                        _ => continue 'next_candidate,
+                    }
+                }
+
+                if budget.check_then_subtract(candidate.as_ref()) == BudgetStatus::Exceeded {
+                    break 'outer;
+                }
+
+                selected.push(candidate.id());
+                continue 'outer;
```
drahnr: Similar to previous `'outer` comment, `let mut iter = `

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+            .take_if(|(proposed, _)| proposed == &to_commit)
+            .expect("block must be in progress to commit");
+
+        self.nodes.committed_blocks.push_back(block);
+        // FIXME: Missing safety
+        self.insert_into_state_unchecked(NodeId::CommittedBlock(to_commit));
+        self.state.chain_tip = to_commit;
+
+        if self.nodes.committed_blocks.len() > self.state_retention.get() {
+            let (number, node) = self.nodes.committed_blocks.pop_front().unwrap();
```
drahnr: We just pushed to the `committed_blocks` and hence `pop_front` must return `Some`.

crates/block-producer/src/mempool/state_dag.rs
```diff
@@ -0,0 +1,442 @@
...
+    /// Blindly inserts the node's state data into [`InflightState`] and associates it with the
+    /// [`NodeId`].
+    ///
+    /// For state data that already exists, this is equivalent to overwriting the [`NodeId`]
+    /// associated with the data.
+    ///
+    /// Callers _must_ ensure that the data is inline with the DAG expectations _and_ that the node
+    /// data is already present in [`Nodes`].
+    fn insert_into_state_unchecked(&mut self, id: NodeId) {
+        todo!()
```
drahnr: These are somewhat critical to judge if the rest works.

crates/block-producer/src/mempool/state_dag.rs
drahnr: It's fixed by the combination of `continue` and `break 'outer` and the fact we keep state in `selected` across iterations

crates/block-producer/src/mempool/state_dag.rs
Mirko-von-Leipzig: Correct yes

crates/block-producer/src/mempool/state_dag.rs
Mirko-von-Leipzig: Not really because things are concurrent. I'll add a more detailed explanation in the `mempool`'s module comments to provide a more general overview but as a shortish summary:

The mempool is used by three primary (but concurrent) processes

1. New txs are appended to the mempool (submitted by users via gRPC)
2. Batch builder(s)
    1. Request next proposed batch, and prove, and
    2. submit the proof back to the mempool
4. Block builder similarly 
    1. Requests the next proposed block, proves it and 
    2. submits the proof back to the mempool

Access is synchronized between these using a `Mutex` aka `SharedMempool` is what is used externally throughout the codebase.

New transactions have to go through the following process before being accepted by the mempool:

1. Proof verification (performed at the RPC component). Note that this only verifies the proof and that the tx contents match the proof.
2. Authenticate the tx against the latest committed state in the store. This does nullifier checks and adds proofs for some of the unauthenticated notes that are present in the store's state.
3. Acquire mempool lock and submit tx.

During steps (2) and (3) the block builder could have committed blocks to the store, which now means there is additional state the tx needs to verify against. We solve this by storing the most recently committed blocks in the mempool as well.

crates/block-producer/src/mempool/state_dag.rs
Mirko-von-Leipzig: The ordering is important i.e. we definitely need a `Vec<Transaction>`

crates/block-producer/src/mempool/state_dag.rs
Mirko-von-Leipzig: That's correct which is why we `panic!` because its unexpected?

I've reworked this now though so no longer applicable.

crates/block-producer/src/mempool/state_dag.rs
Mirko-von-Leipzig: I've refactored/re-organized and added the state mutation.

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,161 @@
...
+        for note in node.unauthenticated_notes() {
+            self.authenticated_notes.remove(&note);
+        }
+
+        for (account, from, to) in node.account_updates() {
+            let Entry::Occupied(entry) =
+                self.accounts.entry(account).and_modify(|entry| entry.remove(from, to))
+            else {
+                // TODO: consider panicking since this shouldn't happen? But we don't assert
+                // about either..
```
drahnr: Proposal: We might want to introduce a `defensive_assert!(condition, keyword oneof(continue | return val | break))` and depending on the build type (`debug_assertions` yay/nay) make it an error or an early return or not. In production it becomes an `::error!(..)`.

crates/block-producer/src/mempool/nodes.rs
```diff
@@ -0,0 +1,302 @@
...
+use miden_objects::note::{NoteHeader, NoteId, Nullifier};
+use miden_objects::transaction::{TransactionHeader, TransactionId};
+
+use crate::domain::transaction::AuthenticatedTransaction;
+
+/// Uniquely identifies a node in the mempool.
+///
+/// This effectively describes the lifecycle of a transaction in the mempool.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub(crate) enum NodeId {
```
drahnr, line 17: Where do we unpack these? Is that outside the current scope?

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: Thing is, this would _always_ indicate the mempool state is corrupt in some fashion and we should abort even in production. But there are a million places we could perform similar assertions, its just that in this case we're forced to unpack the result by the type.

For example, with every insert into a map, we could check that the node/state isn't already present. And with every remove we could assert that the entry actually existed. etc. Might be worth doing I guess but its pretty noisy ito code.

A reasonable approach might be to have this hidden inside functions, but then one needs to be careful on the caller side.

crates/block-producer/src/mempool/nodes.rs
Mirko-von-Leipzig: You mean why is it `pub(crate)`? We use this inside the mempool functions to identify which type of node it is, and therefore which property of `Nodes` to look for it.

Mirko-von-Leipzig: So `pub(super)` would actually be sufficient. The `pub(crate)` was auto-inserted by RA when moving the code into a module. I should probably minimize them to prevent confusion (?)

crates/block-producer/src/mempool/nodes.rs
```diff
@@ -0,0 +1,302 @@
...
+        self.inner.batch_expiration_block_num()
+    }
+}
+
+/// Represents a block - both committed and in-progress.
+#[derive(Clone, Debug, PartialEq, Default)]
+pub(crate) struct BlockNode {
+    txs: Vec<Arc<AuthenticatedTransaction>>,
+    batches: Vec<Arc<ProvenBatch>>,
+}
```
bobbinth, line 120: I haven't looked far enough, but how does this represent a committed block? Wouldn't we need to have the `ProvenBlock` struct in it too?

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Note that the user is responsible for ensuring that the inserted nodes adhere to a DAG
+/// structure. No attempt is made to verify this internally here as it requires more information
+/// than is available at this level.
+#[derive(Clone, Debug, PartialEq, Default)]
+pub(crate) struct InflightState {
+    nullifiers: HashSet<Nullifier>,
+    output_notes: HashMap<NoteId, NodeId>,
+    authenticated_notes: HashMap<NoteId, NodeId>,
+    accounts: HashMap<AccountId, AccountUpdates>,
+}
```
bobbinth, line 48: Could we add brief doc comments to each field explaining what it tracks? For example, for `output_notes: HashMap<NoteId, NodeId>` is not entirely obvious what is the key and what is the value.

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,161 @@
...
+                continue;
+            };
+
+            if entry.get().is_empty() {
+                entry.remove_entry();
+            }
+        }
+    }
+
+    pub(crate) fn insert(&mut self, id: NodeId, node: &dyn Node) {
```
bobbinth, line 119: Could we add doc comments explaining what the parameters are? I'm assuming that `id` is the parent of the `node` - but maybe it isn't so?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -1,141 +1,82 @@
...
+//! reverted.
+//!
+//! The DAG nature of the graph is maintained by:
+//!
+//! - Ensuring incoming transactions are only ever appended to the current graph. This in turn
+//!   implies that the transaction's state transition must build on top of the current mempool
+//!   state.
+//! - Parent/child edges between nodes in the graph are formed via state dependency.
+//! - Transactions are proposed for batch inclusion only once _all_ its ancestors have already been
+//!   included in a batch.
```
bobbinth: Is this different from how it used to work before? I thought if we have two transactions `A` and `B` and `B` depends on `A`, both of them could end up being in the same batch. But this seems to imply that first we'd include `A` in a batch, and then `B` would be included in a following batch.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -1,141 +1,82 @@
...
-    /// The operation remained within the budget.
-    WithinScope,
-    /// The operation exceeded the budget.
-    Exceeded,
+#[derive(Debug, Clone)]
+pub struct MempoolConfig {
+    pub block_budget: BlockBudget,
+    pub batch_budget: BatchBudget,
+    pub expiration_slack: u32,
+    pub state_retention: NonZeroUsize,
```
bobbinth, line 101: Do we have the meaning of these fields documented somewhere? I can kind of guess what they imply - but would be good to know for sure.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: This works the same as before; a less ambiguous phrasing would be something like

```rust
//! - Transactions may be included in a proposed batch only once _all_ its ancestors are themselves included in a batch.
//!   Notably, this includes the batch currently being proposed i.e. if transaction `A` depends on `B` then `A` may be included in the same batch as `B`.
```

I'll do another pass on the module docs; I think they can be improved.

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: Its the ID of the node being inserted. 

I wonder if I shouldn't add a `Node::id(&self) -> NodeId` method to the trait.. The thing is most of the time these nodes are in a hashmap with an unwrapped ID e.g. `TransactionId` and not `NodeId::Transaction` so it didn't seem useful.

crates/block-producer/src/mempool/nodes.rs
Mirko-von-Leipzig: I initially had separate `ProposedBlock` and `ProvenBlock` types, but it turns out the mempool doesn't really care about the block proof since there is no next processing stage that uses it. In practice the difference is that we don't aggregate and erase the state from the batches in a proven block, but that has no real impact.

One could make a similar argument about `ProposedBatch` and `ProvenBatch` except that:

1. There is a next stage that requires the batch proof, and
2. once user batches are introduced there are proven batches that don't come from proposed batches

If it makes the model easier to understand I can add another node type; its not much of a change.

crates/block-producer/src/mempool/nodes.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// may not be strictly necessary but it removes having to worry about reverting batches and
+    /// blocks with erased notes -- since these would otherwise have different state impact than
+    /// the transactions within them.
+    fn nullifiers(&self) -> Box<dyn Iterator<Item = Nullifier> + '_>;
+
+    /// All output notes created by this node, **including** nullifiers for erased notes. This
+    /// may not be strictly necessary but it removes having to worry about reverting batches and
+    /// blocks with erased notes -- since these would otherwise have different state impact than
+    /// the transactions within them.
+    fn output_notes(&self) -> Box<dyn Iterator<Item = NoteId> + '_>;
```
PhilippGackstatter:
```suggestion
    /// All output notes created by this node, **including** erased notes. This
    /// may not be strictly necessary but it removes having to worry about reverting batches and
    /// blocks with erased notes -- since these would otherwise have different state impact than
    /// the transactions within them.
    fn output_notes(&self) -> Box<dyn Iterator<Item = NoteId> + '_>;
```

Is this a copy and paste error?

crates/block-producer/src/mempool/nodes.rs
```diff
@@ -0,0 +1,302 @@
...
+    fn unauthenticated_notes(&self) -> Box<dyn Iterator<Item = NoteId> + '_>;
+    fn account_updates(&self) -> Box<dyn Iterator<Item = (AccountId, Word, Word)> + '_>;
```
PhilippGackstatter, line 157: Nit: May be worth mentioning in a comment what the `Word`s are.

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,161 @@
...
+        self.to.insert(to, id);
+    }
+
+    fn parent(&self, from: &Word) -> Option<&NodeId> {
+        self.to.get(from)
+    }
+
+    fn child(&self, to: &Word) -> Option<&NodeId> {
+        self.from.get(to)
+    }
```
PhilippGackstatter, line 239: If I have a transaction node 0 that goes from account state X to Y, then we insert into the state `to: (Y, 0)` and `from: (X, 0)`, right? The parent of node 0 is then the node that caused the account to go `to` state X (which is the state _from_ which node 0 starts) and node 0's child is the node that starts `from` state `Y` (which is the state _to_ which node 0 goes).

Once written out, I think this makes sense, but took me a bit to understand that calling `self.from.get(to)` is correct. I wonder if we can add comments here to make this a bit easier.

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,190 @@
...
+            self.output_notes.remove(&note);
+        }
+
+        for note in node.unauthenticated_notes() {
+            self.unauthenticated_notes.remove(&note);
+        }
+
+        for (account, from, to) in node.account_updates() {
+            let Entry::Occupied(entry) =
+                self.accounts.entry(account).and_modify(|entry| entry.remove(from, to))
```
PhilippGackstatter, line 106: We only care about unauthenticated notes here, because authenticated notes are authenticated against the chain state at the tx kernel level, which, in the node, would already be part of the committed state and therefore not here in the inflight state, right?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -1,141 +1,112 @@
-use std::collections::BTreeSet;
+//! The [`Mempool`] is responsible for receiving transactions, and proposing transactions for
+//! inclusion in batches, and proposing batches for inclusion in the next block.
+//!
+//! It performs these tasks by maintaining a dependency graph between all inflight transactions,
+//! batches and blocks. A parent-child dependency edge between two nodes exists whenever the child
+//! consumes a piece of state that the parent node created. To be more specific, node `A` is a
+//! child of node `B`:
+//!
+//! - if `B` created an output note which is the input note of `A`, or
```
PhilippGackstatter, line 9: I think it would be good to mention that this only applies to unauthenticated notes.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -1,141 +1,112 @@
...
+//! - if `B` updated account `x` to state `x'`, and `A` is updating the account from `x' -> x''`.
```
PhilippGackstatter:
```suggestion
//! - if `B` updated account from state `x` to `x'`, and `A` is updating the account from `x' -> x''`.
```

Nit: Current phrasing makes it sound like we refer to the account itself as `x`.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -274,55 +218,119 @@ impl Mempool {
...
 
-        let removed_batches =
-            self.batches.remove_batches([batch].into()).expect("Batch was not present");
+        // TODO(mirko): This will be somewhat complicated by the introduction of user batches
+        // since these don't have all inner txs present and therefore must at least partially
+        // revert their own tx descendents.
 
-        let transactions = removed_batches.values().flatten().copied().collect();
+        // Remove all descendents and reinsert their transactions. This is safe to do since
+        // a batch's state impact is the aggregation of its transactions.
```
PhilippGackstatter, line 310: Can you remind me why the tx headers in a `ProvenBatch` aren't sufficient for this?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -363,26 +421,21 @@ impl Mempool {
...
         self.chain_tip = self.chain_tip.child();
 
-        self.subscription.block_committed(header, transactions);
-
-        // Revert expired transactions and their descendents.
-        self.revert_expired_transactions();
+        if self.nodes.committed_blocks.len() > self.config.state_retention.get() {
+            let (_number, node) = self.nodes.committed_blocks.pop_front().unwrap();
+            self.state.remove(&node);
+        }
```
PhilippGackstatter, line 442: Can this pruning make a `NodeId::Block` unavailable that is the parent of a proven batch and thereby prevent that batch from being included in a block? I'm assuming there is something that prevents this from happening.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -401,83 +454,30 @@ impl Mempool {
     ///
     /// Panics if there is no block in flight.
     #[instrument(target = COMPONENT, name = "mempool.rollback_block", skip_all)]
-    pub fn rollback_block(&mut self) {
-        let batches = self.block_in_progress.take().expect("No block in progress to be failed");
-
-        // Revert all transactions. This is the nuclear (but simplest) solution.
+    pub fn rollback_block(&mut self, block: BlockNumber) {
```
PhilippGackstatter, line 462: Nit: "in in" typo in the doc comment of this function.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -401,83 +454,30 @@ impl Mempool {
...
+            .as_ref()
+            .is_none_or(|(proposed, _)| proposed != &block)
+        {
+            return;
+        }
 
-        expired
-    }
+        // Remove all descendents. No re-inserting at all.
+        let _reverted = self.revert_subtree(NodeId::Block(block));
```
PhilippGackstatter: Why aren't these reinserted?

PhilippGackstatter (commented):
Looks good!

The overall structure makes sense to me and the code is well written, imo. I think I mainly left minor comments and questions. I'll have to take another look to think through some other parts, given that this has a large surface area.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: The issue isn't removing them, but rather that we cannot re-queue the transactions since only the transactions that predated the user batch exist (and have proofs).

This means any other nodes that depend on a user batch will need to be reverted and also cannot be re-queued because we cannot trivially tell which state remains intact. We can litigate the details in the follow-up PR that will add user batch support.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: Parents and children edges aren't actually persisted anywhere and instead are recalculated on demand from whatever state remains inflight. As a result, if a block has been pruned then it won't be returned from `state.parents` since it doesn't exist anywhere in the state anymore.

As a demonstrating example, before pruning a batch node `B` depends on block `N` and then `state.parents(&B).contains(N.id())` would be `true`. After pruning, this would be false because `N` is no longer part of the state.

Or even more specifically, if `N` created note `n` then `state.output_notes[n] = N` and this gets found by `state.parents(..)`. After pruning `state.output_notes[n]` doesn't exist, and `N` won't be part of the parent set.

The mempool can only be used `sync`, so pruning/batch stuff can't happen concurrently.

Maybe a different way of thinking about it is, is that not all input state will be present as output state within the mempool. This requirement is only enforced when _new_ data is inserted aka at transaction insertion time.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: An arbitrary business logic decision. 

Since we don't know _why_ the block failed, we assume that there is some bug in the system that a transaction or batch within the block runs into. And to prevent this from causing repeated failures we just nuke everything.

We could apply similar reasoning to batch building, but really we should have smarter approaches like counting number of failures per tx/batch and then evicting the offensive ones after some break point is reached.

I'll expand the comment to something along the above.

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: Correct. And thanks; I've been scratching my on how to write this :sweat_smile:

Mirko-von-Leipzig: I've done my best; maybe TAL and let me know if you have better ideas on phrasing..

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: Correct; and even if we wanted to, we couldn't do anymore because `ProvenTransaction` doesn't contain the node IDs for authenticated notes.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: Tried my best :+1: TAL please.

crates/block-producer/src/mempool/mod.rs
PhilippGackstatter: Ah, makes sense. If a proven batch has no parents, because its parents are part of the committed state, it also cannot be excluded due to a missing parent check. Instead, it would always be included.

crates/block-producer/src/mempool/mod.rs
PhilippGackstatter: Sounds good, thank you!

crates/block-producer/src/mempool/state.rs
```diff
@@ -187,10 +187,18 @@ impl AccountUpdates {
...
+    /// In other words, this returns the node ID where `node.to == from`. This infers the
+    /// parent-child relationship where `parent.to == child.from`.
     fn parent(&self, from: &Word) -> Option<&NodeId> {
         self.to.get(from)
     }
 
+    /// Returns the node ID that consumed the given commitment,
+    ///
+    /// In other words, this returns the node ID where `node.from == to`. This infers the
+    /// parent-child relationship where `parent.to == child.from`.
```
PhilippGackstatter:
```suggestion
    /// Returns the node ID that consumed the given commitment.
    ///
    /// In other words, this returns the ID of the `node` where `node.from == to`. This infers the
    /// parent-child relationship where `parent.to == child.from`.
```

Nit: Maybe like this? Otherwise, sounds great!

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,205 @@
...
+            node.unauthenticated_notes().filter_map(|note| self.output_notes.get(&note));
+
+        let account_parents = node
+            .account_updates()
+            .filter_map(|(account, from, _to)| {
+                self.accounts.get(&account).map(|account| account.parent(&from))
+            })
+            .flatten();
+
+        account_parents.chain(note_parents).copied().collect()
```
bobbinth: Do we always consume the result as a hashmap? An alternative could be to return an iterator and then the caller collects it into a hashmap (if needed).

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    fn is_empty(&self) -> bool {
+        self.from.is_empty() && self.to.is_empty()
+    }
+
+    fn remove(&mut self, from: Word, to: Word) {
+        self.from.remove(&from);
+        self.to.remove(&to);
+    }
```
bobbinth, line 212: IIUC, the node ID associated with both `from` and `to` should be the same, right? If so, should we add an `assert!` to double-check this?

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    fn insert(&mut self, id: NodeId, from: Word, to: Word) {
+        self.from.insert(from, id);
+        self.to.insert(to, id);
+    }
```
bobbinth, line 223: This should fail if `!self.to.contains(from) && !self.from.is_empty()` - right? If so, should we explicitly check this?

crates/block-producer/src/mempool/budget.rs
```diff
@@ -0,0 +1,102 @@
...
+pub struct BatchBudget {
+    /// Maximum number of transactions allowed in a batch.
+    pub transactions: usize,
+    /// Maximum number of input notes allowed.
+    pub input_notes: usize,
+    /// Maximum number of output notes allowed.
+    pub output_notes: usize,
+    /// Maximum number of updated accounts.
+    pub accounts: usize,
+}
```
bobbinth, line 22: Not related to this PR, but I wonder if we should express some of these in terms that are closer to how we think about fees.

For example, instead of the maximum number of transactions, maybe it should be the maximum number of "verification units" (which we currently compute as `log2(num_tx_cycles)`. So, a simple transaction with Falcon signature would be 17 units, and a much more complicated transaction that requires 1M cycles, would be 20 units. And then if we set the budget at 250 verification units, we'd be able to include 15 - 16 regular transactions in a batch, or 11 - 12 "big" transactions etc.

Similar could probably be applied to account updates and input notes - though, this would require more thinking.

cc @PhilippGackstatter.

crates/block-producer/src/mempool/budget.rs
```diff
@@ -0,0 +1,102 @@
...
+
+/// Constraints placed on the blocks proposed by the [`Mempool`](super::Mempool).
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct BlockBudget {
+    /// Maximum number of batches allowed in a block.
+    pub batches: usize,
+}
```
bobbinth, line 29: Related to the previous comment (and also not for this PR), we might have to express this budget in terms of more basic units (e.g., number of verification units).

crates/block-producer/src/mempool/budget.rs
```diff
@@ -0,0 +1,102 @@
...
+    fn default() -> Self {
+        Self { batches: DEFAULT_MAX_BATCHES_PER_BLOCK }
+    }
+}
+
+impl BatchBudget {
+    /// Attempts to consume the transaction's resources from the budget.
+    ///
+    /// Returns [`BudgetStatus::Exceeded`] if the transaction would exceed the remaining budget,
+    /// otherwise returns [`BudgetStatus::Ok`] and subtracts the resources from the budger.
```
bobbinth: typo: "budger" -> "budget".

crates/block-producer/src/mempool/nodes.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub fn calculate_id(&self) -> BatchId {
+        BatchId::from_transactions(
+            self.0
+                .iter()
+                .map(AsRef::as_ref)
+                .map(AuthenticatedTransaction::raw_proven_transaction),
+        )
+    }
+
+    pub fn with_proof(self, proof: Arc<ProvenBatch>) -> ProvenBatchNode {
```
bobbinth: nit: I would maybe rename this to something like `to_proven_batch_node()`. `with_proof()` feels like a constructor method for `ProposedBatchNode`.

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,205 @@
...
+        account_children.chain(note_children).copied().collect()
+    }
+}
+
+/// The commitment updates made to a account.
+#[derive(Clone, Debug, PartialEq, Default)]
+struct AccountUpdates {
+    from: HashMap<Word, NodeId>,
+    to: HashMap<Word, NodeId>,
+}
```
bobbinth, line 182: I wonder if it possible to track the info in a vector - e.g., `Vec<(Word, NodeId)>` - this would basically be equivalent to the current `to` map - just organized sequentially. The only missing part would be the initial state of the account - and this could be just a separate field - e.g., something like:
```rust
struct AccountUpdates {
    init_state: Word,
    transitions: Vec<(Word, NodeId)>,
}
```
Though, maybe doing it this way makes things a bit more cumbersome.

Mirko-von-Leipzig: You can but its a lot more cumbersome. Bear in mind that when updating the graph you often want to remove and re-insert some in the middle. For example, if multiple transactions for this account get put in a batch then each update needs to be removed from the middle, and then one batch update will be placed in the middle. 

All of this is possible, but it just makes the code more involved and error prone because remove & insert are separate actions.

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: The iterator includes duplicates, so collecting to a hashmap made sense to me. Otherwise the caller has to be aware that this can contain duplicates.

crates/block-producer/src/mempool/budget.rs
PhilippGackstatter: We haven't really thought about the fees that a batch needs to pay to be included in a block but the optimal strategy for this would depend on how this is defined, iiuc. Most likely, this will also be based on these "verification units" though, so this would probably be a good strategy long term. Until we have defined that, I'm not sure which "hard facts" we would base the budget on, but maybe I'm missing something.

Mirko-von-Leipzig: #1242

crates/block-producer/src/mempool/budget.rs
Mirko-von-Leipzig: #1242

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: I've added a bunch of state related assertions

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: I've added a bunch of state assertions

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,240 @@
...
+                self.accounts.get(&account).map(|account| account.parent(&from))
+            })
+            .flatten();
+
+        account_parents
+            .chain(note_parents)
+            .copied()
+            // Its possible for a node to have internal state connecting to itself. For example,
+            // a proposed batch has not erased the internally produced and consumed notes.
+            .filter(|parent| parent != &id)
```
bobbinth, line 149: Can this actually happen? If a note is being consumed in a batch, then it is must either be produced in the same batch, or be an authenticated note (i.e., produced in some other batch) - or am I missing something?

Are there any other examples of why a node could also be its own parent?

crates/block-producer/src/mempool/state.rs
```diff
@@ -0,0 +1,240 @@
...
+                self.accounts.get(&account).map(|account| account.child(&to))
+            })
+            .flatten();
+
+        account_children
+            .chain(note_children)
+            .copied()
+            // Its possible for a node to have internal state connecting to itself. For example,
+            // a proposed batch has not erased the internally produced and consumed notes.
+            .filter(|child| child != &id)
```
bobbinth, line 172: Same question as above.

bobbinth (approved):
Looks good! Thank you! I left a couple more questions inline. Overall, I didn't review everything too deeply - so, another set of eyes (or two) looking through this would be appreciated.

We can merge this PR into a new branch (e.g., `mempool-refactoring`) and then merge subsequent PRs into that branch until it is ready to be merged into `next`.

As discussed offline, I think the next steps could be:

1. Re-enabled mempool events.
2. Re-add telemetry.
3. Implement tests - at this point we should be able to merge `mempool-refactoring` into `next`.
4. Add user batch support (which may require some refactoring of mempool events).

Also, not for this PR, but it would be cool to add some mempool related metrics to the `BlockProducerStatus` protobuf message. For example:

- Number of unbatched transactions.
- Number of batches.
- Maybe total number of transactions (batched or unbatched).

Let's create an issue for this (unless we already have one).

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: > If a note is being consumed in a batch, then it is must either be produced in the same batch, or be an authenticated note (i.e., produced in some other batch) - or am I missing something?

The `ProposedBatchNode` and `BlockNode` don't perform any aggregation - that is, they don't erase internal notes, nor aggregate account updates. They only delegate the state their transactions/batches create.

We can change this; but we'll have to allocate extra data containing these unerased notes etc.

Mirko-von-Leipzig:
> Looks good! Thank you! I left a couple more questions inline. Overall, I didn't review everything too deeply - so, another set of eyes (or two) looking through this would be appreciated.

@igamigo @drahnr @PhilippGackstatter I would appreciate another look if you have the time. I'll leave this PR open until early next week but since its merging into a feature branch reviews after the fact can still be addressed.

> Also, not for this PR, but it would be cool to add some mempool related metrics to the `BlockProducerStatus` protobuf message. <...>

Created issue #1244

Mirko-von-Leipzig:
@bobbinth apologies; UI bugged and I pinged you for another review; wasn't intended.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -274,55 +223,119 @@ impl Mempool {
...
+        let Some(proposed) = self.nodes.proposed_batches.remove(&proof.id()) else {
             return;
-        }
+        };
+
+        self.state.remove(&proposed);
+
+        let proven = proposed.to_proven_batch_node(proof);
+        self.state.insert(NodeId::ProvenBatch(proven.id()), &proven);
+        self.nodes.proven_batches.insert(proven.id(), proven);
```
PhilippGackstatter, line 337: When a batch transitions from proposed to proven, we remove the proposed batch and reinsert the proven batch. This essentially re-associates all state it references (e.g. output notes it creates) from the `NodeId::ProposedBatch` (removed) to the new `NodeId::ProvenBatch` (inserted), right?

I'm asking because while they carry the same `BatchId` internally, the `Hash` implementation is different for these IDs, but if the above is correct, then that's fine. Maybe you can add a comment on `NodeId` that the Hash impl is only used for deduplication purposes or something like that, but also fine as is.

PhilippGackstatter (approved):
Looks good to me!

I tried to trace through some cases I thought were interesting and I didn't find any issues, but I also definitely didn't think through all of them. I think a test suite is necessary to cover everything here. In any case, the general abstractions over txs, batches and blocks looks great and I found the structure reasonably (given the complexity involved) easy to follow.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: > When a batch transitions from proposed to proven, we remove the proposed batch and reinsert the proven batch. This essentially re-associates all state it references (e.g. output notes it creates) from the NodeId::ProposedBatch (removed) to the new NodeId::ProvenBatch (inserted), right?

That's correct.

In practice, there is also a slight difference in the state of both because the proposed batch doesn't actually aggregate the txs state i.e. it doesn't erase notes, nor fold account updates together. But that's more an implementation detail, which could be changed if desired.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -336,14 +349,64 @@ impl Mempool {
...
+
+            // A batch is selectable if all parents are already blocks, or if the batch is part of
+            // the current block selection.
+            for parent in self.state.parents(NodeId::ProvenBatch(candidate.id()), candidate) {
+                match parent {
+                    NodeId::Block(_) => {},
+                    NodeId::ProvenBatch(parent) if !selected.contains(parent) => {},
+                    _ => continue 'next,
+                }
+            }
```
igamigo, line 384: Is this right? We want to move forward with the batch whose parent batches are already within the in-progress block.

```suggestion
            for parent in self.state.parents(NodeId::ProvenBatch(candidate.id()), candidate) {
                match parent {
                    NodeId::Block(_) => {},
                    NodeId::ProvenBatch(parent) if selected.contains(parent) => {},
                    _ => continue 'next,
                }
            }
```

igamigo (approved):
Not the most in-depth review, but overall looks solid and fairly easy to follow; might take a deeper look later. Left a comment on what might be a bug.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: Ah yes; thank you! I think I had the logic inverted previously and I missed this during that refactor.

crates/block-producer/src/mempool/state.rs
bobbinth: Let's keep as is for now - but maybe create an issue to see if we want to do anything about this later as it does feel a bit counterintuitive that nodes can have themselves as dependencies.

Mirko-von-Leipzig: #1253

PR #1236, opened 2025-09-19
feat: add remote prover tests run on monitor

SantiagoPittella:
Send periodical `prove` requests to the remote provers.

I ended up refactoring a bit the code for the sake of simplicity and type safety.

### DIFF
--- a/bin/network-monitor/src/remote_prover.rs
+++ b/bin/network-monitor/src/remote_prover.rs
@@ -3,7 +3,7 @@
 //! This module contains the logic for periodically testing remote transaction prover functionality
 //! by sending mock transactions and checking for successful transaction proof generation.
 
-use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use std::time::Duration;
 
 use anyhow::Context;
 use miden_node_proto::clients::{Builder as ClientBuilder, RemoteProverClient};
@@ -18,11 +18,11 @@ use serde::{Deserialize, Serialize};
 use tokio::sync::watch;
 use tokio::time::MissedTickBehavior;
 use tonic::Request;
-use tracing::instrument;
+use tracing::{info, instrument};
 use url::Url;
 
-use crate::COMPONENT;
 use crate::status::{ServiceDetails, ServiceStatus, Status};
+use crate::{COMPONENT, current_unix_timestamp_secs};
 
 // PROOF TYPE
 // ================================================================================================
@@ -94,7 +94,7 @@ pub async fn run_remote_prover_test_task(
     proof_type: ProofType,
     serialized_request_payload: proto::remote_prover::ProofRequest,
     status_sender: watch::Sender<ServiceStatus>,
-) -> anyhow::Result<()> {
+) {
     let mut client = ClientBuilder::new(prover_url)
         .with_tls()
         .expect("TLS is enabled")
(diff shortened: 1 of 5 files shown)


bin/network-monitor/src/remote_prover.rs
```diff
@@ -0,0 +1,318 @@
...
+) -> anyhow::Result<()> {
+    let mut client = ClientBuilder::new(prover_url)
+        .with_tls()
+        .expect("TLS is enabled")
+        .with_timeout(Duration::from_secs(10))
+        .without_metadata_version()
+        .without_metadata_genesis()
+        .connect_lazy::<RemoteProverClient>();
+
+    let mut interval = tokio::time::interval(Duration::from_secs(30)); // Test every 30 seconds
```
sergerad: Just wondering if we should make this relatively large - say 1 or 2 minutes? Particularly if we expect testnet/mainnet provers to be under significant load (I'm not sure how much load this request itself involves).

And it would be good to make this configurable but not required for this PR.

sergerad (approved):

bin/network-monitor/src/remote_prover.rs
SantiagoPittella: Yes, agree that the best is making this configurable. The configuration options is growing and I'm thinking that switching to `clap` might be a good idea, wdyt?

SantiagoPittella: In the meantime, I changed it to 2 mins.

bobbinth:
Not a review yet (I haven't looked at the code) - but curious how does the frontend look now? Specifically, what kind of info do we show for each proof "probe".

Also, how are we building the proof request? For this we need to have a transaction, right? Are we using a transaction against the faucet, or something else?

bin/network-monitor/src/remote_prover.rs
sergerad: Yea clap sounds good. Main thing is to have everything configurable via env vars as always.

SantiagoPittella:
> Nice! Another UI question: in the RPC card, is there not enough space to show the full genesis block hash? (it feels like it should almost fit there).

I'm pasting an image of how it looks with this change. I think that it is too large, we can fit it but making the card that wide will make us to have a lot of free space. Having the short version + the copy button looks better IMO and functionality-wise is the same information.

Screenshot 2025-09-24 at 15 02 09

bin/network-monitor/src/remote_prover.rs
```diff
@@ -0,0 +1,313 @@
...
+    let mut success_count = 0u64;
+    let mut failure_count = 0u64;
+
+    loop {
+        interval.tick().await;
+
+        let current_time = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .context("failed to get current time")?
+            .as_secs();
```
Mirko-von-Leipzig: I wonder if only doing `SystemTime::now()` here, and leaving the unix timestamp conversion to the `frontend.rs` handler call might make sense? That makes all of these functions infallible.

Mirko-von-Leipzig (approved):

bin/network-monitor/src/remote_prover.rs
SantiagoPittella: That would be great, another approach could be making something like:
```rust
    time.duration_since(UNIX_EPOCH)
        .unwrap_or_else(|_| Duration::from_secs(0))  // Fallback to 0 if before Unix epoch
        .as_secs()
```

where we control that the tasks nor the frontend fails with this error. wdyt?

Mirko-von-Leipzig: What I don't really like is casting `as_secs()` here since imo `Duration` is more descriptive and it makes sense that the _frontend_ decides to represent that as seconds

SantiagoPittella: Done! Did this change + replace the return type of the return function.

bobbinth (approved):

PR #1237, opened 2025-09-22
fix: add env var to enable OTel in network monitor

SantiagoPittella:
Create a new env var to enable OTel. The variable name is defined in the `status.rs` file, which doesn't make much sense. A new `config.rs` file was created (along with the remote prover tests feature), and it is currently open for reviews here: https://github.com/0xMiden/miden-node/pull/1236/

### DIFF
--- a/bin/network-monitor/src/status.rs
+++ b/bin/network-monitor/src/status.rs
@@ -34,6 +34,7 @@ const DEFAULT_PORT: u16 = 3000;
 const RPC_URL_ENV_VAR: &str = "MIDEN_MONITOR_RPC_URL";
 const REMOTE_PROVER_URLS_ENV_VAR: &str = "MIDEN_MONITOR_REMOTE_PROVER_URLS";
 const PORT_ENV_VAR: &str = "MIDEN_MONITOR_PORT";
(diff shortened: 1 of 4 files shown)


sergerad (approved):

bin/network-monitor/src/status.rs
```diff
@@ -34,6 +34,7 @@ const DEFAULT_PORT: u16 = 3000;
 const RPC_URL_ENV_VAR: &str = "MIDEN_MONITOR_RPC_URL";
 const REMOTE_PROVER_URLS_ENV_VAR: &str = "MIDEN_MONITOR_REMOTE_PROVER_URLS";
 const PORT_ENV_VAR: &str = "MIDEN_MONITOR_PORT";
+pub(crate) const ENABLE_OTEL_ENV_VAR: &str = "MIDEN_MONITOR_ENABLE_OTEL";
```
bobbinth, line 37: Is there a reason why we do not include this into the `MonitorConfig` struct?

SantiagoPittella: I ended up including it in another PR since this was already merged 👌🏼

PR #1239, opened 2025-09-23
fix(store): align account_storage_map_values PK with migration

MozirDmitriy:
The Diesel schema declared account_storage_map_values primary key as (account_id, block_num, slot, key, is_latest_update), but the authoritative SQL migration defines it as (account_id, block_num, slot, key). Including is_latest_update in the PK is incorrect: it’s a status flag used for toggling latest rows and is not part of the uniqueness constraint. This mismatch could cause Diesel to infer the wrong primary key shape, leading to incorrect query generation and violating the intended update pattern (update latest=false, insert latest=true). Code and tests already rely on the migration’s definition and filter by (account_id, slot, key) with is_latest_update=true. This edit removes is_latest_update from the PK in schema.rs to match the migration and the implemented logic.

### DIFF
--- a/crates/store/src/db/schema.rs
+++ b/crates/store/src/db/schema.rs
@@ -1,7 +1,7 @@
 // @generated automatically by Diesel CLI.
 
 diesel::table! {
-    account_storage_map_values (account_id, block_num, slot, key, is_latest_update) {
+    account_storage_map_values (account_id, block_num, slot, key) {
         account_id -> Binary,
         block_num -> BigInt,
         slot -> Integer,


drahnr:
The issue exists, but by definition we only ever have a single row with `is_latest_update=true`. In cojunction with how we do the insertion (first set all `is_latest_update=false`, then insert the new value with `is_latest_update=true` there is only one failure case, which is inserting a pre-existing block a second time, which we handle already accidentally:

```
    count += insert_block_header(conn, block_header)?;
    count += upsert_accounts(conn, accounts, block_header.block_num())?;
```

the first line is where we would fail if we had duplicate blocks, and hence never make it to the second, which would open us up to the issue, but to my understanding, we cannot get there, since we error out with duplicate key error on the primary key of `table block_headers` which is `block_num`.

tl;dr cannot appear in practice, should merge though

drahnr (approved):

PR #1240, opened 2025-09-23
chore: add faucet check in network monitor

SantiagoPittella:
Periodically checks the faucet workflow.

Screenshot 2025-09-24 at 11 42 47

### DIFF
--- a/bin/network-monitor/src/faucet.rs
+++ b/bin/network-monitor/src/faucet.rs
@@ -127,8 +127,11 @@ pub async fn run_faucet_test_task(
             details: ServiceDetails::FaucetTest(test_details),
         };
 
-        // Send the status update (ignore if no receivers)
-        let _ = status_sender.send(status);
+        // Send the status update; exit if no receivers (shutdown signal)
+        if status_sender.send(status).is_err() {
+            info!("No receivers for faucet status updates, shutting down");
+            return Ok(());
+        }
     }
 }
 
--- a/bin/network-monitor/src/faucet.rs
+++ b/bin/network-monitor/src/faucet.rs
@@ -1,0 +1,237 @@
+//! Faucet testing functionality.
+//!
+//! This module contains the logic for periodically testing faucet functionality
+//! by requesting proof-of-work challenges, solving them, and submitting token requests.
+
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use anyhow::Context;
+use miden_objects::account::AccountId;
+use miden_objects::testing::account_id::ACCOUNT_ID_SENDER;
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+use tokio::sync::watch;
+use tokio::time::MissedTickBehavior;
+use tracing::{debug, info, instrument, warn};
+use url::Url;
+
+use crate::COMPONENT;
+use crate::status::{ServiceDetails, ServiceStatus, Status};
+
+// FAUCET TEST TYPES
+// ================================================================================================
+
+/// Details of a faucet test.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct FaucetTestDetails {
+    pub test_duration_ms: u64,
+    pub success_count: u64,
+    pub failure_count: u64,
+    pub last_tx_id: Option<String>,
+    pub last_note_id: Option<String>,
+    pub challenge_difficulty: Option<u64>,
+}
+
+/// Response from the faucet's `/pow` endpoint.
+#[derive(Debug, Deserialize)]
+struct PowChallengeResponse {
+    challenge: String,
+    target: u64,
+    #[allow(dead_code)] // Timestamp is part of API response but not used
+    timestamp: u64,
+}
+
+/// Response from the faucet's `/get_tokens` endpoint.
+#[derive(Debug, Deserialize)]
+struct GetTokensResponse {
+    tx_id: String,
+    note_id: String,
+}
+
+// FAUCET TEST TASK
+// ================================================================================================
+
+/// Runs a task that continuously tests faucet functionality and updates a watch channel.
+///
+/// This function spawns a task that periodically requests proof-of-work challenges from the faucet,
+/// solves them, and submits token requests to verify the faucet is operational.
+///
+/// # Arguments
+///
+/// * `faucet_url` - The URL of the faucet service to test.
+/// * `status_sender` - The sender for the watch channel.
+///
+/// # Returns
+///
+/// `Ok(())` if the task completes successfully, or an error if the task fails.
+#[instrument(target = COMPONENT, name = "faucet-test-task", skip_all)]
+pub async fn run_faucet_test_task(
+    faucet_url: Url,
+    status_sender: watch::Sender<ServiceStatus>,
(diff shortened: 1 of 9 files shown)


bin/network-monitor/src/main.rs
```diff
@@ -223,6 +228,65 @@ async fn initialize_prover_tasks(
...
+/// # Returns
+///
+/// An optional receiver for the faucet status.
+fn initialize_faucet_task(
+    config: &MonitorConfig,
+    tasks: &mut JoinSet<Result<(), anyhow::Error>>,
+    component_ids: &mut HashMap<Id, String>,
+) -> Option<Receiver<ServiceStatus>> {
+    if let Some(faucet_url) = &config.faucet_url {
+        let current_time = SystemTime::now()
```
Mirko-von-Leipzig:
```suggestion
    let Some(..) else {
        info!(...);
        return None;
    };
```

bin/network-monitor/src/main.rs
```diff
@@ -223,6 +228,65 @@ async fn initialize_prover_tasks(
...
+/// * `tasks` - The join set for the tasks.
+/// * `component_ids` - The component IDs.
+///
+/// # Returns
+///
+/// An optional receiver for the faucet status.
+fn initialize_faucet_task(
+    config: &MonitorConfig,
+    tasks: &mut JoinSet<Result<(), anyhow::Error>>,
+    component_ids: &mut HashMap<Id, String>,
```
Mirko-von-Leipzig, line 250: Future clean up could be to wrap the tasks + IDs in a struct, with impls ala

```rust
#[derive(Default)]
struct Tasks {
   handles: JoinSet<..>,
   names: HashMap<...>,
}

impl Tasks {
    fn spawn_faucet(&mut self, faucet_url: Url) -> anyhow::Result<()> {..}
}
```

bin/network-monitor/src/faucet.rs
```diff
@@ -0,0 +1,237 @@
...
+/// * `status_sender` - The sender for the watch channel.
+///
+/// # Returns
+///
+/// `Ok(())` if the task completes successfully, or an error if the task fails.
+#[instrument(target = COMPONENT, name = "faucet-test-task", skip_all)]
+pub async fn run_faucet_test_task(
+    faucet_url: Url,
+    status_sender: watch::Sender<ServiceStatus>,
+) -> anyhow::Result<()> {
```
Mirko-von-Leipzig, line 82: Do these tasks ever error? I imagine they could only ever panic due to a bug somewhere?

bin/network-monitor/src/faucet.rs
```diff
@@ -0,0 +1,237 @@
...
+            } else {
+                Status::Unhealthy
+            },
+            last_checked: current_time,
+            error: None,
+            details: ServiceDetails::FaucetTest(test_details),
+        };
+
+        // Send the status update (ignore if no receivers)
+        let _ = status_sender.send(status);
```
Mirko-von-Leipzig: Should we not return if there are no receivers? This could be used as a shutdown signal for example.

Mirko-von-Leipzig (approved):

bin/network-monitor/src/faucet.rs
SantiagoPittella: This task should never fail. As you said it should only be affected by a panic. The only **very** unlikely place where it can return an error is if the `current_time < UNIX_EPOCH`.

SantiagoPittella: But I used `anyhow::Result<()>` in all the spawn function to ease the failure handling.

bin/network-monitor/src/faucet.rs
```diff
@@ -0,0 +1,240 @@
...
+/// * `target` - The target value. A solution is valid if H(challenge, nonce) < target.
+///
+/// # Returns
+///
+/// The nonce that solves the challenge, or an error if no solution is found within reasonable
+/// bounds.
+fn solve_pow_challenge(challenge: &str, target: u64) -> anyhow::Result<u64> {
+    debug!("Solving PoW challenge: challenge={}, target={}", challenge, target);
+    // Try up to 10 million nonces - this should be reasonable for most difficulties
+    for nonce in 0..10_000_000u64 {
```
bobbinth: 10M attempts is ~23 bits of difficulty which should be solvable in about 10 seconds (though, depends a lot on the CPU). It is possible we go beyond that - I'd probably put this at 100M.

bin/network-monitor/src/faucet.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    // Step 2: Solve the PoW challenge
+    let nonce = solve_pow_challenge(&challenge_response.challenge, challenge_response.target)
+        .context("Failed to solve PoW challenge")?;
+
+    debug!("Solved PoW challenge with nonce: {}", nonce);
+
+    // Step 3: Request tokens with the solution
+    let tokens_url = faucet_url.join("/get_tokens")?;
+    let asset_amount = 1_000_000u64; // 1 token with 6 decimals
```
bobbinth: Could we create a constant for this? (this way it would be at the top of the file)

bin/network-monitor/src/faucet.rs
```diff
@@ -0,0 +1,240 @@
...
+#[derive(Debug, Deserialize)]
+struct GetTokensResponse {
+    tx_id: String,
+    note_id: String,
+}
+
+// FAUCET TEST TASK
+// ================================================================================================
+
+/// Runs a task that continuously tests faucet functionality and updates a watch channel.
```
bobbinth, line 65: Upon a quick look, I couldn't find how frequently we make the request to the faucet. We probably don't need to do this too frequently. Maybe once every 30 seconds is enough? Either way, I'd document this somewhere (unless it is already documented and I missed it).

bin/network-monitor/src/main.rs
SantiagoPittella: Created https://github.com/0xMiden/miden-node/issues/1241 to track this, as long with the clap addition

bin/network-monitor/src/faucet.rs
SantiagoPittella: In the last commit, i create constants for the before mentioned values, and included the frequency. I set it to 2 minutes to match the frequency of the tests and make sure that we are not disturbing the faucet's challenge difficulty. Also, in a follow up PR i'll add `clap` to the binary since configuration grew too large and management thorugh clap is easer, i plan to add this interval as a configuration option in that PR.

bin/network-monitor/src/faucet.rs
```diff
@@ -0,0 +1,249 @@
...
+        ])
+        .send()
+        .await?;
+
+    let response_text = response.text().await?;
+
+    let tokens_response: GetTokensResponse = serde_json::from_str(&response_text)
+        .with_context(|| format!("Failed to parse tokens response: {response_text}"))?;
+
+    Ok((tokens_response, challenge_response.target))
```
TomasArrachea: You could just use the leading zeros of the target to get an aproximate idea of the difficulty in bits, as @bobbinth suggested above.
```suggestion
    Ok((tokens_response, challenge_response.target.leading_zeros()))
```

PR #1244, opened 2025-09-26
chore: Add tx, block, and batch metrics

### LINKED ISSUE #1118: Additional telemetry metrics
A collection of metrics that would be useful to include.

- [ ] Raw block size (dumped to disk)
- [ ] Potentially raw block folder size
- [ ] Tx IDs in batch
- [ ] Batch IDs in block
- [ ] Tx IDs in block

sergerad:
## Context

There are additional transaction, batch, and block metrics we could be capturing / observing.

Closes #1118.

## Changes
- Add transaction ids to batch attributes
- Add batch ids and transaction ids to block attributes
- Instrument save_block() with block number and size

### DIFF
--- a/crates/block-producer/src/batch_builder/mod.rs
+++ b/crates/block-producer/src/batch_builder/mod.rs
@@ -277,6 +277,7 @@ impl BatchJob {
 
     #[instrument(target = COMPONENT, name = "batch_builder.commit_batch", skip_all)]
     async fn commit_batch(&self, batch: ProvenBatch) {
+        // Log the batch details.
         let transaction_ids = batch
             .transactions()
             .as_slice()
--- a/crates/block-producer/src/batch_builder/mod.rs
+++ b/crates/block-producer/src/batch_builder/mod.rs
@@ -327,7 +327,7 @@ impl TelemetryInjectorExt for SelectedBatch {
         Span::current().set_attribute("transactions.count", self.transactions.len());
         Span::current().set_attribute(
             "transactions.ids",
-            self.transactions()
+            self.transactions
                 .as_slice()
                 .iter()
                 .map(|tx_header| tx_header.id())
--- a/crates/block-producer/src/batch_builder/mod.rs
+++ b/crates/block-producer/src/batch_builder/mod.rs
@@ -277,21 +277,6 @@ impl BatchJob {
 
     #[instrument(target = COMPONENT, name = "batch_builder.commit_batch", skip_all)]
     async fn commit_batch(&self, batch: ProvenBatch) {
-        // Log the batch details.
-        let transaction_ids = batch
-            .transactions()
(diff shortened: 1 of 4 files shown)


crates/block-producer/src/batch_builder/mod.rs
Mirko-von-Leipzig, line 1: I would inject them as part of the root span -- we don't really want to use log events.

If you inspect the `inject_telemetry` impls you'll find that most of this info is already there. You only need to add `transactions.ids` there.

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig, line 1: As with the other builder comment, check the telemetry impls.

crates/store/src/blocks.rs
Mirko-von-Leipzig, line 1: I would instrument the save function, and add these as fields

crates/utils/src/tracing/span_ext.rs
```diff
@@ -15,6 +16,13 @@ pub trait ToValue {
     fn to_value(&self) -> Value;
 }
 
+impl<T: ToValue> ToValue for Vec<T> {
+    fn to_value(&self) -> Value {
+        let string_values = self.iter().map(|v| v.to_value().into()).collect::<Vec<StringValue>>();
```
Mirko-von-Leipzig, line 21: Since you're collecting internally, maybe we should use some `Iterator<Item = T>` type?

sergerad: Sorry I'm not following `Value::Array(Array)` is otel crate type and we can't control the variants of the sub `Array` which are all `Vec<T>`.

Mirko-von-Leipzig: I meant 

```rust
impl<T: ToValue> ToValue for Iterator<Item = T>
```

because right now you're often collecting e.g. the transactions into a `Vec<TransactionId>` and then here you're re-collecting for the array. We should be able to skip the first collect

sergerad:
```rust
impl<T: ToValue, I: Iterator<Item = T>> ToValue for I {
```

Causes conflicts because `I` is generic:
```
conflicting implementations of trait tracing::span_ext::ToValue for type std::time::Duration
upstream crates may add a new impl of trait std::iter::Iterator for type std::time::Duration in future versions
```

Mirko-von-Leipzig: Ah right.. Hmm. 

Another alternative is to add the method to the `SpanExt` trait lower down? But maybe this is overdoing it :)

```rust
fn set_attribute_iter(...)
```

Mirko-von-Leipzig (approved):

PR #1247, opened 2025-09-26
Migrate docs to docusaurus

Keinberger:


docs/external/src/_category_.json
Mirko-von-Leipzig, line 1: Should these `_cateogry_` directories be committed? Or are they generated?

Mirko-von-Leipzig (approved):
I think this is fine ito structuring :)

I haven't checked the content since its 90% just moving the files around.

Thanks!

bobbinth:
Thank you for doing this! A couple of questions:

- Would we still be able to build and show documentation locally via Github pages (i.e., https://0xmiden.github.io/miden-node) - or would this no longer be possible with Docusaurus?
- The diff is pretty big, though from what I can tell, this is mostly driven by `package-lock.json` file. And I guess it is a good idea to check it in, right?

docs/external/src/_category_.json
Keinberger: > Should these _cateogry_ directories be committed? Or are they generated?

These files are Docusaurus specific so that it correctly formats (sub)-directories. They are necessary and have to be committed.

Keinberger:
@bobbinth 

> - Would we still be able to build and show documentation locally via Github pages (i.e., https://0xmiden.github.io/miden-node) - or would this no longer be possible with Docusaurus?

Yes,that is possible. For this repo though, it's a little bit more nuanced. There are md files that are only kept for internal purposes, and can be viewed on the repo-specific Github pages URL. I moved those files into `docs/internal` folder and kept the mdbook architecture there. I also modified the book.yml workflow so that it deploys those files to Github pages.

As for the md files that were exported to the docs: I migrated those to Docusaurus syntax and they will be included in our "Miden Book".

Personally I think it would be good to also migrate the internal md files to use Docusaurus architecture & syntax, so that markdown files and snippets can be easily exported at any time. Nevertheless, we can do that as part of a subsequent PR so that we can finish the core migration for now. Here's what I said in the original PR description in regards to that:
> That said, I believe it would be better to migrate these files to Docusaurus syntax as well. This would allow us to integrate the internal docs into the external documentation in the future, while also ensuring that developers only need to work with a single, consistent syntax (Docusaurus Markdown) instead of switching between mdBook and Docusaurus.

> The diff is pretty big, though from what I can tell, this is mostly driven by package-lock.json file. And I guess it is a good idea to check it in, right?

Yes, I included the package-lock.json file on purpose so that we ensure that the local dev environment builds with the same packages across all engineers that are editing the docs. That way, we can catch bugs more easily and also ensure that everyone has and can view the same version.

Mirko-von-Leipzig (approved):
That seems fine. Only thing I find confusing is external vs internal because it really just depends on the context.

I would prefer `docs/dev` or `docs/contributors` to make things more obvious.

Is there a reason this repo triggers the workflow in the docs repo, instead of the docs repo doing that itself?

Keinberger:
@Mirko-von-Leipzig 

> That seems fine. Only thing I find confusing is external vs internal because it really just depends on the context.
> 
> I would prefer `docs/dev` or `docs/contributors` to make things more obvious.

I understand your point of view, and yes indeed it is correct to say that it depends on the context. Nevertheless, all repos that export docs (and have internal/dev ones) now use this syntax, i.e. having a docs folder, with an external and internal subfolder. We have found consensus on this in all other PRs, not sure if we could find consensus for changing the naming again to your proposal, hence why I think we should keep it as is.

My thinking behind the naming was that "external" refers to the external-facing Miden docs, whereas "internal"  is not exported to the Miden docs.

> Is there a reason this repo triggers the workflow in the docs repo, instead of the docs repo doing that itself?

Yes, there is. The miden-docs repository doesn't really have a way of knowing when changes happen in its source repositories. In the past we worked around this with a scheduled cron job, but that meant changes weren’t propagated right away. With the current setup, redeployment is triggered instantly whenever changes are merged into next, which removes the need for a cron job and ensures that updates are reflected immediately.

PR #1248, opened 2025-09-26
chore: update miden-base version

juan518munoz:
Update dependency to latest commit. Needed for https://github.com/0xMiden/miden-client/pull/1308

Needed for https://github.com/0xMiden/miden-client/pull/1339

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1513,7 +1513,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.0",
+ "socket2 0.5.10",
```
SantiagoPittella: what caused this downgrade?

Cargo.lock
```diff
@@ -3270,7 +3270,7 @@ dependencies = [
  "serde",
  "serde_yaml",
  "sfv",
- "socket2 0.6.0",
+ "socket2 0.5.10",
```
SantiagoPittella: Same here

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - [BREAKING] Replace `GetAccountProofs` with `GetAccountProof` in the public store API (#[1211](https://github.com/0xMiden/miden-node/pull/1211)).
 - Implemented storage map `DataStore` function ([#1226](https://github.com/0xMiden/miden-node/pull/1226)).
 - [BREAKING] Renamed `RemoteProverProxy` to `RemoteProverClient` ([#1236](https://github.com/0xMiden/miden-node/pull/1236)).
+- Updated `miden-base` dependency to latest version ([#1248](https://github.com/0xMiden/miden-node/pull/1248)).
```
SantiagoPittella: I'm not entirely sure that were adding this updates to the changelog, can you double check it please?

juan518munoz: No, we shouldn't. I've removed the changelog changes, and added a `no-changelog` label to this PR.

Cargo.lock
juan518munoz: Lockfile was updated with the command `cargo update miden-objects`, so that may have caused it. Although manually re-setting it back to it previous version causes no issues.

I have reverted back to the version from before (the higher one) so we reduce the diff size of this PR.

Cargo.lock
juan518munoz: Ditto other response.

sergerad (approved):

Issue #1245, opened 2025-09-26
Include mempool stats in block-producer's status


Mirko-von-Leipzig:
Include some mempool metrics in the [block producer's status endpoint](https://github.com/0xMiden/miden-node/blob/881576e2e3f3e3335c54bfa65f8bad9dfe5602d9/proto/proto/block_producer.proto#L15-L16).

This could include:

- number of inflight transactions, both
    - unbatched and
    - in total (including in batches)
- number of inflight batches
- ..maybe note counts? But this is maybe less important

The status endpoint _must not_ lock the mempool to obtain this information as this is an externally accessible endpoint. Instead the mempool can emit these stats using a watch channel.

Mirko-von-Leipzig:
This is blocked by the ongoing work in #1234 and its follow-up work.

PR #1250, opened 2025-09-28
feat(mempool): re-add mempool stats to telemetry

Mirko-von-Leipzig:
This is part of the mempool refactoring to support user batches, first begun in #1234 and is continuing on the `refactor/mempool` branch as base. This PR targets the outstanding telemetry work

- [ ] user batch support
- [x] re-adding the telemetry stuff
- [ ] re-adding mempool events
- [ ] mempool tests

---------------------------

This PR does three orthogonal things

1. Tighten `pub` bounds to `pub(super)` (drive-by refactor)
2. Re-add `mempool.inject_telemetry()`
3. Add events whenever a batch or transaction is removed

-----------------------------

(1) was mostly because I think previously there was some confusion on where these get used in practice. And this makes it more clear that state and nodes are the mempool's internal implementation details. This isn't important; can be pulled out if desired.

---------------------------

(2) is adds the state and node counts as stats to the OTel span data. There is some bike-shedding here on _what_ to expose. I've kept it minimal; feel free to ask for stuff but bear in mind that more complex stuff would require iterating through the nodes themselves. If so desired then we should probably make `Nodes` fields private to allow bookkeeping like total txs within all the nodes etc. The downside here is that its a bunch more code and it means iterating and mutating are less obviously trivial. But its possible.

^^ related to the above is #1245 -- we want similar stats so choose what you want and we can try make it happen.

-------------------------------

(3) I'm on the fence about. I've elected to emit an event for every batch, and every transaction that gets reverted or selected (due to block/batch failure or due to expiration). This would make tracking a single transaction/batch ID very easy; but will be noisy when the mempool gets large. An alternative would be adding a single property with _all_ batch IDs and _all_ tx IDs e.g. we would inject an array of IDs instead of an event per ID.

### DIFF
--- a/crates/block-producer/src/mempool/mod.rs
+++ b/crates/block-producer/src/mempool/mod.rs
@@ -281,6 +281,11 @@ impl Mempool {
             let node =
                 self.nodes.txs.remove(&tx.id()).expect("selected transaction node must exist");
             self.state.remove(&node);
+            tracing::info!(
+                batch.id = %batch_id,
+                transaction.id = %tx.id(),
+                "Transaction selected for inclusion in batch"
+            );
         }
         self.state.insert(NodeId::ProposedBatch(batch_id), &selected);
         self.nodes.proposed_batches.insert(batch_id, selected);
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig:
> Just wondering if any of those we would prefer to put as debug logs instead of info

There is close to zero value in debug logs since we never run in debug mode in prod :shrug: By the time you're interested in a particular transaction ID, the event would already be past.

PR #1251, opened 2025-09-28
feat(mempool): re-add event subscription

Mirko-von-Leipzig:
Part of the mempool refactoring, sibling to #1250, which are refactoring the mempool to support user-submitted-batches.

More specifically, of the follow-up work to #1234, this PR re-adds mempool events

- [ ] user batch support
- [ ] re-adding the telemetry stuff
- [x] re-adding mempool events
- [ ] mempool tests

-------------------

This PR re-adds event support equivalent to `next`; this will require additional changes to enable user batch support since now we have transactions **and** batches that get input, but this PR at least brings the refactor on par with production.

Reviewers can confirm this by comparing against the [current mempool impl](https://github.com/0xMiden/miden-node/blob/next/crates/block-producer/src/mempool/mod.rs) on `next` (searching for `self.subscription`).

### DIFF
--- a/crates/block-producer/src/mempool/mod.rs
+++ b/crates/block-producer/src/mempool/mod.rs
@@ -451,18 +451,18 @@ impl Mempool {
             .proposed_block
             .take_if(|(proposed, _)| proposed == &to_commit.block_num())
             .expect("block must be in progress to commit");
-        let txs = block.1.transactions().map(|tx| tx.id()).collect();
+        let tx_ids = block.1.transactions().map(|tx| tx.id()).collect();
 
         self.nodes.committed_blocks.push_back(block);
         self.chain_tip = self.chain_tip.child();
-        self.subscription.block_committed(to_commit, txs);
+        self.subscription.block_committed(to_commit, tx_ids);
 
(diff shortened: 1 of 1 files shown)


crates/block-producer/src/mempool/mod.rs
```diff
@@ -426,21 +428,24 @@ impl Mempool {
...
     #[instrument(target = COMPONENT, name = "mempool.commit_block", skip_all)]
-    pub fn commit_block(&mut self, to_commit: BlockNumber) {
+    pub fn commit_block(&mut self, to_commit: BlockHeader) {
         let block = self
             .nodes
             .proposed_block
-            .take_if(|(proposed, _)| proposed == &to_commit)
+            .take_if(|(proposed, _)| proposed == &to_commit.block_num())
             .expect("block must be in progress to commit");
+        let txs = block.1.transactions().map(|tx| tx.id()).collect();
```
sergerad: nit: would probably name these tx_ids

sergerad: afaict same with reverted_txs would do reverted_tx_ids

sergerad (approved):

PR #1252, opened 2025-09-29
ci: add a rust-toolchain.toml, MSRV ci & check script

huitseeker:
This is essentially a port of https://github.com/0xMiden/crypto/pull/547 on this repo. 

Copying the PR description from there.

----
This adds a CI job (locally runnable) that will check the MSRV of a PR is set correctly on a crate-by-crate basis. If the MSRV is set incorrectly, due to e.g. some feature being used, it will compute the correct MSRV for that crate and indicate it in its error. 

The script's error message can be emulated by manually changing the MSRV to an incorrect version (e.g. "1.87") and running it (`./scripts/check-msrv.sh`).
<details><summary> Sample output (click to unfold) </summary>

```
Checking MSRV for workspace members...
Checking miden-crypto in /Users/huitseeker/tmp/crypto/miden-crypto
   Current MSRV: 1.87
ERROR: MSRV check failed for miden-crypto
Searching for correct MSRV for miden-crypto...
   Found actual MSRV: 1.88.0

ERROR SUMMARY for miden-crypto:
   Package:   miden-crypto
   Directory: /Users/huitseeker/tmp/crypto/miden-crypto
   Current (incorrect) MSRV: 1.87
   Correct MSRV:             1.88.0

TO FIX:
   Update rust-version in /Users/huitseeker/tmp/crypto/miden-crypto/Cargo.
toml from "1.87" to "1.88.0"

   Or run this command (portable in-place edit):
     sed_i 's/^\s*rust-version\s*=\s*"1.87"/rust-version = "1.88.0"/' "/Us
ers/huitseeker/tmp/crypto/miden-crypto/Cargo.toml"
--------------------------------------------------------------------------
-----
```
</details>

This further adds a rust-toolchain file that will make CI tooling (and local rustup dev envs) follow the latest stable as it updates.

As a consequence, the following behaviors can change from being proactive to being on-demand:
- bumping the MSRV.

As a side effect, we can expect the MSRV setting to (over time) accurately reflect what we can compile with. Right now, the project or any of its dependents cannot be compiled with Rust 1.88 (and there is not technical reason to impose this restriction).

----

TL;DR:
- the file that drives toolchain bumps is `rust-toolchain.toml`, not the MSRV, so no need to bump that any more
- when using a feature that requires a higher MSRV the script (in CI or locally) helps figure out the adjustment necessary to optimally set the MSRV.

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -1,0 +1,33 @@
+name: Check MSRV
+
+on:
+  push:
+    branches: [next]
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+# Limits workflow concurrency to only the latest commit in the PR.
+concurrency:
+  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  # Check MSRV (aka `rust-version`) in `Cargo.toml` is valid for workspace members
+  msrv:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get install -y jq
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
+      - name: Install cargo-msrv
+        run: cargo install cargo-msrv
+      - name: Check MSRV for each workspace member
+        run: |
+          ./scripts/check-msrv.sh
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,5 @@
+[toolchain]
+channel    = "1.90"
+components = ["clippy", "rust-src", "rustfmt"]
+profile    = "minimal"
+targets    = ["wasm32-unknown-unknown"]
--- a/scripts/check-msrv.sh
+++ b/scripts/check-msrv.sh
@@ -1,0 +1,153 @@
+#!/bin/bash
+set -e
+set -o pipefail
+
+# Enhanced MSRV checking script for workspace repository
+# Checks MSRV for each workspace member and provides helpful error messages
+
+# ---- utilities --------------------------------------------------------------
+
+check_command() {
+  if ! command -v "$1" >/dev/null 2>&1; then
+    echo "ERROR: Required command '$1' is not installed or not in PATH"
+    exit 1
+  fi
+}
+
+# Check required commands
+check_command "cargo"
+check_command "jq"
+check_command "rustup"
+check_command "sed"
+check_command "grep"
+check_command "awk"
+
+# Portable in-place sed (GNU/macOS); usage: sed_i 's/foo/bar/' file
+# shellcheck disable=SC2329  # used quoted
+sed_i() {
+  if sed --version >/dev/null 2>&1; then
+    sed -i "$@"
+  else
+    sed -i '' "$@"
+  fi
+}
+
+# ---- repo root --------------------------------------------------------------
+
+# Get the directory where this script is located and change to the parent directory
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$DIR/.."
+
+echo "Checking MSRV for workspace members..."
+
+# ---- metadata --------------------------------------------------------------
+
+metadata_json="$(cargo metadata --no-deps --format-version 1)"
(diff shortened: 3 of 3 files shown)


.github/workflows/msrv.yml
```diff
@@ -0,0 +1,34 @@
...
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+      - name: Install cargo-msrv
+        run: cargo install cargo-msrv
+      - name: Cache rustup toolchains
+        run: rustup update
+      - name: Check MSRV for each workspace member
+        run: |
+          chmod +x scripts/check-msrv.sh
+          ./scripts/check-msrv.sh
```
Mirko-von-Leipzig: Why do we need something so complex? Can we not just run the verify command?

```suggestion
      - name: MSRV check
        run: cargo msrv verify
```

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel    = "1.90"
+components = ["clippy", "rust-src", "rustfmt"]
```
Mirko-von-Leipzig, line 3: My experience with this has usually been pretty bad. What inevitably happens is that I open a repo such as this; some random version of rust gets downloaded, and I don't get rust-analyzer support..

bobbinth: For the record, @huitseeker did not want to specify the exact version - but @bitwalker and I were in favor of it :)

One of the arguments is that it enforces that all of us are building the codebase with the same version of Rust.

bitwalker: @Mirko-von-Leipzig I know both Zed and VSCode allow configuring what rust-analyzer binary is used, both globally and project-local, should you need to. That said, I've never observed the presence of a `rust-toolchain.toml` with a specific toolchain breaking rust-analyzer support in my editor, and I use the default configuration in Zed for Rust.

Something to keep our eye on though for sure.

.github/workflows/msrv.yml
Mirko-von-Leipzig: I'm not sure why this was marked as resolved

huitseeker: I marked it as resolved, because the current state of the PR reflects code that is changed according to your suggestion.

Mirko-von-Leipzig: Maybe I wasn't specific enough, or there is a commit missing; but we are still running a large bash script and not a single `cargo msrv verify` command?

rust-toolchain.toml
Mirko-von-Leipzig: If you're using a globally set version then the toolchain version isn't really helping or no? If we want to develop with the toolchain version then specifically RA should be running that version?

Or does it do something more special and accounts for the toolchain version even though RA is running on version whatever?

.github/workflows/msrv.yml
huitseeker: Oh, I understand your question now. 

`cargo msrv` only works for one single crate and does not provide the for loop that iterates over every crate of the workspace (which is why we implement it). It support for workspaces is limited to understanding MSRV settings that defer to the workspace's settings (which is what we use in most repos).

Mirko-von-Leipzig: Ah I see. That's a bummer but makes sense.

.github/workflows/msrv.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  # Check MSRV (aka `rust-version`) in `Cargo.toml` is valid for workspace members
+  msrv:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get install -y jq
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
```
Mirko-von-Leipzig, line 26: We only cache on `next` because we regularly exceed github's 10GB cache limit otherwise.

https://github.com/0xMiden/miden-node/blob/6e34324243517d5836072e9940de3527a9c49e50/.github/workflows/lint.yml#L39-L41


```suggestion
[      - uses: Swatinem/rust-cache@v2](save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }})
```

.github/workflows/msrv.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Install cargo-msrv
+        run: cargo install cargo-msrv
+      - name: Cache rustup toolchains
+        run: rustup update
```
Mirko-von-Leipzig: Isn't this taken care of by the `dtolnay/rust-toolchain` step?

huitseeker: dtolnay's GH Action doesn't take the `rust-toolchain.toml` into account, but for the purpose of this action, this is irrelevant.

PR #1257, opened 2025-09-30
chore: add pagination to SyncNotes

### LINKED ISSUE #1256: Add pagination to `SyncNotes` endpoint
We should follow the same approach as the other sync endpoints which are paginated using `BLockRange` in the request to define the `from` and `to` blocks and it should return a `PaginationInfo` field with the chain tip and the block up to which it synced

SantiagoPittella:
closes #1256

### DIFF
--- a/crates/store/src/db/models/queries/mod.rs
+++ b/crates/store/src/db/models/queries/mod.rs
@@ -84,13 +84,17 @@ pub(crate) fn get_state_sync(
     from_start_block: BlockNumber,
     account_ids: Vec<AccountId>,
     note_tags: Vec<u32>,
+    chain_tip: BlockNumber,
 ) -> Result<StateSyncUpdate, StateSyncError> {
+    // Sync notes from the starting block to the latest in the chain.
+    let block_range = from_start_block..=chain_tip;
+
     // select notes since block by tag and sender
-    let notes = select_notes_since_block_by_tag_and_sender(
+    let (notes, _) = select_notes_since_block_by_tag_and_sender(
         conn,
-        from_start_block,
         &account_ids[..],
         &note_tags[..],
+        block_range,
     )?;
 
     // select block header by block num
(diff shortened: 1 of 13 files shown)


crates/store/src/server/rpc_api.rs
```diff
@@ -196,16 +196,25 @@ impl rpc_server::Rpc for StoreApi {
     ) -> Result<Response<proto::rpc_store::SyncNotesResponse>, Status> {
         let request = request.into_inner();
 
-        let (state, mmr_proof) = self
+        let chain_tip = self.state.latest_block_num().await;
+
+        let block_range = request.block_range.ok_or(invalid_argument("block_range is required"))?;
```
Mirko-von-Leipzig, line 199: I would put this first so that we avoid hitting the store if the query is invalid.

proto/proto/store/rpc.proto
```diff
@@ -309,18 +309,20 @@ message AccountVaultUpdate {
 // Specifies note tags that client is interested in. The server will return the first block which
 // contains a note matching `note_tags` or the chain tip.
 message SyncNotesRequest {
-    // Last block known by the client. The response will contain data starting from the next block,
-    // until the first block which contains a note of matching the requested tag.
-    fixed32 block_num = 1;
+    // Block range from which to start synchronizing.
+    //
+    // If the `block_to` is specified, this block must be close to the chain tip (i.e., within 30 blocks),
+    // otherwise an error will be returned.
```
bobbinth: This part is not correct, right? `block_to` could be anything for note syncing.

crates/store/src/db/models/queries/mod.rs
bobbinth, line 82: Should we mention the `chain_tip` here?

bobbinth (approved):
Looks good! Thank you! I left a couple of doc-related comments inline. Once these are addressed, we are good to merge.

proto/proto/store/rpc.proto
SantiagoPittella: Yes, does not applies here.

crates/store/src/db/models/queries/mod.rs
SantiagoPittella: Re-thinking it, I probably can get rid of this. Is is not needed since we can get it directly from the DB (as we do always before this function call), so it is probably for the better to remove the parameter.

PR #1260, opened 2025-09-30
chore: Add healthcheck layer to rpc (#1212)

sergerad:


sergerad:
Will make a PR to merge whole of main instead

PR #1264, opened 2025-10-01
fix: predefine otel.name field with empty value

### LINKED ISSUE #1263: gRPC span names aren't working
#1206 simplified the gRPC span names by using `otel.name = <span-name>` which in theory allows us to specify a dynamic span name for open-telemetry purposes. `tracing` does not support dynamic span names, only string literals.

Unfortunately we're not seeing these span names arrive in our honeycomb backend; instead the `tracing` placeholder `grpc_method` root span name is there. This seems to indicate that our `otel.name` solution isn't working even though the [documentation](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/index.html#special-fields) and issues indicate that it should.

Having this functionality work properly is quite essential to our monitoring so we need to figure this out.

I suggest investigating this by creating a small stand-alone binary and hooking it up to some otel collector to see the span name. i.e. create a span using tracing and then override it with `otel.name` and check that the collector actually see's this and the proceed from there.

I believe otel has a stdout logger type collector as well, which I imagine should do the right thing™️. Alternatively, hooking it up to jaeger or some other external otel compatible backend should also suffice.

SantiagoPittella:
fixes #1263

### DIFF
--- a/crates/utils/src/tracing/grpc.rs
+++ b/crates/utils/src/tracing/grpc.rs
@@ -10,18 +10,20 @@ use crate::tracing::OpenTelemetrySpanExt;
 /// Additionally also pulls in remote tracing context which allows the server trace to be connected
 /// to the client's origin trace.
 pub fn grpc_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
-    // Create a span with a generic, static name. Fields to be recorded after needs to be
-    // initialized as empty since otherwise the assignment will have no effect.
-    let span = tracing::info_span!("rpc", rpc.uri = %request.uri(), otel.name = field::Empty);
-
     // A gRPC request's path ends with `../<service>/<method>`.
     let mut path_segments = request.uri().path().rsplit('/');
 
     let service = path_segments.next().unwrap_or_default();
-    span.set_attribute("rpc.service", service);
-
     let method = path_segments.next().unwrap_or_default();
-    span.set_attribute("rpc.method", method);
+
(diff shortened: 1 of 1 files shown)


crates/utils/src/tracing/grpc.rs
```diff
@@ -8,8 +10,22 @@ use crate::tracing::OpenTelemetrySpanExt;
 /// Additionally also pulls in remote tracing context which allows the server trace to be connected
 /// to the client's origin trace.
 pub fn grpc_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
-    // Create a span with a generic, static name.
-    let span = tracing::info_span!("rpc", rpc.uri = %request.uri());
+    // Create a span with a generic, static name. Fields to be recorded after needs to be
+    // initialized as empty since otherwise the assignment will have no effect.
+    let span = tracing::info_span!("rpc", rpc.uri = %request.uri(), otel.name = field::Empty);
```
Mirko-von-Leipzig: Let's also preset the `rpc.service` and `rpc.method` fields like this? aka 

`span.record("rpc.service, service);`

sergerad: Feel free to revert my commit that replaced that logic with the simple uri. I used it to debug on devnet, not required anymore

sergerad: (as in we don't need to have the uri field anymore) I see the service/method logic is below here

PR #1266, opened 2025-10-02
wip: application level errors

### LINKED ISSUE #1106: Refactor Proto Messages for App-level Errors

SantiagoPittella:
closes #1106 

I moved the existing errors from the block producer to the `miden-node-proto` crate, and created the enum for other ones as defined in the original issue.

### DIFF
--- a/crates/proto/src/errors/block_producer.rs
+++ b/crates/proto/src/errors/block_producer.rs
@@ -1,0 +1,48 @@
+// Submit proven transaction error codes for gRPC Status::details
+// =================================================================================================
+
+use crate::errors::store::GrpcError;
+
+#[derive(Debug, Copy, Clone)]
+#[repr(u8)]
+pub enum SubmitProvenTransactionGrpcError {
+    Internal = 0,
+    DeserializationFailed = 1,
+    InvalidTransactionProof = 2,
+    IncorrectAccountInitialCommitment = 3,
+    InputNotesAlreadyConsumed = 4,
+    UnauthenticatedNotesNotFound = 5,
+    OutputNotesAlreadyExist = 6,
+    TransactionExpired = 7,
+}
+
+impl GrpcError for SubmitProvenTransactionGrpcError {
+    fn api_code(self) -> u8 {
+        self as u8
+    }
+
+    fn is_internal(&self) -> bool {
+        matches!(self, SubmitProvenTransactionGrpcError::Internal)
+    }
+}
+
+// Submit proven batch error codes for gRPC Status::details
+// =================================================================================================
+
+#[derive(Debug, Copy, Clone)]
+#[repr(u8)]
+pub enum SubmitProvenBatchGrpcError {
+    #[allow(dead_code)]
+    Internal = 0,
+    DeserializationFailed = 1,
+}
+
+impl GrpcError for SubmitProvenBatchGrpcError {
+    fn api_code(self) -> u8 {
+        self as u8
+    }
+
+    fn is_internal(&self) -> bool {
+        matches!(self, SubmitProvenBatchGrpcError::Internal)
+    }
+}
--- a/crates/proto/src/errors/mod.rs
+++ b/crates/proto/src/errors/mod.rs
@@ -8,6 +8,9 @@ use thiserror::Error;
 
 use crate::domain::note::NetworkNoteError;
 
+pub mod block_producer;
+pub mod store;
+
 #[derive(Debug, Error)]
 pub enum ConversionError {
     #[error("asset error")]
--- a/crates/proto/src/errors/store.rs
+++ b/crates/proto/src/errors/store.rs
@@ -1,0 +1,245 @@
+// SYNC ENDPOINTS
+// ================================================================================================
+
+// Common trait for all gRPC error enums
+pub trait GrpcError {
+    fn api_code(self) -> u8;
+    fn is_internal(&self) -> bool;
+
+    fn tonic_code(&self) -> tonic::Code {
+        if self.is_internal() {
+            tonic::Code::Internal
+        } else {
+            tonic::Code::InvalidArgument
+        }
+    }
+}
+
+// SYNC NULLIFIERS
+// ------------------------------------------------------------------------------------------------
+
+#[derive(Debug, Copy, Clone)]
+#[repr(u8)]
+pub enum SyncNullifiersGrpcError {
+    // Internal Server Error
+    Internal = 0,
+    // Malformed nullifier prefix
+    DeserializationFailed = 1,
+    // Invalid `block_from`/`block_to` parameters
+    InvalidBlockRange = 2,
+    // Unsupported prefix length (only `16` supported)
+    InvalidPrefixLength = 3,
+}
+
+impl GrpcError for SyncNullifiersGrpcError {
+    fn api_code(self) -> u8 {
+        self as u8
+    }
+
+    fn is_internal(&self) -> bool {
+        matches!(self, SyncNullifiersGrpcError::Internal)
+    }
+}
+
+// SYNC ACCOUNT VAULT
+// ------------------------------------------------------------------------------------------------
+
+#[derive(Debug, Copy, Clone)]
+#[repr(u8)]
+pub enum SyncAccountVaultGrpcError {
+    // Internal Server Error
+    Internal = 0,
+    // Malformed account ID
+    DeserializationFailed = 1,
+    // Invalid `block_from`/`block_to` parameters
+    InvalidBlockRange = 2,
+    // Account is not public (no vault sync)
+    AccountNotPublic = 3,
+}
+
+impl GrpcError for SyncAccountVaultGrpcError {
+    fn api_code(self) -> u8 {
+        self as u8
+    }
+
+    fn is_internal(&self) -> bool {
+        matches!(self, SyncAccountVaultGrpcError::Internal)
+    }
+}
+
+// SYNC NOTES
+// ------------------------------------------------------------------------------------------------
+
+#[derive(Debug, Copy, Clone)]
+#[repr(u8)]
+pub enum SyncNotesGrpcError {
+    // Internal Server Error
+    Internal = 0,
+    // Malformed note tags
+    DeserializationFailed = 1,
+    // Invalid `block_from`/`block_to` parameters
+    InvalidBlockRange = 2,
+    // Too many note tags in request
+    TooManyTags = 3,
+}
+
+impl GrpcError for SyncNotesGrpcError {
+    fn api_code(self) -> u8 {
+        self as u8
+    }
+
+    fn is_internal(&self) -> bool {
+        matches!(self, SyncNotesGrpcError::Internal)
+    }
+}
+
+// SYNC STORAGE MAPS
+// ------------------------------------------------------------------------------------------------
+
+#[derive(Debug, Copy, Clone)]
+#[repr(u8)]
+pub enum SyncStorageMapsGrpcError {
+    // Internal Server Error
+    Internal = 0,
+    // Malformed account ID
+    DeserializationFailed = 1,
+    // Invalid `block_from`/`block_to` parameters
+    InvalidBlockRange = 2,
+    // Account ID does not exist
+    AccountNotFound = 3,
+    // Account storage is not publicly accessible
+    AccountNotPublic = 4,
+}
+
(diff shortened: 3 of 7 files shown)


crates/proto/src/errors/store.rs
```diff
@@ -0,0 +1,245 @@
...
+    fn is_internal(&self) -> bool;
+
+    fn tonic_code(&self) -> tonic::Code {
+        if self.is_internal() {
+            tonic::Code::Internal
+        } else {
+            tonic::Code::InvalidArgument
+        }
+    }
+}
```
Mirko-von-Leipzig: This should probably live in the `mod.rs`

crates/proto/src/errors/store.rs
```diff
@@ -0,0 +1,245 @@
...
+
+impl GrpcError for SyncNullifiersGrpcError {
+    fn api_code(self) -> u8 {
+        self as u8
+    }
+
+    fn is_internal(&self) -> bool {
+        matches!(self, SyncNullifiersGrpcError::Internal)
+    }
```
Mirko-von-Leipzig: I think let's add `PartialEq, Eq` to the enum

crates/proto/src/errors/store.rs
Mirko-von-Leipzig, line 1: Should we create a `macro_rules` to create these enums?

```rust
grpc_error!(
    SyncNullifiersGrpcError {
        InvalidBlockRange,
        InvalidPrefixLength,
    }
);
```

which generates

```rust
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
#[repr(u8)]
pub enum SyncNullifiersGrpcError {
    Internal = 0,
    DeserializationFailed = 1,
    InvalidBlockRange = 2,
    InvalidPrefixLength = 3,
}

impl GrpcError for SyncNullifiersGrpcError {
    fn api_code(self) -> u8 {
        self as u8
    }

    fn is_internal(&self) -> bool {
        self == &Self::Internal
    }
}
```

crates/store/src/errors.rs
Mirko-von-Leipzig, line 1: It would be nice if we could ensure we are doing this sort of error handling correctly for the user-facing RPC endpoints.

This would require some further thought, but basically instead of a `Result<T, tonic::Status>` we would want `Result<T, E>` where `E` is our own error type that then gets converted to a `Status`. Doing this cleanly isn't straight-forward though.

Also ideal would be if this error handling was only done at the RPC component since that's the "user" access point. But I think we have to do it in the store and block-producer since they're the only ones with full error info.

This isn't really a concern for this PR; more just musings on my part (so you can ignore  this).

crates/proto/src/domain/block.rs
```diff
@@ -218,18 +219,38 @@ impl From<&FeeParameters> for proto::blockchain::FeeParameters {
...
+            self.block_to.map_or(*fallback, BlockNumber::from),
+        );
+
+        if block_range.start() > block_range.end() {
+            return Err(InvalidBlockRange::StartGreaterThanEnd {
+                start: *block_range.start(),
+                end: *block_range.end(),
+            }
+            .into());
+        }
```
Mirko-von-Leipzig, line 248: Should this be an error? Or can we just ignore this e.g. in Rust its just an empty range/iterator.

```rust
(5..0).len() == 0
```

crates/proto/src/errors/mod.rs
```diff
@@ -74,3 +77,65 @@ impl From<ConversionError> for tonic::Status {
...
+
+                tonic::Status::with_details(
+                    api_error.tonic_code(),
+                    message,
+                    vec![api_error.api_code()].into(),
+                )
+            }
+        }
+    };
+}
```
Mirko-von-Leipzig: This could be a method on `GrpcError` I think?

crates/proto/src/errors/mod.rs
```diff
@@ -74,3 +77,65 @@ impl From<ConversionError> for tonic::Status {
...
+            tonic::Code::InvalidArgument
+        }
+    }
+}
+
+// HELPER MACROS
+// ================================================================================================
+
+/// Implement the `GrpcError` trait for a given error type
+#[macro_export]
```
Mirko-von-Leipzig: Do we need to export this? Its internal to the crate iiuc. Though I always gets confused as to whether export means `pub` or just that its usable outside the module.

crates/store/README.md
Mirko-von-Leipzig, line 1: Should the error handling doc go into the `proto/readme.md` since that is the formal schema?

Mirko-von-Leipzig: As a meta note, I feel like it makes more sense for our internal crates to reference the schema docs rather than having their own?

crates/proto/src/errors/mod.rs
SantiagoPittella: Hmm not sure, `GrpcError` is implemented for this new `*GrpcError`s. While `into_tonic_status` is implemented for the higher-level errors, using the `value.as_report()` to enrich the status

crates/proto/src/errors/mod.rs
SantiagoPittella: The trait methods defined here are called from the `store` and `block-producer` crates, inside each domain error `Into<tonic::Status>` implementation, so we need to export it

crates/store/README.md
SantiagoPittella: For the second comment, should this be done for the endpoints as a whole? or just for the errors?

crates/proto/src/errors/store.rs
SantiagoPittella: Due to the variable length of variants for each enum, I kept the manual enum definition and added the macro for the trait implementation. This is what I did first:

```rust
#[allow(unused_macros)]
macro_rules! grpc_error {
    ($error:ident {
        $($field:ident),* $(,)?
    }) => {
        grpc_error!(@build_enum $error, 2, $($field),*);
    };
    
    // Build the enum with proper discriminant assignment
    (@build_enum $error:ident, $next:expr, $field:ident) => {
        #[derive(Debug, Clone, Copy, PartialEq, Eq)]
        #[repr(u8)]
        pub enum $error {
            // Internal Server Error
            Internal = 0,
            // Malformed data
            DeserializationFailed = 1,
            $field = $next,
        }

        impl GrpcError for $error {
            fn api_code(self) -> u8 {
                self as u8
            }

            fn is_internal(&self) -> bool {
                self == &Self::Internal
            }
        }
    };
    
    (@build_enum $error:ident, $next:expr, $field1:ident, $field2:ident) => {
        #[derive(Debug, Clone, Copy, PartialEq, Eq)]
        #[repr(u8)]
        pub enum $error {
            // Internal Server Error
            Internal = 0,
            // Malformed data
            DeserializationFailed = 1,
            $field1 = $next,
            $field2 = $next + 1,
        }

        impl GrpcError for $error {
            fn api_code(self) -> u8 {
                self as u8
            }

            fn is_internal(&self) -> bool {
                self == &Self::Internal
            }
        }
    };
```

And so on for more variants

crates/store/src/errors.rs
```diff
@@ -351,6 +373,292 @@ pub enum GetBatchInputsError {
...
+            CheckNullifiersError::DatabaseError(_) => CheckNullifiersGrpcError::Internal,
+            CheckNullifiersError::DeserializationFailed(_) => {
+                CheckNullifiersGrpcError::DeserializationFailed
+            },
+            CheckNullifiersError::TooManyNullifiers(..) => {
+                CheckNullifiersGrpcError::TooManyNullifiers
+            },
+        }
+    }
+}
```
sergerad: I'm really worried about the amount of boilerplate required to maintain these grpc errors. Is there any way we could avoid duplicating the error types for the purposes of grpc serialization?

For example instead of manually maintaining this
```rust
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
#[repr(u8)]
pub enum GetNoteScriptByRootGrpcError {
    // Internal Server Error
    Internal = 0,
    // Malformed script root
    DeserializationFailed = 1,
    // Script with given root doesn't exist
    ScriptNotFound = 2,
}
```

Could we create a macro that maps the original error enum to the grpc enum and implements the api_err() fn? The mappings all seem 1-to-1 for the variants apart from the internal ones - which would be doable with a macro. E.G.

```rust
#[derive(Debug, Error, GrpcError)]
pub enum GetNoteScriptByRootError {
    #[error("database error")]
    #[grpc(internal)]
```

drahnr: I can derive a pro-macro to make this annotation based if need be. I agree with serge that this is very quickly to become a frequent synchronization hassle.

SantiagoPittella: Sounds good, i'm on it.

SantiagoPittella: Did it an applied it in almost all the error, I could not use it in the `AddTransactionError` since we look for the internal variants in `VerifyTxError`. To use it there, we should flatten the variants of that error

crates/proto/src/errors/block_producer.rs
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    pub fn tonic_code(&self) -> tonic::Code {
+        if self.is_internal() {
+            tonic::Code::Internal
+        } else {
+            tonic::Code::InvalidArgument
+        }
+    }
+}
```
sergerad: Why couldn't we use the macro for this one? Anything we could change to be able to use it?

SantiagoPittella: Did it an applied it in almost all the error, I could not use it in the AddTransactionError since we look for the internal variants in VerifyTxError. To use it there, we should flatten the variants of that error

sergerad: Right, so the problem is here
```rust
#[derive(Debug, Error)]
pub enum AddTransactionError {
    #[error("transaction verification failed")]
    VerificationFailed(#[from] VerifyTxError),
```

I guess its not ideal that in order to use the macro we need flat error enums. I wonder how complex it would be to support the nesting in the macro. Probably be pretty unwieldy.

If it unblocks using the macro for such errors, I would opt to not worry about `internal` variants and just map them all 1-to-1. But I don't know if that unblocks anything.

crates/store/src/errors.rs
```diff
@@ -3,6 +3,8 @@ use std::io;
 
 use deadpool_sync::InteractError;
 use miden_node_proto::domain::account::NetworkAccountError;
+use miden_node_proto::domain::block::InvalidBlockRange;
+use miden_node_proto::errors::{ConversionError, GrpcError as GrpcErrorDerive};
```
sergerad: nit: why is the alias `GrpcErrorDerive` required?

SantiagoPittella: It is not needed, it was colliding with the trait name but it is now fixed.

crates/proto/src/errors/block_producer.rs
SantiagoPittella: Since with all other error enums we don´t have this problem of nested error where some variants are internal errors, I think that defining the explicit conversions only for this exceptions (only 1, this one ATM) is fine. We can track if this situation changes and more complex variants are added before changing the approach, wdyt?

drahnr: +1 to flattening.

SantiagoPittella: double checking, you mean to flatten the variants from `VerifyTxError` into `AddTransactionError`, right? @drahnr

sergerad: Yea I do think that is probably a reasonable requirement - any error returned via gRPC needs to be flattened.

PR #1267, opened 2025-10-03
chore: Reinstate #[from] for GrpcClientError

sergerad:
Fixes the logging issue in apply_block described in #1269

### DIFF
--- a/crates/block-producer/src/errors.rs
+++ b/crates/block-producer/src/errors.rs
@@ -317,7 +317,7 @@ pub enum StoreError {
     #[error("account Id prefix already exists: {0}")]
     DuplicateAccountIdPrefix(AccountId),
     #[error("gRPC client error")]
-    GrpcClientError(Box<tonic::Status>),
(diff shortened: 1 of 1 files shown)


crates/block-producer/src/errors.rs
```diff
@@ -317,7 +317,7 @@ pub enum StoreError {
     #[error("account Id prefix already exists: {0}")]
     DuplicateAccountIdPrefix(AccountId),
     #[error("gRPC client error")]
-    GrpcClientError(Box<tonic::Status>),
+    GrpcClientError(#[from] Box<tonic::Status>),
```
Mirko-von-Leipzig, line 320: Should the `impl From<Status>` lower down not be removed as well?

sergerad: Need it to map `tonic::Status` to `Box<tonic::Status>` for the purposes of `GrpcClientError(#[from] Box<tonic::Status>)`

PR #1271, opened 2025-10-03
feat(mempool): add tests

Mirko-von-Leipzig:
This PR re-enables previous mempool tests, and adds a bunch of additional state related checks.

This adds confidence to the refactor, however more coverage with more complex batches and transaction graphs would be a good. Its somewhat difficult to create these though, and the better idea is letting it run in production for a bit with integration tests etc.

I think in the future it would have been a good idea to run this as a shadow network i.e. replay all blocks from the current testnet through it, in order.

### DIFF
--- a/crates/block-producer/src/mempool/tests.rs
+++ b/crates/block-producer/src/mempool/tests.rs
@@ -163,7 +163,7 @@ fn block_commit_reverts_expired_txns() {
     uut.add_transaction(tx_to_revert).unwrap();
 
     // Commit the pending block which should revert the above tx.
-    let arb_header = BlockHeader::mock(0, None, None, &[], Word::empty());
+    let arb_header = BlockHeader::mock(block, None, None, &[], Word::empty());
     uut.commit_block(arb_header.clone());
     reference.commit_block(arb_header);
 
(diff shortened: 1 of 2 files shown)


sergerad (approved):

crates/block-producer/src/mempool/tests.rs
```diff
@@ -1,21 +1,34 @@
...
+
 impl Mempool {
-    fn for_tests() -> Self {
-        Self::new(
+    /// Returns an empty [`Mempool`] and a perfect clone intended for use as the Unit Under Test and
+    /// the reference instance.
+    ///
+    /// The clone is important as this guarantees that the internal _hash_ state is the same. This
+    /// is relevant for internal `HashMap`s which would otherwise give different iteration order
+    /// which in turn doesn't let different mempool instances give the same results.
```
drahnr, line 20: 👍

drahnr (approved):

crates/block-producer/src/mempool/tests/add_transaction.rs
```diff
@@ -0,0 +1,520 @@
...
+            .with_authentication_height((oldest_local - 2).into());
+        let tx = Arc::new(tx);
+        uut.add_transaction(tx).unwrap_err();
+    }
+
+    /// Ensures that we guard against authentication height from lying about imaginary blocks beyond
+    /// the chain tip. Since the authentication height is determined by the store, and this is
+    /// considered internal, we panic as this is completely abnormal.
+    #[test]
+    #[should_panic]
```
igamigo, line 170: nit: not many reasons why this could panic but I'd look into adding an `expected` field to this `should_panic`

PR #1273, opened 2025-10-06
chore: update `miden-base`

juan518munoz:
Update `miden-base` dependency to latest version.

Note: this PR currently points to the next tentative merge to `next` https://github.com/0xMiden/miden-base/pull/1973. Leaving as draft until those changes get pushed.

### DIFF
--- a/crates/proto/src/domain/merkle.rs
+++ b/crates/proto/src/domain/merkle.rs
@@ -200,8 +200,8 @@ impl TryFrom<proto::primitives::SmtOpening> for SmtProof {
 
 impl From<SmtProof> for proto::primitives::SmtOpening {
     fn from(proof: SmtProof) -> Self {
-        let (ref path, leaf) = proof.into_parts();
-        let path: SparseMerklePath = path.clone();
+        let (path, leaf) = proof.into_parts();
+
         Self {
             path: Some(path.into()),
             leaf: Some(leaf.into()),


bin/stress-test/src/seeding/mod.rs
```diff
@@ -410,7 +410,7 @@ fn create_consume_note_tx(
         block_ref.commitment(),
         fee_from_block(block_ref).unwrap(),
         u32::MAX.into(),
-        ExecutionProof::new(Proof::new_dummy(), HashFunction::default()),
+        ExecutionProof::new(Proof::new_dummy(), HashFunction::Blake3_192),
```
bobbinth: Could we use `ExecutionProof::new_dummy()` constructor here?

bin/stress-test/src/seeding/mod.rs
```diff
@@ -448,7 +448,7 @@ fn create_emit_note_tx(
         )
         .unwrap(),
         u32::MAX.into(),
-        ExecutionProof::new(Proof::new_dummy(), HashFunction::default()),
+        ExecutionProof::new(Proof::new_dummy(), HashFunction::Blake3_192),
```
bobbinth: Same question as above.

crates/proto/src/domain/account.rs
```diff
@@ -471,11 +472,12 @@ impl TryFrom<proto::account::AccountWitness> for AccountWitnessRecord {
 
 impl From<AccountWitnessRecord> for proto::account::AccountWitness {
     fn from(from: AccountWitnessRecord) -> Self {
+        let merkle_path: MerklePath = from.witness.clone().into_proof().into_parts().0.into();
         Self {
             account_id: Some(from.account_id.into()),
             witness_id: Some(from.witness.id().into()),
             commitment: Some(from.witness.state_commitment().into()),
-            path: Some(from.witness.into_proof().into_parts().0.into()),
+            path: Some(merkle_path.into()),
```
bobbinth: Could we not do just `from.witness.path().clone().into()` here?

crates/proto/src/domain/merkle.rs
```diff
@@ -193,13 +193,14 @@ impl TryFrom<proto::primitives::SmtOpening> for SmtProof {
             .ok_or(proto::primitives::SmtOpening::missing_field(stringify!(leaf)))?
             .try_into()?;
 
-        Ok(SmtProof::new(path, leaf)?)
+        Ok(SmtProof::new(path.try_into()?, leaf)?)
```
bobbinth: Would it make sense to modify the definition of `SmtOpening` to use `SparseMerklePath` instead of converting to/from `MerklePath`? Or does this result in some complex changes?

bobbinth (commented):
Looks good! Thank you! Left some small comments inline.

bin/stress-test/src/seeding/mod.rs
drahnr: `ExecutionProof::new_dummy` is guarded behind the testing feature, to my understanding we'd need to depend on it via
`miden-air  = { features = ["std", "testing"], version = "0.18" }` which should be fine for testing.

crates/store/src/genesis/config/mod.rs
```diff
@@ -341,7 +341,7 @@ impl FungibleFaucetConfig {
         } = self;
         let mut rng = ChaCha20Rng::from_seed(rand::random());
         let secret_key = SecretKey::with_rng(&mut get_rpo_random_coin(&mut rng));
-        let auth = AuthRpoFalcon512::new(secret_key.public_key());
+        let auth = AuthRpoFalcon512::new(secret_key.public_key().into());
```
drahnr, line 344: nit: I'd prefer the `.to_commitment()` call over `.into()` for explicitness

juan518munoz: `to_commitment()` converts into `Word`, but the argument must be of type `PublicKeyCommitment)`.

Best we could do for being more explicit is
```rs
let auth = AuthRpoFalcon512::new(secret_key.public_key().to_commitment().into());
```

drahnr: Mea culpa, you're correct, leave as is.

crates/proto/src/domain/merkle.rs
juan518munoz: Not complex at all (if I understood correctly), I have changed the proto definition to use `SparseMerklePath` in [df7cf5b](https://github.com/0xMiden/miden-node/pull/1273/commits/df7cf5b3c44d17a85e3995f1e31f7567023d3f06).

crates/proto/src/domain/account.rs
juan518munoz: Done in [df7cf5b](https://github.com/0xMiden/miden-node/pull/1273/commits/df7cf5b3c44d17a85e3995f1e31f7567023d3f06)

bin/stress-test/src/seeding/mod.rs
juan518munoz: Added the `testing` featured and replaced the line for:
```rs
ExecutionProof::new_dummy(),
```

[df7cf5b](https://github.com/0xMiden/miden-node/pull/1273/commits/df7cf5b3c44d17a85e3995f1e31f7567023d3f06)

bin/stress-test/src/seeding/mod.rs
juan518munoz: Addressed in [df7cf5b](https://github.com/0xMiden/miden-node/pull/1273/commits/df7cf5b3c44d17a85e3995f1e31f7567023d3f06).

bin/stress-test/Cargo.toml
```diff
@@ -19,17 +19,16 @@ workspace = true
...
-miden-air                 = { workspace = true }
 miden-block-prover        = { features = ["testing"], workspace = true }
 miden-lib                 = { workspace = true }
 miden-node-block-producer = { workspace = true }
 miden-node-proto          = { workspace = true }
 miden-node-store          = { workspace = true }
 miden-node-utils          = { workspace = true }
-miden-objects             = { features = ["testing"], workspace = true }
+miden-objects             = { workspace = true }
+miden-air                 = { features = ["testing"], workspace = true }
```
igamigo: nit: We should order crate dependencies lexicographically

crates/proto/src/domain/merkle.rs
```diff
@@ -183,24 +183,25 @@ impl TryFrom<proto::primitives::SmtOpening> for SmtProof {
...
-        Ok(SmtProof::new(path.try_into()?, leaf)?)
+        Ok(SmtProof::new(path, leaf)?)
     }
 }
 
 impl From<SmtProof> for proto::primitives::SmtOpening {
     fn from(proof: SmtProof) -> Self {
         let (ref path, leaf) = proof.into_parts();
-        let path: MerklePath = path.clone().into();
+        let path: SparseMerklePath = path.clone();
```
igamigo: Is this clone needed now? Seems like it is not

crates/proto/src/domain/account.rs
```diff
@@ -472,12 +471,11 @@ impl TryFrom<proto::account::AccountWitness> for AccountWitnessRecord {
 
 impl From<AccountWitnessRecord> for proto::account::AccountWitness {
     fn from(from: AccountWitnessRecord) -> Self {
-        let merkle_path: MerklePath = from.witness.clone().into_proof().into_parts().0.into();
         Self {
             account_id: Some(from.account_id.into()),
             witness_id: Some(from.witness.id().into()),
             commitment: Some(from.witness.state_commitment().into()),
-            path: Some(merkle_path.into()),
+            path: Some(from.witness.path().clone().into()),
```
igamigo, line 480: Since we are consuming the `AccountWitnessRecord`, would there be any way to avoid this clone?

crates/proto/src/domain/merkle.rs
juan518munoz: Function returns a reference to the value and not the value itself:
```
cannot move out of `*path` which is behind a shared reference
```

I think that we are forced to use clone here.

crates/proto/src/domain/account.rs
juan518munoz: Similar to the comment above, the `path()` function returns a reference to the `path` and not the value itself. We also can't access `path` directly as it is a private field.

bin/stress-test/Cargo.toml
juan518munoz: Forgot to run linting on toml files, it should be resolved now.

crates/proto/src/domain/merkle.rs
igamigo: I think we should be able to remove the `ref` keyword and do something like:

```rust
fn from(proof: SmtProof) -> Self {
    let (path, leaf) = proof.into_parts();

    Self {
        path: Some(path.into()),
        leaf: Some(leaf.into()),
    }
}
```

juan518munoz: I was completely skipping over the `ref` part 😅. I've removed it, thus avoiding the clone.

crates/proto/src/domain/merkle.rs
bobbinth, line 197: We could have probably avoided an extra `.clone()` here by doing something like this:
```rust
use proto::primitives::SmtOpening;
let SmtOpening { path, leaf } = opening;

let path = path.ok_or(SmtOpening::missing_field(stringify!(path)))?.try_into()?;
let leaf = leaf.ok_or(SmtOpening::missing_field(stringify!(leaf)))?.try_into()?;

Ok(SmtProof::new(path, leaf)?)
```

crates/proto/src/domain/account.rs
bobbinth: Maybe be good to add something like `AccountWitness::into_parts()` method in `miden-base`.

PR #1274, opened 2025-10-06
chore: improve grpc spans

SantiagoPittella:
> ! WARNING: this PR aims to `main`

Copies the improved grpc spans handling from `next` to `main`.


igamigo:
These are mainly changes from https://github.com/0xMiden/miden-node/pull/1230, right? If so, would it make more sense to cherry pick the commit instead of creating a new one? This way you'd avoid (I think) the conflicts and the merge back to `next` later

SantiagoPittella:
> These are mainly changes from https://github.com/0xMiden/miden-node/pull/1230, right? If so, would it make more sense to cherry pick the commit instead of creating a new one? This way you'd avoid (I think) the conflicts and the merge back to next later

Done!

sergerad (approved):

igamigo (approved):

PR #1276, opened 2025-10-07
chore: add accounts creation for network monitor

SantiagoPittella:
part of #1190 

Added the account creations and deployment in this first version.
I'm working on another PR in the periodic value increments, and i'm leaving authentication as the last part.

### DIFF
--- a/bin/network-monitor/src/deploy/mod.rs
+++ b/bin/network-monitor/src/deploy/mod.rs
@@ -11,8 +11,8 @@ use anyhow::{Context, Result};
 use miden_node_proto::clients::{Builder, Rpc, RpcClient};
 use miden_node_proto::generated::shared::BlockHeaderByNumberRequest;
 use miden_node_proto::generated::transaction::ProvenTransaction;
-use miden_objects::account::{Account, AccountId, PartialAccount, StorageSlot};
-use miden_objects::asset::{AssetVaultKey, AssetWitness};
+use miden_objects::account::{Account, AccountId, PartialAccount, PartialStorage};
+use miden_objects::asset::{AssetVaultKey, AssetWitness, PartialVault};
 use miden_objects::block::{BlockHeader, BlockNumber};
 use miden_objects::crypto::merkle::{MmrPeaks, PartialMmr};
 use miden_objects::note::NoteScript;
@@ -215,40 +215,28 @@ impl DataStore for MonitorDataStore {
             self.counter_account.lock().await.to_owned()
         };
 
-        Ok(((&account).into(), self.block_header.clone(), self.partial_block_chain.clone()))
+        let partial_storage = PartialStorage::new_full(account.storage().clone());
+        let assert_vault = PartialVault::new_full(account.vault().clone());
+        let partial_account = PartialAccount::new(
+            account_id,
+            account.nonce(),
+            account.code().clone(),
+            partial_storage,
+            assert_vault,
+            account.seed(),
+        )
+        .expect("Partial account be valid");
+
+        Ok((partial_account, self.block_header.clone(), self.partial_block_chain.clone()))
     }
 
     async fn get_storage_map_witness(
         &self,
-        account_id: AccountId,
-        map_root: Word,
-        map_key: Word,
+        _account_id: AccountId,
+        _map_root: Word,
+        _map_key: Word,
     ) -> Result<miden_objects::account::StorageMapWitness, DataStoreError> {
-        let account = self.wallet_account.lock().await;
-
-        let account = if account_id == account.id() {
-            account.to_owned()
-        } else {
-            self.counter_account.lock().await.to_owned()
-        };
-
-        let mut map_witness = None;
-        for slot in account.storage().slots() {
-            if let StorageSlot::Map(map) = slot {
-                if map.root() == map_root {
-                    map_witness = Some(map.open(&map_key));
-                }
-            }
-        }
-
-        if let Some(map_witness) = map_witness {
-            Ok(map_witness)
-        } else {
-            Err(DataStoreError::Other {
-                error_msg: "account storage does not contain the expected root".into(),
-                source: None,
-            })
-        }
+        unimplemented!("Not needed")
     }
 
     async fn get_foreign_account_inputs(


bin/network-monitor/src/cli/commands.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl Cli {
+    /// Execute the parsed command.
+    pub async fn execute(self) -> anyhow::Result<()> {
+        match self.command {
+            Command::Start(config) => start_monitor(config).await,
+            Command::DeployAccount(config) => deploy_accounts(config).await,
+        }
+    }
+}
```
sergerad, line 30: Have we considered making account deployment an automatic / innate aspect of the monitor app? Rather than having to do this manually, the monitor could detect whether the required account exists or not, create it and if all is successful, then start running the relevant monitoring task/reqeusts.

sergerad: There is also this to consider, which doesn't quite give us the ability to match on error variants coming from grpc responses. But I am now wondering if its worth the effort to have that functionality at all.

https://docs.rs/tonic-richer-error/latest/tonic_richer_error/struct.ErrorDetails.html

bin/network-monitor/src/assets/increment_counter.masm
```diff
@@ -0,0 +1,5 @@
+use.external_contract::counter_contract
+
+begin
+    call.counter_contract::increment
```
igamigo, line 8: If this is used as a transaction script on #1280 (should it be moved to that PR?), it will fail every time because the `get_sender` call only applies at the moment of note processing (and in general, any getter under `active_note`)

bin/network-monitor/src/assets/counter_program.masm
```diff
@@ -0,0 +1,60 @@
...
+    # => [AUTHORIZED_ACCOUNT_ID_SUFFIX, sender_id_suffix]
+
+    assert_eq.err="sender suffix not authorized"
+    # => []
+
+    push.0
+    # => [0]
+
+    exec.account::get_item
+    # => [count]
```
igamigo: nit: Some of these could be grouped for better readability:


```suggestion
    push.0 exec.account::get_item
    # => [count]
```

bin/network-monitor/src/cli/commands.rs
SantiagoPittella: Sounds good, I'll remove it as a command and add it implicitly while starting the monitor.

bin/network-monitor/src/assets/increment_counter.masm
SantiagoPittella: I removed the sender validations temporarily, since it was complicating things i wanted to first have a simpler version working and then add it after.
Since this change made the deployment to be easier, I merge #1280 here and now we have the complete workflow. I will now work on sending the increment counter requests and display it in the frontend.

bin/network-monitor/src/deploy/account.rs
```diff
@@ -0,0 +1,349 @@
...
+    file_path: &str,
+) -> Result<()> {
+    let auth_secret_key = AuthSecretKey::RpoFalcon512(secret_key.clone());
+    let account_file = AccountFile::new(account.clone(), vec![auth_secret_key]);
+    account_file.write(file_path)?;
+    Ok(())
+}
+
+/// Save counter program account to file.
+pub fn save_counter_account(account: &Account, file_path: &str) -> Result<()> {
```
sergerad: nit: would use PathBuf / Path instead of String / str

bin/network-monitor/src/config.rs
```diff
@@ -100,4 +97,22 @@ pub struct MonitorConfig {
...
+    pub wallet_file: String,
+
+    /// Path for the counter program account file.
+    #[arg(
+        long = "counter-file",
+        env = "MIDEN_MONITOR_COUNTER_FILE",
+        default_value = "counter_program.bin",
+        help = "Path where the counter program account will be saved"
+    )]
+    pub counter_file: String,
```
sergerad: This should probably be PathBuf. I presume we do that in other parts of the stack but haven't checked

SantiagoPittella: Done! Here and with the wallet file too.

bin/network-monitor/src/deploy/account.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    // Save accounts to files
+    save_wallet_account(&wallet_account, &secret_key, wallet_file)?;
+    save_counter_account(&counter_account, counter_file)?;
+
+    tracing::info!("Successfully created and saved account files");
+
+    Box::pin(deploy_accounts(&wallet_account, &counter_account, rpc_url)).await?;
+
+    Ok(())
```
sergerad: nit: can remove ? and Ok(())

bin/network-monitor/src/monitor/tasks.rs
```diff
@@ -121,22 +121,32 @@ impl Tasks {
...
+            let prover_test_rx = if matches!(proof_type, Some(ProofType::Transaction)) {
                 debug!("Starting transaction proof tests for prover: {}", name);
                 let payload = generate_prover_test_payload().await;
                 let (prover_test_tx, prover_test_rx) =
                     watch::channel(initial_prover_status.clone());
 
                 let prover_url_clone = prover_url.clone();
                 let name_clone = name.clone();
+                // SAFETY: we matched Some above
+                let proof_type = proof_type.unwrap();
```
sergerad: nit: would do `.expect("proof type is not none")` instead of safety comment

bin/network-monitor/src/assets/counter_program.masm
```diff
@@ -0,0 +1,32 @@
...
+
+use.miden::account
+
+use.std::sys
+
+# Increment function with note authentication
+# => []
+export.increment
+    push.0 exec.account::get_item
+    # => [count]
```
bobbinth: nit: `account::get_item` returns a full word (4 elements) - so, technically, the comment should be:
```
# => [count, 0, 0, 0]
```
Also, with the latest update to `miden-base`, `account::get_item` will become `active_account::get_item`.

bin/network-monitor/src/assets/counter_program.masm
```diff
@@ -0,0 +1,32 @@
...
+    
+    push.1 add
+    # => [count+1]
```
bobbinth, line 35: Similarly, the comment here should be:
```
# => [count+1, 0, 0, 0]
```

bin/network-monitor/src/assets/counter_program.masm
```diff
@@ -0,0 +1,32 @@
...
+
+    push.0 exec.account::set_item
+    # => []
```
bobbinth: `account::set_item` returns the previous value of the storage slot. So, technically, the comment should be:
```
# => [count, 0, 0, 0]
```

Also, with the latest update to `miden-base`, `account::set_item` will become `native_account::set_item`.

bin/network-monitor/src/assets/counter_program.masm
```diff
@@ -0,0 +1,32 @@
...
+
+    exec.sys::truncate_stack
```
bobbinth: nit: this is fine, but we should be able to use just a single `dropw` to make it work (since we need to drop only 4 items from the stack to make it empty).

bin/network-monitor/src/assets/counter_program.masm
```diff
@@ -0,0 +1,32 @@
...
+
+# Get the counter (no auth required)
+# => [count]
+export.get_count
+    push.0 exec.account::get_item
+    # => [count]
+
+    exec.sys::truncate_stack
+    # => [count]
+end
```
bobbinth, line 52: Similar comments as above here.

bin/network-monitor/src/assets/increment_counter.masm
bobbinth, line 1: Could we add a comment with a brief description to this file? For example, we should mention that this is a note script.

bin/network-monitor/src/assets/increment_counter.masm
bobbinth: I would maybe put the authentication logic into `counter_contract::increment` since we want the counter contract to decide which notes to accept (rather than for notes to decide which accounts can consume them). To do this, we'd need to store the "owner" account ID in the storage slot together with the counter as well, and then before we increment, check that `owner_id == note.sender`.

bin/network-monitor/src/commands/start.rs
```diff
@@ -0,0 +1,61 @@
...
+    info!("Loaded configuration: {:?}", config);
+
+    if config.enable_otel {
+        miden_node_utils::logging::setup_tracing(OpenTelemetry::Enabled)?;
+    } else {
+        miden_node_utils::logging::setup_tracing(OpenTelemetry::Disabled)?;
+    }
+
+    // Ensure accounts exist before starting monitoring tasks
+    Box::pin(ensure_accounts_exist(
```
bobbinth: Question: why do we need `Box::pin()` here? Maybe explain this briefly in the comment?

bin/network-monitor/src/deploy/account.rs
bobbinth, line 1: nit: could we add some section separators to this file?

Also, it feels like the contents of this file could be in `mod.rs` and we don't actually need `account.rs` file.

bin/network-monitor/src/deploy/account.rs
```diff
@@ -0,0 +1,355 @@
...
+    counter_file: &Path,
+    rpc_url: &Url,
+) -> Result<()> {
+    let wallet_exists = wallet_file.exists();
+    let counter_exists = counter_file.exists();
+
+    if wallet_exists && counter_exists {
+        tracing::info!("Account files already exist, skipping account creation");
+        return Ok(());
+    }
```
bobbinth: If the files exist, should we check if the on-chain account states match the ones in the files? And if they don't match maybe we should download the latest states from the chain and save them? (could also be done in a follow-up).

bin/network-monitor/src/config.rs
```diff
@@ -101,4 +99,22 @@ pub struct MonitorConfig {
...
+    pub wallet_file: PathBuf,
+
+    /// Path for the counter program account file.
+    #[arg(
+        long = "counter-file",
+        env = "MIDEN_MONITOR_COUNTER_FILE",
+        default_value = "counter_program.bin",
+        help = "Path where the counter program account will be saved"
+    )]
+    pub counter_file: PathBuf,
```
bobbinth: nit (feel free to ignore): I'd maybe name the parameters `wallet_filepath` and `counter_filepath`.

bin/network-monitor/src/deploy/account.rs
```diff
@@ -0,0 +1,355 @@
...
+    account_file.write(file_path)?;
+    Ok(())
+}
+
+/// Save counter program account to file.
+pub fn save_counter_account(account: &Account, file_path: &Path) -> Result<()> {
+    let account_file = AccountFile::new(account.clone(), vec![]);
+    account_file.write(file_path)?;
+    Ok(())
+}
```
bobbinth: Should these methods go into their respective modules - i.e., `counter.rs` and `wallet.rs`.

bin/network-monitor/README.md
```diff
@@ -35,6 +35,8 @@ miden-network-monitor --faucet-url http://localhost:8080 --enable-otel
 - `--status-check-interval`: Interval at which to check the status of the services (default: `3s`)
 - `--port, -p`: Web server port (default: `3000`)
 - `--enable-otel`: Enable OpenTelemetry tracing
+- `--wallet-file`: Path where the wallet account will be saved (default: `wallet_account.bin`)
+- `--counter-file`: Path where the counter program account will be saved (default: `counter_program.bin`)
```
bobbinth: nit: "will be saved" implies that we always save the files. Maybe something like "are located" would be better?

bin/network-monitor/README.md
```diff
@@ -50,23 +52,60 @@ If command-line arguments are not provided, the application falls back to enviro
 - `MIDEN_MONITOR_STATUS_CHECK_INTERVAL`: Interval at which to check the status of the services
 - `MIDEN_MONITOR_PORT`: Web server port
 - `MIDEN_MONITOR_ENABLE_OTEL`: Enable OpenTelemetry tracing
+- `MIDEN_MONITOR_WALLET_FILE`: Path where the wallet account will be saved
+- `MIDEN_MONITOR_COUNTER_FILE`: Path where the counter program account will be saved
```
bobbinth: Same nit as above.

bin/network-monitor/src/commands/start.rs
SantiagoPittella: I needed `Box::pin` in the first version that I implemented, I just removed it.

bin/network-monitor/src/deploy/account.rs
SantiagoPittella: I merged account.rs into mod.rs

bin/network-monitor/src/deploy/account.rs
SantiagoPittella: Sure!

bin/network-monitor/src/deploy/account.rs
SantiagoPittella: Yes, or save to file every time that we send a new tx with the account to increment the counter. The second is probably simpler but adds some overhead at every tx

SantiagoPittella:
> I would maybe put the authentication logic into counter_contract::increment since we want the counter contract to decide which notes to accept (rather than for notes to decide which accounts can consume them). To do this, we'd need to store the "owner" account ID in the storage slot together with the counter as well, and then before we increment, check that owner_id == note.sender.

Yes. I did exactly that in the first implementation that I did of the procedures, but ATM I'm using the increment to deploy the network account too, so it was incompatible due to the account ID check failing (the sender was the network account itself and not the wallet)

bin/network-monitor/src/deploy/counter.rs
```diff
@@ -0,0 +1,72 @@
...
+        export.auth__basic
+            exec.account::incr_nonce
+            drop
+        end
+    ";
+
+    let incr_nonce_auth =
+        AccountComponent::compile(INCR_NONCE_AUTH_CODE, TransactionKernel::assembler(), vec![])
+            .context("failed to compile increment nonce auth component")?
+            .with_supports_all_types();
```
igamigo: I wonder if `IncrNonce` component from the mock tools from `miden-testing` can be used here instead (I think they are pretty much equivalent).

bin/network-monitor/src/deploy/counter.rs
```diff
@@ -0,0 +1,72 @@
...
+        TransactionKernel::assembler(),
+        vec![StorageSlot::empty_value(), StorageSlot::Value(owner_word)],
+    )?
+    .with_supports_all_types();
+
+    const INCR_NONCE_AUTH_CODE: &str = r"
+        use.miden::account
+        use.std::sys
+
+        export.auth__basic
```
igamigo: nit: I'm not 100% sure, but I believe having two underscores is not required anymore (ie, `auth_basic` would work)

bin/network-monitor/src/deploy/counter.rs
```diff
@@ -0,0 +1,72 @@
...
+/// Create a counter program account with custom MASM script.
+#[instrument(target = COMPONENT, name = "create-counter-account", skip_all, ret(level = "debug"))]
+pub fn create_counter_account(owner_account_id: AccountId) -> Result<Account> {
+    // Load and customize the MASM script
+    let script =
+        include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/src/assets/counter_program.masm"));
+
+    // Compile the account code
+    let owner_felts: [Felt; 2] = owner_account_id.into();
+    let owner_word = Word::from([owner_felts[0], owner_felts[1], Felt::new(0), Felt::new(0)]);
```
igamigo: Based on how these will be used, would it be better to arrange them differently?


```suggestion
    let owner_word = Word::from([Felt::ZERO, Felt::ZERO, owner_felts[1], owner_felts[0]]);
```

bin/network-monitor/src/deploy/counter.rs
```diff
@@ -0,0 +1,72 @@
...
+    // Create the counter program account
+    let init_seed: [u8; 32] = rand::random();
+    let counter_program = AccountBuilder::new(init_seed)
+        .account_type(AccountType::RegularAccountUpdatableCode)
+        .storage_mode(AccountStorageMode::Network)
+        .with_component(account_code)
+        .with_auth_component(incr_nonce_auth)
+        .build()?;
+
+    Ok(counter_program)
```
igamigo: nit: "program" is a bit misleading, I've seen people name this "counter contract" or "counter account"

bin/network-monitor/src/assets/counter_program.masm
```diff
@@ -0,0 +1,51 @@
...
+    # Ensure the note sender matches the authorized wallet stored in slot 1.
+    push.0 exec.input_note::get_sender
+    # => [sender_prefix, sender_suffix]
+    push.1 exec.account::get_item
+    # => [sender_prefix, sender_suffix, owner_prefix, owner_suffix, 0, 0]
+    movup.3 drop
+    movup.2 drop
+    # => [sender_prefix, sender_suffix, owner_prefix, owner_suffix]
+    exec.account_id::is_equal
+    assert
```
igamigo: I think the ordering on the stack comments may not be correct here. Now that I see how the account is set up, I also think what I commented offline was not fully correct either. If the ordering of the owner word is as in [this other comment](https://github.com/0xMiden/miden-node/pull/1276/files#r2492370115), I believe you should be able to do something like:

```
    push.1 exec.account::get_item
    # => [owner_prefix, owner_suffix, 0, 0]

    exec.active_note::get_sender
    # => [sender_prefix, sender_suffix, owner_prefix, owner_suffix, 0, 0]

    exec.account_id::is_equal
    # => [are_equal, 0, 0]
    
    assert drop drop
```

(but maybe this is not fully correct either, I have not run it)

bin/network-monitor/src/deploy/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+    counter_filepath: &Path,
+    rpc_url: &Url,
+) -> Result<()> {
+    let wallet_exists = wallet_filepath.exists();
+    let counter_exists = counter_filepath.exists();
+
+    if wallet_exists && counter_exists {
+        tracing::info!("Account files already exist, skipping account creation");
+        return Ok(());
+    }
```
igamigo, line 67: I think this shouldn't matter too much outside of testing (where one may want to reset the state of at least one of the accounts), but if somehow only one of these were false, both accounts would be created again. Maybe this was intended but mentioning it just in case.

bin/network-monitor/README.md
```diff
@@ -137,6 +178,48 @@ The web dashboard provides a clean, responsive interface with the following feat
...
+- Uses custom MASM script with account ID-based authorization
+- Automatically created if not present
+
+### Account File Management
+
+The monitor automatically:
+1. Checks for existing account files on startup
+2. Creates new accounts if files don't exist
+3. Deploys accounts to the network via RPC
+4. Uses the specified file paths (default: `wallet_account.bin` and `counter_program.bin`)
```
igamigo: nit: When using `AccountFile`, we've used the `.mac` suffix for files. Also, I'd mention that the files are saved in the paths:
```suggestion
4. Saves the wallet and counter contract account in the specified file paths (default: `wallet_account.mac` and `counter_program.mac`)
```

bin/network-monitor/src/config.rs
igamigo, line 15: I was trying to test this locally and saw that the default port does not match with the node's default (altohugh in the env file it seems to be correct). Should it be changed here as well? 

```suggestion
const DEFAULT_RPC_URL: &str = "http://localhost:57291";
```

bin/network-monitor/src/deploy/counter.rs
SantiagoPittella: Yes! It is equivalent. I;m replacing it since we already use miden testing.

bin/network-monitor/src/deploy/counter.rs
SantiagoPittella: It was removed in the replacement with miden testing

bin/network-monitor/src/deploy/counter.rs
SantiagoPittella: Should it like this? It the prefix first and then the suffix

```suggestion
    let owner_word = Word::from([Felt::new(0), Felt::new(0), owner_felts[0], owner_felts[1]]);
```

SantiagoPittella: Oh no, you're right, i see why.

bin/network-monitor/src/deploy/mod.rs
SantiagoPittella: If only the wallet does not exists, the counter needs to be created again (since it is coupled to a wallet). The other way (wallet exists but not the counter) might be possible and should work

SantiagoPittella: Despite that it should work to deploy a new counter with an existing account, we probably want to use a new wallet to keep some consistency between accounts, wdyt? I can comment it in the check

bin/network-monitor/src/assets/increment_counter.masm
SantiagoPittella: I added it here, but fixed some small things in https://github.com/0xMiden/miden-node/pull/1295 where it is actually tested.

bin/network-monitor/src/config.rs
```diff
@@ -101,4 +99,22 @@ pub struct MonitorConfig {
         help = "Whether to enable OpenTelemetry"
     )]
     pub enable_otel: bool,
+
+    /// Path for the wallet account file.
+    #[arg(
+        long = "wallet-filepath",
+        env = "MIDEN_MONITOR_WALLET_FILEPATH",
+        default_value = "wallet_account.mac",
+        help = "Path where the wallet account will be saved"
```
bobbinth: nit: should probably be "Path where the wallet account is located".

bin/network-monitor/src/config.rs
```diff
@@ -101,4 +99,22 @@ pub struct MonitorConfig {
...
+    )]
+    pub wallet_filepath: PathBuf,
+
+    /// Path for the counter program account file.
+    #[arg(
+        long = "counter-filepath",
+        env = "MIDEN_MONITOR_COUNTER_FILEPATH",
+        default_value = "counter_program.mac",
+        help = "Path where the counter program account will be saved"
```
bobbinth: nit: should probably be "Path where the counter account is located".

bin/network-monitor/src/deploy/wallet.rs
```diff
@@ -0,0 +1,49 @@
...
+    AuthSecretKey,
+};
+use miden_objects::crypto::dsa::rpo_falcon512::SecretKey;
+use rand::{Rng, SeedableRng};
+use rand_chacha::ChaCha20Rng;
+
+/// Create a wallet account with `RpoFalcon512` authentication.
+///
+/// Returns the created account and the secret key for authentication.
+pub fn create_wallet_account() -> Result<(Account, SecretKey)> {
```
bobbinth, line 22: Question: why do we instrument `create_counter_account()` but not this function?

SantiagoPittella: It should be instrumented, i'm adding it in a new commit

bin/network-monitor/src/deploy/mod.rs
igamigo: Ah yes, thanks for the explanation. Makes sense! I think always re-creating both should be fine.

bin/network-monitor/src/deploy/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+        .parse_str(library_path, script, &source_manager)
+        .map_err(|e| anyhow::anyhow!("Failed to parse module: {e}"))?;
+
+    assembler
+        .clone()
+        .assemble_library([module])
+        .map_err(|e| anyhow::anyhow!("Failed to assemble library: {e}"))
+}
+
+// MONITOR DATA STORE
```
igamigo, line 167: nit: I'd probably move this to its own mod

bin/network-monitor/src/deploy/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+/// Deploy accounts to the network.
+///
+/// This function creates both a wallet account and a counter program account,
+/// then saves them to the specified files.
+#[instrument(target = COMPONENT, name = "deploy-accounts", skip_all, ret(level = "debug"))]
+pub async fn deploy_accounts(
+    wallet_account: &Account,
+    counter_account: &Account,
+    rpc_url: &Url,
+) -> Result<()> {
```
igamigo, line 94: I think part of the doc comment is outdated. Also, really only the counter account gets deployed to the network, right? So maybe it could be named `deploy_counter_contract`

bin/network-monitor/src/assets/deploy_counter.masm
```diff
@@ -0,0 +1,8 @@
+# Transaction script used during deployment to bump the counter contract nonce
+# without changing the counter value.
```
igamigo: I _think_ transaction scripts are not needed anymore for this. I'd try removing it and executing without a script.

bin/network-monitor/src/deploy/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+        if let Some(map_witness) = map_witness {
+            Ok(map_witness)
+        } else {
+            Err(DataStoreError::Other {
+                error_msg: "account storage does not contain the expected root".into(),
+                source: None,
+            })
+        }
```
igamigo: In `get_transaction_inputs` you are instantiating the partial accounts with `PartialAccount::from`. If you include all the account data (I think you'd need to use `PartialAccount::new`), I think you should be able to remove the implementations of these methods. I would test removing them. We can also keep them, but it's more maintenance overhead + more code to test.

bin/network-monitor/src/deploy/mod.rs
SantiagoPittella: It was already changed with `deploy_counter_account` in the follow up PR https://github.com/0xMiden/miden-node/pull/1295/files already

bin/network-monitor/src/assets/deploy_counter.masm
SantiagoPittella: Awesome, tested it and works. I remvoed the deploy tx script.

bin/network-monitor/src/deploy/mod.rs
SantiagoPittella: Removing the `get_storage_map_witness` worked when removing while using `PartialAccount::new`, that was not the case for `get_vault_asset_witness`.

bin/network-monitor/src/deploy/mod.rs
SantiagoPittella: Since it is not too large and only used in this same file, I think that leaving it here is fine. Maybe we can consider to move it if we start to use it somewhere else, wdyt?

bin/network-monitor/src/deploy/mod.rs
igamigo: Interesting, I wonder why that was? I don't think assets are involved in these flows.

bin/network-monitor/src/deploy/mod.rs
igamigo: Yeah, sounds good. Not a strong opinion bjut it felt like the executor side of things could be refactored into its own submodule, but I agree it's probably premature

bin/network-monitor/src/deploy/mod.rs
SantiagoPittella: I'll do further investigation after merging this PR 👍🏼

PR #1277, opened 2025-10-07
fix: more features 

drahnr:
Addresses the comment in https://github.com/0xMiden/miden-node/pull/1214#discussion_r2411603520

Does not implement the compression of multiple merkle paths, that'll be done by https://github.com/0xMiden/miden-node/pull/1158

### DIFF
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -3,15 +3,12 @@ use std::fmt::{Debug, Display, Formatter};
 use miden_node_utils::formatting::format_opt;
 use miden_objects::Word;
 use miden_objects::account::{
-    Account,
-    AccountHeader,
-    AccountId,
-    AccountStorageHeader,
-    StorageMap,
+    Account, AccountHeader, AccountId, AccountStorageHeader, StorageMap, StorageMapWitness,
     StorageSlotType,
 };
 use miden_objects::asset::{Asset, AssetVault};
 use miden_objects::block::{AccountWitness, BlockNumber};
+use miden_objects::crypto::merkle::SmtProof;
 use miden_objects::crypto::merkle::SparseMerklePath;
 use miden_objects::note::{NoteExecutionMode, NoteTag};
 use miden_objects::utils::{Deserializable, DeserializationError, Serializable};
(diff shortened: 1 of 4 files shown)


proto/proto/store/rpc.proto
```diff
@@ -209,8 +223,14 @@ message AccountStorageDetails {
...
 
-        // By default we provide all storage entries.
-        MapEntries entries = 3;
+        oneof map_entries {
+            // By default we provide all storage entries.
+            AllMapEntries all = 3;
+
+            // If only a subset is requested, we respond with proofs.
+            // TODO compress, see <https://github.com/0xMiden/miden-node/pull/1158>
+            MapEntriesWithProofs with_proofs = 4;
```
bobbinth: I was thinking for this PR we'd not implement this part yet. This could be done in a follow-up.

Basically, for now, we'd either return all map entries or none. For returning map entries with proofs, let's create a separate issue.

drahnr: We technically do have an issue for the optimization of using a partial Smt in https://github.com/0xMiden/miden-node/issues/617 , which is aligned with the implementation here. I don't quite see the downside to slicing it like this compared to the all-or-no-entries approach?

bobbinth: The main benefit of not adding `MapEntriesWithProofs` is much smaller diff which will let us merge this fix very quickly (and this would unblock updating the client to use the latest node).

Adding `MapEntriesWithProofs` is also a bit more complicated as we may want to use some heuristic to figure out when to set all entries and when to send proofs. For example, if a single key-value pair is requested, we could just send a single proof, even if the map has 100 entries in it. Though, the merits of this optimization are to be discussed.

drahnr: Reduced the impl scope.

crates/proto/src/domain/account.rs
```diff
@@ -424,31 +427,40 @@ impl AccountStorageMapDetails {
...
-        };
-        if exceeds {
-            Self::too_many_entries(slot_index)
-        } else {
-            Self::from_slot_data(slot_index, slot_data, storage_map)
+        match slot_data {
+            SlotData::All => {
+                // For "All" requests, check if total map size exceeds limit
+                // Fix off-by-one error: use skip() instead of nth() to check for entries >= limit
+                if storage_map.entries().nth(Self::MAX_RETURN_ENTRIES).is_some() {
```
bobbinth: Can we not use `StorageMap::num_entries()` here instead of `nth()`?

crates/proto/src/domain/account.rs
```diff
@@ -424,31 +427,40 @@ impl AccountStorageMapDetails {
...
-            },
-        };
+    fn from_all_entries(slot_index: u8, storage_map: &StorageMap) -> Self {
+        let map_entries = Vec::from_iter(storage_map.entries().map(|(k, v)| (*k, *v)));
         Self {
             slot_index,
             too_many_entries: false,
             map_entries,
         }
     }
```
bobbinth, line 447: Would it make sense to do a check for the number of entries in this function? For example:
```rust
fn from_all_entries(slot_index: u8, storage_map: &StorageMap) -> Self {
    let (too_many_entries, map_entries) = if storage_map.num_entries() > Self::MAX_RETURN_ENTRIES {
        (true, Vec::new())
    } else {
        (false, Vec::from_iter(storage_map.entries().map(|(k, v)| (*k, *v))))
    };

    Self { slot_index, too_many_entries, map_entries }
}
```

crates/proto/src/domain/account.rs
drahnr: At the time of opening the base version exposing the API wasn't merged yet, will adjust.

crates/proto/src/domain/account.rs
drahnr: Currently the `if`-condition is a the call-site rather than internal. I don't really have a preference.

Issue #613, opened 2025-01-13
Guardrails: Transaction Re-execution in the RPC


Mirko-von-Leipzig:
We want to (short term) loosen miden's privacy guarantees while we are still actively developing in a centralized fashion. Effectively, miden would have privacy only in the network sense and _not_ from the node operator.

Private transactions will include the transaction data when submitting to the node operator. This data will not appear on-chain (same as it is now), but would be kept separately by the node.

This would act as _guardrails_, and the intention is to drop this sometime before decentralization.

There are some benefits:

- less regulatory issues
- (private) account recovery
- better debugging and support
- trivial full chain history and therefore recovery from some critical failure

and some downsides:

- more work for the node - tx data must be verified, probably by re-execution, negating (some) of the benefits of private transactions
- less complete privacy

ito implementation, it makes sense to store this data completely separately from the current database. This would minimize any actual node IO with the data beyond writing it.

This will of course require changes to RPC and the client.

bobbinth:
One of the other main benefits is that we get to "double-check" the proofs and so any potential soundness bugs in the proving system would not cause catastrophic failures.

Mirko-von-Leipzig:
@igamigo wdyt of this as a starting issue? It touches RPC, storage and will require some integration with the client.

Might be a bit broad, but there is also no rush on this as users are unlikely to care :)

Could also be implemented in stages e.g. (a) update RPC protos & update client, (b) verify via re-execution, (c) figure out long-term storage.

igamigo:
> [@igamigo](https://github.com/igamigo) wdyt of this as a starting issue? It touches RPC, storage and will require some integration with the client.
> 
> Might be a bit broad, but there is also no rush on this as users are unlikely to care :)
> 
> Could also be implemented in stages e.g. (a) update RPC protos & update client, (b) verify via re-execution, (c) figure out long-term storage.

I agree that this sounds quite broad but looks like good one. After a brief chat with Bobbin we settled on tackling #673 first (while he also moves forward with #174). 
Maybe we could split this one up in more explicit tasks in the meantime to tackle next (although the stages you mention sound good to me). Depending on milestone priorities, of course.

sergerad:
What data does this refer to exactly?
>Private transactions will include the transaction data when submitting to the node operator.

Are there endpoints / structs that already contain these for the purposes of understanding what is missing in the private route (RPC and storage)?

sergerad:
[...]

PR #1278, opened 2025-10-08
feat: Update SubmitProvenTransaction RPC with tx replay data

sergerad:
wip

Relates to #613.


crates/ntx-builder/src/transaction.rs
```diff
@@ -116,16 +116,16 @@ impl NtxContext {
...
+            Box::pin(async move {
                 let data_store = NtxDataStore::new(account, chain_tip_header, chain_mmr);
 
                 let notes = notes.into_iter().map(Note::from).collect::<Vec<_>>();
                 let (successful, failed) = self.filter_notes(&data_store, notes).await?;
                 let executed = Box::pin(self.execute(&data_store, successful)).await?;
-                let proven = Box::pin(self.prove(executed)).await?;
-                self.submit(proven).await?;
+                let proven = Box::pin(self.prove(executed.clone())).await?;
+                Box::pin(self.submit(proven, executed)).await?;
```
sergerad: @bobbinth LMK if you have any better ideas for this. Might be worth updating base to have `impl From<ProvenTransaction> for ExecutedTransaction` which can be called deeper in the stack with a feature flag. That way we don't affect the `submit()` API or this part of the stack.

crates/block-producer/src/server/mod.rs
```diff
@@ -385,6 +395,8 @@ impl BlockProducerRpcServer {
                 block_height: block_height.as_u32(),
             }
         })
+
+        // TODO(serge): Store transaction_replay in tangentional store.
```
sergerad: @bobbinth I assume this is where we would simply store the additional tx metadata. Anything else to do apart from store it here?

crates/rpc/src/server/validator.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl TransactionInputsDataStore {
+    fn new(tx_inputs: TransactionInputs) -> Self {
+        let mast_store = TransactionMastStore::new();
+        mast_store.load_account_code(tx_inputs.account().code());
+        Self { tx_inputs, mast_store }
+    }
+}
+
+impl DataStore for TransactionInputsDataStore {
```
sergerad, line 56: @bobbinth IDK what the actual approach would be here. This is modelled around how `NtxDataStore` does it. Maybe we can pull that impl out of ntx-builder and re-use it here? Or maybe there is another way to be doing this. This is assuming we plan to re-execute transactions rather than just re-prove.

sergerad: Perhaps this or similar?

https://github.com/0xMiden/miden-client/blob/57e9d919d38ece367a75f3e2c592ae1be272543d/crates/rust-client/src/store/data_store.rs#L22

crates/rpc/src/server/api.rs
```diff
@@ -392,6 +398,32 @@ impl api_server::Api for RpcService {
...
+                Err(e) => {
+                    warn!(
+                        target = COMPONENT,
+                        tx_id = %tx.id().to_hex(),
+                        error = %e,
+                        "Transaction re-execution failed, but continuing with submission"
+                    );
+                },
+            }
+        }
```
sergerad, line 425: @bobbinth 
1. Do we want this to be blocking as it is now?
2. Do we want this to return an error if it fails? Or just log like it does now?

crates/rpc/src/server/api.rs
```diff
@@ -392,6 +398,32 @@ impl api_server::Api for RpcService {
             ))
         })?;
 
+        // If transaction inputs are provided, re-execute the transaction to validate it.
+        if let Some(tx_inputs_bytes) = &request.transaction_inputs {
+            // Deserialize the transaction inputs.
+            let tx_inputs = TransactionInputs::read_from_bytes(tx_inputs_bytes).map_err(|err| {
+                Status::invalid_argument(err.as_report_context("Invalid transaction inputs"))
+            })?;
```
drahnr, line 406: Do we want to bail here or just log/metrics/spantrace?

crates/rpc/src/server/api.rs
drahnr: Since this is opt in for debugging only, we probably want to create _bounded_ queue but do the re-execution async. This implies it's not going to block by default, but will eventually if we're under high load, mostly to avoid OOM as it would occur in the unbounded channel scenario. This would hint at 2. being logging only.

crates/rpc/src/server/api.rs
sergerad: This is definitely a genuine error situation (mismatched protobuf schema) so would want to bail yea

crates/rpc/src/server/api.rs
sergerad: This re-execution logic will move to the validator component and this RPC stack will talk to that via gRPC.

Having said that, I think the bounded queue and async processing logic would be the same for this PR and the followup work. Its just a question of what we want to put into this PR for now.

crates/rpc/src/server/validator.rs
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    fn get_foreign_account_inputs(
+        &self,
+        foreign_account_id: AccountId,
+        _ref_block: BlockNumber,
+    ) -> impl FutureMaybeSend<Result<AccountInputs, DataStoreError>> {
+        async move { Err(DataStoreError::AccountNotFound(foreign_account_id)) }
+    }
```
bobbinth, line 82: Returning an error is fine for now, but we'll need to figure out how to read `AccountInputs` for foreign accounts from `TransactionInputs` (the data should be in the `advice_inputs` - but it may not be simple to extract).

crates/rpc/src/server/validator.rs
```diff
@@ -0,0 +1,148 @@
...
+    fn get_vault_asset_witness(
+        &self,
+        account_id: AccountId,
+        vault_root: Word,
+        vault_key: Word,
+    ) -> impl FutureMaybeSend<Result<AssetWitness, DataStoreError>> {
+        async move {
+            if self.tx_inputs.account().id() != account_id {
+                return Err(DataStoreError::AccountNotFound(account_id));
+            }
```
bobbinth, line 93: Similar to the above comment, we should be able to serve the foreign account requests. Maybe the issue we need is about the ability to support foreign account invocation during re-execution.

crates/ntx-builder/src/block_producer.rs
```diff
@@ -49,7 +49,10 @@ impl BlockProducerClient {
         &self,
         proven_tx: ProvenTransaction,
     ) -> Result<(), Status> {
-        let request = proto::transaction::ProvenTransaction { transaction: proven_tx.to_bytes() };
+        let request = proto::transaction::ProvenTransaction {
+            transaction: proven_tx.to_bytes(),
+            transaction_inputs: None,
+        };
```
bobbinth, line 55: This would mean that the validator would have no knowledge of the transactions created by the NTB, right? This is fine for now, but once we implement block signing, this will become a problem. So, we'll need to figure out how does NTB communicate its transactions to the validator.

crates/rpc/src/server/api.rs
```diff
@@ -392,6 +398,32 @@ impl api_server::Api for RpcService {
             ))
         })?;
 
+        // If transaction inputs are provided, re-execute the transaction to validate it.
+        if let Some(tx_inputs_bytes) = &request.transaction_inputs {
```
bobbinth, line 402: Would it make sense to move the body of this if-statement into a separate helper function?

crates/rpc/src/server/validator.rs
sergerad: https://github.com/0xMiden/miden-node/issues/1305

crates/ntx-builder/src/block_producer.rs
sergerad: Could ntx-builder submit proven txns via RPC instead of via block-producer?

crates/rpc/src/server/api.rs
sergerad: I could improve the separation but it would be throw-away code given this logic is going to move behind the validator RPC endpoint. So would prefer to keep as is for this PR.

crates/rpc/src/server/validator.rs
sergerad: Put it in the same issue https://github.com/0xMiden/miden-node/issues/1305

crates/ntx-builder/src/block_producer.rs
sergerad: Or maybe this is a reason why @mmagician's idea of pushing tx execution at the block signing stage makes sense? That way we could have a single validation flow for all transactions. Given the ntx builder talks directly to block-producer rather than RPC where we have tx re-execution going on in this PR.

Also, given we don't want to block proven transaction submission on validation (re-execution), I don't see the value of doing it in the submission part of the flow.

I think the value of keeping validator db separate to the rest of the node might be outweighed by these other factors. It might make sense to have all transactions reach block producer with the tx inputs instead of the flow we have discussed so far.

bobbinth: The options have different pros and cons:

Option 1

This is the current approach - i.e., we send the transactions to the validator ahead of time - the main benefits are:

- We don't need to execute transactions at the time of block validation, reducing validation latency. One question is whether this latency reduction is meaningful. For a small number of transactions it is probably not, but if we have 1K transactions to execute, and each of them takes 5 ms, even with parallelization it'll take a couple hundred ms - which could be significant (unless the validator will have a large number of cores).
- The logic of getting transactions to the validator is pretty simple:
  - We forward the transaction from the RPC, before we send it to the block producer.
  - We forward the transaction from the NTB, before we send it to the block producer.

The main downside is that the validator would potentially execute some "extra" transactions (e.g., transactions that make it to the RPC but either don't get accepted into the mempool or never make it into a block).

Option 2

This is the approach suggested by @mmagician - i.e., we send the transactions to the validator together with the block header, right before sending the block to the store.

The advantage here is that there is only one endpoint for the node to interact with the validator and we send only the transactions that made it into the block.

The downsides are:
- Extra latency required to execute the transactions during validation.
- More complex handling of transactions in the mempool as we'll need to carry the transaction replay data through all stages of block production - though, maybe with the latest refactoring of the mempool this may not be too difficult.

I still favor the first option as I think latency reduction is something we'll need to care about eventually anyways, and from the complexity standpoint, both options seem roughly similar to me.

> I think the value of keeping validator db separate to the rest of the node might be outweighed by these other factors.

Validator DB needs to be separate from the node because validator will not be run by the same entity as the node. So, the validator should not know anything about the node store and the node store should not know anything about the validator.

sergerad: @bobbinth I have updated ntx builder to use rpc instead of block producer for submission. Tested locally
https://github.com/0xMiden/miden-node/pull/1278/commits/3ca47076a254dc71b44beaa0b843d170c39804e5

bobbinth: Without having looked at the code, what's the benefit of going through the RPC rather than sending the transaction directly to the validator from the NTB?

sergerad: Just technically the validator component doesn't exist yet. I could block this PR on the other one that adds the validator component and then point NTB to validator?

bobbinth: We could also merge the state before https://github.com/0xMiden/miden-node/commit/3ca47076a254dc71b44beaa0b843d170c39804e5 and then update this part later on (this was the intent of my original comment).

More generally though, maybe sending transactions from NTB to the block producer via the RPC is the right approach - I just haven't thought through pros/cons yet.

sergerad: OK I have made an issue for NTB followup https://github.com/0xMiden/miden-node/issues/1309

PR #1281, opened 2025-10-09
bug: remove tweak pingora http configs

SantiagoPittella:
Tweak proxy's http configurations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.11.3 (2025-10-09)
+
+- Tweaked HTTP configurations on the pingora proxy server ([#1281](https://github.com/0xMiden/miden-node/pull/1281)).
+
 ## v0.11.2 (2025-09-10)
 
 - Added support for keepalive requests against base path `/` of RPC server ([#1212](https://github.com/0xMiden/miden-node/pull/1212)).


bin/remote-prover/src/api/prover.rs
```diff
@@ -215,34 +289,44 @@ impl ProverApi for ProverRpcApi {
...
                 self.prove_tx(tx_witness, &request_id).await
             },
             proto::remote_prover::ProofType::Batch => {
-                let proposed_batch = proof_request.try_into().map_err(invalid_argument)?;
+                let proposed_batch = proof_request.try_into().map_err(|err: DeserializationError| {
+                    let err_report = err.as_report();
+                    error!(target: COMPONENT, request_id = %request_id, proof_type = "batch", error = %err_report, "Failed to parse proposed batch");
+                    tracing::Span::current().record("error_type", "invalid_argument");
+                    tracing::Span::current().record("error_context", &err_report);
+                    Status::invalid_argument(err_report)
```
sergerad: I'm not sure these changes would add any value because the underlying error report should already be captured given, for example:
```
fn invalid_argument<E: ErrorReport>(err: E) -> Status {
    Status::invalid_argument(err.as_report())
}
```

If anything, we could consider adding `tracing::debug!("proving batch...")` which we can enable when we need to debug the networks.

SantiagoPittella: Makes sense. Probably those debug messages are not of much help either, since we can infer that from the span's name.

Ill send just the http config change

PR #1283, opened 2025-10-09
fix: stress test seed

### LINKED ISSUE #1282: Bench store binary is not working
When running:
```
miden-node-stress-test seed-store --data-directory test-data --num-accounts 10000 --public-accounts-percentage 50
```

I'm getting this error logs while applying the block, and the store is panicking:
```
2025-10-09T16:31:04.420832Z ERROR store.block_producer_server.apply_block: miden-store: crates/store/src/server/block_producer.rs:49: error: status: InvalidArgument, message: "block deserialization error: \ncaused by: account ID seed was provided for an existing account", details: [], metadata: MetadataMap { headers: {} }
2025-10-09T16:31:04.420849Z  INFO store.block_producer_server.apply_block: miden-store: crates/store/src/server/block_producer.rs:49: close, time.busy: 265µs, time.idle: 916ns
2025-10-09T16:31:04.420862Z ERROR rpc: tower_http::trace::on_failure: /Users/spittella/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tower-http-0.6.6/src/trace/on_failure.rs:93: response failed, classification: Code: 3, latency: 0 ms rpc.service: "block_producer_store.BlockProducer", rpc.method: "ApplyBlock", otel.name: "block_producer_store.BlockProducer/ApplyBlock"
2025-10-09T16:31:04.420880Z  INFO rpc: miden_node_utils::tracing::grpc: crates/utils/src/tracing/grpc.rs:21: close, time.busy: 345µs, time.idle: 340µs rpc.service: "block_producer_store.BlockProducer", rpc.method: "ApplyBlock", otel.name: "block_producer_store.BlockProducer/ApplyBlock"
2025-10-09T16:31:04.420980Z ERROR store.client.apply_block: miden-block-producer: crates/block-producer/src/store/mod.rs:243: error: gRPC client error
2025-10-09T16:31:04.420989Z  INFO store.client.apply_block: miden-block-producer: crates/block-producer/src/store/mod.rs:243: close, time.busy: 1.72ms, time.idle: 1.07ms
2025-10-09T16:31:04.420997Z ERROR miden_node_utils::logging: crates/utils/src/logging.rs:61: panicked at bin/stress-test/src/seeding/mod.rs:246:51:
called `Result::unwrap()` on an `Err` value: GrpcClientError(Status { code: InvalidArgument, message: "block deserialization error: \ncaused by: account ID seed was provided for an existing account", metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Thu, 09 Oct 2025 16:31:04 GMT"} }, source: None }), panic: true
```

SantiagoPittella:
closes #1282

### DIFF
--- a/bin/stress-test/src/seeding/mod.rs
+++ b/bin/stress-test/src/seeding/mod.rs
@@ -391,6 +391,8 @@ fn create_consume_note_tx(
         account.vault_mut().add_asset(*asset).unwrap();
     });
 
+    account.increment_nonce(ONE).unwrap();
+
     let details = if account.is_public() {
         AccountUpdateDetails::New(account.clone())
     } else {


bin/stress-test/src/seeding/mod.rs
```diff
@@ -391,8 +391,6 @@ fn create_consume_note_tx(
         account.vault_mut().add_asset(*asset).unwrap();
     });
 
-    account.increment_nonce(ONE).unwrap();
```
igamigo, line 394: Should this stay but after wrapping the account in `AccountUpdateDetails::new()`? AFAIK if this is a consume transaction (and is the first for the account), the nonce would go from zero to one so the `ProvenTransaction::final_account_commitment` should reflect this. I think in this case the binary might not be erroring because `account.commitment()` will be the same as `init_hash`, but we'd ideally want it to change. But maybe I'm misunderstanding the issue. 

Also, not related to this PR but I wonder if `increment_nonce` should get rid of the account seed internally.

EDIT: Actually maybe it should be the other way around and the seed should be stripped but this line should be kept?

bobbinth: cc @PhilippGackstatter

PhilippGackstatter: > Also, not related to this PR but I wonder if `increment_nonce` should get rid of the account seed internally.

Oh thanks, that's a bug in `Account::increment_nonce`. I.e. incrementing the nonce by `nonce_delta >= 1` should strip the seed to keep consistency, but I've added that only in `apply_delta`.

Opened https://github.com/0xMiden/miden-base/pull/2003.

bobbinth: https://github.com/0xMiden/miden-base/pull/2003 has now been merged. Do we need to update this PR?

SantiagoPittella: I've put this PR up-to-date and reintroduced the nonce incrementation (which removes the seed 🥳 )

igamigo (approved):

bobbinth (approved):
Looks good! Thank you!

PR #1284, opened 2025-10-09
Mempool refactoring

bobbinth:
This is a tracking PR for mempool refactoring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 - [BREAKING] Renamed `CheckNullifiersByPrefix` endpoint to `SyncNullifiers` ([#1191](https://github.com/0xMiden/miden-node/pull/1191)).
 - [BREAKING] Added `block_range` and `pagination_info` fields to paginated gRPC endpoints ([#1205](https://github.com/0xMiden/miden-node/pull/1205)).
 - Use `tonic` error codes for gRPC errors ([#1208](https://github.com/0xMiden/miden-node/pull/1208)).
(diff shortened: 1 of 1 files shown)


crates/block-producer/src/mempool/tests/add_transaction.rs
```diff
@@ -0,0 +1,520 @@
...
+        Word::new([300u32.into(), 1u32.into(), 2u32.into(), 3u32.into()]),
+    )
+    .private_notes_created_range(1..2)
+    .build();
+
+    let parent_c = MockProvenTxBuilder::with_account(
+        mock_account_id(3),
+        Word::empty(),
+        Word::new([4500u32.into(), 1u32.into(), 2u32.into(), 3u32.into()]),
+    )
```
drahnr, line 36: nit: `impl From<[u32; WORD_SIZE_FELT]> for Word {` exists, so we could replace this with `Word::from([4500u32, 1, 2, 3]);`

crates/block-producer/src/mempool/nodes.rs
```diff
@@ -0,0 +1,313 @@
...
+
+use crate::domain::transaction::AuthenticatedTransaction;
+
+/// Uniquely identifies a node in the mempool.
+///
+/// This effectively describes the lifecycle of a transaction in the mempool.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub(super) enum NodeId {
+    Transaction(TransactionId),
+    // UserBatch(BatchId),
```
drahnr, line 19: Is this going to be added in a follow up I presume?

bobbinth:
I've merged this branch. @igamigo - could you guys update the client to use the latest `next` and then run the integration tests when you get a chance?

PR #1288, opened 2025-10-13
chore: Update base and remove TransactionWitness

sergerad:
## Context

In https://github.com/0xMiden/miden-base/pull/1934 we refactored `TransactionInputs` and `ExecutedTransaction` so that they contained more fields. This meant we were able to remove `TransactionWitness` altogether.

## Changes

Update miden-base and refactor miden-node to use TransactionInputs` instead of `TransactionWitness` for proving calls.

### DIFF
--- a/crates/ntx-builder/src/transaction.rs
+++ b/crates/ntx-builder/src/transaction.rs
@@ -355,6 +355,7 @@ impl DataStore for NtxDataStore {
         &self,
         script_root: Word,
     ) -> impl FutureMaybeSend<Result<miden_objects::note::NoteScript, DataStoreError>> {
+        // TODO: Add implementation for getting note script from NtxDataStore.
         async move { Err(DataStoreError::NoteScriptNotFound(script_root)) }
     }
 }


crates/ntx-builder/src/transaction.rs
```diff
@@ -349,6 +349,13 @@ impl DataStore for NtxDataStore {
             }
         }
     }
+
+    fn get_note_script(
+        &self,
+        _script_root: Word,
+    ) -> impl FutureMaybeSend<Result<miden_objects::note::NoteScript, DataStoreError>> {
+        async move { todo!() }
```
sergerad: @bobbinth any advice on this? Is it fine to leave `unimplemented!()` for ntx builder?

bobbinth: Yes, fine to keep unimplemented for this PR, but we'll need to implement it soon.

crates/ntx-builder/src/transaction.rs
```diff
@@ -349,6 +350,13 @@ impl DataStore for NtxDataStore {
             }
         }
     }
+
+    fn get_note_script(
+        &self,
+        script_root: Word,
+    ) -> impl FutureMaybeSend<Result<miden_objects::note::NoteScript, DataStoreError>> {
+        async move { Err(DataStoreError::NoteScriptNotFound(script_root)) }
```
bobbinth, line 359: I would maybe add a `TODO` here.

Issue #1290, opened 2025-10-14
Implement get_note_script() for NtxDataStore


sergerad:
The `NtxDataStore` implementation of `Datastore` does not properly implement `get_note_script()` at the moment:
```rust
    fn get_note_script(
        &self,
        script_root: Word,
    ) -> impl FutureMaybeSend<Result<miden_objects::note::NoteScript, DataStoreError>> {
        // TODO: Add implementation for getting note script from NtxDataStore.
        async move { Err(DataStoreError::NoteScriptNotFound(script_root)) }
    }
```

If possible, we need to add an implementation for this.

Issue #1227, opened 2025-09-15
feat: Implement Historical AccountTree State Storage for Last N Blocks


drahnr:
We store one instance of `AccountTree` as part of `StateInner`. With #1185 we want/need to keep around the information at the last `n`-block heights (n = 100~ish).

### How it currently works

```
let mut tree: AccountTree = from_somewhere!();
let muta = tree.compute_mutations();
tree.apply_mutations(muta);
```

We create a set of mutations that modify the state.
The mutations can be combined.

---

## Requirements

1. persistence of the history - after a crash reboot we must be able to provide them (needs feeback!)
2. the last `n` account tree states must be retrievable, but we don't need to track anything other than the tree itself
3. 

---

## Approach 1 - Naive

Store `n` instances of the `AccountTree` in a hashmap/tree and discard any older than `n` blocks from the tip.

#### Pros

Simple, indexable FIFO of fixed length 

#### Cons

Depending on the account tree might use excessive amounts of memory

#### Assumptions

None

## Approach 2

Store `2` instances of the `AccountTree`, the one at the chain tip (as currently) and the oldest one we want to provide. The states in between are reconstructed on demand from the additive changesets. We could pre-compute intermediate changesets (i.e. steps of 10) to speed this up or use tree like combination structure.
The account tree deltas need to be stored on-disk/in the database.

#### Pros

Medium complexity to implement.

#### Cons

Some computational overhead for past states

#### Assumptions

Associativity of changesets re addition

## Approach 3 - inverted-delta storage

Store `2` instances of the `AccountTree`, the one at the chain tip (as currently) and the oldest one we want to provide. The states in between are reconstructed on demand from the additive _inverted_ changesets. We could pre-compute intermediate changesets (i.e. steps of 10) to speed this up or use tree like combination structure.

The core difference compared to approach 2, is we do reduce the requirement for caching, since more recent ones only required to get one or two deltas applied, and we can update them cheaply by adding another inverted delta cheaply / easily. The inversion can be amortized at each block insertion.

#### Pros

Cheap on CPU, simpler yet more effective caching, least memory requirements

#### Cons

Unclear how easy inversions are to implement at this point.

#### Assumptions

Invertible mutation sets

Mirko-von-Leipzig:
Would adding a multi-root implementation be possible instead? I'm assuming this will be in-memory only.

[...]

PR #1292, opened 2025-10-14
feat: account tree with history

drahnr:
An alternative to `SmtWithHistory`, but uses a similar pattern.

Depends on ~[#2002 in miden-base](https://github.com/0xMiden/miden-base/pull/2002)~ merged ~and [#596 in miden-crypto](https://github.com/0xMiden/crypto/issues/596)~ closed

See  https://github.com/0xMiden/miden-node/issues/1227

---

The idea here in contrast to `SmtWithHistory` is not to change the underpinnings of `AccountTree` but wrap it.

The advantage is a minimal API surface to be implemented, i.e. we really ever only need `AccountTree'::open_at(block, account)` which is significantly smaller to implement compared to representing a full `Smt` at past block `BlockNumber`.

Impl approach:

Use the latest opening, and override elements in it using the _reversion mutation sets_ (so the inverse of the applied mutation on block number increment / block applied). This allows for traversing the reversion mutation sets if they contain inner nodes and leaf updates for relevant leaves, and as such update the merkle path, and hence opening. 

---

Caveats:

* Currently we panic on some IO errors, this is fine for now~ish, there is a broader issue covering the topic #1302 
* Still depends on https://github.com/0xMiden/miden-base/pull/2006
* The changeset includes a fixup pass through SQL statements in doc comments, which started to deviate from the actual impl.

---

Review approach:

The meat of changes is in `accounts/mod.rs` and some additional changes around the database in the `store/*`. Ignore the fact that we still don't return the actually correct account data, but that of the latest block, the remaining changes would alter the schema some and bloat the PR a little more, which is already large enough.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -792,7 +792,7 @@ dependencies = [
  "anes",
  "cast",
  "ciborium",
- "clap 4.5.50",
+ "clap 4.5.51",
  "criterion-plot",
  "is-terminal",
  "itertools 0.10.5",


crates/ntx-builder/src/transaction.rs
```diff
@@ -123,7 +123,8 @@ impl NtxContext {
                 let notes = notes.into_iter().map(Note::from).collect::<Vec<_>>();
                 let (successful, failed) = self.filter_notes(&data_store, notes).await?;
                 let executed = Box::pin(self.execute(&data_store, successful)).await?;
-                let proven = Box::pin(self.prove(executed.into())).await?;
+                let tx_inputs = executed.into_parts().0;
```
sergerad: Might be worth making `into_tx_inputs()` given we do this 3+ times now.

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+            HistoricalOffset::Future | HistoricalOffset::TooAncient => None,
+        }
+    }
+
+    /// returns account count
+    pub fn num_accounts(&self) -> usize {
+        self.inner.read().unwrap().latest.num_accounts()
+    }
+
+    /// returns history depth
```
sergerad: nit: capital letter + full stop for all doc comments. (and comments in general?)

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+        self.inner.read().unwrap().latest.open(account_id)
+    }
+
+    /// opens account at historical block
+    pub fn open_at(
+        &self,
+        account_id: AccountId,
+        block_number: BlockNumber,
+    ) -> Option<AccountWitness> {
+        let guard = self.inner.read().unwrap();
```
sergerad: expect with explanation instead of unwrap

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                // Reverse latest_nodes because they come in high-to-low depth order
+                // but we need to initialize path_nodes in low-to-high depth order
+                latest_nodes.reverse();
+
+                // initialize the path nodes with the latest state, and update repeatedly.
+                let mut path_nodes = [None; SMT_DEPTH as usize];
+                let mut node_idx = 0;
+                for (depth, path_node) in path_nodes.iter_mut().enumerate().take(SMT_DEPTH as usize)
+                {
+                    if (initial_mask & (1u64 << depth)) == 0 && node_idx < latest_nodes.len() {
```
sergerad: explainer comment would be good here. re bit operations

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                }
+
+                let max_depth =
+                    path_nodes.iter().rposition(std::option::Option::is_some).map_or(0, |d| d + 1);
+                let mut empty_mask = u64::MAX;
+                let nodes: Vec<Word> = (0..max_depth)
+                    .rev()
+                    .filter_map(|d| {
+                        path_nodes[d].inspect(|_| {
+                            empty_mask &= !(1u64 << d);
```
sergerad: explainer comment here and/or above this block of logic would be good

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+        self.inner.read().unwrap().overlays.len()
+    }
+
+    /// opens account at latest block
+    pub fn open(&self, account_id: AccountId) -> AccountWitness {
+        self.inner.read().unwrap().latest.open(account_id)
+    }
+
+    /// opens account at historical block
+    pub fn open_at(
```
sergerad, line 271: This fn is really long. Could we break it up a bit with a few more fns?

crates/store/src/state.rs
```diff
@@ -140,7 +140,13 @@ impl State {
             .map_err(StateInitializationError::DatabaseLoadError)?;
 
         let chain_mmr = load_mmr(&mut db).await?;
-        let account_tree = load_account_tree(&mut db).await?;
+        let block_headers = db.select_all_block_headers().await?;
+        // TODO do we need assurance that the block_headers in the DB are in sync with the account
+        // tree?
```
sergerad: Is this something we will look into in this pr?

drahnr: I am unsure about the invariant that we can assume, to my understanding, it should hold at all times.

crates/store/src/historical/mod.rs
drahnr: The node is present, so we set the bit to 0, which is using the inversion (`!`/`not`) which looks different from C `~`

crates/ntx-builder/src/transaction.rs
drahnr: That's a change that'd have to be done in `miden-base` iiuc

bin/network-monitor/src/remote_prover.rs
```diff
@@ -295,7 +295,7 @@ pub async fn generate_mock_transaction() -> anyhow::Result<TransactionInputs> {
     // Execute the transaction
     let executed_transaction =
         Box::pin(tx_context.execute()).await.context("Failed to execute transaction")?;
-    Ok(executed_transaction.into())
+    Ok(executed_transaction.into_parts().0)
```
bobbinth: nit: here, and in a couple of other places, I think we should be able to keep the previous approach (i.e., just  `executed_transaction.into()`).

Cargo.toml
```diff
@@ -55,7 +55,8 @@ miden-tx              = { branch = "next", default-features = false, git = "http
 miden-tx-batch-prover = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
 
 # Other miden dependencies. These should align with those expected by miden-base.
-miden-air = { features = ["std", "testing"], version = "0.18" }
+miden-air    = { features = ["std", "testing"], version = "0.18.1" }
+miden-crypto = { version = "0.17" }
```
bobbinth: Why do we need to import `miden-crypto` here directly? Ideally, we should be able to get everything that we need form `miden-crypto` via `miden-objects`.

crates/store/Cargo.toml
```diff
@@ -23,11 +23,12 @@ diesel                 = { features = ["numeric", "sqlite"], version = "2.2" }
 diesel_migrations      = { features = ["sqlite"], version = "2.2" }
 hex                    = { version = "0.4" }
 indexmap               = { workspace = true }
+miden-crypto           = { workspace = true }
```
bobbinth: Similar question as above.

crates/store/Cargo.toml
```diff
@@ -50,3 +51,10 @@ pretty_assertions     = { workspace = true }
 rand                  = { workspace = true }
 regex                 = { version = "1.11" }
 termtree              = { version = "0.5" }
+
+[dev-dependencies.criterion]
+version = "0.5"
```
bobbinth: nit: we usually write this as:
```toml
[dev-dependencies]
criterion =  { version = "0.5" }
```

proto/proto/store/rpc.proto
```diff
@@ -164,19 +164,22 @@ message AccountProofResponse {
...
 
         // Account asset vault data; empty if vault commitments matched or the requester
         // omitted it in the request.
         optional AccountVaultDetails vault_details = 4;
     }
 
-    // Account ID, current state commitment, and SMT path
-    account.AccountWitness witness = 1;
+    // The block number at which the witness was observed.
+    blockchain.BlockNumber block_num = 1;
```
bobbinth: nit: I'd mention that this is not only about the witness but also about account details.

crates/store/src/state.rs
```diff
@@ -140,7 +140,17 @@ impl State {
...
+        // TODO: Currently `load_account_tree` ignores the `block_num` parameter and loads all
+        // account commitments from the DB regardless of block. This could lead to inconsistency
+        // if the DB contains account states from blocks beyond `latest_block_num`.
+        // Consider either:
+        // 1. Filtering account commitments by block_num <= latest_block_num, or
+        // 2. Adding a DB constraint to ensure accounts table only contains latest states
+        let latest_block_num = block_headers
+            .last()
+            .map_or(BlockNumber::GENESIS, miden_objects::block::BlockHeader::block_num);
+        let account_tree = load_account_tree(&mut db, latest_block_num).await?;
```
bobbinth: The `accounts` table would have to contain multiple entries - though, in the future these will be only account header + storage header. So, approach 1 is probably what we'll need to go with. But it will be a bit more tricky than just a simple filter - we may have to add `is_latest` field to the `accounts` table to make selection of the latest account as easy as possible.

I am wondering if there is a way to sidestep this for this PR - maybe, for now, we still keep returning just the latest state?

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+            },
+            _ => HistoricalOffset::TooAncient,
+        }
+    }
+}
+
+/// Wraps `AccountTree` with historical query support via reversion overlays.
+#[derive(Debug, Clone)]
+pub struct AccountTreeWithHistory {
+    inner: Arc<RwLock<InnerState>>,
```
bobbinth: Why do we need to put this behind `Arc<RwLock<..>>`? The entire state of the store is locked during a request - so, extra locking here shouldn't be necessary.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+    ) -> Option<AccountWitness> {
+        let latest_witness = guard.latest.open(account_id);
+        let (latest_path, mut leaf) = latest_witness.into_proof().into_parts();
+        let (initial_mask, mut latest_nodes) = latest_path.into_parts();
+
+        // Initialize path_nodes with the latest state.
+        // Reverse latest_nodes because they come in high-to-low depth order
+        // but we need to initialize path_nodes in low-to-high depth order.
+        latest_nodes.reverse();
+        let mut path_nodes = Self::initialize_path_nodes(initial_mask, &latest_nodes);
```
bobbinth: Instead of dealing with a `SparseMerklePath` directly, it may be simpler to convert the `SparseMerklePath` into `MerklePath`, then perform all the updates on the `MerklePath`, and then, at the end, convert the final `MerklePath` into a `SparseMerklePath`. This may be a bit less efficient - but I think the delta would be negligible (we should double-check though).

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+    ///
+    /// We iterate through overlays from newest to oldest (going back in time),
+    /// updating both the path nodes and the leaf value based on the reversion mutations.
+    fn apply_reversion_overlays(
+        overlays: &VecDeque<Arc<HistoricalOverlay>>,
+        overlay_idx: usize,
+        path_nodes: &mut [Option<Word>; SMT_DEPTH as usize],
+        leaf_index: NodeIndex,
+        mut leaf: SmtLeaf,
+    ) -> SmtLeaf {
```
bobbinth: Would it make sene to pass in the iterator rather than `VecDeque` and `overlay_idx`? (since we use `overlay_idx` only in `overlays.iter().take()`.

Also, it is a bit odd that we take ownership of `leaf` but take mutable reference to `path_nodes` - should take ownership for both of these? And then maybe the return type could be `(MerklePath, SmtLeaf)`?

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+    ReversionsIdx(usize),
+    Latest,
+    TooAncient,
+}
+
+/// Captures reversion state for historical queries.
+#[derive(Debug, Clone)]
+struct HistoricalOverlay {
+    block_number: BlockNumber,
+    rev_set: AccountMutationSet,
```
bobbinth: We could probably split `AccountMutationSet` into individual fields. It could look something like:
```rust
struct HistoricalOverlay {
    block_number: BlockNumber,
    root: Word,
    node_mutations: HashMap<NodeIndex, Word>,
    account_updates: HashMap<Word, Word>,
}
```

Using hash maps should be more efficient for lookups, but may take more time to construct.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+    fn new(block_number: BlockNumber, rev_set: AccountMutationSet) -> Self {
+        Self { block_number, rev_set }
+    }
+}
+
+#[derive(Debug)]
+struct InnerState {
+    block_number: BlockNumber,
+    latest: AccountTree,
+    overlays: VecDeque<Arc<HistoricalOverlay>>,
```
bobbinth: nit: this could probably be `BTreeMap<BlockNumber, HistoricalOverlay>`.

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+                overlays: VecDeque::new(),
+            })),
+        }
+    }
+
+    fn cleanup(overlays: &mut VecDeque<Arc<HistoricalOverlay>>) {
+        while overlays.len() > Self::MAX_HISTORY {
+            overlays.pop_back();
+        }
+    }
```
bobbinth: nit: not sure if `cleanup()` is the best name here because this only removes a single entry - so, if we had more than one extra entry in the queue, this would need to be called repeatedly.

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+    /// Returns latest block number.
+    pub fn block_number(&self) -> BlockNumber {
+        self.inner
+            .read()
+            .expect("RwLock poisoned: concurrent thread panicked while holding lock")
+            .block_number
+    }
+
+    /// Returns latest root.
+    pub fn root(&self) -> Word {
```
bobbinth: Unless this creates some complications, I would probably name this `latest_root()` to make things a bit more explicit. This would also apply for other similar methods (e.g., `block_number()`, `open()`).

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+
+        inner.overlays.push_front(Arc::new(overlay));
+        inner.block_number = inner.block_number.child();
+        Self::cleanup(&mut inner.overlays);
+
+        Ok(())
+    }
+
+    /// Gets account commitments in latest state.
+    pub fn account_commitments(&self) -> Vec<(AccountId, Word)> {
```
bobbinth: This method may return too much data - do we actually need it?

bobbinth:
> Access Ops ( against `Smt` @ 10 blocks depth):
> Vanilla: 804 ns
> Historical (d=0): 800 ns
> Historical (d=5): 26.6 µs
> Historical (d=10): 53.8 µs

This is for calling `open()` method, right? How do these compare to `LargetSmt::open()` times?

> #### 500 Accounts:
> | Depth | Time | Per-block history |
> |-------|-------|----------------|
> | 0 | 1.3267 µs | - - |
> | 5 | 51.674 µs | 10.07 µs/block |
> | 10 | 109.32 µs | 10.80 µs/block |
> | 20 | 213.89 µs | 10.63 µs/block |
> | 32 | 349.70 µs | 10.89 µs/block |

What are these benchmarks measuring?

crates/store/src/state.rs
drahnr: The invariant is upheld when calling `apply_block`, removed.

crates/store/src/historical/mod.rs
drahnr: Removed

crates/store/Cargo.toml
drahnr: Removed, dependency patching residue

crates/store/src/accounts/mod.rs
drahnr: If we'd use the `hashmaps` feature of `miden-crypto` that'd be the case. Do we expect that to be enabled by default?

crates/store/src/accounts/mod.rs
drahnr: Iiuc this would simplify / get rid of  `fn historical_offset` at the cost of an additional `log(n)` compares, I don't have a strong preference.

crates/store/src/historical/mod.rs
drahnr: Added some additional details inline

crates/store/src/state.rs
drahnr: > The accounts table would have to contain multiple entries - though, in the future these will be only account header + storage header.

Yes, right now this is only limited by the primary table key being `(account_id)` rather than `(account_id, block_num)`, which limits our ability to lookup by `block_num` (/w foreign key constraint). I think this is a relatively small change.

> So, approach 1 is probably what we'll need to go with. But it will be a bit more tricky than just a simple filter - we may have to add `is_latest` field to the accounts table to make selection of the latest account as easy as possible.

I just had a look at the code for retrieval and insertion, and I am a bit curious why you'd think it'd be more complicated? Why do we require the `is_latest`. Looking at the existing code, that complicates the queries significantly, it only increases the latest lookup speed, but in this case we do know the block_number and the previous block number, so I don't quite see the issue. Could you elaborate?
 
> I am wondering if there is a way to sidestep this for this PR - maybe, for now, we still keep returning just the latest state?

I think that'd be the best approach and I'll create a follow-up PR.

crates/store/src/historical/mod.rs
drahnr: The intent here was to ensure no modifications/increment of the `block_num` and `rev_sets` can occur independently, which would lead to inconsistent return values. Any block import / application would cause a write, asynchronously with the read operations, the `RwLock` is the easiest way to address the contention. Alternatively an asynchronous `BTreeMap<BlockNumber, >` could employed at the cost of additional error handling.

crates/store/src/historical/mod.rs
drahnr: Renamed, the intent was to retain the API as `AccountTree` and add additional API with the `_at` suffix.

crates/store/src/historical/mod.rs
drahnr: Added a doc comment and renamed to `drain_excess`

crates/store/Cargo.toml
drahnr: Re-added to force the unification to include rocksdb, only for the benchmark at this point.

crates/ntx-builder/src/transaction.rs
```diff
@@ -355,8 +355,11 @@ impl DataStore for NtxDataStore {
         &self,
         script_root: Word,
     ) -> impl FutureMaybeSend<Result<miden_objects::note::NoteScript, DataStoreError>> {
-        // TODO: Add implementation for getting note script from NtxDataStore.
-        async move { Err(DataStoreError::NoteScriptNotFound(script_root)) }
+        async move {
+            // For network transactions, note scripts are not cached in the data store
+            // They should be retrieved from the blockchain/note data itself
+            Err(DataStoreError::NoteScriptNotFound(script_root))
```
sergerad: Is this a temporary change or would we close #1290?

crates/store/src/state.rs
```diff
@@ -1088,9 +1113,22 @@ async fn load_mmr(db: &mut Db) -> Result<Mmr, StateInitializationError> {
 }
 
 #[instrument(level = "info", target = COMPONENT, skip_all)]
-async fn load_account_tree(db: &mut Db) -> Result<AccountTree, StateInitializationError> {
+async fn load_account_tree(
+    db: &mut Db,
+    _block_num: BlockNumber,
```
sergerad: Is block num going to be used?

drahnr: I created #1303 to do the full integration, so that'd be a follow up. This provides all means, next step is to integrate them fully.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -85,6 +91,52 @@ pub(crate) fn select_account(
...
+/// LEFT JOIN
+///     account_codes ON accounts.code_commitment = account_codes.code_commitment
+/// WHERE
+///     account_id = ?1 AND block_num = ?2
+/// ```
+pub(crate) fn select_historical_account_at(
+    conn: &mut SqliteConnection,
+    account_id: AccountId,
+    block_num: BlockNumber,
+) -> Result<proto::domain::account::AccountInfo, DatabaseError> {
```
bobbinth: Question: why do we need the full path here - i.e., `proto::domain::account::AccountInfo`? Is this `AccountInfo` different from the one we use elsewhere in this file?

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+    NodeMutation,
+    SMT_DEPTH,
+    SmtLeaf,
+    SmtStorage,
+    SparseMerklePath,
+};
+use miden_objects::{AccountTreeError, EMPTY_WORD, Word};
+
+/// Trait abstracting operations over different account tree backends.
+pub trait AccountTreeBackend {
```
bobbinth: How is this different from the `AccountTreeBackend` that we have in `miden-base`? Do we need both?

crates/store/src/accounts/mod.rs
bobbinth: `log(n)` here should be pretty negligible - so, i'm not too worried about it from the efficiency standpoint. I do think `BTreeMap` makes the readability a bit easier - but also not a strong preference.

crates/store/src/accounts/mod.rs
bobbinth: I think we can enable the `hashmaps` feature by default here, but I think this is separate from that. That is, we could use a hashmap (maybe from hashbrown) here without enabling the `hashmaps` feature. This would still make node looks faster.

But again, we can probably do both.

crates/store/src/historical/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+
+        inner.overlays.push_front(Arc::new(overlay));
+        inner.block_number = inner.block_number.child();
+        Self::drain_excess(&mut inner.overlays);
+
+        Ok(())
+    }
+
+    /// Returns the oldest block still in history.
+    pub fn oldest_block_num(&self) -> BlockNumber {
```
bobbinth: AFAICT, this method does not mutate the state - so, it should probably not be in the "public mutators" section.

Similar comment for `contains_account_id_prefix()` method below.

crates/store/src/historical/mod.rs
bobbinth: As far as I can tell, the only way to mutate `AccountTreeWithHistory` is through the `apply_mutations()` method and when we call this method, we already have a lock on the `AccountTreeWithHistory` (same as we did on `AccountTree` before). Basically, write and read operations should never happen at the same time - we either are reading the tree or writing to it - but never both at the same time. So, `Arc<RwLock<>>` here should not be needed. But maybe I'm missing something?

crates/store/src/state.rs
bobbinth: > I just had a look at the code for retrieval and insertion, and I am a bit curious why you'd think it'd be more complicated? Why do we require the `is_latest`. Looking at the existing code, that complicates the queries significantly, it only increases the latest lookup speed, but in this case we do know the block_number and the previous block number, so I don't quite see the issue. Could you elaborate?

One of the main benefits of having `is_latest` is much simpler cleanup logic - see https://github.com/0xMiden/miden-node/issues/1303#issuecomment-3435394522 for the query example. Another benefit is being able to easily identify the set of latest account states.

Without `is_latest` we'd have to use window functions (or something similar) in SQL which would make the queries more complex and probably impact the performance quite a bit in a negative way.

crates/store/src/lib.rs
```diff
@@ -4,10 +4,12 @@ mod blocks;
 mod db;
 mod errors;
 pub mod genesis;
+pub mod historical;
```
bobbinth: Does this need to be `pub mod` rather than just `mod`? It seems like we already export the relevant pieces on line 12 below.

Also, I'd probably name the module just `accounts` (i.e., `src/accounts/mod.rs` and `src/accounts/tests.rs`).

crates/store/src/state.rs
```diff
@@ -1088,9 +1113,22 @@ async fn load_mmr(db: &mut Db) -> Result<Mmr, StateInitializationError> {
...
+) -> Result<AccountTree<LargeSmt<MemoryStorage>>, StateInitializationError> {
```
bobbinth: Should we create an alias for this (or maybe even a newtype)? Maybe `InMemoryAccountTree`? In the future, we'd maybe have something like `PersistentAccountTree`.

crates/store/src/historical/mod.rs
drahnr: We do, they're different. For better disambiguation I renamed the one here to `trait AccountTreeStorage`. The `miden-base` `trait AccountTreeBackend` we need to parameterize the `AccountTree` over storage backends aka Smt implementations.
Here we do parameterizer the `AccountTreeWithHistory` over implementations over a minimal set of primitives required to allow for it to work, i.e. `AccountTree<S: AccountTreeBackend>`.

The names are far from optimal, my current approach is to call `miden-base::AccounTreeBackend ~~~> AccountTreeStorage` and `minde-node::AccountTreeStorage ~~~> LatestAccountTreeImpl`. I don't really like these names either; open to suggestions

crates/store/benches/account_tree_historical.rs
```diff
@@ -0,0 +1,252 @@
...
+use miden_objects::testing::account_id::AccountIdBuilder;
+
+// Counter for unique benchmark database directories
+static DB_COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+// HELPER FUNCTIONS
+// ================================================================================================
+
+/// Creates a `RocksDB` storage instance in target/bench_* directory for benchmarking.
+fn setup_rocksdb_storage() -> MemoryStorage {
```
drahnr: This should become rocksdb, but is inhibited by https://midengroup.slack.com/archives/C093G3C6MV2/p1761664143312309 for now

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+    ) -> Result<AccountMutationSet, AccountTreeError>;
+
+    /// Applies mutations with reversion data.
+    fn apply_mutations_with_reversion(
+        &mut self,
+        mutations: AccountMutationSet,
+    ) -> Result<AccountMutationSet, AccountTreeError>;
+
+    /// Checks if the tree contains an account with the given prefix.
+    fn contains_account_id_prefix(&self, prefix: miden_objects::account::AccountIdPrefix) -> bool;
```
bobbinth: nit: is there a reason to use a full import path here rather than importing `AccountIdPrefix` at the top of the file?

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+// HISTORICAL OVERLAY
+// ================================================================================================
+
+/// Captures reversion state for historical queries at a specific block.
+#[derive(Debug, Clone)]
+struct HistoricalOverlay {
+    block_number: BlockNumber,
+    root: Word,
+    node_mutations: HashMap<NodeIndex, Word>,
+    account_updates: HashMap<Word, Word>,
```
bobbinth: Should we use `HashMap` from `hashbrown` here? IIRC, it is a bit more efficient than the one from Rust stdlib.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+impl HistoricalOverlay {
+    fn new(block_number: BlockNumber, rev_set: AccountMutationSet) -> Self {
+        let root = rev_set.as_mutation_set().root();
+        let mut_set = rev_set.into_mutation_set();
+
+        let node_mutations =
+            HashMap::from_iter(mut_set.node_mutations().iter().map(|(node_index, mutation)| {
+                match mutation {
+                    NodeMutation::Addition(inner_node) => (*node_index, inner_node.hash()),
+                    NodeMutation::Removal => (*node_index, EMPTY_WORD),
```
bobbinth: Question: should we use `EMPTY_WORD` here or a root of an empty subtree at the depth implied by `node_index`?

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+            latest: account_tree,
+            overlays: BTreeMap::new(),
+        }
+    }
+
+    /// Removes oldest overlays when exceeding the maximum history depth.
+    fn drain_excess(overlays: &mut BTreeMap<BlockNumber, HistoricalOverlay>) {
+        while overlays.len() > Self::MAX_HISTORY {
+            overlays.pop_first();
+        }
```
bobbinth, line 217: nit: we could also use `BTreeMap::split_off()` here - though, not sure if it would be much better.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+        self.latest.get(account_id)
+    }
+
+    /// Checks if the tree contains an account with the given prefix.
+    pub fn contains_account_id_prefix(
+        &self,
+        prefix: miden_objects::account::AccountIdPrefix,
+    ) -> bool {
+        self.latest.contains_account_id_prefix(prefix)
+    }
```
bobbinth, line 295: It doesn't seem like these two methods are being used - do we need them? If so, given that they work with the latest state of the tree, we should probably add "latest" somewhere in the names.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+
+    /// Determines the historical state of a requested block number.
+    pub fn historical_state(&self, desired_block_number: BlockNumber) -> HistoricalState {
```
bobbinth: Does this method need to be `pub`?

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+        let mut path_nodes = [EMPTY_WORD; SMT_DEPTH as usize];
+
+        let path = Vec::from_iter(path.iter());
+
+        path.iter().rev().enumerate().for_each(|(depth, &node)| {
+            let empty_root = *EmptySubtreeRoots::entry(SMT_DEPTH, (depth + 1) as u8); // +1: depth from leaf
+            if node != empty_root {
+                path_nodes[depth] = node;
+            }
+        });
```
bobbinth: Could this not be just something like this:
```rust
let mut path_nodes: [Word; SMT_DEPTH as usize] =
    MerklePath::from(path).to_vec().try_into().expect("some message");
path_nodes.reverse();
```
This would be slightly less efficient but would avoid the need to use `EmptySubtreeRoots`.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+    /// Iterates through overlays from newest to oldest (walking backwards in time),
+    /// updating both the path nodes and the leaf value based on reversion mutations.
+    /// Uses `EMPTY_WORD` as a sentinel to indicate "use empty subtree root for this depth".
+    fn apply_reversion_overlays(
+        overlays: &BTreeMap<BlockNumber, HistoricalOverlay>,
+        block_target: BlockNumber,
+        mut path_nodes: [Word; SMT_DEPTH as usize],
+        leaf_index: NodeIndex,
+        mut leaf: SmtLeaf,
+    ) -> Option<(SparseMerklePath, SmtLeaf)> {
```
bobbinth: We could probably change the signature here to be something like:
```rust
fn apply_reversion_overlays<'a>(
    overlays: impl IntoIterator<Item = &'a HistoricalOverlay>,
    mut path_nodes: [Word; SMT_DEPTH as usize],
    leaf_index: NodeIndex,
    mut leaf: SmtLeaf,
) -> Option<(SparseMerklePath, SmtLeaf)>
```
And then the `for` loop on lines 382-383 becomes just:
```rust
for overlay in overlays.into_iter() {
    ...
}
```
The invocation on lines 332 - 338 then would look something like:
```rust
let (path, leaf) = Self::apply_reversion_overlays(
    self.overlays.range(block_target..).map(|(_, value)| value).rev(),
    path_nodes,
    leaf_index,
    leaf,
)?;
```

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+                    .depth()
+                    .checked_sub(1) // -1: Convert from 1-indexed to 0-indexed
+                    .expect("proof_indices should not include root")
+                    as usize;
+
+                // Apply reversion mutation if this node was modified
+                // EMPTY_WORD sentinel means "use empty subtree root for this depth"
+                if let Some(hash) = overlay.node_mutations.get(&sibling) {
+                    path_nodes[height] = *hash;
+                }
```
bobbinth: Related to one of the above comments: if in `HistoricalOverlay::new()` we use empty subtree roots instead of `EMPTY_WORD`, we should be able to avoid using `EMPTY_WORD` as a sentinel value, and would also not need the `build_dense_path()` helper.

crates/store/src/state.rs
```diff
@@ -88,13 +91,19 @@ pub struct TransactionInputs {
...
 /// Container for state that needs to be updated atomically.
-struct InnerState {
+struct InnerState<S = MemoryStorage>
+where
+    S: SmtStorage + Default,
+{
     nullifier_tree: NullifierTree,
     blockchain: Blockchain,
-    account_tree: AccountTree,
+    account_tree: AccountTree<LargeSmt<S>>,
```
bobbinth: This still needs to change to `AccountTreeWithHistory<AccountTree<LargeSmt<S>>>`, right?

bobbinth (commented):
Thank you! Looks good. Not a full review, but I left some comments inline.

Let's try to get this PR into a mergable state, even if we can't use the RocksDB backend for now.

Also, how do the benchmarks look now for accessing historical data?

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        let node_mutations =
+            HashMap::from_iter(mut_set.node_mutations().iter().map(|(node_index, mutation)| {
+                match mutation {
+                    NodeMutation::Addition(inner_node) => (*node_index, inner_node.hash()),
+                    NodeMutation::Removal => (*node_index, EMPTY_WORD),
+                }
+            }));
+
+        let account_updates = HashMap::from_iter(mut_set.new_pairs().iter().map(|(&k, &v)| (k, v)));
```
bobbinth: I wonder if it would be better to have this hash map be `HashMap<LeafIndex, (Word, Word>>` so that we wouldn't need to use `.find()` to find the right leaf.

crates/store/src/historical/mod.rs
drahnr: We do use it, it wasn't shown as used before since I wanted to hold off on using `AccountTreeWithHistory` with the remaining integration points.

```
crates/store/src/state.rs
870-        let new_account_id_prefix_is_unique = if account_commitment.is_empty() {
871:            Some(!inner.account_tree.contains_account_id_prefix(account_id.prefix()))
872-        } else {
873-            None
874-        };

```

crates/store/src/accounts/mod.rs
drahnr: The default implementation of `std::collections::HashMap` is `hashbrown`. https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#4

crates/store/src/accounts/mod.rs
drahnr: If one has a sentinel value per height level it's just a bit more finnicky to debug if anything goes wrong, with a const sentinel it's easier. I can change to the `EmptyTreeRoot::` approach, I don't have a strong preference.

crates/store/src/accounts/mod.rs
drahnr: Practically we always call this after insertion, so the loop is really only a safeguard and should always run either 0 times or 1 time.

crates/store/src/accounts/mod.rs
drahnr: They are, in `State`.

crates/store/src/accounts/mod.rs
drahnr: Changed to your proposal.

crates/store/src/accounts/mod.rs
drahnr: Changed to your proposal (roughly).

crates/store/src/state.rs
drahnr: Initially I inteded to leave it until it can be fully hooked up, changed.

crates/store/src/accounts/mod.rs
drahnr: It is better, particularly for large updates.

crates/store/src/accounts/mod.rs
drahnr: Changed and removed.

crates/store/src/accounts/mod.rs
drahnr: Implemented the `EmptyTreeRoot` based approach.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+            for sibling in leaf_index.proof_indices() {
+                let height = sibling
+                    .depth()
+                    .checked_sub(1) // -1: Convert from 1-indexed to 0-indexed
+                    .expect("proof_indices should not include root")
+                    as usize;
+
+                // Apply reversion mutation if this node was modified
+                if let Some(hash) = overlay.node_mutations.get(&sibling) {
+                    path_nodes[height] = *hash;
```
sergerad, line 389: Might be worth adding a comment here explaining why height would never be out of bounds

sergerad (approved):

crates/store/src/accounts/mod.rs
bobbinth: Ah - I didn't realize this! I wonder if it makes sense to have `hashbrown` as a dependency in `miden-crypto` then.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    fn apply_mutations_with_reversion(
+        &mut self,
+        mutations: AccountMutationSet,
+    ) -> Result<AccountMutationSet, AccountTreeError> {
+        self.apply_mutations_with_reversion(mutations)
+    }
+
+    fn contains_account_id_prefix(&self, prefix: miden_objects::account::AccountIdPrefix) -> bool {
```
bobbinth: nit: let's remove the full path from here as well (especially since `AccountIdPrefix` is already imported).

crates/store/src/accounts/mod.rs
bobbinth: Ah - I see! I would then add "latest" to the names - e.g., something like `get_latest_account_state()` and `contains_account_id_prefix_in_latest()`.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+        // Check if block exists in overlays
+        if !self.overlays.contains_key(&desired_block_number) {
+            return HistoricalState::TooAncient;
+        }
+
+        HistoricalState::Target(desired_block_number)
+    }
+
+    // PRIVATE HELPERS - HISTORICAL RECONSTRUCTION
+    // --------------------------------------------------------------------------------------------
```
bobbinth: nit: I would move this about line 304 as `historical_state()` is also a helper function.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+
+        let leaf_index = NodeIndex::from(leaf.index());
+
+        // Apply reversion overlays to reconstruct historical state
+        let (path, leaf) = Self::apply_reversion_overlays(
+            self.overlays.range(block_target..).rev().map(|(_, overlay)| overlay),
+            path_nodes,
+            leaf_index,
+            leaf,
+        )?;
```
bobbinth: nit: I would maybe add an explanation to the comment about why we are reversing the order of overlays.

bobbinth:
> So the intent here is to keep public API as is and de-risk planned demos? We indeed could do this and combine it with the schema change.

Yeah, basically keep the schema the same, but error out if the endpoint is invoked for a historical block.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+where
+    S: AccountTreeStorage,
+{
+    /// The current block number (latest state).
+    block_number: BlockNumber,
+    /// The latest account tree state.
+    latest: S,
+    /// Historical overlays indexed by block number, storing reversion data.
+    overlays: BTreeMap<BlockNumber, HistoricalOverlay>,
+}
```
bobbinth, line 192: Maybe not for this PR, but I think it should be possible to define `AccountTreeWithHistory` as:
```rust
pub struct AccountTreeWithHistory<S: AccountTreeBackend> {
    block_number: BlockNumber,
    latest: AccountTree<S>,
    overlays: BTreeMap<BlockNumber, HistoricalOverlay>,
}
```
Because we probably won't have any type besides `AccountTree` to use here for `latest`.

And this should allow us to get rid of `AccountTreeStorage` trait and associated code.

crates/store/Cargo.toml
```diff
@@ -42,11 +44,20 @@ tracing                = { workspace = true }
...
 miden-node-utils      = { features = ["testing", "tracing-forest"], workspace = true }
 miden-objects         = { default-features = true, features = ["testing"], workspace = true }
-pretty_assertions     = { workspace = true }
 rand                  = { workspace = true }
 regex                 = { version = "1.11" }
 termtree              = { version = "0.5" }
+
+[features]
+default = []                       # rocksdb has just too many requirements on CI for now
+rocksdb = ["miden-crypto/rocksdb"]
```
bobbinth: We are not using the `rocksdb` backend yet, right? If so, it may make sense to remove the feature from this PR (and this may also roll back some Makefile changes).

crates/store/src/state.rs
```diff
@@ -1088,9 +1119,21 @@ async fn load_mmr(db: &mut Db) -> Result<Mmr, StateInitializationError> {
 }
 
 #[instrument(level = "info", target = COMPONENT, skip_all)]
-async fn load_account_tree(db: &mut Db) -> Result<AccountTree, StateInitializationError> {
+async fn load_account_tree(
+    db: &mut Db,
+    block_number: BlockNumber,
+) -> Result<AccountTreeWithHistory<AccountTree<LargeSmt<MemoryStorage>>>, StateInitializationError>
```
bobbinth: nit: we could use `AccountTreeWithHistory<InMemoryAccountTree>` instead of `AccountTreeWithHistory<AccountTree<LargeSmt<MemoryStorage>>` - right?

proto/proto/store/rpc.proto
```diff
@@ -164,19 +164,22 @@ message AccountProofResponse {
...
 
         // Account asset vault data; empty if vault commitments matched or the requester
         // omitted it in the request.
         optional AccountVaultDetails vault_details = 4;
     }
 
-    // Account ID, current state commitment, and SMT path
-    account.AccountWitness witness = 1;
+    // The block number at which the account witness and details were observed.
+    blockchain.BlockNumber block_num = 1;
```
bobbinth: I think we should make `block_num` here optional. If `block_num` is not provided, we'd just return the data for the latest block. Also, as I mentioned in one of my previous comments, for now, if `block_num` is provided, we'll return an error and then, in v0.13 we'll enable the full functionality.

crates/store/src/db/mod.rs
```diff
@@ -406,6 +406,25 @@ impl Db {
...
 
+    /// Loads public account details from the DB at a given block height.
+    ///
+    /// Note: Only tracks a limited history of 20~ish blocks.
+    #[allow(
+        dead_code,
+        reason = "Relevant, but requires an additional primary key schema change"
+    )]
+    #[instrument(level = "debug", target = COMPONENT, skip_all, ret(level = "debug"), err)]
+    pub async fn select_historical_account_at(
```
bobbinth: As far as I can tell, this method is not yet used - so, maybe we can remove it from this PR?

bobbinth (commented):
Looks good! Thank you! I left a few more comments inline. The main one is about making `block_num` an optional parameter for the endpoint and, for now, returning an error if it is provided.

Also, would be great to see the latest performance numbers, and I'd love to get one more review either from @sergerad or from @igamigo.

crates/store/src/accounts/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+    ) -> Result<AccountMutationSet, AccountTreeError> {
+        self.apply_mutations_with_reversion(mutations)
+    }
+
+    fn contains_account_id_prefix(&self, prefix: AccountIdPrefix) -> bool {
+        self.contains_account_id_prefix(prefix)
+    }
+}
+
+// ERROR TYPES
```
sergerad: nit: HISTORICAL ERROR

drahnr:
## `InMemory` Benchmark Overview / `LargeSmt<Memory>`

### [Access] Account Tree Vanilla  as _baseline_

| Size | Time |
|------|------|
| 1 | 1.31 µs |
| 10 | 1.35 µs |
| 50 | 1.42 µs |
| 100 | 1.50 µs |
| 500 | 1.51 µs |
| 1000 | 1.48 µs |

### [Access] Account Tree Historical vs Vanilla (n Accounts = 500)

| Depth | Historical | Vanilla |
|-------|------------|---------|
| 0 | 1.61 µs | 1.51 µs |
| 5 | 8.20 µs | 1.51 µs |
| 10 | 13.95 µs | 1.51 µs |
| 20 | 25.18 µs | 1.51 µs |
| 32 | 39.75 µs | 1.51 µs |

**Summary**: Historical access adds ballpark 1µs per depth level

---

### Account Tree Historical Access by Depth and Size

| Depth | Size 10 | Size 100 | Size 500 | Size 2500 |
|-------|---------|----------|----------|-----------|
| 0 | 1.40 µs | 1.47 µs | 1.61 µs | 1.55 µs |
| 5 | 7.69 µs | 7.91 µs | 8.20 µs | 8.30 µs |
| 10 | 13.73 µs | 13.49 µs | 13.95 µs | 14.56 µs |
| 20 | 25.51 µs | 24.36 µs | 25.18 µs | 26.88 µs |
| 32 | 40.10 µs | 37.80 µs | 39.75 µs | 40.92 µs |

**Summary**: Historical access time dominated by depth, size has minimal impact

proto/proto/store/rpc.proto
```diff
@@ -148,7 +148,7 @@ message AccountProofRequest {
     account.AccountId account_id = 1;
 
     // Block at which we'd like to get this data. Must be close to the chain tip.
-    fixed32 block_num = 2;
+    blockchain.BlockNumber block_num = 2;
```
bobbinth: Ah - I think I messed up with one of my last comments. I think block number in the request (i.e., here) should be optional - but the block number in the response doesn't need to be (i.e., on line 177 below). Sorry!

drahnr: I figured, that's what I implemented.

crates/store/benches/account_tree_historical.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    let storage = setup_storage();
+    let smt =
+        LargeSmt::with_entries(storage, entries).expect("Failed to create LargeSmt from entries");
+    let tree = AccountTree::new(smt).expect("Failed to create AccountTree");
+    (tree, account_ids)
+}
+
+/// Sets up `AccountTreeWithHistory` with specified number of accounts and blocks using `RocksDB`
```
igamigo: nit: the RocksDB storage backend part looks misleading, but this seems like it's going to change soon anyway

crates/proto/src/generated/rpc_store.rs
```diff
@@ -18,9 +18,9 @@ pub struct AccountProofRequest {
     /// ID of the account for which we want to get data
     #[prost(message, optional, tag = "1")]
     pub account_id: ::core::option::Option<super::account::AccountId>,
-    /// Block at which we'd like to get this data. Must be close to the chain tip.
-    #[prost(fixed32, tag = "2")]
-    pub block_num: u32,
+    /// Block at which we'd like to get this data. If present, must be close to the chain tip.
+    #[prost(message, optional, tag = "2")]
+    pub block_num: ::core::option::Option<super::blockchain::BlockNumber>,
```
sergerad, line 24: Do we have a corresponding PR in client live or coming up?

crates/store/src/errors.rs
```diff
@@ -117,6 +119,9 @@ pub enum DatabaseError {
     ConnectionManager(#[from] ConnectionManagerError),
     #[error(transparent)]
     SqlValueConversion(#[from] DatabaseTypeConversionError),
+
```
sergerad: nit: whitespace inconsistent with above

proto/proto/store/rpc.proto
```diff
@@ -147,8 +147,8 @@ message AccountProofRequest {
     // ID of the account for which we want to get data
     account.AccountId account_id = 1;
 
-    // Block at which we'd like to get this data. Must be close to the chain tip.
-    fixed32 block_num = 2;
+    // Block at which we'd like to get this data. If present, must be close to the chain tip.
+    optional blockchain.BlockNumber block_num = 2;
```
bobbinth, line 152: I would add to the comment something like `If not present, data from the latest block will be returned."

bobbinth (approved):
All looks good! Thank you!

igamigo (approved):
Looks great! Not the most in-depth review but the critical functionality looks good and so do tests

crates/proto/src/generated/rpc_store.rs
drahnr: Not yet, CC @igamigo do you want to do it, otherwise I'll do it today/tomorrow

PR #1293, opened 2025-10-15
feat: Validator component

sergerad:
## Context

As part of the network guardrails initiative, we require a component of miden-node which will re-execute proven transactions, store transaction metadata for private account retrieval, and perform block attestation of blocks coming out of the block-producer.

This PR adds the initial scaffolding of the validator component with placeholder protobuf and SQL schema definitions.

Relates to https://github.com/0xMiden/miden-node/issues/613.

## Changes

- Add `validator` crate with placeholder protobuf and SQL schemas and server scaffolding.
- Add validator sub-command to binary crate.
- Add `pub` and `pub use` to various types and functions from store crate to be used in validator.

### DIFF
--- a/crates/validator/README.md
+++ b/crates/validator/README.md
@@ -63,12 +63,11 @@ sequenceDiagram
 
 - ✅ Protobuf schema and gRPC server scaffolding.
 - ✅ Transaction submission handler.
-- ✅ Placeholder SQL schema and SQLite database integration.
 
 ### In Development
 
 - 🚧 Transaction re-execution logic.
-- 🚧 Finalized database schema.
+- 🚧 Database schema and impl.
 - 🚧 Block validation logic.
 - 🚧 Block signing logic.
 - 🚧 Integration with RPC.


crates/validator/src/db/migrations.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    let Err(e) = conn.run_pending_migrations(MIGRATIONS) else {
+        return Ok(());
+    };
+    tracing::warn!(target = COMPONENT, "Failed to apply migration: {e:?}");
+    conn.revert_last_migration(MIGRATIONS)
+        .expect("Duality is maintained by the developer");
+
+    Ok(())
+}
```
sergerad: This is duplicated because `EmbeddedMigrations` is not clone and has ownership transferred to `pending_migrations()`. We can work around this if we would probably be fighting the design of the relevant module so we might prefer to duplicate in the way we have here.

crates/validator/src/db/models.rs
```diff
@@ -0,0 +1,65 @@
...
+            account_id: transaction_header.account_id().to_bytes(),
+            block_num: block_num.to_raw_sql(),
+            initial_state_commitment: transaction_header.initial_state_commitment().to_bytes(),
+            final_state_commitment: transaction_header.final_state_commitment().to_bytes(),
+            input_notes: input_notes_binary,
+            output_notes: output_notes_binary,
+            size_in_bytes,
+        }
+    }
+}
```
sergerad: This is duplicated from store crate. Just placeholder for now. I'm unsure what the actual schema will be just yet.

crates/validator/src/db/schema.rs
```diff
@@ -0,0 +1,12 @@
...
+        transaction_id -> Binary,
+        account_id -> Binary,
+        block_num -> BigInt,
+        initial_state_commitment -> Binary,
+        final_state_commitment -> Binary,
+        input_notes -> Binary,
+        output_notes -> Binary,
+        size_in_bytes -> BigInt,
+    }
+}
```
sergerad: This should be generated for us. Maybe I'm missing something from build.rs or maybe we need to add something to the Makefile

crates/validator/src/server/submit.rs
```diff
@@ -0,0 +1,51 @@
...
+    // Insert the transaction into the database
+    let rows_affected = insert_transactions(conn, block_num, &ordered_transactions)?;
+
+    debug!(
+        target = COMPONENT,
+        rows_affected = rows_affected,
+        "Successfully inserted proven transaction into database"
+    );
+
+    Ok(rows_affected)
```
sergerad: This is all placeholder until we know the actual schema

crates/validator/src/db/conv.rs
```diff
@@ -0,0 +1,28 @@
...
+/// Convert from and to it's database representation and back
+///
+/// We do not assume sanity of DB types.
+pub trait SqlTypeConvert: Sized {
+    type Raw: Sized;
+    type Error: std::error::Error + Send + Sync + 'static;
+    fn to_raw_sql(self) -> Self::Raw;
+    #[allow(dead_code)]
+    fn from_raw_sql(_raw: Self::Raw) -> Result<Self, Self::Error>;
+}
```
drahnr: This one feels odd to copy-paste.

crates/validator/src/db/migrations.rs
drahnr: It's fine as long as we pick the right migrations if we end up with two different databases, and not accidentally apply the wrong migration to the wrong database.

crates/validator/src/db/schema.rs
drahnr: The idea is to either create this manually or do it as previously described in https://github.com/0xMiden/miden-node/pull/928#discussion_r2240252953

crates/validator/src/server/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+use crate::COMPONENT;
+use crate::db::load;
+use crate::server::submit::submit_proven_transaction;
+
+pub mod submit;
+
+// VALIDATOR
+// ================================================================================
+
+pub struct Validator {
```
drahnr, line 23: nit: doc comment

crates/validator/src/server/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// The address of the validator component.
+    pub address: SocketAddr,
+    /// Server-side timeout for an individual gRPC request.
+    ///
+    /// If the handler takes longer than this duration, the server cancels the call.
```
drahnr, line 28: nit: what's the default value?

sergerad: All the serveres rely on `const DEFAULT_TIMEOUT: Duration = Duration::from_secs(10);` and have this same comment

crates/validator/README.md
```diff
@@ -0,0 +1,91 @@
...
+The validator component is responsible for:
+
+- **Transaction Validation**: Receiving and re-executing proven transactions from the RPC layer.
+- **Transaction Storage**: Persisting re-executed transaction data for block verification and signing.
+- **Block Validation**: Verifying blocks produced by the block producer using stored transaction information.
+- **Block Signing**: Cryptographically signing valid blocks as part of the consensus process.
+
+## Architecture
+
+The validator operates as a gRPC server and integrates with other Miden node components according to the following flow:
```
bobbinth, line 16: This is fine for this PR, but in the future, we should probably move this diagram into the relevant section of the node docs (in the `docs` directory). In the same future PR, we should probably add a section about the validator component there as well.

crates/validator/README.md
bobbinth, line 1: Let's add a "License" section at the end of this file (as we have in other README files).

crates/validator/src/server/submit.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use crate::COMPONENT;
+use crate::db::insert_transactions;
+
+#[instrument(
+        target = COMPONENT,
+        name = "validator.submit_proven_transaction",
+        skip(conn, proven_tx),
+    )]
+pub fn submit_proven_transaction(
```
bobbinth: For this PR, I'd just stab out this function os that it doesn't do much more than emit telemetry.

crates/validator/src/server/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+            },
+            Err(err) => Err(err.into()),
+        }
+    }
+
+    /// Receives a proven block and validates it.
+    async fn validate_block(
+        &self,
+        request: tonic::Request<proto::blockchain::Block>,
+    ) -> Result<tonic::Response<proto::validator::SignedBlock>, tonic::Status> {
```
bobbinth: For this PR, we can probably omit this endpoint (and also remove it from the protobuf file).

crates/store/src/lib.rs
```diff
@@ -7,6 +7,14 @@ pub mod genesis;
 mod server;
 pub mod state;
 
+pub use db::{
+    ConnectionManager,
+    DatabaseTypeConversionError,
+    Db,
+    SqlTypeConvert,
+    configure_connection_on_creation,
+};
```
bobbinth: For this PR, I would make the validator "stateless" - and so, no changes to the store would be needed.

If we do want to share these components between the store and the validator, they should probably be moved into `node-utils` crate (i.e., the validator should not depend on store).

bin/node/src/commands/validator.rs
```diff
@@ -0,0 +1,64 @@
...
+            long = "grpc.timeout",
+            default_value = &duration_to_human_readable_string(DEFAULT_TIMEOUT),
+            value_parser = humantime::parse_duration,
+            value_name = "DURATION"
+        )]
+        grpc_timeout: Duration,
+
+        /// Directory in which to store the database and raw block data.
+        #[arg(long, env = ENV_DATA_DIRECTORY, value_name = "DIR")]
+        data_directory: PathBuf,
```
bobbinth: Similar to the previous comment: I'd probably remove this from this PR.

crates/validator/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+The validator requires:
+
+- **Address**: Network address for the gRPC server.
+- **gRPC Timeout**: Request timeout duration.
+- **Data Directory**: Path for SQLite database files.
+
+## License
+
+This project is [MIT licensed](./LICENSE).
```
bobbinth: Shouldn't this be `This project is [MIT licensed](../../LICENSE).`?

bobbinth (approved):
Looks good! Thank you!

PR #1295, opened 2025-10-15
feat: increment counter periodically

SantiagoPittella:
We need to change the PR's base after https://github.com/0xMiden/miden-node/pull/1276 gets merged.

TODOs:
- [x] Send a `get_account_details` request to retrieve the counter value.
- [x] Display value in the frontend.
- [x] Update docs and readme.
- [x] Add tracing.
- [x] Do not store wallet account to prevent account state mismatches.
- [x] Add new env vars to `.env`.
- [x] Merge `DataStore` implementations, and reduce duplicated code.

### DIFF
--- a/bin/network-monitor/src/commands/start.rs
+++ b/bin/network-monitor/src/commands/start.rs
@@ -20,11 +20,11 @@ pub async fn start_monitor(config: MonitorConfig) -> Result<()> {
     // Load configuration from command-line arguments and environment variables
     info!("Loaded configuration: {:?}", config);
 
-    if config.enable_otel {
-        miden_node_utils::logging::setup_tracing(OpenTelemetry::Enabled)?;
+    let _otel_guard = if config.enable_otel {
+        miden_node_utils::logging::setup_tracing(OpenTelemetry::Enabled)?
     } else {
-        miden_node_utils::logging::setup_tracing(OpenTelemetry::Disabled)?;
-    }
+        miden_node_utils::logging::setup_tracing(OpenTelemetry::Disabled)?
+    };
 
     let mut tasks = Tasks::new();
 


bin/network-monitor/src/deploy/account.rs
```diff
@@ -317,13 +271,11 @@ impl DataStore for MonitorDataStore {
...
-        let account = if account_id == account.id() {
-            account.to_owned()
-        } else {
-            self.counter_account.lock().await.to_owned()
-        };
+        let account =
+            self.accounts.get(&account_id).cloned().ok_or_else(|| DataStoreError::Other {
+                error_msg: "unknown account".into(),
+                source: None,
+            })?;
```
sergerad: nit: This is repeated 3x
```rust
DataStoreError::Other {
                error_msg: "unknown account".into(),
                source: None,
            })?;
```

Might be worth adding a fn for it instead (`get_account(&self)` or so)

bin/network-monitor/src/monitor/tasks.rs
```diff
@@ -214,7 +219,43 @@ impl Tasks {
...
+            ),
+        };
+
+        // Spawn the counter increment task
+        let (counter_tx, counter_rx) = watch::channel(initial_counter_status);
+        let config = config.clone();
+        let id = self
+            .handles
+            .spawn(async move {
+                Box::pin(run_counter_increment_task(config, counter_tx)).await.unwrap();
```
sergerad: nit: expect() with explanation over unwrap()

bin/network-monitor/src/counter.rs
```diff
@@ -0,0 +1,429 @@
...
+        .without_metadata_version()
+        .without_metadata_genesis()
+        .connect::<Rpc>()
+        .await
+}
+
+async fn get_genesis_block_header(rpc_client: &mut RpcClient) -> Result<BlockHeader> {
+    // Get the latest block header
+    let block_header_request = BlockHeaderByNumberRequest {
+        block_num: Some(BlockNumber::GENESIS.as_u32()), // Get latest block
```
sergerad: Is this really the way to request latest? Rather than `None`?

bin/network-monitor/src/counter.rs
```diff
@@ -0,0 +1,429 @@
...
+        let req = BlockHeaderByNumberRequest {
+            block_num: Some(target),
+            include_mmr_proof: None,
+        };
+        let resp = rpc_client.get_block_header_by_number(req).await?.into_inner();
+        if resp.block_header.is_some() {
+            return Ok(());
+        }
+        sleep(Duration::from_millis(500)).await;
+    }
```
sergerad, line 123: Do we want a timeout for this loop? Perhaps even just a log if the timeout is reached (if we don't want to error out).

bin/network-monitor/src/counter.rs
```diff
@@ -0,0 +1,429 @@
...
+    NoteScript,
+    SecretKey,
+)> {
+    let details = CounterIncrementDetails::default();
+    // Load accounts from files
+    let (wallet_account, secret_key) = match create_wallet_account() {
+        Ok(account) => account,
+        Err(e) => {
+            error!("Failed to create wallet account: {:?}", e);
+            return Err(anyhow::anyhow!("Failed to create wallet account"));
```
sergerad:
```suggestion
            anyhow::bail!("Failed to create wallet account");
```

bin/network-monitor/src/counter.rs
```diff
@@ -0,0 +1,429 @@
...
+        },
+    };
+
+    let counter_account = match load_counter_account(&config.counter_file) {
+        Ok(account) => account,
+        Err(e) => {
+            error!("Failed to load counter account: {:?}", e);
+            return Err(anyhow::anyhow!("Failed to load counter account"));
```
sergerad:
```suggestion
            anyhow::bail!("Failed to load counter account");
```

bin/network-monitor/src/counter.rs
```diff
@@ -0,0 +1,429 @@
...
+    pub current_value: Option<u64>,
+    /// Last transaction ID (if available).
+    pub last_tx_id: Option<String>,
+}
+
+async fn create_rpc_client(config: &MonitorConfig) -> Result<RpcClient> {
+    Builder::new(config.rpc_url.clone())
+        .with_tls()
+        .context("Failed to configure TLS for RPC client")
+        .unwrap()
```
sergerad: nit: expect instead of unwrap

bin/network-monitor/src/counter.rs
```diff
@@ -0,0 +1,429 @@
...
+                Status::Healthy
+            },
+            last_checked: crate::monitor::tasks::current_unix_timestamp_secs(),
+            error: last_error,
+            details: ServiceDetails::CounterIncrement(details.clone()),
+        };
+
+        if tx.send(status).is_err() {
+            error!("Failed to send counter increment status update");
+            return Err(anyhow::anyhow!("Failed to send counter increment status update"));
```
sergerad:
```suggestion
            anyhow::bail!("Failed to send counter increment status update");
```

bin/network-monitor/README.md
```diff
@@ -45,48 +46,50 @@ miden-network-monitor start --faucet-url http://localhost:8080 --enable-otel
...
-1. Checks if wallet and counter account files exist
-2. If files don't exist, creates new accounts:
-   - A wallet account with RpoFalcon512 authentication
-   - A counter program account with custom MASM script
-3. Saves both accounts to the specified files using the Miden AccountFile format
-4. Deploys the accounts to the network via RPC
-5. The counter program includes authentication logic that only allows the wallet account to increment the counter
-6. Both accounts are ready for use in Miden transactions
+**Optional Counter Account Management (only when counter is enabled):**
+When `--enable-counter` is set, the monitor ensures required counter account exist before starting the task:
```
sergerad:
```suggestion
When `--enable-counter` is set, the monitor ensures required counter account exists before starting the task:
```

bin/network-monitor/src/counter.rs
```diff
@@ -0,0 +1,429 @@
...
+        counter_account,
+        block_header,
+        mut data_store,
+        increment_script,
+        secret_key,
+    ) = setup_counter_increment(config.clone(), &mut rpc_client).await?;
+
+    let mut rng = ChaCha20Rng::from_os_rng();
+
+    loop {
```
sergerad, line 264: This is a really large loop and is a bit hard to follow. Could we break up the logic into smaller functions within this loop?

bin/network-monitor/src/counter.rs
SantiagoPittella: The comment is wrong, we want the genesis block (the implementation already does that).

bin/network-monitor/src/counter.rs
SantiagoPittella: I moved the different paths into functions

sergerad (approved):

bin/network-monitor/assets/index.html
bobbinth, line 1: Question: how does the monitor screenshot look like after this change?

SantiagoPittella: This is with the counter + rpc checks.
Screenshot 2025-11-04 at 09 48 46

bobbinth: Very nice! I would maybe change the title to "Network transactions" (as I think that's what we are testing, right?)

Also, probably not for this PR, but would be good to measure network transaction latency - i.e., the time it take before we submit a transaction and the counter value gets updated.

bin/network-monitor/src/assets/counter_program.masm
```diff
@@ -17,13 +18,13 @@ export.increment
     push.1 exec.account::get_item
     # => [owner_prefix, owner_suffix, 0, 0]
 
-    exec.active_note::get_sender
+    push.0 exec.input_note::get_sender
```
bobbinth: I think `active_note::get_sender` should work here - or is it failing for some reason?

SantiagoPittella: `active_note` should work, I'm changing it.

bin/network-monitor/README.md
```diff
@@ -206,11 +221,12 @@ The monitor automatically:
 ### Example Usage
 
 ```bash
-# Start monitor with default account files
-miden-network-monitor start --rpc-url https://testnet.miden.io:443
+# Start monitor with counter task and default account files
+miden-network-monitor start --enable-counter --rpc-url https://testnet.miden.io:443
```
igamigo: Not a strong opinion, but I'd make this enabled by default unless there's a specific reason why we should not. I think it might simplify the ops work by not having to know which things you are able to turn on.

bin/network-monitor/README.md
```diff
@@ -164,7 +172,14 @@ The monitor application provides real-time status monitoring for the following M
   - Response time measurement for challenge completion
   - Challenge difficulty monitoring
   - Transaction and note ID tracking from successful mints
-  - Automated testing every 30 seconds to verify faucet functionality
+  - Automated testing on a configurable interval to verify faucet functionality
+
+### Counter Increment
```
igamigo: nit: If we keep the `--enable-counter` flag (or even if we change it to `--disable-counter`), should it be documented here as well?

bin/network-monitor/src/config.rs
```diff
@@ -117,4 +123,14 @@ pub struct MonitorConfig {
         help = "Path where the counter account is located"
     )]
     pub counter_filepath: PathBuf,
+
+    /// The interval at which to increment the counter.
```
igamigo: "Increment the counter" here means sending a network note to the counter contract, right? If so, would be nice to specify this to explain how the counter is implemented (eg, it's not an account which increases its own counter).

bin/network-monitor/src/config.rs
```diff
@@ -109,6 +105,16 @@ pub struct MonitorConfig {
     )]
     pub wallet_filepath: PathBuf,
 
+    /// Enable the counter increment task (optional; disabled by default).
+    #[arg(
+        long = "enable-counter",
+        env = "MIDEN_MONITOR_ENABLE_COUNTER",
+        action = clap::ArgAction::SetTrue,
+        default_value_t = false,
+        help = "Whether to enable the counter increment task"
```
igamigo: nit: Here too I think it would be nice to specify that the counter contract is a network account that increases a counter through network notes. In general, because this service tests network transactions/accounts, I think all the settings could be named something related to that (eg `enable_ntx_service`)

bin/network-monitor/src/monitor/tasks.rs
```diff
@@ -214,7 +219,45 @@ impl Tasks {
         faucet_rx
     }
 
+    /// Spawn the counter increment task.
+    #[instrument(target = COMPONENT, name = "tasks.spawn-counter", skip_all)]
+    pub fn spawn_counter_increment(&mut self, config: &MonitorConfig) -> Receiver<ServiceStatus> {
```
igamigo: nit: Since the service name is "Network Transactions", maybe this could be named similarly


```suggestion
    pub fn spawn_network_transactions_service(&mut self, config: &MonitorConfig) -> Receiver<ServiceStatus> {
```

bin/network-monitor/src/deploy/mod.rs
igamigo, line 263: I think it would be interesting to investigate why we can't remove this (if the vault is fully loaded on `get_transaction_inputs`), should we create an issue? Or maybe @PhilippGackstatter knows.

bin/network-monitor/src/commands/start.rs
```diff
@@ -45,11 +47,23 @@ pub async fn start_monitor(config: MonitorConfig) -> Result<()> {
         None
     };
 
+    // Initialize the counter increment task only if enabled.
+    let counter_rx = if config.enable_counter {
+        // Ensure accounts exist before starting monitoring tasks
+        ensure_accounts_exist(&config.wallet_filepath, &config.counter_filepath, &config.rpc_url)
+            .await?;
```
igamigo: Unless the accounts are used for something else, I would put the account creation/loading within the service, only for a slightly more encapsulated organization.

igamigo (approved):
Awesome! I left some comments (most of which are minor) but LGTM. I think I'd name the user-facing flags and features in such a way that it reflects that it tests network transactions and accounts. But feel free to disregard if you prefer otherwise (or leave that for a different PR)

bin/network-monitor/src/counter.rs
```diff
@@ -145,6 +145,29 @@ async fn fetch_counter_value(
...
+async fn fetch_wallet_account(
+    rpc_client: &mut RpcClient,
+    account_id: AccountId,
+) -> Result<Option<Account>> {
+    let id_bytes: [u8; 15] = account_id.into();
+    let req = miden_node_proto::generated::account::AccountId { id: id_bytes.to_vec() };
+    let resp = rpc_client.get_account_details(req).await;
+
+    // If the RPC call fails, return None
+    if resp.is_err() {
```
igamigo, line 157: We could check for the error code here (to see that it doesn't catch all errors), but not sure this would make a big difference.

SantiagoPittella: This endpoint should fail only if the account does not exists, or if one of the components is unavailable, but that would be checked in the RPC status checks.

PR #1296, opened 2025-10-15
feat: db account storage cleanup task

### LINKED ISSUE #1355: Add cleanup routine for outdated historical storage entries

drahnr:
~Ref #1185 step 5.~

Closes #1355 

* [x] cleanup accounts sql table
* [x] cleanup vault / storage map smt forest


crates/store/src/db/mod.rs
```diff
@@ -47,8 +47,10 @@ pub(crate) mod schema;
 
 pub type Result<T, E = DatabaseError> = std::result::Result<T, E>;
 
+#[derive(Clone)]
 pub struct Db {
     pool: deadpool_diesel::Pool<ConnectionManager, deadpool::managed::Object<ConnectionManager>>,
+    notify_cleanup_task: tokio::sync::mpsc::Sender<BlockNumber>,
```
sergerad, line 60: Depending on what we do with the validator database (whether we re-use this or not), it would be good to make this task optional/disable-able

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -915,3 +924,169 @@ pub(crate) struct AccountStorageMapRowInsert {
...
+) -> Result<usize, DatabaseError> {
+    // Strategy: For each vault_key, keep the latest entry (is_latest_update=true) and the
+    // most recent MAX_HISTORICAL_ENTRIES_PER_ACCOUNT historical entries, deleting the rest.
+    //
+    // Unfortunately, Diesel doesn't support window functions (ROW_NUMBER OVER) in its type-safe
+    // API, so we must use raw SQL for this complex deletion pattern. This is necessary because:
+    // 1. We need to partition by vault_key and order by block_num
+    // 2. We need to delete based on row ranking within each partition
+    // 3. The table is created WITHOUT ROWID, so we use the primary key (account_id, block_num,
+    //    vault_key)
```
bobbinth: I'm not sure we need to keep `MAX_HISTORICAL_ENTRIES_PER_ACCOUNT` - we just need to make sure we have the data for the last $n$ blocks. So, query could be as simple as:
```sql
DELETE FROM account_vault_assets
WHERE
    block_num < chain_tip - n
    AND is_latest_update = false
```
To make this query efficient, we may need to add indexes on `block_num` and `is_latest_update` fields.

This methodology should also apply to the `account_storage_map_values` table (and in the future to the `accounts` table).

PR #1298, opened 2025-10-19
feat: enforce accept header genesis parameter for write RPC methods

### LINKED ISSUE #1123: Force genesis commitment for write RPC methods
Consider requiring the genesis commitment in the `accept` header for non-read-only RPC methods. At present this is `submitProvenTransaction` but we will also be adding a `submitProvenBatch` endpoint in the future.

https://github.com/0xMiden/miden-base/pull/1659#discussion_r2246621238

radik878:
- Implemented path-based enforcement in the Accept header middleware to require a matching genesis commitment for write methods SubmitProvenTransaction and SubmitProvenBatch. If missing, empty, or mismatched, requests are rejected. Read-only methods retain optional genesis but still validate if provided.
- Added unit tests to cover missing/empty/mismatch rejection and successful pass with matching genesis.

fix #1123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Normalize response size in endpoints to 4 MB ([#1357](https://github.com/0xMiden/miden-node/pull/1357)).
 - [BREAKING] Renamed `ProxyWorkerStatus::address` to `ProxyWorkerStatus::name` ([#1348](https://github.com/0xMiden/miden-node/pull/1348)).
 - Remove `trait AccountTreeStorage` ([#1352](https://github.com/0xMiden/miden-node/issues/1352)).
+- [BREAKING] `SubmitProvenTransaction` now **requires** that the network's genesis commitment is set in the request's `ACCEPT` header ([#1298](https://github.com/0xMiden/miden-node/pull/1298)).
 
 ### Fixes
 


crates/rpc/src/server/accept.rs
sergerad:
```suggestion
            match genesis {
                None if require_genesis => {
                    // For write methods, the genesis parameter must be present.
                    continue;
                },
                Some(genesis) if genesis != self.genesis_commitment => {
                    // Present but mismatched.
                    continue;
                },
                _ => {},
            }
```

crates/rpc/src/server/accept.rs
```diff
@@ -89,13 +89,22 @@ impl AcceptHeaderLayer {
...
 
     /// Parses the `Accept` header's contents, searching for any media type compatible with our
-    /// RPC version and genesis commitment.
-    fn negotiate(&self, accept: &str) -> Result<(), AcceptHeaderError> {
+    /// RPC version and genesis commitment, optionally requiring the presence of the `genesis`
+    /// parameter in the media range.
+    fn negotiate_with_requirement(
+        &self,
+        accept: &str,
+        require_genesis: bool,
```
tindzk: We could use an enum here:

```rust
pub enum NegotiationGenesis {
  Optional,
  Mandatory
}
```

Then, the old function name can be retained:

```rust
fn negotiate(&self, accept: &str, genesis: NegotiationGenesis) { ... }
```

And calls will be more readable without IDE hints:

```rust
self.verifier.negotiate(header, NegotiationGensis::Mandatory)
```

radik878: > We could use an enum here:
> 
> ```rust
> pub enum NegotiationGenesis {
>   Optional,
>   Mandatory
> }
> ```
> 
> Then, the old function name can be retained:
> 
> ```rust
> fn negotiate(&self, accept: &str, genesis: NegotiationGenesis) { ... }
> ```
> 
> And calls will be more readable without IDE hints:
> 
> ```rust
> self.verifier.negotiate(header, NegotiationGensis::Mandatory)
> ```

Ok, did it

crates/rpc/src/server/accept.rs
radik878: > 

i can't commit it, i see this: "Outdated suggestions cannot be applied"

crates/rpc/src/server/accept.rs
```diff
@@ -195,14 +227,34 @@ where
     }
 
     fn call(&mut self, request: http::Request<B>) -> Self::Future {
+        // Determine if this RPC method requires the `genesis` parameter.
+        let path = request.uri().path();
+        let requires_genesis =
+            matches!(path, "/rpc.Api/SubmitProvenTransaction" | "/rpc.Api/SubmitProvenBatch");
```
Mirko-von-Leipzig: It would be nice if this was surfaced at the server itself to make it more visible at the "business logic" level.

What about adding a `require_genesis_methods: Vec<&'static str>` to `AcceptHeaderLayer` which you can then use here?

The caller (i.e. server) then does something like:

```rust
let accept_header = AcceptHeaderLayer::new(...)
   .with_genesis_enforced_method("SubmitProvenTransaction")
   .with_genesis_enforced_method("SubmitProvenBatch");
```

This will make it more obvious for the caller what is going on.

radik878: > It would be nice if this was surfaced at the server itself to make it more visible at the "business logic" level.
> 
> What about adding a `require_genesis_methods: Vec<&'static str>` to `AcceptHeaderLayer` which you can then use here?
> 
> The caller (i.e. server) then does something like:
> 
> ```rust
> let accept_header = AcceptHeaderLayer::new(...)
>    .with_genesis_enforced_method("SubmitProvenTransaction")
>    .with_genesis_enforced_method("SubmitProvenBatch");
> ```
> 
> This will make it more obvious for the caller what is going on.

did it

crates/rpc/src/server/accept.rs
```diff
@@ -10,6 +10,12 @@ use miden_objects::{Word, WordError};
 use semver::{Comparator, Version, VersionReq};
 use tower::{Layer, Service};
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum NegotiationGenesis {
+    Optional,
+    Mandatory,
+}
```
sergerad, line 17: nit: `GenesisNegotiation` might make more sense

crates/rpc/src/server/accept.rs
```diff
@@ -150,16 +179,34 @@ impl AcceptHeaderLayer {
...
+                            continue;
+                        },
+                        _ => {},
+                    }
+                },
+                NegotiationGenesis::Optional => {
+                    if let Some(genesis) = genesis
+                        && genesis != self.genesis_commitment
+                    {
+                        continue;
```
sergerad: nit: would add same comment as above: `// Present but mismatched.`

sergerad (approved):

crates/rpc/src/server/accept.rs
```diff
@@ -150,16 +179,35 @@ impl AcceptHeaderLayer {
...
+                    }
+                },
+                GenesisNegotiation::Optional => {
+                    if let Some(genesis) = genesis
+                        && genesis != self.genesis_commitment
+                    {
+                        // Present but mismatched.
+                        continue;
+                    }
+                },
```
Mirko-von-Leipzig: Nit: we can simplify by collapsing conditions a bit


```suggestion
            match (genesis_mode, genesis) {
                (_, Some(value)) if value != self.genesis_commitment => continue,
                (GenesisNegotiation::Mandatory, None) => continue,
                _ => {},
            }
```

crates/rpc/src/server/accept.rs
```diff
@@ -195,14 +243,34 @@ where
     }
 
     fn call(&mut self, request: http::Request<B>) -> Self::Future {
+        // Determine if this RPC method requires the `genesis` parameter.
+        let path = request.uri().path();
+        let method_name = path.rsplit('/').next().unwrap_or("");
```
Mirko-von-Leipzig:
```suggestion
        let method_name = path.rsplit('/').next().unwrap_or_default();
```

crates/rpc/src/server/accept.rs
```diff
@@ -356,7 +426,41 @@ mod tests {
...
+    fn write_accepts_matching_genesis_param() {
+        let layer = AcceptHeaderLayer::for_tests();
+
+        // Matching genesis only
+        let accept = "application/vnd.miden; genesis=0x00000000000000000000000000000000000000000000000000000000deadbeef";
+        assert!(layer.negotiate(accept, super::GenesisNegotiation::Mandatory).is_ok());
+
+        // Matching genesis with version
+        let accept = "application/vnd.miden; version=0.2.3; genesis=0x00000000000000000000000000000000000000000000000000000000deadbeef";
+        assert!(layer.negotiate(accept, super::GenesisNegotiation::Mandatory).is_ok());
```
Mirko-von-Leipzig: Could you rewrite this as an `rstest` to give each case its own test? And the same for the rejection above.

```suggestion
    #[rstest]
    #[case::matching_network(
        "application/vnd.miden; genesis=0x00000000000000000000000000000000000000000000000000000000deadbeef"
    )]
    #[case::matching_network_and_version(
        "application/vnd.miden; genesis=0x00000000000000000000000000000000000000000000000000000000deadbeef; version=0.2.3"
    )]
    #[test]
    fn request_with_mandadory_genesis_should_pass(
        #[case] accept: &'static str
    ) {
        AcceptHeaderLayer::for_tests()
            .negotiate(accept, super::GenesisNegotiation::Mandatory)
            .unwrap();
    }
```

PR #1301, opened 2025-10-21
feat: create data and accounts dir if not already existing

lima-limon-inc:
Currently, the `miden-node` utility requires for the `data` and `accounts` directories to already exist in order for the `miden-node bundled bootstrap` command to function correctly.

This PR makes it so that `miden-node` CLI will create the passed in directories if they do not exist already, but will error out if they do, following the already present behavior regarding file creation inside said directories. 

This is used in `midenup` in order to auto-initialize the node without requiring previous user input and is related to this discussion: https://github.com/0xMiden/miden-client/issues/1366

### DIFF
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -194,7 +194,7 @@ impl StoreCommand {
                             .file_name()
                             .unwrap_or(std::ffi::OsStr::new("directory"))
                             .display(),
-                        accounts_directory.display()
+                        directory.display()
                     )
                 })?;
             }


bin/node/src/commands/store.rs
```diff
@@ -180,6 +180,9 @@ impl StoreCommand {
         // Write the accounts to disk
         for item in secrets.as_account_files(&genesis_state) {
             let AccountFileWithName { account_file, name } = item?;
+            fs_err::create_dir(accounts_directory).with_context(|| {
+                format!("failed to create accounts_directory at {}", accounts_directory.display())
+            })?;
```
sergerad: Would this not fail if there is more item in this loop? Because the accounts_directory would be created more than once?

sergerad: Even if create_dir doesn't fail on collision, might prefer to take it out of the loop?

lima-limon-inc: Great point, will change ASAP

lima-limon-inc: > Even if create_dir doesn't fail on collision, might prefer to take it out of the loop?

Updated in this commit: https://github.com/0xMiden/miden-node/pull/1301/commits/0d1b08045b45719fb7760b9bfc4b5e6f45a779b8

I added an additional to check if the directory already exists and is empty.

sergerad (approved):

bin/node/src/commands/store.rs
```diff
@@ -177,6 +177,29 @@ impl StoreCommand {
...
+                }
+            } else {
+                fs_err::create_dir(directory).with_context(|| {
+                    format!(
+                        "failed to create {} at {}",
+                        directory
+                            .file_name()
+                            .unwrap_or(std::ffi::OsStr::new("directory"))
+                            .display(),
+                        accounts_directory.display()
```
TomasArrachea: Shouldn't this be:
```suggestion
                        directory.display()
```

lima-limon-inc: Totally! Nice catch, fixed here: https://github.com/0xMiden/miden-node/pull/1301/commits/da1fb40df3387fe17850a66ab8cc9eb857f77df6

PR #1308, opened 2025-10-23
fix: trim MAX_NOTE_ATTEMPTS to 1 - for `main`

drahnr:
Reduces noise significantly on testnet, and we don't have many practical error cases _yet_. In the future we should increment it again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## v0.11.3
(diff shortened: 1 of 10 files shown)


mmagician (approved):
Danke!

igamigo:
This is probably something that is good to add anyway to make things more stable, but are we sure it fixes the crash or are there other things that we should look for before doing a release?

PR #1312, opened 2025-10-24
address node crash in v0.11.3

drahnr:
When doing initial `State::load` we skip any note whoes account state we cannot fetch. In the `add_transaction` scope we do not include the relevant note in the `available_*` to include set, but we keep iterating over the full set of _nullifiers_, including those note's nullifiers we _do not_ have available yet map back to notes which exist for unknown accounts. Later on the node then topples over on an `expect()` call, which relied on the coherence of two mappings.

### Changes:

* Remove the `expect`s, and be defensive in adjacent calls, use `tracing::warn!` instead, be verbose.
* ensure the nullifier is skipped in follow-up loops
* Add a `tracing::warn!` for a potential, unrelated, silent dead-lock. We've seen some hangs, but no indication to what caused it - this is an attempt to get some indication in the logs for that.
* Chore: get some tests out of the way, particularly when searching this makes masking testing instances easy to filter out

### Context:

The next release of the node will do away with the current NTX in it's entirety, as such the current changeset is _good enough_.

### DIFF
--- a/bin/remote-prover/src/proxy/worker.rs
+++ b/bin/remote-prover/src/proxy/worker.rs
@@ -205,7 +205,7 @@ impl Worker {
                         self.address(),
                         e
                     );
-                    return Err(e.to_string());
+                    return Err(e.clone());
                 },
             };
 
--- a/crates/store/src/db/tests.rs
+++ b/crates/store/src/db/tests.rs
@@ -408,7 +408,7 @@ fn sql_unconsumed_network_notes() {
     // An arbitrary public account (network note tag requires public account).
     create_block(conn, block_num);
 
-    let account_notes = vec![
+    let account_notes = [
         make_account_and_note(conn, block_num, [0u8; 32], AccountStorageMode::Public),
         make_account_and_note(conn, block_num, [1u8; 32], AccountStorageMode::Network),
     ];


crates/block-producer/src/mempool/inflight_state/mod.rs
```diff
@@ -106,7 +106,9 @@ impl InflightState {
     ) -> Result<BTreeSet<TransactionId>, AddTransactionError> {
         // Separate verification and state mutation so that a rejected transaction
         // does not impact the state (atomicity).
-        self.verify_transaction(tx)?;
+        self.verify_transaction(tx).inspect_err(|e| {
+            tracing::warn!("Error on auth'd tx submission: {e:?}");
```
sergerad: Just checking we don't want error! instead

crates/ntx-builder/src/state/account.rs
```diff
@@ -180,13 +180,14 @@ impl AccountState {
...
-            .remove(&nullifier)
-            .expect("note must be available to nullify");
-
-        self.nullified_notes.insert(nullifier, note);
+    pub fn add_nullifier(&mut self, nullifier: Nullifier) -> Result<(), ()> {
+        if let Some(note) = self.available_notes.remove(&nullifier) {
+            self.nullified_notes.insert(nullifier, note);
+            Ok(())
+        } else {
+            tracing::warn!(%nullifier, "note must be available to nullify");
```
sergerad, line 186: Same here - maybe error! ?

sergerad (approved):

crates/block-producer/src/mempool/inflight_state/mod.rs
drahnr: I considered it, but it's user input after all, so I don't think we should print errors on the remote, I was leaning more towards `info!` really, but since it's rather helpful to debug I kept it at `warn!`.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -342,14 +342,19 @@ impl State {
             self.in_progress.remove(&prefix);
             tx_impact.account_delta = Some(prefix);
         }
+        // failed to fetch account, so ignore the nullifier
```
tindzk: The comment seems to be on the wrong line.

I would rewrite this loop as follows to make it easier to read:

```rust
        for note in network_notes {
            let prefix = note.account_prefix();

            // Skip and ignore nullifier if note targets a non-existent network account
            let Some(account) = self.fetch_account(prefix).await? else {
                tracing::warn!("could not fetch account from network: {:?}", prefix);
                continue;
            };

            tx_impact.notes.insert(note.nullifier());
            account.add_note(note.clone());
            self.nullifier_idx.insert(note.nullifier(), prefix);
        }
```

crates/ntx-builder/src/state/account.rs
```diff
@@ -195,9 +196,11 @@ impl AccountState {
     ///
     /// Panics if the associated note is not marked as nullified.
```
tindzk: This is not the case anymore.

crates/ntx-builder/src/state/account.rs
```diff
@@ -180,13 +180,14 @@ impl AccountState {
     /// # Panics
     ///
     /// Panics if the note does not exist or was already nullified.
```
tindzk: This is not the case anymore.

crates/block-producer/src/mempool/inflight_state/mod.rs
tindzk: I agree that `warn!` is a better fit here, but I would suggest emitting an event instead:

```rust
tracing::warn!(event = %"VerifyTransactionFailed", op = %"AddTransaction, error = ?e);
```

Then, it will be easier to quantify how frequently this failure occurs.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -343,13 +343,17 @@ impl State {
             tx_impact.account_delta = Some(prefix);
         }
         for note in network_notes {
-            tx_impact.notes.insert(note.nullifier());
             let prefix = note.account_prefix();
+
+            // Skip and ignore nullifier if note targets a non-existent network account
+            let Some(account) = self.fetch_account(prefix).await? else {
+                tracing::warn!("could not fetch account from network: {:?}", prefix);
```
bobbinth, line 351: Technically, this isn't wrong - so, not sure if we should emit a warning here (maybe just `debug!` is better?). Basically, it is OK for transactions to create notes that target non-existent network accounts with the expectation that these accounts will be created later on.

crates/ntx-builder/src/state/mod.rs
```diff
@@ -343,13 +343,17 @@ impl State {
...
+                continue;
+            };
+
+            tx_impact.notes.insert(note.nullifier());
```
bobbinth: Moving this line after we check if the account exists is the main change in this section of the code, right? I don't remember exactly how the list of notes nullified by a given transaction is used, but it feels a bit odd to me that we filter these notes out at this stage. Is there a way to handle this at the point where we iterate over this set of nullifiers instead?

drahnr: > Moving this line after we check if the account exists is the main change in this section of the code, right? 

The important piece is that `self.nullifier_idx.insert(note.nullifier(), prefix);` is not called when the `account` is not available. The one invariant that changed is the line you outlined not happening in case of non-present accounts. Re-reviewing, I am pretty certain that we should retain them regardless since it has an effect if it's considered in-flight or not, which has further effects.

> I don't remember exactly how the list of notes nullified by a given transaction is used, but it feels a bit odd to me that we filter these notes out at this stage. Is there a way to handle this at the point where we iterate over this set of nullifiers instead?

We could, this is a bit later where there is another invariant where we assume resolution from `nullifier_idx` to `account_prefix` works, in `commit_transaction`, but making that failable in addition would be an option.

PR #1318, opened 2025-10-30
chore: update `miden-base`

juan518munoz:
Update `miden-block-prover`, `miden-lib`, `miden-objects`, `miden-testing`, `miden-tx` & `miden-tx-batch-prover` to their latest crates.io release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ miden-tx              = { default-features = false, version = "0.12" }
 miden-tx-batch-prover = { version = "0.12" }
 
 # Other miden dependencies. These should align with those expected by miden-base.
-miden-air = { features = ["std", "testing"], version = "0.19.0" }
+miden-air = { features = ["std", "testing"], version = "0.19" }
 
 # External dependencies
 anyhow             = { version = "1.0" }
--- a/crates/ntx-builder/src/state/account.rs
+++ b/crates/ntx-builder/src/state/account.rs
@@ -250,13 +250,22 @@ impl AccountState {
 // ================================================================================================
 
 #[derive(Clone)]
-pub struct NetworkAccountUpdate(AccountDelta);
+pub enum NetworkAccountEffect {
+    Created(Account),
+    Updated(AccountDelta),
+}
 
-impl NetworkAccountUpdate {
+impl NetworkAccountEffect {
     pub fn from_protocol(update: AccountUpdateDetails) -> Option<Self> {
         let update = match update {
             AccountUpdateDetails::Private => return None,
-            AccountUpdateDetails::Delta(update) => Self(update),
+            AccountUpdateDetails::Delta(update) if update.is_full_state() => {
+                NetworkAccountEffect::Created(
+                    Account::try_from(&update)
+                        .expect("Account should be derivable by full state AccountDelta"),
+                )
+            },
+            AccountUpdateDetails::Delta(update) => NetworkAccountEffect::Updated(update),
         };
 
         update.account_id().is_network().then_some(update)
@@ -267,12 +276,11 @@ impl NetworkAccountUpdate {
         self.account_id().try_into().unwrap()
     }
 
-    pub fn account_delta(&self) -> &AccountDelta {
-        &self.0
-    }
-
     fn account_id(&self) -> AccountId {
-        self.0.id()
+        match self {
+            NetworkAccountEffect::Created(acc) => acc.id(),
+            NetworkAccountEffect::Updated(delta) => delta.id(),
+        }
     }
 }
 
--- a/crates/store/src/db/models/queries/accounts.rs
+++ b/crates/store/src/db/models/queries/accounts.rs
@@ -1,4 +1,3 @@
-use std::borrow::Cow;
 use std::ops::RangeInclusive;
 
 use diesel::prelude::{Queryable, QueryableByName};
@@ -726,7 +725,7 @@ pub(crate) fn upsert_accounts(
             None
         };
 
-        let full_account: Option<Cow<Account>> = match update.details() {
+        let full_account: Option<Account> = match update.details() {
             AccountUpdateDetails::Private => None,
             AccountUpdateDetails::Delta(delta) if delta.is_full_state() => {
                 let account = Account::try_from(delta)?;
@@ -759,7 +758,7 @@ pub(crate) fn upsert_accounts(
                     }
                 }
 
-                Some(Cow::Owned(account))
+                Some(account)
             },
             AccountUpdateDetails::Delta(delta) => {
                 let mut rows = select_details_stmt(conn, account_id)?.into_iter();
@@ -782,9 +781,9 @@ pub(crate) fn upsert_accounts(
                     }
                 }
 
-                // apply delta to the account; we need to do this before we process asset
-                // updates because we currently need to get the
-                // current value of fungible assets from the account
+                // apply delta to the account; we need to do this before we process asset updates
+                // because we currently need to get the current value of fungible assets from the
+                // account
                 let account = apply_delta(account, delta, &update.final_state_commitment())?;
 
                 // --- process asset updates ----------------------------------
@@ -818,11 +817,11 @@ pub(crate) fn upsert_accounts(
                     )?;
                 }
 
-                Some(Cow::Owned(account))
+                Some(account)
             },
         };
 
-        if let Some(code) = full_account.as_ref().map(|account| account.code()) {
+        if let Some(code) = full_account.as_ref().map(Account::code) {
             let code_value = AccountCodeRowInsert {
                 code_commitment: code.commitment().to_bytes(),
                 code: code.to_bytes(),
--- a/crates/ntx-builder/src/state/mod.rs
+++ b/crates/ntx-builder/src/state/mod.rs
@@ -2,7 +2,7 @@ use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque};
 use std::num::NonZeroUsize;
 
-use account::{AccountState, InflightNetworkNote, NetworkAccountUpdate};
+use account::{AccountState, InflightNetworkNote, NetworkAccountEffect};
 use anyhow::Context;
 use miden_node_proto::domain::account::NetworkAccountPrefix;
 use miden_node_proto::domain::mempool::MempoolEvent;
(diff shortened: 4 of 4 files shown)


bobbinth:
With #1320 merged, let's merge the latest `next` into this branch.

crates/rpc/src/server/api.rs
```diff
@@ -374,20 +375,6 @@ impl api_server::Api for RpcService {
...
-        // Verify that the delta commitment matches the actual delta
-        if let AccountUpdateDetails::Delta(delta) = tx.account_update().details() {
-            let computed_commitment = delta.to_commitment();
-
-            if computed_commitment != delta_commitment {
-                return Err(Status::invalid_argument(
-                    "Account delta commitment does not match the actual account delta",
-                ));
-            }
-        }
```
juan518munoz, line 389: This check is now done elsewhere, inside of `ProvenTransaction::validate()`
https://github.com/0xMiden/miden-base/blob/next/crates/miden-objects/src/transaction/proven_tx.rs#L170-L176

Cargo.lock
TomasArrachea, line 1: This seems to be downgrading a bunch of dependencies, we should clean the lockfile to only update the miden related ones.

crates/rpc/Cargo.toml
```diff
@@ -19,6 +19,7 @@ anyhow                 = { workspace = true }
 futures                = { version = "0.3" }
 http                   = { workspace = true }
 mediatype              = { version = "0.20" }
+miden-lib              = { workspace = true }
```
TomasArrachea: I think this should be under dev-dependencies

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -698,98 +698,100 @@ pub(crate) fn upsert_accounts(
...
-                debug_assert_eq!(account_id, account.id());
-
-                if account.commitment() != update.final_state_commitment() {
-                    return Err(DatabaseError::AccountCommitmentsMismatch {
-                        calculated: account.commitment(),
-                        expected: update.final_state_commitment(),
-                    });
-                }
+            AccountUpdateDetails::Delta(delta) => {
+                if delta.is_full_state() {
```
TomasArrachea: nit: we could try to keep the match by handling the cases like:
```rust
match update.details() {
     AccountUpdateDetails::Private => None,
     AccountUpdateDetails::Delta(delta) if delta.is_full_state() => { 
         // new account
     },
     AccountUpdateDetails::Delta(delta) => {
         // update account
     },
}

File: crates/rpc/Cargo.toml (hunk 2 above)
Comment (juan518munoz) [reply crates/rpc/Cargo.toml]: Moved to `[dev-dependencies]` in [dacfda4](https://github.com/0xMiden/miden-node/pull/1318/commits/dacfda42187f53d5273ab3c7f47a9ede5e59d5bf)

File: Cargo.lock
Comment (juan518munoz) [reply Cargo.lock:1]: Yeah good catch, didn't look deeply into the lockfile changes and directly assumed all modifications were consecuential.

Reduced to lockifle diff to its minimun in [dacfda4](https://github.com/0xMiden/miden-node/pull/1318/commits/dacfda42187f53d5273ab3c7f47a9ede5e59d5bf)

File: crates/store/src/db/models/queries/accounts.rs (hunk 3 above)
Comment (juan518munoz) [reply crates/store/src/db/models/queries/accounts.rs]: I like the suggested sintax much more that the one currently used.

Applied it in [dacfda4](https://github.com/0xMiden/miden-node/pull/1318/commits/dacfda42187f53d5273ab3c7f47a9ede5e59d5bf)

File: crates/ntx-builder/src/state/account.rs [hunk 4]
```diff
@@ -251,15 +251,13 @@ impl AccountState {
 
 #[derive(Clone)]
 pub enum NetworkAccountUpdate {
-    New(Account),
     Delta(AccountDelta),
 }
```
bobbinth: This probably doesn't need to be an enum any more (since there is only one variant).

But also, maybe we keep it as it was and do the conversion from `AccountDelta` to `Account` in the `NetworkAccountUpdate::from_protocol()` method?

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -887,12 +888,13 @@ pub(crate) struct AccountAssetRowInsert {
 impl AccountAssetRowInsert {
     pub(crate) fn new(
         account_id: &AccountId,
-        vault_key: &Word,
+        vault_key: &VaultKey,
```
bobbinth: I think in the latest `next` on `miden-base` this is now called `AssetVaultKey`.

bobbinth (commented):
Looks good! Thank you! I left a couple of comments inline - the main one is about updating this to the latest `next` from `miden-base`.

crates/store/src/db/models/queries/accounts.rs
juan518munoz: Updated to latest `miden-base` version in [6c49f05](https://github.com/0xMiden/miden-node/pull/1318/commits/6c49f05dd4510b6727c5ca5580aebefcb7e8d002)

juan518munoz:
The failing CI job appears to be caused by an out of disk error.

https://github.com/0xMiden/miden-node/actions/runs/19076938756/job/54502171493?pr=1318

crates/ntx-builder/src/state/account.rs
bobbinth: Should we address this?

bobbinth:
@huitseeker - any idea why the MSRV checks are failing?

crates/ntx-builder/src/state/account.rs
juan518munoz: I missed the comment, sorry!

I've changed the `enum` for a  tuple `struct`.

```
pub struct NetworkAccountUpdate(AccountDelta);
```

In [e8d2b53](https://github.com/0xMiden/miden-node/pull/1318/commits/e8d2b5323d9a87a140831635aea9c52d98481c40)

Cargo.toml
```diff
@@ -49,15 +49,15 @@ miden-node-validator        = { path = "crates/validator", version = "0.12" }
...
-miden-lib             = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
-miden-objects         = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
-miden-testing         = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
-miden-tx              = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
-miden-tx-batch-prover = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-block-prover    = { version = "0.12" }
+miden-lib             = { version = "0.12" }
+miden-objects         = { version = "0.12" }
+miden-testing         = { version = "0.12" }
+miden-tx              = { version = "0.12" }
```
igamigo: Let's keep the `default-features = false` for `miden-tx` and `miden-objects` (unsure if they were of any specific use but it shouldn't be necessary to change this now)

juan518munoz: Addressed in [edd94b9](https://github.com/0xMiden/miden-node/pull/1318/commits/edd94b9123590db112320d42fee49dcf893f93ca)

Cargo.toml
```diff
@@ -49,15 +49,15 @@ miden-node-validator        = { path = "crates/validator", version = "0.12" }
...
+miden-block-prover    = { version = "0.12" }
+miden-lib             = { version = "0.12" }
+miden-objects         = { version = "0.12", default-features = false }
+miden-testing         = { version = "0.12" }
+miden-tx              = { version = "0.12", default-features = false }
+miden-tx-batch-prover = { version = "0.12" }
 
 # Other miden dependencies. These should align with those expected by miden-base.
-miden-air = { features = ["std", "testing"], version = "0.18.1" }
+miden-air = { features = ["std", "testing"], version = "0.19.0" }
```
bobbinth: nit: let's use `0.19` here instead of `0.19.0`.

crates/ntx-builder/src/state/account.rs
```diff
@@ -250,17 +250,13 @@ impl AccountState {
 // ================================================================================================
 
 #[derive(Clone)]
-pub enum NetworkAccountUpdate {
-    New(Account),
-    Delta(AccountDelta),
-}
+pub struct NetworkAccountUpdate(AccountDelta);
```
bobbinth: Looking at this now, it is not too clear why we need the wrapper here - could be just `AccountDelta`, right?

On the other hand, for our purposes, I think maybe keeping it as enum is still fine and maybe we could make the naming more explicit - e.g., something like:

```rust
pub enum NetworkAccountEffect {
    Created(Account),
    Updated(AccountDelta),
}
```

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -494,7 +494,7 @@ impl TryFrom<AccountVaultUpdateRaw> for AccountVaultValue {
     type Error = DatabaseError;
 
     fn try_from(raw: AccountVaultUpdateRaw) -> Result<Self, Self::Error> {
-        let vault_key = Word::read_from_bytes(&raw.vault_key)?;
+        let vault_key = AssetVaultKey::new_unchecked(Word::read_from_bytes(&raw.vault_key)?);
```
bobbinth, line 496: Not for this PR, but we should probably have a way to serialize/deserialize `AssetVaultKey`. cc @PhilippGackstatter.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -726,9 +726,10 @@ pub(crate) fn upsert_accounts(
             None
         };
 
-        let full_account = match update.details() {
+        let full_account: Option<Cow<Account>> = match update.details() {
```
bobbinth: Do we still need `Cow` here? Seems like all branches return the `Owned` variant.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -781,9 +782,9 @@ pub(crate) fn upsert_accounts(
                     }
                 }
 
-                // apply delta to the account; we need to do this before we process asset updates
-                // because we currently need to get the current value of fungible assets from the
-                // account
+                // apply delta to the account; we need to do this before we process asset
+                // updates because we currently need to get the
+                // current value of fungible assets from the account
```
bobbinth: nit: comment formatting looks off here.

juan518munoz: Reverted back in [a54214c](https://github.com/0xMiden/miden-node/pull/1318/commits/a54214c1572a0d1adf51a85d9469af30cef0b7df)

crates/store/src/db/models/queries/accounts.rs
juan518munoz: We don't, removed in [a54214c](https://github.com/0xMiden/miden-node/pull/1318/commits/a54214c1572a0d1adf51a85d9469af30cef0b7df)

crates/ntx-builder/src/state/account.rs
juan518munoz: Reimpl as `enum` in [a54214c](https://github.com/0xMiden/miden-node/pull/1318/commits/a54214c1572a0d1adf51a85d9469af30cef0b7df)

bobbinth (approved):
All looks good! Thank you!

Issue #1262, opened 2025-10-01
Upgrade dependencies


Mirko-von-Leipzig:
General bumping of dependencies.

I suggest doing a commit per dependency/group so its easier to review.

The `tonic` and `open-telemetry` dependencies should be done asap to see if they solve some of our monitoring issues.

PR #1319, opened 2025-10-30
feat: update dependencies

TomasArrachea:
closes https://github.com/0xMiden/miden-node/issues/1262

Updated all dependencies to the last available version. Notably, the `tonic-build` crate changed to `tonic-prost-build` and it reexports most of the needed structs.

Updated dependencies:
- indexmap 2.7 → 2.12
- prost 0.13 → 0.14
- rstest 0.25 → 0.26
- toml 0.8 → 0.9
- tonic 0.13 → 0.14
- tonic-reflection 0.13 → 0.14
- tonic-build 0.13 -> tonic-prost-build 0.14
- protox 0.8 -> 0.9
- pingora 0.5 → 0.6
- pingora-core 0.5 → 0.6
- pingora-limits 0.5 → 0.6
- pingora-proxy 0.5 → 0.6
- opentelemetry-otlp 0.30 → 0.31
- opentelemetry_sdk 0.30 → 0.31
- tracing-forest 0.1 → 0.2
- tracing-opentelemetry 0.31 → 0.32

New dependencies:
- tonic-prost 0.14

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1538,6 +1538,8 @@ dependencies = [
  "allocator-api2",
  "equivalent",
  "foldhash 0.1.5",
+ "rayon",
+ "serde",
 ]
 
 [[package]]
@@ -2304,7 +2306,7 @@ dependencies = [
 [[package]]
 name = "miden-block-prover"
 version = "0.12.0"
-source = "git+https://github.com/0xMiden/miden-base?branch=next#f23bf31ffff5973969375cd5c379cb9815aa27ab"
+source = "git+https://github.com/0xMiden/miden-base?branch=next#f14d0eb719c6e124e4101eb8a75c279edc00c29c"
 dependencies = [
  "miden-lib",
  "miden-objects",
@@ -2313,9 +2315,9 @@ dependencies = [
 
 [[package]]
 name = "miden-core"
-version = "0.18.2"
+version = "0.18.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "996f8d8e2a4dd4eef54225f6f116c73b14f03256bc23ea5d4260fe803dd103e2"
+checksum = "da3298dd3c75b119f9031631d0087d8d7de66ae5ac4da80ba7f243d2035b2a1f"
 dependencies = [
  "enum_dispatch",
  "miden-crypto",
@@ -2340,6 +2342,7 @@ dependencies = [
  "flume",
  "getrandom 0.2.16",
  "glob",
+ "hashbrown 0.15.5",
  "hkdf",
  "k256",
  "num",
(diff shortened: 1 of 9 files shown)


crates/block-producer/src/mempool/tests.rs
```diff
@@ -46,7 +46,7 @@ async fn add_transaction_traces_are_correct() {
 
     let span_data = rx_export.recv().await.unwrap();
     assert_eq!(span_data.name, "mempool.add_transaction");
-    assert!(span_data.attributes.iter().any(|kv| kv.key == "code.namespace".into()
+    assert!(span_data.attributes.iter().any(|kv| kv.key == "code.module.name".into()
```
bobbinth, line 49: Does this mean that span names are now different? If so, does this have a bigger impact on telemetry data?

bin/network-monitor/Cargo.toml
```diff
@@ -29,6 +29,6 @@ serde            = { features = ["derive"], version = "1.0" }
 serde_json       = { version = "1.0" }
 sha2             = { version = "0.10" }
 tokio            = { features = ["full"], workspace = true }
-tonic            = { features = ["codegen", "prost", "tls-native-roots", "transport"], workspace = true }
+tonic            = { features = ["codegen", "tls-native-roots", "transport"], workspace = true }
```
bobbinth, line 32: Why is `prost` no longer needed?

bobbinth (approved):
Looks good! Thank you! I left just a couple of questions inline.

Before we merge this PR, let's do the following:

- Run client integration tests against this branch to make sure everything still works as expected.
- @TomasArrachea, could you update the PR description with major changes that happened here (e.g., seems like we are now using `tonic-prost-build` instead of a handful of crates). Also, would be good to know if we could update all dependencies to their latest versions or if some dependencies we could not update.

crates/block-producer/src/mempool/tests.rs
TomasArrachea: Span names didn't change, but some attributes did ("The attributes `code.filepath`, `code.lineno`, and `code.namespace` have been renamed to `code.file.path`, and `code.line.number`, and `code.module.name`, to align with the opentelemetry" according to the [release changelog](https://github.com/tokio-rs/tracing-opentelemetry/releases/tag/v0.32.0)). I don't think it's a big impact for us.

bin/network-monitor/Cargo.toml
TomasArrachea: The feature does no longer exist for `tonic`.

crates/block-producer/src/mempool/tests.rs
drahnr: We don't rely on the namespace for alters as far as I am aware, so this is fine.

PR #1219, opened 2025-09-10
feat: NTX Builder Account Actors

sergerad:
wip

PR #1324, opened 2025-11-03
feat: NTX Actor support for notes that predate account creation

### LINKED ISSUE #1317: Ensure Ntx Actors contain mempool events
There seems to be an edge case now where an ntx account actor could receive transactions pertaining to notes that it is not aware of.

This occurs under the following sequence:
1. A transaction is received in the mempool involving a network account that does not yet exist
2. A transaction is received creating the network account (before 1. tx is committed to a block)
3. A transaction is received consuming the aforementioned note

The issue is that the ntx accounts only load notes under the following circumstances
1. on startup by going to the store; and
2. when a new transaction hits the mempool

We need a way to ensure that newly created ntx actors are aware of notes pertaining to transactions in the mempool (inflight).

This issue is a bit WIP and I will come back to fleshing it out properly soon.

Originally formulated by this issue before the actor refactor: https://github.com/0xMiden/miden-node/issues/1055.

sergerad:
## Context

There is an edge case whereby transactions that relate to not yet existent network accounts are never processed by the NTB. See #1317.

Closes #1317.

No changelog as this is being merged into tracking branch.

## Changes

- Move mempool events to `Arc` to reduce cost of cloning.
- Update NTB Coordinator to determine which actors should receive which events rather than broadcasting to all.
- Update NTB Coordinator to keep track of mempool events with transactions that predate corresponding network accounts.

### DIFF
--- a/crates/ntx-builder/src/coordinator.rs
+++ b/crates/ntx-builder/src/coordinator.rs
@@ -254,6 +254,8 @@ impl Coordinator {
     /// Removes any cached events for a given transaction ID from all account prefix caches.
     pub fn drain_predating_events(&mut self, tx_id: &TransactionId) {
         // Remove the transaction from all prefix caches.
+        // This iterates over all predating events which is fine because the count is expected to be
+        // low.
         self.predating_events.retain(|_, prefix_event| {
             prefix_event.shift_remove(tx_id);
             // Remove entries for account prefixes with no more cached events.


crates/ntx-builder/src/builder.rs
```diff
@@ -177,26 +179,46 @@ impl NetworkTransactionBuilder {
...
+                                .spawn_actor(network_account, account_actor_config)
+                                .await?;
+                        }
+                        Ok(())
+                    },
+                    Some(AccountUpdateDetails::Delta(account_delta)) => {
+                        // Cache notes that predate corresponding accounts if there are any.
+                        if let Ok(prefix) = NetworkAccountPrefix::try_from(account_delta.id()) {
+                            self.coordinator.cache_predating_events(prefix, &event, id);
+                        }
```
sergerad: @bobbinth this is probably not sufficient. I suppose I need to search the prefixes involved within every note of the transaction? Not just the one pertaining to the account delta?

bobbinth: Yes, in fact, I don't think the account from account delta is relevant in this context. Basically, for every transaction we need to get a list of output notes (we only care about network notes addressed to a single account), and then route them to their respective accounts. The notes that remain "unassigned" after this process should be treated separately (i.e., put into some pool from which they can be assigned to an account if/when it gets created).

sergerad: The model we have right now is that we broadcast every event to the actors. Initially we did instead filter out irrelevant actors but we removed that logic for some reason (lost in sea of resolved comments in first PR).

This [commit](https://github.com/0xMiden/miden-node/pull/1219/commits/5ec11456e4e036bdf2f3f1b308c78133dad64472) shows the logic we removed. Essentially:
```rust
    fn find_affected_accounts(
        account_delta: Option<&AccountUpdateDetails>,
        network_notes: &[NetworkNote],
    ) -> HashSet<NetworkAccountPrefix> {
```

We might want to go back to this both for this PR and in the future if we decide to only spawn ephemeral tasks rather than keeping a task around for every actor.

crates/ntx-builder/src/coordinator.rs
```diff
@@ -195,11 +214,41 @@ impl Coordinator {
...
+                if let NetworkNote::SingleTarget(note) = note {
+                    // Only cache if no actor exists for this prefix.
+                    let prefix = note.account_prefix();
+                    if !self.actor_registry.contains_key(&prefix) {
+                        self.predating_events.entry(prefix).or_default().insert(*id, event.clone());
+                    }
+                }
+            }
+        }
+    }
```
sergerad, line 252: @bobbinth current impl ^

The events/notes aren't "assigned" in the first place. But we could modify the coordinator to work that way, as mentioned in the other comment.

crates/ntx-builder/src/builder.rs
```diff
@@ -174,29 +176,39 @@ impl NetworkTransactionBuilder {
...
             },
             // Update chain state and broadcast.
-            MempoolEvent::BlockCommitted { header, .. } => {
+            MempoolEvent::BlockCommitted { header, txs } => {
                 self.update_chain_tip(header.clone(), chain_state).await;
-                self.coordinator.broadcast_event(&event).await;
+                self.coordinator.broadcast_event(event.clone()).await;
+                for tx_id in txs {
+                    self.coordinator.drain_predating_events(tx_id);
+                }
```
sergerad, line 210: I am unsure if there is a risk of draining an event that has not been processed when it should have been.

If we can guarantee that all the tx events will come through before the corresponding block committed event does, then this is fine.

crates/ntx-builder/src/coordinator.rs
sergerad: Added targeted_send()

crates/ntx-builder/src/coordinator.rs
```diff
@@ -82,6 +84,10 @@ pub struct Coordinator {
     /// Each actor must acquire a permit from this semaphore before processing a transaction,
     /// ensuring fair resource allocation and system stability under load.
     semaphore: Arc<Semaphore>,
+
+    /// Cache of events received from the mempool that predate corresponding network accounts.
+    /// Grouped by account prefix to allow targeted event delivery to actors upon creation.
+    predating_events: HashMap<NetworkAccountPrefix, HashMap<TransactionId, Arc<MempoolEvent>>>,
```
Mirko-von-Leipzig: This won't quite work because events need to be sent in chronological order. In the mempool's subscription sender this is solved by using a pair of `BTreeMap`s but it only has a single output target so it can be simpler.

https://github.com/0xMiden/miden-node/blob/2cd7b4201486ed4597c7ffbc0fbb25a8500cb959/crates/block-producer/src/mempool/subscription.rs#L138-L163

Two options imo:

1. Copy the mempool impl exactly and iterate over all txs to find ones that apply to an account
2. Keep the current setup, but each internal map must retain the order

In both cases it may make sense to start using a crate for this use case e.g. [indexmap](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html), though I would wrap it to simplify the api (to avoid the `remove` order changing footguns).

crates/ntx-builder/src/builder.rs
```diff
@@ -174,35 +176,42 @@ impl NetworkTransactionBuilder {
...
+                Ok(())
             },
             // Broadcast to all actors.
             MempoolEvent::TransactionsReverted(_) => {
-                self.coordinator.broadcast_event(&event).await;
+                // TODO (current pr): Do we need to cache these?
+                self.coordinator.broadcast(event).await;
```
Mirko-von-Leipzig: You also need to call `drain_predating_events` to ensure these aren't sent to new actors

crates/ntx-builder/src/coordinator.rs
sergerad: Went with 2 and used indexmap

crates/ntx-builder/src/coordinator.rs
```diff
@@ -195,11 +215,57 @@ impl Coordinator {
...
+    }
+
     /// Helper function to send an event to a single account actor.
     async fn send(
         handle: &ActorHandle,
-        event: &MempoolEvent,
+        event: Arc<MempoolEvent>,
     ) -> Result<(), SendError<MempoolEvent>> {
-        handle.event_tx.send(event.clone()).await
+        handle.event_tx.send((*event).clone()).await
```
Mirko-von-Leipzig: Nit: Do we need the clone still?

```suggestion
        handle.event_tx.send(event).await
```

crates/ntx-builder/src/builder.rs
Mirko-von-Leipzig: Mempool events are always ordered chronologically so this is fine 👍

crates/ntx-builder/src/coordinator.rs
```diff
@@ -195,11 +215,57 @@ impl Coordinator {
...
+    /// Sends a mempool event to all network account actors that are found in the corresponding
+    /// transaction's notes.
+    ///
+    /// Caches the mempool event for each network account found in the transaction's notes that does
+    /// not currently have a corresponding actor. If an actor does not exist for the account, it is
+    /// assumed that the account has not been created on the chain yet.
+    ///
+    /// Cached events will be fed to the corresponding actor when the account creation transaction
+    /// is processed.
+    pub async fn send_targeted(
```
drahnr, line 227: nit: suffix `_targeted` doesn't really click from the first read of the doc

crates/ntx-builder/src/coordinator.rs
```diff
@@ -195,11 +215,57 @@ impl Coordinator {
...
+    }
+
+    /// Removes any cached events for a given transaction ID from all account prefix caches.
+    pub fn drain_predating_events(&mut self, tx_id: &TransactionId) {
+        // Remove the transaction from all prefix caches.
+        self.predating_events.retain(|_, prefix_event| {
+            prefix_event.shift_remove(tx_id);
+            // Remove entries for account prefixes with no more cached events.
+            !prefix_event.is_empty()
+        });
```
bobbinth, line 263: This would iterate over all accounts in the `predating_events` map, right? If so, I'd state in the comment that that's OK because we don't expect that `predating_events` map to ever get very big.

crates/ntx-builder/src/coordinator.rs
```diff
@@ -195,11 +215,57 @@ impl Coordinator {
...
+            Self::send(actor, event.clone()).await?;
+        }
+        Ok(())
+    }
+
+    /// Removes any cached events for a given transaction ID from all account prefix caches.
+    pub fn drain_predating_events(&mut self, tx_id: &TransactionId) {
+        // Remove the transaction from all prefix caches.
+        self.predating_events.retain(|_, prefix_event| {
+            prefix_event.shift_remove(tx_id);
```
drahnr, line 260: JFYI: `fn shift_remove` shuffles the map indices, but to my understanding, these are internal, so it's fine, we never track those externally

crates/ntx-builder/src/builder.rs
```diff
@@ -174,35 +176,49 @@ impl NetworkTransactionBuilder {
...
-                            self.coordinator.broadcast_event(&event).await;
+                            self.coordinator
+                                .spawn_actor(network_account, account_actor_config)
+                                .await?;
                         }
                     }
-                } else {
-                    self.coordinator.broadcast_event(&event).await;
                 }
+                self.coordinator.send_targeted(&event).await?;
```
bobbinth, line 198: Was it a bug previously that we didn't broadcast the event for newly created actors?

crates/ntx-builder/src/coordinator.rs
```diff
@@ -195,11 +215,57 @@ impl Coordinator {
...
+    /// is processed.
+    pub async fn send_targeted(
+        &mut self,
+        event: &Arc<MempoolEvent>,
+    ) -> Result<(), SendError<Arc<MempoolEvent>>> {
+        let mut target_actors = HashMap::new();
+        if let MempoolEvent::TransactionAdded { id, network_notes, .. } = event.as_ref() {
+            // Determine target actors for each note.
+            for note in network_notes {
+                if let NetworkNote::SingleTarget(note) = note {
```
drahnr, line 235: I am missing the `NetworkNote::MultiTarget`? Is that intentional?

Mirko-von-Leipzig: Yes, the network tx builder doesn't manage/handle multi target network notes. Those will one day be handled by something else since they can be consumed by.. well many things.

PR #1326, opened 2025-11-03
feat: rocksdb storage for `AccountTree` and `NullifierTree` by default

drahnr:
Currently we only use `LargeSmt<Memory>` with the `Memory` backend. This is not very useful in practice, the entire purpose of `LargeSmt` is having a _disk_ backed `Smt` implementation.

This PR will ensure we have the `rocksdb` feature enabled in `miden-crypto` until it's extracted per #1218 .

### How it's done

We use the dependency feature unification of cargo to ensure we depend on `miden-crypto` with the `rocksdb` feature enabled. However, this incurs both a hefty penalty on compile times by default, compiling `rocksdb-sys` and additional CI dependencies, which leads to _disabling it by default_.

## CI

Since `rocksdb` comes with some overhead / CI time / cache we only want to run it whenever we change OR link against system level libraries statically.
_NOT_ part of this PR since I am unsure of the best path.

## Extras

Some additional type aliases are used to avoid lengthy types.


crates/store/Cargo.toml
```diff
@@ -53,6 +54,14 @@ rand                  = { workspace = true }
 regex                 = { version = "1.11" }
 termtree              = { version = "0.5" }
 
+[features]
+default = []                       # rocksdb has just too many requirements on CI for now
```
sergerad: nit: comment formatting all good? might be formatter

Makefile
```diff
@@ -8,20 +8,21 @@ help:
 
 WARNINGS=RUSTDOCFLAGS="-D warnings"
 BUILD_PROTO=BUILD_PROTO=1
+ALL_FEATURES="--features=tracing-forest,concurrent,tracing-forest,tx-prover,batch-prover,block-prover,std"
```
Mirko-von-Leipzig: Could you add a comment on why this is necessary?

CHANGELOG.md
```diff
@@ -106,6 +106,7 @@
 - Add optional `TransactionInputs` field to `SubmitProvenTransaction` endpoint for transaction re-execution (#[1278](https://github.com/0xMiden/miden-node/pull/1278)).
 - Added `validator` crate with initial protobuf, gRPC server, and sub-command (#[1293](https://github.com/0xMiden/miden-node/pull/1293)).
 - [BREAKING] Added `AccountTreeWithHistory` and integrate historical queries into `GetAccountProof` ([#1292](https://github.com/0xMiden/miden-node/pull/1292)).
+- [BREAKING] Added `rocksdb` feature to enable rocksdb as `LargeSmt` as `SmtStore` ([#1326](https://github.com/0xMiden/miden-node/pull/1326)).
```
Mirko-von-Leipzig: Is this breaking if its an optional feature?

crates/store/Cargo.toml
```diff
@@ -24,6 +24,7 @@ diesel_migrations      = { features = ["sqlite"], version = "2.3" }
 fs-err                 = { workspace = true }
 hex                    = { version = "0.4" }
 indexmap               = { workspace = true }
+miden-crypto           = { features = ["concurrent", "hashmaps"], workspace = true }
```
Mirko-von-Leipzig, line 27: I'm not entirely following. Shouldn't this be optional and only enabled with `rocksdb`?

The feature unification machete exclusion at the bottom hints that there is something going on, but I don't quite grok it.

.github/workflows/lint.yml
Mirko-von-Leipzig, line 1: I'm not loving the CI complexity. 

Is the goal not to _always_ have this enabled? When would we run a node without this once its working? I would imagine that CI caching should keep the compile times in check or no?

crates/store/src/state.rs
```diff
@@ -81,20 +85,44 @@ pub struct TransactionInputs {
...
+#[cfg(feature = "rocksdb")]
+pub type AccountTreeStorage = RocksDbStorage;
+#[cfg(not(feature = "rocksdb"))]
+pub type AccountTreeStorage = MemoryStorage;
+
+/// Helper for constructing storage backends.
+struct StorageBuilder<S>(std::marker::PhantomData<S>);
+
+impl StorageBuilder<MemoryStorage> {
+    #[allow(dead_code, clippy::unnecessary_wraps)]
```
Mirko-von-Leipzig:
```suggestion
    #[allow(dead_code, clippy::unnecessary_wraps, reason = "Result is required by its `RocksDb` counterpart")]
```

crates/store/src/state.rs
```diff
@@ -81,20 +85,44 @@ pub struct TransactionInputs {
     pub new_account_id_prefix_is_unique: Option<bool>,
 }
 
+/// The storage backend for account trees.
+#[cfg(feature = "rocksdb")]
+pub type AccountTreeStorage = RocksDbStorage;
+#[cfg(not(feature = "rocksdb"))]
+pub type AccountTreeStorage = MemoryStorage;
```
Mirko-von-Leipzig: Is there a roadmap or point-in-time when we would always have this enabled?

crates/store/benches/account_tree.rs
Mirko-von-Leipzig, line 1: I found this file name a bit confusing. `_historical` made me think this is some legacy implementation we're testing. But I see now it means an account tree with history in it? Isn't that sort of always the case? 😅 Maybe just `account_tree.rs` is fine?

CHANGELOG.md
drahnr: It's enable it by default since yesterday

crates/store/Cargo.toml
drahnr: We don't use it in `*.rs` files so machete complains. But we use the feature unification process, this one is for `hashmaps` for lookup perf such that the `type Map` is actually a `HashMap` and not a `BTreeMap`. At the time I opened this PR `miden-base` did not expose these feature flags. Let me double check why this shows up as change in this PR, it should only enable `rocksdb` gated behind the `rocksdb` feature (which should become obsolete with #1499 is done, it's still going to be a bit until all pieces are in place)

crates/store/benches/account_tree.rs
drahnr: Correct, legacy naming :melting_face:  I'll rename

.github/workflows/lint.yml
drahnr: I really hate the CI crap that's part of the PR.
The CI caching should keep the CI times in check, but right now `rocksdb` also compiles it's C/CXX dependency, which takes a while.

The rationale: It takes a while to compile rocksdb. If it's cached, it re-claims a good chunk of cache size. I don't necessarily want to cache this _at all_. I'd preferably link statically against host libraries (at least in CI), but this has proven itself to be fragile (llvm compiler version fun).

I am very open to suggestions how to handle this.

Also: Not a fan of GHA, it's a constrained dumpsterfire.

Makefile
drahnr: This plays into CI and _not_ compiling `rocksdb` by default on CI (see absence of it in the list), but only if relevant files changed, i.e. anything in accounts. That was the rough idea to limit the CI roundtrip time to those PRs where it matters. I am aware this is far from optimal and I'd prefer to do it differently (always compile).

.github/workflows/lint.yml
drahnr: I simplified it to use host libraries. As said earlier, no good solutions, only trade-offs.

drahnr: Host libraries are too old on debian 🫠

.github/actions/install-rocksdb/action.yml
```diff
@@ -0,0 +1,14 @@
+name: "Install RocksDB"
```
Mirko-von-Leipzig: And maybe the name of the file as well

```suggestion
name: "Install RocksDB dependencies"
```

.github/workflows/test.yml
```diff
@@ -28,6 +28,10 @@ jobs:
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@main
+      - name: Cleanup large tools for build space
+        uses: ./.github/actions/cleanup-runner
```
Mirko-von-Leipzig, line 32: How do you determine which of these get this action added?

crates/store/src/state.rs
```diff
@@ -8,39 +8,40 @@ use std::ops::RangeInclusive;
...
-    SmtProof,
-    SmtStorage,
-};
+#[cfg(not(feature = "rocksdb"))]
+use miden_protocol::crypto::merkle::smt::MemoryStorage;
+#[cfg(feature = "rocksdb")]
+use tracing::warn;
+#[cfg(feature = "rocksdb")]
+use miden_protocol::crypto::merkle::smt::RocksDbConfig;
+use miden_protocol::crypto::merkle::smt::{LargeSmt, LargeSmtError, SmtProof, SmtStorage};
```
Mirko-von-Leipzig, line 36: Does the formatter let you move all the conditionals to a single separate paragraph below the rest of the imports?


```suggestion
#[cfg(not(feature = "rocksdb"))]
use miden_protocol::crypto::merkle::smt::MemoryStorage;
#[cfg(feature = "rocksdb")]
use {
    miden_crypto::merkle::smt::RocksDbStorage,
    miden_protocol::crypto::merkle::smt::RocksDbConfig,
    tracing::warn,
};
```

crates/store/Cargo.toml
Mirko-von-Leipzig: Oh no. So this is basically another feature flag isn't additive but instead is used to modify behavior in a breaking fashion?

crates/store/Cargo.toml
```diff
@@ -55,6 +56,14 @@ rand                  = { workspace = true }
...
+rocksdb = ["miden-crypto/rocksdb"]
+
 [[bench]]
 harness = false
-name    = "account_tree_historical"
+name    = "account_tree"
+required-features = ["rocksdb"]
+
+[package.metadata.cargo-machete]
+ignored = ["miden-crypto"]
```
Mirko-von-Leipzig, line 72:
```suggestion
[package.metadata.cargo-machete]
# This is an indirect dependency for which we need to enable optimisations
# via feature flags. Because we don't use it directly in code, machete 
# identifies it as unused.
ignored = ["miden-crypto"]
```

crates/store/src/state.rs
```diff
@@ -151,8 +277,16 @@ impl State {
...
-        let nullifier_tree = load_nullifier_tree(&mut db).await?;
+
+        let account_storage = TreeStorage::create(data_path, "accounttree")?;
+        let smt = account_storage.load_account_tree(&mut db).await?;
+        let account_tree =
+            AccountTree::new(smt).map_err(StateInitializationError::FailedToCreateAccountsTree)?;
+        let account_tree = AccountTreeWithHistory::new(account_tree, latest_block_num);
+
+        let nullifier_storage = TreeStorage::create(data_path, "nullifiertree")?;
+        let nullifier_tree = nullifier_storage.load_nullifier_tree(&mut db).await?;
```
Mirko-von-Leipzig, line 289: Does it ever make sense to have a single database for both/all trees, just separate buckets?

crates/store/src/state.rs
```diff
@@ -82,8 +84,135 @@ pub struct TransactionInputs {
...
+    }
+
+    async fn load_nullifier_tree(
+        self,
+        db: &mut Db,
+    ) -> Result<NullifierTree<LargeSmt<Self>>, StateInitializationError> {
+        // If RocksDB storage is empty, populate it from SQLite
+        let is_empty = self
+            .has_leaves()
+            .map_err(|e| StateInitializationError::AccountTreeIoError(e.to_string()))?;
```
Mirko-von-Leipzig: The error variant seems incorrect

crates/store/src/state.rs
Mirko-von-Leipzig, line 289: It feels like there is a more fluid API hiding in here somewhere, feel free to disregard as this isn't super important at this point.

Sort of like the database IO shouldn't be inside the tree impl, but rather something that gets passed in. And maybe separate `load() -> Option<_>` and `create()`.

crates/store/src/state.rs
```diff
@@ -82,8 +84,135 @@ pub struct TransactionInputs {
...
+        if !is_empty {
+            let smt = load_smt(self)?;
+            return Ok(NullifierTree::new_unchecked(smt));
+        }
+
+        warn!(target: COMPONENT, "RocksDB nullifier tree storage is empty, populating from SQLite");
```
Mirko-von-Leipzig: Probably something that is expected; so maybe `info!`

Mirko-von-Leipzig (approved):

crates/store/src/state.rs
drahnr: When writing it, I has the same thought, however the current API doesn't allow for it. In principle a single DB should utilize IO much better

.github/workflows/test.yml
drahnr: With ubuntu-latest there is limited space left, so wherever we compile `librocksdb-sys` we might run into space issues (hypothesis). This shouldn't be the case with custom runners so much (we are in control of the size avail). Rule of thumb: Add it where you see `ubuntu-latest`

crates/store/src/state.rs
drahnr: The abstraction does feel backwards, but it's the most versatile I could come up with, without getting too much in the way.

* `fn x<T>(me: &mut T)` doesn't allow specialization.
* `struct Loader` + `trait StorageLoader` just adds another type for no good reason
* `cfg(feature = "rocksdb")` gated functions are smellier to me than the current appraoch

> Sort of like the database IO shouldn't be inside the tree impl, but rather something that gets passed in. And maybe separate load() -> Option<_> and create().

I might need another hing on what you'd separate? Technically we already have `fn load_*` and `fn create` just with the dependency on the storage type via the trait impl

crates/store/src/state.rs
drahnr: I did what I could, but `cargo fmt` is pretty aggressive.

crates/store/src/state.rs
```diff
@@ -133,7 +260,7 @@ impl State {
     // CONSTRUCTOR
     // --------------------------------------------------------------------------------------------
 
-    /// Loads the state from the `db`.
+    /// Loads the state from the data directory.
     #[instrument(target = COMPONENT, skip_all)]
     pub async fn load(data_path: &Path) -> Result<Self, StateInitializationError> {
```
bobbinth, line 265: Could we move all loading-related methods into a separate file? Basically, I'd split this file into:

- `src/state/mod.rs`
- `src/state/loader.rs`

I think overall, we need 3 loading modes:

1. Load everything from memory (when `rocksdb` feature is disabled).
2. Use persistent data from RockDB (when `rocksdb` feature is enabled). This would probably require passing filepaths to RocksDB databases on startup (or at least assuming some defaults).
3. Rebuild RocksDB data from the database (when `rocksdb` feature is enabled and we are starting a new node, or the user sets some flag that indicates that RocksDB databases are to be re-built).

I think currently we have 1 and 2, and maybe some parts of 3. 

@drahnr - could you tackle this?

drahnr: Done in https://github.com/0xMiden/miden-node/pull/1532

PR #1332, opened 2025-11-04
feat: Implement DataStore:: get_note_script() for NtxDataStore

### LINKED ISSUE #1290: Implement get_note_script() for NtxDataStore

sergerad:
## Context

Closes #1290.

The `NtxDataStore` does not currently have an implementation for `DataStore::get_note_script()`.

It can be implemented by using the `GetNoteScriptByRoot` RPC endpoint.

## Changes
- Implement `DataStore::get_note_script()` for `NtxDataStore` using the RPC.
- Add LRU cache to `NtxDataStore`.

### DIFF
--- a/crates/ntx-builder/src/transaction.rs
+++ b/crates/ntx-builder/src/transaction.rs
@@ -1,5 +1,5 @@
 use std::collections::BTreeSet;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
 use lru::LruCache;
 use miden_node_utils::tracing::OpenTelemetrySpanExt;
@@ -35,6 +35,7 @@ use miden_tx::{
     TransactionMastStore,
     TransactionProverError,
 };
+use tokio::sync::Mutex;
 use tokio::task::JoinError;
 use tracing::{Instrument, instrument};
 
@@ -394,8 +395,7 @@ impl DataStore for NtxDataStore {
         async move {
             // Attempt to retrieve the script from the cache.
             if let Some(cached_script) = {
-                let mut cache_guard =
-                    cache.lock().expect("cache mutex cannot already be held in the current thread");
+                let mut cache_guard = cache.lock().await;
                 cache_guard.get(&script_root).cloned()
             } {
                 return Ok(cached_script);
@@ -413,9 +413,7 @@ impl DataStore for NtxDataStore {
                 Some(script) => {
                     // Cache the retrieved script.
                     {
-                        let mut cache_guard = cache
-                            .lock()
-                            .expect("cache mutex cannot already be held in the current thread");
+                        let mut cache_guard = cache.lock().await;
                         cache_guard.put(script_root, script.clone());
                     }
                     // Return script.


crates/ntx-builder/src/builder/mod.rs
```diff
@@ -30,6 +31,8 @@ pub struct NetworkTransactionBuilder {
     pub store_url: Url,
     /// Address of the block producer gRPC server.
     pub block_producer_url: Url,
+    /// Address of the RPC gRPC server for retrieving note scripts.
+    pub rpc_url: Url,
```
bobbinth: Why do we need to go through the RPC for this? Could we not just get note scripts directly from the store?

crates/ntx-builder/src/builder/mod.rs
```diff
@@ -43,10 +46,39 @@ pub struct NetworkTransactionBuilder {
...
 
+        let rpc_client = ClientBuilder::new(self.rpc_url)
+            .with_tls()
+            .context("Failed to configure TLS")?
+            .with_timeout(Duration::from_secs(10))
+            .without_metadata_version()
+            .without_metadata_genesis()
+            .connect::<Rpc>()
+            .await
+            .context("Failed to connect to RPC server")?;
```
bobbinth: As I mentioned above, I don't think this is necessary - we should be able to get note scripts directly from the store (may need to update protobuf files for this).

crates/ntx-builder/src/transaction.rs
```diff
@@ -73,6 +78,9 @@ pub struct NtxContext {
     /// Defaults to local proving if unset. This should be avoided in production as this is
     /// computationally intensive.
     pub prover: Option<RemoteTransactionProver>,
+
+    /// The RPC client for retrieving note scripts.
+    pub rpc_client: RpcClient,
```
bobbinth: Related to the previous comments: this should probably be `store_client`.

crates/ntx-builder/src/transaction.rs
```diff
@@ -233,18 +246,38 @@ impl NtxContext {
...
 /// This is sufficient for executing a network transaction.
 struct NtxDataStore {
     account: Account,
     reference_header: BlockHeader,
     chain_mmr: PartialBlockchain,
     mast_store: TransactionMastStore,
+    /// RPC client for retrieving note scripts from the RPC server.
+    rpc_client: RpcClient,
+    /// LRU cache for storing retrieved note scripts to avoid repeated RPC calls.
+    script_cache: Arc<Mutex<LruCache<Word, miden_objects::note::NoteScript>>>,
```
bobbinth: nit: I don't think we need the full path for `miden_objects::note::NoteScript` here.

crates/ntx-builder/src/transaction.rs
```diff
@@ -233,18 +246,38 @@ impl NtxContext {
...
 }
 
 impl NtxDataStore {
-    fn new(account: Account, reference_header: BlockHeader, chain_mmr: PartialBlockchain) -> Self {
+    /// Default cache size for note scripts.
+    ///
+    /// Each cached script contains the deserialized `NoteScript` object, so the actual memory usage
+    /// depends on the complexity of the scripts being cached.
+    const DEFAULT_SCRIPT_CACHE_SIZE: usize = 1000;
```
bobbinth, line 267: This is fine for now, but we should make this a config parameter in the future. Let's create an issue for this.

crates/ntx-builder/Cargo.toml
```diff
@@ -16,6 +16,7 @@ workspace = true
 [dependencies]
 anyhow                     = { workspace = true }
 futures                    = { workspace = true }
+lru                        = { default-features = false, version = "0.16" }
```
bobbinth: nit: maybe we should put this at the workspace level? (I anticipate we'll need LRU cache in other places too).

crates/ntx-builder/src/transaction.rs
```diff
@@ -348,12 +386,70 @@ impl DataStore for NtxDataStore {
...
+            }
+
+            // Retrieve the script from the RPC.
+            let note_root = NoteRoot { root: Some(script_root.into()) };
+            match rpc_client.get_note_script_by_root(note_root).await {
+                Ok(response) => {
+                    // Decode the script from the response.
+                    let maybe_script = response.into_inner();
+                    if let Some(script_proto) = maybe_script.script {
+                        match miden_objects::note::NoteScript::read_from_bytes(&script_proto.mast) {
```
bobbinth: nit: I don't think we need to use the full path for `miden_objects::note::NoteScript` here.

crates/ntx-builder/src/transaction.rs
```diff
@@ -348,12 +386,70 @@ impl DataStore for NtxDataStore {
...
+                                Ok(script)
+                            },
+                            // Failed to decode the script.
+                            Err(err) => {
+                                tracing::error!(
+                                    target: COMPONENT,
+                                    ?err,
+                                    "failed to deserialize note script from bytes"
+                                );
+                                Err(DataStoreError::NoteScriptNotFound(script_root))
```
bobbinth: I'm not sure this should return `NoteScriptNotFound` - technically, it was found but was just malformed. Maybe some other error is more appropriate.

crates/ntx-builder/src/transaction.rs
```diff
@@ -348,12 +386,70 @@ impl DataStore for NtxDataStore {
...
+                let mut cache_guard =
+                    cache.lock().expect("cache mutex cannot already be held in the current thread");
+                cache_guard.get(&script_root).cloned()
+            } {
+                return Ok(cached_script);
+            }
+
+            // Retrieve the script from the RPC.
+            let note_root = NoteRoot { root: Some(script_root.into()) };
+            match rpc_client.get_note_script_by_root(note_root).await {
```
bobbinth: nit: I would probably reduce the level of nesting here a bit. For example:
```rust
let note_root = NoteRoot { root: Some(script_root.into()) };
let note_script = match rpc_client.get_note_script_by_root(note_root).await {
    ...
}?

// Cache the retrieved script.
{
    let mut cache_guard = cache.lock().unwrap();
    cache_guard.put(script_root, script.clone());
}

Ok(note_script)
```

crates/ntx-builder/src/transaction.rs
```diff
@@ -348,12 +386,70 @@ impl DataStore for NtxDataStore {
...
+                Ok(response) => {
+                    // Decode the script from the response.
+                    let maybe_script = response.into_inner();
+                    if let Some(script_proto) = maybe_script.script {
+                        match miden_objects::note::NoteScript::read_from_bytes(&script_proto.mast) {
+                            Ok(script) => {
+                                // Cache the retrieved script.
+                                {
+                                    let mut cache_guard = cache.lock().unwrap();
```
bobbinth: If `lock()` fails - should we return an error instead? Or does this mean we can't recover and should crush? (in that case, we should probably log this error better).

bobbinth (commented):
Looks good! Thank you. I left some comments inline - but the general approach is along the lines of what I was thinking.

crates/ntx-builder/src/transaction.rs
sergerad: Forgot to add expect() here, done now - same as the other location

crates/ntx-builder/src/store.rs
```diff
@@ -161,6 +162,15 @@ impl StoreClient {
 
         Ok(account)
     }
+
+    pub async fn get_note_script_by_root(&self, root: Word) -> Result<Option<Vec<u8>>, StoreError> {
```
bobbinth: Should we change this to return `Result<Option<Vec<NoteScript>>, StoreError>`?

For the RPC it probably makes sense to return bytes because we just forward these bytes to the user, but here i think we always convert to `NoteScript`.

crates/ntx-builder/src/transaction.rs
```diff
@@ -348,12 +380,53 @@ impl DataStore for NtxDataStore {
...
-        async move { Err(DataStoreError::NoteScriptNotFound(script_root)) }
+    ) -> impl FutureMaybeSend<Result<NoteScript, DataStoreError>> {
+        let store = self.store.clone();
+        let cache = self.script_cache.clone();
+
+        async move {
+            // Attempt to retrieve the script from the cache.
+            if let Some(cached_script) = {
+                let mut cache_guard =
+                    cache.lock().expect("cache mutex cannot already be held in the current thread");
```
drahnr: Isn't the `lock() -> Result<_>` for `Err` always indicating lock poisoning?

crates/ntx-builder/src/transaction.rs
```diff
@@ -233,18 +240,38 @@ impl NtxContext {
...
 /// This is sufficient for executing a network transaction.
 struct NtxDataStore {
     account: Account,
     reference_header: BlockHeader,
     chain_mmr: PartialBlockchain,
     mast_store: TransactionMastStore,
+    /// Store client for retrieving note scripts.
+    store: StoreClient,
+    /// LRU cache for storing retrieved note scripts to avoid repeated store calls.
+    script_cache: Arc<Mutex<LruCache<Word, NoteScript>>>,
```
drahnr, line 259: don't we want an `async` aware `Mutex`? I.e. the one from `tokio`?

drahnr: practically it doesn't make a difference, but it's a loaded footgun since we use it inside `async move { .. }` blocks

drahnr (approved):
LGTM

crates/ntx-builder/src/transaction.rs
sergerad: yea idk if we want to have standard wording for this expect. in node there is also
```rust
        let mut lock = MOCK_ACCOUNTS.lock().expect("Poisoned mutex");
```

crates/ntx-builder/src/transaction.rs
sergerad: yea mb

crates/ntx-builder/src/transaction.rs
sergerad: irrelevant with tokio mutex 👍

PR #1333, opened 2025-11-04
`AccountTreeWithHistory` - full integration

drahnr:
The first preliminary full integration of recent, historical account states. Part of #1227 

## How is it achieved 

For the purpose of accessing historical data, we add the `fn select_historical_account_at` to query the database accordingly.
For sake of query efficiency, we use `(account_id, block_num)` as primary key instead of `account_id`, and add indices as required.

### In scope

* End to end functionality
* Query optimization

### Out of scope

* SQL schema optimization (split tables/columns into multiple tables)

---

### Implementation details

Previously we had primary keys that were single columns, i.e. `account_id`s, now these have become compound keys. With compound keys one can also use foreign key constraints and we do so :)
Now we add some additional indices for better queries of partial keys.

Initially  I thought we could get away without the `is_latest` column, but since we don't store the latest digest for _every_ single block, we need to find the latest for each i.e. leaf. Having the `is_latest` column simplifies this significantly/removes query complexity.   

### Pitfalls

With the change of a new primary key, a few foreign key relationships don't make sense anymore. These were removed as a consequence. The side effects are affecting a lot of code, also those that were not show/touched as part of the PR, since a) a santiy check is gone that we had before b) we now have more entries per `AccountId`.


bobbinth:
What is left on this to take it out of the draft mode?

bin/network-monitor/src/deploy/mod.rs
```diff
@@ -283,8 +283,11 @@ impl DataStore for MonitorDataStore {
         })
     }
 
-    async fn get_note_script(&self, script_root: Word) -> Result<NoteScript, DataStoreError> {
-        Err(DataStoreError::NoteScriptNotFound(script_root))
```
drahnr: The variant got removed. Unclear if this intended

crates/store/src/state.rs
```diff
@@ -150,11 +150,6 @@ impl State {
 
         let chain_mmr = load_mmr(&mut db).await?;
         let block_headers = db.select_all_block_headers().await?;
-        // TODO: Account tree loading synchronization
-        // Currently `load_account_tree` loads all account commitments from the DB. This could
-        // potentially lead to inconsistency if the DB contains account states from blocks beyond
-        // `latest_block_num`, though in practice the DB writes are transactional and this
-        // should not occur.
```
drahnr, line 157: Impossible since it's all happening in one transaction

bin/network-monitor/src/deploy/mod.rs
bobbinth: We now return `None` if the script is not found.

crates/store/src/errors.rs
```diff
@@ -169,6 +171,8 @@ impl From<DatabaseError> for Status {
 
 #[derive(Error, Debug)]
 pub enum StateInitializationError {
+    #[error("account tree IO error: {0}")]
+    FailedToAccountTreeWrite(String),
```
Mirko-von-Leipzig: Nit

```suggestion
    AccountTreeIoError(String),
```

crates/store/src/errors.rs
```diff
@@ -98,6 +98,8 @@ pub enum DatabaseError {
     AccountNotFoundInDb(AccountId),
     #[error("account {0} state at block height {1} not found")]
     AccountAtBlockHeightNotFoundInDb(AccountId, BlockNumber),
+    #[error("historical block {block_num} not available: {reason}")]
+    HistoricalBlockNotAvailable { block_num: BlockNumber, reason: String },
```
Mirko-von-Leipzig, line 102: I'm wondering what the reasons could be beyond "too old, pruned"

crates/store/src/state.rs
```diff
@@ -922,35 +917,46 @@ impl State {
...
+                    reason: "Block is either in the future or has been pruned from history"
+                        .to_string(),
+                })?;
+            (requested_block, witness)
+        } else {
+            // Latest query: use the latest state
+            let block_num = inner_state.account_tree.block_number_latest();
+            let witness = inner_state.account_tree.open_latest(account_id);
+            (block_num, witness)
+        };
```
Mirko-von-Leipzig, line 948: iiuc we should be able to drop the `inner_state` guard at this point now?

If not, we should work towards being able to do so since this is going to become quite a problem imo since it prevents new blocks.

crates/store/src/state.rs
```diff
@@ -1131,16 +1137,32 @@ async fn load_account_tree(
...
         .map(|(id, commitment)| (account_id_to_smt_key(id), commitment));
 
-    let smt = LargeSmt::with_entries(MemoryStorage::default(), smt_entries)
-        .expect("Failed to create LargeSmt from database account data");
+    let smt = LargeSmt::with_entries(MemoryStorage::default(), smt_entries).map_err(|e| {
+        tracing::error!(
+            target: COMPONENT,
+            error = ?e,
+            "Failed to create LargeSmt from database account data"
+        );
```
Mirko-von-Leipzig: Nit: I would put this in a `.inspect_err`

Mirko-von-Leipzig: Though come to think, won't this get reported higher up in the call stack? Maybe a debug remnant from dev?

crates/store/src/state.rs
```diff
@@ -1131,16 +1137,32 @@ async fn load_account_tree(
...
+        match e {
+            miden_objects::crypto::merkle::LargeSmtError::Merkle(merkle_error) => {
+                StateInitializationError::DatabaseError(DatabaseError::MerkleError(merkle_error))
+            },
+            miden_objects::crypto::merkle::LargeSmtError::Storage(err) => {
+                // large_smt::StorageError is not `Sync` and hence `context` cannot be called
+                // which we want to and do
```
Mirko-von-Leipzig, line 1151: That's.. not great - why is it not `Sync`?

Mirko-von-Leipzig: And should this be `err.to_report()` (or whatever that method is called), to retain the context?

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -34,6 +33,12 @@ CREATE TABLE accounts (
 ) WITHOUT ROWID;
 
 CREATE INDEX idx_accounts_network_prefix ON accounts(network_account_id_prefix) WHERE network_account_id_prefix IS NOT NULL;
+CREATE INDEX idx_accounts_id_block ON accounts(account_id, block_num DESC);
```
Mirko-von-Leipzig, line 36: I didn't know you could order the index. Does this determine the search order priority? So it will scan the index descendingly?

crates/store/src/errors.rs
drahnr: I don't think there is any other, except for we lost it unexpectedly (crash + storage corruption).

crates/store/src/errors.rs
drahnr: The naming between `miden-node` and `miden-base` on error variants differs quite significantly. This was migrated over from there a long time ago.

Do we want to unify naming eventually?

drahnr: I'll fix it here :)

crates/store/src/db/migrations/2025062000000_setup/up.sql
drahnr: https://www.postgresql.org/docs/current/sql-createindex.html

it's not really that relevant, it _should_ safe a few operations for comprision/range operations since we order them by block num descendingly. Iiuc the indices are b-trees and sql engines are seeking for prefixes, so this should help a bit. I didn't measure.

crates/store/src/state.rs
drahnr: Yes, we can. Dropping explicitly.

crates/store/src/state.rs
drahnr: Removed, using `ErrorReport::as_report()`

crates/store/src/state.rs
drahnr:
```
enum StorageError {
    #[error("backend error: {0}")]
    Backend(#[from] Box<dyn core::error::Error + Send + 'static>),
..
```

causes this. I don't know if the concrete error types actually don't impl `+ Sync`.

crates/store/src/state.rs
bobbinth, line 962: We should be able to check whether an account is a public account by just looking at the account ID. This way, we can avoid querying the database before we check if the account is public.

Mirko-von-Leipzig: Added to issue #1378

PR #1336, opened 2025-11-06
fix: properly feature-gate test code in miden-node

huitseeker:
This resolves inconsistent feature gating that caused compilation failures in certain environments, which were detected as MSRV failures.

Before this fix (for every crate concerned by this PR):
- `cargo check -p miden-node-rpc --tests`          # FAILED
- `cargo check -p miden-node-rpc --features testing` # FAILED

After this fix:
- `cargo check -p miden-node-rpc --tests`          # PASSES
- `cargo check -p miden-node-rpc --features testing` # PASSES

The root cause was inconsistent feature gating where:
1. Test imports like `test_fee()` were not feature-gated
2. `ExecutionProof::new_dummy()` was used without feature gates
3. Test code compiled in some environments but failed in others

Changes:
- Add `testing` feature that propagates to dependencies, activate it by default when testing,
- Feature-gate all testing-specific imports and code, so that compiling with `--no-default-features` works,
- Replace unavailable `new_dummy()` with `from_bytes()` fallback.
- Remove non-existent feature import in `miden-node-utils`.

This ensures testing code is only included when explicitly requested, making compilation behavior consistent.

### DIFF
--- a/crates/rpc/Cargo.toml
+++ b/crates/rpc/Cargo.toml
@@ -19,10 +19,11 @@ anyhow                 = { workspace = true }
 futures                = { version = "0.3" }
 http                   = { workspace = true }
 mediatype              = { version = "0.21" }
+miden-air              = { workspace = true }
 miden-node-proto       = { workspace = true }
 miden-node-proto-build = { workspace = true }
-miden-node-utils       = { workspace = true }
-miden-objects          = { default-features = true, workspace = true }
+miden-node-utils       = { features = ["testing"], workspace = true }
+miden-objects          = { default-features = true, features = ["testing"], workspace = true }
 miden-tx               = { default-features = true, workspace = true }
 semver                 = { version = "1.0" }
 thiserror              = { workspace = true }
@@ -37,7 +38,6 @@ tracing                = { workspace = true }
 url                    = { workspace = true }
 
 [dev-dependencies]
-miden-air        = { features = ["testing"], workspace = true }
 miden-lib        = { workspace = true }
 miden-node-store = { workspace = true }
 miden-node-utils = { features = ["testing", "tracing-forest"], workspace = true }


crates/store/Cargo.toml
Mirko-von-Leipzig, line 1: iiuc this weird self-reference feature is required to ensure `--no-default-features` works correctly for tests? But we have no features besides the one(s) you've introduced so I don't quite follow why we need this.

huitseeker: > iiuc this weird self-reference feature is required to ensure --no-default-features works correctly for tests? 

No, it's here so that the "testing" feature is active by default *when testing*.

The gating by `#[cfg(feature = "testing")]` is so that the build still passes when using building tests under `--no-default-features`.

Think of it as the setting required to make "testing" a default-feature, but *only* under the testing scope (contrarily to default-features).

Mirko-von-Leipzig: But why do we want or need a `testing` feature? Who is it for?

huitseeker: A testing feature's goal is to allow code only relevant to tests to be accessible from a different crate, but not compiled in the main binary.

A crate can't turn on the test scope of their dependee, so `#[cfg(test)]` is useless for that purpose.

Mirko-von-Leipzig: Personally I believe a `testing` feature is an anti-pattern and should be avoided (as a general statement). Types, invariant-breaking constructors etc can simply be named appropriately, or placed in a `mod testing` if absolutely necessary.

That aside, the node crates shouldn't be depended on externally at all - these are "internal" crates and we don't use `testing` features. The current exception is the `utils` crate but I'd like to remove that feature as well at some point.

huitseeker: When I say "to allow code only relevant to tests to be accessible from a different crate" I didn't mean to specialize this to *external* crates.

Mirko-von-Leipzig: I think maybe I'm being unclear. What I mean is, could the total diff not just be the following (instead of adding features that no-one is using).

```diff
diff --git a/crates/rpc/Cargo.toml b/crates/rpc/Cargo.toml
index 19058084ad..540f85eb2b 100644
--- a/crates/rpc/Cargo.toml
+++ b/crates/rpc/Cargo.toml
@@ -37,9 +37,10 @@
 url                    = { workspace = true }

 [dev-dependencies]
+miden-air        = { features = ["testing"], workspace = true }
 miden-lib        = { workspace = true }
 miden-node-store = { workspace = true }
-miden-node-utils = { features = ["tracing-forest"], workspace = true }
+miden-node-utils = { features = ["testing", "tracing-forest"], workspace = true }
 miden-objects    = { default-features = true, features = ["testing"], workspace = true }
 reqwest          = { version = "0.12" }
 rstest           = { workspace = true }
```

bobbinth: I applied @Mirko-von-Leipzig's suggestion in 257e0d18a8c673274363a3429edba71dcdbd784c and seems like the MSRV checks are passing - so, we can probably go with that as the easiest solution.

The tests are now failing (something related to LLVM) - but this may be a separate issue?

huitseeker: Sounds good! closing this branch then.

PR #1338, opened 2025-11-06
refactor: validate note commitments before adding them 

### LINKED ISSUE #1329: Validate unauthenticated notes via commitment instead of `NoteId`
Currently, when a transaction gets submitted, unauthenticated notes get validated via [their `NoteId`](https://github.dev/0xMiden/miden-node/blob/32b0faad635b8304f9e3fd7cc1a3c575e80e23af/crates/block-producer/src/store/mod.rs#L170-L171) (and [here](https://github.com/0xMiden/miden-node/blob/32b0faad635b8304f9e3fd7cc1a3c575e80e23af/crates/block-producer/src/mempool/state.rs#L65) if not validated by the store). 
Because `NoteId` does not commit to `NoteMetadata`, I believe one could tamper with a note's `NoteMetadata` and submit a proof that consumes the tampered note as unauthenticated which would get admitted into the mempool as valid. 

I wrote a quick test based on the client the other day and got these results:

```
2025-11-01T17:06:51.561072Z  INFO rpc:block_producer.server.submit_proven_transaction:store.client.get_tx_inputs: miden-block-producer: igamigo/miden-node/crates/block-producer/src/store/mod.rs:174: tx_id: 0xd2d0dd17e0ad23f77fc747018e61f3df9a70b67543cb06ea7ec0c0eb0694b705 rpc.service: "block_producer.Api", rpc.method: "SubmitProvenTransaction", otel.name: "block_producer.Api/SubmitProvenTransaction"
2025-11-01T17:06:51.561214Z  INFO store.block_producer_server.get_transaction_inputs:get_transaction_inputs: miden-store: igamigo/miden-node/crates/store/src/state.rs:839: account_id: 0x81b93a380cc85e8025bf82c6575444, nullifiers: [0x38a3bf60ec32026aa94d3c06ff1c06a7d489c31f483e56a919e2a07034f0679f]
2025-11-01T17:06:51.561350Z  INFO store.block_producer_server.get_transaction_inputs:get_transaction_inputs: miden-store: igamigo/miden-node/crates/store/src/state.rs:832: return: Ok(TransactionInputs { account_commitment: Word([0, 0, 0, 0]), nullifiers: [NullifierInfo { nullifier: 0x38a3bf60ec32026aa94d3c06ff1c06a7d489c31f483e56a919e2a07034f0679f, block_num: BlockNumber(0) }], found_unauthenticated_notes: {NoteId(Word([6734379888269138228, 17083291302760502078, 13649816128495193271, 16249048272441998726]))}, new_account_id_prefix_is_unique: Some(true) })
```

Later, block proposal fails:

```
2025-11-01T17:07:01.367799Z ERROR block_builder.build_block:block_builder.propose_block: miden-block-producer: igamigo/miden-node/crates/block-producer/src/block_builder/mod.rs:196: error: failed to propose block
```

Which I guess bubbles up as part of the consistency checks within [`ProposedBlock::new`](https://github.com/0xMiden/miden-node/blob/next/crates/block-producer/src/block_builder/mod.rs#L204-L205). This is currently sort of exploitable because it allows stopping block producing basically. I didn't test it, but I think a single transaction can both create a note and consume it as unauthenticated but with a tampered metadata and it would halt block building AFAIK, so ideally these transactions would not be allowed into the mempool (should be a simple-ish change I think). But I also didn't review the code too much and I'm basing this off of the tests that I wrote, so maybe I'm missing something.

igamigo:
Closes #1329 

This PR mainly changes transaction checks done between the RPC and mempool and refactors the unauthenticated note's existence checks to be done over `Note::commitment` instead of `Note::id`. The rationale is that the `NoteId` does not commit to `NoteMetadata`, whereas `Note::commitment` does. The required changes are pretty straightforward but did involve changing the store to be able to mirror the lookups that were being done done over the note IDs, but with the note commitments. Maybe there were more clever things (like return `NoteHeader` the note commitments without actually changing the store method) that we could do but upon a quick inspection it did not seem like any other way was substantially better than this. However, not super familiar with the code so not sure.

- [X] test repro and run client integration tests:

Repro that used to trigger block proposal failures now fails with:

```
called `Result::unwrap()` on an `Err` value:
RpcError(GrpcError {
    endpoint: SubmitProvenTx,
    error_kind: InvalidArgument,
    source: Some(
      Status {
        code: InvalidArgument,
        message: "unauthenticated transaction note commitments were not found in the store or in outputs of in-flight transactions: [Word([5664795501264005384, 17605801201601601049, 15971648237439111378, 14675290839910473593])]",
        details: b"\x02",
        metadata: MetadataMap {
          headers: {
            "content-type": "application/grpc",
            "date": "Thu, 06 Nov 2025 20:43:27 GMT",
            "access-control-expose-headers": "grpc-status,grpc-message,grpc-status-details-bin",
            "access-control-allow-credentials": "true",
            "vary": "origin, access-control-request-method, access-control-request-headers",
          },
        },
        source: None,
      }
    ),})

```

which looks correct. Every other test passes: `Summary [ 125.982s] 53 tests run: 52 passed, 1 failed, 1 skipped` (failed one is the repro)

WebClient integration tests also work: 

```
 Running 101 tests using 5 workers
 ...
  101 passed (5.9m)
```

- [ ] maybe add a type alias for `NoteCommitment`?

### DIFF
--- a/crates/store/src/db/models/queries/notes.rs
+++ b/crates/store/src/db/models/queries/notes.rs
@@ -314,7 +314,7 @@ pub(crate) fn select_all_notes(
 ///     note_id IN (?1)
 /// ORDER BY
 ///     committed_at ASC
-///```
+/// ```
 pub(crate) fn select_note_inclusion_proofs(
     conn: &mut SqliteConnection,
     note_commitments: &BTreeSet<Word>,


crates/store/src/state.rs
```diff
@@ -719,15 +727,22 @@ impl State {
...
+            .await
+            .map_err(GetBlockInputsError::SelectNoteInclusionProofError)?;
+        let unauthenticated_note_ids =
+            note_records.iter().map(|note| NoteId::from(note.note_id)).collect();
         let unauthenticated_note_proofs = self
             .db
-            .select_note_inclusion_proofs(unauthenticated_notes)
+            .select_note_inclusion_proofs(unauthenticated_note_ids)
             .await
             .map_err(GetBlockInputsError::SelectNoteInclusionProofError)?;
```
bobbinth, line 733: Should we also select note inclusion proofs by commitment rather than ID?

crates/store/src/state.rs
```diff
@@ -889,8 +904,13 @@ impl State {
...
 
-        let found_unauthenticated_notes =
-            self.db.select_note_ids(unauthenticated_notes.clone()).await?;
+        let found_unauthenticated_notes = self
+            .db
+            .select_notes_by_commitment(unauthenticated_note_commitments)
+            .await?
+            .into_iter()
+            .map(|note| note.note_commitment)
+            .collect();
```
bobbinth, line 899: This is probably fine for now, but we'd probably want to implement something like `select_note_commitments()` not to have to load full notes for this purpose.

crates/store/src/state.rs
```diff
@@ -546,6 +547,13 @@ impl State {
         // First we grab note inclusion proofs for the known notes. These proofs only
         // prove that the note was included in a given block. We then also need to prove that
         // each of those blocks is included in the chain.
+        let note_records = self
+            .db
+            .select_notes_by_commitment(unauthenticated_note_commitments.into_iter().collect())
+            .await
+            .map_err(GetBatchInputsError::SelectNoteInclusionProofError)?;
+        let unauthenticated_note_ids =
+            note_records.iter().map(|note| NoteId::from(note.note_id)).collect();
```
bobbinth: Related to my other comment - if we could get inclusion proofs by commitment, we probably wouldn't need to do this.

crates/block-producer/src/mempool/state.rs
```diff
@@ -32,7 +32,7 @@ pub(super) struct InflightState {
     /// All created notes and the ID of the node that created it.
     ///
     /// This _includes_ erased notes, see `nullifiers` for more information.
-    output_notes: HashMap<NoteId, NodeId>,
+    output_notes: HashMap<Word, NodeId>,
```
bobbinth, line 35: Maybe not for this PR, but I wonder if this should be HashMap<Word, Word> (i.e., mapping commitments to commitments). If we are switching to using commitments as primary note identifiers, may make sense to do it throughout.

crates/block-producer/src/mempool/state.rs
```diff
@@ -41,7 +41,7 @@ pub(super) struct InflightState {
     ///
     /// It is the callers responsibility to ensure that all unauthenticated notes exist as output
     /// notes at the time when a new transaction or batch node is inserted.
-    unauthenticated_notes: HashMap<NoteId, NodeId>,
+    unauthenticated_notes: HashMap<Word, NodeId>,
```
bobbinth, line 44: Same comment as above.

crates/block-producer/src/mempool/nodes.rs
```diff
@@ -163,12 +163,12 @@ impl Node for TransactionNode {
         Box::new(self.0.nullifiers())
     }
 
-    fn output_notes(&self) -> Box<dyn Iterator<Item = NoteId> + '_> {
-        Box::new(self.0.output_note_ids())
+    fn output_note_commitments(&self) -> Box<dyn Iterator<Item = Word> + '_> {
```
bobbinth, line 166: nit: it is not 100% clear to me if adding `_commitments` suffix everywhere is needed - but I guess while we use a generic `Word` to represent commitments, it is helpful. This does make me think that maybe we should create a `NoteCommitment` newtype - though, this would be a part of a bigger refactoring.

crates/block-producer/src/mempool/state.rs
```diff
@@ -118,9 +123,9 @@ impl InflightState {
     /// parent and child relationship lookups.
     pub(super) fn insert(&mut self, id: NodeId, node: &dyn Node) {
         self.nullifiers.extend(node.nullifiers());
-        self.output_notes.extend(node.output_notes().map(|note| (note, id)));
+        self.output_notes.extend(node.output_note_commitments().map(|note| (note, id)));
         self.unauthenticated_notes
-            .extend(node.unauthenticated_notes().map(|note| (note, id)));
+            .extend(node.unauthenticated_note_commitments().map(|note| (note, id)));
```
sergerad: nite: we have `|note|` instead of `|note_commitment|` in a bunch of places but maybe we want that for brevity.

crates/block-producer/src/mempool/nodes.rs
igamigo: Yes, I thought of this as well. It was also ambiguous before where sometimes "id" was appended. I wrote in the PR description that I'd create a type alias at least but didn't end up doing it as it took a bit longer than expected to get the changes out. In this specific case I thought it would be helpful to renmae the function.

crates/block-producer/src/mempool/state.rs
igamigo: You might have misread the second type; I think this is a lookup from note commitment to the graph node that created the note (or I misunderstood something, not sure how we'd map notes to notes here in a way that replaces the functionality of this map).

crates/store/src/server/api.rs
```diff
@@ -153,11 +153,11 @@ where
 
 #[allow(clippy::result_large_err)]
 #[instrument(level = "debug", target = COMPONENT, skip_all, err)]
-pub fn validate_notes(notes: &[proto::primitives::Digest]) -> Result<Vec<NoteId>, Status> {
+pub fn validate_note_commitments(notes: &[proto::primitives::Digest]) -> Result<Vec<Word>, Status> {
     notes
         .iter()
-        .map(|digest| Ok(Word::try_from(digest)?.into()))
-        .collect::<Result<_, ConversionError>>()
+        .map(|digest| Word::try_from(digest).map_err(|_| ()))
```
sergerad:
```suggestion
        .map(Word::try_from)
```

crates/store/src/state.rs
igamigo: I also thought of this but also decided to go with something that replaced what was being done as closely as possible for now

crates/store/src/db/mod.rs
```diff
@@ -454,6 +455,19 @@ impl Db {
         .await
     }
 
+    /// Loads all the [`miden_objects::note::Note`]s matching a certain note commitment from the
```
sergerad: nit: maybe `[`miden_objects::note::NoteRecord`]?

sergerad (approved):
Not super familiar but LGTM. A few nits inline.

crates/store/src/state.rs
igamigo: I thought this would imply yet another call but it was just a matter of changing the previous one (it wasn't used elsewhere), fixed.

crates/block-producer/src/mempool/state.rs
bobbinth: Ah yes! Sorry!

Issue #1200, opened 2025-09-03
Look into reducing vector allocations, especially for fixed-size data like account IDs and nullifiers


drahnr:
We do quite a few allocations when it comes to converting types, using i.e. try_convert and other helpers and collect repeatedly.
Another instance is in the context of #1201

PR #1347, opened 2025-11-10
feat: fetch commitments only

igamigo:
Addresses https://github.com/0xMiden/miden-node/pull/1338#discussion_r2500989020, mainly with the motivation to avoid getting and deserializing more data than needed from the DB.

### DIFF
--- a/crates/store/src/db/models/queries/notes.rs
+++ b/crates/store/src/db/models/queries/notes.rs
@@ -217,6 +217,16 @@ pub(crate) fn select_notes_by_id(
     Ok(records)
 }
 
+/// Select the subset of note commitments that already exist in the notes table
+///
+/// # Raw SQL
+///
+/// ```sql
+/// SELECT
+///     notes.note_commitment
+/// FROM notes
+/// WHERE note_commitment IN (?1)
+/// ```
 pub(crate) fn select_existing_note_commitments(
     conn: &mut SqliteConnection,
     note_commitments: &[Word],


crates/store/src/db/models/queries/notes.rs
Mirko-von-Leipzig, line 1: Would it be possible to check for existence, and then retain the commitment from the inputs instead of re-deserializing it?

I'm unsure how to do this with Diesel, but in raw sqlite I would have done a loop with `exists` to filter the input (in Rust) and build the output hashset.

Mirko-von-Leipzig, line 1: Not for this PR, but just as a thought. Separate (but similar) to the sql query, we may also want to consider minimizing the return message size by returning `true`/`false` for each commitment. This is under the assumption that we expect large note sets and that sometimes there will be many, and sometimes there will be few existing notes.

Though maybe this is premature and we should wait until this shows up as a hotspot. Options as I see them:

1. Return existing notes
2. Return non-existing notes
3. Return flag for each
4. Do any of the above with an enum to optimize dynamically based on whichever is smallest.

cc @bobbinth

igamigo: Something like [this](https://github.com/0xMiden/miden-node/pull/1347/commits/a66baa4565554c449b8748a51f817e57e49dd85d), maybe?

Mirko-von-Leipzig: Not quite; I'm hoping for the database to return a bool per input indicating its existence within the table. @drahnr maybe knows how?

crates/store/src/db/models/queries/notes.rs
```diff
@@ -217,26 +217,29 @@ pub(crate) fn select_notes_by_id(
...
-    let raw: Vec<_> = SelectDsl::select(
-        q,
-        (NoteRecordRawRow::as_select(), schema::note_scripts::script.nullable()),
-    )
-    .load::<(NoteRecordRawRow, Option<Vec<u8>>)>(conn)?;
-    let records = vec_raw_try_into::<NoteRecord, NoteRecordWithScriptRawJoined>(
-        raw.into_iter().map(NoteRecordWithScriptRawJoined::from),
-    )?;
-    Ok(records)
+) -> Result<HashSet<Word>, DatabaseError> {
```
bobbinth, line 233: Could we add doc comments with the SQL statement?

crates/store/src/db/models/queries/notes.rs
```diff
@@ -217,26 +217,29 @@ pub(crate) fn select_notes_by_id(
...
+            .load_iter::<Vec<u8>, _>(conn)?
+            .collect::<diesel::QueryResult<_>>()?;
+
+    let commitments: HashSet<Word> = note_commitments
+        .iter()
+        .zip(&serialized)
+        .filter_map(|(&word, serialized)| {
+            existing_bytes.contains(serialized.as_slice()).then_some(word)
+        })
+        .collect();
```
bobbinth: Question: why do we need to compare against `serialized` here? Wouldn't the data returned from the database already include only the commitments in the `serialized` vec? If so, then the only thing we'd need to do is map bytes to words.

bobbinth (commented):

> Not for this PR, but just as a thought. Separate (but similar) to the sql query, we may also want to consider minimizing the return message size by returning `true`/`false` for each commitment. This is under the assumption that we expect large note sets and that sometimes there will be many, and sometimes there will be few existing notes.
> 
> Though maybe this is premature and we should wait until this shows up as a hotspot. Options as I see them:
> 
> 1. Return existing notes
> 2. Return non-existing notes
> 3. Return flag for each
> 4. Do any of the above with an enum to optimize dynamically based on whichever is smallest.

I think it is a bit too early to work on optimizations like this - but let's create an issue for this (in my mind, this falls into a similar category as https://github.com/0xMiden/miden-node/issues/1200).

crates/store/src/db/models/queries/notes.rs
igamigo: Was mostly an attempt to tackle [this comment](https://github.com/0xMiden/miden-node/pull/1347#discussion_r2509345171), where we avoid deserializing all the commitments again (after having serialized them once already)

bobbinth: I would probably hold off on making this optimization as it is not immediately clear to me that repeated searching through `existing_bytes` would be more performant than just deserializing them.

crates/store/src/db/models/queries/notes.rs
drahnr: I am unsure where this pointed to in the past, both the link and the comment attachment seem to not exist anymore

igamigo: I believe the comment was for the file, and the link does exist, it's a link to a commit I've just reverted. Basically @Mirko-von-Leipzig is referring to [this section of the code](https://github.com/0xMiden/miden-node/pull/1347/files#diff-3a5868c1852c2a410a4336adc4828091c518adb193718e054523baaf998fec01R220-R238), where the input is serialized and then deserialized back to its original form and he's suggesting it could be optimized a bit.

Mirko-von-Leipzig: @drahnr as a summary:

We want to know which of the given input set of note commitments exist in the DB table. This is currently implemented by returning the union of the notes and the input set. This returns the full data for each commitment, which we already have in the input set - and we only want to test for existence actually.

I'm proposing we instead return a bool for each input if it exists in the table.

In raw sql terms:

```sql
-- current
SELECT commitment FROM notes WHERE commitment IN ?;

-- proposed (but maybe there is a better way for multiple rows?)
SELECT EXISTS (SELECT 1 FROM notes where commitment = ?);
```

drahnr: The diesel query would look like this:
```
let sub_query = schema::notes::table
    .select(0.into_sql::<diesel::sql_types::Integer>())
    .filter(schema::notes::commitment.eq(desired_commitment));
let result = schema::notes::table.exists(diesel::dsl::exists(sub_query)).get_results::<bool>(&mut conn)?;
```

We should make sure to have an index for it.

The alt is using an intermediate left join, but I'd not go there unless needed.

Issue #1258, opened 2025-09-30
`SyncTransactions` follow-ups


SantiagoPittella:
After #1224 some follow up took place:

1. There are some inconsistencies in the field's name for `TransactionHeader` proto message, e.g. the `input_notes` field is just `nullifiers`.
2. After retrieving the txs, we make N queries to get the associated notes. We should should be merging into a single query.
3. We should collect the size assumptions for the paginated queries into a single module, were we can be more consistent between endpoint. Also, we should set 4MB as the payload limit for all endpoints.
4. Rename transaction_records -> transaction in the proto message.
5. Rename transaction_header -> header in the proto message.

PR #1357, opened 2025-11-11
chore(breaking): sync transactions follow ups

SantiagoPittella:
closes https://github.com/0xMiden/miden-node/issues/1258

This PR:
- Fixes inconsistencies in the naming of the input note nullifiers for transactions. Renames `input_notes` to `nullifiers` in the transactions table.
- Removes N+1 queries to retrieve associated notes.
- Collects size assumptions for paginated queries in a single file, and set them all to 4MB.
- Renames in the proto messages: `transaction_records -> transactions`, `transaction_header -> header`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 - [BREAKING] Renamed `SyncTransactions` response fields ([#1357](https://github.com/0xMiden/miden-node/pull/1357)).
 - Normalize response size in endpoints to 4 MB ([#1357](https://github.com/0xMiden/miden-node/pull/1357)).
+- [BREAKING] Renamed `ProxyWorkerStatus::address` to `ProxyWorkerStatus::name` ([#1348](https://github.com/0xMiden/miden-node/pull/1348)).
+- Remove `trait AccountTreeStorage` ([#1352](https://github.com/0xMiden/miden-node/issues/1352)).
+
+## v0.12.2 (2025-11-12)
+
+- Fixed `PoW` challenge solving in `miden-network-monitor` binary ([#1363](https://github.com/0xMiden/miden-node/pull/1363)).
 
 ## v0.12.1 (2025-11-08)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -196,7 +196,7 @@ checksum = "c7c24de15d275a1ecfd47a380fb4d5ec9bfe0933f309ed5e705b775596a3574d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.109",
+ "syn 2.0.110",
 ]
 
 [[package]]
@@ -207,7 +207,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.109",
+ "syn 2.0.110",
 ]
 
 [[package]]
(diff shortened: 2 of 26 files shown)


crates/store/src/server/rpc_api.rs
```diff
@@ -566,28 +559,47 @@ impl rpc_server::Rpc for StoreApi {
             .await
             .map_err(SyncTransactionsError::from)?;
 
+        // Collect all note IDs from all transactions to make a single query
+        let mut all_note_ids = Vec::new();
+        for tx_record in &transaction_records_db {
+            all_note_ids.extend(tx_record.output_notes.iter().copied());
+        }
```
Mirko-von-Leipzig: Nit: I would probably do this with iterators


```suggestion
        let all_notes_ids = transaction_records_db
            .iter()
            .flat_map(|tx| tx.output_notes.iter())
            .copied()
            .collect::<Vec<_>>();
```

crates/store/src/server/rpc_api.rs
```diff
@@ -566,28 +559,47 @@ impl rpc_server::Rpc for StoreApi {
...
+            .state
+            .get_notes_by_id(all_note_ids)
+            .await
+            .map_err(SyncTransactionsError::from)?;
+
+        // Create a map from note ID to note record for efficient lookup
+        let note_map: std::collections::HashMap<_, _> = all_note_records
+            .into_iter()
+            .map(|note_record| (note_record.note_id, note_record))
+            .collect();
```
Mirko-von-Leipzig, line 580: This, and the lookup below, makes me wonder if we shouldn't be doing a join in the `state.sync_transactions` database call to include the note data.

cc @drahnr - thoughts? I'm feeling a bit out of touch with the code in question so maybe I'm overlooking something.

Mirko-von-Leipzig (commented):
Code looks good; just the question about whether to inline the note retrieval as part of the original database query.

crates/store/src/server/rpc_api.rs
SantiagoPittella: It might be a good idea, not sure if it will require some schema changes. Though considering that we compact it into a single followup query I'm not sure it would be much more better.

I was waiting for Bernhard input before implementing it, but let me know if you think that I should start implementing it.

Mirko-von-Leipzig: I think lets wait for his input. My concern with multiple, separate DB calls is that it also means they're not the same db transaction, so if we start pruning note data at some point this will run into issues.

drahnr: `        .filter(schema::notes::note_id.eq_any(&note_ids));
`
is done in both SQL queries, this means if we do delete something between those transactions it won't be an issue, it'll simply not show up in the results. For adding an account, this is the same, won't show up in the results.
Deletion is not a problem; missing data is to my best understanding either.

---

Using a join is probably the right approach, but I'd write an issue and do it as a follow-up. I don't think it's urgent.

proto/proto/types/transaction.proto
```diff
@@ -54,7 +54,7 @@ message TransactionHeader {
     primitives.Digest final_state_commitment = 3;
 
     // Nullifiers of the input notes of the transaction.
-    repeated primitives.Digest input_notes = 4;
+    repeated primitives.Digest nullifiers = 4;
```
drahnr, line 57: Thank you ❤️  There's nothing like getting confused by a good old misnomer.

crates/store/src/server/rpc_api.rs
Mirko-von-Leipzig: Issue created #1368

PR #1359, opened 2025-11-11
add a regression test for #1312

drahnr:
Recently we encountered an issue - specificall #1312 - in our test deployment. The component is scheduled for replacement. This is a regression test for the specific condition - it will have to be adapted for the new approach.

### DIFF
--- a/crates/ntx-builder/src/state/tests.rs
+++ b/crates/ntx-builder/src/state/tests.rs
@@ -4,7 +4,6 @@ use miden_node_proto::domain::mempool::MempoolEvent;
 use miden_node_utils::fee::test_fee_params;
 use miden_objects::Word;
 use miden_objects::block::BlockHeader;
-use miden_objects::crypto::merkle::{MmrPeaks, PartialMmr};
 use miden_objects::note::Nullifier;
 use miden_objects::transaction::{PartialBlockchain, TransactionId};
 


crates/rpc/src/server/validator.rs
```diff
@@ -136,9 +136,9 @@ impl DataStore for TransactionInputsDataStore {
 
     fn get_note_script(
         &self,
-        script_root: Word,
-    ) -> impl FutureMaybeSend<Result<NoteScript, DataStoreError>> {
-        async move { Err(DataStoreError::NoteScriptNotFound(script_root)) }
+        _script_root: Word,
+    ) -> impl FutureMaybeSend<Result<Option<NoteScript>, DataStoreError>> {
+        async move { Ok(None) }
```
Mirko-von-Leipzig, line 141: When does one return `None` and when `DataStoreError`? Or rather, what does `None` imply?

crates/ntx-builder/src/state/mod.rs
```diff
@@ -130,6 +133,28 @@ impl State {
         Ok(state)
     }
 
+    /// Creates a new State for testing purposes with minimal setup.
+    ///
+    /// This bypasses the need for a real store connection and provides a mock state
+    /// that can be used in unit tests.
+    #[cfg(test)]
+    pub(crate) fn new_for_testing(
```
Mirko-von-Leipzig, line 141: Ideally we would get an in-memory `Store` implementation but that's a separate discussion :)

crates/ntx-builder/src/state/tests.rs
```diff
@@ -0,0 +1,87 @@
...
+        Word::default(),   // chain_root
+        Word::default(),   // account_root
+        Word::default(),   // nullifier_root
+        Word::default(),   // note_root
+        Word::default(),   // tx_hash
+        Word::default(),   // kernel_root
+        Word::default(),   // proof_hash
+        test_fee_params(), // fee_parameters
+        0_u32,             // timestamp
+    );
```
Mirko-von-Leipzig, line 29: @PhilippGackstatter wen new types 😅

crates/ntx-builder/src/state/tests.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    // Create an empty partial blockchain
+    let chain_mmr = PartialBlockchain::new(PartialMmr::from_peaks(MmrPeaks::default()), Vec::new())
+        .expect("Failed to create empty PartialBlockchain");
```
Mirko-von-Leipzig: I'm surprised there isn't an empty method?

Mirko-von-Leipzig (approved):

crates/rpc/src/server/validator.rs
PhilippGackstatter: See doc comment:
https://github.com/0xMiden/miden-base/blob/d52c65e85949b0b85b38fef04b2e86ab5cbe281a/crates/miden-tx/src/executor/data_store.rs#L74-L76

I generally like `None` as a not found indicator, because it's often an outcome you want to handle whereas errors you usually just propagate.

crates/ntx-builder/src/state/tests.rs
PhilippGackstatter: `PartialBlockchain::default` exists (https://docs.rs/miden-objects/latest/miden_objects/transaction/struct.PartialBlockchain.html#impl-Default-for-PartialBlockchain)

crates/ntx-builder/src/state/tests.rs
PhilippGackstatter: I don't think it makes sense without https://github.com/0xMiden/crypto/issues/430 and that requires a bit of work.

drahnr: I can do that finger exercise, I oftentimes find myself ~cursing~ searching for meaning in word arguments.

Issue #1352, opened 2025-11-10
Enable `RocksDB` backend for `AccountTreeWithHistory`


drahnr:

bobbinth:
As a part of this, we should also consider getting rid of the `AccountTreeStorage` trait. We should be able to do it by defining, the struct as something like this:

```rust
pub struct AccountTreeWithHistory<S: SmtStorage> {
    block_number: BlockNumber,
    latest: AccountTree<S>,
    overlays: BTreeMap<BlockNumber, HistoricalOverlay>,
}
```

bobbinth:
@drahnr - is this addressed already, or is there anything outstanding?

drahnr:
Done, nothing outstanding to my knowledge

PR #1361, opened 2025-11-11
refactor: get rid of account tree storage trait

drahnr:
As commented on #1352 and mentioned in a comment in #1292 

tl;dr the trait is not necessary, `SmtStore` from `miden-crypto` introduced for `LargeSmt` is sufficient, since we always use that for `AccountTree`.


bobbinth (approved):
Looks good! Thank you!

Issue #1343, opened 2025-11-07
Network Monitor: Split counter increment and track functionality


SantiagoPittella:
Currently in the network monitor, we send a transaction to increment the counter of a network account and immediately after we request the network account details to get the counter value. We can split this functionality in two, having a task to send the counter increase transactions, and another to track the counter value. Some coupling would still be required due to https://github.com/0xMiden/miden-node/issues/1342

PR #1362, opened 2025-11-11
chore: split ntx related tasks

### LINKED ISSUE #1342: Network Monitor: track latency in network transaction
We need to track the time between sending a transaction that creates a network note and when the network account consumes it, and display it on the frontend.

SantiagoPittella:
# **This commit aims to the `main` branch.**

closes [#1343](https://github.com/0xMiden/miden-node/issues/1343) & closes #1342 

Splits the network account related cards and backend process into two separated tasks. This PR also adds last and average latency, tracking the time between the tx submit and when the counter is increased. 

Screenshot 2025-11-11 at 19 33 53

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v0.12.4 (TBD)
 
 - Split counter increment and tracking services in `miden-network-monitor` binary ([#1362](https://github.com/0xMiden/miden-node/pull/1362)).
+- Updated the counter account from the `miden-network-monitor` to start at 0 ([#1367](https://github.com/0xMiden/miden-node/pull/1367)).
 
 ## v0.12.3 (2025-11-15)
 
--- a/bin/network-monitor/src/deploy/counter.rs
+++ b/bin/network-monitor/src/deploy/counter.rs
@@ -38,8 +38,7 @@ pub fn create_counter_account(owner_account_id: AccountId) -> Result<Account> {
         owner_account_id_prefix,
     ]));
 
-    let counter_slot =
-        StorageSlot::Value(Word::from([Felt::ZERO, Felt::ZERO, Felt::ZERO, Felt::ONE]));
+    let counter_slot = StorageSlot::Value(Word::empty());
 
     let account_code = AccountComponent::compile(
         script,


bin/network-monitor/src/counter.rs
```diff
@@ -378,6 +328,107 @@ fn send_status(tx: &watch::Sender<ServiceStatus>, status: ServiceStatus) -> Resu
...
+                error!("Failed to fetch counter value: {:?}", e);
+                Some(format!("fetch counter value failed: {e}"))
+            },
+        };
+
+        let status = build_tracking_status(&details, last_error);
+        send_status(&tx, status)?;
+
+        // Fetch faster than the increment interval to ensure we don't miss any increments
+        sleep(config.counter_increment_interval / 2).await;
```
Mirko-von-Leipzig: There could still be weirdness though right? If the ntx builder decides to wait for more counter notes before applying them in a tx the count would skip ahead, causing a desync I think?

bin/network-monitor/src/counter.rs
```diff
@@ -378,6 +328,107 @@ fn send_status(tx: &watch::Sender<ServiceStatus>, status: ServiceStatus) -> Resu
...
+
+                        // Add to recent latencies and calculate average
+                        recent_latencies.push(latency_ms);
+                        if recent_latencies.len() > MAX_LATENCY_SAMPLES {
+                            recent_latencies.remove(0);
+                        }
+                        if !recent_latencies.is_empty() {
+                            let sum: u64 = recent_latencies.iter().sum();
+                            details.avg_latency_ms = Some(sum / recent_latencies.len() as u64);
+                        }
```
Mirko-von-Leipzig: An often simpler alternative is to use an exponentially decaying weighted moving average which avoids storing history. What you do is:

```rust
average = average / (MAX_LATENCY_SAMPLES - 1) + latency_ms;
```

Or in percentages (which may make the mechanism more obvious):

```
average = average * 90% + new_value * 10%
```

where `10%` sort of approximates `1 / MAX_LATENCY_SAMPLES` if the const the set to `10`. The downside is that the initial averages will be inaccurate but that's often irrelevant.

bin/network-monitor/src/counter.rs
SantiagoPittella: It’s incremented in `handle_increment_success` once per successfully submitted increment tx, so it shouldn't _desync_ . Though the sleep comment is misleading, will change that.

Also, I should be using `tick` to be more consistent with the node in general instead of sleep, will change that.

PR #1363, opened 2025-11-12
fix(network monitor: faucet integration

SantiagoPittella:
# **This PR aims to the `main` branch.**

In the new release of the faucet, the challenge encoding was changed. We need to update how we interpret the challenge before searching for a solution.


TomasArrachea (approved):
LGTM

PR #1365, opened 2025-11-12
feat: timeouts for prover clients

SantiagoPittella:
# **This PR aims to the `main` branch**

This PR adds a `with_timeout` method for the remote prover clients (Transaction, batch and block) that allows the configuration of the timeout.

Also, add a `--request-timeout` configuration in the remote prover to set the outgoing requests timeout.

### DIFF
--- a/crates/remote-prover-client/src/remote_prover/tx_prover.rs
+++ b/crates/remote-prover-client/src/remote_prover/tx_prover.rs
@@ -42,7 +42,7 @@ impl RemoteTransactionProver {
         RemoteTransactionProver {
             endpoint: endpoint.into(),
             client: Arc::new(Mutex::new(None)),
-            timeout: Duration::from_millis(10000), // 10 seconds default
+            timeout: Duration::from_secs(10),
         }
     }
 


crates/remote-prover-client/src/remote_prover/batch_prover.rs
```diff
@@ -40,9 +41,25 @@ impl RemoteBatchProver {
         RemoteBatchProver {
             endpoint: endpoint.into(),
             client: Arc::new(Mutex::new(None)),
+            timeout: Duration::from_millis(10000), // 10 seconds default
```
Mirko-von-Leipzig:
```suggestion
            timeout: Duration::from_secs(10),
```

crates/remote-prover-client/src/remote_prover/tx_prover.rs
```diff
@@ -41,9 +42,25 @@ impl RemoteTransactionProver {
         RemoteTransactionProver {
             endpoint: endpoint.into(),
             client: Arc::new(Mutex::new(None)),
+            timeout: Duration::from_millis(10000), // 10 seconds default
```
Mirko-von-Leipzig:
```suggestion
            timeout: Duration::from_secs(10),
```

SantiagoPittella:
> Was the main motivation for this for the network monitor?

Yesterday, the remote prover was taking between 12 and 15 seconds prove, so the operation was being cancelled by the client because of the 10s timeout.

PR #1366, opened 2025-11-13
chore: improve tracing in network monitor

SantiagoPittella:
# **This PR uses `main` as base branch**

Improves the instrumentation in the `miden-network-monitor` binary.

This is the current view of the traces (in jaeger):

Screenshot 2025-11-13 at 17 21 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - The mempool's transaction capacity is now configurable ([#1433](https://github.com/0xMiden/miden-node/pull/1433)).
 - Renamed card's names in the `miden-network-monitor` binary ([#1441](https://github.com/0xMiden/miden-node/pull/1441)).
 - Improved tracing in `miden-network-monitor` binary ([#1366](https://github.com/0xMiden/miden-node/pull/1366)).
+- Integrated RPC stack with Validator component for transaction validation ([#1457](https://github.com/0xMiden/miden-node/pull/1457)).
 
 ### Changes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2499,7 +2499,7 @@ dependencies = [
 [[package]]
 name = "miden-block-prover"
 version = "0.13.0"
-source = "git+https://github.com/0xMiden/miden-base.git?branch=next#4b08b9c916bb3a1ecc7e509d51564e9860c1bbc8"
+source = "git+https://github.com/0xMiden/miden-base.git?branch=next#198e25ee552638caefa356ff5b248a654b119083"
 dependencies = [
  "miden-objects",
  "thiserror 2.0.17",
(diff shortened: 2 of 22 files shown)


SantiagoPittella:
@bobbinth @Mirko-von-Leipzig `mdbook` publishing a new, non compatible, version a few hours ago. I tried removing the multilingual option but other errors arised. Should I lock the Ci version to the previous (`v0.4.52`) until we complete the migration? `mdbook` people provides a migration guide https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#05-migration-guide

bin/network-monitor/src/commands/start.rs
```diff
@@ -27,19 +36,24 @@ pub async fn start_monitor(config: MonitorConfig) -> Result<()> {
     };
 
     let mut tasks = Tasks::new();
+    debug!(target: COMPONENT, "Initializing monitoring tasks");
 
     // Initialize the RPC Status endpoint checker task.
+    debug!(target: COMPONENT, "Initializing RPC status checker");
```
igamigo: nit: Emitting 2 events here feels a bit redundant/unnecessary

bin/network-monitor/src/status.rs
```diff
@@ -398,16 +422,25 @@ pub async fn run_remote_prover_status_task(
...
-#[instrument(target = COMPONENT, name = "check-status.remote-prover", skip_all, ret(level = "info"))]
+#[instrument(
+    parent = None,
+    target = COMPONENT,
+    name = "network_monitor.status.check_remote_prover_status",
+    skip_all,
+    level = "info",
+    ret(level = "debug")
+)]
 pub(crate) async fn check_remote_prover_status(
```
igamigo, line 463: Something that could be nice for troubleshooting is to raise info events for service status changes. Something like 

```rust
  let mut last = Status::Unknown;
  // ...
  let status = check_remote_prover_status(...).await;
  if status.status != last {
      info!(target: COMPONENT, prover = %name, status = ?status.status, "Remote prover status changed");
      last = status.status;
  }
```

But maybe this is already easily filterable with the current traces

igamigo (approved):
Not sure if the plan is still to merge this and patch release it, but LGTM! Sorry for the delay in the review.
Not super familiar with the frequencies of some processes. Doesn't seem like these changes would be very spammy but it's probably something to watch out for.

bin/network-monitor/src/commands/start.rs
SantiagoPittella: I removed the first one. For the debug! level I left one log per component.

bin/network-monitor/src/status.rs
SantiagoPittella: In line 443 we have this:
```rust
   debug!(target: COMPONENT, prover_name = %name, remote_prover_status = ?status, "Remote prover status check successful");
```

Is that what you meant?

igamigo: Mostly meant emitting events when the status changes (so for example, when a service goes from healthy to unhealthy or viceversa). It feels like it could be useful for troubleshooting, but maybe not.

SantiagoPittella: Probably, but the idea of the monitor is not troubleshooting but a quick glance of the overall state of the components

sergerad (approved):

PR #1367, opened 2025-11-16
chore: start counter at zero

SantiagoPittella:
In miden-base v0.12.3 the empty storage value initialization bug was fixed, that bug was making us to start the counter in 1 instead of 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.12.4 (TBD)
+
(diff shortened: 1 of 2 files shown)


igamigo (approved):
Looks good! Not sure if we prefer to merge this to `next` instead of `main`, though.

SantiagoPittella:
Last week @bobbinth  mentioned that I should open network mmonitor-related PRs against `main`, thought the situation might have changed. I can change the base in that case.

bobbinth (approved):
Looks good! Thank you!

Mirko-von-Leipzig (approved):
I'm happy if minor improvements like this go straight to main. So long as its obvious they aren't breaking anything, then its nice to have these improvements sooner.

PR #1370, opened 2025-11-17
fix: RPC client ignores genesis if version is unset

Mirko-von-Leipzig:
While fixing up the tests in external PR #1298 I noticed that our gRPC clients only set the `genesis` metadata value if the `version` field is set.

This PR fixes this, and also reduces the type shenanigans we do in the gRPC client builder.

This PR blocks progress on #1298.

### DIFF
--- a/bin/network-monitor/src/deploy/mod.rs
+++ b/bin/network-monitor/src/deploy/mod.rs
@@ -96,7 +96,7 @@ pub async fn deploy_counter_account(counter_account: &Account, rpc_url: &Url) ->
         .with_timeout(Duration::from_secs(5))
         .without_metadata_version()
         .without_metadata_genesis()
-        .disable_otel_context_injection()
+        .without_otel_context_injection()
         .connect()
         .await
         .context("Failed to connect to RPC server")?;
--- a/crates/proto/src/clients/mod.rs
+++ b/crates/proto/src/clients/mod.rs
@@ -5,16 +5,19 @@
 //!
 //! # Examples
 //!
-//! ```rust,no_run
-//! use miden_node_proto::clients::{Builder, WantsTls, StoreNtxBuilderClient, StoreNtxBuilder};
+//! ```rust
+//! # use miden_node_proto::clients::{Builder, WantsTls, StoreNtxBuilderClient};
+//! # use url::Url;
 //!
 //! # async fn example() -> anyhow::Result<()> {
 //! // Create a store client with OTEL and TLS
-//! let client: StoreNtxBuilderClient = Builder::new("https://store.example.com")?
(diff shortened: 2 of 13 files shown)


crates/proto/src/clients/mod.rs
```diff
@@ -369,30 +423,46 @@ impl Builder<WantsGenesis> {
     }
 }
```
SantiagoPittella, line 430: The return type of this methods should be changed to `Builder<WantsOtel>`, right?

Mirko-von-Leipzig: Ah yes! Thank you :)

bin/network-monitor/src/deploy/mod.rs
```diff
@@ -96,7 +96,8 @@ pub async fn deploy_counter_account(counter_account: &Account, rpc_url: &Url) ->
         .with_timeout(Duration::from_secs(5))
         .without_metadata_version()
         .without_metadata_genesis()
-        .connect::<Rpc>()
+        .disable_otel_context_injection()
```
Mirko-von-Leipzig: I've disabled OTel for the monitor, and enabled it for our node internals which I _think_ is correct.

It might be nice to enable it for the monitor in theory, but since it accesses via the public RPC the RPC server doesn't handle it on its end.

crates/proto/src/clients/mod.rs
```diff
@@ -15,228 +15,277 @@
 //!     .without_timeout()           // or `.with_timeout(Duration::from_secs(10))`
 //!     .without_metadata_version()  // or `.with_metadata_version("1.0".into())`
 //!     .without_metadata_genesis()  // or `.with_metadata_genesis(genesis)`
-//!     .connect::<StoreNtxBuilder>()
+//!     .connect::<StoreNtxBuilderClient>()
```
SantiagoPittella, line 21:
```suggestion
//!     .enable_otel_context_injection()
//!     .connect::<StoreNtxBuilderClient>()
```

crates/proto/src/clients/mod.rs
```diff
@@ -261,7 +310,7 @@ impl GrpcClientBuilder for RemoteProverClient {
 ///     .with_timeout(Duration::from_secs(5)) // or `.without_timeout()`
 ///     .with_metadata_version("1.0".into()) // or `.without_metadata_version()`
 ///     .without_metadata_genesis()           // or `.with_metadata_genesis(genesis)`
-///     .connect::<Rpc>()
+///     .connect::<RpcClient>()
```
SantiagoPittella, line 319:
```suggestion
///     .enable_otel_context_injection()
///     .connect::<RpcClient>()
```

Issue #1371, opened 2025-11-18
Network Monitor: expand faucet information


SantiagoPittella:
Currently the network monitor only displays information related to the mint transaction and challenge difficulty on the frontend. We could use (and even expand) the `get_metadata` endpoint from the faucet API to display information about the token symbol, minted amount and total tokens. Also, we should add information about which version the faucet is running.

PR #1373, opened 2025-11-19
chore: add more faucet information

SantiagoPittella:
Closes https://github.com/0xMiden/miden-node/issues/1371 

Faucet version is on PR and isn't added to the endpoint yet https://github.com/0xMiden/miden-faucet/pull/169

### DIFF
--- a/bin/network-monitor/src/faucet.rs
+++ b/bin/network-monitor/src/faucet.rs
@@ -59,8 +59,6 @@ struct GetTokensResponse {
 }
 
 /// Response from the faucet's `/get_metadata` endpoint.
-///
-/// TODO: We need to add `faucet_version` once available.
 #[derive(Clone, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


bin/network-monitor/src/faucet.rs
```diff
@@ -58,6 +58,21 @@ struct GetTokensResponse {
     note_id: String,
 }
 
+/// Response from the faucet's `/get_metadata` endpoint.
+///
+/// TODO: We need to add `faucet_version` once available.
```
Mirko-von-Leipzig: What's the difference between `faucet_version` and `version`?

Mirko-von-Leipzig: Or did you mean `version` and that it should be made non-optional once its there?

SantiagoPittella: That is an outdated comment, will remove it. THose are the same.

Issue #1378, opened 2025-11-25
Minor `get_account_proof` improvements


Mirko-von-Leipzig:
#1333 merged initial support for account proof history. Some minor improvements were suggested post-merge.

> We should be able to check whether an account is a public account by just looking at the account ID. This way, we can avoid querying the database before we check if the account is public.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-node/pull/1333#discussion_r2557939771_

I also suggest we factor out two functions from `get_account_proof`, 

1. `get_block_witness` (or proof)
  - This only uses `self.inner`, and would make it hard to drop the guard incorrectly
2. `get_account_details`

The current function then becomes something like

```rust
pub async fn get_account_proof(
    &self,
    request: AccountProofRequest
) -> Result<AccountProofResponse, DatabaseError> {
    // Bobbin's suggestion
    if request.details.is_some() && !request.account_id.is_public() {
        return Err(DatabaseError::AccountNotPublic(request.account_id));
    }

    let (block_num, witness) = self.get_block_witness(request.block_num, request.account_id)?;

    let details = if let Some(request) = request.details {
        self.get_account_details(request).await?
    } else {
        None
    };

    Ok(AccountProofResponse {block_num, witness, details})
}
```

PR #1379, opened 2025-11-25
store: refactor/simplify get_account* 

drahnr:
Addresses #1378

### DIFF
--- a/crates/store/src/errors.rs
+++ b/crates/store/src/errors.rs
@@ -104,6 +104,8 @@ pub enum DatabaseError {
     AccountsNotFoundInDb(Vec<AccountId>),
     #[error("account {0} is not on the chain")]
     AccountNotPublic(AccountId),
+    #[error("account {0} details missing")]
+    AccountDetailsMissing(AccountId),
     #[error("invalid block parameters: block_from ({from}) > block_to ({to})")]
     InvalidBlockRange { from: BlockNumber, to: BlockNumber },
     #[error("data corrupted: {0}")]
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -929,7 +929,7 @@ impl State {
         let (block_num, witness) = self.get_block_witness(block_num, account_id).await?;
 
         let details = if let Some(request) = details {
-            Some(self.fetch_account_proof_details(account_id, block_num, request).await?)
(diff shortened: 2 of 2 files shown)


crates/store/src/state.rs
```diff
@@ -946,86 +966,88 @@ impl State {
...
 
-            let storage_header = account.storage().to_header();
+        let account_info = self.db.select_historical_account_at(account_id, block_num).await?;
 
-            let mut storage_map_details =
-                Vec::<AccountStorageMapDetails>::with_capacity(storage_requests.len());
+        // if we get a query for a _private_ account _with_ details requested, we'll error out
+        let Some(account) = account_info.details else {
+            return Err(DatabaseError::AccountNotPublic(account_id));
+        };
```
Mirko-von-Leipzig: If for some reason we corrupt or screw up the database such that a public account is missing the details then this is going to be quite confusing.

Maybe we can rename to `fn fetch_public_account_proof_details` and make this a `panic!` instead, pushing the invariant to the caller (which is our pub fn).

A more type safe approach could be wrapping `AccountDetailRequest` such that the new type can only be constructed if the account ID is public.

drahnr: I'd tend to add another error variant, `"account details missing"` for sake of simplicity. I am not keen on adding panics, from experience, you'll always find them in production.

bobbinth: I think returning an error here is fine, but I'd create a new error variant for this.

Also, I'd probably rename the function to `fetch_public_account_details()`

bobbinth (approved):
Looks good! Thank you!

crates/store/src/state.rs
drahnr: Done

crates/store/src/errors.rs
```diff
@@ -104,6 +104,8 @@ pub enum DatabaseError {
     AccountsNotFoundInDb(Vec<AccountId>),
     #[error("account {0} is not on the chain")]
     AccountNotPublic(AccountId),
+    #[error("account {0} details missing")]
+    AccountDetailsMissing(AccountId),
```
igamigo, line 108: nit: I think it'd be better if this were named `PublicAccountDetailsMissing` and/or if the error message included the fact that the account is expected to be public. On a first look, otherwise, it looks ambiguous because there's also `AccountsNotFoundInDb`

Issue #1316, opened 2025-10-27
Guardrails: Tx and Block Validation


sergerad:
## Context

Following from https://github.com/0xMiden/miden-node/issues/613, we are moving forward with the design and implementation of transaction and block validation (guardrails) and deferred proving of blocks.

See this [comment](https://github.com/0xMiden/miden-base/issues/2009#issuecomment-3450002903) for context on deferred block proving. The TLDR is that signed (validated), _unproven_ blocks will be sent to the store. The store will then be responsible for proving blocks.

## Scope

The following changes will be required to complete this piece of work:

**Validator**
- Transaction submission, re-execution, and storage.
- Block submission, validation, and signing.

**Block Producer**
- Updated flow for deferred proving.
- Integration with Validator.

**Store**
- Updated flow for deferred proving.
- Integration with remote prover.

**Network Transaction Builder**
- Integration with Validator.

### Out of Scope

Miden Client changes will need to be covered in a separate issue.

Miden base changes will be handled in separate issues such as https://github.com/0xMiden/miden-base/issues/2009.

## Error Scenarios

The new flows pertaining to the Validator and deferred block proving create the following new error scenarios that we must handle:

**Transaction validation failure**
- Proven transactions are not able to be re-executed upon submission to the Validator.
- Prevents subsequent Block validation because invalid transactions will be present in them. ❌ 

**Block validation failure**
- Blocks are not able to be validated and signed upon submission to the Validator.
- Prevents Block proving. ❌

**Block proving failure**
- Creates a desync scenario between clients that rely on validated / signed blocks. ❌ 

## Plan

The work will be sliced into the following iterations:
- [x] Define full flow of transactions and blocks involving the Validator and deferred block proving.
- [x] Update miden-base types for block signing and deferred proving.
  - https://github.com/0xMiden/miden-base/pull/2012
  - https://github.com/0xMiden/miden-base/pull/2128
- [x] Create Validator component and integrate with RPC for transaction submission.
  - https://github.com/0xMiden/miden-node/pull/1293
  - https://github.com/0xMiden/miden-node/pull/1457
- [x] Integrate NTB with Validator.
  - https://github.com/0xMiden/miden-node/issues/1309
- [x] Add transaction validation logic to Validator.
   - In memory: #1460
- [x] Add block validation endpoints and logic to Validator.
   - https://github.com/0xMiden/miden-node/pull/1381
   - https://github.com/0xMiden/miden-node/pull/1426
[...]

PR #1381, opened 2025-11-25
feat: Validate and prove blocks

sergerad:
## Context

We have updated base types to support deferred block proving. The node needs to be integrated with the new types.

The deferred proving flow is being worked towards. Eventually the store will make the proving request that the block producer is making currently. In the meantime, we can iterate on the current block production flow by adding block validation via the validator component.

Relates to #1316.

## Changes
- Update base dependancies.
- Add block validation endpoint to validator component.
- Add validator RPC client.
- Integrate validator client to block producer.
- Update block producer to validate blocks through the validator rpc.

### DIFF
--- a/crates/store/Cargo.toml
+++ b/crates/store/Cargo.toml
@@ -15,35 +15,33 @@ version.workspace      = true
 workspace = true
 
 [dependencies]
-anyhow = { workspace = true }
-deadpool = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
-deadpool-diesel = { features = ["sqlite"], version = "0.6" }
-deadpool-sync = { version = "0.1" }
-diesel = { features = ["numeric", "sqlite"], version = "2.2" }
-diesel_migrations = { features = ["sqlite"], version = "2.2" }
-fs-err = { workspace = true }
-hex = { version = "0.4" }
-indexmap = { workspace = true }
-miden-lib = { workspace = true }
-miden-node-proto = { workspace = true }
+anyhow                 = { workspace = true }
+deadpool               = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
+deadpool-diesel        = { features = ["sqlite"], version = "0.6" }
+deadpool-sync          = { version = "0.1" }
+diesel                 = { features = ["numeric", "sqlite"], version = "2.2" }
+diesel_migrations      = { features = ["sqlite"], version = "2.2" }
+fs-err                 = { workspace = true }
+hex                    = { version = "0.4" }
+indexmap               = { workspace = true }
+miden-lib              = { workspace = true }
+miden-node-proto       = { workspace = true }
 miden-node-proto-build = { features = ["internal"], workspace = true }
-miden-node-utils = { workspace = true }
-miden-objects = { features = [
-  "std",
-  "testing",
-], workspace = true } # TODO remove testing, required for `BlockProof::new_dummy`
+miden-node-utils       = { workspace = true }
+# TODO remove `testing` from `miden-objects`, required for `BlockProof::new_dummy`
+miden-objects     = { features = ["std", "testing"], workspace = true }
 pretty_assertions = { workspace = true }
-rand = { workspace = true }
-rand_chacha = { workspace = true }
-serde = { features = ["derive"], version = "1" }
-thiserror = { workspace = true }
-tokio = { features = ["fs", "rt-multi-thread"], workspace = true }
-tokio-stream = { features = ["net"], workspace = true }
-toml = { version = "0.9" }
-tonic = { workspace = true }
-tonic-reflection = { workspace = true }
-tower-http = { features = ["util"], workspace = true }
-tracing = { workspace = true }
+rand              = { workspace = true }
+rand_chacha       = { workspace = true }
+serde             = { features = ["derive"], version = "1" }
+thiserror         = { workspace = true }
+tokio             = { features = ["fs", "rt-multi-thread"], workspace = true }
+tokio-stream      = { features = ["net"], workspace = true }
+toml              = { version = "0.9" }
+tonic             = { workspace = true }
+tonic-reflection  = { workspace = true }
+tower-http        = { features = ["util"], workspace = true }
+tracing           = { workspace = true }
 
 [dev-dependencies]
 assert_matches        = { workspace = true }


crates/block-producer/src/block_builder/mod.rs
```diff
@@ -112,8 +121,10 @@ impl BlockBuilder {
         self.get_block_inputs(selected)
             .inspect_ok(BlockBatchesAndInputs::inject_telemetry)
             .and_then(|inputs| self.propose_block(inputs))
-            .inspect_ok(ProposedBlock::inject_telemetry)
-            .and_then(|inputs| self.prove_block(inputs))
+            .inspect_ok(|(proposed_block, _)| {
+                ProposedBlock::inject_telemetry(proposed_block);
+            })
+            .and_then(|(proposed_block, inputs)| self.validate_and_prove_block(proposed_block, inputs))
```
sergerad: @bobbinth Is this still where the proving flow will run? Do we want to remove proving from here altogether? And move it to the store?

crates/proto/src/domain/proof_request.rs
```diff
@@ -0,0 +1,41 @@
...
+        todo!()
+        //let block = Self {
+        //    tx_batches: OrderedBatches::read_from(source)?,
+        //    block_header: BlockHeader::read_from(source)?,
+        //    // TODO(serge): Add serde for block inputs
+        //    block_inputs: BlockInputs::read_from(source)?,
+        //};
+
+        //Ok(block)
+    }
```
sergerad, line 38: This will be unblocked by https://github.com/0xMiden/miden-base/pull/2112 if we decide to keep this approach

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -197,29 +214,57 @@ impl BlockBuilder {
...
+        proposed_block: ProposedBlock,
+        block_inputs: BlockInputs,
+    ) -> Result<ProvenBlock, BuildBlockError> {
+        // Validate the block.
+        let message = miden_node_proto::generated::blockchain::ProposedBlock {
+            proposed_block: proposed_block.to_bytes(),
+        };
+        tracing::debug!(target: COMPONENT, ?message);
+        let request = tonic::Request::new(message);
+        let _response = self.validator.clone().validate_block(request).await.unwrap(); // todo: rm unwrap
```
sergerad: @bobbinth would the block producer do this here?

crates/block-producer/src/block_builder/mod.rs
bobbinth: We do, but I would do that in a separate PR because this would entail quite a few more changes.

Mirko-von-Leipzig: I would suggest keeping validation and proving as separate steps in this process here. That way they get separate spans and removing the proving stage later on will be cleaner.

But maybe I'm missing something

crates/block-producer/src/block_builder/mod.rs
bobbinth: Yes - though I was imagining the flow a bit differently:

1. We get the proposed block and send it to the validator [as we do now].
2. The validator builds the block and returns the block header and body. As a part of this, the validator would sign the blog header (this part still needs to be implemented in `miden-base`).
3. The block producer checks them to make sure they are consistent.
4. Then, we send the block header, transaction batches, and block inputs to the prover (we currently send the block body to the prove as well, but I'm not sure why this is needed).
5. The prover returns the block proof.
6. We build the proven block.

Once we move the proving to the store, the last 3 steps will need to happen there. For this, instead of sending a proven block to the store, we'll need to send `(block_header, block_body, tx_batches, block_inputs)` to the store (probably organized into a struct). We could also skip sending `block_inputs` because the store should theoretically be able to build them - but it may be easier to just send them.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -197,29 +214,57 @@ impl BlockBuilder {
...
 
     #[instrument(target = COMPONENT, name = "block_builder.prove_block", skip_all, err)]
     async fn prove_block(
         &self,
-        proposed_block: ProposedBlock,
+        tx_batches: OrderedBatches,
+        block_header: BlockHeader,
+        block_body: BlockBody,
+        block_inputs: BlockInputs,
     ) -> Result<ProvenBlock, BuildBlockError> {
```
bobbinth: As mentioned in the other comment, I think we should be returning `BlockProof` from here rather than the full proven block.

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: > Once we move the proving to the store, the last 3 steps will need to happen there. For this, instead of sending a proven block to the store, we'll need to send (block_header, block_body, tx_batches, block_inputs) to the store (probably organized into a struct). We could also skip sending block_inputs because the store should theoretically be able to build them - but it may be easier to just send them.

Just to ensure I/we are on the same page. This is intended for a separate PR/future work correct? Because that would also entail changes on gRPC and client etc to allow for validated but as yet unproven blocks.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -197,31 +214,68 @@ impl BlockBuilder {
...
+        proposed_block: ProposedBlock,
+        block_inputs: BlockInputs,
+    ) -> Result<(OrderedBatches, BlockInputs, BlockHeader, BlockBody), BuildBlockError> {
+        let response = self
+            .validator
+            .validate_block(proposed_block.clone())
+            .await
+            .map_err(BuildBlockError::ValidateBlockFailed)?;
+
+        // TODO: Check that the returned header and body match the proposed block.
```
sergerad, line 239: @bobbinth what checks should we perform here?

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -197,31 +214,68 @@ impl BlockBuilder {
...
 
+    #[instrument(target = COMPONENT, name = "block_builder.construct_proven_block", skip_all, err)]
+    async fn construct_proven_block(
+        &self,
+        ordered_batches: OrderedBatches,
+        header: BlockHeader,
+        body: BlockBody,
+        block_proof: BlockProof,
+    ) -> Result<ProvenBlock, BuildBlockError> {
+        let proven_block = ProvenBlock::new_unchecked(header, body, block_proof);
```
Mirko-von-Leipzig, line 271: Could you add a `// SAFETY` comment explaining why whatever invariants `new_unchecked` requires are upheld?

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -388,15 +442,55 @@ impl BlockProver {
...
                 .map_err(BuildBlockError::RemoteProverClientError),
         }
     }
 }
+
+/// Validates that the proven block's transaction headers are consistent with the transactions
+/// passed in the proposed block.
+///
+/// This expects that transactions from the proposed block and proven block are in the same
+/// order, as define by [`OrderedTransactionHeaders`].
```
Mirko-von-Leipzig, line 467:
```suggestion
/// order, as defined by [`OrderedTransactionHeaders`].
```

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -388,15 +442,55 @@ impl BlockProver {
...
+    proven_block: &ProvenBlock,
+    proposed_txs: &OrderedTransactionHeaders,
+) -> Result<(), BuildBlockError> {
+    if proposed_txs.as_slice().len() != proven_block.body().transactions().as_slice().len() {
+        return Err(BuildBlockError::other(format!(
+            "remote prover returned {} transaction headers but {} transactions were passed as part of the proposed block",
+            proven_block.body().transactions().as_slice().len(),
+            proposed_txs.as_slice().len()
+        )));
+    }
```
Mirko-von-Leipzig, line 478: fyi we could also use [`itertools::equal`](https://docs.rs/itertools/latest/itertools/fn.equal.html) to compare the iterators including length.

Up to you; your impl does give more info

crates/block-producer/src/validator/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        // Send request and receive response.
+        let message = proto::blockchain::ProposedBlock {
+            proposed_block: proposed_block.to_bytes(),
+        };
+        let request = tonic::Request::new(message);
+        let response = self.client.clone().validate_block(request).await?;
+        let response = response.into_inner();
+
+        // Extract header from response (should always be present).
+        let header_proto = response.header.expect("validator always returns a header");
```
Mirko-von-Leipzig, line 74: I wouldn't rely on this expectation, this should be a deserialization error.

In the proto crate we use https://github.com/0xMiden/miden-node/blob/8d76576ef63b1e06402b0c729e836a149507c456/crates/proto/src/errors/mod.rs#L64-L75

which gets uses something like https://github.com/0xMiden/miden-node/blob/8d76576ef63b1e06402b0c729e836a149507c456/crates/proto/src/domain/account.rs#L121-L123

Having the client here is fine for now but I would make these errors instead (and same below).

crates/block-producer/src/validator/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+#[derive(Debug, Clone)]
+pub struct ValidateBlockResponse {
+    pub header: BlockHeader,
+    pub body: BlockBody,
+}
+
+// VALIDATOR CLIENT
+// ================================================================================================
+
+/// Interface to the validator's block-producer gRPC API.
```
Mirko-von-Leipzig, line 36:
```suggestion
/// Interface to the validator's gRPC API.
```

Mirko-von-Leipzig (approved):

proto/proto/validator.proto
```diff
@@ -16,6 +16,9 @@ service Api {
 
    // Submits a transaction to the validator.
    rpc SubmitProvenTransaction(transaction.ProvenTransaction) returns (google.protobuf.Empty) {}
+
+   // Validates a proposed block and returns the block header and body.
+   rpc ValidateBlock(blockchain.ProposedBlock) returns (ValidateBlockResponse) {}
```
bobbinth, line 21: Not for this PR, but I wonder if the naming here can be improved a bit:

- Technically, we are not so much validating the block as constructing and signing it - so, maybe something like `BuildBlock` or `SignBlock` would be better?
- Also, maybe the response should be `SignedBlock`? Technically, we are returning a fully constructed (and in the future) signed block from here.

I don't love the suggestions above either - maybe there are some better options.

bin/node/src/commands/block_producer.rs
bobbinth, line 25: Question not related to this PR: are the valid values for this parameter only `localhost:[port]` and `0.0.0.0:[port]` - or could we pass any URL into this?

bin/remote-prover/src/api/prover.rs
```diff
@@ -165,24 +165,24 @@ impl ProverRpcApi {
...
             return Err(Status::unimplemented("Block prover is not enabled"));
         };
-
-        let proven_block = prover
+        let BlockProofRequest { tx_batches, block_header, block_inputs } = proof_request;
+        let block_proof = prover
             .try_lock()
             .map_err(|_| Status::resource_exhausted("Server is busy handling another request"))?
-            .prove(proposed_block)
+            .prove(tx_batches, block_header.clone(), block_inputs)
```
bobbinth, line 178: nit: we clone the `block_header` here so that we compute `block_id` on line 182 below, right? If so, we could compute `block_id` after line 174 above and avoid the clone.

crates/block-producer/src/lib.rs
```diff
@@ -9,6 +9,7 @@ mod block_builder;
 mod domain;
 mod mempool;
 pub mod store;
+pub mod validator;
```
bobbinth, line 12: Question: why do these need to be public?

crates/proto/src/domain/proof_request.rs
```diff
@@ -0,0 +1,39 @@
...
+    Deserializable,
+    DeserializationError,
+    Serializable,
+};
+
+pub struct BlockProofRequest {
+    pub tx_batches: OrderedBatches,
+    pub block_header: BlockHeader,
+    pub block_inputs: BlockInputs,
+}
```
bobbinth, line 18: Not for this PR, but it would be great to add doc comments to this struct explaining the purpose of its struct and its fields. For example, I think the block header is included here mostly because of the signature as everything else in the header could be derived from `tx_batches` and `block_inputs`.

Also, I wonder if eventually this struct (potentially named differently) should live in `miden-base`.

crates/store/src/state.rs
```diff
@@ -318,12 +324,13 @@ impl State {
         };
 
         // build note tree
-        let note_tree = block.build_output_note_tree();
+        let note_tree = block.body().compute_block_note_tree();
```
bobbinth, line 327: nit (and not for this PR): `build_output_note_tree()` was probably a better name than `compute_block_note_tree()`. Let's remember to update this when we work on follow-up changes in `miden-base`.

crates/block-producer/src/block_builder/mod.rs
bobbinth: Not for this PR, but I think here we'd want to check two things:

- That the `block_header` returned from the validator is consistent with the `proposed_block`.
- That the `block_body` returned from the validator is consistent with the `block_header` and `proposed_block`.

This is protection against the validator becoming malicious/inconsistent and returning an invalid block which will fail to prove later on.

The way to do this is probably by converting `proposed_block` into `block_header` and `block_body` in this function, and checking these against the data we get from the validator (excluding the signature). Since this is a non-negligible operation in terms of performance (would be good to benchmark how long it actually would take at some point), we should do this in parallel with the request to the validator. Meaning:

- We make the request to the validator but don't await on it.
- We build the block header and body locally.
- We await on the validator request - and once it comes back do the comparison.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -197,31 +214,68 @@ impl BlockBuilder {
...
+        let proven_block = ProvenBlock::new_unchecked(header, body, block_proof);
         if proven_block.proof_security_level() < MIN_PROOF_SECURITY_LEVEL {
             return Err(BuildBlockError::SecurityLevelTooLow(
                 proven_block.proof_security_level(),
                 MIN_PROOF_SECURITY_LEVEL,
             ));
         }
-
-        self.simulate_proving().await;
+        validate_tx_headers(&proven_block, &ordered_batches.to_transactions())?;
```
bobbinth, line 278: I'm not sure this is actually needed here - but maybe for now it is fine. Once we have the actual block proving, this would be replaced with verifying the proof returned from the prover against the block header. Maybe worth adding a comment here about this.

bobbinth (approved):
Looks good! Thank you! I left some comments inline, but we can probably merge as is and address the comments in follow-up PRs.

Issue #1349, opened 2025-11-10
Tracking issue for account retrieval refactoring


drahnr:
The full path for enablement of historial account queries

1. Enable historical account queries
2. Merge get account proof and get account details queries
3. Enable RocksDB backend
4. Migrate nullifier tree to LargeSmt + RocksDB
5. Integrate SmtForest once it has support LargeSmt
6. Add cleanup routine for outdated historical storage entries

bobbinth:
Closing this tracking issue as the important parts have been merged already, and the cleanup parts will be worked on separately.

PR #1385, opened 2025-11-27
refactor: [4/4] unify `get_account_details` and `get_account_proof[s]` into `get_account`

drahnr:
## Unify API calls

The new API is flexible enough to address all needs and avoid loading a lot of data from the DB.

### Main changes in:

`crates/proto/src/domain/account.rs`:
  - Renamed domain structs: `AccountProofRequest` → `AccountRequest`, `AccountProofResponse` → `AccountResponse`
  - Dropped `GetAccountDetails*`

### Caveat

Will require client changes!

### Part 2

This is part 1 of two pieces. The second piece is in <insert PR> and does the actual DB changes, as well as populate the `SmtForest`. This is _NOT_ part of this PR. 

### Open Questions

Do we want compatibility on the RPC level to give the client some more migration time, or is the change bounded enough?
---

Part of https://github.com/0xMiden/miden-node/issues/1349
Part 1 of 2 of the remaining pieces in https://github.com/0xMiden/miden-node/issues/1185

### DIFF
--- a/bin/network-monitor/src/counter.rs
+++ b/bin/network-monitor/src/counter.rs
@@ -43,6 +43,7 @@ use tracing::{error, info, instrument, warn};
 
 use crate::COMPONENT;
 use crate::config::MonitorConfig;
+use crate::deploy::counter::COUNTER_SLOT_NAME;
 use crate::deploy::{MonitorDataStore, create_genesis_aware_rpc_client, get_counter_library};
 use crate::status::{
     CounterTrackingDetails,
@@ -112,10 +113,14 @@ async fn fetch_counter_value(
         return Ok(None);
     };
 
-    let first_slot = storage_header.slots.first().context("no storage slots found")?;
+    let counter_slot = storage_header
+        .slots
+        .iter()
+        .find(|slot| slot.slot_name == COUNTER_SLOT_NAME.as_str())
+        .context(format!("counter slot '{}' not found", COUNTER_SLOT_NAME.as_str()))?;
 
     // The counter value is stored as a Word, with the actual u64 value in the last element
-    let slot_value: Word = first_slot
+    let slot_value: Word = counter_slot
         .commitment
         .as_ref()
         .context("missing storage slot value")?


crates/store/src/state.rs
bobbinth: I would rename this into something like `get_account_witness()` - I think this is a bit more clear than `get_block_witness()` (we could do `get_account_block_witness()` but I think that's probably an overkill).

bobbinth: I think in `next` this is named `fetch_public_account_details()`. We should probably merge the latest `next` into this branch to resolve the inconsistencies.

bobbinth: I think this comment still needs to be addressed. But also, maybe this is not ready for review yet?

bin/network-monitor/src/counter.rs
```diff
@@ -94,43 +93,109 @@ async fn fetch_counter_value(
...
+        // For a value slot, the Word directly contains the value
+        let word: Word = (*commitment)
+            .try_into()
+            .map_err(|e| anyhow::anyhow!("Failed to convert commitment to Word: {e}"))?;
+
+        // The counter value is stored in the last element of the Word
+        let value = word
+            .as_elements()
+            .last()
+            .ok_or_else(|| anyhow::anyhow!("Word has no elements"))?
```
Mirko-von-Leipzig: `anyhow::Context` would neaten this up (which also works for `Option::None`), and lowercase messaging


```suggestion
            .context("missing storage details")?;

        let storage_header = storage_details
            .header
            .context("missing storage header")?;

        // The counter value is in the first storage slot (index 0)
        let first_slot = storage_header
            .slots
            .first()
            .context("no storage slots found")?;

        let commitment = first_slot
            .commitment
            .as_ref()
            .context("missing storage slot commitment")?;

        // The commitment is a Word containing the counter value
        // For a value slot, the Word directly contains the value
        let word: Word = (*commitment)
            .try_into()
            .context("failed to convert commitment to Word)?;

        // The counter value is stored in the last element of the Word
        let value = word
            .as_elements()
            .last()
            .context("Word has no elements")?
```

Mirko-von-Leipzig: Is this even possible?

bin/network-monitor/src/counter.rs
```diff
@@ -94,43 +93,109 @@ async fn fetch_counter_value(
...
+                );
+            }
+            // We can't reconstruct the full Account without the actual code, vault, and storage
+            // data The RPC returns only commitments, so we fall back to the file-based account
+            Ok(None)
+        },
+        Err(e) => {
+            warn!("Failed to fetch wallet account from RPC: {:?}", e);
+            Ok(None)
+        },
```
Mirko-von-Leipzig: Doesn't this mean the function does nothing?

bin/network-monitor/src/counter.rs
```diff
@@ -94,43 +93,109 @@ async fn fetch_counter_value(
...
+    match rpc_client.get_account(request).await {
+        Ok(response) => {
+            let resp = response.into_inner();
+            if resp.witness.is_some() {
+                info!(
+                    "Found wallet account on-chain for {}, but cannot reconstruct full account \
+                     from RPC response",
+                    account_id
+                );
+            }
```
Mirko-von-Leipzig: What's the intention for this message?


```suggestion
            if resp.witness.is_some() {
                info!(%account.id, "Wallet found on-chain but cannot reconstruct full account from RPC response");
            }
```

bin/network-monitor/src/counter.rs
```diff
@@ -94,43 +93,109 @@ async fn fetch_counter_value(
...
+            // We can't reconstruct the full Account without the actual code, vault, and storage
+            // data The RPC returns only commitments, so we fall back to the file-based account
+            Ok(None)
+        },
+        Err(e) => {
+            warn!("Failed to fetch wallet account from RPC: {:?}", e);
```
Mirko-von-Leipzig:
```suggestion
            warn!(account.id=%account_id, err=e.as_report(), "Failed to fetch wallet account via RPC");
```

bin/network-monitor/src/counter.rs
Mirko-von-Leipzig, line 1: @SantiagoPittella could you maybe review the changes here to see if the monitoring still makes sense?

bin/network-monitor/src/counter.rs
drahnr: We currently use this function only for the counter, which doesn't require the account details, but only the account header (value entry). So this has no practical limitation. Fixing it properly is done as part of #1394 with the database schema migration which touches the code again.

crates/store/src/state.rs
```diff
@@ -923,32 +923,32 @@ impl State {
...
             Some(self.fetch_public_account_details(account_id, block_num, request).await?)
         } else {
             None
         };
 
-        Ok(AccountProofResponse { block_num, witness, details })
+        Ok(AccountResponse { block_num, witness, details })
     }
 
     /// Gets the block witness (account tree proof) for the specified account
```
igamigo: nit: doc comment is still confusing

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Changes
 
+- [BREAKING] Removed `GetAccountDetails` RPC endpoint. Use `GetAccountProof` instead ([#1185](https://github.com/0xMiden/miden-node/issues/1185)).
```
igamigo: nit: Should be 

```suggestion
- [BREAKING] Removed `GetAccountDetails` RPC endpoint. Use `GetAccount` instead ([#1185](https://github.com/0xMiden/miden-node/issues/1185)).
```

instead, right?

proto/proto/store/rpc.proto
igamigo, line 1: With most of these changes, docs will have to be updated (`docs/external/src/rpc.md`, `crates/rpc/README.md`, etc)

bin/network-monitor/src/counter.rs
```diff
@@ -81,43 +80,98 @@ async fn fetch_counter_value(
...
+
+        // The counter value is in the first storage slot (index 0)
+        let first_slot = storage_header.slots.first().context("no storage slots found")?;
+
+        let commitment =
+            first_slot.commitment.as_ref().context("missing storage slot commitment")?;
+
+        // The commitment is a Word containing the counter value
+        // For a value slot, the Word directly contains the value
+        let word: Word = (*commitment).try_into().context("failed to convert commitment to word")?;
```
igamigo: nit: Conceptually this slot is a value slot (ie, not a storage map slot) so this is not a commitment but rather just a plain `Word` (I think the protobuf comments might also have this detail)

bin/network-monitor/src/counter.rs
```diff
@@ -81,43 +80,98 @@ async fn fetch_counter_value(
...
+
+        // The counter value is stored in the last element of the Word
+        // A Word always has 4 elements, so this is safe
```
igamigo: nit (feel free to disregard): The second line here is a bit redundant with the `expect()` message.

bin/network-monitor/src/counter.rs
```diff
@@ -81,43 +80,98 @@ async fn fetch_counter_value(
...
+/// Fetch the wallet account from RPC to get the latest nonce.
+///
+/// Note: Currently returns None because the RPC API returns commitments but not the full
+/// account data (code, vault assets, storage). To properly sync the account, we would need
+/// either:
+/// 1. The RPC to return full account data, or
+/// 2. To reconstruct the account from commitments + cached data
+///
+/// For now, we rely on the file-based account and update only the nonce after successful
+/// transactions.
```
igamigo: Wouldn't the account always be small enough that you can just request data and pretty much the whole thing would be returned so there's no need to do anything outside of one request?

proto/proto/rpc.proto
```diff
@@ -34,7 +34,7 @@ service Api {
     rpc GetAccountDetails(account.AccountId) returns (account.AccountDetails) {}
```
bobbinth: I thought we were removing `GetAccountDetails` as a part of this PR - but maybe I misunderstood?

proto/proto/internal/store.proto
```diff
@@ -35,7 +35,7 @@ service Rpc {
     rpc GetAccountDetails(account.AccountId) returns (account.AccountDetails) {}
 
     // Returns the latest state proof of the specified account.
-    rpc GetAccountProof(rpc.AccountProofRequest) returns (rpc.AccountProofResponse) {}
+    rpc GetAccount(rpc.AccountRequest) returns (rpc.AccountResponse) {}
```
bobbinth, line 35: Same question as above.

proto/proto/types/account.proto
```diff
@@ -38,11 +38,13 @@ message AccountStorageHeader {
         // The type of the storage slot.
         uint32 slot_type = 2;
 
-        // The commitment (Word) for this storage slot.
+        // The data (Word) for this storage slot.
+        // For value slots (slot_type=0), this is the actual value stored in the slot.
+        // For map slots (slot_type=1), this is the root commitment of the storage map.
```
bobbinth: nit: "root commitment" sounds redundant. We could say just "this is the root of the storage map" or "this is a commitment to the storage map" (the former is probably preferred).

crates/store/src/state.rs
bobbinth: Yes, let's update the doc.

crates/store/src/state.rs
```diff
@@ -1075,11 +1075,11 @@ impl State {
     ///
     /// When `block_num` is provided, this method will return the account state at that specific
     /// block using both the historical account tree witness and historical database state.
-    pub async fn get_account_proof(
+    pub async fn get_account(
```
bobbinth, line 957: Let's update the doc comments for this method as well (especially line 1073).

crates/store/src/server/rpc_api.rs
```diff
@@ -334,21 +334,21 @@ impl rpc_server::Rpc for StoreApi {
...
-    ) -> Result<Response<proto::rpc::AccountProofResponse>, Status> {
+        request: Request<proto::rpc::AccountRequest>,
+    ) -> Result<Response<proto::rpc::AccountResponse>, Status> {
         debug!(target: COMPONENT, ?request);
         let request = request.into_inner();
-        let account_proof_request = request.try_into()?;
+        let account_request = request.try_into()?;
 
-        let proof = self.state.get_account_proof(account_proof_request).await?;
+        let proof = self.state.get_account(account_request).await?;
```
bobbinth: nit: it is weird that we call the variable `proof` here. Maybe something like `account_data` would be better?

crates/proto/src/domain/account.rs
```diff
@@ -238,21 +238,21 @@ pub struct StorageMapRequest {
...
     ) -> Result<Self, Self::Error> {
-        let proto::rpc::account_proof_request::account_detail_request::StorageMapDetailRequest {
+        let proto::rpc::account_request::account_detail_request::StorageMapDetailRequest {
             slot_name,
             slot_data,
         } = value;
 
         let slot_name = StorageSlotName::new(slot_name)?;
-        let slot_data = slot_data.ok_or(proto::rpc::account_proof_request::account_detail_request::StorageMapDetailRequest::missing_field(stringify!(slot_data)))?.try_into()?;
+        let slot_data = slot_data.ok_or(proto::rpc::account_request::account_detail_request::StorageMapDetailRequest::missing_field(stringify!(slot_data)))?.try_into()?;
```
bobbinth, line 214: nit: could we improve formatting here by importing `proto::rpc::account_request::account_detail_request::StorageMapDetailRequest` somewhere above?

Also, similar comments are relevant in a few other places in this file.

bin/network-monitor/src/counter.rs
```diff
@@ -90,45 +89,227 @@ async fn fetch_counter_value(
...
+    // Request account details with storage header (but no storage maps needed)
+    let request = miden_node_proto::generated::rpc::AccountRequest {
+        account_id: Some(account_id_proto),
+        block_num: None,
+        details: Some(miden_node_proto::generated::rpc::account_request::AccountDetailRequest {
+            code_commitment: None,
+            asset_vault_commitment: None,
+            storage_maps: vec![],
+        }),
+    };
```
bobbinth: Instead of handling request/response here, could we encapsulate this in a helper function? - something like:
```rust
async fn fetch_account_storage_header(
    rpc_client: &mut RpcClient,
    account_id: AccountId,
) -> Result<AccountStorageHeader> {
    ...
}
```

bin/network-monitor/src/counter.rs
```diff
@@ -90,45 +89,227 @@ async fn fetch_counter_value(
...
+                .into_iter()
+                .filter_map(|e| {
+                    let key: Word = e.key?.try_into().ok()?;
+                    let value: Word = e.value?.try_into().ok()?;
+                    Some((key, value))
+                })
+                .collect();
+            Some((map_detail.slot_name, converted))
+        })
+        .collect();
```
bobbinth: AFIAK, wallet accounts have no storage maps - do we really need this code?

bin/network-monitor/src/counter.rs
```diff
@@ -90,45 +89,227 @@ async fn fetch_counter_value(
...
+            if entries.is_empty() {
+                warn!(
+                    account.id = %account_id,
+                    slot_name = %slot.slot_name,
+                    "storage map slot has no entries (not requested or empty)"
+                );
+            }
+            let storage_map = miden_objects::account::StorageMap::with_entries(entries)
+                .context("failed to create storage map")?;
+            StorageSlot::with_map(slot_name, storage_map)
```
bobbinth: Similar to the above comment: I think we can probably just error out if there are any storage maps in the account.

bobbinth (commented):
Looks good! Thank you! I left some comments inline - most are pretty minor.

bin/network-monitor/src/counter.rs
drahnr: We could rename the function and remove the codepath. Given the scope this seems to be the a viable approach as well.

bin/network-monitor/src/counter.rs
drahnr: Yes, indeed.

bin/network-monitor/src/counter.rs
drahnr: Removed

crates/store/src/state.rs
```diff
@@ -940,16 +940,19 @@ impl State {
...
-    /// When `block_num` is provided, this method will return the account state at that specific
-    /// block using both the historical account tree witness and historical database state.
-    pub async fn get_account_proof(
+    /// The witness is a Merkle proof of inclusion in the account tree, proving the account's
+    /// state commitment. If `details` is requested, the method also returns the account's code,
+    /// vault assets, and storage data. Account details are only available for public accounts.
+    ///
+    /// If `block_num` is provided, returns the state at that historical block; otherwise, returns
+    /// the latest state.
+    pub async fn get_account(
```
bobbinth, line 957: I would mention in the doc comments that for historical states we can only return recent states (i.e., states close to the chain tip).

bin/network-monitor/src/counter.rs
bobbinth, line 1: nit: not for this PR, but I'd try to break up this file into a couple of smaller files.

bin/network-monitor/src/counter.rs
drahnr: Nope

bin/network-monitor/src/counter.rs
drahnr: Fixed

bin/network-monitor/src/counter.rs
drahnr: Should be covered now

bin/network-monitor/src/counter.rs
```diff
@@ -84,55 +83,216 @@ async fn get_genesis_block_header(rpc_client: &mut RpcClient) -> Result<BlockHea
...
-    let req = miden_node_proto::generated::account::AccountId { id: id_bytes.to_vec() };
-    let resp = rpc_client.get_account_details(req).await?.into_inner();
-    if let Some(raw) = resp.details {
-        let account = Account::read_from_bytes(&raw)
-            .map_err(|e| anyhow::anyhow!("failed to deserialize account details: {e}"))?;
+    let Some(storage_header) = fetch_account_storage_header(rpc_client, account_id).await? else {
+        return Ok(None);
+    };
+
+    let first_slot = storage_header.slots.first().context("no storage slots found")?;
```
igamigo: If the above returns `AccountStorage`, you can look up the slot by name

igamigo: A good validation here may be to check that the name of the slot is as we expect (otherwise, if the account somehow changes you could be reading any other slot and not really realize)

drahnr: I did a direct lookup of the slot by name

Issue #1232, opened 2025-09-18
Network Monitor: Add Block Producer and NTX Builder data


SantiagoPittella:
As Bobbin mentioned in https://github.com/0xMiden/miden-node/pull/1217#pullrequestreview-3237645621 , tracking additional data related to the block producer and the ntx builder is a nice feature, like the mempool size.

We are not able to make requests to internal components so this issue includes expanding the return types for the internal components.

Mirko-von-Leipzig:
Ah yeah that's a fair point then we just periodically do this update instead of per external request.

PR #1388, opened 2025-11-27
feat: add more data to block-producer status endpoint

SantiagoPittella:
first part of https://github.com/0xMiden/miden-node/issues/1232

Adds mempool data to the block producer endpoint. It works by adding a new field to the `BlockProducerRpcServer` struct, that works as a view of the mempool state updated periodically.

### DIFF
--- a/crates/block-producer/src/mempool/mod.rs
+++ b/crates/block-producer/src/mempool/mod.rs
@@ -69,6 +69,9 @@ mod subscription;
 #[cfg(test)]
 mod tests;
 
+// MEMPOOL CONFIGURATION
+// ================================================================================================
+
 #[derive(Clone)]
 pub struct SharedMempool(Arc<Mutex<Mempool>>);
 
@@ -112,12 +115,18 @@ impl Default for MempoolConfig {
     }
 }
 
+// SHARED MEMPOOL
+// ================================================================================================
+
 impl SharedMempool {
     #[instrument(target = COMPONENT, name = "mempool.lock", skip_all)]
     pub async fn lock(&self) -> MutexGuard<'_, Mempool> {
         self.0.lock().await
     }
 }
+
+// MEMPOOL
+// ================================================================================================
 
 #[derive(Clone, Debug)]
 pub struct Mempool {
@@ -142,6 +151,9 @@ impl PartialEq for Mempool {
 }
 
 impl Mempool {
+    // CONSTRUCTORS
+    // --------------------------------------------------------------------------------------------
+
     /// Creates a new [`SharedMempool`] with the provided configuration.
     pub fn shared(chain_tip: BlockNumber, config: MempoolConfig) -> SharedMempool {
         SharedMempool(Arc::new(Mutex::new(Self::new(chain_tip, config))))
@@ -156,6 +168,9 @@ impl Mempool {
             nodes: nodes::Nodes::default(),
         }
     }
+
+    // TRANSACTION & BATCH LIFECYCLE
+    // --------------------------------------------------------------------------------------------
 
     /// Adds a transaction to the mempool.
     ///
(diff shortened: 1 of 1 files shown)


crates/block-producer/src/server/mod.rs
```diff
@@ -312,10 +340,53 @@ impl tokio_stream::Stream for MempoolEventSubscription {
...
 impl BlockProducerRpcServer {
     pub fn new(mempool: SharedMempool, store: StoreClient) -> Self {
-        Self { mempool: Mutex::new(mempool), store }
+        Self {
+            mempool: Mutex::new(mempool),
+            store,
+            cached_mempool_stats: Arc::new(RwLock::new(MempoolStats {
+                unbatched_transactions: 0,
+                proposed_batches: 0,
+                proven_batches: 0,
```
Mirko-von-Leipzig: Consider `#[derive(Default)]`

crates/block-producer/src/server/mod.rs
```diff
@@ -312,10 +340,53 @@ impl tokio_stream::Stream for MempoolEventSubscription {
...
+            })),
+        }
+    }
+
+    /// Starts a background task that periodically updates the cached mempool statistics.
+    ///
+    /// This prevents the need to lock the mempool for each status request.
+    async fn start_mempool_stats_updater(&self) {
```
Mirko-von-Leipzig: Nit: 
```suggestion
    async fn spawn_mempool_stats_updater(&self) {
```

crates/block-producer/src/lib.rs
```diff
@@ -46,6 +47,9 @@ const SERVER_MEMPOOL_STATE_RETENTION: NonZeroUsize = NonZeroUsize::new(5).unwrap
 /// This rejects transactions which would likely expire before making it into a block.
 const SERVER_MEMPOOL_EXPIRATION_SLACK: u32 = 2;
 
+/// The interval at which to update the cached mempool statistics.
+const CACHED_MEMPOOL_STATS_UPDATE_INTERVAL: Duration = Duration::from_secs(1);
```
Mirko-von-Leipzig: Let's make this a bit slower imo; maybe `5s`?

crates/rpc/src/server/api.rs
```diff
@@ -567,6 +567,11 @@ impl api_server::Api for RpcService {
                 proto::block_producer::BlockProducerStatus {
                     status: "unreachable".to_string(),
                     version: "-".to_string(),
+                    mempool_stats: Some(proto::block_producer::MempoolStats {
+                        unbatched_transactions: 0,
+                        proposed_batches: 0,
+                        proven_batches: 0,
+                    }),
```
Mirko-von-Leipzig: Could also use default here

proto/proto/block_producer.proto
bobbinth, line 50: Not for this PR, but it may be good to add something like `chain_tip` to this message so that we can see in the monitor what the block producers thinks is the current chain tip (in case it somehow diverges from the store's chain tip).

crates/block-producer/src/mempool/mod.rs
```diff
@@ -553,6 +553,21 @@ impl Mempool {
         self.subscription.subscribe(chain_tip)
     }
 
+    /// Returns the number of transactions currently waiting to be batched.
+    pub fn unbatched_transactions_count(&self) -> usize {
```
bobbinth, line 578: Maybe not for this PR, but I'd add some section separators to this file to make the code a bit easier to follow (e.g., currently, we have these accessors "interspersed" between state mutator methods).

bobbinth (approved):
Looks great! Thank you! I left a couple of comments for the future inline - but this is good to merge.

crates/block-producer/src/mempool/mod.rs
SantiagoPittella: Done!

proto/proto/block_producer.proto
SantiagoPittella: I opened https://github.com/0xMiden/miden-node/issues/1408 to track this.

PR #1392, opened 2025-11-27
feat: display mempool data in network monitor

SantiagoPittella:
closes https://github.com/0xMiden/miden-node/issues/1232

Displays the following mempool data on the network transaction frontend:
- unbatched_transactions
- proposed_batches
- proven_batches

This is the look of the network monitor with this new fields:
Screenshot 2025-11-27 at 18 31 34

Also, I needed to add the genesis commitment in RPC client in order to be able to submit transactions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - [BREAKING] `SubmitProvenTransaction` now **requires** that the network's genesis commitment is set in the request's `ACCEPT` header ([#1298](https://github.com/0xMiden/miden-node/pull/1298)).
 - Added support for timeouts in the WASM remote prover clients ([#1383](https://github.com/0xMiden/miden-node/pull/1383)).
 - Added support for caching mempool statistics in the block producer server ([#1388](https://github.com/0xMiden/miden-node/pull/1388)).
+- Added mempool statistics to the block producer status in the `miden-network-monitor` binary ([#1392](https://github.com/0xMiden/miden-node/pull/1392)).
 
 ### Fixes
 


bin/network-monitor/src/status.rs
```diff
@@ -166,9 +179,19 @@ impl From<StoreStatus> for StoreStatusDetails {
 
 impl From<BlockProducerStatus> for BlockProducerStatusDetails {
     fn from(value: BlockProducerStatus) -> Self {
+        // We assume all supported nodes expose mempool statistics.
+        let mempool_stats = value
+            .mempool_stats
+            .expect("block producer status must include mempool statistics");
```
Mirko-von-Leipzig, line 185: This means that if we deploy this monitor version, that we must also bump the deployed node version correct?

(this is fine, just something we need to be aware of)

bin/network-monitor/src/deploy/mod.rs
```diff
@@ -40,6 +40,62 @@ use crate::deploy::wallet::{create_wallet_account, save_wallet_account};
...
+/// Create an RPC client configured with the correct genesis metadata in the
+/// `Accept` header so that write RPCs such as `SubmitProvenTransaction` are
+/// accepted by the node.
+pub async fn create_genesis_aware_rpc_client(
+    rpc_url: &Url,
+    timeout: Duration,
+) -> Result<RpcClient> {
+    // First, create a temporary client without genesis metadata to discover the
+    // genesis block header and its commitment.
+    let mut rpc: RpcClient = Builder::new(rpc_url.clone())
```
Mirko-von-Leipzig, line 52: Ohh. That's a consequence that I hadn't really considered. Nice.

Mirko-von-Leipzig (approved):

bin/network-monitor/src/status.rs
SantiagoPittella: Yes. If you prefer a kind of "backward compatibility" I can make all of this optional.

Mirko-von-Leipzig: Nah its fine; just wanted to make sure :)

Issue #1376, opened 2025-11-21
Github actions cache


Mirko-von-Leipzig:
Our github cache has grown too large because we're caching and checking things in CI that we shouldn't be.

```
v0-rust-msrv-Linux-x64-3bcebdd4-89a2857b              2.8 GB cached 19 hours ago
v0-rust-doc-Linux-x64-948915f2-89a2857b               2.1 GB cached 19 hours ago
v0-rust-stress-test-check-Linux-x64-948915f2-89a2857b 1.8 GB cached 19 hours ago 
v0-rust-proto-Linux-x64-948915f2-89a2857b            1000 MB cached 19 hours ago
v0-rust-test-Linux-arm64-1b94ad1b-89a2857b            930 MB cached 19 hours ago
v0-rust-clippy-Linux-x64-948915f2-89a2857b            820 MB cached 19 hours ago
v0-rust-rustfmt-Linux-x64-948915f2-89a2857b           150 MB cached 19 hours ago
v0-rust-check-Linux-x64-948915f2-89a2857b             370 MB cached 19 hours ago
```

Some ideas:

1. I suspect that `docs` can be improved by only building our _own_ docs. Right now its building all of them iiuc.
2. We don't need to run `msrv` checks all the time. Personally I still think its a waste of time, but lets only run it once a day against `next` and without a cache.
3. The stress test.. I'm not really sure why its so large? @SantiagoPittella maybe you have some ideas?
4. The proto check is only required because we're keeping generated build artifacts _in source tree_. We really should get rid of this.
5. #1374 - we should try get docker cached locally within github as well. Something like using cargo chef? I'm not good at docker but it should be possible.

SantiagoPittella:
I'm not sure why the stress test is that large, will take a look.

PR #1393, opened 2025-11-27
chore: use share cache in CI

SantiagoPittella:
part of https://github.com/0xMiden/miden-node/issues/1376

This PR adds a share key for the workflows that compiles all the workspace.

@Mirko-von-Leipzig this is pretty much experimental, lmk if you have any thought about this

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -38,6 +38,8 @@ jobs:
           rustup +nightly component add rustfmt
       - uses: Swatinem/rust-cache@v2
         with:
+          # Share Rust build cache across workflows/jobs on the same OS/arch.
+          shared-key: miden-node
           save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
       - name: Fmt
         run: make format-check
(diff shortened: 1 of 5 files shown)


SantiagoPittella:
This new cache should be saved when merging into `next`

.github/workflows/lint.yml
Mirko-von-Leipzig, line 1: I don't think this will work. The different jobs result in different compilations, either due to features or different binaries.

Won't this just clobber the same cache key each time?

Mirko-von-Leipzig: Some jobs/workflows could share but I'm unsure just how shareable they are

drahnr:
Iiuc the shared is only updated from `next`? If so, I think this is a good baseline for any PR. The size question still remains, how long does it take to stream the cache in vs time saved

SantiagoPittella:
Not sure if we want to continue with this, wdyt @Mirko-von-Leipzig ?

I can start trying another approach if we decide to continue.

Mirko-von-Leipzig:
> Not sure if we want to continue with this, wdyt @Mirko-von-Leipzig ?
> 
> I can start trying another approach if we decide to continue.

I think there are probably more things we could do; but I think with the reductions we already got we're okay for now.

PR #1396, opened 2025-11-28
fix(block-producer): pass through transaction crash

### LINKED ISSUE #1380: Passthrough transaction causes panic
Currently (and [as of recently](https://github.com/0xMiden/miden-base/pull/1982)), in `miden-base`, authentication account components are expected to increase the account's nonce if the nonce is 0 or if the account state has changed (see also https://github.com/0xMiden/miden-base/pull/2060/ for how this was implemented in a couple of the auth components).

In updating https://github.com/0xMiden/miden-client/pull/1471 to test passthrough transactions (ie, txs that do not update the account's state but do output nullifiers or notes), I found that creating a transaction that "deploys" the account (ie, nonce 0 -> nonce 1) and soon after, run a passthrough transaction, the block producer panics:

```
2025-11-25T21:10:07.601927+00:00     ERROR          ┕━ 🚨 [error]: panicked at miden-node-block-producer-0.12.3/src/mempool/state.rs:226:9:
assertion `left == right` failed: Account updates should be removed as a pair with the same node ID
  left: Block(BlockNumber(142))
 right: Transaction(0xe3273b9deb326136fd651036f9bc3584ea6f00439783ffdc001982af9da0fe4c) | panic: true
2025-11-25T20:57:57.600061+00:00     INFO     rpc [ 2.36ms | 99.29% / 100.00% ] rpc.service: "block_producer.Api" | rpc.method: "MempoolSubscription"
2025-11-25T20:57:57.600096+00:00     INFO     ┝━ mempool.lock [ 8.17µs | 0.35% ]
2025-11-25T20:57:57.600129+00:00     INFO     ┕━ mempool.subscribe [ 8.58µs | 0.36% ]
Error: Component block-producer failed

Caused by:
    0: failed while serving block-producer component
    1: error joining block-builder task
    2: task 72 panicked with message "assertion `left == right` failed: Account updates should be removed as a pair with the same node ID\n  left: Block(BlockNumber(142))\n right: Transaction(0xe3273b9deb326136fd651036f9bc3584ea6f00439783ffdc001982af9da0fe4c)"
```

In the client we see

```
Failed tests:
  - test_pass_through (pass_through)
    Error: rpc api error
  Caused by: grpc request failed for submit_proven_transaction: unknown error: grpc-status header missing, mapped from HTTP status code 500
  Caused by: status: 'Unknown error', self: "grpc-status header missing, mapped from HTTP status code 500", metadata: {"content-type": "application/grpc", "date": "Tue, 25 Nov 2025 21:09:43 GMT", "access-control-expose-headers": "grpc-status,grpc-message,grpc-status-details-bin", "access-control-allow-credentials": "true", "vary": "origin, access-control-request-method, access-control-request-headers"}
```

I say "soon after" because the problem seems to happen when cleaning up data from committed blocks. If we wait for an amount of blocks greater than `SERVER_MEMPOOL_STATE_RETENTION` (I thinnk it defaults to 5) between transactions, the test passes. 

To expand a bit:
  - The mempool records account updates in `AccountUpdates` keyed only by `from`/`to` with a single `NodeId` each. See [`AccountUpdates::remove`](https://github.com/0xMiden/miden-node/blob/next/crates/block-producer/src/mempool/state.rs#L217).
  - Deploy tx: inserts commitments (say, `(from=C0, to=C1)`) under `NodeId::Block(..)` when it's added into a block.
  - Passthrough tx: its account update's  `from` and `to` commitments are both `C1`, so `insert` overwrites the `to` entry for commitment `C1` with `NodeId::Transaction(..)` while the `from` entry for `C1` still belongs to the block.
  - Later, when the retained block is popped (`commit_block` -> retention cleanup), `remove` looks up both sides: `from`  resolves to the block, `to` resolves to the tx, so the  “same node ID” assert in `AccountUpdates::remove` fires.

Because as far as the protocol goes, it's fine that transactions do not change an account's commitment, I believe some alternatives to fix this could be not to track updates if commitments do not get updated, or key updates differently, so as to not override previous entries; but not entirely sure about whether this breaks some other invariants.

Mirko-von-Leipzig:
### This PR targets `main`

Fixes #1380 

## Crash cause

This PR addresses a crash in the block-producer caused by not handling pass through transactions. These are transactions which don't update an account's state i.e. the account's initial and final commitment remains the same.

The mempool was unaware of such types transactions, and instead implicitly assumed that every account commitment could at most have one transaction having it as the initial commitment, and one as the final commitment. This invariant was encoded as a mapping of initial/final commitment to tx ID (well node ID).

With pass through transactions this is of course false, and the mappings got overriden. This is turn caused a crash when internal book keeping assertions tripped on this.

## The fix

The solution used here is to keep pass through nodes slightly separate in the account state tracker. This let's us keep the existing code mostly as is. Such nodes are considered siblings of one another and therefore don't have parent/child relations to each other. This makes sense and also let's us re-order them as we wish.

This is much better than my original plan of enforcing keeping the same original sequence.

---------------

Some non-obvious things

- pass through transactions can in turn cause pass through batches and blocks if those are all that is included
- pass through nodes should be treated as siblings and not parent/children within the graph
  - this is achieved by identifying when a parent/child check is done on a pass through node, and then ignoring the siblings

### DIFF
--- a/crates/block-producer/src/domain/batch.rs
+++ b/crates/block-producer/src/domain/batch.rs
@@ -42,6 +42,11 @@ impl SelectedBatch {
         self.txs
     }
 
+    /// The aggregated list of account transitions this batch causes given as tuples of `(AccountId,
+    /// initial commitment, final commitment)`.
+    ///
+    /// Note that the updates are aggregated, i.e. only a single update per account is possible, and
+    /// transaction updates to an account of `a -> b -> c` will result in a single `a -> c`.
     pub(crate) fn account_updates(&self) -> impl Iterator<Item = (AccountId, Word, Word)> {
         self.account_updates.iter().map(|(account, (from, to))| (*account, *from, *to))
     }
--- a/crates/block-producer/src/mempool/state.rs
+++ b/crates/block-producer/src/mempool/state.rs
@@ -199,6 +199,26 @@ impl InflightState {
 struct AccountUpdates {
     from: HashMap<Word, NodeId>,
     to: HashMap<Word, NodeId>,
+    /// This holds updates from nodes where the initial commitment is the same as the final
+    /// commitment aka no actual change was made to the account.
+    ///
+    /// This sounds counter-intuitive, but is caused by so-called pass-through transactions which
+    /// use an account at some state `A` but only consume and emit notes without changing the
+    /// account state itself.
+    ///
+    /// These still need to be tracked as part of account updates since they require that an
+    /// account is in the given state. Since we want these node's to be processed before the
+    /// account state is changed, this implies that they must be considered children of the
+    /// non-pass-through node that created the state. Similarly, they must be considered
+    /// parents of any non-pass-through node that changes to another state as otherwise this
+    /// node might be processed before the pass-through nodes are.
+    ///
+    /// Pass-through nodes with the same state are considered siblings of each as they don't
+    /// actually depend on each other, and may be processed in any order.
+    ///
+    /// Note also that its possible for any node's updates to an account to solely consist of
+    /// pass-through transactions and therefore in turn is a pass-through node from the perspective
+    /// of that account.
     pass_through: HashMap<Word, HashSet<NodeId>>,
 }
 
@@ -273,6 +293,12 @@ impl AccountUpdates {
     /// parent-child relationship where `parent.to == child.from`.
     fn parents(&self, from: Word, to: Word) -> impl Iterator<Item = &NodeId> {
         let direct_parent = self.to.get(&from).into_iter();
+
+        // If the node query isn't for a pass-through node, then it must also consider pass-through
+        // nodes at its `from` commitment as parents.
+        //
+        // This means the query node depends on the pass-through nodes since these must be processed
+        // before the account commitment may change.
         let pass_through_parents = (from != to)
             .then(|| self.pass_through.get(&from).map(HashSet::iter))
             .flatten()
@@ -291,6 +317,12 @@ impl AccountUpdates {
     /// parent-child relationship where `parent.to == child.from`.
     fn children(&self, from: Word, to: Word) -> impl Iterator<Item = &NodeId> {
         let direct_child = self.from.get(&to).into_iter();
+
+        // If the node query isn't for a pass-through node, then it must also consider pass-through
+        // nodes at its `to` commitment as children.
+        //
+        // This means the pass-through nodes depend on the query node since it changes the account
+        // commitment to the state required by the pass-through nodes.
         let pass_through_children = (from != to)
             .then(|| self.pass_through.get(&to).map(HashSet::iter))
             .flatten()


crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig, line 1: I suspect I should add some documentation on why/how this solves the problem here; I'll attempt that next week, I'm a bit deep in it now to describe it clearly.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -239,13 +239,13 @@ impl Mempool {
         //
         // The additional bookkeeping can be implemented once we have fee related strategies. KISS.
 
-        let mut selected = ProposedBatchNode::default();
+        let mut selected = Vec::<Arc<AuthenticatedTransaction>>::default();
```
bobbinth: nit (and maybe not for this PR): I wonder if creating something like `SelectedBatch` newtype would be a bit clearer. It would allow us to still use `selected.contains()` etc. methods.

Mirko-von-Leipzig:
> > @igamigo I ran the client integration tests from the branch as you suggested, and they pass but print some warnings?
> 
> The warnings are output from the assembler when assembling the transaction or note script AFAIK (eg non-used imports). I'll try to update the original test to avoid them, but overall it should not make a difference

No problem; just wanted to check that they weren't something I needed to address

crates/block-producer/src/mempool/state.rs
```diff
@@ -206,53 +209,93 @@ impl AccountUpdates {
...
+            );
+        } else {
+            assert!(
+                self.from.insert(from, id).is_none(),
+                "Account already contained the commitment {from} when inserting {id:?}"
+            );
+            assert!(
+                self.to.insert(to, id).is_none(),
+                "Account already contained the commitment {to} when inserting {id:?}"
+            );
```
bobbinth, line 282: Not from this PR, but should we not check if the set of transitions is contiguous? For example, if we are adding transition `B -> C`, this should be allowed only if:

- This is the first transition (i.e., there are no other transitions in the `AccountUpdates`).
- There is already some transition from `A -> B`.

Or is this not correct thinking and we do allow non-contiguous transitions In this struct?

crates/block-producer/src/mempool/state.rs
```diff
@@ -206,53 +209,93 @@ impl AccountUpdates {
...
+                .expect("should only be removing account updates from nodes that are present");
+            let to_removed = self
+                .to
+                .remove(&to)
+                .expect("should only be removing account updates from nodes that are present");
+            assert_eq!(
+                from_removed, to_removed,
+                "Account updates should be removed as a pair with the same node ID"
+            );
+            assert_eq!(from_removed, id, "Account update removal should match the input node ID",);
```
bobbinth, line 264: Related to the above comment: this doesn't seem to enforce that state transitions remain contiguous, right?

For example, if we have transitions `A -> B -> C -> D`, this would allow us to remove `B->C` leaving two non-contiguous (`A -> B` and `C -> D`) in the map. Would that not cause issues somehow?

crates/block-producer/src/mempool/state.rs
```diff
@@ -197,6 +199,7 @@ impl InflightState {
 struct AccountUpdates {
     from: HashMap<Word, NodeId>,
     to: HashMap<Word, NodeId>,
+    pass_through: HashMap<Word, HashSet<NodeId>>,
 }
```
bobbinth, line 223: Not for this PR (and not related to the current changes), but I do find this structure a bit difficult to reason about. My main issue is that the structure does not enforce that all updates are contiguous (assuming this is the invariant we want to enforce). At the very least, we should expand the doc comment here to make things a bit more explicit - i.e., do we expect the updates to be contiguous? If yes, how this is enforced? If no, why is this OK?

But also, I do think that potentially a more liner structure could be a bit easier to understand - e.g., something like:
```rust
struct AccountUpdates {
    init_state: Word,
    updates: Vec<(NodeId, Word)>,
    pass_through: HashMap<Word, HashSet<NodeId>>,
}
```

bobbinth (approved):
Looks good! Thank you! I left some comments inline - but they are more related to my general understanding of the mechanism. All by the ones about expanding doc comments should probably be left for future PRs.

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: We do _temporarily_ allow non-contiguous transitions. In a practical sense, this occurs whenever nodes are aggregated into another node e.g. txs into a batch, or batches into a block. The way this is currently implemented is that 

- nodes are removed one-by-one,
- we are now in an invalid state,
- new batch node is inserted,
- we should now be back to valid given that the batch is an aggregation of its txs

However this is never asserting after this operation and relies on the calling function to ensure that (a) its aggregation is legal, and (b) that it completes everything properly.

This also applies in situations where we revert a batch(s) and re-insert the transactions. This isn't always a straight-forward replacement, because when a large reversion happens, some transactions within a reverted batch may be invalid and discarded.

I have tried doing something more rigorous using types to enforce that consistency is checked again at the end, but its not trivial e.g. by verifying the integrity of all accounts and notes/nullifiers that were touched.

I would like to get to something like that. The current impl is nice and flexible but does rely on the upper levels being smart.

Mirko-von-Leipzig: There is definitely a smarter way to write all of this, but I haven't quite figured out how just yet.

I've tried the linear approach you describe below, but you often end up operating in the middle of the vector, where you're folding or unfolding nodes into other nodes so it ends up being a bit painful.

If you want to enforce that things _must_ always remain contiguous at all times, you'll need something like

```rust
/// Adds a new transition.
fn push(&mut self, node: NodeId, from: Word, to: Word) { ... }

/// Replaces `target` node with a contiguous series of updates
fn unfold(&mut self, target: &NodeId, to: &[(NodeId, Word)]) {...}

/// Replaces a contiguous range of updates with a single new NodeId
fn fold(&mut self, range: Range<NodeId>, to: NodeId) { ... }

// Similar methods to handle pass through txs
```

and then you'll need smarter top level code to figure out what it should be doing.

crates/block-producer/src/mempool/state.rs
```diff
@@ -206,53 +209,93 @@ impl AccountUpdates {
...
-        );
-        assert!(
-            self.to.insert(to, id).is_none(),
-            "Account already contained the commitment {to} when inserting {id:?}"
-        );
+        if from == to {
+            assert!(
+                self.pass_through.entry(from).or_default().insert(id),
+                "Account already contained the pass through commitment {from} for node {id:?}"
+            );
```
igamigo, line 273: Does this mean the node would now panic in a similar scenario to the one before, only now it would require two passthrough transactions in the same batch? Since they will be seen as independent (afaik) two transactions they can end up in the same batch and the updates could trigger this panic

crates/block-producer/src/mempool/state.rs
igamigo: Maybe until this API rewrite there could be an intermediate `InflightState::assert_contiguous()` called in between selections/commits/rollbacks to ensure state consistency. Unless this already implicitly happens, doesn't sound like it would be too complicated to write and run compared to the value of catching a bug or two (but maybe it is and this is not worth it).

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: Ah. You're right, I think I assumed that account updates for a batch would get merged. This should be true for `ProvenBatch`, but my `ProposedBatch` impl is lazy and doesn't aggregate them.

Thanks, I'll probably fix the proposed batch so that it does aggregation.

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: I tried implementing something like that using a mutation builder of sorts, which at the end checked that it results in a consistent state but it got a bit awkward with multiple borrows.

Getting that right would also solve the issues in total; maybe I should try it again.

bobbinth: > If you want to enforce that things _must_ always remain contiguous at all times, you'll need something like
> 
> ```rust
> /// Adds a new transition.
> fn push(&mut self, node: NodeId, from: Word, to: Word) { ... }
> 
> /// Replaces `target` node with a contiguous series of updates
> fn unfold(&mut self, target: &NodeId, to: &[(NodeId, Word)]) {...}
> 
> /// Replaces a contiguous range of updates with a single new NodeId
> fn fold(&mut self, range: Range<NodeId>, to: NodeId) { ... }
> ```

I really like this API - let's create an issue so that we can come back to it in the future.

crates/block-producer/src/mempool/state.rs
```diff
@@ -206,53 +209,93 @@ impl AccountUpdates {
...
-        assert!(
-            self.from.insert(from, id).is_none(),
-            "Account already contained the commitment {from} when inserting {id:?}"
-        );
-        assert!(
-            self.to.insert(to, id).is_none(),
-            "Account already contained the commitment {to} when inserting {id:?}"
-        );
+        if from == to {
+            assert!(
```
drahnr, line 270: JFYI: yes, they are active in debug _and_ release builds - https://doc.rust-lang.org/std/macro.assert.html#uses
I have to look up the pattern `assert!($expr, text)` every time, it's such a footgun in C++ and I still got PTSD from that 😰

crates/block-producer/src/mempool/state.rs
```diff
@@ -206,53 +209,93 @@ impl AccountUpdates {
...
 
-    /// Returns the node ID that updated this account's commitment to the given value.
+    /// Returns the node IDs that updated this account's commitment to the given value.
     ///
-    /// In other words, this returns the ID of `node` where `node.to == from`. This infers the
+    /// Note that this might be multiple IDs due to pass through transactions. When the input
+    /// is itself a pass through transaction (`from == to`), then its sibling pass through
+    /// transactions are not considered parents as they are siblings.
+    ///
+    /// In other words, this returns the IDs of `node` where `node.to == from`. This infers the
```
drahnr, line 292: nit: This is a tough one to memorize what it yields exactly, but to no effort of trying I couldn't rephrase it more concisely

crates/block-producer/src/mempool/state.rs
```diff
@@ -145,8 +147,8 @@ impl InflightState {
 
         let account_parents = node
             .account_updates()
-            .filter_map(|(account, from, _to)| {
-                self.accounts.get(&account).map(|account| account.parent(&from))
+            .filter_map(|(account, from, to)| {
+                self.accounts.get(&account).map(|account| account.parents(from, to))
```
drahnr, line 151: Iiuc, the key change is this one, avoiding sibling nodes?

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: Agreed. I was hoping siblings would be enough but coming back to it now its still looks like French to me.

crates/block-producer/src/domain/mod.rs
```diff
@@ -1 +1,2 @@
-pub mod transaction;
+pub(crate) mod batch;
+pub(crate) mod transaction;
```
igamigo, line 2: nit: I guess this technically makes this PR breaking. Probably zero impact, but just noting it

crates/block-producer/src/domain/batch.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    pub(crate) fn txs(&self) -> &[Arc<AuthenticatedTransaction>] {
+        &self.txs
+    }
+
+    pub(crate) fn into_transactions(self) -> Vec<Arc<AuthenticatedTransaction>> {
+        self.txs
+    }
+
+    pub(crate) fn account_updates(&self) -> impl Iterator<Item = (AccountId, Word, Word)> {
```
igamigo, line 50: nit: I think it's fair to assume that anyone reading this can easily infer the return type, but I'd maybe still mention it in a doc comment (or maybe at the level of `SelectedBatch`)

igamigo (approved):
LGTM! Did not see any holes in the newest implementation. I believe tests do indeed cover the discussed scenarios as well.

crates/block-producer/src/domain/mod.rs
Mirko-von-Leipzig: We sort of hand-wavingly define our public API as the node's CLI, the public gRPC schema and any major protocol or behavioral changes.

Since our "internal" node crates are so inter-woven they're constantly making semver breaking changes.

crates/block-producer/src/mempool/state.rs
Mirko-von-Leipzig: #1439

PR #1398, opened 2025-11-28
fix: Make NtxDataStore note script cache global

### LINKED ISSUE #1397: Make ntx script cache global
We have a note script cache in the NtxDataStore impl [here](https://github.com/0xMiden/miden-node/blob/next/crates/ntx-builder/src/transaction.rs#L259).

Right now the NtxDataStore is created for every transaction execution ([here](https://github.com/0xMiden/miden-node/blob/next/crates/ntx-builder/src/transaction.rs#L128)) and so is the script cache. We actually intend for the lifecycle of the script cache to be a superset of every NtxDataStore instance.

The implementation need to be changed so that the cache is instantiated outside of the NtxDataStore and fed into it. The cache should be global across all NtxDataStore instances.

sergerad:
## Context

Closes #1397.

## Changes

- Adds LruCache newtype
- Moves note script cache to be global across NtxDataStore instances

### DIFF
--- a/crates/ntx-builder/src/transaction.rs
+++ b/crates/ntx-builder/src/transaction.rs
@@ -342,12 +342,12 @@ impl DataStore for NtxDataStore {
         async move { Err(DataStoreError::AccountNotFound(foreign_account_id)) }
     }
 
-    fn get_vault_asset_witness(
+    fn get_vault_asset_witnesses(
         &self,
         account_id: AccountId,
         vault_root: Word,
-        vault_key: AssetVaultKey,
-    ) -> impl FutureMaybeSend<Result<AssetWitness, DataStoreError>> {
+        vault_keys: BTreeSet<AssetVaultKey>,
+    ) -> impl FutureMaybeSend<Result<Vec<AssetWitness>, DataStoreError>> {
         async move {
             if self.account.id() != account_id {
                 return Err(DataStoreError::AccountNotFound(account_id));
@@ -360,12 +360,14 @@ impl DataStore for NtxDataStore {
                 });
             }
 
-            AssetWitness::new(self.account.vault().open(vault_key).into()).map_err(|err| {
-                DataStoreError::Other {
-                    error_msg: "failed to open vault asset tree".into(),
-                    source: Some(Box::new(err)),
-                }
-            })
+            Result::<Vec<_>, _>::from_iter(vault_keys.into_iter().map(|vault_key| {
+                AssetWitness::new(self.account.vault().open(vault_key).into()).map_err(|err| {
+                    DataStoreError::Other {
+                        error_msg: "failed to open vault asset tree".into(),
+                        source: Some(Box::new(err)),
+                    }
+                })
+            }))
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,13 +12,26 @@
 - [BREAKING] `SubmitProvenTransaction` now **requires** that the network's genesis commitment is set in the request's `ACCEPT` header ([#1298](https://github.com/0xMiden/miden-node/pull/1298)).
 - Added support for timeouts in the WASM remote prover clients ([#1383](https://github.com/0xMiden/miden-node/pull/1383)).
 - Added block validation endpoint to validator and integrated with block producer ([#1382](https://github.com/0xMiden/miden-node/pull/1381)).
+- Added support for caching mempool statistics in the block producer server ([#1388](https://github.com/0xMiden/miden-node/pull/1388)).
+- Added mempool statistics to the block producer status in the `miden-network-monitor` binary ([#1392](https://github.com/0xMiden/miden-node/pull/1392)).
+- Add `S` generic to `NullifierTree` to allow usage with `LargeSmt`s ([#1353](https://github.com/0xMiden/miden-node/issues/1353)).
 
 ### Fixes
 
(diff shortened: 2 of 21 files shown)


crates/ntx-builder/src/transaction.rs
```diff
@@ -70,19 +70,45 @@ type NtxResult<T> = Result<T, NtxError>;
...
+    /// LRU cache for storing retrieved note scripts to avoid repeated store calls.
+    script_cache: Arc<Mutex<LruCache<Word, NoteScript>>>,
 }
 
 impl NtxContext {
+    /// Default cache size for note scripts.
+    ///
+    /// Each cached script contains the deserialized `NoteScript` object, so the actual memory usage
+    /// depends on the complexity of the scripts being cached.
+    const DEFAULT_SCRIPT_CACHE_SIZE: usize = 1000;
```
Mirko-von-Leipzig: I believe you can move the non-zero check here to compile time


```suggestion
    const DEFAULT_SCRIPT_CACHE_SIZE: NonZeroUsize = NonZeroUsize::new(1_000).unwrap();
```

crates/ntx-builder/src/transaction.rs
```diff
@@ -70,19 +70,45 @@ type NtxResult<T> = Result<T, NtxError>;
...
     /// computationally intensive.
-    pub prover: Option<RemoteTransactionProver>,
+    prover: Option<RemoteTransactionProver>,
 
     /// The store client for retrieving note scripts.
-    pub store: StoreClient,
+    store: StoreClient,
+
+    /// LRU cache for storing retrieved note scripts to avoid repeated store calls.
+    script_cache: Arc<Mutex<LruCache<Word, NoteScript>>>,
```
Mirko-von-Leipzig: We should consider creating `struct ScriptCache` because this contains a `tokio::Mutex` which is dangerous if held across an await point. Which can happen by accident rather quickly.

So I'm thinking we move the mutex access inside the struct where we can easily audit and control the lock guard.

e.g.

```rust
#[derive(Clone, Default)]
struct ScriptCache(Arc<Mutex<LruCache<Word, NoteScript>>>);

impl ScriptCache {
    pub fn get(&self, script: Word) -> Option<NoteScript> {
        ...
    }

    pub fn put(&mut self, root: Word, script: NoteScript) {
        ...
    }
}
```

Mirko-von-Leipzig: Although you can probably argue that this is _only_ used by the data store impl and therefore its already quite constrained in scope. Feel free to ignore this suggestion, but then I would recommend we add some warning comments to the data store impl.

bobbinth: I'd be in favor of adding `struct ScriptCache` (or maybe `NoteScriptCache`) newtype as I think it will improve the readability for very little cost.

bobbinth (approved):
Looks good! Thank you!

crates/ntx-builder/src/transaction.rs
sergerad: I made it generic and just called it `LruCache<K, V>`

crates/utils/src/lru_cache.rs
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// Creates a new cache with the given capacity.
+    pub fn new(capacity: NonZeroUsize) -> Self {
+        Self(Arc::new(Mutex::new(InnerCache::new(capacity))))
+    }
+
+    /// Retrieves a value from the cache.
+    pub async fn get(&self, key: &K) -> Option<V> {
+        let mut cache_guard = self.0.lock().await;
+        cache_guard.get(key).cloned()
```
Mirko-von-Leipzig: Nit

```suggestion
        self.0.lock().await.get(key).cloned()
```

crates/utils/src/lru_cache.rs
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    /// Puts a value into the cache.
+    pub async fn put(&mut self, key: K, value: V) {
+        let mut cache_guard = self.0.lock().await;
+        cache_guard.put(key, value);
```
Mirko-von-Leipzig:
```suggestion
        self.0.lock().await.put(key, value);
```

crates/ntx-builder/src/transaction.rs
drahnr: A `tokio::sync::Mutex` would actually be fine, an `std::sync::Mutex`, not so much: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

Mirko-von-Leipzig: Ah misremembered :grin:

crates/ntx-builder/src/transaction.rs
```diff
@@ -70,19 +69,42 @@ type NtxResult<T> = Result<T, NtxError>;
...
+    /// LRU cache for storing retrieved note scripts to avoid repeated store calls.
+    script_cache: LruCache<Word, NoteScript>,
 }
 
 impl NtxContext {
+    /// Default cache size for note scripts.
+    ///
+    /// Each cached script contains the deserialized `NoteScript` object, so the actual memory usage
+    /// depends on the complexity of the scripts being cached.
+    const DEFAULT_SCRIPT_CACHE_SIZE: NonZeroUsize = NonZeroUsize::new(1000).unwrap();
```
bobbinth, line 92: Eventually, we should make this limit configurable. Let's create an issue for this (unless we already have one)

PR #1401, opened 2025-11-28
fix: address service reflection in rpc

### LINKED ISSUE #1391: gRPC reflection exposes internal methods
The node's RPC component has gRPC server reflection enabled which allows users to query and explore available methods and types.

Currently this also exposes our internal only methods and services i.e. store's gRPC interfaces and block-producer's interfaces. This should of course not be the case as these are private.

Its not entirely clear why this is the case as the build script appears to use the right subset of services:

```rust
    let proto_dir = crate_root.join("proto");
    let includes = &[proto_dir];

    let rpc_file_descriptor = protox::compile([RPC_PROTO], includes)?;
```

However, maybe this is simply a repercussion of sharing types and files with inner services?

This can be replicated by using an external service such as postman on a public rpc url.

SantiagoPittella:
Closes #1391 

@Mirko-von-Leipzig I looked a bit into `protox` and `tonic_prost_build` crates and they don't provide any tool to remove services. The thing is that we are importing the whole files for the messages to be present for the RPC API. One thing that we could do is to split services definitions and messages in different files. In this PR, I moved the messages that the RPC service uses from the store services into the `shared` file (which does not implement any service). In this version, the reflection works as we expect.

### DIFF
--- a/proto/proto/README.md
+++ b/proto/proto/README.md
@@ -1,0 +1,19 @@
+# Proto Files Organization
+
+The files are organized by a visibility hierarchy, where the root directory contains the public-facing RPC and remote prover protocols, while the `types` directory contains the data types used by these protocols. The `internal` directory contains the internal protocols used by the node, such as the store, non-transactional data, and block producer protocols.
+
+The organization of the files is as follows:
+
+```
+rpc.proto
+remote_prover.proto
+types/
+├── primitives.proto
+└── xxx.proto
+internal/
+├── store.proto
+├── ntx.proto
+└── block_producer.proto
+```
+
+The public-facing files should only allow the usage of the `types` directory, to avoid service reflection to internal protocols.


proto/proto/rpc.proto
Mirko-von-Leipzig, line 1: Ah I understand the problem now. I guess we can at least use this as a forcing function to think through our types and services properly.

Having a shared set of types makes sense; but I think we should refocus things so that external users don't see or need "internal" things at all.

What I'm thinking is that RPC definitions should be self-sufficient aka internal can use RPC types but RPC can only use RPC.

Maybe something like:

```
rpc.proto
types/
├── primitives.proto
└── xxx.proto
internal/
├── store.proto
├── ntx.proto
├── block_producer.proto
└── types/
    ├── internal.proto
    └── other_internal.proto
```

where we restrict `rpc.proto` to only allow usage of `types/` (this would effectively be something like your `shared`) but in such a way that normal users aren't aware of what we do internally. Basically, we should be able to remove `/internal` and have `rpc.proto` still work as expected.

Mirko-von-Leipzig: File names etc are of course just examples and it might make sense to organise them differently.

Something else I think we should consider is removing the `store` RPC redefinition. This is essentially a 1-1 copy of RPC service definition with the two submit tx/batch methods removed. I feel like we can save ourselves some time and effort by just having the store component implement the RPC service definition and returning an error for those two methods instead of redefining everything.

SantiagoPittella: >What I'm thinking is that RPC definitions should be self-sufficient aka internal can use RPC types but RPC can only use RPC.

Yeah, I think this comment nailed it in what we should have. I will implement this approach.

> Something else I think we should consider is removing the store RPC redefinition. This is essentially a 1-1 copy of RPC service definition with the two submit tx/batch methods removed. I feel like we can save ourselves some time and effort by just having the store component implement the RPC service definition and returning an error for those two methods instead of redefining everything.

Sounds good, an "unimplemented" should be enough, right?

Mirko-von-Leipzig: > Sounds good, an "unimplemented" should be enough, right?

imo yes :+1:

SantiagoPittella: I implemented this approach, and merged the services internal to the store in a single file, I we do not want this, I can create a `store` dir in `/internal` to have them split there. Also, `remote_prover.proto` is also "living" in the same level as `rpc.proto` since it can be used by an external user too.

Mirko-von-Leipzig: I think a single store file is good since the bulk of it was the RPC duplication :+1: Thank you!

proto/build.rs
Mirko-von-Leipzig, line 1: Once we get this change in, I would be interested in a follow-up which moves `/internal` to `crates/proto` aka enforcing the public/internal divorce.

I'm hoping that `crates/proto/build.rs` can somehow re-use the `proto` types and bindings _without_ having to manually import the raw proto files again aka somehow using the file descriptors (as we are now) and combining them with the local files.

This may not be trivial, and we may have the same problem as we do now with the RPC service leaking as part of the file descriptor. Maybe we have to do a separate RPC service definition export and a separate type definition export.

But this isn't super important :)

A nice side effect of this is that it lets us enforce a `changelog` checker since now our public gRPC API is cleanly separated and can be monitored for diffs.

proto/proto/rpc.proto
Mirko-von-Leipzig, line 1: I did not diligently review this file since I assume its mostly a matter of copy-pasting and adjusting the types?

proto/proto/internal/store.proto
```diff
@@ -0,0 +1,342 @@
...
+import "types/transaction.proto";
+import "types/note.proto";
+import "types/primitives.proto";
+import "rpc.proto";
+
+// RPC STORE API
+// ================================================================================================
+
+// Store API for the RPC component
+service Rpc {
```
Mirko-von-Leipzig, line 17: I thought you wanted to remove this in favor of re-using the RPC definition?

No big deal if we keep it; up to you.

Might actually be nicer as a follow-up PR so that the change can be litigated independently - I suspect there may be some counter-arguments :)

proto/proto/rpc.proto
SantiagoPittella: Yes, it is a copy paste of the service definition but  moving the messages from `shared.proto` to this file.

proto/proto/internal/store.proto
SantiagoPittella: While implementing this I realized that besides the two methods that will be left unimplemented, the `Status` method has a different message as a response

proto/proto/internal/block_producer.proto
```diff
@@ -2,6 +2,7 @@
 syntax = "proto3";
 package block_producer;
 
+import "rpc.proto";
```
drahnr, line 5: nit: I find it counterintuitive an internal proto file imports `rpc.proto`. I'd expect it'd be the other way round.

proto/proto/internal/store.proto
```diff
@@ -0,0 +1,342 @@
+// Unified specification of the internal store gRPC APIs.
```
drahnr, line 1: Maybe we should have small readme here `proto/proto/README.md` explaining the rationale?

drahnr (approved):
I understand the rationale, and obviously didn't check every single line for correctness. There are two small questions inline, one might be related to how protobuf does dependency resolution. Feel free to ignore both.

The broad stroke LGTM

crates/block-producer/src/server/mod.rs
```diff
@@ -258,8 +258,7 @@ impl api_server::Api for BlockProducerRpcServer {
     async fn submit_proven_transaction(
         &self,
         request: tonic::Request<proto::transaction::ProvenTransaction>,
-    ) -> Result<tonic::Response<proto::block_producer::SubmitProvenTransactionResponse>, Status>
-    {
+    ) -> Result<tonic::Response<proto::blockchain::BlockNumber>, Status> {
```
igamigo, line 261: Is this replacement of `SubmitProvenBatchResponse` on purpose? Seems fine but I don't think I saw this discussed/mentioned so just double checking

igamigo (commented):
LGTM! If I'm not mistaken this should help with the extra code generation in the client as well, where we are generating quite more types than we need (for the store's API, for example). I recalled there being an issue but upon a quick search I could not find it (I had also investigated settings in `prost` but couldn't avoid this from the client side alone).

proto/proto/internal/store.proto
Mirko-von-Leipzig: Okay; we can tackle that another day then

proto/proto/internal/block_producer.proto
Mirko-von-Leipzig: I mean that was the entire point of this exercise :sweat_smile: To have a free standing _public_ API. What we do with the internal API is our problem/business so if we want to import the public API then that's okay.

Put differently:
- if the _public_ API imports the internal API, then
- the public API exposes the internal API
- which is also demonstrated by #1391

In addition, having the public API be free standing makes it much easier to properly version it since it doesn't magically change when our internal API does.

crates/block-producer/src/server/mod.rs
Mirko-von-Leipzig: I think its just a unification/simplification since the response is just a block number. I think its better this way as well tbh. Unless of course we plan on expanding the output.

Having `XxxRequest` and `XxxResponse` should be the last resort imo since they cannot be re-used for other APIs/methods.

proto/proto/internal/block_producer.proto
drahnr: Understood :)

My intention was to provide i.e.

```
a.proto
b.proto

external.proto (use a.proto)
internal.proto (use a.proto; use b.proto)
```

but the given does the shortcut of `a.proto = external.proto` and going through with `external.proto`.

crates/block-producer/src/server/mod.rs
igamigo: Yeah, I agree. I recall there being a conversation about using `*Request`/`*Response` for all requests, but I prefer this too (also makes code less boilerplatey)

proto/build.rs
SantiagoPittella: I will open an issue to track this further

PR #1402, opened 2025-12-01
chore: Misc followups for block validation

### LINKED ISSUE #1395: follow up to validate and prove blocks #1381
A handful of comments and nits is left over from https://github.com/0xMiden/miden-node/pull/1381 which should be addressed in a follow-up PR.

sergerad:
Closes #1395.

Follows up comments from https://github.com/0xMiden/miden-node/pull/1381.

### DIFF
--- a/crates/block-producer/src/block_builder/mod.rs
+++ b/crates/block-producer/src/block_builder/mod.rs
@@ -28,7 +28,7 @@ use url::Url;
 use crate::errors::BuildBlockError;
 use crate::mempool::SharedMempool;
 use crate::store::StoreClient;
-use crate::validator::BlockProducerValidatorClient;
+use crate::validator::{BlockProducerValidatorClient, BodyDiff, HeaderDiff, ValidatorError};
 use crate::{COMPONENT, TelemetryInjectorExt};
 
 // BLOCK BUILDER
@@ -240,7 +240,7 @@ impl BlockBuilder {
             .validator
             .sign_block(proposed_block.clone())
             .await
-            .map_err(BuildBlockError::ValidateBlockFailed)?;
+            .map_err(|err| BuildBlockError::ValidateBlockFailed(err.into()))?;
         let (expected_header, expected_body) = build_result
             .await
             .map_err(|err| BuildBlockError::other(format!("task join error: {err}")))?
@@ -250,10 +250,20 @@ impl BlockBuilder {
         // proposed block.
         // TODO(sergerad): Update Eq implementation once signatures are part of the header.
         if header != expected_header {
-            return Err(BuildBlockError::ValidateBlockMismatch("header".into()));
+            let diff = HeaderDiff {
+                validator_header: header,
+                expected_header,
+            }
+            .into();
+            return Err(BuildBlockError::ValidateBlockFailed(
+                ValidatorError::HeaderMismatch(diff).into(),
+            ));
         }
         if body != expected_body {
-            return Err(BuildBlockError::ValidateBlockMismatch("body".into()));
+            let diff = BodyDiff { validator_body: body, expected_body }.into();
+            return Err(BuildBlockError::ValidateBlockFailed(
+                ValidatorError::BodyMismatch(diff).into(),
+            ));
         }
 
         let (ordered_batches, ..) = proposed_block.into_parts();
--- a/crates/block-producer/src/errors.rs
+++ b/crates/block-producer/src/errors.rs
@@ -209,9 +209,7 @@ pub enum BuildBlockError {
     #[error("failed to propose block")]
     ProposeBlockFailed(#[source] ProposedBlockError),
     #[error("failed to validate block")]
-    ValidateBlockFailed(#[source] ValidatorError),
-    #[error("the validated {0} does not match the expected {0}")]
-    ValidateBlockMismatch(String),
+    ValidateBlockFailed(#[source] Box<ValidatorError>),
     #[error("failed to prove block")]
     ProveBlockFailed(#[source] BlockProverError),
     /// We sometimes randomly inject errors into the batch building process to test our failure
--- a/crates/block-producer/src/validator/mod.rs
+++ b/crates/block-producer/src/validator/mod.rs
@@ -1,3 +1,5 @@
+use std::fmt::{Display, Formatter};
+
 use miden_node_proto::clients::{Builder, ValidatorClient};
 use miden_node_proto::errors::{ConversionError, MissingFieldHelper};
 use miden_node_proto::generated as proto;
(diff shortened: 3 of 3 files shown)


crates/block-producer/src/block_builder/mod.rs
```diff
@@ -230,13 +231,32 @@ impl BlockBuilder {
...
+        // Request validation concurrently to building the block ourselves.
+        let (validator_result, build_result) = tokio::join!(
+            // Send request to validator.
+            self.validator.validate_block(proposed_block.clone()),
+            // Build the block ourselves.
+            tokio::task::spawn_blocking({
+                let proposed_block = proposed_block.clone();
+                move || build_block(proposed_block)
+            })
+        );
```
bobbinth: Could we not do this without explicitly spawning new threads - something like this:
```rust
let response = self.validator.validate_block(proposed_block.clone());
let local_block = build_block(proposed_block);
let response = response.await.map_err(BuildBlockError::ValidateBlockFailed)?;
```

Or is this undesirable for some reason?

crates/block-producer/src/validator/mod.rs
bobbinth: Should we update naming here as well? i.e., should `ValidateBlockResponse` be `SignedBlock`? Or alternatively, we could just return `(BlockHeader, BlockBody)` tuple from here.

bobbinth: Should we re-name this as well?

crates/block-producer/src/block_builder/mod.rs
sergerad: That example is not doing those operations concurrently. The response future is only polled after the `build_block()` call.

bin/remote-prover/src/api/prover.rs
Mirko-von-Leipzig, line 1: I assume this was just to get the additional `block_id` into the tracing?

crates/block-producer/src/errors.rs
```diff
@@ -210,6 +210,8 @@ pub enum BuildBlockError {
     ProposeBlockFailed(#[source] ProposedBlockError),
     #[error("failed to validate block")]
     ValidateBlockFailed(#[source] ValidatorError),
+    #[error("the validated {0} does not match the expected {0}")]
+    ValidateBlockMismatch(String),
```
Mirko-von-Leipzig: I think let's roll this into the `ValidatorError`?

And rather have a separate variant for header and body rather than stringly typed.

Event more ideal would be knowing _what_ the difference as part of the error report, but I'm unsure how easy that would be so probably not needed now.

I'm thinking something like:

```rust
enum ValidatorError {
    ...
    #[error("validator header does not match the request: {diff}")]
    HeaderMismatch {
        diff: HeaderDiff
    }
}

struct HeaderDiff {
    expected: Header,
    got: Header
}

impl Display for HeaderDiff {
    fn fmt(...) {
        // Maybe this is as simple as subtracting the debug of both?
    }
}
```

Alternatively one would have to iterate through all the fields manually which is maybe not the worst thing either.

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: If you want concurrency then you'll always need to spawn at least one of the tasks. Note that this isn't spawning new _threads_ only off-loading the task to the runtime so it can be scheduled to concurrently with the current task.

We could simplify this a bit since the `spawn_blocking` already does an off-loading by doing:

```rust
// This will start running on the separate blocking thread pool.
//
// Since this is a somewhat compute heavy task (I assume), it makes sense
// to do it as blocking. If its light then it could be regular `spawn`.
let build_block = tokio::task::spawn_blocking({
    let proposed_block = proposed_block.clone();
    move || build_block(proposed_block)
});

// We can run this in-line since the other task is already spawned.
let validator_result = self.validator.validate_block(proposed_block.clone()).await;
let build_result = build_block.await;
```

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -268,13 +288,17 @@ impl BlockBuilder {
         body: BlockBody,
         block_proof: BlockProof,
     ) -> Result<ProvenBlock, BuildBlockError> {
+        // SAFETY: The header and body are assumed valid and consistent with the proof.
```
Mirko-von-Leipzig, line 299: This `fn` doesn't need to be `async` anymore. I'm actually surprised `clippy` doesn't complain about this but maybe we've disabled the lint.

crates/block-producer/src/block_builder/mod.rs
drahnr: Small caveat: The default blocking threadpool is _very_ large, which can lead to starvation and running into timeouts up the call-chain.

crates/block-producer/src/block_builder/mod.rs
sergerad: Would mean I have to wrap it in a future manually in the call chain, probably prefer to keep it in fn signature?
```rust 
.and_then(|(proposed_block, header, body, block_proof)| async {self.construct_proven_block(proposed_block, header, body, block_proof)})
```

crates/block-producer/src/errors.rs
sergerad: Opted to go with simple output rather than iterating over each variable which would be error prone / hard to maintain. We can just take the output and diff it manually when debugging

bin/remote-prover/src/api/prover.rs
sergerad: Yea moved it to avoid clone

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: I've done it similarly in the batch builder 

https://github.com/0xMiden/miden-node/blob/9460b4423f6274441975ced75453b121013cab1b/crates/block-producer/src/batch_builder/mod.rs#L181

Though I also did consider just saying screw it and keeping the async. I think its up to you, at the time I thought its worth indicating to the reader that the function is _not_ async (e.g. it stood out to me on reading just the fn now).

Mirko-von-Leipzig: If you want it to be shorter, you can always leave the tuple in place

```rust
.and_then(|args| async {self.construct_proven_block(args.0, args.1, args.2, args.3)})
```

Its a pity rust doesn't have a spread operator so one could do `args...`

crates/block-producer/src/errors.rs
Mirko-von-Leipzig: Perfect actually :heart:

This would be a pretty unexpected error to get, and hopefully rare, so dumping it all is appropriate. Sorry for over complicating things.

bobbinth (approved):
Looks good! Thank you!

PR #1407, opened 2025-12-01
fix: prover clients WASM lint & block prover

SantiagoPittella:
The client team reported an error were the WASM prover client was failing to compile due to a bad import in the clients. This error wasn't reported before since the Makefile wasn't properly set to run the WASM clients checks.

After fixing the Makefile and the imports, when the lint started to run as it is intended, an error with the `block-prover` arose, because it was using `miden-node-proto` which is unsuited for no-std. I address this error with the usage of `ProposedBlock`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 - RPC client now correctly sets `genesis` value in `ACCEPT` header if `version` is unspecified ([#1370](https://github.com/0xMiden/miden-node/pull/1370)).
 - Pin protobuf (`protox`) dependencies to avoid breaking changes in transitive dependency ([#1403](https://github.com/0xMiden/miden-node/pull/1403)).
+- Fixed no-std compatibility for remote prover clients ([#1407](https://github.com/0xMiden/miden-node/pull/1407)).
 
 ## v0.12.6
 


crates/remote-prover-client/src/remote_prover/block_prover.rs
Mirko-von-Leipzig, line 1: iiuc the prover depended on the proto crate solely for the `BlockProofRequest` definition and instead you now build it manually?

SantiagoPittella: yes, into the remote prover proto type.

crates/remote-prover-client/Cargo.toml
```diff
@@ -38,13 +38,12 @@ tonic-web = { optional = true, version = "0.14" }
...
-miden-node-proto = { optional = true, workspace = true }
-miden-objects    = { optional = true, workspace = true }
-miden-tx         = { optional = true, workspace = true }
-prost            = { default-features = false, features = ["derive"], version = "0.14" }
-thiserror        = { workspace = true }
-tokio            = { default-features = false, features = ["sync"], optional = true, version = "1.44" }
-tonic-prost      = { workspace = true }
+miden-objects = { optional = true, workspace = true }
+miden-tx      = { optional = true, workspace = true }
+prost         = { default-features = false, features = ["derive"], version = "0.14" }
```
Mirko-von-Leipzig, line 43: I notice that this isn't `workspace = true`; we should change that.

For example, #1404 might affect this as well?

SantiagoPittella: I'll open another PR using the workspace version.

PR #1410, opened 2025-12-02
feat: rpc endpoint for parameter limits

### LINKED ISSUE #1405: Expose RPC limits via gRPC method
We have limits for certain RPC methods to restrict how much database cost the query can consume.

These are constants but aren't known to users unless they copy them (and of course this is brittle).

Add a new method to the RPC's gRPC service which exposes these limits.

https://github.com/0xMiden/miden-client/issues/1139#issuecomment-3597790157

### LINKED ISSUE #1473: Document RPC limits
Separately, it may be good to also include limits in the RPC endpoint [docs](https://docs.miden.xyz/miden-node/rpc). Let's create another issue for this (unless we already have one).

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-node/issues/1443#issuecomment-3667859247_

drahnr:
Closes #1405, closes #1473

This PR add a new endpoint (`GetLimits`) to retrieve the requests limits on other endpoints. It adds documentation for the new endpoint and adds docs on each limited endpoint to clarify the maximum amount of items that can be requested.

This PR enables a counterpart in the `miden-client` repository, the issue already exists (https://github.com/0xMiden/miden-client/issues/1597)

### DIFF
--- a/crates/rpc/src/tests.rs
+++ b/crates/rpc/src/tests.rs
@@ -470,7 +470,7 @@ async fn start_store(store_addr: SocketAddr) -> (Runtime, TempDir, Word) {
 }
 
 #[tokio::test]
-async fn test_get_limits_endpoint() {
+async fn get_limits_endpoint() {
     // Start the RPC and store
     let (mut rpc_client, _rpc_addr, store_addr) = start_rpc().await;
     let (store_runtime, _data_directory, _genesis) = start_store(store_addr).await;


proto/proto/rpc.proto
Mirko-von-Leipzig, line 1: I think maybe `method` is a better word than `endpoint`. To me `endpoint` implies the `url` itself, though of course this becomes a bit muddy because a gRPC _method_ is invoked using the `<url>/<method>` endpoint 😅

Up to you.

proto/proto/rpc.proto
```diff
@@ -610,3 +613,22 @@ message TransactionRecord {
...
+
+// RPC LIMITS
+// ================================================================================================
+
+// Represents the query parameter limits for RPC endpoints.
+message RpcLimits {
+    // Maps RPC endpoint names to their parameter limits.
+    // Key: endpoint name (e.g., "CheckNullifiers", "SyncState")
+    // Value: map of parameter names to their limit values
+    map<string, EndpointLimits> endpoints = 1;
```
Mirko-von-Leipzig, line 629: How many limited endpoints are there here? I guess I'm wondering if we should be explicit. Though probably this is better :)

proto/proto/rpc.proto
```diff
@@ -97,6 +97,9 @@ service Api {
 
     // Returns transactions records for specific accounts within a block range.
     rpc SyncTransactions(SyncTransactionsRequest) returns (SyncTransactionsResponse) {}
+
+    // Returns the query parameter limits configured for RPC endpoints.
+    rpc GetLimits(google.protobuf.Empty) returns (RpcLimits) {}
```
Mirko-von-Leipzig, line 106:
```suggestion
    // Returns the query parameter limits configured for RPC endpoints.
    //
    // These define the maximum number of each parameter an endpoint will
    // accept. Exceeding the limit will result in the request being rejected
    // and you should instead send multiple smaller requests.
    rpc GetLimits(google.protobuf.Empty) returns (RpcLimits) {}
```

docs/external/src/rpc.md
Mirko-von-Leipzig, line 1: At some point we should consider some sort of CI check that ensures the limits and docs etc are all in sync.

Doing this isn't super trivial but some options are:

- Doc templates, where the values are filled in from the code values and we "build" template into static docs. CI then checks there's no diff.
- Bash script which lifts the values from both code and docs and ensures they're equal.

Something I also dislike is that we have both this file and the `rpc.proto` gRPC descriptions which should probably be the same thing. Which somewhat further leads down the "build" docs idea.

Mirko-von-Leipzig: I imagine there is some framework that already does something like this.. maybe we already have it in Docusaurus or whatever? cc @Keinberger

crates/rpc/src/tests.rs
```diff
@@ -461,3 +468,58 @@ async fn start_store(store_addr: SocketAddr) -> (Runtime, TempDir, Word) {
         genesis_state.into_block().unwrap().inner().header().commitment(),
     )
 }
+
+#[tokio::test]
+async fn test_get_limits_endpoint() {
```
Mirko-von-Leipzig:
```suggestion
async fn get_limits_endpoint() {
```

crates/rpc/README.md
```diff
@@ -36,6 +37,8 @@ The full gRPC method definitions can be found in the [proto](../proto/README.md)
 
 Returns a nullifier proof for each of the requested nullifiers.
 
+**Limits:** `nullifier` (1000)
```
drahnr, line 40: 👍

proto/proto/rpc.proto
SantiagoPittella: Yeah, `method` is more accurate IMO 👍🏼

proto/proto/rpc.proto
SantiagoPittella: At the moment, there are 5 public and 2 internal methods enforcing this limits

docs/external/src/rpc.md
SantiagoPittella: Agree, it would be great if we can avoid having that much sources of truth. At least, having the docs built from the proto file, and updating the constants just there is a great first milestone

bobbinth (approved):
Looks good! Thank you!

PR #1411, opened 2025-12-02
ci: install cargo-msrv using binstall

Mirko-von-Leipzig:
This should reduce the cache usage of our `msrv` check (#1376).

Its unclear by how much though, so let's see.


Mirko-von-Leipzig:
> How did it shake out?

No change at all :/ 1.6GB before and after. Seems cache in this instance is mostly build deps; I'm wondering if we can't specify a cargo profile for CI that trims the debug symbols from external deps. This should help (probably).

PR #1413, opened 2025-12-02
ci: reflow comments using awk

Mirko-von-Leipzig:
I'm experimenting with using `awk` to reflow comment lines.

The awk script is generated via AI (because doing this manually.. !@#% that :)).

Flow is basically:

1. Run `reflow.sh`
    - This uses the AI generated awk script to merge consecutive comment lines together with some additional rules to account for edge case e.g. code blocks.
    - Comment lines are now too long since they've merged
2. Run `cargo +nightly fmt --all` to truncate the overly long comment lines again
3. ..profit?

It _should_ handle and be considerate of

- Normal comments, doc comments
- Markdown headings
- Lists
- Code fences
- Preserve indentation
- Our `============...` delimiters
- Ignores generated code and target directory
- Preserves empty comment lines

--------------

In addition I've modified the makefile to include the flow. The fmt checker in CI now runs the flow and ensures that the diff is zero.

### DIFF
--- a/crates/store/build.rs
+++ b/crates/store/build.rs
@@ -1,5 +1,4 @@
-// This build.rs is required to trigger the `diesel_migrations::embed_migrations!` proc-macro in
-// `store/src/db/migrations.rs` to include the latest version of the migrations into the binary, see <https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds>.
+// This build.rs is required to trigger the `diesel_migrations::embed_migrations!` proc-macro in `store/src/db/migrations.rs` to include the latest version of the migrations into the binary, see <https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds>.
 fn main() {
     println!("cargo:rerun-if-changed=./src/db/migrations");
     // If we do one re-write, the default rules are disabled,
--- a/crates/store/src/db/migrations.rs
+++ b/crates/store/src/db/migrations.rs
@@ -4,8 +4,7 @@ use tracing::instrument;
 
 use crate::COMPONENT;
 
-// The rebuild is automatically triggered by `build.rs` as described in
-// <https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds>.
+// The rebuild is automatically triggered by `build.rs` as described in <https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds>.
 pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!("src/db/migrations");
 
 // TODO we have not tested this in practice!
--- a/crates/store/src/db/schema.rs
+++ b/crates/store/src/db/schema.rs
@@ -103,8 +103,8 @@ diesel::table! {
 diesel::joinable!(accounts -> account_codes (code_commitment));
 diesel::joinable!(accounts -> block_headers (block_num));
 // Note: Cannot use diesel::joinable! with accounts table due to composite primary key
-// diesel::joinable!(notes -> accounts (sender));
-// diesel::joinable!(transactions -> accounts (account_id));
+// diesel::joinable!(notes -> accounts (sender)); diesel::joinable!(transactions -> accounts
+// (account_id));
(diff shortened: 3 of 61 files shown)


crates/store/src/db/schema.rs
```diff
@@ -103,8 +103,8 @@ diesel::table! {
 diesel::joinable!(accounts -> account_codes (code_commitment));
 diesel::joinable!(accounts -> block_headers (block_num));
 // Note: Cannot use diesel::joinable! with accounts table due to composite primary key
-// diesel::joinable!(notes -> accounts (sender));
-// diesel::joinable!(transactions -> accounts (account_id));
+// diesel::joinable!(notes -> accounts (sender)); diesel::joinable!(transactions -> accounts
+// (account_id));
```
Mirko-von-Leipzig, line 107: This is one outlier which is "malformatted" but I think likely we should place this in a code fence or something.

I'm leaving this as is for now just for discussion's sake.

crates/store/src/db/migrations.rs
```diff
@@ -4,8 +4,7 @@ use tracing::instrument;
 
 use crate::COMPONENT;
 
-// The rebuild is automatically triggered by `build.rs` as described in
-// <https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds>.
+// The rebuild is automatically triggered by `build.rs` as described in <https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds>.
```
Mirko-von-Leipzig, line 7: This line is arguably worse as well now, but its because the link exceeds the limit and rustfmt doesn't know what to do. This might behave better if it was a doc comment.

crates/store/build.rs
```diff
@@ -1,5 +1,4 @@
-// This build.rs is required to trigger the `diesel_migrations::embed_migrations!` proc-macro in
-// `store/src/db/migrations.rs` to include the latest version of the migrations into the binary, see <https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds>.
+// This build.rs is required to trigger the `diesel_migrations::embed_migrations!` proc-macro in `store/src/db/migrations.rs` to include the latest version of the migrations into the binary, see <https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds>.
```
Mirko-von-Leipzig, line 1: Also a failure point; unclear why except rustfmt thought this was the way to do it 🤷

Mirko-von-Leipzig (commented):
I think this actually worked very well.

crates/store/src/db/schema.rs
drahnr: Yeah, should have been a code block in the first place - or deleted.

scripts/reflow.awk
```diff
@@ -0,0 +1,92 @@
...
+        } else {
+            if(in_comment) print out_prefix merged
+            in_comment=1
+            last_type=type
+            merged=text
+            out_prefix = prefix match_space
+        }
+
+        next
+    }
```
drahnr, line 84: End of poor-mans-markdown parser 😰

scripts/reflow.awk
```diff
@@ -0,0 +1,92 @@
...
+
+    # Non-comment line
+    if(in_comment){ print out_prefix merged; in_comment=0; out_space="" }
+    print line
+}
+
+END { if(in_comment) print out_prefix merged }
+
```
drahnr, line 92: I've such an itch to fixup `cargo-spellcheck reflow` now...

drahnr (approved):

crates/store/src/db/migrations.rs
drahnr: It'd be great to have an opt out way for it, i.e. like `clang-format`. But the `awk` is already scarily complex.

Mirko-von-Leipzig: I mean we can probably get it to respect some kind of `// [rustfmt::skip]` pretty easily. But I think in this case its a problem with the `< ... >`

Mirko-von-Leipzig:
I'm going draft this until next year or until all other major ongoing PRs are merged. This PR will conflict with literally everything :)

PR #1414, opened 2025-12-02
doc: improve `CheckNullifiers` documentation

drahnr:
It's non obvious how to hold the API without having a lot of context. This PR attempts to add sufficient context to provide that context without having to read up on non-inclusion proofs and how we implement them.

### DIFF
--- a/crates/proto/src/generated/rpc.rs
+++ b/crates/proto/src/generated/rpc.rs
@@ -1080,7 +1080,6 @@ pub mod api_server {
         ///
         /// * `empty_leaf_index`: Non-inclusion proof (nullifier not in tree)
         /// * `single` or `multiple`: Inclusion proof only if the requested nullifier appears as a key.
-        ///  The value associated with the nullifier key represents the block number at which it was consumed.
         ///
         /// Verify proofs against the nullifier tree root in the latest block header.
         async fn check_nullifiers(
--- a/proto/proto/rpc.proto
+++ b/proto/proto/rpc.proto
@@ -26,7 +26,6 @@ service Api {
     // The `leaf` field indicates the status:
     // * `empty_leaf_index`: Non-inclusion proof (nullifier not in tree)
     // * `single` or `multiple`: Inclusion proof only if the requested nullifier appears as a key.
-    //   The value associated with the nullifier key represents the block number at which it was consumed.
     //
     // Verify proofs against the nullifier tree root in the latest block header.
     rpc CheckNullifiers(NullifierList) returns (CheckNullifiersResponse) {}


proto/proto/store/rpc.proto
```diff
@@ -231,15 +242,23 @@ message AccountStorageDetails {
 // CHECK NULLIFIERS
 // ================================================================================================
 
-// List of nullifiers to return proofs for.
+// List of nullifiers to check for inclusion or non-inclusion proofs.
 message NullifierList {
-    // List of nullifiers to return proofs for.
+    // List of nullifiers to return proofs for. Each nullifier is a Digest (4 field elements).
     repeated primitives.Digest nullifiers = 1;
 }
```
bobbinth: I wonder if we should allow users to query historical nullifier states - similar to what we have for accounts with `AccountTreeWithHistory`. Not a high priority, but given that we have most of the structure working already, shouldn't be too difficult to implement.

drahnr: https://github.com/0xMiden/miden-node/issues/1416

proto/proto/store/rpc.proto
```diff
@@ -232,15 +242,23 @@ message AccountStorageDetails {
 // CHECK NULLIFIERS
 // ================================================================================================
 
-// List of nullifiers to return proofs for.
+// List of nullifiers to check for inclusion or non-inclusion proofs.
```
Mirko-von-Leipzig: Nit: this comment is overly specific since this is just a list of nullifiers. The _usage_ in `CheckNulifiers` adds the context about proofs etc.

proto/proto/store/rpc.proto
```diff
@@ -232,15 +242,23 @@ message AccountStorageDetails {
...
 message NullifierList {
-    // List of nullifiers to return proofs for.
+    // List of nullifiers to return proofs for. Each nullifier is a Digest (4 field elements).
```
Mirko-von-Leipzig: imo this shouldn't have a comment since it would just be repeat of the outer comment.

proto/proto/store/rpc.proto
```diff
@@ -232,15 +242,23 @@ message AccountStorageDetails {
...
     repeated primitives.Digest nullifiers = 1;
 }
 
-// Represents the result of checking nullifiers.
+// Response containing SMT opening proofs for the requested nullifiers.
+//
+// Each proof verifies whether the nullifier exists in the tree (consumed) or not (available).
 message CheckNullifiersResponse {
```
Mirko-von-Leipzig: I'd prefer if we move away from request/response if the type can actually be named properly.

But that isn't really this PRs problem so feel free to disregard.


```suggestion
// A set of Sparse Merkle Tree openings which provide proof of the status
// of each nullifier in the set.
//
// Each nullifier's proof is either a proof of inclusion (consumed) or exclusion (unconsumed)
// which can be verified against the nullifier tree root.
message NullifierProofs {
```

docs/external/src/rpc.md
```diff
@@ -33,7 +33,74 @@ The gRPC service definition can be found in the Miden node's `proto` [directory]
...
+```rust
+use miden_crypto::merkle::SmtProof;
+
+// 1. Get nullifier tree root from block header
+let block_header = get_latest_block_header();
+let nullifier_tree_root = block_header.state_commitment().nullifier_root();
+
+// 2. Verify the proof
+let proof: SmtProof = smt_opening.try_into()?;
+assert_eq!(proof.root(), nullifier_tree_root);
```
PhilippGackstatter: Should this be `compute_root` (https://docs.rs/miden-crypto/latest/miden_crypto/merkle/smt/struct.SmtProof.html#method.compute_root)?

docs/external/src/rpc.md
```diff
@@ -33,7 +33,74 @@ The gRPC service definition can be found in the Miden node's `proto` [directory]
...
+
+// Alternatively:
+
+// non-inclusion
+proof.verify_membership(nullifier, EMPTY_WORD, nullifier_tree_root)
+// inclusion
+proof.verify_membership(nullifier, block_num_to_nullifier_leaf_value(block_header.block_num()), nullifier_tree_root)
+```
 
 ### GetAccountDetails
```
PhilippGackstatter: I think the multiple branch would not compile, `filter` does not return a `bool`.

After https://github.com/0xMiden/crypto/pull/682, I think I would check it this way:

```rust
match proof.verify_non_membership(&key, &root) {
    Ok(()) => println!("note is consumable"),
    // Indicates the proof is valid, but the value is NOT the empty word -> nullifier spent.
    Err(SmtProofError::ValueMismatch { .. }) => println!("note is already consumed"),
    Err(_) => println!("proof is invalid"),
}
```

But I find checking this correctly pretty tricky, so maybe recommending one way to do it is best?

proto/proto/store/rpc.proto
```diff
@@ -20,7 +20,17 @@ service Rpc {
...
-    // Returns a nullifier proof for each of the requested nullifiers.
+    // Returns a Sparse Merkle Tree opening proof for each requested nullifier
+    //
+    // Each proof demonstrates either:
+    // - **Inclusion**: Nullifier exists in the tree (consumed)
+    // - **Non-inclusion**: Nullifier does not exist (not consumed)
+    //
+    // The `leaf` field indicates the status:
+    // - `empty_leaf_index`: Non-inclusion proof
+    // - `single` or `multiple`: Inclusion proof with key-value pair(s)
```
PhilippGackstatter: Edit: My bad, this comment is for the wrong line. It was meant for `rpc.md`.
> - Note has been spent

I read this as: If I request nullifiers `[0xabcd, 0x1234]` and I get back `[SmtProof::Single(_), SmtProof::Multiple(_)]`, then this indicates that both nullifiers are already spent, but this is not the case. We have to check their membership to be sure, because the returned proofs could be for other keys that map to the same leaf.

I had to double-check, because I find the SMT API tricky. These are the tests that I used:

```rust
#[test]
fn nullifier_already_spent_single() {
    let unspent_nullifier = Nullifier::from_raw(Word::from([0, 0, 0, 3u32]));
    let spent_nullifier = Nullifier::from_raw(Word::from([9, 9, 9, 3u32]));

    // Sanity check: Nullifiers map to the same leaf.
    assert_eq!(
        LeafIndex::from(unspent_nullifier.as_word()),
        LeafIndex::from(spent_nullifier.as_word())
    );

    let block1 = BlockNumber::from(1);

    let tree = NullifierTree::with_entries([(spent_nullifier, block1)]).unwrap();

    let witness = tree.open(&unspent_nullifier);

    assert_matches!(witness.proof().leaf(), SmtLeaf::Single(_));

    let is_unspent = witness.proof().verify_membership(
        &unspent_nullifier.as_word(),
        &EMPTY_WORD,
        &tree.root(),
    );
    assert!(is_unspent);
}

#[test]
fn nullifier_already_spent_multiple() {
    let unspent_nullifier = Nullifier::from_raw(Word::from([0, 0, 0, 3u32]));
    let spent_nullifier0 = Nullifier::from_raw(Word::from([8, 9, 8, 3u32]));
    let spent_nullifier1 = Nullifier::from_raw(Word::from([9, 9, 9, 3u32]));

    // Sanity check: Nullifiers map to the same leaf.
    assert_eq!(
        LeafIndex::from(unspent_nullifier.as_word()),
        LeafIndex::from(spent_nullifier0.as_word())
    );
    assert_eq!(
        LeafIndex::from(unspent_nullifier.as_word()),
        LeafIndex::from(spent_nullifier1.as_word())
    );

    let block1 = BlockNumber::from(1);

    let tree =
        NullifierTree::with_entries([(spent_nullifier0, block1), (spent_nullifier1, block1)])
            .unwrap();

    let witness = tree.open(&unspent_nullifier);

    assert_matches!(witness.proof().leaf(), SmtLeaf::Multiple(_));

    let is_unspent = witness.proof().verify_membership(
        &unspent_nullifier.as_word(),
        &EMPTY_WORD,
        &tree.root(),
    );
    assert!(is_unspent);
}
```

docs/external/src/rpc.md
```diff
@@ -33,7 +33,74 @@ The gRPC service definition can be found in the Miden node's `proto` [directory]
...
+    _ => println!("NOT consumed"),
+}
+
+// Alternatively:
+
+// non-inclusion
+proof.verify_membership(nullifier, EMPTY_WORD, nullifier_tree_root)
+// inclusion
+proof.verify_membership(nullifier, block_num_to_nullifier_leaf_value(block_header.block_num()), nullifier_tree_root)
+```
```
PhilippGackstatter: I think the multiple branch in the match statement would not compile, because `filter` does not return a `bool`.

After https://github.com/0xMiden/crypto/pull/682, I think I would check it this way:

```rust
match proof.verify_non_membership(&key, &root) {
    Ok(()) => println!("note is consumable"),
    // Indicates the proof is valid, but the value is NOT the empty word -> nullifier spent.
    Err(SmtProofError::ValueMismatch { .. }) => println!("note is already consumed"),
    Err(_) => println!("proof is invalid"),
}
```

So my point is, I would probably only recommend the `verify_membership` APIs because other methods seem pretty involved to get right.

drahnr: I didn't know the timeline of #682 so didn't want to rely on it yet

proto/proto/store/rpc.proto
drahnr: The `get_word() -> Option<Word>` was the piece that tripped me up.

You're correct, I had it fixed further down already but missed this one

proto/proto/store/rpc.proto
drahnr: Agree on the out of scope piece.

I think it's fine for the definitions inside the `rpc.proto`, but not a strong opinion.

proto/proto/store/rpc.proto
drahnr: Reverted the change. (But what else would you do with a list of proofs in the nullifier trees?)

PR #1415, opened 2025-12-02
fix: acccount upsert failing on insert (new account)

juan518munoz:
[While updating `miden-client` to the latest version of the node](https://github.com/0xMiden/miden-client/pull/1570), I came accross a genesis error caused by failure to insert the genesis account.

This error was introduced in https://github.com/0xMiden/miden-node/pull/1333, both `account_vault_assets` & `note_scripts` now have foreign key contraints on the table `accounts`, but following the logic of `upsert_accounts`, we first insert on the dependent tables, and then on `accounts` leading to the error.

### DIFF
--- a/crates/store/src/db/models/queries/accounts.rs
+++ b/crates/store/src/db/models/queries/accounts.rs
@@ -781,6 +781,10 @@ pub(crate) fn upsert_accounts(
             None
         };
 
+        // NOTE: we collect storage / asset inserts to apply them only after the  account row is
+        // written. The storage and vault tables have FKs pointing to `accounts (account_id,
+        // block_num)`, so inserting them earlier would violate those constraints when inserting a
+        // brand-new account.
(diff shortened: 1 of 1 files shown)


crates/store/src/db/models/queries/accounts.rs
```diff
@@ -774,16 +774,17 @@ pub(crate) fn upsert_accounts(
...
+
         let network_account_id_prefix = if account_id.is_network() {
             Some(NetworkAccountPrefix::try_from(account_id)?)
         } else {
             None
         };
 
-        let full_account: Option<Account> = match update.details() {
-            AccountUpdateDetails::Private => None,
+        let (full_account, pending_storage_inserts, pending_asset_inserts) = match update.details()
```
drahnr, line 788: nit: adding a comment re foreign key constraints as rationale for the extra step would be good

drahnr (approved):

PR #1417, opened 2025-12-02
chore: use dependencies from workspace

SantiagoPittella:


crates/proto/Cargo.toml
```diff
@@ -23,7 +23,7 @@ miden-node-utils            = { workspace = true }
 miden-objects               = { workspace = true }
 prost                       = { workspace = true }
 thiserror                   = { workspace = true }
-tonic                       = { workspace = true }
+tonic                       = { default-features = true, workspace = true }
```
drahnr, line 26: Any particular reason to add `default-features = true`?

crates/ntx-builder/Cargo.toml
```diff
@@ -24,7 +24,7 @@ miden-tx                   = { default-features = true, workspace = true }
 thiserror                  = { workspace = true }
 tokio                      = { features = ["rt-multi-thread"], workspace = true }
 tokio-stream               = { workspace = true }
-tonic                      = { workspace = true }
+tonic                      = { default-features = true, workspace = true }
```
drahnr, line 27: Any particular reason to add `default-features = true`?

bin/stress-test/Cargo.toml
```diff
@@ -31,5 +31,5 @@ miden-objects             = { workspace = true }
 rand                      = { workspace = true }
 rayon                     = { version = "1.10" }
 tokio                     = { workspace = true }
-tonic                     = { workspace = true }
+tonic                     = { default-features = true, workspace = true }
```
drahnr, line 34: Any particular reason to add `default-features = true`?

crates/block-producer/Cargo.toml
```diff
@@ -35,7 +35,7 @@ rand                       = { version = "0.9" }
 thiserror                  = { workspace = true }
 tokio                      = { features = ["macros", "net", "rt-multi-thread"], workspace = true }
 tokio-stream               = { features = ["net"], workspace = true }
-tonic                      = { features = ["transport"], workspace = true }
+tonic                      = { default-features = true, features = ["transport"], workspace = true }
```
drahnr, line 38: Any particular reason to add `default-features = true`?

sergerad (approved):

Mirko-von-Leipzig (approved):
Thanks, just the questions from @drahnr to address.

I think I recall reading that `default-features = true` is the default for workspace inheritance? But its a bit foggy.

crates/proto/Cargo.toml
SantiagoPittella: ~~Not really, will remove them (this and the others that you mentioned)~~

Edit: replying in a new comment at the bottom

SantiagoPittella:
In the rust book, it is say that I shouldn't be able to use `default-features` when using a dep from workspace 😅 : https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

But in practice, and since features are additive, if I use `default-features = false` in the root's Cargo.toml and `default-feature = true` in the crate it uses the features for that specific crate. But if do not set default features or set it to true in root toml, and try to use `default-features = false` in the crate, then I get an error:

```
Caused by:
  `default-features = false` cannot override workspace's `default-features`
```

And to be able to use the workspace version of `prost` I need to disable default features, forcing to then use `default-features = true` in basically every other crate.

Another solution could be avoid using the workspace prost and import it directly from the that crate toml with the version.

cc @drahnr @Mirko-von-Leipzig

Issue #1408, opened 2025-12-01
Block Producer: add chain tip to status endpoint


SantiagoPittella:
Not for this PR, but it may be good to add something like `chain_tip` to this message so that we can see in the monitor what the block producers thinks is the current chain tip (in case it somehow diverges from the store's chain tip).

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-node/pull/1388#discussion_r2572607578_

PR #1419, opened 2025-12-02
feat: add chain tip to block producer status

SantiagoPittella:
closes https://github.com/0xMiden/miden-node/issues/1408

This PR adds the `chain_tip` value to the Block Producer status response. Also, it displays it in the network monitor and improves/rearranges the "RPC" card of the network monitor.

This is the current look:
image

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,6 +16,10 @@ concurrency:
 permissions:
   contents: read
 
+env:
+  # Reduce cache usage by removing debug information.
+  CARGO_PROFILE_DEV_DEBUG: 0
+
 jobs:
   typos:
     runs-on: Linux-ARM64-Runner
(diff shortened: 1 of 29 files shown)


crates/block-producer/src/server/mod.rs
```diff
@@ -360,9 +363,10 @@ impl BlockProducerRpcServer {
             loop {
                 interval.tick().await;
 
-                let (unbatched_transactions, proposed_batches, proven_batches) = {
+                let (chain_tip, unbatched_transactions, proposed_batches, proven_batches) = {
                     let mempool = mempool.lock().await;
                     (
+                        mempool.chain_tip(),
```
Mirko-von-Leipzig, line 369: I wonder if using a watch channel wouldn't be a better fit for this.

Though it would mean that the mempool is then responsible for submitting updates whenever things change.

PR #1421, opened 2025-12-03
ci: disable debug info

Mirko-von-Leipzig:
This disables `debuginfo` for rust builds within CI workflows in the hopes of reducing build time and size and in turn easing cache usage.

Local testing shows quite a large reduction in build artifact size, but we'll have to see how much is actually cached.

Using our `check-msrv.sh` for example:

```
before  after  path
4.0K    4.0K   target/debug/examples
15M     15M    target/debug/.fingerprint
140M    95M    target/debug/incremental
1.7G    1.7G   target/x86_64-unknown-linux-gnu
1.7G    1.7G   target/x86_64-unknown-linux-gnu/debug
2.5G    476M   target/debug/build
3.8G    1.8G   target/debug/deps
6.4G    2.3G   target/debug
8.0G    4.0G   target/
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,6 +16,10 @@ concurrency:
 permissions:
   contents: read
 
+env:
+  # Reduce cache usage by removing debug information.
+  CARGO_PROFILE_DEV_DEBUG: 0
+
 jobs:
   typos:
     runs-on: Linux-ARM64-Runner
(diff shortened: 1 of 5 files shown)


drahnr:
Small caveat: Test panics are now a lot less meaningful

Mirko-von-Leipzig:
> Small caveat: Test panics are now a lot less meaningful

Let's see how it goes. It really depends; maybe we need to tweak it for only deps, or we can try splitting the debug info but I suspect that's still a cache issue.

PR #1422, opened 2025-12-03
fix: make get account proof retrieve latest known state

juan518munoz:
As observed while updating the client with the latest node version, retrieval of account proof is prone to failure when the block number used for the request is different from the block where the last account update happened, as it returns a _not found_ in those cases.

This PRs lighly refactors the logic of proof retrieval to return the account proof of an account **up to** the specified block.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,13 +15,13 @@
 - Added support for caching mempool statistics in the block producer server ([#1388](https://github.com/0xMiden/miden-node/pull/1388)).
 - Added mempool statistics to the block producer status in the `miden-network-monitor` binary ([#1392](https://github.com/0xMiden/miden-node/pull/1392)).
 - Add `S` generic to `NullifierTree` to allow usage with `LargeSmt`s ([#1353](https://github.com/0xMiden/miden-node/issues/1353)).
-- Modify `AccountProofRequest` to retrieve the latest known state in case specified block number (or chain tip) does not contain account updates ([#1422](https://github.com/0xMiden/miden-node/issues/1422)).
 
 ### Fixes
 
 - RPC client now correctly sets `genesis` value in `ACCEPT` header if `version` is unspecified ([#1370](https://github.com/0xMiden/miden-node/pull/1370)).
 - Pin protobuf (`protox`) dependencies to avoid breaking changes in transitive dependency ([#1403](https://github.com/0xMiden/miden-node/pull/1403)).
 - Fixed no-std compatibility for remote prover clients ([#1407](https://github.com/0xMiden/miden-node/pull/1407)).
+- Fixed `AccountProofRequest` to retrieve the latest known state in case specified block number (or chain tip) does not contain account updates ([#1422](https://github.com/0xMiden/miden-node/issues/1422)).
 
 ## v0.12.6
 
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -408,7 +408,7 @@ impl Db {
             .await
     }
 
-    /// Loads account details up to a specific block number from the DB.
+    /// Loads account details at a specific block number from the DB.
     #[instrument(level = "debug", target = COMPONENT, skip_all, ret(level = "debug"), err)]
     pub async fn select_historical_account_at(
         &self,
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -921,7 +921,7 @@ impl State {
     /// Returns the respective account proof with optional details, such as asset and storage
     /// entries.
     ///
-    /// When `block_num` is provided, this method will return the account state up to that specific
+    /// When `block_num` is provided, this method will return the account state at that specific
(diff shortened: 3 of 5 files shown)


igamigo (approved):
LGTM but I was also involved in investigating a fix for this so ideally @drahnr can take a look as well

proto/proto/store/rpc.proto
```diff
@@ -149,6 +149,8 @@ message AccountProofRequest {
 
     // Block at which we'd like to get this data. If present, must be close to the chain tip.
     // If not present, data from the latest block will be returned.
+    // If the specified block does not contain an update for the specified account,
+    // the latest available data will be returned.
```
Mirko-von-Leipzig: I think this is a bit ambiguous and could mean that asking for block `N` could return data after `N` aka `latest > N`.

I think phrasing the entire comment as something like

```
// Optional block height at which to return the proof.
//
// Defaults to current chain tip if unspecified.
```

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Added support for caching mempool statistics in the block producer server ([#1388](https://github.com/0xMiden/miden-node/pull/1388)).
 - Added mempool statistics to the block producer status in the `miden-network-monitor` binary ([#1392](https://github.com/0xMiden/miden-node/pull/1392)).
 - Add `S` generic to `NullifierTree` to allow usage with `LargeSmt`s ([#1353](https://github.com/0xMiden/miden-node/issues/1353)).
+- Modify `AccountProofRequest` to retrieve the latest known state in case specified block number (or chain tip) does not contain account updates ([#1422](https://github.com/0xMiden/miden-node/issues/1422)).
```
Mirko-von-Leipzig: I would argue this is a fix since I would have expected it to work this way to begin with. Since this is a fix for an unreleased part of code I would remove the changelog entry.

crates/store/src/state.rs
```diff
@@ -921,7 +921,7 @@ impl State {
     /// Returns the respective account proof with optional details, such as asset and storage
     /// entries.
     ///
-    /// When `block_num` is provided, this method will return the account state at that specific
+    /// When `block_num` is provided, this method will return the account state up to that specific
```
Mirko-von-Leipzig: I think the original was fine; just not implemented correctly

crates/store/src/state.rs
```diff
@@ -977,7 +977,7 @@ impl State {
         Ok((block_num, witness))
     }
 
-    /// Fetches the account details (code, vault, storage) for a public account at the specified
+    /// Fetches the account details (code, vault, storage) for a public account up to the specified
```
Mirko-von-Leipzig: Same as above; I think original wording is fine

crates/store/src/db/mod.rs
```diff
@@ -408,7 +408,7 @@ impl Db {
             .await
     }
 
-    /// Loads account details at a specific block number from the DB.
+    /// Loads account details up to a specific block number from the DB.
```
Mirko-von-Leipzig: Original `at` is fine

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -93,8 +93,8 @@ pub(crate) fn select_account(
     Ok(info)
 }
 
-/// Select account details at a specific block number from the DB using the given
-/// [`SqliteConnection`].
+/// Select account details at a block number lower or equal than specified from the DB using the
+/// given [`SqliteConnection`].
```
Mirko-von-Leipzig:
```suggestion
/// Select account details as they are at the given block height.
```

Mirko-von-Leipzig (commented):
I'll leave @drahnr to formally approve, but LGTM.

Some suggestions on returning to the original phrasing and removing the changelog entry.

crates/store/src/state.rs
drahnr: +1

crates/store/src/db/mod.rs
drahnr: +1 for `at`

drahnr (approved):

CHANGELOG.md
juan518munoz: Addressed in 04725803b830d6169ad58a0f98ef2c627eea2653

Also renamed the PR accordingly (`fix:`)

crates/store/src/state.rs
juan518munoz: Addressed in 04725803b830d6169ad58a0f98ef2c627eea2653

crates/store/src/state.rs
juan518munoz: Addressed in 04725803b830d6169ad58a0f98ef2c627eea2653

crates/store/src/db/mod.rs
juan518munoz: Addressed in 04725803b830d6169ad58a0f98ef2c627eea2653

PR #1424, opened 2025-12-03
chore: remove internal errors from network monitor

### LINKED ISSUE #1400: Network Monitor: add button to copy NTX error
Recently, we had errors while submitting the network transaction to the node, but the error from the node wasn't being display and the user had no idea what was happening.

We should add a copy button similar to the other cards, that allows to copy the error received from the node.

SantiagoPittella:
closes #1400


Mirko-von-Leipzig (approved):
I'm wondering if we shouldn't begin abstracting some of these behaviors in a trait.

But this isn't a priority so maybe let's just wait and see if patterns emerge.

PR #1425, opened 2025-12-03
Bring up `miden-base` dependencies to the latest `next`

bobbinth:
This PR brings up `miden-base` dependencies to the latest `next`.

### DIFF
--- a/crates/block-producer/src/test_utils/proven_tx.rs
+++ b/crates/block-producer/src/test_utils/proven_tx.rs
@@ -108,7 +108,7 @@ impl MockProvenTxBuilder {
             .map(|index| {
                 let nullifier = Word::from([ONE, ONE, ONE, Felt::new(index)]);
 
-                Nullifier::from(nullifier)
+                Nullifier::new_unchecked(nullifier)
             })
             .collect();
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -338,9 +338,9 @@ checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
 
 [[package]]
 name = "bech32"
-version = "0.11.0"
+version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d965446196e3b7decd44aa7ee49e31d630118f90ef12f97900f262eb915c951d"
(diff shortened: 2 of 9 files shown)


crates/block-producer/src/test_utils/proven_tx.rs
```diff
@@ -108,7 +108,7 @@ impl MockProvenTxBuilder {
             .map(|index| {
                 let nullifier = Word::from([ONE, ONE, ONE, Felt::new(index)]);
 
-                Nullifier::from(nullifier)
+                Nullifier::new_unchecked(nullifier)
```
bobbinth, line 111: I don't really love using `new_unchecked()` as it makes it feel like we are doing something dangerous, while in fact any word can be converted to a nullifier.

@drahnr, @PhilippGackstatter - I believe this resulted from the word-wrapper macro changes. Is there a better way to do this?

sergerad (approved):

crates/block-producer/src/test_utils/proven_tx.rs
PhilippGackstatter: Good point. I think we can revive `From<Word> for Nullifier`. Opened https://github.com/0xMiden/miden-base/pull/2134.

Mirko-von-Leipzig: > while in fact any word can be converted to a nullifier.

Is that true though? This isn't about ensuring byte/bit accuracy but rather that the type system enforces correctness by construction.

Its equivalent to the difference between a pointer and a reference. A reference is much safer _because_ it can only be constructed from a struct and _cannot_ be created by casting a random integer.

In a similar vein, I would expect "safe" nullifiers to originate from other types only e.g. from txs, batches, blocks or directly from its note. Inside these structs we would cast the word using `new_unchecked()` if we don't store the `Nullifier` directly, but this is abstracted away internally.

imo this is exactly correct.. we will have `new_unchecked` in a bunch of places _internally_ to structs where we know the usage is correct.

Mirko-von-Leipzig: Though maybe the wording `new_unchecked` is incorrect, and could be something like `from_raw`. But definitely no `From<Word>` should exist imo.

PR #1426, opened 2025-12-03
feat: Validator block signatures

sergerad:
## Context

The Validator provides an endpoint for signing blocks. This endpoint has so far been stubbed out and needs to be implemented.

The Validator and bootstrap commands need to be configured with ECDSA signers to sign blocks. For now we will use insecure, locally stored secret keys for development and testing while providing an abstraction to facilitate a secure backend for signing in the future.

Relates to #1316.

## Changes

- Update bootstrap command and genesis state to allow genesis block to be signed by validator key.
  - Genesis config TOML is no longer optional (must provide key that is used in subsequent start command).
- Add signer configuration to Validator.
- Add block signing logic to Validator.
- Update Validator `SignBlock` RPC endpoint to return ECDSA `Signature`.
- Add validator key field to block header proto.
- Refactor Dockerfile to not bootstrap at build time.

### DIFF
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -25,6 +25,7 @@ use crate::commands::{
     ENV_ENABLE_OTEL,
     ENV_GENESIS_CONFIG_FILE,
     ENV_VALIDATOR_INSECURE_SECRET_KEY,
+    INSECURE_VALIDATOR_KEY_HEX,
     NtxBuilderConfig,
     duration_to_human_readable_string,
 };
@@ -49,12 +50,15 @@ pub enum BundledCommand {
         #[arg(long, env = ENV_GENESIS_CONFIG_FILE, value_name = "FILE")]
         genesis_config_file: PathBuf,
         /// Insecure, hex-encoded validator secret key for development and testing purposes.
+        ///
+        /// If not provided, a predefined key is used.
         #[arg(
             long = "validator.insecure.secret-key",
             env = ENV_VALIDATOR_INSECURE_SECRET_KEY,
-            value_name = "VALIDATOR_INSECURE_SECRET_KEY"
+            value_name = "VALIDATOR_INSECURE_SECRET_KEY",
+            default_value = INSECURE_VALIDATOR_KEY_HEX
         )]
-        validator_insecure_secret_key: Option<String>,
+        validator_insecure_secret_key: String,
     },
 
     /// Runs all three node components in the same process.
@@ -133,12 +137,9 @@ impl BundledCommand {
                 grpc_timeout,
                 validator_insecure_secret_key,
             } => {
-                let validator_insecure_secret_key = validator_insecure_secret_key.context(
-                    "insecure validator secret key is required until other secret key backends are supported"
-                )?;
-                let signer = SecretKey::read_from_bytes(
-                    hex::decode(validator_insecure_secret_key)?.as_ref(),
-                )?;
+                let secret_key_hex =
+                    validator_insecure_secret_key.unwrap_or(INSECURE_VALIDATOR_KEY_HEX.into());
+                let signer = SecretKey::read_from_bytes(hex::decode(secret_key_hex)?.as_ref())?;
                 Self::start(
                     rpc_url,
                     data_directory,
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -20,6 +20,7 @@ use crate::commands::{
     ENV_ENABLE_OTEL,
     ENV_GENESIS_CONFIG_FILE,
     ENV_VALIDATOR_INSECURE_SECRET_KEY,
+    INSECURE_VALIDATOR_KEY_HEX,
     duration_to_human_readable_string,
 };
 
@@ -43,12 +44,15 @@ pub enum StoreCommand {
         #[arg(long, env = ENV_GENESIS_CONFIG_FILE, value_name = "GENESIS_CONFIG")]
         genesis_config_file: PathBuf,
         /// Insecure, hex-encoded validator secret key for development and testing purposes.
+        ///
+        /// If not provided, a predefined key is used.
         #[arg(
             long = "validator.insecure.secret-key",
             env = ENV_VALIDATOR_INSECURE_SECRET_KEY,
-            value_name = "VALIDATOR_INSECURE_SECRET_KEY"
+            value_name = "VALIDATOR_INSECURE_SECRET_KEY",
+            default_value = INSECURE_VALIDATOR_KEY_HEX
         )]
-        validator_insecure_secret_key: Option<String>,
+        validator_insecure_secret_key: String,
     },
 
     /// Starts the store component.
@@ -179,13 +183,10 @@ impl StoreCommand {
         data_directory: &Path,
         accounts_directory: &Path,
         genesis_config: &PathBuf,
-        validator_insecure_secret_key: Option<String>,
+        validator_insecure_secret_key: String,
     ) -> anyhow::Result<()> {
         // Decode the validator key.
-        let secret_key_hex = validator_insecure_secret_key.context(
-            "insecure validator secret key is required until other secret key backends are supported"
-        )?;
-        let signer = SecretKey::read_from_bytes(&hex::decode(secret_key_hex)?)?;
+        let signer = SecretKey::read_from_bytes(&hex::decode(validator_insecure_secret_key)?)?;
 
         // Read the toml.
         let toml_str = fs_err::read_to_string(genesis_config)?;
--- a/bin/node/src/commands/mod.rs
+++ b/bin/node/src/commands/mod.rs
@@ -15,6 +15,10 @@ pub mod rpc;
 pub mod store;
 pub mod validator;
 
+/// A predefined, insecure validator key for development purposes.
+const INSECURE_VALIDATOR_KEY_HEX: &str =
+    "0101010101010101010101010101010101010101010101010101010101010101";
+
 const ENV_BLOCK_PRODUCER_URL: &str = "MIDEN_NODE_BLOCK_PRODUCER_URL";
 const ENV_VALIDATOR_URL: &str = "MIDEN_NODE_VALIDATOR_URL";
 const ENV_BATCH_PROVER_URL: &str = "MIDEN_NODE_BATCH_PROVER_URL";
--- a/bin/node/src/commands/validator.rs
+++ b/bin/node/src/commands/validator.rs
@@ -12,6 +12,7 @@ use crate::commands::{
     ENV_ENABLE_OTEL,
     ENV_VALIDATOR_INSECURE_SECRET_KEY,
     ENV_VALIDATOR_URL,
+    INSECURE_VALIDATOR_KEY_HEX,
     duration_to_human_readable_string,
 };
 
@@ -40,8 +41,10 @@ pub enum ValidatorCommand {
         grpc_timeout: Duration,
 
         /// Insecure, hex-encoded validator secret key for development and testing purposes.
-        #[arg(long = "insecure.secret-key", env = ENV_VALIDATOR_INSECURE_SECRET_KEY, value_name = "INSECURE_SECRET_KEY")]
-        insecure_secret_key: Option<String>,
+        ///
+        /// If not provided, a predefined key is used.
+        #[arg(long = "insecure.secret-key", env = ENV_VALIDATOR_INSECURE_SECRET_KEY, value_name = "INSECURE_SECRET_KEY", default_value = INSECURE_VALIDATOR_KEY_HEX)]
+        insecure_secret_key: String,
     },
 }
 
@@ -51,14 +54,9 @@ impl ValidatorCommand {
             url, grpc_timeout, insecure_secret_key, ..
         } = self;
 
-        let insecure_secret_key = insecure_secret_key.context(
-            "insecure secret key is required until other secret key backends are supported",
-        )?;
-
         let address =
             url.to_socket().context("Failed to extract socket address from validator URL")?;
 
-        // Read secret key.
         let signer = SecretKey::read_from_bytes(hex::decode(insecure_secret_key)?.as_ref())?;
 
         Validator { address, grpc_timeout, signer }


proto/proto/validator.proto
```diff
@@ -33,14 +33,12 @@ message ValidatorStatus {
...
 
-// Response message for SignBlock RPC.
-message SignedBlock {
-    // The block header.
-    blockchain.BlockHeader header = 1;
-
-    // The block body.
-    blockchain.BlockBody body = 2;
+// ECDSA Signature.
+message Signature {
```
bobbinth: nit: I'd probably name this `BlockSignature` or something like that.

kmk142789 (approved):

proto/proto/types/primitives.proto
```diff
@@ -92,3 +92,13 @@ message Digest {
     fixed64 d2 = 3;
     fixed64 d3 = 4;
 }
+
+// SIGNATURE
+// ================================================================================================
+
+// ECDSA Signature.
+message Signature {
```
bobbinth: I would either call this `BlockSignature` (and move to the appropriate module) or `EcdsaSignature`. The former is probably better as it is more "type-safe".

proto/proto/types/blockchain.proto
```diff
@@ -79,6 +79,16 @@ message BlockHeader {
     fixed32 timestamp = 12;
 }
 
+// PUBLIC KEY
+// ================================================================================================
+
+// ECDSA public key.
+message PublicKey {
```
bobbinth: I would rename this to `ValidatorPublicKey`.

crates/store/src/genesis/config/samples/01-simple.toml
```diff
@@ -28,3 +28,6 @@ storage_mode = "private"
 [[wallet]]
 assets       = [{ amount = 1, symbol = "WHAT" }]
 storage_mode = "private"
+
+[signer]
+Insecure = "/tmp/insecure_2"
```
bobbinth: This is for specifying validator private key, right? If so, I'd change the section to `validator` and allow specifying either private key or public key - e.g.,

```toml
[validator]
private_key = "hex-encoded-private-key"

# or

[validator]
public_key = "hex-encoded-public-key"
```

Then, for dev/test scenarios, the users would be able to specify the private key, and for production scenarios, we could set the public key. Setting both at the same time would be an error.

Also, we should document this (I think we have a section that describes how the genesis file is to be used.

Lastly, I'd move this section up to be before `native_faucet`.

bin/node/src/commands/bundled.rs
```diff
@@ -82,6 +84,15 @@ pub enum BundledCommand {
...
         grpc_timeout: Duration,
+
+        /// Filepath to the insecure validator secret key for signing blocks.
+        ///
+        /// Only used in development and testing environments.
+        #[arg(
+            long = "validator-secret-key-filepath",
+            value_name = "VALIDATOR_SECRET_KEY_FILEPATH"
+        )]
+        validator_secret_key_filepath: Option<PathBuf>,
```
bobbinth: Related to the previous comment, I'm not sure this needs to be a filepath. A secret key is just a 32-byte string that could be hex encoded. But maybe keeping it as a file simplifies things?

crates/store/src/genesis/config/mod.rs
```diff
@@ -49,12 +77,13 @@ mod tests;
...
-    fee_parameters: FeeParameterConfig,
-    wallet: Vec<WalletConfig>,
-    fungible_faucet: Vec<FungibleFaucetConfig>,
+    pub version: u32,
+    pub timestamp: u32,
+    pub native_faucet: NativeFaucet,
+    pub fee_parameters: FeeParameterConfig,
+    pub wallet: Vec<WalletConfig>,
+    pub fungible_faucet: Vec<FungibleFaucetConfig>,
+    pub signer: SignerConfig,
```
bobbinth: Related to some other comments: I would rename this to `validator_config: ValidatorConfig`.

crates/store/src/genesis/config/mod.rs
```diff
@@ -41,6 +46,29 @@ use self::errors::GenesisConfigError;
...
+    /// secret key is stored in the specified filepath.
+    pub fn signer(&self) -> impl BlockSigner + Clone + use<> {
+        match self {
+            SignerConfig::Insecure(key_filepath) => {
+                let secret_key = SecretKey::new();
+                let mut file = File::create(key_filepath).expect("secret key file can be created");
+                file.write_all(&secret_key.to_bytes()).expect("secret key can be written to");
+                secret_key
+            },
+        }
```
bobbinth: As I mentioned in one of the comments, I think we may be able to put the keys into the config file itself rather than use separate files. The secret key is 32 bytes, and the public key, I believe, is 33 bytes.

crates/store/src/genesis/config/mod.rs
```diff
@@ -94,14 +127,18 @@ impl GenesisConfig {
     ///
     /// Also returns the set of secrets for the generated accounts.
     #[allow(clippy::too_many_lines)]
-    pub fn into_state(self) -> Result<(GenesisState, AccountSecrets), GenesisConfigError> {
+    pub fn into_state<S>(
+        self,
+        signer: S,
+    ) -> Result<(GenesisState<S>, AccountSecrets), GenesisConfigError> {
```
bobbinth, line 100: Why do we need to parametrize this with the signer?

bobbinth (commented):
Looks good! Thank you! Not a full review, but I left some comments/questions inline.

crates/store/src/server/mod.rs
```diff
@@ -50,7 +51,10 @@ impl Store {
         skip_all,
         err,
     )]
-    pub fn bootstrap(genesis: GenesisState, data_directory: &Path) -> anyhow::Result<()> {
+    pub fn bootstrap<S: BlockSigner>(
+        genesis: GenesisState<S>,
```
Mirko-von-Leipzig, line 55: Do we not have a concrete implementation for this?

crates/validator/src/server/mod.rs
Mirko-von-Leipzig, line 1: Similar to elsewhere, can we not choose a concrete `BlockSigner` implementation? I'd prefer not having generics flying about unless we can help it.

bin/node/src/commands/bundled.rs
Mirko-von-Leipzig: Given that this is only intended for devnet/local testing I think taking it inline is good enough.

What we can do is give it a more specific name e.g.

```
validator.development.secret_key
```

bin/node/src/commands/bundled.rs
```diff
@@ -110,9 +121,22 @@ impl BundledCommand {
...
                 grpc_timeout,
+                validator_secret_key_filepath,
             } => {
-                Self::start(rpc_url, data_directory, ntx_builder, block_producer, grpc_timeout)
-                    .await
+                let Some(validator_secret_key_filepath) = validator_secret_key_filepath else {
+                    return Err(anyhow::anyhow!(
+                        "secret_key_filepath is required until other secret key backends are supported"
+                    ));
+                };
```
Mirko-von-Leipzig: jfyi `anyhow::Context` is implemented for `Option` so you can do:

```suggestion
                let validator_secret_key_filepath = validator_secret_key_filepath
                    .context("secret_key_filepath is required until other secret key backends are supported")?;
```

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -230,44 +231,31 @@ impl BlockBuilder {
         &self,
         proposed_block: ProposedBlock,
         block_inputs: BlockInputs,
-    ) -> Result<(OrderedBatches, BlockInputs, BlockHeader, BlockBody), BuildBlockError> {
+    ) -> Result<(OrderedBatches, BlockInputs, BlockHeader, Signature, BlockBody), BuildBlockError>
```
Mirko-von-Leipzig, line 233: Its probably time to factor these into a `struct`

crates/ntx-builder/src/state/tests.rs
```diff
@@ -14,18 +15,18 @@ use crate::store::StoreClient;
...
+        1_u8.into(),                   // version
+        Word::default(),               // prev_hash
+        0_u32.into(),                  // block_num (genesis)
+        Word::default(),               // chain_root
+        Word::default(),               // account_root
+        Word::default(),               // nullifier_root
+        Word::default(),               // note_root
+        Word::default(),               // tx_hash
+        Word::default(),               // kernel_root
+        SecretKey::new().public_key(), // validator_key
```
Mirko-von-Leipzig: Idle question: how expensive is generating a secret key?

Mirko-von-Leipzig (commented):
Thanks, looks good in general. The main contention is whether we truly need to be generic over the signer and whether we need a secret key file at all.

crates/ntx-builder/src/state/tests.rs
bobbinth: Secrete key is just a random 32-byte string, I believe. So, should be pretty fast. public key generation is a bit more complicated (we need to do elliptic curve operations), but I think it is also fast (like a few dozen microseconds).

proto/proto/types/primitives.proto
sergerad: I went with EcdsaSignature because it feels odd specifying it as a block signature when the encoding is general ecdsa content

crates/store/src/genesis/config/samples/01-simple.toml
sergerad: Providing the public key alone would be insufficient for performing bootstrap - an entire signer needs to be provided in order to sign the genesis block.

proto/proto/types/primitives.proto
bobbinth: The only concern I have is that ECDSA is not specific enough - i.e., we also need to specify the curve and the hash function used - e.g., something like `EcdsaK256KeccakSignature` but that becomes quite verbose.

crates/store/src/genesis/config/samples/01-simple.toml
bobbinth: Ah yes - good point! I think we can do something like:

```toml
[validator]
private_key = "hex-encoded-private-key"

# or, for production settings:

[validator]
public_key = "hex-encoded-public-key"
block_signature = "hex-encoded-block-signature"
```

But we can start with just the `private_key` option for now, and add the production-specific option later.

sergerad: This approach also means you can no longer do a default bootstrap - you need to wire together the secret key from toml for bootstrap command and then into the subsequent start command via CLI flags

bin/node/src/commands/bundled.rs
sergerad: Have done this. Note that genesis TOML is no longer optional because you need to use the bootstrap key in subsequent node start command.

crates/store/src/genesis/config/mod.rs
sergerad: `GenesisState<S>::into_block()` performs signing

crates/store/src/server/mod.rs
sergerad: How would we support secure + insecure signers without this abstraction?

crates/validator/src/server/mod.rs
sergerad: Can use thread above to discuss https://github.com/0xMiden/miden-node/pull/1426#discussion_r2612230805

crates/store/src/genesis/config/samples/01-simple.toml
bobbinth: I see - so, we can use the private key file for both - genesis block construction and for node startup, right? In this case, probably fine to keep it as a file.

Another option is to output the private key file as a part of genesis construction procedure (similar to how we output the account files).

crates/store/src/server/mod.rs
bobbinth: Can we flash out different scenarios/steps we are trying to cover? For example, we have the dev/testing flow, this flow requires us to have the private key in 2 places:

- When generating the genesis block. Here we need the private key for two reasons:
  - To generate the validator's public key that goes into the block header.
  - To sign the resulting block.
- When starting up the bundled node. Here, we pass the private key to the validator so that it can sign the blocks produced by the block producer.

The production flow will probably be different from this:

- The genesis config file would be pre-populated with the public key of the validator and the signature over the block. Or maybe, we'd have an external component that generates the genesis block out of band (from the genesis config file), and then, the node bootstraps from the pre-built genesis block.
- When we start up the validator component, we'll need to provide the private key to it. There are probably a couple of options to do it:
  - The private key could be provided via an environment variable (probably not very secure).
  - There could be a separate signing services (though, the benefit of such a services is somewhat questionable because it would be doing just "dumb signing" and so, whoever controls the validator can get it to sign anything).
  - Are there other options?

sergerad: >When we start up the validator component, we'll need to provide the private key to it. There are probably a couple of options to do it:
>- The private key could be provided via an environment variable (probably not very secure).
>- There could be a separate signing services (though, the benefit of such a services is somewhat questionable because it would be doing just "dumb signing" and so, whoever controls the validator can get it to sign anything).

For the second point, the main thing is that the corresponding key would be retrieved securely from some remote backend.

>The genesis config file would be pre-populated with the public key of the validator and the signature over the block. Or maybe, we'd have an external component that generates the genesis block out of band (from the genesis config file), and then, the node bootstraps from the pre-built genesis block.

The out-of-band process should be as secure as the validator's process of signing blocks. Which would be an argument for using the same/similar approach that the validator uses above for creating the genesis block.

Having said that, I'm happy to remove the `Signer` trait usage from the bootstrap stack. I don't think we should remove it from the validator/bundled stack, however.

Mirko-von-Leipzig: You can remove it either by using an enum to represent signer options, or taking in a `dyn T`. At some point in the node you'll _need_ to pass in a concrete implementation so we will know what that is.

crates/store/src/genesis/config/samples/01-simple.toml
Mirko-von-Leipzig: I think we've gotten lost by focusing on the `toml` config file. What about keeping the current definition of the `toml` file - as an (optional) way to configure the _contents_ of the genesis block.

We then take in the private key via CLI separately. I would prefer this be in-line aka no key file - the caller can always use a file and echo it in via an arg. imo this is maximally flexible and alleviates us having to manage a file since the key can just be passed in via env var.

sergerad: Yea that makes sense. Will update

crates/store/src/server/mod.rs
sergerad: LMK if we want to go with enum or `dyn T`. If we go with the former then I will update base to remove the `pub trait BlockSigner {` altogether.

crates/store/src/genesis/mod.rs
```diff
@@ -129,25 +135,27 @@ impl GenesisState {
...
-        Ok(GenesisBlock(ProvenBlock::new_unchecked(header, body, block_proof)))
+        Ok(GenesisBlock(ProvenBlock::new_unchecked(header, body, signature, block_proof)))
     }
 }
 
 // SERIALIZATION
 // ================================================================================================
 
-impl Deserializable for GenesisState {
+impl<S: Deserializable> Deserializable for GenesisState<S> {
```
bobbinth: Do we actually need this code? I'm not seeing serialization code for `GenesisState` and without serialization, I'm not sure deserialization is meaningful

bin/node/.env
```diff
@@ -10,6 +10,7 @@ MIDEN_NODE_STORE_RPC_URL=
 MIDEN_NODE_STORE_NTX_BUILDER_URL=
 MIDEN_NODE_STORE_BLOCK_PRODUCER_URL=
 MIDEN_NODE_VALIDATOR_BLOCK_PRODUCER_URL=
+MIDEN_NODE_VALIDATOR_INSECURE_SECRET_KEY=
```
bobbinth, line 13: Not for this PR, but I find the naming a bit confusing here. For example, `MIDEN_NODE_VALIDATOR_BLOCK_PRODUCER_URL` is the URL at which the block producer can find the validator, right? If so, I think a more "intuitive" approach would be to name it `MIDEN_NODE_BLOCK_PRODUCER_VALIDATOR_URL`. Same goes for `MIDEN_NODE_STORE_BLOCK_PRODUCER_URL` etc.

crates/block-producer/src/validator/mod.rs
```diff
@@ -1,10 +1,9 @@
-use std::fmt::{Display, Formatter};
-
 use miden_node_proto::clients::{Builder, ValidatorClient};
-use miden_node_proto::errors::{ConversionError, MissingFieldHelper};
 use miden_node_proto::generated as proto;
-use miden_objects::block::{BlockBody, BlockHeader, ProposedBlock};
+use miden_objects::block::ProposedBlock;
+use miden_objects::crypto::dsa::ecdsa_k256_keccak::Signature;
 use miden_objects::utils::{Deserializable, Serializable};
+use miden_tx::utils::DeserializationError;
```
bobbinth: nit: `DeserializationError` can be imported from `miden_objects::utils`.

crates/store/src/genesis/config/mod.rs
```diff
@@ -49,12 +49,12 @@ mod tests;
...
-    fee_parameters: FeeParameterConfig,
-    wallet: Vec<WalletConfig>,
-    fungible_faucet: Vec<FungibleFaucetConfig>,
+    pub version: u32,
+    pub timestamp: u32,
+    pub native_faucet: NativeFaucet,
+    pub fee_parameters: FeeParameterConfig,
+    pub wallet: Vec<WalletConfig>,
+    pub fungible_faucet: Vec<FungibleFaucetConfig>,
 }
```
bobbinth, line 58: Do these fields need to be `pub`? It doesn't seem like we use this anywhere.

bin/node/src/commands/store.rs
```diff
@@ -38,7 +41,14 @@ pub enum StoreCommand {
...
         #[arg(long, env = ENV_GENESIS_CONFIG_FILE, value_name = "GENESIS_CONFIG")]
-        genesis_config_file: Option<PathBuf>,
+        genesis_config_file: PathBuf,
+        /// Insecure, hex-encoded validator secret key for development and testing purposes.
+        #[arg(
+            long = "validator.insecure.secret-key",
+            env = ENV_VALIDATOR_INSECURE_SECRET_KEY,
+            value_name = "VALIDATOR_INSECURE_SECRET_KEY"
+        )]
+        validator_insecure_secret_key: Option<String>,
```
bobbinth: Question: if this is required, should we make it just `String`, rather than `Option<String>`?

bin/node/src/commands/bundled.rs
```diff
@@ -82,6 +93,14 @@ pub enum BundledCommand {
...
         )]
         grpc_timeout: Duration,
+
+        /// Insecure, hex-encoded validator secret key for development and testing purposes.
+        #[arg(
+            long = "validator.insecure.secret-key",
+            env = ENV_VALIDATOR_INSECURE_SECRET_KEY,
+            value_name = "VALIDATOR_INSECURE_SECRET_KEY"
+        )]
+        validator_insecure_secret_key: Option<String>,
```
bobbinth, line 107: Same question re `Option<String>` as in the previous comment. Also relevant for line 57 above.

crates/store/src/genesis/mod.rs
```diff
@@ -27,11 +28,12 @@ pub mod config;
 
 /// Represents the state at genesis, which will be used to derive the genesis block.
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct GenesisState {
+pub struct GenesisState<S> {
     pub accounts: Vec<Account>,
     pub fee_parameters: FeeParameters,
     pub version: u32,
     pub timestamp: u32,
+    pub block_signer: S,
```
bobbinth, line 35: It is still not clear to me if the genesis state needs to be generic over `BlocSigner` - but I guess it doesn't create too much complexity.

crates/store/src/server/mod.rs
bobbinth: If it doesn't introduce more complexity, I'd probably go with `dyn BockSinger` approach. But keeping as is is probably fine too.

bin/node/.env
sergerad: https://github.com/0xMiden/miden-node/issues/1456

PR #1427, opened 2025-12-04
feat: add genesis regression tests

drahnr:
Adds a few basic regression tests to avoid foreign key constraint related errors, that we could have had caught earlier.

Related to https://github.com/0xMiden/miden-node/pull/1415

CI should fail until we merged 1415


crates/store/src/db/tests.rs
Mirko-von-Leipzig, line 1: Could we remove the `test_` prefix?

And otherwise maybe just some doc comments on what these are testing exactly so that in the future we can know if they're still of value.

igamigo (approved):
Merged `next` to the branch which contained the fixes and now the test is passing

PR #1394, opened 2025-11-28
feat: [2/4] integrate smtforest, avoid ser/de of full account/vault data in database

drahnr:
Part 2/4 of #1185 and lays the ground for https://github.com/0xMiden/miden-node/issues/1354

~:warning: The PR is rather large, a lot of the changes affect large pieces of code across.~

## Intent 

* Provide the baseline for cheap querying partial storage maps, but not implement it
* Prepare for potential deprecation of `AccountInfo`

## core changes

1. Remove the `vault` / `storage_map` BLOB entries from the accounts table.
2. Add `SmtForest` and integrate into `apply_block` and `State::load`

## significant changes in the following files:

* `crates/store/src/db/schema.rs` introduces `account_storage_headers` and removes `storage` (the full serialized account storage) from `accounts` table
* `crates/store/src/state.rs` / `fn State::apply_block` now updates the database, but also the lookup table of roots for the `SmtForest` and the entries in the forest itself, indirect lookup tables

## out of scope

* usage of the `LargeSmtForest` for partial storage maps - #670
* individual of partial storage map queries - #1354
* cleanup for both DB and `(Large|)SmtForest` - #1355 

## how to review

Take the existing TODOs into consideration, if they make sense. This will be the follow up PR.

PR #1428, opened 2025-12-04
feat: [3/4] partial storage map queries

drahnr:
Migration of https://github.com/0xMiden/miden-node/pull/1158

Targets #1394

Outstanding work:

Split https://github.com/0xMiden/miden-node/pull/1178 into a serialization and usage piece, which can replace the current naiive serialization to a byte blob, but use a full PartialSmt protobuf representation 

---

# Scope

Implements the API to query partial storage maps based on an naiive approach (collecting SmtProofs, one per leaf).

# Out of scope

Any optimization as outline by https://github.com/0xMiden/miden-node/pull/1178 - this is follow-up material.

### DIFF
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -448,36 +448,6 @@ impl AccountStorageMapDetails {
         }
     }
 
-    /// Creates storage map details based on the requested slot data.
-    ///
-    /// Handles both "all entries" and "specific keys" requests.
-    /// Returns `LimitExceeded` if too many entries.
-    pub fn new(slot_name: StorageSlotName, slot_data: SlotData, storage_map: &StorageMap) -> Self {
-        match slot_data {
-            SlotData::All => Self::from_all_entries(slot_name, storage_map),
-            SlotData::MapKeys(keys) => {
-                if keys.len() > Self::MAX_RETURN_ENTRIES {
-                    Self {
-                        slot_name,
-                        entries: StorageMapEntries::LimitExceeded,
-                    }
-                } else {
-                    // Query specific keys from the storage map - returns all entries without proofs
-                    // For proofs, use from_specific_keys with SmtForest
-                    let mut entries = Vec::with_capacity(keys.len());
-                    for key in keys {
-                        let value = storage_map.get(&key);
-                        entries.push((key, value));
-                    }
-                    Self {
-                        slot_name,
-                        entries: StorageMapEntries::AllEntries(entries),
-                    }
-                }
-            },
-        }
-    }
-
     /// Creates storage map details from forest-queried entries.
     ///
     /// Returns `LimitExceeded` if too many entries.


proto/proto/store/rpc.proto
Mirko-von-Leipzig, line 1: I think this file no longer exists and things were re-organized

proto/proto/store/rpc.proto
```diff
@@ -0,0 +1,527 @@
...
+
+            repeated StorageMapEntry entries = 1;
+        }
+        // slot index of the storage map
+        uint32 slot_index = 1;
+
+        // A flag that is set to `true` if the number of to-be-returned entries in the
+        // storage map would exceed a threshold. This indicates to the user that `SyncStorageMaps`
+        // endpoint should be used to get all storage map data.
+        bool too_many_entries = 2;
```
Mirko-von-Leipzig: Should this not be an error instead? I think we do have specific error codes now as part of the RPC "spec".

proto/proto/store/rpc.proto
```diff
@@ -0,0 +1,527 @@
...
+        oneof data {
+            // By default we provide all storage entries when `all_entries` is requested
+            // or when the storage map is small.
+            MapEntries entries = 3;
+
+            // When specific keys are requested and the storage map is not small,
+            // we provide a set of SMT proofs (openings) for the requested keys.
+            // This allows the receiver to reconstruct the partial tree or validate individual proofs.
+            SmtProofSet smt_proofs = 4;
+        }
```
Mirko-von-Leipzig: Maybe something like (though it still feels clunky). 

```suggestion
        oneof data {
            // Contains the full key-value entries of the map.
            //
            // This variant is returned if the map is small enough in size.
            //
            // This can be forced by setting `all_entries` in the request, but
            // note that this can return an error if the map has too many entries.
            // In which case the client can sync the full state using `SyncStorageMaps`.
            MapEntries full = 3;

            // Contains SMT proofs for the entries requested.
            SmtProofSet partial = 4;
        }
```

crates/store/src/inner_forest.rs
```diff
@@ -36,233 +29,191 @@ impl InnerForest {
...
-            account_id = %account_id,
-            block_num = %block_num,
-            vault_entries = assets.len(),
-            "Populated vault in forest from DB"
+            "Queried {} storage keys from forest for account {:?}, slot {} at block {}",
+            results.len(),
+            account_id,
+            slot_name,
+            block_num
         );
```
Mirko-von-Leipzig: I don't really care so much since its `debug!` but we should prefer structured logs i.e. the message portion should be a static str 99% of the time.

Similar above.

```suggestion
        tracing::debug!(
            target: crate::COMPONENT,
            keys.count = %results.len(),
            account.id = %account_id,
            slot.name = %slot_name,
            block.number = %block_num,
            "Queried storage keys from forest",
        );
```

crates/store/src/inner_forest.rs
```diff
@@ -36,233 +29,191 @@ impl InnerForest {
...
+                    // Extract the value from the proof
+                    let value = proof.get(key).unwrap_or(EMPTY_WORD);
+                    results.push((*key, value));
+                },
+                Err(e) => {
+                    tracing::debug!(
+                        target: crate::COMPONENT,
+                        "Failed to open proof for key in storage forest: {}. Using empty value.",
+                        e
+                    );
```
Mirko-von-Leipzig: When could this happen? Is this an expected thing?

crates/store/src/inner_forest.rs
```diff
@@ -36,233 +29,191 @@ impl InnerForest {
...
+        keys: &[Word],
+    ) -> Result<Vec<(Word, Word)>, String> {
+        // Get the storage root for this account/slot/block
+        let root = self
+            .storage_roots
+            .get(&(account_id, slot_name.clone(), block_num))
+            .copied()
+            .ok_or_else(|| {
+                format!(
+                    "Storage root not found for account {account_id:?}, slot {slot_name}, block {block_num}"
```
Mirko-von-Leipzig:
```suggestion
                    "storage root not found for account {account_id:?}, slot {slot_name}, block {block_num}"
```

crates/store/src/inner_forest.rs
```diff
@@ -36,233 +29,191 @@ impl InnerForest {
...
+        &self,
         account_id: AccountId,
-        assets: &[Asset],
+        slot_name: &StorageSlotName,
         block_num: BlockNumber,
-    ) {
-        if assets.is_empty() {
-            return;
+        keys: &[Word],
+    ) -> Result<Vec<(Word, Word)>, String> {
```
Mirko-von-Leipzig: Should this be a `Result<_, String>`? Should this not be a `DatabaseError` variant(s)?

crates/proto/src/domain/account.rs
```diff
@@ -498,20 +497,79 @@ impl AccountStorageMapDetails {
...
+    /// Storage map details containing the requested entries or `LimitExceeded` if too many keys.
+    ///
+    /// # Errors
+    ///
+    /// Returns `MerkleError` if the forest doesn't contain sufficient data to provide proofs.
+    pub fn from_specific_keys(
+        slot_name: StorageSlotName,
+        keys: &[Word],
+        storage_forest: &SmtForest,
+        smt_root: Word,
```
Mirko-von-Leipzig: I'm sort of surprised there isn't an `SmtTree<'a>` which you get from `SmtForest::open_tree(&self, root: Word) -> Option<SmtTree<'_>>`

crates/proto/src/domain/account.rs
```diff
@@ -498,20 +497,79 @@ impl AccountStorageMapDetails {
...
+    /// This method queries the forest for specific keys and extracts key-value pairs from
+    /// the SMT proofs. The forest must be available and contain the data for the specified
+    /// SMT root.
+    ///
+    /// # Arguments
+    ///
+    /// * `slot_name` - The name of the storage slot
+    /// * `keys` - The keys to query
+    /// * `storage_forest` - The SMT forest containing the storage data
+    /// * `smt_root` - The root of the SMT for this storage slot
```
Mirko-von-Leipzig: This feels really redundant

crates/proto/src/domain/account.rs
Mirko-von-Leipzig, line 1: I'm getting a bit lost in the sauce here. It feels like there is simply too many similar enums and structs with slight differences. And the file itself is huge which doesn't help.

proto/proto/store/rpc.proto
drahnr: I like that, will rename

crates/store/src/inner_forest.rs
drahnr: Theoretically this should never happen, once we move to a storage based `SmtForest`, we might hit IO errors. Structurally this should never happen right now iiuc.

proto/proto/types/primitives.proto
```diff
@@ -92,3 +92,33 @@ message Digest {
     fixed64 d2 = 3;
     fixed64 d3 = 4;
 }
+
+// PARTIAL SMT
+// ================================================================================================
+
+// Represents a partial Sparse Merkle Tree containing only a subset of leaves and their paths.
+// This allows verifying and updating tracked keys without requiring the full tree.
+message PartialSmt {
```
drahnr: todo: not used in this PR since the PartialSmt usage was deferred to a follow-up

crates/proto/src/domain/account.rs
drahnr: Even if there was, I wouldn't try to use it, since it'd be expensive once we'd move to `LargeSmtForest` hitting IO a lot for larger accounts

crates/store/src/inner_forest.rs
drahnr: Made it fail

proto/proto/store/rpc.proto
drahnr: @bobbinth this came from the original API sketch. I agree with @Mirko-von-Leipzig if we have a size error and a pattern to use it, we should stick to it.

proto/proto/store/rpc.proto
drahnr: It should go away with the next merge/rebase

crates/proto/src/domain/account.rs
drahnr: We should split this up for sure, but I think that's a contentious thing to fold into this PR.

Mirko-von-Leipzig (approved):

bobbinth:
I'm a bit unclear about the purpose of this PR: it introduces some new code but this code doesn't seem to be used anywhere - but maybe I'm missing something?

It is also not clear to me if we need these changes now. Would it make sense to first finish the refactoring and then make these types improvements after? Or does this help with the refactoring somehow?

crates/store/src/state.rs
```diff
@@ -1050,25 +1055,49 @@ impl State {
             None => AccountVaultDetails::empty(),
         };
 
-        // TODO: don't load the entire store at once, load what is required
-        let store = self.db.select_account_storage_at_block(account_id, block_num).await?;
-        let storage_header = store.to_header();
+        // Load storage header from DB (map entries come from forest)
+        let storage_header =
+            self.db.select_account_storage_header_at_block(account_id, block_num).await?;
```
bobbinth: `select_account_header_at_block()` on line 1032 above should bring back storage header as well now, right? If so, we can avoid going to the DB for the storage header again.

crates/store/src/inner_forest/mod.rs
```diff
@@ -82,6 +87,33 @@ impl InnerForest {
...
+    /// Returns `None` if no storage root is tracked for this account/slot/block combination.
+    pub(crate) fn storage_map_forest_with_root(
+        &self,
+        account_id: AccountId,
+        slot_name: &StorageSlotName,
+        block_num: BlockNumber,
+    ) -> Option<(&SmtForest, Word)> {
+        let root = self.storage_roots.get(&(account_id, slot_name.clone(), block_num))?;
+        Some((&self.forest, *root))
+    }
```
bobbinth: My understanding is that we use this to get `SmtProof`s from the forest later on, right? If so, this approach feels backwards to me. Is there a reason not to return all the proofs from here? For example, could this method be something like:
```rust
pub fn open_storage_map(
    &self,
    account_id: AccountId,
    slot_name: &StorageSlotName,
    block_num: BlockNumber,
    keys: Vec<Word>,
) -> Vec<SmtProof> {
    ...
}
```

crates/store/src/inner_forest/mod.rs
```diff
@@ -82,6 +87,33 @@ impl InnerForest {
...
+    /// Returns `None` if no entries are tracked for this account/slot/block combination.
+    pub(crate) fn storage_map_entries(
+        &self,
+        account_id: AccountId,
+        slot_name: &StorageSlotName,
+        block_num: BlockNumber,
+    ) -> Option<Vec<(Word, Word)>> {
+        let entries = self.storage_entries.get(&(account_id, slot_name.clone(), block_num))?;
+        Some(entries.iter().map(|(k, v)| (*k, *v)).collect())
+    }
```
bobbinth: Is this a temporary solution? I don't think we can put all entries for all storage maps into memory - how are you thinking of handling this in the future?

If we are able to provide this functionality in a sustainable way - we probably don't need database table to keep historical data - and this would simplify things quite a bit.

crates/store/src/state.rs
drahnr: Pending the merge-up, yes.

crates/store/src/inner_forest/mod.rs
drahnr: Will return an `Option<Result<Vec<SmtProof>>>` though or the wrapper type `Option<Result<AccountStorageDetails>>` directly.

crates/store/src/inner_forest/mod.rs
drahnr: There must be an earlier check if we exceed the maximum number of entries for the request of "all entries".

bobbinth (commented):
Looks good! Thank you. It seems like some of the changes from the latest review got reverted and some comments that were marked as resolved were not resolved. Other than that, I think we are pretty close.

bobbinth:
> Looks good! Thank you. It seems like some of the changes from the latest review got reverted and some comments that were marked as resolved were not resolved. Other than that, I think we are pretty close.

Actually - ignore this review from me as I reviewed before realizing that there are still some merge conflicts with the base branch. Let's resolve the conflicts and then I can review again.

crates/store/src/inner_forest/mod.rs
bobbinth: I think the bigger problem is that we just can't hold the `self.entries` map in memory. But I think we may be able get rid of this map with the data we get from `LargeSmtForest`. So, let's add this to the list off things to do in #1498.

crates/store/src/state.rs
drahnr: `select_account_header_at_block` became `select_account_header_with_storage_header_at_block`

crates/store/src/db/mod.rs
```diff
@@ -453,23 +453,6 @@ impl Db {
         .await
     }
 
-    /// Reconstructs account storage at a specific block from the database
-    ///
-    /// This method queries the decomposed storage tables and reconstructs the full
-    /// `AccountStorage` with SMT backing for Map slots.
-    // TODO split querying the header from the content
-    #[instrument(level = "debug", target = COMPONENT, skip_all, ret(level = "debug"), err)]
-    pub async fn select_account_storage_at_block(
```
bobbinth, line 462: There are still a couple of tests which seem to be testing `select_account_storage_at_block` functionality. Do we still need them?

crates/proto/src/domain/account.rs
```diff
@@ -969,3 +1108,173 @@ pub enum NetworkAccountError {
 fn get_account_id_tag_prefix(id: AccountId) -> AccountPrefix {
     (id.prefix().as_u64() >> 34) as AccountPrefix
 }
+
+#[cfg(test)]
+mod tests {
```
bobbinth: nit: could we add a "TEST" section header above these lines.

crates/proto/src/domain/account.rs
```diff
@@ -207,26 +215,49 @@ impl TryFrom<proto::rpc::account_storage_details::AccountStorageMapDetails>
...
+                                .key
+                                .ok_or(StorageMapEntryWithProof::missing_field(stringify!(key)))?
+                                .try_into()?;
+                            let _value: Word = entry
+                                .value
+                                .ok_or(StorageMapEntryWithProof::missing_field(stringify!(value)))?
+                                .try_into()?;
+                            let smt_opening = entry.proof.ok_or(
+                                StorageMapEntryWithProof::missing_field(stringify!(proof)),
+                            )?;
```
bobbinth: Question: why do we need to read `_key` and `_value` if they are never used? And more broadly, why do we need them in the struct?

crates/proto/src/domain/account.rs
bobbinth: nit (and not from this PR): the "account storage map details" section is defined below (under line 471) - so, it is a bit odd to have this impl here rather than in that section.

crates/proto/src/domain/account.rs
```diff
@@ -457,92 +489,139 @@ pub enum StorageMapEntries {
...
+    /// Creates storage map details based on the requested slot data.
+    ///
+    /// Handles both "all entries" and "specific keys" requests.
+    /// Returns `LimitExceeded` if too many entries.
     pub fn new(slot_name: StorageSlotName, slot_data: SlotData, storage_map: &StorageMap) -> Self {
         match slot_data {
             SlotData::All => Self::from_all_entries(slot_name, storage_map),
-            SlotData::MapKeys(keys) => Self::from_specific_keys(slot_name, &keys[..], storage_map),
+            SlotData::MapKeys(keys) => {
+                if keys.len() > Self::MAX_RETURN_ENTRIES {
```
bobbinth: Probably not for this PR, but the number of keys that we could return paths for should probably be significantly lower than `MAX_RETURN_ENTRIES`. The main reason is that `SmtProofs` are quite a bit heavier than individual entries. Let's create an issue for this.

crates/proto/src/domain/account.rs
```diff
@@ -457,92 +489,139 @@ pub enum StorageMapEntries {
...
+                entries: StorageMapEntries::AllEntries(entries),
+            }
+        }
+    }
+
+    /// Creates storage map details based on the requested slot data.
+    ///
+    /// Handles both "all entries" and "specific keys" requests.
+    /// Returns `LimitExceeded` if too many entries.
     pub fn new(slot_name: StorageSlotName, slot_data: SlotData, storage_map: &StorageMap) -> Self {
```
bobbinth: From what I can tell, this method is not actually being used. Should we delete it?

crates/proto/src/domain/account.rs
```diff
@@ -457,92 +489,139 @@ pub enum StorageMapEntries {
...
 
-    fn from_specific_keys(
+    /// Creates storage map details with SMT proofs for specific keys.
+    ///
+    /// This method queries the forest for specific keys and returns proofs that
+    /// enable client-side verification of the values.
+    ///
+    /// Returns `LimitExceeded` if too many keys, or `MerkleError` if the forest
+    /// doesn't contain sufficient data.
+    pub fn from_specific_keys(
```
bobbinth: This is used only in tests, right? If so, I'd mark it as such. Also, do we actually need it?

crates/proto/src/domain/account.rs
```diff
@@ -457,92 +489,139 @@ pub enum StorageMapEntries {
...
-                    too_many_entries: false,
-                    entries,
-                }
-            },
+    /// Creates storage map details from pre-computed SMT proofs.
+    ///
+    /// Use this when the caller has already obtained the proofs from an `SmtForest`.
+    /// Returns `LimitExceeded` if too many proofs are provided.
+    pub fn from_proofs(slot_name: StorageSlotName, proofs: Vec<SmtProof>) -> Self {
+        if proofs.len() > Self::MAX_RETURN_ENTRIES {
```
bobbinth, line 473: Probably not for this PR, but the number of proofs that we could return should probably be significantly lower than `MAX_RETURN_ENTRIES`. The main reason is that `SmtProofs` are quite a bit heavier than individual entries. Let's create an issue for this.

crates/proto/src/domain/account.rs
```diff
@@ -707,6 +786,66 @@ impl From<AccountDetails> for proto::rpc::account_proof_response::AccountDetails
     }
 }
 
+impl From<AccountStorageMapDetails>
+    for proto::rpc::account_storage_details::AccountStorageMapDetails
```
bobbinth: This should probably go in the "account storage map detail" section.

Ideally, in the future, we'd split this file into a couple of modules so that they are easier to navigate.

crates/proto/src/domain/account.rs
drahnr: I'll move it to a separate file.

crates/proto/src/domain/account.rs
drahnr: https://github.com/0xMiden/miden-node/issues/1514

crates/proto/src/domain/account.rs
drahnr: It's duplicating the logic of `open_storage_map`, so can be removed and adjusting the tests.

crates/proto/src/domain/account.rs
drahnr: Removed

crates/proto/src/domain/account.rs
drahnr: It's redundant and I adjusted the .proto file too.

crates/store/src/db/mod.rs
drahnr: Kept a subset, renamed that.

PR #1430, opened 2025-12-05
feat: track ntx latency

### LINKED ISSUE #1409: Network Monitor: track NTXs latency
We should research for a way to track latency for network transactions, we can define latency here as the difference between the timestamp of the block in which the transaction that included that specific network note and the timeout of the chain tip returned by the node when we submit the proven transaction that generates the network note.

SantiagoPittella:
closes #1409 

Latency uses blocks as units.

### DIFF
--- a/bin/network-monitor/src/counter.rs
+++ b/bin/network-monitor/src/counter.rs
@@ -271,7 +271,6 @@ pub async fn run_increment_task(
 
         {
             let guard = latency_state.lock().await;
-            details.pending_latency.clone_from(&guard.pending);
             details.last_latency_blocks = guard.last_latency_blocks;
         }
 
--- a/bin/network-monitor/src/status.rs
+++ b/bin/network-monitor/src/status.rs
@@ -80,8 +80,6 @@ pub struct IncrementDetails {
     pub last_tx_id: Option<String>,
     /// Last measured latency in blocks from submission to state update.
     pub last_latency_blocks: Option<u32>,
-    /// Information about the currently pending latency measurement (if any).
-    pub pending_latency: Option<PendingLatencyDetails>,
 }
 
 /// Details about an in-flight latency measurement.
--- a/bin/network-monitor/src/monitor/tasks.rs
+++ b/bin/network-monitor/src/monitor/tasks.rs
@@ -265,7 +265,6 @@ impl Tasks {
                 failure_count: 0,
                 last_tx_id: None,
                 last_latency_blocks: None,
-                pending_latency: None,
             }),
         };
 


bin/network-monitor/src/counter.rs
```diff
@@ -257,7 +288,7 @@ fn handle_increment_success(
     details: &mut IncrementDetails,
     tx_id: String,
     expected_counter_value: &Arc<AtomicU64>,
-) -> Result<Option<String>> {
+) -> Result<u64> {
```
igamigo, line 292: nit: Would be great to update the function's doc comment to mention what it returns since it's not clear at first sight

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@
 - [BREAKING] `SubmitProvenTransaction` now **requires** that the network's genesis commitment is set in the request's `ACCEPT` header ([#1298](https://github.com/0xMiden/miden-node/pull/1298)).
 - Add `S` generic to `NullifierTree` to allow usage with `LargeSmt`s ([#1353](https://github.com/0xMiden/miden-node/issues/1353)).
 - Removed internal errors from the `miden-network-monitor` ([#1424](https://github.com/0xMiden/miden-node/pull/1424)).
+- Added chain tip to the block producer status ([#1419](https://github.com/0xMiden/miden-node/pull/1419)).
+- Track network transactions latency in `miden-network-monitor` ([#1430](https://github.com/0xMiden/miden-node/pull/1430))
```
igamigo: nit for consistency:
```suggestion
- Track network transactions latency in `miden-network-monitor` ([#1430](https://github.com/0xMiden/miden-node/pull/1430)).
```

bin/network-monitor/src/counter.rs
```diff
@@ -405,17 +439,50 @@ pub async fn run_counter_tracking_task(
                     details.pending_increments = Some(0);
                 }
 
-                None
+                // Check for latency completion or timeout
+                let mut guard = latency_state.lock().await;
+                if let Some(pending) = guard.pending.clone() {
+                    if value >= pending.target_value {
+                        match fetch_chain_tip(&mut rpc_client).await {
```
igamigo: I think this could block the increment task from updating the status while the RPC call is still in flight because the `latency_state`'s guard is being held. I think you could take a snapshot of the state, drop the lock, and then continue on with the async work.

bin/network-monitor/src/status.rs
```diff
@@ -78,6 +78,19 @@ pub struct IncrementDetails {
     pub failure_count: u64,
     /// Last transaction ID (if available).
     pub last_tx_id: Option<String>,
+    /// Last measured latency in blocks from submission to state update.
+    pub last_latency_blocks: Option<u32>,
+    /// Information about the currently pending latency measurement (if any).
+    pub pending_latency: Option<PendingLatencyDetails>,
```
igamigo: Is this used for anything right now?

SantiagoPittella: No, i forgot to removed it.

PR #1433, opened 2025-12-08
feat: configurable mempool tx capacity

### LINKED ISSUE #1138: Configurable mempool limit

Node operator

The size of the mempool is currently unlimited. We should limit the capacity so that successfully submitted transactions will be committed in a block within a reasonable amount of time.

### How is this feature used?

Add a new configuration item to the `block-producer` component (and `bundled`) which configures the maximum tx capacity of the mempool. Once this limit is reached, new transactions should be rejected, ideally with an appropriate rate limit-esque error.

The default value should be such that the capacity would be cleared in ~1 minute.

### Anything else?

Consider whether the config item should be the actual capacity, or a maximum time to block estimate which we internally convert into the capacity. 

The latter might be easier to understand conceptually but one could argue the operator needs to know what the absolute capacity means. Capacity will interact with block interval and block capacity to give the time estimate, so setting the time estimate might be simpler from an operator perspective.

The mempool should also emit this capacity as part of its telemetry.

Mirko-von-Leipzig:
This PR adds a new `--mempool.tx-capacity` CLI argument which configures the maximum number of inflight (uncommitted) transactions are allowed in the mempool at any one time.

It defaults to `768` which is roughly the maximum number of transactions we can process within a minute based on a 5s block time if all the stars align and blocks are at capacity.

Closes #1138

### DIFF
--- a/crates/block-producer/src/lib.rs
+++ b/crates/block-producer/src/lib.rs
@@ -60,6 +60,7 @@ pub const DEFAULT_BATCH_INTERVAL: Duration = Duration::from_secs(2);
 ///
 /// The value is selected such that all transactions should approximately be processed within one
 /// minutes with a block time of 5s.
+#[allow(clippy::cast_sign_loss, reason = "Both durations are positive")]
 pub const DEFAULT_MEMPOOL_TX_CAPACITY: NonZeroUsize = NonZeroUsize::new(
     DEFAULT_MAX_BATCHES_PER_BLOCK
         * DEFAULT_MAX_TXS_PER_BATCH


CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 
+### Enhancements
+
+- Added support for timeouts in the WASM remote prover clients ([#1383](https://github.com/0xMiden/miden-node/pull/1383)).
+- Added block validation endpoint to validator and integrated with block producer ([#1382](https://github.com/0xMiden/miden-node/pull/1381)).
+- Added support for caching mempool statistics in the block producer server ([#1388](https://github.com/0xMiden/miden-node/pull/1388)).
+- Added mempool statistics to the block producer status in the `miden-network-monitor` binary ([#1392](https://github.com/0xMiden/miden-node/pull/1392)).
+- Added success rate to the `miden-network-monitor` binary ([#1420](https://github.com/0xMiden/miden-node/pull/1420)).
+- Added chain tip to the block producer status ([#1419](https://github.com/0xMiden/miden-node/pull/1419)).
+- The mempool's transaction capacity is now configurable ([#1433](https://github.com/0xMiden/miden-node/pull/1433)).
```
Mirko-von-Leipzig, line 13: I moved the changelog entries to enhancements where it makes sense.

I do think its confusing having changes & enhancements (as we can see by this).

Maybe `Added`, `Changed` and `Fixed` would be more obvious?

bobbinth: > Maybe `Added`, `Changed` and `Fixed` would be more obvious?

I think `Enhancements` was meant to be the same as `Added` and `Changes` was the same as a combination of `Changed` and `Fixed` (as frequently it is not trivial to categorize things across these two) - but I'm fine with `Added`, `Changed`, `Fixed` categorization as well.

crates/block-producer/src/lib.rs
```diff
@@ -50,6 +50,14 @@ const SERVER_MEMPOOL_EXPIRATION_SLACK: u32 = 2;
 /// The interval at which to update the cached mempool statistics.
 const CACHED_MEMPOOL_STATS_UPDATE_INTERVAL: Duration = Duration::from_secs(5);
 
+/// The default transaction capacity of the mempool.
+///
+/// The value is selected such that all transactions should approximately be processed within one
+/// minutes with a block time of 5s.
```
bobbinth, line 62: Not related to this PR, but I thought we were reducing block times to 2 seconds (didn't we have an issue for this?).

bin/node/src/commands/bundled.rs
```diff
@@ -215,6 +216,7 @@ impl BundledCommand {
                         max_txs_per_batch: block_producer.max_txs_per_batch,
                         production_checkpoint: checkpoint,
                         grpc_timeout,
+                        mempool_tx_capacity: NonZeroUsize::new(1000).unwrap(),
```
bobbinth: Should this not be set to `block_producer.mempool_tx_capacity`?

crates/block-producer/src/lib.rs
Mirko-von-Leipzig: Not that I'm aware of; I saw mumblings about it in chats but will create an issue.

The value is already configurable though but we can change the default to 2s.

bin/node/src/commands/bundled.rs
Mirko-von-Leipzig: Yes, thank you. For some reason I thought I was in a test setup here.

crates/block-producer/src/lib.rs
Mirko-von-Leipzig: This made me realize the mempool capacity can re-use the default block time instead if we move the `const`.

bin/node/src/commands/mod.rs
```diff
@@ -95,4 +100,12 @@ pub struct BlockProducerConfig {
...
     #[arg(long = "max-batches-per-block", env = ENV_MAX_BATCHES_PER_BLOCK, value_name = "NUM", default_value_t = DEFAULT_MAX_BATCHES_PER_BLOCK)]
     pub max_batches_per_block: usize,
+
+    #[arg(
+        long = "mempool.tx-capacity",
+        default_value_t = miden_node_block_producer::DEFAULT_MEMPOOL_TX_CAPACITY,
+        env = ENV_MEMPOOL_TX_CAPACITY,
+        value_name = "NUM"
+    )]
+    mempool_tx_capacity: NonZeroUsize,
```
Mirko-von-Leipzig, line 121: An alternative configuration would be to do it as a maximum time capacity.

i.e. `60s` and then the actual amount is derived from the configured block rate, and block/batch size.

I'll leave it as is for now.

CHANGELOG.md
Mirko-von-Leipzig: Its not trivial but I do think there is value because typically:

- `fixed` implies a patch version release since its a bugfix
- `changed` implies minor/major bump

But overall none of these really seem to match where we are now since things are moving fast.

Issue #1056, opened 2025-07-08
Network transaction builder enhancements


Mirko-von-Leipzig:
This is a tracking issue for enhancements to the network transaction builder. At this point it is unclear how best to tackle these as simply bolting them on top of the existing implementation is unlikely to result in good quality code.

It may be prudent to sketch a more robust design where these enhancements can be added. Its also possible that some of these are mutually exclusive.

This list is not exhaustive, nor are all items necessarily desired.

## Improvements

- [ ] Configurable parameters e.g. `MAX_NOTES_PER_TX`
- [ ] Allow for 3rd party network account updates
- [ ] Consider a persistence layer
- [ ] Reset component without taking down the node
- [ ] Lazy account loading and caching
- [ ] Allow for specialization of specific network accounts

## Design idea: actor-per-account

Explore the idea of having each account be represented by a single actor. This fits neatly with the overall Miden view of accounts, and even more so since network notes are specific to a single account.

There will be issues to tackle and design, primarily around error recovery e.g.
- how do we recover if the actor panics; how do we revive the ephemeral data without coupling the world
- when is it safe to shut down an actor
- how do we trigger network transactions
- should actors go to sleep if their notes warrant a lengthy delay, and how to wake them up again

bobbinth:
Question: how "Consider a persistence layer" and "Local persistence" different?

bobbinth:
Seems like all subtasks of this issue are done. Should we close it it? If there is anything outstanding, @Mirko-von-Leipzig - maybe it makes sense to open a more targeted issue?

Issue #1314, opened 2025-10-26
Improving actor design


tindzk:
I’ve been exploring Miden’s node implementation over the past few days and like its architecture, especially that it is designed around actors.

Based on the code I have read so far, I would suggest a few changes to the current actor design. When working on other distributed systems, I found these constraints particularly effective:

- Give actors full data ownership and avoid shared references
- Using a non-async `on_message` function that mutates the internal state directly in response to events
- Assign each actor a unique identity (e.g. `TypeId`, metadata)
- Spawn actors through an actor system
- Route messages to destination actors via handles (e.g. `ActorRef`) through a bus

These constraints have various benefits: As actors operate only on local state, they can use `&mut self` and thus avoid locks. Actors with this design can be tested in isolation. Not using `async` enables Deterministic Simulation Testing which [TigerBeetle](https://tigerbeetle.com/) used to find subtle bugs.

The Akka framework is worth looking at in this context. It has additional features such as network transparency and automated actor restarts, which might not be relevant for Miden at this stage.

I am currently exploring how these constraints could be applied after #1219 has been merged. Here is a rough sketch of what I have in mind: https://gist.github.com/tindzk/e53772c940ac0fb2e3e886b0431c21e3

If this is a direction you deem worth exploring, I’d be happy to spend more time on it.

sergerad:
@bobbinth how many network accounts do we expect to have in the short term and long term?

bobbinth:
> [@bobbinth](https://github.com/bobbinth) how many network accounts do we expect to have in the short term and long term?

In the shorter term, I'd be surprised if we have more than 100K accounts (our current testnet has about 200K accounts total, this will increase in the future, but I think network accounts will be a minority).

In the longer term, we should expect tens of millions (though, probably not more until the very long term).

SantiagoPittella:
@Mirko-von-Leipzig question here: was the unassignment an error? The GH UI is ...flaky sometimes and I'm not sure if this was an error

Mirko-von-Leipzig:
Closing in favor of #1694

PR #1435, opened 2025-12-08
feat: NTX Builder Actor Refactor

sergerad:
## Context

The Network Transaction Builder is being refactored so that each network account's responsibilities are executed as dedicated tasks.

Relates to #1056.

Some of the additions are due to file location changes (E.G. account_state.rs)

## Changes
- Add `Actor` struct to NTX Builder stack for housing state and logic pertaining to individual network accounts.
- Add `Coordinator` to NTX Builder for coordinating `Actors` and feeding through mempool events.
- Address edge case to handle mempool events that predate network account creation.
- Stopped using NTB acronym and instead NTX Builder to avoid confusion.

### DIFF
--- a/crates/ntx-builder/src/actor/execute.rs
+++ b/crates/ntx-builder/src/actor/execute.rs
@@ -2,7 +2,13 @@ use std::collections::BTreeSet;
 
 use miden_node_utils::lru_cache::LruCache;
 use miden_node_utils::tracing::OpenTelemetrySpanExt;
-use miden_objects::account::{Account, AccountId, PartialAccount, StorageMapWitness, StorageSlot};
+use miden_objects::account::{
+    Account,
+    AccountId,
+    PartialAccount,
+    StorageMapWitness,
+    StorageSlotContent,
+};
 use miden_objects::asset::{AssetVaultKey, AssetWitness};
 use miden_objects::block::{BlockHeader, BlockNumber};
 use miden_objects::note::{Note, NoteScript};
@@ -380,7 +386,7 @@ impl DataStore for NtxDataStore {
 
             let mut map_witness = None;
             for slot in self.account.storage().slots() {
-                if let StorageSlot::Map(map) = slot {
+                if let StorageSlotContent::Map(map) = slot.content() {
                     if map.root() == map_root {
                         map_witness = Some(map.open(&map_key));
                     }


crates/ntx-builder/src/coordinator.rs
```diff
@@ -0,0 +1,273 @@
...
+    ///
+    /// This shared semaphore prevents the system from becoming overwhelmed by limiting the total
+    /// number of transactions that can be processed simultaneously across all account actors.
+    /// Each actor must acquire a permit from this semaphore before processing a transaction,
+    /// ensuring fair resource allocation and system stability under load.
+    semaphore: Arc<Semaphore>,
+
+    /// Cache of events received from the mempool that predate corresponding network accounts.
+    /// Grouped by account prefix to allow targeted event delivery to actors upon creation.
+    predating_events: HashMap<NetworkAccountPrefix, IndexMap<TransactionId, Arc<MempoolEvent>>>,
```
Mirko-von-Leipzig, line 91: This isn't strictly accurate. What these are meant to be are events that are still inflight from transactions that have not yet been committed (nor reverted).

These are replayed to an actor when it starts up so that it doesn't "miss" any events. To this end they need to be stored and replayed in chronological order, as if the actor had been present when they arrived.

crates/ntx-builder/src/coordinator.rs
```diff
@@ -0,0 +1,273 @@
...
+use crate::actor::{AccountActor, AccountActorContext, AccountOrigin, ActorShutdownReason};
+
+// ACTOR HANDLE
+// ================================================================================================
+
+/// Handle to account actors that are spawned by the coordinator.
+#[derive(Clone)]
+struct ActorHandle {
+    event_tx: mpsc::Sender<Arc<MempoolEvent>>,
+    cancel_token: CancellationToken,
```
Mirko-von-Leipzig, line 24: I think this is redundant. The actor should already cancel if the channel gets closed so that's essentially the same thing.

crates/ntx-builder/src/builder.rs
```diff
@@ -0,0 +1,261 @@
...
+            .context("failed to subscribe to mempool events")?;
+
+        // Unlock the block-producer's block production. The block-producer is prevented from
+        // producing blocks until we have subscribed to mempool events.
+        //
+        // This is a temporary work-around until the ntx-builder can resync on the fly.
+        self.bp_checkpoint.wait().await;
+
+        let mut interval = tokio::time::interval(self.ticker_interval);
+        interval.set_missed_tick_behavior(time::MissedTickBehavior::Skip);
```
Mirko-von-Leipzig, line 142: I think this is no longer needed?

crates/ntx-builder/src/builder.rs
```diff
@@ -0,0 +1,261 @@
...
+///
+/// The builder manages the tasks for every network account on the chain through the coordinator.
+pub struct NetworkTransactionBuilder {
+    /// Address of the store gRPC server.
+    store_url: Url,
+    /// Address of the block producer gRPC server.
+    block_producer_url: Url,
+    /// Address of the remote prover. If `None`, transactions will be proven locally, which is
+    /// undesirable due to the performance impact.
+    tx_prover_url: Option<Url>,
```
Mirko-von-Leipzig, line 78: Unrelated to this PR, but I wonder if we shouldn't move this remote/local division by _always_ taking a URL in the components themselves, and then having the node's CLI figure out whether to spawn a prover locally or not.

This would require quite a bit of work though because the prover worker cannot queue requests. So nevermind :)

crates/ntx-builder/src/builder.rs
```diff
@@ -0,0 +1,261 @@
...
+            }
+        }
+
+        // Main loop which manages actors and passes mempool events to them.
+        loop {
+            tokio::select! {
+                // Handle actor result.
+                result = self.coordinator.next() => {
+                    result?;
+                },
```
Mirko-von-Leipzig, line 171: I think the co-ordinator should probably never return an error? It seems like its only on semaphor failure or channel failure but I don't think those are valid reasons to crash the ntx.

crates/ntx-builder/src/builder.rs
```diff
@@ -0,0 +1,261 @@
...
+
+        // Create initial set of actors based on all known network accounts.
+        let account_ids = store.get_network_account_ids().await?;
+        for account_id in account_ids {
+            if let Ok(account_prefix) = NetworkAccountPrefix::try_from(account_id) {
+                self.coordinator
+                    .spawn_actor(AccountOrigin::store(account_prefix), &actor_context)
+                    .await?;
+            }
+        }
```
Mirko-von-Leipzig, line 163: Given that this might take a long while, we probably need to move this above the `bp_checkpoint` as otherwise mempool events will start streaming in without us handling them.

crates/ntx-builder/src/builder.rs
```diff
@@ -0,0 +1,261 @@
...
+        let mut mempool_events = block_producer
+            .subscribe_to_mempool_with_retry(chain_tip_header.block_num())
+            .await
+            .context("failed to subscribe to mempool events")?;
+
+        // Unlock the block-producer's block production. The block-producer is prevented from
+        // producing blocks until we have subscribed to mempool events.
+        //
+        // This is a temporary work-around until the ntx-builder can resync on the fly.
+        self.bp_checkpoint.wait().await;
```
Mirko-von-Leipzig, line 139: A medium term goal is to decouple the block-producer and ntx builder. Currently they are coupled because the ntx builder must first sync to the latest committed state from the store and we need that to align with the block-producer. This is enforced by using the `bp_checkpoint` barrier.

I've been thinking about how we can decouple these so that the ntx builder can start up independently at any time as a separate component. 

There are effectively two things that need to be sync'd:

1. The chain state (latest header & MMR) which is used to align the mempool subscription.
2. Existing network accounts loaded from store.

(1) is trivially solveable by simply repeating it in a loop until it succeeds i.e. (a) read state from store, (b) subscription, if desync'd then repeat.

(2) we can avoid entirely by creating a separate task who's goal it is to fetch network accounts from the store in chronological order until it reaches the chain tip. The task sends these as a new kind of event via a channel. In this way the main loop is eventually informed of all pre-existing accounts and it can decide to spin them up or not, or maybe they're already running.

crates/ntx-builder/src/builder.rs
Mirko-von-Leipzig: In the medium term, this should be paginated/streamed and done as a separate task as per https://github.com/0xMiden/miden-node/pull/1435/files#r2610211009

crates/ntx-builder/src/builder.rs
```diff
@@ -0,0 +1,261 @@
...
+                // All transactions pertaining to predating events should now be available through
+                // the store. So we can now drain them.
+                for tx_id in txs {
+                    self.coordinator.drain_predating_events(tx_id);
+                }
+                Ok(())
+            },
+            // Broadcast to all actors.
+            MempoolEvent::TransactionsReverted(txs) => {
+                self.coordinator.broadcast(event.clone()).await;
```
Mirko-von-Leipzig, line 230: Ideally this wouldn't be a broadcast as this wakes up _all_ sleeping actors.

We technically know which actors these transactions interacts with. We could figure that out and be more specific here.

crates/ntx-builder/src/builder.rs
```diff
@@ -0,0 +1,261 @@
...
+        chain_state.chain_mmr.add_block(&mmr_tip, true);
+
+        // Set the new tip.
+        chain_state.chain_tip_header = tip;
+
+        // Keep MMR pruned.
+        let pruned_block_height =
+            (chain_state.chain_mmr.chain_length().as_usize().saturating_sub(MAX_BLOCK_COUNT))
+                as u32;
+        chain_state.chain_mmr.prune_to(..pruned_block_height.into());
```
Mirko-von-Leipzig, line 259: Remind me why chain state isn't part of `self`? Is it a borrowing issue?

crates/ntx-builder/src/coordinator.rs
```diff
@@ -0,0 +1,273 @@
...
+                ActorShutdownReason::SemaphoreFailed(err) => Err(err).context("semaphore failed"),
+            },
+            Some(Err(err)) => {
+                tracing::error!(err = %err, "actor task failed");
+                Ok(())
+            },
+            None => {
+                // There are no actors to wait for. Wait indefinitely until actors are spawned.
+                std::future::pending().await
+            },
```
Mirko-von-Leipzig, line 214: Nit: since this is so different from the rest, consider moving it out using

```rust
let Some(result) = self.actor_join_set.join_next().await else {
    return std::future::pending().await
};
```

crates/ntx-builder/src/builder.rs
Mirko-von-Leipzig, line 1: I've written about this a bit before https://github.com/0xMiden/miden-node/issues/1314#issuecomment-3598248000, but I'll repeat it here for visibility.

-------

Some thoughts on how we can reduce the number of active account actors. This is largely orthogonal to whether we use an actor framework or just keep it simple with our current approach.

Some of these thoughts might be 💩

### Self shut down

Let's allow accounts to shut themselves down. They can do this by simply exiting with `return Ok(())`. The coordinator then interprets this as a deliberate decision and _does not_ restart the actor automatically.

Accounts could shut themselves down if they have no useful notes and haven't received an event in e.g. 5 minutes.

This technically introduces a race condition, where the actor shut downs and then an event arrives which should have kept it alive. In order to detect this, we could have the actor return the event receiver on exist i.e. `return Ok(rx_events)`. The coordinator can then check if the receiver is empty and if it isn't, restart the actor which would mean it would reprocess things.

### Blacklisting accounts

The co-ordinator can keep a hashmap of accounts that crash or return an error. Once the crash count reaches some limit they get added to a blacklist and are no longer considered. This should be fairly straight-forward to add.

### Standardizing account startup

I think at the moment we have separate flows for starting an account from an event, or starting an account from the store. I think we can connect these by always first getting the account state from the store. If there is no account, then the actor waits until it receives a new account event. We can similarly use the timeout strategy here so that the actor shuts down if the new account event doesn't arrive within a few minutes after start up.

This means we have the same flow for all actors.

PR #1436, opened 2025-12-09
feat: skip check on preflight requests

igamigo:
This PR allows preflight OPTIONS requests to go through without content type negotiation. The context is that browsers enforce these checks and the WebClient's gRPC client goes through the browser's `fetch()`, so this was causing issues with integration tests. An alternative to this was to enforce other requests being performed before calling `SubmitProvenTransaction`, but conceptually this would be more hacky.

This PR also removes a `dbg!()` line that seemed to have been leftover from the PR that implemented the enforcement of the genesis commitment.


crates/rpc/src/server/accept.rs
```diff
@@ -224,12 +224,16 @@ where
     }
 
     fn call(&mut self, request: http::Request<B>) -> Self::Future {
+        // Skip negotiation entirely for CORS preflight/non-gRPC requests.
+        if request.method() == http::Method::OPTIONS {
+            return self.inner.call(request).boxed();
+        }
```
Mirko-von-Leipzig, line 234: Another alternative would be to only have this layer active for `get/post` methods iiuc?

But I think this is fine.

Could you maybe expand on the comment with a _why_ this is necessary?

```suggestion
        // Skip negotiation entirely for CORS preflight/non-gRPC requests.
        //
        // Browsers often automatically perform an `OPTIONS` check _before_ the client
        // SDK can inject the appropriate `ACCEPT` header, causing a rejection.
        // Since an `OPTIONS` request does nothing its safe for us to simply allow them.
        if request.method() == http::Method::OPTIONS {
            return self.inner.call(request).boxed();
        }
```

juan518munoz: Addressed in [3bb0d00](https://github.com/0xMiden/miden-node/pull/1436/commits/3bb0d00b4efc4bb9ec22e476d56e862dc0df38cb)

igamigo: Yeah, only negotiating on `GET`/`POST` requests would have worked as well (probably only really needed for `POST` considering this is gRPC), but the current approach seemed less invasive

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@
 - [BREAKING] Renamed `ProxyWorkerStatus::address` to `ProxyWorkerStatus::name` ([#1348](https://github.com/0xMiden/miden-node/pull/1348)).
 - Added `SyncTransactions` stress test to `miden-node-stress-test` binary ([#1294](https://github.com/0xMiden/miden-node/pull/1294)).
 - Remove `trait AccountTreeStorage` ([#1352](https://github.com/0xMiden/miden-node/issues/1352)).
+- [BREAKING] `SubmitProvenTransaction` now **requires** that the network's genesis commitment is set in the request's `ACCEPT` header ([#1298](https://github.com/0xMiden/miden-node/pull/1298), [#1436](https://github.com/0xMiden/miden-node/pull/1436)).
+- Added support for timeouts in the WASM remote prover clients ([#1383](https://github.com/0xMiden/miden-node/pull/1383)).
+- Added block validation endpoint to validator and integrated with block producer ([#1382](https://github.com/0xMiden/miden-node/pull/1381)).
+- Added support for caching mempool statistics in the block producer server ([#1388](https://github.com/0xMiden/miden-node/pull/1388)).
+- Added mempool statistics to the block producer status in the `miden-network-monitor` binary ([#1392](https://github.com/0xMiden/miden-node/pull/1392)).
```
Mirko-von-Leipzig: I think this got a bit foobar'd in the merge


```suggestion
- [BREAKING] `SubmitProvenTransaction` now **requires** that the network's genesis commitment is set in the request's `ACCEPT` header ([#1298](https://github.com/0xMiden/miden-node/pull/1298), [#1436](https://github.com/0xMiden/miden-node/pull/1436)).
```

crates/rpc/src/server/accept.rs
Mirko-von-Leipzig: The argument for it would be that its more explicit to see from the server config aka like a business level instead of down in the implementation. Not a big deal either way so let's leave as is.

PR #1438, opened 2025-12-09
feat: reduce block interval to `2s`

### LINKED ISSUE #1437: Lower default block time to 2s

Mirko-von-Leipzig:
I've also lowered the batch interval from `2s -> 1s` which at least gives us time for multiple batches per block. There is an argument that it should be even lower?

Closes #1437

### DIFF
--- a/crates/block-producer/src/lib.rs
+++ b/crates/block-producer/src/lib.rs
@@ -51,7 +51,7 @@ const SERVER_MEMPOOL_EXPIRATION_SLACK: u32 = 2;
 const CACHED_MEMPOOL_STATS_UPDATE_INTERVAL: Duration = Duration::from_secs(5);
 
 /// How often a block is created.
-pub const DEFAULT_BLOCK_INTERVAL: Duration = Duration::from_secs(2);
+pub const DEFAULT_BLOCK_INTERVAL: Duration = Duration::from_secs(3);
 
 /// How often a batch is created.
 pub const DEFAULT_BATCH_INTERVAL: Duration = Duration::from_secs(1);
--- a/crates/store/src/accounts/mod.rs
+++ b/crates/store/src/accounts/mod.rs
@@ -118,7 +118,7 @@ pub struct AccountTreeWithHistory<S: SmtStorage> {
 
 impl<S: SmtStorage> AccountTreeWithHistory<S> {
     /// Maximum number of historical blocks to maintain.
-    pub const MAX_HISTORY: usize = 100;
+    pub const MAX_HISTORY: usize = 50;
 
     // CONSTRUCTORS
     // --------------------------------------------------------------------------------------------


bobbinth:
> I've also lowered the batch interval from `2s -> 1s` which at least gives us time for multiple batches per block. There is an argument that it should be even lower?

Batch times mean how frequently we ping the mempool to give us a proposed batch, right? There are not indications of how long it'll take to prove them. So, I think lowering to 1s is fine, and we could potentially lower it even more. I do expect batches to take some time to prove and maybe initially closer to 5 seconds.

Mirko-von-Leipzig:
@igamigo brought up a good point (https://github.com/0xMiden/miden-node/issues/1437#issuecomment-3634981194) which is that this also reduces the absolute time the client has for querying account history unless we also increase the depth there.

cc @drahnr ^^ do you know if the various proof history was chosen with an absolute time in mind, or was it more a spitball of `N` blocks? Should we be adapting this to account for this lowering of block time?

Mirko-von-Leipzig:
> I think it was chosen somewhat arbitrarily as "not too big and not too small". I do think it should be a value that is somewhere between 1 and 10 minutes, and with block times going to 2 seconds, the current setting of 30 is right at ~1 minute. So, maybe we should bump it up to 60 - or maybe even 100. Though, querying things later in history would becomes somewhat more expensive (I don't remember what would be the absolute difference in extra latency between 30 and 100).

I've bumped it to `100`; let's see how it goes.

crates/block-producer/src/lib.rs
```diff
@@ -51,10 +51,10 @@ const SERVER_MEMPOOL_EXPIRATION_SLACK: u32 = 2;
 const CACHED_MEMPOOL_STATS_UPDATE_INTERVAL: Duration = Duration::from_secs(5);
 
 /// How often a block is created.
-pub const DEFAULT_BLOCK_INTERVAL: Duration = Duration::from_secs(5);
+pub const DEFAULT_BLOCK_INTERVAL: Duration = Duration::from_secs(2);
```
bobbinth: After thinking some more about this, let's reduce this to 3 seconds. We will still likely to go to 2 seconds, but I'd like a bit more time to think it through.

crates/store/src/accounts/mod.rs
```diff
@@ -118,7 +118,7 @@ pub struct AccountTreeWithHistory<S: SmtStorage> {
 
 impl<S: SmtStorage> AccountTreeWithHistory<S> {
     /// Maximum number of historical blocks to maintain.
-    pub const MAX_HISTORY: usize = 33;
+    pub const MAX_HISTORY: usize = 100;
```
bobbinth: Given the above comment, let's change this to 50.

PR #1440, opened 2025-12-10
chore: update node to use named storage slots

PhilippGackstatter:
Updates the node to the latest `next` in miden base to use named storage slots.

In essence, this updates slot index to slot name in many places.

## Database

- I chose `BLOB` as the type for slot name, but let me know if you'd like a different one.
- "Slot index in 0..255" constraint in SQL was removed.

### Diesel

This required changing the `up.sql` database scheme, but I couldn't find instructions for how to update the schema automatically. I tried:

```
diesel migration run --config-file crates/store/diesel.toml --migration-dir crates/store/src/db/migrations
```

but that also caused unrelated updates (that also happen when running these migrations without bumping miden base):

```
-        block_num -> BigInt,
+        block_num -> Integer,
```

And this in turn causes compilation errors. So I updated schema.rs manually.

I wonder if there should be a CI job checking that `up.sql` is up to date with `schema.rs`?

It would also be great to add instructions somewhere for how to update this automatically or a `make` command to make this easier.

### DIFF
--- a/bin/network-monitor/src/deploy/counter.rs
+++ b/bin/network-monitor/src/deploy/counter.rs
@@ -23,12 +23,12 @@ use tracing::instrument;
 use crate::COMPONENT;
 
 static OWNER_SLOT_NAME: LazyLock<StorageSlotName> = LazyLock::new(|| {
-    StorageSlotName::new("miden::counter_contract::owner")
+    StorageSlotName::new("miden::monitor::counter_contract::owner")
         .expect("storage slot name should be valid")
 });
 
 static COUNTER_SLOT_NAME: LazyLock<StorageSlotName> = LazyLock::new(|| {
-    StorageSlotName::new("miden::counter_contract::counter")
+    StorageSlotName::new("miden::monitor::counter_contract::counter")
         .expect("storage slot name should be valid")
 });
 
--- a/bin/network-monitor/src/assets/counter_program.masm
+++ b/bin/network-monitor/src/assets/counter_program.masm
@@ -12,7 +12,7 @@ use.miden::tx
 use.std::sys
 
 # The slot in this component's storage layout where the counter is stored.
-const COUNTER_SLOT = word("miden::counter_contract::counter")
+const COUNTER_SLOT = word("miden::monitor::counter_contract::counter")
 
 # Increment function with note authentication
 # => []


crates/proto/src/domain/account.rs
```diff
@@ -169,14 +170,15 @@ impl TryFrom<proto::account::AccountStorageHeader> for AccountStorageHeader {
...
                 let slot_type = storage_slot_type_from_raw(slot.slot_type)?;
                 let commitment =
                     slot.commitment.ok_or(ConversionError::NotAValidFelt)?.try_into()?;
-                Ok((slot_type, commitment))
+                Ok((slot_name, slot_type, commitment))
             })
             .collect::<Result<Vec<_>, ConversionError>>()?;
 
-        Ok(AccountStorageHeader::new(items))
+        AccountStorageHeader::new(items).map_err(ConversionError::AccountError)
```
sergerad: No need for `map_err` because `#[from]` is used in the error enum

sergerad: As in, `Ok(AccountStorageHeader::new(items)?)`. Maybe you don't prefer that

crates/store/src/db/models/conv.rs
```diff
@@ -116,6 +117,24 @@ impl SqlTypeConvert for NoteTag {
...
+impl SqlTypeConvert for StorageSlotName {
+    type Raw = Vec<u8>;
+    type Error = DatabaseTypeConversionError;
+
+    fn from_raw_sql(raw: Self::Raw) -> Result<Self, Self::Error> {
+        String::from_utf8(raw)
+            .map_err(|_| DatabaseTypeConversionError(type_name::<StorageSlotName>()))
+            .and_then(|name| {
+                StorageSlotName::new(name)
+                    .map_err(|_| DatabaseTypeConversionError(type_name::<StorageSlotName>()))
```
sergerad, line 129: Is there no way, or is undesirable to, include the err being ignored here and above?

crates/proto/src/domain/account.rs
Mirko-von-Leipzig: I think the recommendation from the error discussion from way back was to avoid `#[from]` to avoid ambiguity and force people to be more explicit. I somewhat agree with this but we've never gotten around to removing/enforcing this completely in the node.

proto/proto/rpc.proto
```diff
@@ -311,8 +311,9 @@ message AccountStorageDetails {
 
             repeated StorageMapEntry entries = 1;
         }
-        // slot index of the storage map
-        uint32 slot_index = 1;
+
+        // Storage slot name.
+        string slot_name = 1;
```
bobbinth, line 316: This is fine for now, but I wonder if in the requests it would make sense to use storage slot IDs rather than names. Let's create an issue for this.

bin/network-monitor/src/deploy/counter.rs
```diff
@@ -14,12 +14,24 @@ use miden_objects::account::{
...
 
+static OWNER_SLOT_NAME: LazyLock<StorageSlotName> = LazyLock::new(|| {
+    StorageSlotName::new("miden::counter_contract::owner")
+        .expect("storage slot name should be valid")
+});
+
+static COUNTER_SLOT_NAME: LazyLock<StorageSlotName> = LazyLock::new(|| {
+    StorageSlotName::new("miden::counter_contract::counter")
+        .expect("storage slot name should be valid")
+});
```
bobbinth, line 33: nit (feel free to ignore): I'd maybe make the names a bit more "narrow" - i.e., `miden::monitor::counter_contract::counter` and `miden::monitor::counter_contract::owner`.

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -91,13 +91,12 @@ CREATE TABLE note_scripts (
 CREATE TABLE account_storage_map_values (
     account_id          BLOB NOT NULL,
     block_num           INTEGER NOT NULL,
-    slot                INTEGER NOT NULL,
+    slot_name           BLOB NOT NULL,
```
bobbinth, line 94: Related to one of the previous comments: I wonder if we should store both `slot_name` and `slot_id` here - primarily to make querying faster (the assumption here is that querying by `slot_id` would be faster than by `slot_name` - though, if there is a meaningful difference would need to be validated).

bin/network-monitor/src/deploy/counter.rs
Mirko-von-Leipzig, line 33: Bit of a bummer this - can `new` maybe be `const`?

..checking in base..

--------------

`StorageSlotName::new` cannot be `const` because it takes `name: Into<String>`. However maybe we could have a `const fn from_static(s: &'static str)` which `http` does for `HeaderName` etc.

..checking that..

----------------

Okay that doesn't work trivially without changing away from storing a `String` internally because there is no `const` way to convert `'static str` into a `String`. `HeaderName` solves this by storing an enum instead so that it could do something like:

```rust
enum StorageNameRepr {
    Owned(String),
    Static(&'static str),
    Bytes(&'static [u8]),
    ...
}
```

which isn't really worth the effort at the moment? Though maybe it is because fi imagine most names will actually be constant literals?

-----------------

Other alternatives if we keep `String` only:

```rust
/// Our own little macro for formatting into a name.
///
/// This can be made to emulate the `println!` macro etc with `{}` support
/// by delegating that to `format!`.
macro_rules! storage_slot_name {
   ...
}

/// Alternatively create a specific `from_string` impl which can be const.
impl StorageSlotName {
    pub const fn from_string(name: String) -> Result<Self, StorageSlotNameError> {
        Self::validate(&name)?;
        // This has the same problem :(
        let id = StorageSlotId::from_str(name.as_str());
        Ok(Self { name, id })
    }

    pub fn new(name: Into<String>) -> Result<Self, StorageSlotNameError> {
        Self::from_string(s.into())
    }
}
```

.. except I see we have the same issue with the Id etc that they're not const constructable either.

crates/store/src/db/models/conv.rs
Mirko-von-Leipzig: I think the base error type doesn't currently support something like that since its `&'static str` only. Maybe something we should consider i.e. to switch to something like:

```rust
#[error("failed to convert {from_type} into {into_type}"]
struct DatabaseTypeConversionError {
    source: Box<dyn Error>,
    from_type: &'static str,
    into_type: &'static str,
}

// With an automatic helper function implemented inside the trait
trait SqlTypeConvert {
    type Raw;
    // I think we can remove this since we never use anything else.
    // type Error 

    /// We have the type info available here already so we can make it easy.
    fn map_err(source: Box<dyn Error>) -> DatabaseTypeConversionError {
        DatabaseTypeConversionError {
            source,
            from_type: std::any::type_name(Self::Raw),
            into_type: std::any::type_name(Self),
        }
    }
    
    fn from_raw_sql(raw: Self::Raw) -> Result<Self, DatabaseTypeConversionError>;
}
```

Note that this would also let you stack these i.e. if you fail to convert a field within another struct then you would return a nested `DatabaseTypeConversionError` which would be pretty neat imo.

crates/store/Cargo.toml
```diff
@@ -19,8 +19,8 @@ anyhow                 = { workspace = true }
 deadpool               = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
 deadpool-diesel        = { features = ["sqlite"], version = "0.6" }
 deadpool-sync          = { version = "0.1" }
-diesel                 = { features = ["numeric", "sqlite"], version = "2.2" }
-diesel_migrations      = { features = ["sqlite"], version = "2.2" }
+diesel                 = { features = ["numeric", "sqlite"], version = "2.3" }
+diesel_migrations      = { features = ["sqlite"], version = "2.3" }
```
Mirko-von-Leipzig, line 23: Did we want to update these?

bin/network-monitor/src/deploy/counter.rs
Mirko-von-Leipzig: Another alternative is to _always_ store `&'static str` and use `String::leak` if you get an owned one :P

crates/store/src/db/migrations/2025062000000_setup/up.sql
Mirko-von-Leipzig: Is there a reason to need both? If `slot_id` is a unique ID then could we not just store one only?

bobbinth: We can't derive slot name from slot ID and I think in some responses we do want to send slot names to the user.

bin/network-monitor/src/deploy/counter.rs
PhilippGackstatter: It was a goal initially to have `const fn StorageSlotName::from_static_str`, and we had that implemented until two hours ago (https://github.com/0xMiden/miden-base/pull/2153).

The reason to move away from that and accept lazy statics is:
- https://github.com/0xMiden/miden-vm/pull/2287#discussion_r2432197033
- And that caching the storage slot ID in the name is much more convenient than caching it in account storage and storage header.

crates/store/src/db/models/conv.rs
PhilippGackstatter: I find it desirable, but again kept to existing patterns how this is being handled.

Edit: Suggestion looks good.

crates/proto/src/domain/account.rs
PhilippGackstatter: Yeah, I had `#[source]` at first, but then figured I would stay consistent with the existing approach.

As for @sergerad's suggestion, happy to change it if that's preferred.

I think the error handling approach re `#[from]` and not ignoring errors needs an own issue here.

bin/network-monitor/src/deploy/counter.rs
Mirko-von-Leipzig: Eugh :/ I guess I'll live with it.

Though I'll say having any cells, interior mutability, locks, lazy etc in a _library_ type that isn't a singleton is terrible for users of the library. In my opinion its a massive code smell but I see its a deeper underlying issue.

proto/proto/rpc.proto
PhilippGackstatter: That would also avoid the slot name clones in a few places now :+1:

crates/store/Cargo.toml
PhilippGackstatter: The version according to `Cargo.lock` was 2.3.x anyway, and this was a leftover from an attempt to get the diesel thing to work. Can roll back if you want.

bin/network-monitor/src/deploy/counter.rs
Mirko-von-Leipzig: I suppose this follows from requiring two unique but separate identifies (ID and name) for the same thing. Ideally only one of the two would be used internally and then somewhere higher up its just a mapping of the one to the other. Having them both in a single struct seems super weird.

But it looks like this was already discussed

crates/store/Cargo.toml
Mirko-von-Leipzig: Nope that's fine :) I didn't confirm in the lockfile so no stress.

bin/network-monitor/src/deploy/counter.rs
PhilippGackstatter: Just to clarify, we did not end up with any cells, interior mutability, etc. in `StorageSlotName`. It's a very simple struct. The lazy statics are at the application end of things.

This two identifier requirement follows from the split between tx kernel and Rust APIs. In the tx kernel we don't have the space for an entire string identifier per slot, but in Rust APIs we want to retain the human-readable name. We could use slot ID everywhere, but the DevX would be way worse.

Mirko-von-Leipzig: My counter argument would be that I expect 99% of `StorageSlotName` to be `const` definitions so having them all be runtime checked kinda sucks for DevX as well.

But okay I see this isn't a trivial thing; thanks for taking the time :)

crates/store/src/db/migrations/2025062000000_setup/up.sql
drahnr, line 94: Is there are reason this isn't `TEXT`? Do we support non-unicode characters?

crates/store/src/db/models/conv.rs
drahnr: The type conversions so far were rather simple, I left a escape hatch having the `    type Error = ..;` to use your own error type and wrap it as needed in `DatabaseError`

crates/store/src/db/schema.rs
```diff
@@ -1,10 +1,10 @@
 // @generated automatically by Diesel CLI.
 
 diesel::table! {
-    account_storage_map_values (account_id, block_num, slot, key) {
+    account_storage_map_values (account_id, block_num, slot_name, key) {
         account_id -> Binary,
         block_num -> BigInt,
-        slot -> Integer,
+        slot_name -> Binary,
```
drahnr, line 7: Same as early, should we prefer ` -> Text`?

crates/store/src/db/models/conv.rs
sergerad: Generally want to avoid hiding error info. Happy to add a followup issue for this rather than address in this PR

bobbinth:
I'll merge this now to unblock ongoing work.

@PhilippGackstatter - if there are any follow-ups, let's open a new PR for these.

PR #1442, opened 2025-12-10
refactor(store): don't drop source of sql conversion error

Mirko-von-Leipzig:
Inspire by https://github.com/0xMiden/miden-node/pull/1440#discussion_r2605295698.

I've added the error source for the SQL to Rust type conversion error which should just be a strict improvement over the status-quo.

I do think we can improve the situation further because there is at least one other sql-conversion error variant in the top level database error enum. However I think that will require some proper design to get a good abstraction.

### DIFF
--- a/crates/store/src/db/models/conv.rs
+++ b/crates/store/src/db/models/conv.rs
@@ -32,35 +32,46 @@
     on relevant platforms"
 )]
 
-use std::any::type_name;
-
-use miden_node_proto::domain::account::{NetworkAccountError, NetworkAccountPrefix};
+use miden_node_proto::domain::account::NetworkAccountPrefix;
 use miden_objects::Felt;
 use miden_objects::block::BlockNumber;
 use miden_objects::note::{NoteExecutionMode, NoteTag};
 
 #[derive(Debug, thiserror::Error)]
-#[error("failed to convert a database value to it's in memory type {0}")]
-pub struct DatabaseTypeConversionError(&'static str);
+#[error("failed to convert from database type {from_type} into {into_type}")]
+pub struct DatabaseTypeConversionError {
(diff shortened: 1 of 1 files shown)


crates/store/src/db/models/conv.rs
```diff
@@ -32,46 +32,56 @@
...
-use miden_node_proto::domain::account::{NetworkAccountError, NetworkAccountPrefix};
+use miden_node_proto::domain::account::NetworkAccountPrefix;
 use miden_objects::Felt;
 use miden_objects::block::BlockNumber;
 use miden_objects::note::{NoteExecutionMode, NoteTag};
 
 #[derive(Debug, thiserror::Error)]
-#[error("failed to convert a database value to it's in memory type {0}")]
-pub struct DatabaseTypeConversionError(&'static str);
+#[error("failed to convert from database type {from_type} into {into_type}")]
```
sergerad, line 41: Do we not need to explicitly mention `{source}` somewhere here?

sergerad: Or is `source()` somehow picking up on the `source` field via the macros?

Mirko-von-Leipzig: The latter - from the `thiserror` docs (no page anchor nearby unfortunately):

> The Error trait’s [source()](https://doc.rust-lang.org/core/error/trait.Error.html#method.source) method is implemented to return whichever field has a #[source] attribute or is named source, if any. This is for identifying the underlying lower level error that caused your error.
>
> The #[from] attribute always implies that the same field is #[source], so you don’t ever need to specify both attributes.
>
> Any error type that implements std::error::Error or dereferences to dyn std::error::Error will work as a source.

PR #1443, opened 2025-12-10
chore: unify and revisit query limits

### LINKED ISSUE #1389: Param limit mismatch between RPC and Store component
When reviewing https://github.com/0xMiden/miden-client/issues/1557 I was confused because the RPC component limits [note IDs to 1000](https://github.dev/0xMiden/miden-node/blob/8d76576ef63b1e06402b0c729e836a149507c456/crates/utils/src/limiter.rs#L74-L80), whereas they were limited to [100 within the store](https://github.com/0xMiden/miden-node/commit/2b00852f81534f47ce45953f43f8407835a78870#diff-07520853e15646f5a35212cb3b12b32aaf90449483b037845608db9ffa51eb9bR12-R14). 

We should probably look to consolidate these (and there might be more occurrences), unless there's a specific reason why this was done like this in the first place.

SantiagoPittella:
closes #1389 

Centralized limits for the store and rpc queries.


crates/utils/src/limiter.rs
```diff
@@ -1,12 +1,15 @@
-//! Limit the size of a parameter list for a specific parameter
+//! Centralized limits for RPC and store parameters and payload sizes.
```
bobbinth: nit: I'm not sure the word "Centralized" is needed here. Just "Limits for RPC and store parameters and payload sizes." should be fine.

crates/utils/src/limiter.rs
```diff
@@ -37,6 +40,10 @@ pub trait QueryParamLimiter {
     }
 }
 
+/// Maximum payload size (in bytes) for paginated responses returned by the
+/// store.
+pub const MAX_PAGINATED_PAYLOAD_BYTES: usize = 4 * 1024 * 1024;
```
bobbinth: Should this be more like `MAX_RESPONSE_PAYLOAD_BYTES`? This limit is not only for paginated responses, but for all responses - right?

crates/store/src/server/rpc_api.rs
```diff
@@ -545,13 +539,7 @@ impl rpc_server::Rpc for StoreApi {
             read_account_ids::<SyncTransactionsError>(&request.account_ids)?;
 
         // Validate account IDs count
-        if account_ids.len() > MAX_ACCOUNT_IDS {
-            return Err(SyncTransactionsError::TooManyAccountIds(
-                account_ids.len(),
-                MAX_ACCOUNT_IDS,
-            )
```
bobbinth, line 552: Are `SyncTransactionsError::TooManyAccountIds` and other similar error variants still relevant?

bobbinth (approved):

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 - [BREAKING] Added block signing capabilities to Validator component and updated gensis bootstrap to sign blocks with configured signer ([#1426](https://github.com/0xMiden/miden-node/pull/1426)).
 - Reduced default block interval from `5s` to `2s` ([#1438](https://github.com/0xMiden/miden-node/pull/1438)).
 - Increased retained account tree history from 33 to 100 blocks to account for the reduced block interval ([#1438](https://github.com/0xMiden/miden-node/pull/1438)).
+- Increased the maximum query limit for the store ([#1443](https://github.com/0xMiden/miden-node/pull/1443))
```
igamigo: nit for consistency:

```suggestion
- Increased the maximum query limit for the store ([#1443](https://github.com/0xMiden/miden-node/pull/1443)).
```

crates/utils/src/limiter.rs
```diff
@@ -1,12 +1,15 @@
-//! Limit the size of a parameter list for a specific parameter
+//! Centralized limits for RPC and store parameters and payload sizes.
 //!
-//! Used for:
-//! 1. the external facing RPC
-//! 2. limiting SQL statements not exceeding parameter limits
+//! # Rationale
+//! - Parameter limits are kept at 1000 items across all multi-value RPC parameters. This caps
```
igamigo: Is there a way to reference the 1000 const here? So that the doc comment does not accidentally drift if it gets updated

crates/utils/src/limiter.rs
SantiagoPittella: Yes, true that. I'm changing it.

crates/store/src/server/rpc_api.rs
SantiagoPittella: Not needed anymore.

SantiagoPittella:
> LGTM! As we talked offline, if this gets merged we should update the limits on the client, so we should open an issue or a PR directly since it would be a very simple change anyway.

Yes! I'm planning to open a PR for that when this gets merged.

> Also, one question: if get_notes_by_id returns 1000 notes (the limit), could it not potentially be above the maximum current response payload size?

You could request for up to 1000 notes but if the limit is reached you will need to request the "next page"

SantiagoPittella:
As @igamigo point me out, `get_notes_by_id` is not paginated. And from my approximation, a single note can ended up taking too much space in the worst case scenario, over 1 MB. So the 1000 limit can be way to large. Should I reduce it? I don't know how large notes _**usually**_ are. cc @bobbinth

crates/utils/src/limiter.rs
```diff
@@ -37,20 +42,24 @@ pub trait QueryParamLimiter {
     }
 }
 
+/// Maximum payload size (in bytes) for paginated responses returned by the
+/// store.
+pub const MAX_RESPONSE_PAYLOAD_BYTES: usize = 4 * 1024 * 1024;
```
bobbinth, line 47: I think the comment here needs to be updated as well.

bobbinth:
> As @igamigo point me out, `get_notes_by_id` is not paginated. And from my approximation, a single note can ended up taking too much space in the worst case scenario, over 1 MB. So the 1000 limit can be way to large. Should I reduce it? I don't know how large notes _**usually**_ are. cc @bobbinth

A single note should end up being that big. Right now, a note may have up to 1024 input elements (8KB), and we'll probably need to impose script size limit on a note, and we'll also soon have "note attachments", which we'll also probably limit to 8KB or 16KB.

Overall, maybe we impose a general limit on the size of a note to be something like 32KB (@PhilippGackstatter - let's create an issue for this if we don't have one yet). At this limit, we could allow up to 100 notes. So, let's reduce the limit on this endpoint to 100.

PR #1445, opened 2025-12-11
chore: named storage slot update follow-up

PhilippGackstatter:
Addresses the leftover issue from #1440 to use `Text` instead of `Binary` for the slot name DB representation.

Also updated base to address breaking changes from newer PRs, i.e. `SlotNameError` -> `StorageSlotNameError`. Also, used the newly added `From<StorageSlotName> for String` conversion to avoid an extra allocation.


Mirko-von-Leipzig (approved):
tyvm!

PR #1447, opened 2025-12-11
chore: use join for SyncTransaction query

### LINKED ISSUE #1368: Update `SyncTransactions` to use `join` in query
In the https://github.com/0xMiden/miden-node/pull/1357 PR I removed the N queries needed to get notes to complete the `SyncTransactions` response, but as Mirko and Bernhard suggested, we should be able to use one single query using a `join`.

SantiagoPittella:
closes #1368


crates/store/src/db/models/queries/transactions.rs
```diff
@@ -223,66 +230,73 @@ impl TransactionSummaryRowInsert {
...
+///
+/// FROM transactions AS tx
+/// LEFT JOIN notes AS n
+/// ON tx.block_num = n.committed_at
+/// AND tx.account_id = n.sender
+/// LEFT JOIN note_scripts AS s
+/// ON n.script_root = s.script_root
+/// WHERE tx.block_num >= :block_start
+/// AND tx.block_num <= :block_end
+/// AND tx.account_id IN (:account_ids)
```
Mirko-von-Leipzig: Maybe this helps

```suggestion
/// FROM transactions AS tx
///     LEFT JOIN notes AS n ON
///         x.block_num = n.committed_at
///         AND tx.account_id = n.sender
///     LEFT JOIN note_scripts AS s ON
///         n.script_root = s.script_root
/// WHERE
///     tx.block_num >= :block_start
///     AND tx.block_num <= :block_end
///     AND tx.account_id IN (:account_ids)
```

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -223,66 +230,73 @@ impl TransactionSummaryRowInsert {
...
+///     n.sender,
+///     n.tag,
+///     n.aux,
+///     n.execution_hint,
+///     n.assets,
+///     n.inputs,
+///     n.serial_num,
+///     n.inclusion_path,
+///     n.script_root,
+///     s.script  -- nullable
```
Mirko-von-Leipzig: I think this might not be helpful :D Too much, though its accurate.

I wonder if maybe instead we just say.. select stuff..

Not a strong feeling; feel free to ignore.

```suggestion
///     tx.{...}
///     n.{...}
```

crates/store/src/db/models/queries/transactions.rs
Mirko-von-Leipzig: Thinking about this some more, I don't think this works.

If there are sequential transactions from the same account with multiple notes each, then how will you assign the notes to each transaction?

I think we would also need the transaction ID in the notes table to do it this way? The alternative is to use something like `n.note_id in tx.output_notes` but that doesn't work because the latter is a BLOB. I suspect this will need to be two separate db queries - one to fetch the transaction info, and then a second on to fetch the notes+script. Which is effectively more or less what you started with :/

SantiagoPittella: Oh shoot, that's right.

>  I suspect this will need to be two separate db queries - one to fetch the transaction info, and then a second on to fetch the notes+script. Which is effectively more or less what you started with :/

That's exactly the current implementation 😅 .

Probably adding tx ID is the best way if we want to continue the one query path. WDYT?

Mirko-von-Leipzig: I think let's try adding tx ID and see what that looks like? Personally I don't think that increases the table size by much since it already contains many other large columns.

crates/store/src/db/models/queries/mod.rs
Mirko-von-Leipzig, line 1: I think move this logic into `insert_notes` by adding `&OrderedTransactions` as an input to that function.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -867,12 +866,13 @@ pub struct NoteInsertRowInsert {
     pub inclusion_path: Vec<u8>,
 }
 
-impl From<(NoteRecord, Option<Nullifier>)> for NoteInsertRowInsert {
-    fn from((note, nullifier): (NoteRecord, Option<Nullifier>)) -> Self {
+impl From<(NoteRecord, Option<Nullifier>, Vec<u8>)> for NoteInsertRowInsert {
```
Mirko-von-Leipzig, line 869: Would be good if we can end up here

```suggestion
impl From<(NoteRecord, Option<Nullifier>, TransactionId)> for NoteInsertRowInsert {
```

crates/store/src/db/models/queries/mod.rs
Mirko-von-Leipzig: Alternatively, we can move this logic up higher (which I see is how we construct the incoming notes).

However I think that pattern is actually a mistake and we should be minimizing the higher level deconstruction.

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -300,84 +293,137 @@ pub fn select_transactions_records(
...
+        // LEFT JOIN notes AS n ON
+        //   tx.block_num = n.committed_at
+        //   AND tx.account_id = n.sender
+        //   AND tx.transaction_id = n.transaction_id
+        .left_join(
+            notes::table.on(
+                notes::committed_at
+                    .eq(transactions::block_num)
+                    .and(notes::sender.eq(transactions::account_id))
+                    .and(notes::transaction_id.eq(transactions::transaction_id)),
```
Mirko-von-Leipzig, line 307: I think we only need transaction ID?

```suggestion
                    .and(notes::transaction_id.eq(transactions::transaction_id)),
```

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -44,6 +44,7 @@ CREATE TABLE notes (
     committed_at             INTEGER NOT NULL, -- Block number when the note was committed
     batch_index              INTEGER NOT NULL, -- Index of batch in block, starting from 0
     note_index               INTEGER NOT NULL, -- Index of note in batch, starting from 0
+    transaction_id           BLOB    NOT NULL, -- ID of the transaction which emitted this note
```
Mirko-von-Leipzig, line 47: Technically with this addition we can remove `committed_at` and `sender` since these are tracked as part of the transaction data. This might make other queries more painful though so I'm unsure if we should bother at this stage.

Maybe add an issue to look into this at a later stage.

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -63,12 +64,15 @@ CREATE TABLE notes (
...
     CONSTRAINT notes_execution_mode_in_enum CHECK (execution_mode BETWEEN 0 AND 1),
+    CONSTRAINT notes_transaction_id_is_digest CHECK (length(transaction_id) = 32),
     CONSTRAINT notes_consumed_at_is_u32 CHECK (consumed_at BETWEEN 0 AND 0xFFFFFFFF),
     CONSTRAINT notes_batch_index_is_u32 CHECK (batch_index BETWEEN 0 AND 0xFFFFFFFF),
     CONSTRAINT notes_note_index_is_u32 CHECK (note_index BETWEEN 0 AND 0xFFFFFFFF)
 );
 
 CREATE INDEX idx_notes_note_id ON notes(note_id);
+CREATE INDEX idx_notes_transaction_id ON notes(transaction_id);
+CREATE INDEX idx_notes_tx_join ON notes(sender, committed_at, transaction_id);
```
Mirko-von-Leipzig, line 75: This shouldn't be necessary with the suggestion https://github.com/0xMiden/miden-node/pull/1447/changes#r2641977678

crates/store/src/db/models/queries/transactions.rs
Mirko-von-Leipzig, line 1: !@#$%... I think this may have been a waste of time. It somehow didn't click for me that doing this would duplicate the transaction row for every note in that transaction. And since there can 100s of notes per tx, that's incredibly wasteful.

@SantiagoPittella thoughts? Should we just drop this PR? :/ The work is good; I just had (multiple) brain farts.

Mirko-von-Leipzig (commented):
Some improvement suggestions, but mostly this https://github.com/0xMiden/miden-node/pull/1447/changes#r2641998170 which probably means canning this PR and issue.. :/

crates/store/src/db/models/queries/transactions.rs
SantiagoPittella: I don't have a strong preference for any of the options, but IMO having the two queries is not that bad. I found that path to be a bit more readable than the triple join, despite being documented, and less prone to bugs. Wouldn't mind to close this PR & Issue.

Mirko-von-Leipzig: Okay let's close it

bobbinth: Agreed. Duplicating transaction records in the dataset is probably not desirable.

Issue #1255, opened 2025-09-29
Check RPC block height against explorer's


SantiagoPittella:
We can add a check for the explorer that matches (within some error margin) both RPC's height and explorer chain tip. This way we can add a "soft" confirmation that the explorer is working.

Also, we can add a check for the RPC to ensure that the tip is moving forward.

The explorer is working only for testnet currently, we might want to wait for a devnet explorer before implementing this.

PR #1450, opened 2025-12-12
feat: add explorer to network monitor

SantiagoPittella:
closes https://github.com/0xMiden/miden-node/issues/1255

Screenshot 2025-12-12 at 14 28 02


bin/network-monitor/assets/index.html
```diff
@@ -48,6 +48,7 @@
     <script>
         let statusData = null;
         let updateInterval = null;
+        const EXPLORER_LAG_TOLERANCE = 5; // max allowed block delta vs RPC
```
igamigo: Not sure if this was taken into account, but with block times being lowered as of recently I think this amounts to 15 seconds

bin/network-monitor/src/status.rs
```diff
@@ -93,6 +95,20 @@ pub struct CounterTrackingDetails {
...
+pub struct ExplorerStatusDetails {
+    pub block_number: u64,
+    pub timestamp: u64,
+    pub number_of_transactions: u64,
+    pub number_of_nullifiers: u64,
+    pub number_of_notes: u64,
+    pub number_of_account_updates: u64,
+    pub block_commitment: String,
+    pub chain_commitment: String,
+    pub proof_commitment: String,
```
igamigo, line 118: nit: Is it worth it to type these more (ie, making them `Word`)? I think we could catch potential problems with the explorer API itself which might be nice

bin/network-monitor/src/status.rs
```diff
@@ -468,3 +485,265 @@ pub(crate) async fn check_remote_prover_status(
...
+    query: &'static str,
+    variables: V,
+}
+
+fn latest_block_request() -> GraphqlRequest<serde_json::Value> {
+    GraphqlRequest {
+        query: LATEST_BLOCK_QUERY,
+        variables: serde_json::json!({}),
+    }
+}
```
igamigo: nit: This could be a `const` as well, right?

bin/network-monitor/src/status.rs
```diff
@@ -468,3 +485,265 @@ pub(crate) async fn check_remote_prover_status(
...
+            name.clone(),
+            current_time,
+            request_timeout,
+        )
+        .await;
+
+        // Send the status update; exit if no receivers (shutdown signal)
+        if status_sender.send(status).is_err() {
+            println!("No receivers for explorer status updates, shutting down");
+            info!("No receivers for explorer status updates, shutting down");
```
igamigo: nit (feel free to disregard): I'd remove the `println`; the info trace should be enough for the user

bin/network-monitor/src/status.rs
```diff
@@ -468,3 +485,265 @@ pub(crate) async fn check_remote_prover_status(
...
+///                 block_number
+///                 timestamp
+///                 number_of_transactions
+///                 number_of_nullifiers
+///                 number_of_notes
+///             }
+///         }
+///     }
+/// }
+/// ```
```
igamigo: nit: This seems unlikely to drift but you can reference `LATEST_BLOCK_QUERY` here so that it's automatically linked and it's always up to date

bin/network-monitor/src/monitor/tasks.rs
```diff
@@ -82,6 +84,39 @@ impl Tasks {
...
+        &mut self,
+        config: &MonitorConfig,
+    ) -> Result<Receiver<ServiceStatus>> {
+        let explorer_url = config.explorer_url.clone().expect("Explorer URL exists");
+        let name = "Explorer".to_string();
+        let status_check_interval = config.status_check_interval;
+        let request_timeout = config.request_timeout;
+        let (explorer_status_tx, explorer_status_rx) = watch::channel(initial_explorer_status());
+
+        println!("Spawning explorer status checker task for URL: {}", explorer_url.clone());
```
igamigo: Should these be `info` traces?

bin/network-monitor/assets/index.html
```diff
@@ -147,10 +153,39 @@
...
+                    `
+                    : '-';
+
+                const explorerStats = service.details?.ExplorerStatus;
+                const isExplorerService = service.name?.toLowerCase().includes('explorer');
+                const deltaBlock = (isHealthy && explorerStats && rpcChainTip !== null)
+                    ? explorerStats.block_number - rpcChainTip
+                    : null;
+                const deltaText = deltaBlock !== null ? `${deltaBlock >= 0 ? '+' : ''}${deltaBlock}` : '-';
+                const deltaClass = deltaBlock !== null && Math.abs(deltaBlock) > EXPLORER_LAG_TOLERANCE ? 'warning-delta' : '';
```
igamigo: This const doesn't seem used

bin/network-monitor/assets/index.html
```diff
@@ -147,10 +153,39 @@
...
+                const deltaWarning =
+                    deltaBlock !== null && Math.abs(deltaBlock) > EXPLORER_LAG_TOLERANCE
+                        ? `Explorer tip differs from RPC by ${deltaText} blocks`
+                        : null;
```
igamigo, line 196: nit: For UX, could be nice to say "Explorer tip is $N$ blocks _ahead_/_behind_"

bin/network-monitor/src/monitor/tasks.rs
igamigo: Also nit: I think the `.clone()` can be removed here

bin/network-monitor/README.md
```diff
@@ -151,6 +153,13 @@ The monitor application provides real-time status monitoring for the following M
 - **Block Producer Status**:
   - Block producer version and health
 
+### Explorer
+- **Service Health**: Explorer availability and freshness of the latest block
+- **Latest Block Metadata**:
+  - Block height and timestamp
+  - Transactions, nullifiers, notes, and account updates counts
+  - Block, chain, and proof commitments (shortened display with copy-to-clipboard)
```
igamigo, line 163: Ideally we would mention here that the check that compares RPC chain tip and explorer chain tip happens on the frontend.

bin/network-monitor/src/status.rs
```diff
@@ -468,3 +485,265 @@ pub(crate) async fn check_remote_prover_status(
         },
     }
 }
+
+// EXPLORER STATUS CHECKER
+// ================================================================================================
```
igamigo: very minor nit: This feels like enough code/local consts that it could be its own module

bin/network-monitor/assets/index.html
```diff
@@ -147,10 +153,39 @@
...
+                };
+                const commitmentOrDash = (value, label) => isHealthy && value
+                    ? `
+                        ${value.substring(0, 20)}...
+                        <button class="copy-button" onclick="copyToClipboard('${value}', event)" title="Copy full ${label}">
+                            <svg class="copy-icon" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2">
+                                <rect x="9" y="9" width="13" height="13" rx="2" ry="2"></rect>
+                                <path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1"></path>
+                            </svg>
+                        </button>
```
igamigo: This is the same copy button as in other instances, right? If so, I wonder if it is worth creating a function that generates it.

bin/network-monitor/.env
```diff
@@ -16,3 +16,5 @@ MIDEN_MONITOR_DISABLE_NTX_SERVICE=false
 MIDEN_MONITOR_COUNTER_FILEPATH=counter_account.mac
 MIDEN_MONITOR_WALLET_FILEPATH=wallet_account.mac
 MIDEN_MONITOR_COUNTER_INCREMENT_INTERVAL=30s
+# explorer checks
+MIDEN_MONITOR_EXPLORER_URL=https://scan-backend-testnet-miden.eu-central-8.gateway.fm/graphql
```
igamigo: nit: Some defaults above (at least `MIDEN_MONITOR_REMOTE_PROVER_URLS`) uses devnet, would be nice if they were consistent. 
But also, maybe for making ops even easier there could be a holistic `devnet`/`testnet` endpoint that already configures singular components if they were not overridden? Although this may also get confusing (and this is not in the scope of this issue anyway)

bin/network-monitor/src/status.rs
```diff
@@ -468,3 +485,265 @@ pub(crate) async fn check_remote_prover_status(
...
+    match explorer_client
+        .post(explorer_url.to_string())
+        .json(&latest_block_request())
+        .timeout(request_timeout)
+        .send()
+        .await
+    {
+        Ok(response) => match response.json::<serde_json::Value>().await {
+            Ok(status) => match ExplorerStatusDetails::try_from(status) {
+                Ok(explorer_details) => ServiceStatus {
```
igamigo: nit: This feels a bit cumbersome to read due to the very similar errors that take up various lines each, but not sure if there is a better way to do it. Maybe a function that creates an unhealthy response similar to `initial_explorer_status`?

bin/network-monitor/assets/index.html
SantiagoPittella: Should I make it something like 1 minute?

bin/network-monitor/src/status.rs
SantiagoPittella: Oh, yes mb

bin/network-monitor/src/monitor/tasks.rs
SantiagoPittella: This println can be removed since we log when we call the spawn function.

bin/network-monitor/src/status.rs
SantiagoPittella: I'm not sure if this is a validation that we want to do in the monitor, since we want we want to check is mainly availability IMO and not correctness

bin/network-monitor/src/status.rs
SantiagoPittella: I went for this approach with the function that you mentioned, it is much more readable

bin/network-monitor/.env
SantiagoPittella: >nit: Some defaults above (at least MIDEN_MONITOR_REMOTE_PROVER_URLS) uses devnet, would be nice if they were consistent.

I moved to devnet for the example variables.

> But also, maybe for making ops even easier there could be a holistic devnet/testnet endpoint that already configures singular components if they were not overridden? Although this may also get confusing (and this is not in the scope of this issue anyway)

Maybe I'm not fully getting it, but seems a bit confusing

SantiagoPittella:
Current view:

Screenshot 2025-12-17 at 12 10 33

Issue #1299, opened 2025-10-20
Add pagination for get_network_account_ids()


sergerad:
As per [comment](https://github.com/0xMiden/miden-node/pull/1219#discussion_r2443605497):

>Probably not a lot right now - e.g., under a million in the foreseeable future, but in the longer run, it could be a lot (e.g., tens or maybe even hundreds of millions).
>
> As I mentioned in another comment, we should create an issue to enable pagination for this function.

Look for `// TODO: add pagination.` once #1219 is merged.

PR #1452, opened 2025-12-12
feat: add pagination to GetNetworkAccountIds

SantiagoPittella:
I'm working on https://github.com/0xMiden/miden-node/issues/1299 .

I noticed that we do not store the block in which an account is created, so the typical pagination that we are using in the other endpoint does not work here. We would need some kind of cursor to keep track the position of the overall query for the pagination. The problem is that we cannot count on any of the fields in the accounts schema to be ordered. A solution could be to add a `created_at` field storing the block number in which the account was deployed.

cc @Mirko-von-Leipzig @drahnr

### DIFF
--- a/crates/ntx-builder/src/store.rs
+++ b/crates/ntx-builder/src/store.rs
@@ -177,6 +177,10 @@ impl StoreClient {
     ///
     /// Since the `GetNetworkAccountIds` method is paginated, we loop through all pages until we
     /// reach the end.
+    ///
+    /// Each page can return up to `MAX_RESPONSE_PAYLOAD_BYTES / AccountId::SERIALIZED_SIZE`
+    /// accounts (~289,000). With 1000 iterations, this supports up to ~524 million network
+    /// accounts, which is assumed to be sufficient for the foreseeable future.
     #[instrument(target = COMPONENT, name = "store.client.get_network_account_ids", skip_all, err)]
     pub async fn get_network_account_ids(&self) -> Result<Vec<AccountId>, StoreError> {
         const MAX_ITERATIONS: u32 = 1000;
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -433,7 +433,19 @@ impl Db {
         .await
     }
 
-    /// Loads all network account IDs from the DB.
+    /// Returns network account IDs within the specified block range (based on account creation
+    /// block).
+    ///
+    /// The function may return fewer accounts than exist in the range if the result would exceed
+    /// `MAX_RESPONSE_PAYLOAD_BYTES / AccountId::SERIALIZED_SIZE` rows. In this case, the result is
+    /// truncated at a block boundary to ensure all accounts from included blocks are returned.
+    ///
+    /// # Returns
+    ///
+    /// A tuple containing:
+    /// - A vector of network account IDs.
+    /// - The last block number that was fully included in the result. When truncated, this will be
+    ///   less than the requested range end.
     #[instrument(level = "debug", target = COMPONENT, skip_all, ret(level = "debug"), err)]
     pub async fn select_all_network_account_ids(
         &self,
--- a/crates/store/src/db/models/queries/accounts.rs
+++ b/crates/store/src/db/models/queries/accounts.rs
@@ -410,11 +410,19 @@ pub(crate) fn select_all_accounts(
     Ok(account_infos)
 }
 
-/// Returns all network account IDs.
+/// Returns network account IDs within the specified block range (based on account creation
+/// block).
+///
+/// The function may return fewer accounts than exist in the range if the result would exceed
+/// `MAX_RESPONSE_PAYLOAD_BYTES / AccountId::SERIALIZED_SIZE` rows. In this case, the result is
+/// truncated at a block boundary to ensure all accounts from included blocks are returned.
 ///
 /// # Returns
 ///
-/// A vector with network account IDs, or an error.
+/// A tuple containing:
+/// - A vector of network account IDs.
+/// - The last block number that was fully included in the result. When truncated, this will be less
+///   than the requested range end.
 pub(crate) fn select_all_network_account_ids(
     conn: &mut SqliteConnection,
     block_range: RangeInclusive<BlockNumber>,
@@ -449,9 +457,10 @@ pub(crate) fn select_all_network_account_ids(
     )
     .load::<(Vec<u8>, i64)>(conn)?;
 
-    if let Some(&(_, last_created_at_block)) = account_ids_raw.last()
-        && account_ids_raw.len() > MAX_ROWS
-    {
+    if account_ids_raw.len() > MAX_ROWS {
+        // SAFETY: We just checked that len > MAX_ROWS, so the vec is not empty.
+        let last_created_at_block = account_ids_raw.last().expect("vec is not empty").1;
+
         let account_ids = account_ids_raw
             .into_iter()
             .take_while(|(_, created_at_block)| *created_at_block != last_created_at_block)
--- a/crates/store/src/server/ntx_builder.rs
+++ b/crates/store/src/server/ntx_builder.rs
@@ -151,7 +151,14 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
         }))
     }
 
-    // TODO: add pagination.
+    /// Returns network account IDs within the specified block range (based on account creation
+    /// block).
+    ///
+    /// The function may return fewer accounts than exist in the range if the result would exceed
+    /// `MAX_RESPONSE_PAYLOAD_BYTES / AccountId::SERIALIZED_SIZE` rows. In this case, the result is
+    /// truncated at a block boundary to ensure all accounts from included blocks are returned.
+    ///
+    /// The response includes pagination info with the last block number that was fully included.
     #[instrument(
         parent = None,
         target = COMPONENT,
--- a/crates/store/src/state.rs
+++ b/crates/store/src/state.rs
@@ -906,7 +906,14 @@ impl State {
         self.db.select_network_account_by_prefix(id_prefix).await
     }
 
-    /// Returns account IDs for all public (on-chain) network accounts.
+    /// Returns network account IDs within the specified block range (based on account creation
+    /// block).
+    ///
+    /// The function may return fewer accounts than exist in the range if the result would exceed
+    /// `MAX_RESPONSE_PAYLOAD_BYTES / AccountId::SERIALIZED_SIZE` rows. In this case, the result is
+    /// truncated at a block boundary to ensure all accounts from included blocks are returned.
+    ///
+    /// The response includes the last block number that was fully included in the result.
     pub async fn get_all_network_accounts(
         &self,
         block_range: RangeInclusive<BlockNumber>,


bobbinth:
> I noticed that we do not store the block in which an account is created, so the typical pagination that we are using in the other endpoint does not work here. We would need some kind of cursor to keep track the position of the overall query for the pagination. The problem is that we cannot count on any of the fields in the accounts schema to be ordered. A solution could be to add a `created_at` field storing the block number in which the account was deployed.

This is for the `accounts` table, right? And the proposal is to add `created_at` field to the `accounts` table so that we have a stable way to sort accounts (i.e., to avoid accounts getting re-sorted if a new account is added or an existing account is updated) - correct?

crates/ntx-builder/src/store.rs
```diff
@@ -174,8 +176,21 @@ impl StoreClient {
 
     // TODO: add pagination.
```
drahnr: I think this can be removed?

crates/ntx-builder/src/store.rs
```diff
@@ -174,8 +176,21 @@ impl StoreClient {
...
     #[instrument(target = COMPONENT, name = "store.client.get_network_account_ids", skip_all, err)]
-    pub async fn get_network_account_ids(&self) -> Result<Vec<AccountId>, StoreError> {
-        let response = self.inner.clone().get_network_account_ids(()).await?.into_inner();
+    pub async fn get_network_account_ids(
+        &self,
+        block_range: Option<RangeInclusive<BlockNumber>>,
```
drahnr: Can we avoid the `Option`? Nobody should request this without anyways, no?

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -414,21 +414,58 @@ pub(crate) fn select_all_accounts(
 /// A vector with network account IDs, or an error.
 pub(crate) fn select_all_network_account_ids(
     conn: &mut SqliteConnection,
-) -> Result<Vec<AccountId>, DatabaseError> {
-    let account_ids_raw: Vec<Vec<u8>> = QueryDsl::select(
-        schema::accounts::table.filter(schema::accounts::network_account_id_prefix.is_not_null()),
-        schema::accounts::account_id,
+    block_range: RangeInclusive<BlockNumber>,
+) -> Result<(Vec<AccountId>, BlockNumber), DatabaseError> {
+    const ROW_OVERHEAD_BYTES: usize = AccountId::SERIALIZED_SIZE;
```
drahnr, line 430: future work/probably too much: we might want to use `build.rs` that generates this const value

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -414,21 +414,58 @@ pub(crate) fn select_all_accounts(
...
-        .map(|id_bytes| {
-            AccountId::read_from_bytes(&id_bytes).map_err(DatabaseError::DeserializationError)
-        })
-        .collect::<Result<Vec<AccountId>, DatabaseError>>()?;
+    if let Some(&(_, last_created_at_block)) = account_ids_raw.last()
+        && account_ids_raw.len() > MAX_ROWS
+    {
+        let account_ids = account_ids_raw
+            .into_iter()
+            .take_while(|(_, created_at_block)| *created_at_block != last_created_at_block)
```
drahnr, line 466: I am a bit unsure if this is ok, if we have so many accounts for a single block num that we'd exceed the MAX_ROWS and we only query for a single account, wouldn't this return no accounts?

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -800,13 +837,29 @@ pub(crate) fn upsert_accounts(
     let mut count = 0;
     for update in accounts {
         let account_id = update.account_id();
+        let account_id_bytes = account_id.to_bytes();
```
drahnr, line 857: not for this PR: we might want to replace these eventually with the `conv.rs` SQL conversion trait

crates/ntx-builder/src/store.rs
Mirko-von-Leipzig: I actually think we need to retain the original behavior _within this_ method. The ntx builder _currently_ expects this to return **all** network account IDs.

To retain this behavior _for now_ I would

- revert function signature change
- internally loop over and aggregate the returned pages

As currently used this will only return the first page in the main builder https://github.com/0xMiden/miden-node/pull/1452/changes#diff-630c8daca8771b806879dbee6a3059b92341ac7418c345180f537c911dd272a2R15

@SantiagoPittella ^^

crates/store/src/db/models/queries/accounts.rs
Mirko-von-Leipzig: `MAX_ROWS` must be larger than the maximum number of accounts allowed in a block.

We should be able to `const` assert this is the case by doing 

```rust
const _: () = assert!(
    MAX_ROWS > miden_objects::MAX_ACCOUNTS_PER_BLOCK,
    "Block pagination limit must exceed maximum block capacity"
);
```

crates/ntx-builder/src/store.rs
SantiagoPittella: Removed!

crates/ntx-builder/src/store.rs
SantiagoPittella: I implemented this changes

crates/store/src/db/models/queries/accounts.rs
SantiagoPittella: I added this check 👌🏼

crates/store/src/db/models/queries/accounts.rs
SantiagoPittella: I will open an issue to address this in the future

crates/store/src/db/models/queries/accounts.rs
SantiagoPittella: Should I create an issue for this one?

crates/store/src/db/models/queries/accounts.rs
SantiagoPittella: I created https://github.com/0xMiden/miden-node/issues/1477 to search for any other struct that is missing the SQL conversion trait

Mirko-von-Leipzig (approved):

crates/store/src/db/models/queries/accounts.rs
drahnr: I'd like @Mirko-von-Leipzig 's input on it first :)

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -417,21 +417,63 @@ pub(crate) fn select_all_accounts(
...
-        schema::accounts::table.filter(schema::accounts::network_account_id_prefix.is_not_null()),
-        schema::accounts::account_id,
+    block_range: RangeInclusive<BlockNumber>,
+) -> Result<(Vec<AccountId>, BlockNumber), DatabaseError> {
+    const ROW_OVERHEAD_BYTES: usize = AccountId::SERIALIZED_SIZE;
+    const MAX_ROWS: usize = MAX_RESPONSE_PAYLOAD_BYTES / ROW_OVERHEAD_BYTES;
+
+    const _: () = assert!(
+        MAX_ROWS > miden_objects::MAX_ACCOUNTS_PER_BLOCK,
+        "Block pagination limit must exceed maximum block capacity"
```
drahnr: nit: 
```suggestion
        "Block pagination limit must exceed maximum block capacity to uphold assumed logic invariant"
```

crates/store/src/db/models/queries/accounts.rs
Mirko-von-Leipzig: This is an internal only endpoint, so I think we can get away with simplifying this as much as possible such that we don't need a derived `MAX_ROWS` at all, but can simply keep it at some arbitrary value.

With a small bit of refactoring in the ntx builder (briefly touched upon in https://github.com/0xMiden/miden-node/issues/1478#issuecomment-3689619190), we can simplify this function signature to

```rust
pub(crate) fn select_all_network_account_ids(
    conn: &mut SqliteConnection,
    token: ContinuationToken,
) -> Result<(Vec<AccountId>, ContinuationToken), DatabaseError> 
```

This is because the ntx builder won't need to care about block-delimited pages, and instead only needs a mechanism to get _all_ currently committed network account IDs.

So all we need is an iteration key that retains the order, which could be

```sql
ORDER BY rowid                  -- if we don't delete, or
ORDER BY created_at, account_id -- for an agnostic one
```

and then 

```rust
struct ContinuationToken(rowid);
// or
struct ContinuationToken(BlockNumber, AccountId);
```

If we do decide to go the stream route, then `ContinuationToken` would remain internal only, otherwise it forms part of the protobuf.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -417,21 +417,63 @@ pub(crate) fn select_all_accounts(
 /// A vector with network account IDs, or an error.
 pub(crate) fn select_all_network_account_ids(
```
bobbinth: I don't think doc comments here are correct - i.e., this does not return _all_ network account IDs. Let's update it to indicate that this tries to return all IDs within the specified block range (based on account creation block), but could return less if there are too many account IDs (we should specify what's the maximum number of account IDs that can be returned).

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -417,21 +417,63 @@ pub(crate) fn select_all_accounts(
...
 
-    let account_ids = account_ids_raw
-        .into_iter()
-        .map(|id_bytes| {
-            AccountId::read_from_bytes(&id_bytes).map_err(DatabaseError::DeserializationError)
-        })
-        .collect::<Result<Vec<AccountId>, DatabaseError>>()?;
+    if let Some(&(_, last_created_at_block)) = account_ids_raw.last()
+        && account_ids_raw.len() > MAX_ROWS
+    {
```
bobbinth: nit: to improve readability, I think we should be able to write this as something like this:
```rust
if account_ids_raw.len() > MAX_ROWS {
    let Some(&(_, last_created_at_block)) = account_ids_raw.last().expect("...");
    ...
}
```

crates/store/src/db/models/queries/accounts.rs
bobbinth, line 825: Not for this PR, but this method has gotten pretty complicated with multiple requests to the database being made. Let's create an issue to refactor it so that we issue only one read and one write request during this method (I think this should be enough).

crates/store/src/db/mod.rs
```diff
@@ -435,9 +435,14 @@ impl Db {
 
     /// Loads all network account IDs from the DB.
     #[instrument(level = "debug", target = COMPONENT, skip_all, ret(level = "debug"), err)]
-    pub async fn select_all_network_account_ids(&self) -> Result<Vec<AccountId>> {
-        self.transact("Get all network account IDs", queries::select_all_network_account_ids)
-            .await
+    pub async fn select_all_network_account_ids(
+        &self,
+        block_range: RangeInclusive<BlockNumber>,
+    ) -> Result<(Vec<AccountId>, BlockNumber)> {
```
bobbinth: Doc comments here need to be updated.

crates/store/src/server/ntx_builder.rs
```diff
@@ -161,14 +162,28 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
     )]
     async fn get_network_account_ids(
```
bobbinth: Doc comments here need to be updated.

crates/store/src/state.rs
```diff
@@ -907,8 +907,11 @@ impl State {
     }
 
     /// Returns account IDs for all public (on-chain) network accounts.
-    pub async fn get_all_network_accounts(&self) -> Result<Vec<AccountId>, DatabaseError> {
-        self.db.select_all_network_account_ids().await
+    pub async fn get_all_network_accounts(
+        &self,
+        block_range: RangeInclusive<BlockNumber>,
+    ) -> Result<(Vec<AccountId>, BlockNumber), DatabaseError> {
```
bobbinth: Doc comments here need to be updated.

crates/ntx-builder/src/store.rs
```diff
@@ -172,21 +173,56 @@ impl StoreClient {
...
-            .account_ids
-            .into_iter()
-            .map(|account_id| {
-                AccountId::read_from_bytes(&account_id.id)
-                    .map_err(|err| ConversionError::deserialization_error("account_id", err))
-            })
-            .collect();
-
-        Ok(accounts?)
+        const MAX_ITERATIONS: u32 = 1000;
```
bobbinth, line 186: Would be great to add a comment here indicating why 1000 iterations is enough or at least state what kind of assumption we are making here about the maximum number of network accounts.

crates/store/src/db/models/queries/accounts.rs
SantiagoPittella: I created https://github.com/0xMiden/miden-node/issues/1483

PR #1453, opened 2025-12-14
feat: Integrate ntx-builder with validator 

### LINKED ISSUE #1309: Integrate NTB into validator flow
See [comment](https://github.com/0xMiden/miden-node/pull/1278#discussion_r2453921545) chain.

We need to figure out how to involve NTB in the validator flow. Likely solution is to have NTB send the transactions to validator directly before forwarding to block producer.

This would mean that it is blocked on https://github.com/0xMiden/miden-node/pull/1293.

sergerad:
## Context

We want network transactions to go through the same validator flow as all other transactions. Currently the ntx-builder submits proven transactions to the block producer directly which bypasses the validator flow.

Closes #1309.

## Changes
- Add `ValidatorClient` for ntx-builder.
- Remove `BlockProducerClient::submit_proven_transaction` for ntx-builder.
- Use `ValidatorClient::submit_proven_transaction`.

### DIFF
--- a/crates/ntx-builder/src/actor/execute.rs
+++ b/crates/ntx-builder/src/actor/execute.rs
@@ -177,11 +177,11 @@ impl NtxContext {
                 let tx_inputs: TransactionInputs = executed_tx.into();
                 let proven_tx = Box::pin(self.prove(&tx_inputs)).await?;
 
-                // Submit transaction to block producer.
-                self.submit(&proven_tx).await?;
-
                 // Validate proven transaction.
                 self.validate(&proven_tx, &tx_inputs).await?;
+
+                // Submit transaction to block producer.
+                self.submit(&proven_tx).await?;
 
                 Ok((proven_tx.id(), failed_notes))
             })
--- a/crates/ntx-builder/src/actor/mod.rs
+++ b/crates/ntx-builder/src/actor/mod.rs
@@ -4,6 +4,7 @@ mod inflight_note;
 mod note_state;
 
 use std::sync::Arc;
+use std::time::Duration;
 
 use account_state::{NetworkAccountState, TransactionCandidate};
 use execute::NtxError;
@@ -176,7 +177,7 @@ impl AccountActor {
         let block_producer = BlockProducerClient::new(actor_context.block_producer_url.clone());
         let validator = Builder::new(actor_context.validator_url.clone())
             .without_tls()
-            .without_timeout()
+            .with_timeout(Duration::from_secs(10))
             .without_metadata_version()
             .without_metadata_genesis()
             .with_otel_context_injection()


crates/ntx-builder/src/rpc.rs
Mirko-von-Leipzig, line 1: Unfortunately this won't work as the [RPC rejects network transactions](https://github.com/0xMiden/miden-node/blob/9c578f78f1e1593a8276b1a5cb95d70df4db1a8f/crates/rpc/src/server/api.rs#L373-L380) since these are not allowed to be sent publicly (yet).

Options:

1. Add the validator directly to ntx -- this seems unreasonable.
2. Open up network transactions publicly
3. Add authentication somehow
4. An additional internal RPC url which the ntx uses which only supports the ntx submit tx method but skips the rejection

I haven't done (3) with gRPC before, so I would automatically reach for auth via http but https://grpc.io/docs/guides/auth/ has some more info.

drahnr: Can you share some more context on the relationship of auth / network transactions beyond to give some context? (link is enough)

Mirko-von-Leipzig: There isn't really any direct context aside from the fact that currently we only allow our own network transaction builder component to submit network transactions. Allowing external submissions would cause races with our builder and we don't want to deal with that at this stage.

This is currently enforced in the RPC component's `SubmitProvenTransaction` implementation by simply rejecting all transactions that target a network account (except for those which _create_ the account). This means external clients/users can only create a network account and thereafter all transactions for that account cannot go via the public RPC endpoints.

The network tx builder circumvents this by using the block-producer's `SubmitProvenTransaction` directly. This works but now that we're adding additional logic (validator stuff) to the RPC's gRPC implementation it would be ideal if we only have that in a single place. 

Network transactions must also be submitted to the validator, so either we
1. duplicate the validator interaction logic within the network transaction builder as well, or
2. keep a single coherent impl in the RPC component
    - and therefore the RPC component must somehow allow network transactions from the internal network transaction builder _without_ allowing them in from the general public
    - one option for this is some sort of authentication to just say "hey this came from our internal network transaction builder, don't reject it"

bobbinth: I don't think option 2 is viable. Network transactions shouldn't come from the outside at this point.

Why is option 1 unreasonable? NTX builder could send the transaction directly to the validator in the same way as the RPC does. It feel like this would take the least amount of work.

My order of preference would probably be:

- Option 3, if it is not difficult to do. We could send the API key in the headers somehow - though, I remember working with headers was not super straight-forward - but maybe now we have this solved.
- Option 1 - just sent transactions from the NTX builder directly to the validator. As I mentioned above, this should be pretty straight-forward to implement.
- Option 4 - though, running two separate RPCs (one for internal use and one for external use) may introduce more complexity than the other two approaches.

sergerad: Summarizing: go with option 1 - ntx-builder talks directly to validator

sergerad: Moved forward with Option 1 (ready for review again)

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -156,14 +159,24 @@ impl NtxContext {
...
+
+                // Prove transaction.
+                let tx_inputs: TransactionInputs = executed_tx.into();
+                let proven_tx = Box::pin(self.prove(tx_inputs.clone())).await?;
+                let tx_id = proven_tx.id();
+
+                // Validate proven transaction.
+                self.validate(proven_tx, tx_inputs).await?;
+
+                Ok((tx_id, failed_notes))
```
Mirko-von-Leipzig: I think you still need to submit this to the block-producer or no?

SantiagoPittella (commented):
Looking good, besides Mirko's comment

crates/ntx-builder/src/actor/execute.rs
sergerad: Ah I think at some point I was thinking of pairing this PR with the one that moves proving to the store. I need to undo some changes and make sure we still submit to block-producer as you say here

sergerad:
@Mirko-von-Leipzig @SantiagoPittella I have undone the changes that remove the block producer submit logic. Should not have been part of this PR.

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -169,6 +174,13 @@ impl AccountActor {
         cancel_token: CancellationToken,
     ) -> Self {
         let block_producer = BlockProducerClient::new(actor_context.block_producer_url.clone());
+        let validator = Builder::new(actor_context.validator_url.clone())
+            .without_tls()
+            .without_timeout()
```
Mirko-von-Leipzig: We may want a timeout? Or does this default to some value.. can't remember now.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -156,14 +165,26 @@ impl NtxContext {
...
+
+                // Prove transaction.
+                let tx_inputs: TransactionInputs = executed_tx.into();
+                let proven_tx = Box::pin(self.prove(&tx_inputs)).await?;
+
+                // Submit transaction to block producer.
+                self.submit(&proven_tx).await?;
+
+                // Validate proven transaction.
+                self.validate(&proven_tx, &tx_inputs).await?;
```
Mirko-von-Leipzig: Should this be swapped around?

PR #1454, opened 2025-12-14
feat(node): configure ntx script cache size via clap

### LINKED ISSUE #1335: Make NtxDataStore script cache size configurable
The value is currently always set to the default const:
```rust
const DEFAULT_SCRIPT_CACHE_SIZE: usize = 1000;
```

We need to move to an environment variable instead.

sashass1315:
reopen #1340 
Fixes #1335


crates/ntx-builder/src/builder.rs
```diff
@@ -105,7 +105,25 @@ impl NetworkTransactionBuilder {
...
+
+    pub fn with_script_cache_size(
+        store_url: Url,
+        block_producer_url: Url,
+        tx_prover_url: Option<Url>,
+        ticker_interval: Duration,
+        bp_checkpoint: Arc<Barrier>,
+        script_cache_size: NonZeroUsize,
+    ) -> Self {
+        let script_cache = LruCache::new(script_cache_size);
```
sergerad, line 103: I think we should just update `new()` and not have `with_script_cache_size()`.

sashass1315: > I think we should just update `new()` and not have `with_script_cache_size()`.

updated

bin/node/src/commands/mod.rs
```diff
@@ -35,6 +36,7 @@ const ENV_GENESIS_CONFIG_FILE: &str = "MIDEN_GENESIS_CONFIG_FILE";
 const ENV_MAX_TXS_PER_BATCH: &str = "MIDEN_MAX_TXS_PER_BATCH";
 const ENV_MAX_BATCHES_PER_BLOCK: &str = "MIDEN_MAX_BATCHES_PER_BLOCK";
 const ENV_MEMPOOL_TX_CAPACITY: &str = "MIDEN_NODE_MEMPOOL_TX_CAPACITY";
+const ENV_NTX_SCRIPT_CACHE_SIZE: &str = "MIDEN_NTX_DATA_STORE_SCRIPT_CACHE_SIZE";
```
sergerad, line 38: Please add this to `bin/node/.env` like the other env vars.

crates/ntx-builder/src/lib.rs
```diff
@@ -13,6 +13,9 @@ pub use builder::NetworkTransactionBuilder;
 
 const COMPONENT: &str = "miden-ntx-builder";
 
+pub const DEFAULT_SCRIPT_CACHE_SIZE: NonZeroUsize =
+    NetworkTransactionBuilder::DEFAULT_SCRIPT_CACHE_SIZE;
```
sergerad: We don't need this do we?

crates/ntx-builder/src/builder.rs
```diff
@@ -95,7 +95,7 @@ impl NetworkTransactionBuilder {
     ///
     /// Each cached script contains the deserialized `NoteScript` object, so the actual memory usage
     /// depends on the complexity of the scripts being cached.
-    const DEFAULT_SCRIPT_CACHE_SIZE: NonZeroUsize = NonZeroUsize::new(1000).unwrap();
+    pub const DEFAULT_SCRIPT_CACHE_SIZE: NonZeroUsize = NonZeroUsize::new(1000).unwrap();
```
sergerad: This doesn't need to be pub does it?

sashass1315: > This doesn't need to be pub does it?

Made changes

PR #1457, opened 2025-12-15
feat: Integrate RPC with Validator for transaction validation

sergerad:
## Context

The RPC still performs transaction validation from a module in its server crate. Instead, it needs to use the Validator client to allow the Validator component to perform the transaction validation.

Relates to #1316.

## Changes

Note: most of the changes are miscellaneous fixes to bring node up to date with miden-base

- Add validator RPC URL flag to RPC component command.
- Add validator client to RPC component.
- Implemented validator logic for re-executing and validating transactions.


bin/network-monitor/src/deploy/counter.rs
```diff
@@ -50,12 +52,18 @@ pub fn create_counter_account(owner_account_id: AccountId) -> Result<Account> {
...
+    let library_path = LibraryPath::new("counter::program")?;
+    let module = Module::parser(ModuleKind::Library)
+        .parse_str(library_path, script, &source_manager)
+        .map_err(|e| anyhow::anyhow!("Failed to parse module: {e}"))?;
+    let library = assembler
+        .assemble_library([module])
+        .map_err(|e| anyhow::anyhow!("Failed to assemble library: {e}"))?;
+
+    let account_code = AccountComponent::new(library, vec![counter_slot, owner_id_slot])?
+        .with_supports_all_types();
```
sergerad, line 57: Wasn't sure how to do this with CodeBuilder (anything involving a script).

crates/rpc/src/server/mod.rs
bobbinth: We should mention the validator here as well.

crates/validator/src/server/mod.rs
```diff
@@ -101,9 +109,46 @@ impl<S: BlockSigner + Send + Sync + 'static> api_server::Api for ValidatorServer
...
+            })?;
+
+        // Validate that the executed transaction matches the submitted transaction.
+        let executed_tx_header: TransactionHeader = (&executed_tx).into();
+        let proven_tx_header: TransactionHeader = (&proven_tx).into();
+        if executed_tx_header != proven_tx_header {
+            return Err(Status::aborted(
+                "Executed transaction does not match submitted transaction",
+            ));
+        }
```
bobbinth: I would probably move this code into a separate module - similar to how we had it before. The function could be something like:

```rust
pub async fn validate_transaction(
    proven_tx: ProvenTransaction,
    tx_inputs: TransactionInputs,
) -> Result<TransactionHeader, SomeError> {
    ...
}
```

crates/validator/src/tx_validation/data_store.rs
bobbinth, line 1: Related to the previous comment: I'd probably move the code for validating transactions here so that the server code contains just the basic request/response serialization code and the business logic is contained here.

This file could be called `tx_validation.rs` and I'd probably place it under `validator/src/tx_validation.rs`.

Also, let's add section separators to this file.

crates/validator/src/server/mod.rs
```diff
@@ -101,9 +109,46 @@ impl<S: BlockSigner + Send + Sync + 'static> api_server::Api for ValidatorServer
...
-        _request: tonic::Request<proto::transaction::ProvenTransaction>,
+        request: tonic::Request<proto::transaction::ProvenTransaction>,
     ) -> Result<tonic::Response<()>, tonic::Status> {
-        // TODO(sergerad): Implement transaction validation logic.
+        let request = request.into_inner();
+        // Deserialize the transaction.
+        let proven_tx =
+            ProvenTransaction::read_from_bytes(&request.transaction).map_err(|err| {
+                Status::invalid_argument(err.as_report_context("Invalid proven transaction"))
+            })?;
```
bobbinth, line 115: One other thing we need to do is check that the proof associated with the proven transaction is correct (i.e., we need to verify the proof).

crates/validator/src/server/mod.rs
```diff
@@ -101,9 +109,46 @@ impl<S: BlockSigner + Send + Sync + 'static> api_server::Api for ValidatorServer
...
+
+        // Execute the transaction.
+        let (account, block_header, _, input_notes, tx_args) = tx_inputs.into_parts();
+        let executor: TransactionExecutor<'_, '_, _, UnreachableAuth> =
+            TransactionExecutor::new(&data_store);
+        let executed_tx = executor
+            .execute_transaction(account.id(), block_header.block_num(), input_notes, tx_args)
+            .await
+            .map_err(|err| {
+                Status::internal(err.as_report_context("Failed to execute transaction"))
```
bobbinth: I'm not sure if the error status here should be "internal". This could be the result of wrong transaction inputs? So, I'd treat it as input error.

crates/validator/src/server/mod.rs
```diff
@@ -101,9 +109,46 @@ impl<S: BlockSigner + Send + Sync + 'static> api_server::Api for ValidatorServer
...
+            })?;
+
+        // Validate that the executed transaction matches the submitted transaction.
+        let executed_tx_header: TransactionHeader = (&executed_tx).into();
+        let proven_tx_header: TransactionHeader = (&proven_tx).into();
+        if executed_tx_header != proven_tx_header {
+            return Err(Status::aborted(
+                "Executed transaction does not match submitted transaction",
+            ));
```
bobbinth: What's the thinking behind using "aborted" here for the status? Usually we use "invalid_argument" to indicate that the user provided wrong data.

crates/rpc/src/server/api.rs
```diff
@@ -384,21 +400,26 @@ impl api_server::Api for RpcService {
...
                 Status::invalid_argument(err.as_report_context("Invalid transaction inputs"))
             })?;
-            // Re-execute the transaction.
-            match validator::re_execute_transaction(tx_inputs).await {
-                Ok(_executed_tx) => {
+            // Re-execute the transaction via the Validator.
+            let validator_request = proto::transaction::ProvenTransaction {
+                transaction: tx.to_bytes(),
+                transaction_inputs: Some(tx_inputs.to_bytes()),
+            };
```
bobbinth: Do we need to re-serialize the data here? Could we not use the original `request` and just forward it to the validator?

crates/rpc/src/server/api.rs
bobbinth, line 1: Not for this PR, but I'd love to organize this file a bit better - e.g., add section separators, add comments, and maybe move logic from complex methods into separate files. Let's create an issue for this.

bin/network-monitor/src/deploy/counter.rs
bobbinth: Maybe @igamigo or @PhilippGackstatter would have a suggestion here.

igamigo: Something like

```rust
    let component_code = CodeBuilder::default()
        .compile_component_code("counter::program", script)
        .map_err(|e| {
            anyhow::anyhow!("Failed to compile component: {e}")
        })?;
    let account_code = AccountComponent::new(component_code, vec![counter_slot, owner_id_slot])?
        .with_supports_all_types();
```

should work (you can also use `CodeBuilder::with_source_manager(source_manager)`)

crates/rpc/src/server/api.rs
sergerad: https://github.com/0xMiden/miden-node/issues/1458

crates/validator/src/tx_validation/mod.rs
bobbinth, line 1: nit: I would maybe group both this file and the data store file under a common module. For example:

- `src/tx_validation/mod.rs` (for this file).
- `src/tx_validation/data_store.rs` (for the data store file).

crates/validator/src/tx_validation/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+
+// TRANSACTION VALIDATION
+// ================================================================================================
+
+/// Validates a transaction by verifying its proof, executing it and comparing its header with the
+/// provided proven transaction.
+pub async fn validate_transaction(
+    proven_tx: ProvenTransaction,
+    tx_inputs: TransactionInputs,
+) -> Result<(), TransactionValidationError> {
```
bobbinth, line 33: Returning `()` is fine for now, but I do think that we'll probably end up returning `TransactionHeader` from here so that we can store it in the validator's database.

bobbinth (approved):
Looks good! Thank you! I left a couple of optional comments inline.

crates/validator/src/tx_validation/mod.rs
sergerad: https://github.com/0xMiden/miden-node/pull/1459

crates/store/src/db/tests.rs
```diff
@@ -1601,18 +1609,17 @@ fn genesis_with_multiple_accounts() {
...
 
-    let component3 = AccountComponent::compile(
-        "export.baz push.3 end",
-        TransactionKernel::assembler(),
-        component_storage,
-    )
-    .unwrap()
-    .with_supported_type(AccountType::RegularAccountImmutableCode);
+    let account_component_code = CodeBuilder::default()
+        .compile_component_code("baz::interface", "export.baz push.3 end")
```
PhilippGackstatter, line 1613: FYI: After https://github.com/0xMiden/miden-base/pull/2158 is merged, all `export`s will have to become `pub proc`.

PhilippGackstatter (approved):
I reviewed only the `CodeBuilder` parts. Those LGTM.

PR #1460, opened 2025-12-17
feat: Add block validation logic against validated transactions (in-memory)

sergerad:
## Context

We need block validation to involve a check that all the transactions in a given block have previously been validated by the Validator.

We will begin by doing this in memory. Followup PR will add persistence.

Relates to https://github.com/0xMiden/miden-node/issues/1316.

## Changes
- Update Validator transaction validation to store validated transaction headers.
- Update Validator block validation to check for previously validated headers.

### DIFF
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -102,9 +102,10 @@ pub enum BundledCommand {
         #[arg(
             long = "validator.insecure.secret-key",
             env = ENV_VALIDATOR_INSECURE_SECRET_KEY,
-            value_name = "VALIDATOR_INSECURE_SECRET_KEY"
+            value_name = "VALIDATOR_INSECURE_SECRET_KEY",
+            default_value = INSECURE_VALIDATOR_KEY_HEX
         )]
-        validator_insecure_secret_key: Option<String>,
+        validator_insecure_secret_key: String,
     },
 }
 
@@ -137,9 +138,8 @@ impl BundledCommand {
                 grpc_timeout,
                 validator_insecure_secret_key,
             } => {
-                let secret_key_hex =
-                    validator_insecure_secret_key.unwrap_or(INSECURE_VALIDATOR_KEY_HEX.into());
-                let signer = SecretKey::read_from_bytes(hex::decode(secret_key_hex)?.as_ref())?;
+                let secret_key_bytes = hex::decode(validator_insecure_secret_key)?;
+                let signer = SecretKey::read_from_bytes(&secret_key_bytes)?;
                 Self::start(
                     rpc_url,
                     data_directory,


crates/validator/src/server/mod.rs
```diff
@@ -123,9 +142,14 @@ impl<S: BlockSigner + Send + Sync + 'static> api_server::Api for ValidatorServer
         })?;
 
         // Validate the transaction.
-        validate_transaction(proven_tx, tx_inputs).await.map_err(|err| {
-            Status::invalid_argument(err.as_report_context("Invalid transaction"))
-        })?;
+        let validated_tx_header =
+            validate_transaction(proven_tx.clone(), tx_inputs).await.map_err(|err| {
```
igamigo: nit: Is the new `clone()` here necessary? It doesn't look like anything else changed in regards to `proven_tx`

crates/validator/src/block_validation/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+    let (header, body) = build_block(proposed_block)?;
+
+    // Check that all transactions in the proposed block have been validated
+    let validated_txs = validated_transactions.read().await;
+    for tx_header in body.transactions().as_slice() {
+        let tx_id = tx_header.id();
+        if !validated_txs.contains_key(&tx_id) {
+            return Err(BlockValidationError::TransactionNotValidated(tx_id, header.block_num()));
+        }
+    }
```
igamigo: With this now being an error, should the behavior in the node that still continues with the transaction even when re-execution fails be turned into an error instead of a warning?

igamigo (approved):

crates/validator/src/block_validation/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+    // Sign the header.
+    let signature = signer.sign(&header);
+
+    // Remove the validated transactions from the cache.
+    let mut validated_txs = validated_transactions.write().await;
+    for tx_header in body.transactions().as_slice() {
+        validated_txs.remove(&tx_header.id());
+    }
+    // Release the validated transactions write lock.
+    drop(validated_txs);
```
Mirko-von-Leipzig: I suggest not manually managing the locks as it becomes easy to refactor your way into lock issues. How about replacing the lock type alias with a proper `struct ValidatedTransactions` instead?

```rust
#[derive(Default, Clone)]
struct ValidatedTransactions(RwLock<HashMap<TransactionId, TransactionHeader>>);

impl ValidatedTransactions {
    pub async fn check(&self, txs: impl Iterator<TransactionId>) -> Result<(), BlockValidationError>{
        let inner = self.0.read().await;

        ...
    }

    pub async fn remove(&self, txs: ...) {
        let inner = self.0.read().await;
        for tx in txs {
            inner.remove(tx);
        }
    }
}
```

Mirko-von-Leipzig: Although thinking on this some more, I don't think we can actually remove these transactions at this stage.

We should only remove transactions once we _know_ that the block has successfully landed in the store? Right now we're assuming that if this function is called that the caller successfully completes its end i.e. we're being optimistic.

Do we not need two methods, one for signing and one for the store to submit a `BlockCommitted(TransactionIDs)`?

crates/validator/src/server/mod.rs
```diff
@@ -1,26 +1,37 @@
...
 use tonic::Status;
 use tower_http::catch_panic::CatchPanicLayer;
 use tower_http::trace::TraceLayer;
 
 use crate::COMPONENT;
+use crate::block_validation::validate_block;
 use crate::tx_validation::validate_transaction;
 
+/// A type alias for a read-write lock that stores validated transactions.
+pub type ValidatedTransactions = RwLock<HashMap<TransactionId, TransactionHeader>>;
```
bobbinth: I think we won't have the time to implement proper persistence for this before the next release. So, maybe this could be something like an LRU cache? We could set the size to be pretty big (e.g., thousands of transactions), but at least we'll be able to keep it bounded and we won't need to manually manage transaction eviction.

In the longer run, we probably want to store transactions for a long time for audibility purposes - but that's something to implement when we add persistence mechanism.

crates/validator/src/block_validation/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+    // Build the block.
+    let (header, body) = build_block(proposed_block)?;
+
+    // Check that all transactions in the proposed block have been validated
+    let validated_txs = validated_transactions.read().await;
+    for tx_header in body.transactions().as_slice() {
+        let tx_id = tx_header.id();
+        if !validated_txs.contains_key(&tx_id) {
+            return Err(BlockValidationError::TransactionNotValidated(tx_id, header.block_num()));
+        }
```
bobbinth: I would probably first validate the transactions and then build the block - mostly to avoid building the block if any of the transactions is invalid.

I think we should be able to get the transactions from the proposed block just as easily as from the block body - i.e., via `proposed_block.transactions()`.

crates/validator/src/block_validation/mod.rs
bobbinth: I would keep the transactions around for a long time. As I mentioned in my other comment, until we have persistence implemented, we could probably use something like an LRU cache (with fairly generous capacity) to keep transactions around. This way, we should be able to avoid the need for manual dropping.

crates/validator/src/server/mod.rs
sergerad: >I think we won't have the time to implement proper persistence for this before the next release

If we release the in-memory version won't we risk causing a halt if we restart the node? Which is to say that we can't release this until persistence is in place.

crates/validator/src/block_validation/mod.rs
sergerad: Added Lru impl

crates/validator/src/server/mod.rs
Mirko-von-Leipzig: imo just adding a `validator.sqlite` with a basic

```sql
CREATE TABLE transactions (
    id BLOB PRIMARY KEY,
    header BLOB NOT NULL,
    body BLOB NOT NULL,
    timestamp INTEGER NOT NULL,
);
```

is probably quicker to do than the time we spend thinking and discussing this :)

The "hardest" part will be the cli configuration and briefly thinking about how/if concurrent writes are a concern.

sergerad: There was also the issue of code factoring because the validator db wouldn't go through the store. I had a PR a while ago that implemented a basic approach for it. I assume we are still of the mind that it doesn't make sense for the validator db to be tied to the store, but I haven't thought about it for a while now.

Mirko-von-Leipzig: It would be its own separate database

crates/validator/src/server/mod.rs
```diff
@@ -1,25 +1,40 @@
...
 use tonic::Status;
 use tower_http::catch_panic::CatchPanicLayer;
 use tower_http::trace::TraceLayer;
 
 use crate::COMPONENT;
+use crate::block_validation::validate_block;
 use crate::tx_validation::validate_transaction;
 
+/// Number of transactions to keep in the validated transactions cache.
+const NUM_VALIDATED_TRANSACTIONS: NonZeroUsize = NonZeroUsize::new(7000).unwrap();
```
bobbinth, line 33: Let's bump this up to 10K transactions.

PR #1464, opened 2025-12-18
fix: reduce notes limits to 100

SantiagoPittella:
This is a followup of https://github.com/0xMiden/miden-node/pull/1443

In that PR the limits were generalized to 1000, but for notes it can exceed the maximum payload size.

Related to the docs:

> Separately, it may be good to also include limits in the RPC endpoint [docs](https://docs.miden.xyz/miden-node/rpc). Let's create another issue for this (unless we already have one).

There is a draft to get the limits through a RPC call, so it could be documented in that endpoint. Should I add it anyways here? cc @bobbinth

### DIFF
--- a/crates/utils/src/limiter.rs
+++ b/crates/utils/src/limiter.rs
@@ -47,6 +47,9 @@ pub const MAX_RESPONSE_PAYLOAD_BYTES: usize = 4 * 1024 * 1024;
 
 /// Used for the following RPC endpoints
 /// * `state_sync`
+///
+/// Capped at 1000 account IDs to keep SQL `IN` clauses bounded and response payloads under the
+/// 4 MB budget.
 pub struct QueryParamAccountIdLimit;
 impl QueryParamLimiter for QueryParamAccountIdLimit {
     const PARAM_NAME: &str = "account_id";
@@ -55,6 +58,9 @@ impl QueryParamLimiter for QueryParamAccountIdLimit {
 
 /// Used for the following RPC endpoints
 /// * `select_nullifiers_by_prefix`
+///
+/// Capped at 1000 prefixes to keep queries and responses comfortably within the 4 MB payload
+/// budget and to avoid unbounded prefix scans.
 pub struct QueryParamNullifierPrefixLimit;
 impl QueryParamLimiter for QueryParamNullifierPrefixLimit {
     const PARAM_NAME: &str = "nullifier_prefix";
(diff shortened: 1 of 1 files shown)


crates/utils/src/limiter.rs
```diff
@@ -85,7 +84,8 @@ impl QueryParamLimiter for QueryParamNoteTagLimit {
 pub struct QueryParamNoteIdLimit;
 impl QueryParamLimiter for QueryParamNoteIdLimit {
     const PARAM_NAME: &str = "note_id";
-    const LIMIT: usize = GENERAL_REQUEST_LIMIT;
+    ///The approximate maximum size for notes is around 32KB.
```
igamigo:
```suggestion
    /// The approximate maximum size for notes is around 32KB.
```

crates/utils/src/limiter.rs
```diff
@@ -85,7 +84,8 @@ impl QueryParamLimiter for QueryParamNoteTagLimit {
 pub struct QueryParamNoteIdLimit;
 impl QueryParamLimiter for QueryParamNoteIdLimit {
     const PARAM_NAME: &str = "note_id";
-    const LIMIT: usize = GENERAL_REQUEST_LIMIT;
+    /// The approximate maximum size for notes is around 32KB.
+    const LIMIT: usize = 100;
```
bobbinth: I would maybe move the comment into the struct doc comments, and maybe expand it - e.g., "The limit is set to 100 notes to make sure that response size does not exceed 4 MB as note sizes are limited to 32 KB".

It may be nice to add similar comments to other structs here to make size assumptions explicit - but that could be done in a follow-up PR.

SantiagoPittella: I did those changes in this PR

crates/utils/src/limiter.rs
```diff
@@ -56,6 +58,9 @@ impl QueryParamLimiter for QueryParamAccountIdLimit {
 
 /// Used for the following RPC endpoints
 /// * `select_nullifiers_by_prefix`
+///
+/// Capped at 1000 prefixes to keep queries and responses comfortably within the 4 MB payload
+/// budget and to avoid unbounded prefix scans.
 pub struct QueryParamNullifierPrefixLimit;
```
bobbinth, line 64: It would be great to add the underlying assumptions to the comments explaining why 1000 is the right number. For example, 1000 prefixes could return a large number of nullifiers - but I think there is some pagination enforced where whatever is returned would still fit into 4 MB, right?

SantiagoPittella: That's true for some of those endpoints, but does not apply to all of them since some are not paginated. I can add the approximation that was made to determine the limits

bobbinth: For not-paginated ones we should also try to ensure that the response size does not exceed 4 MB. I think it would good to have documentation for every endpoint explaining how this limit is satisfied. This may highlight that we are not satisfying this limit somewhere.

The only endpoints that don't need that are the ones that are used for testing only (though, not sure if we have such user-facing endpoints).

PR #1471, opened 2025-12-19
chore: refresh base crates

igamigo:
Updates crates from `miden-base`

### DIFF
--- a/crates/proto/src/domain/mempool.rs
+++ b/crates/proto/src/domain/mempool.rs
@@ -10,8 +10,6 @@ use super::note::NetworkNote;
 use crate::errors::{ConversionError, MissingFieldHelper};
 use crate::generated as proto;
 
-// TODO: remove this allow, maybe box BlockHeader
-#[allow(clippy::large_enum_variant)]
 #[derive(Debug, Clone)]
 pub enum MempoolEvent {
     TransactionAdded {
@@ -21,7 +19,8 @@ pub enum MempoolEvent {
         account_delta: Option<AccountUpdateDetails>,
     },
     BlockCommitted {
-        header: BlockHeader,
+        // Box'd as this struct is quite large and triggers clippy.
+        header: Box<BlockHeader>,
         txs: Vec<TransactionId>,
     },
     TransactionsReverted(HashSet<TransactionId>),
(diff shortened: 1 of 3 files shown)


crates/proto/src/domain/mempool.rs
```diff
@@ -10,6 +10,8 @@ use super::note::NetworkNote;
 use crate::errors::{ConversionError, MissingFieldHelper};
 use crate::generated as proto;
 
+// TODO: remove this allow, maybe box BlockHeader
+#[allow(clippy::large_enum_variant)]
 #[derive(Debug, Clone)]
 pub enum MempoolEvent {
```
bobbinth: Interesting why this got trigged now.

bobbinth (approved):
Looks good! Thank you!

Once we have a corresponding PR in `miden-client`, let's run the integration tests, and if all looks good, we can merge this in.

crates/proto/src/domain/mempool.rs
Mirko-von-Leipzig: Its a bit confusing but _I think_ what's happened is that the lint name is misleading. A more accurate description is that there is a large difference between the largest and 2nd largest variant. Which in this case is `BlockCommitted` and `TransactionAdded`.

I believe `TransactionAdded` _shrunk_ in size which now triggers this lint. I suspect this is due to the `AccountUpdateDetails` changes in https://github.com/0xMiden/miden-base/pull/2182. Adding a duplicate `account_delta2` field also suppresses the lint conditions again.

It probably makes sense to box the header regardless as it is quite large.

```rust
error: large size difference between variants
  --> crates/proto/src/domain/mempool.rs:16:1
   |
16 | /   pub enum MempoolEvent {
17 | | /     TransactionAdded {
18 | | |         id: TransactionId,
19 | | |         nullifiers: Vec<Nullifier>,
20 | | |         network_notes: Vec<NetworkNote>,
21 | | |         account_delta: Option<AccountUpdateDetails>,
22 | | |     },
   | | |_____- the second-largest variant contains at least 240 bytes
23 | | /     BlockCommitted {
24 | | |         header: BlockHeader,
25 | | |         txs: Vec<TransactionId>,
26 | | |     },
   | | |_____- the largest variant contains at least 456 bytes
27 | |       TransactionsReverted(HashSet<TransactionId>),
28 | |   }
   | |___^ the entire enum is at least 456 bytes
```

Mirko-von-Leipzig: I've box'd the header

PR #1472, opened 2025-12-21
feat: Move block proving to the Store

sergerad:
## Context

We have added the Validator component to perform block validation and signing. As such, blocks now have the following states (conceptually, but also expressed in the code to some extent):
- proposed
- signed
- proven

We now need the block producer to "commit" blocks to the store after they are signed and have the store handle the proving. This PR will move block proving from the block producer to the store. Subsequent PRs will update the store RPC and runtime so that block proving is asynchronous to block committal ("deferred" block proving, so to speak).

Relates to #1316.

## Changes
- Remove block prover URL field from block producer configuration.
- Add block prover URL arg for bundled start command and node start command.
- Move block proving out of block producer to store component.
- Replace `ProvenBlock` usage with constituent parts of block required for proving.

### DIFF
--- a/proto/proto/types/blockchain.proto
+++ b/proto/proto/types/blockchain.proto
@@ -4,15 +4,8 @@ package blockchain;
 import "types/account.proto";
 import "types/primitives.proto";
 
-//// BLOCK
-//// ================================================================================================
-//
-//// Represents a block.
-//message Block {
-//    // Block data encoded using [winter_utils::Serializable] implementation for
-//    // [miden_objects::block::Block].
-//    bytes block = 1;
-//}
+// BLOCK
+// ================================================================================================
 
 // Represents a proposed block.
 message ProposedBlock {


crates/store/src/server/block_producer.rs
```diff
@@ -56,28 +58,45 @@ impl block_producer_server::BlockProducer for StoreApi {
...
 
-        self.state.apply_block(block).await?;
+        // Apply the block to the state.
+        self.state.apply_block(header.clone(), body.clone(), signature.clone()).await?;
+
+        // TODO(currentpr): this need to be a separate task or set of tasks?
+        // Prove the block.
+        self.prove_block(ordered_batches, block_inputs, header, signature, body)
+            .await
+            .map_err(|err| Status::internal(err.as_report_context("failed to prove block")))?;
```
sergerad, line 102: @bobbinth @Mirko-von-Leipzig any thoughts on how we want to factor this now? Do we simply want to be "synchronous" and return error here (and revert the apply_block above?)? Or do we want a separate task that retries block proving failures?

bobbinth: I think this depends on the failure reason:

- If the proving failed because the block is fundamentally unprovable - then we should halt the chain (e.g., by crushing), and then we'll need to roll back the chain to the state priori to the unprovable block.
- If the failure is due to some other reason (e.g., network issues), we should retry.

In general, before we get too far into writing code, could you describe the overall approach you are taking? (probably in the PR description)

sergerad: On second thought it will be best to leave that part of the refactor to a separate PR given this PR is already touching 30+ files.

So the scope of this PR will be to move proving into the store while keeping everything synchronous as it is today. Then we can look at what RPC / runtime changes we want to make in the followup to facilitate the "deferred" nature of proving (so that block committal is asynchronous to block proving).

sergerad: Will specify all of that in the PR description when I'm ready to take it out of draft 👍

crates/proto/src/generated/store.rs
```diff
@@ -1,4 +1,28 @@
...
+    pub header: ::prost::alloc::vec::Vec<u8>,
+    /// Block header encoded using \[winter_utils::Serializable\] implementation for
+    /// \[miden_objects::block::BlockBody\].
+    #[prost(bytes = "vec", tag = "4")]
+    pub body: ::prost::alloc::vec::Vec<u8>,
+    /// Signature encoded using \[winter_utils::Serializable\] implementation for
+    /// \[crypto::dsa::ecdsa_k256_keccak::Signature\].
+    #[prost(bytes = "vec", tag = "5")]
+    pub signature: ::prost::alloc::vec::Vec<u8>,
+}
```
sergerad, line 25: Unsure that this is how we want to structure this (all opaque bytes).

crates/store/src/state.rs
```diff
@@ -225,7 +229,7 @@ impl State {
             return Err(InvalidBlockError::NewBlockInvalidPrevCommitment.into());
         }
 
-        let block_data = block.to_bytes();
+        let block_data = body.to_bytes(); // TODO(currentpr): is this correct?
```
sergerad, line 232: This is used for:
```rust
    async fn get_block_by_number(
        &self,
        request: Request<proto::blockchain::BlockNumber>,
    ) -> Result<Response<proto::blockchain::MaybeBlock>, Status> {
```

I'm not sure what we want to do now that ProvenBlock has been replaced with body and header here. Maybe we change it to SignedBlock when that is added?

Do we want to continue to use this file-based store at all? Could we replace it altogether with data in the db?

Mirko-von-Leipzig: iiuc the separation of raw block data from the structured DB data was originally done so that indexers / explorers could trivially sync this data _without_ locking the database. The long term idea was that this would live on a separate disk / instance and therefore not interfere _at all_ with database IO.

I'm unsure if this usage pattern still holds as it does make it difficult to perform ACID style store IO as this is _external_ to the sqlite. You can see this complication in the store's `apply_block` machinery.

I do think this was done a bit prematurely since its not really something we've used at all (this separation) and its mostly just complicated our code though its not the end of the world.

There are many ways to approach this in general and maybe we should revisit this part of the design. The nice thing about this separation is that it enables syncing "cold" historical data without the main database worrying about retaining it. My view on this is that the sqlite database store's the recent history (where recent is some `N` number of blocks for certain queries), and the file store retains all blocks ever (at least for now).

Now that we need to also retain recent unsigned blocks and then migrate them to signed or potentially rollback invalid blocks this algebra becomes a bit more complex.

---------------------

If we were restarting from scratch I would probably say put it all in the database. And then figure out what the "cold" storage should look like since we now have unsigned and signed blocks. 

As an example design, We could have a cold storage component & API that _only_ stores fully finalized blocks that will never revert or rollback. This allows crawlers/indexers to use this without having to consider rollbacks of this data. The primary store pushes finalized blocks into cold storage and retains some minimal overlap of blocks with it as part of the database itself. What finalized means in our context is debatable -- right now it would mean signed & validated, but it could also eventually mean finalized on agglayer/L1.

Something to consider is that imo horizontally scaling RPC will require horizontal scaling of the database. Whether that means moving to postgres with replication or simply hosting a slave sqlite database with every RPC instance is an open question. Once this is the case, it _may_ also make sense to simply whack the "cold" data in there as well. But probably this makes no sense since cold data should be _relatively_ unused and could "scale" separately.

----------------------------------

I would leave things as they are now, for now, until we settle the above design.

Mirko-von-Leipzig: cc @bobbinth @drahnr

bobbinth: I agree that we should keep the things roughly as they are now and I do think that keeping the "cold data" separately could have advantages in the future (even if done a bit prematurely).

As to what this specific endpoint should return, I think it could still be a `MaybeBlock` message, but it could look like so:

```proto
message MaybeBlock {
    optional Block block = 1;
}

message Block {
    BlockHeader header = 1;
    bytes content = 2;
    BlockSignature signature = 3;
    optional BlockProof proof = 4;
}
```
This implies that blocks could be serialized in two states:

1. Signed but not yet proven.
2. Signed and proven.

Since we first serialize the block without the proof, and then the proof gets added later, we have two options for serializing the proof:

1. We could append proofs to existing block files.
2. We could save the proofs in separate files (e.g., `block134_proof.dat`).

Not sure if there are any significant pros/cons in one approach vs. another - so, don't have a strong opinion on this.

crates/block-producer/src/block_builder/mod.rs
drahnr, line 121: nit/out of scope: we might want to refactor this into a more linear flow with `?`s and maybe a sub function

crates/proto/src/generated/store.rs
drahnr: For signatures and other crypto primitives, bytes are ok; `BlockHeader` is a bag of `Word`s.

`ordered_batches` and `block_inputs` seem better candidates for proper protobuf representation.

I am not sure it needs to be done as part of this PR though.

crates/store/src/db/schema.rs
```diff
@@ -1,5 +1,12 @@
 // @generated automatically by Diesel CLI.
 
+diesel::table! {
+    account_codes (code_commitment) {
+        code_commitment -> Binary,
+        code -> Binary,
+    }
+}
```
drahnr, line 8: was this not present before?

crates/store/src/db/schema.rs
```diff
@@ -100,21 +101,11 @@ diesel::table! {
...
 
-diesel::joinable!(accounts -> account_codes (code_commitment));
-diesel::joinable!(accounts -> block_headers (block_num));
-// Note: Cannot use diesel::joinable! with accounts table due to composite primary key
-// diesel::joinable!(notes -> accounts (sender));
-// diesel::joinable!(transactions -> accounts (account_id));
-diesel::joinable!(notes -> block_headers (committed_at));
-diesel::joinable!(notes -> note_scripts (script_root));
-diesel::joinable!(nullifiers -> block_headers (block_num));
-diesel::joinable!(transactions -> block_headers (block_num));
```
drahnr, line 111: Interesting that we don't need those anymore

PR #1476, opened 2025-12-23
chore: update VM and base version

SantiagoPittella:
This PR updates the VM version to 0.20, and update the `miden-base` dependencies.

### DIFF
--- a/crates/rpc/src/tests.rs
+++ b/crates/rpc/src/tests.rs
@@ -16,6 +16,7 @@ use miden_node_utils::limiter::{
     QueryParamNoteTagLimit,
     QueryParamNullifierLimit,
 };
+use miden_protocol::Word;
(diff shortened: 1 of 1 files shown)


bin/network-monitor/src/deploy/mod.rs
```diff
@@ -174,7 +185,7 @@ pub async fn deploy_counter_account(counter_account: &Account, rpc_url: &Url) ->
     data_store.add_account(counter_account.clone());
 
     let executor: TransactionExecutor<'_, '_, _, BasicAuthenticator> =
-        TransactionExecutor::new(&data_store);
+        TransactionExecutor::new(&data_store).with_debug_mode();
```
bobbinth, line 188: Why do we need to add `.with_debug_mode()` here?

PR #1479, opened 2025-12-23
fix: correct account vault and storage pagination cutoff

phrwlk:
The pagination for account vault assets and storage maps was a bit too aggressive: it treated raw.len() >= MAX_ROWS as “we overflowed”, even though we ask the DB for MAX_ROWS + 1 rows. In the edge case where we get exactly MAX_ROWS rows back, we don't actually know that there's more data, but the code already cuts the last block and moves last_block_included back. This means that we sometimes return less data than we can, and the client has to make an extra request. I changed the condition to raw.len() > MAX_ROWS, as already done for nullifiers. Now we consider a page truncated only when we actually see an extra row, and in other cases we return all available data up to block_range.end(), correctly setting last_block_included.

### DIFF
--- a/crates/store/src/db/models/queries/accounts.rs
+++ b/crates/store/src/db/models/queries/accounts.rs
@@ -301,7 +301,7 @@ pub(crate) fn select_account_vault_assets(
 
     // Discard the last block in the response (assumes more than one block may be present)
     let (last_block_included, values) = if let Some(&(last_block_num, ..)) = raw.last()
-        && raw.len() >= MAX_ROWS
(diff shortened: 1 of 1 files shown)


igamigo (approved):
LGTM! Good catch, thanks. 

I think [originally the constant already had a `+ 1`](https://github.com/0xmiden/miden-node/commit/9b6f108c0a49fdd82a0cb60c74f17c806ae93ae3#diff-0349402692e56351b142741a263b89b477b733382fcc5567eff6bf7e727ebb15R289) but might have been refactored recently and this wasn't caught in that moment

PR #1481, opened 2025-12-29
refactor [1/4]: change accounts db schema

drahnr:
Preparation / carve out from #1394 since the feedback caused additional linecount - ~no~ some minor functional changes were made!

A large chunk of the linecount are testcases

### Core changes

* `schema.rs` - move storage map into separate table, avoid storing full serialized account to `accounts` table, move vault data to separate table (breaking)
* `models/conv.rs` - `StorageSlotType` - change DB representation  (breaking)
* `queries/at_block.rs` - add new `*_at_block` queries 
* adds a few helper functions such as `fn validate_block_exists`

### DIFF
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -71,6 +71,7 @@ impl From<AccountId> for proto::account::AccountId {
 
 // ACCOUNT UPDATE
 // ================================================================================================
+
 #[derive(Debug, PartialEq)]
 pub struct AccountSummary {
     pub account_id: AccountId,
@@ -100,6 +101,30 @@ impl From<&AccountInfo> for proto::account::AccountDetails {
             summary: Some(summary.into()),
             details: details.as_ref().map(Serializable::to_bytes),
         }
+    }
+}
+
+// ACCOUNT STORAGE HEADER
+//================================================================================================
+
+impl TryFrom<proto::account::AccountStorageHeader> for AccountStorageHeader {
+    type Error = ConversionError;
+
+    fn try_from(value: proto::account::AccountStorageHeader) -> Result<Self, Self::Error> {
+        let proto::account::AccountStorageHeader { slots } = value;
+
+        let slot_headers = slots
+            .into_iter()
+            .map(|slot| {
+                let slot_name = StorageSlotName::new(slot.slot_name)?;
+                let slot_type = storage_slot_type_from_raw(slot.slot_type)?;
+                let commitment =
+                    slot.commitment.ok_or(ConversionError::NotAValidFelt)?.try_into()?;
+                Ok(StorageSlotHeader::new(slot_name, slot_type, commitment))
+            })
+            .collect::<Result<Vec<_>, ConversionError>>()?;
+
+        Ok(AccountStorageHeader::new(slot_headers)?)
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -18,17 +18,17 @@ CREATE TABLE accounts (
     block_num                               INTEGER NOT NULL,
     account_commitment                      BLOB NOT NULL,
     code_commitment                         BLOB,
-    storage                                 BLOB,
-    vault                                   BLOB,
     nonce                                   INTEGER,
+    storage_header                          BLOB, -- Serialized AccountStorage from miden-objects
```
bobbinth: Maybe I missed a comment about this somewhere, but why are we serializing full `AccountStorage` into a field named `storage_header`?

If the intention is to serialize full `AccountStorage` (as it was before), then we should probably keep this field named `storage`. If we do want to rename it to `storage_header`, we should have it store `AccountStorageHeader`.

The former is probably acceptable as it means we want to make the updates incrementally - i.e., first migrate the vault structure and then migrate the storage structure. But again, I think the name of the field should reflect its content.

crates/store/src/db/models/queries/accounts/at_block.rs
```diff
@@ -0,0 +1,314 @@
...
+/// # Returns
+///
+/// * `Ok(Some(Vec<u8>))` - The account code bytes if found
+/// * `Ok(None)` - If account doesn't exist or has no code
+/// * `Err(DatabaseError)` - If there's a database error
+pub(crate) fn select_account_code_at_block(
+    conn: &mut SqliteConnection,
+    account_id: AccountId,
+    block_num: BlockNumber,
+) -> Result<Option<Vec<u8>>, DatabaseError> {
```
bobbinth: As I mentioned in my other review, we probably don't need this function. Instead, we need something like `select_account_code_by_commitment()`.

crates/store/src/db/migrations/2025062000000_setup/up.sql
drahnr: The comment is wrong, the actual change of serializing the header only is already fixed/implemented.

crates/store/src/db/models/queries/accounts/at_block.rs
drahnr: Done

crates/store/src/db/models/queries/accounts/at_block.rs
```diff
@@ -0,0 +1,268 @@
...
+
+// ACCOUNT HEADER
+// ================================================================================================
+
+#[derive(Debug, Clone, Queryable)]
+struct AccountHeaderDataRaw {
+    code_commitment: Option<Vec<u8>>,
+    nonce: Option<i64>,
+    storage_header: Option<Vec<u8>>,
+}
```
bobbinth, line 41: Why do we not include `vault_commitment` in `AccountHeaderDataRaw`? If we had it here as well, it would simplify the code below a bit.

crates/store/src/db/models/queries/accounts/at_block.rs
```diff
@@ -0,0 +1,268 @@
...
+/// # Returns
+///
+/// * `Ok(Some(AccountHeader))` - The account header if found
+/// * `Ok(None)` - If account doesn't exist at that block
+/// * `Err(DatabaseError)` - If there's a database error
+pub(crate) fn select_account_header_at_block(
+    conn: &mut SqliteConnection,
+    account_id: AccountId,
+    block_num: BlockNumber,
+) -> Result<Option<AccountHeader>, DatabaseError> {
```
bobbinth: Since we are reading/parsing `AccountStorageHeader` as a part of this function anyways, I wonder if we should also return it. That is, the return type could be `Result<Option<(AccountHeader, AccountStorageHeader)>, DatabaseError>`

crates/store/src/db/models/queries/accounts/at_block.rs
```diff
@@ -0,0 +1,268 @@
...
+
+// ACCOUNT VAULT
+// ================================================================================================
+
+/// Query vault assets at a specific block by finding the most recent update for each `vault_key`.
+pub(crate) fn select_account_vault_at_block(
+    conn: &mut SqliteConnection,
+    account_id: AccountId,
+    block_num: BlockNumber,
+) -> Result<Vec<Asset>, DatabaseError> {
```
bobbinth, line 132: This is probably fine for now, but this function is super expensive. We are basically making one request per asset to the database. For accounts that hold lots of assets, this will be prohibitively expensive.

I wonder if we could somehow leverage the SMT forest data structure for this in the future.

For now, let's add a bit more info to the doc comments explaining that this implementation is a temporary solution.

crates/store/src/db/models/queries/accounts/at_block.rs
```diff
@@ -0,0 +1,268 @@
...
+// ================================================================================================
+
+/// Returns account storage at a given block by reading from `accounts.storage_header`
+/// (which contains the `AccountStorageHeader`) and reconstructing full storage from
+/// map values in `account_storage_map_values` table.
+pub(crate) fn select_account_storage_at_block(
+    conn: &mut SqliteConnection,
+    account_id: AccountId,
+    block_num: BlockNumber,
+) -> Result<AccountStorage, DatabaseError> {
```
bobbinth, line 195: Similar comment as above: this is fine for now, but it will be prohibitively expensive for accounts with any significant number of entries in storage maps.

Let's make a note of this, and start thinking about how we can make this more efficient.

crates/store/src/state.rs
```diff
@@ -987,67 +990,71 @@ impl State {
...
+            Some(_) => {
+                let vault_assets =
+                    self.db.select_account_vault_at_block(account_id, block_num).await?;
+                AccountVaultDetails::from_assets(vault_assets)
+            },
+            None => AccountVaultDetails::empty(),
+        };
 
+        // TODO: don't load the entire store at once, load what is required
+        let store = self.db.select_account_storage_at_block(account_id, block_num).await?;
```
bobbinth: nit: I would use `storage` rather than `store` here (both for variable name and in the comment).

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -873,8 +935,7 @@ pub(crate) fn upsert_accounts(
                 // apply delta to the account; we need to do this before we process asset updates
                 // because we currently need to get the current value of fungible assets from the
                 // account
-                let account_after =
-                    apply_delta(account_before, delta, &update.final_state_commitment())?;
+                let account_after = apply_delta(account, delta, &update.final_state_commitment())?;
```
bobbinth, line 938: Not for this PR, but reading in the full account, applying delta to it, and then saving it back is going to be prohibitively expensive for large accounts. But I think this could be covered by https://github.com/0xMiden/miden-node/issues/1483.

bobbinth (commented):
Looks good! Thank you! I reviewed most non-test code and left some small comments inline.

Overall, I think this structure works for now, but we'll probably need to re-think this shortly after the refactoring is merged as in many places the interactions with the database are likely to be prohibitively expensive.

crates/store/src/db/models/queries/accounts/at_block.rs
drahnr: There is already a TODO at the callsite that the entire function should go away :)

crates/proto/src/domain/account.rs
bobbinth, line 1: Not for this PR: this file is pretty large and difficult to navigate. I think we should break it up into a few smaller files which contain logically grouped functionality.

bobbinth (approved):
Looks good! Thank you! I reviewed all non-test code and left one comment for the future inline.

crates/proto/src/domain/account.rs
drahnr: https://github.com/0xMiden/miden-node/issues/1494

PR #1484, opened 2026-01-05
chore: add grpc-web checks in network monitor

### LINKED ISSUE #1462: Test remote prover through gRPC-Web
Currently, the status monitor shows availability-related data for the remote prover service. This is derived mostly from probing the service by submitting proof requests via the native gRPC client.

An issue we've hit recently is that gRPC-Web clients can fail to connect to the service even when native gRPC works, and we don't detect this today. gRPC-Web uses different request/response framing, is often deployed behind different proxy paths/config, and (since in practice it's mostly used through the browser and cross-origin) commonly triggers CORS preflight checks. As a result, a deployment can look healthy from native gRPC probes while still being unusable from the browser.

To close this gap, we could add a browser-based probe (potentially on-demand) to validate gRPC-Web connectivity. There are a few options here: A simple probe that simply `fetch()`es the prover's status endpoint (even this would catch many CORS/proxy issues) will already yield useful data. A deeper probe that uses the WebClient package (or maybe even the wallet adapter if available) to perform an actual proof request would be great, but not ideal since it adds a soft dependency on the web client stack - but it's worth considering as a more end-to-end signal and also something to keep in mindif we move the monitor codebase to a different repo.

SantiagoPittella:
closes #1462 

Adds an on-demand "Probe gRPC-Web" button to the network monitor dashboard for RPC and Remote Prover services. This allows operators to verify that browser clients can successfully reach these services via gRPC-Web protocol.

This is how it looks (the remote prover are shown as failing due to the usage of incompatible versions):

Screenshot 2026-01-05 at 14 55 06

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,14 @@
 - Added chain tip to the block producer status ([#1419](https://github.com/0xMiden/miden-node/pull/1419)).
 - The mempool's transaction capacity is now configurable ([#1433](https://github.com/0xMiden/miden-node/pull/1433)).
 - Renamed card's names in the `miden-network-monitor` binary ([#1441](https://github.com/0xMiden/miden-node/pull/1441)).
+- Added pagination to `GetNetworkAccountIds` endpoint ([#1452](https://github.com/0xMiden/miden-node/pull/1452)).
 - Improved tracing in `miden-network-monitor` binary ([#1366](https://github.com/0xMiden/miden-node/pull/1366)).
 - Integrated RPC stack with Validator component for transaction validation ([#1457](https://github.com/0xMiden/miden-node/pull/1457)).
+- Added validated transactions check to block validation logc in Validator ([#1460](https://github.com/0xMiden/miden-node/pull/1460)).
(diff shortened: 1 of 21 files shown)


bin/network-monitor/assets/index.js
Mirko-von-Leipzig, line 1: Is the reason for a button just to simplify the trigger conditions? Did you consider a periodic trigger, similar to the others?

SantiagoPittella: There is not an specific issue, I can move it to be periodical if we prefer that way.

bobbinth: Assuming we are pinging only the status endpoint on the prover, I'd go with the periodic pinging approach (rather than requiring the user to click the button).

SantiagoPittella: It is the status endpoint both on the prover and rpc component. Sure, I will move to that approach then.

Mirko-von-Leipzig (approved):

bobbinth (approved):
Looks good! Thank you! Though, this is a very light review from me.

Issue #1305, opened 2025-10-23
Implement get_foreign_account_inputs() for TransactionInputsDataStore


sergerad:
After #1278 is merged.

As per [comment](https://github.com/0xMiden/miden-node/pull/1278#discussion_r2453916387):

>```
>    fn get_foreign_account_inputs(
>       &self,
>       foreign_account_id: AccountId,
>       _ref_block: BlockNumber,
>   ) -> impl FutureMaybeSend<Result<AccountInputs, DataStoreError>> {
>       async move { Err(DataStoreError::AccountNotFound(foreign_account_id)) }
>   }
>```
>
>Returning an error is fine for now, but we'll need to figure out how to read AccountInputs for foreign accounts from TransactionInputs (the data should be in the advice_inputs - but it may not be simple to extract).

And [comment](https://github.com/0xMiden/miden-node/pull/1278#discussion_r2453918699):

>```
>if self.tx_inputs.account().id() != account_id {
>               return >Err(DataStoreError::AccountNotFound(account_id));
>           }
>```
>Similar to the above comment, we should be able to serve the foreign account requests. Maybe the issue we need is about the ability to support foreign account invocation during re-execution.

bobbinth:
The code that adds foreign account inputs into the advice provider is located [here](https://github.com/0xMiden/miden-base/blob/next/crates/miden-protocol/src/transaction/kernel/advice_inputs.rs#L118) - and we'll need to do the reverse of that (i.e., we'll need to read `AccountInputs` from the advice provider). This involves two things:

First, we need to read `PartialAccount` from the advice map. To get the account header we'll need to get the elements located under `account_id_key` in the map, and then parse them into `AccountHeader`. Once we have the header, we can read `AccountStorageHeader` from the map in a similar way, and also get `AccountCode` from `TransactionInputs.foreign_account_code`.

After we build a `PartialAccount`, we'll also need to get `AccountWitness`. The main thing there is the Merkle path, and this one will need to come from the Merkle store. To do this, we'll need to use `MerkleStore::get_path()` method passing in account tree root (from the reference block) as the `root` and deriving index from `account_id` (for this, we should probably add a simple conversion function in `miden-base` - it could build on `account_id_to_smt_key()` function).

Overall, this process is pretty involved, and I'm wondering how much of it should live in `miden-base` because since the logic of putting `AccountInputs` into the advice provider is there, it probably makes the most sense to put the logic for reading `AccountInputs` from the advice provider there too.

PR #1486, opened 2026-01-05
feat: Support foreign accounts 

### LINKED ISSUE #1387: Make FPI call possible during network transaction

End user

For the outlined Agglayer bridge implementation in https://github.com/0xMiden/miden-base/issues/1910, it necessary to be able to make FPI calls in transactions executed by the network transaction builder. 

Example in the context of bridging in using Agglayer:

`CLAIM` note -> Agglayer Faucet (FPI to Agglayer Bridge) -> `P2ID` note

The `CLAIM` note is created by the user and is executed against the Agglayer faucet in a network transaction. During consumption, the P2ID note is created. 

@bobbinth You mentioned this issue was brought up in a separate thread in a different issue, I couldn't find that thread so I am creating this issue. 

### How is this feature used?

This feature would be used in the context of bridging in assets using the Agglayer. The user creates a `CLAIM` note which is consumed by the "Agglayer faucet". During this transaction, the Agglayer faucet makes an FPI call to the Agglayer bridge contract which contains the MMR structure to check if the supplied merkle proof in the `CLAIM` note is valid. If so, the Agglayer faucet issues the note which is commited to in the Agglayer MMR structure. 

### Anything else?

Afaik this is currently possible, but as @igamigo pointed out, the entire `Account` which is FPI called into would be loaded, not the specific data that would need to be read. I think this is the optimization @bobbinth you were referring to that would need to be made.

The absence of the feature **does not** block development of the bridging in components in Miden base, but would be required when testing out the end to end flow on testnet.

sergerad:
## Context

Neither the Validator's `TransactionInputsDataStore` nor the NTX Builder's `NtxDataStore` handle foreign accounts for the purposes of the following functions of `DataStore` trait:
* `get_vault_asset_witnesses()`

NOTE: This will remain in draft until miden-base has the `TransactionIniputs::read_foreign_account_inputs()` functionality merged into next.

Closes #1387 and #1305.

## Changes
* Update `TransactionInputsDataStore` to handle foreign accounts through `self.tx_inputs`.
* Update `NtxDataStore` to handle foreign accounts by accessing `AccountProofResponse` from the store.

### DIFF
--- a/crates/validator/src/tx_validation/data_store.rs
+++ b/crates/validator/src/tx_validation/data_store.rs
@@ -121,7 +121,10 @@ impl DataStore for TransactionInputsDataStore {
                 let foreign_inputs = self
                     .tx_inputs
                     .read_foreign_account_inputs(account_id)
-                    .map_err(|_| DataStoreError::AccountNotFound(account_id))?;
+                    .map_err(|err| DataStoreError::Other {
+                        error_msg: format!("failed to read foreign account inputs: {err}").into(),
+                        source: Some(Box::new(err)),
+                    })?;
 
                 // Search through the foreign account's partial storage maps.
                 let storage_map_witness =


crates/validator/src/tx_validation/data_store.rs
```diff
@@ -53,7 +53,17 @@ impl DataStore for TransactionInputsDataStore {
...
     ) -> impl FutureMaybeSend<Result<AccountInputs, DataStoreError>> {
-        async move { Err(DataStoreError::AccountNotFound(foreign_account_id)) }
+        async move {
+            if foreign_account_id == self.tx_inputs.account().id() {
+                // todo(currentpr): proper error
+                return Err(DataStoreError::AccountNotFound(foreign_account_id));
+            }
+
+            let foreign_inputs =
+                self.tx_inputs.read_foreign_account_inputs(foreign_account_id).unwrap(); // todo unwrap
```
sergerad: @bobbinth do we need the impl for `TransactionInputsDataStore`, rather than just `NtxDataStore`?

If yes, do you imagine the impl looking like this - in the sense that account inputs are derived from tx inputs. The `read_foreign_account_inputs` fn is a stub for now in a branch of mine, just trying to flesh out integration first.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -341,7 +341,17 @@ impl DataStore for NtxDataStore {
...
     ) -> impl FutureMaybeSend<Result<AccountInputs, DataStoreError>> {
-        async move { Err(DataStoreError::AccountNotFound(foreign_account_id)) }
+        async move {
+            if foreign_account_id == self.account.id() {
+                // todo(currentpr): proper error
+                return Err(DataStoreError::AccountNotFound(foreign_account_id));
+            }
+            // TODO(currentpr): We don't have tx inputs here so what do we do?
+            let foreign_inputs =
+                self.tx_inputs.read_foreign_account_inputs(foreign_account_id).unwrap(); // todo unwrap
```
sergerad: @bobbinth assuming we expect to get foreign account inputs from tx inputs, is it possible to do so here? Should we somehow retrieve the tx inputs through the store or something?

bobbinth: I'm not sure we can get foreign account inputs from the original TX inputs - or at least it may not always work.

I think to get foreign account inputs, we'd need to make a request to the store and load the account from there. We may also want to cache the foreign account `NtxDataStore` so that if there are additional requests, we won't need to go to the store again for the same data.

There is one nuance here though: currently, we are able to fetch the full account from the store, but this functionality will go away after the refactoring @drahnr is working on. So, it may make sense to add a dedicated endpoint to the data store that would return the data sufficient to construct `AccountInputs` (i.e., the `PartialAccount` and `AccountWitness` structs).

crates/validator/src/tx_validation/data_store.rs
bobbinth: I think here we should be able to get all the data from the `TransactionInputs` - so, it would work differently from `NtxDataStore`. Basically, the main difference between the validator and NTX builder is:

- NTX builder: executes a transaction from scratch and so it doesn't now what execution path will be taken during transaction execution. For example, it doesn't know which foreign accounts would need to be loaded ahead of time, and so, we need to load them dynamically (i.e., by fetching the requested accounts from the store).
- Validator: re-executes transaction that was previously executed by the user/NTX builder. All foreign account data should already be loaded into the `TransactionInputs`. The main challenge here would be to figure out how to extract the relevant data from `TransactionInputs` as it is not too structured. I believe I left a comment somewhere sketching out how to extract this data (and maybe putting some logic for this in `miden-base` would simplify things).

crates/ntx-builder/src/actor/execute.rs
sergerad: Any advice on how to do this? Seems like we would have to get the full account from store state, convert that to partial account + get account witness somehow.
> it may make sense to add a dedicated endpoint to the data store that would return the data sufficient to construct AccountInputs (i.e., the PartialAccount and AccountWitness structs).

bobbinth: While we have the ability to get the full account from the store, things should be pretty easy:

- The get the partial account, we get the full account and then convert it to partial account (we have `impl From<&Account> for PartialAccount` implemented).
- The witness we'd get from the `AccountTree` - that should be pretty straight-forward too.

I think both of these we should be able to get from the current `GetAccountProof` endpoint. But soon, the ability to get full account will go away.

After a series of @drahnr's PRs gets merged, `GetAccountProof` becomes `GetAccount` and it no longer would return the full account. But it would still return enough info to build a `PartialAccount` + account witness. For partial account, all we really need is `account_id`, vault root, storage header, code, and nonce - and I believe all of this should be available from the `GetAccount` endpoint.

drahnr: > After a series of @drahnr's PRs gets merged, GetAccountProof becomes GetAccount and it no longer would return the full account. But it would still return enough info to build a PartialAccount + account witness. For partial account, all we really need is account_id, vault root, storage header, code, and nonce - and I believe all of this should be available from the GetAccount endpoint.

All of this will be available once #1385 is merged as `GetAccount`

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -339,9 +340,45 @@ impl DataStore for NtxDataStore {
...
+            })?;
+            let account =
+                Account::try_from(&account_details).map_err(|err| DataStoreError::Other {
+                    error_msg: format!("failed to convert account details to account: {err}")
+                        .into(),
+                    source: Some(Box::new(err)),
+                })?;
+
+            // Return partial account and witness.
+            let partial_account = PartialAccount::from(&account);
```
bobbinth: This could be fine for now, but we should also be able to go directly from `AccountDetails` to `PartialAccount` (without having to convert to `Account` in-between). This will also be better long-term because once @drahnr's PRs get merged, we won't be able to get full account from the store.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -339,9 +340,45 @@ impl DataStore for NtxDataStore {
     fn get_foreign_account_inputs(
         &self,
         foreign_account_id: AccountId,
-        _ref_block: BlockNumber,
+        ref_block: BlockNumber,
```
bobbinth, line 343: `ref_block` passed into this function should be the same as the reference block of this data store (i.e., `self.reference_header.block_num()`). We should add an `assert_eq!()` to double-check this.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -350,26 +387,41 @@ impl DataStore for NtxDataStore {
...
+                    })?;
+                let asset_vault =
+                    AssetVault::new(&account_details.vault_details.assets).map_err(|err| {
+                        DataStoreError::Other {
+                            error_msg: format!("failed to create asset vault: {err}").into(),
+                            source: Some(Box::new(err)),
+                        }
+                    })?;
+
+                get_asset_witnesses(vault_keys, &asset_vault)
```
bobbinth: A couple of comments on this approach:

First, when we make a request to the store, we should use the reference block of this data store (i.e., `self.reference_header.block_num()`.

But also, getting the full account proof is very heavy compared to what we need here. Maybe that's OK for this PR, but we should follow it up with using dedicated endpoints on the store to get asset/storage witnesses. Let's create an issue for this.

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -63,50 +82,67 @@ impl DataStore for TransactionInputsDataStore {
...
-            if self.tx_inputs.account().id() != account_id {
-                return Err(DataStoreError::AccountNotFound(account_id));
-            }
-
             if self.tx_inputs.account().vault().root() != vault_root {
                 return Err(DataStoreError::Other {
                     error_msg: "vault root mismatch".into(),
                     source: None,
                 });
             }
```
bobbinth: I'm not sure this check is relevant any more. This basically prevents us from processing foreign accounts.

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -63,50 +82,67 @@ impl DataStore for TransactionInputsDataStore {
...
+                get_asset_witnesses_from_account(self.tx_inputs.account(), vault_keys)
+            } else {
+                let foreign_inputs = self
+                    .tx_inputs
+                    .read_foreign_account_inputs(account_id)
+                    .map_err(|err| DataStoreError::Other {
+                        error_msg: format!("failed to read foreign account inputs: {err}").into(),
+                        source: Some(Box::new(err)),
+                    })?;
+                get_asset_witnesses_from_account(foreign_inputs.account(), vault_keys)
```
bobbinth, line 96: I'm not sure if extracting account inputs first and then trying to get asset witness from the account inputs will work (though, it depends on how `read_foreign_account_inputs()` works).

An alternative approach would be to read the asset witness directly. It should actually be much simpler than reading the whole account inputs object (we should be able to look up the path based on the `vault_root` in the Merkle store).

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -63,50 +82,67 @@ impl DataStore for TransactionInputsDataStore {
...
+                        }
+                    });
+                storage_map_witness
+                    .ok_or_else(|| DataStoreError::Other { error_msg: "todo".into(), source: None })
+            } else {
+                // Get foreign account inputs.
+                let foreign_inputs = self
+                    .tx_inputs
+                    .read_foreign_account_inputs(account_id)
+                    .map_err(|_| DataStoreError::AccountNotFound(account_id))?;
```
bobbinth: Similar comment as above, but about reading storage witnesses directly rather than going though account inputs.

sergerad:
Closing in favour of #1521

Issue #1439, opened 2025-12-09
Improve mempool graph replacement API


Mirko-von-Leipzig:
Some initial context here https://github.com/0xMiden/miden-node/pull/1396#discussion_r2572894647.

### Overview

The mempool is a DAG with nodes representing transactions, batches and blocks. Edges represent a parent/child dependency between the nodes e.g. `a` depends on `b` if it consumes a note created by `b`, or `a`'s account update follows on from `b`s.

High level graph operations include 
- adding new transaction nodes
- replacing a set of nodes with a single node
  - e.g. transforming a set of transaction nodes into a batch node
- replacing a single node with a set of nodes
  - e.g. reverting a batch node and re-inserting its transactions

This is currently facilitated by having _no_ continuity enforcement within an account's state tracking. We allow this so that at the high level node replacements can be implemented by simply removing the old node(s) and inserting the new node(s). This is simple but means the graph state is allowed to be in an invalid, disconnected state while this shuffling happens and there is no check or enforcement in place to double-check for bugs. 

The high level code is responsible for maintaining the invariants which at the very least will be prone to refactoring bugs. It also makes the low-level code look suspicious in isolation.

We want to improve this by changing the API slightly to account for the different use cases. There will be some possible sharp edges so this should begin as an experiment - we do want some flexibility so that adding and removing nodes isn't _that_ difficult.

### Potential API

```rust
/// Adds a new transition.
fn push(&mut self, node: NodeId, from: Word, to: Word) { ... }

/// Replaces `target` node with a contiguous series of updates
fn unfold(&mut self, target: &NodeId, to: &[(NodeId, Word)]) {...}

/// Replaces a contiguous range of updates with a single new NodeId
fn fold(&mut self, range: Range<NodeId>, to: NodeId) { ... }

/// Removes a node transition.
fn remove(&mut, node: NodeId) { ... }
```

The above is an example of the new account state API. This would replace the current `insert/remove` API. The idea is simply that each operation _must_ leave the account transitions in a valid, continuous state. The caller is therefore responsible for ensuring e.g. that `unfold`s target node is equivalent to the set of NodeIds its replacing. Or that the node being removed is either at the top or bottom of the stack.

There will be some friction here since we must also allow for pass-through transactions but it should be doable.

[...]

PR #1488, opened 2026-01-06
Mempool: rework account state updates

Mirko-von-Leipzig:
I've pulled this out of my ongoing mempool improvements as I think this is easier to review in isolation.

This is a rework of the mempool's account state tracking but implemented in a separate file while leaving the original in place. This is because the new implementation isn't directly usable without quite a few high level changes. We can do a feature branch again but I think maybe this way is safer.

## Changes

The structure has been reworked from a map of `(from, to)` pairs to a sequential `[a, b, c, d]` style implementation. This allows us to `assert!` invariants more aggressively and works towards #1439.

I've changed the `insert`/`remove` api to 
- `append` since new transaction nodes can only be appended, and 
- `evict`/`prune` to pop the youngest and prune the oldest nodes

This lets us assert quite a few more invariants as well. I've also changed the removal methods to consume and only return `self` if the account is non-empty. This will help with type safety on the caller side.

I've renamed `children` to `required_by` and `parents` to `depends_on` though now that I'm reading it again I'm unsure if this actually helps. The core problem is that "sibling" pass through nodes make the parent/child names confusing.

## Hiccups

I've run into some issues implementing the `fold` and `unfold` methods. Its fairly painful to do because one has to ensure that they're contiguous _and_ don't leave any dangling pass through nodes behind. However sometimes this is okay still if its on the edges of the mutation.

~~I've left these methods unimplemented for now to at least get the review started on the rest.~~

I've implemented these now in a way I hope makes sense. ~~I'll be adding tests to build some confidence as this was trickier than I had originally thought.~~ I've added property tests for test and the pruning & eviction code. This turned up a bunch of edge cases which are now fixed.

I'm left a bit jaded at how complicated this felt. I'm open to ideas on alternate representations that might simplify things somehow. I do wonder if my overall mempool design is a good idea; or if this is an example of complications arising from poor external choices :D

## Timeline

I've got the surrounding code which uses this implemented; though I'll likely split that up into smaller PRs as well.


crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,267 @@
...
+
+use std::collections::{HashSet, VecDeque};
+
+use itertools::Itertools;
+use miden_protocol::Word;
+use miden_protocol::block::BlockNumber;
+
+use crate::mempool::NodeId;
+
+/// The commitment updates made to a account.
```
igamigo: Surprised `typos` did not catch this ("an account")

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,267 @@
...
+/// Describes a state transition of an account.
+#[derive(Clone, Debug, PartialEq)]
+struct Update {
+    /// The account state commitment.
+    commitment: Word,
+    /// The node which caused the transition to `commitment` state.
+    by: NodeId,
+    /// Nodes which have both initial and final state `commitment`. These are known as pass through
+    /// nodes.
+    pass_through: HashSet<NodeId>,
```
igamigo, line 55: For the purposes of keeping a consistent account commitment state history, it's fine to keep all node IDs in a set, but just making a note that each tx does affect the network's state (they consume and/or create notes), so if there are any semantics (or if they are to be expected int he future) where these would have to be somehow sequenced relative to each other, this could affect that. `evict()` does somewhat imply ordering, but it seems like this ordering would be imposed/handled by users of this struct.

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// - otherwise it returns node `b -> c` and all pass through nodes `b -> b`
+    ///
+    /// # Panics
+    ///
+    /// Panics if the target node does not exist.
+    pub(super) fn required_by(&self, target: NodeId) -> HashSet<NodeId> {
+        for (from, to) in self.updates.iter().tuple_windows() {
+            if from.by == target {
+                let mut output = from.pass_through.clone();
+                output.insert(to.by);
```
igamigo, line 240: This seems to be missing `return output;`, right? Or maybe I'm missing something

crates/block-producer/src/mempool/state/account.rs
Mirko-von-Leipzig: I think it only does words and not grammar

crates/block-producer/src/mempool/state/account.rs
Mirko-von-Leipzig: Yes this strictly only holds from the dependency perspective of the account commitment itself.

Note dependency would be handled external to this struct.

crates/block-producer/src/mempool/state/account.rs
Mirko-von-Leipzig: Correct thank you!

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,582 @@
...
+        // Update the "back" element which should now be next to "front".
+        self.updates[istart + 1].by = folded;
+        for target in &back.pass_through {
+            self.updates[istart + 1].pass_through.remove(target);
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
drahnr, line 398: nit: I prefer to split the tests out into separate files (yes, I am a hypocrite sometimes), so both merge conflicts and file lengths stay in-check

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,582 @@
...
+    fn is_pass_through(&self) -> bool {
+        self.from == self.to
+    }
+}
+
+/// Describes a state transition of an account.
+#[derive(Clone, Debug, PartialEq)]
+struct Update {
+    /// The account state commitment.
+    commitment: Word,
```
drahnr, line 46: ~is this the state _before_ or _after_ applying the update or the update _itself_?~

Adding a comment that this the commitment after the update was applied would be good.

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,582 @@
...
+    ///
+    /// Panics if the node has any descendents i.e. if its not the youngest update.
+    #[must_use = "account may or may not be empty"]
+    pub(super) fn evict(mut self, id: NodeId) -> Option<Self> {
+        // SAFETY: we are guaranteed to always have at least one element.
+        let last = self.updates.back_mut().unwrap();
+
+        if last.pass_through.is_empty() {
+            assert!(last.by == id, "failed to evict {id:?} from account");
+            // This will never pop the [0] element as [0].by is always set to an unmatchable ID.
```
drahnr, line 135: Can't we make it a separate field to enforce presence at construction? It would avoid quite a bit of SAFETY comments at the cost of maybe a helper function or two.

crates/block-producer/src/mempool/state/account.rs
Mirko-von-Leipzig: This is the commitment such that `self.commitment = self.by.final_account_commitment`.

So after this update is applied, the account's commitment would be `self.commitment`.

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,582 @@
...
+
+        // Check the first element which the above loop skips. The first element can only contain
+        // pass through nodes, and by definition these can't depend on anything if they do exist.
+        //
+        // SAFETY: We are guaranteed at least one element.
+        if self.updates.front().unwrap().pass_through.contains(&target) {
+            return HashSet::default();
+        }
+
+        panic!("Target node {target:?} not found");
```
drahnr, line 220: nit: `unreachable!`

crates/block-producer/src/mempool/state/account.rs
Mirko-von-Leipzig: I allude to this [here](https://github.com/0xMiden/miden-node/pull/1488/changes#diff-73f0249bfc132891174549927e6fb13f36a4326d2e0cd866eb9598d0690b80fdR19-R26), but its not really just a helper function or two.

In particular, the `fold` and `unfold` methods become much worse.

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,582 @@
...
+    /// implementation.
+    updates: VecDeque<Update>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub(super) struct AccountDelta {
+    pub(super) id: NodeId,
+    pub(super) from: Word,
+    pub(super) to: Word,
+}
```
sergerad, line 34: The semantics of id/from/to are nebulous to me in this context. Explanatory comments would help me out here 🙏

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,582 @@
...
+
+    /// Appends the account update as the latest account commitment _IFF_ `from` matches the current
+    /// account commitment.
+    ///
+    /// # Errors
+    ///
+    /// Returns the current account commitment if the above precondition does not hold.
+    pub(super) fn append(&mut self, delta: &AccountDelta) -> Result<(), Word> {
+        // SAFETY: We are guaranteed to always have at least one element.
+        let latest = self.updates.back().unwrap().commitment;
```
sergerad, line 105: nit: prefer `expect("We are guaranteed to always have at least one element")` over unwrap and safety comment.

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,582 @@
...
+    /// Prunes the oldest node from the account state. Note that this title may be shared if the
+    /// oldest commitment has multiple pass-through nodes.
+    ///
+    /// Returns `None` if the account is now empty, otherwise returns the modified account state.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the node has any ancestors in the account i.e. if its not the oldest update.
+    #[must_use = "account may or may not be empty"]
+    pub(super) fn prune(mut self, id: NodeId) -> Option<Self> {
```
sergerad, line 153: Is there a reason we wouldn't use pop_front and pop_back semantics instead? Or something clearer than evict/prune (I think neither of those words imply position of removal?).

pop_latest, pop_oldest? or youngest/oldest?

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,582 @@
...
+        // SAFETY: we are guaranteed to always have at least one element.
+        let front = self.updates.front_mut().unwrap();
+
+        if front.pass_through.is_empty() {
```
sergerad, line 157: Comment explaining when this is expected to be the case would be helpful

crates/block-producer/src/mempool/state/account.rs
```diff
@@ -0,0 +1,582 @@
...
+    /// It is guaranteed to have at least one entry. This is enforced by only allowing a non-empty
+    /// set to be created, and removal of an update returns `None` if the account is empty.
+    ///
+    /// The first (`updates[0]`) update's `by` field is always invalid, set to [`Update::INVALID`]
+    /// and is ignored internally. See the [`evict`](Self::evict) and [`prune`](Self::prune)
+    /// methods.
+    ///
+    /// An alternate representation would be splitting `updates[0]` out into its own field without
+    /// `by`, but in practice this caused much more code complexity in the rest of the
+    /// implementation.
```
sergerad, line 25: The alternate representation you mention here would definitely be more expressive. Agree that limiting complexity is high value consideration here too

crates/block-producer/src/mempool/state/account.rs
Mirko-von-Leipzig: I agree; the problem I have with `pop_front` is that technically there are many that could be `front` due to pass through transactions. But maybe semantically that's still better than inventing new lingo.

So far we have been using `prune` to mean removing _old_ committed state that's no longer needed so I think that's okay; though it is still confusing.

I'll think on this some more because I think you're right, and this is quite important to give the correct impression for reading.

crates/block-producer/src/mempool/state/account.rs
Mirko-von-Leipzig: @drahnr implemented a PoC for this better representation while keeping the complexity to somewhat reasonable bounds. I'll probably switch to that.

Mirko-von-Leipzig:
> - Nodes are a combination of account commitment + something else. This something else could be a counter. So, every time we add a transaction to the graph, if the state of the account doesn't change, the counter gets incremented.

This essentially adds our own ordering on pass through txs, which would work for us, but is overly restrictive from the protocol sense. Pass through txs are technically selectable in _any_ order (barring note interactions).

There are a variety of approaches here, but I think simplest is to revert back to a mixture of what we have now, and what the original mempool implementation was. Separate graphs for txs and batches, that way one doesn't have to fold/unfold. This makes _user batches_ a bit weird to implement (which was the original refactoring reason), but I think I've got an alternative there. The difference will be that instead of a global state, we'll have state tracking for each of the graphs which will also add more robustness / checking at each stage.

PR #1489, opened 2026-01-07
fix: increase pagination limit, return chain tip if exceeded

SantiagoPittella:
Currently, starting the node is failing due to max iterations being reached. This was caused by the store returning the `block_from` instead of the chain tip if it was exceeded, and missing validations of the block range. 

I added the same checks that we use for the other paginated endpoints, increased the limit in the client, we now return the chain tip if the reached block number exceeds it, and updated the check to be greater or equal instead of just equal to the chain tip.

### DIFF
--- a/crates/store/src/server/ntx_builder.rs
+++ b/crates/store/src/server/ntx_builder.rs
@@ -173,7 +173,7 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
     ) -> Result<Response<proto::store::NetworkAccountIdList>, Status> {
         let request = request.into_inner();
 
-        let chain_tip = self.state.latest_block_num().await;
+        let mut chain_tip = self.state.latest_block_num().await;
         let block_range =
             read_block_range::<GetNetworkAccountIdsError>(Some(request), "GetNetworkAccountIds")?
                 .into_inclusive_range::<GetNetworkAccountIdsError>(&chain_tip)?;
@@ -187,6 +187,8 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
         if last_block_included > chain_tip {
             last_block_included = chain_tip;
         }
+
+        chain_tip = self.state.latest_block_num().await;
 
         Ok(Response::new(proto::store::NetworkAccountIdList {
             account_ids,


sergerad (approved):

crates/store/src/server/ntx_builder.rs
```diff
@@ -171,19 +171,23 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
...
-        let (account_ids, last_block_included) =
+        let (account_ids, mut last_block_included) =
             self.state.get_all_network_accounts(block_range).await.map_err(internal_error)?;
 
         let account_ids: Vec<proto::account::AccountId> =
             account_ids.into_iter().map(Into::into).collect();
 
+        if last_block_included > chain_tip {
+            last_block_included = chain_tip;
+        }
```
Mirko-von-Leipzig, line 189: iiuc this is handling a race condition and we're essentially lying about the `chain_tip` since `last_block_included` exceeds it?

What _should_ be happening is that `chain_tip` and the result are fetched within the same database transaction so that they're atomic.

In general the API isn't really a good match for what we use it for - there isn't really a need to have a block range, but rather just a starting block. But since we're (hopefully) reworking this soon, let's just patch it by doing an extra call _after_ we've fetched the data.


```suggestion
        chain tip = self.state.latest_block_num().await;
```

Mirko-von-Leipzig (changes requested):

crates/store/src/server/ntx_builder.rs
SantiagoPittella: > iiuc this is handling a race condition and we're essentially lying about the chain_tip since last_block_included exceeds it?

We are always returning the proper chain tip. A few lines under that we have:
```rust
        Ok(Response::new(proto::store::NetworkAccountIdList {
            account_ids,
            pagination_info: Some(proto::rpc::PaginationInfo {
                chain_tip: chain_tip.as_u32(),
                block_num: last_block_included.as_u32(),
            }),
        }))
```
The issue was that in the case that one page was able to return all the entries, the last block included was set to be the one sent by the user, which could be higher than the chain tip. Basically, we were saying that we included blocks past the tip, which is not true since those not even exists. So I add that step to truncate the last included block to the chain tip.

 > What should be happening is that chain_tip and the result are fetched within the same database transaction so that they're atomic.
 
 About the atomicity, you are right on this, a new block can appear between the initial chain tip and the account IDs retrieval. I added the new fetch of the chain tip as you mentioned. Though it makes me think that there is a case when between a function the account IDs and the last chain tip fetch, a new block is created so the chain tip can be `last_included_block + N`, causing the client to be always out of sync and retrying. This case will be solved when using a stream too.

Mirko-von-Leipzig: Ah I see so underneath is also a weird behaviour etc. Pagination is a pain :upside_down_face: 

I think this is good enough for now. But definitely in this internal API we always want to fetch _all_ accounts sequentially until we reach the end. So a streaming API works, as does a pagination _without_ an end point i.e. just a continuation token set to something like `rowid`.

PR #1490, opened 2026-01-07
fix: send tx_inputs, use existing asset witness

SantiagoPittella:
After https://github.com/0xMiden/miden-node/pull/1460 , to verify a block the transactions needs to be verified too. We achieve that by sending the `TransactionInputs` when submitting.

Also, simplifies and fixes the validator's `DataStore::get_vault_asset_witnesses` implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@
 - Fixed missing asset setup for full account initialization ([#1461](https://github.com/0xMiden/miden-node/pull/1461)).
 - Fixed validator to use pre-stored asset witnesses from the transaction inputs instead of trying to open the partial vault ([#1490](https://github.com/0xMiden/miden-node/pull/1490)).
 - Fixed `GetNetworkAccountIds` pagination to return the chain tip ([#1489](https://github.com/0xMiden/miden-node/pull/1489)).
+- Fixed validator to use pre-stored asset witnesses from the transaction inputs instead of trying to open the partial vault ([#1490](https://github.com/0xMiden/miden-node/pull/1490)).
+- Fixed the network monitor counter account to use the storage slot name ([#1492](https://github.com/0xMiden/miden-node/pull/1492)).
 
 ## v0.12.6
 
--- a/bin/network-monitor/src/assets/counter_program.masm
+++ b/bin/network-monitor/src/assets/counter_program.masm
@@ -1,7 +1,7 @@
 # Counter program for network monitoring with note authentication
 # Storage layout:
-# - Slot 0: counter value (u64)
-# - Slot 1: authorized wallet account id as [prefix, suffix, 0, 0]
+# - OWNER_SLOT: authorized wallet account id as [prefix, suffix, 0, 0]
+# - COUNTER_SLOT: counter value (u64)
 
 use miden::core::sys
 use miden::protocol::active_account
@@ -11,14 +11,14 @@ use miden::protocol::account_id
 use miden::protocol::tx
 
 
-# The slot in this component's storage layout where the counter is stored.
 const COUNTER_SLOT = word("miden::monitor::counter_contract::counter")
+const OWNER_SLOT = word("miden::monitor::counter_contract::owner")
 
 # Increment function with note authentication
 # => []
 pub proc increment
-    # Ensure the note sender matches the authorized wallet stored in slot 1.
-    push.1 exec.active_account::get_item
+    # Ensure the note sender matches the authorized wallet.
+    push.OWNER_SLOT[0..2] exec.active_account::get_item
     # => [owner_prefix, owner_suffix, 0, 0]
 
     exec.active_note::get_sender
--- a/bin/network-monitor/src/counter.rs
+++ b/bin/network-monitor/src/counter.rs
@@ -43,6 +43,7 @@ use tracing::{error, info, instrument, warn};
 
 use crate::COMPONENT;
 use crate::config::MonitorConfig;
+use crate::deploy::counter::COUNTER_SLOT_NAME;
 use crate::deploy::{MonitorDataStore, create_genesis_aware_rpc_client, get_counter_library};
 use crate::status::{
     CounterTrackingDetails,
@@ -83,7 +84,7 @@ async fn get_genesis_block_header(rpc_client: &mut RpcClient) -> Result<BlockHea
     Ok(block_header)
 }
 
-/// Fetch the latest nonce of the given account from RPC.
+/// Fetch the latest counter value of the given account from RPC.
 async fn fetch_counter_value(
     rpc_client: &mut RpcClient,
     account_id: AccountId,
@@ -95,9 +96,13 @@ async fn fetch_counter_value(
         let account = Account::read_from_bytes(&raw)
             .map_err(|e| anyhow::anyhow!("failed to deserialize account details: {e}"))?;
 
-        let storage_slot = account.storage().slots().first().expect("storage slot is always value");
-        let word = storage_slot.value();
-        let value = word.as_elements().last().expect("a word is always 4 elements").as_int();
+        // Access the counter slot by name to avoid index-ordering issues
+        let word = account
+            .storage()
+            .get_item(&COUNTER_SLOT_NAME)
+            .context("failed to get counter storage slot")?;
+
+        let value = word.as_elements().first().expect("a word is always 4 elements").as_int();
 
         Ok(Some(value))
     } else {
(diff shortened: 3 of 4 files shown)


crates/validator/src/tx_validation/data_store.rs
```diff
@@ -74,20 +74,24 @@ impl DataStore for TransactionInputsDataStore {
...
+                            source: None,
                         })
-                    },
-                    Err(err) => Err(DataStoreError::Other {
-                        error_msg: "failed to open vault".into(),
-                        source: Some(err.into()),
-                    }),
-                }
-            }))
+                })
```
igamigo, line 93: AFAIK this change is fine because `tx_inputs.asset_witnesses()` only contains prefetched assets. On re-execution, `get_vault_asset_witnesses()` would only run for prefetched assets anyway since `TransactionExecutor` prefetches regardless of context and anything “dynamic” would otherwise be loaded from advice data directly.

cc @PhilippGackstatter is there a reason we keep prefetched asset witnesses separate? If not, could they be removed `TransactionInputs`? Then either:

- the executor can check whether the prefetched assets are already present in the advice map (and avoid the data store call if so), or maybe
- we can add a specific executor entrypoint/flag that skips prefetching because the VM will load them from the advice map

If not, I think the behavior of `get_vault_asset_witnesses()` only being called for prefetched assets when re-executing might be  worth documenting (unless it already is somewhere) because it can be misleading at first sight.

PhilippGackstatter: The original motivation is described here: https://github.com/0xMiden/miden-base/pull/2113#discussion_r2575464984

Now that we have `TransactionAdviceInputs` in `miden-protocol`, at first glance, it looks like we could revisit this and add asset witnesses directly to `TransactionInputs::advice_inputs` in `TransactionInputs::with_asset_witnesses`.
Is this what you mean? If yes, then I can open an issue for this in base.

> I think the behavior of `get_vault_asset_witnesses()` only being called for prefetched assets when re-executing might be worth documenting (unless it already is somewhere) because it can be misleading at first sight.

I understand the cause for confusion. The behavior, that some data is added to advice inputs after execution, seems like a property of the `TransactionExecutor`. So, would it be sufficient to add a doc section on re-execution on `TransactionExecutor` that describes what data fetched from `DataStore` is added to the advice inputs after execution? This would then serve as a reference for this kind of question and we could note that this implies certain methods may not be called during re-execution.

Mirko-von-Leipzig: iiuc then the change above relies on the fact that the provided `TransactionInputs` is the one used to actually execute the `ProvenTransaction`? And that execution adds additional data to "inputs", making it an execution output artifact?

If so, then this is definitely confusing and will lead to issues since this is _passed_ in by users.

igamigo: > Is this what you mean? 

I think so, but not sure if that is sufficient. What I'm mainly trying to get at is that it seems that on the re-execution case, dynamic calls to `DataStore::get_vault_asset_witnesses()` are basically skipped because the Merkle paths are already in the advice data (because you are taking the inputs from an `ExecutedTransaction`). This does not seem to be the case for the prefetched assets call to the `DataStore`, as it's called every time even if the data is already in the advice inputs. Additionally, only the prefetched asset witnesses are part of `TransactionInputs` separately. This means the validator implementor needs to know this `DataStore` call will only be called necessarily once in the whole transaction for the prefetched assets (and now that I see the comment you linked it makes sense why this was the case), and this is what I'd document more. 
Alternatively, it would be great if the asset witnesses prefetch is skipped on re-execution as well. Then if you get a call to `DataStore::get_vault_asset_witnesses()` during re-execution it's basically an error every time and there's no ambiguity; also the implementation itself becomes trivial and can maybe be provided from `miden-base`. But maybe I'm missing something else here.

My assumptions (some of which might be incorrect) are:

- On dynamic calls to the `DataStore` (i.e., not the prefetch of `prepare_tx_inputs()`), the host first checks that the Merkle path is not already on the advice data.
  - If it is, it's loaded from there and the call is not done
- When creating inputs from an `ExecutedTransaction` for the re-execution case, you take the latest state of the advice data (this includes added witnesses)
- Prefetched asset witnesses are retrieved every time regardless of them being in the advice inputs, and later get loaded again into the Merkle store

SantiagoPittella: Should we merge this PR?

igamigo: I think these changes are still valid, so I’m okay with merging this. Though I still have the concerns I mentioned above.

crates/ntx-builder/src/block_producer.rs
```diff
@@ -45,10 +45,11 @@ impl BlockProducerClient {
     pub async fn submit_proven_transaction(
         &self,
         proven_tx: ProvenTransaction,
+        tx_inputs: &TransactionInputs,
     ) -> Result<(), Status> {
         let request = proto::transaction::ProvenTransaction {
             transaction: proven_tx.to_bytes(),
-            transaction_inputs: None,
+            transaction_inputs: Some(tx_inputs.to_bytes()),
         };
```
bobbinth, line 53: I don't think this is needed: block producer ignores this field. Here is how it is supposed to work:

- This `transaction_inputs` field is the optional field that we currently use for guardrails.
- It is used in the RPC to prepare the request and send the request to the validator.
- Once the transaction is OKed by the validator, the RPC forwards the transaction to the block producer, but the `transaction_inputs` are no longer needed.

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -281,7 +281,7 @@ impl AccountActor {
             self.script_cache.clone(),
         );
 
-        let execution_result = context.execute_transaction(tx_candidate).await;
+        let execution_result = Box::pin(context.execute_transaction(tx_candidate)).await;
```
bobbinth, line 284: Why do we need `Box::pin()` here? And same question for line 250 above.

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -74,20 +74,24 @@ impl DataStore for TransactionInputsDataStore {
                 });
             }
 
-            Result::<Vec<_>, _>::from_iter(vault_keys.into_iter().map(|vault_key| {
-                match self.tx_inputs.account().vault().open(vault_key) {
-                    Ok(vault_proof) => {
-                        AssetWitness::new(vault_proof.into()).map_err(|err| DataStoreError::Other {
-                            error_msg: "failed to open vault asset tree".into(),
-                            source: Some(err.into()),
+            let stored_witnesses = self.tx_inputs.asset_witnesses();
```
bobbinth, line 77: I'm not sure this will work. As discussed in another comment, this will only cover the pre-fetched witnesses. As @igamigo mentioned, we should be able to get rid of this field entirely but the change would need to happen in `miden-base`. @igamigo, could you create an issue for this?

For broader context, this should be properly addressed with https://github.com/0xMiden/miden-node/pull/1493.

bobbinth (commented):
Looks good! Thank you. I left a few comments inline. Basically, I think the monitor-related changes are fine, but ntx builder and validator changes may need to be rolled back.

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: It was related to the changes that I did on `Context::execute_transaction`, but since we are rolling back that, the box pin aren't needed anymore.

crates/validator/src/tx_validation/data_store.rs
SantiagoPittella: I'll close this PR then, and open another one only with the changes for the network monitor

SantiagoPittella: I opend https://github.com/0xMiden/miden-node/pull/1501 with the standing changes

igamigo: To be clear, I believe this does fix the issue (and I think @SantiagoPittella did some tests to confirm this). This covers the pre-fetched witnesses only, but AFAICT (might be missing something so take it with a grain of salt), this is enough because on re-execution that's the only time this will be called, as all other witnesses get loaded from advice data directly and do not go through the `DataStore` first. My comment was more pointing out the inconsistency and the fact that this did not seem to be very documented. 

But going to open an issue for potentially removing the field in `miden-base` so it can be discussed there.

bobbinth: The pre-fetched asset witnesses will cover witness of asset of all input notes. So, for transactions that only consume notes, this will be fine. But I don't think these witness will cover assets for output note. These are generated during transaction execution and so, are not added to this filed.

bobbinth:
@SantiagoPittella - now that #1501 has been merged, should we close this PR?

PR #1493, opened 2026-01-09
feat(validator): FPI support in data store

sergerad:
## Context

The Validator's `TransactionInputsDataStore` does not handle foreign accounts for the purposes of the following functions of `DataStore` trait:

NOTE: This will remain in draft until miden-base has the `TransactionIniputs::read_foreign_account_inputs()` functionality merged into next.

A followup PR will be made to add similar functionality to the `NtxDataStore`.

Relates to #1387 and #1305.

## Changes
* Update `TransactionInputsDataStore` to handle foreign accounts through `self.tx_inputs`.

### DIFF
--- a/crates/validator/src/tx_validation/data_store.rs
+++ b/crates/validator/src/tx_validation/data_store.rs
@@ -62,37 +62,17 @@ impl DataStore for TransactionInputsDataStore {
 
     fn get_vault_asset_witnesses(
         &self,
-        account_id: AccountId,
+        _account_id: AccountId,
         vault_root: Word,
         vault_keys: BTreeSet<AssetVaultKey>,
     ) -> impl FutureMaybeSend<Result<Vec<AssetWitness>, DataStoreError>> {
         async move {
-            if self.tx_inputs.account().id() == account_id {
-                // Get asset witnessess from native account.
-                Result::<Vec<_>, _>::from_iter(vault_keys.into_iter().map(|vault_key| {
-                    match self.tx_inputs.account().vault().open(vault_key) {
-                        Ok(vault_proof) => AssetWitness::new(vault_proof.into()).map_err(|err| {
-                            DataStoreError::other_with_source(
-                                "failed to open vault asset tree",
-                                err,
-                            )
-                        }),
-                        Err(err) => {
-                            Err(DataStoreError::other_with_source("failed to open vault", err))
-                        },
-                    }
-                }))
-            } else {
-                // Get asset witnessess from foreign account.
-                self.tx_inputs
-                    .read_vault_asset_witnesses(vault_root, vault_keys)
-                    .map_err(|err| {
-                        DataStoreError::other_with_source(
-                            "failed to read vault asset witnesses",
-                            err,
-                        )
-                    })
-            }
+            // Retrieve native and foreign account asset witnesses from the advice inputs.
+            self.tx_inputs
+                .read_vault_asset_witnesses(vault_root, vault_keys)
+                .map_err(|err| {
+                    DataStoreError::other_with_source("failed to read vault asset witnesses", err)
+                })
         }
     }
 


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - Added `GetLimits` endpoint to the RPC server ([#1410](https://github.com/0xMiden/miden-node/pull/1410)).
 - Added gRPC-Web probe support to the `miden-network-monitor` binary ([#1484](https://github.com/0xMiden/miden-node/pull/1484)).
 - Add DB schema change check ([#1268](https://github.com/0xMiden/miden-node/pull/1485)).
+- Add foreign account support to `TransactionInputsDataStore` ([#1493](https://github.com/0xMiden/miden-node/pull/1493)).
```
```suggestion
- Add foreign account support to validator ([#1493](https://github.com/0xMiden/miden-node/pull/1493)).
```

crates/validator/src/tx_validation/data_store.rs
Mirko-von-Leipzig, line 1: I realized recently that we should probably be using `std::futures::ready` instead of `async move` to create futures for sync code. Though the difference is really just when the bulk of the work gets done.

```rust
// This
return std::futures::ready(x);

// Instead of this
async move {
    // Only sync code
    return x;
}
```

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -53,60 +53,92 @@ impl DataStore for TransactionInputsDataStore {
...
+                    source: None,
+                });
+            }
+
+            let foreign_inputs = self
+                .tx_inputs
+                .read_foreign_account_inputs(foreign_account_id)
+                .map_err(|err| DataStoreError::Other {
+                    error_msg: format!("failed to read foreign account inputs: {err}").into(),
+                    source: Some(Box::new(err)),
```
Mirko-von-Leipzig: This doubles the `err` message if we're printing a report


```suggestion
                    error_msg: "failed to read foreign account inputs",
                    source: Some(Box::new(err)),
```

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -53,60 +53,92 @@ impl DataStore for TransactionInputsDataStore {
...
         _ref_block: BlockNumber,
     ) -> impl FutureMaybeSend<Result<AccountInputs, DataStoreError>> {
-        async move { Err(DataStoreError::AccountNotFound(foreign_account_id)) }
+        async move {
+            if foreign_account_id == self.tx_inputs.account().id() {
+                return Err(DataStoreError::Other {
+                    error_msg: format!(
+                        "requested account with id {foreign_account_id} is local, not foreign"
+                    )
+                    .into(),
```
Mirko-von-Leipzig:
```suggestion
                    error_msg: "requested account with id {foreign_account_id} is local, not foreign".into(),
```

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -53,60 +53,92 @@ impl DataStore for TransactionInputsDataStore {
         foreign_account_id: AccountId,
         _ref_block: BlockNumber,
     ) -> impl FutureMaybeSend<Result<AccountInputs, DataStoreError>> {
-        async move { Err(DataStoreError::AccountNotFound(foreign_account_id)) }
+        async move {
+            if foreign_account_id == self.tx_inputs.account().id() {
```
Mirko-von-Leipzig: Is this actually a problem? Genuine question - I know its a weird thing to do, but maybe someone creates a procedure which always reaches out to account `A` and uses this in account `A` as well?

@PhilippGackstatter is it legal to do an FPI call on yourself? Is this possible, or would the kernel/executioner pick this up and redirect it?

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -53,60 +53,92 @@ impl DataStore for TransactionInputsDataStore {
...
+            // Get asset witnessess from local or foreign account.
+            if self.tx_inputs.account().id() == account_id {
+                get_asset_witnesses_from_account(self.tx_inputs.account(), vault_keys)
+            } else {
+                let foreign_inputs = self
+                    .tx_inputs
+                    .read_foreign_account_inputs(account_id)
+                    .map_err(|err| DataStoreError::Other {
+                        error_msg: format!("failed to read foreign account inputs: {err}").into(),
+                        source: Some(Box::new(err)),
```
Mirko-von-Leipzig:
```suggestion
                        error_msg: "failed to read foreign account inputs.into(),
                        source: Some(Box::new(err)),
```

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -53,60 +53,92 @@ impl DataStore for TransactionInputsDataStore {
...
+                })?;
+            Ok(foreign_inputs)
+        }
     }
 
     fn get_vault_asset_witnesses(
         &self,
         account_id: AccountId,
-        vault_root: Word,
+        _vault_root: Word,
```
Mirko-von-Leipzig: iiuc this no longer matters because we're assuming that transaction inputs was automagically loaded with the correct witness values at this root as a result of the execution?

If so (and I don't fully understand this) but it feels like transaction inputs would need to be validated before this can be done safely.

sergerad: Fixed

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -53,7 +53,25 @@ impl DataStore for TransactionInputsDataStore {
...
+                    error_msg:
+                        "requested account with id {foreign_account_id} is local, not foreign"
+                            .into(),
+                    source: None,
+                });
+            }
+
+            let foreign_inputs = self
+                .tx_inputs
+                .read_foreign_account_inputs(foreign_account_id)
```
PhilippGackstatter: I haven't followed the design of `TransactionInputsDataStore` so I could be missing something.

From what I can tell, there are two use cases involved in this PR and the linked issue:
- Re-execution of transactions submitted to the node by using `TransactionInputs`.
- Network transactions, in particular also supporting FPI calls.

It seems to me these two need different `DataStore` implementations, and from what I can tell this is already the case: `TransactionInputsDataStore` and `NtxDataStore`.

For re-execution, the `TransactionInputs` in this case should already contain all data necessary for execution, for example, the foreign account data in the advice provider needed for an FPI call. So, the only method that really needs to have a meaningful implementation should be `get_transaction_inputs`. This is called by `TransactionExecutor::prepare_tx_inputs` and must succeed. Any of the other `DataStore` methods are not called unless data needs to be requested (except for `get_vault_asset_witnesses`, also called in `prepare_tx_inputs`, but this could just return `Vec::new()` in this case). None of them should actually be called and so they could just return errors in the implementation here, because the data should already be in `TransactionArgs::advice_inputs` and no lazy loading should be required, from what I can tell.

For the FPI calls, the `TransactionInputs` shouldn't be used at all, afaict, so it seems this would be done with `NtxDataStore` and I assume this `DataStore` impl is backed by the actual node databases.

So, in both cases we shouldn't need to extract data from advice inputs (i.e. what https://github.com/0xMiden/miden-base/pull/2246 does). I think the overall flow of data is pretty hard to follow now, and if there's a different way, I would definitely entertain that. But, I'm likely missing something here. Where did I miss something?

crates/validator/src/tx_validation/data_store.rs
PhilippGackstatter: The [tx kernel prevents](https://github.com/0xMiden/miden-base/blob/b82dc0b0f8f220f9c4839309ce11ef13582d2cf7/crates/miden-protocol/asm/kernels/transaction/lib/memory.masm#L931-L934) the native account from FPI-calling itself.

crates/validator/src/tx_validation/data_store.rs
PhilippGackstatter: I checked execution and re-execution in `test_get_initial_item_and_get_initial_map_item_with_foreign_account` in miden-base and during execution we get the following `DataStore` calls:
- get_transaction_inputs
- get_vault_asset_witnesses
- get_foreign_account_inputs
- get_storage_map_witness

During re-execution we get:
- get_transaction_inputs
- get_vault_asset_witnesses
- get_foreign_account_inputs

Notes:
- get_storage_map_witness isn't called during re-execution because we check if the merkle path is in the advice inputs and if so, don't request it. I haven't verified with a test, but the same should apply for asset witnesses as they are handled almost identically.
- The get_vault_asset_witnesses call is for pre-fetching asset witnesses and should go away with https://github.com/0xMiden/miden-base/pull/2274, which skips the call if everything is already present (which it should be on re-execution).
- I need to check for `get_note_script`, but we should be able to skip this as well, following the same logic.
- For get_foreign_account_inputs, the host assumes the event is only emitted when the foreign account wasn't loaded yet, and so there is no redundancy check. When I implemented this, I did not think of re-execution, but we should be able to check for presence of the foreign account inputs in the advice inputs and skip the check as well. Will create an issue for this tomorrow my time.

So, by the end of this, only `get_transaction_inputs` should be left to require an actual implementation. Now, it's pretty weird to implement the `DataStore` methods by returning errors, but this interface really wasn't designed for re-execution, so that's where that mismatch comes from. For this release, I would implement it by returning errors.

For the next release, we can think of having a `TransactionReexecutor` type in miden-base that takes care of all those details for re-execution. This can't sensibly be `TransactionExecutor::reexecute_transaction` because the type itself requires a `DataStore` and `TransactionAuthenticator` impl, and these are the details that can be abstracted away.

It should be possible for this API to just be:

```rust
TransactionReexecutor::reexecute(transaction_inputs)
```

This should make usage in the node very simple.

igamigo: Probably makes sense to go a step further at that point and add a wrapper of `TransactionInputs` that are explicitly for re-execution (e.g, `pub struct ValidationTransactionInputs(TransactionInputs)` with a better name) in order to make them distinct from the normal ones, which does not have much in terms of expressing that the inner data contains everything you need for re-execution.

crates/validator/src/tx_validation/data_store.rs
sergerad: I'll take that to mean that its a good idea to error out here unless any objections

crates/validator/src/tx_validation/data_store.rs
bobbinth: @PhilippGackstatter and @igamigo - thank you for bringing this up! I totally blanked that data store methods may not even be invoked if the data is already in the advice provider.

My thinking about this PR:

- We keep the new implementation of `get_foreign_account_inputs()`.
- We replace `get_storage_map_witness()` with `unimplemented!()` because it should never be called in the re-execution context.
- We keep the new implementation of `get_vault_asset_witnesses()`. But once https://github.com/0xMiden/miden-base/pull/2274 lands, we can also replace it with `unimplemented!()`.
- We still need to double check whether note scripts are requested from the data store during re-execution, and if not, we can put `unimplemented!()` into `get_note_script()` as well. If scripts are quested during re-execution, we'll need to parse them out of the transaction inputs, or better, fix this in `miden-base` to make sure they are not requested.

In the longer term, we can implement a dedicated flow for re-executing transactions which would not require a data store at all. For this, we should create an issue in `miden-base`.

PhilippGackstatter: I agree! I checked for foreign accounts, but those are indeed more complicated. These don't just require the data store call to be skipped, but also require their code to be loaded into the `MastForestStore` during re-execution and the `AccountProcedureIndexMap` to be built with their code. This is something the `TransactionExecutor` currently doesn't deal with because it doesn't consider re-execution. I opened an issue in base that covers this roughly.

I checked `test_public_note_creation_with_script_from_datastore` and added a re-execution at the end and `get_note_script` is not called during re-execution. We already [add the script to the advice map](https://github.com/0xMiden/miden-base/blob/1781efa18ff401157932af4b3ed7f6898813e0de/crates/miden-tx/src/executor/exec_host.rs#L397-L400) when it is missing, so on re-execution it is present. So, we should be able to use `unimplemented!` here, but would be good to test this, like all the other cases, too.

Opened https://github.com/0xMiden/miden-base/issues/2293 for a proper re-execution API in base.

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -53,7 +53,25 @@ impl DataStore for TransactionInputsDataStore {
         foreign_account_id: AccountId,
         _ref_block: BlockNumber,
     ) -> impl FutureMaybeSend<Result<AccountInputs, DataStoreError>> {
-        async move { Err(DataStoreError::AccountNotFound(foreign_account_id)) }
+        async move {
+            if foreign_account_id == self.tx_inputs.account().id() {
+                return Err(DataStoreError::Other {
+                    error_msg:
+                        "requested account with id {foreign_account_id} is local, not foreign"
```
PhilippGackstatter:
```suggestion
                        "requested account with id {foreign_account_id} is native, not foreign"
```

Nit: we usually call this the native account, though I think we don't need this branch at all. As mentioned in the other comment, the kernel already prevents this.

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -53,7 +53,25 @@ impl DataStore for TransactionInputsDataStore {
...
+                    source: None,
+                });
+            }
+
+            let foreign_inputs = self
+                .tx_inputs
+                .read_foreign_account_inputs(foreign_account_id)
+                .map_err(|err| DataStoreError::Other {
+                    error_msg: "failed to read foreign account inputs".into(),
+                    source: Some(Box::new(err)),
```
PhilippGackstatter: FYI: It should be possible to simplify this slightly with `DataStore::other_with_source`.

crates/validator/src/tx_validation/data_store.rs
sergerad: I have replaced `get_storage_map_witness()` impl with `unimplemented!()`. 

>So, we should be able to use unimplemented! here, but would be good to test this, like all the other cases, too.

Are there some tests I should add that involve re-execution for this?

bobbinth: > I have replaced `get_storage_map_witness()` impl with `unimplemented!()`.

Looking at the most current commit, it seems like `get_vault_asset_witnesses()` rather than `get_storage_map_witness()` are `unimplemented!()`. We swap these. Basically, this PR should:

- Set body of `get_storage_map_witness()` to `unimplemented!()`.
- Set the body of `get_note_script()` to `unimplemented()`.
- Use `TransactionInputs::read_vault_asset_witnesses()` in `get_vault_asset_witnesses()`.
- Use `TransactionInputs::read_foreign_account_inputs()` in `get_foreign_account_inputs()` (as it does now).

> Are there some tests I should add that involve re-execution for this?

We could add a test that executes an FPI call and then make sure we don't get panics on re-execution - but I'm not sure if this adds a lot of incremental value compared to the tests we have in `miden-base`.

sergerad: Fixed, thanks

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -63,58 +67,54 @@ impl DataStore for TransactionInputsDataStore {
...
+                                err,
+                            )
+                        }),
+                        Err(err) => {
+                            Err(DataStoreError::other_with_source("failed to open vault", err))
+                        },
+                    }
+                }))
+            } else {
+                let foreign_inputs = self
```
bobbinth: nit: I'd probably rename `foreign_inputs` into just `asset_witnesses` or `foreign_asset_witnesses`.

bobbinth (approved):
Looks good! Thank you! I left just one small nit inline.

Also, would be good to get a review from @igamigo. Maybe we can run client integration tests against this branch to confirm that transactions that do FPI calls can indeed be accepted by the node.

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -63,58 +67,53 @@ impl DataStore for TransactionInputsDataStore {
...
-
-            if self.tx_inputs.account().vault().root() != vault_root {
-                return Err(DataStoreError::Other {
-                    error_msg: "vault root mismatch".into(),
-                    source: None,
-                });
+            if self.tx_inputs.account().id() == account_id {
+                // Get asset witnessess from native account.
+                Result::<Vec<_>, _>::from_iter(vault_keys.into_iter().map(|vault_key| {
+                    match self.tx_inputs.account().vault().open(vault_key) {
```
PhilippGackstatter: The native account's witnesses are handled just like the ones for foreign accounts, i.e. they are [added to the advice inputs](https://github.com/0xMiden/miden-base/blob/9411b6a140e394ff044c5591f6a5d8ac2f6a5118/crates/miden-tx/src/executor/exec_host.rs#L364) on lazy loading, but not to the `PartialVault`. Any asset witnesses that are present in the partial vault at the beginning are also [added to the advice inputs](https://github.com/0xMiden/miden-base/blob/9411b6a140e394ff044c5591f6a5d8ac2f6a5118/crates/miden-protocol/src/transaction/kernel/advice_inputs.rs#L292-L296).

So, the advice inputs contain both the pre-provided asset witnesses and those fetched during lazy loading, and so we should read from the advice inputs and not from the partial vault.

The need for this should go away with https://github.com/0xMiden/miden-base/pull/2298 completely, at which point this could be replaced with `unimplemented!`.

Because of these subtle differences, I still think it would be good to add a test for this, e.g. one that has an FPI call, reads a storage map item lazily and has input notes with some assets, so the pre-fetch is executed.

And then all of these annoying details should be abstracted away once we have a re-execution API exposed in base directly: https://github.com/0xMiden/miden-base/issues/2293.

sergerad: https://github.com/0xMiden/miden-base/pull/2298 has been merged so I have moved this to unimplemented

PhilippGackstatter (approved):
Looks good to me!

crates/validator/src/tx_validation/data_store.rs
bobbinth: https://github.com/0xMiden/miden-base/pull/2298 is not a part of v0.13.0 release - it went into `next`. So, we can't make use of it in this PR.

sergerad: Updated

PR #1495, opened 2026-01-09
chore: load network accounts asynchronously in NTX Builder

### LINKED ISSUE #1487: Decouple ntx builder from block-producer startup
## Summary

Decouple the ntx builder by moving account loading from start-up to a separate concurrent task.

## History

The network tx builder currently has to start-up in lockstep with the block-producer. This is enforced by having the block-producer start the ntx builder and synchronicity is done via a `Barrier` such that both components can proceed together.

This was required initially to ensure the ntx builder and mempool have the same latest chain-tip state:

1. ntx builder loads account states from the store until it reaches the chain tip
2. block-producer starts
3. ntx builder connects to mempool subscription

We have since refactored the ntx builder to no longer pre-load all account state from the store. Instead each account's actor loads its state separately as it starts up. This has its own issues, however it now means we can decouple the startup of these components.

## What to do

Spawn a separate task in the ntx builder which loads the network accounts from the store, page-by-page. Each account is then communicated back to the main event loop via a channel. There is no need to load all accounts - accounts can be submitted as they arrive. Something to keep in mind is that if you end this task the channel receiver will close which may cause issues depending on the implementation. It may be best to simply `pending()` the task forever at this stage.

The event loop will now need a `select!` to differentiate between the above channel, or the mempool subscription. The idea with the new channel is that this notifies the event loop that this account exists and this should follow the same logic as then current event handle. This logic isn't perfect, but we'll tackle that in other work.

Subscribing to the mempool currently requires the block-height, but this requirement can be removed so long as the above task is spawned _after_ the subscription is created.

And of course remove the barrier.

## Follow-ups

#1478 can be considered as well after this. The other thing to note is that this implementation doesn't care about being lined up with block edges; only that we at some point get all committed network accounts. So we could paginate by a constant via rowid (so long as we don't delete accounts from the db) which can simplify the store impl and api.

We can also actually make the ntx builder a proper separate component after this is done. i.e. with its own CLI and so on.

SantiagoPittella:
closes #1487

depends on https://github.com/0xMiden/miden-node/pull/1453 and https://github.com/0xMiden/miden-node/pull/1501 to fully work.

### DIFF
--- a/crates/ntx-builder/src/store.rs
+++ b/crates/ntx-builder/src/store.rs
@@ -202,17 +202,12 @@ impl StoreClient {
         block_range: RangeInclusive<BlockNumber>,
         sender: &tokio::sync::mpsc::Sender<NetworkAccountId>,
     ) -> Result<Option<BlockNumber>, StoreError> {
-        let (accounts, pagination_info) = self
-            .fetch_network_account_ids_page(block_range)
-            .await
-            .inspect_err(|err| tracing::Span::current().set_error(err))?;
+        let (accounts, pagination_info) = self.fetch_network_account_ids_page(block_range).await?;
 
         let chain_tip = pagination_info.chain_tip;
         let current_height = pagination_info.block_num;
 
-        self.send_accounts_to_channel(accounts, sender)
-            .await
-            .inspect_err(|err| tracing::Span::current().set_error(err))?;
+        self.send_accounts_to_channel(accounts, sender).await?;
 
         if current_height >= chain_tip {
             Ok(None)
@@ -223,6 +218,15 @@ impl StoreClient {
 
     #[instrument(target = COMPONENT, name = "store.client.fetch_network_account_ids_page", skip_all, err)]
     async fn fetch_network_account_ids_page(
+        &self,
+        block_range: std::ops::RangeInclusive<BlockNumber>,
+    ) -> Result<(Vec<NetworkAccountId>, proto::rpc::PaginationInfo), StoreError> {
+        self.fetch_network_account_ids_page_inner(block_range)
+            .await
+            .inspect_err(|err| tracing::Span::current().set_error(err))
+    }
+
+    async fn fetch_network_account_ids_page_inner(
         &self,
         block_range: std::ops::RangeInclusive<BlockNumber>,
     ) -> Result<(Vec<NetworkAccountId>, proto::rpc::PaginationInfo), StoreError> {


crates/ntx-builder/src/store.rs
```diff
@@ -173,21 +173,24 @@ impl StoreClient {
...
-    /// foreseeable future.
-    #[instrument(target = COMPONENT, name = "store.client.get_network_account_ids", skip_all, err)]
-    pub async fn get_network_account_ids(&self) -> Result<Vec<AccountId>, StoreError> {
-        const MAX_ITERATIONS: u32 = 100_000;
+    /// accounts (~289,000). With 1000 iterations, this supports up to ~524 million network
+    /// accounts, which is assumed to be sufficient for the foreseeable future.
+    #[instrument(target = COMPONENT, name = "store.client.stream_network_account_ids", skip_all, err)]
+    pub async fn stream_network_account_ids(
+        &self,
+        sender: tokio::sync::mpsc::Sender<Vec<AccountId>>,
```
Mirko-von-Leipzig: I think let's simplify by just sending a single item at a time?


```suggestion
        sender: tokio::sync::mpsc::Sender<AccountId>,
```

crates/ntx-builder/src/builder.rs
```diff
@@ -91,13 +83,16 @@ pub struct NetworkTransactionBuilder {
 }
 
 impl NetworkTransactionBuilder {
+    /// Channel capacity for account loading. Each item in the channel is a batch of account IDs
+    /// loaded from a single page. A small capacity is sufficient since each page can contain
+    /// up to ~289,000 accounts.
+    const ACCOUNT_CHANNEL_CAPACITY: usize = 4;
```
Mirko-von-Leipzig: As mentioned elsewhere, I think lets keep the pagination internal to the gRPC client implementation. This way nothing changes even if we switch pagination to streaming or with different capacity.

In which case we can make this quite a bit larger.

crates/ntx-builder/src/builder.rs
```diff
@@ -176,8 +161,35 @@ impl NetworkTransactionBuilder {
...
+        skip(self, account_ids, actor_context),
+        fields(count = account_ids.len())
+    )]
+    async fn handle_loaded_accounts(
+        &mut self,
+        account_ids: Vec<AccountId>,
+        actor_context: &AccountActorContext,
+    ) -> Result<(), anyhow::Error> {
+        for account_id in account_ids {
+            if let Ok(account_prefix) = NetworkAccountPrefix::try_from(account_id) {
```
Mirko-von-Leipzig: I think we should move this to inside the gRPC client handling

crates/ntx-builder/src/store.rs
```diff
@@ -173,21 +173,24 @@ impl StoreClient {
...
     ///
     /// Each page can return up to `MAX_RESPONSE_PAYLOAD_BYTES / AccountId::SERIALIZED_SIZE`
-    /// accounts (~289,000). With `100_000` iterations, which is assumed to be sufficient for the
-    /// foreseeable future.
-    #[instrument(target = COMPONENT, name = "store.client.get_network_account_ids", skip_all, err)]
-    pub async fn get_network_account_ids(&self) -> Result<Vec<AccountId>, StoreError> {
-        const MAX_ITERATIONS: u32 = 100_000;
+    /// accounts (~289,000). With 1000 iterations, this supports up to ~524 million network
+    /// accounts, which is assumed to be sufficient for the foreseeable future.
+    #[instrument(target = COMPONENT, name = "store.client.stream_network_account_ids", skip_all, err)]
```
Mirko-von-Leipzig: I think we shouldn't instrument at this level considering this might run for a very time.

Instead we should instrument each page. I'm thinking ideally we should have something like:

```
load_committed_accounts
   - fetch_page
   - submit_page
```

and attach the chain tip and current height as fields.

We can do this inline if you don't want to create additional functions -- up to you.

crates/ntx-builder/src/store.rs
Mirko-von-Leipzig, line 1: Something we will also need to address, now that this is decoupled - what do we do for transport errors? Some options:

1. hard error (current), this might still be fine, but this _should_ be a fatal error so that the component resets (not currently implemented)
2. retry forever (loop)
    - this is fine probably, so long as we log each one so we can figure it out
3. retry `N` times and then abort
    - this is probably the more correct way, but eh 🤷 

I tend to (1) or (2). wdyt?

crates/ntx-builder/src/store.rs
```diff
@@ -173,21 +173,24 @@ impl StoreClient {
...
+    pub async fn stream_network_account_ids(
+        &self,
+        sender: tokio::sync::mpsc::Sender<Vec<AccountId>>,
+    ) -> Result<(), StoreError> {
+        const MAX_ITERATIONS: u32 = 1000;
```
Mirko-von-Leipzig: I would remove this. So long as the store is sending us more stuff I don't think we care. And since its an internal connection we don't need to be defensive.

The initial reason for it was to limit super long start up times (since the ntx builder first needs to load).

crates/ntx-builder/src/store.rs
Mirko-von-Leipzig: Actually (2) is probably the best idea for now. Since the `store` can take a _very_ long time to start up, we should spin and wait for it.

SantiagoPittella: I add an exponential backoff limited to 30s max

crates/ntx-builder/src/builder.rs
```diff
@@ -91,13 +82,14 @@ pub struct NetworkTransactionBuilder {
 }
 
 impl NetworkTransactionBuilder {
+    /// Channel capacity for account loading.
+    const ACCOUNT_CHANNEL_CAPACITY: usize = 100_000;
```
Mirko-von-Leipzig: This can be much smaller, since we are in no rush to complete the sync. We can go at whatever pace the main task goes.

```suggestion
    /// Channel capacity for account loading.
    const ACCOUNT_CHANNEL_CAPACITY: usize = 1_000;
```

crates/ntx-builder/src/builder.rs
```diff
@@ -147,15 +128,16 @@ impl NetworkTransactionBuilder {
...
-                    .spawn_actor(AccountOrigin::store(account_prefix), &actor_context)
-                    .await?;
+        // Spawn a background task to load network accounts from the store.
+        // Accounts are sent through a channel in batches and processed in the main event loop.
+        let (account_tx, mut account_rx) =
+            mpsc::channel::<NetworkAccountPrefix>(Self::ACCOUNT_CHANNEL_CAPACITY);
+        let account_loader_store = store.clone();
+        tokio::spawn(async move {
+            if let Err(err) = account_loader_store.stream_network_account_ids(account_tx).await {
+                tracing::error!(%err, "failed to load network accounts from store");
```
Mirko-von-Leipzig: Another question is whether we should be reacting to task failure here. As in, should we abort the ntx builder (or restart the task) if it fails? Or is logging a single error good enough.

Failure here would mean that we never load accounts with existing notes from the store unless they also get a new event coming in. So I think we _should_ abort.

crates/ntx-builder/src/store.rs
```diff
@@ -173,62 +173,130 @@ impl StoreClient {
...
+                .inner
+                .clone()
+                .get_network_account_ids(Into::<BlockRange>::into(block_range.clone()))
+                .await
+            {
+                Ok(response) => break response.into_inner(),
+                Err(err) => {
+                    // Exponential backoff with base 500ms and max 30s.
+                    let backoff = Duration::from_millis(500)
+                        .saturating_mul(1 << retry_counter)
```
Mirko-von-Leipzig: This will fail/rollover after.. 32 or 64 retries 😁 Also need to saturating exponential (or limit counter I guess).

crates/ntx-builder/src/store.rs
```diff
@@ -173,62 +173,130 @@ impl StoreClient {
...
-    /// accounts (~289,000). With `100_000` iterations, which is assumed to be sufficient for the
-    /// foreseeable future.
-    #[instrument(target = COMPONENT, name = "store.client.get_network_account_ids", skip_all, err)]
-    pub async fn get_network_account_ids(&self) -> Result<Vec<AccountId>, StoreError> {
-        const MAX_ITERATIONS: u32 = 100_000;
-
+    /// This method is designed to be run in a background task, sending accounts to the main event
+    /// loop as they are loaded. This allows the ntx-builder to start processing mempool events
+    /// without waiting for all accounts to be preloaded.
+    #[instrument(target = COMPONENT, name = "store.client.load_committed_accounts", skip_all, err)]
```
Mirko-von-Leipzig: There's a bit of a problem with the instrumentation as is. One needs to be careful with long-running tasks and traces. This is because a trace is only complete once the _root_ span completes.

In this case, what we'll have is:

```
start load_committed_accounts
  fetch_page(0)
  submit_page(0)
  fetch_page(1)
  submit_page(1)
  ...
  ...
  ... a long, long, long time later
  fetch_page(chain tip)
  submit_page(chain tip)
close load_committed_accounts
```

Instead we shouldn't instrument this method at all, and each iteration of the loop should be its own root span. This means you'll want to reshuffle things a bit.

See [here](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-retries-examples) for an example with retries. The first example is what we have per each loop iteration. Let me know if this is still ambiguous.

crates/ntx-builder/src/store.rs
SantiagoPittella: I fixed this here, and in some other places where we were using this pattern

crates/ntx-builder/src/store.rs
SantiagoPittella: Ok, perfect. I will check the network-monitor for this too and open an issue if I find something.

crates/ntx-builder/src/builder.rs
SantiagoPittella: I went for the abort approach

crates/ntx-builder/src/store.rs
```diff
@@ -173,62 +174,140 @@ impl StoreClient {
...
+        &self,
+        accounts: Vec<NetworkAccountPrefix>,
+        sender: &tokio::sync::mpsc::Sender<NetworkAccountPrefix>,
+        chain_tip: u32,
+        current_height: u32,
+    ) -> Result<(), StoreError> {
+        for account in accounts {
+            // If the receiver is dropped, stop loading.
+            if sender.send(account).await.is_err() {
+                return Ok(());
```
sergerad, line 304: Might be worth an error/warn log here

sergerad (approved):

crates/ntx-builder/src/block_producer.rs
```diff
@@ -65,9 +65,9 @@ impl BlockProducerClient {
         loop {
             match self.subscribe_to_mempool(chain_tip).await {
                 Err(err) if err.code() == tonic::Code::Unavailable => {
-                    // exponential backoff with base 500ms and max 30s
+                    // Exponential backoff with base 500ms and max 30s.
                     let backoff = Duration::from_millis(500)
-                        .saturating_mul(1 << retry_counter)
+                        .saturating_mul(1 << retry_counter.min(6))
                         .min(Duration::from_secs(30));
```
bobbinth, line 73: We use the same logic in a couple of places (and it took me a bit of thinking to understand what's happening here) - should we create a utility function for this?

Mirko-von-Leipzig: There are some nice crates that let one do this on any future, e.g. [backon](https://docs.rs/backon/latest/backon/). Would probably be preferred over rolling our own, there are always some weird cases that one wants to handle which a DIY impl doesn't anticipate.

Mirko-von-Leipzig: #1550

crates/ntx-builder/src/builder.rs
```diff
@@ -176,10 +160,44 @@ impl NetworkTransactionBuilder {
...
+                // If the task fails, we abort since the builder would be in a degraded state
+                // where existing notes against network accounts won't be processed.
+                Some(result) = async {
+                    match account_loader.as_mut() {
+                        Some(handle) => Some(handle.await),
+                        None => std::future::pending().await,
+                    }
+                } => {
+                    account_loader = None;
+                    result.context("account loader task panicked")??;
```
Mirko-von-Leipzig: We have `flatten` from an error extension trait
```suggestion
                    result.context("account loader task panicked").flatten()?;
```

crates/ntx-builder/src/builder.rs
```diff
@@ -176,10 +160,44 @@ impl NetworkTransactionBuilder {
...
+                },
+                // Handle account loader task completion/failure.
+                // If the task fails, we abort since the builder would be in a degraded state
+                // where existing notes against network accounts won't be processed.
+                Some(result) = async {
+                    match account_loader.as_mut() {
+                        Some(handle) => Some(handle.await),
+                        None => std::future::pending().await,
+                    }
+                } => {
```
Mirko-von-Leipzig: An alternative to `Option<Handle>` is to reassign the handle itself upon completion, something like

```rust
let mut account_loader_handle = tokio::spawn(...);

// ...

select! {
    result = account_loader_handle => {
        result
            .context("panicked while loading accounts from store")
            .flatten()?;

        tracing::info!("account loading from store completed");
        account_loader_handle = std::future::pending();
    }
}
```

crates/ntx-builder/src/store.rs
```diff
@@ -173,62 +174,140 @@ impl StoreClient {
...
+                entity: "NetworkAccountIdList",
+                field_name: "pagination_info",
+            },
+        )?;
 
-            if iterations_count >= MAX_ITERATIONS {
-                return Err(StoreError::MaxIterationsReached("GetNetworkAccountIds".to_string()));
+        tracing::Span::current()
+            .record("chain_tip", pagination_info.chain_tip)
+            .record("current_height", pagination_info.block_num);
```
Mirko-von-Leipzig: This is the info for the _next_ page or no? I would ignore this here since it will get recorded with its page on the next loop.

crates/ntx-builder/src/store.rs
```diff
@@ -173,62 +174,140 @@ impl StoreClient {
...
+
+        Ok((accounts, pagination_info))
+    }
+
+    #[instrument(
+        target = COMPONENT,
+        name = "store.client.submit_page",
+        skip(self, accounts, sender),
+        fields(chain_tip = chain_tip, current_height = current_height)
```
Mirko-von-Leipzig: I think its okay to log this at the root span only

```suggestion
        skip_all
```

crates/ntx-builder/src/store.rs
```diff
@@ -173,62 +174,140 @@ impl StoreClient {
...
-    pub async fn get_network_account_ids(&self) -> Result<Vec<AccountId>, StoreError> {
-        const MAX_ITERATIONS: u32 = 100_000;
+    /// Returns the next block number to fetch from, or `None` if the chain tip has been reached.
+    #[instrument(target = COMPONENT, name = "store.client.load_accounts_page", skip_all, err)]
+    async fn load_page(
+        &self,
+        block_range: RangeInclusive<BlockNumber>,
+        sender: &tokio::sync::mpsc::Sender<NetworkAccountPrefix>,
+    ) -> Result<Option<BlockNumber>, StoreError> {
+        let (accounts, pagination_info) = self.fetch_page(block_range).await?;
```
Mirko-von-Leipzig: Unfortunately `tracing` will only log the error's `Display` impl, while we usually want the full error report.

So instead we have to do this manually :*(

And the same for the error lower down.

```suggestion
        let (accounts, pagination_info) = self.fetch_page(block_range)
            .await
            .inspect_err(|err| tracing::Span::curret().set_error(err))?;
```

A somewhat better way (to prevent missing this) is to construct an async chain so that its all handled in one location, an example https://github.com/0xMiden/miden-node/blob/2cc9742dcdedb2a6f5a5cf572be42f1836f77542/crates/block-producer/src/block_builder/mod.rs#L136-L155

Mirko-von-Leipzig: Though this may not be a good fit depending on how data is broken up.

crates/ntx-builder/src/block_producer.rs
```diff
@@ -65,9 +65,9 @@ impl BlockProducerClient {
         loop {
             match self.subscribe_to_mempool(chain_tip).await {
                 Err(err) if err.code() == tonic::Code::Unavailable => {
-                    // exponential backoff with base 500ms and max 30s
+                    // Exponential backoff with base 500ms and max 30s.
                     let backoff = Duration::from_millis(500)
-                        .saturating_mul(1 << retry_counter)
+                        .saturating_mul(1 << retry_counter.min(6))
```
drahnr, line 72: Iiuc the `.min(6)` is just overflow protection

crates/ntx-builder/src/store.rs
```diff
@@ -173,62 +174,141 @@ impl StoreClient {
...
-    /// Since the `GetNetworkAccountIds` method is paginated, we loop through all pages until we
-    /// reach the end.
+    /// This method is designed to be run in a background task, sending accounts to the main event
+    /// loop as they are loaded. This allows the ntx-builder to start processing mempool events
+    /// without waiting for all accounts to be preloaded.
+    pub async fn stream_network_account_ids(
+        &self,
+        sender: tokio::sync::mpsc::Sender<NetworkAccountPrefix>,
+    ) -> Result<(), StoreError> {
+        let mut block_range = BlockNumber::from(0)..=BlockNumber::from(u32::MAX);
```
drahnr, line 187: nit: do we want to introduce a `BlockNumber::MAX`?

crates/ntx-builder/src/store.rs
```diff
@@ -173,62 +174,140 @@ impl StoreClient {
...
+    }
+
+    #[instrument(
+        target = COMPONENT,
+        name = "store.client.fetch_page",
+        skip_all,
+        fields(chain_tip, current_height),
+        err
+    )]
+    async fn fetch_page(
```
drahnr: nit: names are bit brief

crates/ntx-builder/src/store.rs
drahnr: I am in favor of factoring things to become computation chains, the error handling and flow becomes a bit opaque.
Could we not instrument the called functions appropriately or move a piece of it to an `fn .._inner`?

crates/ntx-builder/src/builder.rs
```diff
@@ -176,10 +159,43 @@ impl NetworkTransactionBuilder {
...
+    #[tracing::instrument(
+        name = "ntx.builder.handle_loaded_accounts",
+        skip(self, account_prefix, actor_context)
+    )]
+    async fn handle_loaded_account(
+        &mut self,
+        account_prefix: NetworkAccountPrefix,
+        actor_context: &AccountActorContext,
+    ) -> Result<(), anyhow::Error> {
+        self.coordinator
```
drahnr, line 198: Out of scope: Did we take a look on how this scales with _lots_ of accounts?

crates/ntx-builder/src/store.rs
```diff
@@ -173,62 +174,140 @@ impl StoreClient {
...
+    /// loop as they are loaded. This allows the ntx-builder to start processing mempool events
+    /// without waiting for all accounts to be preloaded.
+    pub async fn stream_network_account_ids(
+        &self,
+        sender: tokio::sync::mpsc::Sender<NetworkAccountPrefix>,
+    ) -> Result<(), StoreError> {
+        let mut block_range = BlockNumber::from(0)..=BlockNumber::from(u32::MAX);
+
+        while let Some(next_start) = self.load_page(block_range, &sender).await? {
+            block_range = next_start..=BlockNumber::from(u32::MAX);
```
drahnr, line 190: don't we want to load for the next `MAX_CHANNEL_CAPACITY` size range?

crates/ntx-builder/src/store.rs
SantiagoPittella: I went for the first `.await.inspect_err` approach

crates/ntx-builder/src/block_producer.rs
SantiagoPittella: yes it is

crates/ntx-builder/src/store.rs
SantiagoPittella: Maybe @PhilippGackstatter would be a better fit to answer this

crates/ntx-builder/src/store.rs
SantiagoPittella: I went for:
  - fetch_page → fetch_network_account_ids_page
  - submit_page → send_accounts_to_channel

crates/ntx-builder/src/builder.rs
SantiagoPittella: I did not. This is something that we are talking and, in my understanding, will be tackled by next release. Correct me if I'm wrong @Mirko-von-Leipzig

crates/ntx-builder/src/store.rs
SantiagoPittella: hmm i don;t think so. The channel capacity determines the amount of account, not the number of blocks

crates/ntx-builder/src/store.rs
PhilippGackstatter: I don't have a strong preference on either approach. Feel free to open a PR in base if you'd like to add it.

crates/ntx-builder/src/builder.rs
Mirko-von-Leipzig: Scaling isn't super important here (yet). I think testnet only has several thousand accounts and notes at the moment.

That said, this shouldn't be terrible. Each idle account:

- consumes some amount of memory while idle
- has some initial database IO to load account etc

of these, the last point is probably the most egregious. But I do think we have the ability to provide back pressure etc so I'm not concerned at the design level just yet.

SantiagoPittella: Good good, I'm marking this comment as resolved then. thx

crates/ntx-builder/src/store.rs
SantiagoPittella: Might be a good addition. Will open a PR 👌🏼

SantiagoPittella: though I would not change it in this PR, since we probably don't want to make a new release of base for this haha

crates/ntx-builder/src/store.rs
SantiagoPittella: Sorry @drahnr I didn't saw your comment since I didn't reloaded the page. I'm implementing the ..._inner

PR #1496, opened 2026-01-09
fix/db: improve DB query complexity

drahnr:
Previously introduced DB queries around vault and storage maps for accounts are too many / slow.

Relevant comments:
* vault query https://github.com/0xMiden/miden-node/pull/1481#discussion_r2663968253
* ~https://github.com/0xMiden/miden-node/pull/1481#discussion_r2663961444~ `select_account_storage_at_block` has been degraded to a test helper

### DIFF
--- a/crates/store/src/db/models/queries/accounts/at_block.rs
+++ b/crates/store/src/db/models/queries/accounts/at_block.rs
@@ -141,7 +141,7 @@ pub(crate) fn select_account_vault_at_block(
     let block_num_sql = block_num.to_raw_sql();
 
     let entries: Vec<Option<Vec<u8>>> = diesel::sql_query(
-        r#"
+        r"
         SELECT a.asset FROM account_vault_assets a
         INNER JOIN (
             SELECT vault_key, MAX(block_num) as max_block
@@ -150,7 +150,7 @@ pub(crate) fn select_account_vault_at_block(
             GROUP BY vault_key
         ) latest ON a.vault_key = latest.vault_key AND a.block_num = latest.max_block
         WHERE a.account_id = ?
-        "#,
+        ",
     )
     .bind::<Binary, _>(&account_id_bytes)
     .bind::<BigInt, _>(block_num_sql)


crates/store/src/db/models/queries/accounts/at_block.rs
```diff
@@ -135,50 +136,36 @@ pub(crate) fn select_account_vault_at_block(
...
-        )
-        .first(conn)
-        .optional()?;
-        if let Some(Some(asset_bytes)) = result {
-            let asset = Asset::read_from_bytes(&asset_bytes)?;
+    let mut last_vault_key: Option<&[u8]> = None;
+
+    for (vault_key_bytes, asset_bytes) in &all_entries {
+        // Skip if we've already processed this vault_key
+        if last_vault_key == Some(vault_key_bytes.as_slice()) {
```
drahnr: works because of the ordering above using `block_num.desc()`

crates/store/src/db/models/queries/accounts/tests.rs
```diff
@@ -550,3 +550,181 @@ fn test_upsert_accounts_with_empty_storage() {
...
+    let mut conn = setup_test_db();
+    let (account, _) = create_test_account_with_storage();
+    let account_id = account.id();
+
+    // Faucet ID is needed for creating FungibleAssets
+    let faucet_id = AccountId::try_from(ACCOUNT_ID_PUBLIC_FUNGIBLE_FAUCET).unwrap();
+
+    let block_1 = BlockNumber::from_epoch(0);
+    let block_2 = BlockNumber::from_epoch(1);
+    let block_3 = BlockNumber::from_epoch(2);
```
Mirko-von-Leipzig, line 577: This seems a bit odd - is this just to ensure they're really far apart?

crates/store/src/db/models/queries/accounts/at_block.rs
```diff
@@ -135,50 +136,36 @@ pub(crate) fn select_account_vault_at_block(
...
-        .into_iter()
-        .filter_map(|(key, maybe_block)| maybe_block.map(|block| (key, block))),
-    );
-
-    if latest_blocks_per_vault_key.is_empty() {
-        return Ok(Vec::new());
-    }
+    // Query all vault entries for this account up to and including this block.
+    // Order by vault_key ascending, then block_num descending so the first entry
+    // for each vault_key is the latest one.
```
Mirko-von-Leipzig: I assume we truncate this table, or how are rows limited?

crates/store/src/db/models/queries/accounts/at_block.rs
```diff
@@ -125,6 +118,14 @@ pub(crate) fn select_account_header_at_block(
...
 
 /// Query vault assets at a specific block by finding the most recent update for each `vault_key`.
+///
+/// Fetches all vault entries for this account up to and including the target block, ordered by
+/// `vault_key` ascending and `block_num` descending. This allows efficient in-memory deduplication
+/// by taking only the first entry for each `vault_key`.
+///
+/// Note: Diesel's query DSL doesn't support joining with derived tables (subqueries with GROUP BY),
+/// so we use the ordering-based approach which is efficient due to the database index on
+/// `(account_id, vault_key, block_num)`.
```
Mirko-von-Leipzig: Could you give the example you'd want to write ito sql? Is this not possible?

```sql
SELECT vault_key, asset FROM account_vault_assets
    WHERE 
        account_id = ? AND
        block_num <= ?
    GROUP BY vault_key
    ORDER BY block_num DESC
```

crates/store/src/db/models/queries/accounts/tests.rs
drahnr: It doesn't really matter, as long as we have a few distinct blocks to test the "give me the newest up until x" query patterns.

crates/store/src/db/models/queries/accounts/at_block.rs
drahnr: We will, it's not implemented yet, we will either do that as part of the `fn apply_block` or do a cyclic task (this approach) but given recent/current events I tend more towards `apply_block` (with a timeout).

crates/store/src/db/models/queries/accounts/at_block.rs
drahnr: What is not possible is using temporary tables as argument to inner join: 

```
SELECT ac.*
FROM account_vault_assets ac
INNER JOIN (
    SELECT account_id, vault_key, MAX(block_num) AS max_block_num
    FROM account_changes
    GROUP BY (account_id, vault_key)
) latest ON ac.account_id = latest.account_id
AND ac.vault_key = latest.vault_key
AND ac.block_num = latest.max_block_num;
```

drahnr: *which is really what I want to do

Mirko-von-Leipzig: I'm not following why we need the inner join, isn't this equivalent:

```sql
SELECT asset, MAX(block_num) FROM account_vault_assets
WHERE 
  block_num <= ? AND
  account_id = ?
GROUP BY
  vault_key
```

drahnr: tl;dr it's a "feature" of sqlite ( see https://www3.sqlite.org/quirks.html#aggregate_queries_can_contain_non_aggregate_result_columns_that_are_not_in_the_group_by_clause ) but we don't want to bake too much of that into the store/queries.

PR #1499, opened 2026-01-12
feat: move `rocksdb` backend for `LargeSmt` from `-crypto` to `-node`

### LINKED ISSUE #1218: Extract `RocksDbStorage` smt storage backend into a separate crate
Move the `RocksDbStorage` implementation out of miden-`crypto` into a separate crate (e.g. under `miden-node`).

This would simplify `miden-crypto` by removing I/O dependencies and feature-gating complexities, keeping its surface minimal and focused on cryptographic primitives.

drahnr:
## Changes

Copies the RocksDB backend implementation to the `miden-node` repo.

Closes #1218

## Prerequisites

Requires https://github.com/0xMiden/crypto/pull/850 + a release of `miden-crypto @ 0.19.x`.

### DIFF
--- a/crates/large-smt-backend-rocksdb/src/rocksdb.rs
+++ b/crates/large-smt-backend-rocksdb/src/rocksdb.rs
@@ -1173,7 +1173,7 @@ impl RocksDbConfig {
     ///
     /// # Examples
     /// ```
-    /// use miden_crypto::merkle::smt::RocksDbConfig;
+    /// use miden_large_smt_backend_rocksdb::RocksDbConfig;
     ///
     /// let config = RocksDbConfig::new("/path/to/database");
     /// ```
@@ -1196,7 +1196,7 @@ impl RocksDbConfig {
     ///
     /// # Examples
     /// ```
-    /// use miden_crypto::merkle::smt::RocksDbConfig;
+    /// use miden_large_smt_backend_rocksdb::RocksDbConfig;
     ///
     /// let config = RocksDbConfig::new("/path/to/database")
     ///     .with_cache_size(2 * 1024 * 1024 * 1024); // 2GB cache
@@ -1218,7 +1218,7 @@ impl RocksDbConfig {
     ///
     /// # Examples
     /// ```
-    /// use miden_crypto::merkle::smt::RocksDbConfig;
+    /// use miden_large_smt_backend_rocksdb::RocksDbConfig;
     ///
     /// let config = RocksDbConfig::new("/path/to/database")
     ///     .with_max_open_files(1024); // Allow up to 1024 open files


crates/large-smt-backend-rocksdb/src/rocksdb.rs
```diff
@@ -0,0 +1,930 @@
...
+    tbl.set_pin_l0_filter_and_index_blocks_in_cache(true);
+
+    let mut opts = Options::default();
+    opts.set_block_based_table_factory(&tbl);
+    opts.set_write_buffer_size(128 << 20);
+    opts.set_max_write_buffer_number(3);
+    opts.set_min_write_buffer_number_to_merge(1);
+    opts.set_max_write_buffer_size_to_maintain(0);
+    opts.set_compaction_style(DBCompactionStyle::Level);
+    opts.set_level_zero_file_num_compaction_trigger(4);
```
krushimir: nit (pre-existing): `set_level_zero_file_num_compaction_trigger` is called twice - first with 4, then overwritten with 8. The first call is dead code. Since this is a fresh copy, worth cleaning up.

crates/large-smt-backend-rocksdb/src/rocksdb.rs
```diff
@@ -0,0 +1,930 @@
...
+        let db_key = Self::index_db_key(index);
+
+        match self.db.get_cf(cf, db_key).map_rocksdb_err()? {
+            Some(bytes) => Ok(Some(SmtLeaf::read_from_bytes(&bytes)?)),
+            None => Ok(None),
+        }
+    }
+
+    fn set_leaves(&self, leaves: Map<u64, SmtLeaf>) -> Result<(), StorageError> {
+        let leaves_cf = self.cf_handle(LEAVES_CF)?;
```
krushimir: The original also writes `leaf_count` and `entry_count` to metadata here. This version only writes the leaves, so the counts will be stale after calling `set_leaves`.

crates/large-smt-backend-rocksdb/src/rocksdb.rs
```diff
@@ -0,0 +1,930 @@
...
+    fn remove_subtree(&self, index: NodeIndex) -> Result<(), StorageError> {
+        let cf = self.subtree_cf(index);
+        let key = Self::subtree_db_key(index);
+        self.db.delete_cf(cf, key).map_rocksdb_err()?;
+        Ok(())
+    }
+
+    fn get_inner_node(&self, index: NodeIndex) -> Result<Option<InnerNode>, StorageError> {
+        let subtree = self.get_subtree(index)?;
+        Ok(subtree.and_then(|s| s.get_inner_node(index)))
```
krushimir: The method passes index directly to `get_subtree`, but the original first maps to the containing subtree root via `Subtree::find_subtree_root(index)`. Without that, any depth that isn't a subtree root (`24`/`32`/`40`/`48`/`56`) will panic in `cf_for_depth`. 

Same issue in `set_inner_node `and `remove_inner_node` below.

drahnr: Full disclosure: This is an AI change that slipped through my self review.

crates/large-smt-backend-rocksdb/src/rocksdb.rs
```diff
@@ -0,0 +1,930 @@
...
+            match res.map_rocksdb_err()? {
+                Some(bytes) => out.push(Some(Subtree::from_vec(idx, &bytes)?)),
+                None => out.push(None),
+            }
+        }
+        Ok(out)
+    }
+
+    fn set_subtree(&self, subtree: &Subtree) -> Result<(), StorageError> {
+        let index = subtree.root_index();
```
krushimir: In `crypto`, the RocksDB implementation updates the `DEPTH_24_CF `cache in `set_subtree`, `set_subtrees`, and `remove_subtree `(writing the root hash for depth-24 subtrees, and deleting it on removal). In this version, those methods don’t update the cache, so `get_depth24()` can return stale data. Right now, only `apply()` appears to maintain cache consistency.

These methods may not be used in node’s current flow (if operations are always batched), so this might not be user-visible today,  but I think it’s still safer to keep them correct.

crates/large-smt-backend-rocksdb/src/rocksdb.rs
```diff
@@ -0,0 +1,930 @@
...
+            let key = Self::index_db_key(index);
+            match maybe_leaf {
+                Some(leaf) => batch.put_cf(leaves_cf, key, leaf.to_bytes()),
+                None => batch.delete_cf(leaves_cf, key),
+            }
+        }
+
+        // Process subtree updates
+        let subtree_ops: Result<Vec<_>, StorageError> = subtree_updates
+            .into_iter()
```
krushimir: It seems rayon parallelism was removed (`into_par_iter()` → `into_iter()`). I assume the rayon overhead isn't worth it?

edit: same for `get_subtrees`

Cargo.toml
```diff
@@ -45,33 +46,43 @@ debug = true
...
-miden-standards       = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
-miden-testing         = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
-miden-tx              = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
-miden-tx-batch-prover = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+# miden-block-prover    = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+# miden-protocol        = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
+# miden-standards       = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+# miden-testing         = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+# miden-tx              = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
+# miden-tx-batch-prover = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
```
sergerad: Undo this comment/move change?

crates/large-smt-backend-rocksdb/src/rocksdb.rs
```diff
@@ -0,0 +1,1329 @@
...
+    /// * `path` - The filesystem path where the RocksDB database will be stored. This can be any
+    ///   type that converts into a `PathBuf`.
+    ///
+    /// # Default Settings
+    /// * `cache_size`: 1GB (1,073,741,824 bytes)
+    /// * `max_open_files`: 512
+    ///
+    /// # Examples
+    /// ```
+    /// use miden_crypto::merkle::smt::RocksDbConfig;
```
krushimir: minor nit: this doc example still references old location.

Same on lines `1199 `and `1221`.

Cargo.toml
drahnr: (Un-)done

PR #1501, opened 2026-01-12
fix: send tx_inputs in monitor, use named storage slots

SantiagoPittella:
The transactions sent by the monitor were failing due to the missing transaction inputs considering that now are required to be validated.

Also, updates the counter MASM to use named storage slots for the owner of the account.

### DIFF
--- a/bin/network-monitor/src/counter.rs
+++ b/bin/network-monitor/src/counter.rs
@@ -102,7 +102,7 @@ async fn fetch_counter_value(
             .get_item(&COUNTER_SLOT_NAME)
             .context("failed to get counter storage slot")?;
 
-        let value = word.as_elements().first().expect("a word is always 4 elements").as_int();
+        let value = word[0].as_int();
 
         Ok(Some(value))
     } else {


bin/network-monitor/src/counter.rs
```diff
@@ -95,9 +96,13 @@ async fn fetch_counter_value(
...
-        let storage_slot = account.storage().slots().first().expect("storage slot is always value");
-        let word = storage_slot.value();
-        let value = word.as_elements().last().expect("a word is always 4 elements").as_int();
+        // Access the counter slot by name to avoid index-ordering issues
+        let word = account
+            .storage()
+            .get_item(&COUNTER_SLOT_NAME)
+            .context("failed to get counter storage slot")?;
+
+        let value = word.as_elements().first().expect("a word is always 4 elements").as_int();
```
Mirko-von-Leipzig: `word` should be an array or no? Can't we just raw access the element and the compiler will be fine?

Mirko-von-Leipzig (approved):

bin/network-monitor/src/counter.rs
SantiagoPittella: it is, yes. I changed it.

bobbinth (approved):
Looks good! Thank you!

PR #1503, opened 2026-01-12
fix: wait for SQLite connections to close before restarting store in tests

huitseeker:
Tests using `shutdown_background()` to restart the store can fail with "database is locked" because it returns immediately without waiting for connections to close.

Use `shutdown_timeout()` via `spawn_blocking` to properly wait.

Could explain https://github.com/0xMiden/miden-node/actions/runs/20910445122/job/60072127330

---

Edit: found #1510 looking at this

### DIFF
--- a/crates/block-producer/src/server/tests.rs
+++ b/crates/block-producer/src/server/tests.rs
@@ -115,7 +115,7 @@ async fn block_producer_startup_is_robust_to_network_failures() {
 
     // kill the store
     // Use spawn_blocking because shutdown_timeout blocks and can't run in async context
-    task::spawn_blocking(move || store_runtime.shutdown_timeout(Duration::from_secs(5)))
+    task::spawn_blocking(move || store_runtime.shutdown_timeout(Duration::from_millis(500)))
         .await
         .expect("shutdown should complete");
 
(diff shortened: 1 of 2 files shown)


sergerad (approved):

crates/block-producer/src/server/tests.rs
```diff
@@ -114,7 +114,10 @@ async fn block_producer_startup_is_robust_to_network_failures() {
     assert!(response.is_ok());
 
     // kill the store
-    store_runtime.shutdown_background();
+    // Use spawn_blocking because shutdown_timeout blocks and can't run in async context
+    task::spawn_blocking(move || store_runtime.shutdown_timeout(Duration::from_secs(5)))
+        .await
+        .expect("shutdown should complete");
```
drahnr, line 120: Any particular reason to a) choose 5 b) to not inline `store_runtime.shutdown_timeout(_)` with a much smaller value (i.e. 100ms?). The current approach appears to introduce another race with the next request + assert.

huitseeker: The `spawn_blocking(...).await` already waits for completion, so I don't see an obvious race with subsequent code. What did I miss?

Agree the 5s is overkill, reduced to 500ms, as it's the store stress test default.

drahnr: Ah, yes, this is a `tokio::test`; nvm

drahnr (approved):
We might want to make the `shutdown_timeout` a const (nit), otherwise LGTM

PR #1504, opened 2026-01-12
feat(store): improve telemetry and tracing

### LINKED ISSUE #1269: `apply_block` drops errors and trace context
A recent block production failure on `devnet` was difficult to debug because error context was dropped in several locations.

### Investigation summary

The first bug is being fixed in #1267 where the gRPC error wasn't propagated as part of the error report, resulting in

```
failed to apply block to store
caused by: gRPC client error
```

Following the error manually on the store side, we get to

```
status: Internal, message: "block applying was cancelled because of closed channel on database side", details: [], metadata: MetadataMap { headers: {} }
```

which is similarly unhelpful. This is due to the way `apply_store` works at the top-level, which is that two  tasks are spawned and synchronized using a channel instead of a semaphore. And the tasks errors are erased and dropped along the way. 

Manually finding the failing task we finally get

```
table transactions has no column named initial_state_commitment
```

which was due to a breaking database schema change which we didn't reset the node for. This is being tracked in #1268, whereas this issue's focus is the tracing failure.

### Goal

This is strictly for `apply_block`

- Properly instrument the various tracing components for `apply_block` and its tasks.
    - This should include the various db inserts/writes, any waits for locks, and db commit ideally.
- Ensure that the error passed along the gRPC interface is an error report and not just a string. This is an internal interface so we don't have to obscure errors.
- Either remove the concurrent tasks (my preference), or at least replace the channel with a more appropriate sync tool like a `Barrier`, or are joined together.
- Ensure errors from spawned tasks are joined, handled and propagated.

Mirko-von-Leipzig:
> [!CAUTION]
> This targets `main`

Closes #1269.

This PR improves the tracing and telemetry situation in the store; in particular for the entire `apply_block` sequence.

Notably:

- gRPC errors now include the complete error report and not just the top most string
    - This is an internal method so its safe to do so.
- I've instrumented the different parts of the `apply_block` and child functions
    - Its difficult to be certain since `async` makes this non-obvious what gets attached where.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Enhancements
 
 - Added Faucet metadata to the `miden-network-monitor` binary ([#1373](https://github.com/0xMiden/miden-node/pull/1373)).
+- Improve telemetry in the store ([#1504](https://github.com/0xMiden/miden-node/pull/1504)).
 
 ### Fixes
 


crates/store/src/state.rs
```diff
@@ -408,6 +409,7 @@ impl State {
             // change in-memory state, so we return a block applying error and don't proceed with
             // in-memory updates.
             db_update_task
+                .in_current_span()
```
huitseeker, line 412: Nit: The spawned future already has its span it applied at line 375. This one wraps the `JoinHandle` await, so it runs in the "update trees" span. Might confuse readers since the task itself already carries the parent span.

crates/store/src/server/block_producer.rs
```diff
@@ -35,7 +36,6 @@ impl block_producer_server::BlockProducer for StoreApi {
         target = COMPONENT,
         name = "store.block_producer_server.get_block_header_by_number",
         skip_all,
```
huitseeker, line 38: just confirming: removing `ret(level = "debug")` from these endpoints is intentional to reduce log noise?

bobbinth (approved):
Looks good! Thank you! I'll merge as is and we can address @huitseeker's comments in a follow-up (if needed).

crates/store/src/state.rs
Mirko-von-Leipzig: Addressed in #1508

PR #1511, opened 2026-01-13
feat(store): add database size telemetry

Mirko-von-Leipzig:
> [!CAUTION]
> This targets `main`

Once a minute, this now emits database size metrics.

Emits a `database.table.{name}.size = bytes` for every table in the database as well as the total database size.

This is run as a basic loop task.

> [!WARNING]
> This isn't a super fast query.. so its unclear if this should really be used. Current testnet this takes ~20s on a cold start from the CLI, and ~2s thereafter. I'm hoping it would be quicker in-process, but :shrug: Maybe changing from once a minute to once an hour would suffice? Its a read-only tx so it should only block the WAL file and nothing else.

### DIFF
--- a/crates/store/src/server/mod.rs
+++ b/crates/store/src/server/mod.rs
@@ -137,8 +137,9 @@ impl Store {
         join_set.spawn(async move {
             // Manual tests on testnet indicate each iteration takes ~2s once things are OS cached.
             //
-            // 60s seems like a reasonable interval, though it could also be longer.
-            let mut interval = tokio::time::interval(Duration::from_secs(60));
+            // 5 minutes seems like a reasonable interval, where this should have minimal database
+            // IO impact while providing a decent view into table growth over time.
+            let mut interval = tokio::time::interval(Duration::from_secs(5 * 60));
             let database = Arc::clone(&state);
             loop {
                 interval.tick().await;
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -550,7 +550,7 @@ impl Db {
         Ok(())
     }
 
-    /// Runs database optimization.
+    /// Emits size metrics for each table in the database, and the entire database.
     #[instrument(target = COMPONENT, skip_all, err)]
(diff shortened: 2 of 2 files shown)


crates/store/src/server/mod.rs
```diff
@@ -134,6 +134,15 @@ impl Store {
             Ok(())
         });
 
+        join_set.spawn(async move {
+            let database = Arc::clone(&state);
+            let mut interval = tokio::time::interval(Duration::from_secs(60));
```
drahnr: nit: Adding the expected execution time would be useful to avoid future necromancy for the rationale.

crates/store/src/server/mod.rs
```diff
@@ -134,6 +134,18 @@ impl Store {
             Ok(())
         });
 
+        join_set.spawn(async move {
+            // Manual tests on testnet indicate each iteration takes ~2s once things are OS cached.
+            //
+            // 60s seems like a reasonable interval, though it could also be longer.
```
bobbinth: To confirm: it takes two seconds to run "database.analyze_table_sizes()"? Is the database locked during this time?

Maybe, to be on the safer side, we should increase the interval to something like 5 mins? It is unlikely that the size changes too much in 5 mins.

Mirko-von-Leipzig: Correct, and this isn't writing to the db so it shouldn't lock. I'll bump the interval.

Mirko-von-Leipzig:
> Does blocking the WAL file mean that write transactions could not be committed until this query completes? If so, and if it does take 20 seconds, we'd not be able to write blocks into the DB for 20 seconds, right?

No, this only blocks the WAL file from being merged back into the database itself. Writes will still be appended to the WAL file itself. No blocking is done beyond holding a database transaction for that length of time.

> Also, if we run the query once per hour - would it be considered a "cold start" (and take 20 seconds rather than 2)? Could be the same issue for 5-min interval.

Probably yes to both; though its difficult to know for certain. We'll have to try it and see.

PR #1512, opened 2026-01-13
feat: mark the rpc as inhealhy if chain is stale

### LINKED ISSUE #1506: Network monitor: detect stale chain tip and mark RPC as unhealthy
Add health monitoring to detect when the RPC component stops making progress. If the chain tip does not increment within a configurable time threshold, the RPC should be marked as unhealthy.

SantiagoPittella:
closes #1506

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Added gRPC-Web probe support to the `miden-network-monitor` binary ([#1484](https://github.com/0xMiden/miden-node/pull/1484)).
 - Add DB schema change check ([#1268](https://github.com/0xMiden/miden-node/pull/1485)).
 - Improve DB query performance for account queries ([#1496](https://github.com/0xMiden/miden-node/pull/1496).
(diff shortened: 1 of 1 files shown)


bin/network-monitor/src/status.rs
```diff
@@ -377,17 +441,43 @@ pub(crate) async fn check_rpc_status(
...
+                        chain_tip = store_status.chain_tip,
+                        stale_duration_secs = stale_duration,
+                        "Chain tip is stale"
+                    );
+                    return ServiceStatus {
+                        name: "RPC".to_string(),
+                        status: Status::Unhealthy,
+                        last_checked: current_time,
+                        error: Some(format!(
+                            "Chain tip {} has not changed for {} seconds",
```
Mirko-von-Leipzig, line 467: Though you use the current time, and not the block's time correct? Maybe fine as is then.

```suggestion
                            "Latest block {} is {} seconds old",
```

SantiagoPittella: > Though you use the current time, and not the block's time correct?

Yes, I use current time

PR #1516, opened 2026-01-15
feat(store): enable database traces

Mirko-von-Leipzig:
> [!WARNING]
> Targets `main`

This enables the `trace` feature in `deadpool` which is otherwise disabled by default. 

This means that tracing hasn't worked inside any of our database interactions since these are passed into `deadpool` which only passes the span on if `trace` is enabled.

I'm primarily interested in this since this will actually emit database table sizes, though it likely also fixes some other inconsistencies that I've seen in the traces.

In addition this PR bumps version to v0.12.8.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
-## v0.12.7 (2026-01-14)
+## v0.12.8 (2026-01-15)
+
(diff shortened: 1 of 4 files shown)


bobbinth (approved):
Looks good! Thank you!

PR #1517, opened 2026-01-15
store: limit number of requested storage map keys / smt proofs returned

### LINKED ISSUE #1514: Limit number of `SmtProof`s in `get_account` API
When returning `SmtProof`s we should be more restrictive than the `MAX_ENTRIES` threshold, since they're both big in size and need more CPU cycles to obtain.

https://github.com/0xMiden/miden-node/pull/1428#discussion_r2689524207

drahnr:
Closes #1514

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Add DB schema change check ([#1268](https://github.com/0xMiden/miden-node/pull/1485)).
 - Improve DB query performance for account queries ([#1496](https://github.com/0xMiden/miden-node/pull/1496).
 - Limit number of storage map keys in `GetAccount` requests ([#1517](https://github.com/0xMiden/miden-node/pull/1517)).
+- The network monitor now marks the chain as unhealthy if it fails to create new blocks ([#1512](https://github.com/0xMiden/miden-node/pull/1512)).
 
 ### Changes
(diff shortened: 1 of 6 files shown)


crates/proto/src/domain/account.rs
```diff
@@ -435,6 +435,12 @@ impl AccountStorageMapDetails {
     /// Maximum number of storage map entries that can be returned in a single response.
     pub const MAX_RETURN_ENTRIES: usize = 1000;
 
+    /// Maximum number of SMT proofs that can be returned in a single response.
+    ///
+    /// This limit is more restrictive than [`Self::MAX_RETURN_ENTRIES`] because SMT proofs
+    /// are larger (up to 64 inner nodes each) and more CPU-intensive to generate.
+    pub const MAX_SMT_PROOF_ENTRIES: usize = 16;
```
bobbinth, line 442: We should also be checking this on the request side - i.e., if the user requests 1000 proofs, this should be rejected by the RPC rather than being forwarded to the store. And ideally we'd use the same constant in both places. Let's create an issue for this.

bobbinth: Actually, I just re-opened https://github.com/0xMiden/miden-node/issues/1514.

PR #1519, opened 2026-01-15
feat/store: dynamic get account response variant

### LINKED ISSUE #1518: store/GetAccount: Dynamic response based on map leaf number
> Another interesting aspect, somewhat related to this: we can try to get more clever and in some cases returning all entries rather than SMT proofs. For example, if a storage map has 20 entries - returning all 20 key-value pairs will almost always be cheaper than returning 10 SMT proofs. But that's probably for a different issue.

From #1514

drahnr:
Closes #1518

### DIFF
--- a/crates/store/src/inner_forest/tests.rs
+++ b/crates/store/src/inner_forest/tests.rs
@@ -457,8 +457,8 @@ fn test_open_storage_map_returns_all_entries_when_proofs_exceed_entries() {
 
     // Request proofs for both keys. With only 2 entries, proof nodes (2 * 65 = 130)
     // exceeds total entries (2), so we should get AllEntries instead of proofs.
-    let keys = vec![key1, key2];
-    let result = forest.open_storage_map(account_id, slot_name.clone(), block_num, &keys);
+    let requested_keys = vec![key1, key2];
+    let result = forest.open_storage_map(account_id, slot_name.clone(), block_num, &requested_keys);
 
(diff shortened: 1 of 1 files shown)


crates/proto/src/domain/account.rs
```diff
@@ -441,6 +452,33 @@ impl AccountStorageMapDetails {
...
+        if proofs_inefficient {
+            if total_entries <= Self::MAX_RETURN_ENTRIES {
+                StorageMapResponseVariant::AllEntries
+            } else {
+                StorageMapResponseVariant::LimitsExceeded
+            }
+        } else {
+            StorageMapResponseVariant::Proofs
+        }
+    }
```
bobbinth, line 480: I think making this estimate correctly is going to be pretty tricky. For example, `proof_nodes` currently greatly overestimates the size of SMT proofs because:

1. We send compressed Merkle paths (i.e., `SparseMerklePath`) where most nodes would likely take up just 1 bit.
2. Once we start sending `PartialSmt`, we'd also be eliminating more internal nodes.

My guess, is that the current estimate is probably off by 5x - 10x. We could get much more accurate by doing some statistical analysis (e.g., given the size of the map and assuming that keys are randomly distributed, we can expect, on average, this many nodes to be empty in a given path) - but this is much more involved, and I'm not sure we need to spend time on this now (the priority of this issue is pretty low).

PR #1520, opened 2026-01-15
feat(block-producer): detect and abort on chain desync

### LINKED ISSUE #1513: Block producer desync recovery
The `block-producer` should be able to recover if its view of the chain desync's from the canonical chain in the `store`.

------------

The `store` holds the canonical block chain data. The `block-producer`'s local view of this is maintained since it is the sole submitter of new blocks.

Under extreme circumstances, the local view can desync from the canonical chain. Under the current design, this can occur if:

1. `block-producer` submits block `N+1` to `store`
2. `block-producer` gRPC request is cancelled (e.g. timeout exceeded)
    - local view reverts to `N`
3. Block `N+1` is processed and appended in the `store`

Some scenarios in which this can occur:
- gRPC timeout exceeded
- `block-producer` restarts in the middle of processing block `N+1`
  - on startup it would "sync" to block `N`, and miss `N+1`
- bugs, e.g.
  - `store` causes request to be incorrectly cancelled or rejected
  - `block-producer` cancels request

All of these are fairly extreme cases, and _should_ be rare under most circumstances. I therefore propose that we don't need a fully robust mechanism for recovery, but rather one that can be implemented easily i.e. abort and restart.

---------------

We firstly need to identify that something has gone wrong. We could add an error code to the store's block submission API, which would let us know when to check for this desync situation. However, since we expect block submission to succeed in the vast majority of cases, it will be much simpler to simply request the chain tip if block submission fails and then abort if there is a desync.

Mirko-von-Leipzig:
This detects situations where the block-producer disagrees with the store on the current chain tip. This _should_ only be possible under rare circumstances, but one possibility is:

- block-producer submits block `N+1` to store via gRPC
- store processes and persists block `N+1`
- block-producer cancels its gRPC request e.g. due to timeout
- block-producer rolls back local state to `N`
- block production spins endlessly as block `N+1` is now rejected by `store`

There are multiple points to detect this. This code determines by comparing the block number being built against the input's header returned from the `store`. The block builder task panics, which should abort the block-producer process.

Note that this error isn't really recoverable; and requires a full reset of the block-producer. Aborting and restarting is the easiest way to implement this.

Closes #1513.

### DIFF
--- a/crates/block-producer/src/block_builder/mod.rs
+++ b/crates/block-producer/src/block_builder/mod.rs
@@ -1,8 +1,8 @@
 use std::ops::{Deref, Range};
 use std::sync::Arc;
 
+use anyhow::Context;
 use futures::FutureExt;
-use futures::never::Never;
 use miden_block_prover::LocalBlockProver;
 use miden_node_utils::tracing::OpenTelemetrySpanExt;
 use miden_protocol::MIN_PROOF_SECURITY_LEVEL;
@@ -79,13 +79,15 @@ impl BlockBuilder {
     }
     /// Starts the [`BlockBuilder`], infinitely producing blocks at the configured interval.
     ///
+    /// Returns only if there was a fatal, unrecoverable error.
+    ///
     /// Block production is sequential and consists of
     ///
     ///   1. Pulling the next set of batches from the mempool
     ///   2. Compiling these batches into the next block
     ///   3. Proving the block (this is simulated using random sleeps)
     ///   4. Committing the block to the store
-    pub async fn run(self, mempool: SharedMempool) {
+    pub async fn run(self, mempool: SharedMempool) -> anyhow::Result<()> {
         assert!(
             self.failure_rate < 1.0 && self.failure_rate.is_sign_positive(),
             "Failure rate must be a percentage"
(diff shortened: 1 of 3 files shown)


crates/block-producer/src/block_builder/mod.rs
```diff
@@ -207,6 +207,20 @@ impl BlockBuilder {
...
+        // the block-producer's apply_block gRPC request times out, rolling back the block locally,
+        // but the store still committed the block on its end.
+        //
+        // This is not recoverable since the mempool cannot reconstruct the block it rolled back.
+        assert_eq!(
+            block_number,
+            inputs.prev_block_header().block_num().child(),
+            "Desync between mempool and store detected. Block number {block_number} selected from mempool should be the child of the latest committed block in the store {}",
+            inputs.prev_block_header().block_num()
+        );
```
bobbinth: Would the error also be submitted to the telemetry? Or we don't care about it?

CHANGELOG.md
```diff
@@ -23,6 +23,9 @@
 - Add DB schema change check ([#1268](https://github.com/0xMiden/miden-node/pull/1485)).
 - Improve DB query performance for account queries ([#1496](https://github.com/0xMiden/miden-node/pull/1496).
 - The network monitor now marks the chain as unhealthy if it fails to create new blocks ([#1512](https://github.com/0xMiden/miden-node/pull/1512)).
+- Block producer now detects if it is desync'd from the store's chain tip and aborts ([#1520](https://github.com/0xMiden/miden-node/pull/1520)).
+  - This could occur in rare situations where the block is submitted to the store, and the request is only cancelled from the block-producer's perspective.
+  - For example, due to a gRPC timeout.
```
bobbinth: I'm not sure these two lines are necessary.

bobbinth (approved):

crates/block-producer/src/block_builder/mod.rs
Mirko-von-Leipzig: We have code that _should_ be catching panics from any context and logging those.

You make a good point though, maybe instead I should be pushing this back up as an error to `main` and handling it more correctly. I'll see if that isn't easy enough.

PR #1521, opened 2026-01-16
feat: Add FPI support to NtxDataStore

### LINKED ISSUE #1387: Make FPI call possible during network transaction

sergerad:
## Context

The NTX Builder's `NtxDataStore` does not handle foreign accounts for the purposes of the following functions of `DataStore` trait:
* `get_vault_asset_witnesses()`

Closes #1387.

## Changes
* Updates `NtxDataStore` to handle foreign accounts in the above-listed functions.
* Update NTX Builder store client to support `get_account()` like the RPC client.

### DIFF
--- a/crates/ntx-builder/src/actor/execute.rs
+++ b/crates/ntx-builder/src/actor/execute.rs
@@ -343,8 +343,8 @@ struct NtxDataStore {
     /// - When a foreign account is loaded (in `get_foreign_account_inputs`).
     ///
     /// The assumption here are:
-    /// - Once an account is loaded, the mapping between `(account_id, map_root)` and slot names
-    ///   do not change. This is always the case.
+    /// - Once an account is loaded, the mapping between `(account_id, map_root)` and slot names do
+    ///   not change. This is always the case.
     /// - New storage slots created during transaction execution will not be accesses in the same
     ///   transaction. The mechanism for adding new storage slots is not implemented yet, but the
     ///   plan for it is consistent with this assumption.


crates/ntx-builder/src/actor/execute.rs
```diff
@@ -381,9 +384,52 @@ impl DataStore for NtxDataStore {
     fn get_foreign_account_inputs(
         &self,
         foreign_account_id: AccountId,
-        _ref_block: BlockNumber,
+        ref_block: BlockNumber,
     ) -> impl FutureMaybeSend<Result<AccountInputs, DataStoreError>> {
-        async move { Err(DataStoreError::AccountNotFound(foreign_account_id)) }
+        debug_assert_eq!(ref_block, self.reference_header.block_num());
```
Mirko-von-Leipzig: Should this be `assert_eq!`? We will never be running this in debug mode except in `cargo test`.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -381,9 +384,52 @@ impl DataStore for NtxDataStore {
...
+
+            // Retrieve the account proof from the store.
+            let account_request = AccountRequest {
+                account_id: foreign_account_id,
+                block_num: Some(ref_block),
+                details: None,
+            };
+            let account =
+                store.get_account(account_request).await.map_err(|err| DataStoreError::Other {
+                    error_msg: format!("failed to get account proof from store: {err}").into(),
```
Mirko-von-Leipzig:
```suggestion
                    error_msg: "failed to get account proof from store".into(),
```

crates/store/src/server/ntx_builder.rs
```diff
@@ -198,6 +198,28 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
...
+         ret(level = "debug"),
+         err
+     )]
+    async fn get_account(
+        &self,
+        request: Request<proto::rpc::AccountRequest>,
+    ) -> Result<Response<proto::rpc::AccountResponse>, Status> {
+        debug!(target: COMPONENT, ?request);
+        let request = request.into_inner();
+        let account_proof_request = request.try_into()?;
```
bobbinth: nit: I'd rename `account_proof_request` into just `account_request`.

crates/proto/src/errors/mod.rs
```diff
@@ -17,8 +24,14 @@ mod test_macro;
 pub enum ConversionError {
```
bobbinth, line 17: Not for this PR, but we should probably split this error up somehow into more logically grouped variants (right now, it mixes account-related, note-related, SMT-related etc. errors together).

crates/ntx-builder/src/store.rs
```diff
@@ -135,6 +138,43 @@ impl StoreClient {
         Ok(account)
     }
 
+    #[instrument(target = COMPONENT, name = "store.client.get_account", skip_all, err)]
+    pub async fn get_account(
+        &self,
+        request: AccountRequest,
+    ) -> Result<AccountResponse, StoreError> {
```
bobbinth: Should we make this more "typed"? Instead of being a generic `get_account` it could be `get_account_inputs()` and the signature would be:
```rust
pub async fn get_account_inputs(
    &self,
    account_id: AccountId,
    ref_block: BlockNumber,
) -> Result<AccountInputs, StoreError> {
    ...
}
```
Under the hood, we'd still use the `GetAccount` endpoint for this.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -381,9 +384,36 @@ impl DataStore for NtxDataStore {
...
+        debug_assert_eq!(ref_block, self.reference_header.block_num());
+
+        let store = self.store.clone();
+        async move {
+            // Retrieve the account proof from the store.
+            let account_request = AccountRequest {
+                account_id: foreign_account_id,
+                block_num: Some(ref_block),
+                details: None,
+            };
```
bobbinth: Does setting details to `None` work here? Basically, we want to make sure:

- We get the account code back.
- We get the account header and the storage header back.

I think these may not come back if the set the details to `None`.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -392,26 +422,43 @@ impl DataStore for NtxDataStore {
...
+                    return Err(DataStoreError::other("vault root mismatch"));
+                }
+                get_asset_witnesses(vault_keys, self.account.vault())
+            } else {
+                // Get foreign account.
+                let account_request = AccountRequest {
+                    account_id,
+                    block_num: Some(self.reference_header.block_num()),
+                    details: None,
+                };
```
bobbinth: Making a request to `GetAccount` just to get an asset vault witness would be super expensive. Let's add dedicated endpoints to retrieve asset and storage map witnesses from the store (it shouldn't be too difficult).

I would add these endpoints in a separate PR and then rebase this PR on top of that PR. The new endpoints could be:

- `GetVaultAssetWitnesses`
- `GetStorageMapWitness`

These would be NTX builder-specific (i.e., not exposed via the public RPC).

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -421,28 +468,86 @@ impl DataStore for NtxDataStore {
...
+                        None
+                    }
+                })
+            } else {
+                // Get foreign account.
+                let account_request = AccountRequest {
+                    account_id,
+                    block_num: Some(self.reference_header.block_num()),
+                    details: None,
+                };
```
bobbinth: Same comment as above but for storage map witnesses.

crates/proto/src/domain/account.rs
```diff
@@ -783,6 +789,104 @@ impl From<AccountDetails> for proto::rpc::account_response::AccountDetails {
     }
 }
 
+impl TryFrom<&AccountDetails> for PartialAccount {
+    type Error = ConversionError;
```
bobbinth: Instead of implementing a generic conversion from `AccountDetails` to `PartialAccount`, I would maybe do a much more specific one and put the code in the NXT builder somewhere. The main reason is that the NTX builder-specific one could be much simpler because:

- We know we won't be requesting asset vault details - so, that code would not be needed.
- We know we won't be requesting storage map details - so, that code would not be needed either.

The above two assume that we have dedicated endpoints for asset and storage map witnesses as described in one of the previous comments.

bobbinth (commented):
Looks good! Thank you! I left some comments inline - the main ones are about making the code a bit more NTX builder-specific, and also adding helper endpoints on the store to retrieve asset/map witnesses (in a different PR).

crates/proto/src/errors/mod.rs
sergerad: https://github.com/0xMiden/miden-node/issues/1528

crates/ntx-builder/src/store.rs
sergerad: Does that involve making a type that is basically a duplicate of `AccountResponse` but has a different name (`AccountInputs`)? Or am I misinterpreting?

bobbinth: `AccountInputs` is already a struct that exists in `miden-base`. It is basically the same as `AccountResponse` but without the block number. It is the type that is returned from `DataStore::get_foreign_account_inputs()`.

We could add `block_num` to `AccountInputs` in `miden-base` and then get rid of `AccountResponse` here - but I wouldn't do it in this PR.

crates/ntx-builder/src/actor/execute.rs
sergerad: Just checking with @drahnr what content is required for our purposes here (and the other 2 calls).

crates/ntx-builder/src/actor/execute.rs
sergerad: Have started a branch for this. Just need to unblock with https://github.com/0xMiden/miden-base/pull/2305

crates/ntx-builder/src/actor/execute.rs
sergerad: Pasting for reference:


If you need account code, account header, and storage header, use:
```rust
AccountRequest {
    account_id,
    block_num: None, // latest
    details: Some(AccountDetailRequest {
        // Use a dummy/zero commitment to force receiving the code
        code_commitment: Some(Digest::default().into()),
        asset_vault_commitment: None, // don't need vault
        storage_requests: vec![], // no specific storage maps needed
    }),
}
```

crates/proto/src/domain/account.rs
sergerad: This is only used by `NtxDataStore::get_foreign_account_inputs()` to construct `AccountInputs`.

Are we saying that we only need empty `PartialStorage` and `PartialVault` when creating `PartialAccount`?

```rust
impl PartialAccount {
//...
    pub fn new(
        id: AccountId,
        nonce: Felt,
        code: AccountCode,
        partial_storage: PartialStorage,
        partial_vault: PartialVault,
        seed: Option<Word>,
    ) -> Result<Self, AccountError> {
```

bobbinth: Yes, for foreign accounts, we don't need to populate partial vault (only the root will do), and for partial storage, we just need the storage header (i.e., no need to include storage maps).

crates/ntx-builder/src/store.rs
sergerad: I would then keep it as is with `AccountResponse` because we would have to move this conversion elsewhere if we got rid of it:
```rust
impl TryFrom<proto::rpc::AccountResponse> for AccountResponse {
```

Or would you prefer to move that conversion and get rid of AccountResponse?

bobbinth: I'd put this conversion from `proto::rpc::AccountResponse` into `AccountInputs` in the NTX builder as I think it would be much more limited in scope. My assumption here is that it is pretty simple to do - but let me know if not.

crates/store/src/server/ntx_builder.rs
```diff
@@ -209,6 +209,28 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
...
+    #[instrument(
+         parent = None,
+         target = COMPONENT,
+         name = "store.ntx_builder_server.get_account",
+         skip_all,
+         level = "debug",
+         ret(level = "debug"),
+         err
+     )]
+    async fn get_account(
```
bobbinth: I think we no longer need to add the `instrument` macro here (see https://github.com/0xMiden/miden-node/pull/1553) - but cc @Mirko-von-Leipzig to confirm.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -382,9 +399,30 @@ impl DataStore for NtxDataStore {
...
+                details: Some(AccountDetailRequest {
+                    code_commitment: Some(Word::default()),
+                    asset_vault_commitment: None,
+                    storage_requests: vec![],
+                }),
+            };
+            let account_inputs =
+                store.get_account_inputs(account_request).await.map_err(|err| {
+                    DataStoreError::other_with_source("failed to get account inputs", err)
+                })?;
```
bobbinth: Once we get account inputs, we need to add storage slots to `self.storage_slots` (as we do on lines 384 - 390 above).

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -373,6 +380,16 @@ impl DataStore for NtxDataStore {
...
 
+            // Register storage slots for the account.
+            for slot in self.account.storage().slots() {
+                if let StorageSlotContent::Map(map) = slot.content() {
+                    self.storage_slots
+                        .lock()
+                        .await
+                        .insert((account_id, map.root()), slot.name().clone());
+                }
+            }
```
bobbinth: Related to the previous comment, I'd make this into a helper method that takes a reference to a `AccountStorageHeader` and updates `self.storage_slots`. This method can then be called from here and from `get_foreign_account_inputs()`.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -329,6 +333,8 @@ struct NtxDataStore {
     store: StoreClient,
     /// LRU cache for storing retrieved note scripts to avoid repeated store calls.
     script_cache: LruCache<Word, NoteScript>,
+    /// Mapping of storage map roots to storage slot names observed during various calls.
+    storage_slots: Arc<Mutex<BTreeMap<(AccountId, Word), StorageSlotName>>>,
```
bobbinth, line 357: Would be good to explain in the comment why we need these.

crates/ntx-builder/src/store.rs
```diff
@@ -135,6 +143,51 @@ impl StoreClient {
         Ok(account)
     }
 
+    #[instrument(target = COMPONENT, name = "store.client.get_account", skip_all, err)]
+    pub async fn get_account_inputs(
+        &self,
+        request: AccountRequest,
+    ) -> Result<AccountInputs, StoreError> {
```
bobbinth, line 162: Since we know we don't care about storage map and assets, I would change the signature to take account ID and block number. This should simplify the code quite a bit.

Also, let's add doc comments to this method.

crates/ntx-builder/src/store.rs
```diff
@@ -270,4 +400,46 @@ pub enum StoreError {
...
+    // Derive partial storage. Storage maps are not required for foreign accounts.
+    let partial_storage = PartialStorage::new(account_details.storage_details.header.clone(), [])?;
+
+    // Derive partial vault.
+    let assets = match &account_details.vault_details {
+        AccountVaultDetails::LimitExceeded => Err(ConversionError::AssetVaultLimitExceeded),
+        AccountVaultDetails::Assets(assets) => Ok(assets),
+    }?;
+    let asset_vault = AssetVault::new(assets)?;
+    let partial_vault = PartialVault::new(asset_vault.root());
```
bobbinth: Since we are not requesting assets from the store, this won't work. Instead, we can create a `PartialVault` directly by using the root from account header.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -393,26 +431,25 @@ impl DataStore for NtxDataStore {
...
-                    source: None,
-                });
+            if self.account.id() == account_id {
+                if self.account.vault().root() != vault_root {
+                    return Err(DataStoreError::other("vault root mismatch"));
+                }
+                get_asset_witnesses(vault_keys, self.account.vault())
+            } else {
+                let witnesses = store
+                    .get_vault_asset_witnesses(account_id, vault_keys, None)
```
bobbinth: We need to pass the reference block number here - otherwise, we'll get the witness for the latest block (but we want the witness for the reference block of the transaction).

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -422,28 +459,42 @@ impl DataStore for NtxDataStore {
...
+                // Get slot name.
+                let storage_slots = self.storage_slots.lock().await;
+                let Some(slot_name) = storage_slots.get(&(account_id, map_root)) else {
+                    return Err(DataStoreError::other(
+                        "requested storage slot has not been registered",
+                    ));
+                };
+                // Retrieve witness.
+                let witness = store
+                    .get_storage_map_witness(account_id, slot_name.clone(), map_key, None)
```
bobbinth: Same comment as above: we need to pass reference block number here.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -422,28 +459,42 @@ impl DataStore for NtxDataStore {
...
+                self.account.storage().slots().iter().find_map(|slot| {
+                    if let StorageSlotContent::Map(map) = slot.content() {
+                        if map.root() == map_root {
+                            Some(map.open(&map_key))
+                        } else {
+                            None
+                        }
+                    } else {
+                        None
                     }
```
bobbinth: This is fine for now, but as soon as we move to using partial accounts for the native account, this would need to be re-done. An alternative is to treat both native and foreign accounts the same.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -393,26 +431,25 @@ impl DataStore for NtxDataStore {
...
-                return Err(DataStoreError::Other {
-                    error_msg: "vault root mismatch".into(),
-                    source: None,
-                });
+            if self.account.id() == account_id {
+                if self.account.vault().root() != vault_root {
+                    return Err(DataStoreError::other("vault root mismatch"));
+                }
+                get_asset_witnesses(vault_keys, self.account.vault())
+            } else {
```
bobbinth: Similar to the above comment, we could probably treat both foreign and native accounts there same.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -350,6 +359,25 @@ impl NtxDataStore {
...
+        storage_header: &AccountStorageHeader,
+    ) {
+        for slot_header in storage_header.slots() {
+            if let StorageSlotType::Map = slot_header.slot_type() {
+                self.storage_slots
+                    .lock()
+                    .await
+                    .insert((account_id, slot_header.value()), slot_header.name().clone());
+            }
         }
```
bobbinth, line 396: Let's move locking outside of the loop. This way, we only need to lock once rather than doing it on every iteration.

crates/ntx-builder/src/actor/execute.rs
bobbinth: nit (and not related to this PR): I think a better name here would be `reference_block` as we use the word "header" in quite a few other contexts as well.

crates/ntx-builder/src/store.rs
```diff
@@ -137,6 +149,36 @@ impl StoreClient {
...
+        &self,
+        account_id: AccountId,
+        block_num: BlockNumber,
+    ) -> Result<AccountInputs, StoreError> {
+        // Construct proto request.
+        let proto_request = proto::rpc::AccountRequest {
+            account_id: Some(proto::account::AccountId { id: account_id.to_bytes() }),
+            block_num: Some(block_num.into()),
+            details: None,
+        };
```
bobbinth, line 174: If we send `details` as `None`, are we going to get any details back? My understanding was that to get details, we'd need to format the request as you described in https://github.com/0xMiden/miden-node/pull/1521#discussion_r2706023677.

crates/ntx-builder/src/store.rs
```diff
@@ -137,6 +149,36 @@ impl StoreClient {
         Ok(account)
     }
 
+    /// Get the inputs for an account at a given block number from the store.
+    #[instrument(target = COMPONENT, name = "store.client.get_account_inputs", skip_all, err)]
+    pub async fn get_account_inputs(
```
bobbinth, line 158: Let's expand doc comments here. Specifically:

- We should mention that we are using the `GetAccount` endpoint.
- We should explain how we limit the generic `GetAccount` requests/response here (e.g., we shape the request in a way that doesn't return maps or vault etc.).

crates/ntx-builder/src/store.rs
```diff
@@ -345,4 +466,41 @@ pub enum StoreError {
...
+}
+
+// HELPERS
+// =================================================================================================
+
+/// Builds a minimal foreign account from the provided account details.
+///
+/// The account's partial storage does not contain storage maps and the partial vault is constructed
+/// from the asset vault root only.
+pub fn build_minimal_foreign_account(
```
bobbinth: I would maybe rephrase the comments to say that this is building a partial account under these specific assumptions, and that this is useful for getting foreign account data during transaction execution.

crates/proto/src/errors/mod.rs
```diff
@@ -17,8 +25,14 @@ mod test_macro;
...
     #[error("asset error")]
     AssetError(#[from] AssetError),
+    #[error("account code missing")]
+    AccountCodeMissing,
     #[error("account error")]
     AccountError(#[from] AccountError),
+    #[error("asset vault error")]
+    AssetVaultError(#[from] AssetVaultError),
+    #[error("vault has too many assets to return inline")]
+    AssetVaultLimitExceeded,
```
bobbinth: I don't think this variant is used anywhere. Let's remove it and could you also check if other variants are still relevant?

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -423,27 +459,27 @@ impl DataStore for NtxDataStore {
...
-                }
-            }
+            // The slot name that corresponds to the given account ID and map root must have been
+            // registered during previous calls of this data store.
+            let storage_slots = self.storage_slots.lock().await;
+            let Some(slot_name) = storage_slots.get(&(account_id, map_root)) else {
+                return Err(DataStoreError::other(
+                    "requested storage slot has not been registered",
+                ));
+            };
```
Mirko-von-Leipzig, line 486: Are we essentially relying on knowing the order of operations of the caller of data store?

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -366,53 +392,63 @@ impl DataStore for NtxDataStore {
...
+            let account_inputs =
+                self.store.get_account_inputs(foreign_account_id, ref_block).await.map_err(
+                    |err| DataStoreError::other_with_source("failed to get account inputs", err),
+                )?;
+
+            // Register slot names from the foreign account for later use.
+            self.register_storage_map_slots(foreign_account_id, account_inputs.storage().header())
+                .await;
+
+            Ok(account_inputs)
```
bobbinth, line 446: Following up on the conversation we had during the call, I actually don't think we need to add foreign account code to the MAST store in this function (i.e., nothing needs to change). The reason for this is that transaction host does this after calling `DataStore::get_foreign_account_inputs()`.

Technically, I think we didn't have to do this in the validator either - but in the validator, we are planning to change how re-execution works and so, I'd leave it as is.

cc @igamigo to confirm.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -322,20 +326,42 @@ impl NtxContext {
...
+    /// - New storage slots created during transaction execution will not be accesses in the same
+    ///   transaction. The mechanism for adding new storage slots is not implemented yet, but the
+    ///   plan for it is consistent with this assumption.
+    ///
+    /// One nuance worth mentioning: it is possible that there could be a root collision where an
+    /// account has two storage maps with the same root. In this case, the map will contain only a
+    /// single entry with the storage slot name that was added last. Thus, technically, requests
+    /// to the store could be "wrong", but given that two identical maps have identical witnesses
+    /// this does not cause issues in practice.
+    storage_slots: Arc<Mutex<BTreeMap<(AccountId, Word), StorageSlotName>>>,
```
bobbinth, line 357: Expanded this comment. cc @Mirko-von-Leipzig.

crates/ntx-builder/src/actor/execute.rs
igamigo: Looking at the code, I think this is right. But also, not sure why the validator was erroring during client integration FPI tests then (which prompted #1561).

igamigo: Oh actually it's because it never even calls `DataStore::get_foreign_account_inputs()`. 

Although then I wonder how you plan on doing away with adding the MAST forest in the validator.

crates/ntx-builder/src/store.rs
```diff
@@ -332,9 +384,88 @@ impl StoreClient {
...
+                .collect(),
+            block_num: block_num.map(|num| num.as_u32()),
+        };
+
+        // Make the gRPC request.
+        let witness_proto =
+            self.inner.clone().get_vault_asset_witnesses(request).await?.into_inner();
+
+        // Convert the response to domain type.
+        let mut asset_witnesses = Vec::new();
```
igamigo, line 413: nit: can use `Vec::with_capacity` here

crates/ntx-builder/src/actor/execute.rs
igamigo: nit: Seems like you can drop the lock by this point

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -45,6 +48,7 @@ use miden_tx::{
     TransactionMastStore,
     TransactionProverError,
 };
+use tokio::sync::Mutex;
```
igamigo, line 51: Because we know this lock is low-contention and also not held across awaits, I think an `std::sync::Mutex` would be lower overhead and overall faster

PR #1524, opened 2026-01-16
ci: spell check does not need a massive runner

Mirko-von-Leipzig:


.github/workflows/lint.yml
```diff
@@ -22,7 +22,7 @@ env:
 
 jobs:
   typos:
-    runs-on: Linux-ARM64-Runner
+    runs-on: ubuntu-latest
```
drahnr:
```suggestion
    runs-on: ubuntu-24.04
```

PR #1525, opened 2026-01-16
feat/build: allow to override container runtime

drahnr:
## Why

Some distros don't ship with `docker` anymore but use `podman`, the latter also has better integration with Mac OS.

## What

Default to `docker` but allow to override by `CONTAINER_RUNTIME` env var `CONTAINER_RUNTIME=podman` or `make docker-build-node CONTAINER_RUNTIME=podman`

This PR does _not_ rename the makefile target for backwards compat, but adds another `container-..` prefixed one.

## Open Questions

1. prefix the env var with `MIDEN_`  ? Might be misleading for users of the binary :man_shrugging:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ help:
 
 WARNINGS=RUSTDOCFLAGS="-D warnings"
 BUILD_PROTO=BUILD_PROTO=1
+CONTAINER_RUNTIME ?= docker
 
 # -- linting --------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,6 +8,7 @@ help:
 
 WARNINGS=RUSTDOCFLAGS="-D warnings"
 BUILD_PROTO=BUILD_PROTO=1
+CONTAINER_RUNTIME ?= docker
```
Mirko-von-Leipzig, line 11: Supposedly we can do something like this to autoselect `podman` if available, else default to `docker`.

```suggestion
CONTAINER_RUNTIME := $(if $(shell command -v podman),podman,docker)
```

drahnr: I considered that as an alternative, but wanted to stay explicit over implicit

Mirko-von-Leipzig: I'm unsure when we would care about this setting; but I'm not overly invested in this so lets keep it explicit.

PR #1526, opened 2026-01-16
chore: update with latest miden-base

### LINKED ISSUE #1522: Replace network note detection with `NetworkAccountTarget`
When upgrading miden-node to latest miden-base, there are some relevant changes.

- `Note::is_network_note`, `NoteTag::is_single_target` and `NoteTag::NetworkAccount` were removed, so there is no more conversion path from `NoteTag` to `NetworkAccountPrefix`. Instead, notes should be detected as network notes whenever their attachment is `miden_standards::note::NetworkAccountTarget`, which can be decoded with `try_from`.
  - It seems it would make sense to redefine `NetworkAccountPrefix(u32)` as `NetworkAccountId(AccountId)` and then convert `NetworkAccountTarget::target_id` into `NetworkAccountId`.
  - Similarly, `SingleTargetNetworkNote` would conceptually be a `Note` whose attachment is `NetworkAccountTarget`.
- `NoteMetadata::aux` was removed and can be replaced with `NoteAttachment`, but most likely aux was just set to zero anyway in most places.
- `NoteMetadata::execution_hint` was moved to `NetworkAccountTarget`.

I don't see any usages of note tag or metadata in the network monitor that aren't covered by the above, but ping me for questions.

cc @SantiagoPittella

One thing I noticed while looking at usages in node was that we should change `impl TryFrom<NoteAttachment> for NetworkAccountTarget` to `&NoteAttachment`. Changed in https://github.com/0xMiden/miden-base/pull/2289.

SantiagoPittella:
closes #1522 

THis PR update with the latest miden-base changing, affecting primary the network note detection.

I wanted to test this with the network monitor and its counter increment network transaction, but I started to receive the same error that I mentioned in https://github.com/0xMiden/miden-node/pull/1490:

> status: 'Client specified an invalid argument', self: "transaction's initial state commitment 
> 0xfbdb4615d915c83feb32b3943992b305c72b4a3675cacdce407929a7507af245 does not match the account's
> current value of 0x0000000000000000000000000000000000000000000000000000000000000000", metadata:
> {"content-type": "application/grpc", "date": "Fri, 16 Jan 2026 15:25:52 GMT", "access-control-expose-headers":
> "grpc-status,grpc-message,grpc-status-details-bin", "access-control-allow-credentials": "true", "vary": "origin,
> access-control-request-method, access-control-request-headers"}

Update: fixed it by making `get_vault_asset_witnesses` return an empty vector as suggested in https://github.com/0xMiden/miden-node/pull/1493#discussion_r2696590819

### DIFF
--- a/crates/validator/src/tx_validation/data_store.rs
+++ b/crates/validator/src/tx_validation/data_store.rs
@@ -62,7 +62,7 @@ impl DataStore for TransactionInputsDataStore {
         _vault_root: Word,
         _vault_keys: BTreeSet<AssetVaultKey>,
     ) -> impl FutureMaybeSend<Result<Vec<AssetWitness>, DataStoreError>> {
-        async move { Ok(vec![]) }
+        std::future::ready(Ok(vec![]))
     }
 
     fn get_storage_map_witness(


crates/block-producer/src/domain/transaction.rs
```diff
@@ -119,7 +119,7 @@ impl AuthenticatedTransaction {
     pub fn unauthenticated_note_commitments(&self) -> impl Iterator<Item = Word> + '_ {
         self.inner
             .unauthenticated_notes()
-            .copied()
+            .cloned()
             .map(|header| header.commitment())
```
PhilippGackstatter: The clone seems unnecessary here:
```rust
.unauthenticated_notes()
.map(NoteHeader::commitment)
```

crates/ntx-builder/src/actor/inflight_note.rs
```diff
@@ -46,15 +46,9 @@ impl InflightNetworkNote {
...
+    /// The note is available if the backoff period has passed.
     pub fn is_available(&self, block_num: BlockNumber) -> bool {
-        let can_consume = self
-            .to_inner()
-            .metadata()
-            .execution_hint()
-            .can_be_consumed(block_num)
-            .unwrap_or(true);
-        can_consume && has_backoff_passed(block_num, self.last_attempt, self.attempt_count)
+        has_backoff_passed(block_num, self.last_attempt, self.attempt_count)
```
PhilippGackstatter: I think this can be brought back once `SingleTargetNetworkNote` also stores the `NoteExecutionHint`.

crates/proto/src/domain/account.rs
```diff
@@ -971,76 +972,95 @@ impl From<Asset> for proto::primitives::Asset {
...
-    type Error = NetworkAccountError;
-
-    fn try_from(value: u32) -> Result<Self, Self::Error> {
-        if value >> 30 != 0 {
-            return Err(NetworkAccountError::InvalidPrefix(value));
-        }
-        Ok(NetworkAccountPrefix(value))
+    /// Gets the 30-bit prefix of the account ID used for tag matching.
+    pub fn prefix(&self) -> AccountPrefix {
+        get_account_id_tag_prefix(self.0)
```
PhilippGackstatter, line 1048: I think we should no longer use the 30-bit account ID prefix for anything since the full account ID is now provided. This probably requires a deeper refactor. APIs like `Db::select_unconsumed_network_notes` should be able to take a full account ID now, or at least an `AccountIdPrefix` which guarantees uniqueness and that requires a DB schema update. The 30-bit prefix doesn't uniquely identify an account and one advantage of `NetworkAccountTarget` is that it no longer has the space limitations of a note tag. Someone from the node team may want to chime in.

crates/proto/src/domain/note.rs
```diff
@@ -252,10 +249,12 @@ impl TryFrom<Note> for MultiTargetNetworkNote {
     type Error = NetworkNoteError;
 
     fn try_from(note: Note) -> Result<Self, Self::Error> {
-        if note.is_network_note() && !note.metadata().tag().is_single_target() {
-            Ok(Self(note))
+        // Multi-target notes are those that do NOT have a NetworkAccountTarget attachment
+        if NetworkAccountTarget::try_from(note.metadata().attachment().clone()).is_ok() {
+            // This is a single-target note, not multi-target
+            Err(NetworkNoteError::NotMultiTarget)
```
PhilippGackstatter: I think this makes every note without a `NetworkAccountTarget` attachment a multi-target network note then? Doesn't sound quite right to me, but am not sure.

For now, only single target network notes have a standardized attachment. I think multi-target ones should have their own attachment standard, but that is to-be-designed. Iiuc, the node currently doesn't really handle multi-target network notes? cc @Mirko-von-Leipzig @sergerad

crates/proto/src/domain/note.rs
```diff
@@ -292,30 +297,29 @@ impl SingleTargetNetworkNote {
...
-        } else {
-            Err(NetworkNoteError::InvalidExecutionMode(note.metadata().tag()))
-        }
+        // Single-target network notes are identified by having a NetworkAccountTarget attachment
+        let attachment = note.metadata().attachment();
+        let target = NetworkAccountTarget::try_from(attachment.clone())
+            .map_err(|e| NetworkNoteError::InvalidAttachment(e.to_string()))?;
+        let target_id = NetworkAccountId::try_from(target.target_id())
+            .map_err(|e| NetworkNoteError::InvalidAttachment(e.to_string()))?;
+        Ok(Self { note, target_id })
```
PhilippGackstatter: I think this should also extract the `NoteExecutionHint` from `NetworkAccountTarget` and make it accessible from `SingleTargetNetworkNote`. Might make sense to store `NetworkAccountTarget` in `SingleTargetNetworkNote` directly.

Then the previous execution hint check in `InflightNetworkNote::is_available` can be brought back.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -441,11 +429,8 @@ pub(crate) fn select_unconsumed_network_notes_by_tag(
     block_num: BlockNumber,
     mut page: Page,
 ) -> Result<(Vec<NoteRecord>, Page), DatabaseError> {
-    assert_eq!(
-        NoteExecutionMode::Network as u8,
-        0,
-        "Hardcoded execution value must match query"
-    );
+    // Network notes have execution_mode = 0 in the database
+    const NETWORK_NOTE_EXECUTION_MODE: i32 = 0;
```
PhilippGackstatter: `NoteExecutionMode` in miden-protocol was removed, so this most likely needs to be done differently.

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -52,8 +52,6 @@ CREATE TABLE notes (
     sender                   BLOB    NOT NULL,
     tag                      INTEGER NOT NULL,
     execution_mode           INTEGER NOT NULL, -- 0-Network, 1-Local
-    aux                      INTEGER NOT NULL,
-    execution_hint           INTEGER NOT NULL,
```
PhilippGackstatter: I assume the attachment has to be stored in the DB now.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -757,11 +724,8 @@ impl TryInto<NoteMetadata> for NoteMetadataRawRow {
         let sender = AccountId::read_from_bytes(&self.sender[..])?;
         let note_type = NoteType::try_from(self.note_type as u32)
             .map_err(DatabaseError::conversiont_from_sql::<NoteType, _, _>)?;
-        let tag = NoteTag::from(self.tag as u32);
-        let execution_hint = NoteExecutionHint::try_from(self.execution_hint as u64)
-            .map_err(DatabaseError::conversiont_from_sql::<NoteExecutionHint, _, _>)?;
-        let aux = Felt::new(self.aux as u64);
-        Ok(NoteMetadata::new(sender, note_type, tag, execution_hint, aux)?)
+        let tag = NoteTag::new(self.tag as u32);
+        Ok(NoteMetadata::new(sender, note_type, tag))
```
PhilippGackstatter:
```suggestion
        Ok(NoteMetadata::new(sender, note_type, tag).with_attachment(todo!()))
```

I think the attachment needs to be set here as well.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -869,6 +831,10 @@ pub struct NoteInsertRowInsert {
 
 impl From<(NoteRecord, Option<Nullifier>)> for NoteInsertRowInsert {
     fn from((note, nullifier): (NoteRecord, Option<Nullifier>)) -> Self {
+        // Determine execution_mode from tag: 0 if first 2 bits are 00 (account-targeting)
+        let tag_value = note.metadata.tag().as_u32();
+        let execution_mode = i32::from(tag_value >> 30 != 0);
```
PhilippGackstatter: After the recent `NoteTag` refactor (https://github.com/0xMiden/miden-base/pull/2219) it became purely a wrapper around a `u32`. `NoteTag::with_account_target` is purely a convention, nothing that's enforced. So, I think whether a note has execution mode "network" or "local" should be determined through the presence or absence of `NetworkAccountTarget` attachment and not from the note tag.

Maybe it makes sense for the node to reintroduce `NoteExecutionMode` as it was previously defined in miden-protocol and derive it from presence or absence of `NetworkAccountTarget`, but I don't know the codebase well enough here to say if that makes sense - just an idea.

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -52,8 +52,6 @@ CREATE TABLE notes (
...
     inclusion_path           BLOB NOT NULL,    -- Serialized sparse Merkle path of the note in the block's note tree
```
PhilippGackstatter: Aux should likely be replaced by the attachment in the DB now.

Also seems like it would make sense to add a test that ensures that a note metadata _with_ an attachment stored to DB can be correctly fetched _with_ the attachment.

PhilippGackstatter (commented):
Looks good - left a few comments.

crates/proto/src/domain/account.rs
SantiagoPittella: Agree that this is not the best approach. If the node team consider this as required for this PR, I can start working on this immediately. Otherwise I can open an issue.

cc @Mirko-von-Leipzig

crates/proto/src/domain/note.rs
SantiagoPittella: I'm think that the node does _**not**_ handle multi-target notes yet.

I'm not quite sure how to follow this.

crates/store/src/db/models/queries/notes.rs
SantiagoPittella: Yeah, probably `NoteExecutionMode` is the cleanest way to go.

Though I don;t get how are we supposed to establish when a network note is multi or single account targeted

crates/proto/src/domain/account.rs
Mirko-von-Leipzig: I think that would add a lot of noise to this PR, but lets definitely address it in a follow-up. No reason to store more data than needed :)

crates/proto/src/domain/note.rs
Mirko-von-Leipzig: This type has grown a bit out-of-scope I think. We only really care about _SingleTargetNetworkNote_.

We can delete the multi-target type -- we shouldn't care about this yet.

crates/store/src/db/models/queries/notes.rs
Mirko-von-Leipzig: I think we can probably just make this a boolean `is_single_target_network` since we just want to look this up by account.

crates/store/src/db/models/queries/notes.rs
Mirko-von-Leipzig: There is essentially no multi-target yet. Only single target is now defined, and is identifiable via the attachment. 

The node (currently) also only cares about _single_ target so for now its fine if we just store `is_single_target_network_note: bool`.

crates/proto/src/domain/note.rs
```diff
@@ -221,15 +217,7 @@ impl TryFrom<Note> for NetworkNote {
...
-        if note.is_network_note() {
-            if note.metadata().tag().is_single_target() {
-                Ok(NetworkNote::SingleTarget(SingleTargetNetworkNote(note)))
-            } else {
-                Ok(NetworkNote::MultiTarget(MultiTargetNetworkNote(note)))
-            }
-        } else {
-            Err(NetworkNoteError::InvalidExecutionMode(note.metadata().tag()))
-        }
+        SingleTargetNetworkNote::try_from(note.clone()).map(NetworkNote::SingleTarget)
```
PhilippGackstatter:
```suggestion
        SingleTargetNetworkNote::try_from(note).map(NetworkNote::SingleTarget)
```

crates/proto/src/domain/note.rs
```diff
@@ -292,30 +259,31 @@ impl SingleTargetNetworkNote {
...
 
     fn try_from(note: Note) -> Result<Self, Self::Error> {
-        if note.is_network_note() && note.metadata().tag().is_single_target() {
-            Ok(Self(note))
-        } else {
-            Err(NetworkNoteError::InvalidExecutionMode(note.metadata().tag()))
-        }
+        // Single-target network notes are identified by having a NetworkAccountTarget attachment
+        let attachment = note.metadata().attachment();
+        let account_target = NetworkAccountTarget::try_from(attachment.clone())
```
PhilippGackstatter:
```suggestion
        let account_target = NetworkAccountTarget::try_from(attachment)
```

Nit: Since https://github.com/0xMiden/miden-base/pull/2289 is merged, the clone shouldn't be necessary if you update to a later base version.

crates/proto/src/domain/note.rs
```diff
@@ -346,8 +314,8 @@ where
 
 #[derive(Debug, Error)]
 pub enum NetworkNoteError {
-    #[error("note tag {0} is not a valid network note tag")]
-    InvalidExecutionMode(NoteTag),
+    #[error("note does not have a valid NetworkAccountTarget attachment: {0}")]
+    InvalidAttachment(String),
```
PhilippGackstatter, line 318:
```suggestion
    #[error("note does not have a valid NetworkAccountTarget attachment: {0}")]
    InvalidAttachment(#[source] NetworkAccountTargetError),
```

Nit: Should this use asource error instead of stringifying it?

crates/store/src/db/models/queries/notes.rs
```diff
@@ -265,8 +259,6 @@ pub(crate) fn select_existing_note_commitments(
 ///     notes.note_type,
 ///     notes.sender,
 ///     notes.tag,
-///     notes.aux,
-///     notes.execution_hint,
```
PhilippGackstatter, line 270: Nit: I think this includes the attachment now (probably applies to all the modified comments here).

crates/store/src/db/models/queries/notes.rs
```diff
@@ -854,21 +824,27 @@ pub struct NoteInsertRowInsert {
...
+    pub is_single_target_network_note: bool,
     pub inclusion_path: Vec<u8>,
 }
 
 impl From<(NoteRecord, Option<Nullifier>)> for NoteInsertRowInsert {
     fn from((note, nullifier): (NoteRecord, Option<Nullifier>)) -> Self {
+        let attachment = note.metadata.attachment();
+
+        let is_single_target_network_note =
+            NetworkAccountTarget::try_from(attachment.clone()).is_ok();
```
PhilippGackstatter:
```suggestion
        let is_single_target_network_note =
            NetworkAccountTarget::try_from(attachment).is_ok();
```

Nit: After upgrading to a later miden-base version, this clone should also not be needed.

PhilippGackstatter (approved):
Looks good to me from a protocol perspective.

I think the network account ID should eventually contain at least 64 or more bits of the account ID - ideally the full one. Anything less than 64 is not guaranteed to be unique, but probably not critical to do before 0.13 testnet.

crates/proto/src/domain/note.rs
SantiagoPittella: NetworkAccountTargetError is not publicly exposed. In miden-base we have:
```
mod network_account_target;
pub use network_account_target::NetworkAccountTarget;
```

Should we make the whole module public?

crates/ntx-builder/src/actor/account_state.rs
```diff
@@ -209,21 +206,21 @@ impl NetworkAccountState {
...
                     NetworkAccountEffect::Created(_) => {},
                 }
             }
         }
         for note in network_notes {
             assert_eq!(
-                note.account_prefix(),
-                self.account_prefix,
+                note.account_id(),
+                self.account_id,
```
Mirko-von-Leipzig, line 223: Nit: assertion text still references a prefix

crates/validator/src/tx_validation/data_store.rs
```diff
@@ -58,37 +58,11 @@ impl DataStore for TransactionInputsDataStore {
...
-                        })
-                    },
-                    Err(err) => Err(DataStoreError::Other {
-                        error_msg: "failed to open vault".into(),
-                        source: Some(err.into()),
-                    }),
-                }
-            }))
-        }
+        async move { Ok(vec![]) }
```
Mirko-von-Leipzig:
```suggestion
        std::future::ready(Ok(vec![]))
```

crates/ntx-builder/src/coordinator.rs
```diff
@@ -89,7 +89,7 @@ pub struct Coordinator {
 
     /// Cache of events received from the mempool that predate corresponding network accounts.
     /// Grouped by account prefix to allow targeted event delivery to actors upon creation.
-    predating_events: HashMap<NetworkAccountPrefix, IndexMap<TransactionId, Arc<MempoolEvent>>>,
+    predating_events: HashMap<NetworkAccountId, IndexMap<TransactionId, Arc<MempoolEvent>>>,
```
bobbinth, line 92: We should go through comments/names etc. to make sure we no longer use "prefix" terminology.

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -43,28 +43,27 @@ CREATE INDEX idx_accounts_block_num ON accounts(block_num);
...
-    serial_num               BLOB,
+    committed_at                  INTEGER NOT NULL, -- Block number when the note was committed
+    batch_index                   INTEGER NOT NULL, -- Index of batch in block, starting from 0
+    note_index                    INTEGER NOT NULL, -- Index of note in batch, starting from 0
+    note_id                       BLOB    NOT NULL,
+    note_commitment               BLOB    NOT NULL,
+    note_type                     INTEGER NOT NULL, -- 1-Public (0b01), 2-Private (0b10), 3-Encrypted (0b11)
+    sender                        BLOB    NOT NULL,
+    tag                           INTEGER NOT NULL,
+    is_single_target_network_note INTEGER NOT NULL, -- 1 if note has NetworkAccountTarget attachment, 0 otherwise
```
bobbinth, line 54: I would maybe replace this with something like `network_note_type` or `network_note_group` and use the following value:

- `0` - not a network note (same as now).
- `1` - single account target network note (same as now).

In the future, this would allow us to to extend this to other type of network notes.

Mirko-von-Leipzig: We could also repurpose `note_type` I think so that we have:

```rust
enum NoteType {
    Private,
    /// Public, non-network
    Public,
    /// Public, single-target network
    SingleTargetNetwork,
    ...
}
```

or will these not remain orthogonal?

bobbinth: Possible, but may complicate note metadata construction logic because we'd need to convert `SingleTargetNetwork` to `Public`.

PR #1529, opened 2026-01-19
feat: Add store endpoints GetVaultAssetWitnesses and GetStorageMapWitness

sergerad:
## Context

#1521 requires access to witnesses from the store. It would be preferable to retrieve just the witnesses rather than entire account data. The store does not support such retrieval right now.

## Changes
- Add `GetVaultAssetWitnesses` endpoint to store.
- Add `GetStorageMapWitness` endpoint to store.

### DIFF
--- a/crates/proto/src/generated/store.rs
+++ b/crates/proto/src/generated/store.rs
@@ -232,6 +232,8 @@ pub struct VaultAssetWitnessesRequest {
     /// Set of asset vault keys to retrieve witnesses for.
     #[prost(message, repeated, tag = "2")]
     pub vault_keys: ::prost::alloc::vec::Vec<super::primitives::Digest>,
+    /// The witnesses returned correspond to the account state at the specified block number.
+    ///
     /// Optional block number. If not provided, uses the latest state.
     ///
     /// The specified block number should be relatively near the chain tip else an error will be
@@ -275,6 +277,8 @@ pub struct StorageMapWitnessRequest {
     /// Optional block number. If not provided, uses the latest state.
     ///
     /// The witness returned corresponds to the account state at the specified block number.
+    ///
+    /// Optional block number. If not provided, uses the latest state.
     ///
     /// The specified block number should be relatively near the chain tip else an error will be
     /// returned.


proto/proto/internal/store.proto
```diff
@@ -347,3 +353,75 @@ message CurrentBlockchainData {
     // Current block header.
     optional blockchain.BlockHeader current_block_header = 2;
 }
+
+// GET VAULT ASSET WITNESSES
+// ================================================================================================
+
+// Request for vault asset witnesses for a specific account.
+message VaultAssetWitnessesRequest {
```
bobbinth: nit: I am not sure we are always consistent with the terminology, but I usually try to say "asset vault" rather than the other way around. So, for example, I'd call this `AssetVaultWitnessRequest`.

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,75 @@ message CurrentBlockchainData {
...
+    // The account ID for which to retrieve vault asset witnesses.
+    account.AccountId account_id = 1;
+
+    // The vault root hash to verify against.
+    primitives.Digest vault_root = 2;
```
bobbinth: Question: instead of passing `vault_root`, should we not pass `block_num` here?

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,75 @@ message CurrentBlockchainData {
...
+    message StorageWitness {
+        // The storage map key.
+        primitives.Digest key = 1;
+
+        // The value associated with the key.
+        primitives.Digest value = 2;
+
+        // The SMT opening proof for the key-value pair.
+        primitives.SmtOpening proof = 3;
+    }
```
bobbinth: I'm not sure we need to send the `value` here. Using the key we should be able to get the value from the `proof` (I think).

crates/store/src/state.rs
```diff
@@ -1322,6 +1322,110 @@ impl State {
...
+        Ok(witnesses)
+    }
+
+    /// Returns a storage map witness for the specified account and storage entry.
+    pub async fn get_storage_map_witness(
+        &self,
+        account_id: AccountId,
+        map_root: Word,
+        map_key: Word,
+    ) -> Result<miden_protocol::account::StorageMapWitness, crate::errors::GetStorageMapWitnessError>
```
bobbinth: A few comments about this method:

We don't need to go to the database to get the witness. Instead, we should be able to get it from `self.forest`. The current `InnerForest::open_storage_map()` method interface is not ideal for this - but we can add a new one which looks similar (and probably simpler) - e.g.:

```rust
impl InnerForest {

    pub(crate) fn get_storage_map_witness(
        &self,
        account_id: AccountId,
        slot_name: StorageSlotName,
        block_num: BlockNumber,
        key: Word,
    ) -> Result<StorageMapWitness, MerkleError>  {
        ...
    }
}
```

This would also change the signature of this method to be:
```rust
pub async fn get_storage_map_witness(
    &self,
    account_id: AccountId,
    slot_name: StorageSlotName,
    block_num: BlockNumber,
    key: Word,
) -> Result<StorageMapWitness, GetStorageMapWitnessError>
```
And the question is how simple is this to make work on the NTX side (I think it is doable - but haven't checked).

crates/store/src/state.rs
```diff
@@ -1322,6 +1322,110 @@ impl State {
...
         self.db.select_transactions_records(account_ids, block_range).await
     }
+
+    /// Returns vault asset witnesses for the specified account.
+    pub async fn get_vault_asset_witnesses(
+        &self,
+        account_id: AccountId,
+        vault_root: Word,
+        vault_keys: std::collections::BTreeSet<miden_protocol::asset::AssetVaultKey>,
+    ) -> Result<Vec<miden_protocol::asset::AssetWitness>, crate::errors::GetVaultAssetWitnessesError>
```
bobbinth: Similar to the above comment, I'd we should try to get this info from `self.forest` as it is already in-memory and would be much more efficient than getting it from the database. For this, we'd need to add something like:
```rust
impl InnerForest {

    pub(crate) fn get_vault_asset_witnesses(
        &self,
        account_id: AccountId,
        block_num: BlockNumber,
        asset_keys: BTreeSet<AssetVaultKey>,
    ) -> Result<Vec<AssetWitness>, MerkleError>  {
        ...
    }
}
```

And the signature of this method would also change to something like:
```rust
pub async fn get_vault_asset_witnesses(
    &self,
    account_id: AccountId,
    block_num: BlockNumber,
    asset_keys: BTreeSet<AssetVaultKey>,
) -> Result<Vec<AssetWitness>, GetVaultAssetWitnessesError> {
    ...
}
```

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,75 @@ message CurrentBlockchainData {
...
+message VaultAssetWitnessesRequest {
+    // The account ID for which to retrieve vault asset witnesses.
+    account.AccountId account_id = 1;
+
+    // The vault root hash to verify against.
+    primitives.Digest vault_root = 2;
+
+    // Set of asset vault keys to retrieve witnesses for.
+    repeated primitives.Digest vault_keys = 3;
+}
```
bobbinth, line 375: Related to the above comments, we could probably send `block_num` instead of `vault_root` here.

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,75 @@ message CurrentBlockchainData {
...
+    message VaultAssetWitness {
+        // The asset vault key.
+        primitives.Digest vault_key = 1;
+
+        // The asset data.
+        primitives.Asset asset = 2;
+
+        // The Merkle proof for the asset's inclusion in the vault.
+        primitives.SparseMerklePath proof = 3;
+    }
```
bobbinth, line 383: I wonder if we should just re-use `SmtOpening` for this - e.g., something like:
```proto
message VaultAssetWitness {
    primitives.SmtOpening proof = 1;
}
```

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,75 @@ message CurrentBlockchainData {
...
+
+    // The storage map root hash.
+    primitives.Digest map_root = 2;
+
+    // The storage map key for which to retrieve the witness.
+    primitives.Digest map_key = 3;
+
+    // Optional block number. If not provided, uses the latest state.
+    optional fixed32 block_num = 4;
+}
```
bobbinth, line 417: Related to the above comment, here we could send `storage_slot_name` instead of `map_root`. The tricky part would be to map map roots into slot names - but i think we may be able to do this.

proto/proto/internal/store.proto
bobbinth, line 428: I don't think we need `value` here - just `key` should be enough.

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,75 @@ message CurrentBlockchainData {
...
+
+        // The SMT opening proof for the key-value pair.
+        primitives.SmtOpening proof = 3;
+    }
+
+    // The storage map witness.
+    StorageWitness witness = 1;
+
+    // Block number at which the witness was generated.
+    fixed32 block_num = 2;
```
bobbinth, line 434: I'd also adde the `block_num` to `VaultAssetWitnessesResponse`

bobbinth (commented):
Looks good! Thank you! Not a full review, but I left some comments inline. The main one is that i think we should try to get the witnesses from the `InnerForest` - because that's the main reason for to have this structure in the first place (and it would be much more efficient than trying to get them from the database).

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,69 @@ message CurrentBlockchainData {
...
+// Request for vault asset witnesses for a specific account.
+message VaultAssetWitnessesRequest {
+    // The account ID for which to retrieve vault asset witnesses.
+    account.AccountId account_id = 1;
+
+    // Set of asset vault keys to retrieve witnesses for.
+    repeated primitives.Digest vault_keys = 2;
+
+    // Optional block number. If not provided, uses the latest state.
+    optional fixed32 block_num = 3;
```
bobbinth, line 374: I would maybe expand the comment here it make it clear that:

- The witness will be returned for the account state at the specified block number.
- The block number is expected to be relatively close to the chain tip - i.e., within 50 blocks, I believe. Otherwise, and error will be returned.

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,69 @@ message CurrentBlockchainData {
...
+// Response containing vault asset witnesses.
+message VaultAssetWitnessesResponse {
+    // A vault asset witness containing the asset and its proof.
+    message VaultAssetWitness {
+        // The SMT opening proof for the asset's inclusion in the vault.
+        primitives.SmtOpening proof = 1;
+    }
+
+    // Block number at which the witness was generated.
+    fixed32 block_num = 1;
```
bobbinth: typo: "witness was" -> "witnesses were".

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,69 @@ message CurrentBlockchainData {
...
+// Request for a storage map witness for a specific account and storage entry.
+message StorageMapWitnessRequest {
+    // The account ID for which to retrieve the storage map witness.
+    account.AccountId account_id = 1;
+
+    // The storage map key for which to retrieve the witness.
+    primitives.Digest map_key = 2;
+
+    // Optional block number. If not provided, uses the latest state.
+    optional fixed32 block_num = 3;
```
bobbinth, line 413: Similar comment as the one above regarding expanding the comment with more details.

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,69 @@ message CurrentBlockchainData {
...
+
+    // The storage slot name for the map.
+    string slot_name = 4;
```
bobbinth, line 416: Not for this PR: in the future, we should replace slot name with slot ID here and in other places in the API. Let's create an issue for this (unless we have one already).

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,69 @@ message CurrentBlockchainData {
...
+}
+
+// Response containing a storage map witness.
+message StorageMapWitnessResponse {
+    // Storage map witness data.
+    message StorageWitness {
+        // The storage map key.
+        primitives.Digest key = 1;
```
bobbinth: I would expand the comment to make it clear that this is the "raw key" - i.e., the user-provided key.

crates/store/src/inner_forest/mod.rs
```diff
@@ -143,6 +158,39 @@ impl InnerForest {
             .map_or_else(Self::empty_smt_root, |(_, root)| *root)
     }
 
+    pub(crate) fn get_storage_map_witness(
+        &self,
+        account_id: AccountId,
+        slot_name: &StorageSlotName,
+        block_num: BlockNumber,
+        key: Word,
+    ) -> Result<StorageMapWitness, WitnessError> {
```
bobbinth, line 189: Could we add doc comments here? In the comments, we should make sure to mention that the `key` is the raw user-provided key (take a look at the `DataStore` trait docs for the same method to see how it is phrased there).

crates/store/src/inner_forest/mod.rs
```diff
@@ -143,6 +158,39 @@ impl InnerForest {
...
+
+        Ok(StorageMapWitness::new(proof, vec![key])?)
+    }
+
+    pub fn get_vault_asset_witnesses(
+        &self,
+        account_id: AccountId,
+        block_num: BlockNumber,
+        asset_keys: BTreeSet<AssetVaultKey>,
+    ) -> Result<Vec<AssetWitness>, WitnessError> {
```
bobbinth, line 206: Similar to the above comment - let's add doc comments to this method.

crates/store/src/inner_forest/mod.rs
```diff
@@ -143,6 +158,39 @@ impl InnerForest {
...
+        let root = self
+            .vault_roots
+            .get(&(account_id, block_num))
+            .ok_or(WitnessError::RootNotFound)?;
```
bobbinth: Here, we'll need to do something similar to what `get_storage_root()` does (but with one difference). Specifically:

- We want to find the root that is the latest root before the specified `block_num` (inclusive).
- But if there is no such root, return `RootNotFound` error (as we do now).

crates/store/src/inner_forest/mod.rs
```diff
@@ -143,6 +158,39 @@ impl InnerForest {
...
+        let witnessees = asset_keys
+            .into_iter()
+            .map(|key| {
+                let proof = self.forest.open(*root, key.into())?;
+                Ok(AssetWitness::new_unchecked(proof))
```
bobbinth: This is probably fine, but it may be better to use `AssetWitness::new()` here to catch a potential asset error here.

crates/store/src/state/mod.rs
```diff
@@ -1233,4 +1234,35 @@ impl State {
...
+    }
+
+    /// Returns a storage map witness for the specified account and storage entry.
+    pub async fn get_storage_map_witness(
+        &self,
+        account_id: AccountId,
+        slot_name: &StorageSlotName,
+        block_num: BlockNumber,
+        key: Word,
+    ) -> Result<StorageMapWitness, WitnessError> {
```
bobbinth: Similar to some other comments: let's make sure to clarify that the `key` is the raw user-provided key.

crates/store/src/inner_forest/mod.rs
```diff
@@ -143,6 +159,39 @@ impl InnerForest {
...
     }
 
+    pub(crate) fn get_storage_map_witness(
+        &self,
+        account_id: AccountId,
+        slot_name: &StorageSlotName,
+        block_num: BlockNumber,
+        key: Word,
+    ) -> Result<StorageMapWitness, WitnessError> {
+        let root = self.get_storage_root(account_id, slot_name, block_num);
```
bobbinth: We can't use `get_storage_root()` here directly because in case an entry is not found, it'll return an empty root. Instead, if the entry is not found, we should return an error (similar to how we do this for asset vault).

We should still keep the logic of getting the latest root before the `block_num` (inclusive) though.

proto/proto/internal/store.proto
sergerad: >The block number is expected to be relatively close to the chain tip - i.e., within 50 blocks, I believe. Otherwise, and error will be returned.

I don't see why that would be the case. Perhaps `vault_root` but I don't see any eviction logic.

```rust
    pub fn get_vault_asset_witnesses(
        &self,
        account_id: AccountId,
        block_num: BlockNumber,
        asset_keys: BTreeSet<AssetVaultKey>,
    ) -> Result<Vec<AssetWitness>, WitnessError> {
        let root = self
            .vault_roots
            .get(&(account_id, block_num))
            .ok_or(WitnessError::RootNotFound)?;
```

bobbinth: Yeah, the eviction logic is not there yet, but it is something we'll need to do relatively shortly - otherwise, the node will slow down quite a bit. I believe @drahnr is working on this.

proto/proto/internal/store.proto
sergerad: #1556

crates/store/src/inner_forest/mod.rs
sergerad: >We want to find the root that is the latest root before the specified block_num (inclusive).

I understand `get_storage_root()` to do the same with `next_back()`. Am I mistaken there?

bobbinth: Yes, correct. But unlike `get_storage_root()` we shouldn't return a root of an empty tree on "not found".

bobbinth: Or maybe a better option is to modify `get_storage_root()` to return an `Option` and handle it appropriately at call sites.

sergerad: yea good idea. pushed

crates/store/src/inner_forest/mod.rs
```diff
@@ -133,14 +151,61 @@ impl InnerForest {
...
+    pub fn get_vault_asset_witnesses(
+        &self,
+        account_id: AccountId,
+        block_num: BlockNumber,
+        asset_keys: BTreeSet<AssetVaultKey>,
+    ) -> Result<Vec<AssetWitness>, WitnessError> {
+        let (_, root) = self
+            .vault_roots
+            .range((account_id, BlockNumber::GENESIS)..=(account_id, block_num))
+            .next_back()
```
bobbinth: Could we move this out into `get_vault_root()` method? (to keep it parallel with `get_storage_root()` method).

crates/store/src/inner_forest/mod.rs
bobbinth: Not from this PR, but line 144 is incorrect.

crates/store/src/inner_forest/mod.rs
```diff
@@ -133,14 +151,61 @@ impl InnerForest {
...
             .range(
                 (account_id, slot_name.clone(), BlockNumber::GENESIS)
                     ..=(account_id, slot_name.clone(), block_num),
             )
             .next_back()
-            .map_or_else(Self::empty_smt_root, |(_, root)| *root)
+            .map(|(_, root)| *root)
+    }
+
+    /// Retrieves a storage map witness for the specified account and storage entry.
```
bobbinth: nit: "storage entry" -> "storage slot".

crates/store/src/inner_forest/mod.rs
```diff
@@ -133,14 +151,61 @@ impl InnerForest {
...
+    ///
+    /// Finds the most recent storage root entry before the specified block.
```
bobbinth: It should say "... at or before the specified block."

crates/store/src/inner_forest/mod.rs
```diff
@@ -133,14 +151,61 @@ impl InnerForest {
...
+    ///
+    /// Note that the `raw_key` is the raw, user-provided key that needs to be hashed in order to
+    /// get the actual key into the storage map.
+    pub(crate) fn get_storage_map_witness(
```
bobbinth, line 183: This actually makes me realize that we need to hash the key before doing the lookup (see how this is done in `StorageMap`).

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,76 @@ message CurrentBlockchainData {
...
+    // The account ID for which to retrieve vault asset witnesses.
+    account.AccountId account_id = 1;
+
+    // Set of asset vault keys to retrieve witnesses for.
+    repeated primitives.Digest vault_keys = 2;
+
+    // Optional block number. If not provided, uses the latest state.
+    //
+    // The specified block number should be relatively near the chain tip else an error will be returned.
+    optional fixed32 block_num = 3;
```
bobbinth: nit: comment wrapping.

Also, could we add the same line as we have for storage map witness (i.e., line 404).

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,76 @@ message CurrentBlockchainData {
...
+    account.AccountId account_id = 1;
+
+    // The storage map key for which to retrieve the witness.
+    primitives.Digest map_key = 2;
+
+    // Optional block number. If not provided, uses the latest state.
+    //
+    // The witness returned corresponds to the account state at the specified block number.
+    // The specified block number should be relatively near the chain tip else an error will be returned.
+    optional fixed32 block_num = 3;
```
bobbinth, line 413: nit: comment wrapping.

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,76 @@ message CurrentBlockchainData {
...
+// GET STORAGE MAP WITNESS
+// ================================================================================================
+
+// Request for a storage map witness for a specific account and storage entry.
+message StorageMapWitnessRequest {
+    // The account ID for which to retrieve the storage map witness.
+    account.AccountId account_id = 1;
+
+    // The storage map key for which to retrieve the witness.
+    primitives.Digest map_key = 2;
```
bobbinth: Let's make it clear in the comment that this is a raw, user-defined key.

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,79 @@ message CurrentBlockchainData {
...
+    account.AccountId account_id = 1;
+
+    // Set of asset vault keys to retrieve witnesses for.
+    repeated primitives.Digest vault_keys = 2;
+
+    // Optional block number. If not provided, uses the latest state.
+    //
+    // The specified block number should be relatively near the chain tip else an error will be
+    // returned.
+    optional fixed32 block_num = 3;
```
bobbinth, line 374: nit: could we add something like line 405 in this comment too?

crates/store/src/inner_forest/mod.rs
```diff
@@ -123,24 +142,80 @@ impl InnerForest {
             .map_or_else(Self::empty_smt_root, |(_, root)| *root)
     }
 
-    /// Retrieves the vault SMT root for an account at or before the given block.
+    /// Retrieves a vault root witness for the specified account block number.
+    ///
+    /// Finds the most recent vault root entry before the specified block number for the account.
+    pub(crate) fn get_vault_root(
```
bobbinth: typos:

- "vault root witness" -> "vault root".
- "vault root entry" -> "vault root".

crates/store/src/inner_forest/mod.rs
```diff
@@ -123,24 +142,80 @@ impl InnerForest {
...
+        let root = self
+            .get_storage_root(account_id, slot_name, block_num)
+            .ok_or(WitnessError::RootNotFound)?;
+        let proof = self.forest.open(root, key)?;
+
+        Ok(StorageMapWitness::new(proof, vec![raw_key])?)
+    }
+
+    /// Retrieves a vault asset witness for the specified account and asset keys at the specified
+    /// block number.
```
bobbinth: typo: "a vault asset witness" -> "vault asset witnesses" (since we are returning a vector).

crates/store/src/inner_forest/mod.rs
```diff
@@ -123,24 +142,80 @@ impl InnerForest {
...
+    ///
+    /// Finds the latest root before the specified block number (inclusive).
+    pub fn get_vault_asset_witnesses(
```
bobbinth: I'm not sure we need to mention "root" in the comment here.

crates/store/src/inner_forest/mod.rs
```diff
@@ -123,24 +142,80 @@ impl InnerForest {
...
+            .next_back()
+            .map(|(_, root)| *root)
+    }
+
     /// Retrieves the storage map SMT root for an account slot at or before the given block.
     ///
-    /// Finds the most recent storage root entry for the slot, since storage state persists
-    /// across blocks where no changes occur.
+    /// Finds the most recent storage root at or before the specified block number.
     pub(crate) fn get_storage_root(
```
bobbinth: typo: "storage root" -> "storage map root".

Also, not from this PR, but let's rename `get_storage_root()` to `get_storage_map_root()`

crates/store/src/inner_forest/mod.rs
```diff
@@ -157,12 +232,7 @@ impl InnerForest {
         block_num: BlockNumber,
         keys: &[Word],
     ) -> Option<Result<AccountStorageMapDetails, MerkleError>> {
```
bobbinth, line 232: Not from this PR, but this method also has a bug because we are not hashing the keys. Let's fix this in a follow-up PR.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - but they are mostly about cleaning up doc comments.

There is one comment about fixing an existing bug - but we can do it in a follow up.

proto/proto/internal/store.proto
```diff
@@ -347,3 +353,83 @@ message CurrentBlockchainData {
...
+
+    // Optional block number. If not provided, uses the latest state.
+    //
+    // The witness returned corresponds to the account state at the specified block number.
+    //
+    // Optional block number. If not provided, uses the latest state.
+    //
+    // The specified block number should be relatively near the chain tip else an error will be
+    // returned.
+    optional fixed32 block_num = 3;
```
bobbinth, line 413: Lines 405 and 409 are duplicates of each other.

crates/proto/src/generated/store.rs
```diff
@@ -223,6 +223,94 @@ pub struct CurrentBlockchainData {
...
 }
+/// Request for vault asset witnesses for a specific account.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct VaultAssetWitnessesRequest {
+    /// The account ID for which to retrieve vault asset witnesses.
+    #[prost(message, optional, tag = "1")]
+    pub account_id: ::core::option::Option<super::account::AccountId>,
+    /// Set of asset vault keys to retrieve witnesses for.
+    #[prost(message, repeated, tag = "2")]
+    pub vault_keys: ::prost::alloc::vec::Vec<super::primitives::Digest>,
```
drahnr, line 234: do we want to impose limits on this?

sergerad: Not sure it would be necessary but @bobbinth might have thoughts

bobbinth: Since this is an internal-only endpoint, not a high priority. But also doesn't hurt to have it as a sanity-check. I'd create an issue for this but no need to tackle it now.

sergerad: https://github.com/0xMiden/miden-node/issues/1565

PR #1530, opened 2026-01-19
feat: filter excessive numbers of storage map key queries earlier

### LINKED ISSUE #1514: Limit number of `SmtProof`s in `get_account` API

drahnr:
Follow up to #1517 to filter the requests earlier.

Closes #1514

### DIFF
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -487,6 +487,8 @@ impl api_server::Api for RpcService {
         &self,
         request: Request<proto::rpc::AccountRequest>,
     ) -> Result<Response<proto::rpc::AccountResponse>, Status> {
+        use proto::rpc::account_request::account_detail_request::storage_map_detail_request::SlotData::MapKeys as ProtoMapKeys;
+
         let request = request.into_inner();
 
         debug!(target: COMPONENT, ?request);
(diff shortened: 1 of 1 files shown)


crates/rpc/src/server/api.rs
```diff
@@ -490,6 +491,15 @@ impl api_server::Api for RpcService {
 
         debug!(target: COMPONENT, ?request);
 
+        // Validate storage map key limits before forwarding to store
+        if let Some(details) = &request.details {
+            for storage_map in &details.storage_maps {
+                if let Some(proto::rpc::account_request::account_detail_request::storage_map_detail_request::SlotData::MapKeys(keys)) = &storage_map.slot_data {
```
Mirko-von-Leipzig: Nit: maybe a `ProtoMapKeys` alias?

crates/utils/src/limiter.rs
```diff
@@ -120,3 +120,13 @@ impl QueryParamLimiter for QueryParamBlockLimit {
...
+
+/// Used for the following RPC endpoints
+/// * `get_account`
+///
+/// Capped at 16 storage map keys per slot to limit the number of SMT proofs returned.
+pub struct QueryParamStorageMapKeyLimit;
+impl QueryParamLimiter for QueryParamStorageMapKeyLimit {
+    const PARAM_NAME: &str = "storage_map_key";
+    const LIMIT: usize = 16;
+}
```
bobbinth, line 132: This is fine for now, but let's create an issue to impose a limit on the total number of storage keys in a request. The thinking is that if an account has 20 maps, we'd be allowing 20*16 keys in a single request. But for accounts with 1 map, we'd be allowing only 16 keys.

Doing this on per-map bases makes it so that we either have to have the number of keys too low for a single map, or it will be too high in the worst case.

Again, not a priority - but let's create an issue (may be a "good first issue").

PR #1531, opened 2026-01-19
chore: merge main

Mirko-von-Leipzig:
This merges the `main` into `next` which include the `v0.12.7` and `v0.12.8` changes.

This was mostly smooth sailing, except the `ntx` fixes which were ported manually since the original `v0.12` code has been refactored to the actor model. For reference these are the relevant changelog entries:

- Network transaction builder now marks notes from any error as failed ([#1508](https://github.com/0xMiden/miden-node/pull/1508)).
- Network transaction builder now adheres to note limit set by protocol ([#1508](https://github.com/0xMiden/miden-node/pull/1508)).

### DIFF
--- a/crates/store/Cargo.toml
+++ b/crates/store/Cargo.toml
@@ -19,8 +19,8 @@ anyhow                 = { workspace = true }
 deadpool               = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
 deadpool-diesel        = { features = ["sqlite"], version = "0.6" }
 deadpool-sync          = { default-features = false, features = ["tracing"], version = "0.1" }
-diesel                 = { features = ["numeric", "sqlite"], version = "2.2" }
-diesel_migrations      = { features = ["sqlite"], version = "2.2" }
+diesel                 = { features = ["numeric", "sqlite"], version = "2.3" }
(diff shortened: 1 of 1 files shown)


crates/store/Cargo.toml
```diff
@@ -18,9 +18,9 @@ workspace = true
 anyhow                 = { workspace = true }
 deadpool               = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
 deadpool-diesel        = { features = ["sqlite"], version = "0.6" }
-deadpool-sync          = { version = "0.1" }
-diesel                 = { features = ["numeric", "sqlite"], version = "2.3" }
-diesel_migrations      = { features = ["sqlite"], version = "2.3" }
+deadpool-sync          = { default-features = false, features = ["tracing"], version = "0.1" }
+diesel                 = { features = ["numeric", "sqlite"], version = "2.2" }
+diesel_migrations      = { features = ["sqlite"], version = "2.2" }
```
bobbinth: Question: why do we need to downgrade these from `2.3` to `2.2`?

Mirko-von-Leipzig: Copy paste error :) Thanks!

PR #1532, opened 2026-01-19
feat: add a sanity check for sync state of rocksdb vs sqlite, minor partitioning changes

drahnr:
From https://github.com/0xMiden/miden-node/pull/1326#discussion_r2703522248 :

> Could we move all loading-related methods into a separate file? Basically, I'd split this file into:
> 
>     src/state/mod.rs
>     src/state/loader.rs
> 
> I think overall, we need 3 loading modes:
> 
>     * Load everything from memory (when rocksdb feature is disabled).
>     * Use persistent data from RockDB (when rocksdb feature is enabled). This would probably require passing filepaths to RocksDB databases on startup (or at least assuming some defaults).
>     * Rebuild RocksDB data from the database (when rocksdb feature is enabled and we are starting a new node, or the user sets some flag that indicates that RocksDB databases are to be re-built).
> 
> I think currently we have 1 and 2, and maybe some parts of 3.

## Changes

* Adds a `--force-rebuild` flag to the CLI to force re-building of the tree stores from sqlite
* Error out if we detect a mismatch of the tree stores and sqlite

### DIFF
--- a/crates/store/src/state/mod.rs
+++ b/crates/store/src/state/mod.rs
@@ -1087,6 +1087,25 @@ impl State {
             None => AccountVaultDetails::empty(),
         };
 
+        // Check total keys limit upfront before expensive open operations
+        let total_keys: usize = storage_requests
+            .iter()
+            .filter_map(|req| match &req.slot_data {
+                SlotData::MapKeys(keys) => Some(keys.len()),
+                SlotData::All => None,
+            })
+            .sum();
+
+        if total_keys > AccountStorageMapDetails::MAX_SMT_PROOF_ENTRIES {
+            return Ok(AccountDetails::with_storage_limits_exceeded(
+                account_header,
+                account_code,
+                vault_details,
+                storage_header,
+                storage_requests.into_iter().map(|req| req.slot_name),
+            ));
+        }
+
         let mut storage_map_details =
             Vec::<AccountStorageMapDetails>::with_capacity(storage_requests.len());
 
--- a/crates/proto/src/domain/account.rs
+++ b/crates/proto/src/domain/account.rs
@@ -1,7 +1,7 @@
 use std::fmt::{Debug, Display, Formatter};
 
 use miden_node_utils::formatting::format_opt;
-use miden_node_utils::limiter::{QueryParamLimiter, QueryParamStorageMapKeyLimit};
+use miden_node_utils::limiter::{QueryParamLimiter, QueryParamStorageMapKeyTotalLimit};
 use miden_protocol::Word;
 use miden_protocol::account::{
     Account,
@@ -441,9 +441,9 @@ impl AccountStorageMapDetails {
     /// This limit is more restrictive than [`Self::MAX_RETURN_ENTRIES`] because SMT proofs
     /// are larger (up to 64 inner nodes each) and more CPU-intensive to generate.
     ///
-    /// This is defined by [`QueryParamStorageMapKeyLimit::LIMIT`] and used both in RPC validation
-    /// and store-level enforcement to ensure consistent limits.
-    pub const MAX_SMT_PROOF_ENTRIES: usize = QueryParamStorageMapKeyLimit::LIMIT;
+    /// This is defined by [`QueryParamStorageMapKeyTotalLimit::LIMIT`] and used both in RPC
+    /// validation and store-level enforcement to ensure consistent limits.
+    pub const MAX_SMT_PROOF_ENTRIES: usize = QueryParamStorageMapKeyTotalLimit::LIMIT;
 
     /// Creates storage map details with all entries from the storage map.
     ///
@@ -496,6 +496,14 @@ impl AccountStorageMapDetails {
                 slot_name,
                 entries: StorageMapEntries::EntriesWithProofs(proofs),
             }
+        }
+    }
+
+    /// Creates storage map details indicating the limit was exceeded.
+    pub fn limit_exceeded(slot_name: StorageSlotName) -> Self {
+        Self {
+            slot_name,
+            entries: StorageMapEntries::LimitExceeded,
         }
     }
 }
@@ -638,6 +646,21 @@ pub struct AccountStorageDetails {
     pub map_details: Vec<AccountStorageMapDetails>,
 }
 
+impl AccountStorageDetails {
+    /// Creates storage details where all map slots indicate limit exceeded.
+    pub fn all_limits_exceeded(
+        header: AccountStorageHeader,
+        slot_names: impl IntoIterator<Item = StorageSlotName>,
+    ) -> Self {
+        Self {
+            header,
+            map_details: Vec::from_iter(
+                slot_names.into_iter().map(AccountStorageMapDetails::limit_exceeded),
+            ),
+        }
+    }
+}
+
 impl TryFrom<proto::rpc::AccountStorageDetails> for AccountStorageDetails {
     type Error = ConversionError;
 
@@ -731,6 +754,24 @@ pub struct AccountDetails {
     pub account_code: Option<Vec<u8>>,
     pub vault_details: AccountVaultDetails,
     pub storage_details: AccountStorageDetails,
+}
+
+impl AccountDetails {
+    /// Creates account details where all storage map slots indicate limit exceeded.
+    pub fn with_storage_limits_exceeded(
+        account_header: AccountHeader,
+        account_code: Option<Vec<u8>>,
+        vault_details: AccountVaultDetails,
+        storage_header: AccountStorageHeader,
+        slot_names: impl IntoIterator<Item = StorageSlotName>,
+    ) -> Self {
+        Self {
+            account_header,
+            account_code,
+            vault_details,
+            storage_details: AccountStorageDetails::all_limits_exceeded(storage_header, slot_names),
+        }
+    }
 }
 
 impl TryFrom<proto::rpc::account_response::AccountDetails> for AccountDetails {
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -15,7 +15,7 @@ use miden_node_utils::limiter::{
     QueryParamNoteIdLimit,
     QueryParamNoteTagLimit,
     QueryParamNullifierLimit,
-    QueryParamStorageMapKeyLimit,
+    QueryParamStorageMapKeyTotalLimit,
 };
 use miden_protocol::batch::ProvenBatch;
 use miden_protocol::block::{BlockHeader, BlockNumber};
@@ -487,19 +487,27 @@ impl api_server::Api for RpcService {
         &self,
         request: Request<proto::rpc::AccountRequest>,
     ) -> Result<Response<proto::rpc::AccountResponse>, Status> {
-        use proto::rpc::account_request::account_detail_request::storage_map_detail_request::SlotData::MapKeys as ProtoMapKeys;
+        use proto::rpc::account_request::account_detail_request::storage_map_detail_request::{
+            SlotData::MapKeys as ProtoMapKeys,
+            SlotData::AllEntries as ProtoMapAllEntries
+        };
 
         let request = request.into_inner();
 
         debug!(target: COMPONENT, ?request);
 
-        // Validate storage map key limits before forwarding to store
+        // Validate total storage map key limit before forwarding to store
         if let Some(details) = &request.details {
-            for storage_map in &details.storage_maps {
-                if let Some(ProtoMapKeys(keys)) = &storage_map.slot_data {
-                    check::<QueryParamStorageMapKeyLimit>(keys.map_keys.len())?;
-                }
-            }
+            let total_keys: usize = details
+                .storage_maps
+                .iter()
+                .filter_map(|m| m.slot_data.as_ref())
+                .filter_map(|d| match d {
+                    ProtoMapKeys(keys) => Some(keys.map_keys.len()),
+                    ProtoMapAllEntries(_) => None,
+                })
+                .sum();
+            check::<QueryParamStorageMapKeyTotalLimit>(total_keys)?;
         }
 
         self.store.clone().get_account(request).await
@@ -629,7 +637,7 @@ static RPC_LIMITS: LazyLock<proto::rpc::RpcLimits> = LazyLock::new(|| {
         QueryParamNoteIdLimit as NoteId,
         QueryParamNoteTagLimit as NoteTag,
         QueryParamNullifierLimit as Nullifier,
-        QueryParamStorageMapKeyLimit as StorageMapKey,
+        QueryParamStorageMapKeyTotalLimit as StorageMapKeyTotal,
     };
 
     proto::rpc::RpcLimits {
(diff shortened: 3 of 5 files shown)


crates/store/src/state/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+///
+/// # Arguments
+/// * `account_tree_root` - Root of the loaded account tree
+/// * `nullifier_tree_root` - Root of the loaded nullifier tree
+/// * `db` - Database connection to fetch the latest block header
+///
+/// # Errors
+/// Returns `StateInitializationError::TreeStorageDiverged` if any root doesn't match.
+#[instrument(target = COMPONENT, skip_all)]
+pub async fn verify_tree_consistency(
```
drahnr, line 262: Adds an additional verification step 👇

bin/node/src/commands/store.rs
```diff
@@ -93,6 +93,14 @@ pub enum StoreCommand {
...
         )]
         grpc_timeout: Duration,
+
+        /// Force rebuild persistent tree storage from the database on startup.
+        ///
+        /// Use this flag to recover from corrupted tree storage or when the persistent storage
+        /// has diverged from the database. This will delete existing tree storage and rebuild
+        /// it from scratch, which may take some time for large databases.
+        #[arg(long = "force-rebuild-tree-storage", default_value_t = false)]
+        force_rebuild_tree_storage: bool,
```
drahnr: Relevant code to review 👆

drahnr:
Left markers for most relevant code changes

crates/store/src/state/mod.rs
```diff
@@ -83,141 +86,14 @@ pub struct TransactionInputs {
...
-
 /// Container for state that needs to be updated atomically.
 struct InnerState<S>
 where
     S: SmtStorage,
 {
-    nullifier_tree: NullifierTree<LargeSmt<S>>,
+    nullifier_tree: miden_protocol::block::nullifier_tree::NullifierTree<
+        miden_protocol::crypto::merkle::smt::LargeSmt<S>,
+    >,
```
bobbinth, line 96: nit: why do we need to use full paths here?

bin/node/src/commands/bundled.rs
```diff
@@ -215,6 +215,7 @@ impl BundledCommand {
                     ntx_builder_listener: store_ntx_builder_listener,
                     data_directory: data_directory_clone,
                     grpc_timeout,
+                    force_rebuild_tree_storage: false,
```
bobbinth: Shouldn't this also be set to a user-provided parameter?

crates/store/src/state/mod.rs
```diff
@@ -72,6 +65,16 @@ use crate::errors::{
...
 use crate::{COMPONENT, DataDirectory};
 
+mod loader;
+
+pub use loader::{
+    ACCOUNT_TREE_STORAGE_DIR,
+    NULLIFIER_TREE_STORAGE_DIR,
+    StorageLoader,
+    TreeStorage,
+};
```
bobbinth, line 75: Do these all need to be public?

crates/store/src/state/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+    let public_account_ids = db.select_all_public_account_ids().await?;
+
+    // Acquire write lock once for the entire initialization
+    let mut forest = InnerForest::new();
+
+    // Process each account
+    for account_id in public_account_ids {
+        // Get the full account from the database
+        let account_info = db.select_account(account_id).await?;
+        let account = account_info.details.expect("public accounts always have details in DB");
```
bobbinth, line 232: Not related to this PR (and I think we may have an issue for this already), but let's add a `TODO` here to indicate that loading full account from the database is pretty inefficient and will need to go away soon.

crates/store/src/state/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+        db: &mut Db,
+    ) -> impl std::future::Future<Output = Result<LargeSmt<Self>, StateInitializationError>> + Send;
+
+    /// Loads a nullifier tree, either from persistent storage or by rebuilding from DB.
+    fn load_nullifier_tree(
+        self,
+        db: &mut Db,
+    ) -> impl std::future::Future<
+        Output = Result<NullifierTree<LargeSmt<Self>>, StateInitializationError>,
+    > + Send;
```
bobbinth, line 95: nit: I'd just import `std::future::Future` at the top of the file.

Also, why does `load_account_tree()` return `LargeSmt`, but `load_nullifier_tree()` returns `NullifierTree`? Wouldn't it be more consistent if `load_account_tree()` would return `AccountTree`?

crates/store/src/state/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+        if has_data {
+            let smt = load_smt(self)?;
+            return Ok(NullifierTree::new_unchecked(smt));
+        }
+
+        info!(target: COMPONENT, "RocksDB nullifier tree storage is empty, populating from SQLite");
+        let nullifiers = db.select_all_nullifiers().await?;
+        let entries = nullifiers.into_iter().map(|info| (info.nullifier, info.block_num));
+        NullifierTree::with_storage_from_entries(self, entries)
+            .map_err(StateInitializationError::FailedToCreateNullifierTree)
```
bobbinth, line 182: Not for this PR, but would be great to make the methodology of loading account and nullifier trees consistent. Currently, for nullifier tree we use `NullifierTree::with_storage_from_entries()` and for account tree we use `AccountTree::new()`.

Some of these changes may need to happen in `miden-base`. Let's create issues for these (unless we have the already).

crates/store/src/state/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+        // If RocksDB storage has data, load from it directly
+        let has_data = self
+            .has_leaves()
+            .map_err(|e| StateInitializationError::AccountTreeIoError(e.to_string()))?;
+        if has_data {
+            return load_smt(self);
+        }
+
+        info!(target: COMPONENT, "RocksDB account tree storage is empty, populating from SQLite");
+        let account_data = db.select_all_account_commitments().await?;
```
bobbinth, line 157: This is fine for now, but it will break in the longer run because it requires us to load all the data into memory. Instead, we should try to do it using iterators somehow so that we read data in chunks from the database and write it in chunks into `LargeSmt`.

Let's create an issue for this (unless we have one already).

crates/store/src/state/mod.rs
bobbinth, line 213: Not related to this PR, but I'd move this method into a separate file - e.g., `src/state/apply_block.rs`. It could still be a method on this struct though (i.e., no need to change this into a free-floating function).

bobbinth (approved):
Looks good! Thank you! I left some comments inline - many of them are for the future.

crates/store/src/state/loader.rs
drahnr: We do have #1538

crates/store/src/state/loader.rs
drahnr: We do have #1543

crates/store/src/state/loader.rs
drahnr: My thinking is that `with_storage_from_entries` won't work anyways for larger nullifier sets, and hence this will go away in the next PR #1536

bin/node/src/commands/store.rs
Mirko-von-Leipzig: I'm not convinced we want this to be an explicit option. We can also trigger this by manually deleting the database which should be sufficient so long as things are sort of working okay..

From a deployment perspective, this is problematic because we will either have it permanently `true` or `false`. Doing this once-off would require a manual intervention in any case.

bobbinth: Good point! I think manually removing the files is fine. We should probably just document this (as behavior may not be obvious).

drahnr: I think the option is preferred, since it doesn't require having knowledge what all of this means. Deleting a specific subset of files seems to be an unnecessary footgun for an operator.

crates/store/src/errors.rs
```diff
@@ -206,6 +206,22 @@ pub enum StateInitializationError {
     DatabaseLoadError(#[from] DatabaseSetupError),
     #[error("inner forest error")]
     InnerForestError(#[from] InnerForestError),
+    #[error(
+        "{tree_name} root from persistent storage ({persistent_root:?}) does not match expected \
+         root from block {block_num} in database ({database_root:?}). This indicates data \
+         corruption or an incomplete previous shutdown. Consider rebuilding the persistent \
+         tree storage from the database."
```
Mirko-von-Leipzig: Maybe

```suggestion
        "SMT {tree_name} root {persistent_root} does not match {database_root} from block {block_num}'s header. This indicates data \
         corruption or an incomplete previous shutdown. Consider rebuilding the persistent \
         tree storage from the database."
```

crates/store/src/errors.rs
```diff
@@ -206,6 +206,22 @@ pub enum StateInitializationError {
...
+    )]
+    TreeStorageDiverged {
+        tree_name: &'static str,
+        block_num: BlockNumber,
+        persistent_root: Word,
+        database_root: Word,
```
Mirko-von-Leipzig: I assume this is checked against the block's header (will revisit), but I think persistent and database is confusing. They're _both_ from a database.

```suggestion
        tree_name: &'static str,
        tree_root: Word,
        block_num: BlockNumber,
        block_root: Word,
```

bin/node/src/commands/store.rs
Mirko-von-Leipzig: But there is no operator that will be able to use it since you cannot "deploy" with this option? Its only useable in a local context imo.

We similarly don't have an option to rebuild the sqlite database for example. Though technically its possible from raw blocks.

imo the fewer options there are, the better. The rarer the occurrence, the more "manual" it should be. If this becomes a common problem, then sure we can address it differently.

drahnr: I agree, that it should be done as a subcommand, the current form 'd require a separate deployment cycle, which makes it less useful.

crates/store/src/state/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+            persistent_root: nullifier_tree_root,
+            database_root: expected_nullifier_root,
+        });
+    }
+
+    tracing::info!(
+        target: COMPONENT,
+        %block_num,
+        "Tree roots verified: database and tree storage are consistent"
+    );
```
Mirko-von-Leipzig: This isn't necessary since this is already tracked by the span itself not erroring. In general, there should be very few reasons to emit "logs" unless there is an actual event that occurred within the span e.g. you want to notify that transaction `T` was committed as part of a block in `apply_block`.

```suggestion
```

crates/store/src/state/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+#[cfg(feature = "rocksdb")]
+pub fn load_smt<S: SmtStorage>(storage: S) -> Result<LargeSmt<S>, StateInitializationError> {
+    LargeSmt::new(storage).map_err(account_tree_large_smt_error_to_init_error)
+}
+
+// TREE LOADING FUNCTIONS
+// ================================================================================================
+
+/// Loads the blockchain MMR from all block headers in the database.
+#[instrument(level = "info", target = COMPONENT, skip_all)]
```
Mirko-von-Leipzig:
```suggestion
#[instrument(target = COMPONENT, skip_all)]
```

crates/store/src/state/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+
+        // Use the unified update method (will recognize it's a full-state delta)
+        forest.update_account(block_num, &delta)?;
+
+        tracing::debug!(
+            target: COMPONENT,
+            %account_id,
+            %block_num,
+            "Initialized forest for account from DB"
+        );
```
Mirko-von-Leipzig:
```suggestion
```

crates/store/src/state/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+    let latest_header = db.select_block_header_by_block_num(None).await?;
+
+    // If there's no block header (empty database), trees should be empty - no verification needed
+    let Some(header) = latest_header else {
+        return Ok(());
+    };
+
+    let block_num = header.block_num();
+    let expected_account_root = header.account_root();
+    let expected_nullifier_root = header.nullifier_root();
```
Mirko-von-Leipzig: Should probably check that the given roots are `ZERO`/`EMPTY_WORD` or whatever the default root value is. 

```suggestion
    let (block_num, expected_account_root, expected_nullifier_root) = latest_header
        .map(|header| (
            header.block_num(),
            header.account_root(),
            header.nullifier_root(),
        ))
        .unwrap_or_default();
```

crates/store/src/errors.rs
```diff
@@ -206,6 +206,21 @@ pub enum StateInitializationError {
     DatabaseLoadError(#[from] DatabaseSetupError),
     #[error("inner forest error")]
     InnerForestError(#[from] InnerForestError),
+    #[error(
+        "SMT {tree_name} root {persistent_root} does not match {database_root} from block {block_num}'s header. This indicates data \
+         corruption or an incomplete previous shutdown. Consider rebuilding the persistent \
+         tree storage from the database."
```
Mirko-von-Leipzig: Sorry I kinda screwed up your alignment here.

`{tree_name}` also doesn't match the usage anymore :D 

Probably `Account SMT root...` makes more sense. But this is a minor nit.

crates/store/src/state/mod.rs
Mirko-von-Leipzig, line 1: Currently we return an error if we have a root mismatch during `apply_block`. https://github.com/0xMiden/miden-node/blob/a2de1546a965ce36fd652d648a8352d067d370c9/crates/store/src/state.rs#L423-L425

Should we have a similar plan if the corruption happens "live"? Its a bit of a weird situation since we can't know if its an SMT bug, or a block building bug.

drahnr: I think we'd just bubble the error up and find the corrupted tree store on startup, and do it that way.

Mirko-von-Leipzig: Right but this doesn't actually bubble up to anywhere. This is part of the gRPC server on the store side, so it gets sent back to the block-producer which _shrugs_, rolls the block back and tries again.

crates/store/src/errors.rs
drahnr: Done

crates/store/src/state/mod.rs
drahnr: I'll create a follow-up issue, maybe it's time for `fatality` to disambiguate which ones are relevant for the query response and which ones are an _actual_ issue for the service running.

crates/block-producer/src/server/tests.rs
```diff
@@ -94,6 +94,7 @@ async fn block_producer_startup_is_robust_to_network_failures() {
                 ntx_builder_listener,
                 block_producer_listener,
                 data_directory: dir,
+                grpc_timeout: std::time::Duration::from_secs(30),
```
drahnr, line 97: How did this compile previously??

crates/store/src/state/loader.rs
bobbinth: It seems like the second part of this comment wasn't addressed. Let's make sure to address it in the follow-up.

crates/store/src/state/mod.rs
bobbinth, line 96: nit (but let's address in a follow-up): we don't need to use fully-qualified paths here.

Issue #1534, opened 2026-01-19
feat: enforce shutdown sequencing of components


drahnr:
We start to have more and more IO backed datastructures that are kept "alive" by free tasks. Depending on shutdown ordering, artifacts like `pthread invalid arg` and friends show up in CI, and during restart of components in production. We want to avoid this as much as possible and enforce ordering of shutdown tasks.

Commonly this is done via `CancellationToken`s (see https://tokio.rs/tokio/topics/shutdown )

Mirko-von-Leipzig:
Something to also consider is using a more structured runtime aka _not tokio_. The benefit of `tokio` is that we don't have to think very hard about how to allocate resources e.g. thread pools, and they all share. However this is also a downside, as well as the lack of structure concurrency.

This is probably not worth the effort though given the wide `tokio` ecosystem support.

Mirko-von-Leipzig:
Completed in #2306

PR #1536, opened 2026-01-19
feat/db: paged loading of nullifiers and account commitments

drahnr:
Addresses performance and memory constraints in https://github.com/0xMiden/miden-node/issues/1534 when loading `AccountTree` and `NullifierTree`.

## Changes

* Introduces `fn select_account_commitments_paged`
* Introduce `fn select_all_nullifiers_paged`

## Open Questions

* Shall we remove the non-paged versions? My intuition is _yes_
* Shall we rename the `_paged` ones to suffix less? Also _yes_


crates/store/src/errors.rs
```diff
@@ -313,6 +313,16 @@ pub enum ApplyBlockError {
     DbUpdateTaskFailed(String),
 }
 
+impl From<ApplyBlockError> for Status {
+    fn from(err: ApplyBlockError) -> Self {
+        match err {
+            ApplyBlockError::InvalidBlockError(_) => Status::invalid_argument(err.to_string()),
+
+            _ => Status::internal(err.to_string()),
```
Mirko-von-Leipzig: You probably want to `.as_report()` these?

crates/store/src/state/loader.rs
Mirko-von-Leipzig, line 1: Could you instrument all the loading functions? So we can tell what's slow on start up.

crates/store/src/state/loader.rs
```diff
@@ -107,23 +127,77 @@ impl StorageLoader for MemoryStorage {
...
+        // Load account commitments in pages to avoid loading millions of entries at once
+        let mut cursor = None;
+        loop {
+            let page = db
+                .select_account_commitments_paged(ACCOUNT_COMMITMENTS_PAGE_SIZE, cursor)
+                .await?;
+
+            if page.commitments.is_empty() {
+                break;
+            }
```
Mirko-von-Leipzig, line 145: Nit: is this strictly necessary? Applying an empty set if probably close to a no-op I imagine, and we'll exit because `cursor.is_none()`.

Mostly just a style thing but I'd prefer one exit if possible. Similar below.

crates/store/src/db/models/queries/nullifiers.rs
```diff
@@ -137,6 +139,67 @@ pub(crate) fn select_all_nullifiers(
...
+
+    let mut query =
+        SelectDsl::select(schema::nullifiers::table, NullifierWithoutPrefixRawRow::as_select())
+            .order_by(schema::nullifiers::nullifier.asc())
+            .limit(limit)
+            .into_boxed();
+
+    if let Some(cursor) = after_nullifier {
+        query = query.filter(schema::nullifiers::nullifier.gt(cursor.to_bytes()));
+    }
```
Mirko-von-Leipzig, line 187: Nit: probably quicker to just `after_nullifier.unwrap_or_default()` since it will cache the statement on the next call 😁

Mirko-von-Leipzig: Ah nvm I see you're using `after` and not `next` as the token.

bobbinth:
There seem to be some small merge conflicts. @drahnr - could you resolve them when you get a chance?

crates/store/src/db/models/queries/nullifiers.rs
drahnr: > Ah nvm I see you're using after and not next as the token.

I don't think there is a difference?

> after_nullifier.unwrap_or_default()

This only works for an asc sorting and the implicit all-zeros nullifier. I don't like the implicitness.

drahnr: Revisiting, there is a small delta in what the cursor _means_, and I think we should unify eventually.

PR #1551, opened 2026-01-20
fix/query: `GetAccount` should limit on totals, not per keyset

drahnr:
Follow up to https://github.com/0xMiden/miden-node/pull/1530 to limit the _total_ number of key queries.

### DIFF
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -487,7 +487,10 @@ impl api_server::Api for RpcService {
         &self,
         request: Request<proto::rpc::AccountRequest>,
     ) -> Result<Response<proto::rpc::AccountResponse>, Status> {
-        use proto::rpc::account_request::account_detail_request::storage_map_detail_request::SlotData::MapKeys as ProtoMapKeys;
+        use proto::rpc::account_request::account_detail_request::storage_map_detail_request::{
+            SlotData::MapKeys as ProtoMapKeys,
+            SlotData::AllEntries as ProtoMapAllEntries
+        };
 
         let request = request.into_inner();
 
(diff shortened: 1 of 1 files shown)


crates/store/src/state.rs
```diff
@@ -1205,6 +1201,32 @@ impl State {
...
+                storage_details: AccountStorageDetails {
+                    header: storage_header,
+                    map_details: Vec::from_iter(storage_requests.into_iter().map(|req| {
+                        AccountStorageMapDetails {
+                            slot_name: req.slot_name,
+                            entries: StorageMapEntries::LimitExceeded,
+                        }
+                    })),
+                },
+            });
```
Mirko-von-Leipzig: This makes me feel like all of these things should be separate methods 😅

drahnr: I am torn, it spreads the logic quite a bit, but it's less "bobby".

crates/proto/src/domain/account.rs
```diff
@@ -441,9 +441,9 @@ impl AccountStorageMapDetails {
     /// This limit is more restrictive than [`Self::MAX_RETURN_ENTRIES`] because SMT proofs
     /// are larger (up to 64 inner nodes each) and more CPU-intensive to generate.
     ///
-    /// This is defined by [`QueryParamStorageMapKeyLimit::LIMIT`] and used both in RPC validation
-    /// and store-level enforcement to ensure consistent limits.
-    pub const MAX_SMT_PROOF_ENTRIES: usize = QueryParamStorageMapKeyLimit::LIMIT;
+    /// This is defined by [`QueryParamStorageMapKeyTotalLimit::LIMIT`] and used both in RPC
+    /// validation and store-level enforcement to ensure consistent limits.
+    pub const MAX_SMT_PROOF_ENTRIES: usize = QueryParamStorageMapKeyTotalLimit::LIMIT;
```
bobbinth, line 446: I wonder if this limit is still relevant. Basically, `QueryParamStorageMapKeyTotalLimit::LIMIT` ensures that the user doesn't request too many SMT proofs per request, and this one ensures that no specific map returns more than the same number of proofs.

The way it is set currently, this limit should never trigger (because a single map cannot have more SMT proofs than was specified for the whole request) - but also, it is not clear to me that if we check the number of keys, we still need to enforce this at all.

So, unless I'm missing something, I think we should remove `MAX_SMT_PROOF_ENTRIES`.

drahnr: https://github.com/0xMiden/miden-node/pull/1575

PR #1553, opened 2026-01-20
fix(store): connect gRPC traces to handlers

Mirko-von-Leipzig:
Our gRPC servers have a gRPC trace layer which automatically creates a span for each method that is invoked. For our internal APIs this is also linked to the clients which lets us have a single trace representing the entire flow.

This works but we broke the link between this automatic span and the actual method implementation by having the method be its own root span (`parent = None`).

This PR removes this instrumentation entirely. Keeping it is a form of stuttering since it effectively duplicates the span name without doing any work in between.

There is some missing info here, but its not critical imo.


bobbinth (approved):
Looks good! Thank you!

PR #1559, opened 2026-01-20
fix: add attachment to network monitor increase counter note

SantiagoPittella:
While testing next I noticed that after the miden-base update, the network monitor started to failed to increase the counter. The fix is to send the note attachment with the increase request.


fkrause98 (approved):

bobbinth (approved):
Looks good! Thank you!

Issue #1549, opened 2026-01-19
Refactor network note queries to use full account IDs instead of 30-bit prefixes


SantiagoPittella:
I think that would add a lot of noise to this PR, but lets definitely address it in a follow-up. No reason to store more data than needed :)

_Originally posted by @Mirko-von-Leipzig in https://github.com/0xMiden/miden-node/pull/1526#discussion_r2705584433_

PR #1526 introduced `NetworkAccountTarget` attachments that contain the full target account ID for single-target network notes. However, the database queries still use the 30-bit prefix from the `NoteTag` for filtering.

bobbinth:
The concept of `AccountPrefix` should go away after this.

PR #1563, opened 2026-01-21
chore: miden base update followups

SantiagoPittella:
In this PR I:

- Update miden-base version to `v0.13.1`.
- Use `NetworkAccountTargetError` from miden-base for attachment errors (https://github.com/0xMiden/miden-node/pull/1526#discussion_r2707085129 & https://github.com/0xMiden/miden-base/pull/2319 ).
- Remove references to network account prefixes (https://github.com/0xMiden/miden-node/pull/1526#discussion_r2710779983).
- Create the `NetworkNoteType` for database storage and remove `is_single_target_network_note`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Added `GetLimits` endpoint to the RPC server ([#1410](https://github.com/0xMiden/miden-node/pull/1410)).
 - Added gRPC-Web probe support to the `miden-network-monitor` binary ([#1484](https://github.com/0xMiden/miden-node/pull/1484)).
 - Add DB schema change check ([#1268](https://github.com/0xMiden/miden-node/pull/1485)).
+- Decoupled ntx-builder from block-producer startup by loading network accounts asynchronously via a background task ([#????](https://github.com/0xMiden/miden-node/pull/????)).
 - Add foreign account support to validator ([#1493](https://github.com/0xMiden/miden-node/pull/1493)).
 - Improve DB query performance for account queries ([#1496](https://github.com/0xMiden/miden-node/pull/1496).
 - Limit number of storage map keys in `GetAccount` requests ([#1517](https://github.com/0xMiden/miden-node/pull/1517)).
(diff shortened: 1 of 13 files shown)


crates/store/src/db/mod.rs
```diff
@@ -656,9 +656,9 @@ impl Db {
         block_num: BlockNumber,
         page: Page,
     ) -> Result<(Vec<NoteRecord>, Page)> {
```
bobbinth, line 658: This may be a bigger refactoring, and so, probably not for this PR - but we should be using account ID here instead of `network_account_prefix`. Do we have an issue to change this already? If not, let's create one.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -849,15 +868,19 @@ pub struct NoteInsertRowInsert {
     pub serial_num: Option<Vec<u8>>,
     pub nullifier: Option<Vec<u8>>,
     pub script_root: Option<Vec<u8>>,
-    pub is_single_target_network_note: bool,
+    pub network_note_type: i32,
     pub inclusion_path: Vec<u8>,
 }
 
 impl From<(NoteRecord, Option<Nullifier>)> for NoteInsertRowInsert {
```
bobbinth: Not related to this PR, but I find `NoteInsertRowInsert` name a bit confusing. Do we really need two "insert" in there? Maybe just `NoteInsertRow` would be fine?

crates/store/src/db/mod.rs
Mirko-von-Leipzig: We have #1549 which covers this

PhilippGackstatter:
> It seems like we are still relying on note tag to find notes for a network account. Technically, this shouldn't work because network notes shouldn't use tags to encode target accounts (cc @PhilippGackstatter to confirm). Instead, they should be using note attachment data which contains full ID of the target account.

Yes, we should not be relying on the note tag at all for network notes anymore and should use the `NetworkAccountTarget` attachment instead.

crates/store/src/db/models/queries/notes.rs
SantiagoPittella: No, we don't need this redundancy.  Just `NoteInsertRow` is fine. Is a really minor change, I will include it here

Issue #1538, opened 2026-01-19
perf: avoid loading and storing full `Account`s from and to DB


drahnr:
Current we construct a full `struct Account` with all details, which queries multiple tables, while we only update a few elements, and hence don't need to fetch the full account. This make 250ms of the full query.

<img width="745" height="443" alt="Image" src="https://github.com/user-attachments/assets/c0859234-e6e4-4893-b879-26dd146aeb87" />

bobbinth:
Copying from https://github.com/0xMiden/miden-node/issues/1483#issuecomment-3713653494:

As mentioned in https://github.com/0xMiden/miden-node/pull/1481#discussion_r2664067395, we should also move away from the general strategy of reading in the full account from the database, applying delta to it, and then saving it back.

PR #1567, opened 2026-01-22
perf/store: optimize updating existing accounts 

drahnr:
## Context 

Currently applying a partial delta to an existing account requires loading the full account state
consisting of code, all storage map entries, all vault assets, which is very costly - we spend ~250ms on this per block (that's a lot!).

See https://github.com/0xMiden/miden-node/issues/1538

## What

Optimizes account delta updates by avoiding loading full `Account` objects from the database when only applying partial deltas.

### Core changes

- module `crates/store/src/db/models/queries/accounts/delta.rs`
  - `fn select_account_state_for_delta()` - fetches only nonce, code_commitment, storage_header, vault_root
  - `fn select_vault_balances_by_faucet_ids()` - fetches only specific vault balances being updated
  - `fn apply_storage_delta_with_precomputed_roots()` - updates the storage header using precomputed map roots

- `InnerForest::apply_block_updates` now precomputes vault and storage map roots for each account update and passes them to
  `Db::apply_block` for partial delta updates.

- `fn upsert_accounts` uses the optimized path for partial deltas via `enum AccountStateForInsert` with three variants:
  - `::Private` - no public state
  - `::FullAccount` - new account creation == "full-state delta"
  - `::PartialState` - incremental update == "partial delta"

- `AccountRowInsert` constructors: `new_private()`, `new_from_account()`, `new_from_partial()`

### Biggest Change

We  reverse the flow and update - we update the `InnerForest` now and track new roots, then using these in the DB storage header update. The downside is additional tackling.

### DIFF
--- a/.github/actions/debian/action.yml
+++ b/.github/actions/debian/action.yml
@@ -121,8 +121,8 @@ runs:
         Maintainer: Miden <devops@miden.team>
         Description: $pkg binary package
         Homepage: https://miden.xyz
-        Vcs-Git: git@github.com:0xMiden/miden-node.git
-        Vcs-Browser: https://github.com/0xMiden/miden-node
+        Vcs-Git: git@github.com:0xMiden/node.git
+        Vcs-Browser: https://github.com/0xMiden/node
         EOF
 
     - name: Build binaries
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,28 +4,28 @@
 
 ### Enhancements
 
-- [BREAKING] Move block proving from Blocker Producer to the Store ([#1579](https://github.com/0xMiden/miden-node/pull/1579)).
-- [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/miden-node/pull/1595)).
-- Validator now persists validated transactions ([#1614](https://github.com/0xMiden/miden-node/pull/1614)).
-- [BREAKING] Remove `SynState` and introduce `SyncChainMmr` ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
-- Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
-- Fixed `TransactionHeader` serialization for row insertion on database & fixed transaction cursor on retrievals ([#1701](https://github.com/0xMiden/miden-node/issues/1701)).
-- Added KMS signing support in validator ([#1677](https://github.com/0xMiden/miden-node/pull/1677)).
+- [BREAKING] Move block proving from Blocker Producer to the Store ([#1579](https://github.com/0xMiden/node/pull/1579)).
+- [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/node/pull/1595)).
+- Validator now persists validated transactions ([#1614](https://github.com/0xMiden/node/pull/1614)).
+- [BREAKING] Remove `SynState` and introduce `SyncChainMmr` ([#1591](https://github.com/0xMiden/node/issues/1591)).
+- Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1591](https://github.com/0xMiden/node/issues/1591)).
+- Fixed `TransactionHeader` serialization for row insertion on database & fixed transaction cursor on retrievals ([#1701](https://github.com/0xMiden/node/issues/1701)).
+- Added KMS signing support in validator ([#1677](https://github.com/0xMiden/node/pull/1677)).
 
 ### Changes
 
-- [BREAKING] Removed obsolete `SyncState` RPC endpoint; clients should use `SyncNotes`, `SyncNullifiers`, `SyncAccountVault`, `SyncAccountStorageMaps`, `SyncTransactions`, or `SyncChainMmr` instead ([#1636](https://github.com/0xMiden/miden-node/pull/1636)).
-- Added account ID limits for `SyncTransactions`, `SyncAccountVault`, and `SyncAccountStorageMaps` to `GetLimits` responses ([#1636](https://github.com/0xMiden/miden-node/pull/1636)).
-- [BREAKING] Added typed `GetAccountError` for `GetAccount` endpoint, splitting `BlockNotAvailable` into `UnknownBlock` and `BlockPruned`. `AccountNotFound` and `AccountNotPublic` now return `InvalidArgument` gRPC status instead of `NotFound`; clients should parse the error details discriminant rather than branching on status codes ([#1646](https://github.com/0xMiden/miden-node/pull/1646)).
-- Changed `note_type` field in proto `NoteMetadata` from `uint32` to a `NoteType` enum ([#1594](https://github.com/0xMiden/miden-node/pull/1594)).
-- Refactored NTX Builder startup and introduced `NtxBuilderConfig` with configurable parameters ([#1610](https://github.com/0xMiden/miden-node/pull/1610)).
-- Refactored NTX Builder actor state into `AccountDeltaTracker` and `NotePool` for clarity, and added tracing instrumentation to event broadcasting ([#1611](https://github.com/0xMiden/miden-node/pull/1611)).
-- Add #[track_caller] to tracing/logging helpers ([#1651](https://github.com/0xMiden/miden-node/pull/1651)).
-- Added support for generic account loading at genesis ([#1624](https://github.com/0xMiden/miden-node/pull/1624)).
-- Improved tracing span fields ([#1650](https://github.com/0xMiden/miden-node/pull/1650))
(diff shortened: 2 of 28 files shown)


crates/store/src/state/apply_block.rs
```diff
@@ -221,14 +221,25 @@ impl State {
...
             ));
 
+        // Compute roots in InnerForest BEFORE spawning DB task, so DB can use precomputed roots.
+        // This avoids loading all entries from disk to recompute roots during DB writes.
+        let precomputed_roots = self
+            .forest
+            .write()
+            .await
+            .apply_block_updates(block_num, account_deltas.clone())?;
+
```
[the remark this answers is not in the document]
drahnr: Either of them failing will be inconsistent, we rebuild on inconsistency on restart.

crates/store/src/inner_forest/mod.rs
```diff
@@ -433,19 +424,24 @@ impl InnerForest {
...
+        Ok(new_root)
     }
 
     // STORAGE MAP DELTA PROCESSING
     // --------------------------------------------------------------------------------------------
 
-    /// Retrieves the most recent storage map SMT root for an account slot. If no storage root is
-    /// found for the slot, returns an empty SMT root.
+    /// Retrieves the most recent storage map SMT root for an account slot.
+    /// If `is_full_state` is true, returns an empty SMT root.
```
Mirko-von-Leipzig: This feels dodgy. Shouldn't this already be the case though? A full account implies a new account and it should therefore already return `Empty`?

Or is this an optimisation?

crates/store/src/state/mod.rs
```diff
@@ -182,6 +182,9 @@ impl State {
         })
     }
 
+    // STATE MUTATOR
+    // --------------------------------------------------------------------------------------------
```
SantiagoPittella: This looks misplaced.

crates/store/src/state/apply_block.rs
```diff
@@ -284,8 +273,6 @@ impl State {
         .in_current_span()
         .await?;
 
-        self.forest.write().await.apply_block_updates(block_num, account_deltas)?;
```
SantiagoPittella, line 288: Should this be re-added? I couldn't find were are we applying block updates

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -1061,11 +1109,73 @@ pub(crate) fn upsert_accounts(
...
+                )?;
+
+                let new_vault_root = {
+                    let (_last_block, assets) = select_account_vault_assets(
+                        conn,
+                        account_id,
+                        BlockNumber::GENESIS..=block_num,
+                    )?;
+                    let assets: Vec<Asset> =
+                        assets.into_iter().filter_map(|entry| entry.asset).collect();
```
SantiagoPittella: This query can end up being quite expensive, right? We are searching in all blocks basically.

The InnerForest already maintains vault SMT roots in memory and updates them with `batch_insert`, it can compute the new vault root without any DB query.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -1061,11 +1109,73 @@ pub(crate) fn upsert_accounts(
...
+                    for (key, value) in map_delta.entries() {
+                        storage.push((account_id, slot_name.clone(), (*key).into(), *value));
+                    }
+                }
+
+                let mut map_entries = BTreeMap::new();
+                for (slot_name, map_delta) in delta.storage().maps() {
+                    if map_delta.is_empty() {
+                        continue;
+                    }
```
SantiagoPittella: Probably not a big deal but this two could be merged, right?

crates/store/src/state/mod.rs
drahnr: It is :) removed

crates/store/src/db/models/queries/accounts.rs
drahnr: That changes is part of improving #1608 which will be merged back, I can use a raw SQL query instead to improve it and push the logic to the query engine instead.

crates/store/src/db/models/queries/accounts.rs
drahnr: Theoretically yes, but it makes it harder to review. Unless it becomes a practical perf limitation, I'd rather keep it as is.

crates/store/src/state/apply_block.rs
drahnr: This got inlined, I'll undo that.

drahnr: Undone

crates/store/src/inner_forest/mod.rs
```diff
@@ -286,96 +286,50 @@ impl InnerForest {
...
     // --------------------------------------------------------------------------------------------
 
-    /// Retrieves the most recent vault SMT root for an account. If no vault root is found for the
-    /// account, returns an empty SMT root.
-    fn get_latest_vault_root(&self, account_id: AccountId) -> Word {
+    /// Retrieves the most recent vault SMT root for an account.
+    /// If `is_full_state` is true, returns an empty SMT root.
+    fn get_latest_vault_root(&self, account_id: AccountId, is_full_state: bool) -> Word {
+        if is_full_state {
+            return Self::empty_smt_root();
```
drahnr: Always insert an empty root if it's a full account, a full state implies the previous state is all empty.

crates/store/src/inner_forest/mod.rs
drahnr, line 463: this is overly complicated but already changed in #1635  so not changing here

Mirko-von-Leipzig (approved):

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -755,51 +794,95 @@ pub(crate) fn select_latest_account_storage(
...
+        None => AccountStorageHeader::new(Vec::new())?,
     };
 
-    // Deserialize the AccountStorageHeader from the blob
-    let header = AccountStorageHeader::read_from_bytes(&blob)?;
+    let entries = select_latest_storage_map_entries_all(conn, &account_id)?;
+    Ok((header, entries))
+}
+
+// TODO this is expensive and should only be called from tests
```
sergerad, line 806: Do we want to complete this TODO in this PR?

crates/store/src/inner_forest/mod.rs
```diff
@@ -353,12 +353,78 @@ impl InnerForest {
...
+                target: crate::COMPONENT,
+                %account_id,
+                %block_num,
+                ?slot_name,
+                delta_entries = num_entries,
+                "Inserted storage map into forest"
+            );
+        }
+        Ok(())
+    }
```
sergerad, line 412: There are a lot of diffs here and its hard to get my head around functions like this which have no comments in the body. Could you please break this function into logical sections and describe what is going on in each section with a comment above? 🙏

sergerad: (I'm assuming this is a new addition rather than a piece of code moved from elsewhere)

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -940,8 +1023,171 @@ pub(crate) fn insert_account_storage_map_value(
...
+    let new_nonce_value = state_headers
+        .nonce
+        .as_int()
+        .checked_add(delta.nonce_delta().as_int())
+        .ok_or_else(|| {
+            DatabaseError::DataCorrupted(format!("Nonce overflow for account {account_id}"))
+        })?;
+    let new_nonce = Felt::new(new_nonce_value);
+
+    // Create minimal account state data for the row insert
```
sergerad: nit: Some comments start capitalized while others dont. I think we generally capitalize and full stop at end

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -940,8 +1023,171 @@ pub(crate) fn insert_account_storage_map_value(
...
+        };
+        if should_insert {
+            assets.push((account_id, asset.vault_key(), Some(asset)));
+        }
+    }
+
+    Ok((AccountStateForInsert::FullAccount(account), storage, assets))
+}
+
+fn prepare_partial_account_update(
```
sergerad, line 1070: Comment would be good

crates/store/src/db/models/queries/accounts/delta.rs
```diff
@@ -0,0 +1,243 @@
...
+
+    Ok(AccountStateHeadersForDelta { nonce, code_commitment, storage_header })
+}
+
+/// Selects vault balances for specific faucet IDs.
+///
+/// Optimized query that only fetches balances for the faucet IDs
+/// that are being updated by a delta, rather than loading all vault assets.
+///
+/// Returns a map from `faucet_id` to the current balance (0 if not found).
```
sergerad, line 140: Is there a use case where the 0 is unacceptably equivocal? i.e. when there is genuinely 0 balance?

drahnr: Since we cannot have negative balances in accounts, this would fail, for positive ones it'd initialize with that value. As of today, I don't see issues?

crates/store/src/db/models/queries/accounts.rs
drahnr: Definitely not :)

It's used in `State::load` to initialize the `SmtForest` where we still do load the full account from DB, which is the only call-site for this one.

crates/store/src/inner_forest/mod.rs
drahnr: Added some, a lot of code got re-partitioned due to the increased complexity, and git doesn't show all diffs correctly.

PR #1568, opened 2026-01-22
fix(ntx): mempool subscription has a race condition

### LINKED ISSUE #1566: Subscribing to mempool events is flakey
In #1495 we decoupled the ntx and block-producer start up processes. This is now possible because the ntx now streams accounts from the store instead of preloading them all.

This meant we could remove the `Barrier` which ensured the two components started in lock-step from the same chain tip. Now that we've removed this, the ntx's subscription to the mempool can fail if the block-producer has already created another block(s) after the ntx fetches the chain tip, but before it subscribes. Previously this was prevented because blocks were only built _after_ the subscription was created due to the barrier.

We have two options here (afaict):

1. Loop `fetch chain tip` + `subscribe` until it succeeds.
    - Though this has poor semantics since it could fail for other reasons.
2. Remove `chain tip` from the endpoint.
    - This means the ntx doesn't actually _know_ what the mempool's chain tip is until the first block committed event comes through.
    - So we could have start-up consist of buffering events until the first block comes in and only then starting the main event handler loop etc.

On first blush I'm partial to (2) because that feels less dependent on trial and error.

Here is an example error for reference:

```
2026-01-22T06:49:03.610906Z ERROR ntx.block_producer.client.subscribe_to_mempool: miden-ntx-builder: crates/ntx-builder/src/block_producer.rs:61: error: status: 'Client specified an invalid argument', self: "Mempool's chain tip 7 does not match request's 6", metadata: {"content-type": "application/grpc", "date": "Thu, 22 Jan 2026 06:49:03 GMT"}
2026-01-22T06:49:03.610934Z  INFO ntx.block_producer.client.subscribe_to_mempool: miden-ntx-builder: crates/ntx-builder/src/block_producer.rs:61: close, time.busy: 380µs, time.idle: 501ms

Caused by:
    0: failed while serving ntx builder component
    1: failed to subscribe to mempool events
    2: status: 'Client specified an invalid argument', self: "Mempool's chain tip 7 does not match request's 6", metadata: {"content-type": "application/grpc", "date": "Thu, 22 Jan 2026 06:49:03 GMT"}
```

Mirko-von-Leipzig:
This is a quick but hacky solution for #1566.

aka this is solution (1), but more ideal would be (2). (2) involves more of larger refactor so I'm punting on that for now.

Closes #1566.

### DIFF
--- a/crates/ntx-builder/src/builder.rs
+++ b/crates/ntx-builder/src/builder.rs
@@ -112,14 +112,31 @@ impl NetworkTransactionBuilder {
         let store = StoreClient::new(self.store_url.clone());
         let block_producer = BlockProducerClient::new(self.block_producer_url.clone());
 
-        let (chain_tip_header, chain_mmr) = store
-            .get_latest_blockchain_data_with_retry()
-            .await?
-            .expect("store should contain a latest block");
-        let mut mempool_events = block_producer
-            .subscribe_to_mempool_with_retry(chain_tip_header.block_num())
-            .await
-            .context("failed to subscribe to mempool events")?;
+        // Loop until we successfully subscribe.
+        //
+        // The mempool rejects our subscription if we don't have the same view of the chain aka
+        // if our chain tip does not match the mempools. This can occur if a new block is committed
+        // _after_ we fetch the chain tip from the store but _before_ our subscription request is
+        // handled.
+        //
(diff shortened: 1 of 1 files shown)


SantiagoPittella:
Do we want to include the option 2 in the 0.13 release? I can start to work on it and maybe be ready for a patch release or something

PR #1569, opened 2026-01-22
feat: add termination on root mismatch in store

### LINKED ISSUE #1328: `AccountTree`: Establish re-synchronization between sqlite and rocksdb storage
Once we land the `AccountTreeWithHistory<AccountTree<RocksDb>>` implementation, we will have two distinct places for fetching persisted accounts data - the `LargeSmt<_>` and the `SQLite` database. Now whenever we get into a situation where we have a discrepancy between we need a path to resolve.

### Proposal

* drop the `AccountTreeWithHistory` and re-build from the SQL database recent history
* do something smarter™

drahnr:
When the store finds itself in a situation where the nullifier tree update doesn't align with the nullifier root in the block_header

Review notes:

* uses an `mpsc` so we don't have to `Option` wrap, we can have multiple concurrent requests that trigger the code-path, having channel capacity small means the first to hit the path will block all following ones and trigger termination
* `State::load` is _not_ called via RPC, so the error _does_ terminate correctly

Closes #1328

### DIFF
--- a/crates/store/src/state/mod.rs
+++ b/crates/store/src/state/mod.rs
@@ -319,7 +319,7 @@ impl State {
             if nullifier_tree_update.as_mutation_set().root() != header.nullifier_root() {
                 // We do our best here to notify the serve routine, if it doesn't care (dropped the
                 // receiver) we can't do much.
-                let _ = self.termination_ask.send(ApplyBlockError::InvalidBlockError(
+                let _ = self.termination_ask.try_send(ApplyBlockError::InvalidBlockError(
                     InvalidBlockError::NewBlockInvalidNullifierRoot,
                 ));
                 return Err(InvalidBlockError::NewBlockInvalidNullifierRoot.into());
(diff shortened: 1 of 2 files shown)


crates/store/src/state/mod.rs
```diff
@@ -304,6 +317,11 @@ impl State {
                 .map_err(InvalidBlockError::NewBlockNullifierAlreadySpent)?;
 
             if nullifier_tree_update.as_mutation_set().root() != header.nullifier_root() {
+                // We do our best here to notify the serve routine, it it doesn't care (dropped the
```
Mirko-von-Leipzig:
```suggestion
                // We do our best here to notify the serve routine, if it doesn't care (dropped the
```

Mirko-von-Leipzig (approved):
Only question for me is whether this is the correct sync primitive, but its probably just semantics tbh. e.g. https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html.

PR #1578, opened 2026-01-22
fix: add missing is_latest filter to network account query

SantiagoPittella:
Fixes ntx-builder crash on node restart after a network transaction has been processed. After a network transaction updates a network account, restarting the node causes the ntx-builder to fail with:

```
ERROR account actor already exists for account: 0x...
Caused by: event channel closed
```

The `select_all_network_account_ids` query was missing an `is_latest = true` filter. Without filtering for `is_latest`, both historical and current rows were returned, causing duplicate account IDs to be sent to the ntx-builder, which then attempted to spawn multiple actors for the same account.

Added `.filter(schema::accounts::is_latest.eq(true))` to ensure only the current state of each account is returned. I saw that this filter was removed in https://github.com/0xMiden/miden-node/pull/1481 so I want to confirm with @drahnr if we should fix it in another way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,7 @@
 - Fixed `GetNetworkAccountIds` pagination to return the chain tip ([#1489](https://github.com/0xMiden/miden-node/pull/1489)).
 - Fixed the network monitor counter account to use the storage slot name ([#1501](https://github.com/0xMiden/miden-node/pull/1501)).
 - gRPC traces now correctly connect to the method implementation ([1553](https://github.com/0xMiden/miden-node/pull/1553)).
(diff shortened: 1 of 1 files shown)


crates/store/src/db/models/queries/accounts.rs
```diff
@@ -536,7 +536,8 @@ pub(crate) fn select_all_network_account_ids(
     let account_ids_raw: Vec<(Vec<u8>, i64)> = Box::new(
         QueryDsl::select(
             schema::accounts::table
-                .filter(schema::accounts::network_account_id_prefix.is_not_null()),
+                .filter(schema::accounts::network_account_id_prefix.is_not_null())
```
bobbinth, line 539: Not for this PR, but we should probably replace `network_account_id_prefix` with something like `network_account_type`.

PR #1579, opened 2026-01-22
feat: Move block proving to the Store

sergerad:
## Context

The state transitions of blocks have changed such that signed (instead of proven) blocks are applied to Store state, leaving block proving to be done asynchronously some time after. The block proving logic is being moved out of the Block Producer so the the Store can managed asynchronous proving after signed blocks are committed to its state.

NOTE: Block proofs are not used in the PR. Follow-up work will be done to make proving deferred/asynchronous to the rest of the Store's `apply_block()` flow.

Relates to #1316.

## Changes
- Remove block prover client and block proving from the block producer.
- Add block prover client and block proving to the Store.
- Update the Store's `ApplyBlock` RPC endpoint to take data required to perform proving.
- Add `signature` column to `block_header` table.

### DIFF
--- a/crates/store/src/server/block_producer.rs
+++ b/crates/store/src/server/block_producer.rs
@@ -102,7 +102,7 @@ impl block_producer_server::BlockProducer for StoreApi {
                 // SAFETY: The header, body, and signature are assumed to
                 // correspond to each other because they are provided by the Block
                 // Producer.
-                let signed_block = SignedBlock::new_unchecked(header.clone(), body, signature);
+                let signed_block = SignedBlock::new_unchecked(header.clone(), body, signature); // TODO(sergerad): Use `SignedBlock::new()` when available.
                 // Note: This is an internal endpoint, so its safe to expose the full error
                 // report.
                 this.state
--- a/crates/proto/src/generated/store.rs
+++ b/crates/proto/src/generated/store.rs
@@ -7,7 +7,7 @@ pub struct ApplyBlockRequest {
     #[prost(bytes = "vec", tag = "1")]
     pub ordered_batches: ::prost::alloc::vec::Vec<u8>,
     /// Block signed by the Validator.
-    #[prost(message, optional, tag = "3")]
+    #[prost(message, optional, tag = "2")]
     pub block: ::core::option::Option<super::blockchain::SignedBlock>,
 }
 /// Returns data required to prove the next block.


proto/proto/types/blockchain.proto
```diff
@@ -7,13 +7,6 @@ import "types/primitives.proto";
 // BLOCK
 // ================================================================================================
 
-// Represents a block.
-message Block {
-    // Block data encoded using [winter_utils::Serializable] implementation for
-    // [miden_protocol::block::Block].
-    bytes block = 1;
-}
```
bobbinth, line 15: I wonder if we should keep this message and define it as:
```proto
message Block {
    BlockHeader header = 1;
    BlockBody body = 2;
    BlockSignature signature = 3;
    optional BlockProof proof = 4;
}
```

Then, we could use this in both `ApplyBlockRequest` and in `MaybeBlock`.

proto/proto/internal/store.proto
```diff
@@ -112,6 +112,28 @@ service BlockProducer {
...
+    bytes block_inputs = 2;
+    // Block header encoded using [winter_utils::Serializable] implementation for
+    // [miden_objects::block::BlockHeader].
+    bytes header = 3;
+    // Block header encoded using [winter_utils::Serializable] implementation for
+    // [miden_objects::block::BlockBody].
+    bytes body = 4;
+    // Signature encoded using [winter_utils::Serializable] implementation for
+    // [crypto::dsa::ecdsa_k256_keccak::Signature].
+    bytes signature = 5;
```
bobbinth: Related to the previous comment, I'd roll these up into a single message (whether `Block` or something else) and use the `BlockHeader`, `BlockBody`, `BlockSignature` messages.

proto/proto/internal/store.proto
```diff
@@ -112,6 +112,28 @@ service BlockProducer {
...
+// ================================================================================================
+
+// Applies a block to the state.
+message ApplyBlockRequest {
+    // Ordered batches encoded using [winter_utils::Serializable] implementation for
+    // [miden_objects::batch::OrderedBatches].
+    bytes ordered_batches = 1;
+    // Block inputs encoded using [winter_utils::Serializable] implementation for
+    // [miden_objects::block::BlockInputs].
+    bytes block_inputs = 2;
```
bobbinth: Technically, the store has all the required info to generate block inputs from ordered batches (I think). So, sending block inputs is somewhat redundant. At the same time, generating block inputs takes non-trivial amount of time, and probably, there is no good reason to have the store do this twice. So, maybe sending block inputs here is fine.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -140,12 +124,8 @@ impl BlockBuilder {
                 ProposedBlock::inject_telemetry(proposed_block);
             })
             .and_then(|(proposed_block, inputs)| self.validate_block(proposed_block, inputs))
-            .and_then(|(proposed_block, inputs, header, signature, body)| self.prove_block(proposed_block, inputs, header, signature, body))
-            .inspect_ok(ProvenBlock::inject_telemetry)
-            // Failure must be injected before the final pipeline stage i.e. before commit is called. The system cannot
-            // handle errors after it considers the process complete (which makes sense).
-            .and_then(|proven_block| async { self.inject_failure(proven_block) })
-            .and_then(|proven_block| self.commit_block(mempool, proven_block))
+            // TODO(sergerad): Add SignedBlock to miden-base and update validate_block to return it.
```
bobbinth: I'd maybe do this even before merging this PR - but not a strong opinion.

crates/block-producer/src/store/mod.rs
```diff
@@ -238,8 +240,21 @@ impl StoreClient {
...
+        body: &BlockBody,
+        signature: &Signature,
+    ) -> Result<(), StoreError> {
+        let request = tonic::Request::new(proto::store::ApplyBlockRequest {
+            ordered_batches: ordered_batches.to_bytes(),
+            block_inputs: block_inputs.to_bytes(),
+            header: header.to_bytes(),
+            body: body.to_bytes(),
+            signature: signature.to_bytes(),
+        });
```
bobbinth, line 250: Would be nice if this was more like:
```rust
pub async fn apply_block(
    &self,
    block: &SignedBlock,
    ordered_batches: &OrderedBatches,
    block_inputs: &BlockInputs,
) -> Result<(), StoreError> {
    let request = tonic::Request::new(proto::store::ApplyBlockRequest {
        ordered_batches: ordered_batches.to_bytes(),
        block_inputs: block_inputs.to_bytes(),
        block: block.into(),
    });

    ...
}
```

crates/store/src/server/block_prover_client.rs
```diff
@@ -0,0 +1,57 @@
+use miden_block_prover::{BlockProverError, LocalBlockProver};
+use miden_protocol::MIN_PROOF_SECURITY_LEVEL;
+use miden_protocol::batch::OrderedBatches;
+use miden_protocol::block::{BlockHeader, BlockInputs, BlockProof};
+use miden_remote_prover_client::RemoteProverClientError;
+use miden_remote_prover_client::remote_prover::block_prover::RemoteBlockProver;
```
bobbinth, line 5: nit (and not for this PR): the namespacing here feels a bit "too nested" - e.g., could this not be just `miden_remote_prover_client::block_prover::RemoteBlockProver`?

crates/store/src/state/mod.rs
```diff
@@ -219,13 +220,16 @@ impl State {
     // TODO: This span is logged in a root span, we should connect it to the parent span.
     #[allow(clippy::too_many_lines)]
     #[instrument(target = COMPONENT, skip_all, err)]
-    pub async fn apply_block(&self, block: ProvenBlock) -> Result<(), ApplyBlockError> {
+    pub async fn apply_block(
+        &self,
+        header: BlockHeader,
+        body: BlockBody,
+        signature: Signature,
+    ) -> Result<(), ApplyBlockError> {
```
bobbinth: `apply_block()` is a pretty huge method. I would maybe, as a precursor to this PR, open a small PR that just moves this method to a separate file (without adding any new code). The file could be `src/state/apply_block.rs` and in that file, we'd still have it as a method - i.e.,:

```rust
impl State {
    pub async fn apply_block(
        &self,
        header: BlockHeader,
        body: BlockBody,
        signature: Signature,
    ) -> Result<(), ApplyBlockError> {
        ...
    }
}
```

crates/store/src/server/block_prover_client.rs
sergerad: Yea because we are doing `pub mod block_prover;` instead of `pub use block_prover::BlockProver` in `miden_remote_prover_client`. I can make a PR for that after this one.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -278,53 +248,25 @@ impl BlockBuilder {
...
-                MIN_PROOF_SECURITY_LEVEL,
-            ));
-        }
-        // TODO(sergerad): Consider removing this validation. Once block proving is implemented,
-        // this would be replaced with verifying the proof returned from the prover against
-        // the block header.
-        validate_tx_headers(&proven_block, &ordered_batches.to_transactions())?;
-
-        Ok(proven_block)
+        let signed_block = SignedBlock::new_unchecked(header, body, signature);
```
Mirko-von-Leipzig, line 253: Could you add a `// SAFETY` comment

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -140,12 +116,7 @@ impl BlockBuilder {
                 ProposedBlock::inject_telemetry(proposed_block);
             })
             .and_then(|(proposed_block, inputs)| self.validate_block(proposed_block, inputs))
-            .and_then(|(proposed_block, inputs, header, signature, body)| self.prove_block(proposed_block, inputs, header, signature, body))
-            .inspect_ok(ProvenBlock::inject_telemetry)
-            // Failure must be injected before the final pipeline stage i.e. before commit is called. The system cannot
-            // handle errors after it considers the process complete (which makes sense).
-            .and_then(|proven_block| async { self.inject_failure(proven_block) })
-            .and_then(|proven_block| self.commit_block(mempool, proven_block))
+            .and_then(|(ordered_batches, block_inputs, signed_block)| self.commit_block(mempool, ordered_batches, block_inputs, signed_block))
```
Mirko-von-Leipzig: Is there any telemetry from the signing that might be useful?

crates/store/src/server/block_prover_client.rs
```diff
@@ -0,0 +1,57 @@
...
+///
+/// The local proving variant is intended for development and testing purposes.
+/// The remote proving variant is intended for production use.
+pub enum BlockProver {
+    Local(LocalBlockProver),
+    Remote(RemoteBlockProver),
+}
+
+impl BlockProver {
+    pub fn new_local(security_level: Option<u32>) -> Self {
```
Mirko-von-Leipzig: Do we actually use this input anywhere? If not I would remove it.

crates/store/src/server/block_prover_client.rs
```diff
@@ -0,0 +1,57 @@
...
+use tracing::{info, instrument};
+
+use crate::COMPONENT;
+
+#[derive(Debug, thiserror::Error)]
+pub enum StoreProverError {
+    #[error("local proving failed")]
+    LocalProvingFailed(#[from] BlockProverError),
+    #[error("remote proving failed")]
+    RemoteProvingFailed(#[from] RemoteProverClientError),
```
Mirko-von-Leipzig: This isn't an issue here since its fairly localized but we need to get into the habit of avoiding these implicit error mappings. They lead to poor code hygiene over time. We should prefer explicitly mapping `.map_err`

```suggestion
    #[error("local proving failed")]
    LocalProvingFailed(#[source] BlockProverError),
    #[error("remote proving failed")]
    RemoteProvingFailed(#[source] RemoteProverClientError),
```

proto/proto/internal/store.proto
Mirko-von-Leipzig: Current get block inputs testnet numbers are roughly 1-5ms per block but that of course doesn't mean much at the current block size. That's the time taken in the store itself, with gRPC + networking it bumps to 10-50ms.

My gut says that this would be faster to do it locally to reduce message size. But probably the easiest implementation here is to actually just cache it locally at the store. As in, an LRU which stores recent block inputs responses for re-use in apply block. And we just key it by block commitment 🤷 Right now a single block is sufficient since that's all that is ever in flight..

crates/store/src/server/mod.rs
```diff
@@ -35,6 +36,7 @@ pub struct Store {
     pub rpc_listener: TcpListener,
     pub ntx_builder_listener: TcpListener,
     pub block_producer_listener: TcpListener,
+    pub block_prover: Arc<BlockProver>,
```
Mirko-von-Leipzig: I think this should be an `Option<Url>`, or maybe better 

```rust
enum BlockProver {
   Local,
   Remote(Url),
}
```

Then we also don't need to expose the block prover client etc.

proto/proto/internal/store.proto
bobbinth: If it's indeed that cheap, I'd probably skip this parameter and just re-build them internally. As a future optimization, we can add caching which will speed things up but the interface between components will remain the same.

Or if just using `GetBlockInputs` internally causes issues (e.g., because the current implementation assumes that we always call these against the chain tip, and changing this would not be trivial), we can go to the cache approach right away.

crates/block-producer/src/block_builder/mod.rs
sergerad: Nothing as far as the contents of the `SignedBlock` is concerned (only new field is a signature). Only thing I can think of is duration/latency of signing but that should be handled well enough by the `validate_block()` instrumentation as is.

crates/store/src/server/mod.rs
sergerad: Think I prefer `Option<Url>` given we already structure the client as an enum, seems unnecessarily repetitive to add another
```rust
pub enum BlockProver {
    Local(LocalBlockProver),
    Remote(RemoteBlockProver),
}
```

sergerad: If we wanted to tighten up the `Store` construction we could add a builder. Feels overkill for an internal type though.

proto/proto/internal/store.proto
sergerad: Have gone without the cache for now. Let me know whether it is actually required for this PR

Mirko-von-Leipzig: Without is fine :) We'll pick up when this becomes a time problem :+1: Might never actually be needed.

crates/store/src/server/mod.rs
Mirko-von-Leipzig: Its mostly about legibility at the call site, since `fn(None)` doesn't really help understand what its doing. But this is probably a broader issue than just this item :) I'm fine leaving it as an option for now.

crates/store/src/server/block_prover_client.rs
```diff
@@ -0,0 +1,58 @@
...
+impl BlockProver {
+    pub fn new_local() -> Self {
+        info!(target: COMPONENT, "Using local block prover");
+        Self::Local(LocalBlockProver::new(0))
+    }
+
+    pub fn new_remote(endpoint: impl Into<String>) -> Self {
+        info!(target: COMPONENT, "Using remote block prover");
+        Self::Remote(RemoteBlockProver::new(endpoint))
+    }
```
Mirko-von-Leipzig: Nit


```suggestion
    pub fn local() -> Self {
        Self::Local(LocalBlockProver::new(0))
    }

    pub fn remote(endpoint: impl Into<String>) -> Self {
        Self::Remote(RemoteBlockProver::new(endpoint))
    }
```

proto/proto/internal/store.proto
```diff
@@ -112,6 +112,18 @@ service BlockProducer {
...
+// APPLY BLOCK REQUEST
+// ================================================================================================
+
+// Applies a block to the state.
+message ApplyBlockRequest {
+    // Ordered batches encoded using [winter_utils::Serializable] implementation for
+    // [miden_objects::batch::OrderedBatches].
+    bytes ordered_batches = 1;
+    // Block signed by the Validator.
+    blockchain.SignedBlock block = 3;
```
bobbinth: nit: should be `2` rather than `3`.

crates/block-producer/src/block_builder/mod.rs
```diff
@@ -136,16 +112,11 @@ impl BlockBuilder {
...
                 ProposedBlock::inject_telemetry(proposed_block);
             })
-            .and_then(|(proposed_block, inputs)| self.validate_block(proposed_block, inputs))
-            .and_then(|(proposed_block, inputs, header, signature, body)| self.prove_block(proposed_block, inputs, header, signature, body))
-            .inspect_ok(ProvenBlock::inject_telemetry)
-            // Failure must be injected before the final pipeline stage i.e. before commit is called. The system cannot
-            // handle errors after it considers the process complete (which makes sense).
-            .and_then(|proven_block| async { self.inject_failure(proven_block) })
-            .and_then(|proven_block| self.commit_block(mempool, proven_block))
+            .and_then(|proposed_block| self.validate_block(proposed_block))
```
bobbinth: nit (feel free to ignore): I think the name `validate_block()` is not as accurate as it can be. Basically, we are not just validating the block there but also building it. Maybe `build_and_validate_block()` would be better?

crates/store/src/state/apply_block.rs
```diff
@@ -41,13 +41,14 @@ impl State {
...
+    pub async fn apply_block(&self, signed_block: SignedBlock) -> Result<(), ApplyBlockError> {
         let _lock = self.writer.try_lock().map_err(|_| ApplyBlockError::ConcurrentWrite)?;
 
-        let header = block.header();
-
-        let tx_commitment = block.body().transactions().commitment();
+        let header = signed_block.header();
+        let body = signed_block.body();
 
+        // Validate that header and body match.
```
bobbinth, line 50: We should probably validate that the signature is valid since, theoretically, the store could receive a block with an invalid signature.

Not for this PR, but we could do this as a method on `SignedBlock` (in `miden-base`) - something like:

```rust
impl SignedBlock {
    pub fn validate_signature(&self) -> Result<(), SomeError> {
        ...
    }
}
```

We could also do this at `SignedBlock` construction time - but not sure that's much better.

crates/store/src/state/apply_block.rs
bobbinth, line 58: Related to the above comment, we could probably move this into `SignedBlock` as well - e.g., `SignedBlock::validate_transaction_commitment()`. Or this could also be done at construction time.

crates/store/src/state/apply_block.rs
```diff
@@ -78,20 +78,19 @@ impl State {
             return Err(InvalidBlockError::NewBlockInvalidPrevCommitment.into());
         }
```
bobbinth, line 79: Along the same lines as the previous two comments: this could probably be moved into something like `SignedBlock::validate_parent()`.

crates/store/src/server/block_producer.rs
```diff
@@ -40,33 +44,68 @@ impl block_producer_server::BlockProducer for StoreApi {
...
-        request: Request<proto::blockchain::Block>,
+        request: Request<proto::store::ApplyBlockRequest>,
     ) -> Result<Response<()>, Status> {
         let request = request.into_inner();
-
-        let block = ProvenBlock::read_from_bytes(&request.block).map_err(|err| {
-            Status::invalid_argument(err.as_report_context("block deserialization error"))
-        })?;
+        // Read ordered batches and block inputs.
+        let ordered_batches =
```
bobbinth: nit: is the "and block inputs" part in the comment correct here?

bobbinth:
One other thing we should do once the structure is closer to its final shape is update the docs (and related diagrams) with the up-to-date explanation of the node's architecture. @sergerad - could you create an issue for this?

crates/store/src/state/apply_block.rs
sergerad: https://github.com/0xMiden/miden-base/pull/2382

Was thinking we could put it in the non-unchecked constructor and use it a bit earlier in the flow (before apply_block() is called here).

Anything else we would want to check in that constructor (the error enum is overkill if there is nothing else). If there is a fair bit of other stuff to check (compare header/body) then might be worth adding the validate_signature() fn rather than doing unnecessary checks in this stack for example.

crates/store/src/state/apply_block.rs
sergerad: Added to constructor in base PR (above comment)

crates/store/src/state/apply_block.rs
sergerad: Added in same PR as above comments

PR #1580, opened 2026-01-22
refactor: replace network account prefix with type enum and use full account ID for lookups

SantiagoPittella:
Related to https://github.com/0xMiden/miden-node/pull/1578#discussion_r2718147363

- Replaces the `network_account_id_prefix` column with a `network_account_type` enum for identification of network accounts
- Renames `GetNetworkAccountDetailsByPrefix` endpoint to `GetNetworkAccountDetailsById`, now accepting full account ID instead of 30-bit prefix

### DIFF
--- a/crates/store/src/db/models/queries/accounts.rs
+++ b/crates/store/src/db/models/queries/accounts.rs
@@ -64,18 +64,11 @@ type StorageMapValueRow = (i64, String, Vec<u8>, Vec<u8>);
 
 /// Classifies accounts for database storage based on whether they are network accounts.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-#[repr(i32)]
 pub(crate) enum NetworkAccountType {
     /// Not a network account.
-    None = 0,
+    None,
     /// A network account.
-    Network = 1,
-}
-
-impl From<NetworkAccountType> for i32 {
-    fn from(value: NetworkAccountType) -> Self {
-        value as i32
-    }
+    Network,
 }
 
 // ACCOUNT CODE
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -59,6 +59,7 @@
 - [BREAKING] Updated to new miden-base protocol: removed `aux` and `execution_hint` from `NoteMetadata`, removed `NoteExecutionMode`, and `NoteMetadata::new()` is now infallible ([#1526](https://github.com/0xMiden/miden-node/pull/1526)).
 - [BREAKING] Network note queries now use full account ID instead of 30-bit prefix ([#1572](https://github.com/0xMiden/miden-node/pull/1572)).
 - Removed git information from node's `--version` CLI as it was often incorrect ([#1576](https://github.com/0xMiden/miden-node/pull/1576)).
+- [BREAKING] Renamed `GetNetworkAccountDetailsByPrefix` endpoint to `GetNetworkAccountDetailsById` which now accepts full account ID instead of 30-bit prefix ([#????](https://github.com/0xMiden/miden-node/pull/????)).
```
bobbinth: Let's add a PR number here.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -65,6 +59,25 @@ mod tests;
...
+    None = 0,
+    /// A network account.
+    Network = 1,
+}
+
+impl From<NetworkAccountType> for i32 {
+    fn from(value: NetworkAccountType) -> Self {
+        value as i32
+    }
+}
```
Mirko-von-Leipzig: I think we should probably use the sql conversion trait, I'll push a commit for that.

Issue #1581, opened 2026-01-22
DataStore error while re-executing network transactions


SantiagoPittella:
Seems like after https://github.com/0xMiden/miden-node/pull/1521 the NTX builder started to fail to execute the increase counter network transactions.

To reproduce I started the node as usual:

```
$ miden-node bundled bootstrap --data-directory data --accounts-directory accounts --genesis-config-file ./simple.toml

$ miden-node bundled start --rpc.url http://0.0.0.0:57291 --data-directory data --enable-otel
```

And in another terminal:

```
$ miden-network-monitor start --rpc-url http://localhost:57291 --counter-increment-interval 10s
```

Then I get this logs:

```
2026-01-22T22:45:27.434246Z  INFO ntx.state.select_candidate: miden-ntx-builder: crates/ntx-builder/src/actor/account_state.rs:104: close, time.busy: 3.75µs, time.idle: 750ns
2026-01-22T22:45:27.435248Z ERROR rpc: tower_http::trace::on_failure: /Users/spittella/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tower-http-0.6.8/src/trace/on_failure.rs:93: response failed, classification: Code: 13, latency: 0 ms rpc.service: "store.NtxBuilder", rpc.method: "GetVaultAssetWitnesses", otel.name: "store.NtxBuilder/GetVaultAssetWitnesses"
2026-01-22T22:45:27.435264Z  INFO rpc: miden_node_utils::tracing::grpc: crates/utils/src/tracing/grpc.rs:21: close, time.busy: 39.5µs, time.idle: 11.3µs rpc.service: "store.NtxBuilder", rpc.method: "GetVaultAssetWitnesses", otel.name: "store.NtxBuilder/GetVaultAssetWitnesses"
2026-01-22T22:45:27.435353Z ERROR ntx.actor.execute_transactions:ntx.execute_transaction:ntx.execute_transaction.filter_notes:store.client.get_vault_asset_witnesses: miden-ntx-builder: crates/ntx-builder/src/store.rs:388: error: gRPC client error
2026-01-22T22:45:27.435366Z  INFO ntx.actor.execute_transactions:ntx.execute_transaction:ntx.execute_transaction.filter_notes:store.client.get_vault_asset_witnesses: miden-ntx-builder: crates/ntx-builder/src/store.rs:388: close, time.busy: 35.2µs, time.idle: 255µs
[...]

PR #1584, opened 2026-01-23
fix: `InnerForest::get_vault_root` must store empty roots

drahnr:
Closes https://github.com/0xMiden/miden-node/issues/1581

Supersedes https://github.com/0xMiden/miden-node/pull/1583

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -127,3 +127,22 @@ jobs:
         run: make check
       - name: Diff check
         run: git diff --exit-code
+
+  check-features:
+    name: check all feature combinations
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Cleanup large tools for build space
+        uses: ./.github/actions/cleanup-runner
+      - name: Install RocksDB
+        uses: ./.github/actions/install-rocksdb
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
+      - name: Install rust
+        run: rustup update --no-self-update
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@cargo-hack
+      - name: Check all feature combinations
+        run: make check-features
--- a/.github/workflows/publish-dry-run.yml
+++ b/.github/workflows/publish-dry-run.yml
@@ -29,7 +29,6 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y jq
       - name: Update Rust toolchain
         run: rustup update --no-self-update
-      - uses: Swatinem/rust-cache@v2
       - uses: taiki-e/install-action@v2
         with:
           tool: cargo-binstall@1.16.6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,9 @@
 - Fix race condition at DB shutdown in tests ([#1503](https://github.com/0xMiden/miden-node/pull/1503)).
 - [BREAKING] Updated to new miden-base protocol: removed `aux` and `execution_hint` from `NoteMetadata`, removed `NoteExecutionMode`, and `NoteMetadata::new()` is now infallible ([#1526](https://github.com/0xMiden/miden-node/pull/1526)).
 - [BREAKING] Network note queries now use full account ID instead of 30-bit prefix ([#1572](https://github.com/0xMiden/miden-node/pull/1572)).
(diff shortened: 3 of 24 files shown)


crates/store/src/inner_forest/tests.rs
```diff
@@ -377,6 +377,55 @@ fn test_update_storage_map() {
...
+        recorded_root,
+        InnerForest::empty_smt_root(),
+        "empty vault should have the empty SMT root"
+    );
+
+    // Verify `get_vault_asset_witnesses` succeeds (returns empty witnesses for empty keys).
+    let witnesses = forest
+        .get_vault_asset_witnesses(account_id, block_num, std::collections::BTreeSet::new())
+        .expect("get_vault_asset_witnesses should succeed for accounts with empty vaults");
+    assert!(witnesses.is_empty());
```
Mirko-von-Leipzig, line 393: What is the expected result if keys are requested on an empty vault?

crates/store/src/inner_forest/mod.rs
Mirko-von-Leipzig, line 1: Does the storage not need similar treatment?

drahnr: I think we do, let me write a test case

crates/store/src/inner_forest/tests.rs
drahnr: I'd expect an empty list

drahnr: Going through the impl, it'll actually return a non-existence proof using the empty leaf.

crates/store/src/inner_forest/mod.rs
drahnr: I'll create a follow up PR shortly

crates/store/src/inner_forest/mod.rs
```diff
@@ -321,7 +321,9 @@ impl InnerForest {
         let account_id = delta.id();
         let is_full_state = delta.is_full_state();
 
-        if !delta.vault().is_empty() {
+        // Update for new accounts (full-state) or when there are vault changes.
+        // New accounts need their vault root recorded even if the vault is empty.
+        if is_full_state || !delta.vault().is_empty() {
             self.update_account_vault(block_num, account_id, delta.vault(), is_full_state)?;
         }
```
bobbinth: I wonder if we are trying to do too much here in a single method - i.e., maybe we should handle new vs. existing account flows separately. I'll take a stab at separating them and will push a commit soon.

crates/store/src/inner_forest/tests.rs
```diff
@@ -164,39 +164,6 @@ fn test_incremental_vault_updates() {
     assert_ne!(root_1, root_2);
 }
 
-#[test]
-fn test_full_state_delta_starts_from_empty_root() {
```
bobbinth, line 168: I removed this test because I don't think it is testing a valid condition: when we are inserting a delta for a new account (i.e., the full-state delta), there should be no data in the forest for this account.

crates/store/src/inner_forest/mod.rs
```diff
@@ -331,46 +311,105 @@ impl InnerForest {
...
+    /// vault for this account does not yet exist in the forest.
+    fn insert_account_vault(
+        &mut self,
+        block_num: BlockNumber,
+        account_id: AccountId,
+        delta: &AccountVaultDelta,
+    ) {
+        // get the current vault root for the account, and make sure it is empty
+        let prev_root = self.get_latest_vault_root(account_id);
+        assert_eq!(prev_root, Self::empty_smt_root(), "account should not be in the forest");
```
bobbinth, line 308: We could make this check stronger: make sure that there are actually no entries in `vault_roots` for this account (i.e., not just that the most recent entry is an empty vault).

crates/store/src/inner_forest/mod.rs
Mirko-von-Leipzig, line 1: I think this is an improvement, though this still feels complicated to read and audit.

We should take some time to think through how we can structure this differently at some point. Though this applies broadly to the entire `apply_block` sequence.

As an example, having to handle `full`, `full-but-empty`, `not-full`, and `empty` deltas in different branches is a pain. Branches bad for brain. As a dumb example, having a flow like:

```rust
if is_full_state {
    assert!(!self.contains_vault(account_id));
    self.insert_empty_vault(account_id);
}

let mut vault = self.get_latest_vault_mut(account_id).unwrap();

// Delta is _always_ applied regardless if its empty or not.
vault.apply_delta(delta.vault());

// Decide whether to append new root or not.
..
```

PR #1587, opened 2026-01-23
chore(validator): improve instrumentation

Mirko-von-Leipzig:
Adds more fine grained instrumentation to the validator.

The resulting code is a bit ugly but I didn't want to touch the core code flow. I'd like to try a different approach at the gRPC level but I'll leave that for another time.

### DIFF
--- a/crates/store/src/inner_forest/mod.rs
+++ b/crates/store/src/inner_forest/mod.rs
@@ -89,59 +89,6 @@ impl InnerForest {
         *EmptySubtreeRoots::entry(SMT_DEPTH, 0)
     }
 
-    /// Retrieves the most recent vault SMT root for an account.
-    ///
-    /// Returns the latest vault root entry regardless of block number.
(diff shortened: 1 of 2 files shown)


crates/validator/src/block_validation/mod.rs
```diff
@@ -31,9 +32,16 @@ pub async fn validate_block<S: BlockSigner>(
     validated_transactions: Arc<ValidatedTransactions>,
 ) -> Result<Signature, BlockValidationError> {
     // Check that all transactions in the proposed block have been validated
+    let verify_span = info_span!("verify_transactions");
     for tx_header in proposed_block.transactions() {
         let tx_id = tx_header.id();
-        if validated_transactions.get(&tx_id).await.is_none() {
+        // TODO: LruCache is a poor abstraction since it locks many times.
```
bobbinth, line 38: We could add a wrapper - something like `ScriptCache` and provide bulk operations on this struct - to lock the underlying cache only once for such operations.

Issue #1272, opened 2025-10-03
Add `transaction_id` field to `TransactionRecord`


juan518munoz:
### To whom is this feature for?

App developer

### Why is is this feature needed?

[Current proto conversion of `TransactionRecord`](https://github.com/0xMiden/miden-node/blob/f42f0dbd5ff8c5a53bb68aed2a71e3d610d36ad6/crates/store/src/db/mod.rs#L110-L111) skips adding the field `transaction_id` to the proto struct, rendering backwards conversion imposible, as a `TransactionId` cannot be derived from the other parts of the structure.

### How is this feature used?

Related to https://github.com/0xMiden/miden-client/issues/1349, which needs to parse a proto `TransactionRecord` response to implement the `SyncTransaction` enpoint.

bobbinth:
I think I'd prefer the approach where we could compute `TransactionId` from the underlying data. Is the issue that the `input_notes` field of the `TransactionHeader` does not include info needed to compute commitments to unauthenticated notes? If so, it shouldn't be too difficult to add, I think. cc @PhilippGackstatter.

igamigo:
> I think I'd prefer the approach where we could compute `TransactionId` from the underlying data. Is the issue that the `input_notes` field of the `TransactionHeader` does not include info needed to compute commitments to unauthenticated notes? If so, it shouldn't be too difficult to add, I think. cc [@PhilippGackstatter](https://github.com/PhilippGackstatter).

The constructor for `TransactionId` takes

```rust
pub fn new(
        init_account_commitment: Word,
        final_account_commitment: Word,
        input_notes_commitment: Word,
        output_notes_commitment: Word,
    ) 
```

`TransactionHeader` contains `input_notes: Vec<Nullifier>`, but AFAIK you can't derive the input notes commitment from this, because you are missing the `NoteId` for each of the notes, right?

juan518munoz:
If what @igamigo says is true, then from my understanding computing `TransactionId` from the underlying data implies that the `TransactionRecord` needs to at least add a field of input notes commitments. 

This change itself would require that we also make changes to `miden-base`, as we are serializing from a `miden_objects::transaction::TransactionHeader`, which doesn't hold the mentioned commitments.

https://github.com/0xMiden/miden-node/blob/f42f0dbd5ff8c5a53bb68aed2a71e3d610d36ad6/crates/store/src/db/models/queries/transactions.rs#L179-L187

igamigo:
[...]

PR #1593, opened 2026-01-26
chore: add transaction id to TransactionRecord

SantiagoPittella:
closes https://github.com/0xMiden/miden-node/issues/1272

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,8 @@
 
 ### Enhancements
 
-- Added `transaction_id` field to `TransactionRecord` proto message ([#1593](https://github.com/0xMiden/miden-node/pull/1593)).
-## v0.14.0 (TBC)
-
-### Enhancements
-
 - [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/miden-node/pull/1595)).
+- Added `transaction_id` field to `TransactionRecord` proto message ([#1593](https://github.com/0xMiden/miden-node/pull/1593)).
 
 ## v0.13.0 (2026-01-23)
 


proto/proto/rpc.proto
```diff
@@ -650,6 +650,9 @@ message TransactionRecord {
 
     // A transaction header.
     transaction.TransactionHeader header = 2;
+
+    // The unique identifier of the transaction.
+    transaction.TransactionId transaction_id = 3;
```
bobbinth, line 655: Do we actually need to add this filed? I think `TransactionId` can be computed from the `TransactionHeader` struct (at least that's the case in `miden-base` - not sure if `TransactionHeader` message is the same ).

Mirko-von-Leipzig: We can't convert from gRPC to `miden-protocol` for `TransactionHeader` because we're missing `fee` and the `input_notes` (we only have nullifiers).

```rust
    // miden-protocol
    pub fn new(
        account_id: AccountId,
        initial_state_commitment: Word,
        final_state_commitment: Word,
        input_notes: InputNotes<InputNoteCommitment>,
        output_notes: Vec<NoteHeader>,
        fee: FungibleAsset,
    ) -> TransactionHeader {
```

```protobuf
// miden-node-proto
message TransactionHeader {
    // ID of the account against which the transaction was executed.

    // State commitment of the account before the transaction was executed.
    primitives.Digest initial_state_commitment = 2;

    // State commitment of the account after the transaction was executed.
    primitives.Digest final_state_commitment = 3;

    // Nullifiers of the input notes of the transaction.
    repeated primitives.Digest nullifiers = 4;

    // Output notes of the transaction.
    repeated note.NoteSyncRecord output_notes = 5;
}
```

but maybe we can compute `TransactionId` directly already..

```rust
        let id = TransactionId::new(
            initial_state_commitment,
            final_state_commitment,
            input_notes_commitment,
            output_notes_commitment,
        );
```

where we need to compute the input and output note commitment?

Mirko-von-Leipzig: A bit of a pain trying to find where note commitments are calculated because they aren't unique types.

Mirko-von-Leipzig: As far as I can tell there is no way to build the input note commitment from just nullifiers but I can't be sure because a lot of it is garbled internally

bobbinth: Yes, we need full `InputNoteCommitment` (which is nullifier + optional `NoteHeader`) - so, just nullifier is not enough. So, to make it possible to compute the ID from the `TransactionHeader`, we'd need to change `nullifiers` to be "note commitments" - which may be a good idea.

Also, looking at this now, we may want to revisit how we send `NoteMetadata`. Previously, note metadata was limited to 32 bytes, but now is variable in length and could be as big as 8KB (or maybe even 16 KB). To make this "constant size", we could send something more like `NoteMetadataHeader` instead - and send the full metadata only when individual note is requested.

Lastly, I think it may be a good idea to add fees to `TransactionHeader` message. I'm actually not sure why fees are not a part of the `TransactionId` computation. cc @PhilippGackstatter in case he remembers.

PhilippGackstatter: > I'm actually not sure why fees are not a part of the `TransactionId` computation.


I don't think it ever came up, but it seems unproblematic to include it and would make the `TransactionId` commit to the entire `TransactionHeader` contents. For the purpose of the batch kernel, I don't think we need it, since the fee would be committed to by the execution proof, but it also doesn't hurt.

Opened https://github.com/0xMiden/miden-base/issues/2348.

igamigo: > Also, looking at this now, we may want to revisit how we send `NoteMetadata`. Previously, note metadata was limited to 32 bytes, but now is variable in length and could be as big as 8KB (or maybe even 16 KB). To make this "constant size", we could send something more like `NoteMetadataHeader` instead - and send the full metadata only when individual note is requested.

Do you mean changing `NoteMetadata` to be a header struct across all endpoints (ie, sync state/notes as well)?

bobbinth: > Do you mean changing `NoteMetadata` to be a header struct across all endpoints (ie, sync state/notes as well)?

No, I was thinking only for the sync endpoint as these supposed to return small summaries of notes. Before attachments, note metadata was just 32 bytes - but now it could be up to 8KB or 16KB. Instead of returning a full attachment, I think we can just return the 64 bytes which contain the metadata header + attachment word.

The user would still be able to use the `GetNotesById` endpoint to get all relevant note details (including the full attachment).

Mirko-von-Leipzig:
@SantiagoPittella I'm closing this since it looks like we want a different overall approach.

I've opened #1605 to discuss our options.

Issue #624, opened 2025-01-16
Use `enum` for `note_type` field in protobuf representation of `NoteMetadata`


polydez:
Currently we encode `note_type` field in protobuf representation of `NoteMetadata` as `uint32` with additional description of encoding:

https://github.com/0xPolygonMiden/miden-node/blob/3612ef622e3affef87d05bc638b00a7b3de01ab9/proto/note.proto#L14-L15

We could use `enum` support in Protocol Buffers in order to make our code/encoding clearer:
```proto
enum NoteType {
    NOTE_TYPE_UNSPECIFIED = 0;
    NOTE_TYPE_PUBLIC = 1;
    NOTE_TYPE_PRIVATE = 2;
    NOTE_TYPE_ENCRYPTED = 3;
}


// Represents a note's metadata.
message NoteMetadata {
    // ...

    // The type of the note (0b01 = public, 0b10 = private, 0b11 = encrypted).
    NoteType note_type = 2;

    // ....
}
```

PR #1594, opened 2026-01-26
chore: use enum for NoteType

SantiagoPittella:
closes https://github.com/0xMiden/miden-node/issues/624

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,11 @@
 
 - Changed `note_type` field in proto `NoteMetadata` from `uint32` to a `NoteType` enum ([#1594](https://github.com/0xMiden/miden-node/pull/1594)).
 
-## v0.13.2 (2026-01-27)
+## v0.13.3 (2026-01-29)
 
-### Fixes
+- Fixed network monitor faucet test failing to parse `/get_metadata` response due to field type mismatches ([#1612](https://github.com/0xMiden/miden-node/pull/1612)).
+
+## v0.13.2 (2026-01-27)
 
 - Network transaction builder no longer creates conflicting transactions by consuming the same notes twice ([#1597](https://github.com/0xMiden/miden-node/issues/1597)).
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2726,6 +2726,7 @@ dependencies = [
  "hex",
  "humantime",
  "miden-node-proto",
+ "miden-node-rocksdb-cxx-linkage-fix",
  "miden-node-utils",
  "miden-protocol",
  "miden-standards",
@@ -2755,6 +2756,7 @@ dependencies = [
  "humantime",
  "miden-node-block-producer",
  "miden-node-ntx-builder",
+ "miden-node-rocksdb-cxx-linkage-fix",
  "miden-node-rpc",
  "miden-node-store",
  "miden-node-utils",
@@ -2867,6 +2869,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "miden-node-rocksdb-cxx-linkage-fix"
+version = "0.14.0"
+
+[[package]]
 name = "miden-node-rpc"
 version = "0.14.0"
 dependencies = [
@@ -2916,6 +2922,7 @@ dependencies = [
  "miden-crypto",
  "miden-node-proto",
  "miden-node-proto-build",
+ "miden-node-rocksdb-cxx-linkage-fix",
  "miden-node-test-macro",
  "miden-node-utils",
  "miden-protocol",
@@ -2947,6 +2954,7 @@ dependencies = [
  "miden-block-prover",
  "miden-node-block-producer",
  "miden-node-proto",
+ "miden-node-rocksdb-cxx-linkage-fix",
  "miden-node-store",
  "miden-node-utils",
  "miden-protocol",
@@ -3101,6 +3109,7 @@ dependencies = [
  "miden-block-prover",
  "miden-node-proto",
  "miden-node-proto-build",
+ "miden-node-rocksdb-cxx-linkage-fix",
  "miden-node-utils",
  "miden-protocol",
  "miden-standards",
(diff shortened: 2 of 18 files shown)


proto/proto/types/note.proto
```diff
@@ -7,6 +7,18 @@ import "types/account.proto";
...
+// The type of a note.
+enum NoteType {
+    // Unspecified note type (default value, should not be used).
+    NOTE_TYPE_UNSPECIFIED = 0;
+    // Public note - details are visible on-chain.
+    NOTE_TYPE_PUBLIC = 1;
+    // Private note - details are not visible on-chain.
+    NOTE_TYPE_PRIVATE = 2;
+    // Encrypted note - details are encrypted on-chain.
+    NOTE_TYPE_ENCRYPTED = 3;
```
bobbinth: We no longer have `Encrypted` note type in `miden-base` (`next` branch) - so, we probably don't need to introduce it here either.

crates/proto/src/domain/note.rs
Mirko-von-Leipzig, line 1: We're assuming here that the enum _integer_ values are aligned. They do happen to be, though it triggered alarms in me because _why_ wouldn't the protocol type start with `0`. I would prefer if this was implemented as a match, that way we don't have to audit this manually.

Could you implement

```rust
From<miden_protocol::note::NoteTag> for proto::NoteTag {..}

TryFrom<proto::NoteTag> for miden_protocol::note::NoteTag {..}
```

proto/proto/types/note.proto
Mirko-von-Leipzig: With my `match` request we would still have to cater for it in this PR (unless we rebase this PR after the bump).

bobbinth: Since https://github.com/0xMiden/miden-node/pull/1595 is basically already done, I'd merge https://github.com/0xMiden/miden-node/pull/1595 first, and then rebase this PR.

crates/proto/src/domain/note.rs
```diff
@@ -75,7 +104,7 @@ impl From<NetworkNote> for proto::note::NetworkNote {
 impl From<NoteMetadata> for proto::note::NoteMetadata {
     fn from(val: NoteMetadata) -> Self {
         let sender = Some(val.sender().into());
-        let note_type = val.note_type() as u32;
+        let note_type = proto::note::NoteType::from(val.note_type()) as i32;
```
Mirko-von-Leipzig, line 109: That's frustrating that it still uses the integer as the field. Oh well.

crates/proto/src/domain/note.rs
```diff
@@ -29,7 +56,9 @@ impl TryFrom<proto::note::NoteMetadata> for NoteMetadata {
             .sender
             .ok_or_else(|| proto::note::NoteMetadata::missing_field(stringify!(sender)))?
             .try_into()?;
-        let note_type = NoteType::try_from(u64::from(value.note_type))?;
+        let note_type = proto::note::NoteType::try_from(value.note_type)
+            .map_err(|_| ConversionError::EnumDiscriminantOutOfRange)?
```
Mirko-von-Leipzig, line 62: Is this required? Seems like its mapping from `E -> E`?

crates/proto/src/domain/note.rs
SantiagoPittella: At the end of the day for the proto message itself we need the integer 😞 is essentially because of how protobuf works with enums

crates/proto/src/domain/note.rs
SantiagoPittella: This `try_from` is implemented by prost, is the casting from the i32 to an Enum variant and returns a custom error

proto/proto/types/note.proto
```diff
@@ -7,6 +7,16 @@ import "types/account.proto";
 // NOTES
 // ================================================================================================
 
+// The type of a note.
+enum NoteType {
+    // Unspecified note type (default value, should not be used).
+    NOTE_TYPE_UNSPECIFIED = 0;
```
bobbinth, line 13: I'm assuming we need this because we don't define value `0` for `NoteType` enum in `miden-base`, right?

Mirko-von-Leipzig: This is [recommended practice](https://protobuf.dev/best-practices/dos-donts/#unspecified-enum) by protobuf. That this lines up with the definition in `base` is pure coincidence (and also why I wanted to use match instead).

As a summary:

- if this enum is added as a new field to an existing message, then users of the old spec won't know about this field (and therefore won't set it)
- and receiver's would set this "missing" value to the default value of zero
- so by having a dedicated default enum variant we allow for protocol evolution that allows differentiating between "field missing, was defaulted"
- aka protobuf assigns default values to primitives, enums are "integers in a trench coat" and therefore get zero-defaulted and we want to distinguish missing from the real variants

> Enums should include a default FOO_UNSPECIFIED value as the first value in the declaration. When new values are added to an enum, old clients will see the field as unset and the getter will return the default value or the first-declared value if no default exists . For consistent behavior with [proto enums](https://protobuf.dev/programming-guides/proto2#enum), the first declared enum value should be a default FOO_UNSPECIFIED value and should use tag 0. It may be tempting to declare this default as a semantically meaningful value but as a general rule, do not, to aid in the evolution of your protocol as new enum values are added over time. All enum values declared under a container message are in the same C++ namespace, so prefix the unspecified value with the enum’s name to avoid compilation errors. If you’ll never need cross-language constants, an int32 will preserve unknown values and generates less code. Note that [proto enums](https://protobuf.dev/programming-guides/proto2#enum) require the first value to be zero and can round-trip (deserialize, serialize) an unknown enum value.

crates/proto/src/domain/note.rs
Mirko-von-Leipzig: Man that's so weird. What's even the point of the enum if they never actually use it in the code..

Issue #1564, opened 2026-01-21
Replace `BlockNumber::from(u32::MAX)` with `BlockNumber::MAX`


SantiagoPittella:
In https://github.com/0xMiden/miden-base/pull/2324 I added `BlockNumber::MAX` since we were doing `BlockNumber::from(u32::MAX)` in various places here and in the client.

The feature is now added in miden-base and we need to use it in the node.

PR #1595, opened 2026-01-26
chore: update to miden-base latest

SantiagoPittella:
Closes https://github.com/0xMiden/miden-node/issues/1564

The PR was initially intended to use `BlockNumber::MAX` but I needed to update to the latest version of miden-base.

### DIFF
--- a/crates/ntx-builder/src/store.rs
+++ b/crates/ntx-builder/src/store.rs
@@ -236,7 +236,7 @@ impl StoreClient {
         &self,
         sender: tokio::sync::mpsc::Sender<NetworkAccountId>,
     ) -> Result<(), StoreError> {
-        let mut block_range = BlockNumber::from(0)..=BlockNumber::MAX;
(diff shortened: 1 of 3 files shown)


crates/ntx-builder/src/store.rs
```diff
@@ -236,10 +236,10 @@ impl StoreClient {
         &self,
         sender: tokio::sync::mpsc::Sender<NetworkAccountId>,
     ) -> Result<(), StoreError> {
-        let mut block_range = BlockNumber::from(0)..=BlockNumber::from(u32::MAX);
+        let mut block_range = BlockNumber::from(0)..=BlockNumber::MAX;
```
Mirko-von-Leipzig:
```suggestion
        let mut block_range = BlockNumber::GENESIS..=BlockNumber::MAX;
```

Mirko-von-Leipzig: Maybe also do a search through the code base for the same pattern and just replace them.

PR #1598, opened 2026-01-27
fix: load foreign `AccountCode` on ntx builder

igamigo:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.13.2 (TBC)
```
Mirko-von-Leipzig: We actually never released v0.13.1 so you can merge this (and make the date TBD so it doesn't cause more confusion 😓)

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -439,6 +439,10 @@ impl DataStore for NtxDataStore {
                     |err| DataStoreError::other_with_source("failed to get account inputs", err),
                 )?;
 
+            // Ensure foreign account procedures are available to the executor via the mast store.
+            // This assumes the code was not loaded from before
+            self.mast_store.load_account_code(account_inputs.code());
```
bobbinth, line 444: I was incorrect before. I thought `TransactionBaseHost` (in `load_foreign_account_code()`) but that method just update the procedure index map. Though, I think it is somewhat confusing that `TransactionBaseHost` also manages a `mast_store`.

Maybe we should simplify the structure somehow. cc @PhilippGackstatter - maybe we should create an issue about this in `miden-base`?

PR #1599, opened 2026-01-27
fix(ntx): use correct note script serde

Mirko-von-Leipzig:
The ntx builder was parsing the gRPC response as if the note script was written as bytes which didn't match the way it was encoded.

### DIFF
--- a/crates/proto/src/domain/note.rs
+++ b/crates/proto/src/domain/note.rs
@@ -14,7 +14,7 @@ use miden_protocol::note::{
     NoteType,
     Nullifier,
 };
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig:
> Looks good! We use equivalent deserialization code from the client for the same endpoint, so I'm confident this will work, but would be nice to have this tested somehow

I'm hoping to find a better abstraction for our gRPC encoding and decoding. That way these types can be used in pairs (and tested more readily).

With the way this is currently implemented (`From` and `TryFrom`) its impossible to know what's actually happening at the call site. And testing is also a much larger pain.

Issue #1597, opened 2026-01-27
Ntx conflicts with its own transactions


Mirko-von-Leipzig:
Observed on devnet v0.13 on the network monitor's account.

The ntx has failing network transactions rejected by the block-producer due to 

```
input notes with given nullifiers were already consumed by another transaction
```

which indicates that the nullifiers already exist. Since this account is only being interacted with by the ntx itself, we can conclude that the ntx is creating conflicting transactions. 

This implies that the ntx is creating consecutive transactions consuming the same notes. What should happen is that the actor updates its local state to match the transaction going through - including removing the input notes as candidates.

A simple but effective way to implement this is once the actor successfully submits a transaction, it enters a new state where it simply _waits_ for the mempool event for this transaction. And then updates its own state accordingly, and continues looking for the next transaction to submit. An escape hatch (in case a bug causes this transaction event to never arrive) is to add a length timeout to this state e.g. if no matching event arrives after 5 minutes, abort, or move to a resync stage. Probably just aborting the actor is fine for now.

tbh I thought we had code for this but maybe I'm just remembering the previous implementation.

Issue #1596, opened 2026-01-27
Ntx transactions eventually overflow gRPC size limit


Mirko-von-Leipzig:
Observed on devnet v0.13 with the network monitor's counter contract.

After a while of successfully running, all transactions for this account started failing with:

```
failed to submit transaction caused by: status: 'Operation was attempted past the valid range', self: "Error, decoded message length too large: found 5612970 bytes, the limit is: 4194304 bytes"
```

indicating that the submitted transaction + inputs exceeded the validator's gRPC server limit of 4 MB.

The error went away after restarting the node, indicating that something in the transaction "grows" over time (probably).

---------------------------------

Speculation is that this is likely something in transaction inputs since the proven transaction itself _should_ be bounded given the simplicity of the contract and notes. Transaction's were failing even with a single consume note, and did previously succeed.

A likely culprit is `chain_state` though it does have code to keep the mmr pruned to a constant `4` blocks.

If this is the result of growth over time then we should be able to measure this even on node start-up. I recommend debugging by starting a local node + monitor and logging the transaction + inputs size in bytes and then digging further into the sizes of fields from there.

PR #1602, opened 2026-01-27
fix: avoid duplicated consumption

SantiagoPittella:
closes https://github.com/0xMiden/miden-node/issues/1597

> I worked on the assumption that this happened after a node restart to fix https://github.com/0xMiden/miden-node/issues/1596 .

The issue was that we were initializing the `NetworkAccountState` with the proper notes but without the nullifiers.

The existing code has something to handle similar cases but the problem was that when `mempool_update()` processed the event, the `add_nullifier()` call was skipped due to the empty nullifier_idx, so the notes were never actually marked as consumed.

### DIFF
--- a/crates/ntx-builder/src/actor/account_state.rs
+++ b/crates/ntx-builder/src/actor/account_state.rs
@@ -62,8 +62,13 @@ pub struct NetworkAccountState {
     /// an impact are ignored.
     inflight_txs: BTreeMap<TransactionId, TransactionImpact>,
 
-    /// A set of nullifiers which have been registered for the network account.
-    nullifier_idx: HashSet<Nullifier>,
+    /// Nullifiers of all network notes targeted at this account.
+    ///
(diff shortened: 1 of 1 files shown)


crates/ntx-builder/src/actor/account_state.rs
bobbinth: Maybe good to expand the comment here to explain what exactly this means. My understanding is that this is the set of nullifiers of notes that are sent to this account. This is probably used to notify the account when a note that is queued for it to consume has been consumed already (by someone else or by this account).

Also, nit: the naming is a bit confusing here: why is it `nullifier_idx` rather than just `nullifiers`?

SantiagoPittella: The naming is actually weird, I'm assuming that it is left from an older implementation. I renamed it to `known_nullifiers`.

PR #1606, opened 2026-01-28
Move state-sync related methods to a separate file

bobbinth:
This small PR moves sync-related methods from the `State` struct in the store to a separate file. The main motivation is to make the `mod.rs` file a bit smaller and easier to navigate. No actual code changes were made here.

### DIFF
--- a/crates/store/src/state/sync_state.rs
+++ b/crates/store/src/state/sync_state.rs
@@ -1,0 +1,141 @@
+use std::ops::RangeInclusive;
+
+use miden_protocol::account::AccountId;
+use miden_protocol::block::BlockNumber;
+use miden_protocol::crypto::merkle::mmr::{Forest, MmrDelta, MmrProof};
+use tracing::instrument;
+
+use super::State;
+use crate::COMPONENT;
+use crate::db::models::queries::StorageMapValuesPage;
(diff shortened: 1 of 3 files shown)


crates/store/src/state/sync_state.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// block with a matching tag, or the chain tip. All the other values are filter based on this
+    /// block range.
+    ///
+    /// # Arguments
+    ///
+    /// - `note_tags`: The tags the client is interested in, resulting notes are restricted to the
+    ///   first block containing a matching note.
+    /// - `block_range`: The range of blocks from which to synchronize notes.
+    #[instrument(level = "debug", target = COMPONENT, skip_all, ret(level = "debug"), err)]
+    pub async fn sync_notes(
```
bobbinth, line 40: Some methods here are instrumented, while others are not. I couldn't tell why that's the case - but kept things as they were.

Mirko-von-Leipzig (approved):
Long(er) term it would be nice to move to something like file-per-method implementations

crates/store/src/state/sync_state.rs
Mirko-von-Leipzig: Yeah some still have them at `level=debug` which for all purposes means they never get logged in production. We need to figure an approach here still.

PR #1607, opened 2026-01-28
feat: rocksdb linkage of existing static rocksdb library is missing stdc++

drahnr:
## Problem

`rocksdb-rust` has an issue supplying `stdc++` (for linux, differs for platforms) to rustc at linktime. This implies missing C++ standard library features which in turn causes missing symbols when linking an _existing_ `rocksdb.a`.

### It doesn't show for me

Currently we _do_ recompile both locally and in CI rocksdb with the cache (~1GB) and time (~50s) for each of those runs.
If you want to avoid this, the following has worked in the _past_ and shaves off both:

1. compile once without the env vars in step 4
2. find the `rocksdb.a` in your `target` dir
3. copy the `rocksdb.a` 

4.  ```sh
    # assumes linux and home
    export TMP_ARTIFACT="$(fd -i -H --max-results=1 --extension a '.*rocksdb.*' target)"
    cp -v "${TMP_ARTIFACT}" "${HOME}/.local/lib/"
    export ROCKSDB_COMPILE=true
    export ROCKSDB_STATIC
    export ROCKSDB_LIB_DIR=${HOME}/.local/lib/
    make build
    ```
5. the output should show `librocksdb-sys(build)` for less than 100ms instead of 50seconds

### Why do we need this change?

Upstream's last release (10th of August '25) has a bug, it's already fixed in `master`, but there is no release including that, the relevant PR already landed:
https://github.com/rust-rocksdb/rust-rocksdb/pull/1029

tl;dr - this is a stopgap solution to allow us to save both time and space in CI, as well as locally

### Next steps

Separate steps of building our code and rocksdb - the latter should provide a statically linked rocksdb.a in most (all?) CI steps.


---

## Approach

### Possible paths

1. use a `.cargo/config.toml` entry - while it works it will become an issue once trying to link musl or custom toolchain paths since it doesn't support `CXXSTD` env var - we also cannot simply gate for specific targets iiuc
2. implement a `build.rs` for each binary to add the correct linkage flags - it's opt-in and doesn't break any other paths

Choosing the latter, it's more code, but it's both opt-in and very deliberate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,11 @@
 
 - [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/miden-node/pull/1595)).
 
-## v0.13.2 (2026-01-27)
+## v0.13.3 (2026-01-29)
 
-### Fixes
+- Fixed network monitor faucet test failing to parse `/get_metadata` response due to field type mismatches ([#1612](https://github.com/0xMiden/miden-node/pull/1612)).
+
+## v0.13.2 (2026-01-27)
 
 - Network transaction builder no longer creates conflicting transactions by consuming the same notes twice ([#1597](https://github.com/0xMiden/miden-node/issues/1597)).
 
--- a/bin/network-monitor/src/faucet.rs
+++ b/bin/network-monitor/src/faucet.rs
@@ -61,14 +61,14 @@ struct GetTokensResponse {
 /// Response from the faucet's `/get_metadata` endpoint.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct GetMetadataResponse {
+    version: String,
     id: String,
     issuance: u64,
     max_supply: u64,
     decimals: u8,
-    explorer_url: String,
-    pow_load_difficulty: u32,
+    explorer_url: Option<String>,
+    pow_load_difficulty: u64,
     base_amount: u64,
-    version: Option<String>,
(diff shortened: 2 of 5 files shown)


crates/rocksdb-cxx-linkage-fix/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+use std::env;
```
drahnr, line 5: TODO: add doc comment here

Cargo.toml
```diff
@@ -48,6 +49,9 @@ miden-node-utils            = { path = "crates/utils", version = "0.14" }
 miden-node-validator        = { path = "crates/validator", version = "0.14" }
 miden-remote-prover-client  = { path = "crates/remote-prover-client", version = "0.14" }
 
+# Temporary workaround.
+rocksdb-cxx-linkage-fix = { path = "crates/rocksdb-cxx-linkage-fix", version = "0.14" }
```
drahnr: TODO: add more context, add URL https://github.com/rust-rocksdb/rust-rocksdb/pull/1029

crates/rocksdb-cxx-linkage-fix/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+authors.workspace      = true
+description            = "Miden C++ stdlib link helper"
+edition.workspace      = true
+homepage.workspace     = true
+license.workspace      = true
+name                   = "rocksdb-cxx-linkage-fix"
```
Mirko-von-Leipzig: Should we `miden-node-` prefix it so it doesn't pollute crates.io namespace?

sergerad (approved):

Issue #1175, opened 2025-08-21
Prune account storage maps and vault tables


igamigo:
Following #1132, we have a table that holds account storage map updates. With #1140, we would enable querying that table in such a way that users can periodically sync and get updates for public accounts. 

The design (discussed mostly [here](https://github.com/0xMiden/miden-node/discussions/722#discussioncomment-13748611) and in #1095) is based around keeping a set of historical values to allow for syncing up to specific moments with lesser time restrictions. We need to add a way to prune these tables, as well as decide what would be a good threshold of historical values to keep. 
For example, we could decide to prune the table(s) after each `apply_block` and set a cutoff of 100 blocks. So any value in the maps table that has a `block_num` that is smaller than `new_chain_tip - 100` would be deleted.

bobbinth:
I agree with this - but we should make sure we have good telemetry for this step as I suspect it could become a bottleneck.

PR #1608, opened 2026-01-28
feat: prune `InnerForest`

drahnr:
Does cleanup the `InnerForest`, both the lookup tables/maps and the actual `SmtForest`.

Required for bounding the in-memory size growth.

Related #1175

### DIFF
--- a/crates/store/src/inner_forest/mod.rs
+++ b/crates/store/src/inner_forest/mod.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet, HashMap};
+use std::collections::{BTreeMap, BTreeSet};
 
 use miden_node_proto::domain::account::{AccountStorageMapDetails, StorageMapEntries};
 use miden_protocol::account::delta::{AccountDelta, AccountStorageDelta, AccountVaultDelta};
@@ -97,7 +97,7 @@ pub(crate) struct InnerForest {
     vault_roots_by_block: BTreeMap<BlockNumber, Vec<AccountId>>,
 
     /// Tracks storage map roots by block number for pruning.
-    storage_slots_by_block: HashMap<BlockNumber, Vec<(AccountId, StorageSlotName)>>,
+    storage_slots_by_block: BTreeMap<BlockNumber, Vec<(AccountId, StorageSlotName)>>,
 }
 
 impl InnerForest {
@@ -108,7 +108,7 @@ impl InnerForest {
             storage_entries: BTreeMap::new(),
             vault_roots: BTreeMap::new(),
             vault_roots_by_block: BTreeMap::new(),
-            storage_slots_by_block: HashMap::new(),
+            storage_slots_by_block: BTreeMap::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


crates/store/src/db/mod.rs
```diff
@@ -49,6 +49,7 @@ pub(crate) mod schema;
 
 pub type Result<T, E = DatabaseError> = std::result::Result<T, E>;
 
+#[derive(Clone)]
```
drahnr: TODO: remove

crates/store/src/db/mod.rs
```diff
@@ -320,7 +321,9 @@ impl Db {
         );
 
         let me = Db { pool };
+
```
drahnr: TODO: stray

crates/store/src/db/mod.rs
```diff
@@ -570,6 +573,9 @@ impl Db {
         notes: Vec<(NoteRecord, Option<Nullifier>)>,
     ) -> Result<()> {
         self.transact("apply block", move |conn| -> Result<()> {
+            // TODO: This span is logged in a root span, we should connect it to the parent one.
+            let _span = info_span!(target: COMPONENT, "write_block_to_db").entered();
```
drahnr: TODO: rebase artifact

crates/store/src/inner_forest/mod.rs
```diff
@@ -595,4 +609,76 @@ impl InnerForest {
...
+
+        // Prune the underlying SmtForest by telling it which roots are still referenced.
+        // Collect all roots that we're keeping from both vault_roots and storage_map_roots.
+        let roots_to_keep: Vec<Word> = self
+            .vault_roots
+            .values()
+            .chain(self.storage_map_roots.values())
+            .copied()
+            .collect();
+        self.forest.pop_smts(roots_to_keep);
```
drahnr: TODO: logic issue

crates/store/src/inner_forest/mod.rs
```diff
@@ -60,15 +70,34 @@ pub(crate) struct InnerForest {
...
 
     /// Maps (`account_id`, `block_num`) to vault SMT root.
     /// Tracks asset vault versions across all blocks with structural sharing.
+    ///
+    /// Attention: Must be a `BTreeMap`, since not every block contains a value here, so we need to
+    /// be able to query the previous blocks cheaply.
     vault_roots: BTreeMap<(AccountId, BlockNumber), Word>,
+
+    /// Tracks vault roots by block number for pruning.
+    vault_roots_by_block: BTreeMap<BlockNumber, Vec<AccountId>>,
```
drahnr, line 97: we use this to gain a the `.range` of blocks for free further down.

crates/store/src/inner_forest/mod.rs
```diff
@@ -60,15 +70,34 @@ pub(crate) struct InnerForest {
...
+
+    /// Tracks storage map roots by block number for pruning.
+    storage_slots_by_block: HashMap<BlockNumber, Vec<(AccountId, StorageSlotName)>>,
```
drahnr: This is the easy path of cheaply collecting per block entries, there is an alternative using `BTreeMap::range` and fiddling with key ordering a bit, but it's somewhat brittle and subtle. KISS, since this is going to be largely obsoleted by `LargeSmtForest` implementation.

drahnr: We need both to btrees, ignore the above

PR #1610, opened 2026-01-28
refactor: ntx builder start up

### LINKED ISSUE #1573: Improve ntx builder start up
#1568 fixed issue #1566 with a hacky solution (1 in the issue). Ideally we would do (2) instead.

The network transaction builder's start-up is currently done inline with the main event loop. It would be good to refactor this to more clearly distinguish between initialization and running. Configuration is also an issue e.g. the `new` function currently takes in a lot of `Url` but its unclear at the caller side which is meant for which.

I'm thinking an API to enable something like:

```rust
NetworkTransactionBuilder {
    store_url,
    block_producer_url,
    validator_url,
    tx_prover_url,
    script_cache_size,
}
// 1. Subscribe
// 2. Buffer mempool events and wait for 1st block
// 3. Fetch that block chain info from store
// 4. Create type that can be .run
.initialize().await?
// Main event loop
.run().await?
```

SantiagoPittella:
closes #1573 

Refactors NTX Builder startup into distinct phases and introduces typed configuration.

- Split `NetworkTransactionBuilder` initialization phases
- Added `NtxBuilderConfig` struct

Regarding the new `InitializedBuilder`:

The `initialize()` method creates a gRPC subscription stream that needs to be stored for later use in `run()`. Rust's `impl TryStream` return types are opaque, causing lifetime errors when stored in a struct. Boxing the stream with Pin<Box<dyn Stream + 'static>> forces it to own its data, but this requires a separate struct to hold the boxed stream between initialization and execution. Couldn't thought of anything better cc @Mirko-von-Leipzig


crates/ntx-builder/src/builder.rs
```diff
@@ -84,108 +72,157 @@ pub struct NetworkTransactionBuilder {
...
+    /// 1. Spawns a background task to load existing network accounts from the store
+    /// 2. Runs the main event loop, processing mempool events and managing actors
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - The mempool event stream ends unexpectedly
+    /// - An actor encounters a fatal error
+    /// - The account loader task fails
+    pub async fn run(mut self) -> anyhow::Result<()> {
```
sergerad, line 115: Would it make more sense (semantically) if we had `NetworkTransactionBuilder::run()`? I think the `NetworkTransactionBuilder` struct should represent the final, runnable state.

SantiagoPittella: Yeah, makes sense. I changed it.

bin/node/src/commands/mod.rs
```diff
@@ -68,6 +68,7 @@ pub struct NtxBuilderConfig {
     )]
     pub ticker_interval: Duration,
 
+    /// Size of the LRU cache for note scripts.
```
```suggestion
    /// Number of note scripts to cache locally.
    ///
    /// Note scripts not in cache must first be retrieved from the store.
```

bin/node/src/commands/mod.rs
```diff
@@ -77,6 +78,20 @@ pub struct NtxBuilderConfig {
     pub script_cache_size: NonZeroUsize,
```
Mirko-von-Leipzig, line 80: This is more a thought experiment, but I wonder if we ever want to set this to zero? Seems unlikely, but that would effectively be the same as disabling it.

I think `NonZero` is used because that's what LRU requires, so we would have to do `struct MyLru<T>(Option<Lru<T>>)` which is a bit more work but not too much.

I kind of like the idea that we can disable caching e.g. for performance testing. wdyt? As a separate issue though, if anything. Could also easily be a YAGNI thing so maybe we just leave as is until we do need it.

bin/node/src/commands/mod.rs
```diff
@@ -77,6 +78,20 @@ pub struct NtxBuilderConfig {
...
 }
 
+impl NtxBuilderConfig {
+    /// Converts this CLI config into the ntx-builder's internal config.
+    pub fn into_builder_config(
```
Mirko-von-Leipzig, line 85: Bit confusing with the stuttering builder/builder naming, but that's not this PRs fault. We should come up with something better than network transaction builder..

Maybe network transaction..
- producer
- service
- component (bleh)
- handler
- .. hm.

crates/ntx-builder/src/builder.rs
```diff
@@ -286,9 +234,11 @@ impl NetworkTransactionBuilder {
...
 
         // Keep MMR pruned.
-        let pruned_block_height =
-            (chain_state.chain_mmr.chain_length().as_usize().saturating_sub(MAX_BLOCK_COUNT))
-                as u32;
+        let pruned_block_height = (chain_state
+            .chain_mmr
+            .chain_length()
+            .as_usize()
+            .saturating_sub(self.config.max_block_count)) as u32;
```
Mirko-von-Leipzig, line 241: Maybe we have saturating sub on `BlockNumber`? If not, we should add it.

crates/ntx-builder/src/lib.rs
```diff
@@ -13,12 +25,223 @@ pub use builder::NetworkTransactionBuilder;
...
+}
+
+impl NtxBuilderConfig {
+    pub fn new(store_url: Url, block_producer_url: Url, validator_url: Url) -> Self {
+        Self {
+            store_url,
+            block_producer_url,
+            validator_url,
+            tx_prover_url: None,
+            script_cache_size: NonZeroUsize::new(DEFAULT_SCRIPT_CACHE_SIZE).unwrap(),
```
Mirko-von-Leipzig: May as well make the constant a `NonZeroUsize` (and same for `DEFAULT_MAX_NOTES_PER_TX`).

bin/node/src/commands/mod.rs
SantiagoPittella: Maybe to test performance of the underlaying implementation? Otherwise I don't see why we need that

bin/node/src/commands/mod.rs
SantiagoPittella: Yeah, I agree. We tend to use a lot the builder pattern so it is a bit confusing. I like both "producer" and "service", and even "consumer". I'll create a separate  issue to discuss and rename the component since it touches too many files for this PR

crates/ntx-builder/src/builder.rs
SantiagoPittella: It is not implemented on `BlockNumber` :(

bin/node/src/commands/mod.rs
SantiagoPittella: I opened https://github.com/0xMiden/miden-node/issues/1620

PR #1611, opened 2026-01-28
chore: ntx-builder followups

### LINKED ISSUE #1327: NTX Actor Followup
From [comment](https://github.com/0xMiden/miden-node/pull/1219#pullrequestreview-3380432171):

>I also think this structure is probably half-way to where we'd need to get to eventually. The main things to address in follow-ups:
>
>1. How we segment the actor state - the current structure does not seem clean to me - but changing it now will probably result in quite a bit of refactoring.
>2. How we handle event routing in the coordinator. I think this may have significant performance implications - but also, we should confirm if that's the case first.
>3. How to handle actor failures - e.g., under what circumstances we may want to respawn failed actors.
>
>The first point is probably also good to address in combination with introducing lazy account loading as both of these will have impact on how the actor state is structured.

SantiagoPittella:
closes #1327 

Refactors the `NetworkAccountNoteState` into two focused types: `AccountDeltaTracker` for committed/inflight account states, and `NotePool` for available/nullified notes.

Also adds tracing instrumentation to `Coordinator::broadcast()` for monitoring event distribution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,10 +63,12 @@
 - The network monitor now marks the chain as unhealthy if it fails to create new blocks ([#1512](https://github.com/0xMiden/miden-node/pull/1512)).
 - Limited number of storage map keys in `GetAccount` requests ([#1517](https://github.com/0xMiden/miden-node/pull/1517)).
 - Block producer now detects if it is desync'd from the store's chain tip and aborts ([#1520](https://github.com/0xMiden/miden-node/pull/1520)).
+- Pin tool versions in CI ([#1523](https://github.com/0xMiden/miden-node/pull/1523)).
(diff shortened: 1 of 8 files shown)


sergerad (approved):

crates/ntx-builder/src/coordinator.rs
```diff
@@ -154,12 +154,11 @@ impl Coordinator {
     /// message channel and can process it accordingly.
     ///
     /// If an actor fails to receive the event, it will be canceled.
+    #[tracing::instrument(name = "ntx.coordinator.broadcast", skip(self, event), fields(
+        actor_count = self.actor_registry.len(),
+        event_kind = %event.kind()
+    ))]
```
Mirko-von-Leipzig:
```suggestion
    #[tracing::instrument(name = "ntx.coordinator.broadcast", skip_all, fields(
        actor.count = self.actor_registry.len(),
        event.kind = %event.kind()
    ))]
```

PR #1614, opened 2026-01-29
feat: Validator database

sergerad:
## Context

The Validator currently only stores validated transactions in-memory. We want to persist validated transactions for resilience and potential future use cases around debugging / audit-ability w.r.t public transactions.

The Validator will be run as a separate instance to the node so it needs its own database, schema, etc. 

Relates to #1316.

## Changes
- Add data directory CLI arg for Validator command.
- Make various types and functions from Db crate pub for use in Validator crate.
- Add sqlite, diesel scaffolding to Validator crate alongside new schema.
- Fix standalone Validator setup with bundled components.
- Rename insecure key variables to just validator key.
- Merge Validator command args into `ValidatorConfig` struct.

### DIFF
--- a/.github/workflows/cleanup-workflows.yml
+++ b/.github/workflows/cleanup-workflows.yml
@@ -36,106 +36,248 @@ jobs:
         run: |
           git fetch origin main
           WORKFLOWS=$(git ls-tree -r origin/main --name-only | grep '^.github/workflows/')
-          echo $WORKFLOWS
-          echo "workflows=$WORKFLOWS" >> "$GITHUB_OUTPUT"
+          printf "%s\n" $WORKFLOWS
+          {
+            echo "workflows<<EOF"
+            echo "$WORKFLOWS"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
 
       - name: Workflows on next
         id: next
         run: |
           git fetch origin next
           WORKFLOWS=$(git ls-tree -r origin/next --name-only | grep '^.github/workflows/')
-          echo $WORKFLOWS
-          echo "workflows=$WORKFLOWS" >> "$GITHUB_OUTPUT"
-
-      - name: Workflows on github
-        id: github
-        run: |
-          # Note that we filter by `.github` path prefix to ensure we only get locally defined workflows.
-          #
-          # Examples of non-local workflows are `dependabot` and `copilot` which have paths:
-          # - dynamic/dependabot/dependabot-updates
-          # - dynamic/copilot-pull-request-reviewer/copilot-pull-request-reviewer
-          WORKFLOWS=$(gh workflow list \
-            --all \
-            --json path \
-            --jq '.[] | select(.path | startswith(".github")) | .path' \
-          )
-          echo $WORKFLOWS
-          echo "workflows=$WORKFLOWS" >> "$GITHUB_OUTPUT"
+          printf "%s\n" $WORKFLOWS
+          {
+            echo "workflows<<EOF"
+            echo "$WORKFLOWS"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
 
       - name: Filter for deleted workflows
         id: deleted
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
-          # Union of `main` and `next` workflows.
-          EXISTING_FILES=$(                       \
-            printf "%s\n%s\n"                     \
+          set -euo pipefail
+
+          # Union of `main` and `next` workflows as a JSON array of strings (paths)
+          EXISTING=$(printf "%s\n%s\n" \
             "${{ steps.main.outputs.workflows }}" \
             "${{ steps.next.outputs.workflows }}" \
           )
-          EXISTING_FILES=$(echo "$EXISTING_FILES" | sort -u)
-          echo $EXISTING_FILES
+          EXISTING=$(echo "$EXISTING" | sort -u | jq -R . | jq -s .)
 
-          # Find deleted workflows as the items in `WORKFLOWS` but not in the union of main and next.
-          # This assumes that _all_ items in main and next are present in `WORKFLOWS`.
-          DELETED_FILES=$(                          \
-            printf "%s\n%s\n"                       \
-            "$EXISTING_FILES"                       \
-            "${{ steps.github.outputs.workflows }}" \
+          echo "Existing workflows:"
+          echo "$EXISTING"
+
+          # Get workflows currently on GitHub as JSON array of objects
+          GITHUB=$(gh api repos/{owner}/{repo}/actions/workflows \
+            --jq '.workflows[] | select(.path | startswith(".github")) | { name, node_id, path }' \
+              | jq -s '.')
+
+          echo "Workflows on GitHub:"
+          echo "$GITHUB"
+
+          # Find deleted workflows: present on GitHub but not in main/next
+          DELETED=$(echo "$GITHUB" | jq -c \
+            --argjson existing "$EXISTING" '
+              map(select(.path as $p | $existing | index($p) | not))
+            '
           )
-          DELETED_FILES=$(echo "$DELETED_FILES" | sort | uniq -u)
-          echo $DELETED_FILES
-          echo "workflows=$DELETED_FILES" >> "$GITHUB_OUTPUT"
 
+          echo "Deleted workflows:"
+          echo "$DELETED"
+
+          # Output to GitHub Actions
+          {
+            echo "workflows<<EOF"
+            echo "$DELETED"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+      # Performs the actual run deletion.
+      #
+      # This contains a lot of code, but the vast majority is just pretty-printing.
       - name: Delete runs from deleted workflows
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           MODE: ${{ inputs.mode }}
-          DELETED_WORKFLOWS: ${{ steps.deleted.outputs.workflows }}
+          WORKFLOWS: ${{ steps.deleted.outputs.workflows }}
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.repository }}
+        shell: bash --noprofile --norc -euo pipefail {0}
         run: |
-          set -euo pipefail
+          if [ -z "$WORKFLOWS" ]; then
+            echo "No workflows to delete."
+            exit 0
+          fi
 
-          TOTAL_AFFECTED=0
+          # ================================================================================================
+          # Utility functions
+          # ================================================================================================
 
-          echo ""
-          echo "=== Workflow Cleanup Summary ==="
-          echo ""
+          # Fetches a page of workflow runs for a given workflow ID and cursor.
+          #
+          # We use github's graphql API here which allows us to paginate over workflow runs.
+          # Unfortunately `gh run list` does not support pagination, so we use the graphql API instead.
+          gh_workflow_run_page() {
+            local id="$1"
+            local cursor="$2"
 
-          while IFS= read -r workflow; do
-            [ -z "$workflow" ] && continue
+            gh api graphql -F workflowId="$id" -F after="$cursor" \
+              -f query='query($workflowId: ID!, $after: String) {
+                  node(id: $workflowId) {
+                    ... on Workflow {
+                      runs(first: 100, after: $after) {
+                        pageInfo { hasNextPage endCursor }
+                        nodes { databaseId }
+                      }
+                    }
+                  }
+                }'
+          }
 
-            WF_COUNT=0
+          # ================================================================================================
+          # Print helpers for nice progress and table display
+          # ================================================================================================
 
+          # Column widths (table includes three spacers for ' | ' between columns)
+          widths_index=9
+          widths_name=30
+          widths_count=14
+          widths_total=12
+          widths_table=$(( $widths_index + 3 + $widths_name + 3 + $widths_count + 3 + $widths_total ))
+
+          # Repeats a character a given number of times.
+          repeat_char() {
+            local char=$1
+            local count=$2
+            printf "%0.s$char" $(seq 1 $count)
+          }
+
+          # Prints the given header as `==== <header> ====` to match the table layout.
+          print_table_header() {
+            local header="$1"
+            local header_len=${#header}
+            local left_pad=$(( ( $widths_table - header_len - 2) / 2 ))
+            local right_pad=$(( $widths_table - header_len - 2 - left_pad ))
+            printf " \n%s %s %s\n" $(repeat_char = $left_pad) "$header" $(repeat_char = $right_pad)
+          }
+
+          # Prints |---+---+---+---| with appropriate widths to accomodate the table headers.
+          print_table_separator() {
+            printf "%s+%s+%s+%s\n"                     \
+              "$(repeat_char - $((widths_index + 1)))" \
+              "$(repeat_char - $((widths_name + 2)))"  \
+              "$(repeat_char - $((widths_count + 2)))" \
+              "$(repeat_char - $((widths_total + 1)))"
+          }
+
+          # Prints a row of the table (index, workflow name, workflow count, global total)
+          print_table_row() {
+            local index=$1
+            local name=$2
+            local count=$3
+            local total=$4
+            printf "%*s | %-*s | %*s | %*s\n"  \
+              "$widths_index" "$index"          \
+              "$widths_name" "$name"            \
+              "$widths_count" "$count"          \
+              "$widths_total" "$total"
+          }
+
+          # Alias for print_table_row() with empty index and total columns.
+          print_summary_row() {
+            local name=$1
+            local count=$2
+            print_table_row "" "$name" "$count" ""
+          }
+
+          # ================================================================================================
+          # Print progress table header
+          # ================================================================================================
+          print_table_header "Workflow Cleanup Progress"
+          print_table_row "Index" "Workflow" "Workflow Count" "Global Total"
+          print_table_separator
+
+          # ================================================================================================
+          # Core workflow loop, iterate over workflows
+          # ================================================================================================
+
+          n_workflows=$(echo "$WORKFLOWS" | jq -r '. | length')
+          total=0
+          summary=()
+          index=0
+
+          mapfile -t WF_ARRAY < <(echo "$WORKFLOWS" | jq -c '.[]')
+          for wf in "${WF_ARRAY[@]}"; do
+            index=$((index + 1))
+            name=$(echo "$wf" | jq -r '.name')
+            count=0
+            id=$(echo "$wf" | jq -r '.node_id')
+
+            # Safety checks
+            if [ -z "$name" ]; then
+              echo "::error title=Workflow name empty::Resolved workflow name is empty at index $index"
+              exit 1
+            fi
+            if [ -z "$id" ]; then
+              echo "::error title=Workflow ID missing::Workflow '$name' has no ID"
+              exit 1
+            fi
+
+            cursor=""
+
+            # Paginate over workflow runs
             while true; do
-              RUN_IDS=$(gh run list \
-                --workflow "$workflow" \
-                --limit 100 \
-                --json databaseId \
-                --jq '.[].databaseId')
+              response=$(gh_workflow_run_page "$id" "$cursor")
 
-              if [ -z "$RUN_IDS" ]; then
-                break
-              fi
+              run_ids=$(echo "$response" | jq -r '.data.node.runs.nodes[].databaseId')
+              has_next=$(echo "$response" | jq -r '.data.node.runs.pageInfo.hasNextPage')
+              cursor=$(echo "$response" | jq -r '.data.node.runs.pageInfo.endCursor')
 
-              BATCH_COUNT=$(echo "$RUN_IDS" | wc -l | tr -d ' ')
-              WF_COUNT=$((WF_COUNT + BATCH_COUNT))
+              [ -z "$run_ids" ] && break
+
+              deleted=$(echo "$run_ids" | wc -l | tr -d ' ')
+              count=$((count + deleted))
+              total=$((total + deleted))
+
+              # Print progress
+              print_table_row "[$index/$n_workflows]" "$name" "$count" "$total"
 
               if [ "$MODE" = "execute" ]; then
-                for RUN_ID in $RUN_IDS; do
-                  gh run delete "$RUN_ID" --yes >/dev/null
+                for run_id in $run_ids; do
+                  gh run delete "$run_id" >/dev/null
                 done
               fi
+
+              [ "$has_next" != "true" ] && break
             done
 
-            echo "$workflow → $WF_COUNT runs"
-            TOTAL_AFFECTED=$((TOTAL_AFFECTED + WF_COUNT))
+            summary+=("$name|$count")
+          done
 
-          done <<< "$DELETED_WORKFLOWS"
+          # ================================================================================================
+          # Print a summary table
+          # ================================================================================================
+          print_table_header "Workflow Cleanup Summary"
+          print_summary_row "Workflow" "Runs"
+          print_table_separator
+          for entry in "${summary[@]}"; do
+            wf="${entry%%|*}"
+            count="${entry##*|}"
+            print_summary_row "$wf" "$count"
+          done
 
-          echo ""
-          echo "--------------------------------------"
-          echo "Total runs affected: $TOTAL_AFFECTED"
+          # ================================================================================================
+          # Print totals as a footer
+          # ================================================================================================
+          print_table_separator
+          print_summary_row "TOTAL" "$total"
 
-          if [ "$MODE" = "dry run" ]; then
+          if [ "$MODE" != "execute" ]; then
             echo "Dry run complete. No runs were deleted."
           else
             echo "Cleanup complete."
--- a/crates/store/Cargo.toml
+++ b/crates/store/Cargo.toml
@@ -31,7 +31,7 @@ miden-crypto               = { features = ["concurrent", "hashmaps"], workspace 
 miden-node-proto           = { workspace = true }
 miden-node-proto-build     = { features = ["internal"], workspace = true }
 miden-node-utils           = { workspace = true }
-miden-remote-prover-client = { workspace = true }
+miden-remote-prover-client = { features = ["block-prover"], workspace = true }
 miden-standards            = { workspace = true }
 # TODO remove `testing` from `miden-protocol`, required for `BlockProof::new_dummy`
 miden-protocol    = { features = ["std", "testing"], workspace = true }


crates/validator/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -0,0 +1,15 @@
...
+    account_id                   BLOB    NOT NULL,
+    block_num                    INTEGER NOT NULL, -- Block number in which the transaction was included.
+    initial_state_commitment     BLOB    NOT NULL, -- State of the account before applying the transaction.
+    final_state_commitment       BLOB    NOT NULL, -- State of the account after applying the transaction.
+    input_notes                  BLOB    NOT NULL, -- Serialized vector with the Nullifier of the input notes.
+    output_notes                 BLOB    NOT NULL, -- Serialized vector with the NoteId of the output notes.
+    size_in_bytes                INTEGER NOT NULL, -- Estimated size of the row in bytes, considering the size of the input and output notes.
+
+    PRIMARY KEY (transaction_id)
+) WITHOUT ROWID;
```
Mirko-von-Leipzig: Question: could this not be much simpler?

```suggestion
CREATE TABLE transactions (
    id   BLOB    NOT NULL,
    data BLOB NOT NULL,
    PRIMARY KEY (id)
) WITHOUT ROWID;
```

sergerad:
@bobbinth I have validated these behaviours on this branch:
- Validator as separate process works
- Validator going down does not stop Node, can serve read requests
- Node resumes normal functioning when Validator is brought back up

crates/validator/Cargo.toml
Mirko-von-Leipzig, line 1: We should make these workspace deps to ensure the versions stay aligned.

crates/validator/src/server/mod.rs
```diff
@@ -154,14 +149,29 @@ impl<S: BlockSigner + Send + Sync + 'static> api_server::Api for ValidatorServer
...
+            .db
+            .transact("insert_transaction", move |conn| {
+                insert_transaction(conn, &validated_tx_header)
+            })
             .await;
 
-        Ok(tonic::Response::new(()))
+        match result {
+            Ok(rows_affected) => {
+                if rows_affected == 1 {
```
Mirko-von-Leipzig: Do you feel like this check is necessary?

crates/store/src/db/mod.rs
Mirko-von-Leipzig, line 1: I don't like that t he `Validator` now depends on the `Store` like this.

An alternative approach is to create a more dedicated `Database` crate that provides these abstractions for the store and other components to use. We can make that part of the `store` general clean-up cc @drahnr wdyt?

crates/validator/src/db/models.rs
```diff
@@ -0,0 +1,30 @@
...
+            transaction_id: transaction_header.id().to_bytes(),
+            data: transaction_header.to_bytes(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Queryable, Selectable)]
+#[diesel(table_name = schema::transactions)]
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+pub struct TransactionSummaryRowSelect {
```
Mirko-von-Leipzig: I'm not sure where we would need this, will check back after reading the rest of the code.

bin/node/src/commands/mod.rs
```diff
@@ -47,7 +47,26 @@ fn duration_to_human_readable_string(duration: Duration) -> String {
...
+pub struct ValidatorConfig {
+    /// Insecure, hex-encoded validator secret key for development and testing purposes.
+    /// Only used when the Validator URL argument is not set.
+    #[arg(
+        long = "validator.insecure.secret-key",
+        env = ENV_VALIDATOR_INSECURE_SECRET_KEY,
+        value_name = "VALIDATOR_INSECURE_SECRET_KEY",
+        default_value = INSECURE_VALIDATOR_KEY_HEX
+    )]
+    validator_insecure_secret_key: String,
```
Mirko-von-Leipzig: I really think this can just be `validator.key`. The only people that need to worry about this is ourselves - for everyone else (aka _most_ users of this binary), this is a bit alarming/confusing.

bin/node/src/commands/bundled.rs
```diff
@@ -330,6 +319,28 @@ impl BundledCommand {
             component_ids.insert(id, "ntx-builder");
         }
 
+        // Start the Validator if we have bound a socket.
+        if let Some(address) = validator_socket_address {
+            let secret_key_bytes = hex::decode(validator.validator_insecure_secret_key)?;
+            let signer = SecretKey::read_from_bytes(&secret_key_bytes)?;
```
Mirko-von-Leipzig, line 314: I think its time we clean up our configuration. imo this check/conversion should be done _before_ we actually start any tasks etc.

As in, 

1. First parse and validate _all_ configuration properly, and
2. only then begin launching tasks

This is a larger refactor though, so for this PR we can just move on.

Something else to note is that `validator.url` and `validator.key` should be _mutually exclusive_ which we aren't enforcing at the moment.

bin/node/src/commands/validator.rs
```diff
@@ -40,6 +42,10 @@ pub enum ValidatorCommand {
         )]
         grpc_timeout: Duration,
 
+        /// Directory in which to store the database.
```
Mirko-von-Leipzig:
```suggestion
        /// Directory in which to store the validator's data.
```

crates/validator/src/block_validation/mod.rs
```diff
@@ -29,23 +49,51 @@ pub enum BlockValidationError {
...
+
+    // Retrieve all validated transactions pertaining to the proposed block.
+    let proposed_tx_ids =
+        proposed_block.transactions().map(TransactionHeader::id).collect::<Vec<_>>();
+    let query_tx_ids = proposed_tx_ids.clone();
+    let validated_transactions = db
+        .transact("select_transactions", move |conn| {
+            select_validated_transactions(conn, &query_tx_ids)
+        })
+        .await?;
```
Mirko-von-Leipzig: Maybe I've misunderstood, but I thought you should be able to simply check that each transaction in the proposed block:

1. has the correct ID, and
2. that the ID exists in the database

Do we really have to confirm that they match bit for bit?

crates/validator/src/block_validation/mod.rs
```diff
@@ -29,23 +49,51 @@ pub enum BlockValidationError {
...
+
+    // Validated transaction count must match proposed count exactly.
+    if validated_transactions.len() != proposed_tx_ids.len() {
```
Mirko-von-Leipzig: This error isn't very helpful imo. iiuc it means that at least one of the proposed transactions wasn't previously validated. However the error just indicates a length mismatch which

- tells us what equality check failed (but to the caller that's not really meaningful without reading the code here)
- doesn't tell us what data was bad

Instead the error should be something like

```rust
BlockValidationError::UnvalidatedTransactions(Vec<TransactionId>)
```

crates/validator/src/db/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use crate::COMPONENT;
+use crate::db::migrations::apply_migrations;
+use crate::db::models::{TransactionSummaryRowInsert, TransactionSummaryRowSelect};
+
+/// Open a connection to the DB and apply any pending migrations.
+#[instrument(target = COMPONENT, skip_all)]
+pub async fn load(database_filepath: PathBuf) -> Result<miden_node_store::Db, DatabaseSetupError> {
+    let manager = ConnectionManager::new(database_filepath.to_str().unwrap());
+    let pool = deadpool_diesel::Pool::builder(manager).max_size(16).build()?;
```
Mirko-von-Leipzig, line 24: Does `max_size` mean the capacity, or is the actual capacity chosen differently and this is just the upper bound to that algorithm?

crates/validator/src/db/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+        .filter(schema::transactions::transaction_id.eq_any(tx_id_bytes))
+        .order(schema::transactions::transaction_id.asc())
+        .load::<TransactionSummaryRowSelect>(conn)
+        .map_err(DatabaseError::from)?;
+
+    // Deserialize the transaction blobs.
+    let mut transactions: HashMap<TransactionId, TransactionHeader> = HashMap::new();
+    for raw_tx in raw_transactions {
+        let tx_id = TransactionId::read_from_bytes(&raw_tx.transaction_id)?;
+        let tx_header = TransactionHeader::read_from_bytes(&raw_tx.data)?;
```
Mirko-von-Leipzig: We should be storing the entire transaction or no?

crates/store/src/db/mod.rs
drahnr: Agree, we should have a DB utilities crate, keeping all of the plumbing separate from any schema and entity specific queries.

crates/validator/src/db/mod.rs
drahnr, line 24: I'd prefer to have our own builder pattern, with any options we use in practice.

crates/validator/src/db/migrations.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    let Err(e) = conn.run_pending_migrations(MIGRATIONS) else {
+        return Ok(());
+    };
+    tracing::warn!(target = COMPONENT, "Failed to apply migration: {e:?}");
+    conn.revert_last_migration(MIGRATIONS)
+        .expect("Duality is maintained by the developer");
+
+    Ok(())
+}
```
drahnr, line 25: I am not sure we want to duplicate this logic

crates/validator/src/db/schema.rs
```diff
@@ -0,0 +1,6 @@
+diesel::table! {
+    transactions (transaction_id) {
+        transaction_id -> Binary,
+        data -> Binary,
```
drahnr: Are we sure we don't want a k-v-store?

Mirko-von-Leipzig: I'm unsure if it matters much - ito code I think `HashMap<TransactionId, ProvenTransaction>` is sufficient. However it might be nice to be able to do things like query manually by account ID etc.. Eh :shrug: I assume we would use `RocksDB`? Or we could try out `fjall` for this.

drahnr: It doesn't really matter which we'd use, I am just asking if we expect more to be stored here that is "relational" or if that's it? If so, then I'd probably would not want to incur the diesel compile time / complexity.
It does make sense from a unification perspective.

Either way, no strong opinion, it'd be an opportunity to test drive `fjall`.

Mirko-von-Leipzig: Its pretty much just the hashmap I reckon.

crates/validator/src/block_validation/mod.rs
sergerad: probably unnecessary @bobbinth might have thoughts

bin/node/src/commands/mod.rs
sergerad: I think its more alarming to have a private key argument without being very explicit about it being insecure. This argument / env var should never be used in production scenarios. Happy to rename it as you say and just make sure the comments are explicit enough.

crates/validator/src/db/schema.rs
sergerad: @bobbinth wdyt about using k/v db instead?

bobbinth: For the current use case, I think KV is fine.

One feature that we may want to provide in the longer run (while still running with guardrails on) is the ability to give the user their state back - even for private accounts. This would be a "last resort" type of a solution if someone lost their state and didn't not do a proper backup (or use PSM). How this would work exactly, I'm not sure yet. For example:

- We could try to keep an up-to-date state for every private account. This would probably be an overkill though.
- We could keep just the transactions, and in case of an issue i described above, we'd just get all transactions for a given account and then replay them to get the current state. We'd need to build a separate "replay" tool for this, but shouldn't be too difficult.

In case we go with the latter solution, I think the KV store will mostly work too. The only thing we'd need to do is have the ability to query all transactions for a given account - but maybe that's doable too?

bobbinth: If we do go with SQL table, I'd structure it as `(transaction_id, account_id, tx_summary)` with:

- `transaction_id` being a primary key.
- Having an index on `account_id`.
- `tx_summary` containing a serialized version of `TransactionSummary` struct for the transaction.

We could also store the transaction proof and other transaction data - but I'm not sure that's needed.

sergerad: I have pushed up k/v impl to consider. Have removed the logic I had that was comparing proposed vs validated transaction information. Need some direction on what, if anything, we want to compare there.

crates/validator/src/db/database.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        Ok(Self { _db: db, keyspace })
+    }
+
+    /// Stores a transaction header into the database.
+    pub fn put(&self, validated_tx_header: &TransactionHeader) -> Result<(), DatabaseError> {
+        self.keyspace
+            .insert(validated_tx_header.id().to_key(), validated_tx_header.to_value())?;
+        Ok(())
+    }
```
bobbinth: I think we probably want to store more than just a `TransactionHeader`. At the very least, I think we should also store the account delta of the account in question.

crates/validator/src/db/database.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// Retrieves a transaction headers corresponding to the transaction ids from the database.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - Any requested transactions are not found.
+    /// - Retrieved transactions cannot be deserialized.
+    pub fn get(&self, ids: &[TransactionId]) -> Result<Vec<TransactionHeader>, DatabaseError> {
```
bobbinth: Would it be possible to ad a method to also get all account deltas by account ID? I think we'll need this in the longer term, and ideally, we wouldn't need to re-do this work again later on.

crates/validator/src/block_validation/mod.rs
```diff
@@ -29,25 +27,13 @@ pub enum BlockValidationError {
...
-                tx_id,
-                proposed_block.block_num(),
-            ));
-        }
-    }
+    // Retrieve all validated transactions pertaining to the proposed block.
+    let proposed_tx_ids =
+        proposed_block.transactions().map(TransactionHeader::id).collect::<Vec<_>>();
+    // TODO(currentpr): If we don't need to retrieve the data at all we can change this.
+    let _validated_transactions = db.get(&proposed_tx_ids)?;
```
bobbinth: I don't think we actually care about the retrieving transaction details here - what we want to check is that all transactions that are included in the block have been previously validated. This could be done even at the DB level - e.g., `db.contains_transactions(&proposed_tx_ids)`.

crates/validator/src/db/database.rs
sergerad: we could have another keyspace for account deltas. all we can do with the k/v store is duplicate data every time we want to retrieve it by a different key. if we plan on having relational queries then k/v is not a good idea

crates/validator/src/db/mod.rs
sergerad: An upper bound but I don't know what goes on under the hood. @drahnr do you know why we wouldn't just use default?

```rust
    /// Maximum size of the [`Pool`].
    ///
    /// Default: `cpu_count * 4`
    ///
    /// [`Pool`]: super::Pool
    pub max_size: usize,
```

crates/validator/src/db/migrations.rs
sergerad: I believe this and its usage precludes us from re-using the one in the store. 
```rust
pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!("src/db/migrations");
```

If I try do it non-const, then it gets moved and can't be re-used in the fn like it is. The returned type doesn't impl `Clone` or `Copy`. I could just repeat that `embed_migrations!()` macro every time we need an instance?

sergerad: > I could just repeat that embed_migrations!() macro every time we need an instance?

doesn't work because of the macro's lifetime requirements

crates/validator/src/db/schema.rs
sergerad: moved back to sqlite

bin/node/Dockerfile
Mirko-von-Leipzig, line 1: Could you revert please, I think the indentation here was on purpose to show its args as part of the prior command

Mirko-von-Leipzig, line 1: Could you revert this? Indentation is supposed to show they're args to the `install` cmd.

crates/block-producer/src/server/tests.rs
```diff
@@ -48,6 +48,7 @@ async fn block_producer_startup_is_robust_to_network_failures() {
             address: validator_addr,
             grpc_timeout,
             signer: SecretKey::random(),
+            data_directory: "/tmp/".into(),
```
Mirko-von-Leipzig: I think we have `tempdir` as a dependency somewhere

crates/validator/src/block_validation/mod.rs
```diff
@@ -1,52 +1,47 @@
...
+use crate::db::find_unvalidated_transactions;
 
 // BLOCK VALIDATION ERROR
 // ================================================================================================
 
 #[derive(thiserror::Error, Debug)]
 pub enum BlockValidationError {
-    #[error("transaction {0} in block {1} has not been validated")]
-    TransactionNotValidated(TransactionId, BlockNumber),
+    #[error("found unvalidated transactions {0:?}")]
```
Mirko-von-Leipzig:
```suggestion
    #[error("block contains unvalidated transactions {0:?}")]
```

crates/validator/src/server/mod.rs
```diff
@@ -181,11 +174,9 @@ impl<S: BlockSigner + Send + Sync + 'static> api_server::Api for ValidatorServer
 
         // Validate the block.
         let signature =
-            validate_block(proposed_block, &self.signer, self.validated_transactions.clone())
-                .await
-                .map_err(|err| {
-                    tonic::Status::invalid_argument(format!("Failed to validate block: {err}",))
-                })?;
+            validate_block(proposed_block, &self.signer, &self.db).await.map_err(|err| {
+                tonic::Status::invalid_argument(format!("Failed to validate block: {err}",))
```
Mirko-von-Leipzig: This only adds the top level error which is insufficient.

```suggestion
                tonic::Status::invalid_argument(format!("Failed to validate block: {}", err.as_report()))
```

crates/validator/src/db/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+        return Ok(HashMap::new());
+    }
+
+    // Convert TransactionIds to bytes for query.
+    let tx_id_bytes: Vec<Vec<u8>> = tx_ids.iter().map(TransactionId::to_bytes).collect();
+
+    // Query the database for matching transactions.
+    let raw_transactions = schema::transactions::table
+        .filter(schema::transactions::id.eq_any(tx_id_bytes))
+        .order(schema::transactions::id.asc())
```
Mirko-von-Leipzig: Why do we need to order them?

crates/validator/src/db/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+pub(crate) fn select_validated_transactions(
+    conn: &mut SqliteConnection,
+    tx_ids: &[TransactionId],
+) -> Result<HashMap<TransactionId, TransactionSummary>, DatabaseError> {
+    if tx_ids.is_empty() {
+        return Ok(HashMap::new());
+    }
+
+    // Convert TransactionIds to bytes for query.
+    let tx_id_bytes: Vec<Vec<u8>> = tx_ids.iter().map(TransactionId::to_bytes).collect();
```
Mirko-von-Leipzig: We should have a sql implementation for this or no? @drahnr

crates/validator/src/db/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+) -> Result<usize, DatabaseError> {
+    let row = TransactionSummaryRowInsert::new(tx_id, summary);
+    let count = diesel::insert_into(schema::transactions::table).values(row).execute(conn)?;
+    Ok(count)
+}
+
+/// Retrieves validated transactions from the database.
+#[allow(dead_code)]
+pub(crate) fn select_validated_transactions(
+    conn: &mut SqliteConnection,
```
Mirko-von-Leipzig: What is this function for?

crates/validator/src/db/mod.rs
drahnr: We don't,  my intent was to move all conversions to `conv.rs` for easier tracking/coherence. Eventually we might want to move to some diesel based traits, skipping one level, which the `conv.rs` contained traits should have auto trait impls for, eventually.
tl;dr no, not as of today

Cargo.toml
```diff
@@ -80,6 +80,11 @@ lru               = { default-features = false, version = "0.16" }
 pretty_assertions = { version = "1.4" }
 # breaking change `DecodeError::new` is not exposed anymore
 # but is assumed public by some internal dependency
+deadpool           = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
```
bobbinth: Is this comment still relevant? If so, it is probably not for the `deadpool` dependency - right?

crates/validator/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE transactions (
+    id   BLOB NOT NULL,
+    account_id   BLOB NOT NULL,
+    summary BLOB NOT NULL,
+    PRIMARY KEY (id)
+) WITHOUT ROWID;
```
bobbinth: nit: let's align data types (as we do in other SQL files).

Also, would be great to add comments - especially for the `summary` field to indicate what is actually contained there.

crates/validator/src/tx_validation/mod.rs
```diff
@@ -56,7 +61,12 @@ pub async fn validate_transaction(
     let executed_tx_header: TransactionHeader = (&executed_tx).into();
     let proven_tx_header: TransactionHeader = (&proven_tx).into();
     if executed_tx_header == proven_tx_header {
-        Ok(executed_tx_header)
+        let account_delta = executed_tx.account_delta().clone();
+        let input_notes = executed_tx.input_notes().clone();
+        let output_notes = executed_tx.output_notes().clone();
+        let salt = Word::empty();
+        let summary = TransactionSummary::new(account_delta, input_notes, output_notes, salt);
+        Ok(summary)
```
bobbinth: Maybe this is fine for now, but I'm actually not sure that `TransactionSummary` is the right struct here:

1. `salt` field doesn't really have a meaning here.
2. We probably do want to save initial and final account state commitment.

Maybe we create a new struct for this (something like `ValidatedTransactionInfo`) and include the fields we want?

crates/validator/src/db/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+        "Connected to the database"
+    );
+
+    let db = miden_node_store::Db::new(pool);
+    db.query("migrations", apply_migrations).await?;
+    Ok(db)
+}
+
+/// Inserts a new validated transaction into the database.
+pub(crate) fn insert_transaction(
```
bobbinth, line 39: For this and other functions here, could we add the SQL statements that they correspond to in the comments?

crates/validator/Cargo.toml
```diff
@@ -18,8 +18,12 @@ workspace = true
 
 [dependencies]
 anyhow                 = { workspace = true }
+deadpool-diesel        = { workspace = true }
+diesel                 = { workspace = true }
+diesel_migrations      = { workspace = true }
 miden-node-proto       = { workspace = true }
 miden-node-proto-build = { features = ["internal"], workspace = true }
+miden-node-store       = { workspace = true }
```
bobbinth, line 26: I don't think validator should depend on the store. If there is some common functionality needed by both, I think we should move this to the utils crate (or maybe a new crate).

bin/node/src/commands/mod.rs
```diff
@@ -47,7 +47,26 @@ fn duration_to_human_readable_string(duration: Duration) -> String {
...
+        env = ENV_VALIDATOR_KEY,
+        value_name = "VALIDATOR_KEY",
+        default_value = INSECURE_VALIDATOR_KEY_HEX
+    )]
+    validator_key: String,
+
+    /// The remote Validator's gRPC URL. If unset, will default to running a Validator
+    /// in-process. If set, the insecure key argument is ignored.
+    #[arg(long = "validator.url", env = ENV_VALIDATOR_URL, value_name = "URL")]
+    pub validator_url: Option<Url>,
```
bobbinth: Should we enforce that these are mutually-exclusive parameters somehow?

Also, why is `validator_url` `pub` but `validator_key` isn't?

bin/node/src/commands/store.rs
```diff
@@ -48,12 +48,12 @@ pub enum StoreCommand {
...
-            long = "validator.insecure.secret-key",
-            env = ENV_VALIDATOR_INSECURE_SECRET_KEY,
-            value_name = "VALIDATOR_INSECURE_SECRET_KEY",
+            long = "validator.key",
+            env = ENV_VALIDATOR_KEY,
+            value_name = "VALIDATOR_KEY",
             default_value = INSECURE_VALIDATOR_KEY_HEX
         )]
-        validator_insecure_secret_key: String,
+        validator_key: String,
```
bobbinth, line 58: I don't remember: why does the store need to know `validator_key`?

bobbinth (commented):

bin/node/Dockerfile
sergerad: Its based on a formatter. Think its better we align on a format rather than have manual formatting (assuming that's what has happened with this indentation). Otherwise if you have format config you use I can apply the same to my IDE.

Cargo.toml
sergerad: @drahnr

crates/validator/src/db/mod.rs
sergerad: Updated but noting that we don't seem to add raw sql comment for inserts so have left it out for insert_transaction

crates/validator/Cargo.toml
sergerad: Yea planning on doing a followup PR to address that
https://github.com/0xMiden/miden-node/issues/1653

bin/node/src/commands/store.rs
sergerad: For the bootstrap command which signs the genesis block.

sergerad: Added a comment line to explain

bin/node/Dockerfile
Mirko-von-Leipzig: I mean its a docker formatter that's formatting multi-line bash so I'm not sure there is such a thing :) I don't have anything auto-save on save - at the very least I would recommend setting it so it only formats things you've actually touched? aka diffs.

Mirko-von-Leipzig: As is, you're adding deltas to things that weren't changed

Cargo.toml
drahnr: The comment should be tagged to `prost`.

crates/validator/src/db/mod.rs
drahnr: No, I don't remember when or who introduced it

Cargo.toml
drahnr: After some investigation, https://github.com/0xMiden/miden-node/pull/1658

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - [BREAKING] Move block proving from Blocker Producer to the Store ([#1579](https://github.com/0xMiden/miden-node/pull/1579)).
 - [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/miden-node/pull/1595)).
+- Added sqlite database to Validator for transaction persistence ([#1614](https://github.com/0xMiden/miden-node/pull/1614)).
```
Mirko-von-Leipzig: I'd like us to get into the habit of writing these as if for _external_ readers. For example, they don't care _what_ we use as a database.

```suggestion
- Validator now persists validated transactions ([#1614](https://github.com/0xMiden/miden-node/pull/1614)).
```

crates/validator/src/tx_validation/info.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// The data of [`ValidatedTransactionInfo`] which is serialized into a byte array and stored as a
+/// column in the `validated_transactions` table.
+struct ValidatedTransactionInfoBlob {
+    initial_account_state_commitment: Word,
+    final_account_state_commitment: Word,
+    account_delta: AccountDelta,
+    input_notes: InputNotes<InputNote>,
+    output_notes: OutputNotes,
+}
```
Mirko-von-Leipzig: I imagine this structure is derived from the database needs?

I'll revisit this after reading elsewhere, but is there a reason to pack/unpack this instead of just `pub struct ValidatedTransaction(ExcecutedTransaction)`.

------

edit -- okay I think we should simplify this as per above.

crates/validator/src/tx_validation/info.rs
```diff
@@ -0,0 +1,101 @@
...
+        self.blob.account_delta.id()
+    }
+
+    /// Returns the binary representation of the transaction info.
+    pub fn to_bytes(&self) -> Vec<u8> {
+        self.blob.to_bytes()
+    }
+}
+
+impl From<ExecutedTransaction> for ValidatedTransactionInfo {
```
Mirko-von-Leipzig: I don't think this should exist. This should _only_ be constructable by something that determines that this _is a validated_ transaction.

crates/validator/src/server/mod.rs
```diff
@@ -150,17 +145,14 @@ impl<S: BlockSigner + Send + Sync + 'static> api_server::Api for ValidatorServer
...
-        let tx_id = validated_tx_header.id();
-        self.validated_transactions
-            .put(tx_id, validated_tx_header)
-            .instrument(info_span!("validated_txs.insert"))
-            .await;
-
+        // Store the validated transaction.
+        self.db
+            .transact("insert_transaction", move |conn| insert_transaction(conn, &tx_info))
+            .await?;
```
Mirko-von-Leipzig, line 157: We should ensure that this is instrumented (it should be, but async + closures is tricky)

crates/validator/src/db/models.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+impl ValidatedTransactionInfoRowInsert {
+    pub fn new(info: &ValidatedTransactionInfo) -> Self {
+        Self {
+            id: info.tx_id().to_bytes(),
+            block_num: info.block_num().to_raw_sql(),
+            account_id: info.account_id().to_bytes(),
+            info: info.to_bytes(),
+        }
```
Mirko-von-Leipzig: I think we should handle all the serialization here, instead of serializing in `VAlidatedTransactionInfo` as I allude to in [this comment](https://github.com/0xMiden/miden-node/pull/1614/changes#r2797297760)

crates/validator/src/block_validation/mod.rs
```diff
@@ -1,52 +1,47 @@
...
 #[derive(thiserror::Error, Debug)]
 pub enum BlockValidationError {
-    #[error("transaction {0} in block {1} has not been validated")]
-    TransactionNotValidated(TransactionId, BlockNumber),
+    #[error("block contains unvalidated transactions {0:?}")]
+    UnvalidatedTransactions(Vec<TransactionId>),
     #[error("failed to build block")]
     BlockBuildingFailed(#[from] ProposedBlockError),
+    #[error("failed to select transactions")]
+    DatabaseError(#[from] DatabaseError),
```
Mirko-von-Leipzig:
```suggestion
    DatabaseError(#[source] DatabaseError),
```

crates/validator/src/db/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+    // Convert TransactionIds to bytes for query.
+    let tx_id_bytes: Vec<Vec<u8>> = tx_ids.iter().map(TransactionId::to_bytes).collect();
+
+    // Query the database for matching transactions ids.
+    let raw_transaction_ids = schema::validated_transactions::table
+        .select(schema::validated_transactions::id)
+        .filter(schema::validated_transactions::id.eq_any(tx_id_bytes))
+        .order(schema::validated_transactions::id.asc())
+        .load::<Vec<u8>>(conn)
+        .map_err(DatabaseError::from)?;
```
Mirko-von-Leipzig: This is probably not important at the moment, but since this will likely _always_ remain a sqlite local database, I think we should just do an `exists` check in a loop? This would simplify the query, return values and output construction.

Though whether diesel supports this is unclear to me.

crates/validator/src/block_validation/mod.rs
```diff
@@ -1,52 +1,47 @@
...
+    let proposed_tx_ids =
+        proposed_block.transactions().map(TransactionHeader::id).collect::<Vec<_>>();
+    let unvalidated_transactions = db
+        .transact("find_unvalidated_transactions", move |conn| {
+            find_unvalidated_transactions(conn, &proposed_tx_ids)
+        })
+        .await?;
+
+    // All proposed transactions must have been validated.
+    if !unvalidated_transactions.is_empty() {
```
Mirko-von-Leipzig: I think we're really taking the variable names a bit far (in general, not specific to this PR). Its starting to feel like Hungarian notation.

We already have types that tell us what the thing is, and the rest should be inferrable by context. Brevity is also a good thing, aka I'd find this easier to read:


```suggestion
    let txs = block.transactions().map(TransactionHeader::id).collect::<Vec<_>>();
    let unvalidated = db
        .transact("find_unvalidated_transactions", move |conn| {
            find_unvalidated_transactions(conn, &txs)
        })
        .await?;

    // All proposed transactions must have been validated.
    if !unvalidated.is_empty() {
```

crates/validator/src/tx_validation/info.rs
Mirko-von-Leipzig: I think an additional question - why not just store the entire `ExecutedTransaction` as is in the `info` field in the database?

bin/node/src/commands/bundled.rs
```diff
@@ -331,6 +308,28 @@ impl BundledCommand {
             component_ids.insert(id, "ntx-builder");
         }
 
+        // Start the Validator if we have bound a socket.
+        if let Some(address) = validator_socket_address {
```
drahnr, line 312: The configuration decision based on the `validator_socket_address` is a bit obfuscating the fact that the user did not provide a validator URL but a key

crates/validator/src/db/migrations.rs
drahnr: Fair

crates/validator/src/tx_validation/info.rs
sergerad: The only way the Validator gets an executed tx is by re-executing the proposed tx passed to it. So this should be satisfied:
> only be constructable by something that determines that this is a validated transaction.

bin/node/src/commands/bundled.rs
sergerad: Not sure that part of the stack should be considering the key. Key presence is considered elsewhere. This part of the stack just cares about whether to run the local instance or not

crates/validator/src/block_validation/mod.rs
sergerad: Unfortunately I have the exact opposite view haha. Always find it quicker and easier to reason about a snippet when the variables are more explicitly named.

sergerad: `unvalidated_transactions` is extreme though, `unvalidated` is ok, maybe `unvalidated_txs` would be my preference

sergerad: (always abbreviate when it doesn't introduce ambiguity is a good rule)

crates/validator/src/db/models.rs
sergerad: LMK if my latest changes reflect your thinking here

crates/validator/src/db/mod.rs
sergerad: Unsure how many transactions we expect to have in the average block, but feels counterintuitive to scale # of db queries with # of transactions.

crates/validator/src/server/mod.rs
sergerad:
```
2026-02-12T22:43:49.934198Z  INFO rpc:validate_block:find_unvalidated_transactions: miden-validator: crates/validator/src/db/mod.rs:68: close, time.busy: 11.1ms, time.idle: 449µs rpc.service: "validator.Api", rpc.method: "SignBlock", otel.name: "validator.Api/SignBlock" tx_ids: [0x18f8ad1c9312257fe192507336a4b6590d2f0a92127506eb47738c77b18c258b, 0x7c9fed92d7bd767796a7d51f8e9a368bbf2b17a601ff471aa07dfc88bee7dcd1]
```

sergerad:
```
2026-02-12T22:55:33.672764Z  INFO rpc:submit_proven_transaction:insert_transaction: miden-validator: crates/validator/src/db/mod.rs:39: close, time.busy: 6.56ms, time.idle: 35.0µs rpc.service: "validator.Api", rpc.method: "SubmitProvenTransaction", otel.name: "validator.Api/SubmitProvenTransaction" tx_id: 0x99eeffc42a2b6150e6e7fed88f86350e926fd52992911b02bd4162abec0e8670
```

crates/validator/src/db/mod.rs
sergerad: Unsure which builder we would be adding here. Is it something that would be used across both store and validator? Could roll it up into this issue #1653

crates/validator/src/db/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+
+use crate::COMPONENT;
+use crate::db::migrations::apply_migrations;
+use crate::db::models::ValidatedTransactionRowInsert;
+use crate::tx_validation::ValidatedTransaction;
+
+/// Open a connection to the DB and apply any pending migrations.
+#[instrument(target = COMPONENT, skip_all)]
+pub async fn load(database_filepath: PathBuf) -> Result<miden_node_store::Db, DatabaseSetupError> {
+    let manager = ConnectionManager::new(database_filepath.to_str().unwrap());
```
sergerad, line 23: @Mirko-von-Leipzig @drahnr despite this (and the PRAGMA queries it makes for WAL etc) I still get db locked errors when running the integration tests.

```
ERROR rpc:submit_proven_transaction: miden-validator: crates/validator/src/server/mod.rs:126: 
error: status: 'Internal error', self: "database is locked" rpc.service: 
"validator.Api", rpc.method: "SubmitProvenTransaction", otel.name: "validator.Api/SubmitProvenTransaction"
```

Validator could receive parallel requests for `submit_proven_transaction` endpoint so I do think its something we need to solve (not just integration tests' fault).

sergerad: Oddly happens even after adding
```rust
    // Wait up to 5 seconds for writer locks before erroring.
    diesel::sql_query("PRAGMA busy_timeout=5000")
        .execute(conn)
        .map_err(ConnectionManagerError::ConnectionParamSetup)?;
```

sergerad: Adding this does avoid it. ATM I'm reusing the store's connection manager code for validator.

```rust
pub(crate) fn configure_connection_on_creation(
    conn: &mut SqliteConnection,
) -> Result<(), ConnectionManagerError> {
    // Wait up to 5 seconds for writer locks before erroring.
    diesel::sql_query("PRAGMA busy_timeout=5000")
        .execute(conn)
        .map_err(ConnectionManagerError::ConnectionParamSetup)?;
```

crates/validator/src/db/mod.rs
Mirko-von-Leipzig: > but feels counterintuitive to scale # of db queries with # of transactions.

I understand but with sqlite being an embedded database there isn't much difference between looping in rust code, or looping inside the sql code. The downside of the latter is that you cannot (trivially) just do an exists query so you're copying the bytes instead of just a bool.

crates/validator/src/db/mod.rs
drahnr: I think we need to serialize writes. By default you get an error with DB locked if a write is already in process, it's application level responsiblity to deal with this. The timeout fixes this on the surface, since almost all queries will finish under 5 seconds, only under load you'll see the locked again.
I suggest to add an internal bounded channel to serialize the queries (at least the write).

sergerad: Adding a bounded channel would effectively add another "queue" on top of the one in the database only for the benefit of controlling capacity via the size of the channel (instead of timeout seconds or retry count). In both situations we would error out on the basis of too many writes (no difference for the client of the corresponding endpoint(s)).

I agree notionally about serializing writes, but I'm not sure we are actually better off adding the necessary code for the channel rather than relying on a timeout or retry count.

bin/node/src/commands/bundled.rs
sergerad: @Mirko-von-Leipzig please remind me do we have an issue for cleaning up the bundled config/startup stuff?

crates/validator/src/db/mod.rs
drahnr: The main difference is retaining control and the error information we can provide, which to me, given our latest debug adventures, is well worth the additional 100 (fair, 200) loc for queue management.
I see myself debugging something unrelated 6 months down the line, where all we get is a `DB locked`-error again, and the root cause being _somewhere_, but now we also not disambiguate _overload_ based `DB locked` from bug-rooted `DB locked`.

crates/validator/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE validated_transactions (
+    id            BLOB NOT NULL,
+    block_num     INTEGER NOT NULL,
+    account_id    BLOB NOT NULL,
+    "transaction" BLOB NOT NULL, -- Binary encoded ExecutedTransaction.
```
bobbinth, line 5: Storing full executed transaction will quickly add up as these contain proofs and each proof could be close to 100KB. We should change this to store just the relevant info. I'd probably split this up to store:

- account delta
- input note info (need to define what exactly gets stored)
- output note info (need to define what exactly gets stored)
- maybe some other metadata (e.g., initial account state hash, final account state hash, transaction fee).

@sergerad or @Mirko-von-Leipzig - let's create an issue for this.

sergerad: https://github.com/0xMiden/node/issues/1715

PR #1616, opened 2026-01-29
feat: load standard note scripts at `Db::load`

mmagician:
On `Db::load()`, `INSERT OR IGNORE` note scripts present in `miden-standards`, by listing all the `StandardNote` enum variants and inserting their root<>script pairs.

Will be simplified slightly and future-proofed once https://github.com/0xMiden/miden-base/pull/2367 lands.

closes https://github.com/0xMiden/miden-base/issues/2418

### DIFF
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -226,7 +226,7 @@ impl From<NoteRecord> for NoteSyncRecord {
 ///
 /// NOTE: This list must be kept in sync with `miden_standards::note::WellKnownNote` variants.
 /// If `WellKnownNote` adds new variants in the future, they should be added here.
-/// TODO: refactor once (https://github.com/0xMiden/miden-base/pull/2367) lands
+/// TODO: refactor once (<https://github.com/0xMiden/miden-base/pull/2367>) lands
 pub(crate) fn all_well_known_note_types() -> [WellKnownNote; 5] {
     [
         WellKnownNote::P2ID,
@@ -243,7 +243,10 @@ pub(crate) fn all_well_known_note_types() -> [WellKnownNote; 5] {
 /// inserts them into the `note_scripts` table. This allows the node to immediately support
 /// network transactions that output these standard note types.
 fn insert_well_known_note_scripts(conn: &mut SqliteConnection) -> Result<(), DatabaseError> {
-    let scripts: Vec<_> = all_well_known_note_types().iter().map(|n| n.script()).collect();
+    let scripts: Vec<_> = all_well_known_note_types()
+        .iter()
+        .map(miden_standards::note::WellKnownNote::script)
+        .collect();
     let count = queries::insert_note_scripts(conn, &scripts)?;
 
     info!(


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.13.4 (TBD)
+
+### Enhancements
+
+- Adapted store to automatically load standard note scripts from `miden_standards` on startup ([#1616](https://github.com/0xMiden/miden-node/pull/1616)).
```
Mirko-von-Leipzig: Maybe

```suggestion
- Well known note scripts from `miden_standards` are now automatically loaded ([#1616](https://github.com/0xMiden/miden-node/pull/1616)).
```

crates/store/Cargo.toml
```diff
@@ -54,6 +54,7 @@ miden-protocol        = { default-features = true, features = ["testing"], works
 miden-standards       = { features = ["testing"], workspace = true }
 rand                  = { workspace = true }
 regex                 = { version = "1.11" }
+tempfile              = { version = "3" }
```
Mirko-von-Leipzig: We use this in several other toml's now. Could you convert this to a workspace dep so we can keep them in sync?

crates/store/src/db/tests.rs
```diff
@@ -2281,3 +2281,49 @@ fn test_note_metadata_with_attachment_roundtrip() {
...
+    use tempfile::NamedTempFile;
+
+    use crate::errors::DatabaseError;
+    use crate::genesis::GenesisState;
+
+    let db_file = NamedTempFile::new().unwrap();
+    let db_path = db_file.path().to_path_buf();
+
+    // Bootstrap the database with a minimal genesis block
+    let genesis_state = GenesisState::new(vec![], test_fee_params(), 1, 0, SecretKey::random());
```
Mirko-von-Leipzig: An alternative is to use the genesis config's default, which would be less likely to need changing whenever the genesis state is changed. Though this is a minor thing.

```suggestion
    let (genesis_state, _) = GenesisConfig::default().into_state(SecretKey::random()).unwrap();
```

crates/store/src/db/tests.rs
```diff
@@ -2281,3 +2281,49 @@ fn test_note_metadata_with_attachment_roundtrip() {
...
+    // Load the database - this should insert standard note scripts
+    let db = crate::db::Db::load(db_path).await.unwrap();
+
+    // Verify all standard note scripts are present
+    for note_type in all_standard_note_types() {
+        let script = note_type.script();
+        let script_root = note_type.script_root();
+
+        // Verify the script_root() method matches script().root()
+        assert_eq!(script.root(), script_root, "script_root() should match script().root()");
```
Mirko-von-Leipzig: This feels a bit like ceremony - does this need testing?

crates/store/src/db/tests.rs
```diff
@@ -2281,3 +2281,49 @@ fn test_note_metadata_with_attachment_roundtrip() {
...
+            .query("get script", move |conn| queries::select_note_script_by_root(conn, script_root))
+            .await;
+
+        let retrieved_script = retrieved_script.unwrap();
+        assert!(
+            retrieved_script.is_some(),
+            "Standard note script should be present after Db::load()"
+        );
+        assert_eq!(retrieved_script.unwrap(), script, "Retrieved script should match original");
+    }
```
Mirko-von-Leipzig, line 2329:
```suggestion
        // Query the database for the script
        let retrieved_script = db
            .select_note_script_by_root(script_root)
            .await
            .unwrap()
            .expect(format!("Standard note {script_root} should be present after Db::load());

        assert_eq!(retrieved_script, script, "Retrieved script should match original");
    }
```

crates/store/src/db/mod.rs
```diff
@@ -321,6 +372,11 @@ impl Db {
 
         let me = Db { pool };
         me.query("migrations", apply_migrations).await?;
+
+        // Insert any standard note scripts. This ensures that nodes upgrading to a new version of
+        // `miden_standards` will automatically load any new standard note types.
+        me.query("standard note scripts", insert_standard_note_scripts).await?;
```
Mirko-von-Leipzig: While we're here, do you mind renaming `me` to `this`?

crates/store/src/db/mod.rs
```diff
@@ -218,6 +219,55 @@ impl From<NoteRecord> for NoteSyncRecord {
...
+        StandardNote::BURN,
+    ]
+}
+
+/// Inserts all standard note scripts into the database.
+///
+/// This function loads the standard note scripts (P2ID, BURN, etc.) from `miden_standards` and
+/// inserts them into the `note_scripts` table. This allows the node to immediately support
+/// network transactions that output these standard note types.
+fn insert_standard_note_scripts(conn: &mut SqliteConnection) -> Result<(), DatabaseError> {
```
Mirko-von-Leipzig: I guess we need this because the existing script insertion function takes a note instead of a note script?

I think we can refactor the existing function to take an iterator of note scripts instead. This function is used while inserting a block, and otherwise only in tests to ensure the script exists before we use it. We can instead call `insert_note_script` as a call in `insert_notes` which means we can remove it from tests and the insert block fn.

I'd prefer to just have a single place where we do the actual insertion.

crates/store/src/db/mod.rs
```diff
@@ -319,9 +360,14 @@ impl Db {
...
 
-        let me = Db { pool };
-        me.query("migrations", apply_migrations).await?;
-        Ok(me)
+        let this = Db { pool };
+        this.query("migrations", apply_migrations).await?;
+
+        // Insert any standard note scripts. This ensures that nodes upgrading to a new version of
+        // `miden_standards` will automatically load any new standard note types.
+        this.query("well-known note scripts", insert_well_known_note_scripts).await?;
```
bobbinth, line 368: There is a different way to do this: instead of pre-loading standard note scripts into the database, we could instead update our `TransactionExecutorHost` to keep a cache of standard note scripts. Then, in `TransactionExecutorHost::on_note_script_requested()` we'd first check if a note script is a standard script, and if not, only then we'd make the request to the data store.

The main advantage if this approach is that for querying standard note scripts we'd be saving a database trip.

Mirko-von-Leipzig: This would happen in `miden-protocol`?

bobbinth: In `miden-tx`.

Mirko-von-Leipzig: @mmagician @PhilippGackstatter how do you feel about this proposal? This would have the benefit of applying to all execution everywhere.

PhilippGackstatter: Generally sounds fine to me.

I assume this will make memory usage a bit worse, since all standard scripts will be in memory rather than looked up from DB, but that's probably fine.

The other negligible concern I have is that if we do this, `miden-tx` will have the responsibility of loading all possible versions of `StandardsLib` into that internal cache and thus have a dependency on all versions of `miden-standards` beyond 1.0. I don't yet know exactly what upgrading miden::standards will look like, but I assume it will essentially be append-only to not change any stabilized MAST roots, and in that case, there will only ever be one `StandardsLib`, i.e. all future versions are supersets of the earlier ones.

So, I think this should be okay to do.

drahnr: Assuming we will have versioned standard note scripts, we'd need an increasing number of hardwired cache items. I _think_ this works given a very low growth rate of versions, but it's just a hypothesis.

Issue #1429, opened 2025-12-05
Genesis config: custom account types


mmagician:
As part of the AggLayer integration efforts, we would like to have an ability to bootstrap the node with [`AggLayerBridge`](https://github.com/0xMiden/miden-base/issues/1889) account in the genesis.
This would be a network account with a few items in storage, such as the `bridge_owner` (another account that's deployed at genesis time) - probably a multisig.
We further need the bridge contract's registry to point to a few `AggLayerFaucet` accounts (each with its own configuration). The faucet registry is another storage entry. The `AggLayerFaucet` accounts should also be deployed.

We could take either a very specialized approach here and add support for the three account types:
- `AggLayerBridge`
- `MultisigWallet`
- `AggLayerFaucet`

With one custom account type, a specialized approach might have worked. But IMO since we have three account types that we care about, it's better to have infrastructure in place that allows for specifying any custom account type at genesis.

bobbinth:
We do have an ability to instantiate account components from account package + initial storage data (see [here](https://github.com/0xMiden/miden-base/blob/next/crates/miden-objects/src/account/component/mod.rs#L138)). Account package can be read from a file package file, and the initial storage data can be read from a TOML file (this structure is currently being somewhat refactored by @igamigo and @mmagician).

But I think there are a couple of things missing:

- We don't have a declarative way to specify the full account - though, maybe just providing a list of packages and a single initial storage data file would work here, and this could be part of the genesis TOML file.
- We don't have a way of "templating" dependencies between accounts. That is, creating a bunch of independent accounts should be pretty easy, but I'm not sure yet how we'd specify that there is a bridge account, and it is owned by a multisig account. The main difficulty here is that to initialize the storage of the bridge account we need to know the ID owner account, but the ID of the owner would be computed only on account instantiation.

------

An alternative approach could be to generate account files (i.e., `.mac` files) separately using some build script (i.e., Rust code), and, in the genesis config allow to specify these files. My intuition is that this would be the simplest way to do it.

bobbinth:
I think this works!

> optionally, we could make the features marked "*" only under `cfg(dev)` or similar feature flag, or remove them altogether. But I do think it's convenient to keep them around for local testing etc.

Agreed - it's good to have for local testing. Whether or not to make them optional, I think really depends on whether it introduces significant complexity. If not much complexity, disabling these via some flag (even a runtime flag) is probably a good idea.

[...]

PR #1624, opened 2026-01-31
Read general account files at genesis

mmagician:
Changes:
- rename `NativeFaucet` -> `NativeFaucetConfig`, and make it an enum with `Parameters` and `Account` variants
- change `GenesisConfig::read_toml(toml_content)` to `read_toml_path(toml_path)`, 
    - and make it take `Path` instead, so we can lookup the directory and read in account files accordingly, which are referenced relative to the config file
- add a sample config `02-with-account-files.toml` 
    - it references 3 `.mac` files that correspond to AggLayer accounts. The `.mac` files are also committed
    - add a unit test that processes this sample config
    - and a `build.rs` file in `config::store` module that auto-generates these `AccountFile`s 

closes https://github.com/0xMiden/miden-node/issues/1429


bin/node/src/commands/store.rs
```diff
@@ -200,8 +200,7 @@ impl StoreCommand {
         // Parse genesis config (or default if not given).
         let config = genesis_config
             .map(|file_path| {
-                let toml_str = fs_err::read_to_string(file_path)?;
-                GenesisConfig::read_toml(toml_str.as_str()).with_context(|| {
+                GenesisConfig::read_toml_file(file_path).with_context(|| {
```
drahnr, line 205: This swallows some errors, i.e. missing, access rights

crates/store/build.rs
```diff
@@ -1,12 +1,106 @@
...
 
+    // Generate sample agglayer account files for genesis config samples.
+    generate_agglayer_sample_accounts();
     miden_node_rocksdb_cxx_linkage_fix::configure();
 }
+
+/// Generates sample agglayer account files for the `02-with-account-files` genesis config sample.
+///
+/// Creates:
+/// - `bridge.mac` - agglayer bridge account
```
drahnr: Did we document the `.mac` extension somewhere? miden-account-code?

crates/store/src/genesis/config/samples/02-with-account-files.toml
```diff
@@ -0,0 +1,36 @@
+# Genesis configuration example with AggLayer account files
+#
+# This example demonstrates how to include pre-built accounts from .mac files
+# in the genesis block. The account files are generated by the build script
+# using deterministic seeds for reproducibility.
+# They demonstrate interdependencies between accounts:
```
Mirko-von-Leipzig, line 7: Nit
```suggestion
# using deterministic seeds for reproducibility.
#
# They demonstrate interdependencies between accounts:
```

crates/store/src/genesis/config/errors.rs
```diff
@@ -17,13 +19,21 @@ use crate::genesis::config::TokenSymbolStr;
...
+    NativeFaucetNotFungible { path: PathBuf },
     #[error("account translation from config to state failed")]
     Account(#[from] AccountError),
     #[error("asset translation from config to state failed")]
     Asset(#[from] AssetError),
     #[error("adding assets to account failed")]
     AccountDelta(#[from] AccountDeltaError),
-    #[error("the defined asset {symbol:?} has no corresponding faucet")]
+    #[error(
+        "the defined asset {symbol:?} has no corresponding faucet, or the faucet was provided as an account file"
```
Mirko-von-Leipzig: We should `impl Fmt for TokenSymbolStr` so we can do


```suggestion
        "the defined asset '{symbol}' has no corresponding faucet, or the faucet was provided as an account file"
```

crates/store/src/genesis/config/errors.rs
```diff
@@ -17,13 +19,21 @@ use crate::genesis::config::TokenSymbolStr;
...
+    )]
```
Mirko-von-Leipzig, line 35: I don't understand how the faucet being specified as an account file is an error, but I'll return to this once I've read the rest.

crates/store/build.rs
Mirko-von-Leipzig, line 1: Why are generating them if the accounts are checked-in? As a general rule `build.rs` should be used for _out of tree_ code generation, not to generate code that gets checked in.

crates/store/src/genesis/config/mod.rs
```diff
@@ -73,24 +113,92 @@ impl Default for GenesisConfig {
...
     ///
     /// Notice: It will generate the specified case during [`fn into_state`].
-    pub fn read_toml(toml_str: &str) -> Result<Self, GenesisConfigError> {
-        let me = toml::from_str::<Self>(toml_str)?;
-        Ok(me)
+    pub fn read_toml_file(path: &Path) -> Result<Self, GenesisConfigError> {
+        let toml_str =
+            fs_err::read_to_string(path).map_err(|e| GenesisConfigError::ConfigFileRead {
+                path: path.to_path_buf(),
+                reason: e.to_string(),
```
Mirko-von-Leipzig: This will lose context because an errors `ToString` implementation generally only logs the last error and not the full back trace. This is why we generally avoid "stringly" errors.

crates/store/src/genesis/config/mod.rs
```diff
@@ -73,24 +113,92 @@ impl Default for GenesisConfig {
...
+        let native_faucet = match toml_config.native_faucet {
+            NativeFaucetToml::Parameters { symbol, decimals, max_supply } => {
+                NativeFaucet::Parameters { symbol, decimals, max_supply }
+            },
+            NativeFaucetToml::File { path } => {
+                let full_path = config_dir.join(&path);
+                let account_file = AccountFile::read(&full_path).map_err(|e| {
+                    GenesisConfigError::AccountFileRead {
+                        path: full_path.clone(),
+                        reason: e.to_string(),
```
Mirko-von-Leipzig: This drops error context

crates/store/src/genesis/config/mod.rs
```diff
@@ -42,22 +43,61 @@ use self::errors::GenesisConfigError;
...
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+struct AccountToml {
+    /// Path to .mac file (relative to genesis config directory)
+    path: std::path::PathBuf,
+}
+
+/// Intermediate struct for full TOML parsing
+#[derive(Debug, Clone, serde::Deserialize)]
+struct GenesisConfigToml {
```
Mirko-von-Leipzig: Probably? 
```suggestion
#[derive(Debug, Clone, serde::Deserialize)]
#[serde(deny_unknown_fields)]
struct GenesisConfigToml {
```

crates/store/src/genesis/config/mod.rs
```diff
@@ -73,24 +113,92 @@ impl Default for GenesisConfig {
...
+                        path: full_path.clone(),
+                        reason: e.to_string(),
+                    }
+                })?;
+                let account = account_file.account;
+
+                // Validate it's a fungible faucet
+                if account.id().account_type() != AccountType::FungibleFaucet {
+                    return Err(GenesisConfigError::NativeFaucetNotFungible { path: full_path });
+                }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I would also move this into the type's constructor so that we _know_ the type can only be constructed with this invariant.

```rust
impl NativeFaucet {
    pub fn from_account(account: Account) -> Option<Self> { ... }

    pub fn from_parameters(...) -> Self { ... }
}

File: crates/store/src/genesis/config/mod.rs [hunk 10]
```diff
@@ -548,6 +688,14 @@ impl From<TokenSymbolStr> for TokenSymbol {
     }
 }
 
+impl From<TokenSymbol> for TokenSymbolStr {
+    fn from(symbol: TokenSymbol) -> Self {
+        // TokenSymbol guarantees valid format, so to_string should not fail
```
```suggestion
        // SAFETY: TokenSymbol guarantees valid format, so to_string should not fail
```

crates/store/src/genesis/config/mod.rs
```diff
@@ -264,6 +383,9 @@ impl GenesisConfig {
         // Ensure the faucets always precede the wallets referencing them
         all_accounts.extend(wallet_accounts);
 
+        // Append file-loaded accounts as-is
+        all_accounts.extend(file_loaded_accounts);
```
Mirko-von-Leipzig, line 327: I don't think this is safe - do the accounts not need to be verified? e.g. they can contain assets which have no faucets.

crates/store/src/genesis/config/mod.rs
```diff
@@ -281,28 +403,46 @@ impl GenesisConfig {
...
+#[derive(Debug, Clone)]
+enum NativeFaucet {
+    Parameters {
+        symbol: TokenSymbolStr,
+        decimals: u8,
+        max_supply: u64,
+    },
+    Account {
+        account: Box<Account>,
+    },
```
Mirko-von-Leipzig: I think this is somewhat confusing because its not adding much over `NativeFaucetToml` apart from having the account parsed (sometimes).

I would remove this; or at least streamline/refactor things so they make more sense.

crates/store/build.rs
Mirko-von-Leipzig: In general we should avoid using `build.rs` to generate _code in src tree_. Its intended to generate code as part of the build process, not as part of checked in src.

drahnr: Out of scope: We should revisit how we deal with protobuf generated code too.

bin/node/src/commands/store.rs
drahnr: I really don't like this.

The previous variant decoupled parsing from file IO, if we need the error context, then we should use something like:

```
let reader = fs_err::OpenOptions::new().read(true).open(path)?:
let gc = GenesisConfig::parse(reader).with_context(|| { file.path.display().to_string() })?;
..
```

crates/store/build.rs
mmagician: not explicitly documented, but it's become a sort of [de-facto standard](https://github.com/search?q=org%3A0xMiden%20%22.mac%22&type=code) for exported accounts

crates/store/src/genesis/config/mod.rs
mmagician: I refactored things so that:
1. We have a single `GenesisConfig` (prev. also `GenesisConfigToml`)
2. Similarly, there's now a single `NativeFaucetConfig` (prev. `NativeFaucet` + `NativeFaucetToml`)

[feat: unify GenesisConfig & NativeFaucet structs](https://github.com/0xMiden/miden-node/pull/1624/commits/88e943fdf86b7a5f572a8a6de4926943921817ae)

bin/node/src/commands/store.rs
mmagician: Good point, I refactored things so that:
- `read_toml_file` reads the contents of the provided `path` into a str, and then:
    - determines the parent dir
    - calls into `read_toml`, which parses TOML into `GenesisConfig`
- then, the caller uses `GenesisConfig::into_state()` to construct the `GenesisState`

LMK if this is what you had in mind, or would you further separate not just parsing of `GenesisConfig` -> `GenesisState` but also of path to TOML content?

[feat: unify GenesisConfig & NativeFaucet structs](https://github.com/0xMiden/miden-node/pull/1624/commits/88e943fdf86b7a5f572a8a6de4926943921817ae)

crates/store/src/genesis/config/mod.rs
mmagician: [fix: include error context](https://github.com/0xMiden/miden-node/pull/1624/commits/f80c2b33dab955c1d9fc005125c8ec6d98a86acf)

crates/store/src/genesis/config/mod.rs
mmagician: [fix: include error context](https://github.com/0xMiden/miden-node/pull/1624/commits/f80c2b33dab955c1d9fc005125c8ec6d98a86acf)

crates/store/src/genesis/config/mod.rs
mmagician: We have a few options here:
1. Do not verify generic accounts at all - leaving it up to the user to ensure correctness of how they generate generic accounts
2. Verify that generic accounts' assets reference valid `FungibleFaucet` - but this can get complex, because the faucets can be defined explicitly in the TOML, or implicitly as generic accounts:
    1. if provided explicitly, we should ensure that we apply issuance adjustments to these faucets, like we do when a wallet references that faucet
    2. if provided implicitly, we should either:
        1. validate that the total issuance of the faucets matches the total amount in all generic accounts (getting hairy already), or
        2. leave it up to the user

And checking the validity of provided assets in the accounts is just one of the many checks we _could_ run here. For example, for agglayer accounts: does the `AggLayerBridge` account storage reference only valid `AggLayerFaucet`s? And so on.

That's why I'm strongly in favor of Option 1. where we **don't verify** anything about generic accounts.
It DOES leave a disconnect, because we **do verify** wallets <> faucets, but as noted in the [original issue](https://github.com/0xMiden/miden-node/issues/1429), generic accounts are intended for prod, while wallets<>faucets for dev environments.

I added [a note about this](chore: note about NO validation of generic accts)[chore: note about NO validation of generic accts](https://github.com/0xMiden/miden-node/pull/1624/commits/adc3e14eb6407f2a9c53738f371865114a56e34a), but LMK if you disagree with the reasoning, and if so - which level of validation would you want to see enforced?

Mirko-von-Leipzig: My original idea was to use actual transactions to verify everything works as per https://github.com/0xMiden/miden-node/issues/1059#issuecomment-3890395812.

However I think that's probably impossible, though I wish it wasn't.

crates/store/src/genesis/config/mod.rs
```diff
@@ -548,6 +627,14 @@ impl From<TokenSymbolStr> for TokenSymbol {
     }
 }
 
+impl From<TokenSymbol> for TokenSymbolStr {
+    fn from(symbol: TokenSymbol) -> Self {
+        // SAFETY: TokenSymbol guarantees valid format, so to_string should not fail
+        let raw = symbol.to_string().expect("TokenSymbol should always produce valid string");
```
Mirko-von-Leipzig, line 634: I'm actually surprised the method is fallible then. Looking at its code, it really shouldn't be returning an error.

crates/store/src/genesis/config/mod.rs
```diff
@@ -320,6 +343,62 @@ pub struct FeeParameterConfig {
...
+enum NativeFaucetConfig {
+    /// Build from parameters (dev/testing)
+    Parameters {
+        symbol: TokenSymbolStr,
+        decimals: u8,
+        max_supply: u64,
+    },
+    /// Load from pre-built account file (production/multisig)
+    File { path: PathBuf },
+}
```
Mirko-von-Leipzig: Thinking about this some more. Is there a use case where people actually want to configure a basic native faucet? As in, will anyone actually use `Parameters` except the default implementation?

Could we instead have `Option<PathBuf>` and no-one would care?

Not overly important for this PR though.

crates/store/src/genesis/config/mod.rs
```diff
@@ -108,10 +142,20 @@ impl GenesisConfig {
...
+            account: account_entries,
+            config_dir,
         } = self;
 
-        let symbol = native_faucet.symbol.clone();
+        // Load account files from disk
+        let file_loaded_accounts = account_entries
+            .into_iter()
+            .map(|acc| {
+                let full_path = config_dir.join(&acc.path);
```
Mirko-von-Leipzig, line 162: What happens if the account path was specified as an absolute path?

mmagician: not supported ATM

crates/store/src/genesis/config/mod.rs
mmagician: Yes, e.g. to set up multisig authentication for the native faucet. Currently key generation with the native faucet is fixed to rpo

Mirko-von-Leipzig: I think that still agrees with my suggestion. To be more specific, could we do:

```rust
/// Override the native faucet with a custom faucet account.
///
/// If unspecified, a default native faucet will be used with:
///
/// ```json
/// symbol: "MIDEN",
/// decimals: <..>
/// max_supply: <..>
/// ```
native_faucet: Option<PathBuf>,
```

My assumption here is that users will 99% leave the native faucet as default. So there isn't really a use case for configuring from `Parameters`.

mmagician: I see, yes that's a good suggestion. Applied in 55ff6791922caf5519115b2341acc5a34bb4e9ed

crates/store/src/genesis/config/mod.rs
```diff
@@ -73,24 +102,38 @@ impl Default for GenesisConfig {
...
+    /// referenced in the configuration (e.g., `[[account]]` entries with `path` fields).
     ///
     /// Notice: It will generate the specified case during [`fn into_state`].
-    pub fn read_toml(toml_str: &str) -> Result<Self, GenesisConfigError> {
-        let me = toml::from_str::<Self>(toml_str)?;
-        Ok(me)
+    pub fn read_toml_file(path: &Path) -> Result<Self, GenesisConfigError> {
+        let toml_str = fs_err::read_to_string(path)
+            .map_err(|e| GenesisConfigError::ConfigFileRead(e, path.to_path_buf()))?;
+        let config_dir = path.parent().unwrap_or_else(|| Path::new("."));
```
drahnr: I think you can use an expect here, if the user provides i.e. `/` we should not be able to call `read_to_string` without an error.

mmagician: done in [chore: expect path to have parent](https://github.com/0xMiden/miden-node/pull/1624/commits/9811eb81cbec600a9621e8b33df3903ca651353e)

PR #1626, opened 2026-02-01
chore: Remove one level of nesting in remote prover client modules

sergerad:
As per [comment](https://github.com/0xMiden/miden-node/pull/1579#discussion_r2730704971).
>the namespacing here feels a bit "too nested" - e.g., could this not be just miden_remote_prover_client::block_prover::RemoteBlockProver?

### DIFF
--- a/crates/store/src/server/block_prover_client.rs
+++ b/crates/store/src/server/block_prover_client.rs
@@ -1,8 +1,7 @@
 use miden_block_prover::{BlockProverError, LocalBlockProver};
 use miden_protocol::batch::OrderedBatches;
 use miden_protocol::block::{BlockHeader, BlockInputs, BlockProof};
-use miden_remote_prover_client::RemoteProverClientError;
-use miden_remote_prover_client::remote_prover::RemoteBlockProver;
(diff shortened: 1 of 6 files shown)


crates/store/src/server/block_prover_client.rs
```diff
@@ -2,7 +2,7 @@ use miden_block_prover::{BlockProverError, LocalBlockProver};
 use miden_protocol::batch::OrderedBatches;
 use miden_protocol::block::{BlockHeader, BlockInputs, BlockProof};
 use miden_remote_prover_client::RemoteProverClientError;
-use miden_remote_prover_client::remote_prover::block_prover::RemoteBlockProver;
+use miden_remote_prover_client::remote_prover::RemoteBlockProver;
```
bobbinth: Do we also need the `remote_prover` namespace? That is, could this be:
```rust
use miden_remote_prover_client::RemoteBlockProver;
```
Or is this not a good idea for some reason?

sergerad: updated

PR #1631, opened 2026-02-02
ci(docker): use `cargo chef` and cache to github

### LINKED ISSUE #999: Improver docker caching
We currently use S3 as a docker build cache which works as expected; however it doesn't cache any built dependencies. This means it only serves as a cache in the trivial case where no source code is changed.

I'm aware of `cargo chef` as a potential solution, and the linked [hackmd](https://hackmd.io/jgkoQ24YRW6i0xWd73S64A) seems to have a reasonable overview and description of the problem space.

_Originally brought up by @sergerad in https://github.com/0xMiden/miden-node/pull/997#discussion_r2160480396_

### LINKED ISSUE #1374: CI: docker build failing
Our [docker build check](https://github.com/0xMiden/miden-node/actions/workflows/build-docker.yml) has started failing.

Mirko-von-Leipzig:
This PR improves our CI docker build times by:

- using `cargo chef` to cache our dependencies as a separate image layer
- replacing S3 bucket with github cache

Similar to our other github caches, the cache is only updated on _push into `next`_.

The cache usage is around 1 GB which is acceptable I think. I will separately also make changes to how we cache `cargo` stuff to reduce the cache strain in general.

I've tested this on a fork repo (with the free runner), and a cache reduced compile time from 25minutes to 6minutes.

Closes #999 and closes #1374

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2993,6 +2993,7 @@ dependencies = [
  "http-body-util",
  "itertools 0.14.0",
  "lru 0.16.3",
+ "miden-node-rocksdb-cxx-linkage-fix",
  "miden-protocol",
  "opentelemetry",
  "opentelemetry-otlp",
--- a/crates/rocksdb-cxx-linkage-fix/src/lib.rs
+++ b/crates/rocksdb-cxx-linkage-fix/src/lib.rs
@@ -6,6 +6,7 @@ use std::env;
 
 pub fn configure() {
     println!("cargo:rerun-if-env-changed=ROCKSDB_COMPILE");
+    println!("cargo:rerun-if-env-changed=ROCKSDB_LIB_DIR");
     println!("cargo:rerun-if-env-changed=ROCKSDB_STATIC");
     println!("cargo:rerun-if-env-changed=CXXSTDLIB");
     let target = env::var("TARGET").unwrap_or_default();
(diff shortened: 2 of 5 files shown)


bin/node/Dockerfile
```diff
@@ -1,39 +1,47 @@
...
-RUN cargo install --path bin/node --locked
+FROM chef AS planner
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
 
-FROM debian:bullseye-slim
+FROM chef AS builder
+COPY --from=planner /app/recipe.json recipe.json
+# Build dependencies - this is the caching Docker layer!
+RUN cargo chef cook --release --recipe-path recipe.json
```
drahnr, line 25: We are now using a docker layer to cache all dependencies in our build-tree, except for crates that live inside the tree iiuc

Where do we cache that image/the image layers?

Mirko-von-Leipzig: These get cached by the docker invocation, so in this case its the github workflow:

```yaml
- name: Build and push
  uses: docker/build-push-action@v6
  with:
    push: false
    file: ./bin/node/Dockerfile
    cache-from: type=gha
    # Only save cache on push into next
    cache-to: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' && 'type=gha,mode=max' || '' }}
```

and it will upload the various metadata and image blobs to our github cache (`gha`). Previously this was sent to our S3 bucket, but I think 1 GB is small enough to be okay.

Mirko-von-Leipzig: > except for crates that live inside the tree iiuc

The exceptions are the dependencies that are also part of our `build.rs` I think. Though its also only the top layer or parts that have features. Not many. Haven't quite understood why or what precisely.

drahnr: Does the docker cache not count against the gha cache quota?

Mirko-von-Leipzig: It does yes, I'm working to reduce that usage in another PR as well.

But if its an issue still we can revert to caching via S3.

PR #1632, opened 2026-02-02
chore: wrap PartialBlockchain in Arc

### LINKED ISSUE #1119: Reduce cloning of chain_mmr in Network Transaction Builder
From https://github.com/0xMiden/miden-node/pull/1108#discussion_r2238792399:
> We end up cloning chain_mmr twice for each transaction: once here, and then once returning it from the NtxDataStore. Would be great to figure out how we can avoid one of the clones. Let's create an issue for this (this would be a fairly low-priority issue at this point).

SantiagoPittella:
closes #1119

### DIFF
--- a/crates/utils/build.rs
+++ b/crates/utils/build.rs
@@ -1,0 +1,3 @@
+fn main() {
+    miden_node_rocksdb_cxx_linkage_fix::configure();
+}


crates/ntx-builder/src/actor/execute.rs
```diff
@@ -421,7 +422,7 @@ impl DataStore for NtxDataStore {
                 .await;
 
             let partial_account = PartialAccount::from(&self.account);
-            Ok((partial_account, self.reference_block.clone(), self.chain_mmr.clone()))
+            Ok((partial_account, self.reference_block.clone(), (*self.chain_mmr).clone()))
```
bobbinth, line 425: This still means that we'd need to clone the underlying chain MMR once per transaction, right?

But I also don't see a way around this - except for maybe changing the signature of `DataStore::get_transaction_inputs()` to return `Arc<PartialBlockchain>` rather than just `PartialBlockchain`. Maybe that's a good idea regardless.

crates/ntx-builder/src/builder.rs
```diff
@@ -228,7 +231,7 @@ impl NetworkTransactionBuilder {
 
         // Update MMR which lags by one block.
         let mmr_tip = chain_state.chain_tip_header.clone();
-        chain_state.chain_mmr.add_block(&mmr_tip, true);
+        Arc::make_mut(&mut chain_state.chain_mmr).add_block(&mmr_tip, true);
```
bobbinth, line 247: Very cool! I actually didn't know about this option.

crates/ntx-builder/src/builder.rs
bobbinth, line 36: Could we expand the docs a bit here? Specifically, would be great to describe what data the chain MMR tracks. IIUC, we don't keep track of the past block authentication paths because NTX builder executes all notes as "unauthenticated" - and so, doesn't need to keep track of these paths. If so, would be good to explicitly state it.

bobbinth (commented):

crates/ntx-builder/src/actor/execute.rs
SantiagoPittella: Yes, we do need to clone on each transaction.

The `DataStore` store change I think is the only good way to do it properly and remove the full clones. Though now that we prune more properly the block headers this may be overkill.

bobbinth (approved):
Looks good! Thank you!

crates/ntx-builder/src/actor/execute.rs
bobbinth: Could you make a PR in `miden-base` to use `Arc<PartialBlockchain>`? We can merge this PR as is, and then it will naturally propagate here once it is implemented in `miden-base`.

Also, cc @PhilippGackstatter in case he has any thoughts on this.

PhilippGackstatter: Ideally, that data structure wouldn't be very large since it should be only the absolutely necessary set of block headers, and so cloning it shouldn't be that expensive. Cloning should also not incur very expensive operations (e.g. no hashing), "only" memory allocation and copying.

If it tends to be large anyway and cloning is expensive, then we could make this change. I guess I would just avoid it if it isn't absolutely necessary, so maybe benchmarking or checking it would be a good idea first.

igamigo: For the network transaction builder specifically, the structure would be always 4 blocks because it's pruned to that length. Also, only the last block is really necessary (though not sure if there is an advantage to keeping the others).

bobbinth: Yes, agreed - I think in this case, cloning shouldn't actually need to clone more than a few KB of data in the worst case - so, probably negligible.

My thinking was that making the change should be pretty simple, and I don't see a lot of downsides to this - but also, not really a priority at all.

PR #1635, opened 2026-02-03
feat: prune `InnerForest`

drahnr:
> [!IMPORTANT]  
> Targeting `main`

## What

Does cleanup the `InnerForest`, both the lookup tables/maps and the actual `SmtForest`.

Required for bounding the in-memory size growth.

## Context

Related https://github.com/0xMiden/miden-node/issues/1175

## How

We use `LargeSmtForest` - right now with the in-memory backend - to store and track all relevant tree lineages (trees and their respective histories). This vastly simplifies `InnerForest`, it's mainly a simplification to map `(AccountId, ?StorageSlotName) ~> digest` and providing convenient accessors and integrations with `apply_block`.

The general concept is to apply mutations to the `LargeSmtForest` in relation to a lineage, i.e. the vault or storage slot of a given account (the `LineageId`). For queries we can derive a `TreeId` for a given block number - if still present. History gets truncated on every new block to a fixed history entry number.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.13.8 (TBD)
+
+- Private notes with the network note attachment are no longer incorrectly considered as network notes (#[#1736](https://github.com/0xMiden/node/pull/1736)).
+
 ## v0.13.7 (2026-02-25)
 
 - Updated `SyncAccountStorageMaps` and `SyncAccountVault` to allow all accounts with public state, including network accounts ([#1711](https://github.com/0xMiden/node/pull/1711)).
--- a/crates/store/Cargo.toml
+++ b/crates/store/Cargo.toml
@@ -19,7 +19,7 @@ anyhow                 = { workspace = true }
 deadpool               = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
 deadpool-diesel        = { features = ["sqlite"], version = "0.6" }
 deadpool-sync          = { default-features = false, features = ["tracing"], version = "0.1" }
-diesel                 = { features = ["numeric", "sqlite"], version = "2.3" }
+diesel                 = { features = ["numeric", "returning_clauses_for_sqlite_3_35", "sqlite"], version = "2.3" }
 diesel_migrations      = { features = ["sqlite"], version = "2.3" }
 fs-err                 = { workspace = true }
 hex                    = { version = "0.4" }
--- a/crates/store/src/db/mod.rs
+++ b/crates/store/src/db/mod.rs
@@ -4,7 +4,14 @@ use std::ops::RangeInclusive;
 use std::path::PathBuf;
 
 use anyhow::Context;
-use diesel::{Connection, QueryableByName, RunQueryDsl, SqliteConnection};
+use diesel::{
+    BoolExpressionMethods,
+    Connection,
+    ExpressionMethods,
+    QueryableByName,
+    RunQueryDsl,
+    SqliteConnection,
+};
 use miden_node_proto::domain::account::{AccountInfo, AccountSummary};
 use miden_node_proto::generated as proto;
 use miden_node_utils::limiter::MAX_RESPONSE_PAYLOAD_BYTES;
@@ -31,8 +38,8 @@ use crate::COMPONENT;
 use crate::db::manager::{ConnectionManager, configure_connection_on_creation};
 use crate::db::migrations::apply_migrations;
 use crate::db::models::conv::SqlTypeConvert;
-use crate::db::models::queries::StorageMapValuesPage;
-use crate::db::models::{Page, queries};
+use crate::db::models::queries::{NetworkNoteType, StorageMapValuesPage};
+use crate::db::models::{Page, deserialize_raw_vec, queries};
 use crate::errors::{DatabaseError, DatabaseSetupError, NoteSyncError, StateSyncError};
 use crate::genesis::GenesisBlock;
 
@@ -330,7 +337,37 @@ impl Db {
 
         let me = Db { pool };
         me.query("migrations", apply_migrations).await?;
+        me.fixup_network_note_classification().await?;
         Ok(me)
+    }
+
+    /// Temporary fixup of private notes which were misclassified as network notes.
+    #[instrument(target = COMPONENT, skip_all, err)]
+    async fn fixup_network_note_classification(&self) -> Result<()> {
+        let notes = self
+            .transact("fixup network notes", move |conn| {
+                let updated = diesel::update(schema::notes::table)
+                    .filter(
+                        schema::notes::network_note_type
+                            .eq(NetworkNoteType::SingleTarget as i32)
+                            .and(schema::notes::nullifier.is_null()),
+                    )
+                    .set(schema::notes::network_note_type.eq(NetworkNoteType::None as i32))
+                    .returning(schema::notes::note_id)
+                    .get_results::<Vec<u8>>(conn)?;
+
+                deserialize_raw_vec::<_, NoteId>(updated).map_err(DatabaseError::from)
+            })
+            .await?;
+
+        for note in notes {
+            tracing::info!(
+                note.id = %note,
+                "Fixed private note misclassified as network note"
+            );
+        }
+
+        Ok(())
     }
 
     /// Loads all the nullifiers from the DB.
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -12,38 +12,16 @@ permissions:
 
 jobs:
   docker-build:
-    strategy:
-      matrix:
-        component: [node]
     runs-on: Linux-ARM64-Runner
-    name: Build ${{ matrix.component }}
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
-      - name: Configure AWS credentials
-        if: github.event.pull_request.head.repo.fork == false
-        uses: aws-actions/configure-aws-credentials@v4
-        with:
-          aws-region: ${{ secrets.AWS_REGION }}
-          role-to-assume: ${{ secrets.AWS_ROLE }}
-          role-session-name: GithubActionsSession
-
-      - name: Set cache parameters
-        if: github.event.pull_request.head.repo.fork == false
-        run: |
-          echo "CACHE_FROM=type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},name=miden-${{ matrix.component }}" >> $GITHUB_ENV
-          echo "CACHE_TO=type=s3,region=${{ secrets.AWS_REGION }},bucket=${{ secrets.AWS_CACHE_BUCKET }},name=miden-${{ matrix.component }}" >> $GITHUB_ENV
-
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-        with:
-          cache-binary: true
 
-      - name: Build Docker image
-        uses: docker/build-push-action@v5
+      - name: Build and push
+        uses: docker/build-push-action@v6
         with:
           push: false
-          file: ./bin/${{ matrix.component }}/Dockerfile
-          cache-from: ${{ env.CACHE_FROM || '' }}
-          cache-to: ${{ env.CACHE_TO || '' }}
+          file: ./bin/node/Dockerfile
+          cache-from: type=gha
+          # Only save cache on push into next
+          cache-to: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' && 'type=gha,mode=max' || '' }}
--- a/bin/node/Dockerfile
+++ b/bin/node/Dockerfile
@@ -1,39 +1,47 @@
-FROM rust:1.90-slim-bullseye AS builder
-
+FROM rust:1.90-slim-bullseye AS chef
 # Install build dependencies. RocksDB is compiled from source by librocksdb-sys.
 RUN apt-get update && \
     apt-get -y upgrade && \
-    apt-get install -y llvm clang libclang-dev pkg-config libssl-dev libsqlite3-dev ca-certificates && \
+    apt-get install -y \
+        llvm \
+        clang \
+        libclang-dev \
+        cmake \
+        pkg-config \
+        libssl-dev \
+        libsqlite3-dev \
+        ca-certificates && \
     rm -rf /var/lib/apt/lists/*
-
+RUN cargo install cargo-chef
 WORKDIR /app
-COPY ./Cargo.toml .
-COPY ./Cargo.lock .
-COPY ./bin ./bin
-COPY ./crates ./crates
-COPY ./proto ./proto
 
-RUN cargo install --path bin/node --locked
+FROM chef AS planner
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
 
-FROM debian:bullseye-slim
+FROM chef AS builder
+COPY --from=planner /app/recipe.json recipe.json
+# Build dependencies - this is the caching Docker layer!
+RUN cargo chef cook --release --recipe-path recipe.json
+# Build application
+COPY . .
+RUN cargo build --release --locked --bin miden-node
 
-# Update machine & install required packages
-# The installation of sqlite3 is needed for correct function of the SQLite database
+# Base line runtime image with runtime dependencies installed.
+FROM debian:bullseye-slim AS runtime-base
 RUN apt-get update && \
     apt-get -y upgrade && \
-    apt-get install -y --no-install-recommends \
-    sqlite3 \
+    apt-get install -y --no-install-recommends sqlite3 \
     && rm -rf /var/lib/apt/lists/*
 
-COPY --from=builder /usr/local/cargo/bin/miden-node /usr/local/bin/miden-node
-
+FROM runtime-base AS runtime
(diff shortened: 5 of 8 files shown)


crates/store/src/inner_forest/mod.rs
```diff
@@ -60,15 +70,34 @@ pub(crate) struct InnerForest {
 
     /// Maps (`account_id`, `slot_name`, `block_num`) to SMT root.
     /// Populated during block import for all storage map slots.
+    ///
+    /// Note: With potentially large numbers of storage slots per account, this isn't that great.
+    ///
+    /// Attention: Must be a `BTreeMap`, since not every block contains a value here, so we need to
+    /// be able to query the previous blocks cheaply.
```
Mirko-von-Leipzig: Is the idea to compress the fields below to improve the lookup here at some stage?

e.g. to get to something like

```rust 
struct InnerForest {
    forest: SmtForest,
    // Aggregation of `storage_map_roots`, `storage_entries`, `vault_roots`
    accounts: HashMap<AccountId, Update>,
    
    // LUTs for easier pruning.
    vault_roots_by_block: ..,
    storage_slots_by_block: ...,
}
```

crates/store/src/inner_forest/mod.rs
```diff
@@ -229,6 +260,7 @@ impl InnerForest {
     /// # Errors
     ///
     /// Returns an error if applying a vault delta results in a negative balance.
+    #[instrument(target = COMPONENT, skip_all, fields(block_num = %block_num))]
```
Mirko-von-Leipzig:
```suggestion
    #[instrument(target = COMPONENT, skip_all, fields(block.number = %block_num))]
```

crates/store/src/inner_forest/tests.rs
Mirko-von-Leipzig, line 1: Could we remove the `test_` prefix.

crates/store/src/inner_forest/mod.rs
Mirko-von-Leipzig, line 1: This file gives me anxiety

crates/store/src/inner_forest/mod.rs
drahnr: With `LargeSmtForest` all of the inner forest LUTs/side datastructures should disappear.

crates/store/src/state/apply_block.rs
```diff
@@ -0,0 +1,293 @@
...
+        // Validate that header and body match.
+        let tx_commitment = body.transactions().commitment();
+        if header.tx_commitment() != tx_commitment {
+            return Err(InvalidBlockError::InvalidBlockTxCommitment {
+                expected: tx_commitment,
+                actual: header.tx_commitment(),
+            }
+            .into());
+        }
+
```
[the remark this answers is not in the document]
drahnr: Out of scope

crates/store/src/inner_forest/mod.rs
```diff
@@ -597,4 +650,136 @@ impl InnerForest {
...
+    // --------------------------------------------------------------------------------------------
+
+    /// Prunes old entries from the in-memory forest data structures.
+    ///
+    /// Only iterates over blocks in the pruning window (before cutoff). For each affected account
+    /// or slot, checks if there's a newer entry before pruning - preserving the most recent state.
+    ///
+    /// The `SmtForest` itself is not pruned directly as it uses structural sharing and old roots
+    /// are naturally garbage-collected when they become unreachable.
+    #[instrument(target = COMPONENT, skip_all, fields(block.number = %chain_tip), ret)]
```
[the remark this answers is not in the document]
drahnr: It is required.

crates/store/src/inner_forest/mod.rs
```diff
@@ -597,4 +647,136 @@ impl InnerForest {
...
+                    .range((account_id, block.child())..=(account_id, BlockNumber::from(u32::MAX)))
+                    .next()
+                    .is_some();
+
+                if has_newer_entry {
+                    // Prune
+                    if let Some(root) = self.vault_roots.remove(&(account_id, block)) {
+                        roots_to_remove.push(root);
+                    }
+                } else {
```
igamigo: Hmm, would this prune roots globally? That is, another account could be referencing the same root somehow and if one gets pruned, the others potentially won't be able to be served, right?  On the client we have faced a similar issue for keeping roots of local accounts. IIUC some refcounting needs to be in place for this.

igamigo: This should also ideally be tested

crates/store/src/inner_forest/mod.rs
```diff
@@ -597,4 +647,136 @@ impl InnerForest {
...
+                    if let Some(root) = self.storage_map_roots.remove(&key) {
+                        roots_to_remove.push(root);
+                    }
+                    if self.storage_entries.remove(&key).is_some() {
+                        storage_entries_removed += 1;
+                    }
+                } else {
+                    // Keep - this is the most recent state
+                    slots_to_keep.push((account_id, slot_name));
+                }
```
igamigo: I would expect this to grow quite fast as the account tree grows, no? Very likely not the best reference, but according to [this EIP](https://eipsinsight.com/eips/eip-8125), over 60% of storage slots in Ethereum are completely stale (ie they get written to and read from once). And this will increase rescans on every block.

igamigo: Not sure if there is a better alternative for this though given the current setup, but could it help to track latest states and prune them only when a new state gets written for that account?

igamigo (commented):
Overall LGTM, but left a couple of comments. Not sure if they are fully correct (or if they land within the scope of this PR exactly), but worth reviewing for the long term at least.

crates/store/Cargo.toml
Mirko-von-Leipzig, line 1: Should use the workspace one

crates/store/src/state/mod.rs
Mirko-von-Leipzig, line 1: Could you provide an overview of the caching/database approach and the problems this addresses?

My understanding:

- we serve full slot data values from LRU cache, but this only applies to the values at the _latest_ block
  - though technically this would still be valid for a large number of accounts if they haven't been updated?
- everything else is retrieved from the database

crates/store/src/inner_forest/mod.rs
```diff
@@ -214,33 +228,36 @@ impl InnerForest {
...
-            .next_back()
-            .map(|(_, entries)| entries)?;
+        // Get cached snapshot
+        let snapshot = self.latest_storage_cache.get(&(account_id, slot_name.clone()))?;
 
-        if entries.len() > AccountStorageMapDetails::MAX_RETURN_ENTRIES {
+        // Only serve queries for the latest block
+        if snapshot.block_num != block_num {
+            return None; // Historical query - caller should use DB
+        }
```
Mirko-von-Leipzig: Should this not be

```suggestion
        if snapshot.block_num > block_num {
            return None; // Historical query - caller should use DB
        }
```

crates/store/src/inner_forest/mod.rs
drahnr: I don't want to invest too much time on optimization, when we're only weeks away from refactoring it, making use of `LargeSmtForest`.

crates/store/src/inner_forest/mod.rs
drahnr: Added a test and fixed the issue

drahnr: Added a test and fixed the issue, thank you!

crates/store/src/state/mod.rs
drahnr: I'll update the description with the next push to this branch

crates/store/src/inner_forest/mod.rs
drahnr: Obsolete.

crates/store/src/inner_forest/mod.rs
```diff
@@ -597,4 +661,166 @@ impl InnerForest {
...
+        let mut latest_entries = self
+            .storage_entries_per_block_per_account_per_slot
+            .iter()
+            .filter(|((_, cached_account_id, cached_slot_name), _)| {
+                *cached_account_id == account_id && cached_slot_name == slot_name
+            })
+            .map(|((cached_block, ..), snapshot)| (*cached_block, snapshot))
+            .max_by_key(|(cached_block, _)| cached_block.as_u32())
+            .map(|(_, snapshot)| snapshot.entries.clone())
+            .unwrap_or_default();
```
igamigo: Is this correct? Seems like the entry that you are trying to retrieve here could be evicted and then the new delta would be built from a default base.

crates/store/src/db/mod.rs
```diff
@@ -600,13 +601,77 @@ impl Db {
...
+        let values = self
+            .select_storage_map_sync_values(
+                account_id,
+                BlockNumber::GENESIS..=block_num,
+                entries_limit,
+            )
+            .await?;
+        if values.last_block_included != block_num {
+            return Ok(AccountStorageMapDetails::limit_exceeded(slot_name));
+        }
```
igamigo: Should this loop until all values are retrieved instead?

crates/store/src/db/mod.rs
```diff
@@ -600,13 +601,77 @@ impl Db {
...
+        // columns
+        let values = self
+            .select_storage_map_sync_values(
+                account_id,
+                BlockNumber::GENESIS..=block_num,
+                entries_limit,
+            )
+            .await?;
+        if values.last_block_included != block_num {
+            return Ok(AccountStorageMapDetails::limit_exceeded(slot_name));
```
igamigo, line 723: nit: the query does not filter per slot, right? So really the limit is exceeded as a whole

crates/store/src/inner_forest/mod.rs
drahnr: Fixed

crates/store/src/db/mod.rs
drahnr: Fixed

crates/store/src/db/mod.rs
```diff
@@ -600,13 +608,106 @@ impl Db {
...
+        account_id: AccountId,
+        slot_name: miden_protocol::account::StorageSlotName,
+        block_num: BlockNumber,
+        entries_limit: Option<usize>,
+    ) -> Result<miden_node_proto::domain::account::AccountStorageMapDetails> {
+        use miden_node_proto::domain::account::{AccountStorageMapDetails, StorageMapEntries};
+        use miden_protocol::EMPTY_WORD;
+
+        // TODO this remains expensive with a large history until we implement pruning for DB
+        // columns
```
drahnr, line 683: It's already done in `next`

crates/store/src/inner_forest/mod.rs
```diff
@@ -89,38 +99,147 @@ impl InnerForest {
...
+        let lineage = Self::vault_lineage_id(account_id);
+        self.lookup_tree_id(lineage, block_num)
+    }
+
+    #[expect(clippy::unused_self)]
+    fn lookup_tree_id(&self, lineage: LineageId, block_num: BlockNumber) -> TreeId {
+        TreeId::new(lineage, block_num.as_u64())
+    }
+
+    fn storage_lineage_id(account_id: AccountId, slot_name: &StorageSlotName) -> LineageId {
```
drahnr, line 119: Determines the `LineageId` for storage slots

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 
 ### Enhancements
 
+- Added cleanup of old account data from the in-memory forest ([#1175](https://github.com/0xMiden/miden-node/issues/1175)).
```
```suggestion
```

crates/store/src/state/mod.rs
```diff
@@ -1108,25 +1103,53 @@ impl State {
...
+                        )
+                        .ok_or_else(|| DatabaseError::StorageRootNotFound {
+                            account_id,
+                            slot_name: slot_name.to_string(),
+                            block_num,
+                        })?;
+                    match result {
+                        Ok(details) => details,
+                        Err(err) => {
+                            drop(forest_guard);
```
Mirko-von-Leipzig: Why do you need to manually drop the guard?

crates/store/src/state/mod.rs
```diff
@@ -1108,25 +1103,53 @@ impl State {
             Vec::<AccountStorageMapDetails>::with_capacity(storage_requests.len());
 
         // Use forest for storage map queries
-        let forest_guard = self.forest.read().await;
+        let mut forest_guard = self.forest.write().await;
```
Mirko-von-Leipzig: Why do we need `write` access? The only method used is `get_` -- is that mutable?

crates/store/src/state/mod.rs
```diff
@@ -1108,25 +1103,53 @@ impl State {
...
+                        Ok(details) => details,
+                        Err(err) => {
+                            drop(forest_guard);
+                            return Err(DatabaseError::MerkleError(err));
+                        },
+                    }
+                },
+                SlotData::All => {
+                    // we don't want to hold the forest guard for a prolonged time
+                    drop(forest_guard);
```
Mirko-von-Leipzig: Is it possible to split the request into a forest list, and a database list?

As in:

```rust
let with_keys = storage_requests.iter().filter(SlotData::MapKeys);
let with_all  = storage_requests.iter().filter(SlotData::All);

let mut forest_guard = ..;
for (slot, keys) in with_keys {
    forest_guard.get_storage...
}
drop(forest_guard);

for slot in with_all {
    self.db.reconstruct(...)
}
```

crates/store/src/inner_forest/mod.rs
```diff
@@ -559,57 +614,57 @@ impl InnerForest {
...
+    /// Prunes old entries from the in-memory forest data structures.
+    ///
+    /// The `LargeSmtForest` itself is truncated to drop historical versions beyond the cutoff.
+    ///
+    /// Returns the number of _roots_ that got pruned.
+    #[instrument(target = COMPONENT, skip_all, ret, fields(block.number = %chain_tip))]
+    pub(crate) fn prune(&mut self, chain_tip: BlockNumber) -> usize {
+        let cutoff_block = chain_tip
+            .checked_sub(HISTORICAL_BLOCK_RETENTION)
+            .unwrap_or(BlockNumber::GENESIS);
```
Mirko-von-Leipzig, line 662: I guess we don't have `BlockNumber::saturating_sub`?

crates/store/src/inner_forest/mod.rs
```diff
@@ -559,57 +614,57 @@ impl InnerForest {
...
     }
+
+    // PRUNING
+    // --------------------------------------------------------------------------------------------
+
+    /// Prunes old entries from the in-memory forest data structures.
+    ///
+    /// The `LargeSmtForest` itself is truncated to drop historical versions beyond the cutoff.
+    ///
+    /// Returns the number of _roots_ that got pruned.
```
Mirko-von-Leipzig: Is this important?

crates/store/src/inner_forest/mod.rs
Mirko-von-Leipzig: I don't know what a `LineageId` is?

crates/store/src/db/mod.rs
```diff
@@ -34,6 +36,12 @@ use crate::db::models::{Page, queries};
 use crate::errors::{DatabaseError, DatabaseSetupError, NoteSyncError, StateSyncError};
 use crate::genesis::GenesisBlock;
 
+const ROW_OVERHEAD_BYTES: usize = 2 * size_of::<Word>() + size_of::<u32>() + size_of::<u8>();
```
Mirko-von-Leipzig: Overhead of what?

crates/store/src/inner_forest/mod.rs
drahnr: We don't unfortunately

crates/store/src/inner_forest/mod.rs
drahnr: It's not, but I'd rather have an indication how many we prune than flying blind.

crates/store/src/state/mod.rs
drahnr: Yes, we can.

crates/store/src/db/mod.rs
drahnr: bytes per row would be correct, changed

crates/store/src/state/mod.rs
drahnr: Done, but I am not convinced it got any better.

crates/store/src/state/mod.rs
drahnr: Fixed

Mirko-von-Leipzig (approved):

crates/store/src/db/mod.rs
```diff
@@ -600,13 +609,106 @@ impl Db {
         &self,
         account_id: AccountId,
         block_range: RangeInclusive<BlockNumber>,
+        entries_limit: Option<usize>,
```
drahnr, line 649: Required for `reconstruct_storage_map_from_db` to return `LIMIT_EXCEEDED` when a request does a `SlotData::All`.

crates/store/src/inner_forest/mod.rs
Mirko-von-Leipzig: A `lineage` is a term from the forest for the set of trees belonging to a specific entity e.g. the trees for account `A` are the lineage of `A`. `lineage` because these are the updates to that tree over time.

Issue #1591, opened 2026-01-26
Remove the `SynState` endpoint


bobbinth:
Our current `SynState` endpoint does a bit too much, and I think we should be able to replace it with other existing endpoints. Specifically:

- `SyncTransactions` could be used to get all information about updates performed on a specific set of accounts.
- `SyncNotes` could be used to get information on all "incoming" notes for a specific set of accounts.
- `SyncNullifiers` could be used to get information about the status of the notes that a client produced.

The only thing that I think is missing is the ability to sync chain MMR. So, we may need to add an endpoint for this, but it should be pretty straight-forward.

One open question is whether getting rid of the `SyncState` endpoint will make things much more difficult on the client. cc @igamigo.

PR #1636, opened 2026-02-03
feat: initial SyncMmr impl

drahnr:
# Why

`SynState` is too broad and complicated.

# What

1. Obsoletes `SynState` RPC endpoint
2. Adds `SyncChainMmr` RPC endpoint

# Context

Closes https://github.com/0xMiden/miden-node/issues/1591

### DIFF
--- a/bin/stress-test/src/store/mod.rs
+++ b/bin/stress-test/src/store/mod.rs
@@ -457,13 +457,13 @@ pub async fn bench_sync_chain_mmr(
 
     let total_runs = results.len();
     let paginated_runs = results.iter().filter(|r| r.pages > 1).count();
-    #[allow(clippy::cast_precision_loss)]
+    #[expect(clippy::cast_precision_loss)]
     let pagination_rate = if total_runs > 0 {
         (paginated_runs as f64 / total_runs as f64) * 100.0
     } else {
         0.0
     };
-    #[allow(clippy::cast_precision_loss)]
+    #[expect(clippy::cast_precision_loss)]
     let avg_pages = if total_runs > 0 {
         results.iter().map(|r| r.pages as f64).sum::<f64>() / total_runs as f64
     } else {
--- a/crates/store/src/db/models/queries/mod.rs
+++ b/crates/store/src/db/models/queries/mod.rs
@@ -25,11 +25,6 @@
 //! transaction, any nesting of further `transaction(conn, || {})` has no effect and should be
 //! considered unnecessary boilerplate by default.
 
-#![expect(
-    clippy::needless_pass_by_value,
-    reason = "The parent scope does own it, passing by value avoids additional boilerplate"
-)]
-
 use diesel::SqliteConnection;
 use miden_crypto::dsa::ecdsa_k256_keccak::Signature;
 use miden_protocol::block::{BlockAccountUpdate, BlockHeader};


proto/proto/rpc.proto
```diff
@@ -118,7 +118,11 @@ service Api {
     // For preserving some degree of privacy, note tags contain only high
     // part of hashes. Thus, returned data contains excessive notes, client can make
     // additional filtering of that data on its side.
-    rpc SyncState(SyncStateRequest) returns (SyncStateResponse) {}
+    rpc SyncState(SyncStateRequest) returns (SyncStateResponse) {
+        // option deprecated = true;
```
drahnr: TODO, but it'd fail `clippy`

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/miden-node/pull/1595)).
 - [BREAKING] Move block proving from Blocker Producer to the Store ([#1579](https://github.com/0xMiden/miden-node/pull/1579)).
+- [BREAKING] Remove `SynState` and introduce `SyncChainMmr` ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
```
drahnr, line 9: Technically we only add `SyncChainMmr`, removing would be a direct follow up

Mirko-von-Leipzig: Should we make this two separate entries then? One for adding, one for removal?

proto/proto/rpc.proto
```diff
@@ -541,6 +545,29 @@ message SyncStateResponse {
...
+
+// Chain MMR synchronization request.
+message SyncChainMmrRequest {
+    // Last block known by the requester. The response will contain data starting from the next
+    // block, until `block_to` or the chain tip.
+    blockchain.BlockNumber block_num = 1;
+
+    // Optional last block to include in the response (inclusive).
+    optional blockchain.BlockNumber block_to = 2;
+}
```
Mirko-von-Leipzig, line 493: This should be `BlockRange` from higher up in this file

crates/store/src/server/rpc_api.rs
```diff
@@ -196,6 +200,40 @@ impl rpc_server::Rpc for StoreApi {
...
+        request: Request<proto::rpc::SyncChainMmrRequest>,
+    ) -> Result<Response<proto::rpc::SyncChainMmrResponse>, Status> {
+        let request = request.into_inner();
+        let chain_tip = self.state.latest_block_num().await;
+        let block_num = request
+            .block_num
+            .ok_or_else(|| proto::rpc::SyncChainMmrRequest::missing_field(stringify!(block_num)))
+            .map_err(SyncChainMmrError::from)?;
+        let block_num = BlockNumber::from(block_num);
+        let block_to = request.block_to.map_or(chain_tip, BlockNumber::from).min(chain_tip);
```
Mirko-von-Leipzig: iiuc this is okay because the response includes the chain tip (otherwise it would have to be error'd as out of bounds)

crates/store/src/server/rpc_api.rs
```diff
@@ -196,6 +200,40 @@ impl rpc_server::Rpc for StoreApi {
...
+
+        if block_num > block_to {
+            return Err(SyncChainMmrError::InvalidBlockRange(
+                InvalidBlockRange::StartGreaterThanEnd { start: block_num, end: block_to },
+            )
+            .into());
+        }
```
Mirko-von-Leipzig: Do we care about this? `10..8` is a valid but empty range.

crates/store/src/server/rpc_api.rs
```diff
@@ -196,6 +200,40 @@ impl rpc_server::Rpc for StoreApi {
...
+
+        let last_block_included = block_to;
+        let mmr_delta =
+            self.state.sync_chain_mmr(block_num, block_to).await.map_err(internal_error)?;
```
Mirko-von-Leipzig: Should this take a range as input?

crates/store/src/errors.rs
```diff
@@ -348,6 +348,14 @@ pub enum StateSyncError {
     FailedToBuildMmrDelta(#[from] MmrError),
 }
 
+#[derive(Error, Debug, GrpcError)]
+pub enum SyncChainMmrError {
+    #[error("invalid block range")]
+    InvalidBlockRange(#[from] InvalidBlockRange),
+    #[error("malformed block number")]
+    DeserializationFailed(#[from] ConversionError),
```
Mirko-von-Leipzig:
```suggestion
    #[error("invalid block range")]
    InvalidBlockRange(#[source] InvalidBlockRange),
    #[error("malformed block number")]
    DeserializationFailed(#[source] ConversionError),
```

crates/store/src/server/rpc_api.rs
drahnr: I think explicit is better than implicit, as a user I'd prefer an error rather than unexpected behaviour

crates/store/src/server/rpc_api.rs
drahnr: The request range is inclusive, if that's what you mean?
We provide what the user requested, unless it's larger than our chain tip.

proto/proto/rpc.proto
bobbinth: I think we can probably remove this endpoint right away. Though, I'll defer to @igamigo on what's the best sequence to adapt the client to it.

igamigo: Yes, if the MMR endpoint is there, I believe we can remove the `SyncState` endpoint completely already.

CHANGELOG.md
```diff
@@ -5,18 +5,16 @@
 ### Enhancements
 
 - [BREAKING] Move block proving from Blocker Producer to the Store ([#1579](https://github.com/0xMiden/miden-node/pull/1579)).
+- Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
```
```suggestion
- Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1636](https://github.com/0xMiden/miden-node/issues/1636)).
```

CHANGELOG.md
```diff
@@ -5,18 +5,16 @@
...
 - [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/miden-node/pull/1595)).
 
 ### Changes
 
+- [BREAKING] Removed obsolete `SyncState` RPC endpoint; clients should use `SyncNotes`, `SyncNullifiers`, `SyncAccountVault`, `SyncAccountStorageMaps`, `SyncTransactions`, or `SyncChainMmr` instead (#XXXX).
+- Added account ID limits for `SyncTransactions`, `SyncAccountVault`, and `SyncAccountStorageMaps` to `GetLimits` responses (#XXXX).
```
```suggestion
- [BREAKING] Removed obsolete `SyncState` RPC endpoint; clients should use `SyncNotes`, `SyncNullifiers`, `SyncAccountVault`, `SyncAccountStorageMaps`, `SyncTransactions`, or `SyncChainMmr` instead ([#1636](https://github.com/0xMiden/miden-node/issues/1636)).
- Added account ID limits for `SyncTransactions`, `SyncAccountVault`, and `SyncAccountStorageMaps` to `GetLimits` responses ([#1636](https://github.com/0xMiden/miden-node/issues/1636)).
```

proto/proto/rpc.proto
```diff
@@ -494,51 +479,26 @@ message SyncNotesResponse {
...
+    // Block range from which to synchronize the chain MMR.
     //
-    // An account commitment will be included if-and-only-if it is the latest update. Meaning it is
-    // possible there was an update to the account for the given range, but if it is not the latest,
-    // it won't be included in the response.
-    repeated account.AccountId account_ids = 2;
-
-    // Specifies the tags which the requester is interested in.
-    repeated fixed32 note_tags = 3;
+    // The response will contain MMR delta from `block_range.block_from + 1` up to
```
Mirko-von-Leipzig: Is it really `block_range.block_from + 1`? Why plus one? Or are you taking into account the mmr offset?

crates/store/src/state/sync_state.rs
```diff
@@ -25,6 +25,42 @@ impl State {
...
+        }
+
+        // Important notes about the boundary conditions:
+        //
+        // - The Mmr forest is 1-indexed whereas the block number is 0-indexed. The Mmr root
+        //   contained in the block header always lag behind by one block, this is because the Mmr
+        //   leaves are hashes of block headers, and we can't have self-referential hashes. These
+        //   two points cancel out and don't require adjusting.
+        // - Mmr::get_delta is inclusive, whereas the sync request block_from is defined to be
+        //   exclusive, so the from_forest has to be adjusted with a +1.
```
Mirko-von-Leipzig: I worry about this definition. Users will expect (I think) `MMR` to return the values as per the block header number for that block, and get the mmr as at that block. Unless I'm misunderstanding.

crates/store/src/server/rpc_api.rs
```diff
@@ -196,6 +152,42 @@ impl rpc_server::Rpc for StoreApi {
...
+
+        let block_from = BlockNumber::from(block_range.block_from);
+        let block_to = block_range.block_to.map_or(chain_tip, BlockNumber::from).min(chain_tip);
+
+        if block_from > block_to {
+            return Err(SyncChainMmrError::InvalidBlockRange(
+                InvalidBlockRange::StartGreaterThanEnd { start: block_from, end: block_to },
+            )
+            .into());
+        }
```
Mirko-von-Leipzig, line 184: This might be confusing if `block_to` is `None` and `block_from > chain_tip`. Though maybe its fine.

crates/rpc/src/server/api.rs
```diff
@@ -538,12 +535,18 @@ static RPC_LIMITS: LazyLock<proto::rpc::RpcLimits> = LazyLock::new(|| {
...
+            ),
+            (
+                "SyncAccountVault".into(),
+                endpoint_limits(&[(AccountId::PARAM_NAME, AccountId::LIMIT)]),
+            ),
+            (
+                "SyncAccountStorageMaps".into(),
+                endpoint_limits(&[(AccountId::PARAM_NAME, AccountId::LIMIT)]),
             ),
+            ("SyncChainMmr".into(), endpoint_limits(&[])),
```
Mirko-von-Leipzig: Should there be a block range limit?

Mirko-von-Leipzig (approved):
Loving the line delta :)

crates/store/src/server/rpc_api.rs
drahnr: Added another error variant

crates/store/src/state/sync_state.rs
drahnr: Happy to change, kept it in-line with the previous `SynState`.

From a practicality perspective, when calling I have my chain-tip and want everything starting from there, so I think the common call is to do `local_chain_tip..` as the request. The response I'd expect is `{ (local_chain_tip+1)..=(local_chain_tip+PAGE_SIZE), Mmr for <<that range }`

Mirko-von-Leipzig: Thing is then you can't sync from genesis though? Though I guess you have to fetch genesis header and build your own from there which you need to do to compare the hashes in any case

drahnr: Fair, @igamigo is that an issue/a lot of extra work? We can also be redundent and return including the `local_chain_tip` (client perspective).

---

To expand a bit on expected usage:

Assuming there are more entries in the MMR than PAGE_SIZE, the next call would be:
`local_chain_tip+1+PAGE_SIZE..` yielding `(local_chain_tip+1+PAGE_SIZE)..=(local_chain_tip+PAGE_SIZE*2), Mmr for <<that range }`  where `local_chain_tip+PAGE_SIZE` is part of the pagination info returned.

igamigo: I don't think we can return the response here starting on `local_chain_tip`, not without making it more inconvenient to apply the changes within the client anyway. Basically, we need to store peaks for a specific block (e.g, for block number 10 we want to keep the peaks of the MMR with 11 leaves). For this, when syncing, we basically apply the `MmrDelta` (which gets your MMR to block number 9), then apply the individual block header (number 10), and get the peaks to store alongside the block header. If on a subsequent response we got the `MmrDelta` starting block number 10 (whereas currently we would receive it starting block 11); we'd need to "rolllback" the MMR and apply the delta accordingly. 
TLDR, I believe the current way is fine and overlapping the delta with the response/request tips may complicate things.

bin/stress-test/src/store/mod.rs
```diff
@@ -481,6 +416,146 @@ async fn sync_transactions_paginated(
...
+        let (elapsed, response) =
+            sync_chain_mmr(api_client, next_block_from, target_block_to).await;
+        total_duration += elapsed;
+        pages += 1;
+
+        let pagination_info = response.pagination_info.expect("pagination_info should exist");
+        let mmr_delta = response.mmr_delta.expect("mmr_delta should exist");
+        let mmr_delta = MmrDelta::try_from(mmr_delta).expect("mmr_delta conversion should succeed");
+
+        partial_mmr.apply(mmr_delta).expect("mmr delta should apply");
```
igamigo: I'm not sure we want to inlcude MMR building and updating as part of the benchmark (AFAIK the previous one did not do this)

bin/stress-test/src/store/mod.rs
igamigo, line 1: I think one thing that we might be missing (but probably not to be tackled here), is a "full sync" benchmark. Basically, something similar to the `bench_sync_state` that simulates a client sync altogether

crates/store/src/server/rpc_api.rs
```diff
@@ -197,6 +153,57 @@ impl rpc_server::Rpc for StoreApi {
         }))
     }
 
+    /// Returns chain MMR updates within a block range.
+    async fn sync_chain_mmr(
+        &self,
+        request: Request<proto::rpc::SyncChainMmrRequest>,
+    ) -> Result<Response<proto::rpc::SyncChainMmrResponse>, Status> {
+        const MAX_BLOCKS: u32 = 1000;
```
igamigo: Is this limit based on anything? I don't think it meaningfully caps the response size. It would probably be less overhead in terms of walking the MMR but I wonder if this is below what would be sensible since it should be logarithmic.

bin/stress-test/src/store/mod.rs
drahnr: Removed

bin/stress-test/src/store/mod.rs
drahnr: Fair, created https://github.com/0xMiden/miden-node/issues/1676

crates/store/src/server/rpc_api.rs
drahnr: We can always up it to `u32::MAX` (as limited by `PageInfo` response type). The size of `Mmr` is a set of roots + depth of tree (so yes, log2(n)) to be transferred which should be fine for `u32::MAX` (or beyond if we change the representation of `PageInfo`).

proto/proto/rpc.proto
```diff
@@ -103,22 +103,7 @@ service Api {
...
-    //
-    // Each update response also contains info about new notes, accounts etc. created. It also
-    // returns Chain MMR delta that can be used to update the state of Chain MMR. This includes
-    // both chain MMR peaks and chain MMR nodes.
-    //
-    // For preserving some degree of privacy, note tags contain only high
-    // part of hashes. Thus, returned data contains excessive notes, client can make
-    // additional filtering of that data on its side.
-    rpc SyncState(SyncStateRequest) returns (SyncStateResponse) {}
+    rpc SyncChainMmr(SyncChainMmrRequest) returns (SyncChainMmrResponse) {}
```
bobbinth, line 106: Let's add some basic doc comments here (similar to how we do it for other endpoints in this file).

docs/external/src/rpc.md
bobbinth, line 1: We should probably add a subsection for `SynChainMmr` endpoint here.

docs/external/src/rpc.md
```diff
@@ -141,7 +140,9 @@ This endpoint allows clients to discover the maximum number of items that can be
   "endpoints": {
     "CheckNullifiers": { "parameters": { "nullifier": 1000 } },
     "SyncNullifiers": { "parameters": { "nullifier": 1000 } },
-    "SyncState": { "parameters": { "account_id": 1000, "note_tag": 1000 } },
+    "SyncTransactions": { "parameters": { "account_id": 1000 } },
+    "SyncAccountVault": { "parameters": { "account_id": 1000 } },
+    "SyncAccountStorageMaps": { "parameters": { "account_id": 1000 } },
     "SyncNotes": { "parameters": { "note_tag": 1000 } },
     "GetNotesById": { "parameters": { "note_id": 100 } }
```
bobbinth, line 147: Shouldn't `SyncChainMmr` also be in this list?

crates/utils/src/limiter.rs
```diff
@@ -114,13 +113,23 @@ impl QueryParamLimiter for QueryParamNoteCommitmentLimit {
...
 impl QueryParamLimiter for QueryParamBlockLimit {
     const PARAM_NAME: &str = "block_header";
     const LIMIT: usize = GENERAL_REQUEST_LIMIT;
 }
 
+/// Used for the following RPC endpoints:
+/// * `sync_chain_mmr`
+///
+/// Capped at 1000 blocks to keep MMR deltas within the 4 MB payload budget.
+pub struct QueryParamBlockRangeLimit;
```
bobbinth, line 127: Is the computation here correct? Feels like we should be able to query for way more blocks before before we hit the 4 MB limit (and since we are not sending authentication paths, I'm actually not sure we'll ever hit the limit).

Also, 1000 bocks is current about ~1 hour. If we do keep this limit, we should bump it up significantly so that a single request could cover multiple months.

crates/store/README.md
bobbinth, line 1: Similar to my other comment: we should probably add a section for `SyncChainMmr` here.

crates/rpc/README.md
bobbinth, line 1: Similar to some other comments: this file should probably have a section on `SyncChainMmr` endpoint.

crates/rpc/src/server/api.rs
```diff
@@ -192,16 +192,13 @@ impl api_server::Api for RpcService {
...
-    ) -> Result<Response<proto::rpc::SyncStateResponse>, Status> {
+        request: Request<proto::rpc::SyncChainMmrRequest>,
+    ) -> Result<Response<proto::rpc::SyncChainMmrResponse>, Status> {
         debug!(target: COMPONENT, request = ?request.get_ref());
 
-        check::<QueryParamAccountIdLimit>(request.get_ref().account_ids.len())?;
-        check::<QueryParamNoteTagLimit>(request.get_ref().note_tags.len())?;
-
-        self.store.clone().sync_state(request).await
+        self.store.clone().sync_chain_mmr(request).await
```
bobbinth, line 201: Even though we define a limit for block range, it doesn't seem like we enforce it here (unless I'm missing something). But then again, I'm not sure we actually need this limit.

crates/store/src/server/rpc_api.rs
```diff
@@ -197,6 +153,58 @@ impl rpc_server::Rpc for StoreApi {
         }))
     }
 
+    /// Returns chain MMR updates within a block range.
+    async fn sync_chain_mmr(
+        &self,
+        request: Request<proto::rpc::SyncChainMmrRequest>,
+    ) -> Result<Response<proto::rpc::SyncChainMmrResponse>, Status> {
+        // TODO find a reasonable upper boundary
+        const MAX_BLOCKS: u32 = 1 << 20;
```
bobbinth, line 162: Related to some other comments - I'm not sure there needs to be an upper bound.

bobbinth (commented):
I know this has already been merged, but I left some comments inline and it would be god to address them in a follow-up PR.

docs/external/src/rpc.md
drahnr: I don't think it should, since it's paginated over the block range, so whatever requests a piece is returned with pagination info required to resume

crates/utils/src/limiter.rs
drahnr: This accidentally made it in. It's not used.

crates/rpc/src/server/api.rs
drahnr: The block range limit was accidentally committed and doesn't make sense for paged request

crates/store/src/server/rpc_api.rs
drahnr: If there is no upper bound, even pagination, then we should remove pagination altogether.

bobbinth: I think the main intent for specifying the block range was so that the client could request to sync to a specific block height (rather than the chain tip). Basically, the client needs to be able to say: I want chain MMR delta starting from block $x$ and up to block $y$. But this is technically not pagination.

cc @igamigo in case I'm off on this.

drahnr:
Created https://github.com/0xMiden/miden-node/pull/1682 for follow-ups

crates/store/src/server/rpc_api.rs
igamigo: Yes, that's mainly it. I think with the new flow we'd get relevant block headers through `SyncNotes` and then we can `SyncMmr` as needed.

PR #1637, opened 2026-02-03
ci: organize by trigger and improve caching

Mirko-von-Leipzig:
Inspired by the compiler repo's implementation, this PR simplifies our github workflows by:

- aggregating all CI related tasks into `ci.yml`
- moving expensive jobs to the nightly workflow triggered by cron job
    - these don't use caching
- re-organizing CI jobs so they all re-use a single build job as cache

All CI jobs that can benefit from caching now come _after_ a `build` job populates the cache. This means instead of one cache per job, we now have one cache for the entire workflow.

I've tested this on my fork with decent results and build times, though we'll want to just check I didn't typo something, or have the wrong cache save-if condition. This wasn't possible to confirm on the fork.

I've also moved to the smaller normal runner -- I of course couldn't access this on my fork and job times seemed pretty short despite this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,21 +4,23 @@
 
 ### Enhancements
 
-- [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/miden-node/pull/1595)).
 - [BREAKING] Move block proving from Blocker Producer to the Store ([#1579](https://github.com/0xMiden/miden-node/pull/1579)).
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,267 @@
...
+      - next
+    paths-ignore:
+      - "**.md"
+      - "**.txt"
+      - "docs/**"
+  pull_request:
+    paths-ignore:
+      - "**.md"
+      - "**.txt"
+      - "docs/**"
```
drahnr, line 23: Are we sure we don't want to run these on doc only PRs?

Mirko-von-Leipzig: I think these are all `rust` related jobs, so shouldn't be affected by docs folder stuff :shrug: But I don't really mind re-enabling it, they're quite rare.

Mirko-von-Leipzig:
@bobbinth I can't merge this PR because the `test` workflow is required. However, this PR *removes* that workflow and I cannot force it.

PR #1638, opened 2026-02-04
chore: Add validator service file

sergerad:
The Debian packaging automation needed to be updated to support multiple services in a single Debian package.

### DIFF
--- a/.github/actions/debian/action.yml
+++ b/.github/actions/debian/action.yml
@@ -146,6 +146,7 @@ runs:
 
         # Save the .deb files, delete the rest.
         mv packaging/deb/*.deb .
+        rm -rf packaging
 
     - name: Package names
       shell: bash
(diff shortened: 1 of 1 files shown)


packaging/node/miden-validator.service
```diff
@@ -0,0 +1,16 @@
...
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Type=exec
+Environment="OTEL_SERVICE_NAME=miden-validator"
+EnvironmentFile=/lib/systemd/system/miden-validator.env
+ExecStart=/usr/bin/miden-node validator start
+WorkingDirectory=/opt/miden-node
```
Mirko-von-Leipzig: Bit of a meta question, but we should probably distinguish this from the node? In case both node + validator run on the same instance.

```suggestion
WorkingDirectory=/opt/miden-validator
```

A different way of doing things is to have common `/opt/miden/` directory and then each service gets its own `/opt/miden/<service>` folder in there. But that gets dodgy..? Or maybe not.

Essentially, I _think_ I recall that each "org" should have its own `/opt` directory. But I could be mistaken. I think part of the problem would then be that we probably need a `miden` user group to own the `/opt/miden` directory, which would then let the `miden-node`, `miden-validator` users act within that group. Or we just have a single `miden` user which handles _all_ miden services.

Probably I'm overthinking things :)

sergerad: Seems reasonable but would opt to add validator in the current scheme we have and do a migration for all existing packages/services separately to this work

sergerad:
I have tested that everything works up until the point where a release is required to exist
https://github.com/0xMiden/miden-node/actions/runs/21809782138

Issue #1640, opened 2026-02-04
Empty map slots do not get inserted to `InnerForest`


igamigo:
While testing creating non-trivial accounts (eg, accounts with various slots/map data) to test the account workflows in Devnet, @juan518munoz found an issue with the account insertion. It seems that when an account has empty storage map slots, the `InnerForest` correctly inserts the root into `storage_map_roots` but fails to insert an entry into `storage_entries`. This causes `SlotData::All` queries to fail with “storage root not found” errors, even though the storage map slot exists.

In https://github.com/0xMiden/miden-node/blob/6cd595a79e111989618f664489832537d7b08001/crates/store/src/inner_forest/mod.rs#L505-L510
empty maps are inserted into `storage_map_roots` but the code skips inserting into `storage_entries`.

When a client queries with `SlotData::All`, https://github.com/0xMiden/miden-node/blob/6cd595a79e111989618f664489832537d7b08001/crates/store/src/inner_forest/mod.rs#L189-L214 is called, which returns `None` because there’s no entry in `storage_entries` for empty maps. This triggers the [error path in `fetch_public_account_details](https://github.com/0xMiden/miden-node/blob/6cd595a79e111989618f664489832537d7b08001/crates/store/src/state/mod.rs#L1123-L1129)()`.

A probably fix would be to insert an empty `BTreeMap` into `storage_entries` for empty maps:
```
if map_entries.is_empty() {
    self.storage_map_roots
        .insert((account_id, slot_name.clone(), block_num), prev_root);

    // Also insert empty entries so storage_map_entries() queries work
    self.storage_entries
        .insert((account_id, slot_name.clone(), block_num), EMPTY_WORD);

    continue;
}
```

PR #1642, opened 2026-02-04
fix: empty map slot not inserted in forest

juan518munoz:
As discussed in the issue (linked bellow), empty maps where not being inserted into the `storage_entries`, producing a failure when attempting to fetch a full account from the node.

Closes https://github.com/0xMiden/miden-node/issues/1640

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Fixed network monitor displaying explorer URL as a "null" hyperlink when unset ([#1617](https://github.com/0xMiden/miden-node/pull/1617)).
+- Fixed empty storage maps not being inserted into `storage_entries` table when inserting storage delta ([#1642](https://github.com/0xMiden/miden-node/pull/1642)).
 
 ## v0.13.3 (2026-01-29)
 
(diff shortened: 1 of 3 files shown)


crates/store/src/inner_forest/mod.rs
```diff
@@ -500,11 +500,13 @@ impl InnerForest {
...
 
-            // if the delta is empty, make sure we create an entry in the storage map roots map,
-            // but no need to do anything else
+            // if the delta is empty, make sure we create an entry in the storage map roots map
+            // and storage entries map (so storage_map_entries() queries work)
             if map_entries.is_empty() {
                 self.storage_map_roots
                     .insert((account_id, slot_name.clone(), block_num), prev_root);
+                self.storage_entries
+                    .insert((account_id, slot_name.clone(), block_num), BTreeMap::new());
```
bobbinth, line 509: nit (and not from this PR), but probably a more appropriate name for `storage_entries` would have been `storage_map_entries`.

bobbinth (approved):
Looks good! Thank you! Let's also add a changelog entry and increment the crate versions in `Cargo.toml`

PR #1643, opened 2026-02-05
fix: flush OTel traces on panic

SantiagoPittella:
The panic hook adds the panic span to the batch queue, but the program terminates immediately after. The OtelGuard in `main()` is never dropped. Now the flush happens synchronously before the program terminates on panics.

This PR is aiming to `main` since it might be a good addition to have for devnet/testnet as soon as possible.

### DIFF
--- a/crates/utils/src/logging.rs
+++ b/crates/utils/src/logging.rs
@@ -1,4 +1,5 @@
 use std::str::FromStr;
+use std::sync::OnceLock;
 
 use opentelemetry::trace::TracerProvider as _;
 use opentelemetry_otlp::WithTonicConfig;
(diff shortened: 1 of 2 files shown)


crates/utils/src/logging.rs
```diff
@@ -71,8 +86,21 @@ pub fn setup_tracing(otel: OpenTelemetry) -> anyhow::Result<Option<OtelGuard>> {
...
+    // This chains with the default panic hook to preserve backtrace printing.
+    let default_hook = std::panic::take_hook();
+    std::panic::set_hook(Box::new(move |info| {
         tracing::error!(panic = true, "{info}");
+
+        // Flush traces before the program terminates.
+        // This ensures the panic trace is exported even though the OtelGuard won't be dropped.
+        if let Some(provider) = TRACER_PROVIDER.get() {
+            if let Err(err) = provider.force_flush() {
+                eprintln!("Failed to flush traces on panic: {err:?}");
```
Mirko-von-Leipzig, line 98: Nit

```suggestion
                eprintln!("Failed to flush traces on panic: {}", err.as_report());
```

Mirko-von-Leipzig: Heh. dunno why I merged now without this - not a big deal though. My bad.

drahnr:
Do we want to do something similar on signal? i.e. `SIGSEGV` and others?

Mirko-von-Leipzig:
> Do we want to do something similar on signal? i.e. `SIGSEGV` and others?

I think yes - or at least I don't see a downside.

PR #1645, opened 2026-02-06
feat/db: cleanup old account state data db entries

drahnr:
> [!IMPORTANT]  
> Targeting `main`

We recently started tracking historical account state storage ( #1227 ) but that punted on the cleanup, causing unbounded growth of the DB.

The PR adds a cleanup routine to `apply_block`.

Pro:
* simple, all writing DB transactions happen in one place
* no risk of sqlite locked errors popping up

Con:
* prolonged time the mutex is being held
* does not exploit all possible parallelism of `apply_block`


drahnr:
After self review, this still has some implications to be addressed, particularly around storage map root calculation of account updates using ad-hoc smt forests.

crates/store/src/state/mod.rs
```diff
@@ -38,33 +38,40 @@ use tracing::{info, instrument};
...
     NULLIFIER_TREE_STORAGE_DIR,
     StorageLoader,
     TreeStorage,
-    load_mmr,
-    load_smt_forest,
-    verify_tree_consistency,
 };
+use loader::{load_mmr, load_smt_forest, verify_tree_consistency};
 
 mod apply_block;
```
drahnr, line 67: This makes it harder to review, but if we want further changes to `apply_block` - which we do - we should keep the sync on the extra file with `next` to avoid more pain when merging.

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -1252,3 +1252,46 @@ pub(crate) struct AccountStorageMapRowInsert {
...
+pub const HISTORICAL_BLOCK_RETENTION: u32 = 50;
+
+/// Clean up old entries for all accounts, deleting entries older than the retention window.
+///
+/// Deletes rows where `block_num < chain_tip - HISTORICAL_BLOCK_RETENTION` and `is_latest = false`.
+/// This is a simple and efficient approach that doesn't require window functions.
+///
+/// # Returns
+/// A tuple of `(vault_assets_deleted, storage_map_values_deleted)`
+pub fn cleanup_all_accounts(
```
Mirko-von-Leipzig: Organizationally, should we not have:

```rust
fn prune_history(..) {
    let cutoff_block = i64::from(chain_tip.as_u32().saturating_sub(HISTORICAL_BLOCK_RETENTION));

    prune_account_vault_assets()?;
    prune_account_storage_maps()?;
    // Probably more that we will need.
}

fn prune_account_vault_assets(...) {}

fn prune_account_storage_map(...) {}
```

and ideally we should also instrument each of these so we know how long each took.

drahnr: Implemented as suggested

PR #1646, opened 2026-02-06
feat(store): add typed error codes for GetAccount endpoint

### LINKED ISSUE #1641: Add typed error codes for `GetAccount` endpoint
`GetAccount` doesn't have typed error codes like other endpoints do. We should add a `GetAccountError` enum with `#[derive(GrpcError)]` following the pattern of `GetNotesByIdError`, `SyncAccountVaultError`, etc.

Related: 0xMiden/miden-client#1039

Himess:
Closes #1641


crates/store/src/errors.rs
```diff
@@ -489,6 +489,26 @@ pub enum GetBlockByNumberError {
     DeserializationFailed(#[from] DeserializationError),
 }
 
+// GET ACCOUNT ERRORS
+// ================================================================================================
+
+#[derive(Debug, Error, GrpcError)]
+pub enum GetAccountError {
```
huitseeker, line 496: The GrpcError macro maps all non-internal error variants to tonic::Code::InvalidArgument. Previously, DatabaseError::AccountNotPublic and AccountNotFoundInDb mapped to Status::not_found.
This means AccountNotFound and AccountNotPublic will return InvalidArgument instead of NotFound to clients.

That's fine since looking at the broader context, issue #1106 describes the intent that clients should parse error details (the u8 discriminant) rather than branch on gRPC status codes. 

However, if any existing clients branch on NOT_FOUND status code for account queries, this breaks them. So we should probably mark this change as BREAKING in the Changelog.

crates/store/src/state/mod.rs
```diff
@@ -674,26 +676,33 @@ impl State {
...
+            }
+        }
 
         // Query account header and storage header together in a single DB call
         let (account_header, storage_header) = self
             .db
             .select_account_header_with_storage_header_at_block(account_id, block_num)
             .await?
-            .ok_or(DatabaseError::AccountAtBlockHeightNotFoundInDb(account_id, block_num))?;
+            .ok_or(GetAccountError::AccountNotFound(account_id))?;
```
huitseeker: The old error `AccountAtBlockHeightNotFoundInDb(account_id, block_num)` included both the account ID and the block number. The new error `AccountNotFound(account_id)` only includes the account ID.

Per issue #1641 discussion, the client already knows the block number from the request. But for observability and debugging, having the block in logs helps - the message "account X not found" could mislead _operators_ (not the client) into thinking the account never existed rather than it's missing at a specific block height.

I'd consider including `BlockNumber` in the `AccountNotFound` variant, or changing the message to include block context.

crates/store/src/errors.rs
huitseeker, line 496: Nit: no tests cover the new `UnknownBlock` vs `BlockPruned` split or the gRPC status/details for `GetAccountError`.

The split logic in `get_account` correctly distinguishes future blocks from pruned ones by comparing against `latest_block`, but tests should ideally verify this works at the RPC boundary.

Suggested tests:
- Request a future block → expect `UnknownBlock`
- Request a pruned block → expect `BlockPruned`
- Request details for a private account → expect `AccountNotPublic`
- Request a missing account at a valid block → expect `AccountNotFound`

PR #1650, opened 2026-02-07
Improve tracing span fields 

huitseeker:
This PR switches some log messages to structured fields and to add small span fields when `skip_all` otherwise hides context. 

```text
before: INFO Applying 3 migration(s)
after:  INFO Applying migrations migrations=3
```

### DIFF
--- a/crates/utils/src/panic.rs
+++ b/crates/utils/src/panic.rs
@@ -4,6 +4,8 @@ use http::{Response, StatusCode, header};
 use http_body_util::Full;
 pub use tower_http::catch_panic::CatchPanicLayer;
 
+use crate::tracing::OpenTelemetrySpanExt;
+
 /// Custom callback that is used by Tower to fulfill the
 /// [`tower_http::catch_panic::ResponseForPanic`] trait.
 ///
@@ -13,6 +15,10 @@ pub fn catch_panic_layer_fn(err: Box<dyn Any + Send + 'static>) -> Response<Full
     // Log the panic error details.
     let err = stringify_panic_error(err);
     tracing::error!(panic = true, error = %err, "panic");
+
+    // Mark the current span as failed for OpenTelemetry.
+    let wrapped = anyhow::Error::msg(err.clone());
+    tracing::Span::current().set_error(wrapped.as_ref());
 
     // Return generic error response.
     Response::builder()
--- a/crates/utils/src/logging.rs
+++ b/crates/utils/src/logging.rs
@@ -10,6 +10,8 @@ use tracing_opentelemetry::OpenTelemetryLayer;
 use tracing_subscriber::layer::{Filter, SubscriberExt};
 use tracing_subscriber::{Layer, Registry};
 
+use crate::tracing::OpenTelemetrySpanExt;
+
 /// Global tracer provider for flushing traces on panic.
 ///
 /// This is necessary because the panic hook needs access to the tracer provider to flush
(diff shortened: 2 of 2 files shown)


crates/utils/src/panic.rs
```diff
@@ -11,7 +11,7 @@ pub use tower_http::catch_panic::CatchPanicLayer;
 pub fn catch_panic_layer_fn(err: Box<dyn Any + Send + 'static>) -> Response<Full<bytes::Bytes>> {
     // Log the panic error details.
     let err = stringify_panic_error(err);
-    tracing::error!(panic = true, "{err}");
+    tracing::error!(panic = true, error = %err, "panic");
```
Mirko-von-Leipzig, line 17: I'm unsure what OTel translates `error!` to in terms of exception message etc. Will check on Monday.

Since we now also have `[track_caller]` maybe we should also mark the span as failed by adding `tracing::Span::current().set_error(&err);` And similar for the other panic layer.

crates/ntx-builder/src/coordinator.rs
```diff
@@ -123,7 +123,10 @@ impl Coordinator {
 
         // If an actor already exists for this account ID, something has gone wrong.
         if let Some(handle) = self.actor_registry.remove(&account_id) {
-            tracing::error!("account actor already exists for account: {}", account_id);
+            tracing::error!(
+                account_id = %account_id,
```
Mirko-von-Leipzig, line 127: At some point I'll figure out a way to enforce standards for these. For OTel the idea is basically to name things similar to a json mapping (aka structs). So `account.id`, `account.commitment` etc.

This isn't done very thoroughly at the moment, and its a pain to "enforce" via PR reviews only. I was thinking maybe we make our own macros with known defaults so that passing in `&account_id` automatically sets the field name correctly (with overrides allowed).

```suggestion
                account.id = %account_id,
```

Mirko-von-Leipzig: Not really this PRs problem though, feel free to ignore these :)

crates/store/src/server/api.rs
Mirko-von-Leipzig, line 1: Was there a specific reason to include some of the fields? Or just because we may as well if they're small?

Mirko-von-Leipzig (approved):
Thank you! As mentioned in a comment, I think "enforcing" these could be done via our own set of macros which _doesn't_ allow string interpolation. At least that would discourage using the "log" this string approach which is easy to do.

crates/store/src/server/api.rs
huitseeker: Main reason is skip_all hides inputs, so I added small scalar fields (lengths/counts) to keep spans useful beyond their start/end. All are cheap (len() only).

crates/utils/src/panic.rs
huitseeker: `set_error` expects a real `Error`, whereas here we only have a string. I can wrap it with `anyhow::Error::msg(err.clone())` and call `Span::current().set_error(&wrapped)` (and do the same in the panic hook in crates/utils/src/logging.rs?)

crates/ntx-builder/src/coordinator.rs
huitseeker: IIUC

```zsh
huitseeker@gringolet➜~/tmp/miden-node(huitseeker-improve-tokio)» rg -A1 -B2 --pcre2 '(?<!let )\b\w+_(\w+) = .?\w+_\1,?$'
crates/ntx-builder/src/coordinator.rs
125-        if let Some(handle) = self.actor_registry.remove(&account_id) {
126-            tracing::error!(
127:                account_id = %account_id,
128-                "Account actor already exists"
--
170-            if let Err(err) = Self::send(handle, event.clone()).await {
171-                tracing::error!(
172:                    account_id = %account_id,
173-                    error = %err,

crates/store/src/inner_forest/mod.rs
346-            %account_id,
347-            %block_num,
348:            vault_entries = num_entries,
349-            "Inserted vault into forest"
--
431-            %account_id,
432-            %block_num,
433:            vault_entries = num_entries,
434-            "Updated vault in forest"
--
536-                %block_num,
537-                ?slot_name,
538:                delta_entries = num_entries,
539-                "Inserted storage map into forest"
```

And this is of course overly sensitive (e.g. doesn't specify tracing, just the assignment pattern).

crates/utils/src/panic.rs
Mirko-von-Leipzig: Yes please :+1:

PR #1651, opened 2026-02-07
Add track_caller to tracing helpers

### LINKED ISSUE #1618: Use `#[track_caller]` on log/trace helpers
Trace, spans and logs automatically fill in the current code file path and line numbers. This is great but if we're using a helper function this allocates the helper functions details instead of the caller.

For example, all our gRPC spans are created by our helper trace function and get

```
code.file.path  crates/utils/src/tracing/grpc.rs
```

which is less than helpful. Adding `#[track_caller]` would assign a more correct location.

huitseeker:
Fixes #1618.


Mirko-von-Leipzig:
> Did we want this to go into `main` as well?

Probably a good idea to have `main` and `next` track closely until we deploy on testnet?

PR #1652, opened 2026-02-07
fix: Replace writer mutex with semaphore gate

### LINKED ISSUE #1546: fix: get rid of the free-floating `Mutex`
There shouldn't be a free floating Mutex _outside_ of FFI code boundary code, ever. Modeling this better/worse will have a perf impact.

huitseeker:
Fixes #1546.


drahnr:
I don't think this covers the sentiment of the (arguably brief) issue, we want to not only change the type but refactor the approach on how we synchronize parallel streams (yes, the Mutex is a piece of it, but also channels) and revisit _what_ we want and can parallelize.

Issue #1628, opened 2026-02-02
Persist ntx state


Mirko-von-Leipzig:
I think we can simplify our ntx component by adding a Sqlite database to store the current state. This is desired in the long term to reduce store IO in any case, but I actually think this will simplify our state management dramatically.

Our data use already follows a relational pattern which will map naturally to sql database queries.

## Proposal

The database will hold the latest committed network account states and all _unconsumed_ network notes. In addition, the database will also contain all uncommitted network account deltas and notes/nullifiers aka it will also track inflight network stuff.

Whether committed and inflight state should be in separate tables is tbd, and depends largely on how complex queries/joins become.

We should also store the latest chain tip header.

### Start up

On startup the ntx syncs its local database from the canonical state in the store. This would comprise of:

- Purging all local inflight state left-over from the last run.
- Sync from local chain tip to store chain tip.
- We can now start the mempool subscription (and ensure our chain tip matches the mempool one).

### Run time

The database would have a single writer only - the main co-ordinator thread. Each mempool event received it writes the changes to the database.

- Transaction received: add new inflight state (account delta, note nullified, note created)
- Transaction reverted: revert the delta
- Block committed: mark data as committed. Nullified notes can be pruned (or kept if we want to have a bit of history), update chain tip

Instead of passing the mempool event on, we now just need to notify actor accounts if they should wake up (if they're asleep).

### Actors

Actors have a read-only database connection. They can query for available notes and decide on behaviour based on that.

When a transaction is emitted successfully, the actor _should sleep_ until the transaction event was received (actor will be notified by writer and can check the database for the transaction/account update).

### Schema

This will require a bit of design. We have committed notes & account state. And then inflight transactions linked to notes and account deltas.

SantiagoPittella:
- Do we prefer to have separate tables like `committed_accounts` + `inflight_acc_deltas`, `committed_notes` + `inflight_notes` with simpler queries, or fewer tables with a bit more complex queries?
- Should we make the local database of the component configurable via NtxBuilderConfig? Default to a path relative to the node's data directory?

Regarding an implementation plan, we can split this in tasks like:
1. First we migrate the schema and models. This changes are easy to check since everything shouldl still run as expected.
2. Then we can change the NTX builder component management migration.
3. And end up refactoring the component's `DataStore` implementation (`NtxDataStore`) to read committed account/note data from local DB

PR #1654, opened 2026-02-09
feat: add ntx-builder database

### LINKED ISSUE #1655: NTX Builder state(phase 1): Add database to the component
Phase 1 of the plan described in https://github.com/0xMiden/miden-node/issues/1628#issuecomment-3861069572

Add SQLite database infrastructure to the ntx-builder crate. This is the foundation for persisting NTX state locally, replacing the current fully in-memory approach.

SantiagoPittella:
This is the phase 1 of https://github.com/0xMiden/miden-node/issues/1628 

closes #1655

### DIFF
--- a/crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
+++ b/crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
@@ -11,106 +11,49 @@ CREATE TABLE chain_state (
     CONSTRAINT chain_state_block_num_is_u32 CHECK (block_num BETWEEN 0 AND 0xFFFFFFFF)
 );
 
--- One row per network account with its latest committed state.
-CREATE TABLE committed_accounts (
-    -- AccountId serialized bytes (8 bytes). Primary key.
-    account_id      BLOB    PRIMARY KEY,
-    -- Serialized Account.
-    account_data    BLOB    NOT NULL
-) WITHOUT ROWID;
-
--- Ordered chain of uncommitted account updates per account.
--- The auto-incrementing ID preserves insertion order (equivalent to VecDeque ordering).
-CREATE TABLE inflight_account_deltas (
+-- Account states: both committed and inflight.
+-- Committed rows have transaction_id = NULL. Inflight rows have transaction_id set.
+-- The auto-incrementing order_id preserves insertion order (VecDeque semantics).
+CREATE TABLE accounts (
     -- Auto-incrementing ID preserves insertion order.
-    id              INTEGER PRIMARY KEY AUTOINCREMENT,
-    -- FK to committed_accounts. May be NULL for accounts created by inflight txs.
+    order_id        INTEGER PRIMARY KEY AUTOINCREMENT,
+    -- AccountId serialized bytes (8 bytes).
     account_id      BLOB    NOT NULL,
-    -- The transaction that produced this delta.
-    transaction_id  BLOB    NOT NULL,
-    -- Serialized Account state after applying this delta.
-    account_data    BLOB    NOT NULL
+    -- Serialized Account state.
+    account_data    BLOB    NOT NULL,
+    -- NULL if this is the committed state; transaction ID if inflight.
+    transaction_id  BLOB
 );
 
-CREATE INDEX idx_inflight_deltas_account ON inflight_account_deltas(account_id);
+-- At most one committed row per account.
+CREATE UNIQUE INDEX idx_accounts_committed ON accounts(account_id) WHERE transaction_id IS NULL;
+CREATE INDEX idx_accounts_account ON accounts(account_id);
+CREATE INDEX idx_accounts_tx ON accounts(transaction_id) WHERE transaction_id IS NOT NULL;
 
--- Unconsumed notes from committed blocks, keyed by nullifier.
-CREATE TABLE committed_notes (
+-- Notes: committed, inflight, and nullified — all in one table.
+-- created_by = NULL means committed note; non-NULL means created by inflight tx.
+-- consumed_by = NULL means unconsumed; non-NULL means consumed by inflight tx.
+-- Row is deleted once consumption is committed.
+CREATE TABLE notes (
     -- Nullifier bytes (32 bytes). Primary key.
     nullifier       BLOB    PRIMARY KEY,
-    -- FK to the target account.
+    -- Target account ID.
     account_id      BLOB    NOT NULL,
-    -- Serialized SingleTargetNetworkNote (protobuf bytes).
+    -- Serialized SingleTargetNetworkNote.
     note_data       BLOB    NOT NULL,
     -- Backoff tracking: number of failed execution attempts.
     attempt_count   INTEGER NOT NULL DEFAULT 0,
     -- Backoff tracking: block number of the last failed attempt. NULL if never attempted.
     last_attempt    INTEGER,
+    -- NULL if the note came from a committed block; transaction ID if created by inflight tx.
+    created_by      BLOB,
+    -- NULL if unconsumed; transaction ID of the consuming inflight tx.
+    consumed_by     BLOB,
 
-    CONSTRAINT committed_notes_attempt_count_non_negative CHECK (attempt_count >= 0),
-    CONSTRAINT committed_notes_last_attempt_is_u32 CHECK (last_attempt BETWEEN 0 AND 0xFFFFFFFF)
+    CONSTRAINT notes_attempt_count_non_negative CHECK (attempt_count >= 0),
+    CONSTRAINT notes_last_attempt_is_u32 CHECK (last_attempt BETWEEN 0 AND 0xFFFFFFFF)
 ) WITHOUT ROWID;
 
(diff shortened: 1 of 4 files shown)


crates/ntx-builder/Cargo.toml
Mirko-von-Leipzig, line 1: These should probably become `workspace` deps

crates/ntx-builder/src/lib.rs
```diff
@@ -16,6 +16,7 @@ mod actor;
 mod block_producer;
 mod builder;
 mod coordinator;
+pub mod db;
```
Mirko-von-Leipzig: Does this need to be `pub`?

crates/ntx-builder/src/db/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::db::migrations::apply_migrations;
+
+pub mod errors;
+pub(crate) mod manager;
+
+mod migrations;
+mod schema_hash;
+
+/// [diesel](https://diesel.rs) generated schema.
+#[allow(unused)]
```
Mirko-von-Leipzig:
```suggestion
#[expect(unused, "will be used as part of follow-up work")]
```

crates/ntx-builder/src/db/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+impl Db {
+    /// Creates a new database, configures it, and applies migrations.
+    #[instrument(
+        target = COMPONENT,
+        name = "ntx_builder.database.bootstrap",
+        skip_all,
+        fields(path=%database_filepath.display()),
+        err,
+    )]
+    pub fn bootstrap(database_filepath: PathBuf) -> anyhow::Result<()> {
```
Mirko-von-Leipzig, line 39: I'm unsure what this is used for versus `load`.

crates/ntx-builder/src/db/migrations.rs
```diff
@@ -0,0 +1,29 @@
...
+use crate::db::schema_hash::verify_schema;
+
+// The rebuild is automatically triggered by `build.rs` as described in
+// <https://docs.rs/diesel_migrations/latest/diesel_migrations/macro.embed_migrations.html#automatic-rebuilds>.
+pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!("src/db/migrations");
+
+#[instrument(level = "debug", target = COMPONENT, skip_all, err)]
+pub fn apply_migrations(conn: &mut SqliteConnection) -> Result<(), DatabaseError> {
+    let migrations = conn.pending_migrations(MIGRATIONS).expect("In memory migrations never fail");
+    tracing::info!(target = COMPONENT, "Applying {} migration(s)", migrations.len());
```
Mirko-von-Leipzig:
```suggestion
    tracing::info!(target = COMPONENT, migrations=migrations.len(), "Applying migrations");
```

crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
```diff
@@ -0,0 +1,116 @@
...
+    id              INTEGER PRIMARY KEY AUTOINCREMENT,
+    -- FK to committed_accounts. May be NULL for accounts created by inflight txs.
+    account_id      BLOB    NOT NULL,
+    -- The transaction that produced this delta.
+    transaction_id  BLOB    NOT NULL,
+    -- Serialized Account state after applying this delta.
+    account_data    BLOB    NOT NULL
+);
+
+CREATE INDEX idx_inflight_deltas_account ON inflight_account_deltas(account_id);
```
Mirko-von-Leipzig: I imagine we'll also want an index on the `transaction_id` so we can revert/commit it.

crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
```diff
@@ -0,0 +1,116 @@
...
+
+-- Notes created by inflight transactions (not yet committed).
+CREATE TABLE inflight_notes (
+    -- Nullifier bytes (32 bytes). Primary key.
+    nullifier       BLOB    PRIMARY KEY,
+    -- FK to the target account.
+    account_id      BLOB    NOT NULL,
+    -- The transaction that created this note.
+    transaction_id  BLOB    NOT NULL,
+    -- Serialized SingleTargetNetworkNote (protobuf bytes).
```
Mirko-von-Leipzig: I would avoid mentioning protobuf here

crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
```diff
@@ -0,0 +1,116 @@
...
+-- Events cached for network accounts that haven't been spawned yet.
+CREATE TABLE predating_events (
+    -- Composite primary key.
+    account_id      BLOB    NOT NULL,
+    transaction_id  BLOB    NOT NULL,
+    -- Serialized MempoolEvent (protobuf bytes).
+    event_data      BLOB    NOT NULL,
+
+    PRIMARY KEY (account_id, transaction_id)
+) WITHOUT ROWID;
```
Mirko-von-Leipzig: These won't be needed. Reverting and committing a transaction can be done by looking up the transaction ID in the linked tables. Predating events should be possible to store as since these are just _unconsumed_ network notes without an existing account.

crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
Mirko-von-Leipzig, line 1: Looking at this now, it does feel like merging inflight + committed shouldn't be so bad, at least for notes + nullifiers.

I think we can also combine the notes and nullifiers (both inflight and committed) in a single table as:

```sql
CREATE TABLE notes (
  id                -- Note ID
  nullifier         -- Note nullifier
  account_id        -- Target network account
  data              -- Network note data
  attempt_count     -- Failed number of attempts
  last_attempt      -- Last attempt's block number
  created_by        -- Inflight transaction ID (null if committed)
  consumed_by       -- Inflight transaction ID that consumed it
                    --    - Null if unconsumed
                    --    - Row is removed once nullifier is committed
);
```

We can do something similar for the accounts table using your auto-increment trick for ordering:

```sql
CREATE TABLE accounts (
    order_id        -- auto-increment allows chronological ordering
    id              -- network account ID
    data            -- Encoded account data
    transaction_id  -- Inflight tx ID
                    --   - Null if committed
                    --   - Ideally only one committed amount should be kept
                    --   - Reorgs will be a problem, but we can just delete db
);
```

wdyt? Does this work and/or make sense?

crates/ntx-builder/src/db/errors.rs
```diff
@@ -0,0 +1,69 @@
...
+    ConnectionPoolObtainError(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
+    #[error(transparent)]
+    Diesel(#[from] diesel::result::Error),
+    #[error("SQLite pool interaction failed: {0}")]
+    InteractError(String),
+    #[error("schema verification failed")]
+    SchemaVerification(#[from] SchemaVerificationError),
+    #[error(transparent)]
+    ConnectionManager(#[from] ConnectionManagerError),
+}
```
Mirko-von-Leipzig, line 20: Let's remove the `#[from]`

crates/ntx-builder/src/db/errors.rs
```diff
@@ -0,0 +1,69 @@
...
+// ================================================================================================
+
+#[derive(Debug, thiserror::Error)]
+pub enum DatabaseSetupError {
+    #[error("I/O error")]
+    Io(#[from] std::io::Error),
+    #[error("database error")]
+    Database(#[from] DatabaseError),
+    #[error("pool build error")]
+    PoolBuild(#[from] deadpool::managed::BuildError),
```
Mirko-von-Leipzig: Same here `-#[from]` `+#[source]`

crates/ntx-builder/src/lib.rs
SantiagoPittella: There's no need. I made it pub(crate)

crates/ntx-builder/src/db/mod.rs
SantiagoPittella: I improved the docs, this is the difference:

- `bootstrap` is a synchronous one-shot function for node initialization. it creates the database file, configures it, and applies migrations using a single direct SqliteConnection (no pool). It's meant to be called once before the async runtime to ensure the DB exists and is up to date.

- `load` is the async runtime counterpart  and opens a connection pool (deadpool) and returns a Db struct for concurrent query access. It also re-applies migrations (which is a no-op if bootstrap already ran them).

crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
SantiagoPittella: Yes! I like it, it is way more simple

Mirko-von-Leipzig (approved):
Merging as is; I'm sure we'll stumble into other minor issues as part of implementing the follow-up phases.

PR #1660, opened 2026-02-10
feat: network monitor transport layer integration

SantiagoPittella:
> [!WARNING]  
> This PR aims to main.

This PR adds note transport layer support for the network monitor. It works by using the health check endpoint of the given service (https://github.com/0xMiden/miden-note-transport/blob/main/crates/node/src/node/grpc/mod.rs#L87)

Also, add a copy button for the URLs of the services to be easily accesible and re-arranges the order of the cards.


bobbinth:
> The issue is that I refactored `tests` workflow into `ci/tests` job and the branch protection rules for `main` still reference `tests`. This was fixed for `next` but has not yet been done for `main` cc @bobbinth

Should be all good now.

Issue #1656, opened 2026-02-09
NTX Builder state(phase 2): management migration


SantiagoPittella:
Replace in-memory state management with database-backed operations. The coordinator becomes the single DB writer (processing mempool events), and actors become readers (querying DB for account state and available notes).

PR #1662, opened 2026-02-10
chore(ntx): replace in memory with sqlite database

SantiagoPittella:
phase 2 of https://github.com/0xMiden/miden-node/issues/1628
closes https://github.com/0xMiden/miden-node/issues/1656


crates/ntx-builder/src/lib.rs
```diff
@@ -195,8 +205,19 @@ impl NtxBuilderConfig {
     /// - The mempool subscription fails (after retries)
     /// - The store contains no blocks (not bootstrapped)
     pub async fn build(self) -> anyhow::Result<NetworkTransactionBuilder> {
+        // Bootstrap and load the database.
+        Db::bootstrap(self.database_filepath.clone())
+            .context("failed to bootstrap ntx-builder database")?;
+        let db = Db::load(self.database_filepath.clone())
+            .await
+            .context("failed to load ntx-builder database")?;
```
Mirko-von-Leipzig: I think this is what confused me in the previous PR. Should `bootstrap` and `load` not just be a single function?

crates/ntx-builder/src/actor/account_effect.rs
```diff
@@ -0,0 +1,73 @@
...
+    #[case::no_attempts(None, BlockNumber::GENESIS, 0, true)]
+    #[case::one_attempt(Some(BlockNumber::GENESIS), BlockNumber::from(2), 1, true)]
+    #[case::three_attempts(Some(BlockNumber::GENESIS), BlockNumber::from(3), 3, true)]
+    #[case::ten_attempts(Some(BlockNumber::GENESIS), BlockNumber::from(13), 10, true)]
+    #[case::twenty_attempts(Some(BlockNumber::GENESIS), BlockNumber::from(149), 20, true)]
+    #[case::one_attempt_false(Some(BlockNumber::GENESIS), BlockNumber::from(1), 1, false)]
+    #[case::three_attempts_false(Some(BlockNumber::GENESIS), BlockNumber::from(2), 3, false)]
+    #[case::ten_attempts_false(Some(BlockNumber::GENESIS), BlockNumber::from(12), 10, false)]
+    #[case::twenty_attempts_false(Some(BlockNumber::GENESIS), BlockNumber::from(148), 20, false)]
+    fn backoff_has_passed(
```
Mirko-von-Leipzig: Is this test in the right file?

crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
```diff
@@ -27,6 +27,9 @@ CREATE TABLE accounts (
 
 -- At most one committed row per account.
 CREATE UNIQUE INDEX idx_accounts_committed ON accounts(account_id) WHERE transaction_id IS NULL;
+-- At most one inflight row per (account, transaction) pair.
+CREATE UNIQUE INDEX idx_accounts_inflight ON accounts(account_id, transaction_id)
+    WHERE transaction_id IS NOT NULL;
```
Mirko-von-Leipzig, line 32: We probably want to allow this to happen. As in:

- network tx is submitted
    - account actor now goes to sleep
- mempool receives tx and emits mempool event
- we receive the event here
    - account now has an inflight tx
- we should be able to build and submit another tx on top of this
    - otherwise we have to wait for this tx to be committed first

crates/ntx-builder/src/db/models/queries/chain_state.rs
```diff
@@ -0,0 +1,49 @@
...
+// MODELS
+// ================================================================================================
+
+#[derive(Debug, Clone, Insertable)]
+#[diesel(table_name = schema::chain_state)]
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+pub struct ChainStateInsert {
+    /// Singleton row ID. Always `Some(0)` to satisfy the `CHECK (id = 0)` constraint.
+    ///
+    /// Mapped as `Option<i32>` because SQLite's `INTEGER PRIMARY KEY` maps to `Nullable<Integer>`
```
Mirko-von-Leipzig: We should add `NOT NULL` to the `PRIMARY KEY` definition (which should fix this).

PK should imply not null but apparently this is due to an early sqlite bug which is kept for backwards compatibility.

crates/ntx-builder/src/db/models/queries/accounts.rs
```diff
@@ -0,0 +1,92 @@
...
+///
+/// ```sql
+/// INSERT OR REPLACE INTO accounts (account_id, account_data, transaction_id)
+/// VALUES (?1, ?2, NULL)
+/// ```
+pub fn upsert_committed_account(
+    conn: &mut SqliteConnection,
+    account_id: NetworkAccountId,
+    account: &Account,
+) -> Result<(), DatabaseError> {
```
Mirko-von-Leipzig, line 53: This is a bit different to how I imagined it in my head. That's not to say this is incorrect; I just want to make sure I understand correctly.

I _thought_ we would do

```rust
// Technically we could also elide `account_id` by using
// DELETE FROM ... RETURNING account_id; but that seems like more work.
pub fn commit(conn, tx_id, account_id) {
    // -- Remove the previously committed account row
    // DELETE FROM accounts WHERE account_id=account_id AND transaction_id IS NULL;
    // -- Mark the new tx's account state as committed
    // UPDATE accounts SET transaction_id=NULL WHERE transaction_id=tx_id;
}
```

which would also keep the `order_id` in the same order it was originally since the new committed state retains its `order_id`.

Your impl relies on a unique index, which I'm unfamiliar with using like this. Does it retain the original's `order_id`? I think this may also have a subtle bug for newly created accounts iiuc:

```
                           order_id       transaction_id
- New network account:     0              A
- Another tx update        1              B
- A committed              2              NULL
```

Now `latest_account` will return `order_id=2` instead of `order_id=1` as the latest state. This doesn't occur for non-new accounts because they'll always have a `transaction_id=NULL` row at the very start.

crates/ntx-builder/src/actor/account_effect.rs
SantiagoPittella: Moved it to actor/mod.rs where the function is defined

crates/ntx-builder/src/lib.rs
SantiagoPittella: I merged into a single `Db::setup()` that runs the sync bootstrap via spawn_blocking and then opens the async connection pool.

crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
SantiagoPittella: The index already allows multiple inflight rows per account, and the uniqueness is on the (account_id, transaction_id) pair, so each transaction gets its own inflight row.

crates/ntx-builder/src/db/models/queries/accounts.rs
SantiagoPittella: In the original query that I added, SQLite detects the conflict on `idx_accounts_committed`, deletes the old row, and inserts a new one with a higher order_id. The explicit DELETE + INSERT does the exact same thing, but mroe explicitly. I will change to your approach. This was basically handled in `handle_block_commited`, in particular in:

```
            // Promote the inflight row to committed (set transaction_id = NULL).
            // Only promote the row for this specific tx.
            diesel::update(
                schema::accounts::table
                    .filter(schema::accounts::account_id.eq(account_id_bytes))
                    .filter(schema::accounts::transaction_id.eq(&tx_id_bytes)),
            )
```

crates/ntx-builder/src/db/migrations/2026020900000_setup/up.sql
Mirko-von-Leipzig: Ah. My bad. Thank you!

crates/ntx-builder/src/db/models/queries/accounts.rs
Mirko-von-Leipzig: > deletes the old row, and inserts a new one with a higher order_id. The explicit DELETE + INSERT does the exact same thing, but mroe explicitly.

I think the key difference (and only difference) is that the latter _does not_ insert a new row with a higher `order_id`. The former inserting such a row is a bug because it means the `order_id` is no longer chronological for that account. i.e. promoting `tx[i]` would make it the newest entry ito `order_id` which is incorrect.

Mirko-von-Leipzig: e.g.

```
order_id | before  | after (latter)     |  after (former)
------------------------------------------------------
0           tx A      <deleted>            tx A (aka NULL)
1           tx B      tx B                 tx B
2           tx C      tx C                 tx C
3           <empty>   tx A (aka NULL)      <empty>
```

and therefore

```sql
SELECT account FROM accounts .. ORDER BY order_id DESC;

former: tx C (correct)
latter: tx A (incorrect)
```

crates/ntx-builder/src/db/models/conv.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[allow(clippy::cast_possible_wrap)]
+pub fn block_num_to_i32(block_num: BlockNumber) -> i32 {
+    block_num.as_u32() as i32
+}
+
+#[allow(clippy::cast_sign_loss)]
+pub fn block_num_from_i32(val: i32) -> BlockNumber {
+    BlockNumber::from(val as u32)
+}
```
Mirko-von-Leipzig: These should probably be `i64` and not `i32`

crates/ntx-builder/src/db/mod.rs
```diff
@@ -20,38 +30,63 @@ pub(crate) mod schema;
...
+        info!(
+            target: COMPONENT,
+            sqlite = %database_filepath.display(),
+            "Connected to the database"
+        );
+
+        let me = Db { pool };
+        me.query("migrations", apply_migrations)
+            .await
+            .context("failed to apply migrations on pool connection")?;
```
Mirko-von-Leipzig: This is now redundant since its done within `bootstrap`.

crates/ntx-builder/src/db/mod.rs
```diff
@@ -20,38 +30,63 @@ pub(crate) mod schema;
...
+        name = "ntx_builder.database.setup",
         skip_all,
         fields(path=%database_filepath.display()),
         err,
     )]
-    pub fn bootstrap(database_filepath: PathBuf) -> anyhow::Result<()> {
+    pub async fn setup(database_filepath: PathBuf) -> anyhow::Result<Self> {
+        // Run sync bootstrap (create file, configure, migrate) on a blocking thread.
+        let path = database_filepath.clone();
+        tokio::task::spawn_blocking(move || Self::bootstrap_sync(&path))
```
Mirko-von-Leipzig: afaiu bootstrap does three things

1. Open a single database connection
2. Configure that connection
3. Apply migrations (sync)

I don't understand why we need it? Given that this function then essentially does the same thing (I think), just asyncly?

As in, if we keep this function as is but delete this `bootstrap_sync` what changes?

crates/ntx-builder/src/db/mod.rs
```diff
@@ -86,36 +123,164 @@ impl Db {
...
+
+    /// Creates an in-memory SQLite connection for testing with migrations applied.
+    ///
+    /// This bypasses the async connection pool entirely, matching the store crate's test pattern.
+    #[cfg(test)]
+    pub fn test_conn() -> SqliteConnection {
+        use crate::db::manager::configure_connection_on_creation;
+
+        let mut conn =
+            SqliteConnection::establish(":memory:").expect("in-memory sqlite should always work");
```
Mirko-von-Leipzig: Having done this myself before, this can actually be "dangerous". iirc the problem is that by default what happens is that `":memory:"` creates an in-memory file handle - but if you do this concurrently then that handle is shared. So concurrent tests will "connect" to the same in-memory database.

One solution is to [do this](https://github.com/eqlabs/pathfinder/blob/82d9a47179687ee8f60f7f3560861f8affeecbc3/crates/storage/src/lib.rs#L209-L218):

```rust
        // Create a unique database name so that they are not shared between
        // concurrent tests. i.e. Make every in-mem Storage unique.
        static COUNT: std::sync::Mutex<u64> = std::sync::Mutex::new(0);
        let unique_mem_db = {
            let mut count = COUNT.lock().unwrap();
            // &cache=shared allows other threads to see and access the inmemory database
            let unique_mem_db = format!("file:memdb{count}?mode=memory&cache=shared");
            *count += 1;
            unique_mem_db
        };

        let database_path = PathBuf::from(unique_mem_db);
        // This connection must be held until a pool has been created, since an
        // in-memory database is dropped once all its connections are. This connection
        // therefore holds the database in-place until the pool is established.
        let conn = rusqlite::Connection::open(&database_path)?;
```

But as the one that wrote that code, these days I would ignore that (complexity) and use temp files instead:

```rust
impl Db {
    fn for_testing() -> Self {
        // Init using tempdir & friends to create a unique tmp filepath.
    }
}
```

crates/ntx-builder/src/db/mod.rs
Mirko-von-Leipzig, line 1: Something more general that we need to discuss is database writes. This was something I had thought about but now that I'm looking at the code we should maybe align on how to approach this.

We also don't need to address this fully in this PR, since this is already a large delta so we should probably punt this to phase 3. But lets see what you think.

### The problem

We have the coordinator and actors which can perform db writes. We need to know which party is responsible for writing which data.

Broadly speaking we have:

1. Sync account & note data from the store component
2. Insert transaction
3. Commit transaction
4. Revert transaction
5. Update note attempts & potentially delete

(1) is part of `init` which is always done by an exclusive writer.

(2-4) impacts multiple accounts at once and should therefore (imo) always be handled by the coordinator.

(5) I initially thought would be done by the actors themselves but now I'm less sure of that since it can lead to conflicts i.e. some notes are reverted or committed by the coordinator, or the note counts were updated which may cause the coordinator tx to fail? We can work around this though, by knowing that its possible for these to fail, and to simply retry them each time..

An alternative is that we say the coordinator is the _sole_ db writer. This is trivial for (1-4), but (5) would require channel from each actor to the coordinator to submit "bad" notes to bump. I'm actually somewhat partial to this approach now because it means less thinking about what can go wrong. The idea would be that after each tx attempt the actor submits `(failed notes + oneshot)` and waits for that to be processed before proceeding with operations.

wdyt? Or maybe I'm overthinking this, and what actually happens is:

```sql
-- Maybe this never conflicts because the `WHERE` clause uses `in`, so if
-- rows are deleted this doesn't count?
UPDATE notes SET attempt_count=attempt_count+1, last_attempt=?
    WHERE nullifier in ?;

-- And similarly because the coordinator update doesn't touch those columns
-- it might not conflict? I think it will though..
```

crates/ntx-builder/src/db/mod.rs
SantiagoPittella: Ohhh uh. Ok, thanks! Will change it then.

crates/ntx-builder/src/db/mod.rs
SantiagoPittella: I removed `bootstrap_sync` the pool's `ConnectionManager::create()` already configures each connection, and migrations are applied through the pool. This was redundant

crates/ntx-builder/src/db/mod.rs
SantiagoPittella: Yeah, I agree. I have been thinking around this recently, and I'm not a super fun of having multiple write places. I prefer to move it to the coordinator despite the need to a connection to the actor to notify this. If it is ok with you, I would tackle this in phase 3.

Mirko-von-Leipzig: Sure that's good 👍

crates/ntx-builder/src/db/models/queries/accounts.rs
```diff
@@ -0,0 +1,104 @@
...
+        .filter(schema::accounts::account_id.eq(&account_id_bytes))
+        .order(schema::accounts::order_id.desc())
+        .select(AccountRow::as_select())
+        .first(conn)
+        .optional()?;
+
+    match row {
+        Some(r) => Ok(Some(conversions::account_from_bytes(&r.account_data)?)),
+        None => Ok(None),
+    }
```
Mirko-von-Leipzig:
```suggestion
    row.map(conversions::account_from_bytes).transpose()
```

crates/ntx-builder/src/db/models/queries/accounts.rs
Mirko-von-Leipzig: Actually maybe I've misunderstood this the entire time 🤦 

Is this function _only_ used during initialization when loading accounts from the store? If so, then none of what I said before matters..

crates/ntx-builder/src/db/models/queries/accounts.rs
```diff
@@ -0,0 +1,104 @@
...
+/// # Raw SQL
+///
+/// ```sql
+/// SELECT account_data
+/// FROM accounts
+/// WHERE account_id = ?1
+/// ORDER BY order_id DESC
+/// LIMIT 1
+/// ```
+pub fn latest_account(
```
Mirko-von-Leipzig:
```suggestion
pub fn get_account(
```

crates/ntx-builder/src/db/models/queries/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+/// ```
+///
+/// Per nullifier (marks notes as consumed):
+///
+/// ```sql
+/// UPDATE notes
+/// SET consumed_by = ?1
+/// WHERE nullifier = ?2 AND consumed_by IS NULL
+/// ```
+pub fn handle_transaction_added(
```
Mirko-von-Leipzig:
```suggestion
pub fn add_transaction(
```

crates/ntx-builder/src/db/models/queries/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+///
+/// -- Delete consumed notes
+/// DELETE FROM notes WHERE consumed_by = ?1
+///
+/// -- Promote inflight-created notes to committed
+/// UPDATE notes SET created_by = NULL WHERE created_by = ?1
+/// ```
+///
+/// Finally updates chain state (see [`upsert_chain_state`]).
+pub fn handle_block_committed(
```
Mirko-von-Leipzig:
```suggestion
pub fn commit_block(
```

crates/ntx-builder/src/db/models/queries/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+/// -- Check if account creation was fully reverted
+/// SELECT COUNT(*) FROM accounts WHERE account_id = ?1
+///
+/// -- Delete inflight-created notes
+/// DELETE FROM notes WHERE created_by = ?1
+///
+/// -- Restore consumed notes
+/// UPDATE notes SET consumed_by = NULL WHERE consumed_by = ?1
+/// ```
+pub fn handle_transactions_reverted(
```
Mirko-von-Leipzig:
```suggestion
pub fn revert_transaction(
```

crates/ntx-builder/src/db/models/queries/accounts.rs
SantiagoPittella: Yes, it runs before the event loop starts. I think that it achieves the same result with more clarity though

PR #1663, opened 2026-02-11
fix: add ca-certificates to node Docker image for TLS support

### LINKED ISSUE #1661: Node docker image cannot use TLS
This is because the image base excludes the TLS system libs which in turn does not allow the image to connect to `https` externally.

This is a problem for multiple reasons, including the inability to connect to provers or export telemetry which use https (which should be all/most).

We can either choose a less cut back base image, or figure out which libs are required as part of the base to enable TLS/CA support.

avorylli:
Fixes #1661

Adds `ca-certificates` to the node Docker runtime image to enable TLS/HTTPS connections. Required for OpenTelemetry telemetry export and remote prover HTTPS connections.

### DIFF
--- a/bin/node/Dockerfile
+++ b/bin/node/Dockerfile
@@ -31,7 +31,9 @@ RUN cargo build --release --locked --bin miden-node
 FROM debian:bullseye-slim AS runtime-base
 RUN apt-get update && \
     apt-get -y upgrade && \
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig:
Thanks but this is not sufficient. In fact I think the docker image currently doesn't work at all.

PR #1665, opened 2026-02-11
chore/makefile: add `stress-test-smoke` target

drahnr:
We didn't have a good way of invoking it, this is convenience. Allows to override the data-dir with env `STRESS_TEST_DATA_DIR` or uses a default, timestamped one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ install-node: ## Installs node
 install-remote-prover: ## Install remote prover's CLI
 	$(BUILD_PROTO) cargo install --path bin/remote-prover --bin miden-remote-prover --features concurrent --locked
 
-.PHONY: stress-test
+.PHONY: stress-test-smoke
 stress-test: ## Runs stress-test benchmarks
 	${BUILD_PROTO} cargo build --release --locked -p miden-node-stress-test
 	@mkdir -p $(STRESS_TEST_DATA_DIR)


Makefile
```diff
@@ -108,9 +109,14 @@ install-node: ## Installs node
...
-install-stress-test: ## Installs stress-test binary
-	cargo install --path bin/stress-test --locked
+.PHONY: stress-test
+stress-test: ## Runs stress-test benchmarks
+	${BUILD_PROTO} cargo build --release --locked -p miden-node-stress-test
+	@mkdir -p $(STRESS_TEST_DATA_DIR)
+	./target/release/miden-node-stress-test seed-store --data-directory $(STRESS_TEST_DATA_DIR) --num-accounts 500 --public-accounts-percentage 50
+	./target/release/miden-node-stress-test benchmark-store --data-directory $(STRESS_TEST_DATA_DIR) --iterations 10 --concurrency 1 sync-state
+	./target/release/miden-node-stress-test benchmark-store --data-directory $(STRESS_TEST_DATA_DIR) --iterations 10 --concurrency 1 sync-notes
+	./target/release/miden-node-stress-test benchmark-store --data-directory $(STRESS_TEST_DATA_DIR) --iterations 10 --concurrency 1 sync-nullifiers --prefixes 10
```
Mirko-von-Leipzig, line 119: I think a problem here is that you need to control the iteration and concurrency etc. This isn't really a "just run it" kind of command.

The CI job for this (with these values) basically just ensures that the binary can still be run properly.

drahnr: What I really want, is a smoke test locally, so I don't have to check CI for feedback. Maybe renaming the makefile target is a better option?

Mirko-von-Leipzig: I inherently dislike the makefile 😁 Maybe I'm just too used to having an old junky laptop that overheats when I compile everything so I never bother.

drahnr: > I inherently dislike the makefile

Me too, but after all, it's a mean of unification, which bottom-line is positive.

> Maybe I'm just too used to having an old junky laptop that overheats

Used to that too (at one time, I only had an intel i3 rooted chromebook), but I long switched to working on remote, beefy machines, which beats both UX and speed compared to our CI systems.

PR #1666, opened 2026-02-12
feat: enable coredump and debug symbols in Debian packages 

### LINKED ISSUE #1639: Add `debuginfo` to our systemd binaries
This will help debug core dumps.

Mirko-von-Leipzig:
This PR improves the debuggability of our Debian packages.

It adds debug symbols by ~~adding a new cargo profile `release-with-debug` which is used to install the binaries during package production.~~
adding them to the `release` profile.

It enables coredumps by setting `LimitCORE=infinity`. This sets the maximum coredump file size to inifinity. The default is `0` which of course means no dump file at all. This doesn't guarantee success, since there can still be additional system limitations set by the OS, but this at least gives us a chance.

Closes #1639

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,9 @@ version      = "0.13.4"
 [profile.test.package.miden-crypto]
 opt-level = 2
 
+[profile.release]
+debug = true
+
 [workspace.dependencies]
 # Workspace crates.
 miden-node-block-producer   = { path = "crates/block-producer", version = "0.13" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -114,3 +114,7 @@ must_use_candidate           = "allow" # This marks many fn's which isn't helpfu
 needless_for_each            = "allow" # Context dependent if that's useful.
 should_panic_without_expect  = "allow" # We don't care about the specific panic message.
 # End of pedantic lints.
+
+[profile.release-with-debug]
```
drahnr: I'd argue we want debug symbols always for releases

drahnr: In fedora, `*-debuginfo` packages are created automatically which contain the debug info only and can be fetched on demand, I am not sure about how debian handles that.

Mirko-von-Leipzig: Maybe you're right; I'll just update `release` profile instead.

PR #1668, opened 2026-02-13
chore(lint): prefer `#[expect]` over `#[allow]`

Mirko-von-Leipzig:
This PR enables the [clippy::allow_attributes](https://rust-lang.github.io/rust-clippy/master/index.html?versions=lte%3A70#allow_attributes) lint.

This denies `#[allow]` in favor of `#[expect]` where the latter is almost a universal improvement over the former. Notably the latter emits a warning once the expectation is no longer fulfilled, allowing us to remove lint-opt-outs which are no longer applicable.

The affected file set is large, but the deltas are minimal. Mostly just renaming `allow` to `expect` and removing those that are no longer valid.


bin/remote-prover/src/generated/mod.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(clippy::pedantic, reason = "generated by build.rs and tonic")]
+#![allow(clippy::allow_attributes, reason = "generated by build.rs and tonic")]
```
drahnr, line 2: Does it make sense in the context of generated files?

Mirko-von-Leipzig: We don't own that code - it gets generated 🤷 And the generated code uses `#[allow]` and therefore triggers the lint. What we really want to say is that clippy should ignore the entire module 🤷 So I guess we can do that instead?

One of the side effects of checking generated code into source.

PR #1669, opened 2026-02-13
fix/sqlite: ensure we're shipping the node binary with sqlite

drahnr:
Also addresses test failures uncovered with the newer sqlite version.

### DIFF
--- a/crates/store/src/db/models/queries/accounts/tests.rs
+++ b/crates/store/src/db/models/queries/accounts/tests.rs
@@ -654,7 +654,7 @@ fn test_select_account_vault_at_block_historical_with_updates() {
     );
 
     for block in [block_1, block_2, block_3] {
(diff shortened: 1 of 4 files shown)


crates/store/src/db/models/queries/accounts/tests.rs
```diff
@@ -654,7 +654,7 @@ fn test_select_account_vault_at_block_historical_with_updates() {
     );
 
     for block in [block_1, block_2, block_3] {
-        upsert_accounts(&mut conn, &[account_update.clone()], block)
+        upsert_accounts(&mut conn, std::slice::from_ref(&account_update), block)
```
Mirko-von-Leipzig, line 657: TIL -- never knew about `slice::from_ref`

drahnr: I just learned about it too

PR #1670, opened 2026-02-13
[main] fix/sqlite: ensure we're shipping the node binary with sqlite 

drahnr:
> [!IMPORTANT]
> Targets branch `main`

Adapted version of  #1669


crates/store/Cargo.toml
Mirko-von-Leipzig, line 1: What's happening here that its so different?

drahnr: We migrated to workspace dependencies as part Serge's PR

PR #1674, opened 2026-02-14
fix: make build.rs example compile by adding anyhow imports and Resul…

meelon-dev:
### Summary
The `build.rs` example in `proto/README.md` was not compiling:

- `?` was used in `main()` without `main()` returning a `Result`.
- `.context(...)` was used, but `anyhow::Context` was not imported.

### Changes
- Added `use anyhow::{Context, Result};` to imports.
- Changed `fn main()` to `fn main() -> Result<()>`.
- Added `Ok(())` at the end of `main()` to satisfy the Result return type.

### Reason
This makes the example compile correctly and demonstrates modern, idiomatic Rust error handling using `anyhow`.

### Notes
- Only README/example code is affected.
- Works with users’ projects using `anyhow` and `?` for concise error propagation.

### DIFF
--- a/proto/README.md
+++ b/proto/README.md
@@ -13,8 +13,9 @@ generates the RPC component bindings and writes them into the `src/generated` di
 ```rust
 use std::{fs, path::PathBuf, env};
 use miden_node_proto_build::rpc_api_descriptor;
+use anyhow::{Context, Result};
 
-fn main() {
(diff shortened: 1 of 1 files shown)


sergerad:
@Mirko-von-Leipzig does this readme need to spell out a `build.rs` contents at all?

PR #1675, opened 2026-02-14
fix: implement submit_proven_batch endpoint

avorylli:
Implements the submit_proven_batch function that was previously a todo!(). 

Adds a new add_proven_batch method to the mempool to handle user-submitted proven batches that bypass the internal proposed batch stage. Extends SubmitProvenBatchError to handle expiration and other errors. 

The implementation follows the proto documentation which allows batches with transactions that may or may not be in the mempool.


bobbinth:
Thank you - but handling proven batches requires some more design work. Implementing this functionality is a bit premature and we'll come back to a bit later.

PR #1677, opened 2026-02-16
feat: KMS BlockSigner

sergerad:
## Context

The genesis bootstrap and validator processes currently only support ECDSA private keys via command line. We want to be able to perform genesis and validator block signing through more secure, remote backends such as AWS KMS.

Relates to #1316.

Tested locally against AWS KMS (both bootstrap and chain running with separate validator).

## Changes
- Add `BlockSigner` trait (removed from miden-base).
- Add `KmsSigner` struct which implements `BlockSigner`.
- Update all command scaffolding to allow for bootstrap command and validator command to use either local key or KMS key.
- Bump toolchain to Rust 1.91 (for latest aws SDK support).


bin/node/src/commands/bundled.rs
Mirko-von-Leipzig, line 1: We should be more specific with the args. We can use claps' `ArgGroup` to specify relations between arguments (https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#argument-relations).

i.e. that we have 

```
miden-node --validator.key.kms = ..
# Which is mutually exclusive with
miden-node --validator.key.hex = ..
```

bin/node/src/commands/mod.rs
```diff
@@ -40,6 +40,7 @@ const ENV_MAX_BATCHES_PER_BLOCK: &str = "MIDEN_MAX_BATCHES_PER_BLOCK";
 const ENV_MEMPOOL_TX_CAPACITY: &str = "MIDEN_NODE_MEMPOOL_TX_CAPACITY";
 const ENV_NTX_SCRIPT_CACHE_SIZE: &str = "MIDEN_NTX_DATA_STORE_SCRIPT_CACHE_SIZE";
 const ENV_VALIDATOR_KEY: &str = "MIDEN_NODE_VALIDATOR_KEY";
+const ENV_VALIDATOR_KMS_KEY_ID: &str = "MIDEN_NODE_VALIDATOR_KMS_KEY_ID";
```
Mirko-von-Leipzig, line 47:
```suggestion
const ENV_VALIDATOR_KEY_KMS: &str = "MIDEN_NODE_VALIDATOR_KEY_KMS";
```

Mirko-von-Leipzig: And the one above to `_HEX`.

bin/node/src/commands/bundled.rs
Mirko-von-Leipzig: Alternatively, we can make it an enum with a string format..

```
# Hex key with 0x prefix
validator.key = "0xABCDEFG"

# Otherwise default to kms parsing
validator.key = "..."

# Or if that's unsafe because kms key prefix _could_ clash
validator.key = "kms:<key>"
```

Mirko-von-Leipzig: I feel like the enum might be better tbh. And we can add a `ValueEnum` parser to handle it as part of clap.

CHANGELOG.md
```diff
@@ -19,7 +19,8 @@
 - Refactored NTX Builder startup and introduced `NtxBuilderConfig` with configurable parameters ([#1610](https://github.com/0xMiden/miden-node/pull/1610)).
 - Refactored NTX Builder actor state into `AccountDeltaTracker` and `NotePool` for clarity, and added tracing instrumentation to event broadcasting ([#1611](https://github.com/0xMiden/miden-node/pull/1611)).
 - Add #[track_caller] to tracing/logging helpers ([#1651](https://github.com/0xMiden/miden-node/pull/1651)).
-- Improved tracing span fields ([#1650](https://github.com/0xMiden/miden-node/pull/1650))
+- Improved tracing span fields ([#1650](https://github.com/0xMiden/miden-node/pull/1650)).
```
```suggestion
```

crates/validator/src/signers/kms.rs
```diff
@@ -0,0 +1,107 @@
...
+        let client = aws_sdk_kms::Client::new(&config);
+        let key_id = key_id.into();
+
+        // Retrieve DER-encoded SPKI.
+        let pub_key_output = client.get_public_key().key_id(key_id.clone()).send().await?;
+        let spki_der = pub_key_output.public_key().ok_or(KmsSignerError::EmptyBlob)?.as_ref();
+
+        // Decode the DER-encoded SPKI and compress it.
+        let kpub = K256PublicKey::from_public_key_der(spki_der)
+            .map_err(|e| anyhow::anyhow!("failed to parse SPKI as secp256k1: {e}"))?;
```
Mirko-von-Leipzig:
```suggestion
            .context("failed to parse SPKI as secp256k1")?;
```

crates/validator/src/block_validation/mod.rs
```diff
@@ -16,7 +18,9 @@ pub enum BlockValidationError {
     #[error("block contains unvalidated transactions {0:?}")]
     UnvalidatedTransactions(Vec<TransactionId>),
     #[error("failed to build block")]
-    BlockBuildingFailed(#[source] ProposedBlockError),
+    BlockBuildingFailed(#[from] ProposedBlockError),
```
Mirko-von-Leipzig:
```suggestion
    BlockBuildingFailed(#[source] ProposedBlockError),
```

Mirko-von-Leipzig: We really should not use `#[from]`. It makes it too easy to cast things incorrectly.

crates/store/src/genesis/mod.rs
```diff
@@ -134,7 +131,10 @@ impl<S: BlockSigner> GenesisState<S> {
 
         let block_proof = BlockProof::new_dummy();
 
-        let signature = self.block_signer.sign(&header);
+        // Sign and assert verification for sanity (no mismatch between frontend and backend signing
+        // impls).
+        let signature = block_on(self.block_signer.sign(&header))?;
```
Mirko-von-Leipzig: This isn't really safe to do. The reality is that this should really be marked as an async function but that large portions of this are actually `sync` and not `async`.

I'm okay to accept this for now but then this function really requires a clean up and correct handling of async/sync boundaries.

crates/utils/src/signer.rs
Mirko-von-Leipzig, line 1: Can we not just use an enum and a match? We only have two items; I'd prefer less traits unless there is a good reason.

crates/utils/src/signer.rs
```diff
@@ -0,0 +1,35 @@
...
+// SECRET KEY BLOCK SIGNER
+// ================================================================================================
+
+#[async_trait::async_trait]
+impl BlockSigner for SecretKey {
+    type Error = Infallible;
+
+    async fn sign(&self, header: &BlockHeader) -> Result<Signature, Self::Error> {
+        Ok(self.sign(header.commitment()))
+    }
```
Mirko-von-Leipzig, line 31: Is this not a blocking operation? How long does signing take?

crates/validator/src/signers/kms.rs
```diff
@@ -0,0 +1,107 @@
...
+        Ok(Self { key_id, pub_key, client })
+    }
+}
+
+#[async_trait::async_trait]
+impl BlockSigner for KmsSigner {
+    type Error = KmsSignerError;
+
+    async fn sign(&self, header: &BlockHeader) -> Result<Signature, Self::Error> {
+        // KMS backend doesn't support Keccak-256 so we do it ourselves.
```
Mirko-von-Leipzig: Could you expand on what this means?

sergerad: Updated - anything still unclear?

```rust
        // AWS KMS does not support SHA3 (Keccak-256), so we need to produce the digest ourselves.
```

crates/validator/src/block_validation/mod.rs
sergerad: Yea sorry I have been stubborn about this. Any references you have for me to look at on that topic? Like linters/style guides. If we have consensus about not using #[from] at all I'll make sure not to in the future.

bin/node/src/commands/bundled.rs
sergerad: I'm thinking the arg group approach is more expressive and correct. Correct as in `validator.key` implies key content, not key id.

bin/node/src/commands/mod.rs
sergerad: without `KMS_KEY_ID` I think what the env var expects is ambiguous and would need an explanatory comment. Would prefer just to keep it clear in the name of the variable.

crates/utils/src/signer.rs
sergerad: Added a ValidatorSigner enum. I think the BlockSigner trait makes sense so as to avoid importing the ValidatorSigner into the store crate which houses the genesis block construction (signing) logic.

bin/node/src/commands/bundled.rs
sergerad: Did the arg group + separate enum approach

crates/utils/src/signer.rs
sergerad: Its async only because of the trait definition

crates/store/src/genesis/mod.rs
sergerad: I made the fn async. Don't think there was a reason not to

bin/node/src/commands/mod.rs
```diff
@@ -50,9 +54,53 @@ fn duration_to_human_readable_string(duration: Duration) -> String {
...
+        env = ENV_VALIDATOR_KEY,
+        value_name = "VALIDATOR_KEY",
+        default_value = INSECURE_VALIDATOR_KEY_HEX,
+    )]
+    validator_key: String,
+    /// Key ID for the KMS key used by validator to sign blocks.
+    ///
+    /// Cannot be used with `key.hex`.
+    #[arg(
+        long = "key.kms-id",
```
```suggestion
        long = "validator.key.kms-id",
```

bin/node/src/commands/mod.rs
```diff
@@ -50,9 +54,53 @@ fn duration_to_human_readable_string(duration: Duration) -> String {
...
 }
 
-/// Configuration for the Validator component.
+/// Configuration for the Validator key used to sign blocks.
+///
+/// Used by the Validator command and the genesis bootstrap command.
 #[derive(clap::Args)]
-pub struct ValidatorConfig {
+#[group(required = true, multiple = false)]
+pub struct ValidatorKey {
```
Mirko-von-Leipzig, line 63: Have you tested that this works as expected? The default with an optional and group required sounds like it might be wonky. I think its correctly defined, I'm just curious whether `clap` handles this properly.

bin/node/src/commands/validator.rs
```diff
@@ -49,26 +51,63 @@ pub enum ValidatorCommand {
...
-        #[arg(long = "key", env = ENV_VALIDATOR_KEY, value_name = "VALIDATOR_KEY", default_value = INSECURE_VALIDATOR_KEY_HEX)]
+        ///
+        /// Cannot be used with `key.kms-id`.
+        #[arg(
+            long = "key.hex",
+            env = ENV_VALIDATOR_KEY,
+            value_name = "VALIDATOR_KEY",
+            default_value = INSECURE_VALIDATOR_KEY_HEX,
+            group = "key"
+        )]
```
Mirko-von-Leipzig, line 62: Ah. And you have to repeat because its a different arg prefix?

We should take a look at [conf](https://github.com/cbeck88/conf-rs) which builds on top of clap-derive but seems less weird around use cases like this.

But for now this is fine 👍

crates/validator/src/block_validation/mod.rs
Mirko-von-Leipzig: I don't think we have anything explicitly about this; we had this general error discussion https://github.com/0xMiden/miden-base/discussions/966, maybe @PhilippGackstatter has one for `#[from]` vs `#[source]?

crates/utils/src/signer.rs
Mirko-von-Leipzig: Fair point yeah. Organizationally its a bit of a mess with the genesis block. Wonder if there's a cleaner way, though probably not important.

crates/validator/src/signers/kms.rs
Mirko-von-Leipzig: Yes - I don't know how to produce such a digest. Is there an algorithm you followed, or how did you know what to do?

Currently I'm just taking it as being okay; but you could really have written any algorithm here and I wouldn't know any different.

Mirko-von-Leipzig: aka how can I audit this code to ensure its correct.

crates/utils/src/signer.rs
Mirko-von-Leipzig: Yes but we shouldn't make blocking calls in an async function. Apparently we can expect this to take a few ms to maybe 100ms which is _high_ but probably acceptable for us.

Its the downside of having this trait -- it doesn't really match reality; one call is async IO, the other is local blocking sync function.

So really this should be doing something like

```rust
tokio::task::block_in_place(self.sign(header.commitment())).await
```

but that isn't great either.

bin/node/src/commands/store.rs
```diff
@@ -234,7 +222,30 @@ impl StoreCommand {
...
+            ValidatorSigner::Local(signer) => {
+                Self::bootstrap_with_signer(config, signer, accounts_directory, data_directory)
+                    .await
+            },
+        }
+    }
+
+    async fn bootstrap_with_signer(
+        config: GenesisConfig,
+        signer: impl BlockSigner,
```
Mirko-von-Leipzig, line 251: Maybe `&dyn` would make it easier?


```suggestion
        let signer = validator_key.into_signer().await?;
        Self::bootstrap_with_signer(config, &signer, accounts_directory, data_directory).await
    }

    async fn bootstrap_with_signer(
        config: GenesisConfig,
        signer: &dyn BlockSigner,
```

CHANGELOG.md
```diff
@@ -19,7 +19,7 @@
 - Refactored NTX Builder startup and introduced `NtxBuilderConfig` with configurable parameters ([#1610](https://github.com/0xMiden/miden-node/pull/1610)).
 - Refactored NTX Builder actor state into `AccountDeltaTracker` and `NotePool` for clarity, and added tracing instrumentation to event broadcasting ([#1611](https://github.com/0xMiden/miden-node/pull/1611)).
 - Add #[track_caller] to tracing/logging helpers ([#1651](https://github.com/0xMiden/miden-node/pull/1651)).
-- Improved tracing span fields ([#1650](https://github.com/0xMiden/miden-node/pull/1650))
+- Added KMS `BlockSigner` impl and moved `BlockSigner` trait from miden-base ([#1677](https://github.com/0xMiden/miden-node/pull/1677)).
```
```suggestion
- Added KMS signing support in validator ([#1677](https://github.com/0xMiden/miden-node/pull/1677)).
```

Mirko-von-Leipzig (approved):

crates/validator/src/block_validation/mod.rs
PhilippGackstatter: I think this blog post motivates adding context well: https://fast.github.io/blog/stop-forwarding-errors-start-designing-them/. In particular the "The Problem: Error Handling Without Purpose" section:
> When debugging, does “serialization error: expected , or }” tell you which request, which field, which code path led here?

`#[source]` is better than `#[from]` because it forces you to give some thought on how to best add context.

crates/validator/src/signers/kms.rs
sergerad: How about now?
```rust
        // The Validator produces Ethereum-style ECDSA (secp256k1) signatures over Keccak-256
        // digests. AWS KMS does not support SHA-3 hashing for ECDSA keys
        // (ECC_SECG_P256K1 being the corresponding AWS key-spec), so we pre-hash the
        // message and pass MessageType::Digest. KMS signs the provided 32-byte digest
        // verbatim.

```

bin/node/src/commands/mod.rs
sergerad:
```
❮ ./target/debug/miden-node validator start --data-directory /tmp/1 http://0.0.0.0:9191 --key.kms-id bded4b64-636a-4d27-8475-c2b8f8e76e75 --key.hex asdf

error: the argument '--key.kms-id <VALIDATOR_KMS_KEY_ID>' cannot be used with '--key.hex <VALIDATOR_KEY>'
```

```
❯ ./target/debug/miden-node bundled bootstrap --validator.key.hex asdb --validator.key.kms-id asdf

error: the argument '--validator.key.hex <VALIDATOR_KEY>' cannot be used with '--validator.key.kms-id <VALIDATOR_KMS_KEY_ID>'
```

bin/node/src/commands/store.rs
sergerad: Unfortunately the associated (Error) type is problematic for trait object
```
the value of the associated type `Error` in `miden_node_utils::signer::BlockSigner` must be specified
```

We can look at removing the trait instead after this PR maybe.

crates/validator/Cargo.toml
```diff
@@ -18,8 +18,12 @@ workspace = true
 
 [dependencies]
 anyhow                 = { workspace = true }
+async-trait            = { workspace = true }
+aws-config             = { version = "1.8.14" }
+aws-sdk-kms            = { version = "1.100" }
 diesel                 = { workspace = true }
 diesel_migrations      = { workspace = true }
+k256                   = "0.13.4"
```
bobbinth, line 25: We should probably not rely on `k256` directly but get w/e is needed via `miden-crypto`.

sergerad: Does miden-crypto support DER ser/de?

```rust
        let sig = k256::ecdsa::Signature::from_der(sig_der.as_ref())?;
```

sergerad: Or is that something we would want to add to miden-crypto and import here?

bobbinth: If we don't export it from there - we should add it to `miden-crypto` and then import here (otherwise, we could diverge in dependencies).

sergerad: This will have to be a followup PR right? Given node `next` points to crypto 0.19

bobbinth: We could push the required changes into `next` and then cherry-pick them into `v0.19.x` brach and do a patch release (I'm assuming these would be non-breaking changes).

sergerad: https://github.com/0xMiden/crypto/pull/842

crates/validator/src/block_validation/mod.rs
sergerad: Thanks yea the article is insightful.

>#[source] is better than #[from] because it forces you to give some thought on how to best add context.

I agree with the notion that it _could_ prompt you to add more context. But I don't think it ensures that. And I don't think it is necessary to achieve the goal of ensuring context being added when it is needed or formatting errors in such a way that they dont just forward/bubble up info.

Have updated the PR to use `#[source]` in any case!

crates/utils/Cargo.toml
```diff
@@ -20,6 +20,7 @@ testing = ["miden-protocol/testing"]
 
 [dependencies]
 anyhow                = { workspace = true }
+async-trait           = { workspace = true }
```
bobbinth: Question: do we actually need the `async_trait`? I'm assuming regular async traits don't work for some reason?

crates/validator/Cargo.toml
bobbinth: Do we still need this dependency? If so, let's get rid of it in a follow-up PR.

crates/validator/src/signers/kms.rs
```diff
@@ -0,0 +1,126 @@
+use anyhow::Context;
+use aws_sdk_kms::error::SdkError;
+use aws_sdk_kms::operation::sign::SignError;
+use aws_sdk_kms::types::SigningAlgorithmSpec;
+use k256::PublicKey as K256PublicKey;
+use k256::elliptic_curve::sec1::ToEncodedPoint;
+use k256::pkcs8::DecodePublicKey as _;
```
bobbinth, line 7: Related to the previous comment: `k256` dependency should be passed through `miden-crypto`.

crates/validator/src/signers/kms.rs
```diff
@@ -0,0 +1,126 @@
...
+        let pub_key = PublicKey::read_from_bytes(sec1_compressed)?;
+        Ok(Self { key_id, pub_key, client })
+    }
+}
+
+#[async_trait::async_trait]
+impl BlockSigner for KmsSigner {
+    type Error = KmsSignerError;
+
+    async fn sign(&self, header: &BlockHeader) -> Result<Signature, Self::Error> {
```
bobbinth, line 92: Should we double-check (somewhere at the end of this function) that the received signature verifies correctly?

bobbinth (approved):
Looks good! Thank you! I left a few comments inline. Getting rid of the `k256` direct dependency could probably be done in a follow-up PR.

crates/utils/Cargo.toml
Mirko-von-Leipzig: Removed

PR #1678, opened 2026-02-16
ci: add workflow cleanup

Mirko-von-Leipzig:
Github has this unfortunate UI design where deleted workflows continue to exist until _all_ their runs are deleted from history. Common sense would indicate that they have some retention policy and that they'll eventually fade away, but it seems like its much longer than the 90 days I've seen mentioned online.

This is annoying if you want to find or manually run a workflow. Refactoring or renaming is especially punishing because you end up with multiple of very similar entries.

This PR adds a **manual workflow** which deletes all runs of deleted workflows. Deletion is detected as those which don't exist on either `main` or `next`. This could technically be automated but this is a somewhat rare occurrence and its nice to know something won't just yank CI runs out from under you.

There is an argument to also delete existing workflow runs more deliberately i.e. only keep one weeks worth, but I don't see why we should complicate matters just to work around github being annoying. 

Here is a [discussion](https://github.com/orgs/community/discussions/26256) on this topic if you're interested in further frustration.

### DIFF
--- a/.github/workflows/cleanup-workflows.yml
+++ b/.github/workflows/cleanup-workflows.yml
@@ -1,0 +1,142 @@
+# Manual workflow to cleanup deleted workflows runs.
+#
+# Github keeps workflows runs around even if the workflow is deleted.
+# This has the side effect that these still display in the UI which gets cluttered.
+# Once the runs of a workflow are deleted, they also get removed from the UI.
+name: Cleanup Workflow
+
+on:
+  workflow_dispatch:
+    inputs:
+      mode:
+        description: "Choose 'dry run' to preview or 'execute' to delete runs"
+        required: true
+        default: "dry run"
+        type: choice
+        options:
+          - "dry run"
+          - "execute"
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-workflows.yml
```diff
@@ -0,0 +1,142 @@
...
+          echo ""
+
+          while IFS= read -r workflow; do
+            [ -z "$workflow" ] && continue
+
+            WF_COUNT=0
+
+            while true; do
+              RUN_IDS=$(gh run list \
+                --workflow "$workflow" \
```
drahnr, line 110: Q: how is the offset supplied to `gh run list` in case we hit more than 100 but aren't in `$MODE=="execute"`

Mirko-von-Leipzig: Ah. Good question. I didn't even think of that :< Stupid `gh` doesn't actually let you paginate.

PR #1679, opened 2026-02-16
ci: workflow cleanup follow-up

Mirko-von-Leipzig:
Follow-up fixes to #1678.

1. Step outputs don't natively support multiline strings.
2. `GITHUB_TOKEN` is required for `gh` cli.
3. Nicer progress printing ala chatgpt.
4. `gh run delete` has no `--yes` flag.
5. Fix pagination as per issue raised by @drahnr in https://github.com/0xMiden/miden-node/pull/1678#discussion_r2811533967

Pagination fix proof in example `dry-run` [here](https://github.com/0xMiden/miden-node/actions/runs/22066117128/job/63758265844), where the `dry-run` exits after paginating to 112 entries as per the manual total below:

```sh
gh run list --limit 1000 \
  --workflow docker-build-check.yml \
  --json databaseId --jq '.[].databaseId' \
| wc -l                                                                               
     112
```

### DIFF
--- a/crates/store/Cargo.toml
+++ b/crates/store/Cargo.toml
@@ -15,21 +15,24 @@ version.workspace      = true
 workspace = true
 
 [dependencies]
-anyhow                 = { workspace = true }
-deadpool               = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
-deadpool-diesel        = { features = ["sqlite"], version = "0.6" }
-deadpool-sync          = { default-features = false, features = ["tracing"], version = "0.1" }
-diesel                 = { features = ["numeric", "sqlite"], version = "2.3" }
-diesel_migrations      = { features = ["sqlite"], version = "2.3" }
-fs-err                 = { workspace = true }
-hex                    = { version = "0.4" }
-indexmap               = { workspace = true }
-libsqlite3-sys         = { workspace = true }
-miden-crypto           = { features = ["concurrent", "hashmaps"], workspace = true }
-miden-node-proto       = { workspace = true }
-miden-node-proto-build = { features = ["internal"], workspace = true }
-miden-node-utils       = { workspace = true }
-miden-standards        = { workspace = true }
+anyhow                     = { workspace = true }
+deadpool                   = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
+deadpool-diesel            = { features = ["sqlite"], version = "0.6" }
+deadpool-sync              = { default-features = false, features = ["tracing"], version = "0.1" }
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-workflows.yml
```diff
@@ -36,104 +36,228 @@ jobs:
...
-              RUN_IDS=$(gh run list \
-                --workflow "$workflow" \
-                --limit 100 \
-                --json databaseId \
-                --jq '.[].databaseId')
-
-              if [ -z "$RUN_IDS" ]; then
-                break
-              fi
+              # We use github's graphql API here which allows us to paginate over workflow runs.
```
drahnr: 👍

.github/workflows/cleanup-workflows.yml
```diff
@@ -36,104 +36,228 @@ jobs:
...
+          # Output to GitHub Actions
+          {
+            echo "workflows<<EOF"
+            echo "$DELETED"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+      # Performs the actual run deletion.
+      #
+      # This contains a lot of code, but the vast majority is just pretty-printing.
```
drahnr, line 102: Not sure if the trade-off is there, can we shift some (all?) of the formatting to shell functions we can then source?

Mirko-von-Leipzig: I'm also not sure tbh. I was just trying to see how good AI is at it 😁 On the other hand its nice when debugging to at least make sense of it.

When you mean shell functions, do you mean separate script files? Or do you mean some built-in functions from somewhere?

Mirko-von-Leipzig: Or maybe you meant just defining them as functions near the top so we can simplify the loop code?

.github/workflows/cleanup-workflows.yml
```diff
@@ -36,106 +36,248 @@ jobs:
...
-              WF_COUNT=$((WF_COUNT + BATCH_COUNT))
+              run_ids=$(echo "$response" | jq -r '.data.node.runs.nodes[].databaseId')
+              has_next=$(echo "$response" | jq -r '.data.node.runs.pageInfo.hasNextPage')
+              cursor=$(echo "$response" | jq -r '.data.node.runs.pageInfo.endCursor')
+
+              [ -z "$run_ids" ] && break
+
+              deleted=$(echo "$run_ids" | wc -l | tr -d ' ')
+              count=$((count + deleted))
+              total=$((total + deleted))
```
drahnr, line 245: always scared of shell math

Mirko-von-Leipzig: Yeah... me three.

PR #1684, opened 2026-02-17
feat/ci: ensure static linkage of C deps

drahnr:
Adds a CI step to check compiled binaries to be _statically_ linked against `libsqlite` and `librocksdb`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,23 +60,41 @@ jobs:
         run: cargo build --workspace --all-targets --locked
       - name: Check static linkage
         run: |
+          # Ensure database libraries are statically linked to avoid system library dependencies
+          #
+          # It explodes our possible dependency matrix when debugging, particularly
+          # in the case of sqlite and rocksdb as embedded databases, we want them
+          # shipped in identical versions we test with. Those are notoriously difficult
+          # to compile time configure and OSes make very opinionated choices.
           metadata=$(cargo metadata --no-deps --format-version 1)
           mapfile -t bin_targets < <(
             echo "${metadata}" | jq -r '.packages[].targets[] | select(.kind[] == "bin") | .name' | sort -u
           )
-          [[ ${#bin_targets[@]} -ne 0 ]] || { echo "No bin targets found."; exit 1; }
+          if [[ ${#bin_targets[@]} -eq 0 ]]; then
+            echo "error: No binary targets found in cargo manifest."
+            exit 1
+          fi
           for bin_target in "${bin_targets[@]}"; do
+            # Ensure the binary was built by the previous step.
             binary_path="target/debug/${bin_target}"
-            [[ -x "${binary_path}" ]] || { echo "Missing binary: ${binary_path}"; exit 1; }
+            if ! [[ -x "${binary_path}" ]]; then
+              echo "error: Missing binary or missing executable bit: ${binary_path}";
+              exit 2;
+            fi
+            # ldd exits non-zero for static binaries, so we inspect its output instead.
+            # if ldd fails we use an empty string instead
             ldd_output="$(ldd "${binary_path}" 2>&1 || true)"
-            [[ "${ldd_output}" != *"not a dynamic executable"* ]] || continue
-            ! echo "${ldd_output}" | grep -E -q 'librocksdb|libsqlite' || {
-              echo "Dynamic linkage detected for ${bin_target}."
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -58,6 +58,25 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/next' }}
       - name: cargo build
         run: cargo build --workspace --all-targets --locked
+      - name: Check static linkage
+        run: |
+          metadata=$(cargo metadata --no-deps --format-version 1)
+          mapfile -t bin_targets < <(
+            echo "${metadata}" | jq -r '.packages[].targets[] | select(.kind[] == "bin") | .name' | sort -u
+          )
+          [[ ${#bin_targets[@]} -ne 0 ]] || { echo "No bin targets found."; exit 1; }
```
Mirko-von-Leipzig: Does `[ -z $bin_targets ]` not work?

.github/workflows/ci.yml
```diff
@@ -58,6 +58,25 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/next' }}
       - name: cargo build
         run: cargo build --workspace --all-targets --locked
+      - name: Check static linkage
```
Mirko-von-Leipzig, line 61:
```suggestion
      # Ensure our database libraries are statically linked.
      #
      # We've had issues arising from accidentally using system libraries.
      #
      # This is done by inspecting the linker output of each binary.
      - name: Check static linkage
```

.github/workflows/ci.yml
```diff
@@ -58,6 +58,25 @@ jobs:
...
+            binary_path="target/debug/${bin_target}"
+            [[ -x "${binary_path}" ]] || { echo "Missing binary: ${binary_path}"; exit 1; }
+            ldd_output="$(ldd "${binary_path}" 2>&1 || true)"
+            [[ "${ldd_output}" != *"not a dynamic executable"* ]] || continue
+            ! echo "${ldd_output}" | grep -E -q 'librocksdb|libsqlite' || {
+              echo "Dynamic linkage detected for ${bin_target}."
+              echo "${ldd_output}"
+              exit 1
+            }
+          done
```
Mirko-von-Leipzig, line 96: I think some comments for the next poor non-bash soul would be kind.

```suggestion
          # Pull out the binary names from `cargo metadata`.
          metadata=$(cargo metadata --no-deps --format-version 1)
          bin_targets=$(
            echo "${metadata}" | jq -r '.packages[].targets[] | select(.kind[] == "bin") | .name' | sort -u'
          )
          
          # Sanity check, abort if no binaries found.
          [ -z ${bin_targets} ] || {
            echo "::error::No binaries targets found.";
            exit 1;
          }
          
          # Map to an array so we can iterate over entries.
          mapfile -t bin_targets < <$bin_targets
          
          # Inspect each binary's linker output for `rocksdb` and `sqlite`.
          # If they're present then we know they're dynamically linked and
          # are _not_ statically linked (bad).
          for bin_target in "${bin_targets[@]}"; do
            # Sanity check that the binary was built.    
            binary_path="target/debug/${bin_target}"
            [[ -x "${binary_path}" ]] || { 
              echo "::error::Missing binary: ${binary_path}"; 
              exit 1; 
            }
            
            # I'm actually unsure why this is short-circuiting `|| true`
            ldd_output="$(ldd "${binary_path}" 2>&1 || true)"
            
            # Should this not be an unexpected abort? Or are there binaries
            # that we have like this?
            [[ "${ldd_output}" != *"not a dynamic executable"* ]] || continue
            
            # I also don't know this `! echo` syntax?
            ! echo "${ldd_output}" | grep -E -q 'librocksdb|libsqlite' || {
              echo "Dynamic linkage detected for ${bin_target}."
              echo "${ldd_output}"
              exit 1
            }
          done
```

.github/workflows/ci.yml
drahnr: `$bin_targets` is an array, in `zsh` it would fail, in bash _I think_ it evals the first argument of the array

.github/workflows/ci.yml
drahnr: `!` is inverting the bool exit coe of `echo "foo" | grep ..`

drahnr: Simplified to prefer if-cases over `[[ .. ]] || ..` pattern and avoiding brace-less inversions

PR #1685, opened 2026-02-18
chore: Refactor common db capabilities into separate crate

### LINKED ISSUE #1653: Add DB crate that store and validator use
As per [comment](https://github.com/0xMiden/miden-node/pull/1614#discussion_r2776488252):
>I don't think validator should depend on the store. If there is some common functionality needed by both, I think we should move this to the utils crate (or maybe a new crate).

Must be done after #1614 is merged.

sergerad:
## Context

We have started using the sqlite capabilities found in the store crate in other crates (validator and ntx-builder).

The validator crate currently depends on db-related store code. This is a dependancy that we do not want to maintain.

The ntx-builder crate currently duplicates some db-related code that we can consolidate.

Note that the `apply_migration()` fns cannot be consolidated because of the `const` nature of the relevant API.

Closes #1653.

## Changes
- Move common db functionality from store crate into new db crate.
- Update validator crate to no longer depend on store crate.
- Remove duplicate db-related code from validator and ntx-builder.
- Remove unnecessary error enums and variants from the store.

### DIFF
--- a/bin/stress-test/src/store/mod.rs
+++ b/bin/stress-test/src/store/mod.rs
@@ -439,9 +439,7 @@ pub async fn bench_sync_chain_mmr(
 
     let request = |_| {
         let mut client = store_client.clone();
-        tokio::spawn(async move {
-            sync_chain_mmr_paginated(&mut client, chain_tip, block_range_size).await
-        })
+        tokio::spawn(async move { sync_chain_mmr(&mut client, chain_tip, block_range_size).await })
     };
 
     let results = stream::iter(0..iterations)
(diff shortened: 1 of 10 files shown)


crates/db/src/conv.rs
```diff
@@ -0,0 +1,183 @@
...
+//! Notice: All of them are infallible. The invariant is a sane content of the database
+//! and humans ensure the sanity of casts.
+//!
+//! Notice: Keep in mind if you _need_ to expand the datatype, only if you require sorting this is
+//! mandatory!
+//!
+//! Notice: Ensure you understand what casting does at the bit-level before changing any.
+//!
+//! Notice: Changing any of these are _backwards-incompatible_ changes that are not caught/covered
+//! by migrations!
```
drahnr, line 15: I really struggled formulating these..

crates/ntx-builder/src/db/manager.rs
```diff
@@ -1,86 +0,0 @@
...
-    // Enable the WAL mode. This allows concurrent reads while a write is in progress.
-    diesel::sql_query("PRAGMA journal_mode=WAL")
-        .execute(conn)
-        .map_err(ConnectionManagerError::ConnectionParamSetup)?;
-
-    // Enable foreign key checks.
-    diesel::sql_query("PRAGMA foreign_keys=ON")
-        .execute(conn)
-        .map_err(ConnectionManagerError::ConnectionParamSetup)?;
-    Ok(())
```
drahnr, line 85: These pragmas got lost while we added the timeout for all of them. I don't think that's right.

crates/db/src/errors.rs
drahnr: is there a particular reason to _not_ have `+ 'static` here? If not, please add it

crates/ntx-builder/src/db/manager.rs
sergerad: Sorry I'm not following - those pragmas are all still in `configure_connection_on_creation()` crates/db/src/manager.rs? Or is that not what you mean?

drahnr: Oh, I see, this is just `ntx`, and it's intentional you drop these?

sergerad: Yea ntx builder should use the same consolidated fns as store and validator

Issue #1680, opened 2026-02-16
Update dependencies


Mirko-von-Leipzig:
Update dependencies etc.

bobbinth:
I think this would go beyond just updating the lockfile (i.e., we'd want to bring up dependencies to their latest versions, wherever possible).

PR #1686, opened 2026-02-18
chore: update dependencies and fix store manifest error

namedfarouk:
This commit updates several external dependencies to their latest versions as requested in #1680. It also fixes a manifest error in crates/store/Cargo.toml that was causing cargo to fail.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,22 +92,22 @@ pretty_assertions = { version = "1.4" }
 # to avoid future breakage.
 prost              = { default-features = false, version = "=0.14.3" }
 protox             = { version = "=0.9.1" }
-rand               = { version = "0.9" }
-rand_chacha        = { version = "0.9" }
+rand               = { version = "0.10" }
+rand_chacha        = { version = "0.10" }
 rstest             = { version = "0.26" }
 serde              = { features = ["derive"], version = "1" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
Mirko-von-Leipzig, line 1: Some of these have actually gone backwards?

crates/store/Cargo.toml
Mirko-von-Leipzig, line 1: What command was causing the build to fail?

Cargo.toml
namedfarouk: Actually, I've checked and these are the latest versions. For example, `toml` is now at `1.0.3` (which is newer than `0.9`), `tower-http` is at `0.6.8` (newer than `0.5`), and `rand` is at `0.10.0` (newer than `0.9`). The version numbers in Cargo follow semver, so `0.10` is indeed a step forward from `0.9`.

crates/store/Cargo.toml
namedfarouk: The `cargo build` command was failing because `miden-node-rocksdb-cxx-linkage-fix` was being included as a regular build-dependency instead of an optional one, which caused issues when the `rocksdb` feature wasn't enabled. Making it optional in both `[dependencies]` and `[build-dependencies]` resolved the manifest error.

Issue #1080, opened 2025-07-11
Revisit limits for list parameters in proto messages


SantiagoPittella:
> Also, maybe not for this PR, but we should document how many tags can be passed in and adjust if needed. For example, I think currently the limit may be something like 1000, but we probably need to restrict it quite a bit more. This applies to most endpoints which take lists as parameters.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-node/pull/1045#discussion_r2199993860_
            
We should determine if we want to reduce the limit of list's sizes, and document it.

namedfarouk:
I’d like to work on this issue. I opened a PR with the scoped changes here: https://github.com/0xMiden/node/pull/1986\n\nCould a maintainer please assign this issue to me?

namedfarouk:
I checked this on current `next` and it looks like it has already been addressed.

The work appears to have landed in PR #1986 (`rpc/store: align list-limit metadata and document query limits`), which was merged on April 27, 2026.

Current evidence:
- `docs/internal/src/rpc.md` documents the query/list limits exposed by `GetLimits`
- `miden_node_utils::limiter` defines the current centralized list limits
- RPC/store tests cover the `GetLimits` metadata and enforcement paths
- the changelog entry for #1986 explicitly mentions aligning list-limit metadata and documenting query parameter limits

Unless there is still a desired follow-up to further reduce some of these limits, this issue may be safe to close as already resolved.

Mirko-von-Leipzig:
@SantiagoPittella could you confirm?

SantiagoPittella:
Yes, this can be marked as complete. In case we decide to revisit the limits we can create a new issue.

PR #1687, opened 2026-02-18
fix: add missing list validation for sync_transactions and remove inc…

johnh4098:
noticed while reading through the rpc code that sync_transactions was missing the account_ids list validation at the RPC layer - all other endpoints with repeated list params already had this check. the store layer was validating it, but requests should be rejected early at the RPC boundary.

also removed the RPC_LIMITS entries for SyncAccountVault and SyncAccountStorageMaps since they accept a single account_id, not a repeated list - having list limits for them was misleading for clients calling get_limits.

partially addresses #1080

cargo check passes.


drahnr:
Please adjust the CI check for the get limits endpoint

PR #1688, opened 2026-02-18
refactor(prover): combine proxy and worker

### LINKED ISSUE #935: Investigate about grouping metrics by status code
> One thought for querying metrics is that maybe we want to group all responses under a single metric with a "status code" label. Then queries become more versatile (`status_code >= 400`, etc.). But maybe this is overkill for now because the amount of different responses is narrow enough that it can be handled separately (and maybe this was discussed already).

_Originally posted by @igamigo in https://github.com/0xPolygonMiden/miden-base/issues/1017#issuecomment-2568247494_

### LINKED ISSUE #1171: Replace Pingora proxy with custom tonic-based service
We want to remove the dependency on `Pingora` and replace it with our own gRPC service built with `tonic`.  
This new service will maintain shared state across "workers" and handle request delegation.  
The service should be fully compatible with the existing `ProverApi` interface, so clients do not need to be aware of which worker is actually handling the request.  

## Tasks
- [ ] **Core service implementation**  
  - Create a tonic service that:  
    - Maintains shared state of worker availability.  
    - Queues requests and assigns them to free workers.  
    - Exposes the same API (`ProverApi`) as the workers.  

- [ ] **Status endpoint**  
  - Add an endpoint to expose the overall service status.  

- [ ] **Rate limiting**  
  - Add configurable rate limiting to protect the workers from overload.  

- [ ] **Health checks**  
  - Implement worker health checks to automatically detect and exclude unhealthy workers.  

- [ ] **Worker management**  
  - Add functionality to dynamically add or remove workers.  

- [ ] **Metrics, logging, and tracing**  
  - Integrate metrics collection, structured logs, and tracing for observability.  

- [ ] **Self-reflection**  
  - Add an endpoint for self-reflection / introspection (e.g., list workers, queue length, recent activity).

### LINKED ISSUE #1341: Consider removing proxy from remote prover binary
At the moment, the worker and proxy of the remote prover are coupled in the same binary. We need to discuss if we want to move it into a different binary or even another repository.

Mirko-von-Leipzig:
This PR effectively rewrites the `miden-remote-prover`, removing the distinction between `proxy` and `worker` mode by adding a configurable request queue to the "worker".

This removes the need for `pingora` as we now essentially replace it with a basic queuing mechanism. This was proposed in https://github.com/0xMiden/miden-node/issues/1171#issuecomment-3897328852.

Given the scale of the refactor, I don't think the diff view will be of much use. Instead a from-scratch review is probably necessary, though I think (hope) it should be quite straight-forward.

### Prover gRPC schema

I didn't touch the gRPC schema of the prover. The only real problem here is that we have both a proxy and worker status definition and its unclear now to which they refer.

This isn't such a big deal, though I will note that the gRPC reflection on the `miden-remote-prover` now erroneously thinks it supports the proxy status because its part of the same file.

One potential option is to only have a single status implementation i.e. the proxy one, and the "worker" just always sends an array of one element in its impl. This would also let us add this as a method instead of as a separate service. The original intention here was to host it on internal ports but I think at this point that intention no longer matters (imo).

Closes #1171 
Closes #1341
Closes #935


bin/remote-prover/src/server/prover.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl ProveRequest for LocalBlockProver {
+    type Input = BlockProofRequest;
+    type Output = BlockProof;
+
+    fn prove(&self, input: Self::Input) -> Result<Self::Output, tonic::Status> {
+        let BlockProofRequest { tx_batches, block_header, block_inputs } = input;
+        self.prove(tx_batches, &block_header, block_inputs)
+            .map_err(|e| tonic::Status::internal(e.as_report_context("failed to prove batch")))
```
SantiagoPittella:
```suggestion
            .map_err(|e| tonic::Status::internal(e.as_report_context("failed to prove block")))
```

bin/remote-prover/README.md
```diff
@@ -68,274 +71,73 @@ make install-remote-prover
...
-miden-remote-prover add-workers --control-port 8083 0.0.0.0:8085,200.58.70.4:50051
-# To remove 158.12.12.3:8080 and 122.122.6.6:50051 from the workers list:
-miden-remote-prover remove-workers --control-port 8083 158.12.12.3:8080,122.122.6.6:50051
-```
+The server implements the following health and status related gRPC services:
 
-These commands can receive the list of workers to update as a comma-separated list of addresses through the `MRP_PROXY_WORKERS_LIST` environment variable, or as command-line arguments:
+- [gRPC Health Check](https://grpc.io/docs/guides/health-checking/)
+- [gRPC Reflection](https://grpc.io/docs/guides/reflection/) 
+- [WorkerStatusApi](https://github.com/miden-io/miden-prover/blob/main/proto/proto/remote_prover.proto)
```
SantiagoPittella:
```suggestion
- [WorkerStatusApi](https://github.com/0xmiden/miden-prover/blob/main/proto/proto/remote_prover.proto)
```

packaging/prover/miden-prover.service
SantiagoPittella, line 1: We need to update the `ExecStart` variable too. It is using the old `start-worker` command.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 ### Changes
 
+- [BREAKING] Reworked `miden-remote-prover`, removing the `worker`/`proxy` distinction and simplifying to a `worker` with a request queue ([#1688](https://github.com/0xMiden/miden-node/pull/1688)).
+
+### Changes
```
SantiagoPittella: The `changes` subtitle is now duplicated.

bin/remote-prover/src/server/prover.rs
```diff
@@ -0,0 +1,122 @@
...
+/// This trait abstracts over proof request handling by providing a common interface for our
+/// different provers.
+///
+/// It standardizes the proving process by providing default implementations for the decoding of
+/// requests, and encoding of response. Notably it also standardizes the instrumentation, though
+/// implementations should still add attributes that can only be known post-decoding of the request.
+///
+/// Implementations of this trait only need to provide the input and outputs types, as well as the
+/// proof implementation.
+trait ProveRequest {
```
SantiagoPittella, line 54: I love this trait, really simplified things 🚀

Mirko-von-Leipzig: I've been experimenting with a generalized version of this as well, but I'm not quite happy with it yet.

The idea is to have a grpc method trait which sort of looks like this but with more generics.

And then the grpc crate generates a server trait for each api, where each method is specified as such a grpc method. Then you can implement method by method like this.

bin/remote-prover/src/server/service.rs
```diff
@@ -0,0 +1,88 @@
...
+        let prover = self.acquire_prover().await;
+
+        // Blocking in place is fairly safe since we guarantee that only a single request is
+        // processed at a time.
+        //
+        // This has the downside that requests being proven cannot be cancelled since we are now
+        // outside the async runtime. This could occur if the server timeout is exceeded, or
+        // the client cancels the request. A different approach is technically possible, but
+        // would require more complex logic to handle cancellation in tandem with sync.
+        tokio::task::block_in_place(|| prover.prove(request)).map(tonic::Response::new)
```
Mirko-von-Leipzig, line 86: So this is something I'd like to address somehow. But we can try that in a follow-up PR because I think it would complicate this review.

As a summary:

- We have a `server.timeout` which _should_ cancel a request if its been enqueued for longer than the timeout.
- This only works for tasks in the async runtime.
- `block_in_place` moves the _current_ thread off the async runtime, making it uncancelable.

So right now what happens is that if there is a very long prove request, then it completes but likely all other requests in-queue get timed out. Which is of course unfair and not really what we want.

-------

I _think_ what we want is to bound the time a single request spends actually proving. We already have "back-pressure" in the form of the bounded queue capacity so there is no real benefit to punishing those already in the queue.

Cancelling a sync task isn't trivial, but I think we can achieve this using a dedicated prover thread + a manager task which aborts the thread if it takes too long.

bobbinth: Question: why do we need the task here to be sync? Is this because the `prove()` function is sync? If so, this will change shortly (in v0.21 VM, `prove()` is an async function).

packaging/prover/postinst
```diff
@@ -2,27 +2,9 @@
...
-
-# Configuration folder
-if [ -d "/etc/opt/miden-prover" ]
-then
-    echo "Directory /etc/opt/miden-prover exists."
-else
-    mkdir -p /etc/opt/miden-prover
-fi
-sudo chown -R miden-prover /etc/opt/miden-prover
-
```
sergerad, line 25: How come we no longer need this?

Mirko-von-Leipzig: We never needed it - the prover has no state

bin/remote-prover/src/server/service.rs
Mirko-von-Leipzig: Oh.. Yes because `prove()` is (expensive) sync and would otherwise block the runtime.

In which case I'm happy to leave this as is until we get async prove.

PR #1689, opened 2026-02-18
chore(ntx-builder): DataStore & DB writers refactor 

### LINKED ISSUE #1628: Persist ntx state

### LINKED ISSUE #1657: NTX Builder state (phase 3): Replace `DataStore` impl with DB queries
Refactor NtxDataStore to read committed account and note data from the local SQLite database instead of making remote gRPC calls to the store, reducing network traffic and latency during transaction execution.

SantiagoPittella:
closes #1657 
closes #1628 

This PR includes two major changes:

1. Persistent note script cache.
Adds a `note_script` table to the database so that fetched note scripts survive restarts.

2. Centralized DB writes via actor notification channel.
In the phase 2 PR, when we switched from in-memory to the SQLite db,  actors were still writing to the DB to mark notes as failing. In this PR I added a channel between actor and coordinator so that the actor can emit an event to the coordinator when it is failing to consume a note.

### DIFF
--- a/crates/ntx-builder/src/actor/execute.rs
+++ b/crates/ntx-builder/src/actor/execute.rs
@@ -78,6 +78,12 @@ pub enum NtxError {
 
 type NtxResult<T> = Result<T, NtxError>;
 
+/// The result of a successful transaction execution.
+///
+/// Contains the transaction ID, any notes that failed during filtering, and note scripts fetched
+/// from the remote store that should be persisted to the local DB cache.
+pub type NtxExecutionResult = (TransactionId, Vec<FailedNote>, Vec<(Word, NoteScript)>);
+
 // NETWORK TRANSACTION CONTEXT
 // ================================================================================================
 
@@ -138,8 +144,9 @@ impl NtxContext {
     ///
     /// # Returns
     ///
-    /// On success, returns the [`TransactionId`] of the executed transaction and a list of
-    /// [`FailedNote`]s representing notes that were filtered out before execution.
+    /// On success, returns an [`NtxExecutionResult`] containing the transaction ID, any notes
+    /// that failed during filtering, and note scripts fetched from the remote store that should
+    /// be persisted to the local DB cache.
     ///
     /// # Errors
     ///
@@ -152,7 +159,7 @@ impl NtxContext {
     pub fn execute_transaction(
         self,
         tx: TransactionCandidate,
-    ) -> impl FutureMaybeSend<NtxResult<(TransactionId, Vec<FailedNote>)>> {
+    ) -> impl FutureMaybeSend<NtxResult<NtxExecutionResult>> {
         let TransactionCandidate {
             account,
             notes,
@@ -185,6 +192,9 @@ impl NtxContext {
                 // Execute transaction.
                 let executed_tx = Box::pin(self.execute(&data_store, successful_notes)).await?;
 
+                // Collect scripts fetched from the remote store during execution.
+                let scripts_to_cache = data_store.take_fetched_scripts().await;
+
                 // Prove transaction.
                 let tx_inputs: TransactionInputs = executed_tx.into();
                 let proven_tx = Box::pin(self.prove(&tx_inputs)).await?;
@@ -195,7 +205,7 @@ impl NtxContext {
                 // Submit transaction to block producer.
                 self.submit(&proven_tx).await?;
 
-                Ok((proven_tx.id(), failed_notes))
+                Ok((proven_tx.id(), failed_notes, scripts_to_cache))
             })
             .in_current_span()
             .await
@@ -341,8 +351,11 @@ struct NtxDataStore {
     store: StoreClient,
     /// LRU cache for storing retrieved note scripts to avoid repeated store calls.
     script_cache: LruCache<Word, NoteScript>,
-    /// Local database for persistent note script caching.
+    /// Local database for persistent note script.
     db: Db,
+    /// Scripts fetched from the remote store during execution, to be persisted by the
+    /// coordinator.
+    fetched_scripts: Arc<Mutex<Vec<(Word, NoteScript)>>>,
     /// Mapping of storage map roots to storage slot names observed during various calls.
     ///
     /// The registered slot names are subsequently used to retrieve storage map witnesses from the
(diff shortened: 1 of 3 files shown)


crates/ntx-builder/src/builder.rs
```diff
@@ -98,9 +98,12 @@ pub struct NetworkTransactionBuilder {
     actor_context: AccountActorContext,
     /// Stream of mempool events from the block producer.
     mempool_events: MempoolEventStream,
+    /// Receiver for notifications from account actors (e.g., note failures).
+    notification_rx: mpsc::UnboundedReceiver<ActorNotification>,
```
Mirko-von-Leipzig: I think this should be bounded - actors can wait to submit.

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -273,9 +296,10 @@ impl AccountActor {
                             let chain_state = self.chain_state.read().await.clone();
 
                             // Drop notes that have failed too many times.
-                            if let Err(err) = self.db.drop_failing_notes(account_id, self.max_note_attempts).await {
-                                tracing::error!(err = %err, "failed to drop failing notes");
-                            }
+                            let _ = self.notification_tx.send(ActorNotification::DropFailingNotes {
+                                account_id,
+                                max_attempts: self.max_note_attempts,
+                            });
```
Mirko-von-Leipzig: I think we can ignore this. As in, keep the notes around - our query logic will avoid them.

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -507,28 +518,40 @@ impl DataStore for NtxDataStore {
...
+                    DataStoreError::other_with_source(
+                        "failed to retrieve note script from store",
+                        err,
+                    )
                 })?;
-            // Handle response.
+
             if let Some(script) = maybe_script {
+                // Best-effort persist — don't fail the transaction if DB write fails.
+                let _ = self.db.insert_note_script(script_root, &script).await;
```
Mirko-von-Leipzig: This could be another message to the co-ordinator instead?

This shouldn't happen that often..

crates/ntx-builder/src/builder.rs
SantiagoPittella: Should we limit it to 1 message at a time?

Mirko-von-Leipzig: The actor needs to wait until the notification has been handled in order to proceed. So having it be a single one at a time makes sense.

So we might also want a return `oneshot` channel for the actor to know its been completed.

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: Done!

crates/ntx-builder/src/actor/execute.rs
SantiagoPittella: I added the new message on [9198f08](https://github.com/0xMiden/miden-node/pull/1689/commits/9198f08fa7d0f2c3d5e0913486f368c423193e37)

crates/ntx-builder/src/builder.rs
SantiagoPittella: hmmm in that case we should keep one receiver per actor in the coordinator, since oneshot is single producer, right?

SantiagoPittella:
@Mirko-von-Leipzig should I wait for any other review to merge this one?

crates/ntx-builder/src/builder.rs
Mirko-von-Leipzig: There are several patterns we can use here; but the one I'm proposing looks like this:

```rust
struct Coordinator {
    rx_notes_failed: tokio::sync::mpsc::Receiver<
        // Notes that failed.
        HashSet<NoteId>,
        // Return channel to inform the note sender that processing is done.
        tokio::sync::oneshot::Sender<()>,
    >,
    ...
}

impl Coordinator {
    fn handle_note_failure(&mut self, notes: HashSet<NoteId>, notify: oneshot::Sender<()> {
        self.db.increment_note_failures(notes).await;
        let _ = notify.send().await;
    }
}

struct Actor {
    tx_notes_failed: tokio::sync::mpsc::Sender < ..as above..>,
}

impl Actor {
    // Channel closure handling elided as demo code.
    async fn process_note_failure(&self, notes: HashSet<NoteId>) {
        let (tx, rx) = tokio::sync::oneshot::channel();

        let _ = self.tx_notes_failed.send(notes, tx).await;
        // Wait for coordinator to inform us that processing completed.
        // If we don't wait, we risk a data race with the note failure
        // being updated in the database.
        rx.await;
    }
}
```

There are alternatives, but this has the advantage that the oneshot return channels usage is obviously tied to the note failure request. You _could_ have an additional channel per actor, or try to re-use the notification channel, but then you run into other problems e.g. that channel will have other signals on it already maybe.

PR #1691, opened 2026-02-19
chore: migrate to base 0904e2c61

mmagician:
preparing for the 0.14-alpha release


crates/store/src/db/tests.rs
```diff
@@ -350,12 +350,8 @@ fn sql_unconsumed_network_notes() {
...
             note_id: num_to_word(i.into()),
             note_commitment: num_to_word(i.into()),
-            metadata: NoteMetadata::new(
-                account_note.0,
-                NoteType::Public,
-                NoteTag::with_account_target(account_note.0),
-            )
-            .with_attachment(attachment.clone()),
+            metadata: NoteMetadata::new(account_note.0, NoteType::Public)
+                .with_attachment(attachment.clone()),
```
juan518munoz, line 354: Looking at other parts of the code update, shouldn't we `with_tag( NoteTag::with_account_target(account_note.0))`?

mmagician: not for network accounts as of https://github.com/0xMiden/miden-base/pull/2257

crates/store/src/db/tests.rs
```diff
@@ -1409,7 +1415,7 @@ fn genesis_with_account_assets() {
         .unwrap();
 
     let genesis_state =
-        GenesisState::new(vec![account], test_fee_params(), 1, 0, SecretKey::random());
+        GenesisState::new(vec![account], test_fee_params(), 1, 0, random_secret_key());
```
Mirko-von-Leipzig, line 1418: Do we have any idea why we now need a free standing function?

`SecretKey::random()` is so much easier to find..

crates/validator/src/server/mod.rs
```diff
@@ -42,13 +43,13 @@ pub struct Validator<S> {
     pub grpc_timeout: Duration,
 
     /// The signer used to sign blocks.
-    pub signer: S,
+    pub signer: SecretKey,
```
Mirko-von-Leipzig: @sergerad is this compatible? I don't think it is right..

mmagician: you're right - I think this PR should build on top of https://github.com/0xMiden/miden-node/pull/1677 to be meaningful

crates/store/src/db/tests.rs
mmagician: was changed [here](https://github.com/0xMiden/miden-base/pull/2447/changes#diff-4fec4124609517d10347d11b17f986ce8566e1ddb41db479a77ff0a4c9010e49R9), but I'm not sure as to why

crates/store/src/genesis/config/mod.rs
```diff
@@ -220,11 +219,16 @@ impl GenesisConfig {
...
 
             if total_issuance != 0 {
-                // slot 0
-                storage_delta.set_item(
-                    AccountStorage::faucet_sysdata_slot().clone(),
-                    [ZERO, ZERO, ZERO, Felt::new(total_issuance)].into(),
-                )?;
+                let metadata_slot_name = BasicFungibleFaucet::metadata_slot();
+                let current_metadata =
+                    faucet_account.storage().get_item(metadata_slot_name).unwrap();
```
mmagician: @cursor we should use `TokenMetadata` here:
we can try to build it from `Word`

and then construct `updated_metadata: TokenMetadata = TokenMetadata::with_supply` as well, using the values from the current metadata

crates/store/src/genesis/config/mod.rs
```diff
@@ -442,10 +446,10 @@ impl AccountSecrets {
     ///
     /// If no name is present, a new one is generated based on the current time
     /// and the index in
-    pub fn as_account_files<S>(
+    pub fn as_account_files(
         &self,
-        genesis_state: &GenesisState<S>,
-    ) -> impl Iterator<Item = Result<AccountFileWithName, GenesisConfigError>> + use<'_, S> {
+        genesis_state: &GenesisState<impl miden_node_utils::signer::BlockSigner>,
```
mmagician: @cursor import at the top of file, here we should only see `BlockSigner`

crates/store/src/genesis/config/tests.rs
```diff
@@ -45,7 +45,7 @@ fn parsing_yields_expected_default_values() -> TestResult {
 
     // check total issuance of the faucet
     assert_eq!(
-        native_faucet.storage().get_item(AccountStorage::faucet_sysdata_slot()).unwrap()[3],
+        native_faucet.storage().get_item(BasicFungibleFaucet::metadata_slot()).unwrap()[0],
```
mmagician: use `TokenMetadata` (try to build from Word) and its `token_supply` getter

crates/store/src/db/tests.rs
Mirko-von-Leipzig: I think it should still be available from `miden-crypto`, its just protocol that stopped using it

mmagician: it was never in `miden-crypto` - we had a trait `RandomBlockSigner` in protocol with a `random()` method: https://github.com/0xMiden/miden-base/pull/2447/changes#diff-f183681be8c6e7e0ae0df81a671d744f8dd2fc396f74e798ad4916e95b8bc3e2L9 

something similar could be added back (to Serge's PR which brings back `BlockSigner`?) but not sure if it makes that much sense, it was used for testing purposes only anyway

Mirko-von-Leipzig: Ah I see; then I understand yeah.

Issue #1694, opened 2026-02-20
Ntx actor refactoring


Mirko-von-Leipzig:
With #1628 largely complete, we can hopefully also simplify the actor side of the equation.

I touched on some ideas in https://github.com/0xMiden/miden-node/issues/1314#issuecomment-3598248000, but the foundations have since shifted a bit so lets consolidate our efforts in this issue. I'll jot my ideas down and maybe @SantiagoPittella can split them out into child issues as you see fit? 

Overarching goals are to:

1. Limit number of network transactions being built.
2. Reap inactive network accounts.
3. Eliminate race conditions between actors and co-ordinator, and state updates.
4. Simplify and re-organize the code so its easier to understand.

(1) has already been solved using the semaphor + permit system, and I think that's quite effective.

### Deactivating a network actor

I think there are two options.

1. Actor shuts itself down, and
2. co-ordinator kills the actor

(2) is already possible if we interpret closing the channel to the actor down. I think we just need to add logic to the actor that its an expected signal to receive, and not an error. However, this isn't important to implement yet because I don't think we have any need to do this from the coordinator side.

I'm mostly interested in implementing (1). The idea is simply that if an account's state is sterile for some amount of time, then that actor shuts down. Sterile meaning no viable transaction to create e.g. no possible notes to consume.

There is a race condition here where the coordinator sends a message, but the actor has shutdown. We can "detect" this by having the actor return its message channel on exit. The coordinator then receives this as part of the actor task ending, and can inspect it. If its empty, then the actor can stay dead, if its non-empty then maybe we should start the actor up again so it can handle these. If we don't do this then its possible for new network notes to be missed.

I _think_ this solves the problem quite neatly, though maybe I've missed something :)

### Simplifying coordinator -> actor messages

Since state is now all in the database, we no longer need to explicitly inform the actor about _what_ changed. i.e. new notes would be queried from the database, account updates would be present as new transaction ID in the database etc.

This is still a fuzzy idea, but maybe its enough to just have a [notification channel](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) now? Where the actor can see if any changes have occurred and it should re-sync / check the database. Or maybe a different sync primitive, but I think you get the drift.

The downside of this is that constant database checks might be overly expensive? But we can "cheaply" query the account by checking the latest commitment (or transaction ID), and only fetching the update if its changed. It might be that sending deltas and applying them is still better.. but 🤷 

[...]

PR #1699, opened 2026-02-23
refactor(ntx-builder): simplify coordinator-actor messaging with Notify

SantiagoPittella:
First task of #1694 

- Replace per-actor `mpsc<Arc<MempoolEvent>>` channels with `Arc<Notify>`. The DB is already the source of truth (since #1662), so actors only need a "re-check your state" signal and not the event payload. This removes all `SendError` handling, failed-actor cleanup, and the `send()` helper from the coordinator.
- In `TransactionInflight` mode, actors now query the DB (`is_transaction_resolved`) to check if their awaited transaction was committed/reverted, since `Notify` doesn't carry payload.
- Unify actor -> coordinator communication into a single `ActorRequest` enum over one `mpsc` channel. `NotesFailed` carries a oneshot ack to prevent a race condition where the actor could re-select notes before failure counts are persisted ([context](https://github.com/0xMiden/miden-node/pull/1689#discussion_r2840426323)). `CacheNoteScript` remains fire-and-forget.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4321,9 +4321,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.13"
+version = "0.11.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1906b49b0c3bc04b5fe5d86a77925ae6524a19b816ae38ce1e426255f1d8a31"
+checksum = "434b42fec591c96ef50e21e886936e66d3cc3f737104fdb9b737c40ffb94c098"
 dependencies = [
  "aws-lc-rs",
  "bytes",
--- a/crates/store/src/account_state_forest/mod.rs
+++ b/crates/store/src/account_state_forest/mod.rs
@@ -41,7 +41,7 @@ mod tests;
 // ================================================================================================
 
 #[derive(Debug, Error)]
-pub enum InnerForestError {
+pub enum AccountStateForestError {
     #[error(transparent)]
     Asset(#[from] AssetError),
     #[error(transparent)]
@@ -60,17 +60,17 @@ pub enum WitnessError {
     AssetError(#[from] AssetError),
 }
 
-// INNER FOREST
+// ACCOUNT STATE FOREST
 // ================================================================================================
 
 /// Container for forest-related state that needs to be updated atomically.
-pub(crate) struct InnerForest {
+pub(crate) struct AccountStateForest {
     /// `LargeSmtForest` for efficient account storage reconstruction.
     /// Populated during block import with storage and vault SMTs.
     forest: LargeSmtForest<ForestInMemoryBackend>,
 }
 
-impl InnerForest {
+impl AccountStateForest {
     pub(crate) fn new() -> Self {
         Self { forest: Self::create_forest() }
     }
@@ -311,7 +311,7 @@ impl InnerForest {
         &mut self,
         block_num: BlockNumber,
         account_updates: impl IntoIterator<Item = AccountDelta>,
-    ) -> Result<(), InnerForestError> {
+    ) -> Result<(), AccountStateForestError> {
         for delta in account_updates {
             self.update_account(block_num, &delta)?;
 
@@ -345,7 +345,7 @@ impl InnerForest {
         &mut self,
         block_num: BlockNumber,
         delta: &AccountDelta,
-    ) -> Result<(), InnerForestError> {
+    ) -> Result<(), AccountStateForestError> {
         let account_id = delta.id();
         let is_full_state = delta.is_full_state();
 
@@ -383,7 +383,7 @@ impl InnerForest {
         block_num: BlockNumber,
         account_id: AccountId,
         vault_delta: &AccountVaultDelta,
-    ) -> Result<(), InnerForestError> {
+    ) -> Result<(), AccountStateForestError> {
         let prev_root = self.get_latest_vault_root(account_id);
         let lineage = Self::vault_lineage_id(account_id);
         assert_eq!(prev_root, Self::empty_smt_root(), "account should not be in the forest");
@@ -523,7 +523,7 @@ impl InnerForest {
         block_num: BlockNumber,
         account_id: AccountId,
         vault_delta: &AccountVaultDelta,
-    ) -> Result<(), InnerForestError> {
+    ) -> Result<(), AccountStateForestError> {
         assert!(!vault_delta.is_empty(), "expected the delta not to be empty");
 
         // get the previous vault root; the root could be for an empty or non-empty SMT
--- a/crates/store/src/account_state_forest/tests.rs
+++ b/crates/store/src/account_state_forest/tests.rs
@@ -58,21 +58,21 @@ fn dummy_full_state_delta(account_id: AccountId, assets: &[Asset]) -> AccountDel
 fn empty_smt_root_is_recognized() {
     use miden_crypto::merkle::smt::Smt;
 
-    let empty_root = InnerForest::empty_smt_root();
+    let empty_root = AccountStateForest::empty_smt_root();
 
     assert_eq!(Smt::default().root(), empty_root);
 }
 
 #[test]
-fn inner_forest_basic_initialization() {
-    let forest = InnerForest::new();
+fn account_state_forest_basic_initialization() {
+    let forest = AccountStateForest::new();
     assert_eq!(forest.forest.lineage_count(), 0);
     assert_eq!(forest.forest.tree_count(), 0);
 }
 
 #[test]
 fn update_account_with_empty_deltas() {
-    let mut forest = InnerForest::new();
+    let mut forest = AccountStateForest::new();
     let account_id = dummy_account();
     let block_num = BlockNumber::GENESIS.child();
 
@@ -99,7 +99,7 @@ fn vault_partial_vs_full_state_produces_same_root() {
     let asset = dummy_fungible_asset(faucet_id, 100);
 
     // Partial delta (block application)
-    let mut forest_partial = InnerForest::new();
+    let mut forest_partial = AccountStateForest::new();
     let mut vault_delta = AccountVaultDelta::default();
     vault_delta.add_asset(asset).unwrap();
     let partial_delta =
@@ -107,7 +107,7 @@ fn vault_partial_vs_full_state_produces_same_root() {
     forest_partial.update_account(block_num, &partial_delta).unwrap();
 
     // Full-state delta (DB reconstruction)
-    let mut forest_full = InnerForest::new();
+    let mut forest_full = AccountStateForest::new();
     let full_delta = dummy_full_state_delta(account_id, &[asset]);
     forest_full.update_account(block_num, &full_delta).unwrap();
 
@@ -120,7 +120,7 @@ fn vault_partial_vs_full_state_produces_same_root() {
 
 #[test]
 fn vault_incremental_updates_with_add_and_remove() {
-    let mut forest = InnerForest::new();
+    let mut forest = AccountStateForest::new();
     let account_id = dummy_account();
     let faucet_id = dummy_faucet();
 
@@ -153,7 +153,7 @@ fn vault_incremental_updates_with_add_and_remove() {
     assert_ne!(root_after_150, root_after_120);
 
     // Verify by comparing to full-state delta
-    let mut fresh_forest = InnerForest::new();
+    let mut fresh_forest = AccountStateForest::new();
     let full_delta = dummy_full_state_delta(account_id, &[dummy_fungible_asset(faucet_id, 120)]);
     fresh_forest.update_account(block_3, &full_delta).unwrap();
     let root_full_state_120 = fresh_forest.get_vault_root(account_id, block_3).unwrap();
@@ -168,7 +168,7 @@ fn forest_versions_are_continuous_for_sequential_updates() {
     use assert_matches::assert_matches;
     use miden_protocol::account::delta::{StorageMapDelta, StorageSlotDelta};
 
-    let mut forest = InnerForest::new();
+    let mut forest = AccountStateForest::new();
     let account_id = dummy_account();
     let faucet_id = dummy_faucet();
     let slot_name = StorageSlotName::mock(9);
@@ -201,7 +201,7 @@ fn forest_versions_are_continuous_for_sequential_updates() {
 
 #[test]
 fn vault_state_is_not_available_for_block_gaps() {
-    let mut forest = InnerForest::new();
+    let mut forest = AccountStateForest::new();
     let account_id = dummy_account();
     let faucet_id = dummy_faucet();
 
@@ -229,7 +229,7 @@ fn witness_queries_work_with_sparse_lineage_updates() {
     use assert_matches::assert_matches;
     use miden_protocol::account::delta::{StorageMapDelta, StorageSlotDelta};
 
-    let mut forest = InnerForest::new();
+    let mut forest = AccountStateForest::new();
     let account_id = dummy_account();
     let faucet_id = dummy_faucet();
     let slot_name = StorageSlotName::mock(6);
@@ -282,14 +282,14 @@ fn witness_queries_work_with_sparse_lineage_updates() {
             .open(forest.tree_id_for_vault_root(account_id, block_3), asset_key.into()),
         Ok(_)
     );
-    assert_ne!(vault_root_at_3, InnerForest::empty_smt_root());
+    assert_ne!(vault_root_at_3, AccountStateForest::empty_smt_root());
 }
 
 #[test]
 fn vault_full_state_with_empty_vault_records_root() {
     use miden_protocol::account::{Account, AccountStorage};
 
-    let mut forest = InnerForest::new();
+    let mut forest = AccountStateForest::new();
     let account_id = dummy_account();
     let block_num = BlockNumber::GENESIS.child();
 
@@ -306,7 +306,7 @@ fn vault_full_state_with_empty_vault_records_root() {
     forest.update_account(block_num, &full_delta).unwrap();
 
     let recorded_root = forest.get_vault_root(account_id, block_num);
-    assert_eq!(recorded_root, Some(InnerForest::empty_smt_root()));
+    assert_eq!(recorded_root, Some(AccountStateForest::empty_smt_root()));
 
     let witnesses = forest
         .get_vault_asset_witnesses(account_id, block_num, std::collections::BTreeSet::new())
@@ -316,7 +316,7 @@ fn vault_full_state_with_empty_vault_records_root() {
 
 #[test]
 fn vault_shared_root_retained_when_one_entry_pruned() {
-    let mut forest = InnerForest::new();
+    let mut forest = AccountStateForest::new();
     let account1 = AccountId::try_from(ACCOUNT_ID_REGULAR_PUBLIC_ACCOUNT_IMMUTABLE_CODE).unwrap();
     let account2 = AccountId::try_from(ACCOUNT_ID_REGULAR_PUBLIC_ACCOUNT_IMMUTABLE_CODE_2).unwrap();
     let faucet_id = dummy_faucet();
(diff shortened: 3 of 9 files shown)


crates/ntx-builder/src/actor/mod.rs
```diff
@@ -261,28 +259,22 @@ impl AccountActor {
...
-                        if should_wake {
+                // Handle coordinator notifications. On notification, re-evaluate state from DB.
+                _ = self.notify.notified() => {
+                    match self.mode {
+                        ActorMode::TransactionInflight(awaited_id) => {
+                            // Check DB: is the inflight tx still pending?
+                            let resolved = self.db
+                                .is_transaction_resolved(account_id, awaited_id)
+                                .await
+                                .expect("should be able to check tx status");
```
sergerad: Do we really want to stop everything if a single actor fails on this basis?

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -395,25 +387,31 @@ impl AccountActor {
...
         }
     }
 
-    /// Sends a notification to the coordinator to mark notes as failed.
+    /// Sends a request to the coordinator to mark notes as failed and waits for the DB write to
+    /// complete. This prevents a race condition where the actor could re-select the same notes
+    /// before the failure counts are updated in the database.
     async fn mark_notes_failed(&self, nullifiers: &[Nullifier], block_num: BlockNumber) {
+        let (ack_tx, ack_rx) = tokio::sync::oneshot::channel();
         let _ = self
```
sergerad: `_` is type `Result<(), SendError<ActorRequest>>`. Are we sure we don't want to be handling that error here?

sergerad: Same with `let _ = ack_rx.await;` below

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: If an actor panics, the coordinator catches the error in `coordinator.next()` and logs it, but the system keeps running without the actor. Though know that you mentioned this, probably the best would be to return an specific error for this Db fail case.

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: Replaced them with `.is_err` checks

crates/ntx-builder/src/db/models/queries/accounts.rs
```diff
@@ -100,3 +101,30 @@ pub fn get_account(
...
+/// Returns `true` when no inflight account row exists with the given `transaction_id`, meaning
+/// the transaction was committed or reverted.
+///
+/// # Raw SQL
+///
+/// ```sql
+/// SELECT COUNT(*)
+/// FROM accounts
+/// WHERE account_id = ?1 AND transaction_id = ?2
+/// ```
```
Mirko-von-Leipzig, line 112: I think this might be better? So long as we have a index on `transaction_id`.

The argument being, there will be _few_ inflight transactions but _many_ accounts.

```suggestion
/// ```sql
/// SELECT EXISTS (
///   SELECT 1 FROM accounts WHERE transaction_id = ?1
/// )
/// ```
```

crates/ntx-builder/src/db/models/queries/accounts.rs
```diff
@@ -100,3 +101,30 @@ pub fn get_account(
...
+pub fn is_transaction_resolved(
```
Mirko-von-Leipzig: I think rather call the function what it _does_. The actual meaning or interpretation should be done by the caller.

i.e. `transaction_exists`

crates/ntx-builder/src/builder.rs
```diff
@@ -98,8 +98,8 @@ pub struct NetworkTransactionBuilder {
     actor_context: AccountActorContext,
     /// Stream of mempool events from the block producer.
     mempool_events: MempoolEventStream,
-    /// Receiver for notifications from account actors (e.g., note failures).
-    notification_rx: mpsc::Receiver<ActorNotification>,
+    /// Receiver for requests from account actors (note failures, script caching).
```
Mirko-von-Leipzig:
```suggestion
    /// Database update requests from account actors.
    ///
    /// We keep database writes centralized so this is how actors communicate
    /// items to write.
```

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -395,25 +411,45 @@ impl AccountActor {
...
-            .await;
+            .await
+            .is_err()
+        {
+            tracing::warn!("failed to send notes failed request, coordinator is shutting down");
+            return;
+        }
+        // Wait for the coordinator to confirm the DB write.
+        if ack_rx.await.is_err() {
+            tracing::warn!("failed to receive notes failed ack from coordinator");
```
Mirko-von-Leipzig: Personally I don't think we want these as logs. If the coordinator shuts down then we _already_ know this because it should log that itself.

And the actor should handle and log this in a single place ideally -- as an expected thing that can happen. Right now we'll get `N=len(actors)` warnings popping off everywhere which is just noise.

SantiagoPittella: I removed them.

crates/ntx-builder/src/db/models/queries/accounts.rs
SantiagoPittella: Okok, makes sense since we do have the index

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -29,16 +28,31 @@ use crate::builder::ChainState;
...
-// ACTOR NOTIFICATION
+/// Converts a database result into an `ActorShutdownReason` error, logging the error on failure.
+fn db_query<T>(
+    account_id: NetworkAccountId,
+    result: Result<T, miden_node_db::DatabaseError>,
+    context: &str,
+) -> Result<T, ActorShutdownReason> {
+    result.map_err(|err| {
+        tracing::error!(err = err.as_report(), account_id = %account_id, "{context}");
+        ActorShutdownReason::DbError(account_id)
```
igamigo: Should this error variant include the `miden_node_db::DatabaseError` ?

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -29,16 +28,31 @@ use crate::builder::ChainState;
 use crate::db::Db;
 use crate::store::StoreClient;
 
-// ACTOR NOTIFICATION
+/// Converts a database result into an `ActorShutdownReason` error, logging the error on failure.
```
igamigo: Docs are a bit misleading (it's an internal fn so feel free to disregard):

```suggestion
/// Converts a database result mapping the error into an `ActorShutdownReason`, logging it on failure.
```

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -237,11 +249,14 @@ impl AccountActor {
...
-        let has_notes = self
-            .db
-            .has_available_notes(account_id, block_num, self.max_note_attempts)
-            .await
-            .expect("actor should be able to check for available notes");
+        let has_notes = match db_query(
+            account_id,
+            self.db.has_available_notes(account_id, block_num, self.max_note_attempts).await,
+            "failed to check for available notes",
+        ) {
```
igamigo: minor nit, feel free to disregard: A pattern similar to

```rust
self.db.has_available_notes(...).map_err(|err| map_db_err(acc_id, "failed to check notes"))
```

reads a bit better IMO

crates/ntx-builder/src/builder.rs
```diff
@@ -294,15 +291,16 @@ impl NetworkTransactionBuilder {
...
-        match notification {
-            ActorNotification::NotesFailed { nullifiers, block_num } => {
+    /// Processes a request from an account actor.
+    async fn handle_actor_request(&mut self, request: ActorRequest) {
+        match request {
+            ActorRequest::NotesFailed { nullifiers, block_num, ack_tx } => {
                 if let Err(err) = self.db.notes_failed(nullifiers, block_num).await {
                     tracing::error!(err = %err, "failed to mark notes as failed");
                 }
+                let _ = ack_tx.send(());
```
igamigo, line 291: Is this supposed to be acked even if the `db.notes_failed()` call failed?

crates/ntx-builder/src/coordinator.rs
```diff
@@ -194,10 +161,11 @@ impl Coordinator {
...
                     Ok(())
                 },
-                ActorShutdownReason::EventChannelClosed => {
-                    anyhow::bail!("event channel closed");
-                },
                 ActorShutdownReason::SemaphoreFailed(err) => Err(err).context("semaphore failed"),
+                ActorShutdownReason::DbError(account_id) => {
+                    tracing::error!(account_id = %account_id, "Account actor shut down due to DB error");
+                    Ok(())
+                },
```
igamigo: Is the idea not to update the registry and remove the actor handle when this happens? Took a quick look at the other branches and could nto find it either

crates/ntx-builder/src/coordinator.rs
```diff
@@ -129,50 +119,27 @@ impl Coordinator {
...
-                self.actor_registry.remove(&account_id).expect("actor found in send loop above");
-            handle.cancel_token.cancel();
+    /// Each actor will re-evaluate its state from the DB on the next iteration of its run loop.
+    /// Notifications are coalesced: multiple notifications while an actor is busy result in a
+    /// single wake-up.
+    pub fn broadcast(&self) {
+        for handle in self.actor_registry.values() {
+            handle.notify.notify_one();
         }
     }
```
igamigo: Will this scale well? Feels like broadcasting to all actors will cost a lot of unnecessary queries. For instance, I saw that `select_candidate_from_db` queries account state and notes through 2 different queries in the DB. The account state one is probably not trivial though the other one might be. 
I know this is what the PR was partially trying to change but wonder if sending the data here to filter in memory is worth it.

igamigo: Also, I believe there is an low-hanging optimization opportunity with the 2 queries in the DB. You can check for notes first and bail if none are found.

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: I ended up removing it entirely

crates/ntx-builder/src/builder.rs
SantiagoPittella: I propgated the error instead before sending the ack

SantiagoPittella: Now making the builder to halt if it is failing to write to the database, this behavior is more consistent with all the other processes in the sewrvice

crates/ntx-builder/src/coordinator.rs
SantiagoPittella: This might be wasteful at scale, i'm lookking for ways to always send targeted

SantiagoPittella: I sent a new commit addressing this, let me know what do you think

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -261,28 +268,30 @@ impl AccountActor {
...
+                                .await
+                                .map_err(|err| {
+                                    tracing::error!(err = err.as_report(), account_id = %account_id, "failed to check transaction status");
+                                    ActorShutdownReason::DbError(account_id, err)
+                                })
+                            {
+                                Ok(v) => v,
+                                Err(reason) => return reason,
+                            };
+                            if !exists {
```
Mirko-von-Leipzig: I think this is inverted maybe? We're waiting for the tx to arrive, so once it arrives (aka `exists`) we should exit this mode?

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -293,10 +302,13 @@ impl AccountActor {
...
                             // Query DB for latest account and available notes.
-                            let tx_candidate = self.select_candidate_from_db(
+                            let tx_candidate = match self.select_candidate_from_db(
                                 account_id,
                                 chain_state,
-                            ).await;
+                            ).await {
+                                Ok(candidate) => candidate,
+                                Err(shutdown_reason) => return shutdown_reason,
+                            };
```
Mirko-von-Leipzig:
```suggestion
                            ).await?;
```

crates/ntx-builder/src/coordinator.rs
SantiagoPittella: We should update it. I replaced it.

crates/ntx-builder/src/db/models/queries/mod.rs
```diff
@@ -247,13 +262,21 @@ pub fn commit_block(
...
-    Ok(())
+    Ok(affected_accounts.into_iter().collect())
+}
+
+/// Result of reverting transactions.
+pub struct RevertResult {
+    /// All accounts affected by the reverted transactions (need notification).
+    pub affected_accounts: Vec<NetworkAccountId>,
+    /// Accounts whose creation was fully reverted (need cancellation).
+    pub reverted_accounts: Vec<NetworkAccountId>,
```
Mirko-von-Leipzig: The actor should cancel themselves. We shouldn't cancel them.

This way all we need is _always_ a list of affected accounts and nothing else.

crates/ntx-builder/src/db/models/queries/mod.rs
```diff
@@ -290,6 +317,10 @@ pub fn revert_transaction(
             .select(schema::accounts::account_id)
             .load(conn)?;
 
+        for account_id_bytes in &affected_account_ids {
+            affected_accounts.insert(conversions::network_account_id_from_bytes(account_id_bytes)?);
+        }
+
```
Mirko-von-Leipzig: I think we can simplify (or rather reduce) the number of sql queries, by using `RETURNING account_id` in these:

```sql
DELETE FROM accounts WHERE transaction_id = ?1 RETURNING account_id;
DELETE FROM notes WHERE created_by = ?1 RETURNING account_id;
UPDATE notes SET consumed_by = NULL WHERE consumed_by = ?1 RETURNING account_id;
```

crates/ntx-builder/src/builder.rs
```diff
@@ -245,61 +244,64 @@ impl NetworkTransactionBuilder {
...
                     .write_event(&event)
                     .await
                     .context("failed to write TransactionsReverted to DB")?;
 
-                self.coordinator.broadcast(event.clone()).await;
+                self.coordinator.notify_accounts(&result.accounts_to_notify);
 
                 // Cancel actors for reverted account creations.
-                for account_id in &reverted_accounts {
+                for account_id in &result.accounts_to_cancel {
```
Mirko-von-Leipzig: Just again to clarify -- the actor can just be notified along with the rest. When they lookup their account, they'll get `None` and should exit.

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: If the transaction exits means that it is wan't committed/reverted yet, so `!exists` signals that we can take on a new task

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: I replaced the return type of the function to result in order to avoid this matches.

crates/ntx-builder/src/actor/mod.rs
Mirko-von-Leipzig: Not quite -- we submitted a tx to the node and now we're waiting for it to be acknowledged via the mempool subscription and therefore `exists`? Once its committed or reverted, its already past that stage.

Though that does raise some further questions. This might be quite racey if a tx comes in and is reverted or committed before we handle it.

crates/ntx-builder/src/db/models/queries/mod.rs
SantiagoPittella: Done, I removed `RevertResult` and `reverted_accounts` entirely. `revert_transaction` now returns `Vec<NetworkAccountId>` with all affected accounts. The actor self-detects removal in `select_candidate_from_db`: when the account is None we now return `Err(ActorShutdownReason::AccountRemoved)` and shuts itself down.

crates/ntx-builder/src/builder.rs
SantiagoPittella: I addressed this with https://github.com/0xMiden/node/pull/1699#discussion_r2905205750

crates/ntx-builder/src/actor/mod.rs
Mirko-von-Leipzig: Hmm. I'm unsure how to handle this tbh. The idea was to wait for the mempool ack event to arrive before attempting another tx. But now with the database + notify its possible to miss the tx arriving.

Instead of checking for the tx, we could wait for the account commitment changing. But that only works if the tx is committed, and fails if its reverted.

We could add a `last_updated`, or update count field, but that seems extreme..

Mirko-von-Leipzig: But we can punt on this in this PR - we must just ensure we address this before the next release.

crates/ntx-builder/src/db/models/queries/mod.rs
SantiagoPittella: Done!

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -260,55 +266,48 @@ impl AccountActor {
...
+                _ = self.notify.notified() => {
+                    match self.mode {
+                        ActorMode::TransactionInflight(awaited_id) => {
+                            // Check DB: is the inflight tx still pending?
+                            let exists = self
+                                .db
+                                .transaction_exists(awaited_id)
+                                .await
+                                .map_err(|err| {
+                                    tracing::error!(err = err.as_report(), account_id = %account_id, "failed to check transaction status");
```
Mirko-von-Leipzig, line 281: Minor nit: I would put the error tracing inside `inspect_err`

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: ohohhh, shoot. you're right. Until the txadded event `exists` doesn't hold the value that I expected. So using `exists == true` fixes it. It was working because it was stalling the actor until the next event, but since i was testing with the network monitor there was always a new event

Mirko-von-Leipzig: Yeah 😬 But my initial idea is also incorrect so ideally we figure out an alternative path which avoids this entirely somehow.

Maybe the actor should just assume the tx succeeds and proceed as if that's the case. This does complicate matters though because now there is data "missing" in the database so I don't really like that.

The safe and simple solution for now is actually what you've done - which is to wait until the tx is reverted or committed. But that means we can't stack txs which is a bummer - but we can punt on this :)

crates/ntx-builder/src/db/models/queries/mod.rs
```diff
@@ -247,78 +262,69 @@ pub fn commit_block(
...
         diesel::delete(schema::notes::table.filter(schema::notes::created_by.eq(&tx_id_bytes)))
             .execute(conn)?;
 
-        // Un-nullify consumed notes (set consumed_by = NULL where consumed_by = tx_id).
-        diesel::update(schema::notes::table.filter(schema::notes::consumed_by.eq(&tx_id_bytes)))
-            .set(schema::notes::consumed_by.eq(None::<Vec<u8>>))
-            .execute(conn)?;
+        // Restore consumed notes and collect affected account IDs.
+        let restored_accounts: Vec<AccountIdRow> = diesel::sql_query(
+            "UPDATE notes SET consumed_by = NULL WHERE consumed_by = ?1 RETURNING account_id",
```
Mirko-von-Leipzig, line 312: @drahnr is there a way to do this the "diesel" way?

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: I opened https://github.com/0xMiden/node/issues/1769 to revisitr this

crates/ntx-builder/src/db/models/queries/mod.rs
SantiagoPittella: should we wait for a response here or can I merge it?

Mirko-von-Leipzig: I'll merge it, can always follow up

Issue #1697, opened 2026-02-23
Improve node startup time


Mirko-von-Leipzig:
Recent testnet restarts took ~21s, of which 15s was `load_mmr` and 4s `load_smt_forest`.

<img width="805" height="252" alt="Image" src="https://github.com/user-attachments/assets/32cfa486-ac85-4fe3-9686-4ad630d2b3a1" />

`load_mmr` is likely dominated by reading all block headers from the database. However, we don't actually need the entire header -- only the ordered commitments. We can improve this by storing the header fields as columns. At minimum we should store the commitment in its own field, even if we decide to keep the headers as blobs for performance in other areas.

I would also take a look at what else we are using the headers for; if there is other usage where we only require partial fields then we should consider flattening it completely.

Mirko-von-Leipzig:
@kkovaacs I think this is largely complete now?

Mirko-von-Leipzig:
ping @kkovaacs ^^

kkovaacs:
Well, sort of. Node startup time is certainly much better with the persistent backend for `AccountStateForest`. _Populating_ the forest though is still unchanged and PR #1802 certainly improves that.

sergerad:
With persistent RocksDB backends landed, startup on testnet is now ~16s.
Two costs dominate, and both grow linearly with chain state:
<img width="1023" height="330" alt="Image" src="https://github.com/user-attachments/assets/b1de48f0-a791-458a-94ed-ae1287120550" />

The instrumented spans sum to ~11.9s, leaving **~4.4s unattributed**. The main suspects are the
three uninstrumented RocksDB opens (`TreeStorage::create` ×2 and
`ForestPersistentBackend::load`), plus SQLite open/migrations and `BlockStore::load`.

## Problems

1. **`verify_account_state_forest_consistency` (6.9s, ~42% of startup)** full-scans every
   public account on every boot: paged SQLite reads, storage-header deserialization, and
   per-account RocksDB forest lookups — all serial. O(#public accounts) per startup.
2. **~4.4s of unattributed time**, likely dominated by three serial, uninstrumented RocksDB
   opens. Each open replays unflushed WAL segments (up to `max_total_wal_size`) into memtables
   and fills the table cache across five column families — WAL recovery alone can take seconds
   after a run without a clean flush. Notably, `load_account_state_forest` at 0.10s *excludes*
   its backend's open. These opens are also synchronous calls that block the async runtime.
[...]

PR #1700, opened 2026-02-23
feat: improve node startup time

drahnr:
Ref #1697 

# Motivation

The node startup is still rather long, partially because we load more data than actually required.

# Changes

1. avoid loading the full `BlockHeader` and instead adds a column `commitment` ( `crates/store/src/db/models/queries/block_headers.rs` and `crates/store/src/db/schema.rs` )
3. adds `BlockHeaderCommitment` wrapper type, it's a start for avoiding more `Word`-type(les)s

# Caveat

Folds all migrations into the initial one, since it cannot work with the current migration pattern and the changes 'd be too large for this PR, see #1704 for further discussions on that.

### DIFF
--- a/crates/store/src/db/models/queries/block_headers.rs
+++ b/crates/store/src/db/models/queries/block_headers.rs
@@ -212,7 +212,7 @@ impl From<(&BlockHeader, &Signature)> for BlockHeaderInsert {
             block_num: header.block_num().to_raw_sql(),
             block_header: header.to_bytes(),
             signature: signature.to_bytes(),
-            commitment: header.commitment().to_bytes(),
+            commitment: BlockHeaderCommitment::new(header).to_raw_sql(),
         }
     }
 }


crates/store/src/db/migrations/20260223160000_add_block_header_commitment/down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS migrations;
```
drahnr: we should find a better name for the table used for tracking code migrations, open to suggestions

crates/store/src/db/migrations/mod.rs
Mirko-von-Leipzig, line 1: If I understand the setup here correctly, can we not massively simplify matters by assigning every migration an ordered integer index (which they already have implictly), and then force every migration to also have what you call a data migration? Then we don't have to play funny macro / DNS-like resolution of diesel migration to function look up games.

You end up with something like https://github.com/eqlabs/pathfinder/blob/cb5c42b005562fca59c2ef044e83aa1464d4f58f/crates/storage/src/schema.rs#L43-L94

drahnr: That is correct, the main motivation was to not spend too much time on the migration, but get to a functional minimal impl.

I am tempted to just carve it out and squash all migrations and punt on the migrations impl/feat/refactor, to get #1697 done.

Mirko-von-Leipzig: If you're happy to do that; maybe let's start with that -- migrations will need to be solid; so I expect a lot of bikeshedding even in just the design portion.

drahnr: A brief brain-dump of my current thinking https://github.com/0xMiden/node/issues/1704 it's similar to https://github.com/eqlabs/pathfinder/blob/cb5c42b005562fca59c2ef044e83aa1464d4f58f/crates/storage/src/schema.rs#L43-L94 with a bit more automation and keeping the date-as-version approach.

crates/store/src/db/models/queries/block_headers.rs
```diff
@@ -158,13 +204,15 @@ pub struct BlockHeaderInsert {
...
-    fn from(from: (&BlockHeader, &Signature)) -> Self {
+    fn from((header, signature): (&BlockHeader, &Signature)) -> Self {
         Self {
-            block_num: from.0.block_num().to_raw_sql(),
-            block_header: from.0.to_bytes(),
-            signature: from.1.to_bytes(),
+            block_num: header.block_num().to_raw_sql(),
+            block_header: header.to_bytes(),
+            signature: signature.to_bytes(),
+            commitment: header.commitment().to_bytes(),
```
Mirko-von-Leipzig:
```suggestion
            commitment: BlockHeaderCommitment::new(header).to_raw_sql(),
```

PR #1701, opened 2026-02-23
fix: TransactionHeader serialization

juan518munoz:
Debugging #1691 with the miden client, found the following errors:

### The `TransactionHeader` type changes from [miden-base #1973](https://github.com/0xMiden/miden-base/pull/1973/changes#diff-b5fe86f21f143a6db466be20c0f3213b0fb82b15a1e354566a7a3c89e37f673bR30-R31) introduce a serialization mismatch in `TransactionSummaryRowInsert::new`.

The read side (`TransactionRecordRaw::try_into`) deserializes stored bytes as `Vec<Nullifier>` and `Vec<NoteId>`, but after the base update:

- `transaction_header.input_notes()` returns `InputNotes<InputNoteCommitment>` (was `Vec<Nullifier>`), where `InputNoteCommitment` includes an `Option<NoteHeader>` alongside the nullifier.
- `transaction_header.output_notes()` returns `&[NoteHeader]` (was `Vec<NoteId>`), where `NoteHeader` includes `NoteMetadata` alongside the note ID.

`Calling .to_bytes()` on these richer types produces a different binary format than what the read side expects, so we need to extract just the nullifiers and note IDs before serializing.

### Cursor/push logic

Old code accumulated transactions in the loop but only stored the last one in `last_added_tx`. After the loop, it pushed that single transaction and updated the cursor. This was a bug — only the last transaction from each chunk was actually pushed to `all_transactions`, thus all earlier transactions in the chunk were silently dropped.

Fix pushes each transaction and updates the cursor immediately inside the loop, so every transaction that fits within the size limit is correctly added to `all_transactions`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Validator now persists validated transactions ([#1614](https://github.com/0xMiden/miden-node/pull/1614)).
 - [BREAKING] Remove `SynState` and introduce `SyncChainMmr` ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
 - Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
+- Fixed `TransactionHeader` serialization for row insertion on database & fixed transaction cursor on retrievals ([#1701](https://github.com/0xMiden/miden-node/issues/1701)).
 
 ### Changes
 


drahnr:
This also needs a migration which is about to be introduced in #1700 , or do we never query the TX header from DB?

docs/external/src/rpc.md
```diff
@@ -217,12 +216,6 @@ Caller specifies the `account_id` of the public account and the block range (`bl
 
 This endpoint enables clients to maintain an updated view of account storage.
 
-### SyncChainMmr
```
bobbinth, line 220: Question: why did we remove this section (and also the entry on line 26 above)?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -125,11 +125,25 @@ impl TransactionSummaryRowInsert {
     ) -> Self {
         const HEADER_BASE_SIZE: usize = 4 + 32 + 16 + 64; // block_num + tx_id + account_id + commitments
 
-        // Serialize input notes using binary format (store nullifiers)
-        let nullifiers_binary = transaction_header.input_notes().to_bytes();
+        // Extract nullifiers from input notes and serialize them.
+        // We only store the nullifiers (not the full InputNoteCommitment) since
```
drahnr: nit: I'd appreciate backticks for types in comments

bin/node/src/commands/mod.rs
```diff
@@ -118,7 +118,7 @@ pub struct BundledValidatorConfig {
 }
 
 impl BundledValidatorConfig {
-    /// Converts the [`ValidatorConfig`] into a URL and an optional [`SocketAddr`].
+    /// Converts the `ValidatorConfig` into a URL and an optional [`SocketAddr`].
```
drahnr:
```suggestion
    /// Converts the [`crate::commands::ValidatorConfig`] into a URL and an optional [`SocketAddr`].
```

docs/external/src/rpc.md
juan518munoz: It was mistakenly removed, fixed!

bin/node/src/commands/mod.rs
juan518munoz: Linting was not part of actual code changes, as of rebasing into `next`, suggestion no longer applies.

PR #1703, opened 2026-02-24
fix: move proto code generation output to `OUT_DIR`, out of tree

drahnr:
# Motivation

Committing generated protobuf code is discouraged and creates review churn. Moving artifacts to `OUT_DIR` removes ~10k LOC from the repo and avoids frequent diffs when protobuf definitions change. `rust-analyzer` resolves `include!` paths, so location is no longer a blocker when it comes to symbol resolution / editor support.

# Implementation

We keep the existing integration wiring by continuing to use i.e. `generated/mod.rs` and `include!(concat!(env!("OUT_DIR"), "/generated/…"))`. This minimizes import-path changes across the codebase and keeps the diff concise.

# Side effects

- Removes the `docs.rs`-specific `BUILD_PROTO=1` special casing.
- Proto code is no longer tracked in git -> requires a working protobuf compiler installed

# Risks

- **Build reproducibility**: if protobuf compiler has a breaking change we depend on the host installed compiler


Mirko-von-Leipzig (commented):
I think is better and minimizes the review diffs when proto changes are involved.

It also makes sense to me to drop the `generated/mod.rs` codegen but as you say that probably makes more sense as a separate PR. This file can be created manually.

This looks correct; but PR is still in draft and description looks semi-done so I'm not approving yet.

.github/workflows/publish-main.yml
```diff
@@ -18,8 +18,8 @@ jobs:
         with:
           fetch-depth: 0
           ref: main
-      - name: Install RocksDB
-        uses: ./.github/actions/install-rocksdb
+      - uses: ./.github/actions/install-rocksdb
+      - uses: ./.github/actions/install-protobuf-compiler
```
Mirko-von-Leipzig, line 22: Given that these are now married at the hip, should we have a single `install-system-prerequisites`?

.github/actions/install-protobuf-compiler/action.yml
Mirko-von-Leipzig, line 1: This isn't actually installing anything though?

bin/remote-prover/src/server/prover.rs
```diff
@@ -31,8 +31,8 @@ impl Prover {
         }
     }
 
-    /// Proves a [`ProofRequest`] using the appropriate prover implementation as specified during
-    /// construction.
+    /// Proves a [`proto::ProofRequest`] using the appropriate prover implementation as specified
```
Mirko-von-Leipzig, line 34: Any idea why this wasn't caught by clippy or doc lints?

drahnr: No idea, it's definitely popping up with `make doc` locally. I can bisect where it got introduced, but also not sure if that's worth it.

.github/workflows/publish-main.yml
drahnr: One `CTRL-K` too much. Fixed

I considered merging them, but wasn't sure if won't get more changes in rocksdb when I migrate the `LargeSmtForest` backend (!= `LargeSmt` backend).

.github/actions/install-protobuf-compiler/action.yml
drahnr: Fixed

PR #1705, opened 2026-02-24
chore: ntx builder actor deactivation

SantiagoPittella:
Closes the third task of https://github.com/0xMiden/node/issues/1694

- Add actor sterility timeout: default 5 min
- Re-activate deactivated actors when new notes target them via send_targeted, which now returns inactive target account IDs so the builder can re-spawn actors.
- Add coordinator unit tests for shutdown approval/rejection and targeted notification routing.
- Improves a bit the tests helpers org.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - NTX Builder actors now deactivate after being idle for a configurable idle timeout (`--ntx-builder.idle-timeout`, default 5 min) and are re-activated when new notes target their account ([#1705](https://github.com/0xMiden/node/pull/1705)).
 - [BREAKING] Modified `TransactionHeader` serialization to allow converting back into the native type after serialization ([#1759](https://github.com/0xMiden/node/issues/1759)).
 - Removed `chain_tip` requirement from mempool subscription request ([#1771](https://github.com/0xMiden/node/pull/1771)).
+- Moved bootstrap procedure to `miden-node validator bootstrap` command ([#1764](https://github.com/0xMiden/node/pull/1764)).
 
 ### Fixes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3227,6 +3227,7 @@ dependencies = [
  "anyhow",
  "bytes",
  "clap",
+ "fs-err",
  "governor",
  "http 1.4.0",
  "http-body-util",
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -96,16 +96,21 @@ impl BundledCommand {
                 genesis_config_file,
                 validator_key,
             } => {
-                // Currently the bundled bootstrap is identical to the store's bootstrap.
-                crate::commands::store::StoreCommand::Bootstrap {
-                    data_directory,
-                    accounts_directory,
-                    genesis_config_file,
+                // Run validator bootstrap to create genesis block + account files.
+                crate::commands::validator::ValidatorCommand::bootstrap_genesis(
+                    &data_directory,
+                    &accounts_directory,
+                    genesis_config_file.as_ref(),
                     validator_key,
-                }
-                .handle()
+                )
                 .await
-                .context("failed to bootstrap the store component")
+                .context("failed to bootstrap genesis block")?;
+
+                // Feed the genesis block file into the store bootstrap.
+                let genesis_block_path =
+                    data_directory.join(crate::commands::validator::GENESIS_BLOCK_FILENAME);
+                crate::commands::store::bootstrap_store(&data_directory, &genesis_block_path)
+                    .context("failed to bootstrap the store component")
             },
             BundledCommand::Start {
                 rpc_url,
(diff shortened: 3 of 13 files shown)


bin/node/src/commands/mod.rs
```diff
@@ -125,6 +126,18 @@ pub struct NtxBuilderConfig {
     )]
     pub script_cache_size: NonZeroUsize,
 
+    /// Duration an actor must remain idle before requesting shutdown.
+    ///
+    /// When an actor has no viable notes for this duration, it will be deactivated to free
+    /// resources. It will be re-activated when new notes target it.
+    #[arg(
+        long = "ntx-builder.sterility-timeout",
```
Mirko-von-Leipzig: Having now read this; `sterility` is maybe not the best term 😅 

Let's just go with `idle-timeout`?

bin/node/src/commands/mod.rs
```diff
@@ -125,6 +126,18 @@ pub struct NtxBuilderConfig {
     )]
     pub script_cache_size: NonZeroUsize,
 
+    /// Duration an actor must remain idle before requesting shutdown.
+    ///
+    /// When an actor has no viable notes for this duration, it will be deactivated to free
+    /// resources. It will be re-activated when new notes target it.
```
```suggestion
    /// Duration after which an idle network account will deactivate.
    ///
    /// An account is considered idle once it has no viable notes to consume.
    /// A deactivated account will reactivate if targetted with new notes.
```

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -325,10 +349,31 @@ impl AccountActor {
                         }
                     }
                 }
+                // Sterility timeout: actor has been idle too long, request shutdown.
+                _ = sterility_sleep => {
+                    match self.initiate_shutdown(account_id).await {
+                        Ok(()) => return ActorShutdownReason::Sterile(account_id),
+                        Err(()) => self.mode = ActorMode::NotesAvailable,
+                    }
+                }
```
Mirko-von-Leipzig: I understand that this is following the established pattern, but I think we're needlessly complicating matters.

If the actor wants to exit, just exit the task. We don't actually need a shutdown handle or token either.

`ActorShutdownReason` can be replaced with a conventional `-> Result<(), Error>`. This may complicate this PR, so I'm okay with simply returning, and cleaning that up separately.


```suggestion
                // Sterility timeout: actor has been idle too long, deactivate account
                _ = sterility_sleep => return ActorShutdownReason::Sterile,
```

crates/ntx-builder/src/coordinator.rs
```diff
@@ -261,10 +284,161 @@ impl Coordinator {
...
+            tracing::debug!(
+                %account_id,
+                "Rejected actor shutdown: notes available in DB"
+            );
+        } else {
+            self.actor_registry.remove(&account_id);
+            let _ = ack_tx.send(());
+        }
+    }
+
```
Mirko-von-Leipzig: I would avoid this entirely. Let the actor deactivate itself.

In the `select!` arm that reaps actor tasks, we can then check if the deactivated actor should be restarted by inspecting that actor's notify channel (which we have a copy of of here). If there is a pending notification, then we know the actor didn't handle it. 

This should be _rare_ because it implies the actor idle timed out just as we notified it. And therefore, even though this code is technically more optimal because the actor doesn't have to respawn, I wouldn't be concerned about the performance angle.

SantiagoPittella: Ok, sounds good. I will address this with the previous comment too

crates/ntx-builder/src/coordinator.rs
```diff
@@ -151,25 +158,42 @@ impl Coordinator {
...
+                    // Remove the actor from the registry, but check if a notification arrived
+                    // just as the actor timed out. If so, the caller should respawn it.
+                    let should_respawn =
+                        self.actor_registry.remove(&account_id).is_some_and(|handle| {
+                            let notified = handle.notify.notified();
+                            tokio::pin!(notified);
+                            notified.enable()
+                        });
+
+                    Ok(should_respawn.then_some(account_id))
```
Mirko-von-Leipzig, line 211: This works, but I really still think we shouldn't need `ActorShutdownReason`. The actor return value can just be a `anyhow::Result<()>`. You'll also notice that `next` returns a result, but we never actually return a result.

SantiagoPittella: We use the `ActorShutdownReason` in the coordinator side to determinate if we should re-spawn it

Mirko-von-Leipzig: Yes but there is only one reason to respawn it (currently, and for the foreseeable future imo)

- if the actor itself decides to shutdown, and
   - if the actor had unhandled notifications
   - then we should restart it

An actor currently shuts down (on purpose) if:
- it idles for too long
- its creation tx was reverted

Everything else is a crash/error/bug.

We can therefore in theory say that all `Ok(())` means check notify and possibly respawn, and all `Err(_)` means increment crash counter and respawn.

I'm okay with merging as is, and we can litigate/discuss after. Basically I want to reduce the number of branches and conditionals because these lead to edge cases and bugs.

SantiagoPittella: Oh ok, I misunderstood before. It should be a simple change. To avoid generating conflicts in this PR https://github.com/0xMiden/node/pull/1712 I will address this there since it is related to that PR too

PR #1707, opened 2026-02-24
fix(rpc): move CORS layer above AcceptHeaderLayer

WiktorStarczewski:
## Summary

Reorders the tower middleware stack so that `cors_for_grpc_web_layer()` is applied **before** `AcceptHeaderLayer`.

## Problem

When a browser-based gRPC-Web client (e.g. the Miden web SDK) connects to the RPC server with an SDK version the server doesn't recognize, `AcceptHeaderLayer` rejects the request by returning a `tonic::Status::invalid_argument` response directly via `futures::future::ready()`. This short-circuits all downstream middleware — including the CORS layer.

Because the CORS layer sat below `AcceptHeaderLayer` in the middleware stack, rejection responses were missing `Access-Control-Allow-Origin` and other CORS headers entirely. The browser then blocked the response at the network level, giving the user an opaque `net::ERR_FAILED` / "CORS policy" error instead of the actual gRPC error message ("server does not support any of the specified application/vnd.miden content types").

This made it impossible to diagnose version mismatches from a browser environment — the real error was silently swallowed by the browser's CORS enforcement.

## Root Cause

In tower's `.layer()` builder, layers listed earlier are **outer** layers — they see responses last on the way out. The previous ordering was:

```
.layer(HealthCheckLayer)
.layer(AcceptHeaderLayer)         // rejects here → returns response directly
.layer(cors_for_grpc_web_layer()) // never sees rejection responses
.layer(GrpcWebLayer)
```

When `AcceptHeaderLayer` rejected a request, the response traveled back out through `HealthCheckLayer` and `TraceLayer` but never passed through `CorsLayer`, since it was an inner layer that the request never reached.

## Fix

```
.layer(HealthCheckLayer)
.layer(cors_for_grpc_web_layer()) // now wraps AcceptHeaderLayer
.layer(AcceptHeaderLayer)         // rejections still get CORS headers
.layer(GrpcWebLayer)
```

By moving the CORS layer above the accept header check, all responses — including version rejections — are wrapped with proper CORS headers. The browser can then read and surface the actual gRPC error to the application.

## Verification

Tested with `curl` against the two scenarios:

**Before (no CORS headers on rejection):**
```
$ curl -X POST "https://rpc.testnet.miden.io/rpc.Api/GetBlockHeaderByNumber" \
  -H "Origin: http://localhost:3000" \
  -H "Content-Type: application/grpc-web+proto" \
  -H "accept: application/vnd.miden; version=0.14.0"

HTTP/2 200
content-type: application/grpc
grpc-message: server does not support any of the specified application/vnd.miden content types
grpc-status: 3
# ← no Access-Control-Allow-Origin
```

**After (CORS headers present on rejection):**
The same request would include `Access-Control-Allow-Origin`, `Access-Control-Allow-Credentials`, and `Access-Control-Expose-Headers`, allowing the browser to read the gRPC error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.13.6 (TBD)
 
-- Fixed CORS headers missing from version-rejection responses by reordering the CORS middleware above `AcceptHeaderLayer` ([#1707](https://github.com/0xMiden/node/pull/1707)).
+- Fixed CORS headers missing from version-rejection responses ([#1707](https://github.com/0xMiden/node/pull/1707)).
 
 ## v0.13.5 (2026-02-19)
 


crates/rpc/src/server/mod.rs
```diff
@@ -84,12 +84,12 @@ impl Rpc {
...
             .layer(HealthCheckLayer)
+            .layer(cors_for_grpc_web_layer())
             .layer(
                 AcceptHeaderLayer::new(&rpc_version, genesis.commitment())
                     .with_genesis_enforced_method("SubmitProvenTransaction")
                     .with_genesis_enforced_method("SubmitProvenBatch"),
             )
-            .layer(cors_for_grpc_web_layer())
             // Enables gRPC-web support.
             .layer(GrpcWebLayer::new())
```
Mirko-von-Leipzig, line 97: Should this one travel with it? I suppose not because you've tested it; but it may make sense to keep the gRPC web layer's close together. 

In the future we may combine these into a single layer and then the implementer might choose the wrong location.

crates/rpc/src/server/mod.rs
```diff
@@ -84,12 +84,12 @@ impl Rpc {
             .layer(CatchPanicLayer::custom(catch_panic_layer_fn))
             .layer(TraceLayer::new_for_grpc().make_span_with(grpc_trace_fn))
             .layer(HealthCheckLayer)
+            .layer(cors_for_grpc_web_layer())
```
Mirko-von-Leipzig, line 90: Maybe you can phrase it more betterer.

```suggestion
            // Note: must come before the accept layer, as otherwise accept rejections
            // do _not_ get CORS headers applied, masking the accept error in
            // web-clients (which would experience CORS rejection).
            .layer(cors_for_grpc_web_layer())
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.13.6 (TBD)
+
+- Fixed CORS headers missing from version-rejection responses by reordering the CORS middleware above `AcceptHeaderLayer` ([#1707](https://github.com/0xMiden/node/pull/1707)).
```
Mirko-von-Leipzig: Nit, readers don't need to know the internals

```suggestion
- Fixed CORS headers missing from version-rejection responses ([#1707](https://github.com/0xMiden/node/pull/1707)).
```

crates/rpc/src/server/mod.rs
WiktorStarczewski: Fair point, but no it should not move, because:
 1. They serve different purposes in the middleware stack. GrpcWebLayer translates the wire protocol (converting gRPC-Web requests into standard gRPC for the inner service). It needs to be as close to the service as possible — it must run before tonic's gRPC handling sees the request. cors_for_grpc_web_layer adds CORS headers to responses, and needs to be above AcceptHeaderLayer so rejections still get CORS headers.
  2. The ordering constraint is different for each. The CORS layer must wrap AcceptHeaderLayer. But GrpcWebLayer must wrap the actual gRPC services - it's doing protocol translation that the AcceptHeaderLayer and other middleware expect to already be done (or not care about).
  3. Moving GrpcWebLayer up could break things. If GrpcWebLayer runs before AcceptHeaderLayer, the accept header checking would see already translated requests, which could change behavior.

PR #1709, opened 2026-02-25
feat: expose rpc listener

juan518munoz:
After #1688 the miden-client lost the capability of using the testing prover due to it's previously used structures/methods becoming private or removed. This PR re-adds this functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Validator now persists validated transactions ([#1614](https://github.com/0xMiden/miden-node/pull/1614)).
 - [BREAKING] Remove `SynState` and introduce `SyncChainMmr` ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
 - Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
-- Fixed `TransactionHeader` serialization for row insertion on database & fixed transaction cursor on retrievals ([#1701](https://github.com/0xMiden/miden-node/issues/1701)).
+- Re-Introduce `RpcListener` & expose remote prover as a library to allow instantiation on other crates ([#1709](https://github.com/0xMiden/miden-node/issues/1709)).
 
 ### Changes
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - Improved tracing span fields ([#1650](https://github.com/0xMiden/miden-node/pull/1650))
  - Replaced NTX Builder's in-memory state management with SQLite-backed persistence; account states, notes, and transaction effects are now stored in the database and inflight state is purged on startup ([#1662](https://github.com/0xMiden/miden-node/pull/1662)).
 - [BREAKING] Reworked `miden-remote-prover`, removing the `worker`/`proxy` distinction and simplifying to a `worker` with a request queue ([#1688](https://github.com/0xMiden/miden-node/pull/1688)).
+- Fixed `TransactionHeader` serialization for row insertion on database & fixed transaction cursor on retrievals ([#1701](https://github.com/0xMiden/miden-node/issues/1701)).
```
juan518munoz, line 27: Moved this previous entry here as it is better suited here.

bin/remote-prover/src/lib.rs
Mirko-von-Leipzig, line 1: I'm going to need some more context in how this is being used? If its just for testing why not just implement a very basic proving server?

As a general rule I'm not very comfortable having "libraries" exposed from the node since we're a binary application. Changes here would not be considered breaking.

juan518munoz: > I'm going to need some more context in how this is being used

[It's used on the testing prover of miden-client](https://github.com/0xMiden/miden-client/blob/f3e9c93edf9f3664b9be8a0b21812cea4d20b296/crates/testing/prover/src/main.rs#L20-L21)

> If its just for testing why not just implement a very basic proving server

It's just for testing, so implementing everything from scratch is a possible solution. I've pushed a possible implementation to the miden-client update PR [755123e](https://github.com/0xMiden/miden-client/pull/1815/commits/755123ef9a9ee2f1acfebea57219831040568b61). It's not that more complicated so we can close this PR and just roll with it.

SantiagoPittella: We were using this a library in the client also to benefit for rust artifacts caching in the CI, the same we do with the node using its components as library. I agree that it is not completely necessary, best case scenario we should use the actual binary from the node to ensure that the integration works.

igamigo: Yeah, I think we briefly discussed this with @juan518munoz: We could clone the binary and run it, it's just a bit more cumbersome to manage revisions and artifact caching (with the upside that it tests the integration better)

Mirko-von-Leipzig: I mean if its for our own internal testing purposes; I don't mind it so much. 

I'm just thinking about our "internal" RPC client that then got used by actual teams and I'd prefer to not have that be an option. If exposing this is easiest for you then lets do it, and just note in the readme and doc comments that its not stable.

PR #1711, opened 2026-02-25
fix(rpc): check for public state

igamigo:
Updated storage map sync validation to use `has_public_state()` instead of `is_public()`, allowing network accounts to sync storage maps. Applied this in both RPC and database query guards, and added a regression test.

### DIFF
--- a/crates/store/src/db/models/queries/accounts.rs
+++ b/crates/store/src/db/models/queries/accounts.rs
@@ -666,7 +666,7 @@ pub(crate) fn select_account_storage_map_values(
         2 * size_of::<Word>() + size_of::<u32>() + size_of::<u8>(); // key + value + block_num + slot_idx
     pub const MAX_ROWS: usize = MAX_RESPONSE_PAYLOAD_BYTES / ROW_OVERHEAD_BYTES;
 
-    if !account_id.is_public() {
+    if !account_id.has_public_state() {
         return Err(DatabaseError::AccountNotPublic(account_id));
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.13.7 (2026-02-25)
+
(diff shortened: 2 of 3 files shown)


crates/store/src/db/models/queries/accounts.rs
```diff
@@ -370,7 +370,7 @@ pub(crate) fn select_account_vault_assets(
     const ROW_OVERHEAD_BYTES: usize = 2 * size_of::<Word>() + size_of::<u32>(); // key + asset + block_num
     const MAX_ROWS: usize = MAX_RESPONSE_PAYLOAD_BYTES / ROW_OVERHEAD_BYTES;
 
-    if !account_id.is_public() {
+    if !account_id.has_public_state() {
         return Err(DatabaseError::AccountNotPublic(account_id));
     }
```
bobbinth, line 375: This issue came up a few times now. I wonder if we need to come up with a better name. For example, something like `has_onchain_state()` or `is_onchain()`.

bobbinth (approved):
Looks good! Thank you!

Could you also add a changelog entry?

crates/store/src/db/models/queries/accounts.rs
igamigo: I think `has_public_state()` did not exist when this was first introduced (would have had to check for `is_public() || is_network()`), though not sure. But indeed it can be confusing.

Mirko-von-Leipzig: imo the problem is `is_public` is ambiguous, and changing `has_public_state` won't really help. The core issue is that when typing one goes `account.is_` and IDE proposes `is_public` and that's the "obviously" correct result we are looking for.

Ideally we would have an API where devs fall into the "pit of success". In a way this derives from our categories being overly flat i.e.

```rust
enum AccountKind {
    /// Private implies there is public
    Private,
    /// And public completes the "set"
    Public,
    /// Network is "public", but also different.
    Network,
}

/// This is better "English" categories imo, if we're using public.
enum AccountKind {
    Private,
    Public(User | Network)
}
```

If we want to keep things similar as is, I would propose something like `is_public_user` and `is_public_network`. Ideally though, `AccountKind::Public` would be called something else.

PR #1712, opened 2026-02-26
feat(ntx-builder): deactivate accounts which crash repeatedly

SantiagoPittella:
Implements account blacklisting for the NTX Builder (4th task from #1694).

Track crash counts per account in the coordinator. When an actor shuts down due to a `DbError`, its crash count is incremented. Once it reaches a configurable threshold, the account is blacklisted and `spawn_actor` skips it.

Only DbError shutdowns count as crashes because other shutdown reasons (`Cancelled`, `IdleTimeout`, `SemaphoreFailed`) are either intentional or system-wide and not indicative of a per-account bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Enhancements
 
+- Expose per-tree RocksDB tuning options ([#1782](https://github.com/0xMiden/node/pull/1782)).
 - Added verbose `info!`-level logging to the network transaction builder for transaction execution, note filtering failures, and transaction outcomes ([#1770](https://github.com/0xMiden/node/pull/1770)).
 - [BREAKING] Move block proving from Blocker Producer to the Store ([#1579](https://github.com/0xMiden/node/pull/1579)).
 - [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/node/pull/1595)).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3234,6 +3234,7 @@ dependencies = [
  "humantime",
  "itertools 0.14.0",
  "lru",
+ "miden-large-smt-backend-rocksdb",
  "miden-protocol",
  "opentelemetry",
  "opentelemetry-otlp",
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -5,7 +5,7 @@ use anyhow::Context;
 use miden_node_block_producer::BlockProducer;
 use miden_node_rpc::Rpc;
 use miden_node_store::Store;
-use miden_node_utils::clap::GrpcOptionsExternal;
+use miden_node_utils::clap::{GrpcOptionsExternal, StorageOptions};
 use miden_node_utils::grpc::UrlExt;
 use miden_node_validator::{Validator, ValidatorSigner};
 use miden_protocol::crypto::dsa::ecdsa_k256_keccak::SecretKey;
@@ -84,6 +84,9 @@ pub enum BundledCommand {
 
         #[command(flatten)]
         grpc_options: GrpcOptionsExternal,
+
+        #[command(flatten)]
+        storage_options: StorageOptions,
     },
 }
 
@@ -121,6 +124,7 @@ impl BundledCommand {
                 validator,
                 enable_otel: _,
                 grpc_options,
+                storage_options,
             } => {
                 Self::start(
                     rpc_url,
@@ -130,13 +134,14 @@ impl BundledCommand {
                     ntx_builder,
                     validator,
                     grpc_options,
+                    storage_options,
                 )
                 .await
             },
         }
     }
 
-    #[expect(clippy::too_many_lines)]
+    #[expect(clippy::too_many_lines, clippy::too_many_arguments)]
     async fn start(
         rpc_url: Url,
         block_prover_url: Option<Url>,
@@ -145,6 +150,7 @@ impl BundledCommand {
         ntx_builder: NtxBuilderConfig,
         validator: BundledValidatorConfig,
         grpc_options: GrpcOptionsExternal,
+        storage_options: StorageOptions,
     ) -> anyhow::Result<()> {
         // Start listening on all gRPC urls so that inter-component connections can be created
         // before each component is fully started up.
@@ -202,6 +208,7 @@ impl BundledCommand {
                     data_directory: data_directory_clone,
                     block_prover_url,
                     grpc_options: grpc_options.into(),
+                    storage_options,
                 }
                 .serve()
                 .await
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 use anyhow::Context;
 use miden_node_store::Store;
 use miden_node_store::genesis::GenesisBlock;
-use miden_node_utils::clap::GrpcOptionsInternal;
+use miden_node_utils::clap::{GrpcOptionsInternal, StorageOptions};
 use miden_node_utils::fs::ensure_empty_directory;
 use miden_node_utils::grpc::UrlExt;
 use miden_protocol::block::ProvenBlock;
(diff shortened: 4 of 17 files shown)


crates/ntx-builder/src/coordinator.rs
```diff
@@ -117,6 +130,18 @@ impl Coordinator {
     pub fn spawn_actor(&mut self, origin: AccountOrigin, actor_context: &AccountActorContext) {
         let account_id = origin.id();
 
+        // Skip spawning if the account has been blacklisted due to repeated crashes.
+        if let Some(&count) = self.crash_counts.get(&account_id) {
+            if count >= self.max_actor_crashes {
+                tracing::warn!(
+                    %account_id,
+                    crash_count = count,
+                    "Account blacklisted due to repeated crashes, skipping actor spawn"
```
drahnr: Q: should we make this more explicit? `tracing::error!`, since it signifies a bug in our impl or add "BUG" deliberately to the message?

crates/ntx-builder/src/coordinator.rs
```diff
@@ -117,6 +130,18 @@ impl Coordinator {
...
+                );
+                return;
+            }
+        }
+
```
drahnr, line 151: Q: should we first check for running actors and terminate them?

crates/ntx-builder/src/coordinator.rs
```diff
@@ -358,4 +427,47 @@ mod tests {
         assert_eq!(inactive_targets.len(), 1);
         assert_eq!(inactive_targets[0], inactive_id);
     }
+
+    // BLACKLIST TESTS
```
drahnr: nit: we might want to find a different term to "blacklist", i.e. `repeated_failure_block_list` to avoid any confusion / fud

bin/node/src/commands/mod.rs
```diff
@@ -138,6 +138,13 @@ pub struct NtxBuilderConfig {
     )]
     pub idle_timeout: Duration,
 
+    /// Maximum number of crashes before an account actor is blacklisted.
+    ///
+    /// Once an actor for a given account exceeds this crash count, no new actor will be
+    /// spawned for that account.
```
```suggestion
    /// Maximum number of crashes before an account deactivated.
    ///
    /// Once this limit is reached, no new transactions will be created for this account.
```

bin/node/src/commands/mod.rs
```diff
@@ -138,6 +138,13 @@ pub struct NtxBuilderConfig {
...
+    #[arg(long = "ntx-builder.max-actor-crashes", default_value_t = 10, value_name = "NUM")]
+    pub max_actor_crashes: usize,
```
Mirko-von-Leipzig: I would avoid mentioning `actor` in the public docs -- its more an implementation detail.


```suggestion
    #[arg(long = "ntx-builder.max-account-crashes", default_value_t = 10, value_name = "NUM")]
    pub max_account_crashes: usize,
```

crates/ntx-builder/src/coordinator.rs
Mirko-von-Leipzig: We should already be logging the crash itself, so this is probably okay at `warn` since it will repeat often for each account.

crates/ntx-builder/src/coordinator.rs
```diff
@@ -117,6 +130,18 @@ impl Coordinator {
     pub fn spawn_actor(&mut self, origin: AccountOrigin, actor_context: &AccountActorContext) {
         let account_id = origin.id();
 
+        // Skip spawning if the account has been blacklisted due to repeated crashes.
+        if let Some(&count) = self.crash_counts.get(&account_id) {
+            if count >= self.max_actor_crashes {
+                tracing::warn!(
+                    %account_id,
```
Mirko-von-Leipzig:
```suggestion
                    account.id = %account_id,
```

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
  - Replaced NTX Builder's in-memory state management with SQLite-backed persistence; account states, notes, and transaction effects are now stored in the database and inflight state is purged on startup ([#1662](https://github.com/0xMiden/miden-node/pull/1662)).
 - [BREAKING] Reworked `miden-remote-prover`, removing the `worker`/`proxy` distinction and simplifying to a `worker` with a request queue ([#1688](https://github.com/0xMiden/miden-node/pull/1688)).
 - NTX Builder actors now deactivate after being idle for a configurable idle timeout (`--ntx-builder.idle-timeout`, default 5 min) and are re-activated when new notes target their account ([#1705](https://github.com/0xMiden/miden-node/pull/1705)).
+- NTX Builder now blacklists network accounts whose actors crash repeatedly (configurable via `--ntx-builder.max-actor-crashes`, default 10) ([#1712](https://github.com/0xMiden/miden-node/pull/1712)).
```
```suggestion
- NTX Builder now deactivates network accounts which crash repeatedly (configurable via `--ntx-builder.max-actor-crashes`, default 10) ([#1712](https://github.com/0xMiden/miden-node/pull/1712)).
```

crates/ntx-builder/src/coordinator.rs
Mirko-von-Leipzig: I'd prefer something much shorter -- we're already overly verbose and stuttery throughout the codebase imo. Maybe `deactivated_accounts`

crates/ntx-builder/src/coordinator.rs
```diff
@@ -305,17 +336,55 @@ impl Coordinator {
...
         let (db, dir) = Db::test_setup().await;
-        (Coordinator::new(4, db), dir)
+        (Coordinator::new(4, 10, db), dir)
+    }
+
+    /// Creates a minimal `AccountActorContext` suitable for unit tests.
+    ///
+    /// The URLs are fake and actors spawned with this context will fail on their first gRPC call,
+    /// but this is sufficient for testing coordinator logic (registry, blacklisting, etc.).
+    fn test_actor_context(db: &Db) -> AccountActorContext {
```
Mirko-von-Leipzig: Maybe have this be a method on `AccountActorContext::test`?

crates/ntx-builder/src/coordinator.rs
```diff
@@ -305,17 +336,55 @@ impl Coordinator {
...
+    use crate::chain_state::ChainState;
+    use crate::clients::StoreClient;
     use crate::db::Db;
     use crate::test_utils::*;
 
     /// Creates a coordinator with default settings backed by a temp DB.
     async fn test_coordinator() -> (Coordinator, tempfile::TempDir) {
         let (db, dir) = Db::test_setup().await;
-        (Coordinator::new(4, db), dir)
+        (Coordinator::new(4, 10, db), dir)
```
Mirko-von-Leipzig: This could also be a method on `Coordinator`

crates/ntx-builder/src/coordinator.rs
```diff
@@ -173,7 +198,13 @@ impl Coordinator {
                 },
                 ActorShutdownReason::SemaphoreFailed(err) => Err(err).context("semaphore failed"),
                 ActorShutdownReason::DbError(account_id) => {
-                    tracing::error!(account_id = %account_id, "Account actor shut down due to DB error");
+                    let count = self.crash_counts.entry(account_id).or_insert(0);
+                    *count += 1;
+                    tracing::error!(
+                        %account_id,
```
Mirko-von-Leipzig:
```suggestion
                        account.id = %account_id,
```

crates/ntx-builder/src/coordinator.rs
SantiagoPittella: I dont think so. By the time the crash threshold is reached, the actor has already shut itself down. `DbError` shutdowns are handled in `next()`, which removes the actor from the registry before any re-spawn attempt. So when `spawn_actor` is called for a deactivated account, there's no running actor to terminate.

crates/ntx-builder/src/coordinator.rs
```diff
@@ -91,7 +89,7 @@ pub struct Coordinator {
     /// This join set allows the coordinator to wait for actor task completion and handle
     /// different shutdown scenarios. When an actor task completes (either successfully or
     /// due to an error), the corresponding entry is removed from the actor registry.
-    actor_join_set: JoinSet<ActorShutdownReason>,
+    actor_join_set: JoinSet<(NetworkAccountId, anyhow::Result<()>)>,
```
Mirko-von-Leipzig: I suspect we should be tracking `task ID -> `NetworkAccountId` because I don't think we can get the account ID when the task panics, but still need to review the rest.

PR #1716, opened 2026-02-27
chore: use `build-rs` in build scripts

Mirko-von-Leipzig:
[`build-rs`](https://docs.rs/build-rs/latest/build_rs/) makes it harder to @#$%-up our build scripts. This is more a nice-to-have than anything else, ran across this while trying to add more complex codegen :)

> provides a strongly typed interface around the Cargo build script protocol. Cargo provides inputs to the build script by environment variable and accepts commands by printing to stdout.
>
>    This crate is maintained by the Cargo team for use by the wider ecosystem. This crate follows semver compatibility for its APIs.

### DIFF
--- a/crates/proto/build.rs
+++ b/crates/proto/build.rs
@@ -1,5 +1,4 @@
-use std::env;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 use fs_err as fs;
 use miden_node_proto_build::{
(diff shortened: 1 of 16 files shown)


crates/proto/build.rs
```diff
@@ -16,12 +15,9 @@ use tonic_prost_build::FileDescriptorSet;
 
 /// Generates Rust protobuf bindings using `miden-node-proto-build`.
 fn main() -> miette::Result<()> {
-    println!("cargo::rerun-if-changed=../../proto/proto");
```
Mirko-von-Leipzig, line 19: Not actually needed because this will auto-trigger since its a crate dependency.

drahnr: I don't quite follow how this is a crate dependency, the directory only contains `*.proto` files?

Mirko-von-Leipzig: This _used_ to be a direct dependency because this crate _used_ to copy-paste the protobuf schema files from `/proto/proto` and then do codegen here.

I then refactored that to instead expose file descriptors from `/proto/` as a crate. And I guess this `rerun` got left behind.

aka any proto schema changes will trigger a rebuild of `proto`, which is build dependency of this crate (`crates/proto`), and therefore will automatically trigger this.

Mirko-von-Leipzig: Also petition to rename `proto` crates..

bobbinth: What would we rename them to?

PR #1718, opened 2026-02-27
chore: simplify grpc file descriptor codegen

Mirko-von-Leipzig:
This is an experiment using the [`codegen`](https://docs.rs/codegen/latest/codegen/) crate.

Its a more type-safe way of generating rust code.

### Other changes

We iterate through the gRPC service files, instead of hard-coding them. I'm not married to this; it was just easier to do at the time.

I've also merged the three store file descriptors into a single one. This is a remnant of when these three were in separate files, but since they got merged, the three were identical.

> [!NOTE]
> We don't need to merge this; I'm experimenting with gRPC codegen using this crate.


proto/src/lib.rs
```diff
@@ -1,57 +1 @@
...
-        .expect("bytes should be a valid file descriptor created by build.rs")
-}
-
-/// Returns the Protobuf file descriptor for the validator API.
-pub fn validator_api_descriptor() -> FileDescriptorSet {
-    let bytes = include_bytes!(concat!(env!("OUT_DIR"), "/", "validator_file_descriptor.bin"));
-    FileDescriptorSet::decode(&bytes[..])
-        .expect("bytes should be a valid file descriptor created by build.rs")
-}
+include!(concat!(env!("OUT_DIR"), "/file_descriptors.rs"));
```
drahnr, line 1: comment: it's unfortunate `build-rs` doesn't provide this as well

Mirko-von-Leipzig:
Superseded by #1946

PR #1721, opened 2026-02-27
chore: extract misc changes from #1567

drahnr:
Chore changes from https://github.com/0xMiden/node/pull/1567

* makes `fn apply_vault_delta` failable rather than asserting
* fixes two doclinks
* removes `is_full_state: bool` from getter APIs and where it can be asserted/assumed/is an invariant

### Intent

Reduce the size of #1567 to make it easier to review

### DIFF
--- a/crates/store/src/inner_forest/mod.rs
+++ b/crates/store/src/inner_forest/mod.rs
@@ -13,7 +13,7 @@ use miden_protocol::asset::{Asset, AssetVaultKey, AssetWitness, FungibleAsset};
 use miden_protocol::block::BlockNumber;
 use miden_protocol::crypto::merkle::smt::{SMT_DEPTH, SmtForest};
 use miden_protocol::crypto::merkle::{EmptySubtreeRoots, MerkleError};
-use miden_protocol::errors::{AssetError, StorageMapError};
+use miden_protocol::errors::{AccountError, AssetError, StorageMapError};
 use miden_protocol::{EMPTY_WORD, Word};
 use thiserror::Error;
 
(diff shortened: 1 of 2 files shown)


crates/store/src/inner_forest/mod.rs
```diff
@@ -25,6 +25,12 @@ mod tests;
 
 #[derive(Debug, Error)]
 pub enum InnerForestError {
+    #[error(transparent)]
+    Account(#[from] AccountError),
+    #[error(transparent)]
+    Asset(#[from] AssetError),
+    #[error(transparent)]
+    Merkle(#[from] MerkleError),
```
Mirko-von-Leipzig, line 33: Could you `#[source]` these for call site clarity

crates/store/src/inner_forest/mod.rs
```diff
@@ -268,8 +282,20 @@ impl InnerForest {
         let account_id = delta.id();
         let is_full_state = delta.is_full_state();
 
+        #[cfg(debug_assertions)]
```
Mirko-von-Leipzig, line 285: Is this an expensive check? If not, maybe we should enable it fully?

drahnr: It's not particularly expensive, but it's still a full iteration over all storage maps and entries - for large accounts this could take a bit extra, but I don't think it's excessive.

crates/store/src/inner_forest/mod.rs
drahnr: We should use `InnerForestError(Box<dyn std::error::Error + Send + Sync + 'static>)` but I'd hold off on it until all my PRs are done, similarly for adding `map_err(ForestInnerError::$variant)` conversions in quite a few places.

PR #1722, opened 2026-02-27
Rename `NoteRoot` message into `NoteScriptRoot`

bobbinth:
This PR fixes an incorrectly named protobuf message. Specifically, it renames `NoteRoot` into `NoteScriptRoot` (there is not such thing as `NoteRoot`).


proto/proto/rpc.proto
Mirko-von-Leipzig, line 1: I would add a changelog entry since this affects the public gRPC API.

bobbinth: Done!

PR #1735, opened 2026-03-03
chore: Refactor validator tx schema

### LINKED ISSUE #1715: Split Validator DB data
Storing full executed transaction will quickly add up as these contain proofs and each proof could be close to 100KB. We should change this to store just the relevant info. I'd probably split this up to store:

- account delta
- input note info (need to define what exactly gets stored)
- output note info (need to define what exactly gets stored)
- maybe some other metadata (e.g., initial account state hash, final account state hash, transaction fee).

_Originally posted by @bobbinth in https://github.com/0xMiden/node/pull/1614#discussion_r2861550380_

sergerad:
The validator currently stores overly large data in the DB that is not necessary.

Closes #1715.

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -1,4 +1,4 @@
-name: Publish (main)
+name: Publish crates
 
 permissions:
   contents: read
@@ -17,25 +17,13 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          ref: main
+          ref: ${{ github.event.release.tag_name }}
       - uses: ./.github/actions/install-rocksdb
       - uses: ./.github/actions/install-protobuf-compiler
-      # Ensure the release tag refers to the latest commit on main.
-      # Compare the commit SHA that triggered the workflow with the HEAD of the branch we just
-      # checked out (main).
-      - name: Verify release was triggered from main HEAD
+      - name: Log release info
         run: |
-          tag_sha="${{ github.sha }}"
-          main_sha="$(git rev-parse HEAD)"
-
-          echo "Tag points to:        $tag_sha"
-          echo "Current main HEAD is: $main_sha"
-
-          if [ "$tag_sha" != "$main_sha" ]; then
-            echo "::error::The release tag was not created from the latest commit on main. Aborting."
-            exit 1
-          fi
-          echo "Release tag matches main HEAD — continuing."
+          echo "Publishing release ${{ github.event.release.tag_name }}"
+          echo "Commit: $(git rev-parse HEAD)"
       - name: Cleanup large tools for build space
         uses: ./.github/actions/cleanup-runner
       - name: Install dependencies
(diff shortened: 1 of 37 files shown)


crates/validator/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -1,8 +1,13 @@
...
-    id            BLOB NOT NULL,
-    block_num     INTEGER NOT NULL,
-    account_id    BLOB NOT NULL,
-    "transaction" BLOB NOT NULL, -- Binary encoded ExecutedTransaction.
+    id                    BLOB NOT NULL,
+    block_num             INTEGER NOT NULL,
+    account_id            BLOB NOT NULL,
+    account_delta         BLOB NOT NULL,
+    input_notes           BLOB NOT NULL,
+    output_notes          BLOB NOT NULL,
```
Mirko-von-Leipzig: afaik these can all be empty

- account delta can be empty for pass through txs, and
- no input notes, and
- no output notes

Mirko-von-Leipzig: Or is that fine as not null, and it will simply be an empty vector? cc @drahnr is `Option::<Vec<u8>>::None` and `Vec<u8>::new()` equivalent in sql/diesel land?

bobbinth: Yes, all of these could be null.

Regarding what to store for notes - ideally, we should store full note data (even for private notes). For input notes, I think this should be pretty straight forward (I believe input notes for a transaction are already full notes). For output notes, this would be tricky as we actually don't have full info available in some cases. But maybe that's OK as an output note would need to become an input note at some point.

So, maybe for this PR, we just make sure we store full input notes (i.e., `Vec<InputNote>`) and w/e we get from the executed transaction as output notes (i.e., `Vec<OutputNote>`).

sergerad: >So, maybe for this PR, we just make sure we store full input notes (i.e., Vec<InputNote>) and w/e we get from the executed transaction as output notes (i.e., Vec<OutputNote>).

And do we want to store these or not?
```rust
    pub initial_account_hash: Vec<u8>,
    pub final_account_hash: Vec<u8>,
    pub fee: Vec<u8>,
```

Mirko-von-Leipzig: Yes we want to store those

crates/validator/src/tx_validation/validated_tx.rs
```diff
@@ -28,11 +35,36 @@ impl ValidatedTransaction {
...
+
+    /// Returns the notes consumed in this transaction.
+    pub fn input_notes(&self) -> &InputNotes<InputNote> {
+        self.0.input_notes()
+    }
+
+    /// Returns the notes created in this transaction.
+    pub fn output_notes(&self) -> &OutputNotes {
+        self.0.output_notes()
+    }
```
bobbinth, line 54: I would maybe have these return `Vec<InputNote>` and `Vec<OutputNote>` - but maybe these would serialize the same way.

sergerad: That would mean we do an additional clone which is unnecessary for our current use case

Issue #1726, opened 2026-03-02
Network note missing public details


Mirko-von-Leipzig:
Discovered on testnet v0.13.7, we appear to have a network note stored that has no public data. This causes the store query to panic which unwraps network note details since it expects them to exist:

https://github.com/0xMiden/node/blob/e44ea2bfef032d8da1c40379dae66c50c875d14b/crates/store/src/server/ntx_builder.rs#L111-L117

Example trace: https://ui.honeycomb.io/miden-e3/environments/testnet/datasets/gateway-node/result/EiEZW3y4gRC/trace/3KW7bVxBzkG?fields[]=s_name&fields[]=s_serviceName&span=c38bd3dd1c65cf3e&zoom=c38bd3dd1c65cf3e

When network notes changed to using note attachments, its now technically possible to construct a private note, and add the network note attachment to it. Which is potentially what caused this problem. We can fix this by _also_ checking that a note is public before declaring it a network note.

PR #1736, opened 2026-03-03
fix(store): network notes must be public [main]

Mirko-von-Leipzig:
#1726 fix on `main`.

This includes automatically fixing existing misclassified entries in the database on startup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.13.8 (TBD)
 
-- Private notes with the network note attachment are no longer incorrectly considered as network notes (#[]()).
+- Private notes with the network note attachment are no longer incorrectly considered as network notes (#[#1736](https://github.com/0xMiden/node/pull/1736)).
 
 ## v0.13.7 (2026-02-25)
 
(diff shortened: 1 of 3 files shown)


crates/rpc/src/server/api.rs
```diff
@@ -519,13 +519,11 @@ fn endpoint_limits(params: &[(&str, usize)]) -> proto::rpc::EndpointLimits {
...
-        QueryParamNoteIdLimit as NoteId,
-        QueryParamNoteTagLimit as NoteTag,
-        QueryParamNullifierLimit as Nullifier,
-        QueryParamStorageMapKeyTotalLimit as StorageMapKeyTotal,
-    };
+    use QueryParamAccountIdLimit as AccountId;
+    use QueryParamNoteIdLimit as NoteId;
+    use QueryParamNoteTagLimit as NoteTag;
+    use QueryParamNullifierLimit as Nullifier;
+    use QueryParamStorageMapKeyTotalLimit as StorageMapKeyTotal;
```
drahnr, line 526: depending on your toolchain / component version this diff flip-flops

PR #1738, opened 2026-03-03
fix(store): network notes must be public

### LINKED ISSUE #1726: Network note missing public details

Mirko-von-Leipzig:
The counterpart of #1736 for the `next` branch.

This PR updates `miden-protocol` to include https://github.com/0xMiden/protocol/pull/2365 which added network note support. This PR then replaces our adhoc (and faulty) implementation with the one from protocol.

This does leave our types and conversions in a bit of a wonky state -- things could be cleaned up further, but since we're already actively refactoring the ntx-builder I've elected to not do this.

Fixes #1726

### DIFF
--- a/.github/actions/debian/action.yml
+++ b/.github/actions/debian/action.yml
@@ -121,8 +121,8 @@ runs:
         Maintainer: Miden <devops@miden.team>
         Description: $pkg binary package
         Homepage: https://miden.xyz
-        Vcs-Git: git@github.com:0xMiden/miden-node.git
-        Vcs-Browser: https://github.com/0xMiden/miden-node
+        Vcs-Git: git@github.com:0xMiden/node.git
+        Vcs-Browser: https://github.com/0xMiden/node
         EOF
 
     - name: Build binaries
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,28 +4,28 @@
 
 ### Enhancements
 
-- [BREAKING] Move block proving from Blocker Producer to the Store ([#1579](https://github.com/0xMiden/miden-node/pull/1579)).
-- [BREAKING] Updated miden-base dependencies to use `next` branch; renamed `NoteInputs` to `NoteStorage`, `.inputs()` to `.storage()`, and database `inputs` column to `storage` ([#1595](https://github.com/0xMiden/miden-node/pull/1595)).
-- Validator now persists validated transactions ([#1614](https://github.com/0xMiden/miden-node/pull/1614)).
-- [BREAKING] Remove `SynState` and introduce `SyncChainMmr` ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
-- Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1591](https://github.com/0xMiden/miden-node/issues/1591)).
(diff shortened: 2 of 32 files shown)


crates/ntx-builder/src/db/models/queries/notes.rs
```diff
@@ -54,13 +55,16 @@ pub struct NoteInsert {
...
-    notes: &[SingleTargetNetworkNote],
+    notes: &[AccountTargetNetworkNote],
 ) -> Result<(), DatabaseError> {
     for note in notes {
         let row = NoteInsert {
-            nullifier: conversions::nullifier_to_bytes(&note.nullifier()),
-            account_id: conversions::network_account_id_to_bytes(note.account_id()),
-            note_data: conversions::single_target_note_to_bytes(note),
+            nullifier: conversions::nullifier_to_bytes(&note.as_note().nullifier()),
+            account_id: conversions::network_account_id_to_bytes(
```
SantiagoPittella, line 63: nit: there's a bit of an inconsistency on how we are casting to bytes. In https://github.com/0xMiden/node/pull/1738/changes#diff-18f92718b4914ffa99386fe38041dec4064cde98763edefa1f8750462449af70R142 we are using:
```
            nullifier: note.as_note().nullifier().to_bytes(),
            account_id: note.target_account_id().to_bytes(),
```

Mirko-von-Leipzig: I've reverted to prefer the `conversions::` functions. It would be nice to find a way to _force_ a single way of doing this. This isn't unique to this PR, or even crate, its something we'd ideally address more globally somehow.

crates/block-producer/src/mempool/subscription.rs
```diff
@@ -83,7 +83,13 @@ impl SubscriptionProvider {
...
             .output_notes()
             .filter_map(|note| match note {
-                OutputNote::Full(inner) => NetworkNote::try_from(inner.clone()).ok(),
+                // We check first to avoid cloning non-network notes.
+                OutputNote::Full(inner) => inner.is_network_note().then_some(
+                    inner
+                        .clone()
+                        .into_account_target_network_note()
+                        .expect("we just checked that this is a network note"),
+                ),
```
igamigo, line 92: @juan518munoz found an issue here where this panics. `then_some()` will imply evaluating the inner value regardless, whereas something like `then()` would do it lazily. In any case we also found that it's probably not even necessary to check for `if note.is_network_note()` because the inner conversion already implicitly checks so you can probably mtach/check for `Ok(network_note)` instead. He'll open a PR soon.

Mirko-von-Leipzig: Ah of course 🤦 Thank you!

PR #1741, opened 2026-03-03
ci(docker): use `cargo chef` and cache to github (#1631)

Mirko-von-Leipzig:
Fixes the docker CI issues on main.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.13.8 (TBD)
+
+- Private notes with the network note attachment are no longer incorrectly considered as network notes (#[#1736](https://github.com/0xMiden/node/pull/1736)).
+
 ## v0.13.7 (2026-02-25)
 
 - Updated `SyncAccountStorageMaps` and `SyncAccountVault` to allow all accounts with public state, including network accounts ([#1711](https://github.com/0xMiden/node/pull/1711)).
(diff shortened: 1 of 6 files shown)


bin/node/Dockerfile
```diff
@@ -1,39 +1,47 @@
...
 
-# Update machine & install required packages
-# The installation of sqlite3 is needed for correct function of the SQLite database
+# Base line runtime image with runtime dependencies installed.
+FROM debian:bullseye-slim AS runtime-base
 RUN apt-get update && \
     apt-get -y upgrade && \
-    apt-get install -y --no-install-recommends \
-    sqlite3 \
+    apt-get install -y --no-install-recommends sqlite3 \
```
drahnr, line 34: we should not install sqlite3 anymore, or do we need it for anything? We do build from source everytime.

bin/node/Dockerfile
```diff
@@ -1,39 +1,47 @@
...
+        clang \
+        libclang-dev \
+        cmake \
+        pkg-config \
+        libssl-dev \
+        libsqlite3-dev \
+        ca-certificates && \
     rm -rf /var/lib/apt/lists/*
-
+RUN cargo install cargo-chef
```
drahnr, line 15: note: We might want to install `cargo-binstall` if we add more cargo / rust tools

Mirko-von-Leipzig: They get cached so its not the biggest pain

bin/node/Dockerfile
Mirko-von-Leipzig: The docker file in general needs a revamp; but since this is `main` its really more just to get the CI passing

PR #1742, opened 2026-03-03
PoC: gRPC server standardization

Mirko-von-Leipzig:
Proof of concept for gRPC server implementations.

Each gRPC method gets its own trait (generated by `build.rs`):

```rust
trait <Method> {
    type Input;
    type Output;

    fn decode(request: proto::RequestType) -> tonic::Result<Self::Input>;

    fn encode(output: Self::Output) -> tonic::Result<proto::ResponseType>;

    async fn handle(&self, input: Self::Input) -> tonic::Result<Self::Output>;

    // Implementers can override the behaviour e.g. if its just a proxy shim, then we
    // can just do proxy_client.request(request).await
    async fn full(
        &self,
        request: proto::RequestType,
    ) -> tonic::Result<proto::ResponseType> {
        let input = Self::decode(request)?;
        let output = self.handle(input).await?;
        Self::encode(output)
    }
}
```

The implementation is not fully complete; I don't support the mempool subscription stream here for example. ~~The comments in the `build.rs` are also outdated and reference an initial implementation I had.~~

Some benefits of doing this, over implementing the tonic generated traits:

- We can easily add instrumentation to encode and decode and handle in one location.
- This lays the foundation for `GrpcDecode` and `GrpcEncode` traits as a follow-on, which I think will let us be more consistent with our errors here.
- This will also let us do something like `Server::public_facing` and `Server::internal`, which changes how errors are encoded (i.e. hide internal errors for public facing).
- Because each method is its own trait, we can implement one method per file.
- Standard implementation of each method.

I've implemented the validator api as an example.

Looking for feedback if this is worth pursuing and cleaning up :) (imo yes, but maybe its not worth it).


crates/proto/build.rs
```diff
@@ -50,30 +72,411 @@ fn generate_bindings(file_descriptors: FileDescriptorSet, dst_dir: &Path) -> mie
...
 
+fn rustfmt_generated(dir: &Path) -> miette::Result<()> {
+    let mut rs_files = Vec::new();
+    collect_rs_files(dir, &mut rs_files)?;
+
+    if rs_files.is_empty() {
+        return Ok(());
+    }
+
+    let status = Command::new("rustfmt")
```
drahnr, line 83: we should try to avoid calling external funbi aries, particularly when it's impossible to pass on flags (i.e. which version / edition does it expect?)

I suggest to use something like `prettyplease` for generated code https://github.com/drahnr/expander/blob/master/src/lib.rs#L239-L246

crates/validator/src/server/sign_block.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    async fn handle(&self, input: Self::Input) -> tonic::Result<Self::Output> {
+        let signature = validate_block(input, &self.signer, &self.db).await.map_err(|err| {
+            tonic::Status::invalid_argument(format!(
+                "Failed to validate block: {}",
+                err.as_report()
+            ))
+        })?;
+        Ok(signature)
+    }
```
drahnr, line 32: Minor stylistic proposal: I find it a bit awkward to have many traits, I'd rather have a single, parameterized trait that uses a concrete type as parameter:

```
impl WireCodec for MySuperDuperInteraction {
type Input = ..;
type Output = ..;
fn decode_input(grpc::input::Type) -> Result<Self::Input> {
..
}
fn encode_output(Self::Output) ->  Result<grpc::output::Type> {
..
}
}


impl GrpcInteraction<MySuperDuperInteraction> for ValidatorServer {
    async fn handle(&self, input: MySuperDuperInteraction::Input) -> tonic::Result<Self::Output> {
..
}
```

crates/proto/build.rs
```diff
@@ -50,30 +72,411 @@ fn generate_bindings(file_descriptors: FileDescriptorSet, dst_dir: &Path) -> mie
...
+        }
+    }
+    Ok(())
+}
+
 /// Generate `mod.rs` which includes all files in the folder as submodules.
 fn generate_mod_rs(dst_dir: impl AsRef<Path>) -> std::io::Result<()> {
-    let mod_filepath = dst_dir.as_ref().join("mod.rs");
+    // I couldn't find any `codegen::` function for `mod <module>;`, so we generate it manually.
+    let mut modules = Vec::new();
```
drahnr, line 112: There should be `Module::new("we_are_here").get_or_new_module("foo")`

crates/validator/src/server/sign_block.rs
SantiagoPittella: Agree, we can achieve the same separation of concerns with fewer generated types

crates/proto/build.rs
Mirko-von-Leipzig: I used `prettyplease` originally, and it gave notably worse output - and also worse dx because if you enter the generated file manually and hit save, your ide may autofmt it again.

I figured using `rustfmt` would be much better? I don't really care what version or edition it uses - its just to make it legible. It is actually possible to pass edition and other flags in; I removed it because I thought it added no value.

crates/validator/src/server/sign_block.rs
Mirko-von-Leipzig: I'm not quite understanding why you think it will result in less types? You can either use more traits, and slightly less traits and many structs and impls. The trait variant is much less code, and less indirection imo.

But maybe I've missed an approach? @SantiagoPittella I don't quite follow your proc macro suggestion -- on what do you place the proc macro?

@drahnr I actually had your suggestion implemented [here](https://github.com/0xMiden/node/blob/d9fd24757bbadc50018333ec470667d52920a448/crates/proto/src/server/mod.rs), in commit [d9fd24](https://github.com/0xMiden/node/pull/1742/changes/d9fd24757bbadc50018333ec470667d52920a448).

Mirko-von-Leipzig:
@sergerad yeah this overlaps -- but I think we can begin working on that in parallel if we want. I've given some more thought to the API I'd like -- but I don't actually know how to achieve it. I'll write that down in the issue.

crates/validator/src/server/sign_block.rs
Mirko-von-Leipzig: But I agree simplifying would be great. Create a new thread in one of the files with a sketch of the code you want to generate, and we can hash it out

Mirko-von-Leipzig:
> Also, I'm not sure how streams will be accommodated into this pipeline.

We have a couple of options; it's not much more complex than the normal method - we just need to decide what to do with the stream type.

The `tonic` adds an associated stream type for each streaming method e.g.

```rust
trait tonic::BlockProducerApi {
    type MempoolSubscriptionStream: tonic::codegen::tokio_stream::Stream<
         Item = tonic::Result<MempoolEvent>
     > + Send + Sync + 'static;

     async fn mempool_subscription(request: ...) -> tonic::Result<Self::MempoolSubscriptonStream>);
}
```

Some options:

1. Hardcode the associated type to `Pin<Box<dyn Stream>>` and let the method return any stream
2. Have the method trait also contain an additional associated stream type which gets passed along

crates/proto/build.rs
drahnr: It will error if you're using a version that doesn't know about certain idioms, i.e. `2018` will bail on `let Some(foo) = bar else {..};`

drahnr:
I am in favor of splitting the concerns of type mapping to be split from the RPC endpoint, which would allow re-use of the encoding across types. I'd propose two new traits. 

This is so boring it's hardly worth typing it out. The question to me is _how_ we want to pass the in

```
pub trait WireCodecIn {
    type GrpcInput;
    fn decode_input(input: Self::GrpcInput) -> tonic::Result<Self>;
}
pub trait WireCodecOut {
    type GrpcOutput;
    fn encode_output(output: Self) -> tonic::Result<Self::GrpcOutput>;
}

pub trait GrpcInteraction<Input: WireCodecIn, Output: WireCodecOut> {
    async fn handle(&self, input: Input) -> tonic::Result<Output>;

    async fn full(&self, request: <M as WireCodecIn>::GrpcInput) -> tonic::Result<<M as WireCodecOut>::GrpcOutput> {
        let input = <Input as WireCodecIn>::decode_input(request)?;
        let output = self.handle(input).await?;
        <Output as WireCodecOut>::encode_output(output)
    }
}

pub struct RpcEndpointChicken;

impl GrpcInteraction<FetchEggs, SomeOrNone> for RpcEndpointChicken {
    async fn handle(&self, input: Input) -> tonic::Result<Output> {
    ..
    }
}
```

What I do like about it is the modularity and the potential to opt-in to custom impls of pieces of it.

Mirko-von-Leipzig:
Superseded by #1946 and #1950

PR #1743, opened 2026-03-03
chore: prepare release `v0.14.0-alpha.1`

mmagician:
- bump workspace and dependency versions to `v0.14.0-alpha.1`
- adapt the CI job to release from other branch than just `main` (in particular, after merging this to `next`, we'll branch off to a new `release/v0.14.0-alpha` from which the alpha release will be triggered
- protocol introduced a new `StorageMapEntry` (& `StorageMapKeyHash`) wrapper around `Word`, which required a few changes to migrate to that new interface

### DIFF
--- a/scripts/check-msrv.sh
+++ b/scripts/check-msrv.sh
@@ -1,0 +1,153 @@
+#!/bin/bash
+set -e
+set -o pipefail
+
+# Enhanced MSRV checking script for workspace repository
+# Checks MSRV for each workspace member and provides helpful error messages
+
+# ---- utilities --------------------------------------------------------------
+
+check_command() {
+  if ! command -v "$1" >/dev/null 2>&1; then
+    echo "ERROR: Required command '$1' is not installed or not in PATH"
+    exit 1
+  fi
+}
+
+# Check required commands
+check_command "cargo"
+check_command "jq"
+check_command "rustup"
+check_command "sed"
+check_command "grep"
+check_command "awk"
+
+# Portable in-place sed (GNU/macOS); usage: sed_i 's/foo/bar/' file
+# shellcheck disable=SC2329  # used quoted
+sed_i() {
+  if sed --version >/dev/null 2>&1; then
+    sed -i "$@"
+  else
+    sed -i '' "$@"
+  fi
+}
+
+# ---- repo root --------------------------------------------------------------
+
+# Get the directory where this script is located and change to the parent directory
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$DIR/.."
+
+echo "Checking MSRV for workspace members..."
+
+# ---- metadata --------------------------------------------------------------
+
+metadata_json="$(cargo metadata --no-deps --format-version 1)"
+workspace_root="$(printf '%s' "$metadata_json" | jq -r '.workspace_root')"
+
+failed_packages=""
+
+# Iterate actual workspace packages with manifest paths and (maybe) rust_version
+# Fields per line (TSV): id  name  manifest_path  rust_version_or_empty
+while IFS=$'\t' read -r pkg_id package_name manifest_path rust_version; do
+  # Derive package directory (avoid external dirname for portability)
+  package_dir="${manifest_path%/*}"
+  if [[ -z "$package_dir" || "$package_dir" == "$manifest_path" ]]; then
+    package_dir="."
+  fi
+
+  echo "Checking $package_name ($pkg_id) in $package_dir"
+
+  if [[ ! -f "$package_dir/Cargo.toml" ]]; then
+    echo "WARNING: No Cargo.toml found in $package_dir, skipping..."
+    continue
+  fi
+
+  # Prefer cargo metadata's effective rust_version if present
+  current_msrv="$rust_version"
+  if [[ -z "$current_msrv" ]]; then
+    # If the crate inherits: rust-version.workspace = true
+    if grep -Eq '^\s*rust-version\.workspace\s*=\s*true\b' "$package_dir/Cargo.toml"; then
+      # Read from workspace root [workspace.package]
+      current_msrv="$(grep -Eo '^\s*rust-version\s*=\s*"[^"]+"' "$workspace_root/Cargo.toml" | head -n1 | sed -E 's/.*"([^"]+)".*/\1/')"
+      if [[ -n "$current_msrv" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-main.yml
```diff
@@ -17,25 +17,26 @@ jobs:
...
+          ref: ${{ github.event.release.target_commitish }}
       - uses: ./.github/actions/install-rocksdb
       - uses: ./.github/actions/install-protobuf-compiler
-      # Ensure the release tag refers to the latest commit on main.
+      # Ensure the release tag refers to the latest commit on the release branch.
       # Compare the commit SHA that triggered the workflow with the HEAD of the branch we just
-      # checked out (main).
-      - name: Verify release was triggered from main HEAD
+      # checked out.
+      - name: Verify release was triggered from release branch HEAD
```
Mirko-von-Leipzig: Isn't this _always_ true now? The check can probably just be yanked?

crates/proto/src/domain/digest.rs
```diff
@@ -136,6 +137,18 @@ impl From<&Word> for proto::primitives::Digest {
...
+    fn from(value: StorageMapKey) -> Self {
+        Into::<Word>::into(value).into()
+    }
+}
+
+impl From<&StorageMapKey> for proto::primitives::Digest {
+    fn from(value: &StorageMapKey) -> Self {
+        (*value).into()
+    }
+}
```
Mirko-von-Leipzig, line 150: Should we be mirroring these types in our gRPC schema instead of just using `Digest`?

Cargo.toml
Mirko-von-Leipzig, line 1: Should these be pinned to the precise version, or does cargo's semver support doing `"0.14.0-alpha.1"` which then also allows `"0v.14.0-alpha.2"` etc?

Mirko-von-Leipzig, line 1: More of a meta question, but how did we land on `alpha` as a suffix? Are we also planning on `beta` etc, or should we have done release-candidate aka `-rc.0`

crates/store/src/server/ntx_builder.rs
```diff
@@ -262,8 +262,9 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
             read_account_id::<GetWitnessesError>(request.account_id).map_err(invalid_argument)?;
 
         // Read the map key.
-        let map_key =
-            read_root::<GetWitnessesError>(request.map_key, "MapKey").map_err(invalid_argument)?;
+        let map_key = StorageMapKey::new(
+            read_root::<GetWitnessesError>(request.map_key, "MapKey").map_err(invalid_argument)?,
+        );
```
Mirko-von-Leipzig:
```suggestion
        let map_key = read_root::<GetWitnessesError>(request.map_key, "MapKey")
             .map(StorageMapKey::new)
             .map_err(invalid_argument)?;
```

crates/proto/src/domain/account.rs
```diff
@@ -548,10 +552,12 @@ impl TryFrom<proto::rpc::account_storage_details::AccountStorageMapDetails>
...
-                            let key = entry
-                                .key
-                                .ok_or(StorageMapEntry::missing_field(stringify!(key)))?
-                                .try_into()?;
+                            let key = StorageMapKey::new(
+                                entry
+                                    .key
+                                    .ok_or(StorageMapEntry::missing_field(stringify!(key)))?
+                                    .try_into()?,
+                            );
```
Mirko-von-Leipzig: Nit: prefer iteration over nesting 

```suggestion
                            let key = entry
                                    .key
                                    .ok_or(StorageMapEntry::missing_field(stringify!(key)))?
                                    .try_into()
                                    .map(StorageMapKey::new)?;
```

Cargo.toml
mmagician: I think release candidate is _expected not to have breaking changes_, but they _might_ be introduced in case something is discovered during the testing / pre-release process.

Alpha in my mind is when things are _expected to break_.

But this is just a rough guideline I've got for myself.

crates/proto/src/domain/digest.rs
mmagician: Hmm I had a comment about this somewhere, but I think it got lost in the process:
Yes, I would be in favor of introducing different proto definitions for `Word`, `StorageMapKey` etc., even if under the hood they hold the same data structure. But since we're already using `Digest` when we really mean just `Word`, this would be a rather broader refactor to make things consistent across the board.

Cargo.toml
Mirko-von-Leipzig: Sounds fair

> A **release candidate (RC)** is a build that’s *feature complete* and is being evaluated for final release. An **alpha** is an early, *feature‑incomplete* build used to validate initial functionality.
>
> Here’s a quick comparison:
>
> - **Alpha**
>   - Early stage
>   - Features still missing or unstable
>   - Lots of known bugs are expected
>   - Used for internal testing and rapid iteration
>
> - **Release Candidate (RC)**
>   - Late stage
>   - Feature complete
>   - Only critical bug fixes allowed
>   - If no blockers appear, it can become the final release
>
> In short: **alpha is early and rough**, **RC is late and almost final**.

crates/proto/src/domain/digest.rs
mmagician: Tracked in https://github.com/0xMiden/node/issues/1745

Cargo.toml
mmagician: they should be pinned to precise versions, [again](https://github.com/0xMiden/node/pull/1743#discussion_r2882769072) because there might be breaking changes between `alpha` versions

.github/workflows/publish-main.yml
mmagician: it was not strictly always true (e.g. when the release was triggered from a branch, but then a new commit was pushed to that branch before this step run).

However, while at it, I figured out that this was probably overcomplicating things too much overall. I changed the workflow to always check out the tag that the release came from (disregarding commit SHAs and branch names altogether). I think this is cleaner overall, WDYT @Mirko-von-Leipzig?

https://github.com/0xMiden/node/pull/1743/changes/7e5d1a8762b2539780e5a1ccffd8194a429d8356

---

I also [brought](https://github.com/0xMiden/node/pull/1743/commits/c5242a276cffecec103ab3a3d4dc6c496984a8f6) in the MSVR script which was only on `main`

mmagician: We can always iterate on this if needed

PR #1748, opened 2026-03-04
feat: add self-healing in network monitor

### LINKED ISSUE #1744: Self-Healing mechanism for network monitor
When the `miden-node` restarts, the in-memory mempool is cleared and the commitment in the network monitor account becomes invalid. We can add a mechanism to refresh the account state syncing from the node if we receive this error.

SantiagoPittella:
closes #1744 

When the node restarts and clears its mempool, the monitor's local wallet drifts from the on-chain state, causing every subsequent transaction to fail with a commitment mismatch. After 3 consecutive failures, the monitor now re-fetches the wallet account from the RPC and updates the local state, allowing it to self-heal.

I tested this by stoping the local node right after the local transaction is sent to the node, and managed to get commitment mismatch errors. After 3 tries the account states was refetched and the healing worked properly.


bin/network-monitor/src/counter.rs
```diff
@@ -433,7 +439,21 @@ pub async fn run_increment_task(
...
+                        &mut rpc_client,
+                        &mut wallet_account,
+                        &mut data_store,
+                    )
+                    .await
+                    .is_ok()
+                    {
+                        consecutive_failures = 0;
+                    }
+                }
```
igamigo, line 456: Should this reset `consecutive_failures` anyway (ie, even if the resync fails)?

bin/network-monitor/src/counter.rs
```diff
@@ -433,7 +439,21 @@ pub async fn run_increment_task(
...
+                consecutive_failures += 1;
                 last_error = Some(handle_increment_failure(&mut details, &e));
+
+                if consecutive_failures >= RESYNC_FAILURE_THRESHOLD {
+                    if try_resync_wallet_account(
+                        &mut rpc_client,
+                        &mut wallet_account,
+                        &mut data_store,
+                    )
+                    .await
```
igamigo, line 451: One thing to keep in mind is that this fix really would only apply for when you get the "commitment mismatch" error where the mempool or blockchain state reflects a different account commitment, so maybe it can be narrowed down to that case so that it doesn't mask other behaviors. But maybe there is an argument for doing it regardless, not sure.

igamigo (approved):
LGTM!

bin/network-monitor/src/counter.rs
SantiagoPittella: If the re-sync fails then the account keeps the mismatch state with the node. If it was a transient error then the next iteration trying to increase the counter will succeed and the consecutive_failures will be reseted anyways. If the error persist, then we want to retry syncing anyways

bin/network-monitor/src/counter.rs
SantiagoPittella: The threshold of 3 already guards against transient issues, and resync is pretty cheap in our case. I cannot think of another kind of issue other than the node failing and causing the state to mismatch after that will be fixed "silently" by re-fetching the state.

bin/network-monitor/src/counter.rs
```diff
@@ -478,6 +498,37 @@ fn handle_increment_success(
...
+#[instrument(
+    parent = None,
+    target = COMPONENT,
+    name = "network_monitor.counter.try_resync_wallet_account",
+    skip_all,
+    fields(account.id = %wallet_account.id()),
+    level = "warn",
+    err,
+)]
+async fn try_resync_wallet_account(
```
Mirko-von-Leipzig, line 511: Nit 
```suggestion
async fn resync_wallet_account(
```

PR #1750, opened 2026-03-04
feat: add rate limiter

drahnr:
Adds a rate-limiter with quotas using `tower_governor` and `tower::limit::GlobalConcurrencyLimitLayer`.

 Done in scope of our recent fd consumption, generally a useful thing to do.
 
 Makes all parameters configurable and group them into `GrpcOptions` and moves them to `miden-node-utils` to avoid duplication, similarly for limiting layer setup helpers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 
 ### Fixes
 
+- Fixed network monitor looping on stale wallet nonce after node restarts by re-syncing wallet state from RPC after repeated failures ([#1748](https://github.com/0xMiden/node/pull/1748)).
 - Fixed `bundled start` panicking due to duplicate `data_directory` clap argument name between `BundledCommand::Start` and `NtxBuilderConfig` ([#1732](https://github.com/0xMiden/node/pull/1732)).
 - Fixed `bundled bootstrap` requiring `--validator.key.hex` or `--validator.key.kms-id` despite a default key being configured ([#1732](https://github.com/0xMiden/node/pull/1732)).
 - Fixed incorrectly classifying private notes with the network attachment as network notes ([#1378](https://github.com/0xMiden/node/pull/1738)).
@@ -82,6 +83,7 @@
 ### Enhancements
 
 - Cleanup old account data from the database on apply block ([#1304](https://github.com/0xMiden/miden-node/issues/1304)).
+- Added cleanup of old account data from the in-memory forest ([#1175](https://github.com/0xMiden/miden-node/issues/1175))
 - Added block validation endpoint to validator and integrated with block producer ([#1382](https://github.com/0xMiden/miden-node/pull/1381)).
 - Added support for timeouts in the WASM remote prover clients ([#1383](https://github.com/0xMiden/miden-node/pull/1383)).
 - Added mempool statistics to the block producer status in the `miden-network-monitor` binary ([#1392](https://github.com/0xMiden/miden-node/pull/1392)).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2836,9 +2836,9 @@ dependencies = [
 
 [[package]]
 name = "miden-crypto-derive"
-version = "0.19.4"
+version = "0.19.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f40e95b9c7c99ed6bbf073d9e02721d812dedd2c195019c0a0e0a3dbb9cbf034"
+checksum = "3550b5656b791fec59c0b6089b4d0368db746a34749ccd47e59afb01aa877e9e"
 dependencies = [
  "quote",
  "syn 2.0.114",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,7 +107,7 @@ rand_chacha        = { default-features = false, version = "0.9" }
 reqwest            = { version = "0.13" }
 rstest             = { version = "0.26" }
 serde              = { features = ["derive"], version = "1" }
-tempfile           = { version = "3" }
+tempfile           = { version = "3.12" }
 thiserror          = { default-features = false, version = "2.0" }
 tokio              = { features = ["rt-multi-thread"], version = "1.46" }
 tokio-stream       = { version = "0.1" }
--- a/bin/network-monitor/src/counter.rs
+++ b/bin/network-monitor/src/counter.rs
@@ -41,6 +41,9 @@ use rand_chacha::ChaCha20Rng;
 use tokio::sync::{Mutex, watch};
 use tracing::{error, info, instrument, warn};
 
+/// Number of consecutive increment failures before re-syncing the wallet account from the RPC.
+const RESYNC_FAILURE_THRESHOLD: usize = 3;
+
 use crate::COMPONENT;
 use crate::config::MonitorConfig;
 use crate::deploy::counter::COUNTER_SLOT_NAME;
@@ -395,6 +398,7 @@ pub async fn run_increment_task(
 
     let mut rng = ChaCha20Rng::from_os_rng();
     let mut interval = tokio::time::interval(config.counter_increment_interval);
+    let mut consecutive_failures: usize = 0;
 
     loop {
         interval.tick().await;
@@ -414,6 +418,8 @@ pub async fn run_increment_task(
         .await
         {
             Ok((tx_id, final_account, block_height)) => {
+                consecutive_failures = 0;
+
                 let target_value = handle_increment_success(
                     &mut wallet_account,
                     &final_account,
@@ -433,7 +439,21 @@ pub async fn run_increment_task(
                 }
             },
             Err(e) => {
+                consecutive_failures += 1;
                 last_error = Some(handle_increment_failure(&mut details, &e));
+
+                if consecutive_failures >= RESYNC_FAILURE_THRESHOLD {
+                    if try_resync_wallet_account(
+                        &mut rpc_client,
+                        &mut wallet_account,
+                        &mut data_store,
+                    )
+                    .await
+                    .is_ok()
+                    {
+                        consecutive_failures = 0;
+                    }
+                }
             },
         }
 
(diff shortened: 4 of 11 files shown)


crates/rpc/src/server/mod.rs
```diff
@@ -78,12 +79,32 @@ impl Rpc {
...
+            tower_governor::GovernorLayer::new(config)
+        };
+
         tonic::transport::Server::builder()
             .accept_http1(true)
             .timeout(self.grpc_timeout)
             .layer(CatchPanicLayer::custom(catch_panic_layer_fn))
             .layer(TraceLayer::new_for_grpc().make_span_with(grpc_trace_fn))
             .layer(HealthCheckLayer)
+            .layer(rate_limiter)
```
Mirko-von-Leipzig: Do we run into [this](https://docs.rs/tower_governor/latest/tower_governor/#common-pitfalls)

> Be careful to create your server with [`.into_make_service_with_connect_info::<SocketAddr>`](https://docs.rs/axum/latest/axum/struct.Router.html#method.into_make_service_with_connect_info) instead of `.into_make_service()` if you are using the default PeerIpKeyExtractor. Otherwise there will be no peer ip address for Tower to find!

crates/rpc/src/server/mod.rs
```diff
@@ -78,12 +79,32 @@ impl Rpc {
         let rpc_version =
             semver::Version::parse(rpc_version).context("failed to parse crate version")?;
 
+        let rate_limiter = {
+            let config = tower_governor::governor::GovernorConfigBuilder::default()
+                .per_second(10)
+                .burst_size(20)
```
Mirko-von-Leipzig: iiuc each IP can have 20 requests, and every 10s a permit is replenished?

I'm also unsure what reasonable defaults would be. imo the default should be _very_ lenient so that local node users don't notice. And then in production we must actually set them -- and here we'll likely have to experiment?

crates/rpc/src/server/mod.rs
Mirko-von-Leipzig, line 1: This still leaves us with an unbound total connections correct? Or is that limited by default already?

drahnr: `tower` doesn't provide a connection limit in itself, at least I couldn't find one.

drahnr: It's not limited, I added another limiter for the total global concurrent connections

crates/rpc/src/server/mod.rs
drahnr: Yes, it's replenish "API tokens" after we received the burst number of requests, I made this configurable with what I consider sane defaults (they're per IP)

crates/rpc/src/server/mod.rs
drahnr: Yes and no, we need the connection info, but since we're behind a proxy we should have the `X-Real-IP` anyways, which is what `SmartIpKey..` uses as one of it's first sources to extract the key.

crates/block-producer/src/server/tests.rs
```diff
@@ -130,6 +134,8 @@ async fn start_store(
     data_directory: &std::path::Path,
 ) -> runtime::Runtime {
     let genesis_state = GenesisState::new(vec![], test_fee_params(), 1, 1, random_secret_key());
+    fs::create_dir_all(data_directory.join(ACCOUNT_TREE_STORAGE_DIR))
+        .expect("account tree directory should be created");
```
sergerad: Is this diff related and intentional?

crates/utils/src/grpc/connect_info.rs
```diff
@@ -0,0 +1,19 @@
...
+
+impl Interceptor for ConnectInfoInterceptor {
+    fn call(&mut self, mut request: Request<()>) -> Result<Request<()>, Status> {
+        let addr = request
+            .remote_addr()
+            .ok_or_else(|| Status::failed_precondition("Expected TCP connection"))?;
+        // TODO double check how to address proxy rate limiting based on i.e. `X-Real-IP`.
+        request
+            .metadata_mut()
+            .insert("forwarded", format!("for={addr}").try_into().unwrap());
```
sergerad, line 15: we shouldn't panic here right? this is a callback for every request?

crates/utils/src/grpc/layers.rs
```diff
@@ -0,0 +1,66 @@
...
+    tokio::spawn(async move {
+        let mut interval = tokio::time::interval(Duration::from_secs(60));
+        loop {
+            interval.tick().await;
+            // avoid a DoS vector
+            limiter.retain_recent();
+        }
+    });
+    Ok(tower_governor::GovernorLayer::new(config))
+}
```
sergerad, line 63: Just checking this only actually gets called once (`rate_limit_per_ip()`)

crates/block-producer/src/server/tests.rs
drahnr: I don't think it's required, let me remove it

crates/utils/src/grpc/connect_info.rs
drahnr: It is a callback per request, unless someone manages to inject a broken/malicious IP in the Linux Kernel and manage to pop it up in our end, it's virtually impossible to break this invariant.

Since we're using `SmartIPKeyExtractor*` we might get away without this. It's here for ensuring we have a fallback

crates/utils/src/grpc/layers.rs
drahnr: I get 7 hits in the diff? It creates a layer so this is fine, unless I am missing smth?

crates/store/src/server/mod.rs
Mirko-von-Leipzig, line 1: I don't think we should be applying these to all servers? Only the RPC server or no?

The rest are internal; and we should be bounding the calls we make. e.g. we don't want to disconnect our own gRPC connections after some time..

crates/utils/src/grpc/connect_info.rs
```diff
@@ -0,0 +1,19 @@
...
+// Extracts the IP for `Governor`
+#[derive(Debug, Clone)]
+pub struct ConnectInfoInterceptor;
+
+impl Interceptor for ConnectInfoInterceptor {
+    fn call(&mut self, mut request: Request<()>) -> Result<Request<()>, Status> {
+        let addr = request
+            .remote_addr()
+            .ok_or_else(|| Status::failed_precondition("Expected TCP connection"))?;
+        // TODO double check how to address proxy rate limiting based on i.e. `X-Real-IP`.
```
Mirko-von-Leipzig: Was this addressed?

crates/utils/src/grpc/layers.rs
```diff
@@ -0,0 +1,66 @@
...
+) -> anyhow::Result<
+    tower_governor::GovernorLayer<
+        SmartIpKeyExtractor,
+        StateInformationMiddleware,
+        tonic::body::Body,
+    >,
+> {
+    let config = GovernorConfigBuilder::default()
+        .key_extractor(SmartIpKeyExtractor)
+        .per_second(grpc_options.replenish_per_sec)
```
Mirko-von-Leipzig: I think we're using this invertedly. `.per_second` specifies the number of _seconds_ to wait until _a single_ permit is replenished. And not the number of permits to replenish each second.

crates/utils/src/grpc/layers.rs
```diff
@@ -0,0 +1,66 @@
...
+        .burst_size(grpc_options.burst_size as u32)
```
Mirko-von-Leipzig: Should we make `burst_size` a `u32` then?

crates/utils/src/grpc/layers.rs
```diff
@@ -0,0 +1,66 @@
...
+        .use_headers()
+        .finish()
+        .context("config parameters are inconsistent, i.e. burst < per second")?;
```
Mirko-von-Leipzig: I don't think `burst < per second` is an error condition. Only that they're non-zero. Which we can enforce with types.

crates/utils/src/grpc/layers.rs
```diff
@@ -0,0 +1,66 @@
...
+    let limiter = std::sync::Arc::clone(config.limiter());
+    tokio::spawn(async move {
+        let mut interval = tokio::time::interval(Duration::from_secs(60));
+        loop {
+            interval.tick().await;
+            // avoid a DoS vector
+            limiter.retain_recent();
```
Mirko-von-Leipzig, line 59: What does this comment mean?

crates/utils/src/grpc/connect_info.rs
drahnr: This is addressed with the `SmartKeyExtractor*` which tests for 3 distinct metadatadata entries before using the raw incoming IP

crates/utils/src/grpc/layers.rs
drahnr: You are correct.

crates/utils/src/grpc/layers.rs
drahnr: If we don't call retain, using a lot of IP addresses can move the service into an OOM by using a large number of source IP addresses

crates/utils/src/grpc/layers.rs
```diff
@@ -46,13 +44,21 @@ pub fn rate_limit_per_ip(
...
     >,
 > {
+    ensure!(
+        grpc_options.replenish_per_sec > 0,
+        "grpc.replenish_per_sec must be greater than zero"
+    );
+    ensure!(
+        grpc_options.burst_size > 0,
+        "grpc.burst_size must be greater than zero"
+    );
```
Mirko-von-Leipzig, line 45: Lets use the `NonZero` variants?

crates/utils/src/clap.rs
```diff
@@ -0,0 +1,155 @@
...
+impl GrpcOptionsExternal {
+    pub fn test() -> Self {
+        Self {
+            request_timeout: TEST_REQUEST_TIMEOUT,
+            max_connection_age: DEFAULT_MAX_CONNECTION_AGE,
+            burst_size: DEFAULT_BURST_SIZE,
+            replenish_n_per_second_per_ip: DEFAULT_REPLENISH_N_PER_SECOND_PER_IP,
+            max_global_concurrent_connections: DEFAULT_MAX_GLOBAL_CONNECTIONS,
+        }
+    }
```
Mirko-von-Leipzig, line 122:
```suggestion
    pub fn test() -> Self {
        Self {
            request_timeout: TEST_REQUEST_TIMEOUT,
            ..Default::default()
        }
    }
```

crates/validator/src/server/mod.rs
```diff
@@ -39,7 +39,7 @@ pub struct Validator {
     /// Server-side timeout for an individual gRPC request.
     ///
     /// If the handler takes longer than this duration, the server cancels the call.
-    pub grpc_timeout: Duration,
+    pub grpc_options: GrpcOptionsInternal,
```
Mirko-von-Leipzig, line 42: Doc comment is inaccurate now. We also ignore the max connections setting then?

crates/utils/src/clap.rs
```diff
@@ -0,0 +1,155 @@
...
+    )]
+    pub request_timeout: Duration,
+
+    /// Number of global concurrent connections.
+    #[arg(
+        long = "grpc.max_global_connections",
+        default_value_t = DEFAULT_MAX_GLOBAL_CONNECTIONS,
+        value_name = "MAX_GLOBAL_CONNECTIONS"
+    )]
+    pub max_global_concurrent_connections: u64,
```
Mirko-von-Leipzig: I think `global` is confusing -- this is just the maximum number of connections?

crates/utils/src/grpc/connect_info.rs
```diff
@@ -0,0 +1,18 @@
+use tonic::service::Interceptor;
+use tonic::{Request, Status};
+
+// Extracts the IP for `Governor`
```
Mirko-von-Leipzig: Its unclear in this context what a `Governor` is.

crates/store/src/server/mod.rs
```diff
@@ -160,12 +159,15 @@ impl Store {
             }
         });
 
+        let concurrency_semaphore = grpc::concurrency_semaphore(self.grpc_options);
```
Mirko-von-Leipzig: I think this is incorrect. We don't want to share concurrency between these because it means external RPC requests can DoS our internal requests. I still don't think we want to apply any limits internally at all.

crates/utils/src/clap.rs
drahnr: Rename to `_concurrent_` if that's any better, open to suggestions

crates/utils/src/grpc/connect_info.rs
drahnr: Extended doc comment

crates/store/src/server/mod.rs
drahnr: Removed, only limiting the RPC component now

PR #1751, opened 2026-03-05
sanity check VaultAssetWitnessesRequest vault key len

### LINKED ISSUE #1565: Sanity check or limit vault key count coming via VaultAssetWitnessesRequest
As per [comment](https://github.com/0xMiden/miden-node/pull/1529#discussion_r2713606190):

Consider adding a sanity check or enforce an upper limit to number of vault keys processed based on `VaultAssetWitnessesRequest` requests.

sergerad:
Closes #1565

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1591](https://github.com/0xMiden/node/issues/1591)).
 - Fixed `TransactionHeader` serialization for row insertion on database & fixed transaction cursor on retrievals ([#1701](https://github.com/0xMiden/node/issues/1701)).
 - Added KMS signing support in validator ([#1677](https://github.com/0xMiden/node/pull/1677)).
+- Added finality field for `SyncChainMmr` requests ([#1725](https://github.com/0xMiden/miden-node/pull/1725)).
 - Added per-IP gRPC rate limiting across services as well as global concurrent connection limit ([#1746](https://github.com/0xMiden/node/issues/1746)).
 - Added limit to execution cycles for a transaction network, configurable through CLI args (`--ntx-builder.max-tx-cycles`) ([#1801](https://github.com/0xMiden/node/issues/1801)).
 - Added monitor version and network name to the network monitor dashboard, network name is configurable via `--network-name` / `MIDEN_MONITOR_NETWORK_NAME` ([#1838](https://github.com/0xMiden/node/pull/1838)).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3323,7 +3323,6 @@ dependencies = [
  "diesel",
  "diesel_migrations",
  "fs-err",
- "futures",
  "hex",
  "indexmap",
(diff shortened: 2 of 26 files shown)


crates/store/src/server/ntx_builder.rs
```diff
@@ -203,8 +203,17 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
...
+        const MAX_VAULT_KEYS: usize = 1000;
+
         let request = request.into_inner();
 
+        // Sanity check the number of vault keys in the request
+        if request.vault_keys.len() > MAX_VAULT_KEYS {
+            return Err(Status::invalid_argument(format!(
+                "number of vault keys in request cannot exceed {MAX_VAULT_KEYS}"
+            )));
+        }
```
sergerad, line 223: @bobbinth is this where we would want to put the sanity check?

crates/store/src/server/ntx_builder.rs
```diff
@@ -203,8 +203,17 @@ impl ntx_builder_server::NtxBuilder for StoreApi {
         &self,
         request: Request<proto::store::VaultAssetWitnessesRequest>,
     ) -> Result<Response<proto::store::VaultAssetWitnessesResponse>, Status> {
+        const MAX_VAULT_KEYS: usize = 1000;
```
Mirko-von-Leipzig: Maybe a question for @partylikeits1983 or @mmagician - what is a reasonable limit to set for network transactions? This would mean any network tx which requires more than this amount of witnesses would never succeed

crates/store/src/server/ntx_builder.rs
bobbinth: Yes, something like this.

Regarding a reasonable limit - most of the time, the number of requests should be in single digits. In some cases, when a transaction is consuming lots of notes, and each of these notes has many different assets, it could be higher - but I think we could set the limit at something like 100 and see if it ever gets tripped.

I think the important thing here would, in case the limit does get tripped, to make sure we clearly know that it's happening (via logs or otherwise).

Mirko-von-Leipzig: @sergerad I think ensure this gets logged, lets add a `warn!` for this case?

We'll need to move this check down so we can include the account ID.

```rust
tracing::warn!(
    limit=%MAX_VAULT_KEYS,
    request=%request.vault_keys.len(),
    account.id=%account_id,
    "maximum vault key limit exceeded",
);
```

sergerad: @Mirko-von-Leipzig why would we not see an error log for this in the ntx-builder stack?

Mirko-von-Leipzig: We would, but we wouldn't be able to trivially trigger an alert for specifically this error because it would be a general gRPC error and we would have to do a stringly match.

This is redundant, but since we want to _definitely_ know that specifically this occurred, I thought its prudent to do that here so we cannot misconfigure or miss it.

bobbinth: I added the warning in 0942ca91e90138f8616bea67bff62811e661364f.

PR #1752, opened 2026-03-05
fix: network accounts should be loaded at genesis

mmagician:
simple as per title

We'll need to create another release

### DIFF
--- a/crates/store/src/genesis/mod.rs
+++ b/crates/store/src/genesis/mod.rs
@@ -73,10 +73,10 @@ impl<S: BlockSigner> GenesisState<S> {
             .accounts
             .iter()
             .map(|account| {
-                let account_update_details = if !account.id().is_private() {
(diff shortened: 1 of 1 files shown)


crates/store/src/genesis/config/tests.rs
```diff
@@ -354,7 +355,30 @@ async fn parsing_agglayer_sample_with_account_files() -> TestResult {
     assert_eq!(secrets.secrets.len(), 1, "Only native faucet should generate a secret");
 
     // Verify the genesis state can be converted to a block
-    let _block = state.into_block().await?;
+    let block = state.into_block().await?;
+
+    // Verify that non-private accounts (Public and Network) get full Delta details,
+    // not Private. This is the key check: Network accounts like the AggLayer bridge
+    // and faucets must have their full details stored so the ntx-builder can find them.
```
mmagician:
```suggestion
    // Verify that non-private accounts (Public and Network) get full Delta details.
```

PR #1754, opened 2026-03-05
chore: Refactor ConversionError

### LINKED ISSUE #1528: chore: Split `ConversionError` enum
As per [comment](https://github.com/0xMiden/miden-node/pull/1521#discussion_r2702702905):
>split this error up somehow into more logically grouped variants (right now, it mixes account-related, note-related, SMT-related etc. errors together).

Should be done after #1521 is merged.

sergerad:
Resolves #1528

Refactors `ConversionError` from a public enum to an opaque struct, and introduces layered abstractions to progressively eliminate proto-to-domain conversion boilerplate.

### `ConversionError` struct

Converts `ConversionError` to an opaque struct with a field path stack and a boxed error source. The `Display` impl renders dotted paths like `"header.account_root: value is not in range 0..MODULUS"`.

Adds `ConversionResultExt` for ergonomic `.context("field_name")` chaining on `Result<T, ConversionError>`.

### `DecodeBytesExt`

Extension trait on `Deserializable` types to replace the repeated pattern:
```rust
T::read_from_bytes(&bytes).map_err(|source| ConversionError::deserialization("T", source))
```
with:
```rust
T::decode_bytes(&bytes, "T")
```

### `GrpcStructDecoder` + `GrpcDecodeExt`

Replaces the `TryConvertFieldExt` trait (which required specifying the parent proto type on every field) with a serde-style struct decoder. The parent type is specified once via `.decoder()`, then each field is decoded with `decoder.decode_field("name", value.field)`:

```rust
let decoder = value.decoder();
let field = decoder.decode_field("field", value.field)?;
```

`decode_field` handles `Option` unwrapping, `TryInto` conversion, and field path context in a single call.

### Other changes

- Fixed incorrect `missing_field` type parameters throughout
- Made `read_account_id` generic over parent proto message types
- Replaced `stringify!` macros with string literals
- Re-exported `prost` from `miden-node-proto`

### DIFF
--- a/crates/store/src/server/api.rs
+++ b/crates/store/src/server/api.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 
 use miden_node_proto::decode::{ConversionResultExt, GrpcStructDecoder};
 use miden_node_proto::errors::ConversionError;
-use miden_node_proto::generated as proto;
+use miden_node_proto::{decode, generated as proto};
 use miden_node_utils::ErrorReport;
 use miden_protocol::Word;
 use miden_protocol::account::AccountId;
@@ -162,14 +162,13 @@ where
 }
 
 pub fn read_account_id<M: miden_node_proto::prost::Message, E>(
-    id: Option<proto::account::AccountId>,
+    account_id: Option<proto::account::AccountId>,
 ) -> Result<AccountId, E>
 where
     E: From<ConversionError>,
 {
-    GrpcStructDecoder::<M>::default()
-        .decode_field("account_id", id)
-        .map_err(|e: ConversionError| e.into())
+    let decoder = GrpcStructDecoder::<M>::default();
+    decode!(decoder, account_id).map_err(|e: ConversionError| e.into())
 }
 
 #[instrument(


crates/proto/src/domain/block.rs
Mirko-von-Leipzig, line 1: Top level overview; attaching to an arbitrary file so we can discuss inline.

I think this gives the correct end result, but I worry about the ergonomics of it. Having to specify the field, type and operation is a problem and will be error prone. I'll leaving some thoughts on possible things to explore.

At a higher level, I want to move away from `TryFrom` for something this specific. We really should use a dedicated `GrpcDecode` trait. This would be a rather large diff so I would suggest we do this in tandem with #1742. This allows us to have the current implementation, where we can then slowly move over to the #1742 piece by piece without requiring _all_ things to change at once. So as we implement a method for #1742, we also implement the `GrpcDecode` parts that we need.

crates/proto/src/domain/block.rs
```diff
@@ -75,48 +75,46 @@ impl TryFrom<proto::blockchain::BlockHeader> for BlockHeader {
...
+            .try_convert_field::<proto::blockchain::BlockHeader>("tx_commitment")?;
+        let tx_kernel_commitment = value
+            .tx_kernel_commitment
+            .try_convert_field::<proto::blockchain::BlockHeader>("tx_kernel_commitment")?;
+        let validator_key = value
+            .validator_key
+            .try_convert_field::<proto::blockchain::BlockHeader>("validator_key")?;
+        let fee_parameters = value
+            .fee_parameters
+            .try_convert_field::<proto::blockchain::BlockHeader>("fee_parameters")?;
```
Mirko-von-Leipzig: `serde` solves the redundancy of specifying the parent type (`proto::blockchain::BlockHeader`) by having dedicated sub-decoders. Something like this (I forget the exact details):

```rust
/// This gives a single place to inject the parent struct name.
struct GrpcStructDecoder<T: GrpcMessage>;

impl GrpcStructDecoder<T: GrpcMessage> {
    fn decode_field<U: GrpcDecode, V>(name: &'static str, value: U) -> Result<V> {
        value.decode().context(name)?
    }
}

let mut decoder = value.decode_struct();
let account_root = decoder.decode_field("account_root", value.account_root)?;
```

and they have similar helper structs for arrays (which inject indices), and enums etc.

This still requires manually specifying the field name which is a bummer, but I wonder if we can't write a proc-macro that injects that intelligently for us to give something along

```rust
#[GrpcDecode::struct]
impl GrpcDecode<BlockHeader> for proto::blockchain::BlockHeader {
    fn decode(self) -> Result<BlockHeader, GrpcDecodeError> {
        let prev_block_commitment = self.prev_block_commitment.decode()?;
        let chain_commitment = self.chain_commitment.decode()?;
        let account_root = self.account_root.decode()?;
        let nullifier_root = self.nullifier_root.decode()?;

        ...
        Ok(BlockHeader::new(...))
    }
}
```

crates/proto/src/domain/block.rs
```diff
@@ -335,9 +327,13 @@ impl From<&Signature> for proto::blockchain::BlockSignature {
...
     fn try_from(fee_params: proto::blockchain::FeeParameters) -> Result<Self, Self::Error> {
-        let native_asset_id = fee_params.native_asset_id.map(AccountId::try_from).ok_or(
-            proto::blockchain::FeeParameters::missing_field(stringify!(native_asset_id)),
-        )??;
+        let native_asset_id = fee_params
+            .native_asset_id
+            .map(AccountId::try_from)
+            .ok_or(ConversionError::missing_field::<proto::blockchain::FeeParameters>(
+                "native_asset_id",
+            ))?
```
Mirko-von-Leipzig, line 313: Maybe we can add a `decode_required()` which the proc macro can fill in appropriately?

crates/proto/src/domain/block.rs
```diff
@@ -149,7 +147,7 @@ impl TryFrom<proto::blockchain::BlockBody> for BlockBody {
     type Error = ConversionError;
     fn try_from(value: proto::blockchain::BlockBody) -> Result<Self, Self::Error> {
         BlockBody::read_from_bytes(&value.block_body)
-            .map_err(|source| ConversionError::deserialization_error("BlockBody", source))
+            .map_err(|source| ConversionError::deserialization("BlockBody", source))
```
Mirko-von-Leipzig: Since this occurs a lot, I wonder if we can't add our own extension trait that does this more trivially somehow. e.g. `Vec<u8>::decode_bytes` for gRPC specially.

crates/proto/src/domain/block.rs
sergerad: Latest changes use the proc macro approach with the decoder struct you also mentioned. I have kept the `TryFrom` instead of introducing another trait, however. Just because there are already a lot of changes for now.

crates/proto/src/domain/block.rs
sergerad: Have done the decoder + proc macro. But kept the TryFrom for now

crates/rpc/src/server/api.rs
```diff
@@ -412,12 +412,10 @@ impl api_server::Api for RpcService {
             ))
         })?;
 
-        // Transaction inputs must be provided in order to allow for transaction re-execution via
-        // the Validator.
+        // If transaction inputs are provided, re-execute the transaction to validate it.
```
SantiagoPittella: Don't we always want to re-execute?

crates/store/src/server/api.rs
```diff
@@ -111,8 +111,7 @@ where
     E: From<ConversionError>,
 {
     block_range.ok_or_else(|| {
-        ConversionError::MissingFieldInProtobufRepresentation { entity, field_name: "block_range" }
-            .into()
+        ConversionError::message(format!("{entity}: missing field `block_range`")).into()
```
SantiagoPittella, line 115: Any reason not to use `ConversionError::missing_field::<T>("block_range")`?

crates/store/src/server/api.rs
```diff
@@ -125,12 +124,10 @@ pub fn read_root<E>(
 where
     E: From<ConversionError>,
 {
-    root.ok_or_else(|| ConversionError::MissingFieldInProtobufRepresentation {
-        entity,
-        field_name: "root",
-    })?
-    .try_into()
-    .map_err(Into::into)
+    root.ok_or_else(|| ConversionError::message(format!("{entity}: missing field `root`")))?
```
SantiagoPittella, line 128: Same here.

crates/store/src/server/api.rs
sergerad: Because entity type is not available, only a String of its name

crates/store/src/server/api.rs
sergerad: Same as above - entity type not available

crates/rpc/src/server/api.rs
sergerad: Good catch this diff is not intended. Strange

PR #1755, opened 2026-03-06
fix: lenient pre-release matching in accept header validation

mmagician:
## Summary

Alternative to #1753. Instead of requiring an exact pre-release match, this matches only on the pre-release **label** (e.g. `alpha`) and ignores the numeric suffix. This means:

- A server at `0.14.0-alpha.3` accepts clients at `alpha.1`, `alpha.2`, etc.
- Different labels are still incompatible (`alpha` vs `beta`)
- Stable and pre-release remain incompatible
- Patch version flexibility is preserved (`0.14.0` and `0.14.1` both accepted)

This allows client and node codebases to evolve somewhat independently within the same pre-release phase - a bugfix release on one side doesn't force a re-deployment of the other.

Based on the discussion in #1753.

---

Note: the side effect is that an `alpha.3` is acceptably incompatible with `alpha.2` w.r.t. to semver rules - but it would be considered compatible according to our header validation logic, which can also be confusing.

Typically, a `beta` release indicates something more stable already, but maybe that will anyway be the case for us (generally the later throughout the release cycle we are, the more stable the release?)

### DIFF
--- a/crates/rpc/src/server/accept.rs
+++ b/crates/rpc/src/server/accept.rs
@@ -220,9 +220,7 @@ impl AcceptHeaderLayer {
                 }
                 // Pre-release versions must also match the patch version exactly
                 // (patch flexibility only applies to stable versions).
-                if self.expected_pre_label.is_some()
-                    && version.patch != self.expected_patch
-                {
+                if self.expected_pre_label.is_some() && version.patch != self.expected_patch {
                     continue;
                 }
             }


crates/rpc/src/server/accept.rs
```diff
@@ -498,4 +536,42 @@ mod tests {
...
+        }
+
+        #[rstest::rstest]
+        #[case::empty("")]
+        #[case::wildcard("*/*")]
+        #[case::media_type_only("application/vnd.miden")]
+        #[case::exact_prerelease("application/vnd.miden; version=0.14.0-alpha.3")]
+        #[case::different_patch_same_prerelease("application/vnd.miden; version=0.14.1-alpha.3")]
+        #[case::different_prerelease_number("application/vnd.miden; version=0.14.0-alpha.1")]
+        #[case::different_patch_different_number("application/vnd.miden; version=0.14.2-alpha.5")]
```
Mirko-von-Leipzig: `different_patch` should be rejected or no?

mmagician: good catch

PR #1757, opened 2026-03-09
ci: tune ci profile; add nextest ci config and sharding

huitseeker:
Improve CI latency using [nextest shards](https://nexte.st/docs/ci-features/partitioning/) (locally -30%), disable incremental compilation for CI, introduce nextest config.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,11 +116,14 @@ jobs:
         run: cargo clippy --locked --all-targets --all-features --workspace -- -D warnings
 
   tests:
+    name: Tests - shard ${{ matrix.shard }}/${{ matrix.total }}
     runs-on: ubuntu-24.04
     needs: [build]
     timeout-minutes: 30
     strategy:
       fail-fast: false
+      # Split the test suite into multiple shards so they can run in parallel.
+      # We use hash-based partitioning in the test step to balance long tests.
       matrix:
         shard: [1, 2, 3, 4]
         total: [4]


.github/workflows/ci.yml
```diff
@@ -29,6 +29,9 @@ env:
   RUST_CACHE_KEY: rust-cache-2026.02.02
   # Reduce cache usage by removing debug information.
   CARGO_PROFILE_DEV_DEBUG: 0
+  # Disable incremental compilation in CI to avoid extra work for builds
+  # that will not be incrementally reused, reducing build time and disk use.
+  CARGO_INCREMENTAL: 0
```
Mirko-von-Leipzig, line 34: This is technically already the case because we use `[Swatinem/rust-cache@v2`](https://github.com/Swatinem/rust-cache?tab=readme-ov-file#cache-details)

> In particular, the workspace crates themselves are not cached since doing so is [generally not effective](https://github.com/Swatinem/rust-cache/issues/37#issuecomment-944697938). For this reason, this action automatically sets `CARGO_INCREMENTAL=0` to disable incremental compilation, so that the Rust compiler doesn't waste time creating the additional artifacts required for incremental builds.

I don't mind setting it explicitly; just mentioning that this was already the case.

.github/workflows/ci.yml
```diff
@@ -129,9 +137,11 @@ jobs:
           prefix-key: ${{ env.RUST_CACHE_KEY }}
           save-if: false
       - name: Build tests
-        run: cargo nextest run --all-features --workspace --no-run
+        # Shared build per shard; compilation is not partitioned.
+        run: cargo nextest run --profile ci --all-features --workspace --no-run
```
Mirko-von-Leipzig, line 144: Should we re-use the build by archiving it in a separate job?

Building is technically the most expensive part afaik. Overall this might be slower but we use less resources.

.github/workflows/ci.yml
```diff
@@ -116,6 +119,11 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [build]
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3, 4]
+        total: [4]
```
Mirko-von-Leipzig, line 129: Maybe a more descriptive name and some explanation 

```suggestion
    strategy:
      fail-fast: false
      # We segment our tests into multiple shards.
      matrix:
        shard: [1, 2, 3, 4]
        total: [4]
    name: Test - shard ${{ matrix.shard }}/${{ matrix.total }}
```

.config/nextest.toml
```diff
@@ -0,0 +1,11 @@
+[profile.ci]
+# CI-specific behavior: fail fast wastes runner time on large suites.
+fail-fast = false
+# Retry once to reduce flakes without hiding persistent failures.
+retries = 1
```
Mirko-von-Leipzig, line 5: Does this still show the test failure as a warning or something?

.config/nextest.toml
```diff
@@ -0,0 +1,11 @@
...
+# Report long runners and terminate them to cap wall-clock time.
+slow-timeout = { period = "300s", terminate-after = 4 }
+# Surface failures early to reduce rerun cycles.
+failure-output = "immediate-final"
```
Mirko-von-Leipzig, line 9: What is the difference to the default?

.config/nextest.toml
huitseeker: Yes https://nexte.st/docs/features/retries/

.github/workflows/ci.yml
huitseeker: This may cut CPU usage (though not latency) for test compilation roughly 3x (build once instead of 4x), but may not improve wall-clock due to large artifact upload/download & IO.

.config/nextest.toml
huitseeker: The failure output is printed at the end of the summary, which helps when you are looking for the failure on GH's web interface.

.github/workflows/ci.yml
Mirko-von-Leipzig: I don't mind much either way :)

PR #1759, opened 2026-03-09
chore: refactor `TransactionHeader` serialization

juan518munoz:
Align serialized `TransactionHeader` to it's native counterpart, allowing backwards convertion for this type, allowing local computation of `TransactionId`.

Closes https://github.com/0xMiden/node/issues/1719


crates/proto/src/domain/note.rs
```diff
@@ -219,6 +219,35 @@ impl TryFrom<proto::note::Note> for Note {
     }
 }
 
+// NOTE HEADER
+// ================================================================================================
+
+impl From<miden_protocol::note::NoteHeader> for proto::note::NoteHeader {
```
drahnr: For consistency, we should `use miden_protocol::note::NoteHeader` / `NoteHeader` rather than using the full-path inline.

crates/proto/src/domain/transaction.rs
```diff
@@ -56,3 +58,39 @@ impl TryFrom<proto::transaction::TransactionId> for TransactionId {
...
+            .ok_or_else(|| {
+                proto::transaction::InputNoteCommitment::missing_field(stringify!(nullifier))
+            })?
+            .try_into()?;
+
+        let header: Option<miden_protocol::note::NoteHeader> =
+            value.header.map(TryInto::try_into).transpose()?;
+
+        // InputNoteCommitment has private fields, so we reconstruct it via
+        // serialization roundtrip using its Serializable/Deserializable impls.
```
drahnr, line 90: Eek, can we expose the private fields?

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -38,33 +38,32 @@ pub struct TransactionRecordRaw {
     transaction_id: Vec<u8>,
     initial_state_commitment: Vec<u8>,
     final_state_commitment: Vec<u8>,
-    nullifiers: Vec<u8>,
+    input_notes: Vec<u8>,
     output_notes: Vec<u8>,
     size_in_bytes: i64,
+    fee: Vec<u8>,
```
drahnr, line 44: Implies this is a breaking change

crates/store/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -146,9 +146,10 @@ CREATE TABLE transactions (
     block_num                    INTEGER NOT NULL, -- Block number in which the transaction was included.
     initial_state_commitment     BLOB    NOT NULL, -- State of the account before applying the transaction.
     final_state_commitment       BLOB    NOT NULL, -- State of the account after applying the transaction.
-    nullifiers                   BLOB    NOT NULL, -- Serialized vector with the Nullifier of the input notes.
-    output_notes                 BLOB    NOT NULL, -- Serialized vector with the NoteId of the output notes.
+    input_notes                  BLOB    NOT NULL, -- Serialized Vec<InputNoteCommitment> (nullifier + optional NoteHeader).
+    output_notes                 BLOB    NOT NULL, -- Serialized Vec<NoteHeader> (NoteId + NoteMetadata).
```
drahnr, line 150: Given the absence of a migration solution _yet_, this is fine

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1591](https://github.com/0xMiden/node/issues/1591)).
 - Fixed `TransactionHeader` serialization for row insertion on database & fixed transaction cursor on retrievals ([#1701](https://github.com/0xMiden/node/issues/1701)).
 - Added KMS signing support in validator ([#1677](https://github.com/0xMiden/node/pull/1677)).
+- Modified `TransactionHeader` serialization to allow converting back into the native type after serialization ([#1759](https://github.com/0xMiden/node/issues/1759)).
```
drahnr: This is breaking public API iiuc? So should carry `[BREAKING]` and be listed under `### Changes`

drahnr (approved):
Two minor things (please address pre merge!), otherwise LGTM

crates/store/src/db/models/queries/transactions.rs
```diff
@@ -124,50 +124,37 @@ impl TransactionSummaryRowInsert {
...
         // - 64 bytes for initial + final state commitments (32 bytes each)
-        // - 32 bytes per input note (nullifier size)
-        // - 500 bytes per output note (estimated size when converted to NoteSyncRecord)
-        //
-        // Note: 500 bytes per output note is an over-estimate but ensures we don't
-        // exceed memory limits when these transactions are later converted to proto records.
-        let nullifiers_size = (transaction_header.input_notes().num_notes() * 32) as usize;
-        let output_notes_size = transaction_header.output_notes().len() * 500;
-        let size_in_bytes = (HEADER_BASE_SIZE + nullifiers_size + output_notes_size) as i64;
+        // - ~64 bytes per input note (nullifier + optional NoteHeader)
```
SantiagoPittella, line 142: Is this correct?

juan518munoz: - `NoteId` = `Word` = 4 Felts = 32 bytes
- `NoteMetadata` serializes to a `Word` = 32 bytes
- Total: **64 bytes** per output note

crates/proto/src/domain/transaction.rs
juan518munoz: Seems like the most reasonable approach, but being miden-base types implies opening a PR there (and subsequently updating the miden-base version and so on).

Mirko-von-Leipzig: @PhilippGackstatter I imagine you're okay with adding a `InputNoteCommitment::from_parts_unchecked` or similar?

I'm okay with merging as is; but this is something we should fix.

crates/store/src/db/models/queries/transactions.rs
juan518munoz: I have added the `[BREAKING]` changelog entry.

crates/proto/src/domain/transaction.rs
drahnr: I am _very_ aware of the pita process to get it done :)

If you don't want to tackle it as part of this PR, file an issue, leave a `// TODO <https://gh issueurl>` inline and do a follow-up

juan518munoz: I've added the tracking issue in [ee5b0f1](https://github.com/0xMiden/node/pull/1759/commits/ee5b0f1051f8e9c11f6ce286972955cc56bf8758).

I think @SantiagoPittella (correct me if wrong) is working on updating the node with the latest from base, so can address this if this PR gets merged before that.

SantiagoPittella: Yes, i'm working on it.

PR #1760, opened 2026-03-09
chore: update validator development status

mmagician:
While getting familiar with how the validator works, I found that the components marked "in progress" have already been addressed.


Mirko-von-Leipzig:
@sergerad does that section of the readme make sense to keep?

PR #1763, opened 2026-03-09
feat: add rate limiter for [`main`]

drahnr:
> [!IMPORTANT]
> Targets `main`

Port of https://github.com/0xMiden/node/pull/1750 for `main`.
Also contains #1727 to avoid more test failures due to how we do port allocations for tests

### DIFF
--- a/crates/store/Cargo.toml
+++ b/crates/store/Cargo.toml
@@ -15,21 +15,21 @@ version.workspace      = true
 workspace = true
 
 [dependencies]
-anyhow                     = { workspace = true }
-deadpool                   = { default-features = false, features = ["managed", "rt_tokio_1"], version = "0.12" }
-deadpool-diesel            = { features = ["sqlite"], version = "0.6" }
-deadpool-sync              = { default-features = false, features = ["tracing"], version = "0.1" }
-diesel                     = { features = ["numeric", "returning_clauses_for_sqlite_3_35", "sqlite"], version = "2.3" }
-diesel_migrations          = { features = ["sqlite"], version = "2.3" }
-fs-err                     = { workspace = true }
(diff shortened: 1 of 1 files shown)


crates/block-producer/src/server/mod.rs
```diff
@@ -258,10 +261,12 @@ impl BlockProducerRpcServer {
             .context("failed to build reflection service")?;
 
         // Build the gRPC server with the API service and trace layer.
+
         tonic::transport::Server::builder()
+            .accept_http1(true)
```
drahnr, line 266: Not sure if this slipped through in the other PR too?

crates/rpc/src/tests.rs
```diff
@@ -363,14 +402,29 @@ async fn send_request(
...
+
 /// Binds a socket on an available port, runs the RPC server on it, and
 /// returns a client to talk to the server, along with the socket address.
-async fn start_rpc() -> (RpcClient, std::net::SocketAddr, std::net::SocketAddr) {
-    let store_addr = {
-        let store_listener =
-            TcpListener::bind("127.0.0.1:0").await.expect("store should bind a port");
-        store_listener.local_addr().expect("store should get a local address")
-    };
+async fn start_rpc() -> (RpcClient, std::net::SocketAddr, TcpListener) {
```
drahnr, line 419: The change to retain the port allocation rather than to allocate -> free -> attempt to re-allocate the initial one.

crates/block-producer/src/server/mod.rs
Mirko-von-Leipzig: `next` also has `accept_http1(true)` if that's what you mean?

PR #1764, opened 2026-03-09
feat: Validator bootstrap

### LINKED ISSUE #1761: Dedicated validator bootstrap command
### Motivation

Currently, `store bootstrap` creates AND signs the genesis block, requiring the validator's private key as a CLI argument. In a production deployment where the validator and store run as separate instances operated by different parties, the store operator should not need access to the validator's private key.

The validator operator should own the entire genesis process: define the genesis config, create the block, sign it, and hand off a portable file to the store operator.

### How it differs from today

Today, the store operator runs `store bootstrap` with the validator's private key to create the genesis block and immediately ingest it into the store's database and block store. This couples genesis creation (which requires the signing key) with genesis ingestion (which doesn't).

In the new flow, these are split into two independent steps. The validator produces a `genesis.dat` file containing raw `ProvenBlock` bytes - the exact same format already written to `blocks/0000/block_00000000.dat` by the current bootstrap. The store then imports that file without needing any signing key.

### Proposed Flow

1. Validator operator writes a `genesis.toml` defining faucets, wallets, and accounts.
2. Validator operator runs `miden-node validator bootstrap`, which creates accounts from the genesis config, builds and signs the genesis block header, writes `genesis.dat` (serialized `ProvenBlock`), and writes account secret files (`.mac`).
3. Validator operator hands off `genesis.dat` to the store operator.
4. Store operator runs `miden-node store bootstrap --genesis-block genesis.dat`, which reads the `ProvenBlock` from the file, validates that block_num == 0 and the signature is correct, then stores the block in `BlockStore` and inserts it into the SQLite DB. No signing key is needed.

### Changes Required

**1. Add `bootstrap` subcommand to `miden-node validator`**

New command with flags:
- `--data-directory` - where to write `genesis.dat`
- `--accounts-directory` - where to write account secret `.mac` files
- `--genesis-config-file` - optional TOML (defaults to `GenesisConfig::default()`)
- Existing validator key flags (`--validator.key.hex` / `--validator.key.kms-id`)

The handler reuses existing logic from `store.rs` (`bootstrap_accounts_and_store`): parses config, creates accounts via `GenesisConfig::into_state(signer)`, calls `into_block()` to sign, and writes the inner `ProvenBlock` bytes to `genesis.dat`.

**2. Add `TryFrom<ProvenBlock>` for `GenesisBlock`**

The store needs to validate the imported file before ingestion. `TryFrom<ProvenBlock>` for `GenesisBlock` checks:
- `block_num == 0`
- Signature verifies against the public key embedded in the header

This preserves the type-safety guarantee of `GenesisBlock` while allowing construction from a deserialized file.

**3. Refactor `Store::bootstrap()` to accept a `GenesisBlock` directly**

Current signature requires `GenesisState<S: BlockSigner>` and calls `into_block()` internally. New signature accepts a pre-built `GenesisBlock`, removing the `BlockSigner` bound and async requirement from the store's bootstrap path.

**4. Update `store bootstrap` CLI**

Replace the current flags (`--genesis-config-file`, `--accounts-directory`, validator key flags) with a single `--genesis-block` flag pointing to the file produced by the validator.

**5. Keep `bundled bootstrap` self-contained**

The bundled command continues to work as-is for local dev/testing (takes validator key + genesis config, does everything in one shot).

### Notes

- The genesis block file is raw `ProvenBlock` bytes - the same format already used by `BlockStore` for `block_00000000.dat`. No new serialization format.
- Account secret files (`.mac`) are produced by the validator and distributed separately from the genesis block file.
- The validator's own database (`validator.sqlite3`) is auto-created on first `serve()` - no changes needed there.

sergerad:
Closes #1761

Tested via the following commands locally:
```sh
function bootstrapnode() {
    # Bootstrap validator
    rm -rf /tmp/validator /tmp/accounts
    cfg="/Users/sergeradinovich/Source/miden-node/crates/store/src/genesis/config/samples/01-simple.toml"
    export AWS_REGION=REDACTED
    miden-node validator bootstrap --data-directory /tmp/validator --accounts-directory /tmp/accounts --genesis-config-file "$cfg" --validator.key.kms-id REDACTED
    # Bootstrap store
     rm -rf /tmp/store
    miden-node store bootstrap --data-directory /tmp/store --genesis-block /tmp/validator/genesis.dat
}

function startnode() {
   miden-node bundled start --rpc.url http://0.0.0.0:57291 --data-directory /tmp/store --enable-otel --validator.url http://127.0.0.1:9191
}

function startvalidator() {
    export AWS_REGION=eu-north-1
   miden-node validator start --enable-otel --data-directory /tmp/validator http://0.0.0.0:9191 --key.kms-id REDACTED
}
```

### DIFF
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -1,7 +1,8 @@
 use std::path::{Path, PathBuf};
 
 use anyhow::Context;
-use miden_node_store::{GenesisBlock, Store};
+use miden_node_store::Store;
+use miden_node_store::genesis::GenesisBlock;
 use miden_node_utils::clap::GrpcOptionsInternal;
 use miden_node_utils::fs::ensure_empty_directory;
 use miden_node_utils::grpc::UrlExt;
--- a/crates/store/src/lib.rs
+++ b/crates/store/src/lib.rs
@@ -13,7 +13,7 @@ pub use accounts::{AccountTreeWithHistory, HistoricalError, InMemoryAccountTree}
 pub use db::Db;
 pub use db::models::conv::SqlTypeConvert;
 pub use errors::DatabaseError;
-pub use genesis::{GenesisBlock, GenesisState};
+pub use genesis::GenesisState;
 pub use server::block_prover_client::BlockProver;
 pub use server::{DataDirectory, Store};
 
--- a/crates/store/src/server/mod.rs
+++ b/crates/store/src/server/mod.rs
@@ -23,8 +23,9 @@ use url::Url;
 use crate::blocks::BlockStore;
 use crate::db::Db;
 use crate::errors::ApplyBlockError;
+use crate::genesis::GenesisBlock;
 use crate::state::State;
-use crate::{BlockProver, COMPONENT, GenesisBlock};
+use crate::{BlockProver, COMPONENT};
 
 mod api;
 mod block_producer;


sergerad:
Noting that this innocuous error does occur on the first start of the bundled node:
```
2026-03-10T02:19:48.260606Z ERROR ntx.block_producer.client.subscribe_to_mempool: miden-ntx-builder: crates/ntx-builder/src/block_producer.rs:61: error: status: 'Client specified an invalid argument', self: "Mempool's chain tip 1 does not match request's 0", metadata: {"content-type": "application/grpc", "date": "Tue, 10 Mar 2026 02:19:48 GMT"}
```

bin/node/src/commands/store.rs
```diff
@@ -172,93 +147,37 @@ impl StoreCommand {
...
-            let accountpath = accounts_directory.join(name);
-            // do not override existing keys
-            fs_err::OpenOptions::new()
-                .create_new(true)
-                .write(true)
-                .open(&accountpath)
-                .context("key file already exists")?;
-            account_file.write(accountpath)?;
+/// Validates that a directory either does not exist (and creates it) or exists and is empty.
+pub fn ensure_empty_directory(directory: &Path) -> anyhow::Result<()> {
```
mmagician: It feels a little awkward for this helper to live in `store` now that it’s used by both `store` and `validator`. Since this crate doesn’t have a shared `utils.rs` yet, maybe it’s worth introducing one here?

bin/node/src/commands/validator.rs
```diff
@@ -1,25 +1,51 @@
...
 
+/// The filename used for the genesis block file.
+pub const GENESIS_BLOCK_FILENAME: &str = "genesis.dat";
+
 #[derive(clap::Subcommand)]
 pub enum ValidatorCommand {
+    /// Bootstraps the genesis block.
+    ///
+    /// Creates accounts from the genesis configuration, builds and signs the genesis block,
+    /// and writes the signed block and account secret files to disk.
```
mmagician, line 34: While I think it's fine to keep writing the secret files to disk for now (probably it would need to change eventually in prod where we'd externally provide the pubkeys for each bootstrapped account) because they're meant to stay "with us", I would provide an optional flag that allows the caller to specify the output filepath for the genesis block.

bin/node/src/commands/validator.rs
```diff
@@ -112,7 +155,83 @@ impl ValidatorCommand {
...
+            Self::Bootstrap { .. } => false,
+        }
     }
 }
+
+/// Bootstraps the genesis block: creates accounts, signs the block, and writes artifacts to disk.
+///
+/// This is extracted as a free function so it can be reused by the bundled bootstrap command.
+pub async fn bootstrap_genesis(
+    data_directory: &Path,
```
mmagician: following up on the previous comment, this could then be renamed to be more explicitly tied to the "genesis block directory" rather than the current generic "data directory"

crates/store/src/genesis/mod.rs
```diff
@@ -48,6 +48,27 @@ impl GenesisBlock {
...
+        );
+
+        anyhow::ensure!(
+            block
+                .signature()
+                .verify(block.header().commitment(), block.header().validator_key()),
+            "genesis block signature verification failed",
+        );
+
+        Ok(Self(block))
```
mmagician, line 68: This will work for any valid signature<> key pair, which means the store will accept any genesis block as valid.

I wonder if we should instead keep the `validator_key` argument on the `store bootstrap` command - but change it to only mean the **public key** of the validator (AFAIU, currently a private key is provided).
E.g. `validator_key` -> `validator_pk`?

This way, the store is sure that the genesis block comes from the canonical validator.

bin/node/src/commands/store.rs
Mirko-von-Leipzig: We have `miden_node_utils` for shared stuff (I do sometimes wish it was named `core` or `common`).

crates/store/src/genesis/mod.rs
Mirko-von-Leipzig: We'll need to think about how this _should_ work though.

afaik currently the pub key for each block isn't derived from the config, but rather its the key specified in the genesis block. As in, by definition, the node _cannot_ be configured with a key -- otherwise the node operator can determine the "valid" key.

bin/node/src/commands/validator.rs
sergerad: We can add an issue to push those files to s3 or something. Would do separately to this PR.


>I would provide an optional flag that allows the caller to specify the output filepath for the genesis block.

That is handled by this, no?
```rust
    Bootstrap {
        /// Directory in which to write the genesis block file.
        #[arg(long, env = ENV_DATA_DIRECTORY, value_name = "DIR")]
        data_directory: PathBuf,
```

crates/store/src/genesis/mod.rs
sergerad: >I wonder if we should instead keep the validator_key argument on the store bootstrap command - but change it to only mean the public key of the validator (AFAIU, currently a private key is provided).
E.g. validator_key -> validator_pk?

The store bootstrap command does not take a private key:
```
Usage: miden-node store bootstrap --data-directory <DIR> --genesis-block <FILE>

Options:
      --data-directory <DIR>  Directory in which to store the database and raw block data [env: MIDEN_NODE_DATA_DIRECTORY=]
      --genesis-block <FILE>  Path to the pre-signed genesis block file produced by the validator
  -h, --help                  Print help (see more with '--help')
```

>This will work for any valid signature<> key pair, which means the store will accept any genesis block as valid.

I think this is the point - that the node operator relies on the validator as the source of truth for this.

sergerad: @bobbinth the validator currently doesn't ensure the "canonical chain" at all. As in, the node operator could theoretically modify a historical block and the validator would still sign the subsequent ones (because the validator doesn't check the chain of commitments and signatures).

Is this something we want to add to the validator? Any of these:
- Track the chain tip
- Verify `prev_block_commitment` links to the last block it signed
- Verify `block_num` is sequential
- Verify `chain_commitment` (the MMR) is consistent
- Verify the previous block's signature

bobbinth: > @bobbinth the validator currently doesn't ensure the "canonical chain" at all. As in, the node operator could theoretically modify a historical block and the validator would still sign the subsequent ones (because the validator doesn't check the chain of commitments and signatures).
> 
> Is this something we want to add to the validator?

I think the validator should make sure the block it is signing is the valid next block in the chain (for some reason, I thought that was already the case, but if not, let's create an issue for this).

The way it could work:

- We use the validator to "bootstrap" the genesis block (I think this is what this PR is about, but I haven't reviewed the details yet). This results in the validator keeping track of the chain tip (which is at first just the genesis block).
- For any new block that the validator needs to sign, the validator:
  - Makes sure that all transactions in the block are the ones it has seen/validated before (we already do this, I believe).
  - Makes sure that given this transactions, we get the block header that we need to sign (we already do this, I believe).
  - Makes sure that the new block header is built on top of the previous block. This may also require checking consistency of account and nullifier tree updates (it may be possible to incorporate this in the above two checks, but I'm not 100% sure).
- If all looks good, the validator signs the block, and sets the new block's header as the chain tip.

sergerad: #1773

bin/node/src/commands/store.rs
```diff
@@ -172,93 +148,18 @@ impl StoreCommand {
...
-
-        // Bootstrap store.
-        Store::bootstrap(genesis_state, data_directory).await
-    }
+/// Reads a genesis block from disk, validates it, and bootstraps the store.
+pub fn bootstrap_store(data_directory: &Path, genesis_block_path: &Path) -> anyhow::Result<()> {
+    // Read and deserialize the genesis block file.
+    let bytes = fs_err::read(genesis_block_path).with_context(|| {
+        format!("failed to read genesis block from {}", genesis_block_path.display())
+    })?;
```
Mirko-von-Leipzig: `fs_err` actually already includes the file path (which is why we use it instead of `std::fs`).

```suggestion
    let bytes = fs_err::read(genesis_block_path)
      .context("failed to read genesis block")?;
```

bin/node/src/commands/validator.rs
Mirko-von-Leipzig: Is there a reason we can't just dump them to the current directory?

I don't mind having these args, but they're not really necessary imo since these are "once-off" operations. Its not a persistent data directory which must be the same on each restart.

Mirko-von-Leipzig: > While I think it's fine to keep writing the secret files to disk for now (probably it would need to change eventually in prod where we'd externally provide the pubkeys for each bootstrapped account) because they're meant to stay "with us",

@mmagician do you mean that the account secrets should be predefined somehow, and passed in? Similar to how the validator key via KMS works?

bin/node/src/commands/validator.rs
```diff
@@ -70,26 +97,43 @@ pub enum ValidatorCommand {
...
+                url,
+                grpc_options,
+                validator_key,
+                data_directory,
+                kms_key_id,
+                ..
+            } => {
+                let address = url
+                    .to_socket()
+                    .context("Failed to extract socket address from validator URL")?;
```
Mirko-von-Leipzig:
```suggestion
                    .context("failed to extract socket address from validator URL")?;
```

bin/node/src/commands/validator.rs
```diff
@@ -112,7 +156,85 @@ impl ValidatorCommand {
...
+            Self::Start { enable_otel, .. } => *enable_otel,
+            Self::Bootstrap { .. } => false,
+        }
     }
 }
+
+/// Bootstraps the genesis block: creates accounts, signs the block, and writes artifacts to disk.
+///
+/// This is extracted as a free function so it can be reused by the bundled bootstrap command.
+pub async fn bootstrap_genesis(
```
Mirko-von-Leipzig: Could technically still be a `ValidatorCommand::bootstrap` static function :)

That would be my preference; though its not a strong one. I just dislike free functions from an organisational perspective.

bin/node/src/commands/validator.rs
```diff
@@ -112,7 +156,85 @@ impl ValidatorCommand {
...
+    // Build the signed genesis block.
+    let genesis_block =
+        genesis_state.into_block().await.context("failed to build the genesis block")?;
+
+    // Serialize and write the genesis block to disk.
+    let block_bytes = genesis_block.inner().to_bytes();
+    let genesis_block_path = genesis_block_directory.join(GENESIS_BLOCK_FILENAME);
+    fs_err::write(&genesis_block_path, block_bytes).with_context(|| {
+        format!("failed to write genesis block to {}", genesis_block_path.display())
+    })?;
```
Mirko-von-Leipzig:
```suggestion
    fs_err::write(&genesis_block_path, block_bytes)
       .context("failed to write genesis block")?;
```

bin/node/src/commands/validator.rs
```diff
@@ -112,7 +156,85 @@ impl ValidatorCommand {
...
+                .await
+        },
+    }
+}
+
+/// Builds the genesis state, writes account secret files, signs the genesis block, and writes it
+/// to disk.
+async fn build_and_write_genesis(
+    config: GenesisConfig,
+    signer: impl BlockSigner,
```
Mirko-von-Leipzig, line 209: Would suggest we use `&dyn BlockSigner` since this isn't really performance critical.

crates/store/src/lib.rs
Mirko-von-Leipzig, line 1: I imagine this diff isn't needed

crates/utils/src/fs.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use anyhow::Context;
+
+/// Validates that a directory either does not exist (and creates it) or exists and is empty.
+pub fn ensure_empty_directory(directory: &Path) -> anyhow::Result<()> {
+    if fs_err::exists(directory)? {
+        let is_empty = fs_err::read_dir(directory)?.next().is_none();
+        if !is_empty {
+            anyhow::bail!(format!("{} exists but it is not empty.", directory.display()));
+        }
```
Mirko-von-Leipzig: `anyhow::ensure!`

crates/utils/src/fs.rs
```diff
@@ -0,0 +1,22 @@
...
+    } else {
+        fs_err::create_dir(directory).with_context(|| {
+            format!(
+                "failed to create {} at {}",
+                directory.file_name().unwrap_or(std::ffi::OsStr::new("directory")).display(),
+                directory.display()
+            )
+        })?;
```
Mirko-von-Leipzig: `fs_err` already includes the path information

bin/node/src/commands/validator.rs
mmagician: To take a concrete example, we currently have a GER manager which is a `BasicWallet`. This will eventually be managed by Gateway and they should hold the secret key, providing us only with the public key that we would use to initialize the account.

bin/node/src/commands/validator.rs
sergerad: Think it gets gnarly because of the associated error type
```rust
pub trait BlockSigner {
    type Error: error::Error + Send + Sync + 'static;
```

mmagician (approved):

PR #1767, opened 2026-03-10
chore: rename `InnerForest` to `AccountStateForest`

drahnr:
`InnerForest` .. well, you see, is garbage.

Concludes and closes https://github.com/0xMiden/node/issues/1498

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4321,9 +4321,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.13"
+version = "0.11.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1906b49b0c3bc04b5fe5d86a77925ae6524a19b816ae38ce1e426255f1d8a31"
+checksum = "434b42fec591c96ef50e21e886936e66d3cc3f737104fdb9b737c40ffb94c098"
 dependencies = [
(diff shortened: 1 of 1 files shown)


crates/store/src/state/mod.rs
```diff
@@ -116,7 +116,7 @@ pub struct State {
     inner: RwLock<InnerState<TreeStorage>>,
 
     /// Forest-related state `(SmtForest, storage_map_roots, vault_roots)` with its own lock.
-    forest: RwLock<InnerForest>,
+    forest: RwLock<AccountStateForest>,
```
bobbinth, line 119: Not for this PR, but would it make sense to move the account state forest into the `InnerState`? Basically, I think of the store as 3 components:

1. The `block_store` that stores raw blocks.
2. The `db` which stores the "canonical" chain data.
3. The `inner` state, which stores in-memory and auxiliary data structures to help the store fulfill the RPC requests. This inner state can always be re-built from the canonical chain data in the `db`.

Under this categorization, the account state forest would go under the inner state, and I think the locking pattern for them is the same (but I didn't double-check this).

-----

Separately, I never loved the `InnerState` name - so, if anyone has better ideas, I'd be very happy to rename it.

Mirko-von-Leipzig: https://github.com/0xMiden/node/issues/1777

PR #1770, opened 2026-03-10
chore(ntx): log check and transaction results

igamigo:
Addresses https://github.com/0xMiden/node/discussions/1609#discussioncomment-16055870

### DIFF
--- a/crates/ntx-builder/src/actor/execute.rs
+++ b/crates/ntx-builder/src/actor/execute.rs
@@ -251,15 +251,13 @@ impl NtxContext {
         .await
         {
             Ok(NoteConsumptionInfo { successful, failed, .. }) => {
-                if !failed.is_empty() {
-                    for failed_note in &failed {
-                        tracing::info!(
-                            note_id = %failed_note.note.id(),
-                            nullifier = %failed_note.note.nullifier(),
-                            err = %failed_note.error,
-                            "note failed consumability check",
-                        );
-                    }
+                for failed_note in &failed {
+                    tracing::info!(
+                        note_id = %failed_note.note.id(),
+                        nullifier = %failed_note.note.nullifier(),
+                        err = %failed_note.error,
+                        "note failed consumability check",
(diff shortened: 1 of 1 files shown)


mmagician (approved):
LGTM thanks!

crates/ntx-builder/src/actor/execute.rs
```diff
@@ -251,6 +251,17 @@ impl NtxContext {
...
+                if !failed.is_empty() {
+                    for failed_note in &failed {
+                        tracing::info!(
+                            note_id = %failed_note.note.id(),
+                            nullifier = %failed_note.note.nullifier(),
+                            err = %failed_note.error,
+                            "note failed consumability check",
+                        );
+                    }
+                }
```
SantiagoPittella: the `if !failed.is_empty()` check could be removed, since if the iterator is empty we wont do anything anyways

SantiagoPittella: Also, should we use `tracing::error!` for this?

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -377,6 +386,12 @@ impl AccountActor {
             },
             // Execution completed with some failed notes.
             Ok((tx_id, failed, scripts_to_cache)) => {
+                tracing::info!(
```
SantiagoPittella, line 389: Same comment on the log level here.

crates/ntx-builder/src/actor/execute.rs
igamigo: It was suggested to use `info` level for these ([here](https://github.com/0xMiden/node/discussions/1609#discussioncomment-16055870) and in the Slack threads). They are also not really error conditions on the node itself, so I'd say info/trace level is more appropriate.

SantiagoPittella: sounds good

PR #1771, opened 2026-03-10
chore: remove chain tip from mempool subscription

### LINKED ISSUE #1766: Remove chain tip requirement from mempool subscription
> Noting that this innocuous error does occur on the first start of the bundled node:
> 
> ```
> 2026-03-10T02:19:48.260606Z ERROR ntx.block_producer.client.subscribe_to_mempool: miden-ntx-builder: crates/ntx-builder/src/block_producer.rs:61: error: status: 'Client specified an invalid argument', self: "Mempool's chain tip 1 does not match request's 0", metadata: {"content-type": "application/grpc", "date": "Tue, 10 Mar 2026 02:19:48 GMT"}
> ```

Ah that's not unexpected; though we can fix that now.

We currently have a loop of:

```rust
loop {
    let chain_tip = store.get_latest_header().await.block_num();
    if let Ok(subscription) = block_producer.subscribe(chain_tip).await {
        break subscription;
    }
}
```

which means there is a race condition between us fetching the chain tip and subscribing.

@SantiagoPittella I think we now remove the `chain_tip` from the mempool subscription gRPC as per #1487, which will "fix" this.

> Subscribing to the mempool currently requires the block-height, but this requirement can be removed so long as the above task is spawned after the subscription is created.

_Originally posted by @Mirko-von-Leipzig in https://github.com/0xMiden/node/issues/1764#issuecomment-4029045722_

SantiagoPittella:
closes #1766

Removes chain tip requirement from the mempool subscription.


proto/proto/internal/block_producer.proto
```diff
@@ -35,27 +35,19 @@ service Api {
...
+    // currently inflight events that have not yet been committed to the chain.
     //
-    // The event stream will contain all events after the chain tip. This includes all currently inflight
-    // events that have not yet been committed to the chain.
-    //
-    // Currently only a single active subscription is supported. Subscription requests will cancel the active
-    // subscription, if any.
+    // Currently only a single active subscription is supported. Subscription requests will cancel
+    // the active subscription, if any.
     rpc MempoolSubscription(MempoolSubscriptionRequest) returns (stream MempoolEvent) {}
```
Mirko-von-Leipzig: Should we change the request type tp `.google.proto.Empty` (or whatever that name is)

SantiagoPittella: Yes, that's better

PR #1772, opened 2026-03-10
feat/store: prune account code

drahnr:
We currently prune outdated storage items and vault entries, but not yet account code.

The PR adds that cleanup.

### DIFF
--- a/crates/store/src/db/models/queries/accounts/tests.rs
+++ b/crates/store/src/db/models/queries/accounts/tests.rs
@@ -1121,7 +1121,7 @@ fn test_select_account_vault_at_block_with_deletion() {
 // ACCOUNT CODE PRUNING TESTS
 // ================================================================================================
 
-/// Counts the number of rows in `account_codes` for the given code commitments.
+/// Counts the number of rows in `account_codes`.
 fn count_account_codes(conn: &mut SqliteConnection) -> usize {
     use schema::account_codes;
 


crates/store/src/db/models/queries/accounts.rs
```diff
@@ -1530,3 +1532,47 @@ fn prune_account_storage_map_values(
...
+)]
+fn prune_account_codes(
+    conn: &mut SqliteConnection,
+    cutoff_block: i64,
+) -> Result<usize, DatabaseError> {
+    use diesel::sql_types::BigInt;
+
+    diesel::sql_query(
+        "DELETE FROM account_codes \
+         WHERE code_commitment NOT IN ( \
```
drahnr, line 1566: We do delete everything that is not referenced by accounts in the latest 50 blocks or `is_latest=1`

crates/store/src/db/models/queries/accounts/tests.rs
```diff
@@ -1117,3 +1117,217 @@ fn test_select_account_vault_at_block_with_deletion() {
     assert_eq!(assets_at_block_3.len(), 1, "Should have 1 asset at block 3");
     assert_matches!(&assets_at_block_3[0], Asset::Fungible(f) if f.amount() == 2000);
 }
+
+// ACCOUNT CODE PRUNING TESTS
+// ================================================================================================
+
+/// Counts the number of rows in `account_codes` for the given code commitments.
+fn count_account_codes(conn: &mut SqliteConnection) -> usize {
```
Mirko-von-Leipzig: There are no given code commitments?

drahnr: It's a test helper only, it lives in `tests` and we create a new in-memory database per test, so I think it's best to assert global state that lives in the database.

Mirko-von-Leipzig: Just saying the comment is misleading

drahnr: Oh, I missed that, fixing.

PR #1774, opened 2026-03-11
Validator chain invariants

### LINKED ISSUE #1773: Validator ensures canonical chain
## Current state of the Validator

When asked to sign a block (`SignBlock` RPC), it:

1. Deserializes the `ProposedBlock`
2. Checks that every transaction ID in the block exists in its local `validated_transactions` table (i.e. was previously submitted via `SubmitProvenTransaction`)
3. Calls `proposed_block.into_header_and_body()` to build the header
4. Signs the header

It does **not**:

- Track the chain tip
- Verify `prev_block_commitment` links to the last block it signed
- Verify `block_num` is sequential
- Verify `chain_commitment` (the MMR) is consistent
- Verify the previous block's signature

The validator is completely stateless with respect to the chain. Its database only stores validated transactions — no block headers, no chain tip, no history.

## Proposed Changes
The original comment chain explains the suggested changes in order to complete this issue:

>I think the validator should make sure the block it is signing is the valid next block in the chain (for some reason, I thought that was already the case, but if not, let's create an issue for this).
>
>The way it could work:
>
>- We use the validator to "bootstrap" the genesis block (I think this is what this PR is about, but I haven't reviewed the details yet). This results in the validator keeping track of the chain tip (which is at first just the genesis block).
>- For any new block that the validator needs to sign, the validator:
>  - Makes sure that all transactions in the block are the ones it has seen/validated before (we already do this, I believe).
>  - Makes sure that given this transactions, we get the block header that we need to sign (we already do this, I believe).
>  - Makes sure that the new block header is built on top of the previous block. This may also require checking consistency of account and nullifier tree updates (it may be possible to incorporate this in the above two checks, but I'm not 100% sure).
>- If all looks good, the validator signs the block, and sets the new block's header as the chain tip.

_Originally posted by @bobbinth in https://github.com/0xMiden/node/pull/1764#discussion_r2914751343_

sergerad:
## Context

Closes #1773.

In #1764 we are adding a validator subcommand to construct and sign the genesis block. In this PR, we are building on top of that and storing the genesis block as the initial chain tip in the validator DB.

The validator is updated to maintain a chain tip as it continues to sign blocks. It uses this state to validate chain continuity (block number sequence, commitment references, etc).

## Changes
- Added a `block_headers` table to the validator database.
- Bootstrap now initializes the validator database with the genesis block header as the chain tip (new `--data-directory` argument).
- `validate_block` now enforces chain continuity: sequential block numbers and matching previous block commitments.
- `validate_block` allows for the current chain tip in `block_headers` table to be overwritten to allow for idempotence.
- Made the `db` module public so the bootstrap command can call `save_chain_tip` and `load` directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Fixed `TransactionHeader` serialization for row insertion on database & fixed transaction cursor on retrievals ([#1701](https://github.com/0xMiden/node/issues/1701)).
 - Added KMS signing support in validator ([#1677](https://github.com/0xMiden/node/pull/1677)).
 - Added per-IP gRPC rate limiting across services as well as global concurrent connection limit ([#1746](https://github.com/0xMiden/node/issues/1746)).
+- Network transaction actors now share the same gRPC clients, limiting the number of file descriptors being used ([#1806](https://github.com/0xMiden/node/issues/1806)).
 
 ### Changes
 
--- a/crates/ntx-builder/src/actor/execute.rs
+++ b/crates/ntx-builder/src/actor/execute.rs
@@ -1,8 +1,6 @@
 use std::collections::{BTreeMap, BTreeSet};
 use std::sync::Arc;
 
-use miden_node_proto::clients::ValidatorClient;
-use miden_node_proto::generated::{self as proto};
 use miden_node_utils::ErrorReport;
 use miden_node_utils::lru_cache::LruCache;
 use miden_node_utils::tracing::OpenTelemetrySpanExt;
@@ -35,7 +33,6 @@ use miden_protocol::transaction::{
 use miden_protocol::vm::FutureMaybeSend;
 use miden_remote_prover_client::RemoteTransactionProver;
 use miden_tx::auth::UnreachableAuth;
-use miden_tx::utils::Serializable;
 use miden_tx::{
     DataStore,
     DataStoreError,
@@ -56,7 +53,7 @@ use tracing::{Instrument, instrument};
 
 use crate::COMPONENT;
 use crate::actor::candidate::TransactionCandidate;
-use crate::clients::{BlockProducerClient, StoreClient};
+use crate::clients::{BlockProducerClient, StoreClient, ValidatorClient};
 use crate::db::Db;
 
 #[derive(Debug, thiserror::Error)]
@@ -327,16 +324,10 @@ impl NtxContext {
         proven_tx: &ProvenTransaction,
         tx_inputs: &TransactionInputs,
     ) -> NtxResult<()> {
-        let request = proto::transaction::ProvenTransaction {
-            transaction: proven_tx.to_bytes(),
-            transaction_inputs: Some(tx_inputs.to_bytes()),
-        };
         self.validator
-            .clone()
-            .submit_proven_transaction(request)
+            .submit_proven_transaction(proven_tx, tx_inputs)
             .await
-            .map_err(NtxError::Submission)?;
-        Ok(())
+            .map_err(NtxError::Submission)
     }
 }
 
--- a/crates/ntx-builder/src/actor/mod.rs
+++ b/crates/ntx-builder/src/actor/mod.rs
@@ -8,7 +8,6 @@ use std::time::Duration;
 use anyhow::Context;
 use candidate::TransactionCandidate;
 use futures::FutureExt;
-use miden_node_proto::clients::{Builder, ValidatorClient};
 use miden_node_proto::domain::account::NetworkAccountId;
 use miden_node_utils::ErrorReport;
 use miden_node_utils::lru_cache::LruCache;
@@ -21,11 +20,10 @@ use miden_remote_prover_client::RemoteTransactionProver;
 use miden_tx::FailedNote;
 use tokio::sync::{Notify, RwLock, Semaphore, mpsc};
 use tokio_util::sync::CancellationToken;
-use url::Url;
 
 use crate::NoteError;
 use crate::chain_state::ChainState;
-use crate::clients::{BlockProducerClient, StoreClient};
+use crate::clients::{BlockProducerClient, StoreClient, ValidatorClient};
 use crate::db::Db;
 
 // ACTOR REQUESTS
@@ -54,13 +52,13 @@ pub enum ActorRequest {
 pub struct AccountActorContext {
     /// Client for interacting with the store in order to load account state.
     pub store: StoreClient,
-    /// Address of the block producer gRPC server.
-    pub block_producer_url: Url,
-    /// Address of the Validator server.
-    pub validator_url: Url,
-    /// Address of the remote prover. If `None`, transactions will be proven locally, which is
-    // undesirable due to the performance impact.
-    pub tx_prover_url: Option<Url>,
+    /// Client for interacting with the block producer.
+    pub block_producer: BlockProducerClient,
+    /// Client for interacting with the validator.
+    pub validator: ValidatorClient,
+    /// Client for remote transaction proving. If `None`, transactions will be proven locally,
+    /// which is undesirable due to the performance impact.
+    pub prover: Option<RemoteTransactionProver>,
     /// The latest chain state that account all actors can rely on. A single chain state is shared
     /// among all actors.
     pub chain_state: Arc<RwLock<ChainState>>,
@@ -101,9 +99,9 @@ impl AccountActorContext {
         let (request_tx, _request_rx) = mpsc::channel(1);
 
         Self {
-            block_producer_url: url.clone(),
-            validator_url: url.clone(),
-            tx_prover_url: None,
+            block_producer: BlockProducerClient::new(url.clone()),
+            validator: ValidatorClient::new(url.clone()),
+            prover: None,
             chain_state,
             store: StoreClient::new(url),
             script_cache: LruCache::new(NonZeroUsize::new(1).unwrap()),
@@ -229,15 +227,6 @@ impl AccountActor {
         notify: Arc<Notify>,
         cancel_token: CancellationToken,
     ) -> Self {
-        let block_producer = BlockProducerClient::new(actor_context.block_producer_url.clone());
-        let validator = Builder::new(actor_context.validator_url.clone())
-            .without_tls()
-            .with_timeout(Duration::from_secs(10))
-            .without_metadata_version()
-            .without_metadata_genesis()
-            .with_otel_context_injection()
-            .connect_lazy::<ValidatorClient>();
-        let prover = actor_context.tx_prover_url.clone().map(RemoteTransactionProver::new);
         Self {
             origin,
             store: actor_context.store.clone(),
@@ -245,9 +234,9 @@ impl AccountActor {
             mode: ActorMode::NoViableNotes,
             notify,
             cancel_token,
-            block_producer,
-            validator,
-            prover,
+            block_producer: actor_context.block_producer.clone(),
+            validator: actor_context.validator.clone(),
+            prover: actor_context.prover.clone(),
             chain_state: actor_context.chain_state.clone(),
             script_cache: actor_context.script_cache.clone(),
             max_notes_per_tx: actor_context.max_notes_per_tx,
--- a/crates/ntx-builder/src/clients/mod.rs
+++ b/crates/ntx-builder/src/clients/mod.rs
@@ -1,5 +1,7 @@
 mod block_producer;
 mod store;
+mod validator;
 
 pub use block_producer::BlockProducerClient;
 pub use store::StoreClient;
+pub use validator::ValidatorClient;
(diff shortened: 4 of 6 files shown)


crates/validator/src/server/mod.rs
```diff
@@ -61,6 +62,24 @@ impl Validator {
...
+        // bootstrap; if it's missing, the validator cannot start.
+        let chain_tip = db
+            .query("load_chain_tip", load_chain_tip)
+            .await
+            .context("failed to load chain tip from database")?
+            .ok_or_else(|| {
+                anyhow::anyhow!(
+                    "validator database has no chain tip — run `validator bootstrap` first"
+                )
+            })?;
```
Mirko-von-Leipzig: `anyhow::Context` is implemented for `Option` as well

```suggestion
            .context("validator database has no chain tip — run `validator bootstrap` first")?;
```

crates/validator/src/server/mod.rs
```diff
@@ -61,6 +62,24 @@ impl Validator {
...
+
+        tracing::info!(
+            target: COMPONENT,
+            block_num = chain_tip.block_num().as_u32(),
```
Mirko-von-Leipzig:
```suggestion
            block.number = chain_tip.block_num().as_u32(),
```

Mirko-von-Leipzig: Might suggest also adding the block commitment field

crates/validator/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -13,3 +13,9 @@ CREATE TABLE validated_transactions (
 
 CREATE INDEX idx_validated_transactions_account_id ON validated_transactions(account_id);
 CREATE INDEX idx_validated_transactions_block_num ON validated_transactions(block_num);
+
+CREATE TABLE chain_tip (
+    id           INTEGER PRIMARY KEY CHECK (id = 0),
+    block_num    INTEGER NOT NULL,
+    block_header BLOB NOT NULL
+) WITHOUT ROWID;
```
Mirko-von-Leipzig: Should we perhaps store all of them to make it easier to rollback?

```suggestion
CREATE TABLE block_headers (
    block_num    INTEGER PRIMARY KEY CHECK (id = 0),
    block_header BLOB NOT NULL
) WITHOUT ROWID;
```

crates/validator/src/block_validation/mod.rs
```diff
@@ -21,19 +21,32 @@ pub enum BlockValidationError {
...
-/// the validator in the past.
-#[instrument(target = COMPONENT, skip_all, err)]
+/// Validates a proposed block by checking:
+/// 1. All transactions have been previously validated by this validator.
+/// 2. The block header can be successfully built from the proposed block.
+/// 3. The block is the valid next block in the chain (sequential block number, matching previous
+///    block commitment).
+///
+/// On success, returns the signature and the new block header (which becomes the new chain tip).
+#[instrument(target = COMPONENT, skip_all, err, fields(chain_tip = chain_tip.block_num().as_u32()))]
```
Mirko-von-Leipzig: We should also log `block.number` and `block.commitment`

crates/validator/src/block_validation/mod.rs
```diff
@@ -54,11 +67,41 @@ pub async fn validate_block(
...
+    if header.prev_block_commitment() != chain_tip.commitment() {
+        return Err(BlockValidationError::PrevBlockCommitmentMismatch);
+    }
+
+    let signature = sign_header(signer, &header).await?;
+    Ok((signature, header))
+}
+
+/// Signs a block header using the validator's signer.
+#[instrument(target = COMPONENT, name = "sign_block", skip_all, err, fields(header = header.block_num().as_u32()))]
```
Mirko-von-Leipzig: Calling the field `header` seems a bit weird. We're never going to find that. `block.number` - though we already should be doing that on the parent span.

crates/validator/src/block_validation/mod.rs
```diff
@@ -54,11 +67,41 @@ pub async fn validate_block(
...
-    let signature = info_span!("sign_block")
-        .in_scope(async move || signer.sign(&header).await)
-        .await
-        .map_err(|err| BlockValidationError::BlockSigningFailed(err.to_string()))?;
+    // If the proposed block produces a header identical to the current chain tip, this is a
+    // retry of an already-signed block. Re-sign and return without advancing state.
+    if header.commitment() == chain_tip.commitment() {
+        let signature = sign_header(signer, &header).await?;
+        return Ok((signature, header));
+    }
```
Mirko-von-Leipzig: This is one edge case, but there is also the situation where _after_ we signed the block (and updated the local chain-tip), that the block gets lost or dropped _before_ it gets committed.

In that situation, only `header.block_num()` would match the `chain tip`, but not necessarily the commitment - it could be made of different transactions.

imo we need to allow for a "fork" of a single block i.e. 

```suggestion
    if header.block_num() == chain_tip.block_num() {
        let signature = sign_header(signer, &header).await?;
        return Ok((signature, header));
    }
```

crates/validator/src/server/mod.rs
```diff
@@ -176,19 +206,38 @@ impl api_server::Api for ValidatorServer {
...
+                .await
+                .map_err(|err| {
+                    tonic::Status::invalid_argument(format!(
+                        "Failed to validate block: {}",
+                        err.as_report()
+                    ))
+                })?;
+
+        // Update the chain tip only if the block is new (not a retry of the current tip).
+        if new_header.commitment() != chain_tip.commitment() {
```
Mirko-von-Leipzig: nit: imo this edge case isn't worth optimising - just always save it.

crates/validator/src/server/mod.rs
```diff
@@ -176,19 +206,38 @@ impl api_server::Api for ValidatorServer {
...
-        // Validate the block.
-        let signature =
-            validate_block(proposed_block, &self.signer, &self.db).await.map_err(|err| {
-                tonic::Status::invalid_argument(format!(
-                    "Failed to validate block: {}",
-                    err.as_report()
-                ))
-            })?;
+        // Acquire the chain tip lock for the duration of validation and update.
+        let mut chain_tip = self.chain_tip.lock().await;
```
Mirko-von-Leipzig: I'm unsure if its worth "caching" this value instead of just reading it from the database each time - the database transaction would also act as a lock, and we wouldn't have something that can desync if we screw it up.

Not a strong opinion though.

crates/validator/src/block_validation/mod.rs
Mirko-von-Leipzig: I guess you'll _also_ need to check against `chain_tip.parent.commitment()` which will be awkward without the database, or caching that field in the chain tip as well. You essentially need to "rollback" by one block.

Mirko-von-Leipzig: An alternative if you don't want to allow forks, is to have a more complicated gRPC setup, with a two-phase request-response-ack. But even then you can desync if things crash.

crates/validator/src/block_validation/mod.rs
sergerad: those aren't available from ProposedBlock, if thats what you mean. I could add an event in the fn? After
```rust
    let (proposed_header, _) = proposed_block
```

Mirko-von-Leipzig: Avoid events, but we can inject the value within the function using `span.set_attribute`

crates/db/src/errors.rs
```diff
@@ -43,6 +43,8 @@ pub enum DatabaseError {
         inner: Option<Box<dyn std::error::Error + Send + Sync + 'static>>,
         to: &'static str,
     },
+    #[error("conflict: {0}")]
+    Conflict(String),
```
Mirko-von-Leipzig: This feels a bit weird - should we not have more specific variants?

crates/validator/src/db/migrations/2025062000000_setup/up.sql
```diff
@@ -13,3 +13,8 @@ CREATE TABLE validated_transactions (
 
 CREATE INDEX idx_validated_transactions_account_id ON validated_transactions(account_id);
 CREATE INDEX idx_validated_transactions_block_num ON validated_transactions(block_num);
+
+CREATE TABLE block_headers (
+    block_num    INTEGER PRIMARY KEY,
```
Mirko-von-Leipzig, line 18: If we also store `block_commitment` can we avoid pulling in the entire header each check?

crates/validator/src/server/mod.rs
Mirko-von-Leipzig, line 1: I think this breaks under concurrent requests. This is compounded by having separate database transactions/queries instead of using a single transaction throughout.

I'm unsure if using a single database tx would help either, since its possible to concurrently perform an upsert on block `N`, while storing `N+1` based on the old value.

A simple fix is adding a `Semaphor` with exactly _one_ permit after deserialization. I still don't like that we keep opening new database transactions multiple times per request though. This is also a problem in the `store`, but that's a `diesel` problem.

crates/validator/src/db/migrations/2025062000000_setup/up.sql
sergerad: That depends on whether we want to do any additional checks on top of the block number + commitment ones I have now.

@bobbinth may have had some ideas on that point - whether there is more to compare with a new / proposed chain tip (not signed) against the previous signed chain tip beyond just block number and commitment continuity.

crates/validator/src/server/mod.rs
sergerad: Yea in the initial impl we had the mutex for this reason. Added the semaphore

Mirko-von-Leipzig (approved):
Thank you!

bin/node/src/commands/bundled.rs
```diff
@@ -100,6 +100,7 @@ impl BundledCommand {
                 crate::commands::validator::ValidatorCommand::bootstrap_genesis(
                     &data_directory,
                     &accounts_directory,
+                    &data_directory,
```
mmagician, line 106: We're passing the same `data_directory` twice - I assume this means that we can't distinguish the two directories in `bundled` mode?

bin/node/src/commands/validator.rs
```diff
@@ -31,14 +31,18 @@ pub enum ValidatorCommand {
...
     Bootstrap {
         /// Directory in which to write the genesis block file.
         #[arg(long, value_name = "DIR")]
         genesis_block_directory: PathBuf,
         /// Directory to write the account secret files (.mac) to.
         #[arg(long, value_name = "DIR")]
         accounts_directory: PathBuf,
+        /// Directory in which to store the validator's database.
+        #[arg(long, env = ENV_DATA_DIRECTORY, value_name = "DIR")]
+        data_directory: PathBuf,
```
mmagician, line 45: This might be for a separate discussion/PR, but I find the argument names a little misleading: it's not immediately clear whether these are input directories or output directories, for example:
- `genesis_block_directory` is an output location, but
- `genesis_config_file` is an input location (here, path, not a directory, but the same argument applies)

crates/validator/src/block_validation/mod.rs
```diff
@@ -21,19 +22,35 @@ pub enum BlockValidationError {
...
 
-/// Validates a block by checking that all transactions in the proposed block have been processed by
-/// the validator in the past.
-#[instrument(target = COMPONENT, skip_all, err)]
+/// Validates a proposed block by checking:
+/// 1. All transactions have been previously validated by this validator.
+/// 2. The block header can be successfully built from the proposed block.
+/// 3. The block is either: a. The valid next block in the chain (sequential block number, matching
+///    previous block commitment), or b. A replacement block at the same height as the current chain
+///    tip, validated against the previous block header.
```
mmagician, line 44: what are the situations where we would run into scenario b) ?

Mirko-von-Leipzig: The order of operations for block `N+1` is roughly

1. [node]      block is built
2. [validator] block is signed
     1. block is signed 
     2. block is now in validator database as latest
     3. signed block is returned to [node]
4. [node]      block is stored (and is now considered committed)

So there are several steps between (2ii) and (3) where things can go wrong e.g. node or validator crash, or network IO issue, or node database issue, etc, where the validator would have the signed block `N+1` but the node is still on `N`.

The validator can only consider `N+1` finalized once the node requests `N+2`. The alternative would be a two-phase ack `node submit -> validator sign -> node ack`. But that's more complicated than this.

bin/node/src/commands/bundled.rs
sergerad: Yea no need to distinguish. TBH I don't think we ever needed `bundled bootstrap` subcommand at all. And we are working on removing `bundled` command altogether in #1786.

bin/node/src/commands/validator.rs
sergerad: Wouldn't hurt to specify `_output_` for the directories I guess

mmagician: created https://github.com/0xMiden/node/issues/1807

PR #1778, opened 2026-03-11
chore: bring in changes from `v0.14.0-alpha` into `next`

mmagician:
to be merged w/o squashing


crates/ntx-builder/src/actor/execute.rs
```diff
@@ -251,6 +251,15 @@ impl NtxContext {
         .await
         {
             Ok(NoteConsumptionInfo { successful, failed, .. }) => {
+                for failed_note in &failed {
+                    tracing::info!(
+                        note_id = %failed_note.note.id(),
+                        nullifier = %failed_note.note.nullifier(),
+                        err = %failed_note.error,
+                        "note failed consumability check",
```
Mirko-von-Leipzig:
```suggestion
                        note.id = %failed_note.note.id(),
                        nullifier = %failed_note.note.nullifier(),
                        err = %failed_note.error.as_report(),
                        "note failed consumability check",
```

crates/ntx-builder/src/actor/mod.rs
Mirko-von-Leipzig, line 1: This is okay for short-term, but we should be using spans to communicate these sorts of things.

PR #1782, opened 2026-03-11
feat/rocksdb: expose rocksdb config per `LargeSmt` tree

drahnr:
As per discussion, lowers the default number of max open file descriptors to 64 from 512 per isntance and exposes them for host specific tuning.

There's a bit of boilerplate due to the conditional `rocksdb` feature gating.  

Ref https://github.com/0xMiden/node/issues/1776

### DIFF
--- a/crates/utils/src/clap.rs
+++ b/crates/utils/src/clap.rs
@@ -141,7 +141,7 @@ impl GrpcOptionsExternal {
 /// Collection of per usage storage backend configurations.
 ///
 /// Note: Currently only contains `rocksdb` related configuration.
-#[derive(clap::Args, Clone, Debug, PartialEq, Eq)]
+#[derive(clap::Args, Clone, Debug, Default, PartialEq, Eq)]
 pub struct StorageOptions {
     #[cfg(feature = "rocksdb")]
     #[clap(flatten)]
@@ -149,17 +149,6 @@ pub struct StorageOptions {
     #[cfg(feature = "rocksdb")]
     #[clap(flatten)]
     pub nullifier_tree: NullifierTreeRocksDbOptions,
-}
-
-impl Default for StorageOptions {
-    fn default() -> Self {
-        Self {
-            #[cfg(feature = "rocksdb")]
-            account_tree: RocksDbOptions::default().into(),
-            #[cfg(feature = "rocksdb")]
-            nullifier_tree: RocksDbOptions::default().into(),
-        }
-    }
 }
 
 impl StorageOptions {
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -141,7 +141,7 @@ impl BundledCommand {
         }
     }
 
-    #[expect(clippy::too_many_lines)]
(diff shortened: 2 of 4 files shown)


crates/utils/src/clap.rs
```diff
@@ -3,6 +3,10 @@
 use std::num::{NonZeroU32, NonZeroU64};
 use std::time::Duration;
 
+#[cfg(feature = "rocksdb")]
+mod rocksdb;
+pub use rocksdb::*;
```
Mirko-von-Leipzig, line 9:
```suggestion
#[cfg(feature = "rocksdb")]
pub use rocksdb::*;
```

crates/utils/src/clap.rs
```diff
@@ -132,3 +136,49 @@ impl GrpcOptionsExternal {
...
+                max_open_fds: self::rocksdb::BENCH_ROCKSDB_MAX_OPEN_FDS,
+                cache_size_in_bytes: self::rocksdb::DEFAULT_ROCKSDB_CACHE_SIZE,
+            };
+            let nullifier_tree = NullifierTreeRocksDbOptions {
+                max_open_fds: BENCH_ROCKSDB_MAX_OPEN_FDS,
+                cache_size_in_bytes: DEFAULT_ROCKSDB_CACHE_SIZE,
+            };
+            Self { account_tree, nullifier_tree }
+        }
+        #[cfg(not(feature = "rocksdb"))]
```
Mirko-von-Leipzig, line 171: Can we not drop the `rocksdb` feature?

docs/internal/src/store.md
```diff
@@ -14,6 +14,20 @@ Note that the migration logic includes both a schema number _and_ a hash based o
...
+## RocksDB tree storage
+
+The account and nullifier trees are persisted in separate RocksDB instances under
+`<data-directory>/accounttree` and `<data-directory>/nullifiertree`, managed by
+`crates/large-smt-backend-rocksdb`. Column families: `leaves`, `st24`–`st56` (subtrees at each
+depth), `metadata` (root/counts), `depth24` (cached depth-24 hashes for fast startup).
+
+Compaction parallelism and background jobs are set to `rayon::current_num_threads()` automatically.
+WAL sync per write is disabled for throughput; a 512 MiB WAL cap bounds recovery time. Bloom filter
+bits vary by depth (8.0–12.0) and memtables are 128 MiB per CF. See `RocksDbStorage::open` for the
```
Mirko-von-Leipzig: What is `CF`

crates/utils/src/clap.rs
drahnr: Technically, right now we could, but we'd re-intro it shortly once I am freed to investigate another backend :upside_down_face: - unless we make it a runtime choice.

docs/internal/src/store.md
drahnr: column family, sorry

crates/utils/src/clap.rs
drahnr: Thanks!

PR #1789, opened 2026-03-12
fix(store): query only latest vault rows when applying account delta

TomasArrachea:
While testing for https://github.com/0xMiden/miden-client/pull/1843, I made some changes to the node seed store binary and bumped into this bug.

When applying a partial account delta, the store reconstructed the `AssetVault` by querying all historical vault rows instead of only the current state. This caused `DuplicateAsset` error when the same vault key had entries across multiple blocks. i.e., after 2+ balance updates to the same faucet for an account. This is way the current test did not catch the error, I updated it to update the account 3 times and it fails.

The fix is to replace the historical range query with a new `select_latest_vault_assets` function that filters on `is_latest = true`, returning only the current vault state for reconstruction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@
 - [BREAKING] Modified `TransactionHeader` serialization to allow converting back into the native type after serialization ([#1759](https://github.com/0xMiden/node/issues/1759)).
 - Removed `chain_tip` requirement from mempool subscription request ([#1771](https://github.com/0xMiden/node/pull/1771)).
 - Moved bootstrap procedure to `miden-node validator bootstrap` command ([#1764](https://github.com/0xMiden/node/pull/1764)).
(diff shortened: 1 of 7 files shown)


crates/store/src/db/models/queries/accounts/delta.rs
```diff
@@ -189,6 +189,36 @@ pub(super) fn select_vault_balances_by_faucet_ids(
...
+            .filter(vault::is_latest.eq(true))
+            .load(conn)?;
+
+    let mut assets = Vec::new();
+    for (_vault_key_bytes, maybe_asset_bytes) in entries {
+        if let Some(asset_bytes) = maybe_asset_bytes {
+            assets.push(Asset::read_from_bytes(&asset_bytes)?);
+        }
+    }
+    Ok(assets)
```
drahnr:
```suggestion
  Result::from_iter(entries.into_iter().filter_map(|(_vault_key_bytes, maybe_asset_bytes)| { maybe_asset_bytes.map(Asset::read_from_bytes).transpose()}));
```

PR #1792, opened 2026-03-13
feat: add GetNoteError endpoint in ntx builder

### LINKED ISSUE #1758: Expose network note failures
Debugging network notes is painful because these aren't executed locally and therefore developers have very little feedback on what's going wrong.

We could make this easier by storing the note's latest execution error in the database and exposing this via gRPC.

There is a downside to this, which is that the ntx-builder now requires a gRPC server; though at least this also lets us add a status endpoint in the future...

Open questions:
- Do we expose status in general, or only latest error? The latter is much simpler; I would prefer to start there.

cc @mmagician @partylikeits1983 for further inputs.

SantiagoPittella:
closes #1758 

- Add a GetNoteError gRPC endpoint that returns the latest execution error for a network note, including attempt count and last attempt block number.
- The NTX builder now runs its own internal gRPC server. The RPC component proxies GetNoteError requests to it.
- Store last_error and last_attempt_block_num for each note in the NTX builder database.

### DIFF
--- a/crates/ntx-builder/src/server.rs
+++ b/crates/ntx-builder/src/server.rs
@@ -38,11 +38,6 @@ impl NtxBuilderRpcServer {
             .build_v1()
             .context("failed to build reflection service")?;
 
-        let reflection_service_alpha = server::Builder::configure()
-            .register_file_descriptor_set(ntx_builder_api_descriptor())
-            .build_v1alpha()
-            .context("failed to build reflection service")?;
-
         tracing::info!(
             target: COMPONENT,
             endpoint = ?listener.local_addr(),
@@ -54,7 +49,6 @@ impl NtxBuilderRpcServer {
             .layer(TraceLayer::new_for_grpc().make_span_with(grpc_trace_fn))
             .add_service(api_service)
             .add_service(reflection_service)
-            .add_service(reflection_service_alpha)
             .serve_with_incoming(TcpListenerStream::new(listener))
             .await
             .context("failed to serve NTX builder gRPC API")


crates/ntx-builder/src/actor/mod.rs
```diff
@@ -411,23 +410,66 @@ impl AccountActor {
...
+                                let note_error = f.error.as_report();
+                                tracing::info!(
+                                    note.id = %f.note.id(),
+                                    nullifier = %f.note.nullifier(),
+                                    err = %note_error,
+                                    "note failed: consumability check",
+                                );
+                                (f.note.nullifier(), note_error)
+                            })
+                            .collect()
```
sergerad: nit: might be worth deduping this and its use above. Maybe not though

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -449,12 +491,16 @@ impl AccountActor {
     /// Sends a request to the coordinator to mark notes as failed and waits for the DB write to
     /// complete. This prevents a race condition where the actor could re-select the same notes
     /// before the failure counts are updated in the database.
-    async fn mark_notes_failed(&self, nullifiers: &[Nullifier], block_num: BlockNumber) {
+    async fn mark_notes_failed(
+        &self,
+        failed_notes: &[(Nullifier, String)],
```
sergerad: Would maybe be nice to replace `String` with `Box<dyn ErrorReport>` but the problem is that one of the branches doesn't provide an `Error`: `(note.nullifier(), error_msg.clone())`.

proto/proto/rpc.proto
```diff
@@ -105,6 +105,15 @@ service Api {
...
     rpc SyncChainMmr(SyncChainMmrRequest) returns (SyncChainMmrResponse) {}
+
+    // NOTE DEBUGGING ENDPOINTS
+    // --------------------------------------------------------------------------------------------
+
+    // Returns the latest execution error for a network note, if any.
+    //
+    // This is useful for debugging notes that are failing to be consumed by
+    // the network transaction builder.
+    rpc GetNoteError(note.NoteId) returns (GetNoteErrorResponse) {}
```
sergerad, line 116: Its starting to feel like we are implementing a manual gateway/reverse proxy. But maybe thats not a bad thing.

docs/internal/src/ntx-builder.md
```diff
@@ -53,3 +53,14 @@ coordinator (via the `send_targeted` path).
 
 The block-producer remains blissfully unaware of network transactions. From its perspective a
 network transaction is simply the same as any other.
+
+## gRPC Server
+
+The NTB exposes an internal gRPC server for querying its state. The RPC component proxies public
```
sergerad: nit: NTX Builder, not NTB

crates/ntx-builder/src/actor/mod.rs
sergerad: The match statement is pretty long in general. If we could reduce it that would be good.

proto/proto/rpc.proto
Mirko-von-Leipzig: Do you mean the `RPC` proxy into the "internal" components? It is essentially a gateway yeah.. If this was non gRPC we could probably use some off the shelf implementation for this, though maybe not, because there are some non-trivial validation steps happening.

docs/internal/src/ntx-builder.md
juan518munoz: Addressed in 25aee9df5cee

crates/ntx-builder/src/actor/mod.rs
juan518munoz: It's doable, but instead of using `Box`, we need to wrap with `Arc`, and also restrain to `Sync + Send`, creating the following type:
```rs
type NoteError = Arc<dyn ErrorReport + Send + Sync>
```

Which may be an overcomplicated solution just to get rid of the `String`, so we can roll it back if prefered.

25aee9df5cee

crates/ntx-builder/src/actor/mod.rs
juan518munoz: Addressed the dedup in 25aee9df5cee

The match is still present, but it doesn't look as large now, we can still try to reduce it further tho.

crates/ntx-builder/src/server.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        tonic::transport::Server::builder()
+            .layer(CatchPanicLayer::custom(catch_panic_layer_fn))
+            .layer(TraceLayer::new_for_grpc().make_span_with(grpc_trace_fn))
+            .add_service(api_service)
+            .add_service(reflection_service)
+            .add_service(reflection_service_alpha)
+            .serve_with_incoming(TcpListenerStream::new(listener))
+            .await
+            .context("failed to serve NTX builder gRPC API")
```
Mirko-von-Leipzig: Support for this was removed recently in #1795 


```suggestion
        tracing::info!(
            target: COMPONENT,
            endpoint = ?listener.local_addr(),
            "NTX builder gRPC server initialized",
        );

        tonic::transport::Server::builder()
            .layer(CatchPanicLayer::custom(catch_panic_layer_fn))
            .layer(TraceLayer::new_for_grpc().make_span_with(grpc_trace_fn))
            .add_service(api_service)
            .add_service(reflection_service)
            .serve_with_incoming(TcpListenerStream::new(listener))
            .await
            .context("failed to serve NTX builder gRPC API")
```

proto/proto/internal/ntx_builder.proto
```diff
@@ -0,0 +1,30 @@
...
+// ================================================================================================
+
+// API for querying network transaction builder state.
+service Api {
+    // Returns the latest execution error for a network note, if any.
+    //
+    // This is useful for debugging notes that are failing to be consumed by
+    // the network transaction builder.
+    rpc GetNoteError(note.NoteId) returns (GetNoteErrorResponse) {}
+}
```
bobbinth, line 17: It may be good to generalize this to something like `GetNoteStatus`. This way, we can return status of successfully executed notes as well. Otherwise, it may not be clear if the note just had a transient error and then was consumed correctly vs. the error is still preventing it from being consumed.

This could be something like `GetNetworkNoteStatus` endpoint, and it could return:

- "Not found" if the note doesn't exist or is not a network note.
- Status of the note - maybe:
  - `pending` - we are still trying to execute this note.
  - `processed` - the note made it into a transaction which was sent to the block producer.
  - `committed` - the network is now in the store
  - `discarded` - the note failed too many times and won't be retried in the future.
- The last error associated with the note execution - basically, what we have here now (this is applicable for all status above).

I'll create an issue for this.

PR #1797, opened 2026-03-16
feat: add miden-genesis tool for canonical genesis state

### LINKED ISSUE #1788: Define the canonical genesis state
Some canonical genesis incl. AggLayer accounts should be defined in this repo, so that a genesis block can be generated by the validator similar to [here](https://github.com/0xMiden/miden-client/pull/1882/changes#diff-56f7e62b96a92735c096f6c857b1445ca0aead3d5dedff70ee65b29369227a7b).

mmagician:
## Summary

- Adds a new `miden-genesis` binary crate that generates canonical AggLayer genesis accounts and a `genesis.toml` config file
- Generates three accounts: bridge admin & GER manager (`BasicWallet`s), and `AggLayerBridge`
- Only the bridge account (nonce=1) is included in `genesis.toml` for the genesis block; bridge admin and GER manager are local accounts (nonce=0) to be deployed later via transactions. But, they are "implicitly" included in the genesis block in that the `AggLayerBridge` account makes reference to them in its storage.

Closes #1788

## Operator workflow

```bash
# Generate canonical genesis accounts
miden-genesis --output-dir ./genesis

# Bootstrap the node
miden-node validator bootstrap \
  --genesis-block-directory ./data \
  --accounts-directory ./accounts \
  --genesis-config-file ./genesis/genesis.toml \
  --validator.key.hex <key>
```

🤖 Generated by hand & with [Claude Code](https://claude.com/claude-code)

---

I wanted to add this as a separate minimal binary (non-publishable) because I do see this as a separate role from the validator (even though for now they'll be run by the same entity). But it should be trivial to move this in a new command on the validator if desired.

### DIFF
--- a/bin/genesis/src/main.rs
+++ b/bin/genesis/src/main.rs
@@ -197,99 +197,58 @@ fn bump_nonce_to_one(mut account: Account) -> anyhow::Result<Account> {
 
 #[cfg(test)]
 mod tests {
-    use miden_protocol::ZERO;
+    use miden_node_store::genesis::config::GenesisConfig;
+    use miden_protocol::crypto::dsa::ecdsa_k256_keccak::SecretKey;
     use miden_protocol::utils::Serializable;
 
     use super::*;
 
-    #[test]
-    fn default_mode_generates_accounts_with_secret_keys() {
-        let dir = tempfile::tempdir().unwrap();
+    /// Parses the generated genesis.toml, builds a genesis block, and asserts the bridge account
+    /// is included with nonce=1.
+    async fn assert_valid_genesis_block(dir: &Path) {
+        let bridge_id = AccountFile::read(dir.join("bridge.mac")).unwrap().account.id();
 
-        run(dir.path(), None, None).expect("run should succeed");
+        let config = GenesisConfig::read_toml_file(&dir.join("genesis.toml")).unwrap();
+        let signer = SecretKey::read_from_bytes(&[0x01; 32]).unwrap();
+        let (state, _) = config.into_state(signer).unwrap();
 
-        // All 4 files should exist.
-        assert!(dir.path().join("bridge_admin.mac").exists());
-        assert!(dir.path().join("ger_manager.mac").exists());
-        assert!(dir.path().join("bridge.mac").exists());
-        assert!(dir.path().join("genesis.toml").exists());
+        let bridge = state.accounts.iter().find(|a| a.id() == bridge_id).unwrap();
+        assert_eq!(bridge.nonce(), ONE);
 
-        // Bridge account should have nonce=1 (genesis account).
-        let bridge = AccountFile::read(dir.path().join("bridge.mac")).unwrap();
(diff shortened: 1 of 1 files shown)


bin/genesis/README.md
```diff
@@ -0,0 +1,60 @@
...
+## Output
+
+The tool generates the following files in the output directory:
+
+- `bridge_admin.mac` - Bridge admin wallet (nonce=0, deployed later via transaction)
+- `ger_manager.mac` - GER manager wallet (nonce=0, deployed later via transaction)
+- `bridge.mac` - AggLayer bridge account (nonce=1, included in genesis block)
+- `genesis.toml` - Genesis configuration referencing only `bridge.mac`
+
+When public keys are omitted, the `.mac` files for bridge admin and GER manager include generated secret keys. When public keys are provided, no secret keys are included.
```
Mirko-von-Leipzig, line 30: If this is intended for internal use only, why do we have two different "modes"?

bin/genesis/README.md
```diff
@@ -0,0 +1,60 @@
+# Miden Genesis
+
+A tool for generating canonical Miden genesis accounts and configuration.
```
Mirko-von-Leipzig, line 3: If this is intended for internal use only, then I would add that disclaimer here, and repeat it where applicable.

I'm concerned that local node users will attempt to use this.

bin/genesis/README.md
```diff
@@ -0,0 +1,60 @@
...
+- `bridge_admin.mac` - Bridge admin wallet (nonce=0, deployed later via transaction)
+- `ger_manager.mac` - GER manager wallet (nonce=0, deployed later via transaction)
+- `bridge.mac` - AggLayer bridge account (nonce=1, included in genesis block)
+- `genesis.toml` - Genesis configuration referencing only `bridge.mac`
+
+When public keys are omitted, the `.mac` files for bridge admin and GER manager include generated secret keys. When public keys are provided, no secret keys are included.
+
+The bridge account always uses NoAuth and has no secret keys.
+
+## Bootstrapping a node
```
Mirko-von-Leipzig, line 34: This is going to get outdated pretty often and we'll forget to update this.

It would be better to keep this in our operator docs.

bin/genesis/README.md
Mirko-von-Leipzig, line 1: Could you describe the different deployment flows and where this fits in?

1. Local node
2. Testnet deploy
3. Automated devnet deploy
4. ...

#### Some bikeshedding

Should this be `miden-node make-genesis`? Bear in mind that we probably want to access this binary somewhere where we can't (or don't want to) compile code. And therefore this _should_ likely be installable and versionable.

#### Some thought experiments

Maybe we should have a `miden-node genesis` subcommand? 

```
# This binary
miden-node genesis generate-accounts ...

# miden-node validator bootstrap
miden-node genesis sign

# miden-node store bootstrap
miden-node genesis bootstrap-store
```

PR #1801, opened 2026-03-17
feat: limit execution cycles for network transactions

### LINKED ISSUE #1800: Limit the number of execution cycles for network txs
We currently use the default executor, which defaults to `1 << 29` cycles.

We should limit this to something more reasonable, like 64k. It likely makes sense to make this a configurable parameter.

Some additional context https://github.com/0xMiden/protocol/pull/2618

juan518munoz:
Default to 64k (1 << 16) cycles instead of the protocol maximum (1 << 29). 

Note: This is configurable via `--ntx-builder.max-tx-cycles` CLI flag or `MIDEN_NTX_MAX_TX_CYCLES` env var.

Closes #1800

### DIFF
--- a/bin/node/src/commands/bundled.rs
+++ b/bin/node/src/commands/bundled.rs
@@ -103,6 +103,7 @@ impl BundledCommand {
                 crate::commands::validator::ValidatorCommand::bootstrap_genesis(
                     &data_directory,
                     &accounts_directory,
+                    &data_directory,
                     genesis_config_file.as_ref(),
                     validator_key,
                 )
--- a/bin/node/src/commands/validator.rs
+++ b/bin/node/src/commands/validator.rs
@@ -31,7 +31,8 @@ pub enum ValidatorCommand {
     /// Bootstraps the genesis block.
     ///
     /// Creates accounts from the genesis configuration, builds and signs the genesis block,
-    /// and writes the signed block and account secret files to disk.
+    /// and writes the signed block and account secret files to disk. Also initializes the
+    /// validator's database with the genesis block as the chain tip.
     Bootstrap {
         /// Directory in which to write the genesis block file.
         #[arg(long, value_name = "DIR")]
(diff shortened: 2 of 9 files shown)


bin/node/src/commands/mod.rs
```diff
@@ -195,6 +197,17 @@ pub struct NtxBuilderConfig {
...
+    /// Maximum number of VM execution cycles allowed for a single network transaction.
+    ///
+    /// Network transactions that exceed this limit will fail. Defaults to 64k (65536) cycles.
+    #[arg(
+        long = "ntx-builder.max-cycles",
+        env = ENV_NTX_MAX_CYCLES,
+        default_value_t = DEFAULT_NTX_MAX_CYCLES,
+        value_name = "NUM"
+    )]
+    pub max_cycles: u32,
```
bobbinth: I would maybe call this `max_tx_cycles` for clarity.

Also, on line 202, I'd say "... Defaults to 2^16 (65536) cycles."

Mirko-von-Leipzig: The default should be displayed by `clap` as part of the help message

juan518munoz: > I would maybe call this max_tx_cycles for clarity.
> Also, on line 202, I'd say "... Defaults to 2^16 (65536) cycles."

Addressed in [96481f2](https://github.com/0xMiden/node/pull/1801/commits/96481f26b0cc45a32f4bb01a3e81a472cf860da0)

> The default should be displayed by clap as part of the help message

This is already done by default on the version of clap being used:
```bash
      --ntx-builder.max-cycles <NUM>
          Maximum number of VM execution cycles allowed for a single network transaction.

          Network transactions that exceed this limit will fail. Defaults to 2^16 (65536) cycles.

          [env: MIDEN_NTX_MAX_CYCLES=]
          [default: 65536]
```

PR #1802, opened 2026-03-17
feat/store: improve node store startup time - 2nd round

drahnr:
Ref #1697 

Adds a bench with progress bars and prints closing spans, which shows "the numbers" for `State::load` pieces.


Mirko-von-Leipzig:
Largely implemented now thanks for persistent backends and improved account reads.

PR #1816, opened 2026-03-21
fix(store): handle AllEntries pagination for genesis block storage maps

mmagician:
## Summary

- Fixes a bug where `GetAccount` RPC with `AllEntries(true)` returns an internal error for storage maps with entries exceeding the pagination limit when all entries are in genesis block 0
- Root cause: `select_account_storage_map_values_paged` computes `last_block_num.saturating_sub(1)` = -1 (i64) when all entries share the same block, which fails `BlockNumber::from_raw_sql`
- Fix: when `take_while` yields empty results (single-block overflow), return `block_range.start()` to signal no pagination progress, triggering `limit_exceeded` in the caller

Discovered via 0xMiden/miden-client#1926

## Test plan

- [x] Added test `select_storage_map_sync_values_all_entries_in_genesis_block` that reproduces the bug (fails without fix, passes with fix)
- [x] Full `miden-node-store` test suite passes (121 tests)

### DIFF
--- a/crates/store/src/db/models/queries/accounts.rs
+++ b/crates/store/src/db/models/queries/accounts.rs
@@ -727,7 +727,7 @@ pub(crate) fn select_account_storage_map_values_paged(
             .map(StorageMapValue::from_raw_row)
             .collect::<Result<Vec<_>, DatabaseError>>()?;
 
-        let last_block_included = values.last().unwrap_or_else(|| *block_range.start());
+        let last_block_included = values.last().map_or(*block_range.start(), |v| v.block_num);
 
         (last_block_included, values)
     } else {


crates/store/src/db/models/queries/accounts.rs
```diff
@@ -721,17 +721,20 @@ pub(crate) fn select_account_storage_map_values_paged(
...
 
-        (BlockNumber::from_raw_sql(last_block_num.saturating_sub(1))?, values)
+        if values.is_empty() {
+            // All entries are in the same block and exceed the limit.
+            // Return the range start to signal no progress was made,
+            // which the caller interprets as limit_exceeded.
+            (*block_range.start(), values)
+        } else {
+            (BlockNumber::from_raw_sql(last_block_num.saturating_sub(1))?, values)
+        }
```
Mirko-von-Leipzig: I think the original code had the correct idea, but didn't realise that the sql value was signed integer.

This has come up a few time, might be worth adding `saturating_sub` and friends to block number @PhilippGackstatter?


```suggestion
(BlockNumber::from_raw_sql(last_block_num)?.parent().unwrap_or_default(), values)
```

mmagician: The latest commit e0a7ba8 takes a different approach now and derives `last_block_included` from the last kept row:

```rust
let kept: Vec<_> = raw.into_iter()
    .take_while(|(bn, ..)| *bn != last_block_num)
    .collect();

let last_block_included = match kept.last() {
    Some(&(bn, ..)) => BlockNumber::from_raw_sql(bn)?,
    None => *block_range.start(),
};
```

The `block_num - 1` approach assumed the previous block number is meaningful - but after `take_while`, we already have the actual rows we're keeping, so we can just read the block number directly and avoid any arithmetic on block numbers altogether.

---

(independently adding a `saturating_sub` method to `BlockNumber` can still be useful in general).

bobbinth: Doesn't the latest approach require two allocations vs. just one allocation with the original approach?

mmagician: Ah yes, good catch, this ends up with one allocation for raw values and a second for `StorageMapValue`. But they can be collapsed into a single allocation - fixed in a6016ed

crates/store/src/db/models/queries/accounts.rs
```diff
@@ -716,22 +716,26 @@ pub(crate) fn select_account_storage_map_values_paged(
...
             .collect::<Result<Vec<_>, DatabaseError>>()?;
 
-        (BlockNumber::from_raw_sql(last_block_num.saturating_sub(1))?, values)
+        let last_block_included = match values.last() {
+            Some(v) => v.block_num,
+            // All rows are in the same block and exceed the limit.
+            // Return the range start to signal no progress was made,
+            // which the caller interprets as limit_exceeded.
+            None => *block_range.start(),
+        };
```
Mirko-von-Leipzig:
```suggestion
        let last_block_included = values.last().unwrap_or_else(|| *block_range.start());
```

Issue #1242, opened 2025-09-25
Express mempool constraints ito fees


Mirko-von-Leipzig:
The mempool has a batch and block budget which it uses to constrain the content of the batches and blocks it  proposes. Currently, this is expressed in absolute note, transaction and account update amounts but in the future it likely makes more sense to consider them similarly to fee structure.

> Not related to this PR, but I wonder if we should express some of these in terms that are closer to how we think about fees.
>
>For example, instead of the maximum number of transactions, maybe it should be the maximum number of "verification units" (which we currently compute as `log2(num_tx_cycles)`. So, a simple transaction with Falcon signature would be 17 units, and a much more complicated transaction that requires 1M cycles, would be 20 units. And then if we set the budget at 250 verification units, we'd be able to include 15 - 16 regular transactions in a batch, or 11 - 12 "big" transactions etc.
>
> Similar could probably be applied to account updates and input notes - though, this would require more thinking.
>
>cc @PhilippGackstatter.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-node/pull/1234#discussion_r2378196519_

PR #1820, opened 2026-03-23
refactor(mempool): revert to separate DAGs

### LINKED ISSUE #537: [block-producer]: improve `InflightState` error handling

### LINKED ISSUE #1253: Consider erasing data in proposed mempool nodes
https://github.com/0xMiden/miden-node/pull/1234#discussion_r2381394144

The `ProposedBatchNode` and `BlockNode` don't perform any aggregation - that is, they don't erase internal notes, nor aggregate account updates. They only delegate the state their transactions/batches create.

This means that a proposed batch and proven batch differ in the state they emit since the latter erases internal-only state. This has no material impact; but does this can be unexpected since it means a proposed batch can be dependent on itself (which we account for in the children/parent functions).

We can avoid this by performing the aggregation ourselves.

### LINKED ISSUE #1439: Improve mempool graph replacement API

Mirko-von-Leipzig:
This PR effectively rewrites the mempool implementation (again). The result is a combination of the first implementation and the current implementation, hopefully resulting in a best of both.

#### Background

The former was simple -- separate graphs for txs and batches, with nodes moving from one to the next as they are selected. It also had a singular inflight state which was checked when txs were added, but never after again. This meant it was had complex internal transitions which were never checked as nodes moved between graphs. However, within a single graph things were simple, and the graph impl could be shared.

The current implementation has a single graph, with transaction nodes being folded into batch nodes, which fold into block nodes. This graph impl is much more complex, in particular the fold and unfold functions are difficult to grok and test, in part due to pass-through nodes causing DAG properties to no longer hold. The positive is that state cannot really be corrupted since there aren't multiple graphs to make mistakes between.

#### Why

The driving motivation was to do something about `fold` and `unfold`. I couldn't convince myself that the implementations were ever 100% correct, nor that they would survive changes over time.

Eventually I realised that folding and unfolding could be achieved by first _reverting_ all the node subgraphs, and then re-inserting with the new replacement nodes. Reverting and inserting are simple (relatively speaking), and we can lean on the existing insertion and reverting checks as additional assertions during this process.

I then realised there wasn't much need for a single graph anymore, since the original motivation was to support folding and unfolding. _However_ we can use the above to add additional safety to the underlying graph by enforcing that nodes can only ever be reverted if they're leaves, and pruned if they're roots aka either no parents _or_ no children remain in the graph.

#### This PR

A single graph type is created, which holds its state, and edges based on that state. Nodes can only be appended, reverted (removed from the top/leaves), or pruned (removed from the bottom/roots). Nodes can be marked as selected (to mark a tx/batch as included in batch/block).

Transaction and batch graphs wrap this underlying graph type. This is _very_ similar to the original implementation, with some improvements and additional state checks due to the lessons learnt from the current implementation.

The main `Mempool` functions should be much easier to read (imo). This implementation is a pure refactor - user batches, and tx reverting improvement strategies will be in follow-up PRs so they can be reviewed without this huge diff.

#### How to review

There is nearly no current code that applies ito graph. They're also in separate files (which is a positive imo).

I suggest:

- Review `mempool`; since this business level logic shouldn't change
  - It should be correct wrt to the claims made in the batch and transaction top-level graphs.
- Review the state and edge files, then merge that logically with the assumptions made in the graph impl.

-----------

Closes #1439, closes #1253, closes #537

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Introduce `SyncChainMmr` RPC endpoint to sync chain MMR deltas within specified block ranges ([#1591](https://github.com/0xMiden/node/issues/1591)).
 - Fixed `TransactionHeader` serialization for row insertion on database & fixed transaction cursor on retrievals ([#1701](https://github.com/0xMiden/node/issues/1701)).
 - Added KMS signing support in validator ([#1677](https://github.com/0xMiden/node/pull/1677)).
+- Added finality field for `SyncChainMmr` requests ([#1725](https://github.com/0xMiden/miden-node/pull/1725)).
 - Added per-IP gRPC rate limiting across services as well as global concurrent connection limit ([#1746](https://github.com/0xMiden/node/issues/1746)).
 - Added limit to execution cycles for a transaction network, configurable through CLI args (`--ntx-builder.max-tx-cycles`) ([#1801](https://github.com/0xMiden/node/issues/1801)).
 - Added monitor version and network name to the network monitor dashboard, network name is configurable via `--network-name` / `MIDEN_MONITOR_NETWORK_NAME` ([#1838](https://github.com/0xMiden/node/pull/1838)).
@@ -51,6 +52,8 @@
 - Fixed `bundled bootstrap` requiring `--validator.key.hex` or `--validator.key.kms-id` despite a default key being configured ([#1732](https://github.com/0xMiden/node/pull/1732)).
 - Fixed incorrectly classifying private notes with the network attachment as network notes ([#1378](https://github.com/0xMiden/node/pull/1738)).
 - Fixed accept header version negotiation rejecting all pre-release versions; pre-release label matching is now lenient, accepting any numeric suffix within the same label (e.g. `alpha.3` accepts `alpha.1`) ([#1755](https://github.com/0xMiden/node/pull/1755)).
+- Fixed `GetAccount` returning an internal error for `AllEntries` requests on storage maps where all entries are in a single block (e.g. genesis accounts) ([#1816](https://github.com/0xMiden/node/pull/1816)).
+- Fixed `GetAccount` returning empty storage map entries instead of `too_many_entries` when a genesis account's map exceeds the pagination limit ([#1816](https://github.com/0xMiden/node/pull/1816)).
 
 ## v0.13.8 (2026-03-12)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,27 +143,12 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
-dependencies = [
- "anstyle",
- "anstyle-parse 0.2.7",
- "anstyle-query",
- "anstyle-wincon",
- "colorchoice",
- "is_terminal_polyfill",
- "utf8parse",
-]
-
-[[package]]
-name = "anstream"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
- "anstyle-parse 1.0.0",
+ "anstyle-parse",
  "anstyle-query",
  "anstyle-wincon",
  "colorchoice",
@@ -176,15 +161,6 @@ name = "anstyle"
 version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
-
-[[package]]
-name = "anstyle-parse"
-version = "0.2.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4e7644824f0aa2c7b9384579234ef10eb7efb6a0deb83f9630a49594dd9c15c2"
-dependencies = [
- "utf8parse",
-]
 
 [[package]]
 name = "anstyle-parse"
@@ -915,9 +891,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.2.57"
+version = "1.2.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a0dd1ca384932ff3641c8718a02769f1698e7563dc6974ffd03346116310423"
+checksum = "e1e928d4b69e3077709075a938a05ffbedfa53a84c8f766efbf8220bb1ff60e1"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
@@ -1054,7 +1030,7 @@ version = "4.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "714a53001bf66416adb0e2ef5ac857140e7dc3a0c48fb28b2f10762fc4b5069f"
 dependencies = [
- "anstream 1.0.0",
+ "anstream",
  "anstyle",
  "clap_lex",
  "strsim",
@@ -1080,9 +1056,9 @@ checksum = "c8d4a3bb8b1e0c1050499d1815f5ab16d04f0959b233085fb31653fbfc9d98f9"
 
 [[package]]
 name = "cmake"
-version = "0.1.57"
+version = "0.1.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "75443c44cd6b379beb8c5b45d85d0773baf31cce901fe7bb252f4eff3008ef7d"
+checksum = "c0f78a02292a74a88ac736019ab962ece0bc380e3f977bf72e376c5d78ff0678"
 dependencies = [
  "cc",
 ]
@@ -1614,9 +1590,9 @@ dependencies = [
 
 [[package]]
 name = "env_filter"
-version = "1.0.0"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a1c3cc8e57274ec99de65301228b537f1e4eedc1b8e0f9411c6caac8ae7308f"
+checksum = "32e90c2accc4b07a8456ea0debdc2e7587bdd890680d71173a15d4ae604f6eef"
 dependencies = [
  "log",
  "regex",
@@ -1624,11 +1600,11 @@ dependencies = [
 
 [[package]]
 name = "env_logger"
-version = "0.11.9"
+version = "0.11.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b2daee4ea451f429a58296525ddf28b45a3b64f1acf6587e2067437bb11e218d"
+checksum = "0621c04f2196ac3f488dd583365b9c09be011a4ab8b9f37248ffcc8f6198b56a"
 dependencies = [
- "anstream 0.6.21",
+ "anstream",
  "anstyle",
  "env_filter",
  "jiff",
@@ -2440,9 +2416,9 @@ checksum = "d98f6fed1fde3f8c21bc40a1abb88dd75e67924f9cffc3ef95607bad8017f8e2"
 
 [[package]]
 name = "iri-string"
-version = "0.7.10"
+version = "0.7.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c91338f0783edbd6195decb37bae672fd3b165faffb89bf7b9e6942f8b1a731a"
+checksum = "d8e7418f59cc01c88316161279a7f665217ae316b388e58a0d10e29f54f1e5eb"
 dependencies = [
  "memchr",
  "serde",
@@ -2517,7 +2493,7 @@ dependencies = [
  "cesu8",
  "cfg-if",
  "combine",
- "jni-sys",
+ "jni-sys 0.3.1",
  "log",
  "thiserror 1.0.69",
  "walkdir",
@@ -2526,9 +2502,31 @@ dependencies = [
 
 [[package]]
 name = "jni-sys"
-version = "0.3.0"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8eaf4bc02d17cbdd7ff4c7438cafcdf7fb9a4613313ad11b4f8fefe7d3fa0130"
+checksum = "41a652e1f9b6e0275df1f15b32661cf0d4b78d4d87ddec5e0c3c20f097433258"
+dependencies = [
+ "jni-sys 0.4.1",
+]
+
+[[package]]
+name = "jni-sys"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c6377a88cb3910bee9b0fa88d4f42e1d2da8e79915598f65fb0c7ee14c878af2"
+dependencies = [
+ "jni-sys-macros",
+]
+
+[[package]]
+name = "jni-sys-macros"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "38c0b942f458fe50cdac086d2f946512305e5631e720728f2a61aabcd47a6264"
+dependencies = [
+ "quote",
+ "syn 2.0.117",
+]
 
 [[package]]
 name = "jobserver"
@@ -3325,7 +3323,6 @@ dependencies = [
  "diesel",
  "diesel_migrations",
  "fs-err",
- "futures",
  "hex",
  "indexmap",
  "libsqlite3-sys",
@@ -3351,7 +3348,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tokio-stream",
- "toml 1.0.7+spec-1.1.0",
+ "toml 1.1.0+spec-1.1.0",
  "tonic",
  "tonic-reflection",
  "tower-http",
@@ -3492,7 +3489,7 @@ dependencies = [
  "semver 1.0.27",
  "serde",
  "thiserror 2.0.18",
- "toml 1.0.7+spec-1.1.0",
+ "toml 1.1.0+spec-1.1.0",
  "walkdir",
 ]
 
@@ -3903,9 +3900,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.2.0"
+version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf97ec579c3c42f953ef76dbf8d55ac91fb219dde70e49aa4a6b7d74e9919050"
+checksum = "c6673768db2d862beb9b39a78fdcb1a69439615d5794a1be50caa9bc92c81967"
 
 [[package]]
 name = "num-derive"
@@ -4674,9 +4671,9 @@ dependencies = [
 
 [[package]]
 name = "proptest"
-version = "1.10.0"
+version = "1.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "37566cb3fdacef14c0737f9546df7cfeadbfbc9fef10991038bf5015d0c80532"
+checksum = "4b45fcc2344c680f5025fe57779faef368840d0bd1f42f216291f0dc4ace4744"
 dependencies = [
  "bit-set",
  "bit-vec",
@@ -5565,9 +5562,9 @@ dependencies = [
 
 [[package]]
 name = "serde_spanned"
-version = "1.0.4"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f8bbf91e5a4d6315eee45e704372590b30e260ee83af6639d64557f51b067776"
+checksum = "876ac351060d4f882bb1032b6369eb0aef79ad9df1ea8bc404874d8cc3d0cd98"
 dependencies = [
  "serde_core",
 ]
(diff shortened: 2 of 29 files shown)


crates/block-producer/src/mempool/mod.rs
```diff
@@ -533,50 +365,31 @@ impl Mempool {
...
         //
         // This is done to prevent a system bug from causing repeated failures if we keep retrying
         // the same transactions. Since we can't trivially identify the cause of the block
         // failure, we take the safe route and nuke all associated state.
         //
         // A more refined approach could be to tag the offending transactions and then evict them
         // once a certain failure threshold has been met.
-        let reverted = self.revert_subtree(NodeId::Block(block));
         let mut reverted_txs = HashSet::default();
+        let (_, batches) = self.pending_block.take().unwrap();
```
sergerad: nit: expect

crates/block-producer/src/mempool/mod.rs
```diff
@@ -533,50 +365,31 @@ impl Mempool {
...
+        if self.pending_block.as_ref().is_none_or(|(num, _)| num != &block) {
             return;
         }
 
-        // Remove all descendents _without_ reinserting the transactions.
+        // Remove all descendants _without_ reinserting the transactions.
         //
         // This is done to prevent a system bug from causing repeated failures if we keep retrying
         // the same transactions. Since we can't trivially identify the cause of the block
         // failure, we take the safe route and nuke all associated state.
```
sergerad, line 386: Just wondering if this is actually a trade-off we want to make. There could be a situations where a transient error gets us here and then users need to resubmit transactions right? Which would be a confusing / unexpected UX?

I also don't follow why the `rollback_batch()` requeues transactions while this doesn't. Why does that make sense and what is the intention?

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.inner.append(tx)
+    }
+
+    pub fn select_batch(&mut self, mut budget: BatchBudget) -> Option<SelectedBatch> {
+        let mut selected = SelectedBatch::builder();
+
+        while let Some((id, tx)) = self.inner.selection_candidates().pop_first() {
+            if budget.check_then_subtract(tx) == BudgetStatus::Exceeded {
+                break;
+            }
```
sergerad, line 86: Would it be possible to order tx candidates by size so that a single fat tx doesn't block a series of smaller ones unnecessarily? Unsure if this is a real problem.

crates/block-producer/src/mempool/graph/dag.rs
```diff
@@ -0,0 +1,349 @@
...
+    nodes: HashMap<N::Id, N>,
+    /// The aggregate state of all nodes in the graph.
+    state: State<N::Id>,
+    /// The relation between the nodes formed by their dependencies on each others state.
+    edges: Edges<N::Id>,
+    /// Nodes that have been selected.
+    selected: HashSet<N::Id>,
+    /// Nodes that are available for selection.
+    ///
+    /// These are nodes who's parents have all been selected.
```
sergerad:
```suggestion
    /// These are nodes whose parents have all been selected.
```

crates/block-producer/src/mempool/graph/dag.rs
```diff
@@ -0,0 +1,349 @@
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
+use std::hash::Hash;
```
sergerad, line 2: nit: there are instances of both `Hash` and `std::hash::Hash` below.

crates/block-producer/src/mempool/graph/batch.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub fn submit_proof(&mut self, proof: Arc<ProvenBatch>) {
+        if self.inner.contains(&proof.id()) {
+            self.proven.insert(proof.id(), proof);
+        }
+    }
+
+    pub fn select_block(&mut self, mut budget: BlockBudget) -> Vec<Arc<ProvenBatch>> {
+        let mut selected = Vec::default();
+
+        // Only root's which are proven can be selected for inclusion in a block.
```
sergerad:
```suggestion
        // Only roots which are proven can be selected for inclusion in a block.
```

crates/block-producer/src/mempool/graph/batch.rs
```diff
@@ -0,0 +1,142 @@
...
+        self.inner.prune(batch);
+        self.proven.remove(&batch);
+    }
+
+    pub fn proven_count(&self) -> usize {
+        self.proven.len()
+    }
+
+    pub fn proposed_count(&self) -> usize {
+        self.inner.node_count() - self.proven_count()
```
sergerad: `.saturating_sub()`?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -617,121 +430,78 @@ impl Mempool {
...
         let span = tracing::Span::current();
 
-        self.nodes.inject_telemetry(&span);
-        self.state.inject_telemetry(&span);
+        let committed_txs = self
+            .committed_blocks
+            .iter()
+            .flat_map(|block| block.iter())
+            .map(|batch| batch.transactions().as_slice().len())
+            .sum::<usize>();
```
sergerad, line 451: Don't think its a problem because O(blocks*batches) is fine but just noting that this `inject_telemetry` function appears to be called frequently.

crates/block-producer/src/mempool/graph/dag.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    /// Returns the set of nodes which can be selected.
+    ///
+    /// Candidates are nodes that are not currently selected, have all parents selected, and can be
+    /// handed directly to [`select_candidate`](Self::select_candidate).
+    pub fn selection_candidates(&self) -> BTreeMap<&N::Id, &N> {
+        self.selection_candidates
+            .iter()
+            .map(|id| (id, self.nodes.get(id).unwrap()))
```
sergerad: nit: expect

crates/block-producer/src/mempool/graph/dag.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// # Panics
+    ///
+    /// Panics if the node does not exist or if the graph invariants (such as acyclicity) are
+    /// violated while unwinding descendants. The latter indicates graph corruption.
+    pub fn revert_node_and_descendants(&mut self, id: N::Id) -> Vec<N> {
+        let mut descendants = self.descendants(&id);
+
+        let mut reverted = Vec::new();
+        'outer: while !descendants.is_empty() {
+            for id in descendants.iter().copied() {
```
sergerad, line 188: Might be worth noting this fn is O(n^2)

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: Its an arbitrary decision, and we could do either. At the time we decided that block failure was much worse and should be avoided at all costs, whereas batches would be more likely to shuffle the same transactions in different orders.

We do have #594 to address this, and I'm opening a PR hopefully today that basically gives each transactions some number of allowed failures before they are nuked.

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: This is a concern; will be addressed once fees are more realistic. In that case we would prioritise txs based on fees vs effort (or some other strategy).

#1242 is also related to this.

Mirko-von-Leipzig: Consider `pop_first` a placeholder for a more complex selection strategy

crates/block-producer/src/mempool/graph/batch.rs
Mirko-von-Leipzig: Hmm. This would constitute a bug so I've made a checked + expect.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: I'm trying to avoid as much additional book keeping as possible; to ensure we don't desync somewhere. This could be replaced by a counter which is incremented and decremented appropriately though.

Might be worth doing at some point.

crates/block-producer/src/mempool/graph/dag.rs
Mirko-von-Leipzig: We can improve this by tracking the insertion order and reverting in reverse. However it is quite a bit more code to maintain, and might never actually be faster until we are huge.

Will add a comment to this effect.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: See #1832 which reverts transactions only after they've received four infractions.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -203,56 +219,27 @@ impl Mempool {
...
+        clippy::needless_pass_by_value,
+        reason = "Not impactful, and we may want ownership in the future"
+    )]
     #[instrument(target = COMPONENT, name = "mempool.add_transaction", skip_all, fields(tx=%tx.id()))]
     pub fn add_transaction(
         &mut self,
         tx: Arc<AuthenticatedTransaction>,
     ) -> Result<BlockNumber, AddTransactionError> {
-        if self.nodes.uncommitted_tx_count() >= self.config.tx_capacity.get() {
+        if self.unbatched_transactions_count() >= self.config.tx_capacity.get() {
```
igamigo, line 234: `unbatched_transactions_count()` here yields a different result than what `uncommitted_tx_count()` used to return right? It used to be

```rust
    pub(super) fn uncommitted_tx_count(&self) -> usize {
        self.txs.len()
            + self.proposed_batches.values().map(|b| b.0.txs().len()).sum::<usize>()
            + self.proven_batches.values().map(|b| b.txs.len()).sum::<usize>()
            + self.proposed_block.as_ref().map(|b| b.1.txs.len()).unwrap_or_default()
    }
```

which meant that selected transactions (but not yet committed) were counting toward the limit, whereas now, once hte transactions get selected into a batch they don't count anymore.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -265,132 +252,30 @@ impl Mempool {
...
     /// Drops the proposed batch and all of its descendants.
     ///
     /// Transactions are re-queued.
     #[instrument(target = COMPONENT, name = "mempool.rollback_batch", skip_all)]
     pub fn rollback_batch(&mut self, batch: BatchId) {
-        // Due to the distributed nature of the system, its possible that a proposed batch was
-        // already proven, or already reverted. This guards against this eventuality.
-        if !self.nodes.proposed_batches.contains_key(&batch) {
-            return;
-        }
```
igamigo, line 352: It seems like a late rollback could still arrive (I think?), but nothing is done about it in the new version (though I might be missing something). Could there be some scenario where a transaction gets pushed into a pending block which gets committed, and a late rollback re-queues the tx?

crates/block-producer/src/mempool/graph/batch.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// Reverts expired batches and their descendants.
+    ///
+    /// Only unselected batches are considered, the assumption being that selected batches
+    /// are in committed blocks and should not be reverted.
+    ///
+    /// Batches are returned in reverse-chronological order.
+    pub fn revert_expired(&mut self, chain_tip: BlockNumber) -> Vec<SelectedBatch> {
+        self.inner.revert_expired_unselected(chain_tip)
+    }
```
igamigo, line 108: Do you need to prune from `self.proven` here?

igamigo (commented):
Leaving some comments. The logic looks great and I agree that the `Mempool` code is quite nicer and easier to read. But comparing it to the previous implementation I did find some of the previous models to be more explicit in terms of the state of things in their lifecycles (eg, `ProvenBatchNode`, `ProposedBatchNode`) and this seems a bit more loose now (a batch is in the the batch graph but also may be within `proven` and in the pending block etc.)

crates/block-producer/src/mempool/budget.rs
```diff
@@ -18,6 +18,9 @@ pub struct BatchBudget {
     /// Maximum number of output notes allowed.
     pub output_notes: usize,
     /// Maximum number of updated accounts.
+    ///
+    /// Authenticated transactions are assumed to update at most one account; this field enforces
+    /// how many such single-account updates can fit into a batch.
     pub accounts: usize,
```
bobbinth: Why do we say "Authenticated transactions" here? Isn't this true for all transactions?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -142,28 +158,26 @@ impl SharedMempool {
...
+    transactions: graph::TransactionGraph,
+    /// Tracks the dependency graph for batches awaiting inclusion in a block.
+    batches: graph::BatchGraph,
+    /// The block currently being built, if any.
+    pending_block: Option<(BlockNumber, Vec<Arc<ProvenBatch>>)>,
+    /// The most recently committed blocks in chronological order.
+    ///
+    /// Limited to the state retention amount defined in the config. Once a pending block is
+    /// committed it is appended here, and the oldest block's state is pruned.
+    committed_blocks: VecDeque<Vec<Arc<ProvenBatch>>>,
```
bobbinth: nit: it may make senes to introduce newtype to warp `Vec<Arc<ProvenBatch>>` to improve readability a bit.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -478,32 +312,32 @@ impl Mempool {
...
             .expect("block must be in progress to commit");
-        let tx_ids = block.1.transactions().map(|tx| tx.id()).collect();
+        let tx_ids = batches
+            .iter()
+            .flat_map(|batch| batch.transactions().as_slice().iter())
+            .map(miden_protocol::transaction::TransactionHeader::id)
+            .collect();
 
-        self.nodes.committed_blocks.push_back(block);
         self.chain_tip = self.chain_tip.child();
```
bobbinth, line 348: Should we assert somewhere that `self.chain_tip.child() == to_commit.block_num()`?

Also, not from this PR, but I find `to_commit` name a bit confusion. Maybe just `block_header` would be more clear?

crates/block-producer/src/mempool/mod.rs
bobbinth: I think this comment may be outdated.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -742,8 +512,10 @@ impl Mempool {
         &self,
         authentication_height: BlockNumber,
     ) -> Result<(), AddTransactionError> {
-        let oldest = self.nodes.oldest_committed_block().unwrap_or_default();
-        let limit = oldest.parent().unwrap_or_default();
+        let limit = self
+            .chain_tip
+            .checked_sub(self.committed_blocks.len() as u32)
+            .expect("amount of committed blocks cannot exceed the chain tip");
```
bobbinth: typo: "amount of committed blocks..." should probably be "number of committed blocks...".

crates/block-producer/src/mempool/graph/node.rs
bobbinth, line 1: nit: I'd add section separators in this file (like we have in other files).

crates/block-producer/src/mempool/graph/edges.rs
bobbinth, line 1: nit: I'd add section separators in this file (like we have in other files).

crates/block-producer/src/mempool/graph/dag.rs
bobbinth, line 1: nit: I'd add section separators in this file (like we have in other files).

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+        let selected = selected.build();
+
+        Some(selected)
+    }
+
+    /// Reverts expired transactions and their descendants.
+    ///
+    /// Only unselected transactions are considered; selected transactions are assumed to be in
+    /// committed blocks and should not be reverted.
```
bobbinth, line 103: It may be worth to clarify why we make such an assumption: my understanding is that this happens because we first revert expired batches, and only then revert expired transactions. This means that there should be no transactions in expired batches by the time we get here.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: Ah yes you're right I think. I'll add some tests covering this.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: It does, I'm not sure its terribly important - in both cases I just picked the one that was readily available. But I can update this to match the original.

crates/block-producer/src/mempool/graph/batch.rs
Mirko-von-Leipzig: Yes, thank you. Ideally we would have a single point of removal but that's turned out to be difficult.

crates/block-producer/src/mempool/graph/batch.rs
bobbinth, line 1: nit: I'd add section separators in this file (like we have in other files).

crates/block-producer/src/mempool/graph/batch.rs
```diff
@@ -0,0 +1,156 @@
...
+            self.proven.insert(proof.id(), proof);
+        }
+    }
+
+    /// Returns `true` if the batch has been proven previously.
+    pub fn is_proven(&mut self, batch: &BatchId) -> bool {
+        self.proven.contains_key(batch)
+    }
+
+    pub fn select_block(&mut self, mut budget: BlockBudget) -> Vec<Arc<ProvenBatch>> {
```
bobbinth, line 129: Let's add brief doc comments here.

crates/block-producer/src/mempool/budget.rs
Mirko-von-Leipzig: I asked an LLM to expand some doc comments, didn't notice it went here as well :) I'll revert.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: > Should we assert somewhere that self.chain_tip.child() == to_commit.block_num()?

This is done sort of indirectly because `pending_block` is always `self.chain_tip.child()`, and that's what we check in the `take_if`. I'll add the additional assert though, doesn't hurt.

Also renamed.

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: I always get confused :D

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: The main reason we need to assume this, is because we don't actually handle _committed_ transactions - we only insert, revert, or prune them. And the latter only occurs once the transaction's block itself is pruned.

So if we revert selected transactions (which are probably in committed blocks), we'd be corrupting the state. I'll add this to the explanation here.

Mirko-von-Leipzig: Oh, apologies, I see what you mean. Yes you're right, uncommitted but selected transactions are deselected as part of the expired batches which runs prior to this being called. I'll add that snippet as well, as a pre-requisite.

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        self.inner
+            .revert_node_and_descendants(transaction)
+            .into_iter()
+            .map(|tx| tx.id())
+            .collect()
+    }
+
+    /// Marks the batch's transactions are ready for selection again.
```
bobbinth, line 144: Potential typo: was "are ready" supposed to be "as ready"?

PR #1826, opened 2026-03-23
fix: network transaction executor + network monitor

igamigo:
This PR contains 2 fixes:

- on the NTX builder, creates the executor with a valid core trace fragment size
- updates the monitor code for the new Word stack orientation changes in the new VM revision

Found while running client integration tests:
```
thread 'tokio-rt-worker' (15182569) panicked at .cargo/git/checkouts/miden-node-98c06b5d084e5405/b59b408/crates/ntx-builder/src/actor/execute.rs:143:18:
max_cycles should be within valid range: CoreTraceFragmentSizeTooSmall
```

### DIFF
--- a/bin/network-monitor/src/assets/counter_program.masm
+++ b/bin/network-monitor/src/assets/counter_program.masm
@@ -1,6 +1,6 @@
 # Counter program for network monitoring with note authentication
 # Storage layout:
-# - OWNER_SLOT: authorized wallet account id as [prefix, suffix, 0, 0]
+# - OWNER_SLOT: authorized wallet account id as [suffix, prefix, 0, 0]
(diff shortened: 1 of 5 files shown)


bin/network-monitor/src/assets/counter_program.masm
Mirko-von-Leipzig, line 1: Could we avoid this by writing it in rust instead of masm?

That would also make it easier to maintain for people like me that don't know `masm` very well.

igamigo: Yeah, I'm not familiar with the compiler but I thought of this as well as a possible improvement. I'm not sure if there are any drawbacks to the approach other than some potentially wasted cycles but I doubt that's prohibitive. I'll open an issue to track this

PR #1832, opened 2026-03-24
feat(mempool): revert transactions after 3 failures

### LINKED ISSUE #594: [mempool]: improve tx reverting strategy

Mirko-von-Leipzig:
This PR changes the mempool behaviour when rolling back blocks and batches. Transactions are now reverted once they have participated in more than three failed batches and blocks.

The current behaviour is that transactions are **never** reverted for batch failures, and **always** reverted for block failures.

The intention here is to weed out transactions which _cause_ these failures e.g. due to a bug in the prover/mempool/node. The new implementation is a bit more forgiving and gives innocent bystander transactions a couple of chances to avoid the buggy one.

The limit of three is arbitrary, but I also don't see a good reason to make this more configurable. Fight me :)

This PR builds on the mempool refactor in #1820. 

Closes #594


crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -122,24 +134,59 @@ impl TransactionGraph {
...
+        &mut self,
+        txs: impl Iterator<Item = TransactionId>,
+    ) -> HashSet<TransactionId> {
+        let mut to_revert = Vec::default();
+
+        for tx in txs {
+            let count = self.failures.entry(tx).or_default();
+            *count += 1;
+
+            if *count > Self::FAILURE_LIMIT {
```
sergerad:
```suggestion
            if *count >= Self::FAILURE_LIMIT {
```

crates/block-producer/src/mempool/mod.rs
```diff
@@ -369,25 +380,21 @@ impl Mempool {
...
             for batch in reverted {
-                for tx in batch.into_transactions() {
-                    reverted_txs.extend(self.transactions.revert_tx_and_descendants(tx.id()));
-                }
+                self.transactions.requeue_transactions(&batch);
             }
         }
+        let failed_txs = block
+            .iter()
+            .flat_map(|batch| batch.transactions().as_slice().iter().map(TransactionHeader::id));
```
sergerad, line 407: Would it be possible / desirable for us to pinpoint which transactions actually failed rather than treating all as failed?

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: I think its correct as `>` no? You're allowed to fail `N` times. Though I guess title and docs sort of conflict :D

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: iiuc this is supposed to only occur if there is a bug in the proving system, or one of the kernels, or the batch was invalid. As in, this is a really unexpected outcome once we are at this stage.

But maybe there is some information that could be surfaced. cc @PhilippGackstatter

PhilippGackstatter: I'm missing / have forgotten the context in what scenarios we get here, could you remind me?

Mirko-von-Leipzig: We select transactions for inclusion in a batch, and select batches for inclusion in a block. The mempool ensures that this selection process is coherent and that the state _should_ be in order.

However, in the scenario where a batch or block fails, we want to prevent a buggy transaction from halting the network e.g. if it causes every batch/block it is in to fail then we want to evict that transaction and investigate wtf later.

How or why this could happen is.. likely only due to an internal bug. I think either in:
- mempool selection / DAG
- batch kernel
- block kernel
- node database corruption

Since we cannot (trivially? at all?) pinpoint which exact transaction, or set thereof, triggered this, the mempool assigns each participating transaction a strike, and after three strikes they're evicted.

PhilippGackstatter: Thanks!

> Since we cannot (trivially? at all?) pinpoint which exact transaction, or set thereof, triggered this, the mempool assigns each participating transaction a strike, and after three strikes they're evicted.

Yeah, I agree. I think each transaction in the mempool is internally consistent and valid from the protocol perspective, so we can indeed not really pinpoint which transaction was the issue by looking at them. So, I think the current strategy is coarse but the best thing we can do - afaict.

crates/block-producer/src/mempool/graph/batch.rs
```diff
@@ -92,7 +92,14 @@ impl BatchGraph {
     ///
     /// Batches are returned in reverse-chronological order.
     pub fn revert_expired(&mut self, chain_tip: BlockNumber) -> Vec<SelectedBatch> {
-        self.inner.revert_expired_unselected(chain_tip)
+        let expired = self.inner.expired_and_unselected(chain_tip);
+        let mut reverted = Vec::new();
+
+        for tx in expired {
```
igamigo: naming nit: `tx` is a batch here, right?

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -70,9 +70,17 @@ impl GraphNode for Arc<AuthenticatedTransaction> {
 #[derive(Clone, Debug, PartialEq, Default)]
 pub struct TransactionGraph {
     inner: Graph<Arc<AuthenticatedTransaction>>,
+    /// The number of failures a transaction has participated in.
+    ///
+    /// These are batch or block proving errors in which the transaction was a part of. This is
+    /// used to identify potentially buggy transactions that should be evicted.
+    failures: HashMap<TransactionId, u32>,
```
igamigo, line 77: Does this leak? I couldn't find where it would get cleared for a transaction that fails and then succeeds

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -70,9 +70,17 @@ impl GraphNode for Arc<AuthenticatedTransaction> {
...
 }
 
 impl TransactionGraph {
+    /// The maximum number of times a transaction is allowed to fail before being evicted.
```
igamigo: nit: As already mentioned, this is a bit conflicting as it can technically fail 4 times right?

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: You're right, I needed to also clear it in `prune`. Bit painful ito book keeping with the multiple mutation methods.

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: I've updated code and comment to be less ambiguous. Now reverts after third failure.

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: I've corrected this, and the comment to be less ambiguous. Txs are reverted after third failure.

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -102,19 +110,7 @@ impl TransactionGraph {
...
     /// Returns the identifiers of transactions that were removed from the graph.
-    ///
-    /// # Note
-    ///
-    /// Since this _ignores_ selected transactions, and the purpose is to revert expired
-    /// transactions after a block is committed, the caller **must** ensure that selected
-    /// transactions from expired batches (and therefore not committed) are deselected
-    /// _before_ calling this function. i.e. first revert expired batches and deselect their
-    /// transactions, then call this.
     pub fn revert_expired(&mut self, chain_tip: BlockNumber) -> HashSet<TransactionId> {
```
bobbinth, line 118: I think these comments may still be relevant. Were they removed on purpose?

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -134,24 +130,59 @@ impl TransactionGraph {
...
 
+    /// Increments each transaction's failure counter, and reverts transactions which exceed the
+    /// failure limit.
+    ///
+    /// This weeds out transactions which participate in batch and block failures, and might be the
+    /// root cause.
+    pub fn increment_failure_count(
+        &mut self,
+        txs: impl Iterator<Item = TransactionId>,
+    ) -> HashSet<TransactionId> {
```
bobbinth, line 166: It may be good to mention what is returned from this method (I think it is a set of reverted transactions).

crates/block-producer/src/mempool/mod.rs
bobbinth, line 268: We should probably update the doc comments here to mention that for the provided batch we also increment tx failure count, and for the transactions which have failed more times than the allowed threshold, we also revert these transactions. I believe this would affect only the transactions of this batch, so, the net effect may be that not all of the batch transactions are re-queued.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -277,9 +277,20 @@ impl Mempool {
...
-        for reverted in reverted_batches {
+        for reverted in &reverted_batches {
             self.transactions.requeue_transactions(reverted);
         }
+
+        // Find rolled back batch to mark its txs as failed.
+        //
+        // Note that its possible it doesn't exist, since this batch could have already been
+        // reverted as part of a separate rollback.
+        if let Some(batch) = reverted_batches.iter().find(|reverted| reverted.id() == batch) {
```
bobbinth, line 288: typo: "that its possible" -> "that it's possible".

But also, it may be worth expanding this comment a bit. It is not immediately obvious why there could be a separate rollback.

crates/block-producer/src/mempool/mod.rs
bobbinth, line 383: Is the doc comment here fully accurate? Specifically, I don't think we remove all the block content from the pool now - but rather increment failure count for all transactions that were in the block, and then only remove the transactions that exceeded the failure threshold.

Also, I'm not sure if the panic condition here is described correctly: AFAICT, if the specified block is not in flight, we just return (but maybe we should panic indeed?)

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: Fixed in #1846

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: Addressed in #1846

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: Fixed in #1846

Mirko-von-Leipzig: Addressed in #1846

PR #1834, opened 2026-03-25
chore: Add validator unit tests

sergerad:


crates/validator/src/server/tests.rs
```diff
@@ -0,0 +1,409 @@
...
+/// `NoPrevBlockHeader`.
+///
+/// This covers the genesis replacement case: genesis (block 0) has no parent, so
+/// `chain_tip.block_num().parent()` returns `None` and the replacement is rejected.
+/// Since `ProposedBlock::new` cannot produce `block_num` 0, we simulate the scenario by
+/// building a block at height 1 (`proposed_block_num` == `chain_tip block_num`), and using a
+/// database that does not contain the parent (genesis). This triggers the same
+/// `NoPrevBlockHeader` error via `load_block_header` returning `None`.
+#[tokio::test]
+async fn genesis_replacement_rejected() {
```
sergerad: @Mirko-von-Leipzig note that because the API takes in a `ProposedBlock`, its impossible for genesis block to be overridden (a proposed block can only be constructed with reference to some parent block).

So this test is a contrived way of testing the related logic.

crates/validator/src/server/tests.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    assert!(result.is_ok(), "chain tip + 1 should succeed, got: {:?}", result.err());
+}
+
+/// A replacement block at the same height as the current chain tip should be accepted.
+#[tokio::test]
+async fn chain_tip_replacement_succeeds() {
+    let mut tv = TestValidator::new().await;
+
+    // Apply block 1 to advance the chain tip.
```
Mirko-von-Leipzig: Nit: explain the why not the what


```suggestion
    // The genesis block can never be replaced, so we advance the chain 
    // to block 1, which we can then replace.
```

crates/validator/src/server/tests.rs
```diff
@@ -0,0 +1,409 @@
...
+    assert!(result.is_err(), "replacement with missing parent should be rejected");
+    assert!(
+        matches!(result.unwrap_err(), BlockValidationError::NoPrevBlockHeader),
+        "expected NoPrevBlockHeader error"
+    );
+}
+
+/// A block with the wrong previous block commitment should be rejected.
+#[tokio::test]
+async fn commitment_mismatch_rejected() {
```
Mirko-von-Leipzig, line 232: Could we add a similar test for replacement + commitment mismatch?

crates/validator/src/server/tests.rs
Mirko-von-Leipzig: Ah. I see, for some reason I assumed we were getting the proposed block's header, but its actually the parents..

I was essentially worried one could construct a proposed genesis block via bytes (since that's what we receive), but the literal definition prohibits that.

I'm also okay with removing this test, its testing the code path but not really something practical. Up to you - we could also name it a database protection test I guess :D In case we lose the block header somehow.

Or actually -- I guess this tests that the database must be properly initialized with genesis which is good. In which case just a test rename?

Up to you.

Mirko-von-Leipzig (approved):
This is great! Thank you!

crates/validator/src/server/tests.rs
sergerad: Sorry I don't quite follow the case. Do you mean we replace the tip, then attempt to add a new block with the commitment matching that of the one that was replaced? I'll add that case now

Mirko-von-Leipzig: The `commitment_mismatch_rejected` tests

```rust
proposed.parent_commitment == chain_tip.commitment
```

aka that proposed block N+1 follows correctly from N.

I want the same test, except for the proposed block being N again. Basically just ensuring that the commitment is checked correctly when we do a replacement block.

PR #1835, opened 2026-03-25
feat: add custom tracing macro to use reports (external version)

drahnr:
Formerly https://github.com/0xMiden/node/pull/1500


crates/store/src/blocks.rs
```diff
@@ -74,13 +67,7 @@ impl BlockStore {
...
     }
 
-    #[instrument(
-        target = COMPONENT,
-        name = "store.block_store.save_block",
-        skip(self, data),
-        err,
-        fields(block_size = data.len())
-    )]
+    #[instrument(COMPONENT: err)]
```
sergerad: @Mirko-von-Leipzig for this kind of diff are we not losing the block_size field we want?

drahnr: That's correct, I haven't done a full self review yet 🫣

crates/tracing-macro/src/allowed.rs
```diff
@@ -0,0 +1,38 @@
...
+});
+
+/// Checks `dotted` against the open telemetry field allowlist.
+///
+/// Succeeds when the given name is present in `allowlist.txt` verbatim,
+/// otherwise provides suggestions in increasing levenshtein distance.
+pub(crate) fn check(dotted: &str) -> Result<(), Vec<String>> {
+    if ALLOWED_OPENTELEMETRY_NAMES.contains(&dotted) {
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
drahnr: :point_up: is a false positive

drahnr:
The implementation changed again, there were more limitations with how `tracing::event!` was parsing input, so went one level deeper doing the implementation using `tracing::span!` which resulted in ~200 additional LOC but allows for both `LitStr` and `Ident` in the component position and a few other goodies.

Issue #1720, opened 2026-02-27
Replace full `NoteMetadata` with a fixed-size header in `NoteSyncRecord`


SantiagoPittella:
With the recent addition of note attachments, `NoteMetadata` is no longer bounded. It can be up to 8-16 KB per note. Since `NoteSyncRecord` is embedded in sync responses that return summaries of many notes, this makes response sizes unbounded.

Replace the full `NoteMetadata` in `NoteSyncRecord` with a fixed-size note metadata header (64 bytes: metadata header + attachment commitment word). Users who need full note details including attachments can use the `GetNotesById` endpoint.

bobbinth:
I think there are several related issues here:

First, we have several slightly different protobuf messages: `NoteSyncRecord`, `NoteInclusionInBlockProof`, `CommittedNote`. I'd love to find a way to normalize these. For example, maybe we need `CommittedNote` and `CommittedNoteInfo` (replacing `NoteSyncRecord`), and both should use `NoteInclusionInBlockProof`.

Second, `NoteMetadata` now contains all attachment data. It is not clear what's the best solution is here but we need to refactor this somehow so that we can:
- Have a small, fixed-size object that contains `sender`, `tag`, `note_type`, and some information about the attachments.
- A way to get full attachments for a note (either as a part of full metadata object - or maybe somehow separately).

Maybe, the `Note` message could be refactored to look something like:
```proto
message Note {
    NoteMetadata metadata = 1;
    optional bytes details = 2;
    optional NoteAttachment attachment = 3;
}
```
It would be good to align these changes with `miden-base` though. cc @PhilippGackstatter in case he has some ideas.

----

Thinking about the future, we will probably transition to having multiple attachments per note. Then, the `Note` message would look something like:
```proto
message Note {
    NoteMetadata metadata = 1;
    optional bytes details = 2;
    repeated NoteAttachment attachments = 3;
}
```

bobbinth:
Yes, I think using `NoteMetadataHeader` in `NoteSyncRecord` (and in other similar messages, if relevant) is probably the right way to go now.

Once we enable support for multiple attachments per note, things may look a bit different - but we can cross this bridge when we get there. @PhilippGackstatter - do we already have an issue for multiple note attachments? (I remember we discussed them somewhere, but not sure where it was).

SantiagoPittella:
Yes, since `protocol` is is already using 0.22 of the crypto repository.

SantiagoPittella:
[...]

PR #1837, opened 2026-03-25
refactor(rpc): normalize NoteSyncRecord with NoteMetadataHeader

SantiagoPittella:
closes [#1720](https://github.com/0xMiden/miden-node/issues/1720)

- Refactor `NoteSyncRecord` proto message to be `NoteMetadataHeader` + `NoteInclusionInBlockProof` instead of carrying the full `NoteMetadata`.
- Add `NoteMetadataHeader` proto message containing the fixed-size metadata fields.
- Add `NoteAttachmentKind` proto enum mapping to `miden_protocol::note::NoteAttachmentKind`.
- Clients that need the full `NoteMetadata` (including attachment) should use `GetNotesById`.

### DIFF
--- a/bin/network-monitor/assets/index.js
+++ b/bin/network-monitor/assets/index.js
@@ -367,8 +367,20 @@ function updateDisplay() {
         rpcService?.details?.RpcStatus?.block_producer_status?.chain_tip ??
         null;
 
+    // Compute effective health for a service, considering all signals for remote provers.
+    const isServiceHealthy = (s) => {
+        if (s.details && s.details.RemoteProverStatus) {
+            const statusOk = s.status === 'Healthy';
+            const testOk = s.testStatus == null || s.testStatus === 'Healthy';
+            const probeResult = grpcWebProbeResults.get(s.details.RemoteProverStatus.url);
+            const probeOk = !probeResult || probeResult.ok;
+            return statusOk && testOk && probeOk;
+        }
+        return s.status === 'Healthy';
+    };
+
     // Count healthy vs unhealthy services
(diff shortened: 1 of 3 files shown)


proto/proto/types/note.proto
```diff
@@ -105,19 +140,16 @@ message NoteInclusionInBlockProof {
     primitives.SparseMerklePath inclusion_path = 4;
 }
 
-// Represents proof of a note inclusion in the block.
+// Represents a note's metadata header together with proof of inclusion in a block.
+//
+// To get the full `NoteMetadata` (including
+// attachment), use `GetNotesById`.
 message NoteSyncRecord {
```
bobbinth: nit: comment wrapping.

crates/store/src/db/mod.rs
```diff
@@ -197,17 +197,14 @@ impl From<NoteRecord> for proto::note::CommittedNote {
...
-            inclusion_path: Some(inclusion_path),
-        }
+        let metadata_header = Some(value.metadata.to_header().into());
+        let inclusion_proof = Some(proto::note::NoteInclusionInBlockProof {
+            note_id: Some(value.note_id.into()),
+            block_num: value.block_num.as_u32(),
+            note_index_in_block: value.note_index.leaf_index_value().into(),
+            inclusion_path: Some(value.inclusion_path.into()),
+        });
+        Self { metadata_header, inclusion_proof }
```
bobbinth: Do we actually use this conversion anywhere? If not, we should probably remove it.

SantiagoPittella: we do not! I'm removing it.

PR #1843, opened 2026-03-25
feat: make `SyncNotes` return multiple blocks

TomasArrachea:
Closes https://github.com/0xMiden/node/issues/1809

This PR updates the RPC component to batch multiple blocks into a single `SyncNotes` response.

## Changes
- RPC `sync_notes` handler now loops over store calls, accumulating blocks until the 4MB budget is exceeded or the range is exhausted
- Proto: changed `SyncNotesResponse` proto from single-block (header + mmr_path + notes) to multi-block (repeated `NoteSyncBlock` blocks + `BlockRange`). Added `StoreSyncNotesResponse` for the store-internal single-block response, using `BlockRange` instead of `PaginationInfo`.
- Store's `sync_notes` uses `open_at(block_num, checkpoint)` instead of `open(block_num)` so MMR proofs are relative to `block_to`

## Follow ups:
- Instead of batching the sync note responses at the RPC component, batch them on the store and minimize the gRPC requests.
- Deduplicate the authentication nodes between returned note updates.

### DIFF
--- a/proto/proto/rpc.proto
+++ b/proto/proto/rpc.proto
@@ -492,7 +492,7 @@ message SyncNotesResponse {
     //
     // `chain_tip` is the current chain tip.
     // `block_num` is the last block checked. If it equals the requested `block_to` (or the chain
-    // tip when unset), the sync is complete. Otherwise the response was truncated and the client
+    // tip when unset), the sync is complete. Otherwise the response was paginated and the client
     // should resume from `block_num + 1`.
     PaginationInfo pagination_info = 1;
 
--- a/crates/store/README.md
+++ b/crates/store/README.md
@@ -209,11 +209,9 @@ When account vault synchronization fails, detailed error information is provided
 
 Returns info which can be used by the client to sync up to the tip of chain for the notes they are interested in.
 
-Client specifies the `note_tags` they are interested in, and the block range from which to search for matching notes. The request will then return the next block containing any note matching the provided tags within the specified range.
+Client specifies the `note_tags` they are interested in, and the block range to search. The response contains all blocks with matching notes that fit within the response payload limit, along with each note's metadata, inclusion proof, and MMR authentication path.
 
-The response includes each note's metadata and inclusion proof.
-
-A basic note sync can be implemented by repeatedly requesting the previous response's block until reaching the tip of the chain.
+If `response.pagination_info.block_num` is less than the requested range end, make another request starting from `response.pagination_info.block_num + 1` to continue syncing.
 
 #### Error Handling
 


proto/proto/rpc.proto
```diff
@@ -473,20 +472,35 @@ message SyncNotesRequest {
...
+    //
+    // `block_from` matches the request's `block_range.block_from`.
+    // `block_to` is the chain tip at the time of the response.
     //
-    // An MMR proof can be constructed for the leaf of index `block_header.block_num` of
-    // an MMR of forest `chain_tip` with this path.
-    primitives.MerklePath mmr_path = 3;
+    // If `block_to` equals the requested range end (or the chain tip when unset), the sync
+    // is complete. Otherwise, the client should request again starting from `block_to + 1`.
+    BlockRange block_range = 5;
```
bobbinth: Would `block_to` here be set to the chain tip or to the requested `block_to`? My thinking is:

- If the user is requesting a specific `block_to`, they probably don't really care about what the current chain tip is.
- If they don't specify the `block_to`, then we do return the cain tip because we know they want to sync to tip of the chain.

Also, and not for this PR, but I wonder if we should make `block_to` mandatory for all sync requests except for the `SyncChainMmr`. There, the request would allow specifying the finality level as described in https://github.com/0xMiden/node/issues/1842.

TomasArrachea: Makes sense, updated it ([2d16d93](https://github.com/0xMiden/node/pull/1843/commits/2d16d9307d469e68414ea71e30701fa063fc68ec)) so `block_to` now returns the requested `block_to` when specified, or the chain tip when unset. 

One detail: when the response is truncated due to the size limit, `block_to` still reflects the requested range end (or chain tip), not the last block included. The client can detect truncation by checking the last entry in blocks

proto/proto/rpc.proto
```diff
@@ -473,20 +472,36 @@ message SyncNotesRequest {
...
 
-    // List of all notes together with the Merkle paths from `response.block_header.note_root`.
-    repeated note.NoteSyncRecord notes = 4;
+    // Blocks containing matching notes, ordered by block number ascending.
+    // May be empty if no notes matched in the range.
+    repeated NoteSyncBlock blocks = 6;
```
bobbinth: The field numbering looks a bit off here. Could `block_range` not start at `1`?

crates/store/src/state/sync_state.rs
```diff
@@ -11,6 +12,17 @@ use crate::db::models::queries::StorageMapValuesPage;
...
+/// Estimated byte size of a [`NoteSyncBlock`] excluding its notes.
+///
+/// `BlockHeader` (~341 bytes) + MMR proof with 32 siblings (~1216 bytes).
+const BLOCK_OVERHEAD_BYTES: usize = 1600;
+
+/// Estimated byte size of a single [`NoteSyncRecord`].
+///
+/// Note ID (~38 bytes) + index + metadata (~26 bytes) + sparse merkle path with 16
+/// siblings (~608 bytes).
+const NOTE_RECORD_BYTES: usize = 700;
```
bobbinth, line 25: This is fine for now, but most likely will be a significant overestimate because sparse Merkle paths get compressed, and in most cases shouldn't be more than a couple hundred bytes. But the compression depends on how many paths there are (the more paths, the worse the compression) - so, taking the worst case is fine for now.

crates/store/src/state/sync_state.rs
```diff
@@ -11,6 +12,17 @@ use crate::db::models::queries::StorageMapValuesPage;
 use crate::db::{AccountVaultValue, NoteSyncUpdate, NullifierInfo};
 use crate::errors::{DatabaseError, NoteSyncError, StateSyncError};
 
+/// Estimated byte size of a [`NoteSyncBlock`] excluding its notes.
+///
+/// `BlockHeader` (~341 bytes) + MMR proof with 32 siblings (~1216 bytes).
+const BLOCK_OVERHEAD_BYTES: usize = 1600;
```
bobbinth, line 19: Not for this PR, but it would be good to make this based on the actual serialization sizes. For example, we could have `BlockHeader::SERIALIZED_SIZE` constant and use it here.

Let's create an issue for this (and include `NOTE_RECORD_BYTES` in this issue too).

bobbinth (approved):

proto/proto/rpc.proto
```diff
@@ -473,20 +472,36 @@ message SyncNotesRequest {
...
 
-    // Merkle path to verify the block's inclusion in the MMR at the returned `chain_tip`.
+    // The block range covered by this response.
     //
-    // An MMR proof can be constructed for the leaf of index `block_header.block_num` of
-    // an MMR of forest `chain_tip` with this path.
-    primitives.MerklePath mmr_path = 3;
+    // `block_from` matches the request's `block_range.block_from`.
+    // `block_to` matches the request's `block_range.block_to`, or the chain tip if it was
+    // not specified.
```
igamigo: nit: This is not always true, right? The text below mentions that the `block_to` may be smaller than what the user requested (and the chain tip as well)

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +76,48 @@ impl State {
...
-    /// The client's request contains a list of tags, this method will return the first
-    /// block with a matching tag, or the chain tip. All the other values are filter based on this
-    /// block range.
-    ///
-    /// # Arguments
-    ///
-    /// - `note_tags`: The tags the client is interested in, resulting notes are restricted to the
-    ///   first block containing a matching note.
-    /// - `block_range`: The range of blocks from which to synchronize notes.
+    /// Returns as many blocks with matching notes as fit within the response payload
```
igamigo: nit: You can reference the max payload size constant here

crates/store/src/server/rpc_api.rs
```diff
@@ -130,26 +130,34 @@ impl rpc_server::Rpc for StoreApi {
         let request = request.into_inner();
 
         let chain_tip = self.state.latest_block_num().await;
```
igamigo, line 132: nit (feel free to disregard): There's no need to read the latest block number if the user set a `block_to`

crates/store/src/server/rpc_api.rs
```diff
@@ -130,26 +130,34 @@ impl rpc_server::Rpc for StoreApi {
...
+        let requested_block_to = request.block_range.as_ref().and_then(|r| r.block_to);
```
igamigo: nit: I would move this down to be below line 134

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +76,48 @@ impl State {
...
+            };
 
-        let (note_sync, last_included_block) =
-            self.db.get_note_sync(block_range, note_tags).await?;
+            accumulated_size += BLOCK_OVERHEAD_BYTES + note_sync.notes.len() * NOTE_RECORD_BYTES;
 
-        let mmr_proof = inner.blockchain.open(note_sync.block_header.block_num())?;
+            if accumulated_size > MAX_RESPONSE_PAYLOAD_BYTES {
+                break;
+            }
```
igamigo: Returning a single block is always guaranteed, right? We should make sure this is the case because otherwise it might be problematic

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +76,48 @@ impl State {
...
         let inner = self.inner.read().await;
+        let checkpoint = *block_range.end();
+
+        let mut results = Vec::new();
+        let mut accumulated_size: usize = 0;
+        let mut current_from = *block_range.start();
+
+        loop {
+            let range = current_from..=checkpoint;
+            let Some(note_sync) = self.db.get_note_sync(range, note_tags.clone()).await? else {
```
igamigo: Wonder if `get_note_sync` could change to take something like `&[u32]` to avoid cloning every time here

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +76,48 @@ impl State {
...
+    /// Returns as many blocks with matching notes as fit within the response payload
+    /// limit. Each block includes its header and MMR proof at `block_range.end()`.
     #[instrument(level = "debug", target = COMPONENT, skip_all, ret(level = "debug"), err)]
     pub async fn sync_notes(
         &self,
         note_tags: Vec<u32>,
         block_range: RangeInclusive<BlockNumber>,
-    ) -> Result<(NoteSyncUpdate, MmrProof, BlockNumber), NoteSyncError> {
+    ) -> Result<Vec<(NoteSyncUpdate, MmrProof)>, NoteSyncError> {
         let inner = self.inner.read().await;
```
igamigo: I think we want to take the lock in the loop to avoid keeping it for the whole duration

igamigo: (and it can be moved right next to the `open_at()` call)

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +76,48 @@ impl State {
...
-        let mmr_proof = inner.blockchain.open(note_sync.block_header.block_num())?;
+            if accumulated_size > MAX_RESPONSE_PAYLOAD_BYTES {
+                break;
+            }
+
+            let block_num = note_sync.block_header.block_num();
+
+            if block_num >= checkpoint {
+                break;
+            }
```
igamigo: We break here but AFAIK we are supposed to return notes for the checkpoint block, right? I think this might be a special case. If the user calls `SyncMmr` which returns the chain tip header and the `MmrDelta`, the user already has the header (and it's already validated by default). So when `block_num == checkpoint`, we might to just returnr the notes, and not return the merkle path at all (`open_at` would otherwise fail, I think). cc @bobbinth

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +76,48 @@ impl State {
...
+
+            let mmr_proof = inner.blockchain.open_at(block_num, checkpoint)?;
```
igamigo: Before, I think if the user set a `block_to` beyond the chain tip, the response would be automatically valid because the store would find no note after the chain tip, but now because we `open_at` the user-set `checkpoint`, this might fail. I think this is fine because it's a wrong argument, but maybe at the RPC level we should check that the `block_to` is not beyond the chain tip and return early.

proto/proto/rpc.proto
TomasArrachea: I removed the text below; as mentioned [in this discussion](https://github.com/0xMiden/node/pull/1843#discussion_r2996824856) the `block_to` will always return either the requested block or the chain tip.

crates/store/src/state/sync_state.rs
TomasArrachea: Added an extra check so the first block is always included, even if the size is exceeded.

crates/store/src/state/sync_state.rs
TomasArrachea: The issue is that `get_note_sync` then uses `db.transact` with takes a `move` closure, so we cannot pass a `&[u32]`. I changed it to pass an `Arc<[u32]>` that is cheaper to clone.

crates/store/src/state/sync_state.rs
TomasArrachea: The RPC component doesn't have the chain tip so we would need to add a request to the Store for that check. Instead, I added a check on the Store component so the block_to is set to the chain tip if needed - this should mimic the prior behaviour.

Another detail, this means that the response block_num will be the chain tip if the requested number is beyond the tip. Updated the comments to clarify that.

crates/store/src/state/sync_state.rs
TomasArrachea: Yeah, the `open_at` would fail if the `block_num == checkpoint`.

I changed it so the last note sync is returned without the MMR proof.

proto/proto/rpc.proto
```diff
@@ -473,20 +472,33 @@ message SyncNotesRequest {
...
+        blockchain.BlockHeader block_header = 1;
 
-    // Block header of the block with the first note matching the specified criteria.
-    blockchain.BlockHeader block_header = 2;
+        // Merkle path to verify the block's inclusion in the MMR at the returned
+        // `block_range.block_to`.
+        //
+        // An MMR proof can be constructed for the leaf of index `block_header.block_num` of
+        // an MMR of forest `block_range.block_to` with this path.
+        primitives.MerklePath mmr_path = 2;
```
bobbinth: Is the comment on line 480-481 correct here? I thought the MMR path is done against the requested `block_range.block_to` rather than the returned one.

crates/store/src/server/rpc_api.rs
```diff
@@ -133,23 +133,31 @@ impl rpc_server::Rpc for StoreApi {
         let block_range =
             read_block_range::<NoteSyncError>(request.block_range, "SyncNotesRequest")?
                 .into_inclusive_range::<NoteSyncError>(&chain_tip)?;
+        let block_from = block_range.start();
+        // Clamp block_to to the chain tip to avoid erroring when opening the MMR proof
+        let block_to = block_range.end().min(&chain_tip);
+        let clamped_block_range = *block_from..=*block_to;
```
bobbinth: I think things like this should be enforced in `BlockRange::into_inclusive_range()`. But also, I'm not sure we want to clamp here. Maybe returning an error is a better approach?

crates/store/src/state/sync_state.rs
bobbinth: I wonder if there is an off-by-one issue somewhere here. Let's say the current chain tip is at block 100 (i.e., there are 101 blocks in the chain). The user sends a request either specifying `block_to = 100` or omitting `block_to` (which should have the same effect). This means, the user wants to get notes up to, and including, block 100.

Assuming there are some relevant notes in block 100, the user needs to get the authentication for this block as well. To do this, they need to say `blockchain.open_at(100, 101)`. I Believe the second value needs to be `101` because the second parameter should the `forest` which is the total number of leaves in the MMR.

The problem, may be in the `Blockchain::open_at()` implementation, where we say that we return the MMR proof at `checkpoint` but I think we are actually returning it at `checkpoint-1`. If that's the case, it'll have to be fixed in `miden-protocol`. cc @PhilippGackstatter - in case I'm off here.

igamigo: Yeah, at first signt it's confusing, but I think it's fully cohesive. Specifically:

[`Blockchain` docs](https://github.com/0xMiden/protocol/blob/next/crates/miden-protocol/src/block/blockchain.rs#L25-L29) mention

```
/// Some APIs take a _checkpoint_ which is equivalent to the concept of _forest_ of the underlying
/// MMR. As an example, if the blockchain has 20 blocks in total, and the checkpoint is 10, then
/// the API works in the context of the chain at the time it had 10 blocks, i.e. it contains blocks
/// 0..=9. This is useful, for example, to retrieve proofs that are valid when verified against the
/// chain commitment of block 10.
```

So `checkpoint` is not the actual block number that was expected to be the last included in the MMR when the proof is returned. Which also makes sense because the chain commitment at that point is the one that goes into the block number `checkpoint`. 
From the client perspective, I think the API for this works if we basically:

- `SyncMmr` to chain tip
- Apply the `MmrDelta`, but defer `add`ing the chain tip block header
- `SyncNotes` and apply all returned merkle proofs to the MMR
- (`add` the chain tip block header)

For the `chain_tip` case, there would be no need to return the header merkle proof and so it should be an optional field in the response.

bobbinth: Ah yes - then we don't need to modify the `Blockchain` struct. If I'm understanding correctly, what you describe basically would work like this:

Let's say the chain tip is block 11 (i.e., there are 12 blocks in the chain). If we make the `SyncNotes` request with `block_to = 11`, we'll get back MMR proofs for a chain of 11 blocks (because we'd use `checkpoint = 11`), and the last block, would come with no MMR proof, right?

If so, I do find this a bit confusing. I understand that it would work on the client by deferring adding the chain tip - but it feels like it could be a lot to remember (even if it works efficiently in practice).

I guess, why not return MMR proofs for a chain of 12 blocks (i.e. use `checkpoint = block_to + 1`)? Basically, here, we'd just set the checkpoint to be `block_to + 1`, and then the logic is the same for all blocks. Or does this complicate things too much on the client?

igamigo: Yeah, that would be equivalent to `open(block)` at that point, so it should work as expected.

crates/store/src/server/rpc_api.rs
TomasArrachea: Changed this to error instead https://github.com/0xMiden/node/pull/1843/commits/889fc04398541cab53f27c1ff6a6106e28f98701

crates/store/src/state/sync_state.rs
TomasArrachea: Thanks! Made this change in [59c4658](https://github.com/0xMiden/node/pull/1843/commits/59c465859c10f138e78ad483bc2e333375b6827f)

crates/store/src/server/rpc_api.rs
TomasArrachea: The `read_block_range` utility needs the chain tip as a fallback, the same is done in other endpoints. A slight improvement could be to take a closure as fallback that reads the tip only when needed, but not sure if its a big win.

crates/store/src/server/rpc_api.rs
```diff
@@ -133,23 +133,33 @@ impl rpc_server::Rpc for StoreApi {
...
+            })
+            .collect();
 
         Ok(Response::new(proto::rpc::SyncNotesResponse {
-            pagination_info: Some(proto::rpc::PaginationInfo {
-                chain_tip: chain_tip.as_u32(),
-                block_num: last_block_included.as_u32(),
+            block_range: Some(proto::rpc::BlockRange {
+                block_from: block_from.as_u32(),
+                block_to: Some(block_to.as_u32()),
```
igamigo: `block_to` here is probably not what we want, as it is just the same number that the user might have requested. Instead, what we need (AFAIK) is to return the largest block num in the `results` (or have it returned separately on `sync_notes`). Or, if no valid sync notes response was found, then we reached the user's `block_to`

crates/store/src/state/sync_state.rs
igamigo: Reminder to create this issue later

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +77,47 @@ impl State {
...
     pub async fn sync_notes(
         &self,
         note_tags: Vec<u32>,
         block_range: RangeInclusive<BlockNumber>,
-    ) -> Result<(NoteSyncUpdate, MmrProof, BlockNumber), NoteSyncError> {
-        let inner = self.inner.read().await;
+    ) -> Result<Vec<(NoteSyncUpdate, MmrProof)>, NoteSyncError> {
+        let block_end = *block_range.end();
+        // Use block_end + 1 as the MMR checkpoint so that block_end itself can be proven.
+        let mmr_checkpoint = block_end + 1;
```
igamigo: nit: I'd move this down closer to the `open_at` call

proto/proto/rpc.proto
```diff
@@ -82,14 +82,12 @@ service Api {
...
+    // Client specifies the `note_tags` they are interested in, and the block range to search.
+    // The response contains all blocks with matching notes that fit within the response payload
+    // limit, along with each note's metadata and inclusion proof.
     //
-    // The response includes each note's metadata and inclusion proof.
-    //
-    // A basic note sync can be implemented by repeatedly requesting the previous response's block
-    // until reaching the tip of the chain.
+    // If `response.block_range.block_to` is less than the requested range end, make another
+    // request starting from `response.block_range.block_to + 1` to continue syncing.
```
igamigo: I think the underlying query is `committed_at > block_from`, so I think prompting for `+ 1` is not correct here.

crates/store/src/server/rpc_api.rs
TomasArrachea: I changed this so the `block_to` returned marks the last block included in the response. This makes sense for the proposed flow where the client calls `SyncChainMmr` and then uses that chain tip as parameter for the sync notes. But if a user made a sync notes request without specifying the chain tip, it would have no way of finding out which is the chain tip used for the MMR paths included in the response. It might not be the last included block if the response is truncated.

TomasArrachea: A simple solution would be to add a `chain_tip` field to the response

Edit: actually, using `PaginationInfo` would be a better fit.

igamigo: Agree, that should work well

crates/store/src/server/rpc_api.rs
```diff
@@ -133,23 +133,31 @@ impl rpc_server::Rpc for StoreApi {
         let block_range =
             read_block_range::<NoteSyncError>(request.block_range, "SyncNotesRequest")?
                 .into_inclusive_range::<NoteSyncError>(&chain_tip)?;
+        if *block_range.end() > chain_tip {
+            Err(NoteSyncError::FutureBlock { chain_tip, block_to: *block_range.end() })?;
+        }
```
bobbinth, line 138: Not for this PR, but I wonder if we should modify `into_inclusive_range()` to perform this check. This may be a part of a bigger refactoring though.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -158,14 +158,14 @@ pub(crate) fn select_notes_since_block_by_tag_and_sender(
                     .eq_any(&desired_note_tags[..])
                     .or(schema::notes::sender.eq_any(&desired_senders[..])),
             )
-            .filter(schema::notes::committed_at.gt(start_block_num))
+            .filter(schema::notes::committed_at.ge(start_block_num))
```
bobbinth, line 161: Does `ge` mean "greater or equal then"? If so, was `gt` a bug previously? And if so, we should probably update the doc comments (the SQL statement still says `block_num > ?3`.

crates/store/src/db/models/queries/notes.rs
```diff
@@ -141,7 +141,7 @@ pub(crate) fn select_notes_since_block_by_tag_and_sender(
     account_ids: &[AccountId],
     note_tags: &[u32],
     block_range: RangeInclusive<BlockNumber>,
-) -> Result<(Vec<NoteSyncRecord>, BlockNumber), DatabaseError> {
+) -> Result<Vec<NoteSyncRecord>, DatabaseError> {
```
bobbinth, line 144: Not for this PR, but it seems like this function is never invoked with `account_ids` supplied. If so, we should remove the `account_ids` parameter and simplify the associated SQL. Let's create an issue for this.

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +77,52 @@ impl State {
...
+            accumulated_size += BLOCK_OVERHEAD_BYTES + note_sync.notes.len() * NOTE_RECORD_BYTES;
+
+            if !results.is_empty() && accumulated_size > MAX_RESPONSE_PAYLOAD_BYTES {
+                break;
+            }
+
+            let block_num = note_sync.block_header.block_num();
+            let mmr_checkpoint = block_end + 1;
+            let mmr_proof =
+                self.inner.read().await.blockchain.open_at(block_num, mmr_checkpoint)?;
```
bobbinth, line 119: Let's add a comment explaining why `mmr_checkpoint = block_end + 1` is safe.

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +77,52 @@ impl State {
...
+            let Some(note_sync) = self.db.get_note_sync(range, Arc::clone(&note_tags)).await?
+            else {
+                break;
+            };
+
+            accumulated_size += BLOCK_OVERHEAD_BYTES + note_sync.notes.len() * NOTE_RECORD_BYTES;
+
+            if !results.is_empty() && accumulated_size > MAX_RESPONSE_PAYLOAD_BYTES {
+                break;
+            }
```
bobbinth, line 110: Why do we have `!results.is_empty()` as a condition here? Basically, if the result is empty, why do we need to fall through to getting the MMR proof?

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +77,52 @@ impl State {
...
+
+            let block_num = note_sync.block_header.block_num();
+            let mmr_checkpoint = block_end + 1;
+            let mmr_proof =
+                self.inner.read().await.blockchain.open_at(block_num, mmr_checkpoint)?;
+            results.push((note_sync, mmr_proof));
+
+            current_from = block_num + 1;
```
bobbinth, line 122: Why is this not just `current_from = block_num`? It seems like this may result in some inconsistencies - but if not, I'd add a comment here explaining why we are doing this.

To lay out a scenario: let's say we are doing a sync with `block_from = 10` and `block_to = 20`. The first request to `get_note_sync` will be `10..20`. If the lower bound is exclusive here, we may get notes from block 11 - and so, `block_num` would be 11, and the next request would be for range `12..20`, which would skip block 12 (if the range is exclusive).

If the range is inclusive, then, in the first request, we could get notes from block 10 - but everything else should be fine.

bobbinth: Thinking more about this, it seems like the behavior is consistent, but it is different from how this used to work before. Specifically:

Previously, `block_to` was "exclusive" - i.e., if we send `block_to = 10`, we'd get back notes from blocks 11, 12, 13 etc. And to continue the sync, we'd make another request with `request.block_from = response.block_num`.

Now, `block_to` is "inclusive" - i.e., if we send `block_to = 10`, we'd get back notes from blocks 10, 11, 12, etc. And to continue the sync, we'd make another request with `request.block_from = response.block_num + 1`.

So, both work - but isn't the old approach more consistent with how all other similar endpoints work (e.g., `SyncTransactions`)? If so, any particular reason to change this?

crates/store/src/state/sync_state.rs
TomasArrachea: This was added after this dicussion https://github.com/0xMiden/node/pull/1843#discussion_r2998915065 to guarantee that at least one update is included in the response. If the `db.get_note_sync` returns `None` we do break .

crates/store/src/db/models/queries/notes.rs
SantiagoPittella, line 161: The doc comment SQL above still says `block_num > ?3` (exclusive) but the code now uses `.ge` (inclusive, `>=`). This should be updated to match the docs.

bin/stress-test/src/store/mod.rs
```diff
@@ -163,25 +173,15 @@ pub async fn bench_sync_nullifiers(
                 .into_inner()
                 .notes;
```
SantiagoPittella, line 174: This breaks when `blocks` is empty, but doesn't check whether the sync actually reached the chain tip. If there are no matching notes in the current range but more blocks exist beyond it, this stops syncing prematurely.

The old code compared `chain_tip == current_block_num` before breaking. Consider:
```rust
if response.blocks.is_empty() {
    // No matching notes — sync is complete if we checked up to the tip.
    break;
}
```
This happens to be fine here because the stress test controls the data, but it's a subtle difference from the old behavior worth being aware of.

crates/store/src/state/sync_state.rs
TomasArrachea: Right, I changed this to use inclusive ranges to follow the existing documentation of the `PaginationInfo` response: https://github.com/0xMiden/node/blob/08e47a4e8a1319e8d4bb39d522c8c4eda3d91cea/proto/proto/rpc.proto#L595-L597

Actually, the other endpoints also work with inclusive ranges, so my thinking was to make the sync notes follow the same behaviour. I think this was inconsistent before.

Sorry for not pointing this out explicitly.

crates/store/src/db/models/queries/notes.rs
SantiagoPittella: Using `ge` is the most consistent across the codebase. I'm in favor of keeping it.

TomasArrachea: It was not a bug, rather the endpoint handled the lower bound as exclusive. Changed this as mentioned in https://github.com/0xMiden/node/pull/1843#discussion_r3016522341. 

Updated the SQL comment.

bin/stress-test/src/store/mod.rs
TomasArrachea: > This breaks when blocks is empty, but doesn't check whether the sync actually reached the chain tip

With the new implementation of the sync notes it's not strictly needed. When `blocks` is empty, it means the store already searched the entire range up to the chain tip. Still, added the check to compare the returned `block_num` with the chain tip so the last extra request is avoided.

crates/store/src/server/rpc_api.rs
TomasArrachea: Created https://github.com/0xMiden/node/issues/1866

crates/store/src/state/sync_state.rs
TomasArrachea: Created https://github.com/0xMiden/node/issues/1867

crates/store/src/state/sync_state.rs
igamigo: I checked and all the sync endpoints seem aligned in this regard, so `SyncNotes` was the exception before.

proto/proto/rpc.proto
```diff
@@ -473,20 +472,33 @@ message SyncNotesRequest {
...
 
-    // Merkle path to verify the block's inclusion in the MMR at the returned `chain_tip`.
+    // Pagination information.
     //
-    // An MMR proof can be constructed for the leaf of index `block_header.block_num` of
-    // an MMR of forest `chain_tip` with this path.
-    primitives.MerklePath mmr_path = 3;
+    // `chain_tip` is the current chain tip.
+    // `block_num` is the last block checked. If it equals the requested `block_to` (or the chain
+    // tip when unset), the sync is complete. Otherwise the response was truncated and the client
```
igamigo: nit:
```suggestion
    // tip when unset), the sync is complete. Otherwise the response was paginated and the client
```

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +77,56 @@ impl State {
...
     pub async fn sync_notes(
         &self,
         note_tags: Vec<u32>,
         block_range: RangeInclusive<BlockNumber>,
-    ) -> Result<(NoteSyncUpdate, MmrProof, BlockNumber), NoteSyncError> {
-        let inner = self.inner.read().await;
-
-        let (note_sync, last_included_block) =
-            self.db.get_note_sync(block_range, note_tags).await?;
+    ) -> Result<(Vec<(NoteSyncUpdate, MmrProof)>, BlockNumber), NoteSyncError> {
```
igamigo, line 91: With some of these changes we might need to update the store's `README.md`

crates/store/src/db/models/queries/notes.rs
igamigo, line 139: Could this be modified to return multiple responses at once? I think this was the original suggestion (although we should do it in a separate issue/PR)

crates/store/src/state/sync_state.rs
```diff
@@ -64,29 +77,56 @@ impl State {
...
+                self.inner.read().await.blockchain.open_at(block_num, mmr_checkpoint)?;
+            results.push((note_sync, mmr_proof));
+
+            current_from = block_num + 1;
+        }
 
-        let mmr_proof = inner.blockchain.open(note_sync.block_header.block_num())?;
+        // if results is empty, return `block_end` since the sync is complete.
+        let last_block_checked =
+            results.last().map_or(block_end, |(update, _)| update.block_header.block_num());
```
igamigo, line 127: There is a small optimization here where we return the block number we dropped, minus one. For example, if we match against blocks 1, 5, 10 and 15 but block 15 does not make it into the response, instead of returning 10 as the last included block, we return 14 (so the next query avoids blocks 10-14 because the user will start at 15 already). I doubt this is significant due to how the tables are probably indexed, so I would not bother doing it unless it's really trivial, but maybe it can be done in a follow up PR or issue

crates/store/src/state/sync_state.rs
bobbinth: Makes sense! Thank you for making this endpoint consistent with the others!

crates/store/src/state/sync_state.rs
bobbinth: I see. Basically, the idea here is that if somehow `accumulated_size` is over the limit on the first iteration (i.e., when `esults.is_empty()`) we still want to send the response. Practically, this shouldn't happen (unless we are off in `accumulated_size` computations), but it's better to be on the safe side.

crates/store/src/state/sync_state.rs
TomasArrachea: Created https://github.com/0xMiden/node/issues/1870

PR #1846, opened 2026-03-26
feat: user batch support

### LINKED ISSUE #1112: Support submitting `ProvenBatch`

### LINKED ISSUE #1859: Update mempool doc comments
This is basically to track the review I left in https://github.com/0xMiden/node/pull/1832#pullrequestreview-4027288007.

Mirko-von-Leipzig:
This PR is the third and final part of the mempool refactoring PR stack. Part 1 (#1820) performs the broad mempool refactoring to simplify this PR. Builds on part 2 (#1832).

Batch submissions must include their transaction inputs since we currently require this for the validator to verify them before inclusion in a block. This PR abuses this by treating the batch as a set of normal transactions at the mempool level. This simplifies the mempool implementation, which is currently built around a DAG of transactions - so having to insert a batch directly would be more complex. This will need to change once we stop requiring transaction inputs as part of the validator; but it won't be too bad.

The way this is implemented here, is that the transaction DAG tracks user batches and ensures that when a batch is selected, that transactions from user batches are not mixed with conventional transactions. That is, `select_batch` outputs either a user batch, _or_ a conventional batch.

Effectively, the transaction DAG internally ensures that the user batch's transactions remain coherent even though the batch has been deconstructed into individual transactions. The benefit is that this doesn't require any major structural changes to the mempool. The rest of the mempool then treats the user batch as per normal.

Closes #1112 and closes #1859

### DIFF
--- a/crates/block-producer/src/mempool/graph/batch.rs
+++ b/crates/block-producer/src/mempool/graph/batch.rs
@@ -148,7 +148,7 @@ impl BatchGraph {
         selected
     }
 
-    /// Prunes the given batch.
+    /// Prunes the given batch and returns it.
     ///
     /// # Panics
     ///
--- a/crates/block-producer/src/mempool/graph/batch.rs
+++ b/crates/block-producer/src/mempool/graph/batch.rs
@@ -21,15 +21,15 @@ impl GraphNode for SelectedBatch {
     type Id = BatchId;
 
     fn nullifiers(&self) -> Box<dyn Iterator<Item = Nullifier> + '_> {
-        Box::new(self.txs().iter().flat_map(|tx| tx.nullifiers()))
+        Box::new(self.transactions().iter().flat_map(|tx| tx.nullifiers()))
     }
 
     fn output_notes(&self) -> Box<dyn Iterator<Item = Word> + '_> {
-        Box::new(self.txs().iter().flat_map(|tx| tx.output_note_commitments()))
+        Box::new(self.transactions().iter().flat_map(|tx| tx.output_note_commitments()))
     }
 
     fn unauthenticated_notes(&self) -> Box<dyn Iterator<Item = Word> + '_> {
-        Box::new(self.txs().iter().flat_map(|tx| tx.unauthenticated_note_commitments()))
+        Box::new(self.transactions().iter().flat_map(|tx| tx.unauthenticated_note_commitments()))
     }
 
     fn account_updates(
@@ -98,10 +98,14 @@ impl BatchGraph {
     ///
     /// Batches are returned in reverse-chronological order.
     pub fn revert_expired(&mut self, chain_tip: BlockNumber) -> Vec<SelectedBatch> {
-        let reverted = self.inner.revert_expired_unselected(chain_tip);
+        // We only revert transactions which are _not_ included in batches.
+        let mut to_revert = self.inner.expired(chain_tip);
+        to_revert.retain(|batch| !self.inner.is_selected(batch));
 
-        for batch in &reverted {
-            self.proven.remove(&batch.id());
+        let mut reverted = Vec::with_capacity(to_revert.len());
+
+        for batch in to_revert {
+            reverted.extend_from_slice(&self.revert_batch_and_descendants(batch));
         }
 
         reverted
@@ -150,9 +154,9 @@ impl BatchGraph {
     ///
     /// Panics if the batch does not exist, or has existing ancestors in the batch
     /// graph.
-    pub fn prune(&mut self, batch: BatchId) {
-        self.inner.prune(batch);
+    pub fn prune(&mut self, batch: BatchId) -> SelectedBatch {
         self.proven.remove(&batch);
+        self.inner.prune(batch)
     }
 
     pub fn proven_count(&self) -> usize {
--- a/crates/block-producer/src/mempool/graph/dag.rs
+++ b/crates/block-producer/src/mempool/graph/dag.rs
@@ -199,28 +199,13 @@ where
         reverted
     }
 
-    /// Reverts nodes (and their descendants) which have expired and which are _not_ selected.
-    ///
-    /// Returns the reverted nodes in **reverse** chronological order.
-    pub fn revert_expired_unselected(&mut self, chain_tip: BlockNumber) -> Vec<N> {
-        let mut reverted = Vec::default();
-
-        let expired = self
-            .nodes
+    /// Returns the set of nodes that expired at the given block height.
+    pub fn expired(&self, chain_tip: BlockNumber) -> HashSet<N::Id> {
+        self.nodes
             .iter()
-            .filter(|(id, _)| !self.is_selected(id))
             .filter_map(|(id, node)| (node.expires_at() <= chain_tip).then_some(id))
             .copied()
-            .collect::<HashSet<_>>();
-
-        for id in expired {
-            // Its possible the node is already reverted by a previous loop iteration.
-            if self.contains(&id) {
-                reverted.extend(self.revert_node_and_descendants(id));
-            }
-        }
-
-        reverted
+            .collect()
     }
 
     /// Returns `true` if the given node is a leaf node aka has no children.
@@ -233,14 +218,14 @@ where
     /// # Panics
     ///
     /// Panics if this node has any ancestor nodes, or if this node was not selected.
-    pub fn prune(&mut self, id: N::Id) {
+    pub fn prune(&mut self, id: N::Id) -> N {
         assert!(
             self.edges.parents_of(&id).is_empty(),
             "Cannot prune node {id} as it still has ancestors",
         );
         assert!(self.selected.contains(&id), "Cannot prune node {id} as it was not selected");
 
-        self.remove(id);
+        self.remove(id)
     }
 
     /// Unconditionally removes the given node from the graph, deleting its edges and state.
@@ -330,31 +315,5 @@ mod tests {
         let candidates_after_parent: Vec<u32> =
             graph.selection_candidates().keys().map(|id| **id).collect();
         assert_eq!(candidates_after_parent, vec![2]);
-    }
-
-    #[test]
-    fn revert_expired_unselected_removes_descendants() {
-        let mut graph = Graph::<TestNode>::default();
-
-        graph
-            .append(
-                TestNode::new(1).with_output_notes([1]).with_expires_at(BlockNumber::from(2u32)),
-            )
-            .unwrap();
-        graph
-            .append(
-                TestNode::new(2)
-                    .with_output_notes([2])
-                    .with_unauthenticated_notes([1])
-                    .with_expires_at(BlockNumber::from(3u32)),
-            )
-            .unwrap();
-
-        let reverted = graph.revert_expired_unselected(BlockNumber::from(3u32));
-        let reverted_ids: Vec<u32> = reverted.into_iter().map(|node| node.id).collect();
-
-        assert_eq!(reverted_ids, vec![2, 1]);
-        assert_eq!(graph.node_count(), 0);
-        assert_eq!(graph.selection_candidates().len(), 0);
     }
 }
--- a/crates/block-producer/src/mempool/graph/dag.rs
+++ b/crates/block-producer/src/mempool/graph/dag.rs
@@ -213,7 +213,7 @@ where
         self.edges.children_of(id).is_empty()
     }
 
-    /// Removes the node _IFF_ it has no ancestor nodes.
+    /// Removes the node _IFF_ it has no ancestor nodes and returns the pruned node.
     ///
     /// # Panics
     ///
--- a/crates/block-producer/src/mempool/graph/transaction.rs
+++ b/crates/block-producer/src/mempool/graph/transaction.rs
@@ -77,7 +77,11 @@ pub struct TransactionGraph {
     /// used to identify potentially buggy transactions that should be evicted.
     failures: HashMap<TransactionId, u32>,
 
+    /// Defines the transactions that belong to a user batch.
     user_batch_txs: HashMap<BatchId, Vec<TransactionId>>,
+    /// A mapping of transactions to their user batch (if any).
+    ///
+    /// Inverse map of `user_batch_txs`.
     txs_user_batch: HashMap<TransactionId, BatchId>,
 }
 
@@ -195,10 +199,19 @@ impl TransactionGraph {
 
     /// Reverts expired transactions and their descendants.
     ///
-    /// Only unselected transactions are considered; selected transactions are assumed to be in
-    /// committed blocks and should not be reverted.
+    /// This is because we don't distinguish between committed and selected transactions. If we
+    /// didn't ignore selected transactions here, we would revert committed ones as well, which
+    /// breaks the state.
     ///
     /// Returns the identifiers of transactions that were removed from the graph.
+    ///
+    /// # Note
+    ///
+    /// Since this _ignores_ selected transactions, and the purpose is to revert expired
+    /// transactions after a block is committed, the caller **must** ensure that selected
+    /// transactions from expired batches (and therefore not committed) are deselected
+    /// _before_ calling this function. i.e. first revert expired batches and deselect their
+    /// transactions, then call this.
     pub fn revert_expired(&mut self, chain_tip: BlockNumber) -> HashSet<TransactionId> {
         // We only revert transactions which are _not_ included in batches.
         let mut to_revert = self.inner.expired(chain_tip);
@@ -272,6 +285,10 @@ impl TransactionGraph {
     ///
     /// This weeds out transactions which participate in batch and block failures, and might be the
     /// root cause.
+    ///
+    /// # Returns
+    ///
+    /// Returns the set of reverted transactions.
     pub fn increment_failure_count(
         &mut self,
         txs: impl Iterator<Item = TransactionId>,
--- a/crates/block-producer/src/mempool/graph/transaction.rs
+++ b/crates/block-producer/src/mempool/graph/transaction.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 
 use miden_protocol::Word;
 use miden_protocol::account::AccountId;
+use miden_protocol::batch::BatchId;
 use miden_protocol::block::BlockNumber;
 use miden_protocol::note::Nullifier;
 use miden_protocol::transaction::TransactionId;
@@ -75,6 +76,9 @@ pub struct TransactionGraph {
     /// These are batch or block proving errors in which the transaction was a part of. This is
     /// used to identify potentially buggy transactions that should be evicted.
     failures: HashMap<TransactionId, u32>,
+
+    user_batch_txs: HashMap<BatchId, Vec<TransactionId>>,
+    txs_user_batch: HashMap<TransactionId, BatchId>,
 }
 
 impl TransactionGraph {
@@ -85,16 +89,100 @@ impl TransactionGraph {
         self.inner.append(tx)
     }
 
-    pub fn select_batch(&mut self, mut budget: BatchBudget) -> Option<SelectedBatch> {
+    pub fn append_user_batch(
+        &mut self,
+        batch: &[Arc<AuthenticatedTransaction>],
+    ) -> Result<(), StateConflict> {
+        let batch_id =
+            BatchId::from_transactions(batch.iter().map(|tx| tx.raw_proven_transaction()));
+
+        // Append each transaction, but revert atomically on error.
+        for (idx, tx) in batch.iter().enumerate() {
+            if let Err(err) = self.append(Arc::clone(tx)) {
+                // We revert in reverse order because inner.revert panics if the node doesn't exist.
+                for tx in batch.iter().take(idx).rev() {
+                    let reverted = self.inner.revert_node_and_descendants(tx.id());
+                    assert_eq!(reverted.len(), 1);
+                    assert_eq!(&reverted[0], tx);
+                }
+
+                return Err(err);
+            }
+        }
+
+        let txs = batch.iter().map(GraphNode::id).collect::<Vec<_>>();
+        for tx in &txs {
+            self.txs_user_batch.insert(*tx, batch_id);
+        }
+        self.user_batch_txs.insert(batch_id, txs);
+
+        Ok(())
+    }
+
+    pub fn select_batch(&mut self, budget: BatchBudget) -> Option<SelectedBatch> {
+        self.select_user_batch().or_else(|| self.select_conventional_batch(budget))
+    }
+
+    fn select_user_batch(&mut self) -> Option<SelectedBatch> {
+        // Comb through all user batch candidates.
+        let candidate_batches = self
+            .inner
+            .selection_candidates()
+            .values()
+            .filter_map(|tx| self.txs_user_batch.get(&tx.id()))
+            .copied()
+            .collect::<HashSet<_>>();
+
+        'outer: for candidate in candidate_batches {
+            let mut selected = SelectedBatch::builder();
+
+            let txs = self
+                .user_batch_txs
+                .get(&candidate)
+                .cloned()
+                .expect("bi-directional mapping should be coherent");
+
+            for tx in txs {
+                let Some(tx) = self.inner.selection_candidates().get(&tx).copied() else {
+                    // Rollback this batch selection since it cannot complete.
+                    for tx in selected.txs.into_iter().rev() {
+                        self.inner.deselect(tx.id());
+                    }
+
+                    continue 'outer;
+                };
+                let tx = Arc::clone(tx);
+
+                self.inner.select_candidate(tx.id());
+                selected.push(tx);
+            }
+
+            assert!(!selected.is_empty(), "User batch should not be empty");
+            return Some(selected.build());
+        }
+
+        None
+    }
+
+    fn select_conventional_batch(&mut self, mut budget: BatchBudget) -> Option<SelectedBatch> {
         let mut selected = SelectedBatch::builder();
 
-        while let Some((id, tx)) = self.inner.selection_candidates().pop_first() {
-            if budget.check_then_subtract(tx) == BudgetStatus::Exceeded {
+        loop {
+            // Select arbitrary candidate which is _not_ part of a user batch.
+            let candidates = self.inner.selection_candidates();
+            let Some(candidate) =
+                candidates.values().find(|tx| !self.txs_user_batch.contains_key(&tx.id()))
+            else {
+                break;
+            };
+
+            if budget.check_then_subtract(candidate) == BudgetStatus::Exceeded {
                 break;
             }
 
-            selected.push(Arc::clone(tx));
-            self.inner.select_candidate(*id);
+            let candidate = Arc::clone(candidate);
+            self.inner.select_candidate(candidate.id());
+            selected.push(candidate);
         }
 
         if selected.is_empty() {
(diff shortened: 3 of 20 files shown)


proto/proto/types/transaction.proto
```diff
@@ -18,10 +18,15 @@ message ProvenTransaction {
     optional bytes transaction_inputs = 2;
 }
 
-message ProvenTransactionBatch {
-    // Encoded using [winter_utils::Serializable] implementation for
-    // [miden_protocol::transaction::proven_tx::ProvenTransaction].
-    bytes encoded = 1;
+message TransactionBatch {
```
Mirko-von-Leipzig, line 23: The suggestion here I think was to re-add the `proven_batch` property, and have the others be optional so we can drop them at some point.

crates/rpc/src/server/api.rs
Mirko-von-Leipzig, line 1: @PhilippGackstatter if you could throw an eye on the process here to ensure I'm checking the correct things.

The state itself is checked in the mempool, so here we really just want to ensure that the batch and its transactions are valid and the reference block is correct iiuc.

crates/rpc/src/server/api.rs
```diff
@@ -387,47 +388,81 @@ impl api_server::Api for RpcService {
...
+            .await?
+            .into_inner()
+            .block_header
+            .expect("store should always send block header");
+        let reference_commitment: Word = reference_header
+            .chain_commitment
+            .expect("store should always fill block header")
+            .try_into()
+            .expect("store Word should be okay");
+        if reference_commitment != proof.reference_block_commitment() {
```
Mirko-von-Leipzig: Is this correct?

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -127,22 +215,40 @@ impl TransactionGraph {
...
+                continue;
+            }
 
-        let reverted = self
-            .inner
-            .revert_node_and_descendants(transaction)
-            .into_iter()
-            .map(|tx| tx.id())
-            .collect();
+            let x = self.inner.revert_node_and_descendants(transaction);
```
sergerad:
```suggestion
            let x = self.inner.revert_node_and_descendants(revert);
```

Should this be revert instead of transaction?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -229,22 +230,59 @@ impl Mempool {
...
 
         Ok(self.chain_tip)
     }
 
+    #[instrument(target = COMPONENT, name = "mempool.add_user_batch", skip_all)]
+    pub fn add_user_batch(
+        &mut self,
+        txs: &[Arc<AuthenticatedTransaction>],
+    ) -> Result<BlockNumber, MempoolSubmissionError> {
+        assert!(!txs.is_empty(), "Cannot have a batch with no transactions");
```
sergerad, line 257: Just checking we want to crash here instead of return error

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -85,16 +89,100 @@ impl TransactionGraph {
...
+        for tx in &txs {
+            self.txs_user_batch.insert(*tx, batch_id);
+        }
+        self.user_batch_txs.insert(batch_id, txs);
+
+        Ok(())
+    }
+
+    pub fn select_batch(&mut self, budget: BatchBudget) -> Option<SelectedBatch> {
+        self.select_user_batch().or_else(|| self.select_conventional_batch(budget))
```
sergerad, line 127: Might want some doc comments to make it clear that budget is intended to only relevant for conventional batches.

sergerad: Also are we OK with user batches always taking priority over conventional here?

crates/block-producer/src/mempool/mod.rs
```diff
@@ -229,22 +230,59 @@ impl Mempool {
...
+
+        if self.unbatched_transactions_count() + txs.len() >= self.config.tx_capacity.get() {
```
sergerad: nit: we reject if we are at capacity, rather than over capacity

crates/block-producer/src/mempool/graph/transaction.rs
sergerad: Also wondering if we need to prevent user batches of size 1 (or some other limit). Unsure if that is relevant to this PR just a general thought.

crates/block-producer/src/server/mod.rs
```diff
@@ -361,14 +355,33 @@ impl BlockProducerRpcServer {
...
+        // We assume that the rpc component has verified everything, including the transaction
+        // proofs.
+
+        let mut txs = Vec::with_capacity(batch.transactions().len());
+        for tx in batch.transactions() {
+            let inputs = self
+                .store
+                .get_tx_inputs(tx)
+                .await
+                .map_err(MempoolSubmissionError::StoreConnectionFailed)?;
```
sergerad, line 367: Would it be worth to do these queries concurrently with something like `futures::future::try_join_all`?

sergerad: Or a batch endpoint. Probably not worth it based on the number of txns in a batch?

crates/block-producer/src/mempool/mod.rs
Mirko-von-Leipzig: I assumed that one cannot build a `ProvenBatch` without one, so this would indicate an internal bug somewhere. But maybe that's a poor assumption.

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: > Also are we OK with user batches always taking priority over conventional here?

I'm unsure, but at the moment it doesn't matter much. If its a concern we can make it random -- I was thinking maybe that's best.

> Also wondering if we need to prevent user batches of size 1 (or some other limit). Unsure if that is relevant to this PR just a general thought.

Good question. I'm unsure 😬 I wonder if that makes some user loop more difficult i.e. they always submit user batches, but sometimes they don't have many transactions to bundle..

Probably we would want _some_ limit even in the future? cc @bobbinth

crates/block-producer/src/server/mod.rs
Mirko-von-Leipzig: Good point, tbh I didn't even consider it. I think simple is better for now, there's also the thought that this makes it equivalent to normal transactions i.e. submitting a batch of `N` isn't unfairly faster, or consumes more resources, than `N` single ones.

I don't feel strongly about this.

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: Yes, thank you!

Mirko-von-Leipzig: Thanks! Good catch.

proto/proto/types/transaction.proto
bobbinth: This is probably fine for now, but in the next release, we'll need to change this to something like `ProvenTransactionBatch` which would contain the `ProvenBatch` struct + optional data that would allow us to re-execute all transactions in the batch and validate that the proven batch is correct (similar to how we do it for proven transactions).

crates/block-producer/src/mempool/graph/batch.rs
```diff
@@ -134,9 +134,9 @@ impl BatchGraph {
     ///
     /// Panics if the batch does not exist, or has existing ancestors in the batch
     /// graph.
-    pub fn prune(&mut self, batch: BatchId) {
-        self.inner.prune(batch);
+    pub fn prune(&mut self, batch: BatchId) -> SelectedBatch {
```
igamigo, line 157: nit: Would be nice to update the doc comments to say that this returns the pruned batch now

crates/block-producer/src/mempool/graph/dag.rs
```diff
@@ -219,14 +219,14 @@ where
     /// # Panics
     ///
     /// Panics if this node has any ancestor nodes, or if this node was not selected.
-    pub fn prune(&mut self, id: N::Id) {
```
igamigo, line 236: Same here

crates/block-producer/src/mempool/graph/transaction.rs
igamigo: > Also are we OK with user batches always taking priority over conventional here?

Would having fees solve this?

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -85,16 +89,100 @@ impl TransactionGraph {
...
+            let mut selected = SelectedBatch::builder();
+
+            let txs = self
+                .user_batch_txs
+                .get(&candidate)
+                .cloned()
+                .expect("bi-directional mapping should be coherent");
+
+            for tx in txs {
+                let Some(tx) = self.inner.selection_candidates().get(&tx).copied() else {
```
igamigo, line 150: Maybe this calls for a `get_selection_candidate()`? So that you avoid re-allocating the map every iteration through `selection_candidates()`

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -85,16 +89,100 @@ impl TransactionGraph {
...
+    }
+
+    fn select_conventional_batch(&mut self, mut budget: BatchBudget) -> Option<SelectedBatch> {
         let mut selected = SelectedBatch::builder();
 
-        while let Some((id, tx)) = self.inner.selection_candidates().pop_first() {
-            if budget.check_then_subtract(tx) == BudgetStatus::Exceeded {
+        loop {
+            // Select arbitrary candidate which is _not_ part of a user batch.
+            let candidates = self.inner.selection_candidates();
```
igamigo, line 176: This is not important now and I know this was pre-existing but re-allocating on every iteration here as well is probably not ideal

crates/block-producer/src/server/mod.rs
```diff
@@ -361,14 +355,33 @@ impl BlockProducerRpcServer {
...
-    ) -> Result<proto::blockchain::BlockNumber, SubmitProvenBatchError> {
-        let _batch = ProvenBatch::read_from_bytes(&request.encoded)
-            .map_err(SubmitProvenBatchError::Deserialization)?;
+        request: proto::transaction::TransactionBatch,
+    ) -> Result<proto::blockchain::BlockNumber, MempoolSubmissionError> {
+        let batch = ProposedBatch::read_from_bytes(&request.proposed_batch)
+            .map_err(MempoolSubmissionError::TransactionDeserializationFailed)?;
+
+        // We assume that the rpc component has verified everything, including the transaction
+        // proofs.
```
igamigo, line 359: nit, feel free to disregard: this is a bit redundant considering the safety comment near the end of the function

crates/rpc/src/server/api.rs
```diff
@@ -387,47 +388,81 @@ impl api_server::Api for RpcService {
...
+        // Verify the reference header matches the canonical chain.
+        let reference_header = self
+            .get_block_header_by_number(Request::new(proto::rpc::BlockHeaderByNumberRequest {
+                block_num: proof.reference_block_num().as_u32().into(),
+                include_mmr_proof: false.into(),
+            }))
+            .await?
+            .into_inner()
+            .block_header
+            .expect("store should always send block header");
```
igamigo: I don't think this is true for blocks beyond the chain tip, so should this be an error?

crates/rpc/src/server/api.rs
igamigo: I don't think it is, here `chain_commitment` is the root of the blockchain MMR, whereas the `reference_block_commitment` sounds more like it's the block hash.

igamigo (approved):
LGTM! I left some mostly minor, non-blocking comments. Not sure if you tested already but jsut in case, I'm going to run the client integration tests and report back.

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: It would give us a strategy to use, so yes fees solve this imo. Though it will be potentially complex to implement.

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: I agree; I think I've got a potential solve for this, but since this not performance critical at our size (yet), I'll try it out in a follow-up PR.

crates/block-producer/src/mempool/graph/transaction.rs
Mirko-von-Leipzig: I'll refactor this in a follow-up PR.

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: Okay I think I fixed it. The proto header elides the block commitment which confused me :)

proto/proto/types/transaction.proto
Mirko-von-Leipzig: I've added it in now. Naming might still be off, e.g. `TransactionBatch` but I think maybe its more useful to end users than `Batch`.. maybe not though.

crates/rpc/src/server/api.rs
```diff
@@ -427,44 +428,100 @@ impl api_server::Api for RpcService {
...
+        // really check that the calculated proof matches the one given in the request.
+        let expected_proof = LocalBatchProver::new(MIN_PROOF_SECURITY_LEVEL)
+            .prove(proposed_batch.clone())
+            .map_err(|err| {
+                Status::invalid_argument(err.as_report_context("proposed block proof failed"))
+            })?;
+
+        if expected_proof != proven_batch {
+            return Err(Status::invalid_argument("batch proof did not match proposed batch"));
+        }
```
Mirko-von-Leipzig, line 486: Is this the idea? I'm unsure on how else to align the proof with the batch.. unless I also compare headers and then just assume?

Can a proof differ based on some other variables e.g. time, rng? Or is this safe to do.

cc @PhilippGackstatter

crates/rpc/src/server/api.rs
```diff
@@ -427,44 +428,100 @@ impl api_server::Api for RpcService {
...
+            .transpose()?
+            .ok_or_else(|| {
+                Status::invalid_argument(format!(
+                    "unknown reference block {}",
+                    expected_proof.reference_block_num()
+                ))
+            })?;
+        if reference_header.commitment() != expected_proof.reference_block_commitment() {
+            return Err(Status::invalid_argument(format!(
+                "batch reference commitment {} at block {} does not match canonical chain's commitemnt of {}",
```
sergerad:
```suggestion
                "batch reference commitment {} at block {} does not match canonical chain's commitment of {}",
```

crates/block-producer/src/server/mod.rs
```diff
@@ -353,12 +347,34 @@ impl BlockProducerRpcServer {
      )]
     async fn submit_proven_batch(
```
sergerad, line 348: The doc comment for this function may be stale

crates/rpc/src/server/api.rs
Mirko-von-Leipzig: Strange that it wasn't picked up as a typo. Maybe because its in a string? Thanks!

crates/block-producer/src/server/mod.rs
Mirko-von-Leipzig: There is no doc comment? Or do you mean in the proto file?

crates/rpc/src/server/api.rs
PhilippGackstatter: That seems fine and sufficient. The "proof" (which is not a cryptographic one now anyway) should be deterministic based on the input (the proposed batch), since it's really just destructuring the proposed batch, verifying transaction proofs and constructing the proven batch.

crates/rpc/src/server/api.rs
```diff
@@ -427,44 +428,100 @@ impl api_server::Api for RpcService {
...
+                    "unknown reference block {}",
+                    expected_proof.reference_block_num()
+                ))
+            })?;
+        if reference_header.commitment() != expected_proof.reference_block_commitment() {
+            return Err(Status::invalid_argument(format!(
+                "batch reference commitment {} at block {} does not match canonical chain's commitment of {}",
+                expected_proof.reference_block_num(),
+                expected_proof.reference_block_commitment(),
+                reference_header.commitment()
```
PhilippGackstatter, line 510:
```suggestion
                "batch reference commitment {} at block {} does not match canonical chain's commitment of {}",
                expected_proof.reference_block_commitment(),
                expected_proof.reference_block_num(),
                reference_header.commitment()
```

Nit: I think these need to be swapped

crates/rpc/src/server/api.rs
```diff
@@ -427,44 +428,100 @@ impl api_server::Api for RpcService {
...
+            )));
+        }
+
```
PhilippGackstatter, line 513: Is the batch expiration deliberately not checked? This would also be a pretty cheap check, assuming the latest block is easily retrievable here. E.g. if the batch expiration is already older than the latest block, we should be able to discard it directly.

crates/rpc/src/server/api.rs
PhilippGackstatter: Looks good to me. I left another comment re batch expiration, but your call where/if to do that.

crates/block-producer/src/server/mod.rs
sergerad: I got the wrong `async fn submit_proven_batch(`

proto/proto/types/transaction.proto
```diff
@@ -10,70 +10,83 @@ import "types/primitives.proto";
...
-    // [miden_protocol::transaction::TransactionInputs].
-    optional bytes transaction_inputs = 2;
+  // The transaction proof.
+  //
+  // Encoded using [miden_protocol::transaction::ProvenTransaction::to_bytes].
+  bytes transaction = 1;
+  // The inputs used for the transaction proof.
+  //
+  // Encoded using [miden_protocol::transaction::TransactionInputs::to_bytes].
+  optional bytes transaction_inputs = 2;
```
bobbinth, line 20: I would maybe expand on the comment here to explain why we need to provide `transaction_inputs`.

proto/proto/types/transaction.proto
```diff
@@ -10,70 +10,83 @@ import "types/primitives.proto";
...
+  // The batch contents of the given proof.
+  //
+  // Encoded using [miden_protocol::batch::ProposedBatch::to_bytes].
+  optional bytes proposed_batch = 2;
+  // The transaction inputs for each transaction in the batch.
+  //
+  // Must match the transaction ordering in the batch.
+  //
+  // Encoded using [miden_protocol::transaction::TransactionInputs::to_bytes].
+  repeated bytes transaction_inputs = 3;
```
bobbinth, line 37: Similar comment as above: would be great to expand the comments to explain how these fields are used.

Separately, I do think eventually this should be a single field that is either a separate structured message or a single serialized struct that contains all the data that we need to re-execute the batch.

crates/block-producer/src/mempool/graph/batch.rs
```diff
@@ -98,10 +98,14 @@ impl BatchGraph {
     ///
     /// Batches are returned in reverse-chronological order.
     pub fn revert_expired(&mut self, chain_tip: BlockNumber) -> Vec<SelectedBatch> {
-        let reverted = self.inner.revert_expired_unselected(chain_tip);
+        // We only revert transactions which are _not_ included in batches.
+        let mut to_revert = self.inner.expired(chain_tip);
+        to_revert.retain(|batch| !self.inner.is_selected(batch));
```
bobbinth, line 103: Is the comment here correct? Should it be "We only revert batches which are not included in blocks"?

crates/rpc/src/server/api.rs
```diff
@@ -427,44 +428,100 @@ impl api_server::Api for RpcService {
...
+            .into_inner()
+            .block_header
+            .map(BlockHeader::try_from)
+            .transpose()?
+            .ok_or_else(|| {
+                Status::invalid_argument(format!(
+                    "unknown reference block {}",
+                    expected_proof.reference_block_num()
+                ))
+            })?;
```
bobbinth, line 504: We don't do a similar check for transactions - i.e., we don't check if the transaction's reference block is actually in the chain (this is done by the block producer). So, if we do want to keep it here, probably makes to add a similar check for transactions.

But also, going to the store from here introduces non-trivial amount of work. So, maybe we should cache some number of recent block headers (e.g,. 1 million) in the RPC so that for most transactions/batches, we don't need to go to the store.

Separately, I wonder if we should keep some kind of an LRU cache for invalid batches/transactions. The attack vector here is that someone could generate a single proven tx/batch that references some invalid block, and then submit it many times. This way, they attacker pays the cost of generating the proof only once, but the node would be forced to verify it many times (maybe thousands) which may also include going to the store etc.

Let's create issues for these.

crates/rpc/src/server/api.rs
```diff
@@ -427,44 +428,100 @@ impl api_server::Api for RpcService {
...
 
+        // Verify batch transaction proofs.
+        //
+        // Need to do this because ProvenBatch has no real kernel yet, so we can only
+        // really check that the calculated proof matches the one given in the request.
+        let expected_proof = LocalBatchProver::new(MIN_PROOF_SECURITY_LEVEL)
+            .prove(proposed_batch.clone())
+            .map_err(|err| {
+                Status::invalid_argument(err.as_report_context("proposed block proof failed"))
+            })?;
```
bobbinth, line 482: Should we create an issue for this so that we come back and implement it properly once batch kernel proofs are done?

crates/rpc/src/server/api.rs
bobbinth: I think this is something we should add in a follow-up PR, potentially together with keeping a cache of most recent block headers that I mentioned in another comment.

crates/block-producer/src/server/mod.rs
```diff
@@ -353,12 +347,34 @@ impl BlockProducerRpcServer {
...
+            // SAFETY: We assume that the rpc component has verified the transaction proofs, as well
+            // as the batch integrity itself.
+            let tx = AuthenticatedTransaction::new_unchecked(Arc::clone(tx), inputs)
+                .map(Arc::new)
+                .map_err(MempoolSubmissionError::StateConflict)?;
+            txs.push(tx);
+        }
 
-        todo!();
+        self.mempool.lock().await.lock().await.add_user_batch(&txs).map(Into::into)
```
bobbinth, line 377: This is fine for now, but it has two issues:

- First, we are not respecting all of the batch parameters. Specifically, a user-submitted batch as a specific reference block number, but IIUC, we are ignoring it here. So, a user-submitted batch is different from a batch that is built by the node because the node always picks the latest block to "anchor" the batch against, while the user may have a different "anchor" block. In the future, we may have additional "user-defined" properties for a batch that may not be inferred from just the transaction list.
- Second, when we eventually remove the training wheels, we will not be able to represent a batch as a list of regular transactions. In a way a batch is atomic and we can "unwrap" it back into a list of transactions because we won't have all the underlying individual transaction details (for a proven batch, once we remove the training wheels, the only thing we'll have per transaction is `TransactionHeader`, which for example, would not have a transaction proof).

The first point we can address relatively shortly, but the second point could probably wait even util after mainnet.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -230,22 +231,61 @@ impl Mempool {
...
+        }
+
+        // Ensure the batch doesn't exceed the mempool budget for batches.
+        let mut budget = self.config.batch_budget;
+        for tx in txs {
+            if budget.check_then_subtract(tx) == BudgetStatus::Exceeded {
+                // TODO: better error plox.
+                return Err(MempoolSubmissionError::CapacityExceeded);
+            }
+        }
```
bobbinth, line 270: A couple of thoughts on this:

- Ideally, we should move this check into the RPC component so that batches that exceed the limit would not make it all the way to the mempool. This would be similar to check the batch proof, consistency against the chain etc.
- Once we have batch proofs, the budgets for "internal" vs. "user-provided" batches would be different. Specifically, because we don't need to generate batch proofs ourselves, we could be much less conservative on the number of txs that people include in their batches. We will need to check things like the number of output notes produced, number of accounts updated etc. but if people want to batch 1K transactions into their batches while staying within these limits, we don't really care too much.

Let's create issues for these.

crates/block-producer/src/mempool/mod.rs
```diff
@@ -230,22 +231,61 @@ impl Mempool {
...
+
+        for tx in txs {
+            self.authentication_staleness_check(tx.authentication_height())?;
+            self.expiration_check(tx.expires_at())?;
+        }
```
bobbinth, line 275: This is fine for now, but eventually (once we have batch proofs), we should check these at the batch level as per-transaction checks may not always be correct here. For example, in the future, we may allow batches to include expired transactions - and these transactions would just be ignored at the block level (we don't allow this currently because this would introduce significant complexity).

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -75,6 +76,13 @@ pub struct TransactionGraph {
     /// These are batch or block proving errors in which the transaction was a part of. This is
     /// used to identify potentially buggy transactions that should be evicted.
     failures: HashMap<TransactionId, u32>,
+
+    /// Defines the transactions that belong to a user batch.
+    user_batch_txs: HashMap<BatchId, Vec<TransactionId>>,
+    /// A mapping of transactions to their user batch (if any).
+    ///
+    /// Inverse map of `user_batch_txs`.
+    txs_user_batch: HashMap<TransactionId, BatchId>,
```
bobbinth, line 85: nit: we might want to encapsulate these into a single struct as I'm assuming these get updated atomically.

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -85,16 +93,100 @@ impl TransactionGraph {
         self.inner.append(tx)
     }
 
-    pub fn select_batch(&mut self, mut budget: BatchBudget) -> Option<SelectedBatch> {
+    pub fn append_user_batch(
+        &mut self,
+        batch: &[Arc<AuthenticatedTransaction>],
+    ) -> Result<(), StateConflict> {
```
bobbinth, line 99: Would be good to add some doc comments here.

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -85,16 +93,100 @@ impl TransactionGraph {
...
+        for tx in &txs {
+            self.txs_user_batch.insert(*tx, batch_id);
+        }
+        self.user_batch_txs.insert(batch_id, txs);
+
+        Ok(())
+    }
+
+    pub fn select_batch(&mut self, budget: BatchBudget) -> Option<SelectedBatch> {
+        self.select_user_batch().or_else(|| self.select_conventional_batch(budget))
```
bobbinth, line 127: nit: I would probably use "user" and "internal" batches (rather than "conventional" batches).

crates/block-producer/src/mempool/graph/transaction.rs
bobbinth: I think it is OK to have batches of 1 transactions as people are unlikely to to do them on purpose (they would first need to prove TXs and then prove batches - so, it is non-negligible cost). The fees would also address this as some of the tx-related fees would get amortized over large batches.

Regarding batch selection strategy - I think this is fine for now, but it'll need to change once we have fees so that we prioritize based on some value metric. But that's a whole other discussion.

And agreed that adding some doc comments would be nice.

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -85,16 +93,100 @@ impl TransactionGraph {
...
+                .get(&candidate)
+                .cloned()
+                .expect("bi-directional mapping should be coherent");
+
+            for tx in txs {
+                let Some(tx) = self.inner.selection_candidates().get(&tx).copied() else {
+                    // Rollback this batch selection since it cannot complete.
+                    for tx in selected.txs.into_iter().rev() {
+                        self.inner.deselect(tx.id());
+                    }
```
bobbinth, line 154: Would be good to explain why a batch may not be able to complete here (maybe we can do this once we break this code out into a separate function).

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -85,16 +93,100 @@ impl TransactionGraph {
...
+            }
+
+            assert!(!selected.is_empty(), "User batch should not be empty");
+            return Some(selected.build());
+        }
+
+        None
+    }
+
+    fn select_conventional_batch(&mut self, mut budget: BatchBudget) -> Option<SelectedBatch> {
```
bobbinth, line 171: nit: maybe `select_internal_batch()`?

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -107,44 +199,77 @@ impl TransactionGraph {
...
+    /// transactions, then call this.
     pub fn revert_expired(&mut self, chain_tip: BlockNumber) -> HashSet<TransactionId> {
-        self.inner
-            .revert_expired_unselected(chain_tip)
-            .into_iter()
-            .map(|tx| tx.id())
-            .collect()
+        // We only revert transactions which are _not_ included in batches.
+        let mut to_revert = self.inner.expired(chain_tip);
+        to_revert.retain(|tx| !self.inner.is_selected(tx));
```
bobbinth, line 218: Is the comment still correct here? Wouldn't this revert transactions in user-defined batches (and the whole batches themselves)?

crates/block-producer/src/mempool/graph/transaction.rs
```diff
@@ -107,44 +199,77 @@ impl TransactionGraph {
...
+                self.failures.remove(&tx.id());
+
+                // Note that this is a pretty rough shod approach. We just dump the entire batch of
+                // transactions in, which will result in at least the current
+                // transaction being duplicated in `to_revert`. This isn't a concern
+                // though since we skip already processed transactions at the top of the loop.
+                if let Some(batch) = self.txs_user_batch.remove(&tx.id()) {
+                    if let Some(batch) = self.user_batch_txs.remove(&batch) {
+                        to_revert.extend(batch);
+                    }
```
bobbinth, line 262: nit: using `batch` in both `Some()` expressions here is a bit confusing. I'd maybe use `batch_id` in the outer one, and `batch_txs` in the inner one.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - but I think these can be addressed in follow-up PRs. So, I'll merge this as is.

Mirko-von-Leipzig:
I've created #1995 to track the comments left post-merge.

PR #1852, opened 2026-03-27
fix: Return error if no tx inputs provided in submit_proven_transaction

### LINKED ISSUE #1850: Require TransactionInputs or error on None in submit_proven_transaction
We need to either:
1. Update client to always sent tx inputs and update RPC to not have optional tx inputs; or
2. Update RPC to reject requests with tx inputs == None

https://github.com/0xMiden/node/blob/ff2a1cdfdcdbf2690ab638545fb811c829684375/crates/rpc/src/server/api.rs#L415-L419

sergerad:
Closes #1850.

### DIFF
--- a/crates/rpc/src/server/api.rs
+++ b/crates/rpc/src/server/api.rs
@@ -412,7 +412,8 @@ impl api_server::Api for RpcService {
             ))
         })?;
 
-        // Transaction inputs must be provided in order to allow for re-execution.
+        // Transaction inputs must be provided in order to allow for transaction re-execution via
+        // the Validator.
         if request.transaction_inputs.is_some() {
             self.validator.clone().submit_proven_transaction(request.clone()).await?;
         } else {


crates/rpc/src/server/api.rs
```diff
@@ -412,10 +412,12 @@ impl api_server::Api for RpcService {
...
         })?;
 
-        // If transaction inputs are provided, re-execute the transaction to validate it.
+        // Transaction inputs must be provided in order to allow for transaction re-execution via
+        // the Validator.
         if request.transaction_inputs.is_some() {
-            // Re-execute the transaction via the Validator.
             self.validator.clone().submit_proven_transaction(request.clone()).await?;
+        } else {
+            return Err(Status::invalid_argument("Transaction inputs must be provided"));
```
Mirko-von-Leipzig, line 420: fyi this can be written as 

```rust
let request = request.transaction_inputs.ok_or(Err(...))?;
```

sergerad: Yes but we don't actually want to extract `transaction_inputs` here, we just pass on the request
```rust
            self.validator.clone().submit_proven_transaction(request.clone()).await?;
```

PR #1857, opened 2026-03-29
chore: Reduce block proving timeout worst case

### LINKED ISSUE #1855: Improve worst case block proof timeout
Currently a block proof is only considered failed after `10` retries, with each retry timing out after `4 minutes`. While we _should_ timeout each gRPC/proof attempt, we likely want to bound the _overall_ block in terms of time as well. So retries, attempt time and overall time should all be limited.

> Note that this combination means we have a worst case of 40minutes allowed for a proof.
>
> An alternative is to have two separate timeouts. One for each block proof attempt, and one overall one (instead of max retries). We can discuss/do that in a separate PR though, its not really impactful yet.

_Originally posted by @Mirko-von-Leipzig in https://github.com/0xMiden/node/pull/1725#discussion_r3001149521_

sergerad:
Closes #1855.

## Changes
- Manage block proving timeouts with per-proving request timeout + overall across all retries for one block.
- Adds a commit missed in last block proving PR (return new_tip instead of vec<BlockNumber>).
- Use AtomicU32 newtype for sending proven tip from proof scheduler (fixes possibility of sending old tip)

### DIFF
--- a/crates/store/src/server/proof_scheduler.rs
+++ b/crates/store/src/server/proof_scheduler.rs
@@ -41,8 +41,8 @@ const BLOCK_PROVE_ATTEMPT_TIMEOUT: Duration = Duration::from_mins(4);
 /// Overall timeout for proving a single block (across all retries).
 const BLOCK_PROVE_OVERALL_TIMEOUT: Duration = Duration::from_mins(12);
 
-/// Delay between retry attempts on transient errors.
-const RETRY_DELAY: Duration = Duration::from_secs(5);
+/// Maximum number of proving attempts per block before giving up.
+const MAX_PROVE_ATTEMPTS: u32 = 3;
 
 /// Default maximum number of blocks being proven concurrently.
 pub const DEFAULT_MAX_CONCURRENT_PROOFS: NonZeroUsize = NonZeroUsize::new(8).unwrap();
@@ -242,7 +242,9 @@ async fn prove_block(
                 },
             }
 
-            tokio::time::sleep(RETRY_DELAY).await;
+            if attempt >= MAX_PROVE_ATTEMPTS {
+                anyhow::bail!("block {} failed after {attempt} attempts", block_num.as_u32());
+            }
         }
     })
     .await


crates/store/src/db/mod.rs
```diff
@@ -829,41 +829,42 @@ impl Db {
...
+        if candidate <= current_tip {
             return Err(DatabaseError::DataCorrupted(format!(
-                "block {candidate} is proven but not marked in-sequence while the tip is at {tip}"
+                "block {candidate} is proven but not marked in-sequence while the tip is at {current_tip}"
             )));
         }
-        if candidate == tip + 1 {
-            tip = candidate;
-            newly_in_sequence.push(candidate);
+        if candidate == new_tip + 1 {
```
Mirko-von-Leipzig:
```suggestion
        if candidate == new_tip.child() {
```

crates/store/src/server/proof_scheduler.rs
```diff
@@ -173,45 +176,47 @@ async fn prove_block(
...
+
+                    // Mark the block as proven and advance the sequence in the database.
+                    let new_tip = db.mark_proven_and_advance_sequence(block_num).await?;
+                    if new_tip == block_num {
+                        info!(target = COMPONENT, block.number = %block_num, "Block proven");
+                    } else {
+                        info!(
+                            target = COMPONENT,
+                            block.number = %block_num,
+                            %new_tip,
```
Mirko-von-Leipzig: We should settle on a conventional name for this e.g. `proven_chain_tip`

crates/store/src/server/proof_scheduler.rs
```diff
@@ -166,52 +169,59 @@ async fn run(
...
+            let attempt_span = tracing::info_span!(
+                target: COMPONENT,
+                "prove_attempt",
+                attempt,
+                error = tracing::field::Empty,
+                timed_out = tracing::field::Empty,
+            );
+            let result = tokio::time::timeout(
+                BLOCK_PROVE_ATTEMPT_TIMEOUT,
+                generate_block_proof(db, block_prover, block_num),
```
SantiagoPittella, line 217: If this fails fast, consideraing the now we don't have a maximum amount of retries, this could spam the DB.

sergerad: Looked at adding a retry delay but think failing on a maximum attempts is a better approach (fail harder/faster).

PR #1858, opened 2026-03-29
fix: Genesis(SignedBlock)

### LINKED ISSUE #1844: Move GenesisBlock(ProvenBlock) to GenesisBlock(SignedBlock))
I think this is fine for now, but why is the genesis block defined as a proven block? It can't have a ZK proof and technically should always be just a "signed block".

Let's create an issue to fix this.

_Originally posted by @bobbinth in https://github.com/0xMiden/node/pull/1725#discussion_r2992449639_

```rust
// Current state
pub struct GenesisBlock(ProvenBlock);
```

sergerad:
Closes #1844.

### DIFF
--- a/bin/node/src/commands/store.rs
+++ b/bin/node/src/commands/store.rs
@@ -7,7 +7,7 @@ use miden_node_store::{DEFAULT_MAX_CONCURRENT_PROOFS, Store};
 use miden_node_utils::clap::{GrpcOptionsInternal, StorageOptions};
 use miden_node_utils::fs::ensure_empty_directory;
 use miden_node_utils::grpc::UrlExt;
-use miden_protocol::block::ProvenBlock;
+use miden_protocol::block::SignedBlock;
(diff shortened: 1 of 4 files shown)


bobbinth (commented):
Looks good! Thank you! 

One thing to think through: what are the implications of this from an external user standpoint? I think there may be a couple:

- The `GetBlockByNumber` endpoint would return a `SignedBlock` for the genesis block and `ProvenBlock` for everything else, right? If so, we probably should make the `MaybeBlock` message more structured to contain this information. If that's not the case (e.g., if we always get the `SignedBlock` from this endpoint), we should document this explicitly and maybe add a separate endpoint for retrieving block proofs (in a different PR). Or maybe we can change this endpoint to return both the `SignedBlock` and the `BlockProof`?
- Does this impact the `GetBlockHeaderByNumber` in any way? I don't think it should, but let's double-check.
- Nothing else comes to mind, but maybe there is some other implication?

PR #1860, opened 2026-03-30
fix: Update SyncChainMmr endpoint upper bound logic

### LINKED ISSUE #1842: Make finality and block_to mutually exclusive in sync_chain_mmr endpoint
I still think it would be good to make `finality` and `block_to` mutually exclusive - but I would leave this for a follow-up PR.

To illustrate the idea, in Rust, it would look something like this:
```rust
struct SyncChainMmrRequest {
    block_from: u32,
    block_to: UpperBlockBound,
}

// the name is just illustrative
enum UpperBlockBound {
    Block(u32),
    LastCommitted,
    LastProven,
}
```

_Originally posted by @bobbinth in https://github.com/0xMiden/node/pull/1725#discussion_r2985825543_

sergerad:
Closes #1842.

### DIFF
--- a/proto/proto/rpc.proto
+++ b/proto/proto/rpc.proto
@@ -82,14 +82,12 @@ service Api {
     // Returns info which can be used by the client to sync up to the tip of chain for the notes
     // they are interested in.
     //
-    // Client specifies the `note_tags` they are interested in, and the block height from which to
-    // search for new for matching notes for. The request will then return the next block containing
-    // any note matching the provided tags.
+    // Client specifies the `note_tags` they are interested in, and the block range to search.
+    // The response contains all blocks with matching notes that fit within the response payload
+    // limit, along with each note's metadata and inclusion proof.
     //
-    // The response includes each note's metadata and inclusion proof.
-    //
-    // A basic note sync can be implemented by repeatedly requesting the previous response's block
-    // until reaching the tip of the chain.
+    // If `response.pagination_info.block_num` is less than the requested range end, make
+    // another request starting from `response.pagination_info.block_num + 1` to continue syncing.
     rpc SyncNotes(SyncNotesRequest) returns (SyncNotesResponse) {}
 
     // Returns a list of nullifiers that match the specified prefixes and are recorded in the node.
@@ -461,8 +459,9 @@ message AccountVaultUpdate {
 
 // Note synchronization request.
 //
-// Specifies note tags that requester is interested in. The server will return the first block which
-// contains a note matching `note_tags` or the chain tip.
+// Specifies note tags that the requester is interested in. The server will return blocks
+// containing notes matching `note_tags` within the specified block range, batching as many
+// blocks as fit within the response payload limit.
 message SyncNotesRequest {
     // Block range from which to start synchronizing.
     BlockRange block_range = 1;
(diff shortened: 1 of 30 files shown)


proto/proto/rpc.proto
```diff
@@ -492,30 +492,23 @@ message SyncNotesResponse {
...
-    //
-    // When set to `FINALITY_UNSPECIFIED` or `FINALITY_COMMITTED`, the upper bound is clamped to the chain tip.
-    // When set to `FINALITY_PROVEN`, the upper bound is clamped to the latest proven block.
-    Finality finality = 2;
+    reserved 2;
+
+    // Upper bound for the block range. Determines how far ahead to sync.
+    oneof upper_bound {
+        // Sync up to this specific block number (inclusive), clamped to the effective tip.
+        fixed32 block_num = 3;
```
Mirko-von-Leipzig: nit
```suggestion
        fixed32 block_number = 3;
```

proto/proto/rpc.proto
```diff
@@ -492,30 +492,23 @@ message SyncNotesResponse {
...
-    // The finality level to use when clamping the upper bound of the block range.
-    //
-    // When set to `FINALITY_UNSPECIFIED` or `FINALITY_COMMITTED`, the upper bound is clamped to the chain tip.
-    // When set to `FINALITY_PROVEN`, the upper bound is clamped to the latest proven block.
-    Finality finality = 2;
+    reserved 2;
+
+    // Upper bound for the block range. Determines how far ahead to sync.
+    oneof upper_bound {
+        // Sync up to this specific block number (inclusive), clamped to the effective tip.
```
Mirko-von-Leipzig: What does `effective tip` mean?

--edit-- ah the chain tip.


```suggestion
        // Sync up to this specific block number (inclusive), clamped to the committed chain tip.
```

proto/proto/rpc.proto
```diff
@@ -492,30 +492,23 @@ message SyncNotesResponse {
...
+        fixed32 block_num = 3;
+        // Sync up to the latest committed block (chain tip).
+        bool last_committed = 4;
+        // Sync up to the latest proven block.
+        bool last_proven = 5;
```
Mirko-von-Leipzig: Nit: `_chain_tip` or `latest_` rather?

```suggestion
        bool committed_chain_tip = 4;
        // Sync up to the latest proven block.
        bool proven_chain_tip = 5;
```

proto/proto/rpc.proto
```diff
@@ -492,30 +492,23 @@ message SyncNotesResponse {
...
+    }
```
Mirko-von-Leipzig, line 528: Is this the recommended way to emulate sum types in protobuf?

Haven't seen `bool` used as a placeholder before - what happens if they're manually set to false?

proto/proto/rpc.proto
```diff
@@ -492,30 +492,23 @@ message SyncNotesResponse {
...
+    // Block number from which to synchronize (inclusive). Set this to the last block
+    // already present in the caller's MMR so the delta begins at the next block.
+    fixed32 block_from = 1;
 
-    // The finality level to use when clamping the upper bound of the block range.
-    //
-    // When set to `FINALITY_UNSPECIFIED` or `FINALITY_COMMITTED`, the upper bound is clamped to the chain tip.
-    // When set to `FINALITY_PROVEN`, the upper bound is clamped to the latest proven block.
-    Finality finality = 2;
+    reserved 2;
```
bobbinth: Why do we need `reserved 2` here? Maybe worth adding a comment explaining this?

proto/proto/rpc.proto
sergerad: I see block_num as the convention in the proto
```
➜ grep -r 'fixed32 block_num.*' .
./proto/proto/types/note.proto:    fixed32 block_num = 2;
./proto/proto/types/blockchain.proto:    fixed32 block_num = 1;
./proto/proto/types/blockchain.proto:    optional fixed32 block_num = 1;
./proto/proto/types/blockchain.proto:    fixed32 block_num = 3;
./proto/proto/types/transaction.proto:    fixed32 block_num = 2;
./proto/proto/rpc.proto:        fixed32 block_num = 2;
./proto/proto/rpc.proto:    fixed32 block_num = 3;
./proto/proto/rpc.proto:        fixed32 block_num = 3;
./proto/proto/rpc.proto:    fixed32 block_num = 1;
./proto/proto/rpc.proto:    fixed32 block_num = 2;
./proto/proto/rpc.proto:    fixed32 block_num = 1;
./proto/proto/internal/store.proto:        fixed32 block_num = 2;
./proto/proto/internal/store.proto:     fixed32 block_num = 4;
./proto/proto/internal/store.proto:    optional fixed32 block_num = 3;
./proto/proto/internal/store.proto:    fixed32 block_num = 1;
./proto/proto/internal/store.proto:    optional fixed32 block_num = 3;
./proto/proto/internal/store.proto:    fixed32 block_num = 2;
```

proto/proto/rpc.proto
sergerad: No need because its just for backwards compat. Removed

proto/proto/rpc.proto
sergerad: The value of the boolean is ignored
```rust
        match value {
            UpperBound::BlockNum(block_num) => Ok(Self::BlockNumber(block_num.into())),
            UpperBound::CommittedChainTip(_) => Ok(Self::CommittedChainTip),
            UpperBound::ProvenChainTip(_) => Ok(Self::ProvenChainTip),
        }
```

proto/proto/rpc.proto
Mirko-von-Leipzig: Bleg. Yeah okay, though we should address that imo.

proto/proto/rpc.proto
Mirko-von-Leipzig: I know. But since these are numbered fields, and protobuf _default_ inits fields at the receiver, what actually happens if I send `last_proven = false` -- does it not still get `last_committed = false`.

sergerad: Not entirely sure. Depends on what the generator does under the hood but I thought better to change oneof + enum anyway.

PR #1861, opened 2026-03-30
chore: Cache proven chain tip in State

### LINKED ISSUE #1841: Keep chain tip in memory for sync_chain_mmr endpoint
Not for this PR, but it may be good to keep this info in memory somehow so that we could avoid going to the DB just to look up just to figure out which block number to use. Maybe this could be returned from `State::latest_block_num()` - e.g., as something like:
```rust
let chain_tip = self.state.latest_block_num(finality_level).await;
```

_Originally posted by @bobbinth in https://github.com/0xMiden/node/pull/1725#discussion_r2985667722_

```rust
        // Determine the effective tip based on the requested finality level.
        let effective_tip = match request.finality() {
            proto::rpc::Finality::Unspecified | proto::rpc::Finality::Committed => chain_tip,
            proto::rpc::Finality::Proven => self
                .state
                .db()
                .select_latest_proven_block_num()
                .await
                .map_err(SyncChainMmrError::DatabaseError)?,
        };
```

sergerad:
Closes #1841.

## Cache proven-in-sequence tip to avoid repeated DB queries

Adds a `watch::channel`-based cache for the proven-in-sequence block number, so RPC endpoints requesting `Finality::Proven` no longer hit the database on every call.

### Changes

- Introduces a `Finality` enum (`Committed` / `Proven`) in the store's `State`
- Replaces `State::latest_block_num()` with `State::chain_tip(finality)` which returns the committed tip (from the in-memory MMR) or the proven tip (from the cached watch channel) depending on the requested finality
- `State::load()` initializes the proven tip from the DB and returns a `watch::Sender` that the proof scheduler uses to push updates after `mark_proven_and_advance_sequence` advances the sequence
- Simplifies `sync_chain_mmr` from a match + DB query into a single `chain_tip(finality)` call
- Splits up `Store::serve()` into helper functions `spawn_proof_scheduler()` and `spawn_grpc_servers()`

### DIFF
--- a/crates/store/src/server/rpc_api.rs
+++ b/crates/store/src/server/rpc_api.rs
@@ -235,19 +235,24 @@ impl rpc_server::Rpc for StoreApi {
 
     async fn get_block_by_number(
         &self,
-        request: Request<proto::blockchain::BlockNumber>,
+        request: Request<proto::blockchain::BlockRequest>,
     ) -> Result<Response<proto::blockchain::MaybeBlock>, Status> {
         let request = request.into_inner();
 
         debug!(target: COMPONENT, ?request);
 
-        let block = self
-            .state
-            .load_block(request.block_num.into())
-            .await
-            .map_err(GetBlockByNumberError::from)?;
+        // Load block from state.
+        let block_num = BlockNumber::from(request.block_num);
+        let block = self.state.load_block(block_num).await.map_err(GetBlockByNumberError::from)?;
 
-        Ok(Response::new(proto::blockchain::MaybeBlock { block }))
+        // Load proof from state.
+        let proof = if request.include_proof.unwrap_or_default() {
+            self.state.load_proof(block_num).await.map_err(GetBlockByNumberError::from)?
+        } else {
+            None
+        };
+
+        Ok(Response::new(proto::blockchain::MaybeBlock { block, proof }))
     }
 
     async fn get_account(
--- a/crates/store/src/state/mod.rs
+++ b/crates/store/src/state/mod.rs
@@ -855,6 +855,18 @@ impl State {
         })
     }
 
+    /// Returns the effective chain tip for the given finality level.
+    ///
+    /// - [`Finality::Committed`]: returns the latest committed block number (from in-memory MMR).
+    /// - [`Finality::Proven`]: returns the latest proven-in-sequence block number (cached via watch
+    ///   channel, updated by the proof scheduler).
+    pub async fn chain_tip(&self, finality: Finality) -> BlockNumber {
+        match finality {
+            Finality::Committed => self.inner.read().await.latest_block_num(),
+            Finality::Proven => *self.proven_tip_rx.borrow(),
+        }
+    }
+
     /// Loads a block from the block store. Return `Ok(None)` if the block is not found.
     pub async fn load_block(
         &self,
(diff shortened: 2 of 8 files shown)


crates/store/src/server/mod.rs
```diff
@@ -94,33 +95,90 @@ impl Store {
...
+            }
+        }
+    }
+
+    /// Initializes the block prover client and spawns the proof scheduler as a background task.
+    ///
+    /// Returns the scheduler task handle and the chain tip sender (needed by gRPC services to
+    /// notify the scheduler of new blocks).
+    async fn spawn_proof_scheduler(
+        state: &Arc<State>,
```
Mirko-von-Leipzig: Nit: seems like this could be `&State`, though that doesn't really matter.

crates/store/src/server/proof_scheduler.rs
```diff
@@ -59,13 +59,16 @@ impl ProofTaskJoinSet {
...
     ) {
         let db = Arc::clone(db);
         let block_prover = Arc::clone(block_prover);
         let block_store = Arc::clone(block_store);
-        self.0
-            .spawn(async move { prove_block(&db, &block_prover, &block_store, block_num).await });
+        let proven_tip_sender = proven_tip_sender.clone();
+        self.0.spawn(async move {
+            prove_block(&db, &block_prover, &block_store, &proven_tip_sender, block_num).await
+        });
```
Mirko-von-Leipzig: Nit: often easier to read by moving the clone inside

```suggestion
        self.0.spawn({
            let proven_tip = prove_tip.clone();
            async move {
            prove_block(&db, &block_prover, &block_store, &proven_tip, block_num).await
        }});
```

crates/store/src/server/proof_scheduler.rs
```diff
@@ -98,9 +101,17 @@ pub fn spawn(
     block_prover: Arc<BlockProver>,
     block_store: Arc<BlockStore>,
     chain_tip_rx: watch::Receiver<BlockNumber>,
+    proven_tip_sender: watch::Sender<BlockNumber>,
```
Mirko-von-Leipzig: We should consider standardizing on naming conventions for channels (e.g. above is `xxx_rx`.

I vote `tx|rx` prefix with the suffix for the thing


```suggestion
    tx_proven_tip: watch::Sender<BlockNumber>,
```

crates/store/src/server/rpc_api.rs
```diff
@@ -41,6 +41,18 @@ use crate::server::api::{
     read_root,
     validate_nullifiers,
 };
+use crate::state::Finality;
+
+impl From<proto::rpc::Finality> for Finality {
+    fn from(finality: proto::rpc::Finality) -> Self {
+        match finality {
+            proto::rpc::Finality::Unspecified | proto::rpc::Finality::Committed => {
```
Mirko-von-Leipzig: I'm not sure how I feel about defaulting here for `Unspecified`, but its probably not a train smash.

crates/store/src/state/mod.rs
```diff
@@ -183,14 +198,25 @@ impl State {
...
-            db,
-            block_store,
-            inner,
-            forest,
-            writer,
-            termination_ask,
-        })
+        // Initialize the proven tip from database.
+        let proven_tip = db
+            .select_latest_proven_in_sequence_block_num()
```
Mirko-von-Leipzig: We should consider less of a mouthful - this basically describes what its doing internally, but really we should just call it `db.proven_chain_tip()`

SantiagoPittella (commented):
Could we benefit of using `AtomicU32` inside `State` instead of the channel? Considering that we are only borrowing to get the latest value

Issue #1839, opened 2026-03-25
Add Doc Comments for Raw Block Data Storage


sergerad:
... add doc comments (maybe at the module level), explaining how we save raw block data. I believe currently it is something like this:

- Raw block data is saved under `[store_dir]/[epoch]/block_[block_num].dat` file.
- For proven blocks, the proof data is saved under `[store_dir]/epoch[epoch_num]/proof_[block_num].dat` file.

Where `epoch_num` is the 16 most significant bits of the block number, and both `epoch_num` and `block_num` are formatted as hex strings.

_Originally posted by @bobbinth in https://github.com/0xMiden/node/pull/1725#discussion_r2985650196_

Also this comment: https://github.com/0xMiden/node/pull/1725#discussion_r2985655307

PR #1864, opened 2026-03-30
feat: Add proof to GetBlockByNumber

### LINKED ISSUE #1863: Add block proof to GetBlockByNumber endpoint
Modify the endpoint to allow users to specify whether they want a block proof to also be returned.

Both the fn argument and return type must be updated for this purpose:

```proto
    // Returns raw block data for the specified block number.
    rpc GetBlockByNumber(blockchain.BlockNumber) returns (blockchain.MaybeBlock) {}
```

Relevant comments
- https://github.com/0xMiden/node/pull/1858#pullrequestreview-4027396614
- https://github.com/0xMiden/node/pull/1858#issuecomment-4153441643

sergerad:
### GetBlockByNumber: support optional block proof in response

Closes #1863 and #1839

Updates the `GetBlockByNumber` gRPC endpoint to accept a `BlockRequest` (with optional `include_proof` flag) and return a `MaybeBlock` containing the block and an optional block proof. 

Adds `load_proof` to the store's `BlockStore` and `State` to read proof files from disk when requested.

### DIFF
--- a/crates/store/src/server/rpc_api.rs
+++ b/crates/store/src/server/rpc_api.rs
@@ -239,7 +239,7 @@ impl rpc_server::Rpc for StoreApi {
     async fn get_block_by_number(
         &self,
         request: Request<proto::blockchain::BlockRequest>,
-    ) -> Result<Response<proto::blockchain::Block>, Status> {
+    ) -> Result<Response<proto::blockchain::MaybeBlock>, Status> {
         let request = request.into_inner();
 
         debug!(target: COMPONENT, ?request);
@@ -255,7 +255,7 @@ impl rpc_server::Rpc for StoreApi {
             None
         };
 
-        Ok(Response::new(proto::blockchain::Block { block, proof }))
+        Ok(Response::new(proto::blockchain::MaybeBlock { block, proof }))
     }
 
     async fn get_account(
--- a/proto/proto/types/blockchain.proto
+++ b/proto/proto/types/blockchain.proto
@@ -21,13 +21,6 @@ message ProposedBlock {
     bytes proposed_block = 1;
 }
 
-// Represents a block or nothing.
-message MaybeBlock {
-    // The requested block data encoded using [miden_serde_utils::Serializable] implementation for
-    // [miden_protocol::block::Block].
-    optional bytes block = 1;
-}
-
 // Request for retrieving a block by its number, optionally including the block proof.
 message BlockRequest {
     // The block number of the target block.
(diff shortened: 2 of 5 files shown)


crates/store/src/server/rpc_api.rs
```diff
@@ -237,19 +237,24 @@ impl rpc_server::Rpc for StoreApi {
...
-            .state
-            .load_block(request.block_num.into())
-            .await
-            .map_err(GetBlockByNumberError::from)?;
+        // Load block from state.
+        let block_num = BlockNumber::from(request.block_num);
+        let block = self.state.load_block(block_num).await.map_err(GetBlockByNumberError::from)?;
+
+        // Load proof from state.
+        let proof = if request.include_proof.unwrap_or(false) {
```
Mirko-von-Leipzig:
```suggestion
        let proof = if request.include_proof.unwrap_or_default() {
```

proto/proto/types/blockchain.proto
```diff
@@ -28,6 +28,24 @@ message MaybeBlock {
     optional bytes block = 1;
 }
 
+// Request for retrieving a block by its number, optionally including the block proof.
+message GetBlockByNumberRequest {
```
bobbinth: I thought we tried to move away from naming messages using action verbs. Could this be just `BlockRequest` or something like that?

proto/proto/types/blockchain.proto
```diff
@@ -28,6 +28,24 @@ message MaybeBlock {
...
+    // The block number of the target block.
+    fixed32 block_num = 1;
+    // Whether to include the block proof in the response.
+    optional bool include_proof = 2;
+}
+
+// Response containing the block data and optionally its proof.
+message GetBlockByNumberResponse {
```
bobbinth: Same comment as above here.

Also, it may be good to clarify what happens if:
- If we could find a block with the requested number.
- If the block for which we requested a proof is not proven yet.

proto/proto/types/blockchain.proto
```diff
@@ -28,6 +28,24 @@ message MaybeBlock {
...
+    // The requested block data encoded using [miden_serde_utils::Serializable] implementation for
+    // [miden_protocol::block::Block].
+    optional bytes block = 1;
```
bobbinth, line 39: I don't think we have `miden_protocol::block::Block`. Was this meant to be `SignedBlock`?

crates/store/src/blocks.rs
```diff
@@ -96,6 +105,14 @@ impl BlockStore {
         fs_err::write(block_path, data)
     }
 
+    pub async fn load_proof(&self, block_num: BlockNumber) -> std::io::Result<Option<Vec<u8>>> {
```
bobbinth: Shouldn't this go into the "Proof storage" section below?

PR #1872, opened 2026-04-01
chore: Split apply_block

sergerad:
Splits the apply_block fn to add more instrumentation

```
rpc [ 3.48ms | 0.00% / 100.00% ] rpc.service: "store.BlockProducer" | rpc.method: "ApplyBlock"
┕━ apply_block [ 3.48ms | 0.00% / 100.00% ]
   ┝━ validate_block_header [ 1.43ms | 41.18% ]
   ┝━ compute_tree_mutations [ 1.14ms | 32.69% ]
   ┝━ build_note_records [ 12.5µs | 0.36% ]
   ┝━ store.block_store.save_block [ 63.2µs | 1.82% ] block_num: BlockNumber(8) | block_size: 357
   ┝━ await_db_readiness [ 12.2µs | 0.35% ]
   ┝━ acquire_inner_write_lock [ 9.62µs | 0.28% ]
   ┝━ apply_block [ 802µs | 13.59% / 23.06% ]
   │  ┝━ insert_block_header [ 88.3µs | 2.54% ]
   │  ┝━ upsert_accounts [ 5.08µs | 0.15% ]
   │  ┝━ insert_scripts [ 18.3µs | 0.53% ]
   │  ┝━ insert_notes [ 14.6µs | 0.42% ]
   │  ┝━ insert_transactions [ 13.4µs | 0.38% ]
   │  ┝━ insert_nullifiers_for_block [ 71.5µs | 2.06% ]
   │  ┕━ acquire_write_lock [ 118µs | 3.40% ]
   ┝━ acquire_forest_write_lock [ 9.12µs | 0.26% ]
   ┕━ ｉ [info]: apply_block successful | block_commitment:
```

### DIFF
--- a/crates/store/src/state/mod.rs
+++ b/crates/store/src/state/mod.rs
@@ -40,7 +40,7 @@ use miden_protocol::note::{NoteDetails, NoteId, NoteScript, Nullifier};
 use miden_protocol::transaction::{OutputNote, PartialBlockchain};
 use miden_protocol::utils::Serializable;
 use tokio::sync::{Mutex, RwLock, oneshot};
-use tracing::{Instrument, info, instrument};
+use tracing::{Instrument, info, info_span, instrument};
 
 use crate::accounts::{AccountTreeWithHistory, HistoricalError};
 use crate::blocks::BlockStore;
@@ -277,7 +277,10 @@ impl State {
         );
 
         // Wait for the message from the DB update task, that we ready to commit the DB transaction
-        acquired_allowed.await.map_err(ApplyBlockError::ClosedChannel)?;
+        acquired_allowed
+            .instrument(info_span!(target: COMPONENT, "await_db_readiness"))
+            .await
+            .map_err(ApplyBlockError::ClosedChannel)?;
 
         // Awaiting the block saving task to complete without errors
         block_save_task.await??;
@@ -287,7 +290,11 @@ impl State {
             // We need to hold the write lock here to prevent inconsistency between the in-memory
             // state and the DB state. Thus, we need to wait for the DB update task to complete
             // successfully.
-            let mut inner = self.inner.write().await;
+            let mut inner = self
+                .inner
+                .write()
+                .instrument(info_span!(target: COMPONENT, "acquire_inner_write_lock"))
+                .await;
 
             // We need to check that neither the nullifier tree nor the account tree have changed
             // while we were waiting for the DB preparation task to complete. If either of them
@@ -329,7 +336,12 @@ impl State {
         .in_current_span()
         .await?;
 
-        self.forest.write().await.apply_block_updates(block_num, account_deltas)?;
+        let mut forest = self
+            .forest
+            .write()
+            .instrument(info_span!(target: COMPONENT, "acquire_forest_write_lock"))
+            .await;
+        forest.apply_block_updates(block_num, account_deltas)?;
 
         info!(%block_commitment, block_num = block_num.as_u32(), COMPONENT, "apply_block successful");
 


crates/store/src/state/mod.rs
```diff
@@ -471,13 +336,188 @@ impl State {
...
+        let header = block.header();
+
+        let tx_commitment = block.body().transactions().commitment();
+        if header.tx_commitment() != tx_commitment {
+            return Err(InvalidBlockError::InvalidBlockTxCommitment {
+                expected: tx_commitment,
+                actual: header.tx_commitment(),
+            }
+            .into());
+        }
```
bobbinth, line 364: Not for this PR, but would be great to use `ProvenBlock::validate()` here. Or maybe, we should make sub-validators of `ProvenBlock::validate()` public and invoke them separately.

crates/store/src/state/mod.rs
```diff
@@ -471,13 +336,188 @@ impl State {
...
+        if block_num != expected_block_num {
+            return Err(InvalidBlockError::NewBlockInvalidBlockNum {
+                expected: expected_block_num,
+                submitted: block_num,
+            }
+            .into());
+        }
+        if header.prev_block_commitment() != prev_block.commitment() {
+            return Err(InvalidBlockError::NewBlockInvalidPrevCommitment.into());
+        }
```
bobbinth, line 385: Also not for this PR, but it would be great if this was behind something like `BlockHeader::is_parent_of()` or `BlockHeader::is_child_of()`.

crates/store/src/state/mod.rs
```diff
@@ -471,13 +336,188 @@ impl State {
...
+            .collect();
+        if !duplicate_nullifiers.is_empty() {
+            return Err(InvalidBlockError::DuplicatedNullifiers(duplicate_nullifiers).into());
+        }
+
+        // new_block.chain_root must be equal to the chain MMR root prior to the update
+        let peaks = inner.blockchain.peaks();
+        if peaks.hash_peaks() != header.chain_commitment() {
+            return Err(InvalidBlockError::NewBlockInvalidChainCommitment.into());
+        }
```
bobbinth, line 417: Technically, this should be a part of `validate_block_header()`.

crates/store/src/state/mod.rs
```diff
@@ -471,13 +336,188 @@ impl State {
...
+    fn build_note_records(
+        block: &ProvenBlock,
+    ) -> Result<Vec<(NoteRecord, Option<Nullifier>)>, ApplyBlockError> {
+        let header = block.header();
+        let block_num = header.block_num();
+
+        let note_tree = block.body().compute_block_note_tree();
+        if note_tree.root() != header.note_root() {
+            return Err(InvalidBlockError::NewBlockInvalidNoteRoot.into());
+        }
```
bobbinth, line 484: Technically, this part should be in `validate_block_header()`. We'll need to change the signatures somewhat - e.g., `validate_block_header()` would need to return the note tree, and then `build_note_records()` would need to take it as a parameter. Let's do this in a follow-up PR though.

crates/store/src/state/mod.rs
```diff
@@ -471,13 +336,188 @@ impl State {
...
+                let (details, nullifier) = match note {
+                    OutputNote::Full(note) => {
+                        (Some(NoteDetails::from(note)), Some(note.nullifier()))
+                    },
+                    OutputNote::Header(_) => (None, None),
+                    note @ OutputNote::Partial(_) => {
+                        return Err(InvalidBlockError::InvalidOutputNoteType(Box::new(
+                            note.clone(),
+                        )));
+                    },
```
bobbinth, line 499: This will be simpler in v0.14+ because we'd only have two variants `Public` and `Private`.

crates/store/src/state/mod.rs
```diff
@@ -268,130 +242,14 @@ impl State {
...
-                    OutputNote::Header(_) => (None, None),
-                    note @ OutputNote::Partial(_) => {
-                        return Err(InvalidBlockError::InvalidOutputNoteType(Box::new(
-                            note.clone(),
-                        )));
-                    },
-                };
-
-                let inclusion_path = note_tree.open(note_index);
+        ) = self.compute_tree_mutations(&block).await?;
```
bobbinth, line 250: Maybe not from this PR, but I'm curious if this actually works as desired. Specifically:

We start the `block_save_task` above to run in parallel with `compute_tree_mutations()` but since we await on `compute_tree_mutations()`, would this work? Or should this also be a task and then we'd await on both of these tasks somehow?

PR #1876, opened 2026-04-01
fix: filter unauthenticated notes created in the batch in `SelectedBatch`

TomasArrachea:
Closes https://github.com/0xMiden/node/issues/1875

Running the integration-tests of the client against a node with this fix runs correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.14.1 (2025-04-02)
+
+- Fixed batch building issue with unauthenticated notes consumed in the same batch as they were created ([#1875](https://github.com/0xMiden/node/issues/1875)).
+
 ## v0.14.0 (2025-04-01)
 
 ### Enhancements


igamigo (approved):
LGTM, but would be nice if anyone else could take a look.

crates/block-producer/src/mempool/graph/batch.rs
```diff
@@ -29,7 +29,14 @@ impl GraphNode for SelectedBatch {
     }
 
     fn unauthenticated_notes(&self) -> Box<dyn Iterator<Item = Word> + '_> {
-        Box::new(self.transactions().iter().flat_map(|tx| tx.unauthenticated_note_commitments()))
+        // Filter notes that are produced within this batch.
+        let output_notes: HashSet<Word> = self.output_notes().collect();
```
bobbinth, line 33: This works, but I wonder how expensive it would be to do an allocation every time. Let's create an issue to investigate this in the future.

bobbinth (approved):

Also, I'm not 100% sure about all the implications - so, would be good for @Mirko-von-Leipzig to take a look at this as well. In the interest of time, I'll merge this now, and then, if we discover better ways of addressing this, we can make another PR.

crates/block-producer/src/mempool/graph/batch.rs
TomasArrachea: Created https://github.com/0xMiden/node/issues/1878

PR #1881, opened 2026-04-03
feat: include block header in `SyncChainMmr`

igamigo:
With this PR, `SyncChainMmr` now returns the `block_to` block header in the response, so clients can obtain it alongside the MMR delta without a separate request.

Addresses the smaller point of #1890

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v0.14.2
 
+- Added inclusion proofs to `SyncTransactions` output notes ([#1893](https://github.com/0xMiden/node/pull/1893)).
 - Added `block_header` field to `SyncChainMmrResponse` so clients can obtain the `block_to` block header without a separate request ([#1881](https://github.com/0xMiden/node/pull/1881)).
 
 ## v0.14.1 (2025-04-02)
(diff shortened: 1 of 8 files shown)


crates/store/src/db/models/queries/notes.rs
```diff
@@ -521,10 +521,15 @@ pub(crate) fn get_note_sync(
 ) -> Result<Option<NoteSyncUpdate>, NoteSyncError> {
```
TomasArrachea, line 574: Should we change the return type so it's no longer optional? we never actually return None.

igamigo: Good point, changed

CHANGELOG.md
```diff
@@ -49,6 +53,7 @@
 - Removed gRPC reflection v1-alpha support ([#1795](https://github.com/0xMiden/node/pull/1795)).
 - [BREAKING] Rust requirement bumped from `v1.91` to `v1.93` ([#1803](https://github.com/0xMiden/node/pull/1803)).
 - [BREAKING] Updated `SyncNotes` endpoint to returned multiple note updates (([#1843](https://github.com/0xMiden/node/pull/1843))).
+- `SyncNotes` now always includes the range-end block header and MMR proof when the scan completes, even if the block has no matching notes ([#1881](https://github.com/0xMiden/node/pull/1881)).
```
TomasArrachea: Wouldn't this also fall under v0.14.2?

igamigo: Yeah, I moved it to `0.14.2` and forgot to remove this line.

igamigo:
Updated the PR with the alternative approach

PR #1901, opened 2026-04-08
feat: Refactor apply_block() locks

sergerad:
### Problem

On `next`, the store's `apply_block` uses a two-phase commit with `oneshot` channels to synchronize DB and in-memory updates. It acquires a `Mutex` to prevent concurrent writes, then coordinates between a DB update task and the main apply path using channels (`acquired_allowed`, `inform_acquire_done`). During the commit phase, it acquires `RwLock::write()` on `inner` (which holds the nullifier tree, account tree, and blockchain) and a separate `RwLock::write()` on `forest`, blocking all readers for the duration of the DB commit and in-memory mutations.

This means:
- **Write throughput is limited by read latency**: the writer must wait for all in-flight reads to release their `RwLock` read guards before it can acquire the write lock. On Testnet we have observed `store.Rpc/GetAccount` having P99 ~4.4s over 5.6M calls in a 3 day period. This can cause the write path to slow down significantly.
- **Read latency spikes during writes**: readers block on the write lock while the DB transaction commits and in-memory state is updated.
- **Complex coordination logic**: the two-phase channel handshake (`acquired_allowed` / `inform_acquire_done`) adds error-prone complexity and extra `ConcurrentWrite` error paths for root staleness checks.

### Solution

Replace the `RwLock`/`Mutex`/channel coordination with a single `ArcSwap<InMemoryState>` and a dedicated writer task serialized by an `mpsc` channel.

**Readers** call `snapshot()` which returns `Arc<InMemoryState>` via `ArcSwap::load_full()` — a wait-free atomic refcount increment with no data cloning and no locks. The returned `Arc` is a frozen view: readers are completely unaffected if the writer publishes new state while they hold it.

**The writer** is a dedicated `tokio::spawn`ed task that owns the writable trees directly as local variables — no `Arc`, no locks, no interior mutability. It receives blocks via an `mpsc` channel (one at a time), validates against the current snapshot, commits to DB with **no locks held**, applies mutations to its owned trees, then builds a new `InMemoryState` with snapshot-backed read-only copies of the trees and atomically publishes via `ArcSwap::store()`.

Both the nullifier tree and account tree are RocksDB-backed. The writer holds writable `LargeSmt<RocksDbStorage>` instances. After applying mutations (which write to RocksDB), the writer creates `LargeSmt<RocksDbSnapshotStorage>` instances for the new `InMemoryState` — these use native RocksDB snapshots for point-in-time consistent reads with zero data copying.

### Changes

- **`InMemoryState`**: New struct holding `block_num`, nullifier tree, account tree, blockchain MMR, and forest. Lives behind `ArcSwap` on `State`.
- **`Scoped<T>`**: New wrapper pairing query results with the `BlockNumber` of the snapshot that produced them, so callers always know which block the data corresponds to. Applied to `get_transaction_inputs`, `sync_nullifiers`, `sync_notes`, `sync_account_vault`, `sync_account_storage_maps`, `sync_transactions`.
- **Writer task**: Replaced the two-phase commit (`apply_block.rs`) with a dedicated `mpsc`-based writer loop (`writer.rs`). Owns writable trees as local variables — no locks held during DB commit.
- **`RocksDbSnapshotStorage`**: New read-only, `Clone`able storage backend using native RocksDB snapshots. Used for the in-memory snapshot trees so readers don't block the writer.
- **Trait bound relaxation**: `SmtStorage` split into `SmtStorageReader` (read) and `SmtStorage` (read+write). Read-only snapshot trees only need `SmtStorageReader`. `AccountTreeWithHistory` and loader functions relaxed accordingly.
- **`Clone` derives** added to `AccountTreeWithHistory`, `AccountStateForest`, and upstream `LargeSmt`/`LargeSmtForest` (via patches) to support snapshot construction.
- **Removed**: `RwLock<InnerState>`, `RwLock<AccountStateForest>`, `Mutex<()>` write lock, `apply_block.rs`, all `.await` on snapshot calls.

### Further changes required

We will want to do the following either in this PR or followups:
- Replace `LargeSmtForest<ForestInMemoryBackend>` with a RocksDb backend.
- Dynamic in-memory cache size for `LargeSmt` so that we can reduce memory footprint of `NullifierTree`, `AccountTree`, and `AccountStateForest` in `InMemoryState`.
- Observability as to number of separate `Arc<InMemoryState>` instances at any one time.

### Performance characteristics

| | `next` | This PR |
|---|---|---|
| Reader cost | `RwLock` read guard (may block on writer) | Atomic refcount bump (wait-free) |
| Writer blocked by readers | Yes (waits for read guards to drain) | No |
| Locks held during DB commit | `RwLock` write lock on `inner` | None |
| Write coordination | Two-phase channel handshake + root staleness check | Single writer task via `mpsc` |
| Deep clone per block | None | Once (writer clones `InMemoryState` — trees use RocksDB snapshots, no tree data copied) |
| Unsafe code | None | None |


crates/store/src/state/writer.rs
```diff
@@ -0,0 +1,343 @@
...
+    pub signed_block: SignedBlock,
+    pub proving_inputs: Option<BlockProofRequest>,
+    pub result_tx: oneshot::Sender<Result<(), ApplyBlockError>>,
+}
+
+/// Runs the single writer loop. Receives blocks through the channel and applies them
+/// sequentially. The writer owns the writable trees — no locks or interior mutability needed.
+pub(crate) async fn writer_loop(
+    mut rx: mpsc::Receiver<WriteRequest>,
+    state: Arc<State>,
```
SantiagoPittella: I think that we have a circular reference here since `State` contains `writer_tx: mpsc::Sender`.  This task keeps the `Sender` alive causing this to never terminate. A solution could be to move out the write handle into its own self contained struct:
```
struct WriteHandle {
  tx: mpsc::Sender
}
```

sergerad: Thanks, good call. Added a few more abstractions and pulled things out of State.

crates/store/src/state/writer.rs
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    /// Builds a snapshot-backed nullifier tree for the new in-memory state.
+    fn build_snapshot_nullifier_tree(&self) -> NullifierTree<LargeSmt<SnapshotTreeStorage>> {
+        #[cfg(feature = "rocksdb")]
+        {
+            let snapshot_storage = miden_large_smt_backend_rocksdb::RocksDbSnapshotStorage::new(
+                std::sync::Arc::clone(&self.nullifier_db),
+            );
+            let snapshot_smt = crate::state::loader::load_smt(snapshot_storage)
```
kkovaacs: As far as I understand this is a fairly costly operation, re-building the in-memory cache of the top layers of the tree.

Can't we re-use the in-memory state of the writable tree?

crates/store/src/state/writer.rs
```diff
@@ -0,0 +1,392 @@
...
+        new_forest.apply_block_updates(block_num, account_deltas)?;
+
+        let new_state = InMemoryState {
+            block_num,
+            nullifier_tree: snapshot_nullifier_tree,
+            account_tree: snapshot_account_tree,
+            blockchain: new_blockchain,
+            forest: new_forest,
+        };
+
```
kkovaacs: This is a large chunk of non-trivial, CPU / I/O intensive code that we're running on an async task. I think we should be more careful about not blocking on an async task for a long period, so this probably should be moved to a blocking task instead.

kkovaacs (commented):
I've added some comments, but I think the overall concept is solid.

crates/store/src/state/writer.rs
```diff
@@ -0,0 +1,372 @@
...
+        self.validate_block_header(header, body).await?;
+
+        // Load the current in-memory state snapshot for validation (wait-free).
+        let snapshot = self.in_memory.load_full();
+
+        // Tree mutation, RocksDB writes, and snapshot construction are all CPU- and I/O-bound
+        // synchronous workloads. Run them inside block_in_place so tokio can evacuate other tasks
+        // from this thread for the duration.
+        let (snapshot_nullifier_tree, snapshot_account_tree, notes, account_deltas) =
```
SantiagoPittella: Does the order of this sequence matter here? The tree mutations could be applied here and then the Sqlite tx may fail later

sergerad: Yea good point. I have refactored `write_block()` to have two fns - one `&self` and one `&mut self`. The latter is responsible for mutations and any error coming out of that is treated as fatal and causes the store to shut down.

Mirko-von-Leipzig:
@sergerad what's the status of this? Are we still waiting on other PRs to land elsewhere or can we review and merge this?

(you may have mentioned this elsewhere already; if so, apologies)

sergerad:
> @sergerad what's the status of this? Are we still waiting on other PRs to land elsewhere or can we review and merge this?
> 
> (you may have mentioned this elsewhere already; if so, apologies)

We can review this as is. But waiting on PRs
- https://github.com/0xMiden/crypto/pull/967
  - Has an approval but need to point it to next / v0.25 instead of main because introduced a breaking change in later commit
- https://github.com/0xMiden/protocol/pull/2755
  - Blocked on the crypto PR above
  
Also need protocol and node to then use crypto v0.25 which might take some time.

Mirko-von-Leipzig:
> > @sergerad what's the status of this? Are we still waiting on other PRs to land elsewhere or can we review and merge this?
> > (you may have mentioned this elsewhere already; if so, apologies)
> 
> We can review this as is. But waiting on PRs
> 
>     * [feat: SmtStorageReader and SparseMerkleTreeReader traits crypto#967](https://github.com/0xMiden/crypto/pull/967)
>       
>       * Has an approval but need to point it to next / v0.25 instead of main because introduced a breaking change in later commit
> 
>     * [feat: AccountTreeBackendReader and NullifierTreeBackendReader traits protocol#2755](https://github.com/0xMiden/protocol/pull/2755)
>       
>       * Blocked on the crypto PR above
> 
> 
> Also need protocol and node to then use crypto v0.25 which might take some time.

I suspect that especially the last point would result in this diff no longer being applicable as is. Should we move this to draft until we're ready?

kkovaacs:
FYI: if we implement a persistent account state forest backend using `miden_crypto::merkle::smt::large_forest::backend::persistent::PersistentBackend` I think we'll require the same separation for the large forest backend.

sergerad:
Moved to #2175

PR #1922, opened 2026-04-10
fix: explorer field types

SantiagoPittella:
Looks like the Explorer's GraphQL API changed and now returns some numeric fields as JSON strings instead of numbers:

{"block_number": "30629", "timestamp": "1775830836", "number_of_transactions": "1", ...}

The monitor code used `serde_json::Value::as_u64()` to extract these, which only works for actual JSON numbers. So it returned None for every block, and the health check always failed.

Screenshot 2026-04-10 at 11 35 12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.14.6 (2026-04-10)
+
+- Fixed network monitor explorer health check failing to parse string-encoded numeric fields from the Explorer GraphQL API ([#1922](https://github.com/0xMiden/node/pull/1922)).
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.14.6 (2026-04-10)
```
igamigo, line 3: Probably also good to bump the `Cargo.toml` version in this PR

bobbinth (approved):
Not a very deep review - but looks good! Thank you!

Agree with @igamigo that would be good to bump up the version in `Cargo.toml` as well. Once this is done, I can merge and release the next node version.

PR #1923, opened 2026-04-10
Port crypto RocksDB budgeted deserialization

huitseeker:
This ports the RocksDB backend change from 0xMiden/crypto#846 “Use budgeted deserialization for untrusted bytes.”

The change updates the migrated RocksDB backend to use `read_from_bytes_with_budget` for SMT leaves and depth-24 hashes read from RocksDB. The budget is the stored byte length.

During validation, `make check-features` also found a rocksdb feature-gating gap in store state loading. This PR gates the `RocksDbOptions` import and uses the memory-storage config when the `rocksdb` feature is off.


kkovaacs (approved):
Looks good! Thank you!

Mirko-von-Leipzig (approved):

PR #1926, opened 2026-04-13
fix(store): bump SQLite connection pool size to 128

kkovaacs:
We allow up to 1k concurrent RPC requests, and each request may require a connection to the database. The default SQLite connection pool size is 16, which is too small for our use case and can lead to timeouts when the pool is exhausted. By increasing the pool size to 128, we can better accommodate concurrent requests and reduce the likelihood of timeouts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.13.13 (2026-04-13)
+
+- Increase SQLite connection pool size to 128.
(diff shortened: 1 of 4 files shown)


bobbinth (approved):
Looks good! Thank you!

When we apply this change to v0.14 (`main`) and `next`, would be great to make this an operator-configurable setting so that it could be adjusted at node startup.

PR #1929, opened 2026-04-13
chore: remove InflightNetworkNote

### LINKED ISSUE #1887: Remove `InflightNetworkNote`
The ntx builder has the `InflightNetworkNote` which wraps a network note with some additional builder related metadata. This metadata was useful when we stored everything in-memory e.g. tracked number of failures.

I believe this information is now be tracked inside the database, and there is therefore no reason to include this in a custom struct. We should remove this wrapper and use the targeted note directly.

Note failures can simply be incremented in-place in the database, as can the last attempted block height. (if that isn't the case already).

SantiagoPittella:
closes #1887 

Removes the `InflightNetworkNote` struct in the NTX builder, since we weren't using the metadata stored in the struct.


sergerad (approved):

bobbinth (approved):
Looks good! Thank you!

Issue #1920, opened 2026-04-10
Propagate v0.13 changes to v0.14 and v0.15


bobbinth:
We recently released v0.13.12 which contains changes from https://github.com/0xMiden/node/pull/1916 and https://github.com/0xMiden/node/pull/1905. We should make sure these changes are both in v0.14 (`main`) and in v0.15 (`next`) branches. The changes from #1905 may already be there, but I don't think #1916 has been propagated yet.

PR #1936, opened 2026-04-13
chore: Add spans to get_account stack

sergerad:
Relates to #1920.

### DIFF
--- a/crates/store/src/account_state_forest/mod.rs
+++ b/crates/store/src/account_state_forest/mod.rs
@@ -272,6 +272,7 @@ impl AccountStateForest {
     ///
     /// Returns `None` if no storage root is tracked for this account/slot/block combination.
     /// Returns a `MerkleError` if the forest doesn't contain sufficient data for the proofs.
+    #[instrument(target = COMPONENT, skip_all)]
     pub(crate) fn get_storage_map_details_for_keys(
         &self,
         account_id: AccountId,
--- a/crates/store/src/account_state_forest/mod.rs
+++ b/crates/store/src/account_state_forest/mod.rs
@@ -231,6 +231,7 @@ impl AccountStateForest {
     ///
     /// Note that the `raw_key` is the raw, user-provided key that needs to be hashed in order to
     /// get the actual key into the storage map.
+    #[instrument(target = COMPONENT, skip_all)]
     pub(crate) fn get_storage_map_witness(
         &self,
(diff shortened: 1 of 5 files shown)


bobbinth:
I think v0.14 "master" is currently in `main` - or are we keeping these two branches in sync?

sergerad:
> I think v0.14 "master" is currently in `main` - or are we keeping these two branches in sync?

Pointing at main now. I will delete the release/v0.14 branch I made.

crates/store/src/state/mod.rs
```diff
@@ -771,7 +774,7 @@ impl State {
         let mut storage_map_details_by_index = vec![None; storage_request_slots.len()];
 
         if !map_keys_requests.is_empty() {
-            let forest_guard = self.forest.read().await;
+            let forest_guard = self.forest.read().instrument(tracing::Span::current()).await;
```
SantiagoPittella: Why was this needed?

SantiagoPittella (approved):

crates/store/src/state/mod.rs
sergerad: To have a dedicated span and duration for reading this lock.

crates/store/src/account_state_forest/mod.rs
bobbinth, line 235: Should we also instrument this method?

bobbinth, line 253: Should we also instrument this method?

bobbinth (approved):

Mirko-von-Leipzig (approved):
Thanks @sergerad. Once we've figured out wtf is going on, we should drop these down to `debug` again. But we can has this sort of thing out in the #1949

crates/store/src/state/mod.rs
Mirko-von-Leipzig: I think @SantiagoPittella means that this _should_ already be auto-instrumented by the `#[instrument]` on the function. If it wasn't then this is.. unexpected.

If you want a dedicated span then we should give this a new span, not re-use the current one.

sergerad: Sorry had meant to `.instrument(tracing::info_span!("something"))`.

I think/hope we don't need `instrument(tracing::Span::current())` on such futures to ensure they fall under the relevant span of the stack they are called from.

PR #1939, opened 2026-04-13
chore: improve error diagnostics

### LINKED ISSUE #1931: Monitor: Improve error diagnostics when external service responses change
## Problem

When external services (Explorer and Faucet basically) change their response format, the monitor produces errors that don't indicate which field actually changed or what was received vs expected. This has happened twice recently:

- The Explorer changed a field type and the error was just `"missing field: block_number"`, which was misleading because the field was present but with a different type.
- The Faucet's `/get_metadata` response changed a field, and the deserialization error didn't clearly indicate which field caused the failure.

### Explorer

Replace the manual `node.get("field").and_then(value_as_u64).ok_or_else(...)` pattern with an approach that includes the actual received value in the error message. For example:
```
  "field 'block_number': expected u64, got string '12345'"
  "field 'block_commitment': expected string, got null"
```

Extend `ExplorerStatusError` with a `TypeMismatch` variant that captures the field name, expected type, and actual JSON value snippet.

### Fauce

The `GetMetadataResponse`, `PowChallengeResponse`, and `GetTokensResponse` structs use `serde_json::from_str()` which already produces decent errors, but they're wrapped in a generic `"Failed to parse metadata response: {text}"` context. Improve by:
- Logging the raw response body at debug level
- Including the specific serde error (which names the field) prominently in the status error

SantiagoPittella:
closes https://github.com/0xMiden/node/issues/1931

When the Explorer or Faucet change their response format, the monitor now produce an error messages with the actual field that changed. The Explorer previously reported "missing field: block_number" when the field was present but had a different type. It now distinguishes between missing fields ("field 'block_number': not present in response") and type mismatches ("field 'block_number': expected u64-compatible value, got \"abc\""). Faucet deserialization errors now identify which endpoint failed.


bin/network-monitor/src/faucet.rs
```diff
@@ -241,8 +242,8 @@ async fn perform_faucet_test(
 
     let response_text = response.text().await?;
 
-    let metadata: GetMetadataResponse = serde_json::from_str(&response_text)
-        .with_context(|| format!("Failed to parse metadata response: {response_text}"))?;
+    let metadata: GetMetadataResponse =
```
JereSalo, line 237: For `get_metadata` we didn't add the `debug!` line printing `response_text`. Should we add it?

bin/network-monitor/src/explorer.rs
```diff
@@ -174,83 +174,84 @@ fn unhealthy(name: &str, current_time: u64, err: &impl ToString) -> ServiceStatu
...
+            field: field.into(),
+            expected: "string",
+            got: truncate_json(value),
+        })
+}
+
+/// Returns a short string representation of a JSON value for error messages.
+fn truncate_json(value: &serde_json::Value) -> String {
+    let s = value.to_string();
+    if s.len() > 60 { format!("{}...", &s[..60]) } else { s }
```
JereSalo: There's an edge case where the slicing lands in the middle of a multi-byte character and would panic. We could prevent this by truncating at char boundary, like this:

```suggestion
    match s.char_indices().nth(60) {
        Some((idx, _)) => format!("{}...", &s[..idx]),
        None => s,
    }
```

JereSalo (approved):
Left some tiny suggestions, the rest is perfect though.

bin/network-monitor/src/explorer.rs
```diff
@@ -174,83 +174,84 @@ fn unhealthy(name: &str, current_time: u64, err: &impl ToString) -> ServiceStatu
...
+        .map(String::from)
+        .ok_or_else(|| ExplorerStatusError::TypeMismatch {
+            field: field.into(),
+            expected: "string",
+            got: truncate_json(value),
+        })
+}
+
+/// Returns a short string representation of a JSON value for error messages.
+fn truncate_json(value: &serde_json::Value) -> String {
```
sergerad, line 217: This should probably have a unit test. Probably one above too

Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
Makes me wonder if we shouldn't switch to `json` for these..

bin/network-monitor/src/explorer.rs
```diff
@@ -150,83 +150,90 @@ pub(crate) async fn check_explorer_status(
 
 #[derive(Debug)]
 pub enum ExplorerStatusError {
-    MissingField(String),
+    /// A required field was not present in the response.
+    NotPresent(String),
```
igamigo: nit: I think I'd still add the received response here, it makes debugging a bit more straightforward (easier to tell whether the response was valid or something else)

igamigo (approved):
LGTM!

SantiagoPittella:
> Makes me wonder if we shouldn't switch to json for these..

I think that we could benefit, though not that much probably. I will add this to the original issue to do some further investigation

PR #1940, opened 2026-04-13
feat: return erased notes ids on SyncTransactions

TomasArrachea:


TomasArrachea:
Solved in https://github.com/0xMiden/node/pull/1941

PR #1941, opened 2026-04-14
refactor: align `TransactionHeader` and add inclusion proofs

### LINKED ISSUE #1927: Erased notes and their nullifiers are not stored in the node DB
### Problem

When a network note is created and consumed within the same batch/block, the protocol performs "note erasure" - removing both the note from `BlockBody::output_notes()` and its nullifier from `BlockBody::created_nullifiers()`. The node store only inserts notes from `output_notes()` and nullifiers from `created_nullifiers()`, so erased notes leave no trace in the DB.

This means a client tracking such a note (e.g. a note reader monitoring an account) can never discover it was consumed - `sync_nullifiers` won't return the nullifier, and `sync_notes` won't return the note.

Discovered by @TomasArrachea in [0xMiden/miden-client#1971](https://github.com/0xMiden/miden-client/issues/1971#issuecomment-4227069351).

igamigo:
See motivation [here](https://github.com/0xMiden/node/issues/1927#issuecomment-4246027607).

Closes #1927
Closes #1898

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.15.0 (TBD)
+
+- [BREAKING] Moved enriched output note data (inclusion proofs, erased note IDs) from `TransactionHeader` to `TransactionRecord` in `SyncTransactions` response. `TransactionHeader.output_notes` now uses `NoteHeader` (aligned with domain type) instead of `NoteSyncRecord` ([#1927](https://github.com/0xMiden/node/issues/1927), [#1898](https://github.com/0xMiden/node/issues/1898)).
+
 ## v0.14.6 (2026-04-10)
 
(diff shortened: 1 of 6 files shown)


bobbinth (approved):
Looks good! Thank you!

A bit unfortunate that this is a breaking change - but I think this works. Once merged, we'll need to release a new patch versions and yank the previous ones and do the same with the client crates too.

Mirko-von-Leipzig:
> Looks good! Thank you!
> 
> A bit unfortunate that this is a breaking change - but I think this works. Once merged, we'll need to release a new patch versions and yank the previous ones and do the same with the client crates too.

As a note (post-merge): we should practice non-breaking changes by having something like this be a _new_ gRPC method. On the next breaking release we then delete and rename to get the current state.

PR #1942, opened 2026-04-14
feat: auto-regenerate accounts on persistent failures

### LINKED ISSUE #1930: Monitor: Auto-regenerate accounts after persistent failures
## Problem:

The counter increment task in the network monitor currently re-syncs the wallet account from the RPC after some consecutive failures. However, if the accounts themselves are fundamentally outdated or corrupted (e.g. after a network reset or protocol upgrade), re-syncing doesn't help and the monitor keeps failing indefinitely and reports the NTX services as unhealthy.

This is a common issue in our current environment.

## Proposed Solution

After N consecutive failures where re-sync also fails, the monitor should automatically:

  1. Redeploy the counter account via `deploy_counter_account()`
  2. Recreate and save the wallet and counter account files
  3. Re-initialize the increment task state (`setup_increment_task`)
  4. Reset failure counters and resume normal operation

This should be bounded (e.g. max 1 regeneration attempt per hour) to avoid infinite loops if the network itself is down.

SantiagoPittella:
closes #1930

When the counter increment task fails repeatedly and wallet re-sync from RPC is ineffective (e.g., after a network reset or protocol upgrade), the monitor now automatically creates fresh wallet and counter accounts, deploys them, and re-initializes the increment task. This triggers after 10 consecutive failures and is rate-limited to once per hour to avoid loops when the network itself is down.


Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
In theory an alternative would be to separately monitor the genesis hash (or the latest known local chain tip commitment)? This wouldn't catch a protocol upgrade, but in that situation won't we remain broken even if we regenerate?

SantiagoPittella:
> In theory an alternative would be to separately monitor the genesis hash (or the latest known local chain tip commitment)? This wouldn't catch a protocol upgrade, but in that situation won't we remain broken even if we regenerate?

The thing is that sometimes we restart the service but not the accounts after a protocol upgrade, causing the service to continue with the old accounts.

bin/network-monitor/src/counter.rs
```diff
@@ -444,16 +451,41 @@ pub async fn run_increment_task(
...
+
+                // If re-sync keeps failing, regenerate accounts from scratch (rate-limited).
+                let cooldown_elapsed =
+                    last_regeneration.is_none_or(|t| t.elapsed() >= REGENERATE_COOLDOWN);
+                if consecutive_failures >= REGENERATE_FAILURE_THRESHOLD && cooldown_elapsed {
+                    warn!(
+                        consecutive_failures,
+                        "re-sync ineffective, regenerating accounts from scratch"
+                    );
+                    last_regeneration = Some(Instant::now());
```
JereSalo, line 474: We could move this inside the `Ok` path of `try_regenerate_accounts`, so that if, for some reason, fails with an error we can try again in a short period of time instead of having to wait for an hour.

JereSalo: Perhaps it was intentional though and we still want to wait upon failure, but I'll leave this JIC

JereSalo (approved):
I see that in `run_counter_tracking_task` we [assign the counter_account variable](https://github.com/0xMiden/node/blob/3393a1769f5fe0224e5f3b617d119c65a73a175c/bin/network-monitor/src/counter.rs#L668-L669), I wonder if after re-creating the accounts this keeps pointing to the previous account and we should update it. Perhaps there's something I'm missing though.

Not for this PR but maybe we should consider adding tests for these kinds of behavior (if you find it's worth it and not too complex).

bin/network-monitor/src/counter.rs
SantiagoPittella: The cooldown exists precisely because the failure mode is pretty expensive expensive, and the trigger condition is already rare by design. Also If regeneration fails, the failure is almost certainly correlated with whatever broke increments.

PR #1943, opened 2026-04-15
fix: Various run-node.sh issues

sergerad:
Changes to make `run-node.sh` more robust.

### DIFF
--- a/crates/utils/src/grpc.rs
+++ b/crates/utils/src/grpc.rs
@@ -17,24 +17,6 @@ impl UrlExt for url::Url {
     }
 }
 
-/// Binds a TCP listener to the given address with `SO_REUSEADDR` enabled.
-///
-/// This allows the listener to bind to a port that is in the `TIME_WAIT` state, which is common
-/// when rapidly restarting services.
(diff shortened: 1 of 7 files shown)


crates/utils/src/grpc.rs
```diff
@@ -17,6 +17,24 @@ impl UrlExt for url::Url {
...
+        tokio::net::TcpSocket::new_v6()
+    }
+    .context("Failed to create TCP socket")?;
+
+    socket.set_reuseaddr(true).context("Failed to set SO_REUSEADDR")?;
+    socket.bind(addr).context("Failed to bind socket")?;
+
+    socket.listen(1024).context("Failed to listen on socket")
+}
+
```
kkovaacs: I don't quite understand why we need this. According to the [documentation](https://docs.rs/tokio/latest/tokio/net/struct.TcpListener.html#impl-TcpListener) `TcpListener::bind()` _does_ set the `SO_REUSEADDR` socket option.

sergerad: Thanks yea removed. Things were flakey and this was never the real fix clearly

kkovaacs (approved):
Looks good to me, thanks!

PR #1946, opened 2026-04-15
refactor(proto): Use `codegen` for proto code generation

kkovaacs:
As a preparation for GRPC server standardization (see PR #1742) this change moves our `build.rs` to use the [codegen](https://docs.rs/codegen/latest/codegen/) crate for generating `file_descriptors.rs`.

What we're generating now is really trivial, but the changes outlined in PR #1742 will absolutely need something better than concatenating strings.


proto/build.rs
```diff
@@ -1,87 +1,105 @@
...
+
+    fs::write(out_dir.join("file_descriptors.rs"), code.to_string()).into_diagnostic()?;
 
     Ok(())
 }
+
+/// The list of `*.proto` files in the given directory.
+///
+/// Does _not_ recurse into folders; only top level files are returned.
+fn proto_files_in_directory(directory: &PathBuf) -> Result<Vec<PathBuf>, miette::Error> {
```
sergerad: nit: all &PathBuf args could/should be &Path?

sergerad (approved):

proto/build.rs
kkovaacs: Thanks, fixed in https://github.com/0xMiden/node/pull/1946/commits/af136ddf67ed851bda950a63e8f9d8294351637a.

proto/build.rs
```diff
@@ -1,87 +1,105 @@
...
+        proto_files.push(entry.path());
+    }
+    Ok(proto_files)
+}
+
+/// Creates a function which emits the file descriptor of the given gRPC service file.
+///
+/// The function looks as follows:
+///
+/// ```rust
```
Mirko-von-Leipzig:
```suggestion
/// ```rust, ignore
```

proto/build.rs
```diff
@@ -1,87 +1,105 @@
...
-const VALIDATOR_DESCRIPTOR: &str = "validator_file_descriptor.bin";
-const NTX_BUILDER_DESCRIPTOR: &str = "ntx_builder_file_descriptor.bin";
-
-/// Generates Rust protobuf bindings from .proto files.
+/// Compiles each gRPC service definitions into a
+/// [`FileDescriptorSet`](tonic_prost_build::FileDescriptorSet) and exposes it as a function:
 ///
-/// This is done only if `BUILD_PROTO` environment variable is set to `1` to avoid running the
-/// script on crates.io where repo-level .proto files are not available.
+/// ```rust
```
Mirko-von-Leipzig:
```suggestion
/// ```rust, ignore
```

Mirko-von-Leipzig (commented):
Don't think we currently run doc tests, or even if `build.rs` is considered for them. But otherwise lgtm.

Mirko-von-Leipzig (approved):

PR #1947, opened 2026-04-15
feat(rocksdb): tune LargeSmt profile and config

huitseeker:
The new default uses cached index and filter blocks, 16 KiB block tables, tuned Bloom filters, and bottommost ZSTD compression for SMT data. Metadata stays uncompressed. The RocksDB config also keeps support for tuning overrides, memory-budget controls, and durability mode selection.

The main issue this helps is 0xMiden/node#1776. It should also help the write-heavy performance work in 0xMiden/node#1537. In earlier quick Criterion runs, the tuned block-table profile improved vanilla insertion/100 by 14.13%, with_history insertion/10 by 9.44%, and with_history insertion/500 by 12.62%. Bottommost ZSTD improved vanilla insertion/1 by 12.39%. Read-path changes in those runs were mostly within noise, so this is best treated as a better default write profile.

This PR does not try to fix the RocksDB correctness issues in 0xMiden/crypto#1095, 0xMiden/node#1729, or 0xMiden/miden-vm#3350. It changes defaults and config control, not error handling or storage semantics.

References:
- https://github.com/facebook/rocksdb/wiki/Setup-Options-and-Basic-Tuning

### DIFF
--- a/crates/large-smt-backend-rocksdb/Cargo.toml
+++ b/crates/large-smt-backend-rocksdb/Cargo.toml
@@ -18,7 +18,7 @@ workspace = true
 miden-crypto   = { features = ["concurrent", "std"], workspace = true }
 miden-protocol = { features = ["std"], workspace = true }
 rayon          = { version = "1.10" }
-rocksdb        = { default-features = false, features = ["bindgen-runtime", "lz4"], version = "0.24" }
+rocksdb        = { default-features = false, features = ["bindgen-runtime", "lz4", "zstd"], version = "0.24" }
 
 [build-dependencies]
 miden-node-rocksdb-cxx-linkage-fix = { workspace = true }
--- a/crates/large-smt-backend-rocksdb/src/rocksdb.rs
+++ b/crates/large-smt-backend-rocksdb/src/rocksdb.rs
@@ -1439,11 +1439,11 @@ impl Default for RocksDbBloomFilterBitsPerKey {
         Self {
             leaves: DEFAULT_BLOOM_FILTER_BITS_PER_KEY,
             depth_24: DEFAULT_BLOOM_FILTER_BITS_PER_KEY,
-            subtree_24: 8.0,
+            subtree_24: DEFAULT_BLOOM_FILTER_BITS_PER_KEY,
             subtree_32: DEFAULT_BLOOM_FILTER_BITS_PER_KEY,
             subtree_40: DEFAULT_BLOOM_FILTER_BITS_PER_KEY,
-            subtree_48: 12.0,
-            subtree_56: 12.0,
+            subtree_48: DEFAULT_BLOOM_FILTER_BITS_PER_KEY,
+            subtree_56: DEFAULT_BLOOM_FILTER_BITS_PER_KEY,
         }
     }
 }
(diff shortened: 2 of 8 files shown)


crates/large-smt-backend-rocksdb/Cargo.toml
```diff
@@ -18,7 +18,7 @@ workspace = true
 miden-crypto   = { features = ["concurrent", "std"], workspace = true }
 miden-protocol = { features = ["std"], workspace = true }
 rayon          = { version = "1.10" }
-rocksdb        = { default-features = false, features = ["bindgen-runtime", "lz4"], version = "0.24" }
+rocksdb        = { default-features = false, features = ["bindgen-runtime", "lz4", "zstd"], version = "0.24" }
```
Mirko-von-Leipzig, line 21: Do we still need `lz4` if we're using `zstd`?

Mirko-von-Leipzig (commented):
I'm not entirely sure how to review something like this. Beyond saying that the code still compiles so there's that 😅

crates/large-smt-backend-rocksdb/Cargo.toml
huitseeker: The setting is not “use ZSTD instead of LZ4.” It is “use LZ4 for active SST files, and use ZSTD only for bottommost files.” Those serve different parts of the LSM tree:

-  compression = Lz4 applies to normal/newer levels. This keeps write-path compaction and hot reads cheap while still avoiding uncompressed SST files. In this repo, that is set for the SMT column families, and also directly for the depth-24 CF.

-  bottommost_compression = Zstd is for the coldest, most stable level. That is where stronger compression is most useful, because the data is rewritten less often and the extra CPU cost is easier to justify. 

So the dependency line also needs both features since we keep both algorithms.

kkovaacs (approved):
Looks good to me.

huitseeker:
I've tuned the PR to be a bit more rigorous given low confidence signals in review:
- I have not changed the persistence/durability setting from its default of "relaxed" — Leaving the main PR contribution to make it configurable,
- I have tightened the ZSTD config, it had a footgun of needing an "enable" value to be passed, this is fixed,
- I've removed the most exotic compaction tuning options, with that removal benched at no regression,
- left the tuning of bits per key on the Bloom filters to the recommended default (10).

Overall, this makes the PR more "vanilla" with more defensible 'best practice' defaults — still improving the tuning, but more of "setting a good basis for iteration" than making a strong claim to "optimized settings".

Mirko-von-Leipzig (approved):

crates/large-smt-backend-rocksdb/src/rocksdb.rs
```diff
@@ -20,13 +20,21 @@ use rocksdb::{
...
 use crate::helpers::{insert_into_leaf, map_rocksdb_err, remove_from_leaf};
 use crate::{EMPTY_WORD, Word};
 
 const IN_MEMORY_DEPTH: u8 = 24;
+const DEFAULT_CACHE_SIZE: usize = 1 << 30;
+const DEFAULT_MAX_OPEN_FILES: i32 = 512;
+const DEFAULT_BLOCK_SIZE: usize = 16 << 10;
+const DEFAULT_MAX_TOTAL_WAL_SIZE: u64 = 512 * 1024 * 1024;
+const DEFAULT_BOTTOMMOST_ZSTD_MAX_TRAIN_BYTES: i32 = 1 << 20;
+const DEFAULT_BLOOM_FILTER_BITS_PER_KEY: f64 = 10.0;
```
bobbinth, line 37: It seems like these changes (and probably other changes from this PR) have not been backported to the RocksDB backend in `miden-crypto`. So, now we have some divergence between the implementation here and there.

I think we should make sure that both backends are in-sync. We could also consider dropping one of the backends - though, I'm not entirely sure if we can drop one from `miden-crypto` just yet.

Issue #1886, opened 2026-04-04
Docker compose extensions


Mirko-von-Leipzig:
Add and expose additional components to the node's docker compose (#1786)

- `jaeger` for telemetry support
- block explorer
- note transport layer
- native faucet
- monitor

SantiagoPittella:
We could add `tempo` and `grafana` to the list.

Reposting my message in https://github.com/0xMiden/node/pull/1902 :

> Additionally, and did not commit it, to monitor the actgor's spawning I added a local observability setup using Grafana and Tempo, using the already exported traces from the node.
> 
> This adds a docker-compose.yml that spins up Tempo as the trace backend and Grafana as the dashboard frontend. Tempo receives traces on port 4317 and Grafana is pre-configured with Tempo as a datasource so there is zero manual setup.
> 
> This is useful for debugging NTX actor behavior during development since you can see the full span tree for actor spawns, transaction execution, block commits, and gRPC calls; and also Tempo can derive rate, error rate, and duration metrics from traces, which was really useful to track actors spawning.

kkovaacs:
We could maybe just use `tempo` and `grafana` instead of `jaeger`?

sergerad:
Block explorer might be a bit heavy handed for local deployment? I will just use this https://crates.io/crates/miden-watch.

Anyone have experience with the note transport layer? @SantiagoPittella

Mirko-von-Leipzig:
> Block explorer might be a bit heavy handed for local deployment? 

Its hard to say, depends on how they've set it up. If its a fully decentralised system itself, then yeah we shouldn't. But it could also just a database file with a UI and work acceptably imo. We should check?

It would be nice for users since that's what they're used to.

SantiagoPittella:
>Anyone have experience with the note transport layer? @SantiagoPittella

Just from the user side, and note starting up the service

Mirko-von-Leipzig:
We're only missing the note transport service at this stage. @mmagician is it worth adding this to the local development environment?

This would require it being a nice docker image to simply add in here.

Mirko-von-Leipzig:
Ping @mmagician ^^

PR #1948, opened 2026-04-16
chore: Add network-monitor image and update compose with tempo and grafana

sergerad:
Relates to #1886.


crates/utils/src/logging.rs
```diff
@@ -114,10 +114,19 @@ pub fn setup_tracing(otel: OpenTelemetry) -> anyhow::Result<Option<OtelGuard>> {
...
 fn init_tracer_provider() -> anyhow::Result<SdkTracerProvider> {
-    let exporter = opentelemetry_otlp::SpanExporter::builder()
-        .with_tonic()
-        .with_tls_config(tonic::transport::ClientTlsConfig::new().with_native_roots())
-        .build()?;
+    let mut builder = opentelemetry_otlp::SpanExporter::builder().with_tonic();
+
+    // Only enable TLS if the endpoint uses HTTPS.
+    let needs_tls = std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT")
+        .map(|ep| ep.starts_with("https://"))
```
Mirko-von-Leipzig: Maybe a dumb question, but is this sufficient as a check? Its probably good enough for now.

Or maybe alternative question: what's the downside of always building with TLS config? Is it so we don't require a cert authority inside docker for no reason?

compose/telemetry.yml
```diff
@@ -0,0 +1,49 @@
+services:
+  tempo:
+    image: grafana/tempo:latest
+    volumes:
+      - ./compose/tempo/tempo.yaml:/etc/tempo.yaml
+    command: ["-config.file=/etc/tempo.yaml"]
+    ports:
+      - "3200:3200"
```
Mirko-von-Leipzig, line 8: Should this match the `grafana: 3000` port?

crates/utils/src/logging.rs
kkovaacs: Yeah, I think think should just work without this change? (`ca-certificates` seems to be installed anyway in the Docker image).

compose/telemetry.yml
sergerad: No, I think they're different services:

3200 Tempo race storage/query backend
3000 Grafana UI you open in a browser

docker-compose.yml
```diff
@@ -102,6 +120,47 @@ services:
...
+      - miden-network-monitor
+      - start
+    environment:
+      - MIDEN_MONITOR_RPC_URL=http://rpc:57291
+      - MIDEN_MONITOR_PORT=3001
+      - MIDEN_MONITOR_NETWORK_NAME=Localhost
+      - MIDEN_MONITOR_DISABLE_NTX_SERVICE=true
+      - MIDEN_MONITOR_ENABLE_OTEL=true
+      - OTEL_EXPORTER_OTLP_ENDPOINT=http://tempo:4317
+      - OTEL_SERVICE_NAME=monitor
```
SantiagoPittella, line 160: We probably would want to enable some more checks, like validator and network transactions. This can be done in a follow up PR, if you create an issue feel free to assign it to me.

SantiagoPittella (approved):
Overall looks good and works properly. The default dashboard can be improved but we can tackle that later I guess

kkovaacs (approved):
Works great for me. Thanks!

PR #1953, opened 2026-04-16
fix: chain tip race condition

### LINKED ISSUE #1951: Ntx chain tip race condition
There is a race condition in the network transaction builder's startup.

The ntxb subscribes to mempool events which might be at block `N`. It then fetches the chain tip from the store, lets call it `N'`. `N' != N` if a block was committed _after_ the subscription, but _before_ the store call.

The result is a panic since the next block committed event will conflict with the `chain_tip` from the store.

Some solutions.

Process mempool events _until_ we arrive at `N'`, and only then proceed with proper startup.

Alternatively, wait until the _first_ block committed event and use that to fetch the chain MMR from the store.

SantiagoPittella:
closes #1951

If the event is for a block we already received from the store, we skip it. If it's genuinely new, we  proceed as we used to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.14.8 (TBD)
+
+- Fixed a startup race in the network transaction builder that could panic the chain MMR when a block committed between subscribing to the mempool and fetching the chain tip from the store ([#1953](https://github.com/0xMiden/node/pull/1953)).
+
 ## v0.14.7 (2026-04-15)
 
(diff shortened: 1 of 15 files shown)


sergerad (approved):
Yea I encountered this in https://github.com/0xMiden/node/pull/1901

Mirko-von-Leipzig (approved):
Oh that's much cleaner than my suggestions, thank you!

SantiagoPittella:
@Mirko-von-Leipzig before merging this, should I mark the changelog section of 0.14.8 with today's date?

bobbinth (approved):
Looks good! Thank you!

PR #1954, opened 2026-04-17
chore(docker): ignore rust-toolchain.toml

kkovaacs:
Copying `rust-toolchain.toml` results in `rustup` installing a new Rust toolchain even if we already have one in the base image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -34,3 +34,4 @@ accounts
 genesis.dat
 miden-store.*
 store.*
+/rust-toolchain.toml


Mirko-von-Leipzig (approved):
Is this why caching never worked properly?

kkovaacs:
> Is this why caching never worked properly?

No, this just results in an additional download -- but otherwise should have no effect on caching.

sergerad (approved):

PR #1955, opened 2026-04-17
chore: simplify actor's context

### LINKED ISSUE #1889: Simplify actor context
Some suggested simplifications

```rust
pub struct AccountActor {
    // Delete as per #1888 
    origin: AccountOrigin,
    // Combine into GrpcClients.store
    store: StoreClient,
    // Combine into State
    db: Db,
    mode: ActorMode,
    notify: Arc<Notify>,
    // Delete: same functionality can be achieved by dropping `notify` sender
    cancel_token: CancellationToken,
    // Combine into GrpcClients.block_producer
    block_producer: BlockProducerClient,
    // Combine into GrpcClients.validator
    validator: ValidatorClient,
    // Combine into GrpcClients.prover
    prover: Option<RemoteTransactionProver>,
    // Combine into State
    chain_state: Arc<RwLock<ChainState>>,
    // Combine into State
    script_cache: LruCache<Word, NoteScript>,
    // Combine into Config
    max_notes_per_tx: NonZeroUsize,
    // Combine into Config
    max_note_attempts: usize,
    // Combine into Config
    idle_timeout: Duration,
    /// Channel for sending requests to the coordinator.
    request_tx: mpsc::Sender<ActorRequest>,
}
```

which should give something like

```rust
struct AccountActor {
    clients: GrpcClients,
    config: ActorConfig,
    state: State,
    notify: Arc<Notify>,
    request: mpsc::Sender<ActorRequest>,
}
```

SantiagoPittella:
closes #1889


crates/ntx-builder/src/coordinator.rs
```diff
@@ -24,16 +23,35 @@ pub struct WriteEventResult {
...
+    fn notify(&self) -> Result<(), mpsc::error::TrySendError<()>> {
+        self.notify.try_send(())
+    }
+
+    /// Returns `true` if a notification is queued but not yet consumed by the actor.
+    ///
+    /// Used after an actor has shut down to detect the race where a notification arrived just as
+    /// the actor timed out; if so, the coordinator should respawn the actor.
+    fn has_pending_notification(&self) -> bool {
+        self.notify.max_capacity() > self.notify.capacity()
```
Mirko-von-Leipzig: I assume you changed from `Notify` because you couldn't determine if it had a pending notification?

I _think_ you should be able to something like the following

```rust
/// Returns `true` if there is an unread notification by the actor.
fn has_pending_notification(&self) -> bool {
    use futures::FutureExt;

    let mut fut = self.notify.notified();
    if fut.now_or_never().is_ready() {
        // Replace the notification to keep the state the same as before.
        // Only required because fn signature doesn't consume.
        self.notified.notify_one();
        true
    } else {
        false
    }
}
```

Mirko-von-Leipzig (commented):
Looks good, barring the `notify` change

Mirko-von-Leipzig (changes requested):

crates/ntx-builder/src/coordinator.rs
SantiagoPittella: Yeah, exactly. The `now_or_never` trick didn't occur to me, so I went with mpsc::channel(1).
With this approach I can go back to `Arc<Notify>` with a little twist.

Mirko-von-Leipzig (approved):
Thank you!

crates/ntx-builder/src/coordinator.rs
```diff
@@ -206,12 +227,10 @@ impl Coordinator {
                 // Actor shut down intentionally (idle timeout, cancelled, account removed).
```
igamigo: nit: Looks like it cannot be cancelled anymore


```suggestion
                // Actor shut down intentionally (idle timeout or account removed).
```

crates/ntx-builder/src/actor/mod.rs
```diff
@@ -391,30 +388,37 @@ impl AccountActor {
                         return Ok(true);
                     }
                 }
+                _ = tokio::time::sleep(self.config.idle_timeout) => {
+                    tracing::info!(
+                        %account_id,
+                        "Account actor deactivated while waiting for account commit",
+                    );
+                    return Ok(false);
+                }
```
igamigo, line 397: Mostly asking to fully understand this change: Previously, the cancel token flow would wait indefinitely, right? If so, what is the motivation for it? I'm wondering if there is any new scenario where an actor could exit and later misses notifications

crates/ntx-builder/src/coordinator.rs
```diff
@@ -151,20 +173,19 @@ impl Coordinator {
...
+        // If an actor already exists for this account ID, something has gone wrong. Reject the
+        // spawn rather than replacing.
+        if self.actor_registry.contains_key(&account_id) {
             tracing::error!(
                 account_id = %account_id,
                 "Account actor already exists"
             );
-            handle.cancel_token.cancel();
+            return;
         }
```
igamigo, line 184: Should the existing actor in the registry be notified here?

igamigo (approved):
LGTM, left some minor comments/questions

crates/ntx-builder/src/actor/mod.rs
SantiagoPittella: > Previously, the cancel token flow would wait indefinitely, right?

Yes

> what is the motivation for it? I'm wondering if there is any new scenario where an actor could exit and later misses notifications

Notifications can't be missed, the `has_pending_notification` handshake in the coordinator catches this

crates/ntx-builder/src/coordinator.rs
SantiagoPittella: Hmm I'm not sure about this. why?

PR #1957, opened 2026-04-17
chore: simplify network monitor counter script

SantiagoPittella:
Due to upstream changes, and considering that we are about to do a new release, this PR adds a simplification to the counter creation and fixes the source manager handling following the changes from https://github.com/0xMiden/miden-client/pull/2006

### DIFF
--- a/bin/network-monitor/src/counter.rs
+++ b/bin/network-monitor/src/counter.rs
@@ -14,7 +14,6 @@ use miden_node_proto::generated::rpc::BlockHeaderByNumberRequest;
 use miden_node_proto::generated::transaction::ProvenTransaction;
 use miden_protocol::account::auth::AuthSecretKey;
 use miden_protocol::account::{Account, AccountFile, AccountHeader, AccountId};
-use miden_protocol::assembly::Library;
 use miden_protocol::block::{BlockHeader, BlockNumber};
 use miden_protocol::crypto::dsa::falcon512_poseidon2::SecretKey;
 use miden_protocol::note::{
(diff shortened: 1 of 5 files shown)


bin/network-monitor/src/counter.rs
```diff
@@ -880,11 +878,12 @@ async fn create_and_submit_network_note(
...
 
 /// Create the increment procedure script.
-fn create_increment_script() -> Result<(NoteScript, Library)> {
-    let library = get_counter_library()?;
+fn create_increment_script() -> Result<NoteScript> {
+    let script =
+        include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/src/assets/counter_program.masm"));
 
     let script_builder = CodeBuilder::new()
-        .with_dynamically_linked_library(&library)
```
igamigo, line 887: To give some more context (correct me if I'm wrong @SantiagoPittella), this was failing due to https://github.com/0xMiden/miden-vm/issues/2986, as this `CodeBuilder` was not sharing the same source manager as the library. On Linux for some reason this is not failing (which is why [the monitor](https://status.devnet.miden.io/) did not show errors AFAIK) but when testing this locally on MacOS it would panic cc @bitwalker

igamigo (approved):
LGTM! Quick question, should we change the criteria for marking the card as unhealthy? The numbers are signaling that something is off:

Image

bin/network-monitor/src/counter.rs
SantiagoPittella: Yes, this is correct.

SantiagoPittella: (and regarding the review comment above this, I will add it to the existing issue of improvements)

SantiagoPittella:
@bobbinth this prevents us of having the issue that Ignacio described in his comment; it would be great to get that merge and released.

bobbinth (approved):
Looks good! Thank you!

PR #1950, opened 2026-04-16
feat(proto): generate generic GRPC API implementations

kkovaacs:
This PR adds the build-time generation of generic GRPC API implementations for _all_ services discussed in PR #1742.

For each method we're generating a trait like this:
```rust
#[tonic::async_trait]
pub trait Status {
    type Input;
    type Output;

    fn decode(request: ()) -> tonic::Result<Self::Input>;

    fn encode(output: Self::Output) -> tonic::Result<crate::generated::validator::ValidatorStatus>;

    async fn handle(&self, input: Self::Input) -> tonic::Result<Self::Output>;

    async fn full(
        &self,
        request: (),
    ) -> tonic::Result<crate::generated::validator::ValidatorStatus> {
        let input = Self::decode(request)?;
        let output = self.handle(input).await?;
        Self::encode(output)
    }
}
```

The idea is that the _actual_ implementation should consist of implementing all per-method traits for the API server (most of the time just `decode` / `encode` / `handle`), and then the provided generic implementation for the API can be used:

```rust
pub trait ApiService: Status + SubmitProvenTransaction + SignBlock {}

impl<T> ApiService for T
where
    T: Status,
    T: SubmitProvenTransaction,
    T: SignBlock,
{
}

#[tonic::async_trait]
impl<T> crate::generated::validator::api_server::Api for T
where
    T: ApiService,
    T: Send,
    T: Sync,
    T: 'static,
{
    async fn status(
        &self,
        request: tonic::Request<()>,
    ) -> tonic::Result<tonic::Response<crate::generated::validator::ValidatorStatus>> {
        #[allow(clippy::unit_arg)]
        <T as Status>::full(self, request.into_inner()).await.map(tonic::Response::new)
    }

    async fn submit_proven_transaction(
        &self,
        request: tonic::Request<crate::generated::transaction::ProvenTransaction>,
    ) -> tonic::Result<tonic::Response<()>> {
        #[allow(clippy::unit_arg)]
        <T as SubmitProvenTransaction>::full(self, request.into_inner())
            .await
            .map(tonic::Response::new)
    }

    async fn sign_block(
        &self,
        request: tonic::Request<crate::generated::blockchain::ProposedBlock>,
    ) -> tonic::Result<tonic::Response<crate::generated::blockchain::BlockSignature>> {
        #[allow(clippy::unit_arg)]
        <T as SignBlock>::full(self, request.into_inner())
            .await
            .map(tonic::Response::new)
    }
}
```

[...]

Issue #1885, opened 2026-04-04
Refactor gRPC servers


Mirko-von-Leipzig:
#1742 introduces a codegen ontop of the tonic codegen for our gRPC servers.

We should get this across the line, and then being implementing it for each service's methods.

I propose that each method should get its own file within the server's module. This also lets us keep unit tests (if desired) inline/close by.

Once all servers are migrated, we can then make the tonic server traits private, only exposing our server/method traits.

It would also be nice if we could move our server configuration into the proto crate. So maybe something like `ServerXX::external` (and internal), which would create the proper tower service.

---------------------

Something I've also thought of, but not solved concretely, is that most of the RPC components methods do very basic verification before sending it on to the store aka a proxy of sorts. It may make sense to create an abstraction there where you define only the verification, and it automatically implements the method. Might also not be worth it.

Mirko-von-Leipzig:
#1742 was superseded by #1946 and #1950. The left-over work is therefore migrating to the new codegen, and then deprecating the existing tonic servers. I'll add sub-issues to track these.

PR #1959, opened 2026-04-20
refactor(validator): refactor gRPC server implementation

### LINKED ISSUE #1970: Migrate validator gRPC server to new codegen

kkovaacs:
As outlined in #1885, after having the build-time generation of related traits and blanket implementations (PR #1950) we can start moving our GRPC API implementations to the new framework.

This PR does this for the validator GRPC API. As suggested in the issue, each method is moved to its own file.

Closes #1970

### DIFF
--- a/crates/validator/src/server/sign_block.rs
+++ b/crates/validator/src/server/sign_block.rs
@@ -16,8 +16,11 @@ impl grpc::server::validator_api::SignBlock for ValidatorServer {
     type Output = Signature;
 
     fn decode(request: grpc::blockchain::ProposedBlock) -> tonic::Result<Self::Input> {
-        ProposedBlock::read_from_bytes(&request.proposed_block)
-            .map_err(|err| tonic::Status::invalid_argument(err.as_report()))
+        ProposedBlock::read_from_bytes(&request.proposed_block).map_err(|err| {
+            tonic::Status::invalid_argument(
+                err.as_report_context("Failed to deserialize proposed block"),
+            )
+        })
     }
 
     fn encode(output: Self::Output) -> tonic::Result<grpc::blockchain::BlockSignature> {
--- a/crates/validator/src/server/sign_block.rs
+++ b/crates/validator/src/server/sign_block.rs
@@ -1,0 +1,72 @@
+use std::sync::atomic::Ordering;
+
+use miden_node_proto::generated as grpc;
+use miden_node_utils::ErrorReport;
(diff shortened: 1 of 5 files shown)


crates/validator/src/server/sign_block.rs
```diff
@@ -0,0 +1,75 @@
...
+        ProposedBlock::read_from_bytes(&request.proposed_block).map_err(|err| {
+            tonic::Status::invalid_argument(
+                err.as_report_context("Failed to deserialize proposed block"),
+            )
+        })
+    }
+
+    fn encode(output: Self::Output) -> tonic::Result<grpc::blockchain::BlockSignature> {
+        Ok(grpc::blockchain::BlockSignature { signature: output.to_bytes() })
+    }
```
Mirko-von-Leipzig, line 28: I suspect all of our `encode` calls are infallible.. Should we change the API to replace `Result<T>` with `T`? Maybe in the future we may need fallible encoding? Can't imagine why though..

crates/validator/src/server/status.rs
```diff
@@ -0,0 +1,33 @@
...
+    type Output = ();
+
+    async fn full(&self, _request: ()) -> tonic::Result<grpc::validator::ValidatorStatus> {
+        Ok(grpc::validator::ValidatorStatus {
+            version: env!("CARGO_PKG_VERSION").to_string(),
+            status: "OK".to_string(),
+            chain_tip: self.chain_tip.load(Ordering::Relaxed),
+            validated_transactions_count: self.validated_transactions_count.load(Ordering::Relaxed),
+            signed_blocks_count: self.signed_blocks_count.load(Ordering::Relaxed),
+        })
```
Mirko-von-Leipzig, line 19: We just need to be careful when we're adding instrumentation at the top level, that this doesn't override anything important.

Mirko-von-Leipzig (approved):
I didn't inspect the code too much with the assumption that its largely a copy-paste from the original.

sergerad (approved):

crates/validator/src/server/sign_block.rs
kkovaacs: That's a good point. I'd rather leave it as `Result<T>` though, because just putting `Ok()` around the encoded result should be trivial -- on the other hand, serialization frameworks like `serde` are fallible, so should we ever move to that we'd actually need to be able to handle errors in `encode()`.

PR #1980, opened 2026-04-21
chore: merge RemoteProver status and tests in backend

### LINKED ISSUE #1908: Monitor: Server-side prover merge
## Server-side prover merge

Problem: The backend sends prover status and test results as separate `ServiceStatus` entries. The frontend has a complex 60-line `mergeProverStatusAndTests()` to recombine them.

Solution: Merge on the server before serialization.

SantiagoPittella:
closes #1908 

This PR merges the Remote Prover status + test results in a same component. Previously, this were sent individually and the frontend was merging them.

Screenshot 2026-04-21 at 14 39 11


bobbinth:
Not related to this PR, but I thought about it while looking at the screenshot: could we truncate worker IDs in the remote prover card? Right now, these span 3 lines, which makes it a bit difficult to read. We could still show the full ID on hover or something like that.

SantiagoPittella:
> Not related to this PR, but I thought about it while looking at the screenshot: could we truncate worker IDs in the remote prover card? Right now, these span 3 lines, which makes it a bit difficult to read. We could still show the full ID on hover or something like that.

Yeah, we could. We can do something similar to what we do with commitments were we provide the copy button to get the full commitment. I will add it to this PR

SantiagoPittella:
Frontend with the copy button
Screenshot 2026-04-21 at 16 32 57

bobbinth:
Looks great! I'd try to format it so that the status (i.e., "healthy") fits on the same line as the ID.

SantiagoPittella:
Update it to:

image

sergerad (approved):

Mirko-von-Leipzig (approved):

PR #1982, opened 2026-04-22
feat: trace relevant http headers

### LINKED ISSUE #1979: Record `X-Forwarded-For`for telemetry
We currently record a client's IP address as part of our gRPC traces. Unfortunately, for requests forwarded by other servers e.g. load-balancers or proxies, this means the original IP address is obfuscated by the other server's.

We should also record the `X-Forwarded-For` header, which is conventionally how such servers indicate the original IP address.

Mirko-von-Leipzig:
This PR adds headers to our telemetry fields for received gRPC requests.

Notably this includes the actual client IPs from `x-forwaded-for`, `x-real-ip`, and `forwarded` headers, which are set by proxies. Currently we only see the proxy IP in production which is not helpful.

I've also removed a layer which seems to be unused now.

Closes #1979


kkovaacs (approved):
Looks good to me!

Mirko-von-Leipzig:
Looks like that layer did actually do something...

crates/utils/src/grpc/layers.rs
```diff
@@ -61,3 +52,43 @@ pub fn rate_limit_per_ip(
...
     Ok(tower_governor::GovernorLayer::new(config))
 }
+
+/// Wraps [`SmartIpKeyExtractor`] by providing a fallback to the client IP address provided by the
+/// gRPC transport.
+///
+/// [`SmartIpKeyExtractor`]'s own fallback of checking the peer IP directly fails because we are in
+/// a gRPC transport and not the typical `SocketAddr` as it expects.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct GrpcIpExtractor(SmartIpKeyExtractor);
```
Mirko-von-Leipzig, line 62: Okay so I figured out what that extra layer was for, and this new extractor improves on it.

`SmartIpExtractor` falls back to deriving the peer IP address using

```rust
request
    .extensions()
    .get::<SocketAddr>()
```

which fails because we are in a gRPC transport. The previous layer "fixed" this, by injecting the peer IP into the `forwarded` header, which `SmartIpExtractor` then picks up.

This of course only works if the proxy is not using `forwarded` header, since otherwise we're overwriting it. This new extractor is the proper way of doing this.

Mirko-von-Leipzig: By removing the extra layer (without the new extractor), tests were failing because IP extraction failed entirely (since peer IP was not found).

kkovaacs: Ah, makes sense. And yes, this should be handled in the extractor.

kkovaacs (approved):
Looks good to me!

